PR #3964, opened 2023-02-14
SALTO-3252: make change validator message on undeployable values attached to the problematic value

shayc331:
_make change validator message on undeployable values attached to the problematic value_

---
_Release Notes_: 

Core:
* change validator message on undeployable values attached to the undeployable value instead of the entire instance

---
_User Notifications_: 
None


alonstern (approved):
🔥 🔥 
Please just add release notes.
@netama do you want to have a look too before merging?

netama (commented):
LGTM, @adamro might be good if you take a look as well to make sure this doesn't break any assumptions in the SAAS that we're aware of?

PR #3968, opened 2023-02-15
SALTO-3651: fix getFullName of undefined in zendesk one_translation_per_locale validator

edenhassid:
fix getFullName of undefined in zendesk one_translation_per_locale validator

---

---
_Release Notes_: 
zendesk:
* fix issue in the validation of Guide translations that could cause failures in deployments.

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/change_validators/one_translation_per_locale.ts
+++ b/packages/zendesk-adapter/src/change_validators/one_translation_per_locale.ts
@@ -77,54 +77,57 @@ const findDuplicateTranslations = async (
     return []
   }
   const locales = await awu(parent.value.translations)
-    .map(referenceExpression => referenceExpression.value)
+    .map(referenceExpression => (isReferenceExpression(referenceExpression)
+      ? referenceExpression.getResolvedValue(elementSource)
+      : referenceExpression))
     .filter(isTranslation)
     .map(translation => translation.value.locale)
     .map(async locale => (isReferenceExpression(locale) && locale.elemID.idType === 'instance'
-      ? (await elementSource.get(locale.elemID)).value.locale ?? ''
+      ? (await elementSource.get(locale.elemID))?.value.locale ?? ''
       : locale))
     .toArray()
   return findDuplicates(locales)
 }
 
-export const oneTranslationPerLocaleValidator: ChangeValidator = async (changes, elementSource) => {
-  if (elementSource === undefined) {
-    log.error('Failed to run oneTranslationPerLocale because no element source was provided')
-    return []
-  }
-  const relevantInstances = changes
-    .filter(isAdditionOrModificationChange)
-    .map(getChangeData)
-    .filter(isInstanceElement)
-    .filter(instance =>
-      [ARTICLE_TRANSLATION_TYPE_NAME,
-        CATEGORY_TRANSLATION_TYPE_NAME,
-        SECTION_TRANSLATION_TYPE_NAME].includes(instance.elemID.typeName))
+export const oneTranslationPerLocaleValidator: ChangeValidator = async (changes, elementSource) =>
+  log.time(async () => {
+    if (elementSource === undefined) {
+      log.error('Failed to run oneTranslationPerLocale because no element source was provided')
+      return []
+    }
+    const relevantInstances = changes
+      .filter(isAdditionOrModificationChange)
+      .map(getChangeData)
+      .filter(isInstanceElement)
+      .filter(instance =>
+        [ARTICLE_TRANSLATION_TYPE_NAME,
+          CATEGORY_TRANSLATION_TYPE_NAME,
+          SECTION_TRANSLATION_TYPE_NAME].includes(instance.elemID.typeName))
 
-  const parentInstanceIds = _.uniqBy(relevantInstances
-    .filter(instance => isTranslation(instance))
-    .flatMap(getParents)
-    .flatMap(parent => (isReferenceExpression(parent) ? parent.elemID : undefined))
-    .filter(isDefined), id => id.getFullName())
+    const parentInstanceIds = _.uniqBy(relevantInstances
+      .filter(instance => isTranslation(instance))
+      .flatMap(getParents)
+      .flatMap(parent => (isReferenceExpression(parent) ? parent.elemID : undefined))
+      .filter(isDefined), id => id.getFullName())
 
-  const parentInstances = await awu(parentInstanceIds)
-    .map(async id => elementSource.get(id))
-    .filter(isInstanceElement)
-    .toArray()
+    const parentInstances = await awu(parentInstanceIds)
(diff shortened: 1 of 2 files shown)


packages/zendesk-adapter/src/change_validators/one_translation_per_locale.ts
```diff
@@ -81,6 +89,10 @@ const findDuplicateTranslations = async (
 }
 
 export const oneTranslationPerLocaleValidator: ChangeValidator = async (changes, elementSource) => {
+  if (elementSource === undefined) {
+    log.error('Failed to run findDuplicateTranslations because no element source was provided')
```
netama, line 93: nit: `findDuplicateTranslations` -> `oneTranslationPerLocale`

packages/zendesk-adapter/src/change_validators/one_translation_per_locale.ts
```diff
@@ -90,18 +102,19 @@ export const oneTranslationPerLocaleValidator: ChangeValidator = async (changes,
         CATEGORY_TRANSLATION_TYPE_NAME,
         SECTION_TRANSLATION_TYPE_NAME].includes(instance.elemID.typeName))
 
-  // it is a set to avoid duplicate of parents from translations which have the same parent
-  const parentInstances = relevantInstances
+  const parentInstancesId = relevantInstances
     .filter(instance => isTranslation(instance))
-    .flatMap(getParents)
-    .map(translationParentInstance => translationParentInstance.value)
+    .flatMap(instance => instance.annotations[CORE_ANNOTATIONS.PARENT])
```
netama, line 107: why not keep it as `getParents` ? without it we shouldn't assume this is an array, and it can also be undefined

packages/zendesk-adapter/src/change_validators/one_translation_per_locale.ts
```diff
@@ -90,18 +102,19 @@ export const oneTranslationPerLocaleValidator: ChangeValidator = async (changes,
         CATEGORY_TRANSLATION_TYPE_NAME,
         SECTION_TRANSLATION_TYPE_NAME].includes(instance.elemID.typeName))
 
-  // it is a set to avoid duplicate of parents from translations which have the same parent
-  const parentInstances = relevantInstances
+  const parentInstancesId = relevantInstances
```
netama, line 105: nit: `parentInstancesId` -> `parentInstanceIds`

packages/zendesk-adapter/src/change_validators/one_translation_per_locale.ts
```diff
@@ -90,18 +102,19 @@ export const oneTranslationPerLocaleValidator: ChangeValidator = async (changes,
...
+    .map(async id => elementSource.get(id))
+    .filter(isInstanceElement)
+    .toArray()
 
-  if (elementSource === undefined) {
-    log.error('Failed to run findDuplicateTranslations because no element source was provided')
-    return []
-  }
+  // it is a record to avoid duplicate of parents from translations which have the same parent
+  const parentInstancesObj = _.keyBy(parentInstances, instance => instance.elemID.getFullName())
```
netama, line 117: if the goal is uniqueness, I think we will benefit more from dedup'ing before calling the element source - maybe use
```suggestion
  const parentInstances = await awu(parentInstancesId)
    .uniqBy(id => id.getFullName())
    .map(async id => elementSource.get(id))
    .filter(isInstanceElement)
    .toArray()
```
and then we won't need the `parentInstancesObj` at all?

packages/zendesk-adapter/test/change_validators/one_translation_per_locale.test.ts
```diff
@@ -149,7 +149,7 @@ describe('oneTranslationPerLocalValidator',
 
       const errors = await oneTranslationPerLocaleValidator(
         [toChange({ after: enTranslation }), toChange({ after: enTranslation2 })],
-        elementSource.createInMemoryElementSource([enusLocale]),
+        elementSource.createInMemoryElementSource([enusLocale, article]),
```
netama, line 152: let's also change the parent references to not have the value and make sure this breaks for the old code?

netama (commented):
looks good :) added some comments.
for the release notes - is there a way to phrase that in a more user-facing way (e.g. "fix issue in a validation around Guide translations that could cause fialures in deployments of ...")

packages/zendesk-adapter/src/change_validators/one_translation_per_locale.ts
edenhassid: sounds good! it didn't work exactly like that so I did something a bit different.

packages/zendesk-adapter/src/change_validators/one_translation_per_locale.ts
```diff
@@ -90,20 +101,20 @@ export const oneTranslationPerLocaleValidator: ChangeValidator = async (changes,
...
         SECTION_TRANSLATION_TYPE_NAME].includes(instance.elemID.typeName))
 
-  // it is a set to avoid duplicate of parents from translations which have the same parent
-  const parentInstances = relevantInstances
+  const parentInstanceIds = _.uniqBy(relevantInstances
     .filter(instance => isTranslation(instance))
     .flatMap(getParents)
-    .map(translationParentInstance => translationParentInstance.value)
+    .flatMap(parent => (isReferenceExpression(parent) ? parent.elemID : undefined))
+    .filter(isDefined), id => id.getFullName())
```
netama, line 108: readability nit:
```suggestion
  const parentInstanceIds = _.uniqBy(
    relevantInstances
      .filter(instance => isTranslation(instance))
      .flatMap(getParents)
      .flatMap(parent => (isReferenceExpression(parent) ? parent.elemID : undefined))
      .filter(isDefined),
    id => id.getFullName()
  )
```
(the `id => id.getFullName()` part should be at the same "level" as `relevantInstances.<...>`)

packages/zendesk-adapter/src/change_validators/one_translation_per_locale.ts
```diff
@@ -90,20 +101,20 @@ export const oneTranslationPerLocaleValidator: ChangeValidator = async (changes,
...
 
-  const parentInstancesObj = _.keyBy(parentInstances, instance => instance.elemID.getFullName())
+  const parentInstances = await awu(parentInstanceIds)
```
netama, line 110: since we have a few checks that may "lose" instances along the way - let's add a log in case the length of `parentInstances` is smaller than `parentInstanceIds`? (maybe also log the specific missing ids)

netama (approved):
LGTM with a few nits (+ the test fix).
let's benchmark the time on a medium-size env+deployment to make sure using the element source didn't deteriorate the performance too much? (I think it should be ok, but worth verifying)

PR #3970, opened 2023-02-16
SALTO-3656 Fix field parents resolving

dantal4:
Today we assume that if we got a field and its parent type in the same resolve call, the field's parent will point to the same object type that we got to resolve, so we don't need to resolve both.

The problem is that we clone the elements before we resolve them, so the field’s parent don’t point to the same object type that we got to resolve.

---

---
_Release Notes_: 
Workspace:
- Fix field parents resolving

---
_User Notifications_: 
None

### DIFF
--- a/packages/workspace/src/expressions.ts
+++ b/packages/workspace/src/expressions.ts
@@ -247,12 +247,44 @@ const getResolveFunctions = ({
   }
 }
 
+const getClonedElements = (elements: Element[]): Element[] => {
+  const [fields, restOfElements] = _.partition(elements, isField)
+  const clonedRestOfElements = restOfElements.map(e => e.clone())
+  const clonedTypesMap = new Map(
+    clonedRestOfElements.filter(isObjectType)
+      .map(type => [type.elemID.getFullName(), type]),
+  )
+  const clonedMissingParentsMap = new Map(
+    fields.map(field => field.parent)
+      .filter(parent => !clonedTypesMap.has(parent.elemID.getFullName()))
+      .map(parent => [parent.elemID.getFullName(), parent.clone()])
+  )
+  // We want to get the fields from the cloned parents so we keep them pointing the parents
+  // that will be resolved and we don't use an uncloned parent type.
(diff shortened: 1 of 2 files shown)


packages/workspace/src/expressions.ts
```diff
@@ -247,12 +247,44 @@ const getResolveFunctions = ({
   }
 }
 
+const getClonedElements = (elements: Element[]): Element[] => {
+  const [fields, restOfElements] = _.partition(elements, isField)
+  const clonedRestOfElements = restOfElements.map(e => e.clone())
+  const clonedTypesMap = new Map(
```
adamro, line 253: Style question - Do you have a reason to use Map here and not Object.fromEntries?

packages/workspace/src/expressions.ts
```diff
@@ -247,12 +247,44 @@ const getResolveFunctions = ({
...
+      .filter(parent => !clonedTypesMap.has(parent.elemID.getFullName()))
+      .map(parent => [parent.elemID.getFullName(), parent.clone()])
+  )
+  // We want to get the fields from the cloned parents so we keep them pointing the parents
+  // that will be resolved and we don't use an uncloned parent type.
+  const clonedFields = fields.map(field => {
+    const fieldFromClonedParent = clonedTypesMap.get(field.parent.elemID.getFullName())?.fields[field.name]
+      ?? clonedMissingParentsMap.get(field.parent.elemID.getFullName())?.fields[field.name]
+    if (fieldFromClonedParent === undefined) {
+      log.error(
```
adamro, line 268: Shouldn't this be a warning if you are not actually crashing?

adamro (approved):
🔥

packages/workspace/src/expressions.ts
dantal4: I like the typesafety of `map.get` that returns `T | undefined`

packages/workspace/src/expressions.ts
dantal4: Will change 👍

PR #3971, opened 2023-02-16
SALTO-3634: Basic ScriptRunner Workflow Support

IdoZakk:
Added initial support for script runner on the cloud.
The current change enables the user to track changes on script runner workflows, and also edit them

---

The Script runner supports 3 workflow extensions- post functions, validators, and conditions. 

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/jira-adapter/src/filters/script_runner/workflow_scripts.ts
```diff
@@ -0,0 +1,182 @@
...
+  try {
+    const compressedObject = JSON.parse(Buffer.from(scriptRunnerString, 'base64').toString('utf8'))
+    if (!isCompressedObject(compressedObject)) {
+      return scriptRunnerString
+    }
+    const zipBuffer = Buffer.from(compressedObject.compressed)
+    const dataString = (await unzip(zipBuffer)).toString()
+    return JSON.parse(dataString)
+  } catch (e) {
+    log.warn('Could not decode script runner')
```
alonstern, line 82: I think we should use `error` here so it will be monitored by oncall

packages/jira-adapter/src/filters/script_runner/workflow_scripts.ts
```diff
@@ -0,0 +1,182 @@
...
+const encodeScriptRunner = async (object: Value): Promise<Value> => {
+  try {
+    const dataString = safeJsonStringify(object)
+    const zipBuffer = await zip(Buffer.from(dataString))
+    const compressedObject = {
+      compressed: zipBuffer.toJSON().data,
+    }
+    return Buffer.from(safeJsonStringify(compressedObject)).toString('base64')
+  } catch (e) {
+    log.warn('Could not encode script runner object')
```
alonstern, line 99: I think we should use `error` here so it will be monitored by oncall

packages/jira-adapter/src/filters/script_runner/workflow_scripts.ts
```diff
@@ -0,0 +1,182 @@
...
+    if (e instanceof Error) {
+      log.warn('error due to  %s', e.message)
+    }
+    return object
```
alonstern, line 103: I think in this case we really want to throw an error since deploying it as an object might mess up the workflow

packages/jira-adapter/src/filters/script_runner/workflow_scripts.ts
```diff
@@ -0,0 +1,182 @@
...
+import { createSchemeGuard, safeJsonStringify } from '@salto-io/adapter-utils'
+import Joi from 'joi'
+import { isInstanceElement, Element, isInstanceChange, isAdditionOrModificationChange, getChangeData, InstanceElement, Value } from '@salto-io/adapter-api'
+import { FilterCreator } from '../../filter'
+import { WORKFLOW_TYPE_NAME } from '../../constants'
+
+const { awu } = collections.asynciterable
+const { makeArray } = collections.array
+const log = logger(module)
+const scriptRunnerPostFunctionType = 'com.onresolve.jira.groovy.groovyrunner__script-postfunction'
```
alonstern, line 29: super nit: consts convention:
```suggestion
const SCRIPT_RUNNER_POST_FUNCTION_TYPE = 'com.onresolve.jira.groovy.groovyrunner__script-postfunction'
```

alonstern (commented):
Looks great 🔥 
- This only supports Cloud right? Can you have a look at DC to see if it's a minimal effort to add support for it as well?
- We should check with @liorn if UX-wise we want to keep it in the same element as it is currently or put it in a new instance

packages/jira-adapter/src/filters/script_runner/workflow_scripts.ts
```diff
@@ -0,0 +1,182 @@
...
+const filter: FilterCreator = () => ({
+  name: 'scriptRunnerWorkflowFilter',
+  onFetch: async (elements: Element[]) => {
+    await awu(elements)
+      .filter(isInstanceElement)
+      .filter(instance => instance.elemID.typeName === WORKFLOW_TYPE_NAME)
+      .forEach(async instance => {
+        await applyOnScriptRunnersFields({ instance, func: decodeScriptRunner, field: 'postFunctions', typeName: scriptRunnerPostFunctionType })
+        await applyOnScriptRunnersFields({ instance, func: fallBackJsonParse, field: 'validators', typeName: scriptRunnerValidatorType })
+        await applyOnScriptRunnersFields({ instance, func: fallBackJsonParse, field: 'conditions', typeName: scriptRunnerConditionType })
```
alonstern, line 153: in `Alon_s_Test_Workflow_22e@tsss` in ori-salto-test there are some scripts that are still a string of a json for some reason

packages/jira-adapter/src/filters/script_runner/workflow_scripts.ts
netama: not a nit actually :)

packages/jira-adapter/src/filters/script_runner/workflow_scripts.ts
IdoZakk: I changed to walk on value to account for that

packages/jira-adapter/src/filters/script_runner/workflow_scripts.ts
```diff
@@ -0,0 +1,170 @@
...
+
+const typeToDecodeFuncMap: TypeToCodeFuncMap = new Map([
+  [SCRIPT_RUNNER_POST_FUNCTION_TYPE, decodeScriptRunner],
+  [SCRIPT_RUNNER_VALIDATOR_TYPE, fallBackJsonParse],
+  [SCRIPT_RUNNER_CONDITION_TYPE, fallBackJsonParse],
+])
+
+const walkOnWorkflow = (typeMap: TypeToCodeFuncMap): WalkOnFunc => (
+  ({ value }): WALK_NEXT_STEP => {
+    if (Array.from(typeMap.keys()).includes(value.type) && value.configuration !== undefined) {
```
alonstern, line 116: nit: `Array.from(typeMap.keys()).includes(value.type)` -> `typeMap.has(value.type)`

alonstern (dismissed):
👌 🏃 
Added some comments
Please remove the user notifications since this is now under a flag

IdoZakk:
Sorry for that, I first deployed the changes for DC and then the PR comments (applied also on DC...)

packages/jira-adapter/src/filters/script_runner/workflow_scripts_dc.ts
```diff
@@ -0,0 +1,146 @@
...
+
+const transformConfigFields = (funcMap: FieldToCodeFuncMap): WalkOnFunc => (
+  ({ value }): WALK_NEXT_STEP => {
+    if (SCRIPT_RUNNER_DC_TYPES.includes(value.type) && value.configuration !== undefined) {
+      funcMap.forEach((_value, fieldName) => {
+        // Field comment is base64 encoded only in some cases
+        if (value.configuration[fieldName] !== undefined
+          && (fieldName !== 'FIELD_COMMENT' || value.configuration[CANNED_SCRIPT] === FIELD_COMMENT_TYPE)) {
+          value.configuration[fieldName] = funcMap.get(fieldName)(value.configuration[fieldName])
+        }
```
alonstern, line 92: So `FIELD_COMMENT` is base64 encoded only when `value.configuration[CANNED_SCRIPT] === FIELD_COMMENT_TYPE` and when it is not base64 no encoding is needed at all? How does it look in the NaCl when it is not base64

packages/jira-adapter/src/filters/script_runner/workflow_scripts_dc.ts
```diff
@@ -0,0 +1,146 @@
...
+const SCRIPT_RUNNER_DC_TYPES = ['com.onresolve.jira.groovy.GroovyFunctionPlugin',
+  'com.onresolve.jira.groovy.GroovyValidator',
+  'com.onresolve.jira.groovy.GroovyCondition']
+const DC_BASE64_PREFIX = 'YCFg'
+const CANNED_SCRIPT = '"canned-script"'
+const FIELD_COMMENT_TYPE = 'com.onresolve.scriptrunner.canned.jira.workflow.postfunctions.CommentIssue'
+
+
+const decodeBase64 = (base64: string): string => {
+  if (!base64.startsWith(DC_BASE64_PREFIX)) {
```
alonstern, line 33: Why is this check needed? I am not sure we should rely on that

packages/jira-adapter/src/filters/script_runner/workflow_scripts_dc.ts
```diff
@@ -0,0 +1,146 @@
...
+          && (fieldName !== 'FIELD_COMMENT' || value.configuration[CANNED_SCRIPT] === FIELD_COMMENT_TYPE)) {
+          value.configuration[fieldName] = funcMap.get(fieldName)(value.configuration[fieldName])
+        }
+      })
+      return WALK_NEXT_STEP.SKIP
+    }
+    return WALK_NEXT_STEP.RECURSE
+  })
+
+const filter: FilterCreator = ({ client, config }) => ({
```
alonstern, line 99: WDYT about creating one filter for DC and Cloud in a common file, and creating a file for dc and a file for cloud that will export each `decodeConfigFields: WalkOnFunc = transformConfigFields(fieldToDecodeMap)` and `encodeConfigFields: WalkOnFunc = transformConfigFields(fieldToEncodeMap)` ?

packages/jira-adapter/src/filters/script_runner/workflow_scripts_dc.ts
```diff
@@ -0,0 +1,146 @@
...
+
+const decodeBase64 = (base64: string): string => {
+  if (!base64.startsWith(DC_BASE64_PREFIX)) {
+    log.warn(`Could not decode DC ScriptRunner script, expected base64, got: ${base64}`)
+    return base64
+  }
+  return Buffer.from(base64.substring(DC_BASE64_PREFIX.length), 'base64').toString('utf8')
+}
+
+const encodeBase64 = (script: string): string => DC_BASE64_PREFIX + Buffer.from(script).toString('base64')
```
alonstern, line 40: I think it's better to do 
```suggestion
const encodeBase64 = (script: string): string => Buffer.from('`!`' + script).toString('base64')
```

Concatenating base64 parts like that feels dangerous

alonstern (commented):
👌

packages/jira-adapter/src/filters/script_runner/workflow_scripts_dc.ts
IdoZakk: You can take a look here: https://github.com/salto-io/ScriptRunnerNaCls
In general, this is weird, in some cases it is just plain text and in others it is base64 encoded

packages/jira-adapter/src/filters/script_runner/workflow_scripts_dc.ts
IdoZakk: What are the alternatives? 
As I need to put it back on during encoding and up till now all the base64 strings in DC I saw had that prefix storing it in a hidden variable did not seem wise

packages/jira-adapter/src/filters/script_runner/workflow_scripts_dc.ts
IdoZakk: As it is the start and exactly 4 base64 chars it seems ok to me and cleaner this way, do you have a specific scenario in mind?

IdoZakk:
Fixed one comment and replied to the rest. 
I split it into two commits for easier tracking, first one is the changes, the second one is name changes

packages/jira-adapter/src/filters/script_runner/workflow_dc.ts
```diff
@@ -0,0 +1,98 @@
...
+import { logger } from '@salto-io/logging'
+import { safeJsonStringify, WalkOnFunc, WALK_NEXT_STEP } from '@salto-io/adapter-utils'
+import { Value } from '@salto-io/adapter-api'
+
+const log = logger(module)
+const SCRIPT_RUNNER_DC_TYPES = ['com.onresolve.jira.groovy.GroovyFunctionPlugin',
+  'com.onresolve.jira.groovy.GroovyValidator',
+  'com.onresolve.jira.groovy.GroovyCondition']
+const DC_BASE64_PREFIX = 'YCFg'
+const CANNED_SCRIPT = '"canned-script"'
```
alonstern, line 26: Are you sure this needs to be with `"` inside? the NaCl parse adds the `"` because of the `-` in the middle but IIRC it also removes it when loading the nacl

packages/jira-adapter/src/filters/script_runner/workflow_scripts_dc.ts
alonstern: I'm not sure it is cleaner, the encoded part does not have any meaning so it's not clear what you are actually adding to the string. I needed to decode the string to understand what's happening in this part

packages/jira-adapter/src/filters/script_runner/workflow_scripts_dc.ts
alonstern: Gotcha about why this is needed. Same as the other comment I think it is clearer to decode the base64 and check if it starts with \`!\`. Also a comment might be helpful here about that that all the base64 encoded things in DC start with that string and that we remove it so the rest of the string will a valid json

IdoZakk: If a string starts with the base64 prefix, you are pretty safe to try and decode it from base64. Otherwise, you decode first, which might fail, so you must wrap it in try-catch. I added the requested change, if you think it is cleaner I will keep it

alonstern (approved):
👍 👌 🏃

PR #3976, opened 2023-02-16
Return only added elements in getReferencedElements

dantal4:
Instead of returning the referenced elements with the input elements, we should return only the referenced elements.
That let us control the return type of the function (it can be `ObjectType | InstanceElement`), and avoid a case of returning an equivalent to an input element but not the exact same object (if it is also referenced, and we return the referenced element instead of the input element).
The thought behind that change is that we want to use a cloned version of the added changes in `additional_changes.ts` - but we don't want to clone the added field parents, because it will separate the `field.parent` property from the added field parent.

---

---
_Release Notes_: 
Netsuite Adapter:
- Return only added elements in getReferencedElements

---
_User Notifications_: 
None

### DIFF
--- a/packages/netsuite-adapter/src/reference_dependencies.ts
+++ b/packages/netsuite-adapter/src/reference_dependencies.ts
@@ -17,7 +17,7 @@ import _ from 'lodash'
 import { logger } from '@salto-io/logging'
 import {
   isInstanceElement, isPrimitiveType, ElemID, getFieldType,
-  isReferenceExpression, Value, isServiceId, isObjectType, ChangeDataType,
+  isReferenceExpression, Value, isServiceId, isObjectType, ChangeDataType, ObjectType, InstanceElement,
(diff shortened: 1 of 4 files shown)


packages/netsuite-adapter/src/reference_dependencies.ts
```diff
@@ -30,14 +30,18 @@ const { awu } = collections.asynciterable
 const { isDefined } = lowerDashValues
 const log = logger(module)
 
+type TopLevelElement = ObjectType | InstanceElement
```
omrilit, line 33: Don't we have this type elsewhere?

omrilit (approved):
Looks good!

packages/netsuite-adapter/src/reference_dependencies.ts
dantal4: No..

PR #3981, opened 2023-02-19
Salto-3671/validate saved searches dates

guyhamo:
filter system note/ saved search query results to not include changes from the future.

---

---
_Release Notes_: 
netsuite adapter: 
* changed at annotations will no longer be able to be from the future.

---
_User Notifications_:


Shulik95 (approved):
💯

PR #3985, opened 2023-02-19
SALTO-3663: deploy warning when modifying managed Elements in Salesforce

tamtamirr:
Added a deploy warning when modifying managed Elements in Salesforce.

---

Previous to this PR, the `package` Change Validator had a set of rules that validated if a change on an `Element` from a managed installed package is valid. As the time went by, this filter turned out to be way too restrictive, and blocked many valid changes.

I've changed the current implementation to yield a warning when attempting to modify Element from managed installed package.

---
_Release Notes_: 
Salesforce Adapter:
- Modification of Elements from managed installed package won't be blocked, and instead an informative warning message will be created.
---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/change_validators/package.ts
```diff
@@ -56,133 +52,25 @@ export const hasNamespace = async (customElement: Element): Promise<boolean> =>
...
-  detailedMessage?: string,
-): Promise<ChangeError> => {
-  const packageNamespace = await getNamespace(element)
-  return ({
-    elemID: element.elemID,
-    severity: 'Error',
-    message: `Cannot change a managed package using Salto. Package namespace: ${packageNamespace}`,
-    detailedMessage: detailedMessage ?? `Cannot ${action} ${element.elemID.getFullName()} because it is part of a package namespace: ${packageNamespace}`,
-  })
+export const getNamespace = async (customElement: Element): Promise<string | undefined> => {
```
tamtamirr, line 55: This implementation is old, but as for the time I'm implementing this PR, my fixes to `getNamespace` and `getNamespaceFromString` are not merged. When they do, we can use `getNamespace` from utils instead.

packages/salesforce-adapter/src/change_validators/package.ts
```diff
@@ -13,32 +13,28 @@
...
 
+const createPackageElementModificationChangeWarning = (
+  { elemID, annotations }: Element,
+  namespace: string
+): ChangeError => ({
+  elemID,
+  severity: 'Warning',
+  message: 'Modification of element from managed package may not be allowed',
+  detailedMessage: `Modification of element ${elemID.getFullName()} from managed package with namespace ${namespace} may not be allowed. `
+  + `For more information refer to ${annotations[CORE_ANNOTATIONS.SERVICE_URL]} https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/packaging_packageable_components.htm`,
```
amitgoldarb, line 35: I think this is missing a comma:
```suggestion
  + `For more information refer to ${annotations[CORE_ANNOTATIONS.SERVICE_URL]}  and https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/packaging_packageable_components.htm`,
```

amitgoldarb, line 35: Also I think the service_url might not be sufficient for some elements. For example for flows, we send to flow builders. Do they have the message on top?
I'm not sure we want to couple these 2 links. But then again, it in _most_ cases has the most relevant infomration. Also I think we don't support the url for all elements. Maybe we need some fallback if the url doesn't exist? Maybe the message itself could help?
For example - 
```
For more information refer to the element's link <element_url>, if you cannot find the information there please look https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/packaging_packageable_components.htm for more information.
```
@tomermevorach WDYT?

amitgoldarb (approved):
LGTM!
Thanks for this PR!
I think we should update the message a bit though before merging. I added a few suggestions, please go through @tomermevorach for final approval.

PR #3987, opened 2023-02-20
SALTO-3602/issue type scheme migration

guyhamo:
handled a migration error for issue type schemes when we either change a project issue type scheme or when we remove an issue type from a scheme that is being used by a project with the same typed issues

---

added tests for the new validator.

---
_Release Notes_: 
jira adapter:
* Add change validator to block removal of issue type from a scheme if that issue type is being used in assigned projects.
* Block Issue type scheme changes in active projects.

---
_User Notifications_:

### DIFF
--- a/packages/jira-adapter/src/change_validators/issue_type_scheme_migration.ts
+++ b/packages/jira-adapter/src/change_validators/issue_type_scheme_migration.ts
@@ -98,7 +98,8 @@ export const issueTypeSchemeMigrationValidator = (
       .map(id => elementSource.get(id))
       .toArray()
     const issueTypeSchemesToProjects = _.groupBy(
-      projects.filter(project => project.value.issueTypeScheme !== undefined),
+      projects.filter(project => project.value.issueTypeScheme !== undefined
+        && isReferenceExpression(project.value.issueTypeScheme)),
       project => project.value.issueTypeScheme.elemID.getFullName(),
     )
     const errors = await awu(relevantChanges).map(async change => {
--- a/packages/jira-adapter/src/change_validators/issue_type_scheme_migration.ts
+++ b/packages/jira-adapter/src/change_validators/issue_type_scheme_migration.ts
@@ -13,7 +13,7 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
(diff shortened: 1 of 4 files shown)


packages/jira-adapter/src/change_validators/issue_type_scheme_migration.ts
```diff
@@ -0,0 +1,122 @@
...
+      return []
+    }
+    const idsIterator = awu(await elementSource.list())
+    const projects = await awu(idsIterator)
+      .filter(id => id.typeName === PROJECT_TYPE)
+      .filter(id => id.idType === 'instance')
+      .map(id => elementSource.get(id))
+      .toArray()
+    const issueTypeSchemesToProjects = _.groupBy(
+      projects.filter(project => project.value.issueTypeScheme !== undefined),
```
alonstern, line 101: Better check that `issueTypeScheme` is a refernce

packages/jira-adapter/src/change_validators/issue_type_scheme_migration.ts
```diff
@@ -0,0 +1,122 @@
...
+    )
+    const errors = await awu(relevantChanges).map(async change => {
+      const issueTypeScheme = getChangeData(change)
+      const linkedProjectNames = issueTypeSchemesToProjects[issueTypeScheme.elemID.getFullName()]
+        ?.map(project => project.value.name) ?? []
+      if (linkedProjectNames.length === 0) {
+        return undefined
+      }
+      const removedIssueTypeNames = getRemovedIssueTypeIds(change)
+        .map(issueTypeId => issueTypeId.elemID.name)
```
alonstern, line 112: Should it be `issueTypeId.value.value.name`?

packages/jira-adapter/src/change_validators/issue_type_scheme_migration.ts
```diff
@@ -0,0 +1,122 @@
...
+    response = await client.getSinglePage({
+      url: '/rest/api/3/search',
+      queryParams: {
+        jql,
+        maxResults: '0',
+      },
+    })
+  } catch (e) {
+    log.error(`Received an error Jira search API, ${e.message}. Assuming issue type "${issueType}" has issues.`)
+    return true
```
alonstern, line 76: I think when something went wrong it's better to assume the issue types are not being used and worst case the user will get an error from the service

alonstern (approved):
👌 👌

PR #3988, opened 2023-02-20
SALTO-3454 add quick-deploy fallback to deploy

Uri-Bar:
SALTO-3454 add quick-deploy fallback to deploy
---
_Release Notes_: 
None

---
_User Notifications_: 
None


adamro (approved):
Just add a unit test

Approving to save a cycle

PR #3990, opened 2023-02-20
CV wording changes

benisraeladi:
Wording changes for Salesforce change validators

---

---
_Release Notes_: 

---
_User Notifications_: _None_

### DIFF
--- a/packages/salesforce-adapter/src/change_validators/cpq_trigger.ts
+++ b/packages/salesforce-adapter/src/change_validators/cpq_trigger.ts
@@ -28,7 +28,7 @@ const getCpqError = (
   elemID,
   severity: 'Info',
   message: 'CPQ data changes detected',
-  detailedMessage: 'CPQ data changes detected',
+  detailedMessage: '',
   deployActions: {
     preAction: {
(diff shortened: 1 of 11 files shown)


packages/salesforce-adapter/src/change_validators/cpq_trigger.ts
```diff
@@ -28,7 +28,7 @@ const getCpqError = (
   elemID,
   severity: 'Info',
   message: 'CPQ data changes detected',
-  detailedMessage: '',
+  detailedMessage: 'CPQ data changes detected',
```
dorsev, line 31: I think this must stay "Empty".
lets discuss f2f

dorsev (approved):
LGTM. I have one blocker (IMO).
lets discuss it f2f

packages/salesforce-adapter/src/change_validators/cpq_trigger.ts
benisraeladi: If detailedMessage is not empty, this will show in the UI, which is not the required behavior. @tomermevorach please ack this

PR #3992, opened 2023-02-21
SALTO-3681: `getPath` does not correctly when `rootElement` is a field

alonstern:
Fixed getPath to work with fields and the rootElement

---
_Release Notes_: 
None (no effect on users currently)

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-utils/src/utils.ts
+++ b/packages/adapter-utils/src/utils.ts
@@ -647,6 +647,9 @@ export const getPath = (
   rootElement: Element,
   fullElemID: ElemID,
 ): string[] | undefined => {
+  // If rootElement is an objectType or an instance (i.e., a top level element),
+  // we want to compare the top level id of fullElemID.
+  // If it is a field we want to compare the base id.
   const { parent, path } = rootElement.elemID.isTopLevel()
     ? fullElemID.createTopLevelParentID()
     : fullElemID.createBaseID()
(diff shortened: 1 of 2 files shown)


packages/adapter-utils/test/utils.test.ts
```diff
@@ -1534,6 +1534,12 @@ describe('Test utils.ts', () => {
       expect(clonedMockInstance.value.str).toEqual('new val')
     })
 
+    it('should set field annotation', () => {
+      const clonedMockField = new Field(mockType, 'field', BuiltinTypes.STRING)
+      setPath(clonedMockField, clonedMockField.elemID.createNestedID('str'), 'new val')
```
netama, line 1539: if the fix is on `getPath` maybe we can have a direct test as well, and not only based on `setPath`?

packages/adapter-utils/src/utils.ts
```diff
@@ -647,7 +647,10 @@ export const getPath = (
   rootElement: Element,
   fullElemID: ElemID,
 ): string[] | undefined => {
-  const { parent, path } = fullElemID.createTopLevelParentID()
+  const { parent, path } = rootElement.elemID.isTopLevel()
```
netama, line 650: continuing from IRL - I think we should add an explanation here, as it's easy to miss the fact that toe condition is on `rootElement` and not on `fullElemID`. I'm not convinced this is a scenario we should handle here at all tbh, but if we do then we should explain the specific use case of the `rootElement` being an object type and then path being nested under a field

netama (approved):
I think we might have been able to consolidate the cases more, but not sure it's critical (and maybe I'm missing something and we couldn't). 
LGTM - please add a dedicated test as well if possible

PR #3993, opened 2023-02-21
Attempt to fix flaky timeout test

alonstern:
An attempt to fix a flaky timeout test

---

Tried:
- creating the P variable separately in each test in case of a race between the tests
- increased the timeout in the tests

---
_Release Notes_: 
None

---
_User Notifications_: 
None


tamtamirr (approved):
Thanks @alonstern ! Can we make sure there's no more places where we `sleep` inside `beforeEach`?

PR #3994, opened 2023-02-21
SALTO-3669/user list permission failure

guyhamo:
users without permission to view other members in the service will fail running "getUserMapFuncCreator"
this change will make this function throw a specific error when it happens and everywhere that uses this code is now expecting this error and fail gracefully

added a log when it happens.

---

---
_Release Notes_: 
jira adapter:
* Fixed bug where missing permissions to view other users will fail deploy/fetch

---
_User Notifications_:

### DIFF
--- a/packages/jira-adapter/src/change_validators/account_id.ts
+++ b/packages/jira-adapter/src/change_validators/account_id.ts
@@ -255,7 +255,6 @@ export const accountIdValidator: (
         userMap = getUsersMapByVisibleId(await getUserMapFunc(), client.isDataCenter)
       } catch (e) {
         if (e instanceof MissingUsersPermissionError) {
-          config.fetch.convertUsersIds = false
           return []
         }
         throw e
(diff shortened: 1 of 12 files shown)


packages/jira-adapter/src/change_validators/account_id.ts
```diff
@@ -246,11 +246,19 @@ export const accountIdValidator: (
...
   ChangeValidator = (client, config, getUserMapFunc) => async changes =>
     log.time(async () => {
+      let userMap: UserMap
       if (!(config.fetch.convertUsersIds ?? true)) {
         return []
       }
       const { baseUrl, isDataCenter } = client
-      const userMap = getUsersMapByVisibleId(await getUserMapFunc(), client.isDataCenter)
+      try {
+        userMap = getUsersMapByVisibleId(await getUserMapFunc(), client.isDataCenter)
```
alonstern, line 255: I understand why you threw an error in this case but I tend to think the code will be clearer if you return undefined from `getUserMapFunc` in the `getUserMapFunc`. Also it will force the user to handle this case so we won't miss calls to this function. But your call

alonstern (approved):
👌

packages/jira-adapter/src/change_validators/account_id.ts
guyhamo: I changed it to undefined.

PR #3997, opened 2023-02-22
SALTO-3050: Clear secrets from response data before writing to log

shir-reifenberg:
Clear secrets from response data before writing to log

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/client/client.ts
+++ b/packages/okta-adapter/src/client/client.ts
@@ -36,34 +36,6 @@ const DEFAULT_PAGE_SIZE: Required<clientUtils.ClientPageSizeConfig> = {
   get: 50,
 }
 
-/**
- * Clear response data values might contain secrets, returns a new object
- */
-const clearSecrets = (
-  object: Values,
-  secretFieldNames: string[]
-): Values => {
-  const SECRET_PLACEHOLER = '<SECRET>'
-  const replaceValuesInplace = (values: Values): void => {
-    if (_.isPlainObject(values)) {
-      Object.keys(values).forEach(key => {
-        if (secretFieldNames.includes(key)) {
-          values[key] = SECRET_PLACEHOLER
-        } else if (_.isPlainObject(values[key]) || Array.isArray(values[key])) {
-          replaceValuesInplace(values[key])
-        }
-      })
-    }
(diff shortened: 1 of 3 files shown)


packages/okta-adapter/src/config.ts
```diff
@@ -708,14 +708,6 @@ const DEFAULT_TYPE_CUSTOMIZATIONS: OktaApiConfig['types'] = {
       ],
     },
   },
-  IdentityProviderCredentialsClient: {
```
shir-reifenberg, line 711: Nor necessary cause it's nested under `credentials` in `Protocol` type which we already omit in line 703

packages/okta-adapter/src/client/client.ts
```diff
@@ -34,6 +36,34 @@ const DEFAULT_PAGE_SIZE: Required<clientUtils.ClientPageSizeConfig> = {
...
 
+/**
+ * Clear response data values might contain secrets, returns a new object
+ */
+const clearSecrets = (
+  object: Values,
+  secretFieldNames: string[]
+): Values => {
+  const SECRET_PLACEHOLER = '<SECRET>'
+  const replaceValuesInplace = (values: Values): void => {
```
netama, line 47: tiny nit: `replaceValuesInplace` -> `replaceValuesInPlace`

packages/okta-adapter/src/client/client.ts
```diff
@@ -34,6 +36,34 @@ const DEFAULT_PAGE_SIZE: Required<clientUtils.ClientPageSizeConfig> = {
...
+          replaceValuesInplace(values[key])
+        }
+      })
+    }
+    if (Array.isArray(values)) {
+      values.forEach(item => replaceValuesInplace(item))
+    }
+  }
+
+  const clonedObject = _.cloneDeep(object)
```
netama, line 62: I think we can use `cloneDeepWith` and have the "steps" of `replaceValuesInplace` in the `customizer` and avoid the double traversal/recursion, wdyt?

netama (commented):
looks good :) added a suggestion that I _think_ can simplify the logic, if you prefer this approach lmk and I'll approve

netama (approved):

PR #3998, opened 2023-02-22
SALTO-3050: Clear secrets from response data before writing to log (adapter-components changes)

shir-reifenberg:
Just the adapter-components changes from https://github.com/salto-io/salto/pull/3997

---
_Release Notes_: 
None

---
_User Notifications_: 
None


netama (approved):
🎈  approving but please add a test before merging

PR #3999, opened 2023-02-22
SALTO-3690/fixed bug with unresolved references

guyhamo:
fixed bug that was caused over an unresolved reference for multiple change validators.

---

---
_Release Notes_: 

---
_User Notifications_:

### DIFF
--- a/packages/jira-adapter/src/change_validators/field_contexts/field_contexts.ts
+++ b/packages/jira-adapter/src/change_validators/field_contexts/field_contexts.ts
@@ -13,7 +13,7 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
-import { ChangeValidator, getChangeData, InstanceElement, isReferenceExpression, ReadOnlyElementsSource, ReferenceExpression } from '@salto-io/adapter-api'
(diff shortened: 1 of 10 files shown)


packages/jira-adapter/src/change_validators/field_contexts/field_contexts.ts
```diff
@@ -37,6 +37,9 @@ const getFieldContexts = async (
       }
       return true
     })
+    .filter(async (ref: ReferenceExpression): Promise<boolean> =>
+      (await elementSource.has(ref.elemID)
+      && !(await ref.getResolvedValue(elementSource) instanceof PlaceholderObjectType)))
```
alonstern, line 42: I don't think the `PlaceholderObjectType` is needed and I think we better not rely on that
```suggestion
      await elementSource.has(ref.elemID)
```

alonstern (approved):
👌

packages/jira-adapter/src/change_validators/field_contexts/field_contexts.ts
guyhamo: changed to is instance element on ref.value

PR #4000, opened 2023-02-22
SALTO-3689 fix broken fetch operation

omrifarkash:
---
The bug was: 
In the old code, core saved the fetchFunction as const and call it. If inside the fetch function there are references to some of the adapter propertise (e.g this.someAttribute) 'this' was undefined and thrown an error.
We should call the function directly from the adapter it self (adapter.fetch(...args))

Behavior change:
When calling fetchWithChange detection with adapter that do not support this operation
In the old code a single error were thrown, contains all the unsupported fetchWithChangeDetection adapter names.
In the new code a single error will be thrown, contains only the first unsupported adapter name.

IMO this is good enough since this case shouldn't happen, and if it does the odds that there are more than a single adapter that does not support the operation is even smaller.
---
_Release Notes_: 

---
_User Notifications_: 
None

### DIFF
--- a/packages/core/src/core/fetch.ts
+++ b/packages/core/src/core/fetch.ts
@@ -375,7 +375,7 @@ type RunAdapterFetchFunctionParams = {
   adapter: AdapterOperations
   fetchOptions: FetchOptions
   adapterName: string
-  withChangeDetection?: boolean
+  withChangeDetection: boolean | undefined
 }
 
 const runAdapterFetchFunction = ({
@@ -463,7 +463,7 @@ const fetchAndProcessMergeErrors = async (
       (_adapter, accountName) => createAdapterProgressReporter(accountName, 'fetch', progressEmitter)
     )
     const fetchResults = await Promise.all(
-      Object.entries((accountsToAdapters))
+      Object.entries(accountsToAdapters)
         .map(async ([accountName, adapter]) => {
           const fetchOptions = { progressReporter: progressReporters[accountName] }
           const fetchResult = await runAdapterFetchFunction({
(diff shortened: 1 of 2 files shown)


packages/core/src/core/fetch.ts
```diff
@@ -371,26 +371,26 @@ const runPostFetch = async ({
...
 
-type CreateAdapterToFetchFunctionMapParams = {
-  accountsToAdapters: Record<string, AdapterOperations>
-  accountToServiceNameMap: Record<string, string>
-  withChangeDetection: boolean | undefined
+type RunAdapterFetchFunctionParams = {
+  adapter: AdapterOperations
+  fetchOptions: FetchOptions
+  adapterName: string
+  withChangeDetection?: boolean
```
dantal4, line 378: My personal preference is not to have optional args if it's not needed, and since it is called only once it can be required IMO

packages/core/src/core/fetch.ts
```diff
@@ -462,14 +462,15 @@ const fetchAndProcessMergeErrors = async (
...
     )
-    const accountNameToFetchFunction = createAdapterToFetchFunctionMap({
-      accountsToAdapters, accountToServiceNameMap, withChangeDetection,
-    })
     const fetchResults = await Promise.all(
-      Object.entries((accountNameToFetchFunction))
-        .map(async ([accountName, fetchFunction]) => {
-          const fetchResult = await fetchFunction({
-            progressReporter: progressReporters[accountName],
+      Object.entries((accountsToAdapters))
```
dantal4, line 466:
```suggestion
      Object.entries(accountsToAdapters)
```

dantal4 (approved):
LGMT

PR #4001, opened 2023-02-22
SALTO-3691: Remove all User and AppUser related code in Okta

shir-reifenberg:
Remove all custom `User` and `AppUser` code from Okta adapter

---

As we currently not managing users in Okta, I deleted all code related to `User` and `AppUser` types cause it will no longer be maintained.

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/okta-adapter/src/filters/replace_object_with_id.ts
```diff
@@ -19,9 +19,8 @@ import { FilterCreator } from '../filter'
 
 const TYPE_TO_REFERENCES_TYPES: Record<string, string[]> = {
   Application: ['assignedGroups'],
-  Group: ['users', 'roles'],
-  IdentityProvider: ['users'],
-  Feature: ['featureDependents', 'featureDependencies'],
```
shir-reifenberg, line 24: Not really related to this PR but it's a leftover from old code so I removed this as well

netama (approved):

PR #4004, opened 2023-02-22
SALTO-3696/workflow schemes without items

guyhamo:
Fixed bug with workflow scheme change validator that assums that items field is always populated 

---

---
_Release Notes_: 
Jira adapter:
* bug fix for failing deployments over workflow schemes with no items. 

---
_User Notifications_:


netama (approved):
LGTM for the fix - please add release notes before merging.
can you please work with @alonstern to see if there are other assumptions we are making in other places that should be validated with a schemeguard?

PR #4005, opened 2023-02-22
SALTO-1730 - Salesforce: Fix handling of references to instances in formulas

tomsellek:
Following a training session with @tamtamirr I realized the current way we handle custom labels and custom metadata type records is wrong. This is the fix.

---

Workspace diff: https://github.com/salto-io/tomsellek-sf/pull/3

---
_Release Notes_: 
Salesforce: Improved handling of references in formulas

---
_User Notifications_: 
N/A


tamtamirr (approved):
LGTM :)

PR #4006, opened 2023-02-23
SALTO-3698/missing fields

guyhamo:
fixed potential bug with workflow scheme migration change validator potentially using workflow statuses when its undefined 
and another one in issue type scheme that might not have issue type ids.

---

---
_Release Notes_: 
jira adapter:
* Fixed cases in some change validators that failed on missing fields in instances.

---
_User Notifications_:


alonstern (approved):
👌 
Please add release notes

PR #4007, opened 2023-02-23
Downgraded node version to 14.15

alonstern:
Downgraded node version to 14.15 since it looks like the UTs are still flaky on the current version

---
_Release Notes_: 
None

---
_User Notifications_: 
None


adamro (approved):
Approving but @alonstern but hold with merging since @netama wanted a short sync before we do it

netama (approved):

PR #4011, opened 2023-02-23
SALTO-3660, SALTO-3661: add alias to core and instance annotation, add alias index

edenhassid:
add alias to core and instance annotation
add alias index
add prettifyName function to nacl_case_utils

---

---
_Release Notes_: 
core:
* add alias to core and instance annotation

workspace:
* add alias index

---
_User Notifications_: 


packages/workspace/src/workspace/alias_index.ts
```diff
@@ -0,0 +1,131 @@
...
+import { logger } from '@salto-io/logging'
+import _ from 'lodash'
+import { prettifyName } from '@salto-io/adapter-utils'
+import { ElementsSource } from './elements_source'
+import { getAllElementsChanges } from './index_utils'
+import { RemoteMap } from './remote_map'
+
+
+const log = logger(module)
+export const ALIAS_INDEX_VERSION = 4
```
netama, line 35: should start at 1 :) (I am not sure we need it here, but probably won't hurt either and will make it easier to re-calculate if we need to)

packages/workspace/src/workspace/alias_index.ts
```diff
@@ -0,0 +1,131 @@
...
+
+const getFieldsElemIDsFullName = (objectType: ObjectType): string[] =>
+  Object.values(objectType.fields).map(field => field.elemID.getFullName())
+
+const getChangedFields = (change: ModificationChange<ObjectType>): Change<Element>[] => {
+  const afterFields = Object.values(change.data.after.fields)
+  const beforeFields = Object.values(change.data.before.fields)
+  const additionFields = afterFields.filter(field => !beforeFields.find(f => f.elemID.isEqual(field.elemID)))
+  const removalFields = beforeFields.filter(field => !afterFields.find(f => f.elemID.isEqual(field.elemID)))
+  const aliasModificationFields = afterFields
```
netama, line 46: nit: maybe `addedFields`, `removedFields`, `aliasModifiedFields`?

packages/workspace/src/workspace/alias_index.ts
```diff
@@ -0,0 +1,131 @@
...
+const log = logger(module)
+export const ALIAS_INDEX_VERSION = 4
+const ALIAS_INDEX_KEY = 'alias_index'
+
+const getFieldsElemIDsFullName = (objectType: ObjectType): string[] =>
+  Object.values(objectType.fields).map(field => field.elemID.getFullName())
+
+const getChangedFields = (change: ModificationChange<ObjectType>): Change<Element>[] => {
+  const afterFields = Object.values(change.data.after.fields)
+  const beforeFields = Object.values(change.data.before.fields)
```
netama, line 43: I think we should also create sets of the elem ids of each of these, so that the calculations are more efficient. it probably won't make a huge difference but still since we will run this code a lot I think it's worth it.
also - since these are all fields of the same object, we can probably compare just the field names and not the full elem id for finding additions/removals (so - either create a set from the object keys, or check if `Object.prototype.hasOwnProperty.call(objectType.fields, fieldName)` is true)

packages/workspace/src/workspace/alias_index.ts
```diff
@@ -0,0 +1,131 @@
...
+  const additionFields = afterFields.filter(field => !beforeFields.find(f => f.elemID.isEqual(field.elemID)))
+  const removalFields = beforeFields.filter(field => !afterFields.find(f => f.elemID.isEqual(field.elemID)))
+  const aliasModificationFields = afterFields
+    .filter(afterField => {
+      const beforeField = beforeFields.find(f => f.elemID.isEqual(afterField.elemID))
```
netama, line 48: can be 
```suggestion
      const beforeField = change.data.before.fields[afterField.name]
```
(but - then we should check `isField(beforeField)` and not just `!== undefined`, to avoid some edge cases with builtin functions)

packages/workspace/src/workspace/alias_index.ts
```diff
@@ -0,0 +1,131 @@
...
+      return (beforeField !== undefined)
+        && afterField.annotations[CORE_ANNOTATIONS.ALIAS] !== beforeField.annotations[CORE_ANNOTATIONS.ALIAS]
```
netama, line 50: (will need to change if we switch to arrays)

packages/workspace/src/workspace/alias_index.ts
```diff
@@ -0,0 +1,131 @@
...
+    const fieldsNames = isObjectType(element)
+      ? getFieldsElemIDsFullName(element)
+      : []
+    return [element.elemID.getFullName(), ...fieldsNames]
+  }
+  const [additions, removals] = _.partition(changes.flatMap(change => {
+    if (isModificationChange(change)) {
+      if (isObjectTypeChange(change)) {
+        const changedFields = getChangedFields(change)
+        const objAliasModification = change.data.before.annotations[CORE_ANNOTATIONS.ALIAS]
```
netama, line 74: I think we might need to be a little more accurate here and check the "final" annotation including the fallback to the prettified id?

packages/workspace/src/workspace/alias_index.ts
```diff
@@ -0,0 +1,131 @@
...
+        change,
+      ]
+    }
+    return change
+  }).filter(change => !isModificationChange(change)),
+  isAdditionChange)
+  const elemById = _.keyBy(additions.map(getChangeData), elem => elem.elemID.getFullName())
+  const additionsNames = Object.keys(elemById)
+
+  await index.setAll(additionsNames.map(name => ({
```
netama, line 98: I think for clarity we should distinguish better between `id` and `name` - here these are elem ids

packages/workspace/src/workspace/alias_index.ts
```diff
@@ -0,0 +1,131 @@
...
+    key: name,
+    value: elemById[name].annotations[CORE_ANNOTATIONS.ALIAS] ?? prettifyName(name.split('.').slice(-1)[0]),
```
netama, line 100: 1. continuing from above - I think we can get here with just the name part of the elem id, and then we won't need to `.split().slice()`
2. continuing from another comment - I think we should probably add a helper function that does the "alias or prettify" logic, since we should use it for determining if the alias changed in each of the changes as well (e.g. if we added a new alias annotation on an existing field that didn't have one until now - maybe it is still identical)

packages/workspace/src/workspace/alias_index.ts
```diff
@@ -0,0 +1,131 @@
...
+      const addedFields = Object.values(change.data.after.fields)
+      return [
+        ...addedFields.map(f => toChange({ after: f })),
+        change,
+      ]
+    }
+    return change
+  }).filter(change => !isModificationChange(change)),
+  isAdditionChange)
+  const elemById = _.keyBy(additions.map(getChangeData), elem => elem.elemID.getFullName())
```
netama, line 95: seems we only use the keys here for setting the index - can we just use the elem id `.name` part instead, to avoid the string manipulation later?

netama (commented):
looks good! added some comments, but I think if we change this to a string array some of the checks will change so will need another pass.
also let's wait for @ori-moisis / @adamro to take a look as well

packages/workspace/src/workspace/alias_index.ts
```diff
@@ -0,0 +1,131 @@
...
+    ...aliasModificationFields.map(f => toChange({ after: f })),
+    ...removalFields.map(f => toChange({ before: f })),
+  ]
+}
+
+const updateChanges = async (
+  changes: Change<Element>[],
+  index: RemoteMap<string>
+): Promise<void> => {
+  const getRelevantNamesFromChange = (change: Change<Element>): string[] => {
```
adamro, line 63: nit: Why's this function inside this scope?

packages/workspace/src/workspace/alias_index.ts
edenhassid: @netama but i think we can have duplicates in that way, if we have two instances of different types but same name

packages/workspace/src/workspace/workspace.ts
```diff
@@ -257,6 +258,7 @@ export type Workspace = {
     filePaths?: Set<string>,
     envName?: string
   ): Promise<Record<string, string>>
+  getAliasByElemId(envName?: string): Promise<Record<string, string>>
```
adamro, line 261: I would go with `getAliases`

ori-moisis: I don't have a strong opinion on the name as there are already examples that would be consistent with both of your suggestions...

I do think we may want to allow ourselves to option to read only a part of the index and not the whole thing at once, so either expose the remote map as @adamro suggests or add an optional input of element ids maybe? (raising this just because of the optional `envName` parameter which would make adding the option to get specific ids later look really bad...)

adamro (approved):
🔥 

Pending final decision on the api but approving to save a cycle

packages/adapter-api/src/builtins.ts
```diff
@@ -191,6 +191,7 @@ export const InstanceAnnotationTypes: TypeMap = {
   [INSTANCE_ANNOTATIONS.CREATED_AT]: BuiltinTypes.HIDDEN_STRING,
   [INSTANCE_ANNOTATIONS.CHANGED_BY]: BuiltinTypes.HIDDEN_STRING,
   [INSTANCE_ANNOTATIONS.CHANGED_AT]: BuiltinTypes.HIDDEN_STRING,
+  [INSTANCE_ANNOTATIONS.ALIAS]: StandardBuiltinTypes.STRING,
```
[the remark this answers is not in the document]
netama: @edenhassid let's add the annotation in `syntax.md`? (can also be in a follow-up PR)

packages/workspace/src/workspace/alias_index.ts
```diff
@@ -0,0 +1,144 @@
...
+  const afterFields = Object.values(change.data.after.fields)
+  const beforeFields = Object.values(change.data.before.fields)
+  const afterFieldsNamesSet = new Set(afterFields.map(field => field.name))
+  const beforeFieldsNamesSet = new Set(beforeFields.map(field => field.name))
+
+  const addedFields = afterFields.filter(afterField => !beforeFieldsNamesSet.has(afterField.name))
+  const removedFields = beforeFields.filter(beforeField => !afterFieldsNamesSet.has(beforeField.name))
+  const aliasModifiedFields = afterFields.filter(afterField => {
+    const field = change.data.before.fields[afterField.name]
+    return (isField(field) && field !== undefined) && (calcAlias(afterField) !== calcAlias(field))
```
netama, line 51: nit: I think `isField` also guarantees `field !== undefined`, no?

packages/workspace/src/workspace/alias_index.ts
```diff
@@ -0,0 +1,144 @@
...
+  })
+  return (addedFields.map(field => toChange({ after: field })) ?? [])
+    .concat(aliasModifiedFields.map(field => toChange({ after: field })) ?? [])
+    .concat(removedFields.map(field => toChange({ before: field })) ?? [])
+}
+
+const getRelevantNamesFromChange = (change: Change<Element>): string[] => {
```
netama, line 58: looks like this is a re-implementation of the function from `nacl_file_source.ts` - can we move this to a shared location (e.g. `adapter-utils`) and avoid the code duplication?

packages/workspace/src/workspace/workspace.ts
netama: :+1: I think we can just use named parameters to unblock this PR and add this separately when needed?

netama (approved):
looks great :) added a few more comments but nothing major so pre-approving

packages/adapter-utils/src/compare.ts
```diff
@@ -375,3 +375,11 @@ const sortChanges = (a: Change, b: Change): number =>
 
 export const calculateChangesHash = (changes: ReadonlyArray<Change>): string =>
   hashUtils.toMD5(safeJsonStringify(Array.from(changes).sort(sortChanges)))
+
+export const getRelevantNamesFromChange = (change: Change<Element>): string[] => {
```
netama, line 379: not a blocker, but I'm still not sure I understand the separation between this function and `getFieldsElemIDsFullName` and why one is here and the other on the object type. since this came out of your discussion with @adamro , leaving this up to you to finalize (and just to clarify, I think we can merge it this way as there are different opinions here and it's not critical)

packages/adapter-utils/src/nacl_case_utils.ts
```diff
@@ -161,3 +165,47 @@ export const invertNaclCase = (name: string): string => {
       ? specialCharsMappingList[0]
       : specialCharsMappingList.shift() ?? ''))
 }
+
+const prettifyWord = (str: string): string[] => {
```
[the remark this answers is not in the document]
netama: @edenhassid probably still worth asking @guykr to review this, but I think it's not a blocker either :)

netama (approved):

PR #4014, opened 2023-02-23
SALTO-3148 throw on SDF deploy error in other languages

dantal4:
In case that SDF deploy failed on an non-english NS account, no error is thrown from SDF so we need to look for the error format of `***<error in any language>***` and throw an exception.

---

---
_Release Notes_: 
Netsuite Adapter:
- Throw on SDF deploy error in other languages

---
_User Notifications_: 
None

### DIFF
--- a/packages/netsuite-adapter/src/client/sdf_client.ts
+++ b/packages/netsuite-adapter/src/client/sdf_client.ts
@@ -14,7 +14,7 @@
 * limitations under the License.
 */
 import { collections, decorators, objects as lowerdashObjects, promises, values as valuesUtils, strings } from '@salto-io/lowerdash'
-import { Values, AccountId, Value } from '@salto-io/adapter-api'
+import { Values, AccountId } from '@salto-io/adapter-api'
 import { mkdirp, readDir, readFile, writeFile, rm, rename } from '@salto-io/file'
 import { logger } from '@salto-io/logging'
 import {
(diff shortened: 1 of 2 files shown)


packages/netsuite-adapter/src/client/sdf_client.ts
```diff
@@ -338,7 +345,10 @@ export default class SdfClient {
       log.error(`SDF command ${commandName} has failed.`)
       throw Error(ActionResultUtils.getErrorMessagesString(actionResult))
     }
-    if (commandName === COMMANDS.DEPLOY_PROJECT) {
+    if ([
+      COMMANDS.DEPLOY_PROJECT,
+      COMMANDS.VALIDATE_PROJECT,
```
omrilit, line 350: Are you sure about this one?
I don't see any problem, just wondering whether there was a reason for not doing it before

omrilit (approved):
Looks good!
@tomermevorach @dantal4 do you think we could wrap the thrown message with some disclaimer about our partial support for non-english partial deploys?
Do you think we should map all inner messages for other languages or at least French where we have such customer?

Approving since as a first step this is good enough.

packages/netsuite-adapter/src/client/sdf_client.ts
dantal4: `verifySuccessfulDeploy` was used until now for extacting errors in enabled features deploy, which wasn't relevant for the validation command.

PR #4016, opened 2023-02-24
SALTO-3702 - Added maxLogTags config option to logger

amitgoldarb:
Added maxLogTags config to logger

---

There have been a few documented cases of crashes due to a large amount of logTags being logged.

---
_Release Notes_: 
Stability fix to our logger.

---
_User Notifications_: 
None.

### DIFF
--- a/packages/logging/src/internal/log-tags.ts
+++ b/packages/logging/src/internal/log-tags.ts
@@ -45,10 +45,10 @@ const formatKeyValue = (key: string, value: string): string =>
 
 const formatObjectLogTag = (obj: Record<string, unknown>): string =>
   Object.entries(obj)
-    .map(([key, value]) => formatKeyValue(key, safeStringify(value)))
+    .map(([key, value]) => formatKeyValue(key, typeof value === 'string' ? value : safeStringify(value)))
     .join(' ')
 
-export const formatLogTags = (logTags: Record<string, unknown>, baseKeys: string[]): string => {
+export const formatTextFormatLogTags = (logTags: Record<string, unknown>, baseKeys: string[]): string => {
   const tagsWithoutBaseKeys = _.omit(logTags, ...baseKeys)
   return Object.entries(tagsWithoutBaseKeys)
     .map(([logTagKey, logTagValue]) => {
--- a/packages/logging/src/internal/pino.ts
+++ b/packages/logging/src/internal/pino.ts
@@ -34,7 +34,7 @@ import {
 import { BaseLoggerRepo, BaseLoggerMaker } from './logger'
 import { LogLevel, toHexColor as levelToHexColor } from './level'
 import { Config } from './config'
-import { LogTags, formatLogTags, LOG_TAGS_COLOR, mergeLogTags, isPrimitiveType, PrimitiveType, normalizeLogTags } from './log-tags'
+import { LogTags, formatTextFormatLogTags, LOG_TAGS_COLOR, mergeLogTags, isPrimitiveType, PrimitiveType, normalizeLogTags } from './log-tags'
 
 const toPinoLogLevel = (level: LogLevel | 'none'): LevelWithSilent => (
   level === 'none' ? 'silent' : level
@@ -49,6 +49,7 @@ type FormatterBaseInput = {
   name: string
   stack?: string
   excessArgs: unknown[]
+  logTags?: LogTags
   type?: 'Error'
 }
 
@@ -57,6 +58,8 @@ type FormatterInput = FormatterBaseInput & Record<string, unknown>
 const formatterBaseKeys: (keyof FormatterBaseInput)[] = [
   'level', MESSAGE_KEY, 'time', 'name', 'stack', 'type', 'excessArgs',
 ]
+// Removing 2 for 'stack' which in most cases isn't part of the tags and excessArgs
+const COUNT_DEFAULT_TAGS = formatterBaseKeys.length - 2
 const excessDefaultPinoKeys = ['hostname', 'pid']
 
 type Formatter = (input: FormatterInput) => string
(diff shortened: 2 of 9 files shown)


packages/logging/src/internal/log-tags.ts
```diff
@@ -45,10 +45,10 @@ const formatKeyValue = (key: string, value: string): string =>
 
 const formatObjectLogTag = (obj: Record<string, unknown>): string =>
   Object.entries(obj)
-    .map(([key, value]) => formatKeyValue(key, safeStringify(value)))
+    .map(([key, value]) => formatKeyValue(key, typeof value === 'string' ? value : safeStringify(value)))
```
amitgoldarb, line 48: This is already being done earlier in the flow.

packages/logging/src/internal/pino.ts
```diff
@@ -57,6 +58,8 @@ type FormatterInput = FormatterBaseInput & Record<string, unknown>
 const formatterBaseKeys: (keyof FormatterBaseInput)[] = [
   'level', MESSAGE_KEY, 'time', 'name', 'stack', 'type', 'excessArgs',
 ]
+// Removing 2 for 'stack' which in most cases isn't part of the tags and excessArgs
```
amitgoldarb, line 61: I didn't check if the logged message was an error, just always subtracted 2, didn't think it was important enough to add more logic there, I could fix if reviewer thinks otherwise.

packages/logging/src/internal/pino.ts
```diff
@@ -79,7 +82,25 @@ const formatPrimitiveExcessArg = (
   [formatArgumentKey(i)]: value,
 })
 
-const formatExcessArgs = (excessArgs?: FormatterBaseInput['excessArgs']): LogTags => {
+const formatExcessTagsPerMessage = (
+  maxTagsPerLogMessage: number,
+  logTags?: LogTags,
+  excessArgs?: LogTags,
+): LogTags => {
+  // We slice first the excess args and then the logTags
```
amitgoldarb, line 90: I prioritiezed logTags above excess args, I think that in most cases this is what the user will want.

packages/logging/src/internal/pino.ts
```diff
@@ -79,7 +82,25 @@ const formatPrimitiveExcessArg = (
...
+  const entriesObject = [...Object.entries(excessArgs ?? {}), ...Object.entries(logTags ?? {})]
+  const countCurrentTags = entriesObject.length + COUNT_DEFAULT_TAGS
+  if (countCurrentTags > maxTagsPerLogMessage) {
+    return {
+      ...Object.fromEntries(entriesObject.slice(countCurrentTags - maxTagsPerLogMessage)),
+      countRemovedTags: countCurrentTags - maxTagsPerLogMessage,
```
amitgoldarb, line 96: I'm willing to change the name if someone can think of a better option :)

packages/logging/src/internal/pino.ts
```diff
@@ -200,24 +221,19 @@ export const loggerRepo = (
     } : false,
     formatters: {
       level: (level: string) => ({ level: level.toLowerCase() }),
-      log: (obj: object) => {
-        // When config is text leave the formatting for prettifier.
-        if (config.format === 'json') return formatJsonLog(obj as FormatterBaseInput)
```
amitgoldarb, line 205: I moved this logic to be higher up the in the logging flow so that it can be shared with text format and also so that it could have the context of the logTags when doing the formatting.

packages/logging/src/internal/pino.ts
dorsev, line 61: Nit:  ( and remove the comment?)

```
const errorKeys = ['message','stack']
const COUNT_DEFAULT_TAGS = formatterBaseKeys.length - errorKeys.length
```

packages/logging/src/internal/pino.ts
```diff
@@ -79,7 +82,25 @@ const formatPrimitiveExcessArg = (
...
+  maxTagsPerLogMessage: number,
+  logTags?: LogTags,
+  excessArgs?: LogTags,
+): LogTags => {
+  // We slice first the excess args and then the logTags
+  const entriesObject = [...Object.entries(excessArgs ?? {}), ...Object.entries(logTags ?? {})]
+  const countCurrentTags = entriesObject.length + COUNT_DEFAULT_TAGS
+  if (countCurrentTags > maxTagsPerLogMessage) {
+    return {
+      ...Object.fromEntries(entriesObject.slice(countCurrentTags - maxTagsPerLogMessage)),
```
dorsev, line 95: Nit: extract `countCurrentTags - maxTagsPerLogMessage` to a const.
Nit2: are all the `...` here safe? should we use `array.concat`? 
Nit3: I'm not sure the if else is needed in this code, aka I think it could be simplified. something like ( `combine logTags + excessArgs, and take just maxTagsPerLogMessage` from them?

if it's too big, it will slice, and if it's too small, it will do nothing.. no need for if else ( I think)

packages/logging/src/internal/pino.ts
```diff
@@ -230,53 +246,46 @@ export const loggerRepo = (
...
     }
   }
 
-  const logJson = (
+  const logMessage = (
     pinoLogger: pino.Logger,
     level: LogLevel,
     unconsumedArgs: unknown[],
-    message: string
+    logTags: LogTags,
```
dorsev, line 284: Q\Nit: I see that we pass the log tags to a lot of functions in this file... is there a way to pass it just once to an upper scope, and reuse across the file? I think it could make the signatures a bit easier to read.

packages/logging/test/internal/pino_logger.test.ts
```diff
@@ -1098,6 +1133,32 @@ describe('pino based logger', () => {
           expect(nonChunkedLog.logId).toBeUndefined()
         })
       })
+      describe('when there are too many tags', () => {
+        let lines: string[]
+        beforeEach(() => {
+          initialConfig.maxTagsPerLogMessage = 8
+          logger = createLogger()
+          logger.assignTags(logTags)
+          logger.log('warn', 'some log', { one: 1, two: 2, three: 3 })
```
dorsev, line 1142: Q: should we add a test with 100000 tags maybe? just to make our point and verify that this issue really is resolved?

dorsev (commented):
LGTM. Thank you for this PR!

Added some comments & nits inside. LMK your thoughts

ori-moisis:
@amitgoldarb is this still relevant?
can we close it?

amitgoldarb:
@ori-moisis This caused a few crashes. I should probably merge it.. I'll try and circle back to it next week.

packages/logging/src/internal/pino.ts
amitgoldarb: 1. Done 
2. Done
3. I find this readable even a few months after. Not sure a simplification is needed.

packages/logging/src/internal/pino.ts
amitgoldarb: These are a lot of static functions. How?

packages/logging/test/internal/pino_logger.test.ts
amitgoldarb: I tested this manually, I'm concerned just creating such a large object in a test will be very slow. I think this is sufficient.

packages/logging/src/internal/pino.ts
dorsev: been a few months already and I'm not sure what I meant 👍

dorsev (approved):
LGTM

PR #4018, opened 2023-02-26
SAAS-7316: Add validate capability to the dummy-adapter

dorsev:
Add `validate` capability to the dummy-adapter

I'm not 100% sure I'm doing this right, or if this is the right approach, so this is WIP.

---

None

---

_Release Notes_: 
None

---

_User Notifications_: 
None


tamtamirr (approved):
LGTM

PR #4019, opened 2023-02-26
SALTO-3671/un-revert fixed import error

guyhamo:
this pr was merged and reverted after failed deployment, I found the problem to be a bad import in system note.

---

---
_Release Notes_: 
netsuite adpater:
* ignore changed at values from the future.

---
_User Notifications_:


Shulik95 (approved):
💯

PR #4021, opened 2023-02-26
SALTO-3708: zendesk typeError cannot read property toString of undefined

edenhassid:
_zendesk typeError cannot read property toString of undefined_

---

---
_Release Notes_: 
zendesk:
* fix error of cannot read property toString of undefined in view filter

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/filters/view.ts
+++ b/packages/zendesk-adapter/src/filters/view.ts
@@ -24,7 +24,7 @@ import { applyforInstanceChangesOfType } from './utils'
 
 export const VIEW_TYPE_NAME = 'view'
 
-const valToString = (val: Value): string | string[] => (_.isArray(val) ? val.map(String) : val.toString())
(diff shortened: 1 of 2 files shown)


packages/zendesk-adapter/src/filters/view.ts
```diff
@@ -24,7 +24,7 @@ import { applyforInstanceChangesOfType } from './utils'
 
 export const VIEW_TYPE_NAME = 'view'
 
-const valToString = (val: Value): string | string[] => (_.isArray(val) ? val.map(String) : val.toString())
+const valToString = (val: Value): string | string[] => (_.isArray(val) ? val.map(String) : val?.toString())
```
netama, line 27: let's also change the return value
```suggestion
const valToString = (val: Value): string | string[] | undefined => (_.isArray(val) ? val.map(String) : val?.toString())
```

netama (approved):
LGTM with one nit

PR #4022, opened 2023-02-26
salto-2883 netsuite partial fetch

kami-delaroz:
Changes in Netsuite partial fetch:
Quick fetch false by default
Don't allow partial fetch for first fetch

---

_None_

---
_Release Notes_: 
Netsuite Fetches will not be "quick" fetches by default, but only if quick fetch was requested.
Partial fetch as first fetch is not supported

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/netsuite-adapter/src/adapter.ts
+++ b/packages/netsuite-adapter/src/adapter.ts
@@ -345,6 +345,12 @@ export default class NetsuiteAdapter implements AdapterOperations {
    */
 
   public async fetch({ progressReporter, withChangesDetection = false }: FetchOptions): Promise<FetchResult> {
+    const isFirstFetch = async () : Promise<boolean> =>
+      !(await awu(await this.elementsSource.list()).find(e => !e.isConfig()))
+    if (this.fetchTarget !== undefined && await isFirstFetch()) {
+      throw new Error('Can\'t define fetchTarget for the first fetch. Remove fetchTarget from adapter config file')
+    }
+
     const explicitIncludeTypeList = [REPORT_DEFINITION, FINANCIAL_LAYOUT]
       .filter(typeName => !this.fetchInclude?.types.some(type => type.name === typeName))
     const deprecatedSkipList = buildNetsuiteQuery(convertToQueryParams({
@@ -360,7 +366,7 @@ export default class NetsuiteAdapter implements AdapterOperations {
       notQuery(deprecatedSkipList),
     ].filter(values.isDefined).reduce(andQuery)
 
-    const isPartial = withChangesDetection || this.fetchTarget !== undefined
+    const isPartial = (withChangesDetection || this.fetchTarget !== undefined) && !(await isFirstFetch())
 
     const {
       failedToFetchAllAtOnce,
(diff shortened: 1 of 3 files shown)


packages/netsuite-adapter/src/adapter.ts
```diff
@@ -353,6 +353,13 @@ export default class NetsuiteAdapter implements AdapterOperations {
       isPartial: boolean
     }
   ): Promise<FetchResult> {
+    const isFirstFetch = await awu(await this.elementsSource.list())
+      .filter(e => !e.isConfig())
+      .isEmpty()
```
omrilit, line 358: performance nit: I wonder whether we can identify it in a more efficient way, for example after encountering the first element which is not a Config, and not iterating over all elements.
If you think that it's negligible performance-wise, you can ignore it

packages/netsuite-adapter/src/config.ts
```diff
@@ -36,7 +36,7 @@ export const DEFAULT_MAX_ITEMS_IN_IMPORT_OBJECTS_REQUEST = 40
 export const DEFAULT_DEPLOY_REFERENCED_ELEMENTS = false
 export const DEFAULT_WARN_STALE_DATA = false
 export const DEFAULT_VALIDATE = true
-export const DEFAULT_USE_CHANGES_DETECTION = true
+export const DEFAULT_USE_CHANGES_DETECTION = false
```
omrilit, line 39: Not sure that I understand this part.
IIUC if we only merge this part, it means that if the user will specify only fetchTargets then we won't use useChangesDetection which is not really what we want.
IIUC we'd like that if the user specifies fetchTargets we'll use useChangesDetection unless it was explicitly set to false.
In addition, if useChangesDetection is set to true without setting fetchTargets -> we'll treat it as if fetchTargets is set with ".*" for all sub queries.
Am I right here?

omrilit (commented):
Looks good!

omrilit:
Missing UT for "Don't allow partial fetch for first fetch"

packages/netsuite-adapter/src/adapter.ts
kami-delaroz: yes, I guess I could use "find"
anyway the code is not final yet, I should have put it as "draft", sorry.
we still have to discuss whether this is the right place and behavior, and then of curse I'll add a UT

packages/netsuite-adapter/src/adapter.ts
```diff
@@ -353,6 +353,12 @@ export default class NetsuiteAdapter implements AdapterOperations {
       isPartial: boolean
     }
   ): Promise<FetchResult> {
+    const isFirstFetch = !(await awu(await this.elementsSource.list())
+      .find(e => !e.isConfig()))
+    if (isFirstFetch && isPartial) {
```
omrilit, line 358: Maybe we can start with checking `isPartial` and only if it returns true -> calculate `isFirstFetch`

omrilit (approved):
Looks good!
I think that only the "Partial fetch as first fetch is not supported" is relevant in the release notes, right?

PR #4025, opened 2023-02-26
change validation messages for SDF elements removal

kami-delaroz:
Change the message of SDF elements removal change validators

---

_None_

---
_Release Notes_: 
change-validator messages in case of a failure in removing an SDF element are different

---
_User Notifications_: 
_None._


omrilit (approved):
Thanks for that!

PR #4026, opened 2023-02-26
SALTO-3688: Add fetch warning if talking to an old version of Jira DC Plugin

IdoZakk:
relevant PR in the plugin:  https://github.com/salto-io/jira-dc-app/pull/95

---
_Release Notes_: 
Jira adapter:
Warn in DC if the plugin and OSS versions do no fit

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/data_center/plugin_version.ts
+++ b/packages/jira-adapter/src/filters/data_center/plugin_version.ts
@@ -14,6 +14,7 @@
 * limitations under the License.
 */
 import Joi from 'joi'
+import semver from 'semver'
 import { logger } from '@salto-io/logging'
 import { createSchemeGuard } from '@salto-io/adapter-utils'
 import { FilterCreator } from '../../filter'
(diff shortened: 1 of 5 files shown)


packages/jira-adapter/src/filters/data_center/plugin_version.ts
```diff
@@ -0,0 +1,100 @@
...
+*/
+import Joi from 'joi'
+import { logger } from '@salto-io/logging'
+import { createSchemeGuard } from '@salto-io/adapter-utils'
+import { FilterCreator } from '../../filter'
+
+const log = logger(module)
+export const PLUGIN_VERSION_NUMBER = '1.0.4'
+
+const versionNumberCompare = (version1: string, version2: string): number => {
```
alonstern, line 24: I think we can use a library for that. E.g. https://stackoverflow.com/questions/6832596/how-can-i-compare-software-version-number-using-javascript-only-numbers `semver` which we also use in some other packages

packages/jira-adapter/src/filters/data_center/plugin_version.ts
```diff
@@ -0,0 +1,100 @@
...
+  return 0
+}
+
+type InfoResponse = {
+  version: string
+}
+
+const INFO_RESPONSE_SCHEME = Joi.object({
+  version: Joi.string().required(),
+}).required()
```
alonstern, line 44: add `unknown(true)` so if we add more things it the future this won't break

packages/jira-adapter/src/filters/data_center/plugin_version.ts
```diff
@@ -0,0 +1,100 @@
...
+ */
+const filter: FilterCreator = ({ client }) => ({
+  name: 'jiraDcPluginVerionNumberFilter',
+  onFetch: async () => {
+    if (!client.isDataCenter) {
+      return undefined
+    }
+    try {
+      const response = await client.getSinglePage({
+        url: '/rest/api/3/plugininfo',
```
alonstern, line 59: nit: I think we can use here the real URL `/rest/salto/1.0/plugininfo` since this does not have a `rest/api/3/` version

packages/jira-adapter/src/filters/data_center/plugin_version.ts
```diff
@@ -0,0 +1,100 @@
...
+              severity: 'Info',
+            },
+          ],
+        }
+      }
+      if (compareResult < 0) {
+        return {
+          errors: [
+            {
+              message: `Your Jira instance is running an old version ${response.data.version} of Salto Configuration Manager for Jira Data Center. Please update the app by setting up automatic updates, or visit https://marketplace.atlassian.com/apps/1225356/salto-configuration-manager-for-jira to download the latest version.`,
```
alonstern, line 79: `Please update the app by setting up automatic updates` do we know how to do that?

IdoZakk: Yes, https://confluence.atlassian.com/upm/updating-apps-273875710.html#:~:text=.-,Enabling%20automatic%20updates,-Administrators%20can%20choose

alonstern (approved):
🔥

PR #4027, opened 2023-02-27
skip convert-lists for file-cabinet instances

dantal4:
file&folder instances have no list fields so we can skip them

---

---
_Release Notes_: 
Netsuite Adapter:
- Skip convert-lists for file-cabinet instances

---
_User Notifications_: 
None


omrilit (approved):
Thanks for that!

PR #4029, opened 2023-02-27
SALTO-3662: add zendesk alias

edenhassid:
add zendesk alias

---

---
_Release Notes_: 
zendesk:
* add alias to zendesk elements

---
_User Notifications_: 
None


packages/zendesk-adapter/src/filters/add_alias.ts
```diff
@@ -0,0 +1,279 @@
...
+const STRING_SIGN = '$'
+const CATEGORY_ORDER = `${STRING_SIGN}Category Order${STRING_SIGN}`
+const SECTION_ORDER = `${STRING_SIGN}Section Order${STRING_SIGN}`
+const ARTICLE_ORDER = `${STRING_SIGN}Article Order${STRING_SIGN}`
+const LANGUAGE_SETTINGS = `${STRING_SIGN}language settings${STRING_SIGN}`
+const SETTINGS = `${STRING_SIGN}Settings${STRING_SIGN}`
+
+
+type AliasData = {
+  aliasFields: string[]
```
netama, line 39: (summarizing a suggestion here from IRL and comments below)
wdyt about something like this type instead of a string?
```
type AliasComponent = {
  fieldName: string // can be optional as well?
  referencedFieldName?: string // when available, we access this field in the resolved reference
  prefix?: string
  suffix?: string
}
```
(can be instead of the separator or on top of it - we can probably decide when we see more cases)

packages/zendesk-adapter/src/filters/add_alias.ts
```diff
@@ -0,0 +1,279 @@
...
+
+const calculateAlias = (
+  instance: InstanceElement, instById: Record<string, InstanceElement>,
+): string | undefined => {
+  const currentType = instance.elemID.typeName
+  const { aliasFields } = aliasMap[currentType]
+  const separator = aliasMap[currentType].separator ?? ' '
+  const aliasParts = aliasFields
+    .map(field => {
+      if (field.startsWith(STRING_SIGN) && field.endsWith(STRING_SIGN)) {
```
netama, line 219: will be replaced by the `prefix` / `suffix`

packages/zendesk-adapter/src/filters/add_alias.ts
```diff
@@ -0,0 +1,279 @@
...
+        return field.slice(1, -1)
+      }
+      if (field.startsWith(CORE_ANNOTATIONS.PARENT)) {
```
netama, line 222: nit (since not strictly needed now): can we generalize this treatment to look for instance annotations and not specifically parent?

packages/zendesk-adapter/src/filters/add_alias.ts
```diff
@@ -0,0 +1,279 @@
...
+        const parentRoute = route.slice(2).join('.')
+        if (parentRoute.startsWith(CORE_ANNOTATIONS.ALIAS)) {
+          return parentInstance.annotations[CORE_ANNOTATIONS.ALIAS]
+        }
+        return _.get(parentInstance.value, parentRoute)
+      }
+      return _.get(instance.value, field)
+    })
+  if (!aliasParts.every(isDefined)) {
+    return undefined
```
netama, line 244: let's log what happened (maybe even which one was missing) for debugging purposes?

netama (commented):
very cool :) added some comments, mostly around defining the structure more explicitly

netama (commented):
looks great! added some nits. I think the main remaining part is finalizing the config

packages/zendesk-adapter/src/filters/add_alias.ts
edenhassid: @netama instead of `isDefined` I added the function `isValidAlias`. a log is added there :)

packages/zendesk-adapter/src/filters/add_alias.ts
```diff
@@ -0,0 +1,279 @@
...
+  routing_attribute_value: {
+    aliasFields: ['name'],
+  },
+  sla_policy: {
+    aliasFields: ['title'],
+  },
+  support_address: {
+    aliasFields: ['name'],
+  },
+  tag: { // should we add alias? there is only id
```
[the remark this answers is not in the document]
tomermevorach: I tend to prefer consistency in this case, it makes things more simple.

Also, I understand from @edenhassid that these aliases are visible in the nacl, I thought it supposed to be hidden.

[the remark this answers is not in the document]
netama: they will be visible. so the decision is to create the alias?

packages/zendesk-adapter/src/filters/add_alias.ts
```diff
@@ -0,0 +1,279 @@
...
+  ticket_field__custom_field_options: {
+    aliasFields: ['name'],
+  },
+  ticket_form: {
+    aliasFields: ['name'],
+  },
+  trigger: {
+    aliasFields: ['title'],
+  },
+  trigger_category: { // should we add alias? there is only name
```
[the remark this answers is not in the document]
netama: just making sure - we'll remove these comments with questions before merging, right? (if there are leftovers - let's create a ticket with the remaining decisions and link to it from the comments)

packages/zendesk-adapter/src/filters/add_alias.ts
```diff
@@ -0,0 +1,500 @@
...
+  if (!isValidAlias(aliasParts, instance)) {
+    return undefined
+  }
+  return aliasParts.join(separator)
+}
+
+const filterCreator: FilterCreator = ({ config }) => ({
+  name: 'addAlias',
+  onFetch: async (elements: Element[]): Promise<void> => {
+    if (config[FETCH_CONFIG].addAlias === false) {
```
netama, line 469: just making sure - by default we will reach this case (temporarily until we finish testing), right?

packages/zendesk-adapter/test/filters/add_alias.test.ts
```diff
@@ -0,0 +1,177 @@
...
+            brands: ['.*'],
+          },
+          addAlias: true,
+        },
+      },
+    })) as FilterType
+  })
+
+  describe('onFetch', () => {
+    it('should add alias annotation correctly', async () => {
```
netama, line 81: let's add some test cases for "failure" modes? (not adding when one of the values is undefined / not crashing when the structure is not as expected / handling cases where a value that was expected to be a reference is not and vice versa)?

netama (approved):
code LGTM :) I think we can add some more test cases + remove some leftover comments before merging. but no need for another review (unless there's something from your side)

PR #4031, opened 2023-02-27
SALTO-3711 - Zendesk Adapter - Filter organization names from the logs

idoser96:
If resolveOrganizationIDs is false, we will filter the organization names before logging them, so we won't save them anywhere and we will keep the user's privacy requests

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/client/client.ts
+++ b/packages/zendesk-adapter/src/client/client.ts
@@ -30,7 +30,7 @@ const {
 } = clientUtils
 const log = logger(module)
 
-const ORG_ENDPOINTS_TO_FILTER = ['organizations/show_many', 'organizations/autocomplete']
+const OMIT_REPLACEMENT = '<OMITTED>'
 
 type FilterLogsConfig = {
   allowOrganizationNames?: boolean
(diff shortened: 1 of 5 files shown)


packages/zendesk-adapter/src/client/client.ts
```diff
@@ -29,6 +30,13 @@ const {
 } = clientUtils
 const log = logger(module)
 
+const ORG_ENDPOINTS_TO_FILTER = ['organizations/show_many', 'organizations/autocomplete']
+
+type FilterLogsConfig = {
```
netama, line 35: nit: maybe `LogFilterConfig`?

packages/zendesk-adapter/src/client/client.ts
```diff
@@ -143,4 +159,14 @@ export default class ZendeskClient extends clientUtils.AdapterHTTPClient<
       throw e
     }
   }
+
+  protected clearValuesFromResponseData(
+    responseData: Values,
+    url: string
+  ): Values {
+    if (this.filterLogsConfig.allowOrganizationNames !== true && ORG_ENDPOINTS_TO_FILTER.some(ep => url.includes(ep))) {
```
netama, line 167: nit: if it's boolean it's ok to check `!this.filterLogsConfig.allowOrganizationNames`

netama (approved):
LGTM with one remaining comment

PR #4032, opened 2023-02-28
SALTO-3522: Make AuthorizationServerPolicy and AuthorizationServerPolicyRule standalone fields

shir-reifenberg:
Make AuthorizationServerPolicy and AuthorizationServerPolicyRule standalone fields
---

Until now we fetched `AuthorizationServerPolicy` and `AuthorizationServerPolicyRule` as part of `AuthorizationServer` type. This PR makes them standalone fields as we split all other policy and policy rules type & we want to hide fields like `id`

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/config.ts
+++ b/packages/okta-adapter/src/config.ts
@@ -423,6 +423,7 @@ const DEFAULT_TYPE_CUSTOMIZATIONS: OktaApiConfig['types'] = {
       fieldsToOmit: DEFAULT_FIELDS_TO_OMIT.concat({ fieldName: '_links' }),
       fieldsToHide: [{ fieldName: 'id' }],
       serviceIdField: 'id',
+      standaloneFields: [{ fieldName: 'policies' }],
     },
   },
   AuthorizationServerPolicy: {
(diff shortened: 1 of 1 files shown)


packages/okta-adapter/src/config.ts
```diff
@@ -431,6 +432,16 @@ const DEFAULT_TYPE_CUSTOMIZATIONS: OktaApiConfig['types'] = {
...
       serviceIdField: 'id',
+      fieldsToHide: [{ fieldName: 'id' }],
+      fieldsToOmit: DEFAULT_FIELDS_TO_OMIT.concat({ fieldName: '_links' }),
+      standaloneFields: [{ fieldName: 'policyRules' }],
+    },
+  },
+  AuthorizationServerPolicyRule: {
+    transformation: {
+      serviceIdField: 'id',
+      fieldsToHide: [{ fieldName: 'id' }],
```
alonstern, line 443: Should we make these the default?

alonstern (approved):
👌

packages/okta-adapter/src/config.ts
shir-reifenberg: We can't yet cause there places where we have list of nested types with `id` as well (so hiding this is not supported)
Once I'll extract those to standalone fields as well I will change it

PR #4034, opened 2023-02-28
SALTO-3725 improve parse-report-types filter

dantal4:
Improve parse-report-types filter by not parsing the current `definition` from the nacl, but taking the already parsed values.
This improves the run time by 50%.

---

---
_Release Notes_: 
Netsuite Adapter:
- Improve parse-report-types filter

---
_User Notifications_: 
None


omrilit (approved):
Great initiative!

PR #4036, opened 2023-03-01
SALTO-3591: Support status changes (Okta)

shir-reifenberg:
To change instance status in Okta, we must make a request to a different endpoint.

---

This PR adds general support in status changes in Okta. 
As this is relevant to many types (`GroupRule`, `Application`, `Policy`...), I added it as part of the default deploy change in Okta, currently for `ModificaionChange` only.
For now the relevant types we already deploy are `GroupRule` and `Application`, we will add more types to the config once we support deployment of other types.

Status change will happen before of after other values changes, depends on if it's an activation(INACTIVE --> ACTIVE) or deactivation (ACTIVE --> INACTIVE). That's because for some types, we can't make changes when the status is 'ACTIVE'.

Also added unit tests for `deployment.ts`
**Please review adapter-components changes here** https://github.com/salto-io/salto/pull/4048 🙏 

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/okta-adapter/src/deployment.ts
```diff
@@ -63,13 +66,50 @@ const getOktaError = (elemID: ElemID, error: Error): Error => {
...
+  }
+  const url = createUrl({
+    instance,
+    baseUrl: endpoint.url,
+    urlParamsToFields: endpoint.urlParamsToFields,
+  })
+  try {
+    await client[endpoint.method]({ url, data: {} })
+  } catch (err) {
+    log.error(`Status could not be updated in instance: ${getChangeData(change).elemID.getFullName()}`)
```
netama, line 101: if we're logging let's also log the error?

netama (approved):
🎉

PR #4039, opened 2023-03-01
SALTO-3715 Make CLI support withChangeDetection flag on fetch

omrifarkash:
---

---
_Release Notes_: 
CLI:
When passing -cd flag on fetch operation a 'withChangeDetection' will passed to core with the value 'true' default value is false
---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-api/src/adapter.ts
+++ b/packages/adapter-api/src/adapter.ts
@@ -57,7 +57,7 @@ export type ProgressReporter = {
 
 export type FetchOptions = {
   progressReporter: ProgressReporter
-  withChangesDetection?: boolean
+  withChangeDetection?: boolean
 }
 
 export type DeployOptions = {
--- a/packages/adapter-api/src/adapter.ts
+++ b/packages/adapter-api/src/adapter.ts
@@ -57,7 +57,7 @@ export type ProgressReporter = {
 
 export type FetchOptions = {
   progressReporter: ProgressReporter
-  withChangeDetection?: boolean
+  withChangesDetection?: boolean
 }
 
 export type DeployOptions = {
--- a/packages/cli/src/commands/fetch.ts
+++ b/packages/cli/src/commands/fetch.ts
@@ -60,6 +60,7 @@ export type FetchCommandArgs = {
   stateOnly: boolean
   accounts: string[]
   regenerateSaltoIds: boolean
+  withChangeDetection?: boolean
 }
 
 const createFetchFromWorkspaceCommand = (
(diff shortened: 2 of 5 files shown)


packages/adapter-api/src/adapter.ts
```diff
@@ -58,7 +58,7 @@ export type ProgressReporter = {
 
 export type FetchOptions = {
   progressReporter: ProgressReporter
-  withChangesDetection?: boolean
```
omrifarkash, line 61: Just aligning the terminology to be in singular change and not changes

packages/core/src/core/fetch.ts
```diff
@@ -445,7 +445,7 @@ const fetchAndProcessMergeErrors = async (
         .map(async ([accountName, adapter]) => {
           const fetchResult = await adapter.fetch({
             progressReporter: progressReporters[accountName],
-            withChangesDetection,
+            withChangeDetection,
```
omrifarkash, line 448: Just aligning the terminology to be in singular change and not changes

packages/core/src/core/fetch.ts
```diff
@@ -717,7 +717,7 @@ export const fetchChanges = async (
   accountToServiceNameMap: Record<string, string>,
   currentConfigs: InstanceElement[],
   progressEmitter?: EventEmitter<FetchProgressEvents>,
-  withChangesDetection?: boolean
+  withChangeDetection?: boolean
```
omrifarkash, line 720: Just aligning the terminology to be in singular change and not changes

packages/core/src/core/fetch.ts
```diff
@@ -732,7 +732,7 @@ export const fetchChanges = async (
     accountToServiceNameMap,
     getChangesEmitter,
     progressEmitter,
-    withChangesDetection
+    withChangeDetection
```
omrifarkash, line 735: Just aligning the terminology to be in singular change and not changes

packages/cli/src/commands/fetch.ts
```diff
@@ -378,6 +382,14 @@ const fetchDef = createWorkspaceCommand({
         type: 'boolean',
         default: false,
       },
+      {
+        name: 'withChangeDetection',
+        alias: 'cd',
+        required: false,
+        description: 'Detect changes before running full fetch',
```
adamro, line 389: @tomermevorach @rozenmanofer  Can you please review the text here please?

adamro (approved):
Need a PM eye on the text there but otherwise LGTM

packages/cli/src/commands/fetch.ts
omrifarkash: @tomermevorach  ?

tomermevorach: To be honest I am not in favor of the term 'with ChangeDetection', not sure why we decided not to use the term 'quick' instead.
As for the description, it needs to be also rephrase to something in the spirit of:
"Improve fetch performance by relying on the service audit trail in order to fetch only elements that were modified or created since the last fetch"

PR #4040, opened 2023-03-01
SALTO-3694 - Salesforce: Validate recipients of WorkflowAlerts

tomsellek:
Validate that if the recipients are Salesforce users, they actually exist

---

This required a slight modification to the change validator to allow it to extract multiple users from a single field.

---
_Release Notes_: 
Salesforce: Improved validation of WorkflowAlerts recipients

---
_User Notifications_: 
N/A


tomsellek:
image

amitgoldarb (approved):
LGTM
Thanks for this one!

PR #4041, opened 2023-03-02
SALTO-3620 add filter to remove unix time zero (1/1/1970) from created and modified dates

benisraeladi:
add filter to remove unix time zero (1/1/1970) from created and modified dates

---

This does not change the workspace, since the annotations are hidden. 

element print before:
image

element print after:
image

---
_Release Notes_: _None_
---
_User Notifications_: _None_

### DIFF
--- a/packages/salesforce-adapter/src/filters/remove_unix_time_zero.ts
+++ b/packages/salesforce-adapter/src/filters/remove_unix_time_zero.ts
@@ -13,7 +13,7 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
-import { Element, isElement } from '@salto-io/adapter-api'
+import { CORE_ANNOTATIONS, Element, isElement } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
 import { collections } from '@salto-io/lowerdash'
 import { LocalFilterCreator } from '../filter'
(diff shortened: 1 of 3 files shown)


ori-moisis:
why is this a separate filter and not a fix to the original logic that populates this value?

benisraeladi:
> why is this a separate filter and not a fix to the original logic that populates this value?

@ori-moisis This is not added by any filter, rather this comes directly from the fetch itself. For example, you can see from the following that Salesforce returns this when requesting the instance:
image

If you believe that it's more correct to do it at the fetch level and not at the filter level then I can try to do that, but given that this can technically also happen on custom object instances that we only fetch as part of filters I believe this is a more complete solution

ori-moisis:
so you are saying that adding a condition in [this code](https://github.com/salto-io/salto/blob/main/packages/salesforce-adapter/src/transformers/transformer.ts#L1493:L1493) to make sure we don't populate 1970 in the first place won't cover this?

benisraeladi:
In short, yes.
It's possible to add the logic both to the transformer and [here in this filter](https://github.com/salto-io/salto/blob/b03d71acc11642d0d92b4739ed96e31caa45656d/packages/salesforce-adapter/src/filters/author_information/data_instances.ts#L43) but it makes it so that these changes are done in multiple places because of the way we work with data instances. I'm partial to something that is all in one place, but open to discussion

packages/salesforce-adapter/src/filters/remove_unix_time_zero.ts
```diff
@@ -0,0 +1,59 @@
...
+  elements: Element[],
+): Promise<void> => {
+  await awu(elements)
+    .filter(isElement)
+    .forEach(async e => {
+      // eslint-disable-next-line no-underscore-dangle
+      if (e.annotations._changed_at === UNIX_TIME_ZERO_STRING) {
+        // eslint-disable-next-line no-underscore-dangle
+        delete e.annotations._changed_at
+        log.debug(`Removed unix time 0 last modified of ${await apiName(e)}`)
```
ori-moisis, line 38: might be better to have this as `trace` as this is O(number of instances in the account) which can be quite a lot.

also, logs support format strings which is (very slightly) more efficient, so
```suggestion
        log.trace('Removed unix time 0 last modified of %s', await apiName(e))
```

ori-moisis: @benisraeladi you changed the other log but missed this one

ori-moisis (approved):

PR #4043, opened 2023-03-02
SALTO-3724: zendesk revert nacl case from channel name

edenhassid:
zendesk revert nacl case from channel name

---

---
_Release Notes_: 
zendesk:
* in channel, name will no longer be in nacl case

---
_User Notifications_: 
zendesk:
* in channel, name will no longer be in nacl case


netama (approved):
LGTM, can we cover this in a test (can edit the existing instance)?

PR #4046, opened 2023-03-03
SALTO-3751 - Salesforce: Improve modelling of track feed history and enable feeds annotations

tomsellek:
Keep a list of fields whose history/feed-history is tracked at the object level instead of keeping Salesforce's per-field annotation.

---
Salesforce enables history tracking per field + an additional "master switch" per object. This causes a lot of noise when deploying objects between environments where history tracking is enabled to environments where it's disabled (basically every single field is marked as modified). We want to make it easier to determine what the actual change is, so we centralize the list of tracked field as an annotation in the object.

---
_Release Notes_: 
Salesforce Adapter:

Improved modelling of `History` and `Feed History Tracking` on Standard and Custom Objects.

---
_User Notifications_: 
Salesforce Adapter:

The annotation enableHistory will be removed from all custom objects
The annotation enableFeeds will be removed from all custom objects

### DIFF
--- a/packages/salesforce-adapter/src/adapter.ts
+++ b/packages/salesforce-adapter/src/adapter.ts
@@ -83,7 +83,7 @@ import formulaDepsFilter from './filters/formula_deps'
 import removeUnixTimeZeroFilter from './filters/remove_unix_time_zero'
 import organizationWideDefaults from './filters/organization_wide_sharing_defaults'
 import { FetchElements, FETCH_CONFIG, SalesforceConfig } from './types'
-import historyTrackingFilter from './filters/history_tracking'
+import centralizeTrackingInfoFilter from './filters/centralize_tracking_info'
 import { getConfigFromConfigChanges } from './config_change'
 import { LocalFilterCreator, Filter, FilterResult, RemoteFilterCreator, LocalFilterCreatorDefinition, RemoteFilterCreatorDefinition } from './filter'
 import { addDefaults } from './filters/utils'
@@ -155,8 +155,8 @@ export const allFilters: Array<LocalFilterCreatorDefinition | RemoteFilterCreato
   { creator: xmlAttributesFilter },
   { creator: minifyDeployFilter },
   { creator: formulaDepsFilter },
-  // historyTrackingFilter depends on customObjectsToObjectTypeFilter and must run before customTypeSplit
-  { creator: historyTrackingFilter },
+  // centralizeTrackingInfoFilter depends on customObjectsToObjectTypeFilter and must run before customTypeSplit
+  { creator: centralizeTrackingInfoFilter },
   // The following filters should remain last in order to make sure they fix all elements
   { creator: convertListsFilter },
   { creator: convertTypeFilter },
--- a/packages/salesforce-adapter/src/adapter.ts
+++ b/packages/salesforce-adapter/src/adapter.ts
@@ -83,6 +83,7 @@ import formulaDepsFilter from './filters/formula_deps'
 import removeUnixTimeZeroFilter from './filters/remove_unix_time_zero'
 import organizationWideDefaults from './filters/organization_wide_sharing_defaults'
 import { FetchElements, FETCH_CONFIG, SalesforceConfig } from './types'
+import historyTrackingFilter from './filters/history_tracking'
 import { getConfigFromConfigChanges } from './config_change'
 import { LocalFilterCreator, Filter, FilterResult, RemoteFilterCreator, LocalFilterCreatorDefinition, RemoteFilterCreatorDefinition } from './filter'
 import { addDefaults } from './filters/utils'
@@ -154,6 +155,8 @@ export const allFilters: Array<LocalFilterCreatorDefinition | RemoteFilterCreato
   { creator: xmlAttributesFilter },
   { creator: minifyDeployFilter },
   { creator: formulaDepsFilter },
+  // historyTrackingFilter depends on customObjectsToObjectTypeFilter and must run before customTypeSplit
+  { creator: historyTrackingFilter },
   // The following filters should remain last in order to make sure they fix all elements
   { creator: convertListsFilter },
   { creator: convertTypeFilter },
--- a/packages/salesforce-adapter/src/constants.ts
+++ b/packages/salesforce-adapter/src/constants.ts
@@ -157,6 +157,7 @@ export const FOLDER_CONTENT_TYPE = 'folderContentType'
 // must have the same name as INTERNAL_ID_FIELD
 export const INTERNAL_ID_ANNOTATION = INTERNAL_ID_FIELD
 export const HISTORY_TRACKED_FIELDS = 'historyTrackedFields'
+export const FEED_HISTORY_TRACKED_FIELDS = 'feedHistoryTrackedFields'
 
 // Salesforce annotations
 export const LABEL = 'label'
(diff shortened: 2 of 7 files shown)


packages/salesforce-adapter/src/filters/centralize_tracking.ts
```diff
@@ -0,0 +1,241 @@
...
+  customObject.annotations[trackingDef.aggregate] = _(customObject.fields)
+    .pickBy(field => (field.annotations[trackingDef.fieldLevelEnable] === true))
+    .mapValues(field => (field !== undefined ? new ReferenceExpression(field.elemID) : undefined))
+    .value()
+
+  Object.values(customObject.fields).forEach(field => delete field.annotations[trackingDef.fieldLevelEnable])
+  delete customObject.annotations[trackingDef.objectLevelEnable]
+}
+
+const createFieldChangeIfApplicable = async (
```
tamtamirr, line 85:
```suggestion
const createFieldChange = async (
```
Nit: The `IfApplicable` is redundant as it's clear we won't always be able to create the `Field Change` since the return type is possibly undefined.

packages/salesforce-adapter/src/filters/centralize_tracking.ts
```diff
@@ -0,0 +1,241 @@
...
+  const fieldApiName = await apiName(field)
+  const trackedBefore = Object.values(typeBefore.annotations[HISTORY_TRACKED_FIELDS] ?? {}).includes(fieldApiName)
+  const trackedAfter = Object.values(typeAfter.annotations[HISTORY_TRACKED_FIELDS] ?? {}).includes(fieldApiName)
+  if (trackedBefore === trackedAfter) {
+    return undefined
+  }
+
+  const fieldBefore = field.clone()
+  const fieldAfter = field.clone()
+  if (!trackedBefore && trackedAfter) {
```
tamtamirr, line 99: Nit: Consider exporting this logic to `wasAddedToTrackedFields` method

packages/salesforce-adapter/src/filters/centralize_tracking.ts
```diff
@@ -0,0 +1,241 @@
...
+    // field was added to the annotations
+    fieldBefore.annotations[FIELD_ANNOTATIONS.TRACK_HISTORY] = false
```
tamtamirr, line 101: I'm confused, where do we handle the other types of annotations? Isn't this logic specific to `History Tracking`?

packages/salesforce-adapter/test/filters/centralize_tracking.test.ts
```diff
@@ -0,0 +1,416 @@
...
+import { FilterWith } from '../../src/filter'
+import { Types } from '../../src/transformers/transformer'
+import {
+  API_NAME,
+  FIELD_ANNOTATIONS,
+  HISTORY_TRACKED_FIELDS,
+  OBJECT_HISTORY_TRACKING_ENABLED,
+} from '../../src/constants'
+
+const filter = filterCreator({ config: defaultFilterContext }) as FilterWith<'onFetch' | 'preDeploy' | 'onDeploy'>
```
tamtamirr, line 40: Remove as we decided not to do it in the end 😢

packages/salesforce-adapter/test/filters/centralize_tracking.test.ts
```diff
@@ -0,0 +1,416 @@
...
+
+describe(filter.name, () => {
```
tamtamirr, line 42: Where are the tests around `Feed History Tracking`?

tamtamirr (commented):
I'm a little bit confused about this iteration. We can take this F2F as-well :) Not sure about how we integrated the support in `Feed History Tracking` upon deployment.

- Can we also attach a new Workspace diff with the `Feed History Tracking` support?

packages/salesforce-adapter/src/filters/centralize_tracking.ts
tomsellek: Sorry, I made this change to the history tracking PR, but it didn't carry over to this one (presumably because the file was renamed).

packages/salesforce-adapter/test/filters/centralize_tracking.test.ts
tomsellek: Ack

packages/salesforce-adapter/src/filters/centralize_tracking.ts
```diff
@@ -0,0 +1,242 @@
...
+  FEED_HISTORY_TRACKED_FIELDS,
+  FIELD_ANNOTATIONS,
+  HISTORY_TRACKED_FIELDS,
+  OBJECT_FEED_HISTORY_TRACKING_ENABLED,
+  OBJECT_HISTORY_TRACKING_ENABLED,
+} from '../constants'
+
+const { awu } = collections.asynciterable
+
+type trackedFieldsDefinition = {
```
tamtamirr, line 45:
```suggestion
type TrackedFieldsDefinition = {
```

packages/salesforce-adapter/src/filters/centralize_tracking.ts
```diff
@@ -0,0 +1,242 @@
...
+        .filter(isField)
+        .filter(isFieldOfCustomObject)
+
+      const objectTypeChanges = await awu(changes)
+        .filter(isAdditionOrModificationChange)
+        .filter(isObjectTypeChange)
+        .filter(change => isCustomObject(getChangeData(change)))
+        .toArray()
+
+      // 1. For all CustomObjects, set the correct object-level enable value
```
tamtamirr, line 152: This comment is outdated already, since this section also handles `field level tracking` (L157)

packages/salesforce-adapter/src/filters/centralize_tracking.ts
```diff
@@ -0,0 +1,242 @@
...
+
+      const changedFieldNames = changedCustomObjectFields.map(field => field.elemID.getFullName())
+
+      const additionalChanges = await awu(objectTypeChanges)
+        .filter(isModificationChange)
+        .flatMap(change => awu(Object.values(getChangeData(change).fields))
+          .filter(field => !changedFieldNames.includes(field.elemID.getFullName()))
+          .map(field => createTrackingFieldChange(field, change, trackingDef))
+          .toArray())
+        .filter(valueUtils.isDefined)
```
tamtamirr, line 191: Nit: We usually import isDefined as follows across all of the OSS:
```
import { values } from '@salto-io/lowerdash'
const { isDefined } = values
```

packages/salesforce-adapter/test/filters/centralize_tracking.test.ts
tamtamirr: A bug in this filter can be really fragile. I would modify the tested `CustomObject` to include `feed history tracking` related annotations on the type & it's fields (the same as you do for `track history`) just to make sure it's covered.

tamtamirr (commented):
Looking good in general, added some small comments.

Please note the previous discussion we have regarding the UTs around `Feed History Tracking`.

tamtamirr:
@tomsellek Also, attach a workspace diff to this PR as-well.

tamtamirr:
@tomsellek Please update the PR description to include the description of the previous PR. You can close the other PR as it doesn't make sense to merge both (You've been working on the exact same code, and you may run into conflicts as-well)

packages/salesforce-adapter/src/filters/centralize_tracking.ts
tomsellek: Ack

packages/salesforce-adapter/test/filters/centralize_tracking.test.ts
tomsellek: Ack

tamtamirr (approved):
LGTM 👑 ❤️  Some things to note before merging:

- Make sure we tested for edge cases on both tracking fields.
- Update the description with a Workspace diff that includes the remodeling on both the fields.
- Notify CSMs that we can fire away the PROD ticket so we can merge this one soon.

ori-moisis:
is this replacing https://github.com/salto-io/salto/pull/3902?
which one is the real one?

ori-moisis:
so I reviewed the wrong one 😢 
I assume most of the comments there apply here though

tomsellek:
Superseded by https://github.com/salto-io/salto/pull/4400

PR #4047, opened 2023-03-05
SALTO-3154: User facing errors for errors thrown upon login

tamtamirr:
Seems like errors that were thrown from [requiresLogin](https://github.com/salto-io/salto/blob/d614e86396266818cf987c8f38d3e32efe1a29b3/packages/adapter-components/src/client/decorators.ts#L22) were not mapped due to the order of the [decorators](https://www.typescriptlang.org/docs/handbook/decorators.html) on each client method. It explains why when users received `invalid_grant` when attempting to login the error was not mapped to our custom user friendly error message.

Nice to see that the UT I've added to validate the scenario fails on the code that is currently in main 😄 

Screen Shot 2023-03-05 at 9 17 46


roeeneg (approved):
LGTM 🥇

PR #4050, opened 2023-03-05
filter out unresolved references

idoser96:

---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/zendesk-adapter/src/dependency_changers/ticket_form_change.ts
+++ b/packages/zendesk-adapter/src/dependency_changers/ticket_form_change.ts
@@ -19,7 +19,7 @@ import {
   getChangeData,
   InstanceElement,
   isAdditionChange,
-  isInstanceChange, isReferenceExpression, isRemovalChange, ReferenceExpression,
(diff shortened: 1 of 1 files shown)


packages/zendesk-adapter/src/dependency_changers/ticket_form_change.ts
```diff
@@ -54,7 +54,9 @@ export const ticketFormDependencyChanger: DependencyChanger = async changes => {
 
   const ticketFormOrderValue = getChangeData(ticketFormOrderChange.change).value
   const orderTicketForms = new Set((ticketFormOrderValue.active ?? []).concat(ticketFormOrderValue.inactive ?? [])
-    .filter(isReferenceExpression).map((ref: ReferenceExpression) => ref.value.elemID.getFullName()).flat())
+    .filter(isReferenceExpression).filter((ref: ReferenceExpression) => isInstanceElement(ref.value))
```
netama, line 57: please add a comment explaining what this is catching (since it may happen in other places as well and we will likely want to generalize it eventually)

netama (approved):
LGTM with a small nit, please add release notes as well

PR #4052, opened 2023-03-05
SALTO-3741 Add missing features to manifest.xml as optional first

dantal4:
Add missing features as optional first, and if the deploy fails then add it as required.
This way we don't force the target account to have an enable feature, when it is not required.

In the adapter config, features are included as optional by default. In order to include a required feature add the `":required"` suffix to it.
Example:
```
netsuite {
  deploy = {
    additionalDependencies = {
      include = {
        features = [
          "DEPARTMENTS", // will be added as optional
          "SUBSCRIPTIONBILLING:required", // will be added as required
        ]
      }
    }
  }
}
```

---

---
_Release Notes_: 
Netsuite Adapter:
- Add missing features to manifest.xml as optional first
- In the adapter config, features are included as optional by default. In order to include a required feature add the `":required"` suffix to it.

---
_User Notifications_: 
None


packages/netsuite-adapter/src/config.ts
```diff
@@ -383,27 +393,45 @@ const additionalDependenciesConfigPath: string[] = [
...
 const validateAdditionalDependencies = (
-  { include, exclude }: Record<keyof AdditionalDependencies, unknown>
+  { include, exclude }: Partial<Record<keyof AdditionalDependencies, unknown>>
 ): void => {
+  let validatedInclude: Partial<AdditionalSdfDeployDependencies> = {}
   if (include !== undefined) {
     validatePlainObject(include, additionalDependenciesConfigPath.concat('include'))
     validateAdditionalSdfDeployDependencies(include, 'include')
+    validatedInclude = include
   }
```
omrilit, line 417: nit: you can extract this code to a typeguard to remove the need of the `validatedInclude`.
```suggestion
  isPartialAdditionalSdfDeployDependencies(include, 'include')
```

packages/netsuite-adapter/src/client/client.ts
```diff
@@ -217,6 +215,63 @@ export default class NetsuiteClient {
...
+      } else if (feature.status === 'optional' && !feature.canBeRequired) {
+        log.warn('The %s feature is required but it is excluded', featureName)
+        missingExcludedFeatures.add(featureName)
+      }
+    })
+    if (missingExcludedFeatures.size > 0) {
+      const error = new Error(`The following features are required but they are excluded: ${Array.from(missingExcludedFeatures).join(', ')}.`)
+      return { failToUpdate: true, error }
+    }
+    return { failToUpdate }
```
omrilit, line 272:
```suggestion
    return { failedToUpdate }
```

packages/netsuite-adapter/src/client/client.ts
```diff
@@ -217,6 +215,63 @@ export default class NetsuiteClient {
...
+  ): FeaturesMap {
+    return Object.fromEntries([
+      ...additionalDependencies.include.features
+        .map(featureName => [featureName, { status: 'optional', canBeRequired: true }]),
+      ...additionalDependencies.exclude.features
+        .map(featureName => [featureName, { status: 'excluded' }]),
+    ])
+  }
+
+  private static updateFeaturesMap(
```
omrilit, line 246: NICE!

packages/netsuite-adapter/src/client/client.ts
```diff
@@ -217,6 +215,63 @@ export default class NetsuiteClient {
...
+    let failToUpdate = false
+    missingFeaturesError.missingFeatures.forEach(featureName => {
+      const feature = featuresMap[featureName]
+      if (feature === undefined) {
+        featuresMap[featureName] = { status: 'optional', canBeRequired: true }
+      } else if (feature.status === 'excluded') {
+        featuresMap[featureName] = { status: 'optional', canBeRequired: false }
+      } else if (feature.status === 'required') {
+        log.error('The %s feature is already required, but sdf returned an error: %o', featureName, missingFeaturesError)
+        failToUpdate = true
```
omrilit, line 260: Why did you decide not to return an error to the caller in this case?
If you'll decide we should return such, then it makes the `failToUpdate` returned value redundant.

packages/netsuite-adapter/src/client/manifest_utils.ts
```diff
@@ -137,34 +137,35 @@ const fixDependenciesObject = (dependencies: Value): void => {
...
   features.feature = _(makeArray(features.feature))
-    // remove all required features
-    .differenceBy(requiredFeatures, item => item[TEXT_ATTRIBUTE])
-    // re-add all required features with "required=true"
-    .unionBy(requiredFeatures, item => item[TEXT_ATTRIBUTE])
-    .filter(item => !additionalDependencies.exclude.features.includes(item[TEXT_ATTRIBUTE]))
+    // remove all additional features
+    .differenceBy(additionalFeatures, getTagText)
+    // re-add all additional features with the desired "required" value for each feature
+    .unionBy(additionalFeatures, getTagText)
```
omrilit, line 161: Just to make sure that I follow, is it ok to put only the feature name when it's optional and not using the `{ [REQUIRED_ATTRIBUTE]: 'false', [TEXT_ATTRIBUTE]: feature }` object?
Or I am missing something here?

omrilit (commented):
Very nice work!

packages/netsuite-adapter/src/client/client.ts
dantal4: The error that will be returned is the error from SDF, so IMO we shouldn't create another error (not like when a feature was manually excluded, so we want the user to know that).

omrilit: You can still return the original error from here, no?

packages/netsuite-adapter/src/client/manifest_utils.ts
dantal4: You're right-
```
An error occurred during manifest validation.
Details: The required attribute for the SUBSCRIPTIONBILLING feature must be set to true or false in the manifest.
File: ~/manifest.xml
```

packages/netsuite-adapter/src/config.ts
```diff
@@ -404,6 +421,18 @@ const validateAdditionalDependencies = (
     validatePlainObject(exclude, additionalDependenciesConfigPath.concat('exclude'))
     validateAdditionalSdfDeployDependencies(exclude, 'exclude')
   }
+  if (include?.features && exclude?.features) {
+    const conflictedFeatures = _.intersection(include.features, exclude.features)
```
omrilit, line 425: The `REQUIRED_FEATURE_SUFFIX` should be removed from the included features, no?

omrilit (approved):
Looks good!
Approving to save cycle.

packages/netsuite-adapter/src/config.ts
dantal4: right

PR #4053, opened 2023-03-06
SALTO-3768: store users during fetch in a hidden NaCl

IdoZakk:
Now storing users' info during fetch in a hidden NaCl.
On deploy we will no longer fetch the users.

---

I verified that we could handle the case in which the nacl does not exist- we simply do not issue CV errors

---
_Release Notes_: 
Jira Adapter:
Deploy plan will become faster thanks to a change in the way we handle users' data

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/change_validators/account_id.ts
+++ b/packages/jira-adapter/src/change_validators/account_id.ts
@@ -14,13 +14,13 @@
 * limitations under the License.
 */
 import { ChangeError, ChangeValidator, ElemID, getChangeData, InstanceElement,
-  isAdditionOrModificationChange, isInstanceChange } from '@salto-io/adapter-api'
+  isAdditionOrModificationChange, isInstanceChange, ReadOnlyElementsSource } from '@salto-io/adapter-api'
 import { config as configUtils } from '@salto-io/adapter-components'
 import { logger } from '@salto-io/logging'
 import { walkOnElement } from '@salto-io/adapter-utils'
 import _ from 'lodash'
 import { isDeployableAccountIdType, walkOnUsers, WalkOnUsersCallback } from '../filters/account_id/account_id_filter'
-import { getUserIdFromEmail, GetUserMapFunc, getUsersMapByVisibleId, UserMap } from '../users'
+import { getUserIdFromEmail, getUserMapFunc, getUsersMapByVisibleId, UserMap } from '../users'
 import { JiraConfig } from '../config/config'
 import JiraClient from '../client/client'
 import { JIRA_USERS_PAGE, PERMISSION_SCHEME_TYPE_NAME } from '../constants'
@@ -242,15 +242,15 @@ const createChangeErrorsForAccountIdIssues = (
 export const accountIdValidator: (
   client: JiraClient,
   config: JiraConfig,
-  getUserMapFunc: GetUserMapFunc
+  elementsSource: ReadOnlyElementsSource,
 ) =>
-  ChangeValidator = (client, config, getUserMapFunc) => async changes =>
+  ChangeValidator = (client, config, elementsSource) => async changes =>
     log.time(async () => {
       if (!(config.fetch.convertUsersIds ?? true)) {
         return []
       }
       const { baseUrl, isDataCenter } = client
-      const rawUserMap = await getUserMapFunc()
+      const rawUserMap = await getUserMapFunc(elementsSource)
       if (rawUserMap === undefined) {
         return []
       }
--- a/packages/jira-adapter/src/change_validators/wrong_user_permission_scheme.ts
+++ b/packages/jira-adapter/src/change_validators/wrong_user_permission_scheme.ts
@@ -13,13 +13,13 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
-import { ChangeError, ChangeValidator, getChangeData, InstanceElement, isAdditionOrModificationChange, isInstanceChange } from '@salto-io/adapter-api'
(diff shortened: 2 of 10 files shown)


packages/jira-adapter/src/change_validators/account_id.ts
```diff
@@ -242,15 +242,15 @@ const createChangeErrorsForAccountIdIssues = (
 export const accountIdValidator: (
   client: JiraClient,
   config: JiraConfig,
-  getUserMapFunc: GetUserMapFunc
+  elementsSource: ReadOnlyElementsSource,
 ) =>
-  ChangeValidator = (client, config, getUserMapFunc) => async changes =>
+  ChangeValidator = (client, config, elementsSource) => async changes =>
```
alonstern, line 247: We already get the elements source in the change validator function, meaning you can do
```
  ChangeValidator = (client, config) => async (changes, elementsSource) =>
```

packages/jira-adapter/src/users.ts
```diff
@@ -140,6 +141,13 @@ export const getUserMapFuncCreator = (paginator: clientUtils.Paginator, isDataCe
   }
 }
 
+export const getUserMapFunc = async (
```
alonstern, line 144: nit: maybe `getUserMapFromElementSource`? to differ it from the other getUserMapFunc

packages/jira-adapter/src/users.ts
```diff
@@ -140,6 +141,13 @@ export const getUserMapFuncCreator = (paginator: clientUtils.Paginator, isDataCe
...
+  elementSource: ReadOnlyElementsSource
+): Promise<UserMap | undefined> => {
+  const usersElement = await elementSource.get(new ElemID('jira', 'Users', 'instance', 'users'))
+  return usersElement?.value?.users as (UserMap | undefined)
```
alonstern, line 148: probably worth adding a log here if it is undefined

packages/jira-adapter/src/change_validators/wrong_user_permission_scheme.ts
```diff
@@ -45,13 +45,13 @@ Check ${new URL(JIRA_USERS_PAGE, url).href} to see valid users and account IDs.`
...
-  getUserMapFunc: GetUserMapFunc
-  ) => ChangeValidator = (client, config, getUserMapFunc) => async changes => {
+  elementsSource: ReadOnlyElementsSource,
+  ) => ChangeValidator = (client, config, elementsSource) => async changes => {
     if (!(config.fetch.convertUsersIds ?? true)) {
       return []
     }
     const { baseUrl } = client
-    const rawUserMap = await getUserMapFunc()
+    const rawUserMap = await getUserMapFunc(elementsSource)
```
alonstern, line 54: You also should change it in the `wrong_user_permission_scheme_filter.ts`
Make sure when you run a deployment that `jira:client.getSinglePage(/rest/api/3/users/search)` does not appear in the logs

packages/jira-adapter/src/users.ts
IdoZakk: That was my initial name, but it seems too long. I thought that passing the elementSource makes it clear that it is from the element Source. Do you think I should change?

packages/jira-adapter/src/filters/store_users.ts
```diff
@@ -0,0 +1,71 @@
...
+    displayName: { refType: BuiltinTypes.STRING },
+    email: { refType: BuiltinTypes.STRING },
+    username: { refType: BuiltinTypes.STRING },
+  },
+  annotations: {
+    [CORE_ANNOTATIONS.HIDDEN]: true,
+  },
+})
+
+const usersType = new ObjectType({
```
netama, line 37: we shouldn't define the object types hard-coded here as they may be shared across accounts - can we create this in the relevant function instead?

packages/jira-adapter/src/filters/store_users.ts
```diff
@@ -0,0 +1,71 @@
...
+    { [CORE_ANNOTATIONS.HIDDEN]: true },
+  )
+
+
+/*
+ * stores the fetched users to an hidden nacl
+*/
+const filter: FilterCreator = ({ getUserMapFunc }) => ({
+  name: 'storeUsers',
+  onFetch: async elements => log.time(async () => {
```
netama, line 62: why do we need `log.time()` here? I think we already log it in the filter

packages/jira-adapter/src/filters/store_users.ts
```diff
@@ -0,0 +1,71 @@
...
+  elemID: new ElemID(JIRA, 'Users'),
+  fields: {
+    users: { refType: new MapType(userInfoType) },
+  },
+  annotations: {
+    [CORE_ANNOTATIONS.HIDDEN]: true,
+  },
+})
+
+const getUsers = (userMap: UserMap | undefined): InstanceElement =>
```
netama, line 47: nit: can we give this a more specific name? this function is creating an instance, not getting the users

packages/jira-adapter/src/users.ts
```diff
@@ -140,6 +141,13 @@ export const getUserMapFuncCreator = (paginator: clientUtils.Paginator, isDataCe
   }
 }
 
+export const getUserMapFunc = async (
+  elementSource: ReadOnlyElementsSource
+): Promise<UserMap | undefined> => {
+  const usersElement = await elementSource.get(new ElemID('jira', 'Users', 'instance', 'users'))
```
netama, line 147: all of these should be constants (except maybe for `instance` which is validated from the enum)

packages/jira-adapter/src/users.ts
alonstern: Maybe `loadUserMap`?

alonstern (approved):
LGTM, let's just verify the compliance thing before merging

PR #4056, opened 2023-03-06
SALTO-3774 allow configuring idFields that only exist in some elements

netama:
Until now we assumed that if a field configured as an id field does not exist in an element, it is a bug and we should not attempt to generate the id based on the configuration at all. 
That behavior was not ideal - and we now have cases where we want to use id fields as differentiators even if they only exist in some elements (for example, in Zendesk views we may want to rely on restrictions even though they do not always exist).

---

The PR includes some fixes in specific adapters, and is not supposed to cause different ids to be generated in any of the existing element types (even when ids are regenerated) - so no upgrade is needed.

Still need to add some tests

---
_Release Notes_: 
None (not user facing)

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/filters/referenced_instance_names.ts
+++ b/packages/adapter-components/src/filters/referenced_instance_names.ts
@@ -107,7 +107,7 @@ const createInstanceNameAndFilePath = (
   const { idFields, nameMapping } = idConfig
   const newNameParts = createInstanceReferencedNameParts(instance, idFields)
   const newName = joinInstanceNameParts(newNameParts) ?? instance.elemID.name
-  const parentName = getFirstParentElemId(instance)?.name
+  const parentName = idConfig.extendsParentId ? getFirstParentElemId(instance)?.name : undefined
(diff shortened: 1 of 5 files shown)


shir-reifenberg (commented):
Cool 😎  added 2 questions

packages/adapter-components/src/filters/referenced_instance_names.ts
```diff
@@ -107,7 +107,7 @@ const createInstanceNameAndFilePath = (
   const { idFields, nameMapping } = idConfig
   const newNameParts = createInstanceReferencedNameParts(instance, idFields)
   const newName = joinInstanceNameParts(newNameParts) ?? instance.elemID.name
-  const parentName = getFirstParentElemId(instance)?.name
+  const parentName = idConfig.extendsParentId ? getFirstParentElemId(instance)?.name : undefined
```
netama, line 110: @shir-reifenberg I think this was a bug fix, we don't need it anymore?

PR #4057, opened 2023-03-06
SALTO-3775 update zendesk config doc

netama:
Update the config with the latest changes

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/zendesk-adapter/src/config.ts
```diff
@@ -1465,6 +1465,7 @@ export const DEFAULT_TYPES: ZendeskApiConfig['types'] = {
       idFields: ['&locale_id'],
       fieldsToHide: FIELDS_TO_HIDE.concat({ fieldName: 'id', fieldType: 'number' }),
       fieldTypeOverrides: [{ fieldName: 'id', fieldType: 'number' }],
+      extendsParentId: true,
```
netama, line 1468: this was missing - ran into it as part of another PR, but fixing here as well to avoid having to update the config twice

edenhassid (approved):

PR #4035, opened 2023-02-28
SALTO-3716 create core fetch config as input to core fetch

omrifarkash:
- Reverting https://github.com/salto-io/salto/pull/3963
- Introduce new flag for core fetch 'withChangeDetection'
- Make Netsuite adapter use 'withChangeDetection' received from core to run quick fetch
---

_Additional context for reviewer_

---
_Release Notes_: 
Netsuite_adapter:
When receiving 'withChangeDetection' from core on fetch, will run quick fetch. (not supported on the CLI yet)

---
_User Notifications_: 
_Replace me with a short sentence that describes changes that will appear in NaCls and are not caused by user actions (e.g. None

PR #4059, opened 2023-03-07
SAAS-6749 support quick fetch when not selecting fetch target

kami-delaroz:
support quick fetch when not selecting fetch target. remove the default value and use only explicit settings

---

https://github.com/salto-io/salto/pull/4035
https://github.com/salto-io/salto/pull/4022

---
_Release Notes_: 
Support quick fetch when not selecting fetch target

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/netsuite-adapter/src/adapter.ts
+++ b/packages/netsuite-adapter/src/adapter.ts
@@ -62,7 +62,7 @@ import customRecordsFilter from './filters/custom_records'
 import currencyUndeployableFieldsFilter from './filters/currency_omit_fields'
 import additionalChanges from './filters/additional_changes'
 import { Filter, FilterCreator } from './filter'
-import { getConfigFromConfigChanges, NetsuiteConfig, DEFAULT_DEPLOY_REFERENCED_ELEMENTS, DEFAULT_WARN_STALE_DATA, DEFAULT_USE_CHANGES_DETECTION, DEFAULT_VALIDATE } from './config'
+import { getConfigFromConfigChanges, NetsuiteConfig, DEFAULT_DEPLOY_REFERENCED_ELEMENTS, DEFAULT_WARN_STALE_DATA, DEFAULT_VALIDATE } from './config'
 import { andQuery, buildNetsuiteQuery, NetsuiteQuery, NetsuiteQueryParameters, notQuery, QueryParams, convertToQueryParams, getFixedTargetFetch } from './query'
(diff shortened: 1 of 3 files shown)


packages/netsuite-adapter/src/change_validators/safe_deploy.ts
```diff
@@ -103,7 +102,7 @@ const getMatchingServiceElements = async (
   const fetchQuery = buildNetsuiteQuery(convertToQueryParams(fetchTarget))
   const {
     elements: fetchedElements,
-  } = await fetchByQuery(fetchQuery, { reportProgress: () => null }, false, true)
+  } = await fetchByQuery(fetchQuery, { reportProgress: () => null }, false) // TODO: verify
```
kami-delaroz, line 105: @dantal4

packages/netsuite-adapter/src/adapter.ts
```diff
@@ -339,15 +338,26 @@ export default class NetsuiteAdapter implements AdapterOperations {
     }
   }
 
+  private shouldFetchWithChangesDetection(
+    withChangesDetection: boolean,
+    hasFetchTarget: boolean,
+    isFirstFetch: boolean
+  ) : boolean {
```
dantal4, line 345: nit - a function that receives multiple boolean params should get them in one object param.

packages/netsuite-adapter/src/adapter.ts
```diff
@@ -339,15 +338,26 @@ export default class NetsuiteAdapter implements AdapterOperations {
...
+    return !isFirstFetch && (
+      this.useChangesDetection === true
+      || withChangesDetection
+      || (hasFetchTarget && !(this.useChangesDetection === false)))
```
dantal4, line 349: nit
```suggestion
      || (hasFetchTarget && this.useChangesDetection !== false))
```
Also IMO it's worth adding a comment about the implicit decision that if `this.useChangesDetection` is undefined than the default is true.

dantal4 (commented):
Nice work!
Added some nits. Please add a test too 🙏

packages/netsuite-adapter/src/adapter.ts
kami-delaroz: even if it's a private single used function? :)

dantal4: yes, because next time someone will use it they can make a mistake.

dantal4 (approved):
Very nice!

PR #4061, opened 2023-03-07
SALTO-3783 Avoid logging unneeded data in client

netama:
* Avoid logging headers except where needed (currently only Okta)
* Avoid logging large buffer data

---
_Release Notes_: 
None

---
_User Notifications_: 
None


shir-reifenberg (approved):
LGTM 👯 
WDYT about logging rate limit headers as well for debugging purposes?

PR #4063, opened 2023-03-08
add SDF missing feature regex

dantal4:
---

---
_Release Notes_: 
Netsuite-adapter:
- Add SDF missing feature regex

---
_User Notifications_: 
None


omrilit (approved):
Thanks!

PR #4066, opened 2023-03-08
SALTO-3788/top level element path index

guyhamo:
created a new index which is identical to path index apart from the fact that it only hold paths of elements that are top level.

---

current known issue that is not fixed here is that partial fetch will break the path index and the top level path index, 
it will be fixed in a different pr.

---
_Release Notes_: 
core:
* add a new index that will hold top level elements current paths.

---
_User Notifications_:

### DIFF
--- a/packages/cli/test/commands/deploy.test.ts
+++ b/packages/cli/test/commands/deploy.test.ts
@@ -335,6 +335,7 @@ describe('deploy command', () => {
       return state.buildInMemState(async () => ({
         elements: createInMemoryElementSource(),
         pathIndex: new InMemoryRemoteMap<pathIndex.Path[]>(),
+        topLevelPathIndex: new InMemoryRemoteMap<pathIndex.Path[]>(),
         accountsUpdateDate: dateMap ?? new InMemoryRemoteMap(),
         saltoMetadata,
         staticFilesSource: mocks.mockStateStaticFilesSource(),
--- a/packages/workspace/src/workspace/path_index.ts
+++ b/packages/workspace/src/workspace/path_index.ts
@@ -181,10 +181,8 @@ export const overridePathIndex = async (
 const getTopLevelPathHints = (unmergedElements: Element[]): PathHint[] => {
   const topLevelElementsWithPath = unmergedElements
     .filter(e => e.path !== undefined)
-    .filter(e => e.elemID.isTopLevel())
   const elementsByID = _.groupBy(topLevelElementsWithPath, e => e.elemID.getFullName())
   return Object.entries(elementsByID)
-    .filter(([_key, value]) => value.length > 0)
     .map(([key, value]) => ({
       key,
       value: value.map(e => e.path as Path),
--- a/packages/workspace/src/workspace/path_index.ts
+++ b/packages/workspace/src/workspace/path_index.ts
@@ -182,8 +182,16 @@ export const overrideTopLevelPathIndex = async (
   current: PathIndex,
   unmergedElements: Element[],
 ): Promise<void> => {
-  const entries = getElementsPathHints(unmergedElements)
-    .filter(e => ElemID.fromFullName(e.key).isTopLevel())
+  const topLevelElementsWithPath = unmergedElements
+    .filter(e => e.path !== undefined)
+    .filter(e => e.elemID.isTopLevel())
+  const elementsByID = _.groupBy(topLevelElementsWithPath, e => e.elemID.getFullName())
+  const entries = Object.entries(elementsByID)
+    .filter(([_key, value]) => value.length > 0)
+    .map(([key, value]) => ({
+      key,
+      value: value.map(e => e.path as Path),
+    }))
   await current.clear()
   await current.setAll(entries)
 }
--- a/packages/workspace/src/workspace/path_index.ts
+++ b/packages/workspace/src/workspace/path_index.ts
@@ -198,27 +198,34 @@ export const overrideTopLevelPathIndex = async (
   await current.setAll(entries)
 }
 
+type UpdateIndexParms = {
+  index: PathIndex
+  elements: Element[]
+  accountsToMaintain: string[]
+  isTopLevel: boolean
+}
+
 export const updatePathIndex = async (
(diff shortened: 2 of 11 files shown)


packages/workspace/src/workspace/path_index.ts
```diff
@@ -178,6 +178,24 @@ export const overridePathIndex = async (
...
+  const topLevelElementsWithPath = unmergedElements
+    .filter(e => e.path !== undefined)
+    .filter(e => e.elemID.isTopLevel())
+  const elementsByID = _.groupBy(topLevelElementsWithPath, e => e.elemID.getFullName())
+  const entries = Object.entries(elementsByID)
+    .filter(([_key, value]) => value.length > 0)
+    .map(([key, value]) => ({
+      key,
+      value: value.map(e => e.path as Path),
+    }))
```
ori-moisis, line 194: since we only really want one path per-element, feels like this can be simpler...
e.g
```typescript
const entries = unmergedElements
  .filter(elem => values.isDefined(elem.path))
  .map(elem => ({ key: elem.elemID.getFullName(), value: elem.path })) 
```
not sure if it is important to remove duplicate keys before going to rocksdb, if it is (check performance) we could add a `_.uniqBy` after this...

packages/cli/test/commands/deploy.test.ts
```diff
@@ -335,6 +335,7 @@ describe('deploy command', () => {
       return state.buildInMemState(async () => ({
         elements: createInMemoryElementSource(),
         pathIndex: new InMemoryRemoteMap<pathIndex.Path[]>(),
+        topLevelPathIndex: new InMemoryRemoteMap<pathIndex.Path[]>(),
```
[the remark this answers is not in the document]
guyhamo: there are still multiple paths for some elements, account object in salesforce will have 3 different paths. annotations, fields and custom fields.

packages/workspace/src/workspace/state/in_mem.ts
```diff
@@ -115,13 +116,19 @@ export const buildInMemState = (
       await updatePathIndex(
         currentStateData.pathIndex, unmergedElements, servicesNotToChange
       )
+      await overrideTopLevelPathIndex(
+        currentStateData.topLevelPathIndex, unmergedElements
+      )
```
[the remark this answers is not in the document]
guyhamo: servicesNotToChange was ignored on purpose but I did not think about that option, I will check out how partial fetches and multiple account environments behave with this. thanks

packages/workspace/src/workspace/path_index.ts
guyhamo: elementsByID will have multiple elements for some ids like account object in salesforce.

packages/workspace/src/workspace/path_index.ts
```diff
@@ -178,16 +178,43 @@ export const overridePathIndex = async (
   await current.setAll(entries)
 }
 
+const getTopLevelPathHints = (unmergedElements: Element[]): PathHint[] => {
+  const topLevelElementsWithPath = unmergedElements
+    .filter(e => e.path !== undefined)
+    .filter(e => e.elemID.isTopLevel())
+  const elementsByID = _.groupBy(topLevelElementsWithPath, e => e.elemID.getFullName())
+  return Object.entries(elementsByID)
+    .filter(([_key, value]) => value.length > 0)
```
ori-moisis, line 187: I think this filter is redundant, pretty sure `groupBy` cannot create keys without any value...

packages/workspace/src/workspace/path_index.ts
```diff
@@ -178,16 +178,43 @@ export const overridePathIndex = async (
   await current.setAll(entries)
 }
 
+const getTopLevelPathHints = (unmergedElements: Element[]): PathHint[] => {
+  const topLevelElementsWithPath = unmergedElements
+    .filter(e => e.path !== undefined)
+    .filter(e => e.elemID.isTopLevel())
```
ori-moisis, line 184: nit: this last filter is redundant, all elements in this flow are top level elements

packages/workspace/src/workspace/state/in_mem.ts
```diff
@@ -106,22 +111,29 @@ export const buildInMemState = (
...
+      await overrideTopLevelPathIndex(currentStateData.topLevelPathIndex, unmergedElements)
     },
     updatePathIndex: async (
       unmergedElements: Element[],
       servicesNotToChange: string[]
     ): Promise<void> => {
       const currentStateData = await stateData()
       await updatePathIndex(
-        currentStateData.pathIndex, unmergedElements, servicesNotToChange
+        currentStateData.pathIndex, unmergedElements, servicesNotToChange, false
```
ori-moisis, line 122: unnamed boolean parameters are not very readable (it is not easy to understand what `false` means here)
it would probably be best to change the function signature to accept a named parameters object as input...

packages/workspace/test/workspace/state.test.ts
```diff
@@ -155,26 +160,34 @@ describe('state', () => {
     it('getPathIndex', async () => {
       expect(await state.getPathIndex()).toEqual(pathIndex)
     })
+    it('getTopLevelPathIndex', async () => {
+      expect(await state.getTopLevelPathIndex()).toEqual(topLevelPathIndex)
+    })
     it('overridePathIndex', async () => {
-      const elements = [elem, newElem]
+      const elements = [elem, newElem, newField]
```
ori-moisis, line 167: AFAIK fields cannot be inputs to `overridePathIndex` as they are not top level elements and only top level elements are passed into / out of the state.

if I am wrong about this please let me know, but if I am not, then this test is not valid

ori-moisis (commented):
overall looks good
missing unit tests for the new functions

packages/workspace/test/workspace/state.test.ts
guyhamo: it matches this [comment](https://github.com/salto-io/salto/pull/4066#discussion_r1135254812) so i will remove that filter and this specific check.

PR #4067, opened 2023-03-08
SALTO-3766 improve elementSourceIndexes

dantal4:
Improve the running time of `elementSourceIndexes` indexes creation:
- Separate to fetch/deploy indexes
- In full fetch - calculate only changedBy/changedAt indexes

---

---
_Release Notes_: 
Netsuite Adapter:
- Improve elementSourceIndexes

---
_User Notifications_: 
None


omrilit (commented):
Looks great and I assume it will be felt in users' deploy performance (BTW please elaborate a bit more in the Release Notes).

I wonder whether you thought about modifying the indexes to be lazy separately from each other, i.e. an index will be calculated only when needed, without classifying them to Fetch / Deploy indexes.
The thing is that with the current impl, it's not really type safe, and one can call getDeployIndexes from Fetch flow.
WDYT?

dantal4:
@omrilit I'm working on another 2 PRs that will let us not use the indexes in deploy at all, so this one probably won't be necessary,

PR #4069, opened 2023-03-09
INCIDENT-3657: Fixed a bug where the fetch would fail when an holder is missing from a security level

alonstern:
Fixed a bug where the fetch would fail when an holder is missing from a security level

---
_Release Notes_: 
_Jira Adapter_:
- Fixed a bug where the fetch would fail when a holder is missing from a security level

---
_User Notifications_: 
None


IdoZakk (approved):
:fire:

PR #4073, opened 2023-03-09
SALTO-3800: Jira DC user scheme does not match a valid user

alonstern:
We got a user in a customer account with an empty emailAddress

---
_Release Notes_: 
_Jira Adapter_:
- Fixed failure in getting the current user in some cases

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/test/filters/locale.test.ts
+++ b/packages/jira-adapter/test/filters/locale.test.ts
@@ -45,6 +45,26 @@ describe('localeFilter', () => {
       })
     })
 
+    it('should work when email is an empty string', async () => {
+      connection.get.mockResolvedValue({
+        status: 200,
+        data: {
+          key: 'key',
+          name: 'name',
+          displayName: 'displayName',
+          locale: 'es_ES',
+          emailAddress: '',
+        },
+      })
+      const filterRes = await filter.onFetch([])
+      expect(filterRes).toEqual({
+        errors: [{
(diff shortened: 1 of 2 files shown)


packages/jira-adapter/test/filters/locale.test.ts
```diff
@@ -45,6 +45,26 @@ describe('localeFilter', () => {
...
+      })
+      const filterRes = await filter.onFetch([])
+      expect(filterRes).toEqual({
+        errors: [{
+          message: 'Your Jira Data Center instance is not set to English-US language. Salto currently only supports accessing Jira DC through users with their default language set to English-US. Please change the user’s language, or create another user with English as its Jira language, and change Salto\'s credentials to use it. After doing that, make sure you re-fetch your environment using an advanced fetch, with “Regenerate Salto IDs” turned on. You only need to do this once. For help on how to change Jira users\' language, go to https://confluence.atlassian.com/adminjiraserver/choosing-a-default-language-938847001.html',
+          severity: 'Warning',
+        }],
+      })
+    })
+
```
IdoZakk, line 67: Why not use en_us locale and expect no errors?

IdoZakk (approved):
👌 but I think it is better to change the test

packages/jira-adapter/test/filters/locale.test.ts
alonstern: When we fail to check the locale we return no warning, so checking that there are no errors in this case will pass even without my fix

alonstern: I am not sure I agree. The bug we had is that when the email is empty we would not get a warning about the locale if it is wrong, so I think the best way to test to bug is solved is to verify that we get the warning the email is empty. Whether we write a log or not is less interesting IMO

PR #4087, opened 2023-03-13
SALTO-3766 improve convert-lists-to-maps preDeploy filter

dantal4:
Do not use `elementSourceIndex` to transform maps back to lists.

---

---
_Release Notes_: 
Netsuite Adapter:
- Improve convert-lists-to-maps preDeploy filter

---
_User Notifications_: 
None

### DIFF
--- a/packages/netsuite-adapter/src/mapped_lists/utils.ts
+++ b/packages/netsuite-adapter/src/mapped_lists/utils.ts
@@ -266,6 +266,9 @@ export const convertDataInstanceMapsToLists = async (instance: InstanceElement):
   })
 }
 
+// We use hardcoded types when running transformMapsToLists,
+// so the fields don't have the `map_key_field` annotation.
+// Because of that, we add that annotation on the mapped list itself.
 const setListMappedByFieldToValue = <T extends ChangeDataType>(element: T): Promise<T> =>
   transformElement({
     element,
@@ -283,6 +286,7 @@ const transformMapsToLists = (element: ChangeDataType): Promise<ChangeDataType> 
     transformFunc: async ({ value, field }) => (
       field !== undefined && await isMappedList(value, field)
         ? _.sortBy(
+          // Removing the `map_key_field` annotation that was used to identify the mapped list.
           Object.values(_.omit(value, [LIST_MAPPED_BY_FIELD])),
           [INDEX, value[LIST_MAPPED_BY_FIELD]].flat()
         ).map(item => (_.isObject(item) ? _.omit(item, INDEX) : item))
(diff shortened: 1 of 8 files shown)


packages/netsuite-adapter/src/mapped_lists/utils.ts
```diff
@@ -191,13 +190,10 @@ export const convertAnnotationListsToMaps = async (
...
   field: Field,
-  mapKeyFieldsIndex?: Record<string, string | string[]>,
 ): Promise<boolean> =>
-  (mapKeyFieldsIndex === undefined
-    ? field.annotations[LIST_MAPPED_BY_FIELD]
-    : mapKeyFieldsIndex[field.elemID.getFullName()])
-  && _.isPlainObject(value)
+  _.isPlainObject(value)
   && isContainerType(await field.getType())
+  && (field.annotations[LIST_MAPPED_BY_FIELD] || value[LIST_MAPPED_BY_FIELD])
```
omrilit, line 196: Is it really possible to have a situation where value[LIST_MAPPED_BY_FIELD] is evaluated to true while field.annotations[LIST_MAPPED_BY_FIELD] is evaluated to false?

If not, then setListMappedByFieldToValue is redundant.
From my understanding in order to set `value[LIST_MAPPED_BY_FIELD` we always first call `isMappedList`.
`isMappedList` return true only if `field.annotations[LIST_MAPPED_BY_FIELD]` is evaluated to true (and then it could be identified also in `transformMapsToLists`, or because `value[LIST_MAPPED_BY_FIELD]` is evaluated to true, but that would happen after we previously identified that `field.annotations[LIST_MAPPED_BY_FIELD]` is evaluated to true.

Please LMK if I am missing something here

dantal4: We run `transformMapsToLists` with the hardcoded standard types, so we don't have the `map_key_field` annotation on fields there. We do it in order to have lists in the type (and inner types) and not maps, which we have in the `refType` of the input element.
Because of that, we somehow need to know that a map should be transformed back to list, without having the annotation on the fields. That is why we add the `map_key_field` to `value`.

packages/netsuite-adapter/src/mapped_lists/utils.ts
```diff
@@ -191,13 +190,10 @@ export const convertAnnotationListsToMaps = async (
 export const isMappedList = async (
```
omrilit, line 190: Is this one used outside of this file? maybe we can remove the export?

packages/netsuite-adapter/src/mapped_lists/utils.ts
omrilit: OK now I understand it.
I would add a comment explaining it, maybe here or in row 305.

Is it better than using the refType and assuming it's a list?
I guess your solution is easier, but efficiency wise, WDYT?

packages/netsuite-adapter/src/mapped_lists/utils.ts
dantal4: Only in the test.

packages/netsuite-adapter/src/mapped_lists/utils.ts
dantal4: I will add a comment.
We can't use the `refType` when it's a map, because it makes `transformElement` to act different, and it messes all the element value/annotations.

omrilit (approved):

PR #4088, opened 2023-03-13
SALTO-3766 Use elementSource in remove_sdf_elements

dantal4:
Use `elementSource.list()` instead of `elementSourceIndex` to see if there are existing instances for removed custom record type.

---

---
_Release Notes_: 
Netsuite Adapter:
- Use elementSource in remove_sdf_elements

---
_User Notifications_: 
None

### DIFF
--- a/packages/netsuite-adapter/src/change_validator.ts
+++ b/packages/netsuite-adapter/src/change_validator.ts
@@ -15,7 +15,7 @@
 */
 import _ from 'lodash'
 
-import { ChangeError, getChangeData, ChangeValidator, Change, ChangeDataType, isFieldChange, isAdditionOrRemovalChange } from '@salto-io/adapter-api'
+import { ChangeError, getChangeData, ChangeValidator, Change, ChangeDataType, isFieldChange, isAdditionOrRemovalChange, ReadOnlyElementsSource } from '@salto-io/adapter-api'
 import { deployment } from '@salto-io/adapter-components'
 import accountSpecificValuesValidator from './change_validators/account_specific_values'
 import dataAccountSpecificValuesValidator from './change_validators/data_account_specific_values'
@@ -44,7 +44,6 @@ import currencyUndeployableFieldsValidator from './change_validators/currency_un
 import NetsuiteClient from './client/client'
 import { AdditionalDependencies } from './config'
 import { Filter } from './filter'
-import { LazyElementsSourceIndexes } from './elements_source_index/types'
 import { NetsuiteChangeValidator } from './change_validators/types'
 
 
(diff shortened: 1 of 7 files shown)


packages/netsuite-adapter/src/change_validator.ts
```diff
@@ -145,7 +144,7 @@ const getChangeValidator: ({
 
       const validatorChangeErrors: ChangeError[] = _.flatten(await Promise.all([
         ...changeValidators.map(validator => validator(changes, elementSource)),
-        ...validators.map(validator => validator(changes, deployReferencedElements, elementsSourceIndex)),
+        ...validators.map(validator => validator(changes, deployReferencedElements, elementsSource)),
```
adamro, line 147: Notice the use of `buildLazyShallowTypeResolverElementsSource` in Salesforce for a similar use-case cause if you need resolved types in the Change Validators and not just list for example it's the optimal way

dantal4: We need only the list, so the regular `elementSource` is fine here 👍

adamro: Maybe better to use it in advanced so when someone needs not just the list it will be there :)

packages/netsuite-adapter/src/change_validators/remove_sdf_elements.ts
```diff
@@ -55,10 +54,10 @@ const validateRemovableChange = async (
       }
     }
     // will be redundant once SALTO-3534 introduced
-    if (elementsSourceIndex !== undefined) {
-      const recordTypeIndexInstances = Object.keys((await elementsSourceIndex.getIndexes()).internalIdsIndex)
-        .some(elemInternalId => elemInternalId.startsWith(element.elemID.typeName.concat('-')))
-      if (recordTypeIndexInstances) {
+    if (elementsSource !== undefined) {
+      const hasCustomRecordInstances = await awu(await elementsSource.list())
+        .some(elemId => elemId.idType === 'instance' && elemId.typeName === element.elemID.name)
```
omrilit, line 59: How bad the performance of elementsSource.list()?
I wonder whether this is something we should do in one pass for all customRecordTypes removals.
In the current implementation, if you have 10 deletions of customrecordtypes you go over elementsSource.list() 10 times.

omrilit (commented):
LGTM!

packages/netsuite-adapter/src/change_validators/remove_sdf_elements.ts
dantal4: IDK. We can add a log.time to see that, and change in the future if we see it's more efficient.

dantal4: You were right- it's much efficient to run over `elementsSource.list()` only once - I committed a fix.

omrilit (approved):

PR #4090, opened 2023-03-13
SALTO-7487 fix hash calculation for quick deploy

Uri-Bar:
SALTO-7487 fix hash calculation for quick deploy

---
calc the hash on the xml package instead of the changes

---

_Release Notes_: 
None

---

_User Notifications_: 
None

### DIFF
--- a/packages/salesforce-adapter/src/metadata_deploy.ts
+++ b/packages/salesforce-adapter/src/metadata_deploy.ts
@@ -300,7 +300,7 @@ const quickDeployOrDeploy = async (
     try {
       return await client.quickDeploy(quickDeployParams.requestId)
     } catch (e) {
-      log.info(`preforming regular deploy instead of quick deploy due to error: ${e.message}`)
+      log.warn(`preforming regular deploy instead of quick deploy due to error: ${e.message}`)
     }
   }
   return client.deploy(pkgData, { checkOnly })
@@ -332,7 +332,6 @@ export const deployMetadata = async (
 
   const pkgData = await pkg.getZip()
   const planHash = hashUtils.toMD5(pkgData)
-
   if (quickDeployParams !== undefined) {
     if (quickDeployParams.hash !== planHash) {
       return {
--- a/packages/salesforce-adapter/src/metadata_deploy.ts
+++ b/packages/salesforce-adapter/src/metadata_deploy.ts
@@ -14,9 +14,10 @@
 * limitations under the License.
 */
 import _ from 'lodash'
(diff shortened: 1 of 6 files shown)


packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -359,13 +359,21 @@ export const quickDeploy = async (
...
   if (validChanges.length === 0) {
     // Skip deploy if there are no valid changes
     return { appliedChanges: [], errors: validationErrors }
   }
-  if (quickDeployParams.hash !== calculateChangesHash(validChanges)) {
+  await awu(validChanges).forEach(async change => {
+    await addChangeToPackage(pkg, change, nestedMetadataTypes)
+  })
+
+  const pkgData = await pkg.getZip()
```
ori-moisis, line 375: building the package in a different flow than the one used for deploy is a mistake IMO.
this is creating implicit coupling between the flow here and the flow in `deployMetadata`.

I think we should not duplicate the logic

packages/salesforce-adapter/src/transformers/xml_transformer.ts
```diff
@@ -448,6 +448,13 @@ export const createDeployPackage = (deleteBeforeUpdate?: boolean): DeployPackage
       if (deleteManifest.size !== 0) {
         zip.file(`${PACKAGE}/${deletionsPackageName}`, toPackageXml(deleteManifest))
       }
+      // as part of the QuickDeploy feature
+      // we need a fixed date in  the timeStamp order to the hash to be consistent
+      // we chose a significant date for Salto
```
ori-moisis, line 453:
```suggestion
      // Set a constant date for all files in the zip in order to keep the zip hash constant when
      // the contents are the same.
      // this is important for the "quickDeploy" feature
```

packages/salesforce-adapter/src/transformers/xml_transformer.ts
```diff
@@ -448,6 +448,13 @@ export const createDeployPackage = (deleteBeforeUpdate?: boolean): DeployPackage
...
+      const date = new Date(2023, 6, 15)
+      Object.values(zip.files).forEach(info => {
+        info.date = date
+      })
```
ori-moisis, line 457: can we add a line break here?
it would be easier to visually separate logical blocks (e.g - the comment above applies to this, but does not apply to the `return` statement below)

ori-moisis (approved):
looks good

PR #4092, opened 2023-03-13
Salto-2329: generate generated dependencies for suite scripts

Shulik95:
_generate generated dependencies from suite scripts content_

---

---
_Release Notes_: 
_references from suitescripts will now be included as a generated dependency in salto_

---
_User Notifications_: 
_This will add new generated dependencies for existing suitescript nacls_


packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -13,19 +13,27 @@
...
 import { isSdfCreateOrUpdateGroupId } from '../group_changes'
 import { getLookUpName } from '../transformer'
+import { getGroupItemFromRegex } from '../client/sdf_client'
+import { getContent } from '../suiteapp_file_cabinet'
 
 const { awu } = collections.asynciterable
+const { isDefined } = values
+const NETSUITE_MODULE_PREFIX = 'N/'
+const OPTIONAL_REFS = 'optionalReferences'
+const semanticReferenceRegex = new RegExp(`(?<![a-zA-Z])(?<${OPTIONAL_REFS}>("[^"]*"|'[^']*'))`, 'gm')
```
dantal4, line 35: 1. why is `(?<![a-zA-Z])` needed?
2. this is going to catch empty strings too, is it fine?
3. can you add a comment above with some examples?

packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -13,19 +13,27 @@
...
+const nsConfigRegex = new RegExp(`\\*\\s@N\\w+\\s*(?<${OPTIONAL_REFS}>.*)`, 'gm')
```
dantal4, line 36: can you add a comment above with some examples?

packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -108,6 +116,51 @@ const generateServiceIdToElemID = async (
...
+  (isDefined(absolutePath) ? osPath.resolve(osPath.dirname(absolutePath), relativePath) : '')
+
+const updateDependenciesToAdd = (
+  foundReferences: string[],
+  fetchedElementsServiceIdToElemID: ServiceIdRecords,
+  elementsSourceServiceIdToElemID: ServiceIdRecords,
+  element: InstanceElement,
+): ElemID[] =>
+  foundReferences
+    .map(ref => (isFilePath(ref) ? resolveRelativePath(element.value?.path, ref) : ref))
```
dantal4, line 132:
```suggestion
    .map(ref => (isFilePath(ref) ? resolveRelativePath(element.value.path, ref) : ref))
```
also, what about folders or files without extension?

packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -108,6 +116,51 @@ const generateServiceIdToElemID = async (
   .map(elem => getElementServiceIdRecords(elem))
   .reduce<ServiceIdRecords>((acc, records) => Object.assign(acc, records), {})
 
+const isFilePath = (path: string): boolean =>
+  osPath.extname(osPath.basename(path)) !== ''
+
+const resolveRelativePath = (absolutePath: string | undefined, relativePath: string): string =>
```
dantal4, line 122: When will `absolutePath` be `undefined`? That is running on file/folder instances only, right?

packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -108,6 +116,51 @@ const generateServiceIdToElemID = async (
...
+  foundReferences: string[],
+  fetchedElementsServiceIdToElemID: ServiceIdRecords,
+  elementsSourceServiceIdToElemID: ServiceIdRecords,
+  element: InstanceElement,
+): ElemID[] =>
+  foundReferences
+    .map(ref => (isFilePath(ref) ? resolveRelativePath(element.value?.path, ref) : ref))
+    .map(ref => {
+      const serviceIdRecord = fetchedElementsServiceIdToElemID[ref] ?? elementsSourceServiceIdToElemID[ref]
+      if (isDefined(serviceIdRecord) && typeof serviceIdRecord !== 'function') {
```
dantal4, line 135: how will it be a function?

packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -108,6 +116,51 @@ const generateServiceIdToElemID = async (
...
+      }
+      return undefined
+    })
+    .filter(isDefined)
+
+
+const getSuiteScriptReferences = async (
+  element: InstanceElement,
+  fetchedElementsServiceIdToElemID: ServiceIdRecords,
+  elementsSourceServiceIdToElemID: ServiceIdRecords
```
dantal4, line 146: nit - these maps can be unified and used as one map all over the file.
In line 258:
```
const serviceIdToElemID = Object.assign(
   await generateServiceIdToElemID(elements),
   await createElementsSourceServiceIdToElemID(elementsSourceIndex, isPartial)
)
```

packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -108,6 +116,51 @@ const generateServiceIdToElemID = async (
...
+): Promise<ElemID[]> => {
+  const content = (await getContent(element.value.content)).toString()
+
+  const nsConfigReferences = getGroupItemFromRegex(content, nsConfigRegex, OPTIONAL_REFS)
+  const semanticReferences = getGroupItemFromRegex(content, semanticReferenceRegex, OPTIONAL_REFS)
+    .map(semanticRef => semanticRef.slice(1, -1))
```
dantal4, line 152: nit - change the regex to take only the inside part of a string, so this won't be necessary.

packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -108,6 +116,51 @@ const generateServiceIdToElemID = async (
   .map(elem => getElementServiceIdRecords(elem))
   .reduce<ServiceIdRecords>((acc, records) => Object.assign(acc, records), {})
 
+const isFilePath = (path: string): boolean =>
+  osPath.extname(osPath.basename(path)) !== ''
+
+const resolveRelativePath = (absolutePath: string | undefined, relativePath: string): string =>
+  (isDefined(absolutePath) ? osPath.resolve(osPath.dirname(absolutePath), relativePath) : '')
+
+const updateDependenciesToAdd = (
```
dantal4, line 125: maybe `getServiceElemIDsFromPaths` ?

packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -108,6 +116,51 @@ const generateServiceIdToElemID = async (
...
+  element: InstanceElement,
+  fetchedElementsServiceIdToElemID: ServiceIdRecords,
+  elementsSourceServiceIdToElemID: ServiceIdRecords
+): Promise<ElemID[]> => {
+  const content = (await getContent(element.value.content)).toString()
+
+  const nsConfigReferences = getGroupItemFromRegex(content, nsConfigRegex, OPTIONAL_REFS)
+  const semanticReferences = getGroupItemFromRegex(content, semanticReferenceRegex, OPTIONAL_REFS)
+    .map(semanticRef => semanticRef.slice(1, -1))
+    .filter(path => !path.startsWith(NETSUITE_MODULE_PREFIX))
```
dantal4, line 153: Should we maybe validate that the path starts with one of `/`, `./`, `../` ?
Because a string can be "abc" and by chance there can be a file "abc" in the same folder, and we'll have a false positive.
@omrilit WDYT?

packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -152,6 +205,11 @@ const replaceReferenceValues = async (
 
     return returnValue
   }
+  if (isFileCabinetInstance(element) && isFileInstance(element)) {
```
dantal4, line 208: `isFileInstance` should be enough.

dantal4 (commented):
Great work!

packages/netsuite-adapter/src/filters/element_references.ts
omrilit: Sounds like it would be easy to enforce it and better safe than sorry.
Let's do it!

packages/netsuite-adapter/src/filters/element_references.ts
Shulik95: 1. It is needed to avoid cases where an apostrophe appears as a part of a word (in possession or  plural) and doesn't really wrap word.
2. Empty strings are fine, all strings which aren't a scripid or a path won't have a value for them in the map and will be filtered out.
3. will do 👌

packages/netsuite-adapter/src/filters/element_references.ts
Shulik95: In cases where the reference is a string that matches a function name of Object.prototype the maps return a function, encountered it when the string 'toString' appeared in one of the files.

packages/netsuite-adapter/src/filters/element_references.ts
Shulik95: `isFileInstance` doesn't serve as a typeguard by itself so I think `isFileCabinetInstance` is needed here

packages/netsuite-adapter/src/filters/element_references.ts
dantal4: Wow. I didn't know it works that way haha. Maybe this is a nicer way-
```suggestion
      if (_.isPlainObject(serviceIdRecord)) {
```

packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -108,10 +121,47 @@ const generateServiceIdToElemID = async (
...
+const resolveRelativePath = (absolutePath: string, relativePath: string): string =>
+  osPath.resolve(osPath.dirname(absolutePath), relativePath)
+
+const getServiceElemIDsFromPaths = (
+  foundReferences: string[],
+  serviceIdToElemID: ServiceIdRecords,
+  element: InstanceElement,
+): ElemID[] =>
+  foundReferences
+    .map(ref => (pathPrefixRegex.test(ref) ? resolveRelativePath(element.value.path, ref) : ref))
```
dantal4, line 133:
```suggestion
    .map(ref => (pathPrefixRegex.test(ref) ? resolveRelativePath(element.value[PATH], ref) : ref))
```

dantal4 (approved):
Nice work!
Approving in advanced, but please fix the comments 🙏

packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -13,19 +13,32 @@
...
+import { getGroupItemFromRegex } from '../client/utils'
+import { getContent } from '../suiteapp_file_cabinet'
 
 const { awu } = collections.asynciterable
+const { isDefined } = values
+const NETSUITE_MODULE_PREFIX = 'N/'
+const OPTIONAL_REFS = 'optionalReferences'
+// matches strings in single/double quotes (paths and scriptids) where the apostrophes aren't a part of a word
+// e.g: 'custrecord1' "./someFolder/someScript.js"
+const semanticReferenceRegex = new RegExp(`(?<![a-zA-Z])("|')(?<${OPTIONAL_REFS}>.*?)\\1[\\s,\\]]`, 'gm')
```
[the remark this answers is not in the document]
Shulik95: I couldn't make it work with `\b`, but the regex can be simplified to  - `("|')(?<${OPTIONAL_REFS}>.*?)\\1` which isn't affected by whatever character comes before/after and only captures the value in quotes

PR #4095, opened 2023-03-14
SALTO-3405 - test failure stack trace can be string or object

benisraeladi:
Apex test failure showed stack trace as [object Object], so changed stack trace interface to be `string | object`

---

---
_Release Notes_: 
Salesforce:
* Stack traces for apex test failures will no longer show as [object Object]

---
_User Notifications_: _None_


tomsellek (approved):
LGTM!

PR #4097, opened 2023-03-14
SALTO-3825 - CLI: Deprecation message is shown when using the account command

tomsellek:
Flipped the account and service actions for some reason.

---

---
_Release Notes_: 
CLI: 'account' command will no longer show a deprecation warning.

---
_User Notifications_: 
N/A


yuribro (approved):
Approving to save cycles, but is there a way to catch this with tests?

tomsellek:
> Approving to save cycles, but is there a way to catch this with tests?

@yuribro presumably yes, but right now the tests use the CLI action directly, so the change will have to be pretty significant.

PR #4099, opened 2023-03-14
add logs on sdf errors

dantal4:
add logs on feature map update & log content of manifest.xml & deploy.xml on unknown sdf error.

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/netsuite-adapter/src/client/client.ts
```diff
@@ -326,9 +331,6 @@ export default class NetsuiteClient {
           }
         }
         if (error instanceof MissingManifestFeaturesError) {
-          if (_.isEmpty(error.missingFeatures)) {
-            return { errors, appliedChanges: [] }
-          }
```
dantal4, line 331: This was not reachable (we don't throw `MissingManifestFeaturesError` if `missingFeatures` is empty).

omrilit (approved):

PR #4101, opened 2023-03-14
SALTO-3820 - Refactor RemoteMap location cache pool

tomsellek:
Replace the LRU-of-LRUs with a refcounted pool for location caches

---

This is the first step in our journey towards proper resource management in the RemoteMap implementation.
This change makes the locationCache pool no longer be LRU. If we determine this is a problem we can modify the implementation as needed.
https://www.notion.so/saltoio/Notes-SALTO-2490-34a1b434f10a4eacb74a441d8339e6ea

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/core/src/local-workspace/remote_map.ts
+++ b/packages/core/src/local-workspace/remote_map.ts
@@ -208,8 +208,8 @@ const readonlyDBConnections: ConnectionPool = {}
 const tmpDBConnections: Record<string, ConnectionPool> = {}
 const readonlyDBConnectionsPerRemoteMap: Record<string, ConnectionPool> = {}
 let currentConnectionsCount = 0
-/* eslint-disable-next-line @typescript-eslint/no-explicit-any */
-const locationCaches = new LRU<string, LRU<string, any>>({ max: 10 })
+
+
 const COUNTER_TYPES = [
   'LocationCacheCreated',
   'LocationCacheReuse',
@@ -224,6 +224,59 @@ const COUNTER_TYPES = [
 ] as const
 type CounterType = typeof COUNTER_TYPES[number]
 const counters: Record<string, Record<CounterType, number>> = {}
+
+/* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+class LocationCache extends LRU<string, any> {
+  readonly location: string
+
+  constructor(location: string, cacheSize: number) {
+    super({ max: cacheSize })
+    this.location = location
+  }
+}
+
+type LocationCachePool = {
+  get: (location: string, cacheSize: number) => LocationCache
+
+  // The 'string' overload is temporary to allow the implementation of closeRemoteMapsOfLocation.
+  // Once we remove closeRemoteMapOfLocation, the 'string' overload should be removed.
+  put: (cache: LocationCache | string) => void
+}
+
+const createLocationCachePool = (): LocationCachePool => {
+  // TODO: LRU if we determine too many locationCaches are created.
+  const pool = new Map<string, {cache: LocationCache; refcnt: number}>()
+  return {
+    get: (location, cacheSize) => {
+      const cachePoolEntry = pool.get(location)
+      if (cachePoolEntry) {
+        counters[location].LocationCacheReuse += 1
+        cachePoolEntry.refcnt += 1
+        return cachePoolEntry.cache
+      }
+      counters[location].LocationCacheCreated += 1
+      const newCache:LocationCache = new LocationCache(location, cacheSize)
+      pool.set(location, { cache: newCache, refcnt: 1 })
+      return newCache
+    },
+    put: cacheOrLocation => {
+      const location = _.isString(cacheOrLocation) ? cacheOrLocation : cacheOrLocation.location
+      const poolEntry = pool.get(location)
+      if (!poolEntry || poolEntry.refcnt === 0) {
+        log.error('Bug in LocationCachePool refcounting!')
+        return
+      }
+      poolEntry.refcnt -= 1
+
+      if (poolEntry.refcnt === 0) {
+        pool.delete(location)
+      }
+    },
+  }
+}
+
+/* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+const locationCaches = createLocationCachePool()
 
 const deleteLocation = async (location: string): Promise<void> => {
   try {
@@ -287,7 +340,7 @@ export const closeRemoteMapsOfLocation = async (location: string): Promise<void>
     delete readonlyDBConnectionsPerRemoteMap[location]
     log.debug('closed read-only connections per remote map of location %s', location)
   }
-  locationCaches.del(location)
+  locationCaches.put(location)
   log.debug('Remote Map Stats for location \'%s\': %o', location, counters[location])
   delete counters[location]
 }
@@ -457,22 +510,7 @@ remoteMap.RemoteMapCreator => {
     ) as Record<CounterType, number>
   }
 
-  // Note: once we set a non-zero cache size,
-  // we won't change the cache size even if we give different value
-  /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
-  let locationCache: LRU<string, any>
-  if (locationCaches.has(location)) {
-    counterInc('LocationCacheReuse')
-    /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
-    locationCache = locationCaches.get(location) as LRU<string, any>
-  } else {
-    counterInc('LocationCacheCreated')
-    /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
-    locationCache = new LRU<string, any>({ max: cacheSize })
-    if (cacheSize > 0) {
-      locationCaches.set(location, locationCache)
-    }
-  }
+  const locationCache = locationCaches.get(location, cacheSize)
 
   let persistentDB: rocksdb
   let tmpDB: rocksdb
--- a/packages/core/src/local-workspace/remote_map/counters.ts
+++ b/packages/core/src/local-workspace/remote_map/counters.ts
@@ -37,6 +37,10 @@ export const counterInc = (location: string, counter: CounterType): void => {
   counters[location][counter] += 1
 }
 
+export const counterValue = (location: string, counter: CounterType): number => (
+  counters[location][counter]
+)
+
 export const countersInit = (location: string): void => {
   counters[location] = Object.fromEntries(
     COUNTER_TYPES.map(counterName => [counterName, 0])
--- a/packages/core/src/local-workspace/remote_map/counters.ts
+++ b/packages/core/src/local-workspace/remote_map/counters.ts
@@ -15,7 +15,9 @@
 */
 
 import { logger } from '@salto-io/logging'
+import { collections } from '@salto-io/lowerdash'
 
+const { DefaultMap } = collections.map
 const log = logger(module)
 
 const COUNTER_TYPES = [
@@ -31,23 +33,46 @@ const COUNTER_TYPES = [
   'TmpDbConnectionReuse',
 ] as const
 type CounterType = typeof COUNTER_TYPES[number]
-export const counters: Record<string, Record<CounterType, number>> = {}
 
-export const counterInc = (location: string, counter: CounterType): void => {
-  counters[location][counter] += 1
+type Counter = {
+  inc: () => void
+  value: () => number
 }
 
-export const counterValue = (location: string, counter: CounterType): number => (
-  counters[location][counter]
-)
-
-export const countersInit = (location: string): void => {
-  counters[location] = Object.fromEntries(
-    COUNTER_TYPES.map(counterName => [counterName, 0])
-  ) as Record<CounterType, number>
+type LocationCounters = Record<CounterType, Counter> & {
+  dump: () => void
 }
 
-export const countersDumpAndClose = (location: string): void => {
-  log.debug('Remote Map Stats for location \'%s\': %o', location, counters[location])
-  delete counters[location]
+type StatCounters = {
+  locationCounters: (location: string) => LocationCounters
+  deleteLocation: (location: string) => void
 }
+
+const createCounter = (): Counter => {
+  let counterValue = 0
+  return {
+    inc: () => { counterValue += 1 },
+    value: () => counterValue,
+  }
+}
+
+const createLocationCounters = (location: string): LocationCounters => {
+  const counters = Object.fromEntries([
+    ...COUNTER_TYPES.map(counterType => [counterType, createCounter()]),
+    ['dump', () => {
+      log.debug('Remote Map Stats for location \'%s\': %o',
+        location, Object.fromEntries(COUNTER_TYPES.map(counterType => [counterType, counters[counterType]])))
+    }],
+  ])
+  return counters
+}
+
+const createStatCounters = (): StatCounters => {
+  const locations = new DefaultMap<string, LocationCounters>(createLocationCounters)
+  return {
+    locationCounters: location => locations.get(location),
+    deleteLocation: location => { locations.delete(location) },
+  }
+}
+
+export const counters = createStatCounters()
--- a/packages/core/src/local-workspace/remote_map/counters.ts
+++ b/packages/core/src/local-workspace/remote_map/counters.ts
@@ -1,0 +1,49 @@
+/*
+*                      Copyright 2023 Salto Labs Ltd.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+import { logger } from '@salto-io/logging'
+
+const log = logger(module)
+
+const COUNTER_TYPES = [
+  'LocationCacheCreated',
+  'LocationCacheReuse',
+  'LocationCacheHit',
+  'LocationCacheMiss',
+  'RemoteMapCreated',
+  'RemoteMapHit',
+  'RemoteMapMiss',
+  'PersistentDbConnectionCreated',
+  'PersistentDbConnectionReuse',
+  'TmpDbConnectionReuse',
+] as const
+type CounterType = typeof COUNTER_TYPES[number]
+export const counters: Record<string, Record<CounterType, number>> = {}
+
+export const counterInc = (location: string, counter: CounterType): void => {
+  counters[location][counter] += 1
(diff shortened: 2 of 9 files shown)


packages/core/src/local-workspace/remote_map.ts
```diff
@@ -225,6 +225,59 @@ const COUNTER_TYPES = [
...
+      counters[location].LocationCacheCreated += 1
+      const newCache:LocationCache = new LocationCache(location, cacheSize)
+      pool.set(location, { cache: newCache, refcnt: 1 })
+      return newCache
+    },
+    put: cacheOrLocation => {
+      const location = _.isString(cacheOrLocation) ? cacheOrLocation : cacheOrLocation.location
+      const poolEntry = pool.get(location)
+      if (!poolEntry || poolEntry.refcnt === 0) {
+        log.error('Bug in LocationCachePool refcounting!')
```
yuribro, line 266: I think this error is a bit misleading, it can be also a bug in the caller - a double free. 
To make the debugging easier, maybe add a debug print each time we get or put a location?

packages/core/src/local-workspace/remote_map.ts
```diff
@@ -225,6 +225,59 @@ const COUNTER_TYPES = [
...
+
+type LocationCachePool = {
+  get: (location: string, cacheSize: number) => LocationCache
+
+  // The 'string' overload is temporary to allow the implementation of closeRemoteMapsOfLocation.
+  // Once we remove closeRemoteMapOfLocation, the 'string' overload should be removed.
+  put: (cache: LocationCache | string) => void
+}
+
+const createLocationCachePool = (): LocationCachePool => {
```
yuribro, line 246: I think we should add UTs for this, this looks very testable

packages/core/src/local-workspace/remote_map.ts
```diff
@@ -225,6 +225,59 @@ const COUNTER_TYPES = [
...
+  // TODO: LRU if we determine too many locationCaches are created.
```
yuribro, line 247: To determine this, we need to know how many locationCaches are in use at the same time. Maybe add a LocationCacheFreed counter? or a metric/log of the current size of the cache on each LocationCacheCreated event?

packages/core/src/local-workspace/remote_map.ts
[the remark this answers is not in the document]
tomsellek: Ack.

tomsellek: Ack.

packages/core/src/local-workspace/remote_map.ts
tomsellek: Ack.

packages/core/src/local-workspace/remote_map.ts
tomsellek: Ack.

packages/core/index.ts
```diff
@@ -60,5 +60,5 @@ export {
   closeRemoteMapsOfLocation,
   replicateDB,
   createReadOnlyRemoteMapCreator,
-} from './src/local-workspace/remote_map'
+} from './src/local-workspace/remote_map/remote_map'
```
yuribro, line 63: Maybe re-export these from `remote_map/index.ts`? This will signal what is really exported what is only for tests

packages/core/test/workspace/local/remote_map/location_cache.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    expect(counterValue(LOCATION1, 'LocationCacheCreated')).toEqual(1)
+    expect(counterValue(LOCATION1, 'LocationCacheReuse')).toEqual(0)
+    pool.put(cache)
+  })
+
+  it('should reuse caches where possible', () => {
+    const caches = _.times(10, () => pool.get(LOCATION1, 5000))
+    expect(counterValue(LOCATION1, 'LocationCacheCreated')).toEqual(1)
+    expect(counterValue(LOCATION1, 'LocationCacheReuse')).toEqual(caches.length - 1)
+    caches.forEach(cache => pool.put(cache))
```
yuribro, line 47: Maybe check that the cache is empty at the end?

packages/core/index.ts
tomsellek: Ooh, nice

packages/core/src/local-workspace/remote_map/counters.ts
```diff
@@ -0,0 +1,53 @@
...
+  'LocationCacheMiss',
+  'RemoteMapCreated',
+  'RemoteMapHit',
+  'RemoteMapMiss',
+  'PersistentDbConnectionCreated',
+  'PersistentDbConnectionReuse',
+  'TmpDbConnectionReuse',
+] as const
+type CounterType = typeof COUNTER_TYPES[number]
+export const counters: Record<string, Record<CounterType, number>> = {}
```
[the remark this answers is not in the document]
tomsellek: > Is there a possibility to not keep it as a global? In general, it looks like we moved to keeping only per-location counters, so maybe keep it with (inside) the location cache.
We don't only keep stats that are related to locationCache, though.
> And this approach is very procedural, why to return a more OOP interface where we don't have to pass the location each time?
Indeed
>Also, maybe use DefaultMap<CounterType, number> instead, so there is not need to init?
Yuppers

packages/core/src/local-workspace/remote_map/location_cache.ts
```diff
@@ -0,0 +1,80 @@
...
+  get: (location: string, cacheSize: number) => LocationCache
+
+  // The 'string' overload is temporary to allow the implementation of closeRemoteMapsOfLocation.
+  // Once we remove closeRemoteMapOfLocation, the 'string' overload should be removed.
+  put: (cache: LocationCache | string) => void
+}
+export const createLocationCachePool = (): LocationCachePool => {
+  // TODO: LRU if we determine too many locationCaches are created.
+  const pool = new Map<string, { cache: LocationCache; refcnt: number }>()
+  let maxPoolSize = 0
```
yuribro, line 43: nit: This sounds like a limitation on the size, not the observed max

packages/core/test/workspace/local/remote_map/location_cache.test.ts
yuribro: Good question.

1. You can try to put it again, and see that we get the error
2. I guess we could expose more of the pool for testing, either the  pool itself or "length", but I the implementation is simple enough that we probably shouldn't for the sake of testing
3. You could mock `new Map`, but maybe it's an overkill

tomsellek: I dunno, I'm not crazy about relying on logs in tests, and as you said the other two alternatives seem like overkill for this sort of thing

tomsellek: One way to make things more testable would be to convert the cache pool to a class, thereby making its state accessible by the tests. However, it seems like we generally prefer closures to classes in our code. @ori-moisis WDYT?

ori-moisis: this is indeed an issue with non-functional code and more so with global singletons.
there are several ways to work around this, I think the most common one we use to make things testable is dependency injection.
in this case, the easiest change to make in order to have this testable is to let `createLocationCachePool` get a map as input.
the test instance can keep a reference to the map it passes into `createLocationCachePool` and the tests can check the status of this map. it creates coupling of the implementation and the tests which is not great, but it is a decent compromise between not exposing unneeded capabilities and still making the code testable.
(there can be other injections that can be used here like injecting the "create" and "release" functions, but I think that would be too generic for this...)

in this case btw, it sounds like the desired behavior is that when the reference count reaches zero, a location will be removed from the map.
so, this should be a specific test, not mixed with a test that checks the "reuse" behavior of the same location.
it can be under the same `describe` block in the spirit of
```typescript
describe('when the same location is used multiple times', () => {
  it('should create one cache for the location', () => {}) // check the number of caches before closing
  it('should close the location after all connections are released', () => {}) // check the number of caches after closing
})
```
(my terminology might be off, use the terms that are consistent with the code)

btw, I think using the counters here is not a good test because it doesn't actually check the desired behavior, it only checks a proxy / side effect.
there can be more tests that check the behavior of the counters specifically, they can use the same scenario as the other tests, e.g
```typescript
describe('when the same location is used multiple times', () => {
  beforeEach(() => {
    // open a bunch of connections
  })
  it('should create one cache for the location', () => {}) // check the number of caches
  it('should close the location after all connections are released', () => {}) // closes the location and then check the number of caches
  it('should increase the LocationCacheCreated counter once', () => {})
  it('should increase the LocationCacheReuse counter for each use after the first', () => {}) 
})
```

packages/core/src/local-workspace/remote_map/location_cache.ts
```diff
@@ -0,0 +1,80 @@
...
+  return {
+    get: (location, cacheSize) => {
```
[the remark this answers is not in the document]
tomsellek: I absolutely agree. The only reason I'm doing things this way is because it mimics the existing code as closely as possible. Changing this interface would require changing the interface of createRemoteMapCreator, or ignoring the cacheSize parameter in remoteMapCreator. 
If you think it's worth it, I'm happy to do it.

packages/core/src/local-workspace/remote_map/remote_map.ts
```diff
@@ -57,7 +58,7 @@ let rocksdbImpl: any
 const getRemoteDbImpl = (): any => {
   if (rocksdbImpl === undefined) {
     // eslint-disable-next-line global-require, @typescript-eslint/no-var-requires
-    rocksdbImpl = require('./rocksdb').default
+    rocksdbImpl = require('../rocksdb').default
```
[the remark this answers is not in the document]
tomsellek: Love it.

packages/core/src/local-workspace/remote_map/location_cache.ts
```diff
@@ -0,0 +1,80 @@
...
+      const cachePoolEntry = pool.get(location)
+      if (cachePoolEntry) {
```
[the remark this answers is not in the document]
tomsellek: Ack

packages/core/src/local-workspace/remote_map/remote_map.ts
```diff
@@ -287,9 +273,9 @@ export const closeRemoteMapsOfLocation = async (location: string): Promise<void>
     delete readonlyDBConnectionsPerRemoteMap[location]
     log.debug('closed read-only connections per remote map of location %s', location)
   }
-  locationCaches.del(location)
-  log.debug('Remote Map Stats for location \'%s\': %o', location, counters[location])
-  delete counters[location]
+  locationCaches.put(location)
+  counters.locationCounters(location).dump()
+  counters.deleteLocation(location)
```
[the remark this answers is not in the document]
tomsellek: Counters are decoupled from location cache. The only relationship here is that if we `closeRemoteMapsOfLocation` then we need to close all the location caches and counters that are associated with said location (as well as the DB connections, etc, which will happen in the next PR).

packages/core/test/workspace/local/remote_map/location_cache.test.ts
tomsellek: > let createLocationCachePool get a map as input

But won't the caller need to know the type of values that are stored in the map, thereby exposing an internal implementation detail?

packages/core/src/local-workspace/remote_map/location_cache.ts
tomsellek: Welp, you just made me break out the term "watermark", which I haven't used in 20 years. Hope you're happy.

packages/core/test/workspace/local/remote_map/location_cache.test.ts
ori-moisis: that is part of the signature of the function
```
const createLocationCachePool(pool: Map<string, ???> = new Map()): ...
```
yes, it exposes some implementation detail but it should only be used by the tests (the rest of the code should not be familiar with the interface of the function anyway since the rest of the code uses the singleton) but as I wrote above, this is a compromise:

> it creates coupling of the implementation and the tests which is not great, but it is a decent compromise between not exposing unneeded capabilities and still making the code testable. 

it is the least exposure we've been able to figure out so far that still allows testing to be meaningful.
open to other ideas if you have them though

packages/core/src/local-workspace/remote_map/location_cache.ts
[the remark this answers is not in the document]
ori-moisis: from a quick search, doesn't seem like anything is actually passing that parameter

packages/core/src/local-workspace/remote_map/location_cache.ts
```diff
@@ -0,0 +1,80 @@
...
+*/
+import _ from 'lodash'
+import { logger } from '@salto-io/logging'
+import LRU from 'lru-cache'
+import { counters } from './counters'
+
+const log = logger(module)
+
+/* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+class LocationCache extends LRU<string, any> {
```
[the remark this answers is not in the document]
ori-moisis: I will gladly correct
types can be as complex as needed (i.e, if for some reason the map value can be either string or number, then `V = string | number`)

packages/core/src/local-workspace/remote_map/remote_map.ts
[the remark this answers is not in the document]
ori-moisis: there is a certain asymmetry here that usually points to a design issue.
location counters are increased inside the cache pool code, but the code that handles dumping the when the cache is deleted happens outside the cache pool code.
it is unclear which code is supposed to be responsible for handling the counters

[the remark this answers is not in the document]
tomsellek: Counters are also increased inside remoteMapCreator and (in the next PR) in the DB connection pool.
Note that the reason for this whole chain of changes is to get rid of `closeRemoteMapsOfLocation` - so its implementation will be awkward during the transition, but making the design of `closeRemoteMapsOfLocation` elegant was never a design goal.

packages/core/test/workspace/local/remote_map/location_cache.test.ts
tomsellek: Not really. Done.

packages/core/src/local-workspace/remote_map/location_cache.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export type LocationCachePool = {
+  get: (location: string, cacheSize: number) => LocationCache
+
+  // The 'string' overload is temporary to allow the implementation of closeRemoteMapsOfLocation.
+  // Once we remove closeRemoteMapOfLocation, the 'string' overload should be removed.
+  put: (cache: LocationCache | string) => void
+}
+export const createLocationCachePool = (): LocationCachePool => {
+  // TODO: LRU if we determine too many locationCaches are created.
```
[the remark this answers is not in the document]
tomsellek: I could add another counter to the existing code and see what the highest watermark is

packages/core/src/local-workspace/remote_map/location_cache.ts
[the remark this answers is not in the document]
tomsellek: Aight, I went ahead and did it.

packages/core/src/local-workspace/remote_map/location_cache.ts
```diff
@@ -0,0 +1,88 @@
...
+export type LocationCachePoolContents = Map<string, { cache: LocationCache; refcnt: number }>
+
+const DEFAULT_LOCATION_CACHE_SIZE = 5000
+
+export const createLocationCachePool = (
+  initialContents?: LocationCachePoolContents,
+  cacheSize: number = DEFAULT_LOCATION_CACHE_SIZE
+): LocationCachePool => {
+  // TODO: LRU if we determine too many locationCaches are created.
+  const pool: LocationCachePoolContents = initialContents ?? new Map<string, { cache: LocationCache; refcnt: number }>()
```
ori-moisis, line 50: I think typescript will allow this so you don't have to repeat the type
```suggestion
  const pool: LocationCachePoolContents = initialContents ?? new Map()
```

ori-moisis (commented):
given the movements for rocksdb stuff, I suggest to test that the packaged CLI is still able to run fetch (the CI tests only run `salto -h` so I don't think they would catch an issue with the packaging of rocksdb).

packages/core/test/workspace/local/remote_map/location_cache.test.ts
```diff
@@ -0,0 +1,67 @@
...
+  })
+
+  it('should not reuse caches of a different location', () => {
+    const cache = pool.get(LOCATION1)
+    const anotherCache = pool.get(LOCATION2)
+    expect(cache.location).toEqual(LOCATION1)
+    expect(anotherCache.location).toEqual(LOCATION2)
+    expect(poolContents.size).toEqual(2)
+    expect(poolContents.get(LOCATION1)?.refcnt).toEqual(1)
+    expect(poolContents.get(LOCATION2)?.refcnt).toEqual(1)
```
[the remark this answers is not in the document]
tomsellek: 🤷

tomsellek:
> given the movements for rocksdb stuff, I suggest to test that the packaged CLI is still able to run fetch (the CI tests only run `salto -h` so I don't think they would catch an issue with the packaging of rocksdb).

Confirmed that the packaged CLI ran fetch successfully

packages/core/src/local-workspace/remote_map/remote_map.ts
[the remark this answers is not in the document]
ori-moisis: I think you are ok with this because you are operating under the assumption that no long-living process will ever open workspaces or remote maps.
I don't think that is the case in reality.
I think we can't allow such memory leaks in our code, when we close a workspace, it should not leave any allocated memory behind.

packages/core/src/local-workspace/remote_map/location_cache.ts
```diff
@@ -0,0 +1,88 @@
...
+        pool.delete(location)
+        if (pool.size === 0) {
+          log.debug('Max location cache pool size: %o', poolSizeWatermark)
+          poolSizeWatermark = 0
+        }
+      }
+    },
+  }
+}
+/* eslint-disable-next-line @typescript-eslint/no-explicit-any */
```
ori-moisis, line 87: unneeded

packages/core/src/local-workspace/remote_map/location_cache.ts
```diff
@@ -0,0 +1,88 @@
...
+* limitations under the License.
+*/
+import _ from 'lodash'
+import { logger } from '@salto-io/logging'
+import LRU from 'lru-cache'
+import { counters } from './counters'
+
+const log = logger(module)
+
+/* eslint-disable-next-line @typescript-eslint/no-explicit-any */
```
ori-moisis, line 23: unneeded

ori-moisis (approved):

ori-moisis:
can we close this one then?

tomsellek:
superseded by https://github.com/salto-io/salto/pull/4284

PR #4105, opened 2023-03-15
SALTO-3826: Salesforce Package CV bugfix

tamtamirr:
Bugfix

### DIFF
--- a/packages/salesforce-adapter/src/change_validators/package.ts
+++ b/packages/salesforce-adapter/src/change_validators/package.ts
@@ -52,23 +52,17 @@ export const hasNamespace = async (customElement: Element): Promise<boolean> => 
   return cleanFullName.includes(NAMESPACE_SEPARATOR)
 }
 
-export const getNamespace = async (customElement: Element): Promise<string | undefined> => {
-  const parts = (await apiName(customElement, true))?.split(NAMESPACE_SEPARATOR) ?? []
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/src/change_validators/package.ts
```diff
@@ -52,23 +52,17 @@ export const hasNamespace = async (customElement: Element): Promise<boolean> =>
...
 }
 
-export const getNamespace = async (customElement: Element): Promise<string | undefined> => {
-  const parts = (await apiName(customElement, true))?.split(NAMESPACE_SEPARATOR) ?? []
-  return parts.length > 1 ? parts[0] : undefined
-}
-
-
-export const PACKAGE_VERSION_FIELD_NAME = 'version_number'
+const getNamespace = async (customElement: Element): Promise<string> => (
```
amitgoldarb, line 55: Please add a comment that this will be fixed in the upcoming PR for installed packages.
Maybe also add a comment that this function assumes that hasNamespace has already been called on it.

amitgoldarb (approved):
Thanks for this!

packages/salesforce-adapter/src/change_validators/package.ts
amitgoldarb: I agree.
We discussed this one f2f and @tamtamirr preferred to align this in the installed packages PR.

PR #4110, opened 2023-03-15
SALTO-3713: Hide Okta, Zuora and Stripe types by default

shir-reifenberg:
Hide Okta, Zuora and Stripe types by default

---
_Release Notes_: 

_Stripe adapter_:
* The types folder will be hidden by default

_Okta adapter_:
* The types folder will be hidden by default

_Zuora-Billing adapter_:
* The system types folder will be hidden by default. Standard and custom object definitions will still be visible.

---

_User Notifications_: 

_Stripe adapter_:
* The types folder will be hidden by default

_Okta adapter_:
* The types folder will be hidden by default

_Zuora-Billing adapter_:
* The system types folder will be hidden by default. Standard and custom object definitions will still be visible.


netama (approved):
good catch! approving, let's add it in zuora+stripe as well as you mentioned :)

PR #4114, opened 2023-03-16
INCIDENT-3687: problem fix, start node changed to top level element

Shulik95:
_fix for incident-3687_

---

_Link to the incident - https://salto-io.atlassian.net/browse/INCIDENT-3687_

---
_Release Notes_: 
_When creating the graph, edges are now created between top level changes only_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/netsuite-adapter/src/client/client.ts
+++ b/packages/netsuite-adapter/src/client/client.ts
@@ -175,7 +175,7 @@ export default class NetsuiteClient {
         const serviceIdInfoArray = captureServiceIdInfo(val)
         serviceIdInfoArray.forEach(serviceIdInfo => {
           currSet.add(serviceIdInfo.serviceId)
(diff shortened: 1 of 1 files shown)


packages/netsuite-adapter/src/client/client.ts
```diff
@@ -175,7 +175,7 @@ export default class NetsuiteClient {
         const serviceIdInfoArray = captureServiceIdInfo(val)
         serviceIdInfoArray.forEach(serviceIdInfo => {
           currSet.add(serviceIdInfo.serviceId)
-          const startNode = dependencyGraph.findNodeByField('serviceid', serviceIdInfo.serviceId)
+          const startNode = dependencyGraph.findNodeByField('serviceid', serviceIdInfo.serviceId.split('.')[0])
```
omrilit, line 178: I think that it may become incorrect for FileCabinet path, right?

PR #4122, opened 2023-03-19
SALTO-3857/missing items bug

guyhamo:
missing items on workflow scheme changes caused a bug when we tried to iterate the items.
there was a check for it that missed this flow so i changed it as well.

---

---
_Release Notes_: 

---
_User Notifications_:

### DIFF
--- a/packages/jira-adapter/src/change_validators/workflow_scheme_migration.ts
+++ b/packages/jira-adapter/src/change_validators/workflow_scheme_migration.ts
@@ -95,7 +95,7 @@ const getDefaultWorkflowIssueTypes = (
   workflowScheme: InstanceElement,
   assignedIssueTypes: ReferenceExpression[]
 ): ReferenceExpression[] => {
-  workflowScheme.value.item
+  workflowScheme.value.items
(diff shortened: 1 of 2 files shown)


packages/jira-adapter/src/change_validators/workflow_scheme_migration.ts
```diff
@@ -95,8 +95,8 @@ const getDefaultWorkflowIssueTypes = (
   workflowScheme: InstanceElement,
   assignedIssueTypes: ReferenceExpression[]
 ): ReferenceExpression[] => {
-  workflowScheme.value.items
-    .filter(isWorkflowSchemeItem)
+  workflowScheme.value.item
```
IdoZakk, line 98: Seems like you dropped the `s` by mistake

IdoZakk (approved):
I approved, but I think there is an error there, look at the comment

packages/jira-adapter/src/change_validators/workflow_scheme_migration.ts
guyhamo: wow thank you

PR #4124, opened 2023-03-19
SALTO-3835/partial fetch path index bug

guyhamo:
on partial fetches there was a bug that removed the path index and top level path index entries that were not fetched in the partial fetch.
we already had a logic to maintain accounts not fetched that collected all of their elements and then pushed them back inside the index after we cleared it.
I used the same logic to update the partially fetched accounts with a slight change to how we merge the old entires and new entries so we enforce that the new changes make it into the index.

---

tested manually, looking into adding a ut.

---
_Release Notes_: 
core
* fixed bug with partial fetches removing entries from path index.

---
_User Notifications_:

### DIFF
--- a/packages/core/src/api.ts
+++ b/packages/core/src/api.ts
@@ -239,7 +239,8 @@ const updateStateWithFetchResults = async (
   workspace: Workspace,
   mergedElements: Element[],
   unmergedElements: Element[],
-  fetchedAccounts: string[]
+  fetchedAccounts: string[],
+  partiallyFetchedAccounts: string[],
 ): Promise<void> => {
   const fetchElementsFilter = shouldElementBeIncluded(fetchedAccounts)
   const stateElementsNotCoveredByFetch = await awu(await workspace.state().getAll())
@@ -247,8 +248,11 @@ const updateStateWithFetchResults = async (
   await workspace.state()
     .override(awu(mergedElements)
       .concat(stateElementsNotCoveredByFetch), fetchedAccounts)
-  await workspace.state().updatePathIndex(unmergedElements,
-    (await workspace.state().existingAccounts()).filter(key => !fetchedAccounts.includes(key)))
+  const accountsToMaintain = [
+    ...(await workspace.state().existingAccounts()).filter(key => !fetchedAccounts.includes(key)),
+    ...partiallyFetchedAccounts,
+  ]
+  await workspace.state().updatePathIndex(unmergedElements, accountsToMaintain)
   log.debug(`finish to override state with ${mergedElements.length} elements`)
 }
 
@@ -280,6 +284,7 @@ export const fetch: FetchFunc = async (
     const {
       changes, elements, mergeErrors, errors, updatedConfig,
       configChanges, accountNameToConfigMessage, unmergedElements,
+      partiallyFetchedAccounts,
     } = await fetchChanges(
       accountToAdapter,
       await workspace.elements(),
@@ -290,7 +295,13 @@ export const fetch: FetchFunc = async (
       withChangeDetection,
     )
     log.debug(`${elements.length} elements were fetched [mergedErrors=${mergeErrors.length}]`)
-    await updateStateWithFetchResults(workspace, elements, unmergedElements, fetchAccounts)
+    await updateStateWithFetchResults(
+      workspace,
+      elements,
+      unmergedElements,
+      fetchAccounts,
+      Array.from(partiallyFetchedAccounts),
+    )
     return {
       changes,
       fetchErrors: errors,
@@ -347,7 +358,7 @@ export const fetchFromWorkspace: FetchFromWorkspaceFunc = async ({
   )
 
   log.debug(`${elements.length} elements were fetched from a remote workspace [mergedErrors=${mergeErrors.length}]`)
-  await updateStateWithFetchResults(workspace, elements, unmergedElements, fetchAccounts)
+  await updateStateWithFetchResults(workspace, elements, unmergedElements, fetchAccounts, [])
   return {
     changes,
     fetchErrors: errors,
--- a/packages/core/src/api.ts
+++ b/packages/core/src/api.ts
@@ -248,10 +248,8 @@ const updateStateWithFetchResults = async (
   await workspace.state()
     .override(awu(mergedElements)
       .concat(stateElementsNotCoveredByFetch), fetchedAccounts)
-  const accountsToMaintain = [
-    ...(await workspace.state().existingAccounts()).filter(key => !fetchedAccounts.includes(key)),
-    ...partiallyFetchedAccounts,
-  ]
+  const accountsToMaintain = partiallyFetchedAccounts
+    .concat((await workspace.state().existingAccounts()).filter(key => !fetchedAccounts.includes(key)))
   await workspace.state().updatePathIndex(unmergedElements, accountsToMaintain)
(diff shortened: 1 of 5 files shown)


packages/core/src/core/fetch.ts
```diff
@@ -726,7 +726,7 @@ export const fetchChanges = async (
   currentConfigs: InstanceElement[],
   progressEmitter?: EventEmitter<FetchProgressEvents>,
   withChangesDetection?: boolean
-): Promise<FetchChangesResult> => {
+): Promise<FetchChangesResult & { partiallyFetchedAccounts: Set<string>} > => {
```
ori-moisis, line 729: `FetchChangesResult` describes the result of the function `fetchChanges`, that is its purpose.
it doesn't make sense that `fetchChanges` would return something different

I think it might be easier to let `createFetchChanges` return this new field instead of what you did here

packages/workspace/src/workspace/path_index.ts
```diff
@@ -222,10 +222,10 @@ export const updatePathIndex = async (
   const entries = isTopLevel ? getTopLevelPathHints(elements) : getElementsPathHints(elements)
   const oldPathHintsToMaintain = await awu(index.entries())
     .filter(e => accountsToMaintain.includes(ElemID.fromFullName(e.key).adapter))
-    .concat(entries)
     .toArray()
+  const updatedEntries = _.unionBy(entries, oldPathHintsToMaintain, e => e.key)
```
ori-moisis, line 226: this is generally not a very efficient solution, but can we at least skip this extra iteration on the entire data if there are no accounts to maintain (which is the more common case I think)?

packages/core/src/core/fetch.ts
guyhamo: changed FetchChangesResult to include partiallyFetchedAccounts

packages/workspace/src/workspace/path_index.ts
guyhamo: what do you mean? we already skip this if there are no accounts to maintain.
we used to run concat on the elements, it might have been faster but i wanted to enforce the updated entries to stay.
i checked it locally and for an env with dummy salesforce and jira it took only a few(4ms for top level 16 for normal) milliseconds to run the unionby, on a thousands of elements. I dont think its the problem here.

I like how it does not change the flow of the function and use an existing logic.

packages/core/src/api.ts
```diff
@@ -239,16 +239,20 @@ const updateStateWithFetchResults = async (
...
     .filter(element => !fetchElementsFilter(element.elemID)).toArray()
   await workspace.state()
     .override(awu(mergedElements)
       .concat(stateElementsNotCoveredByFetch), fetchedAccounts)
-  await workspace.state().updatePathIndex(unmergedElements,
-    (await workspace.state().existingAccounts()).filter(key => !fetchedAccounts.includes(key)))
+  const accountsToMaintain = [
+    ...(await workspace.state().existingAccounts()).filter(key => !fetchedAccounts.includes(key)),
+    ...partiallyFetchedAccounts,
+  ]
```
ori-moisis, line 254: should not be very large lists, but we should still try to avoid using `...` to concat lists when possible
```suggestion
  const accountsToMaintain = partiallyFetchedAccounts
    .concat(await workspace.state().existingAccounts()).filter(key => !fetchedAccounts.includes(key))
```

packages/workspace/test/workspace/path_index.test.ts
```diff
@@ -154,14 +178,17 @@ describe('topLevelPathIndex', () => {
...
+    await topLevelPathIndex.setAll(getTopLevelPathHints([singlePathObject, oldPartiallyFetchedObject]))
     await updatePathIndex({
       index: topLevelPathIndex,
       elements: [
         multiPathAnnoObj,
         multiPathFieldsObj,
         multiPathInstanceA,
         multiPathInstanceB,
+        // when there is a partial fetch, the adapter name will be in accountsToMaintain
+        // but there will also be elements of that adapter in the elements array
```
ori-moisis, line 190:
```suggestion
        // when there is a partial fetch, the account name will be in accountsToMaintain
        // but there will also be elements of that account in the elements array
```

packages/workspace/test/workspace/path_index.test.ts
```diff
@@ -191,14 +214,17 @@ describe('updatePathIndex', () => {
...
+    await index.setAll(getElementsPathHints([singlePathObject, oldPartiallyFetchedObject]))
     await updatePathIndex({
       index,
       elements: [
         multiPathAnnoObj,
         multiPathFieldsObj,
         multiPathInstanceA,
         multiPathInstanceB,
+        // when there is a partial fetch, the adapter name will be in accountsToMaintain
+        // but there will also be elements of that adapter in the elements array
```
ori-moisis, line 226:
```suggestion
        // when there is a partial fetch, the account name will be in accountsToMaintain
        // but there will also be elements of that account in the elements array
```

packages/workspace/test/workspace/path_index.test.ts
ori-moisis, line 1: shouldn't there be a test that checks the bug symptom?
I expected to see a test that verifies items are not deleted from the path index on a partial fetch

guyhamo: there is already a check for it using "singlePathObject"
we put it in the index before we update (and use "salto" in accountsToMaintain) without putting it inside the elements.
we then check for it inside the index.

ori-moisis: I am missing something.
if this test really tested this scenario and the bug really exists, why did the test not fail until now?

packages/core/test/api.test.ts
```diff
@@ -139,6 +139,7 @@ describe('api.ts', () => {
   describe('fetch', () => {
     const mockFetchChanges = fetch.fetchChanges as jest.MockedFunction<typeof fetch.fetchChanges>
     const objType = new ObjectType({ elemID: new ElemID(mockService, 'dummy') })
+    let partiallyFetchedAccounts = new Set<string>()
```
ori-moisis, line 142: I think we should avoid sharing mutable context between tests like this.

I think this context sharing is worse than having two separate calls to
```
mockFetchChanges.mockResolvedValue...
```
(one in each `describe` block).

which I see why having to duplicate the mock setup may be less than ideal, I think the risk in sharing mutable context for the readability of the tests is a bigger problem

guyhamo: yeah, it wasnt my first choice but this api.test has annoying beforeAll that supposed to be beforeEach and they ruin everything.

packages/core/test/api.test.ts
```diff
@@ -157,7 +159,11 @@ describe('api.ts', () => {
         elements: fetchedElements,
         mergeErrors: [],
         accountNameToConfigMessage: {},
-      })
+        partiallyFetchedAccounts: mockPartiallyFetchedAccounts(),
+      }))
+    })
+    beforeEach(() => {
+      mockPartiallyFetchedAccounts.mockReturnValue(new Set())
```
ori-moisis, line 166: there is a better way
if you look at the other tests you can see it being used in some places

in the `beforeAll` of the test where you need to return a different value, add a
```typescript
mockFetchChanges.mockResolvedValueOnce(...)
```
similar to how it is done in the tests for `fetch failed`

ori-moisis (approved):

PR #4126, opened 2023-03-19
SALTO-3016 skip NS validation when there are only file cabinet instances

dantal4:
When there are SDF objects being deployed we validate them with the file cabinet instances, but when there are only file cabinet instances being deploy we shouldn't run NS validation using SDF.

---

---
_Release Notes_: 
Netsuite Adapter:
- Skip NS validation when there are only file cabinet instances

---
_User Notifications_: 
None

### DIFF
--- a/packages/netsuite-adapter/src/change_validators/client_validation.ts
+++ b/packages/netsuite-adapter/src/change_validators/client_validation.ts
@@ -83,20 +83,28 @@ const changeValidator: ClientChangeValidator = async (
   additionalDependencies,
   filtersRunner,
 ) => {
+  const changesMap = new Map(changes.map(change => [changeId(change), change]))
   // SALTO-3016 we can validate only SDF elements because
   // we need FileCabinet references to be included in the SDF project
-  const getChangeGroupIds = getChangeGroupIdsFunc(false)
-  const { changeGroupIdMap } = await getChangeGroupIds(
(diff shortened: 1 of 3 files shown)


packages/netsuite-adapter/src/change_validators/client_validation.ts
```diff
@@ -83,20 +83,28 @@ const changeValidator: ClientChangeValidator = async (
...
 
+  const {
+    changeGroupIdMap: realChangesGroupIdMap,
+  } = await getChangeGroupIdsFunc(client.isSuiteAppConfigured())(changesMap)
+
   return awu(changesByGroupId)
     .flatMap(async ([groupId, groupChanges]) => {
+      // SALTO-3016 if the real change group of all changes is different than the one used for validation
+      // (e.g. only FileCabinet instances) we should skip the validation.
+      if (groupChanges.every(change => realChangesGroupIdMap.get(changeId(change)) !== groupId)) {
```
omrilit, line 104: Don't we have a more explicit way to find this one? e.g. "isSDFChangeGroup"

dantal4: We have, but IMO this is more generic - if the change group used for validation is not the real group for any of the changes then we should skip the validation. When won't it be correct?

omrilit (approved):

PR #4132, opened 2023-03-20
SAAS-7388: support ScriptRunner checkbox in UI

IdoZakk:
Added support to receive the ScriptRunner configuration from the UI

---

_Same as zendesk and salesforce_

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/config_creator.ts
+++ b/packages/jira-adapter/src/config_creator.ts
@@ -46,14 +46,8 @@ export const getConfig = async (
   options?: InstanceElement
 ): Promise<InstanceElement> => {
   const defaultConf = await createDefaultInstanceFromType(ElemID.CONFIG_NAME, configType)
-  if (options === undefined || !isOptionsTypeInstance(options)) {
-    return defaultConf
-  }
-  if (options.value.enableScripRunnerAddon === true) {
-    const configWithScriptRunner = defaultConf.clone()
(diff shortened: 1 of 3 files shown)


packages/jira-adapter/src/config_creator.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export const getConfig = async (
+  options?: InstanceElement
+): Promise<InstanceElement> => {
+  const defaultConf = await createDefaultInstanceFromType(ElemID.CONFIG_NAME, configType)
+  if (options === undefined || !isOptionsTypeInstance(options)) {
+    return defaultConf
+  }
+  if (options.value.enableScripRunnerAddon === true) {
+    const configWithScriptRunner = defaultConf.clone()
```
alonstern, line 53: nit: do you actually need to clone it? It doesn't look like `defaultConf` is being used anywhere else

packages/jira-adapter/src/config_creator.ts
```diff
@@ -0,0 +1,64 @@
...
+    configWithScriptRunner.value.fetch = { ...configWithScriptRunner.value.fetch,
+      enableScripRunnerAddon: true }
+    return configWithScriptRunner
+  }
+  return defaultConf
+}
+
+export const configCreator: ConfigCreator = {
```
alonstern, line 61: do you also need to export this from the package `index.ts`?

alonstern (approved):
👍

PR #4133, opened 2023-03-20
SALTO-3823: fix a bug with Automation deployment - Jira

shayc331:
_fix a bug with Automation deployment_

---

The bug occur when the user add new Automation with projects. We use identifier in two places that got the projects in different order so the identifier was different and caused the failure. So I added sorting to prevent it.

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/automation/automation_deployment.ts
+++ b/packages/jira-adapter/src/filters/automation/automation_deployment.ts
@@ -125,8 +125,13 @@ const importAutomation = async (
   })
 }
 
+const sortProjects = (projects: Values[]): Values[] => (
+  _.sortBy(projects, project => project.projectId)
+)
+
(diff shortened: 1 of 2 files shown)


packages/jira-adapter/src/filters/automation/automation_deployment.ts
```diff
@@ -125,8 +125,12 @@ const importAutomation = async (
   })
 }
 
+const sortProjects = (projects: Values[]): Values[] => (
+  _.sortBy(projects, project => project.projectId)
+)
+
 const getAutomationIdentifier = (values: Values): string =>
```
alonstern, line 132: Let's also log the identifier we deployed and all the identifiers in case it will happen again

alonstern (approved):
👌

PR #4135, opened 2023-03-21
SALTO-3793: Log reasons of config suggestion

idoamit8:
**Log the reasons of Config Suggestions**

---

I added a log in `fetch.ts` in core that describes the reasons for config suggestions if needed. 

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/core/src/core/fetch.ts
+++ b/packages/core/src/core/fetch.ts
@@ -448,6 +448,9 @@ const fetchAndProcessMergeErrors = async (
             withChangesDetection,
           })
           const { updatedConfig, errors } = fetchResult
+          if (updatedConfig !== undefined) {
+            log.debug(`Salto got config suggestions in ${accountName} account: ${updatedConfig.message}`)
+          }
           if (
             fetchResult.elements.length > 0 && accountName !== accountToServiceNameMap[accountName]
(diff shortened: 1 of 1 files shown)


packages/core/src/core/fetch.ts
```diff
@@ -448,6 +448,9 @@ const fetchAndProcessMergeErrors = async (
             withChangesDetection,
           })
           const { updatedConfig, errors } = fetchResult
+          if (updatedConfig !== undefined) {
+            log.debug(`Salto got config suggestions in ${accountName} account: ${updatedConfig.message}`)
```
shir-reifenberg, line 452: nit: as we're logging the reasons for the config update, maybe write something similar to this ? 
```suggestion
            log.debug(`In account: ${accountName}, received config suggestions for the following reasons: ${updatedConfig.message}`)
```

shir-reifenberg (approved):
Great 💯 added one nit

packages/core/src/core/fetch.ts
idoamit8: I Didn't see the nit comment and already merged. I will change it on Sunday and ask for PR again.

PR #4139, opened 2023-03-21
SALTO-2957 Support filtering by name in all simple adapters

netama:
Support "naive" filtering by the name field (with one exception in workato).
Based on the Jira support.

---

Note that this may not always behave as expected - e.g. filtering out a folder instance does not mean instances that are "contained" in it are removed. We can later extend this to recursively remove instances based on their parents as well (will be done in a follow-up PR).

---
_Release Notes_: 

_Zendesk adapter_:
* It is now possible to filter the fetched resources by their name.

_Workato adapter_:
* It is now possible to filter the fetched resources by their name.

_Zuora Billing adapter_:
* It is now possible to filter the fetched resources by their name.

_Stripe adapter_:
* It is now possible to filter the fetched resources by their name.

_Okta adapter_:
* It is now possible to filter the fetched resources by their name.

---
_User Notifications_: 
None


packages/adapter-components/src/elements/query/fetch_criteria.ts
```diff
@@ -13,13 +13,9 @@
...
-import { FilterAdditionParams, FilterCreator, FilterResult } from '../filter'
+import { regex } from '@salto-io/lowerdash'
+import { QueryCriterion } from './query'
 
-const filter: FilterCreator = params =>
-  filters.queryFilterCreator<JiraClient, JiraConfig, FilterResult, FilterAdditionParams>(
-  )(params)
-
-export default filter
+export const nameCriterion: QueryCriterion = ({ instance, value }): boolean => (
```
alonstern, line 19: WDYT about exporting 
```
export const defaultFetchCriteria: Record<keyof DefaultFetchCriteria, ElementsQuery>> = {
  name: nameCriterion
}
```
?
This will make it easier to add another default one

alonstern (commented):
LGTM 👌

packages/adapter-components/src/elements/query/fetch_criteria.ts
netama: sounds good, will add this in https://github.com/salto-io/salto/pull/4141

packages/jira-adapter/config_doc.md
```diff
@@ -6,7 +6,6 @@ jira {
     retry = {
       maxAttempts = 5
       retryDelay = 5000
-      retryStrategy = "NetworkError"
```
ori-moisis, line 9: is this change supposed to be here?
seems unrelated...

packages/adapter-components/src/config/shared.ts
```diff
@@ -47,12 +47,16 @@ export type AdapterApiConfig<
...
+  name?: string
+}
+
 type FetchEntry<T extends Record<string, unknown> | undefined> = {
   type: string
   criteria?: T
 }
 
-export type UserFetchConfig<T extends Record<string, unknown> | undefined = undefined> = {
+export type UserFetchConfig<T extends Record<string, unknown> | undefined = DefaultFetchCriteria> = {
```
ori-moisis, line 59: shouldn't the default change in `createElementQuery` as well?
seems like it still has a default of `{}` (no predicates) there...

I might be wrong but IIUC the change here means the config assumes that name is a possible criterion, but this should be coupled with the actual filtering done in `createElementQuery`, so if an adapter does not use the same type in both places, there would be a mis-match between the "interface" it exposes and what is actually supported

packages/okta-adapter/src/adapter.ts
```diff
@@ -102,6 +106,7 @@ export default class OktaAdapter implements AdapterOperations {
 
     this.fetchQuery = elementUtils.query.createElementQuery(
       this.userConfig.fetch,
+      fetchCriteria,
```
ori-moisis, line 109: this is what drove me to the comment on changing the default argument in `createElementQuery`
feels like if an adapter doesn't override the default type, it should not have to define its own fetch criteria structure...

packages/zuora-billing-adapter/src/adapter.ts
```diff
@@ -42,26 +43,22 @@ const { createPaginator } = clientUtils
...
 
 export const DEFAULT_FILTERS = [
   // hideTypes should run before creating custom objects, so that it doesn't hide them
-  hideTypes,
+  hideTypesFilter,
+  queryFilter,
   // objectDefsFilter should run before everything else
   objectDefsFilter,
-
+  queryFilter,
```
ori-moisis, line 54: query filter appears twice on purpose?
seems like the one on line 51 makes more sense than this one (I can see why we need to filter before changing instances to types if the filter only works on instances...)

packages/jira-adapter/config_doc.md
netama: yes

ori-moisis (approved):

PR #4141, opened 2023-03-21
SALTO-2957 Support filtering by name and omitting instances recursively in all simple adapters

netama:
The next step for https://github.com/salto-io/salto/pull/4139 (there are a few additional improvements we should make).

* When an instance is omitted, also recursively omit its children (based on the parent annotation + additional customizations when needed)
* Move the filter to run after the `fieldReferences` filter (in all adapters except jira), both so that the traversal can rely on the references created in the filter, and to allow breaking references when an instance relies on another instance that has been omitted (this is useful especially in workato and zendesk, where the references are by internal id)

---
_Release Notes_: 

_Zendesk adapter_:
* When filtering fetched resources by their name, child instances of omitted resources will be omitted as well.

_Jira adapter_:
* When filtering fetched resources by their name, child instances of omitted resources will be omitted as well

_Okta adapter_:
* When filtering fetched resources by their name, child instances of omitted resources will be omitted as well

_Workato adapter_:
* When filtering fetched resources by their name, child instances of omitted resources will be omitted as well

_Zuora Billing adapter_:
* When filtering fetched resources by their name, child instances of omitted resources will be omitted as well

---
_User Notifications_: 

* If include/exclude by name is used in the fetch configuration, additional instances may be omitted in the next fetch if they are children of instances that have previously been omitted. For example, omitting a zendesk brand will also omit the associated brand logo.


packages/adapter-components/src/filters/query.ts
```diff
@@ -13,33 +13,78 @@
...
+ */
+const createGraph = (
+  instances: InstanceElement[],
+  additionalParentFields?: Record<string, string[]>,
+): DAG<InstanceElement> => {
+  const graph = new DAG<InstanceElement>()
+  instances.forEach(instance => {
+    const parents = getParents(instance)
+    const additionalParents = additionalParentFields?.[instance.elemID.typeName]
+      ?.map(fieldName => instance.value[fieldName]) ?? []
```
ori-moisis, line 37: This is making an assumption that the value of these additional parents field is a single reference.
is this assumption documented somewhere?

packages/adapter-components/src/filters/query.ts
```diff
@@ -13,33 +13,78 @@
...
 const log = logger(module)
 
+/*
+ * Create a graph with instance ids as nodes and parent annotations as edges
+ */
+const createGraph = (
+  instances: InstanceElement[],
+  additionalParentFields?: Record<string, string[]>,
+): DAG<InstanceElement> => {
+  const graph = new DAG<InstanceElement>()
```
ori-moisis, line 33: I think this doesn't need to be a DAG, IIUC there is nothing here that relies on this having no cycles, so this can be a `DataNodeMap`

I would say the data isn't actually used as well so this could even be an `AbstractNodeMap` but that would require moving the implementation of `getComponent` to it and IMO also renaming that class in a separate PR (it should not be too hard I think)

packages/zendesk-adapter/src/filters/common.ts
```diff
@@ -22,7 +22,16 @@ import { FilterCreator } from '../filter'
 const filterCreators: Record<string, FilterCreator> = {
   hideTypes: filters.hideTypesFilterCreator(),
   referencedInstanceNames: filters.referencedInstanceNamesFilterCreator(),
-  query: filters.queryFilterCreator(),
+  query: filters.queryFilterCreator({ typesToIgnore: [
```
ori-moisis, line 25: this is a behavior change right?
if so, I think it might be best to put it in a separate PR

ori-moisis (commented):
I think the main risk here is moving the location of the query filter in all the adapters.
I wonder what effect this might have on the other filters that might now do extra work and if this might break some filtering like in zoura where the filter loses the ability to filter out objectDefs IIUC (because they are no longer instances by the time the filter runs)

packages/adapter-components/src/filters/query.ts
netama: switching to `DataNodeMap` to unblock this PR, will follow up separately with the switch to `AbstractNodeMap`

packages/adapter-components/src/filters/query.ts
netama: you're right, I'll just convert these to arrays instead

packages/zendesk-adapter/src/filters/common.ts
netama: I'll double check now but IIRC it wasn't really (see the comment in the tests). we'd need some more changes in order to actually allow omitting these singletons so I think I'll just leave it this way for now (there's some value in doing it, but not super urgent)

packages/adapter-components/src/filters/query.ts
netama: ended up adding here

packages/workato-adapter/src/adapter.ts
```diff
@@ -42,13 +41,12 @@ const { getAllElements } = elementUtils.ducktype
...
 export const DEFAULT_FILTERS = [
   addRootFolderFilter,
-  queryFilter,
   jiraProjectIssueTypeFilter,
   // fieldReferencesFilter should run after all element manipulations are done
   fieldReferencesFilter,
   recipeCrossServiceReferencesFilter,
   serviceUrlFilter,
-  // referencedIdFieldsFilter should run after element references are resolved
+  // referencedIdFieldsFilter and queryFilter should run after element references are resolved
```
ori-moisis, line 49: this comment exists on a bunch of the adapters, I guess it should be changed on all of them?

packages/zendesk-adapter/src/filters/common.ts
ori-moisis: not sure I understand the answer here, lets talk about this f2f?

packages/adapter-components/src/filters/query.ts
```diff
@@ -13,33 +13,80 @@
...
+}) => FilterCreator<TClient, TContext, TResult, TAdditional> = ({ additionalParentFields, typesToIgnore }) => ({
+  fetchQuery,
+}) => ({
   name: 'queryFilter',
   onFetch: async elements => {
-    const removedInstances = _.remove(
+    const ignoredTypes = new Set(typesToIgnore ?? [])
+    const removedInstances = _.remove(elements, element => (
+      isInstanceElement(element)
+      && !ignoredTypes.has(element.elemID.typeName)
```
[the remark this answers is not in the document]
netama: discussed IRL

packages/okta-adapter/src/adapter.ts
```diff
@@ -85,9 +82,9 @@ export const DEFAULT_FILTERS = [
   serviceUrlFilter,
   appDeploymentFilter,
   // should run after fieldReferences
-  ...Object.values(otherCommonFilters),
-  privateApiDeployFilter,
+  ...Object.values(commonFilters),
```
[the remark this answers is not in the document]
ori-moisis: for future reference, this was done on purpose
also, this could be somewhat important because the new queryFilter logic relies on having references in the instances

ori-moisis (approved):

PR #4142, opened 2023-03-22
PROD-76: enable fetching reports and layouts by default for all users

Shulik95:
_enable fetching reports and layouts by default for all users_

---

_None_

---
_Release Notes_: 
_Netsuite Adapter_:
* reportDefinitions and financialLayouts will be fetched by default and no longer need to be specified explicitly in the config

---
_User Notifications_: 
_reportDefinitions and financialLayouts nacls will be added to the WS/Env_


dantal4 (approved):
💯

PR #4144, opened 2023-03-22
SALTO-3877 add configChanges to the client call in fetch flows filter

Uri-Bar:
SALTO-3877 add configChanges to the client call in fetch flows filter

---
None

---
_Release Notes_: 
None
---

_User Notifications_: 
None


ori-moisis (approved):
please add some context in the ticket for what this change is

PR #4150, opened 2023-03-22
SALTO-3856 improve NS quickFetch changed-objects queries

dantal4:
Should not query system notes specifically in order to know which objects have been changed - we can get the info from the queries of each type.

---

---
_Release Notes_: 
Netsuite Adapter:
- Improve NS quickFetch changed-objects queries

---
_User Notifications_: 
None


alonstern (approved):
LGTM
This is a big improvement 👍

PR #4152, opened 2023-03-23
Fix date format for future dates validation in saved searches

dantal4:
Refactor & use the account's date&time format to check if a date is a future date.

---

---
_Release Notes_: 
Netsuite Adapter:
- Fix date format for future dates validation in saved searches

---
_User Notifications_: 
None


omrilit (approved):
Looks good!!

PR #4155, opened 2023-03-26
SALTO-3896: Fixed fetch error when there are no tabs

alonstern:
Fixed fetch error when there are no tabs

---
_Release Notes_: 
_Jira Adapter_:
- Fixed a fetch error when there is a screen with no tabs

---
_User Notifications_: 
None


IdoZakk (approved):
:goat:

PR #4156, opened 2023-03-26
SALTO-3700 - Zendesk Adapter - Warn the user about bad format of trigger & automation actions

idoser96:
On deploy, if we see that the value type of trigger/automation notification_webhook is not an array, we will warn the user about it

---

None

---
_Release Notes_: 
Zendesk Adapter:
* In case a notification_webhook action of a trigger or an automation, is fetched in a bad format, the user will be warned on the deploy

---
_User Notifications_: 
None


packages/zendesk-adapter/src/change_validators/bad_format_webhook_action.ts
```diff
@@ -0,0 +1,65 @@
...
+    value: unknown
+}
+
+const notificationWebhookAction = joi.object({
+  field: joi.string().valid('notification_webhook').required(),
+  value: joi.any().required(),
+}).unknown(true)
+
+const isNotificationWebhookAction = (value: unknown): value is notificationWebhookAction =>
+  notificationWebhookAction.validate(value).error === undefined
```
edenhassid, line 41: maybe add a log.error if its not true?

edenhassid (approved):
Added a small comment. Besides that looks good! 🤩

packages/zendesk-adapter/src/change_validators/bad_format_webhook_action.ts
idoser96: Its ok if its not true, this is used as filter, not a validation

PR #4158, opened 2023-03-26
SALTO-3872: fix default alias for instances with config name

edenhassid:
fix default alias for instances with config name

---

currently when no alias is provided we calculate the alias from the prettified name. however for setting instances for which their name is `_config` the default alias received is `config` and we don’t want that. So I am changing the default to be the type.

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/workspace/src/workspace/alias_index.ts
+++ b/packages/workspace/src/workspace/alias_index.ts
@@ -37,8 +37,8 @@ export const ALIAS_INDEX_VERSION = 2
 const ALIAS_INDEX_KEY = 'alias_index'
 
 
-const calcAlias = (elem: Element): string => elem.annotations[CORE_ANNOTATIONS.ALIAS]
-  ?? prettifyName(elem.elemID?.isConfigInstance() ? elem.elemID.typeName : elem.elemID.name)
+const getAlias = (element: Element): string => element.annotations[CORE_ANNOTATIONS.ALIAS]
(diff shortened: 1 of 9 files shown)


packages/workspace/src/workspace/alias_index.ts
```diff
@@ -33,10 +33,14 @@ import { RemoteMap } from './remote_map'
 
 const { isDefined } = lowerdashValues
 const log = logger(module)
-export const ALIAS_INDEX_VERSION = 1
+export const ALIAS_INDEX_VERSION = 2
 const ALIAS_INDEX_KEY = 'alias_index'
 
-const calcAlias = (elem: Element): string => elem.annotations[CORE_ANNOTATIONS.ALIAS] ?? prettifyName(elem.elemID.name)
+const isSetting = (elemId: ElemID | undefined): boolean =>
```
adamro, line 39: The fact elemID has a `isConfig` function that has different logic throws me away here o_O

ori-moisis: maybe rename the function on elemID to `isConfigType`?
maybe add this logic to the elemID class (not sure if under `isSettings` or `isConfig`...), or keep this as an external function (`isConfigInstance`) that will check
```typescript
elemId.isConfig() && elemId.idType === 'instance'
```

adamro (approved):
🔥

PR #4159, opened 2023-03-26
SALTO-3893 - Convert url fields with reference expressions

idoser96:
If url is not omitted, replace the brand and the instance Id with reference expressions to support clone

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
Zendesk Adapter:
* Url fields that are not omitted will include reference expressions instead of strings only

### DIFF
--- a/packages/zendesk-adapter/src/adapter.ts
+++ b/packages/zendesk-adapter/src/adapter.ts
@@ -132,6 +132,7 @@ import hideAccountFeatures from './filters/hide_account_features'
 import auditTimeFilter from './filters/audit_logs'
 import { isCurrentUserResponse } from './user_utils'
 import addAliasFilter from './filters/add_alias'
+import urlReferenceExpressionFilter from './filters/url_reference_expression'
 
 const { makeArray } = collections.array
 const log = logger(module)
@@ -159,6 +160,7 @@ export const DEFAULT_FILTERS = [
   // omitInactiveFilter should be before:
   //  order filters, collisionErrorsFilter and fieldReferencesFilter
   omitInactiveFilter,
+  urlReferenceExpressionFilter,
   ticketFormOrderFilter,
   userFieldOrderFilter,
   organizationFieldOrderFilter,
--- a/packages/zendesk-adapter/src/filters/url_reference_expression.ts
+++ b/packages/zendesk-adapter/src/filters/url_reference_expression.ts
@@ -1,0 +1,72 @@
+/*
+*                      Copyright 2023 Salto Labs Ltd.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+import { isInstanceElement, ReferenceExpression, TemplateExpression } from '@salto-io/adapter-api'
+import { FilterCreator } from '../filter'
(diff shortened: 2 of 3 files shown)


packages/zendesk-adapter/src/adapter.ts
```diff
@@ -159,6 +160,7 @@ export const DEFAULT_FILTERS = [
   // omitInactiveFilter should be before:
   //  order filters, collisionErrorsFilter and fieldReferencesFilter
   omitInactiveFilter,
+  urlReferenceExpressionFilter,
```
edenhassid, line 163: is there a reason why you added it after the omitInactiveFilter?

packages/zendesk-adapter/src/filters/url_reference_expression.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    instances.forEach(instance => {
+      const { url } = instance.value
+      // url is usually omitted, if it's not - we need to try and update it
+      if (!url) {
+        return
+      }
+      const { brandUrl, path, id, suffix } = url.match(urlRegex)?.groups ?? {}
+      // Don't do anything if the url doesn't match the expected format
+      if (!brandUrl || !path || !id) {
```
edenhassid, line 52: why don't you check the suffix as well?

packages/zendesk-adapter/src/filters/url_reference_expression.ts
```diff
@@ -0,0 +1,72 @@
...
+        return
+      }
+      const brandInstance = brands.find(brand => brand.value.brand_url === brandUrl)
+
+      // Replace the brand and the instance id with reference expressions
+      instance.value.url = new TemplateExpression({
+        parts: [
+          brandInstance
+            ? new ReferenceExpression(brandInstance.elemID, brandInstance)
```
edenhassid, line 61: I think this should be a reference to the brand_url itself no? like we do in the article body
https://github.com/salto-io/salto/blob/main/packages/zendesk-adapter/src/filters/article/article_body.ts#:~:text=return%20%5Bnew%20ReferenceExpression(urlBrandInstance.elemID.createNestedID(%27brand_url%27)%2C%20urlBrandInstance%3F.value.brand_url)%5D

packages/zendesk-adapter/src/filters/url_reference_expression.ts
```diff
@@ -0,0 +1,72 @@
...
+            : createMissingBrand({ brandUrl, config }),
+          path,
+          new ReferenceExpression(instance.elemID, instance),
```
edenhassid, line 64: isn't it wired to have a reference from the instance to itself? how does it work in deploy?

packages/zendesk-adapter/src/filters/url_reference_expression.ts
```diff
@@ -0,0 +1,72 @@
...
+  }
+  return brandUrl
+}
+
+/**
+ * Updates element's url to reference expressions of the brand and the instance
+ */
+const filterCreator: FilterCreator = ({ config }) => ({
+  name: 'urlReferenceExpression',
+  onFetch: async elements => {
```
edenhassid, line 40: what about predeploy and ondeploy? or maybe. make sure it is ommited in deploy?

edenhassid (commented):
looks good! I think we should also add a preDeploy and onDeploy. Don't forget to add noise reduction.

packages/zendesk-adapter/src/filters/url_reference_expression.ts
idoser96: in case it can be empty

packages/zendesk-adapter/src/filters/url_reference_expression.ts
idoser96: i gotta say i was sceptical as well but it seems to work, but i'll give it another look

packages/zendesk-adapter/src/filters/url_reference_expression.ts
edenhassid: I mean that if you don't add a preDeploy than you will have to ignore this field in the deploy

PR #4161, opened 2023-03-26
SALTO-3845: cancel sorting of workflow post-functions

IdoZakk:
Revert the revert.....
---

_Release Notes_: 
None

---
_User Notifications_: 
None


alonstern (approved):
👌

PR #4162, opened 2023-03-26
SALTO-3640: Add missing InstalledPackage instances

tamtamirr:
Some installed package instances are not retrieved neither through `read` or `retrieve`. This PR creates the missing ones taking advantage of the `namespaces` returned from `listMetadataObjects(InstalledPackage)`

---

Noise Suppression: https://github.com/salto-io/salto_private/pull/5395 (`activateRSS` removal)
Workspace diff: https://github.com/salto-io/tamir-sf/commit/d5cfea10a51e6bf5aa531615d969d40b7998dc8f

---
_Release Notes_: 
Salesforce Adapter:
- Missing `InstalledPackage` instances will be fetched.

---
_User Notifications_: 
Salesforce:
- The `activateRSS` value will be removed from existing `InstalledPackage` instances.
- You may encounter new `InstalledPackage` instances that were not fetched before.


packages/salesforce-adapter/src/filters/create_missing_installed_packages_instances.ts
```diff
@@ -0,0 +1,63 @@
...
+const filterCreator: RemoteFilterCreator = ({ client, config }) => ({
+  name: 'createMissingInstalledPackagesInstancesFilter',
+  onFetch: async (elements: Element[]): Promise<FilterResult | undefined> => {
+    const installedPackageType = await awu(elements)
+      .filter(isObjectType)
+      .find(async objectType => await apiName(objectType) === INSTALLED_PACKAGE_METADATA)
+    if (installedPackageType === undefined) {
+      return
+    }
+    const { elements: listResult } = await listMetadataObjects(client, INSTALLED_PACKAGE_METADATA)
```
tamtamirr, line 46: No need to handle the `errors` here. They would have been handled before if any occurred.

packages/salesforce-adapter/test/filters/create_missing_installed_packages_instances.test.ts
```diff
@@ -0,0 +1,150 @@
...
+          [INSTANCE_FULL_NAME_FIELD]: namespace,
+          version: '1.0.0',
+        },
+        mockTypes.InstalledPackage,
+      )
+    )
+
+
+    beforeEach(() => {
+      listMetadataObjectsSpy = jest.spyOn(fetchModule, 'listMetadataObjects')
```
ori-moisis, line 73: we should not use mocks when not necessary, it is more conventional is our tests to use the client as the mocking layer.
this helps avoiding tests that don't catch errors due to over-mocking, which is an issue we used to have

ori-moisis (commented):
overall looks good
just a few comments on the tests

packages/salesforce-adapter/src/filters/create_missing_installed_packages_instances.ts
ori-moisis: good point

packages/salesforce-adapter/test/filters/create_missing_installed_packages_instances.test.ts
tamtamirr: I agree, but isn't it better to always mock the highest function you use in your code? When it's an exported utility (listMetadataObjects) it's right to assume it's logic is already tested and the code itself is safe. Creating code that will trigger a flow relying on the internal implementation is less clear IMO and will require to dive deeper into the code. Yes, we do have two mocks here, but I've only mocked `Salesforce Client` just because it's a required input to my filter (It could also be a stub that does nothing and that would satisfy my needs)

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -134,7 +134,7 @@ const getNamespace = (obj: FileProperties): string => (
   obj.namespacePrefix === undefined || obj.namespacePrefix === '' ? DEFAULT_NAMESPACE : obj.namespacePrefix
 )
 
-const notInSkipList = (metadataQuery: MetadataQuery, file: FileProperties, isFolderType: boolean): boolean => (
+export const notInSkipList = (metadataQuery: MetadataQuery, file: FileProperties, isFolderType: boolean): boolean => (
```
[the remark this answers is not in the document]
tamtamirr: Opened [SALTO-3950](https://salto-io.atlassian.net/browse/SALTO-3905)

packages/salesforce-adapter/test/filters/create_missing_installed_packages_instances.test.ts
tamtamirr: v

packages/salesforce-adapter/test/filters/create_missing_installed_packages_instances.test.ts
```diff
@@ -0,0 +1,145 @@
...
+        mockTypes.InstalledPackage,
+      )
+    )
+
+
+    beforeEach(() => {
+      jest.spyOn(client, 'listMetadataObjects').mockResolvedValue({
+        result: EXISTING_NAMESPACES.map(createInstalledPackageFileProperties),
+        errors: [],
+      })
```
ori-moisis, line 70: the client is already a mock interface, no need to spy on

ori-moisis: right, the connection is the mock interface, that is the one that should be used like we do in the other tests (e.g - [here](https://github.com/salto-io/salto/blob/main/packages/salesforce-adapter/test/filters/author_information/custom_objects.test.ts#L75))

packages/salesforce-adapter/src/filters/create_missing_installed_packages_instances.ts
ori-moisis: maybe add a comment saying why we don't need to handle the config suggestions here specifically?
just so future readers would know it is on purpose, because at first look this would seem to a bug

packages/salesforce-adapter/test/filters/create_missing_installed_packages_instances.test.ts
tamtamirr: @ori-moisis The linked tests don't use `Client` as-well. Am I missing something? Is this comment resolved?

tamtamirr: Didn't we agree that the level of mocking should be the `Client` and not the `JSForce Connection` though?

ori-moisis: it is basically the same in practice (remember the client has minimal logic)
the important part is working the same way the rest of the code does

ori-moisis (approved):

PR #4163, opened 2023-03-26
SALTO-3127: Unknown Picklist Values Change Validator (not from _restriction)

tamtamirr:
Since we decided to revert [SALTO-3479](https://salto-io.atlassian.net/browse/SALTO-3479), I've needed to change the implementation of the CV to calculate the `allowedValues` from either the `valueSet` annotation or the referenced `Global Value Set` instance.

---

You can see the yielded Change Errors for both cases (Global and normal ValueSet):

Screen Shot 2023-03-26 at 15 31 44

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

[SALTO-3479]: https://salto-io.atlassian.net/browse/SALTO-3479?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/salesforce-adapter/src/change_validators/unknown_picklist_values.ts
```diff
@@ -19,17 +19,60 @@ import {
...
 
 const { isDefined } = values
 const { awu } = collections.asynciterable
 
+type ValueSetValue = {
+  [INSTANCE_FULL_NAME_FIELD]: string
+}[]
+
+type GlobalValueSetValue = InstanceElement['value'] & {
+  customValue: {[INSTANCE_FULL_NAME_FIELD]: string}[]
```
ori-moisis, line 39: is this not
```suggestion
  customValue: ValueSetValue
```
?

ori-moisis (approved):

packages/salesforce-adapter/src/change_validators/unknown_picklist_values.ts
tamtamirr: @ori-moisis I've done the separation on purpose. The Salto types for both of them is different.

For `customValue` of `GlobalValueSet` the type is `List<salesforce.CustomValue>`
For `valueSet` of `Picklist` field the type is `List<salesforce.valueSet>`

My opinion is not too strong here, but we are dealing with two different types thay may not be aligned 🤔

ori-moisis: but you are not reusing those types, you built completely decoupled new types...
I think there is a choice, if you want to say these are different because you are trying to refer to the salto types, then the types should be coupled with the salto types (e.g. - the value set type should have all the fields of a value set and the salto object definition should be done with `createMatchingObjectType` maybe?)
I originally read this code as if you are only defining the types as needed in this code and not coupling them with the salto types. in which case, there is no reason not to reuse them...

I can see reasons to go with either direction.
personally I think I would have gone with saying the type used here is for the purpose of this filter and I would reuse it but this is not a strong feeling...

tamtamirr: v, since this is being used in the context of the filter

PR #4166, opened 2023-03-26
Reverts

omrilit:
---

---
_Release Notes_: 

---
_User Notifications_: 


dantal4 (approved):

omrilit:
As part of https://salto-io.atlassian.net/browse/PROD-76?focusedCommentId=73612 we (@dantal4 and myself) had to make sure we quickly fix the OSS, so we opened https://github.com/salto-io/salto/pull/4165 which fixes the issue.
Since it's late at night and we already tested staging and MUST deploy tomorrow morning at 9am, we didn't want to take the risk of testing the code that is reverted here or test it so we decided to revert it.
FYI @alonstern @idoser96 @IdoZakk @shir-reifenberg

PR #4170, opened 2023-03-27
SALTO-2834: Support account settings types

shir-reifenberg:
Add support in fetch & deploy for account setting types

---

Stil TODO:

- [ ] Consolidate `OktaApiDefinition` with `DucktypeApiDefinition`
- [ ] Add tests

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/okta-adapter/src/config.ts
```diff
@@ -31,12 +31,20 @@ export type OktaActionName = ActionName | 'activate' | 'deactivate'
 export type OktaFetchConfig = configUtils.UserFetchConfig & {
   convertUsersIds?: boolean
 }
-export type OktaApiConfig = configUtils.AdapterSwaggerApiConfig<OktaActionName>
+
+export type OktaSwaggerApiConfig = configUtils.AdapterSwaggerApiConfig<OktaActionName>
+export type OktaDuckTypeApiConfig = configUtils.AdapterDuckTypeApiConfig
+
+export type OktaApiConfig = {
+  swaggerApiConfig: OktaSwaggerApiConfig
```
alonstern, line 39: nit: maybe just `swagger` and `ducktyope` since they already under `apiConfig`

packages/okta-adapter/src/config.ts
```diff
@@ -920,16 +970,150 @@ export const SUPPORTED_TYPES = {
...
+    deployRequests: {
+      modify: {
+        url: '/api/internal/org/settings/reauth-expiration',
+        method: 'post',
+      },
+    },
+  },
 }
 
+export const ADDITIONAL_SUPPORTED_TYPES = {
```
alonstern, line 1094: maybe `DUCKTYPE_SUPPORTED_TYPES`?

packages/okta-adapter/src/config.ts
```diff
@@ -941,9 +1125,37 @@ export const DEFAULT_CONFIG: OktaConfig = {
...
 
+const oktaApiConfigType = createMatchingObjectType<OktaApiConfig>({
+  elemID: new ElemID(OKTA, 'OktaApiConfig'),
+  fields: {
+    swaggerApiConfig: {
+      refType: createSwaggerAdapterApiConfigType({
+        adapter: OKTA,
+      }),
+      annotations: {
+        _required: true,
```
alonstern, line 1142: I am actually not sure it should be required since it does not appear in the NaCl.
You can remove the required and replace the `OktaApiConfig` with `Partial<OktaApiConfig>`

packages/okta-adapter/src/adapter.ts
```diff
@@ -177,9 +213,17 @@ export default class OktaAdapter implements AdapterOperations {
     progressReporter.reportProgress({ message: 'Fetching instances' })
     const { errors, elements: instances } = await this.getInstances(allTypes, parsedConfigs)
 
+    const privateApiElements = await getPrivateApiElements({
```
netama, line 216: 1. nit: I think it will be nice if we can still "hide" the element generation in a more standard way - e.g. something like
```
const { errors, elements } = await this.getAllElements()
```
where `getAllElements` has the logic that we now have here - e.g. something like `this.getAllSwaggerElements()` which will run the current `this.getAllTypes()` and `this.getInstances()` (maybe with different names), and `this.getPrivateApiElements()` (which is the new function - though I think for consistency it should also be a method inside the class and not an external function).
that way when reading the main `fetch()` function, this is all hidden away and the flow is easier to read.
wdyt?

2. there is a risk here of having overlapping object types between the swagger and the ducktype, which can result in unintended conflicts - might be worth adding some prefix to the ducktype config to make sure we avoid it. e.g. `settings_` for each of the types like we do in zuora (it won't be impossible to have overlaps then, but much less likely)

packages/okta-adapter/src/adapter_creator.ts
```diff
@@ -51,12 +53,13 @@ const adapterConfigFromConfig = (config: Readonly<InstanceElement> | undefined):
   )
 
   validateClientConfig(CLIENT_CONFIG, config?.value?.client)
-  validateSwaggerApiDefinitionConfig(API_DEFINITIONS_CONFIG, apiDefinitions)
+  validateSwaggerApiDefinitionConfig(API_DEFINITIONS_CONFIG, apiDefinitions.swagger)
   validateSwaggerFetchConfig(
     FETCH_CONFIG,
     fetch,
-    apiDefinitions
+    apiDefinitions.swagger
```
netama, line 60: we can start with this structure and change it later, but I am not sure this specific additional nesting is the most likely to be what we do going forward - since there are places where we have multiple swaggers as well. I'm guessing eventually we will have separate "sections" one level up (combining a `swagger` section with an `apiDefinitions` section and giving both of them a name that will be used as the prefix for the types).
another small advantage to the "less structured" approach we have in zuora is that then users don't need to be aware of which types are coming from which "part" when overriding e.g. `idFields`.
let's discuss the structure IRL before merging?

packages/okta-adapter/src/adapter_creator.ts
```diff
@@ -69,6 +72,17 @@ const adapterConfigFromConfig = (config: Readonly<InstanceElement> | undefined):
   return adapterConfig
 }
 
+// In order to use private APIs, '-admin' should be added to the account subdomain
+const createAdminClient = (credentials: Credentials, config: OktaConfig): OktaClient => {
+  const { baseUrl } = credentials
+  const domainIndex = baseUrl.indexOf('.okta.com')
```
netama, line 78: 1. nit: if we're keeping this approach, let's use a constant for the `okta.com` part
2. I think we might need some validations if we use this, and/or to avoid crashing the fetch if we cannot generate the private-API domain correctly. it might be a little easier if we split the domain by `.`, ensure the length is (at least?) as expected, then add `-admin` to the subdomain. but any approach where we validate the structure more will be ok :)

netama (commented):
looks great :)
I think there might be some advantages to using a different nesting structure, but not 100% sure - let's discuss IRL before merging. other comments are nits

packages/okta-adapter/src/client/client.ts
```diff
@@ -95,3 +97,24 @@ export default class OktaClient extends clientUtils.AdapterHTTPClient<
       : undefined
   }
 }
+
+/**
+ * Each Okta org has an administrator URL which is used to sign in to the admin console
+ * and used in order to fetch data from private APIs
+ * For more info: https://developer.okta.com/docs/concepts/okta-organizations/#org-urls
+ */
+export const getAdminUrl = (baseUrl: string): string | undefined => {
```
netama, line 106: nit: maybe add this in a separate file under the `client` folder (e.g. `admin.ts` / `private_api.ts`?

packages/okta-adapter/src/config.ts
```diff
@@ -14,29 +14,34 @@
...
+const { createUserFetchConfigType, createSwaggerAdapterApiConfigType, createDucktypeAdapterApiConfigType } = configUtils
 
 export const CLIENT_CONFIG = 'client'
 export const FETCH_CONFIG = 'fetch'
 export const API_DEFINITIONS_CONFIG = 'apiDefinitions'
+export const PRIVATE_API_DEFINITIONS_CONFIG = 'privateApiDefinitions'
 
-export type OktaClientConfig = clientUtils.ClientBaseConfig<clientUtils.ClientRateLimitConfig>
+export type OktaClientConfig = clientUtils.ClientBaseConfig<clientUtils.ClientRateLimitConfig> & {
+  usePrivateAPI: boolean
```
netama, line 30: should be optional?

netama (commented):
looks great :) not approving only because you said you're still making changes

packages/okta-adapter/src/config.ts
shir-reifenberg: It's part of the default config, so I think it doesn't need to be optional (same in jira), but I can change it if u prefer

netama (approved):

packages/okta-adapter/src/filters/user.ts
```diff
@@ -84,12 +85,17 @@ const replaceValues = (instance: InstanceElement, mapping: Record<string, string
   const paths = USER_MAPPING[instance.elemID.typeName]
   paths.forEach(
     path => {
-      const valuesPath = instance.elemID.createNestedID(...path)
-      const values = resolvePath(instance, valuesPath) ?? []
-      values.forEach((value: string, i: number) => {
+      const usersPath = instance.elemID.createNestedID(...path)
+      const pathValues = resolvePath(instance, usersPath) ?? []
+      const userValues = _.isArray(pathValues) ? pathValues : [pathValues]
```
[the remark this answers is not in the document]
shir-reifenberg: Cool!

packages/okta-adapter/src/filters/private_api_deploy.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+    if (adminClient === undefined) {
+      log.error('Skip deployment of private API types because adminClient does not exist')
+      const error = new Error(`The following changes were not deployed, due to error with the private API client: ${relevantChanges.map(c => getChangeData(c).elemID.getFullName()).join(', ')}`)
+      return {
+        leftoverChanges,
+        deployResult: { appliedChanges: [], errors: [error] },
+      }
+    }
+    if (config[CLIENT_CONFIG]?.usePrivateAPI !== true) {
```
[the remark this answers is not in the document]
shir-reifenberg: We're gonna use it for the service url as well 😄

PR #4177, opened 2023-03-28
SALTO-3663: add alias to jira elements

edenhassid:
add alias to jira elements

---

moved the function from zendesk add_alias filter to adapter utils so that the all filters use the same code

---
_Release Notes_: 
Jira: 
* add alias to jira elements

---
_User Notifications_: 
None


packages/jira-adapter/src/filters/add_alias.ts
```diff
@@ -0,0 +1,91 @@
...
+      {
+        fieldName: 'id',
+        referenceFieldName: '_alias',
+      }],
+    separator: '/',
+  },
+  ProjectComponent: {
+    aliasComponents: [
+      {
+        fieldName: '_parent.0',
```
alonstern, line 65: Are we sure the `referenceFieldName` is needed? We can put in the field name `_parent.0.value.value.name` and it will work the same

packages/jira-adapter/src/filters/add_alias.ts
```diff
@@ -0,0 +1,91 @@
...
+    aliasComponents: [{
+      fieldName: 'name',
+    }],
+  },
+  DashboardGadget: {
+    aliasComponents: [{
+      fieldName: 'title',
+    }],
+  },
+  // CustomFieldContext: {
```
alonstern, line 45: Why commneted out?

packages/jira-adapter/src/filters/add_alias.ts
```diff
@@ -0,0 +1,91 @@
...
+        fieldName: 'name',
+      }],
+    separator: '/',
+  },
+}
+
+const filterCreator: FilterCreator = ({ config }) => ({
+  name: 'addAlias',
+  onFetch: async (elements: Element[]): Promise<void> => {
+    if (config.fetch.addAlias === false || config.fetch.addAlias === undefined) {
```
alonstern, line 78: Shouldn't we run it when `config.fetch.addAlias` is undefined? Otherwise it will never run unless the user will add this flag to the config manually

packages/jira-adapter/src/filters/add_alias.ts
edenhassid: @netama wanted it this way

packages/jira-adapter/src/filters/add_alias.ts
edenhassid: I think we will remove the flag once alias is in for all the adapters and it is ready in the UI. for now the flag is undefined by default so that we don't cause noise for customers.

alonstern (approved):
👌

PR #4178, opened 2023-03-28
Omit dates from report definition criteria

dantal4:
Date values in report definition criteria are changing every day if it's not a "CUSTOM" date criteria. In this case we should omit the date values on fetch and leave only the type (today/this month/previous year/etc.).

---

This PR is rebased on https://github.com/salto-io/salto/pull/4176. Please review this PR from [Omit dates from report definition criteria](https://github.com/salto-io/salto/commit/9721ac1a7fe840b26455a9a51d8045de434a867d)

---
_Release Notes_: 
Netsuite Adapter:
- Omit dates from non "CUSTOM" report definition criteria

---
_User Notifications_: 
Netsuite Adapter:
- Date values in report definition criteria will be removed if it's not a "CUSTOM" date criteria.
- Change field name in `reportdefinition_criteria_values` to `FIELD_VALUE`


packages/netsuite-adapter/src/type_parsers/report_definition_parsing/parsed_report_definition.ts
```diff
@@ -144,8 +144,12 @@ type ReportDefinitionInnerFields = {
 type ReportCriteriaValuesType = {
   FIELD_DATE_FILTER_INDEX?: number
   SEQ_NUMBER?: number
-  FILED_VALUE?: string
+  FIELD_VALUE?: string
 }
+// TODO: use ReportCriteriaValuesType for the reportdefinition_criteria_values type-
+// It will change the FILED_VALUE field to FIELD_VALUE (which is the correct value that we see in instances).
```
omrilit, line 150: We said we'll do it also now, no?

packages/netsuite-adapter/src/type_parsers/report_definition_parsing/report_definition_parser.ts
```diff
@@ -85,12 +85,21 @@ const getReportParameters = (reportParameter: { Map?: ParameterObject[] } | unde
     return [i.key._text, i.value._text]
   }).filter(lowerdashValues.isDefined))
 
+const shouldOmitDates = ({ descriptor, values }: ReportCriteriaType): boolean =>
+  descriptor?.FILED_TYPE === 'DATE'
```
omrilit, line 89: FIELD_TYPE

packages/netsuite-adapter/src/type_parsers/report_definition_parsing/report_definition_parser.ts
```diff
@@ -85,12 +85,21 @@ const getReportParameters = (reportParameter: { Map?: ParameterObject[] } | unde
...
+  && values?.length === 3
+  && values[0].FIELD_VALUE !== undefined
+  && values[0].FIELD_VALUE !== 'CUSTOM'
```
omrilit, line 92: What should we do in case it's CUSTOM, isn't it interesting to remove as well? or otherwise we lose some visibility on relevant data?

omrilit (approved):
Looks good!

packages/netsuite-adapter/src/type_parsers/report_definition_parsing/report_definition_parser.ts
dantal4: the actual field name from the definition is FILED_TYPE haha...

packages/netsuite-adapter/src/type_parsers/report_definition_parsing/report_definition_parser.ts
dantal4: If it's custom, it means that the dates doesn't change, and they are used in the definition. If we'll omit them, so the user won't be able to see it and worse - we won't update the definition in fetch if the a custom date was changed.

PR #4180, opened 2023-03-28
SALTO-3801 add missing references to DuplicateRuleFilterItem

benisraeladi:
Do not merge without noise reduction
---

[workspace diff](https://github.com/salto-io/adi-ws/commit/f4416ae56d93263c34830b4142fc698ae31ed0a3)

---
_Release Notes_: 
Salesforce:
* Add references to table and field in DuplicateRuleFilterItem
---
_User Notifications_:  _None_

### DIFF
--- a/packages/salesforce-adapter/src/transformers/reference_mapping.ts
+++ b/packages/salesforce-adapter/src/transformers/reference_mapping.ts
@@ -638,6 +638,16 @@ export const defaultFieldNameToTypeMappingDefs: FieldReferenceDefinition[] = [
     src: { field: 'milestoneName', parentTypes: ['EntitlementProcessMilestoneItem'] },
     target: { type: 'MilestoneType' },
   },
+  {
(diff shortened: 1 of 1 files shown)


packages/salesforce-adapter/src/transformers/reference_mapping.ts
```diff
@@ -634,6 +634,16 @@ export const defaultFieldNameToTypeMappingDefs: FieldReferenceDefinition[] = [
     serializationStrategy: 'customLabel',
     target: { type: CUSTOM_LABEL_METADATA_TYPE },
   },
+  {
```
tomsellek, line 637: For my own education: isn't it enough to have a reference to the field? Why do we need the explicit reference to the type?

benisraeladi: The field is not the full name, and not even a partial name. The `field` value refers to a field inside the `table` table, so in this case it's important to make sure that the reference mapper knows to create a field

PR #4182, opened 2023-03-28
SALTO-3547 - Salesforce: missing reference for recordTypeId field

tomsellek:

---

---
_Release Notes_: 

---
_User Notifications_: 


tomsellek:
This is a reference from a field annotation, which we don't support right now

PR #4186, opened 2023-03-29
SALTO-3917/exporting all config types

guyhamo:
in order to better validate config changes we need to export each adapter config type to generate a schema out of

---

---
_Release Notes_: 
none

---
_User Notifications_:


guyhamo:
I noticed i can import adapter and use it to get the configType

ori-moisis:
don't re-invent the wheel
validation on config instances is already implemented in the adapter config source

PR #4189, opened 2023-03-29
SALTO-2915: Netsuite - add change validator for a unique title field in saved search instance

shoshana-houri:
[Netsuite Adapter]
Add a new change validator to the Netsuite adapter.
This validator validates uniqueness of:
-  savedSearch titles (FIELD_DEFAULT_NAME in nacl files).
- financialLayout name
- customrecordtype fields scriptid 

---

Netsuite does not allow duplicated savedSearch titles, duplicated financiallayout name field and inner customrecordtype fields scriptids, even between different customrecordtypes.
Currently there isn't a validator for that and the deployment will fail.

---
_Release Notes_: 

Netsuite - validate unique title field in saved search instances, financial layout instances and custom record types during deployment.

### DIFF
--- a/packages/netsuite-adapter/src/change_validators/unique_fields.ts
+++ b/packages/netsuite-adapter/src/change_validators/unique_fields.ts
@@ -1,0 +1,207 @@
+/*
+*                      Copyright 2023 Salto Labs Ltd.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+import { ElemID, getChangeData, isAdditionOrModificationChange, ChangeError,
+  ReadOnlyElementsSource, ChangeDataType, isObjectType } from '@salto-io/adapter-api'
+import { values, collections, promises } from '@salto-io/lowerdash'
+import _ from 'lodash'
+import { resolvePath } from '@salto-io/adapter-utils'
+import { isCustomFieldName, isCustomRecordType } from '../types'
+import { NAME_FIELD, FINANCIAL_LAYOUT, SAVED_SEARCH, SCRIPT_ID } from '../constants'
+import { NetsuiteChangeValidator } from './types'
+
+const { awu } = collections.asynciterable
+const { isDefined } = values
+const { mapValuesAsync } = promises.object
+
+
+const FIELD_DEFAULT_NAME = 'FIELD_DEFAULT_NAME'
+const CUSTOM_RECORD_TYPE_NAME_PREFIX = 'customrecord'
+
+type RestrictedType = 'savedSearch' | 'financialLayout' | 'customRecordField'
+
+type GetterParams = {
+  elemID: ElemID
+  elementsSource: ReadOnlyElementsSource
+}
+
+type ElementGetter = (params: GetterParams) => Promise<string[]>
+type ChangeGetter = (change: ChangeDataType) => string
+
+type RestrictedTypeGetters = {
+  getChangeRestrictedField: ChangeGetter
+  getSourceRestrictedFields: ElementGetter
+  getMessage: () => string
+  getDetailedMessage: (field: string) => string
+}
+
+const getNestedField = async ({ elemID, elementsSource }: GetterParams, field: string) : Promise<string[]> =>
+  [await elementsSource.get(elemID.createNestedID(field))]
+
+const getChangeNestedField = (
+  change: ChangeDataType,
+  field: string
+) : string => resolvePath(change, change.elemID.createNestedID(field))
+
+const getCustomRecordRestrictedData = async ({ elemID, elementsSource }: GetterParams
+): Promise<string[]> => {
+  const element = await elementsSource.get(elemID)
+  if (!isObjectType(element) || !isCustomRecordType(element)) {
+    return []
+  }
+
+  return Object.keys(element.fields)
+    .filter(field => isCustomFieldName(field))
+    .map(field => element.fields[field].annotations[SCRIPT_ID])
+}
+
+
+const savedSearchGetters: RestrictedTypeGetters = {
+  getChangeRestrictedField: change => getChangeNestedField(change, FIELD_DEFAULT_NAME),
+  getSourceRestrictedFields: params => getNestedField(params, FIELD_DEFAULT_NAME),
+  getMessage: () => 'A Saved Search with that title already exists',
+  getDetailedMessage: title => `Can't deploy this Saved Search because there is already a Saved Search with the title "${title}" in the target account.`
+    + ' To deploy it, change its title to a unique one.',
+}
+
+const financialLayoutGetters: RestrictedTypeGetters = {
+  getChangeRestrictedField: change => getChangeNestedField(change, NAME_FIELD),
+  getSourceRestrictedFields: params => getNestedField(params, NAME_FIELD),
+  getMessage: () => 'A Financial Layout with that name already exists',
+  getDetailedMessage: name => `Can't deploy this Financial Layout because there is already a Financial Layout with the name "${name}" in the target account.`
+    + ' To deploy it, change its name to a unique one.',
+}
+
+const customRecordGetters: RestrictedTypeGetters = {
+  getChangeRestrictedField: change => getChangeNestedField(change, SCRIPT_ID),
+  getSourceRestrictedFields: getCustomRecordRestrictedData,
+  getMessage: () => 'A Custom Record Type Field with that ID already exists',
+  getDetailedMessage: scriptID => `Can't deploy this Custom Record Type Field because there is already a Custom Record Type Field with the ID "${scriptID}" in the target account.`
+    + ' To deploy it, change its ID to a unique one.',
+}
+
+const restrictedTypeGettersMap: Record<RestrictedType, RestrictedTypeGetters> = {
+  savedSearch: savedSearchGetters,
+  financialLayout: financialLayoutGetters,
+  customRecordField: customRecordGetters,
+}
+
+const getRestrictedType = (elemID: ElemID, includeFieldElements: boolean): RestrictedType | undefined => {
+  if (elemID.idType === 'instance') {
+    if (elemID.typeName === SAVED_SEARCH) {
+      return 'savedSearch'
+    }
+
+    if (elemID.typeName === FINANCIAL_LAYOUT) {
+      return 'financialLayout'
+    }
+  } else if ((includeFieldElements && elemID.idType === 'field') || (!includeFieldElements && elemID.idType === 'type')) {
+    if (elemID.typeName.startsWith(CUSTOM_RECORD_TYPE_NAME_PREFIX)) {
+      return 'customRecordField'
+    }
+  }
+
+  return undefined
+}
+
+const getEmptyRestrictedTypeRecord = <T>() : Record<RestrictedType, T[]> => ({
+  savedSearch: [],
+  financialLayout: [],
+  customRecordField: [],
+})
+
+const getTypeToDataRecord = <T> (
+  elements: T[],
+  getElemID: (element: T) => ElemID,
+  includeFieldElements = true,
+  typesToInclude?: Set<RestrictedType>,
+): Record<RestrictedType, T[]> => {
+  const typeToDataRecord = getEmptyRestrictedTypeRecord<T>()
+
+  const addToGroup = (elem: T): void => {
+    const elemID = getElemID(elem)
+    const restrictedType = getRestrictedType(elemID, includeFieldElements)
+    if (isDefined(restrictedType)
+      && (_.isUndefined(typesToInclude) || typesToInclude.has(restrictedType))) {
+      typeToDataRecord[restrictedType].push(elem)
+    }
+  }
+
+  elements.forEach(addToGroup)
+
+  return typeToDataRecord
+}
+
+const getTypeToRestrictedFields = (
+  elementsSource: ReadOnlyElementsSource,
+  typeToElementsRecord: Record<RestrictedType, ElemID[]>
+): Promise<Record<RestrictedType, string[]>> =>
+  mapValuesAsync(typeToElementsRecord,
+    (elemIDs, type) =>
+      awu(elemIDs)
+        .flatMap(elemID => restrictedTypeGettersMap[type as RestrictedType]
+          .getSourceRestrictedFields({ elemID, elementsSource }))
+        .toArray())
+
+const validateDuplication = (
+  changesData: ChangeDataType[],
+  uniqueFieldToID: Record<string, number>,
+  type: RestrictedType
+): Promise<Array<ChangeError>> => {
+  const getters = restrictedTypeGettersMap[type]
+
+  return awu(changesData)
+    .map(change => ({
+      elemID: change.elemID,
+      field: getters.getChangeRestrictedField(change),
+    }))
+    .filter(({ field }) => (uniqueFieldToID[field] ?? 0) > 1)
+    .map(({ elemID, field }): ChangeError => ({
+      elemID,
+      severity: 'Error',
+      message: getters.getMessage(),
+      detailedMessage: getters.getDetailedMessage(field),
+    }))
+    .toArray()
+}
+
+const changeValidator: NetsuiteChangeValidator = async (changes, _deployReferencedElements, elementsSource) => {
+  const changeElements = changes
+    .filter(isAdditionOrModificationChange)
+    .map(getChangeData)
+
+  const typesToChanges = getTypeToDataRecord(changeElements, element => element.elemID)
+  const relevantTypes = new Set(
+    Object.keys(_.pickBy(typesToChanges, mappedChanges => mappedChanges.length > 0)) as RestrictedType[]
+  )
+
+  if (!elementsSource || relevantTypes.size === 0) {
+    return []
+  }
+
+  const typeToElements = getTypeToDataRecord(
+    (await awu(await elementsSource.list()).toArray()), elem => elem, false, relevantTypes
+  )
+  const typeToRestrictedFields = await getTypeToRestrictedFields(elementsSource, typeToElements)
+
+  const getErrors = (type: RestrictedType, changesSingleType: ChangeDataType[]): Promise<Array<ChangeError>> =>
+    validateDuplication(changesSingleType, _.countBy(typeToRestrictedFields[type]), type)
+
+  return awu(Object.entries(typesToChanges))
+    .flatMap(([type, changesOfType]) => getErrors(type as RestrictedType, changesOfType))
+    .toArray()
+}
+
+export default changeValidator
--- a/packages/netsuite-adapter/test/change_validators/unique_fields.test.ts
+++ b/packages/netsuite-adapter/test/change_validators/unique_fields.test.ts
@@ -1,0 +1,390 @@
+/*
+*                      Copyright 2023 Salto Labs Ltd.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+import { buildElementsSourceFromElements } from '@salto-io/adapter-utils'
+import { ElemID, InstanceElement, ObjectType, toChange, ReadOnlyElementsSource, Field, Element, ChangeDataType } from '@salto-io/adapter-api'
+import uniqueFields from '../../src/change_validators/unique_fields'
+import { CUSTOM_RECORD_TYPE, CUSTOM_RECORD_TYPE_PREFIX, FINANCIAL_LAYOUT, METADATA_TYPE, NAME_FIELD, NETSUITE, SAVED_SEARCH } from '../../src/constants'
+
+const DUPLICATED_FIELD = 'test uniqueness'
+const UNIQUE_FIELD = 'test uniqueness 2'
+
+const FIELD_DEFAULT_NAME = 'FIELD_DEFAULT_NAME'
+
+describe('unique fields validator', () => {
+  type TestElements = {
+    basic: ChangeDataType
+    sameField: ChangeDataType
+    diffField: ChangeDataType
+  }
+
+  const getTestElements = (
+    typeName: string,
+    createElem: (name: string, type: ObjectType, uniqueField: string) => ChangeDataType
+  ): TestElements => ({
+    basic: createElem(
+      'test',
+      new ObjectType({ elemID: new ElemID(NETSUITE, typeName) }),
+      DUPLICATED_FIELD
+    ),
+    sameField: createElem(
+      'test_same_title_diff_id',
+      new ObjectType({ elemID: new ElemID(NETSUITE, typeName) }),
+      DUPLICATED_FIELD
+    ),
+    diffField: createElem(
+      'test_diff_title_diff_id',
+      new ObjectType({ elemID: new ElemID(NETSUITE, typeName) }),
+      UNIQUE_FIELD
+    ),
+  })
+
+  const getIDToVal = (
+    testInstances: TestElements,
+    nestedField: string
+  ): Map<string, string> => new Map(Object.values(testInstances)
+    .map(elem =>
+      [elem.elemID.createNestedID(nestedField).getFullName(),
+        (elem as InstanceElement).value[nestedField]]))
+
+  const buildMockElementsSource = (
+    elementSource: ReadOnlyElementsSource,
+    idToVal: Map<string, string>
+  ): ReadOnlyElementsSource => {
+    const mockGet = jest.fn().mockImplementation((id: ElemID) => idToVal.get(id.getFullName()) ?? elementSource.get(id))
+    const mockElementsSource = {
+      list: elementSource.list,
+      getAll: elementSource.getAll,
+      has: elementSource.has,
+      get: mockGet,
+    } as unknown as ReadOnlyElementsSource
+
+    return mockElementsSource
+  }
+
+  const getSavedSearchElements = (): TestElements => getTestElements(SAVED_SEARCH,
+    (name: string, elemID: ObjectType, uniqueField: string) =>
+      new InstanceElement(name, elemID, { FIELD_DEFAULT_NAME: uniqueField }))
+
+  const getFinancialLayoutElements = (): TestElements => getTestElements(FINANCIAL_LAYOUT,
+    (name: string, elemID: ObjectType, uniqueField: string) =>
+      new InstanceElement(name, elemID, { name: uniqueField }))
+
+  const getCustomRecordElements = (): TestElements => getTestElements(CUSTOM_RECORD_TYPE_PREFIX,
+    (name: string, elemID: ObjectType, uniqueField: string) =>
+      new ObjectType({
+        elemID: new ElemID(elemID.elemID.adapter, elemID.elemID.typeName + uniqueField),
+        annotations: {
+          name,
+          [METADATA_TYPE]: CUSTOM_RECORD_TYPE,
+        },
+        fields: {
+          custom_custrecord1: {
+            refType: elemID,
+            annotations: {
+              scriptid: uniqueField,
+            },
+          },
+        },
+      }))
+
+  const getCustomRecordChanges = (customRecordTestObjects: TestElements): TestElements => {
+    const getFieldFromElement = (element: ObjectType): Field =>
+      new Field(element, element.annotations.name, element, element.fields.custom_custrecord1.annotations)
+    return {
+      basic: getFieldFromElement(customRecordTestObjects.basic as ObjectType),
+      sameField: getFieldFromElement(customRecordTestObjects.sameField as ObjectType),
+      diffField: getFieldFromElement(customRecordTestObjects.diffField as ObjectType),
+    }
+  }
+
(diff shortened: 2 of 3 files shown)


packages/netsuite-adapter/src/change_validators/saved_search_unique_title.ts
```diff
@@ -0,0 +1,85 @@
...
+import { logger } from '@salto-io/logging'
+import { ElemID, getChangeData, InstanceElement, isAdditionOrModificationChange, ReadOnlyElementsSource } from '@salto-io/adapter-api'
+import { collections } from '@salto-io/lowerdash'
+import { NetsuiteChangeValidator } from './types'
+
+const { awu } = collections.asynciterable
+const log = logger(module)
+
+const isSavedSearchInstanceElemID = (elemID: ElemID): boolean =>
+  (elemID.idType === 'instance') && (elemID.typeName === 'savedsearch')
```
kami-delaroz, line 25: please use isInstanceElement and isSavedSearchInstance directly

packages/netsuite-adapter/src/change_validators/saved_search_unique_title.ts
```diff
@@ -0,0 +1,85 @@
...
+
+/* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+const isSavedSearchInstanceElement = (element: any): element is InstanceElement =>
+  isSavedSearchInstanceElemID(element.elemID)
```
kami-delaroz, line 29: I'm not sure there is a need here for two functions

packages/netsuite-adapter/src/change_validators/saved_search_unique_title.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  return titleToIds
+}
+
+const changeValidator: NetsuiteChangeValidator = async (changes, _deployReferencedElements, elementsSource) => {
+  const savedSearchChanges = changes
+    .filter(isAdditionOrModificationChange)
+    .map(getChangeData)
+    .filter(isSavedSearchInstanceElement)
```
kami-delaroz, line 61: you can continue directly here with 
`    .filter(changeElem => {
      const uniqueIds = titleToIds.get(changeElem.value.FIELD_DEFAULT_NAME)?.size
      return uniqueIds !== undefined && uniqueIds > 1
    })
    .map(({ value, elemID }) => ({
      elemID,
      severity: 'Error',
      message: 'Saved search must have a unique title',
      detailedMessage: `Saved search title (default name) ${value.FIELD_DEFAULT_NAME} already exists`,
    }))`

packages/netsuite-adapter/src/change_validators/saved_search_unique_title.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  if (!elementsSource || savedSearchChanges.length === 0) {
+    return []
+  }
+
+  const titleToIds = await getTitleToIds(elementsSource)
```
kami-delaroz, line 67: in order to have this param being calculated only if needed (lazy loading) and only once,
see `typesWithInstances` calculation in `change_validators/remove_sdf_elements.ts`

packages/netsuite-adapter/src/change_validators/saved_search_unique_title.ts
```diff
@@ -0,0 +1,85 @@
...
+    .map(getChangeData)
+    .filter(isSavedSearchInstanceElement)
+    .filter(changeElem => {
+      const uniqueIds = titleToIds.get(changeElem.value.FIELD_DEFAULT_NAME)?.size
+      return uniqueIds !== undefined && uniqueIds > 1
+    })
+    .map(({ value, elemID }) => ({
+      elemID,
+      severity: 'Error',
+      message: 'Saved search must have a unique title',
```
kami-delaroz, line 80: change msg to validators format. something like
"Can\'t deploy Saved search with an ununiq title"
please also verify the messages with Ofer Rozenman in the ticket

packages/netsuite-adapter/src/change_validators/saved_search_unique_title.ts
```diff
@@ -0,0 +1,85 @@
...
+      detailedMessage: `Saved search title (default name) ${value.FIELD_DEFAULT_NAME} already exists`,
```
kami-delaroz, line 81: also should start with "Can't deploy .."
and add an optional fix ("change element title")

packages/netsuite-adapter/src/change_validators/saved_search_unique_title.ts
```diff
@@ -0,0 +1,85 @@
...
+const log = logger(module)
+
+const isSavedSearchInstanceElemID = (elemID: ElemID): boolean =>
+  (elemID.idType === 'instance') && (elemID.typeName === 'savedsearch')
+
+/* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+const isSavedSearchInstanceElement = (element: any): element is InstanceElement =>
+  isSavedSearchInstanceElemID(element.elemID)
+
+const getTitleToIds = async (elementsSource: ReadOnlyElementsSource): Promise<Map<string, Set<string>>> => {
```
kami-delaroz, line 31: please notice elementsSource might be undefined, as it's an optional parameter in NetsuiteChangeValidator.

packages/netsuite-adapter/src/change_validators/saved_search_unique_title.ts
```diff
@@ -0,0 +1,85 @@
...
+  interface savedSearchPartialData {
+    title: string
+    scriptID: string
+  }
+  const sourceSavedSearchPartialData = await log.time(async () =>
+    awu(await elementsSource.list())
```
kami-delaroz, line 37: haven't we said we're going with get instead of list?

packages/netsuite-adapter/src/change_validators/saved_search_unique_title.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  const titleToIds: Map<string, Set<string>> = new Map()
+  let title: string
+  let scriptID: string
+  for ({ title, scriptID } of sourceSavedSearchPartialData.values()) {
+    const ids = titleToIds.get(title) ?? new Set()
+
+    ids.add(scriptID)
+    titleToIds.set(title, ids)
+  }
```
dantal4, line 52: We really prefer using consts and `.map` / `.forEach` instead of `let` and for loop, and I think that in this case it can changed to `forEach` 😃

packages/netsuite-adapter/src/change_validators/saved_search_unique_title.ts
```diff
@@ -0,0 +1,85 @@
...
+  const sourceSavedSearchPartialData = await log.time(async () =>
+    awu(await elementsSource.list())
+      .filter(isSavedSearchInstanceElemID).map(async (elem): Promise<savedSearchPartialData> => {
+        const relevantData = { title: await elementsSource.get(elem.createNestedID('FIELD_DEFAULT_NAME')),
+          scriptID: await elementsSource.get(elem.createNestedID('scriptid')) }
+        return relevantData
+      }).toArray(), 'fetching saved searches')
+
+  const titleToIds: Map<string, Set<string>> = new Map()
```
dantal4, line 44: Consider using lodash's `countBy` function - https://lodash.com/docs/4.17.15#countBy

packages/netsuite-adapter/src/change_validators/saved_search_unique_title.ts
shoshana-houri: Changed in the new version.
As we discussed f2f, the new verification is done using ElemID param, therefore `isInstanceElement` and `isSavedSearchInstance` cannot be used. However, I do use constants instead of string literals.

packages/netsuite-adapter/src/change_validators/saved_search_unique_title.ts
shoshana-houri: Changed in the new version.

packages/netsuite-adapter/src/change_validators/saved_search_unique_title.ts
shoshana-houri: Same as in [https://github.com/salto-io/salto/pull/4189#discussion_r1161282009](url)

packages/netsuite-adapter/src/change_validators/saved_search_unique_title.ts
shoshana-houri: Asked Ofer, waiting for his response :)

packages/netsuite-adapter/src/change_validators/saved_search_unique_title.ts
shoshana-houri: Changed that, thanks for letting me know :)

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -0,0 +1,235 @@
...
+    .map(field => element.fields[field].annotations.scriptid)
+}
+
+/* If you wish validate another type's unique fields, just:
+      1. Add its type definition (of your choice) to RestrictedTypes.
+      2. Define its RestrictedTypeGetters.
+      3. Update the getters map.
+      4. Update getRestrictedType to match the new type as well. */
+
+type RestrictedTypes = typeof SAVED_SEARCH_TYPE | typeof FINANCIAL_LAYOUT_TYPE | typeof CUSTOM_RECORD_FIELD_TYPE
```
dantal4, line 83: You can use here the strings directly (and not `typeof SOME_CONST`), and then wherever you expect/return `RestrictedTypes` it will be restricted (🥁 ) only to these strings, so the constants aren't necessary.

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -0,0 +1,235 @@
...
+  }
+
+  return undefined
+}
+
+/* From this place forward the functions are general.
+   Change them carefully only if the new type's logic requires that. */
+
+const getTypeToDataMap = <T extends ElemID | ChangeDataType>(
+  elements: T[],
```
dantal4, line 137: It's wrong to expect for different unrelated types. The proper way is to pass only one type (`ElemID` in this case), and transform the input outside of the function:
```
const changesElemIDs = changes
    .map(getChangeData)
    .map(elem => elem.elemID)
const typeToChanges = getTypeToDataMap(changesElemIDs)
```

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -0,0 +1,235 @@
...
+
+const getCustomRecordRestrictedData = async ({ elemID, elementsSource }: GetterParams):
+  Promise<string[]> => {
+  const element = await elementsSource.get(elemID)
+  if (!isObjectType(element) || !isCustomRecordType(element)) {
+    return []
+  }
+
+  return Object.keys(element.fields)
+    .filter(field => field.startsWith(CUSTOM_RECORD_FIELD_PREFIX))
```
dantal4, line 73:
```suggestion
    .filter(isCustomFieldName)
```

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -0,0 +1,235 @@
...
+const restrictedTypesGettersMap: Record<RestrictedTypes, RestrictedTypeGetters> = {
+  savedSearch: savedSearchGetters,
+  financialLayout: financialLayoutGetters,
+  customRecordField: customRecordGetters,
+}
+
+const getRestrictedType = (elemID: ElemID, isChange = true): RestrictedTypes | undefined => {
+  if (elemID.idType === INSTANCE_TYPE) {
+    if (elemID.typeName === SAVED_SEARCH) {
+      return SAVED_SEARCH_TYPE
```
dantal4, line 118: example of [this](https://github.com/salto-io/salto/pull/4189/files#r1165054318) comment
```suggestion
      return 'savedSearch'
```

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -0,0 +1,235 @@
...
+/* From this place forward the functions are general.
+   Change them carefully only if the new type's logic requires that. */
+
+const getTypeToDataMap = <T extends ElemID | ChangeDataType>(
+  elements: T[],
+  isChange = true,
+  typeToChanges?: Map<RestrictedTypes, ChangeDataType[]>
+):
+  Map<RestrictedTypes, T[]> => {
+  const restrictedGroups: Map<RestrictedTypes, T[]> = new Map()
```
dantal4, line 142: why not using a `Record`? you can initiate it with empty lists so the condition in line 150 will be unnecessary: 
```suggestion
  const restrictedGroups: Record<RestrictedTypes, T[]> = {
       savedSearch: [],
       financialLayout: [],
       customRecordField: [],
  }
```

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -0,0 +1,235 @@
...
+
+const getTypeToRestrictedFields = async (
+  elementsSource: ReadOnlyElementsSource,
+  typeToElementsMap: Map<RestrictedTypes, ElemID[]>):
+  Promise<Map<RestrictedTypes, string[]>> => {
+  const typeToFields: Map<RestrictedTypes, string[]> = new Map()
+
+  const setMap = async (elementsIDs: ElemID[], type: RestrictedTypes): Promise<void> => {
+    const getters = restrictedTypesGettersMap[type]
+    if (isDefined(getters)) {
```
dantal4, line 172: how can it be undefined?

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -0,0 +1,235 @@
...
+      const fields: string[][] = await awu(elementsIDs).map(elemID =>
```
dantal4, line 173:
```suggestion
      const fields: string[] = await awu(elementsIDs).flatMap(elemID =>
```

packages/netsuite-adapter/src/change_validators/unique_fields.ts
dantal4: In general, anywhere that uses `Map<RestrictedTypes, ...>` should be replaced with `Record<...>` IMO - because the keys are well-defined.

shoshana-houri: The places I used Map instead of Record are ones where we don't necessarily need a mapping of **all** the RestrictedTypes.
In addition, using Record requires initialization with all the RestrictedTypes, which extends the areas in the code that require changes when adding a new restricted type. I wanted to avoid that and make the code as 'generic' as I could.

WDYT?

dantal4: IMO It's not bad that adding a new restrictedType requires changes in the code - it's an advantage that TS force us to add it, and this way prevent missing some of the places. You can still avoid code duplication by having a function that generates an empty Record.

packages/netsuite-adapter/src/change_validators/unique_fields.ts
dantal4: I understand now why you used the mixed type input, and I still think that we can have a more elegant solution.
Your suggestion is good, and here's another suggestion that you can use:
```
const getTypeToDataMap = <T>(
  items: T[],
  toElemId: (item: T) => ElemID,
  ...
```
and then inside `addToGroup` use:
```
const matchedRestriction = getRestrictedType(toElemId(item), isChange)
```
This way the function doesn't care at all about the input type, and uses `toElemId(..)` supplied by the caller to get the elemID, which is the only thing interesting in `T`.

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -0,0 +1,218 @@
...
+
+const changeValidator: NetsuiteChangeValidator = async (changes, _deployReferencedElements, elementsSource) => {
+  const changeElements = changes
+    .filter(isAdditionOrModificationChange)
+    .map(getChangeData)
+
+  const typesToChanges = getTypeToDataRecord(changeElements, (element: ChangeDataType) => element.elemID)
+  const relevantTypes = new Set(Object.entries(typesToChanges)
+    .filter(typeToChanges => typeToChanges[1].length > 0)
+    .map(typeToChanges => typeToChanges[0] as RestrictedType))
```
dantal4, line 199: nit - lodash have some helpful functions
```suggestion
  const relevantTypes = new Set(
    Object.keys(_.pickBy(typesToChanges, changes => changes.length > 0) as RestrictedType[]
  )
```

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -0,0 +1,218 @@
...
+  if (!elementsSource || relevantTypes.size === 0) {
+    return []
+  }
+
+  const typeToElements = getTypeToDataRecord(
+    (await awu(await elementsSource.list()).toArray()), elem => elem, false, relevantTypes
+  )
+  const typeToRestrictedFields = await getTypeToRestrictedFields(elementsSource, typeToElements)
+
+  const getErrors = async (type: RestrictedType, changesSingleType: ChangeDataType[]): Promise<Array<ChangeError>> =>
```
dantal4, line 210: `async` is unnecessary since there's no `await` in this function.

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -0,0 +1,218 @@
...
+    typeToFields[type] = fields
+  }
+
+  await awu(Object.entries(typeToElementsRecord)).forEach(typeToElements =>
+    setMap(typeToElements[0] as RestrictedType, typeToElements[1]))
+
+  return typeToFields
+}
+
+const validateDuplication = async (
```
dantal4, line 170: `async` is unnecessary since there's no `await` in this function.

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -0,0 +1,218 @@
...
+  const typeToElements = getTypeToDataRecord(
+    (await awu(await elementsSource.list()).toArray()), elem => elem, false, relevantTypes
+  )
+  const typeToRestrictedFields = await getTypeToRestrictedFields(elementsSource, typeToElements)
+
+  const getErrors = async (type: RestrictedType, changesSingleType: ChangeDataType[]): Promise<Array<ChangeError>> =>
+    validateDuplication(changesSingleType, countBy(typeToRestrictedFields[type]), type)
+
+  return awu(Object.entries(typesToChanges))
+    .flatMap(typeToChanges => getErrors(typeToChanges[0] as RestrictedType, typeToChanges[1]))
```
dantal4, line 214: nit
```suggestion
    .flatMap(([type, changesOfType]) => getErrors(type as RestrictedType, changesOfType))
```

packages/netsuite-adapter/test/change_validators/unique_fields.test.ts
```diff
@@ -0,0 +1,222 @@
...
+import { buildElementsSourceFromElements } from '@salto-io/adapter-utils'
+import { ElemID, InstanceElement, ObjectType, toChange, ReadOnlyElementsSource, Field, Element, ChangeError } from '@salto-io/adapter-api'
+import uniqueFields from '../../src/change_validators/unique_fields'
+import { CUSTOM_RECORD_TYPE, CUSTOM_RECORD_TYPE_PREFIX, FINANCIAL_LAYOUT, METADATA_TYPE, NETSUITE, SAVED_SEARCH } from '../../src/constants'
+
+const DUPLICATED_FIELD = 'test uniqueness'
+const UNIQUE_FIELD = 'test uniqueness 2'
+
+describe('unique fields validator', () => {
+  type ValidTypes = InstanceElement | ObjectType | Field
```
dantal4, line 25: this is `ChangeDataType`

packages/netsuite-adapter/test/change_validators/unique_fields.test.ts
```diff
@@ -0,0 +1,222 @@
...
+
+  type TestElements = {
+    BASIC: ValidTypes
+    SAME_FIELD: ValidTypes
+    DIFF_FIELD: ValidTypes
+  }
+
+  const getTestElements = (typeName: string, createElem:
+    (name: string, elemID: ObjectType, uniqueField: string) => ValidTypes): TestElements => {
```
dantal4, line 34: the common style is that if the args don't fit into one line then they should be in separate lines for each arg:
```suggestion
  const getTestElements = (
    typeName: string,
    createElem: (name: string, elemID: ObjectType, uniqueField: string) => ValidTypes
  ): TestElements => {
```

dantal4: and I think that the second arg should be called `type`?

packages/netsuite-adapter/test/change_validators/unique_fields.test.ts
```diff
@@ -0,0 +1,222 @@
...
+    const elemID = new ObjectType({
```
dantal4, line 35: also `type`

packages/netsuite-adapter/test/change_validators/unique_fields.test.ts
```diff
@@ -0,0 +1,222 @@
...
+const UNIQUE_FIELD = 'test uniqueness 2'
+
+describe('unique fields validator', () => {
+  type ValidTypes = InstanceElement | ObjectType | Field
+
+  type TestElements = {
+    BASIC: ValidTypes
+    SAME_FIELD: ValidTypes
+    DIFF_FIELD: ValidTypes
+  }
```
dantal4, line 31: just curious, why did you choose upper case?
if it wasn't a test, it wasn't valid 😺

packages/netsuite-adapter/test/change_validators/unique_fields.test.ts
```diff
@@ -0,0 +1,222 @@
...
+
+    return {
+      BASIC: createElem('test', elemID, DUPLICATED_FIELD),
+      SAME_FIELD: createElem('test_same_title_diff_id', elemID, DUPLICATED_FIELD),
+      DIFF_FIELD: createElem('test_diff_title_diff_id', elemID, UNIQUE_FIELD),
+    }
+  }
+
+  const buildMockElementsSource = (elementSource: ReadOnlyElementsSource, idToVal: Map<string, string>):
+  ReadOnlyElementsSource => {
```
dantal4, line 47:
```suggestion
  const buildMockElementsSource = (
    elementSource: ReadOnlyElementsSource,
    idToVal: Map<string, string>
  ): ReadOnlyElementsSource => {
```

packages/netsuite-adapter/test/change_validators/unique_fields.test.ts
```diff
@@ -0,0 +1,222 @@
...
+      new InstanceElement(name, elemID, { name: uniqueField }))
+
+  const getCustomRecordElements = (): TestElements => getTestElements(CUSTOM_RECORD_TYPE_PREFIX,
+    (name: string, elemID: ObjectType, uniqueField: string) =>
+      new ObjectType({ elemID: new ElemID(elemID.elemID.adapter, elemID.elemID.typeName + uniqueField),
+        annotations:
+          { name,
+            [METADATA_TYPE]: CUSTOM_RECORD_TYPE },
+        fields:
+          { custom_custrecord1: { refType: elemID, annotations: { scriptid: uniqueField } } } }))
```
dantal4, line 74: please fix the formatting here - each key in a different line and nested too, like:
```
new ObjectType({
  elemID: ...,
  annotations: {
     name,
     ....
  },
  ...
})
```

packages/netsuite-adapter/test/change_validators/unique_fields.test.ts
```diff
@@ -0,0 +1,222 @@
...
+    return {
+      BASIC: getFieldFromElement(customRecordTestObjects.BASIC as ObjectType),
+      SAME_FIELD: getFieldFromElement(customRecordTestObjects.SAME_FIELD as ObjectType),
+      DIFF_FIELD: getFieldFromElement(customRecordTestObjects.DIFF_FIELD as ObjectType),
+    }
+  }
+
+  const getIDToVal = (testInstances: TestElements, nestedField: string): Map<string, string> => {
+    const idToVal: Map<string, string> = new Map()
+    for (const elem of Object.values(testInstances)) {
```
dantal4, line 88: no for loops!

packages/netsuite-adapter/test/change_validators/unique_fields.test.ts
```diff
@@ -0,0 +1,222 @@
...
+      idToVal.set(elem.elemID.createNestedID(nestedField).getFullName(), (elem as InstanceElement).value[nestedField])
+    }
+    return idToVal
+  }
+
+  const testUniqueness = (
```
dantal4, line 94: Tests should be much more simpler and easier to understand. Please write the tests in blocks of `describe`s & `it`s, and not in another function. We prefer readability and simplicity in tests over code duplication.
After that I'll be able to review the tests, and validate that all scenarios are covered correctly. Now it's too complicated for me to understand.

dantal4 (commented):
Nice and clean work!

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -0,0 +1,218 @@
...
+const getRestrictedType = (elemID: ElemID, includeTypeForFields = true): RestrictedType | undefined => {
+  if (elemID.idType === 'instance') {
+    if (elemID.typeName === SAVED_SEARCH) {
+      return 'savedSearch'
+    }
+
+    if (elemID.typeName === FINANCIAL_LAYOUT) {
+      return 'financialLayout'
+    }
+  } else if ((includeTypeForFields && elemID.idType === 'field') || (!includeTypeForFields && elemID.idType === 'type')) {
```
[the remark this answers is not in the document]
shoshana-houri: I think that's because it's not clear whether the 'type' in `includeTypeForFields` indicates the `idType` field or one of the types, which is 'type' - confusing indeed.
I can rename it as `includeFieldElements`, wdyt?

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -0,0 +1,218 @@
...
+      .flatMap(elemID => getters.getSourceRestrictedFields({ elemID, elementsSource }))
+      .toArray()
+    typeToFields[type] = fields
+  }
+
+  await awu(Object.entries(typeToElementsRecord)).forEach(typeToElements =>
+    setMap(typeToElements[0] as RestrictedType, typeToElements[1]))
+
+  return typeToFields
+}
```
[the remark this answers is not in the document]
shoshana-houri: Loved that, thanks!

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -0,0 +1,218 @@
...
+
+const validateDuplication = async (
+  changes: ChangeDataType[],
+  uniqueFieldToID: Dictionary<number>,
```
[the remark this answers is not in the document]
shoshana-houri: `uniqueFieldToID` is the the output of `countBy` function, which its return type is Dictionary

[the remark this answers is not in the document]
shoshana-houri: I see now it's a safe casting, changing to Record

packages/netsuite-adapter/test/change_validators/unique_fields.test.ts
shoshana-houri: They started as constants and I continued thinking about them like that :sweat_smile:
I'll change that

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -0,0 +1,218 @@
...
+    .map(field => element.fields[field].annotations.scriptid)
+}
+
+
+const savedSearchGetters: RestrictedTypeGetters = {
+  getChangeRestrictedField: change => getChangeNestedField(change, FIELD_DEFAULT_NAME),
+  getSourceRestrictedFields: params => getNestedField(params, FIELD_DEFAULT_NAME),
+  getMessage: () => 'A Saved Search with that title already exists',
+  getDetailedMessage: field => `Can't deploy the Saved Search: ${field} because there is already a Saved Search with that title in the target account.`
+    + ' Please select a Saved Search with a unique title.',
```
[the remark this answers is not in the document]
shoshana-houri: It's the rephrasing suggested by @rozenmanofer [here](https://salto-io.atlassian.net/browse/SALTO-2915), but I agree this one may be less confusing, I'll ask.

packages/netsuite-adapter/test/change_validators/unique_fields.test.ts
```diff
@@ -0,0 +1,384 @@
...
+        { elemID: new ElemID(elemID.elemID.adapter, elemID.elemID.typeName + uniqueField),
+          annotations:
+            { name,
+              [METADATA_TYPE]: CUSTOM_RECORD_TYPE },
+          fields:
+            { custom_custrecord1:
+              { refType: elemID,
+                annotations:
+                  { scriptid: uniqueField } } } }
+      ))
```
dantal4, line 97: should be formatted

packages/netsuite-adapter/test/change_validators/unique_fields.test.ts
```diff
@@ -0,0 +1,384 @@
...
+          [savedSearchTestInstances.basic, savedSearchTestInstances.sameField,
+            financialLayoutTestInstances.basic, financialLayoutTestInstances.sameField,
+            customRecordTestObjects.basic, customRecordTestObjects.sameField]
+        ))
+        expect(changeErrors).toHaveLength(3)
+        expect(changeErrors.map(changeError => changeError.severity)).toEqual(['Error', 'Error', 'Error'])
+        const errorsElemIDs = changeErrors.map(changeError => changeError.elemID)
+        expect(errorsElemIDs.includes(savedSearchTestInstances.sameField.elemID)).toBeTruthy()
+        expect(errorsElemIDs.includes(financialLayoutTestInstances.sameField.elemID)).toBeTruthy()
+        expect(errorsElemIDs.includes(customRecordTestChanges.sameField.elemID)).toBeTruthy()
```
dantal4, line 362: nit-
```suggestion
        expect(errorsElemIDs).toHaveLength(3)
        expect(errorsElemIDs).toEqual(expect.arrayContaining([
           savedSearchTestInstances.sameField.elemID,
           financialLayoutTestInstances.sameField.elemID,
           customRecordTestChanges.sameField.elemID,
        ])
```

dantal4 (commented):
Nice work!
Last comments, and then you're done 👍

dantal4 (approved):
💯

PR #4191, opened 2023-03-30
SALTO-3922: FieldContext projects and issue types are not paginated

alonstern:
Fixed some projects and issue types of field contexts not being paginated and therefore limited to 50

---
_Release Notes_: 
__Jira Adapter__:
- Fixed a bug when fetching a field context projects and issue types that only the first 50 were fetched

---
_User Notifications_: 
__Jira Adapter__:
- Some additional projects and issue types of field contexts might be fetched (only if there a field contexts with more than 50 issue types or projects)


IdoZakk (approved):
👍

PR #4192, opened 2023-03-30
SALTO-3923: fieldContextValidator can return errors on contexts that are not in the plan

alonstern:
Fixed field contexts validator to not return errors about elements in the plan

---
_Release Notes_: 
__Jira Adapter__:
- Fixed field contexts validator to not return errors about elements in the plan

---
_User Notifications_: 
None


IdoZakk (approved):
🌃

PR #4193, opened 2023-03-30
SALTO-3928: 403 when fetching a dashboard fails the entire fetch

alonstern:
Fixed a bug that when we fail to fetch a certain dashboard layout value the entire fetch would fail

---
_Release Notes_: 
__Jira Adapter__:
- Fixed a bug that when we fail to fetch a certain dashboard layout value the entire fetch would fail

---
_User Notifications_: 
None


IdoZakk (approved):
🔨 Look at the comment

PR #4194, opened 2023-03-30
SALTO-3121 Add validation for script id when adding new elements

adfineberg:
_Add validation for script id when adding new elements_
---

---
_Release Notes_: 
Netsuite adapter: 
* Add validation for script id when adding new elements

---
_User Notifications_: 
None

### DIFF
--- a/packages/netsuite-adapter/src/change_validators/immutable_changes.ts
+++ b/packages/netsuite-adapter/src/change_validators/immutable_changes.ts
@@ -139,7 +139,7 @@ const toModificationInstanceErrors = async (
     )
   }
 
-  if (before.value[APPLICATION_ID] !== after.value[APPLICATION_ID]) {
+  if (modificationServiceIdCondition(change, APPLICATION_ID)) {
     modifiedImmutableFields.push(
       after.elemID.createNestedID(APPLICATION_ID).getFullName()
     )
--- a/packages/netsuite-adapter/src/change_validators/immutable_changes.ts
+++ b/packages/netsuite-adapter/src/change_validators/immutable_changes.ts
@@ -61,9 +61,9 @@ const modificationServiceIdCondition = (change: ModificationChange<ChangeDataTyp
   !== getElementValueOrAnnotations(change.data.after)[annoName]
 
 const toModificationTypeErrors = async (change: ModificationChange<ObjectType>): Promise<ChangeError[]> => {
-  const { before, after } = change.data
+  const { after } = change.data
   const modifiedImmutableAnnotations = await typeServiceIdConditions(change, modificationServiceIdCondition)
-  if (before.annotations[APPLICATION_ID] !== after.annotations[APPLICATION_ID]) {
+  if (modificationServiceIdCondition(change, APPLICATION_ID)) {
     modifiedImmutableAnnotations.push(APPLICATION_ID)
   }
   return modifiedImmutableAnnotations.map(modifiedAnno => toModifiedAnnotationChangeError(after, modifiedAnno))
@@ -178,13 +178,13 @@ const toAdditionInstanceErrors = async (
   change: AdditionChange<InstanceElement>
 ): Promise<ChangeError[]> => {
   const { after } = change.data
-  const missingServiceIdAnnotations = await instanceServiceIdConditions(change, additionServiceIdCondition)
+  const missingServiceIdFields = await instanceServiceIdConditions(change, additionServiceIdCondition)
 
   return missingServiceIdFields.map(addedField => ({
     elemID: after.elemID,
     severity: 'Error',
     message: 'Can\'t deploy a field without a ServiceID',
-    detailedMessage: `The ${addedAnno} field is missing a ServiceID.\n`
+    detailedMessage: `The ${addedField} field is missing a ServiceID.\n`
       + 'In order to proceed with this deployment, please edit the element in Salto and add a valid ServiceID.',
   } as ChangeError))
 }
--- a/packages/netsuite-adapter/src/change_validators/immutable_changes.ts
+++ b/packages/netsuite-adapter/src/change_validators/immutable_changes.ts
@@ -44,7 +44,7 @@ const toModifiedAnnotationChangeError = (
     + 'In order to proceed with this deployment, please edit the element in Salto and remove this annotation change.',
 })
 
-const typeServiceIdConditions = <T extends (AdditionChange<ObjectType> | ModificationChange<ObjectType>)>(
+const typeServiceIdConditions = <T extends AdditionChange<ObjectType> | ModificationChange<ObjectType>>(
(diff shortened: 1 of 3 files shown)


packages/netsuite-adapter/src/change_validators/immutable_changes.ts
```diff
@@ -44,39 +44,64 @@ const toModifiedAnnotationChangeError = (
...
 
-const toTypeErrors = async (change: ModificationChange<ObjectType>): Promise<ChangeError[]> => {
-  const { before, after } = change.data
-  const modifiedImmutableAnnotations = await awu(Object.entries(after.annotationRefTypes))
+const typeServiceIdConditions = <T extends (AdditionChange<ObjectType> | ModificationChange<ObjectType>)>(
+  change: T,
+  condition: (change: T, annoName: string) => boolean
+): Promise<string[]> => {
+  const { after } = change.data
+  return awu(Object.entries(after.annotationRefTypes))
```
dantal4, line 52: There can be a case where the service id annotation refType doesn't exist too. In this case we won't catch a missing service id annotation. I think that we should return an error in case that no annotation is the service id (specifically in types).

packages/netsuite-adapter/src/change_validators/immutable_changes.ts
```diff
@@ -44,39 +44,64 @@ const toModifiedAnnotationChangeError = (
...
     modifiedImmutableAnnotations.push(APPLICATION_ID)
   }
   return modifiedImmutableAnnotations.map(modifiedAnno => toModifiedAnnotationChangeError(after, modifiedAnno))
 }
 
-const toFieldErrors = (change: ModificationChange<Field>): ChangeError[] => {
-  const { before, after } = change.data
-  return [SCRIPT_ID]
-    .filter(annoName => before.annotations[annoName] !== after.annotations[annoName])
+const fieldServiceIdConditions = <T extends (AdditionChange<Field> | ModificationChange<Field>)>(
```
dantal4, line 72:
```suggestion
const fieldServiceIdConditions = <T extends AdditionChange<Field> | ModificationChange<Field>>(
```

packages/netsuite-adapter/src/change_validators/immutable_changes.ts
```diff
@@ -44,39 +44,64 @@ const toModifiedAnnotationChangeError = (
...
+}
+
+const modificationServiceIdCondition = (change: ModificationChange<ChangeDataType>, annoName: string): boolean =>
+  getElementValueOrAnnotations(change.data.before)[annoName]
+  !== getElementValueOrAnnotations(change.data.after)[annoName]
+
+const toModificationTypeErrors = async (change: ModificationChange<ObjectType>): Promise<ChangeError[]> => {
+  const { before, after } = change.data
+  const modifiedImmutableAnnotations = await typeServiceIdConditions(change, modificationServiceIdCondition)
   if (before.annotations[APPLICATION_ID] !== after.annotations[APPLICATION_ID]) {
```
dantal4, line 66: This can be replaced with `modificationServiceIdCondition` too

dantal4 (commented):
Nice work!

@rozenmanofer can you help rephrasing the error messages?

dantal4 (commented):
Looks great!
Let's wait for @rozenmanofer  say on the error messages.
Small thing that I saw a couple of times in the code - `async` is unnecessary when there's no `await` inside the function - even when the result type is `Promise<...>`.

packages/netsuite-adapter/src/change_validators/immutable_changes.ts
```diff
@@ -105,23 +130,81 @@ const toInstanceErrors = async (
...
+
+const toAddedMissingAnnotationError = (
+  after: Field | ObjectType,
+  addedAnno: string
+): ChangeError => ({
+  elemID: after.elemID,
+  severity: 'Error',
+  message: 'Can\'t deploy an annotation without a ServiceID',
+  detailedMessage: `This ${addedAnno} annotation is missing a ServiceID.\n`
+    + 'In order to proceed with this deployment, please edit the element in Salto and add a valid ServiceID.',
```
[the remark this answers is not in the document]
rozenmanofer: What's after.elemID.idType?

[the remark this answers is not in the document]
dantal4: It's replaced with "type" / "field" so the message that the user will see would be:
- Can't deploy a type without a service id annotation
- Can't deploy a field without a service id annotation

@rozenmanofer  Do you think we should have one message for both cases?

packages/netsuite-adapter/src/change_validators/immutable_changes.ts
```diff
@@ -39,44 +39,77 @@ const toModifiedAnnotationChangeError = (
...
+  const { after } = change.data
+  const serviceIdRefTypes = await awu(Object.entries(after.annotationRefTypes))
+    .filter(async ([_annoName, refType]) => isServiceId(await refType.getResolvedValue())).toArray()
+  if (serviceIdRefTypes.length === 0) {
+    return [{ type: 'missingRefType' }]
   }
-  return modifiedImmutableAnnotations.map(modifiedAnno => toModifiedAnnotationChangeError(after, modifiedAnno))
+  return serviceIdRefTypes
+    .filter(([annoName, _refType]) => condition(change, annoName))
+    .map(([annoName, _refType]) => ({ type: 'missingAnnotation', value: annoName }))
```
dantal4, line 60: Since the user have no idea what is `script id refType` from the messages, I think that we just need a fallback here that in case there's no `serviceid` annotation we'll check for missing/modified `SCRIPT_ID` and that's it.

packages/netsuite-adapter/src/change_validators/immutable_changes.ts
```diff
@@ -104,24 +137,83 @@ const toInstanceErrors = async (
...
+  change: AdditionChange<InstanceElement>
+): Promise<ChangeError[]> => {
+  const { after } = change.data
+  const missingServiceIdFields = await instanceServiceIdConditions(change, additionServiceIdCondition)
+
+  return missingServiceIdFields.map(addedField => ({
+    elemID: after.elemID,
+    severity: 'Error',
+    message: 'Can\'t deploy a field without a script ID',
+    detailedMessage: `Missing field script ID.\n In order to deploy this ${addedField} field, make sure it has a valid script ID`,
```
dantal4, line 197:
```suggestion
    message: 'Can\'t deploy an instance without its ${addedField}',
    detailedMessage: `Missing ${addedField}.\n In order to deploy this instance, make sure it has a valid ${addedField}`,
```

dantal4 (commented):
Looks good 👌

dantal4 (approved):
💯

PR #4195, opened 2023-03-30
SALTO-3793: Log reasons of config suggestion

idoamit8:
Fixing a nit suggestion from Shir about log content. 

---
This is followup for https://github.com/salto-io/salto/pull/4135
---
_Release Notes_: 
None
---
_User Notifications_: 
None


shir-reifenberg (approved):
👌

PR #4196, opened 2023-03-30
SALTO-1730 - Salesforce: ignore self-references in formulas

tomsellek:
The code we use for parsing extracts all references, which means if the formula refers to a field in the formula field's parent type, then we'll end up with a reference to the field's parent type. This is redundant.

---

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A


packages/salesforce-adapter/src/filters/formula_deps.ts
```diff
@@ -138,8 +138,10 @@ const addDependenciesAnnotation = async (field: Field, allElements: ReadOnlyElem
     const [validReferences, invalidReferences] = await promises.array.partition(references, isValidReference)
     logInvalidReferences(invalidReferences, formula, identifiersInfo)
 
-    log.info(`Extracted ${validReferences.length} valid references`)
-    const depsAsRefExpr = validReferences.map(elemId => ({ reference: new ReferenceExpression(elemId) }))
+    log.debug(`Extracted ${validReferences.length} valid references`)
```
ori-moisis, line 141: is this log actually useful? doesn't seem to provide any information that could be used later on as there doesn't seem to be any context to what is happening here, extracted from where? in what scenario is just the number useful?

I would vote to remove this log or merge it with another log if we have one that says what happened on a field.
unrelated, logs should use `%d` and not `${...}` for string formatting (slightly more efficient, but mostly for convention...)

ori-moisis (commented):
didn't have time to sufficiently understand the code around this change so not approving yet.
if someone else has enough context to approve this, feel free to do so

packages/salesforce-adapter/src/filters/formula_deps.ts
```diff
@@ -138,8 +138,9 @@ const addDependenciesAnnotation = async (field: Field, allElements: ReadOnlyElem
     const [validReferences, invalidReferences] = await promises.array.partition(references, isValidReference)
     logInvalidReferences(invalidReferences, formula, identifiersInfo)
 
-    log.info(`Extracted ${validReferences.length} valid references`)
-    const depsAsRefExpr = validReferences.map(elemId => ({ reference: new ReferenceExpression(elemId) }))
+    const depsAsRefExpr = validReferences
+      .filter(elemId => !elemId.isEqual(field.parent.elemID))
```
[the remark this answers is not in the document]
tomsellek: > Why are we treating it differently than "invalidReferences"? isn't these two are just different reasons to omit the reference?
(isValidReference checks if the reference is not dangling, and this is checking that it's not a self reference?)

Really we want to do a 3-way partition between valid, omitted and invalid references, because they have different meanings as far as the user is concerned. I can replace the partition with 3 calls to `filter`.

>What about references to other fields in the same element? 

Quoting from the comment in the ticket: "Unless @Tomer Mevorach or @Pablo Gonzalez say otherwise, we will keep the dependency on fields but discard the dependency on the parent object."
I don't actually remember why we decided to do it this way.

packages/salesforce-adapter/src/filters/formula_deps.ts
```diff
@@ -135,11 +146,12 @@ const addDependenciesAnnotation = async (field: Field, allElements: ReadOnlyElem
       References: ${references.map(ref => ref.getFullName()).join(', ')}`)
     }
 
-    const [validReferences, invalidReferences] = await promises.array.partition(references, isValidReference)
-    logInvalidReferences(invalidReferences, formula, identifiersInfo)
+    const referencesWithValidity = await groupByAsync(references, referenceValidity)
+
+    logInvalidReferences(referencesWithValidity.invalid ?? [], formula, identifiersInfo)
```
yuribro, line 151: nit: I still think that we should have a log about the `omitted` references (at least a count)

yuribro (approved):

PR #4199, opened 2023-04-02
SALTO-3600, SALTO-3393: Add missing references to Okta

shir-reifenberg:
Add missing references to Okta

---

Add missing references to Okta & move shared code from zendesk to adapter-components
Didn't add config flag as we have in zendesk, if u think it's necessary lmk

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/references/field_references.ts
+++ b/packages/adapter-components/src/references/field_references.ts
@@ -14,8 +14,8 @@
 * limitations under the License.
 */
 import _ from 'lodash'
-import { Field, Element, isInstanceElement, Value, Values, ReferenceExpression, InstanceElement, ElemID, cloneDeepWithoutRefs, isElement, ObjectType } from '@salto-io/adapter-api'
-import { GetLookupNameFunc, GetLookupNameFuncArgs, TransformFunc, transformValues, safeJsonStringify, resolvePath, naclCase } from '@salto-io/adapter-utils'
+import { Field, Element, isInstanceElement, Value, Values, ReferenceExpression, InstanceElement, ElemID, cloneDeepWithoutRefs, isElement } from '@salto-io/adapter-api'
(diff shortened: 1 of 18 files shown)


packages/adapter-components/src/references/field_references.ts
```diff
@@ -38,9 +38,25 @@ const doNothing: ContextFunc = async () => undefined
 const emptyContextStrategyLookup: Record<string, ContextFunc> = {}
 
 export const MISSING_ANNOTATION = 'salto_missing_ref'
+const MISSING_REF_PREFIX = 'missing_'
+
 const checkMissingRef = (element: Element): boolean =>
   element.annotations?.[MISSING_ANNOTATION] === true
 
+export const createMissingInstance = (
```
netama, line 46: nit: I think it might be good to move this to a separate file (e.g. `missing_instances.ts` / `missing_references.ts`), even if we keep the `MISSING_ANNOTATION` here (though we can move it as well). 
the main reason being that there might be some other ways to define these instances (this approach specifically is good when we have instances where there is no info besides the referenced internal id, there might be some other "strategies" we will want to use in adapters that have more info) -  so I think we shouldn't tie it too closely with the "main" logic, and start having a "bank" of potential implementations (like we have for example for pagination functions and auth methods).

packages/adapter-components/src/references/reference_mapping.ts
```diff
@@ -101,7 +101,7 @@ ReferenceSourceTransformation
 export type MissingReferenceStrategy = {
   create: CreateMissingRefFunc
 }
-export type MissingReferenceStrategyName = 'typeAndValue' | 'startsWith'
```
netama, line 104: good catch :)

netama (approved):
looks great :) one small nit about file structure

packages/adapter-components/src/references/field_references.ts
shir-reifenberg: Cool, thanks 🙏

PR #4201, opened 2023-04-02
SALTO-3932: Add deploy for AuthorizationServerPolicy & AuthorizationServerPolicyRule

shir-reifenberg:
Support deployment of `AuthorizationServerPolicy` and `AuthorizationServerPolicyRule` instances

---

In order to deploy `AuthorizationServerPolicyRule`, we need the `AuthorizationServerPolicy` id and the `AuthorizationServer` id, which is not a direct parent of `AuthorizationServerPolicyRule`. Therefore, during fetch we add for each `AuthorizationServerPolicyRule` the related `AuthorizationServer` as a second parent.

---
_Release Notes_: 
None

---
_User Notifications_: 
None


edenhassid (approved):
looks great! 🔥 🔥

PR #4208, opened 2023-04-03
SALTO-3950 All instances with namespace must add it to their fullName if it doesn't exist

benisraeladi:
All instances with namespace must add it to their fullName if it doesn't exist

---

`PermissionSet` was one instance that was found to not have the namespace in its fullName even if the namespace exists

---
_Release Notes_: 
Salesforce:
* Instance elements in namespace will automatically add the namespace to the instance's fullName if `addNamespacePrefix` flag is set
---
_User Notifications_: _None_


packages/salesforce-adapter/src/fetch.ts
```diff
@@ -239,28 +245,31 @@ export const fetchMetadataInstances = async ({
...
-        isFolderType: isDefined(metadataType.annotations[FOLDER_CONTENT_TYPE]),
-      })
-    ).map(({ name }) => name)
+    filePropsToRead.map(({ name }) => name)
   )
 
+  const fullNamesFromRead = new Set(metadataInfos.map(info => info?.fullName))
   const filePropertiesMap = _.keyBy(fileProps, getFullName)
+  const missingMetadata = Object.keys(filePropertiesMap).filter(name => !fullNamesFromRead.has(name))
+  log.debug('Missing metadata with valid fileProps: %o', missingMetadata)
```
ori-moisis, line 269: this will just log the names, I think we want the full properties in this case

also, important, we should not log if there is no missing metadata (which is supposed to be the common case)

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -173,21 +173,27 @@ const listMetadataObjectsWithinFolders = async (
...
-    ? `${obj.namespacePrefix}${NAMESPACE_SEPARATOR}` : ''
-  if (obj.type === LAYOUT_TYPE_ID_METADATA_TYPE && obj.namespacePrefix) {
+  if (obj.namespacePrefix === undefined) {
+    return obj.fullName
+  }
+
+  const namePrefix = `${obj.namespacePrefix}${NAMESPACE_SEPARATOR}`
+  // In some cases, obj.fullName does not contain the namespace prefix even though
+  // obj.namespacePrefix is defined. In these cases, we want to add the prefix manually
+  const fullName = obj.fullName.startsWith(namePrefix) ? obj.fullName : `${namePrefix}${obj.fullName}`
```
ori-moisis, line 183: I think the unit test that fails on this is right
we should not add the prefix to layouts (because a package could contain a layout for a standard object, in which case, the namespace will be only on the layout name, not on the object name which is at the beginning)
sorry for misleading you on that one

ori-moisis:
missing unit tests for the new scenario

packages/salesforce-adapter/src/fetch.ts
benisraeladi: Done, and checked that it would've shown the missing instances if we did have the log

packages/salesforce-adapter/src/fetch.ts
benisraeladi: Done

benisraeladi:
> missing unit tests for the new scenario

Created unit test, it's a little late for me to refactor the test so that it reads better, but now all tests passed when I ran `yarn test` so 🤞

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -184,12 +187,22 @@ const getFullName = (obj: FileProperties): string => {
     if (layoutName.length !== 0 && !layoutName[0].startsWith(obj.namespacePrefix)) {
```
ori-moisis, line 187: unrelated fix that we should add IMO is to change this to
```typescript
if (layoutName.length !== 0 && !layoutName[0].startsWith(namePrefix)) {
```
this is because it does not catch a case where we are missing the namespace prefix but the layout name just happens to start with the namespace prefix.
e.g - if the namespace prefix is "bla" and the layout name is "bla layout", we should change it to "bla__bla layout" but we won't because the condition currently doesn't contain the namespace separator

we can do this in a separate PR if you prefer

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -239,25 +254,32 @@ export const fetchMetadataInstances = async ({
 
   const metadataTypeName = await apiName(metadataType)
 
+  const filePropsToRead = fileProps
+    .map(prop => ({
+      ...prop,
+      fullName: getFullName(prop, addNamespacePrefixToFullName),
+      namespacePrefix: getNamespace(prop),
```
ori-moisis, line 261: this is actually incorrect behavior.
looking at what `getNamespace` does, it will replace the namespace prefix in a way that we probably don't want it to do for anything other than the call to `isInstanceMatch`

I think we should move this to be called in the `filter` part (i.e - remove this line and change line 264 to `namespace: getNamespace(namespacePrefix)`)

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -350,7 +374,7 @@ export const retrieveMetadataInstances = async ({
       const chunkSize = Math.ceil(fileProps.length / 2)
       log.debug('reducing retrieve item count %d -> %d', fileProps.length, chunkSize)
       configChanges.push({ type: MAX_ITEMS_IN_RETRIEVE_REQUEST, value: chunkSize })
-      return (await Promise.all(_.chunk(fileProps, chunkSize).map(retrieveInstances))).flat()
+      return (await Promise.all(_.chunk(fileProps, chunkSize).map(file => retrieveInstances(file)))).flat()
```
ori-moisis, line 377: I think we don't need this change...

packages/salesforce-adapter/test/fetch_installed_packages.test.ts
```diff
@@ -0,0 +1,222 @@
...
+  const fetch = async ({
+    fileProp,
+    mockType,
+    addNamespacePrefixToFullName,
+  }:
+    {
+        fileProp: MockFilePropertiesInput
+        mockType: ObjectType
+        addNamespacePrefixToFullName: boolean
+    }): Promise<InstanceElement> => {
```
ori-moisis, line 34: style nit
```suggestion
  type MockFetchArgs = {
    fileProp: MockFilePropertiesInput
    mockType: ObjectType
    addNamespacePrefixToFullName: boolean
  }
  const fetch = async (
    { fileProp, mockType, addNamespacePrefixToFullName }: MockFetchArgs
  ): Promise<InstanceElement> => {
```

packages/salesforce-adapter/test/fetch_installed_packages.test.ts
```diff
@@ -0,0 +1,222 @@
...
+
+    const { elements } = await fetchMetadataInstances({
+      client,
+      fileProps: [mockFileProperties(fileProp)],
+      metadataType: mockType,
+      metadataQuery,
+      addNamespacePrefixToFullName,
+    })
+    return elements
+      .filter(isInstanceElement)[0]
```
ori-moisis, line 54: style nit:
```suggestion
    return elements.filter(isInstanceElement)[0]
```

packages/salesforce-adapter/test/fetch_installed_packages.test.ts
ori-moisis: also, technically, this may return `undefined`, so the return type should be `Promise<InstanceElement | undefined>`

packages/salesforce-adapter/test/fetch_installed_packages.test.ts
```diff
@@ -0,0 +1,222 @@
...
+        const fileProp = mockFileProperties({ fullName: fullNameFromList, type: 'PermissionSet', namespacePrefix: 'Test' })
+        const instance = await fetch({ fileProp, mockType: mockTypes.PermissionSet, addNamespacePrefixToFullName })
+
+        expect(instance).toBeDefined()
+        expect(instance?.value).toHaveProperty('fullName', expectedFullName)
+      })
+    })
+  })
+  describe('Test fetching layouts of installed packages', () => {
+    // const fetch = async ({
```
ori-moisis, line 84: remove commented out code

packages/salesforce-adapter/test/fetch_installed_packages.test.ts
```diff
@@ -0,0 +1,222 @@
...
+    //     .filter(isInstanceElement)
+    //     .find(inst => inst.elemID.typeName === constants.LAYOUT_TYPE_ID_METADATA_TYPE)
+    //   const targetFullName = `${apiName}-SBQQ__Test Layout`
+    //   expect(instance).toBeDefined()
+    //   expect(instance?.elemID).toEqual(LAYOUT_TYPE_ID.createNestedID('instance', naclCase(targetFullName)))
+    // }
+    describe('if the API name already includes namespacePrefix', () => {
+      describe('if layout name does not include prefix', () => {
+        it('should add prefix to layout name', async () => {
+          const apiName = 'SBQQ__TestApiName__c'
```
ori-moisis, line 168: nit: `apiName` is a bit misleading here, this is not the full api name of the layout, it is just the object name.
I would call this `layoutObjectName` or just `objectName`

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -133,6 +133,9 @@ export const listMetadataObjects = async (
 const getNamespace = (obj: FileProperties): string => (
   obj.namespacePrefix === undefined || obj.namespacePrefix === '' ? DEFAULT_NAMESPACE : obj.namespacePrefix
 )
+const getNamespaceFromNamespacePrefix = (namespacePrefix: string|undefined): string => (
```
ori-moisis, line 136: we don't need this function
also, we should not repeat logic (from `getNamespace`)

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -239,25 +257,31 @@ export const fetchMetadataInstances = async ({
...
+    .map(prop => ({
+      ...prop,
+      fullName: getFullName(prop, addNamespacePrefixToFullName),
+    }))
+    .filter(({ fullName, namespacePrefix }) => metadataQuery.isInstanceMatch({
+      namespace: getNamespaceFromNamespacePrefix(namespacePrefix),
+      metadataType: metadataTypeName,
+      name: fullName,
+      isFolderType: isDefined(metadataType.annotations[FOLDER_CONTENT_TYPE]),
+    }))
```
ori-moisis, line 270:
```suggestion
    .filter(prop => metadataQuery.isInstanceMatch({
      namespace: getNamespace(prop),
      metadataType: metadataTypeName,
      name: prop.fullName,
      isFolderType: isDefined(metadataType.annotations[FOLDER_CONTENT_TYPE]),
    }))
```

ori-moisis (approved):

PR #4211, opened 2023-04-04
Salto 3670 redundant ns service url queries

adfineberg:
_Removed unnecessary queries to Netsuite from the service url filter._
Also added internal ids for saved searches which allows them to be used in other filters.

---

When the service url filter is run, the internal ids are already fetched into the system, there is no need to fetch them again.

---
_Release Notes_: 
_Removed unnecessary queries to Netsuite from the service url filter._

---
_User Notifications_:


packages/netsuite-adapter/src/service_url/file_cabinet.ts
```diff
@@ -14,32 +14,22 @@
...
-  if (id === undefined) {
-    log.warn(`Did not find the internal id of ${element.elemID.getFullName()}`)
-    return undefined
-  }
+import { setElementsUrls } from './elements_urls'
+import { ServiceUrlSetter } from './types'
 
-  return isFileInstance(element)
+const generateUrl = (id: number, element: InstanceElement): string | undefined =>
+  (isFileInstance(element)
```
dantal4, line 23: a little bit nicer IMO
```suggestion
const generateUrl = (id: number, element: InstanceElement): string | undefined => (
  isFileInstance(element)
```

packages/netsuite-adapter/src/service_url/file_cabinet.ts
```diff
@@ -14,32 +14,22 @@
...
     ? `app/common/media/mediaitem.nl?id=${id}`
-    : `app/common/media/mediaitemfolder.nl?id=${id}`
-}
+    : `app/common/media/mediaitemfolder.nl?id=${id}`)
```
dantal4, line 25:
```suggestion
    : `app/common/media/mediaitemfolder.nl?id=${id}`
)
```

packages/netsuite-adapter/test/filters/internal_ids/sdf_internal_ids.test.ts
```diff
@@ -244,6 +271,7 @@ describe('sdf internal ids tests', () => {
             toChange({ before: accountInstance, after: accountInstance }),
             toChange({ before: customRecordType, after: customRecordType }),
             toChange({ before: customScriptInstance, after: customScriptInstance }),
+            toChange({ after: savedSearchInstance }),
```
dantal4, line 274: I think that this should be a modification change?
Since it's working, I think that there's a problem with the code/test, as we would expect an added saved search to get an internal id, right?

dantal4 (commented):
Nice work!

packages/netsuite-adapter/test/filters/internal_ids/sdf_internal_ids.test.ts
adfineberg: The problem was that there was no mocked value, so no internal id was created (which is the right flow, just not what we wanted to test here). I updated the test for the right case.

dantal4 (approved):
Very clean 💯

PR #4213, opened 2023-04-09
SALTO-3913: Identify successful global value set deploy

ori-moisis:
When deploying a change to a user defined global value set that was created before version 57.0, the success response may come back with a different fullName (the salesforce API will sometimes append a __gvs suffix, but not always)

Changed the code to handle this case and identify the deployed component as successful

---

---
_Release Notes_: 
_Salesforce adapter_:
- Fixed issue where deploying changes to global value sets would sometimes not be considered successful even if the change was successful

---
_User Notifications_: 
_None_


tamtamirr (commented):
@ori-moisis Since `GlobalValueSet` instances are always custom (The standard ones would be of type `StandardValueSet`), wouldn't it be better to append the `__gvs` suffix on all of the instances upon fetch? This will avoid inconsistency, by always having the correct `fullName` of these instances.

ori-moisis:
I don't know if the claims you make are correct.
as long as the salesforce API (in `list` and `read` and `retrieve` and sometimes in `deploy` as well) says the fullNames of these global value sets don't have `__gvs` in them, I don't think we can decide they do

packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -255,6 +257,39 @@ const getChangeError = async (change: Change): Promise<string | undefined> => {
...
+    .map(deployedIds => deployedIds[GLOBAL_VALUE_SET])
+    .filter(values.isDefined)
+    .forEach(valueSetIds => {
+      const idsToAdd = wu(valueSetIds)
+        .filter(id => !id.endsWith('__gvs'))
+        .map(id => `${id}__gvs`)
+        // If there is another global value set with the same name and the __gvs suffix
+        // in this deployment, it means the one we are currently looking at must be a "standard"
+        // global value set, and we should really not add the __gvs suffix to it
+        .filter(idWithSuffix => !allDeployedGlobalValueSetIDs.has(idWithSuffix))
```
tamtamirr, line 287: There's no such thing as standard `GlobalValueSet` instances. I think we can safely remove this lines

packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -255,6 +257,39 @@ const getChangeError = async (change: Change): Promise<string | undefined> => {
...
+  const allDeployedGlobalValueSetIDs = new Set(
+    globalValueSetIDDeployedIds.flatMap(deployedIds => Array.from(deployedIds[GLOBAL_VALUE_SET]))
+  )
+
+  globalValueSetIDDeployedIds
+    .map(deployedIds => deployedIds[GLOBAL_VALUE_SET])
+    .filter(values.isDefined)
+    .forEach(valueSetIds => {
+      const idsToAdd = wu(valueSetIds)
+        .filter(id => !id.endsWith('__gvs'))
```
tamtamirr, line 282: - nit: Consider `const GLOBAL_VALUE_SET_INSTANCE_SUFFIX = 'gvs'`
- In the same context (But unrelated) please add `gvs` to [INSTANCE_SUFFIXES](https://github.com/salto-io/salto/blob/4284afff82ab0fc8bfa86a530c07b37f7bdd156c/packages/salesforce-adapter/src/types.ts#L52). Otherwise, we might have weird behavior around logics that consider this list (Like `hasNamespace`)

packages/salesforce-adapter/test/adapter.crud.test.ts
```diff
@@ -1725,5 +1726,57 @@ describe('SalesforceAdapter CRUD', () => {
         expect(resultProfile.appliedChanges).toHaveLength(1)
       })
     })
+
+    describe('when deploying a global value set', () => {
```
tamtamirr, line 1730:
```suggestion
    describe('when deploying a GlobalValueSet', () => {
```
nit

packages/salesforce-adapter/test/adapter.crud.test.ts
```diff
@@ -1725,5 +1726,57 @@ describe('SalesforceAdapter CRUD', () => {
...
+            componentSuccess: [{ fullName, componentType: GLOBAL_VALUE_SET }],
+          }))
+          result = await adapter.deploy({ changeGroup })
+        })
+        it('should apply the change', () => {
+          expect(changesToFullNames(result.appliedChanges)).toEqual(
+            changesToFullNames(changeGroup.changes)
+          )
+        })
+      })
```
tamtamirr, line 1758: Is `describe.each` really necessary here? Won't it be clear to explicitly define two `it`s instead?

packages/salesforce-adapter/test/adapter.crud.test.ts
```diff
@@ -1725,5 +1726,57 @@ describe('SalesforceAdapter CRUD', () => {
...
+      describe('when deploying both changes with and without __gvs suffix', () => {
+        let changeGroup: ChangeGroup
+        beforeEach(() => {
+          changeGroup = createChangeGroup(['MyValueSet', 'MyValueSet__gvs'])
+        })
+        describe.each([
+          { successFullNames: ['MyValueSet'] },
+          { successFullNames: ['MyValueSet__gvs'] },
+          { successFullNames: ['MyValueSet', 'MyValueSet__gvs'] },
```
tamtamirr, line 1767: This scenario is not possible

tamtamirr (commented):
Good work! Added some comments.

packages/salesforce-adapter/src/metadata_deploy.ts
ori-moisis: `INSTANCE_SUFFIXES`, while strangely named, does not actually list all possible metadata instance suffixes, it only has the ones that relate to custom objects and their fields.
do you think we should extend this to assume we know all suffixes for all metadata instances?
if so, I will add it there...

tamtamirr: @ori-moisis I think we should as I already noticed a bug when trying to deploy a new `GlobalValueSet` instance with full name `globalValueSet__gvs` in the `package.ts` Change Validator. I got a warning that I'm trying to deploy instance from a managed package with name `globalValueSet`.

I'm not sure the name is misleading, since `CustomObject` and `CustomField` are kind of `Metadata Instances` themselves, but I definitely think this same list should include all the possible custom suffixes. (Perhaps we could rename it to `CUSTOM_SUFFIXES` instead?)

packages/salesforce-adapter/src/metadata_deploy.ts
ori-moisis: hmmm
not sure about that, but this comment did get me to test what would happen if we try to actually deploy this type of conflict (i.e - the "same" value set twice in the same deployment) and it seems like that causes salesforce to crash).
so I guess there isn't too much of a point in handling this case, I'll remove that code (which should make this change much smaller)

packages/salesforce-adapter/test/adapter.crud.test.ts
ori-moisis: it would have to repeat the setup and the test, the `each` seemed readable enough to me, but if you think this is not readable I can change test...

packages/salesforce-adapter/src/metadata_deploy.ts
ori-moisis: added

packages/salesforce-adapter/test/adapter.crud.test.ts
ori-moisis: it crashes (https://github.com/salto-io/salto/pull/4213#discussion_r1161629509)

packages/salesforce-adapter/test/adapter.crud.test.ts
ori-moisis: changed to copy-paste instead of `each`
if you think the `each` is better let me know and I'll change it back

tamtamirr: @ori-moisis I personally don't mind the trade-off between copy-pastes and readability in tests :)

tamtamirr (approved):
LGTM 👑

PR #4215, opened 2023-04-09
SALTO-3969 use safeJsonStringify in merge logs

netama:
Avoid logging inside reference expressions etc.

we lose _some_ information, but I think it is usually better than the current state.

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/workspace/src/merger/internal/common.ts
+++ b/packages/workspace/src/merger/internal/common.ts
@@ -14,9 +14,9 @@
 * limitations under the License.
 */
 import _ from 'lodash'
-import { inspect } from 'util'
 import { types } from '@salto-io/lowerdash'
 import { ElemID, SaltoElementError, SeverityLevel } from '@salto-io/adapter-api'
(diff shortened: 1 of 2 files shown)


ori-moisis (approved):
IIRC inspect stops at a certain depth whereas json stringify will not stop.
this might mean we produce much bigger logs.
not sure if that is the intention here...

a possible alternative would be to call `cloneDeepWithoutRefs` on the value before passing to inspect.

anyway, in case this was the intention, approving the PR

netama:
the intention is to make the logs smaller (by avoiding huge logs inside reference expressions), I _think_ this is worth the additional depth (this is what we did in other places), was wondering more if we lose something functionally in these logs by not having the full values - but I think given that we have the ids we can usually resurrect the info

netama:
discussed IRL - merging, we can add some max-level replacer as well if we see too-large logs after the change

PR #4218, opened 2023-04-09
SALTO-3809: Deploy addition change of instance in different status from the default status

shir-reifenberg:
Allow creating instances in any status

---

In Okta, each type has its default status when created. For example `GroupRule` instances are created in status `INACTIVE` by default, and `AccessPolicyRule` created in status `ACTIVE`.
Until now, instances were created with their default status. This PR add the option to created instance in a different status from the default one.

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/deployment.ts
+++ b/packages/okta-adapter/src/deployment.ts
@@ -41,6 +41,10 @@ type OktaError = {
   }[]
 }
 
+type ResponseWithStatus = {
+  status: string
+}
+
 const OKTA_ERROR_SCHEMA = Joi.object({
   errorSummary: Joi.string().required(),
   errorCauses: Joi.array().items(Joi.object({
@@ -48,7 +52,13 @@ const OKTA_ERROR_SCHEMA = Joi.object({
   }).unknown(true).optional()),
 }).unknown(true)
 
+const RESPONSE_WITH_STATUS = Joi.object({
+  status: Joi.string().valid(ACTIVE_STATUS, INACTIVE_STATUS).allow(),
+}).unknown(true)
+
 const isOktaError = createSchemeGuard<OktaError>(OKTA_ERROR_SCHEMA, 'Received an invalid error')
+
+const isResponseWithStatus = createSchemeGuard<ResponseWithStatus>(RESPONSE_WITH_STATUS, 'Response does not have a valid status field')
 
 export const getOktaError = (elemID: ElemID, error: Error): Error => {
   if (!(error instanceof clientUtils.HTTPError)) {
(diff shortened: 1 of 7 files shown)


packages/okta-adapter/src/deployment.ts
```diff
@@ -68,26 +68,20 @@ export const getOktaError = (elemID: ElemID, error: Error): Error => {
...
-}
-
-export const isDeactivationChange = (change: ModificationChange<InstanceElement>): boolean => {
-  const statusBefore = change.data.before.value.status
-  const statusAfter = change.data.after.value.status
-  return statusBefore === ACTIVE_STATUS && statusAfter === INACTIVE_STATUS
-}
+export const isActivationChange = ({ before, after }: {before: string; after: string}): boolean =>
+  before === INACTIVE_STATUS && after === ACTIVE_STATUS
+export const isDeactivationChange = ({ before, after }: {before: string; after: string}): boolean =>
```
netama, line 73: tiny nit: missing newline

packages/okta-adapter/src/deployment.ts
```diff
@@ -186,9 +181,25 @@ export const defaultDeployWithStatus = async (
       fieldsToIgnore,
       queryParams
     )
+
+    // Update status for the created instance if necessary
+    if (isAdditionChange(change) && !Array.isArray(response)) {
```
netama, line 186: what's the meaning of the `isArray` check? when do we expect the response to be an array? (can we either add a comment with an explanation, or make the check more explicit / add logs for unexpected cases)?

packages/okta-adapter/src/deployment.ts
```diff
@@ -186,9 +181,25 @@ export const defaultDeployWithStatus = async (
...
+      const returnedStatus = response?.status
+      if (typeof returnedStatus !== 'string') {
```
netama, line 188: technically it can also be a string with an unexpected value, right? may be worth logging a warning in that case? (an error might be too noisy)

packages/okta-adapter/src/deployment.ts
```diff
@@ -186,9 +181,25 @@ export const defaultDeployWithStatus = async (
...
+        return response
+      }
+      const changeStatus = getChangeData(change).value.status
+      if (isActivationChange({ before: returnedStatus, after: changeStatus })) {
+        await deployStatusChange(change, client, apiDefinitions, 'activate')
```
netama, line 193: maybe add a debug log of why we're making the extra call(s)?

netama (approved):
looks good 🎉  added a question and minor comments, but LGTM otherwise so approving

packages/okta-adapter/src/deployment.ts
shir-reifenberg: That's because `ResponseResult` (which returned by `defaultDeployChange`) can be either `ResponseValue` or `ResponseValue[]`.
There is already log in `defaultDeployChange` for unexpected result, do you think we should add here as well?

packages/okta-adapter/src/deployment.ts
shir-reifenberg: Sure :)

packages/okta-adapter/src/deployment.ts
shir-reifenberg: @netama actually replaced with joi validation

PR #4219, opened 2023-04-09
SALTO-3901: fix deployment of macro restriction modification

edenhassid:
fix deployment of macro restriction modification

---

To modify the macro restriction to all agents we need to pass in the request `restriction = null`

---
_Release Notes_: 
zendesk: 
* fix bug where restrictions were not removed from macros that already had restrictions defined

---
_User Notifications_: 
None


netama (approved):
🎈

netama:
maybe make the release notes more specific? something like "fix bug where restrictions were not removed from macros that already had restrictions defined"?

PR #4220, opened 2023-04-09
SALTO-3607 Add go to service in Okta 

idoamit8:
1. I added service_url filter in order to add go to service in Salto SaaS using Okta. 
2. I added service_url tests. 
3. I added a function in the client that adds '-admin' to the base-url. 
4. I modified the config file to include service url. 
---

1. We might want to change `baseUrlWithAdmin()` in the future. 
---

_Release Notes_: 
None

---
_User Notifications_: 
None


packages/okta-adapter/src/client/client.ts
```diff
@@ -58,6 +58,12 @@ export default class OktaClient extends clientUtils.AdapterHTTPClient<
     )
   }
 
+  public get baseUrlWithAdmin(): string {
```
shir-reifenberg, line 61: As we have another usecase for adding the `-admin` to the `baseUrl` ([here](https://github.com/salto-io/salto/pull/4170)), let's separate this to 2 different functions - 
1. `getBaseUrl` which will be a class method and would return just the `baseUrl`
2.  another function (not part of the class) that gets the `baseUrl` and return it with the `-admin` - already implemented in [this PR](https://github.com/salto-io/salto/pull/4170/files#diff-0e2b7020d1ed8ca4a0628e02dd1f26705307e102acf23ca43e601d59e23fb5b0) so you can just copy for now

packages/okta-adapter/src/config.ts
```diff
@@ -56,32 +56,41 @@ type PolicyTypeNames = 'AccessPolicy' | 'IdentityProviderPolicy' | 'MultifactorE
 type PolicyParams = {
   queryParam: string
   ruleName: string
+  policyUrl: string
+  policyRuleUrl?: string
```
shir-reifenberg, line 60: I think we should name it:
`policyServiceUrl` and `ruleServiceUrl`

packages/okta-adapter/src/filters/service_url.ts
```diff
@@ -0,0 +1,75 @@
...
+  try {
+    const userTypeId = getParent(instance).value.id
+    const url = `/admin/universaldirectory#okta/${userTypeId}`
+    instance.annotations[CORE_ANNOTATIONS.SERVICE_URL] = (new URL(url, baseUrl)).href
+  } catch {
+    log.warn(`Failed to create serviceUrl for ${instance.elemID.name}`)
+  }
+}
+
+const serviceUrlFilterCreator: FilterCreator = ({ client, config }) => ({
```
shir-reifenberg, line 46: After seeing the code, I think it would be best to avoid the code duplication here and create 
 a `FilterCreator` map that would contain the adapter-components service url filter and the okta service url filter for the more complex cases,  similar to this: https://github.com/salto-io/salto/blob/main/packages/okta-adapter/src/filters/common.ts

Then we can make sure both filters are running one after the other, like in [here](https://github.com/salto-io/salto/blob/main/packages/okta-adapter/src/adapter.ts#L77) 
WDYT?

packages/okta-adapter/test/filters/service_url.test.ts
```diff
@@ -0,0 +1,115 @@
...
+    beforeEach(async () => {
+      jest.clearAllMocks()
+      const mockCli = mockClient()
+      client = mockCli.client
+      filter = serviceUrlFilterCreator(getFilterParams({ client })) as typeof filter
+    })
+
+    describe('cheking object types that use addUrlToInstance() and can be filtered using the config', () => {
+      const appType = new ObjectType({ elemID: new ElemID(OKTA, APPLICATION_TYPE_NAME) })
+      const appInstance = new InstanceElement(
```
shir-reifenberg, line 38: Can you please add another app that is a custom application? and validate the test pass for both cases?
That's cause wer'e relying of `name` field in the `serviceUrl` which doesn't exist for custom apps.

packages/okta-adapter/test/filters/service_url.test.ts
```diff
@@ -0,0 +1,115 @@
...
+          await filter.onFetch(elements)
+          const [instance] = elements
+          expect(instance.annotations?.[CORE_ANNOTATIONS.SERVICE_URL]).toEqual(
+            `${client.baseUrlWithAdmin}/admin/app/${appInstance.value.name}/instance/${appInstance.value.id}/#tab-general`,
+          )
+        })
+      })
+      describe('onDeploy', () => {
+        it('should add service url annotation for application if it is exist in the config', async () => {
+          const changes = [appInstance].map(e => e.clone()).map(inst => toChange({ after: inst }))
```
shir-reifenberg, line 55: same here :) please add a custom app as well

packages/okta-adapter/test/filters/service_url.test.ts
```diff
@@ -0,0 +1,115 @@
...
+    let filter: FilterType
+
+    beforeEach(async () => {
+      jest.clearAllMocks()
+      const mockCli = mockClient()
+      client = mockCli.client
+      filter = serviceUrlFilterCreator(getFilterParams({ client })) as typeof filter
+    })
+
+    describe('cheking object types that use addUrlToInstance() and can be filtered using the config', () => {
```
shir-reifenberg, line 36:
```suggestion
    describe('checking object types that use addUrlToInstance() and can be filtered using the config', () => {
```

tiny nit: it can be shortened to `types using serviceUrl config`

packages/okta-adapter/src/client/client.ts
shir-reifenberg: Also let's add a test for (1) as well :)

packages/okta-adapter/src/filters/service_url.ts
```diff
@@ -0,0 +1,75 @@
...
+} from '../constants'
+
+const log = logger(module)
+const { addUrlToInstance } = filters
+const createServiceUrlUserSchema = (instance: InstanceElement, baseUrl:string): void => {
+  try {
+    const userTypeId = getParent(instance).value.id
+    const url = `/admin/universaldirectory#okta/${userTypeId}`
+    instance.annotations[CORE_ANNOTATIONS.SERVICE_URL] = (new URL(url, baseUrl)).href
+  } catch {
```
shir-reifenberg, line 41:
```suggestion
  } catch (error) {
```

packages/okta-adapter/src/filters/service_url.ts
```diff
@@ -0,0 +1,75 @@
...
+    log.warn(`Failed to create serviceUrl for ${instance.elemID.name}`)
```
shir-reifenberg, line 42: let's use `instance.elemID.getFullName` (so we would have the type name) and log the error as well

shir-reifenberg (commented):
Looks good! I think we can make sure both filters run together without duplicating code (suggested inside), rest of the comments are mainly small things 👌

packages/okta-adapter/test/utils.ts
```diff
@@ -58,7 +58,7 @@ export const mockClient = (): ClientWithMockConnection => {
   const connection = mockConnection()
   const client = new OktaClient({
     credentials: {
-      baseUrl: 'https://bla.okta.net/',
+      baseUrl: 'https://dev-00000000.okta.com',
```
shir-reifenberg, line 61: nit: do we need this? :)

packages/okta-adapter/test/filters/additional_service_url.test.ts
```diff
@@ -0,0 +1,87 @@
...
+    let filter: FilterType
+
+    beforeEach(async () => {
+      jest.clearAllMocks()
+      const mockCli = mockClient()
+      client = mockCli.client
+      filter = additionalServiceUrlFilter(getFilterParams({ client })) as typeof filter
+    })
+
+    describe('cheking user schame object type', () => {
```
shir-reifenberg, line 36:
```suggestion
    describe('checking UserSchema type', () => {
```

packages/okta-adapter/test/filters/additional_service_url.test.ts
```diff
@@ -0,0 +1,87 @@
...
+        it('should not add service url annotation for userSchema if it has no parents annotation', async () => {
+          const elements = [userSchemaInstaceWithoutParents.clone()]
+          delete elements[0].annotations[CORE_ANNOTATIONS.PARENT]
+          await filter.onFetch(elements)
+          const [instance] = elements
+          expect(instance.annotations?.[CORE_ANNOTATIONS.SERVICE_URL]).toBeUndefined()
+        })
+      })
+      describe('onDeploy', () => {
+        it('should add service url annotation for userSchema if it has userType as reference expression', async () => {
```
shir-reifenberg, line 77:
```suggestion
        it('should add service url annotation for userSchema if it has userType as parent', async () => {
```

packages/okta-adapter/test/filters/additional_service_url.test.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    describe('cheking user schame object type', () => {
+      const userSchemaType = new ObjectType({ elemID: new ElemID(OKTA, USER_SCHEMA_TYPE_NAME) })
+      const userTypeType = new ObjectType({ elemID: new ElemID(OKTA, USERTYPE_TYPE_NAME) })
+      const userTypeInstance = new InstanceElement(
+        'user1',
+        userTypeType,
+        { id: 11, name: 'user1' }
+      )
+      const userSchemaInstace = new InstanceElement(
```
shir-reifenberg, line 44:
```suggestion
      const userSchemaInstance = new InstanceElement(
```

packages/okta-adapter/src/filters/additional_service_url.ts
```diff
@@ -0,0 +1,73 @@
...
+
+const log = logger(module)
+
+const createServiceUrlUserSchema = (instance: InstanceElement, baseUrl:string): void => {
+  try {
+    const userTypeId = getParent(instance).value.id
+    const url = `/admin/universaldirectory#okta/${userTypeId}`
+    instance.annotations[CORE_ANNOTATIONS.SERVICE_URL] = (new URL(url, baseUrl)).href
+  } catch (error) {
+    log.warn(`Failed to create serviceUrl for ${instance.elemID.getFullName()}`)
```
shir-reifenberg, line 42: Please log the error here as well :)

packages/okta-adapter/src/filters/service_url.ts
```diff
@@ -0,0 +1,30 @@
...
+*/
+
+
+import { filters } from '@salto-io/adapter-components'
+import OktaClient, { getAdminUrl } from '../client/client'
+import { OktaConfig } from '../config'
+import { FilterCreator, FilterResult } from '../filter'
+
+const filter: FilterCreator = params => {
+  const baseUrl = getAdminUrl(params.client.baseUrl) as string
```
shir-reifenberg, line 24: We shouldn't use casting here and we should validate `baseUrl` is not undefined

shir-reifenberg (commented):
Great :) added a few more comments

packages/okta-adapter/src/filters/additional_service_url.ts
idoamit8: I logged the error in the same message. Is it good? or did you mean to do it in another message?

packages/okta-adapter/src/client/client.ts
```diff
@@ -95,3 +99,9 @@ export default class OktaClient extends clientUtils.AdapterHTTPClient<
       : undefined
   }
 }
+
+/**
+ * Each Okta org has an administrator URL which is used to sign in to the admin console
+ * and used in order to fetch data from private APIs
+ * For more info: https://developer.okta.com/docs/concepts/okta-organizations/#org-urls
+ */
```
shir-reifenberg, line 107: please remove :)

packages/okta-adapter/src/filters/service_url.ts
```diff
@@ -0,0 +1,84 @@
...
+import { FilterCreator } from '../filter'
+import {
+  USER_SCHEMA_TYPE_NAME,
+} from '../constants'
+import { getAdminUrl } from '../client/admin'
+
+const log = logger(module)
+const { addUrlToInstance } = filters
+
+const createServiceUrlUserSchema = (instance: InstanceElement, baseUrl:string): void => {
```
shir-reifenberg, line 37:
```suggestion
const createServiceUrlUserSchema = (instance: InstanceElement, baseUrl: string): void => {
```

packages/okta-adapter/src/filters/service_url.ts
```diff
@@ -0,0 +1,84 @@
...
+    instance.annotations[CORE_ANNOTATIONS.SERVICE_URL] = (new URL(url, baseUrl)).href
+  } catch (error) {
+    log.warn(`Failed to create serviceUrl for ${instance.elemID.getFullName()}. Error: ${error.message}}`)
+  }
+}
+
+const serviceUrlFilter: FilterCreator = ({ client, config }) => ({
+  name: 'serviceUrlFilter',
+  onFetch: async (elements: Element[]) => {
+    const baseUrl = getAdminUrl(client.baseUrl) as string
```
shir-reifenberg, line 50: please remove casing here :)
also, I think we can move `if (baseUrl === undefined)` check here, cause we can check this once instead of checking for each instance (this doesn't change)

packages/okta-adapter/src/filters/service_url.ts
```diff
@@ -0,0 +1,84 @@
...
+        }
+        if (instance.elemID.typeName === USER_SCHEMA_TYPE_NAME) {
+          createServiceUrlUserSchema(instance, baseUrl)
+          return
+        }
+        addUrlToInstance(instance, baseUrl, config)
+      })
+  },
+  onDeploy: async (changes: Change<InstanceElement>[]) => {
+    const baseUrl = getAdminUrl(client.baseUrl) as string
```
shir-reifenberg, line 66: same as the comment above :)

packages/okta-adapter/test/filters/service_url.test.ts
```diff
@@ -0,0 +1,119 @@
...
+import { filterUtils } from '@salto-io/adapter-components'
+import { CORE_ANNOTATIONS, ElemID, InstanceElement, ObjectType, ReferenceExpression, getChangeData, toChange } from '@salto-io/adapter-api'
+import { getFilterParams, mockClient } from '../utils'
+import OktaClient from '../../src/client/client'
+import { getAdminUrl } from '../../src/client/admin'
+import serviceUrlFilter from '../../src/filters/service_url'
+import { APPLICATION_TYPE_NAME, OKTA, USERTYPE_TYPE_NAME, USER_SCHEMA_TYPE_NAME } from '../../src/constants'
+
+describe('serviceUrlFilter', () => {
+  let client: OktaClient
```
shir-reifenberg, line 26: nit: missing indentation

packages/okta-adapter/src/filters/service_url.ts
```diff
@@ -47,14 +47,14 @@ const createServiceUrlUserSchema = (instance: InstanceElement, baseUrl:string):
 const serviceUrlFilter: FilterCreator = ({ client, config }) => ({
   name: 'serviceUrlFilter',
   onFetch: async (elements: Element[]) => {
-    const baseUrl = getAdminUrl(client.baseUrl) as string
+    const baseUrl = getAdminUrl(client.baseUrl)
+    if (baseUrl === undefined) {
+      log.warn('Failed to create baseUrl for instances')
```
shir-reifenberg, line 52: I think we should change message to something like
`'Failed to run serviceUrlFilter, because baseUrl could not be found'`

packages/okta-adapter/src/filters/service_url.ts
```diff
@@ -63,15 +63,15 @@ const serviceUrlFilter: FilterCreator = ({ client, config }) => ({
       })
   },
   onDeploy: async (changes: Change<InstanceElement>[]) => {
-    const baseUrl = getAdminUrl(client.baseUrl) as string
+    const baseUrl = getAdminUrl(client.baseUrl)
+    if (baseUrl === undefined) {
+      log.warn('Failed to create baseUrl for instances')
```
shir-reifenberg, line 68: same

shir-reifenberg (approved):
YAY 💯

PR #4221, opened 2023-04-10
SALTO-3974: move alias code to adapter component

edenhassid:
move alias code to adapter component

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/adapter-components/index.ts
```diff
@@ -21,3 +21,4 @@ export * as elements from './src/elements'
 export * as filters from './src/filters'
 export * as filterUtils from './src/filter_utils'
 export * as references from './src/references'
+export * from './src/add_alias'
```
netama, line 24: nit: maybe `aliases.ts` / `alias.ts`?

netama (approved):

PR #4222, opened 2023-04-10
SALTO-3705: isCondition schema validation should allow values to be references

edenhassid:
isCondition schema validation should allow values to be references

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/change_validators/automation_all_conditions.ts
+++ b/packages/zendesk-adapter/src/change_validators/automation_all_conditions.ts
@@ -17,11 +17,12 @@
 
 import {
   ChangeValidator, getChangeData, InstanceElement,
-  isAdditionOrModificationChange, isInstanceElement,
+  isAdditionOrModificationChange, isInstanceElement, ReferenceExpression,
 } from '@salto-io/adapter-api'
 
 import { resolveValues } from '@salto-io/adapter-utils'
 import { collections } from '@salto-io/lowerdash'
+import _ from 'lodash'
 import { isConditions } from '../filters/utils'
 import { lookupFunc } from '../filters/field_references'
 import { AUTOMATION_TYPE_NAME } from '../constants'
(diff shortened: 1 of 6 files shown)


packages/zendesk-adapter/src/change_validators/automation_all_conditions.ts
```diff
@@ -17,20 +17,21 @@
...
 import { lookupFunc } from '../filters/field_references'
 import { AUTOMATION_TYPE_NAME } from '../constants'
 
 
 const { awu } = collections.asynciterable
 
-const fieldExist = (field: string): boolean =>
-  ['status', 'type', 'group_id', 'assignee_id', 'requester_id'].includes(field)
+const fieldExist = (field: string | ReferenceExpression): boolean =>
+  _.isString(field) && ['status', 'type', 'group_id', 'assignee_id', 'requester_id'].includes(field)
```
netama, line 34: discussed IRL - let's add a comment explaining that these are standard fields so they will never be references + mention that this may change in SALTO-2283 (here and in the other places)

also nit (not directly related to this PR) - let's rename to either `fieldExists`, or something with more meaning (e.g. `hasRequiredAllField`)?

packages/zendesk-adapter/src/replacers_utils.ts
```diff
@@ -44,7 +44,7 @@ export const replaceConditionsAndActionsCreator = (
       .flatMap((condition, i) => {
         const fieldNamesToReplace = replacerParams.fieldsToReplace.map(f => f.name)
         const conditionValue = conditionFieldValue(condition, typeName)
-        if (!fieldNamesToReplace.includes(conditionValue)) {
+        if (isReferenceExpression(conditionValue) || !fieldNamesToReplace.includes(conditionValue)) {
```
netama, line 47: is this for the same reason (all fields currently used in `fieldsToReplace` are standard)?

netama (approved):
LGTM (with a small comment) assuming all these cases only use standard fields 🎈

packages/zendesk-adapter/src/replacers_utils.ts
edenhassid: yes

PR #4225, opened 2023-04-10
SALTO-3866: fix deploy of custom_role modification

edenhassid:
fix deploy of custom_role modification

---

---
_Release Notes_: 
zendesk:
* fix bug in the deployment of custom_role modification (specifically modifications in roles that cannot be fully edited such as `light_agent`)

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/filters/custom_role_deploy.ts
+++ b/packages/zendesk-adapter/src/filters/custom_role_deploy.ts
@@ -1,0 +1,80 @@
+/*
+*                      Copyright 2023 Salto Labs Ltd.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+import {
+  Change, DeployResult, getChangeData,
+  InstanceElement, isInstanceChange, isModificationChange, ModificationChange,
+} from '@salto-io/adapter-api'
+import _ from 'lodash'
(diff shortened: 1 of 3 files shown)


packages/zendesk-adapter/src/filters/custom_role_deploy.ts
```diff
@@ -0,0 +1,80 @@
...
+  applyDetailedChanges(relevantChangesInstance, detailedChanges)
+  return relevantChangesInstance
+}
+
+const editCustomRoleModificationChange = (change: ModificationChange<InstanceElement>)
+  : ModificationChange<InstanceElement> => ({
+  ...change,
+  data: {
+    before: change.data.before,
+    after: returnEditedInstanceFromChange(change),
```
netama, line 42: hmm this structure is a little misleading, because the format of changes assumes we can compare the before and after. let's add some disclaimer explaining that these changes should never be used outside the context of this filter?

packages/zendesk-adapter/src/filters/custom_role_deploy.ts
```diff
@@ -0,0 +1,80 @@
...
+  InstanceElement, isInstanceChange, isModificationChange, ModificationChange,
+} from '@salto-io/adapter-api'
+import _ from 'lodash'
+import { applyDetailedChanges, detailedCompare } from '@salto-io/adapter-utils'
+import { FilterCreator } from '../filter'
+import { CUSTOM_ROLE_TYPE_NAME } from '../constants'
+import { deployChange, deployChanges } from '../deployment'
+
+
+const returnEditedInstanceFromChange = (change: ModificationChange<InstanceElement>): InstanceElement => {
```
netama, line 27: naming nit: the function names here are describing what the "technical" action is, but I think it can help if they explain more about what the logic is - e.g. `createDiffOnlyInstance` or something like that (and similarly below `editCustomRoleModificationChange` can be `createDiffOnlyCustomRoleChange` or even `createDiffOnlyChange`) - that way when skimming through the filter's logic, you can get a better understanding of what it does without diving into the implementation. wdyt?
(btw - there's a good chance both of these will move together to the shared deployment utility functions soon, this seems like useful logic and it's great that it's generic :) )

netama (approved):
🚤

PR #4227, opened 2023-04-10
SALTO-3977 - On fetch, update the changed elements instead of overriding the whole state

idoser96:
* On fetch, instead of overriding the whole state, update the specific changed elements according to the changes between the service and the state
* Delete elements from pathIndex using removal changes instead of overriding it (temporary solution)

---

Currently, changes to element's path are not calculated as a change on plan calculation
Because of that, we are unable to update the changed elements on the pathIndex (because if only the path of an element was changed, we won't know it and won't update_
The solution for now is to keep overriding the pathIndex, and delete elements from it using removal changes (this is in order to delete usage of partial fetched accounts in our code)

---
_Release Notes_: 
Core:
* On fetch, only the changed elements will be updated in the state, instead of overriding all of the elements

---
_User Notifications_: 
None


packages/core/src/api.ts
```diff
@@ -245,9 +273,22 @@ const updateStateWithFetchResults = async (
   const fetchElementsFilter = shouldElementBeIncluded(fetchedAccounts)
   const stateElementsNotCoveredByFetch = await awu(await workspace.state().getAll())
     .filter(element => !fetchElementsFilter(element.elemID)).toArray()
-  await workspace.state()
-    .override(awu(mergedElements)
-      .concat(stateElementsNotCoveredByFetch), fetchedAccounts)
+
+  // If the number of changes matches the number of elements, this is the first fetch, and it is faster to just override
```
ori-moisis, line 277: there can be multiple detailed changes on each element
this also renders the code above in `updateStateFromChanges` very much incorrect (kind of surprised it doesn't crash actually)

packages/workspace/src/workspace/state/in_mem.ts
```diff
@@ -167,5 +221,38 @@ export const buildInMemState = (
...
+        unmergedElements?: Element[]
+        fetchAccounts?: string[]
+}) => {
+      await log.time(async () => {
+        await updateStateElements(changes)
+        if (!_.isEmpty(fetchAccounts)) {
+          await updateAccounts(fetchAccounts)
+        }
+      }, 'state update')
+      log.debug('finished updating state elements with %d changes', changes.length)
```
ori-moisis, line 235: style comment
running anonymous functions in `log.time` isn't great for readability (puts too much emphasis on the fact that something is timed, taking the focus away from what it does).
we still do this sometimes if the timing needs context (e.g - if the same function called from different places needs different log text), but generally if it can be avoided, the code looks better IMO.
when possible, if something needs to be timed, the `log.time` should be in the function definition so it does not interfere with the code flow (like [here](https://github.com/salto-io/salto/blob/main/packages/workspace/src/validator.ts#L757:L757) for example)
In this case I think we can put this on the `updateStateElements` and on the `updatePathIndex` functions

also, the `log.time` already emits a log at the end, there is no need to have a debug log immediately afterwards...

packages/workspace/src/workspace/path_index.ts
```diff
@@ -169,6 +169,19 @@ RemoteMapEntry<Path[]>[] => {
...
+  unmergedElementIDs: Set<string>,
+): Promise<void> => {
+  // Entries that exists in the index but not in the unmerged elements were deleted and should be removed from the index
+  const entriesToDelete = await awu(current.keys()).filter(key => !unmergedElementIDs.has(key)).toArray()
+  const entriesToSet = getElementsPathHints(changedUnmergedElements)
+
+  await current.deleteAll(entriesToDelete)
+  await current.setAll(entriesToSet)
+}
+
```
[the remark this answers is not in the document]
idoser96: Temp because i did not switch all of the original functions with this once yet (and the name without 'temp' is taken)

once i'll finish switching ill rename this and delete the other

packages/core/src/core/fetch.ts
```diff
@@ -121,17 +121,24 @@ type WorkspaceDetailedChange = {
...
-  new collections.treeMap.TreeMap(
-    wu(await getDetailedChanges(before, after, topLevelFilters))
-      .map(change => [change.id.getFullName(), [{ change, origin }]])
+): Promise<detailedChangesResult> => {
+  const changes = wu(await getDetailedChanges(before, after, topLevelFilters)).toArray()
+  const changesTree = new collections.treeMap.TreeMap(
+    changes.map(change => [change.id.getFullName(), [{ change, origin }]])
   )
-)
+  return { changesTree, changes }
```
ori-moisis, line 140: this is a mis-match with the return type, `changes` are claimed to be `DetailedChange<Element>` but you are returning `DetailedChange<any>`
and in practice, the detailed changes contain specific values, not entire elements.

packages/workspace/src/workspace/path_index.ts
```diff
@@ -189,45 +183,47 @@ export const getTopLevelPathHints = (unmergedElements: Element[]): PathHint[] =>
...
-  { index,
-    elements,
-    accountsToMaintain,
-    isTopLevel }: UpdateIndexParams
+/** Currently because a change in an element's path is not a change, we are unable to detect it
+ *  Until this is solved, we have to override the path index with all the elements, and delete the elements that
+ *  were removed
+* */
+const updateIndex = async (
+  { pathIndex, unmergedElements, topLevelRemovalFullNames, getHintsFunction }:
```
ori-moisis, line 197: if we are only looking at top level removals, would we be missing the case where, for example, a single field is deleted?
specifically in salesforce that scenario I think would actually affect the path index since fields are split across two files

I am afraid that for completeness we may have to look at any level of removal and check it with `isParentOf` on the element ID

packages/workspace/src/workspace/state/state.ts
```diff
@@ -23,6 +23,12 @@ import { StateStaticFilesSource } from '../static_files/common'
 
 export type StateMetadataKey = 'version' | 'hash'
 
+export type updateStateElementsArgs = {
```
ori-moisis, line 26: type name conventions
```suggestion
export type UpdateStateElementsArgs = {
```

packages/workspace/src/workspace/path_index.ts
```diff
@@ -189,45 +183,49 @@ export const getTopLevelPathHints = (unmergedElements: Element[]): PathHint[] =>
...
+  const entriesToSet = getHintsFunction(unmergedElements)
+
+  // If no removals were passed, override the index with the new elements
+  if (removedElementsFullNames === undefined) {
+    await pathIndex.clear()
+  } else {
+    // Entries that are related to an element that was removed should be deleted
+    const entriesToDelete = await awu(pathIndex.keys()).filter(key => {
+      // Needed for createTopLevelParentID() function
+      const tempElemID = ElemID.fromFullName(key)
```
ori-moisis, line 210: for performance - do this only after the check if
```
removedElementsFullNames.has(key)
```

packages/workspace/src/workspace/state/in_mem.ts
```diff
@@ -71,6 +74,55 @@ export const buildInMemState = (
...
+    const data = await stateData()
+    const newAccounts = accounts ?? await awu(getUpdateDate(data).keys()).toArray()
+    return getUpdateDate(data).setAll(
+      awu(newAccounts.map(s => ({ key: s, value: new Date(Date.now()) })))
+    )
+  }
+
+  const updateStatePathIndex = async (
+    unmergedElements: Element[],
+    removedElementsFullNames?: Set<string>,
```
ori-moisis, line 87: does not need to be optional (AFAICT, only one place calls this, and it passes this argument)
```suggestion
    removedElementsFullNames: Set<string>,
```

packages/workspace/src/workspace/state/in_mem.ts
```diff
@@ -71,6 +74,59 @@ export const buildInMemState = (
...
+      removedElementsFullNames,
+    })
+    await updatePathIndex({
+      pathIndex: currentStateData.pathIndex,
+      unmergedElements,
+      removedElementsFullNames,
+    })
+  }
+
+  const updateStateElements = async (changes: DetailedChange<Element>[]): Promise<void> => log.time(async () => {
```
ori-moisis, line 102: by moving away from `override` we are not running the line
```typescript
await data.staticFilesSource.clear()
```
anymore

this is a problem as it means we break the behavior of the state static files source (it won't remove files that are no longer in the state)

the old behavior also covered for a bug in `set` where we would not remove static files that were deleted from the element, we will be exposing that bug, so we probably need to fix it (that can be in a separate PR if you want)

packages/core/src/api.ts
```diff
@@ -175,15 +194,17 @@ export const deploy = async (
...
+    const updatedElements = awu(appliedChanges)
+      .filter(change => (isAdditionOrModificationChange(change) || isFieldChange(change)))
+      .map(getUpdatedElement)
+
+    const detailedChanges = appliedChanges.map(change => getDetailedChangesFromChanges(change)).flat()
+    await workspace.state().updateStateFromChanges({ changes: detailedChanges })
+
+    await updatedElements.forEach(async updatedElement => {
+      await changedElements.set(updatedElement)
+    })
```
ori-moisis, line 207: why is this separate from the calculation of the updated elements?
also, and quite important, this is still running the calculation of `getUpdatedElement` AFTER the state has be updated

packages/core/src/api.ts
```diff
@@ -175,15 +194,17 @@ export const deploy = async (
...
-    }))
+    // create updated elements before updating the state, to prevent any unwanted behavior
+    const updatedElements = awu(appliedChanges)
+      .filter(change => (isAdditionOrModificationChange(change) || isFieldChange(change)))
+      .map(getUpdatedElement)
+
+    const detailedChanges = appliedChanges.map(change => getDetailedChangesFromChanges(change)).flat()
+    await workspace.state().updateStateFromChanges({ changes: detailedChanges })
+
+    await updatedElements.forEach(async updatedElement => {
```
ori-moisis, line 205:
```suggestion
    // update changedElements before updating the state because we rely on the state data for the calculation
    await awu(appliedChanges)
      .filter(change => (isAdditionOrModificationChange(change) || isFieldChange(change)))
      .map(getUpdatedElement)
      .forEach(async updatedElement => {
        await changedElements.set(updatedElement)
      })

    const detailedChanges = appliedChanges.flatMap(getDetailedChangesFromChanges)
    await workspace.state().updateStateFromChanges({ changes: detailedChanges })
```

as an alternative, we could keep just the IDs of the changed elements and use the state as the element source for comparison in [line 218](https://github.com/salto-io/salto/pull/4227/files#diff-7a68aa0c5146864913c9bb8b990a25deb8715ead9aea7f9aa9e0f44d23d57448R218)

packages/core/src/core/fetch.ts
```diff
@@ -549,71 +557,127 @@ export const getAdaptersFirstFetchPartial = async (
...
+}
+
 // Calculate the fetch changes - calculation should be done only if workspace has data,
 // o/w all account elements should be consider as "add" changes.
 export const calcFetchChanges = async (
   accountElements: ReadonlyArray<Element>,
-  mergedAccountElements: ReadOnlyElementsSource,
-  stateElements: ReadOnlyElementsSource,
+  mergedAccountElements: ReadonlyArray<Element>,
+  stateElements: elementSource.ElementsSource,
```
ori-moisis, line 577: why did this become a writeable element source?

packages/core/src/core/fetch.ts
```diff
@@ -549,71 +557,127 @@ export const getAdaptersFirstFetchPartial = async (
...
   workspaceElements: ReadOnlyElementsSource,
   partiallyFetchedAccounts: Set<string>,
   allFetchedAccounts: Set<string>
-): Promise<FetchChange[]> => {
+): Promise<CalcFetchChangesResult> => {
+  const mergedAccountElementsSource = elementSource.createInMemoryElementSource(mergedAccountElements)
+  // When we init a new env, state will be empty. We fallback to the workspace
+  // elements since they should be considered a part of the env and the diff
+  // should be calculated with them in mind.
+  const isStateEmpty = await stateElements.isEmpty()
```
ori-moisis, line 586: this check should not be so far away from the place where it is used

packages/workspace/src/workspace/state/in_mem.ts
```diff
@@ -71,6 +75,78 @@ export const buildInMemState = (
     await (await stateData()).elements.delete(id)
   }
 
+  const updateAccounts = async (accounts?: string[]): Promise<void> => {
+    const data = await stateData()
+    const newAccounts = accounts ?? await awu(getUpdateDate(data).keys()).toArray()
+    return getUpdateDate(data).setAll(
+      awu(newAccounts.map(s => ({ key: s, value: new Date(Date.now()) })))
```
ori-moisis, line 82: `awu` not needed here
```suggestion
      newAccounts.map(s => ({ key: s, value: new Date(Date.now()) }))
```

packages/workspace/src/workspace/state/in_mem.ts
```diff
@@ -71,6 +75,78 @@ export const buildInMemState = (
...
+  const setElement = async (element: Element): Promise<void> => {
+    const state = await stateData()
+    const beforeElement = await state.elements.get(element.elemID)
+
+    const filesToDelete = beforeElement !== undefined
+      ? getDanglingStaticFiles(detailedCompare(beforeElement, element))
+      : []
+
+    await state.elements.set(element)
+    await Promise.all(filesToDelete.map(async f => state.staticFilesSource.delete(f)))
```
ori-moisis, line 147: unused `async`
```suggestion
    await Promise.all(filesToDelete.map(f => state.staticFilesSource.delete(f)))
```

packages/workspace/src/workspace/state/in_mem.ts
```diff
@@ -83,58 +159,15 @@ export const buildInMemState = (
       )
       return (await stateData()).elements.deleteAll(ids)
     },
-    set: async (element: Element): Promise<void> =>
-      (await stateData()).elements.set(element),
+    set: setElement,
     setAll: async (elements: ThenableIterable<Element>): Promise<void> =>
       (await stateData()).elements.setAll(elements),
```
ori-moisis, line 164: should this be calling our `set` so it supports the whole static file deletion thing?
otherwise, this is a possible gap...

probably not a critical gap since nobody should ever use this function, but since actually removing this function is a bigger task right now I think we are not going to remove it completely...
the fact to nothing should call this function makes me less worried about how inefficient it will be if we make it just call `setElement` one by one

packages/workspace/test/workspace/state.test.ts
```diff
@@ -202,6 +198,102 @@ describe('state', () => {
     it('should return the salto version that was provided in load data', async () => {
       expect(await state.getStateSaltoVersion()).toEqual('0.0.1')
     })
+
+    describe('updateStateFromChanges', () => {
+      it('should update changed elements correctly without changing non-changed elements', async () => {
```
ori-moisis, line 203: this style of test is not very useful for future readers (putting at least 9 different scenarios in a single test).
it makes it very hard to understand how we expect the code to behave in each scenario.

the big red flag here is the word "correctly" in the test name - the test is kind of the only place that can define what the "correct" behavior is, so if the test says "it should behave correctly", then what "correct" means remains undefined.

this in turn means that when a future developer breaks this test, it would be much harder for them to figure out what they broke and whether it is important or not

packages/core/src/api.ts
idoser96: I wanted to keep the running order the same, but i don't think it really matters with state is being set first, i'll merge the calculation and the updating

packages/core/src/core/fetch.ts
idoser96: it was like that until a week ago

it should be readonly, but readonly doesn't implement isEmpty (i think it should, but when i went to change it i noticed it required a lot of changes in other places)

packages/core/src/api.ts
idoser96: That means that 'adaptersElementSource' won't have changedElements as a a fallback source
Im not sure in what case changedElement will include elements that that aren't in workspace.elements(), but I don't have enough knowledge to be sure

packages/workspace/test/workspace/state.test.ts
idoser96: I combined them to make sure the code handles all cases parallel, i did move it into a new describe and had every case in a single 'it', let me know what you think

packages/workspace/src/workspace/state/in_mem.ts
idoser96: Changed, but are you sure nothing calls it? i see a few usages and they seem to be part of our regular flow

ori-moisis: where did you find a use of this?

packages/core/src/api.ts
ori-moisis: ok, so not the alternative then

packages/core/src/core/fetch.ts
ori-moisis: ok

packages/core/src/api.ts
```diff
@@ -193,19 +193,17 @@ export const deploy = async (
...
+    const updatedElements = await awu(appliedChanges)
       .filter(change => (isAdditionOrModificationChange(change) || isFieldChange(change)))
-      .map(getUpdatedElement)
+      .map(getUpdatedElement).toArray()
+    await changedElements.setAll(updatedElements)
 
-    const detailedChanges = appliedChanges.map(change => getDetailedChangesFromChanges(change)).flat()
+    const detailedChanges = appliedChanges.flatMap(change => getDetailedChangesFromChanges(change))
     await workspace.state().updateStateFromChanges({ changes: detailedChanges })
+  }, 'postDeployAction')
```
idoser96, line 205: added log.time for future debugging

packages/core/src/api.ts
```diff
@@ -193,19 +193,17 @@ export const deploy = async (
...
 
-  const postDeployAction = async (appliedChanges: ReadonlyArray<Change>): Promise<void> => {
-    // create updated elements before updating the state, to prevent any unwanted behavior
-    const updatedElements = awu(appliedChanges)
+  const postDeployAction = async (appliedChanges: ReadonlyArray<Change>): Promise<void> => log.time(async () => {
+    // update changedElements before updating the state because we rely on the state data for the calculation
+    const updatedElements = await awu(appliedChanges)
       .filter(change => (isAdditionOrModificationChange(change) || isFieldChange(change)))
-      .map(getUpdatedElement)
+      .map(getUpdatedElement).toArray()
```
ori-moisis, line 200: be consistent with the newlines
```suggestion
      .map(getUpdatedElement)
      .toArray()
```

packages/workspace/src/workspace/path_index.ts
```diff
@@ -189,45 +183,67 @@ export const getTopLevelPathHints = (unmergedElements: Element[]): PathHint[] =>
...
-    .filter(e => accountsToMaintain.includes(ElemID.fromFullName(e.key).adapter))
-    .toArray()
-  const updatedEntries = _.unionBy(entries, oldPathHintsToMaintain, e => e.key)
-  await index.clear()
-  await index.setAll(awu(updatedEntries))
+    // Needed for createTopLevelParentID() function
+    const tempElemID = ElemID.fromFullName(key)
+    const parentFullName = tempElemID.createTopLevelParentID().parent.getFullName()
+    // If any of the levels above the element was removed, delete the element
+    return getPathsFromParent(parentFullName, key).some(path => removedElementsFullNames.has(path))
```
ori-moisis, line 232: You don't have to re-invent the wheel


```suggestion
    const keyElemID = ElemID.fromFullName(key)
    // If any of the levels above the key was removed, delete the key
    return keyElemID.createAllElemIdParents()
      .map(id => id.getFullName())
      .some(fullName => removedElementsFullNames.has(fullName))
```

packages/workspace/test/workspace/state.test.ts
```diff
@@ -200,7 +200,7 @@ describe('state', () => {
     })
 
     describe('updateStateFromChanges', () => {
-      it('should update changed elements correctly without changing non-changed elements', async () => {
+      describe('should update changed elements correctly without changing non-changed elements', async () => {
```
ori-moisis, line 203: describe should describe the scenario, not the intended behavior
the intended behavior is in the `it('should ...`

also - describe must not be async
also - we should not run code in describe, only in `beforeEach` / `beforeAll`

packages/workspace/test/workspace/state.test.ts
```diff
@@ -13,17 +13,40 @@
...
+import { PathIndex, getElementsPathHints } from '../../src/workspace/path_index'
 import { createInMemoryElementSource } from '../../src/workspace/elements_source'
 import { InMemoryRemoteMap, RemoteMapCreator } from '../../src/workspace/remote_map'
 import { StaticFilesSource } from '../../src/workspace/static_files/common'
 import { mockStaticFilesSource } from '../utils'
 
+const updatePathMock = jest.fn()
+const updateTopLevelPathMock = jest.fn()
+
+jest.mock('../../src/workspace/path_index', () => {
```
ori-moisis, line 41: we should not over-mock things, it makes tests hit impossible scenarios (e.g - anything in the code that may need the result of a mocked function would get an unrealistic response when running in a test vs. running in real code)

if there is a need to check the path index is updated correctly, it should be checked through the state interface (e.g - read the path index after updating to make sure it is correct)

we have created too many bugs with this type of excessive mocking...

packages/workspace/test/workspace/state.test.ts
```diff
@@ -202,6 +199,118 @@ describe('state', () => {
...
+              { action: 'remove', data: { before: fieldToRemove }, id: fieldToRemoveElemID }, // Field to be removed
+              { action: 'modify', data: { before: fieldToModify, after: fieldToModify }, id: fieldToModifyElemID }, // Field to be modified
+            ],
+          })
+
+          allElements = await awu(await state.getAll()).toArray()
+        })
+
+
+        it('should not add or remove unexpected elements', async () => {
```
ori-moisis, line 239: what does this test mean? what are "unexpected elements'?
also, why is it async? same for all other tests I think

packages/workspace/src/workspace/path_index.ts
```diff
@@ -189,45 +183,48 @@ export const getTopLevelPathHints = (unmergedElements: Element[]): PathHint[] =>
...
-    return
-  }
-  const entries = isTopLevel ? getTopLevelPathHints(elements) : getElementsPathHints(elements)
-  const oldPathHintsToMaintain = await awu(index.entries())
-    .filter(e => accountsToMaintain.includes(ElemID.fromFullName(e.key).adapter))
-    .toArray()
-  const updatedEntries = _.unionBy(entries, oldPathHintsToMaintain, e => e.key)
-  await index.clear()
-  await index.setAll(awu(updatedEntries))
+    // Needed for createTopLevelParentID() function
```
ori-moisis, line 208: the comment doesn't make sense

packages/workspace/test/workspace/state.test.ts
idoser96: no reason for async, leftover from copy-paste
i'll put more attention to my async and awu uses from now on

packages/core/src/api.ts
```diff
@@ -14,18 +14,37 @@
...
 import { EventEmitter } from 'pietile-eventemitter'
 import { logger } from '@salto-io/logging'
 import _ from 'lodash'
 import { promises, collections, values } from '@salto-io/lowerdash'
 import { Workspace, ElementSelector, elementSource, expressions, merger } from '@salto-io/workspace'
 import { EOL } from 'os'
-import { buildElementsSourceFromElements } from '@salto-io/adapter-utils'
+import {
+  buildElementsSourceFromElements,
+  getDetailedChanges as getDetailedChangesFromChanges,
```
ori-moisis, line 46: why the rename?
if it is needed, at least lets make this reflect the correct input, it is from change, not from changes

packages/workspace/src/workspace/state/in_mem.ts
```diff
@@ -167,5 +198,22 @@ export const buildInMemState = (
...
     getStateSaltoVersion: async () => (await stateData()).saltoMetadata.get('version'),
     setVersion: async (version: string) => (await stateData()).saltoMetadata.set('version', version),
+    updateStateFromChanges: async (
+      { changes, unmergedElements, fetchAccounts }: UpdateStateElementsArgs) => {
+      await updateStateElements(changes)
+      if (!_.isEmpty(fetchAccounts)) {
+        await updateAccounts(fetchAccounts)
+      }
+
+      if (unmergedElements === undefined || _.isEmpty(unmergedElements)) {
```
ori-moisis, line 208: shouldn't we let it pass to the path index update at least to remove the things that need to be removed?
otherwise when we deploy a deletion for example, the path index for that element will never be cleaned I think

we should add a test for it

packages/core/src/api.ts
idoser96: renames because we already have a getDetailedChanges imported, will rename

packages/workspace/src/workspace/state/in_mem.ts
idoser96: This is the equivalent of just calling 'updatePathIndex' that happened before
Now that we merged updateState and updatePathIndex, some calls to this functions doesn't really need to touch the pathIndex (and it doesn't do it today as well)

Changing it is changing the behaviour, so i guess we should talk about it in person

ori-moisis (approved):
I think this is ready 🥳 
we should consider the timing of when to merge this though as this is a risky change.
I suggest we don't merge this before we talk about it

PR #4230, opened 2023-04-10
SALTO-3978 Avoid stopping pagination on empty filtered page

netama:
Fix bug in the pagination logic, where we would stop pagination not only if a page was empty, but also if the filtered results of the page (when using a custom entry extractor) were empty. this can cause pagination to stop mid-way (e.g. in jira - if a full page of project-scope entries is received).

---

This only impacted Jira, since this is currently the only adapter using a `customEntryExtractor`.

---
_Release Notes_: 
_Jira Adapter_:
* Fix bug where in some cases not all instances would be returned during fetch, when a full page of project-scoped items was returned.

---
_User Notifications_: 
_Jira Adapter_:
* Fix bug where in some cases not all instances would be returned during fetch

### DIFF
--- a/packages/adapter-components/src/client/pagination.ts
+++ b/packages/adapter-components/src/client/pagination.ts
@@ -127,12 +127,15 @@ export const traverseRequests: (
         ? response.data.items
         : makeArray(response.data)
     ).flatMap(extractPageEntries)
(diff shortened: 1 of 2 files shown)


IdoZakk (approved):
🥇

PR #4232, opened 2023-04-13
SALTO-3390: Salesforce fetch with changes detection (Fetch of updated instances)

tamtamirr:
Added support for fetch with changes detection in Salesforce Adapter.

---

ChangedAtSingleton:  https://github.com/salto-io/tamir-sf/blob/main/ChangedAtSingleton.txt
Optimized fetch verbose logs: https://github.com/salto-io/tamir-sf/blob/main/withCD.txt

Note:
- `StandardValueSet` has empty string as `_changed_at` annotation, since we have special handling to retrieve them (They must be retrieved explicitly by their names in a filter that has `empty file properties`) [See this](https://github.com/salto-io/tamir-sf/blob/3eab48d54441e31c367ce7e14e55feae9452afd2/ChangedAtSingleton.txt#L1885) for reference. It shouldn't bother us though as there are no edge cases with it. We can also explicitly validate the values in the instance, but I'm not sure it's urgent.
- This will be a partial fetch (similar to `targeted fetch`) so the same behavior applies here. (For example, deletions are not supported)
---
_Release Notes_: 
Salesforce Adapter:
- Support fetch with changes detection 🎉  Retrieve Elements that were modified or added since the last fetch.

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
```diff
@@ -28,6 +29,8 @@ const VALID_FOLDER_PATH_RE = /^[a-zA-Z\d_/]+$/
 export type MetadataQuery = {
   isTypeMatch: (type: string) => boolean
   isInstanceMatch: (instance: MetadataInstance) => boolean
+  isTargetedFetch: () => boolean
+  isFetchWithChangeDetection: () => boolean
```
tamtamirr, line 33: Even though both are considered `Partial fetch` they should be separated in some use-cases. (For example, the filter [createMissingInstalledPackageInstances](https://github.com/salto-io/salto/blob/976e8be0df168e21035eccb7a0efe9b0e2a97d6a/packages/salesforce-adapter/src/filters/create_missing_installed_packages_instances.ts#L37-L36) can't run in `fetch withChangesDetection` mode)

packages/salesforce-adapter/src/filters/create_missing_installed_packages_instances.ts
```diff
@@ -37,6 +37,9 @@ const createMissingInstalledPackageInstance = (file: FileProperties, installedPa
 const filterCreator: RemoteFilterCreator = ({ client, config }) => ({
   name: 'createMissingInstalledPackagesInstancesFilter',
   onFetch: async (elements: Element[]): Promise<FilterResult | undefined> => {
+    if (config.fetchProfile.metadataQuery.isFetchWithChangeDetection()) {
```
tamtamirr, line 40: This filter should not run in `fetch withChangeDetection` since it will find all the `InstalledPackages` as missing at this point. The alternative would be to read them from the `elementsSource` but it seems redundant.

packages/salesforce-adapter/test/mock_elements.ts
```diff
@@ -604,10 +608,16 @@ export const mockDefaultValues = {
...
-)
+export const mockInstances = () => ({
+  ..._.mapValues(
+    mockDefaultValues,
+    (values, typeName) => createInstanceElement(
+      values,
+      mockTypes[typeName as keyof typeof mockDefaultValues],
+    )
+  ),
+  [CHANGED_AT_SINGLETON]: new InstanceElement(
```
tamtamirr, line 619: Could not add it to **mockDefaultValues** as this is not a **Metadata Instance**

tamtamirr (commented):
Added some comments :) Can take this F2F as-well, this PR has some logic

packages/salesforce-adapter/src/constants.ts
```diff
@@ -378,6 +379,18 @@ export const DATA_CATEGORY_GROUP_METADATA_TYPE = 'DataCategoryGroup'
 
 // Artifitial Types
 export const CURRENCY_CODE_TYPE_NAME = 'CurrencyIsoCodes'
+export const CHANGED_AT_SINGLETON = 'ChangedAtSingleton'
+
+export const ArtificialTypes = {
```
ori-moisis, line 384: why is this in `constants`?
I would not expect type definitions to be here, we have other places that have hard coded types...

also, this is not a general constant, it has a very specific purpose, I would expect it to be defined somewhere that has the code that is related to the feature of storing and using this information.

in general this PR seems to have put code all over and touched many files, this is not a good sign.
features ideally should be kept to a few locations so it would be easy to follow a feature and, more importantly, so it would be easy to ignore a feature if I don't care about it and want to read other things...

packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
```diff
@@ -115,6 +119,15 @@ export const buildMetadataQuery = (
     }
     return false
   }
+  const wasUpdated = (instance: MetadataInstance): boolean => {
+    if (changedAtSingleton === undefined || instance.changedAt === undefined) {
+      return true
+    }
+    const lastChangedAt = _.get(changedAtSingleton.value, [instance.metadataType, instance.name])
+    return _.isString(lastChangedAt)
+      ? new Date(lastChangedAt).getTime() < new Date(instance.changedAt).getTime()
```
ori-moisis, line 128: not sure we need to parse this, I think any difference here should make us fetch again, e.g
```
return !_.isEqual(lastChangedAt, instance.changedAt)
```
should be sufficient, no?

packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
```diff
@@ -115,6 +119,15 @@ export const buildMetadataQuery = (
     }
     return false
   }
+  const wasUpdated = (instance: MetadataInstance): boolean => {
+    if (changedAtSingleton === undefined || instance.changedAt === undefined) {
+      return true
+    }
+    const lastChangedAt = _.get(changedAtSingleton.value, [instance.metadataType, instance.name])
```
ori-moisis, line 126: I don't think we should work with an untyped value like this in our code.
we should get rid of the instance at the highest abstraction level possible and the rest of the code should work with typed values

packages/salesforce-adapter/src/filters/changed_at_singleton.ts
```diff
@@ -0,0 +1,73 @@
...
+const filterCreator: LocalFilterCreator = ({ config }): FilterWith<'onFetch'> => ({
+  name: 'changedAtSingletonFilter',
+  onFetch: async (elements: Element[]) => {
+    const metadataInstancesByType = await groupByAsync(
+      await awu(elements)
+        .filter(isInstanceElement)
+        .filter(isMetadataInstanceElement)
+        .filter(instance => _.isString(instance.annotations[CORE_ANNOTATIONS.CHANGED_AT]))
+        .toArray(),
+      async instance => await safeApiName(await instance.getType()) ?? '_unsorted'
```
ori-moisis, line 55: should be `metadataType(instance)`?

packages/salesforce-adapter/src/filters/changed_at_singleton.ts
```diff
@@ -0,0 +1,73 @@
...
+  const instanceValues: Values = {}
+  Object.entries(metadataInstancesByType).forEach(([metadataType, instances]) => {
+    instanceValues[metadataType] = {}
+    instances.forEach(instance => {
+      const instanceName = instance.value[INSTANCE_FULL_NAME_FIELD]
+      instanceValues[metadataType][instanceName] = instance.annotations[CORE_ANNOTATIONS.CHANGED_AT]
+    })
+  })
+  return instanceValues
+}
```
ori-moisis, line 44: why use stateful code for a simple transformation?

```suggestion
const createChangedAtSingletonInstanceValues = (metadataInstancesByType: Record<string, InstanceElement[]>): Values => _.mapValues(
  metadataInstancesByType,
  instances => Object.fromEntries(
    instances.map(instance => [
      instance[INSTANCE_FULL_NAME_FIELD],
      instance.annotations[CORE_ANNOTATIONS.CHANGED_AT],
    ])
  )
)
```

also, is there a reason we don't have a type for the value?
I'd expect that we would have a type of it given that we need to use it in multiple places and I assume we rely on its structure in the code that needs to decide what to fetch

packages/salesforce-adapter/test/mock_elements.ts
ori-moisis: so maybe it shouldn't be here?
seems like it shouldn't be here...

packages/salesforce-adapter/src/filters/create_missing_installed_packages_instances.ts
```diff
@@ -37,6 +37,9 @@ const createMissingInstalledPackageInstance = (file: FileProperties, installedPa
 const filterCreator: RemoteFilterCreator = ({ client, config }) => ({
   name: 'createMissingInstalledPackagesInstancesFilter',
   onFetch: async (elements: Element[]): Promise<FilterResult | undefined> => {
+    if (config.fetchProfile.metadataQuery.isFetchWithChangesDetection()) {
+      return
+    }
```
ori-moisis, line 42: why should this filter be aware of this feature?
we already make every fetch go through the fetch profile thingy to ensure we control what is fetched, we should not have to also deal with specific inner features inside the fetch profile in every filter...

packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
```diff
@@ -131,9 +144,14 @@ export const buildMetadataQuery = (
     isInstanceMatch: instance => (
       include.some(params => isInstanceMatchQueryParams(instance, params))
       && !fullExcludeList.some(params => isInstanceMatchQueryParams(instance, params))
+      && wasUpdated(instance)
     ),
 
-    isPartialFetch: () => target !== undefined,
+    isTargetedFetch: () => target !== undefined,
+
+    isFetchWithChangesDetection: () => changedAtSingleton !== undefined,
```
ori-moisis, line 152: why are these needed?
is there a reason for this information to be known outside this implementation?

ori-moisis (commented):
didn't fully review everything, but some initial comments

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -276,6 +282,33 @@ export const UNSUPPORTED_SYSTEM_FIELDS = [
...
+ * {@link convertListsFilter}, {@link convertTypeFilter}, {@link convertMapsFilter}
+ * This logic can be removed once [SALTO-3985]{@link https://salto-io.atlassian.net/browse/SALTO-3985} is implemented.
+ */
+const getMetadataTypesWithInstances = async (
+  metadataTypes: TypeElement[],
+  metadataInstances: InstanceElement[],
+  topLevelTypeToSubtypes: Record<string, string[]>): Promise<TypeElement[]> => {
+  const metadataInstancesByType = await groupByAsync(
+    metadataInstances,
+    async instance => await safeApiName(await instance.getType()) ?? '_unsorted',
```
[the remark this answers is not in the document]
tamtamirr: Thanks 🔥

packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
tamtamirr: true!

packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
tamtamirr: Not sure I understand the suggestion. How can we type the **ChangedAtSingleton**? It's value is dynamic.

packages/salesforce-adapter/src/filters/changed_at_singleton.ts
tamtamirr: v

packages/salesforce-adapter/src/filters/changed_at_singleton.ts
tamtamirr: What would you expect this type to be? `Record<string, Record<string,string>>`? I personally don't think it adds anything, as we simply use **_.get(instanceValue, typeName, instanceName)**

packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
tamtamirr: Why would we expose **isTargetedFetch** but not **isFetchWithChangesDetection**? This is two different fetch modes of the adapter, no?

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -336,6 +369,7 @@ export default class SalesforceAdapter implements AdapterOperations {
     systemFields = SYSTEM_FIELDS,
     unsupportedSystemFields = UNSUPPORTED_SYSTEM_FIELDS,
     config,
+    changedAtSingleton,
```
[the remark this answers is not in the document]
tamtamirr: Trying to figure out what's our alternative? Wouldn't it make the most sense to enforce this logic within the **metadataQuery** object?

packages/salesforce-adapter/src/filters/changed_at_singleton.ts
ori-moisis: you can say that about any type...
we want to have types because they make it easier to know what we are looking at, we want to know that when we `_.get(value, typeName, instanceName)` we will get back a string and not a list or an object or nothing. that is the whole point of writing typescript and not plain javascript (plus, not sure why we would need to use `_.get` to access a value in an object).

`Values` is basically `any`, and we want to minimize our use of `any` because it removes the value we get from using typescript and not writing in plain javascript...

packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
ori-moisis: typeguard

and don't use `InstanceElement` as an input parameter to the query, pass in the structure you really want to have

packages/salesforce-adapter/src/adapter.ts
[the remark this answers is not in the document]
ori-moisis: lets talk about it f2f when you get back to working on this PR.
I think it would be easier to write this part of the code together

tamtamirr:
Closing due to blocker.

PR #4236, opened 2023-04-13
remove fixDeployXml call in sdf_client to revert SALTO-2865

Shulik95:
_Removed fixDeployXml call in sdf_client to revert SALTO-2865_

---

_None_

---
_Release Notes_: 
_Netsuite Adapter:_
* deploy.xml will no longer be ordered according to reference level
---
_User Notifications_: 
_None_


omrilit (approved):
Thanks for this one!

PR #4237, opened 2023-04-13
SALTO-3779 Always add remaining types on fetch

netama:
Fix bug in Zendesk where we would sometimes run into merge conflicts on deploy, due to the `id` values not being hidden properly.

The scenario is very specific:
1. The fetch config has Guide enabled, but no matching brand has Guide enabled in the service
2. An instance is added during deploy for a type that has 0 existing instances (and has an `id` field)

---

The reason this happens is that we use `addRemainingTypes` to ensure all hidden fields are defined properly - and this function did not run in this scenario due to an early return. 

---
_Release Notes_: 
_Zendesk Adapter_:
* Fix bug where deploying the first instance of its kind might result in merge errors, for environments with Guide enabled but without Guide elements

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/test/adapter.test.ts
+++ b/packages/zendesk-adapter/test/adapter.test.ts
@@ -27,7 +27,7 @@ import {
   BuiltinTypes,
   CORE_ANNOTATIONS,
   isRemovalChange,
-  getChangeData, TemplateExpression,
+  getChangeData, TemplateExpression, isObjectType,
 } from '@salto-io/adapter-api'
 import { buildElementsSourceFromElements } from '@salto-io/adapter-utils'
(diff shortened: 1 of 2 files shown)


packages/zendesk-adapter/test/adapter.test.ts
```diff
@@ -679,7 +679,14 @@ describe('adapter', () => {
           'zendesk.automation.instance.Tag_tickets_from_Social@s',
           'zendesk.automation_order.instance',
           'zendesk.custom_role.instance.Advisor',
+          'zendesk.organization_field_order.instance',
```
netama, line 682: this was "fixed" by the change. we might want to remove these, but I don't think it should be tied to this change

shir-reifenberg (approved):
🚀

PR #4238, opened 2023-04-16
SALTO-3630: Partial Fetch Bugfix

tamtamirr:
As part of testing out Partial Fetch with my recent `Fetch With Change Detection` locally, I've noticed that a partial fetch on a certain type, will remove the `generatedDependencies` to `InstalledPackage` instances if they were not part of the `fetch targets`.

You can see the Workspace diff when attempting to partially fetch `ApexClass`: https://github.com/salto-io/tamir-sf/commit/f864ea77c6cb9ec3624a93f47d4cdecc2d7974b8

---

_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
+++ b/packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
@@ -15,16 +15,7 @@
 */
 import { regex, values } from '@salto-io/lowerdash'
 import _ from 'lodash'
-import {
-  DEFAULT_NAMESPACE,
-  SETTINGS_METADATA_TYPE,
-  TOPICS_FOR_OBJECTS_METADATA_TYPE,
-  CUSTOM_OBJECT,
-  MAX_TYPES_TO_SEPARATE_TO_FILE_PER_FIELD,
-  FLOW_DEFINITION_METADATA_TYPE,
-  FLOW_METADATA_TYPE,
-  INSTALLED_PACKAGE_METADATA,
-} from '../constants'
+import { DEFAULT_NAMESPACE, SETTINGS_METADATA_TYPE, TOPICS_FOR_OBJECTS_METADATA_TYPE, CUSTOM_OBJECT, MAX_TYPES_TO_SEPARATE_TO_FILE_PER_FIELD, FLOW_DEFINITION_METADATA_TYPE, FLOW_METADATA_TYPE } from '../constants'
 import { validateRegularExpressions, ConfigValidationError } from '../config_validation'
 import { MetadataInstance, MetadataParams, MetadataQueryParams, METADATA_INCLUDE_LIST, METADATA_EXCLUDE_LIST, METADATA_SEPARATE_FIELD_LIST } from '../types'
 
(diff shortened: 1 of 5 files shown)


packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
```diff
@@ -97,12 +110,11 @@ export const buildMetadataQuery = (
       ? isFolderMetadataTypeNameMatch(instance, name)
       : regex.isFullRegexMatch(instance.name, name)
   }
-
   const isIncludedInPartialFetch = (type: string): boolean => {
     if (target === undefined) {
       return true
     }
-    if (target.includes(type)) {
+    if (target.includes(type) || ALWAYS_INCLUDED_IN_PARTIAL_FETCH.includes(type)) {
```
tamtamirr, line 117: Since **MANY** MetadataTypes can possibly come from InstalledPackage, it's better to always retrieve them in partial fetch.

packages/salesforce-adapter/src/filters/create_missing_installed_packages_instances.ts
```diff
@@ -37,6 +37,9 @@ const createMissingInstalledPackageInstance = (file: FileProperties, installedPa
 const filterCreator: RemoteFilterCreator = ({ client, config }) => ({
   name: 'createMissingInstalledPackagesInstancesFilter',
   onFetch: async (elements: Element[]): Promise<FilterResult | undefined> => {
+    if (config.fetchProfile.metadataQuery.isPartialFetch()) {
+      return
+    }
```
ori-moisis, line 42: why was this added?
what if the fetch target includes installed packages?

packages/salesforce-adapter/src/filters/create_missing_installed_packages_instances.ts
```diff
@@ -37,6 +37,9 @@ const createMissingInstalledPackageInstance = (file: FileProperties, installedPa
 const filterCreator: RemoteFilterCreator = ({ client, config }) => ({
   name: 'createMissingInstalledPackagesInstancesFilter',
   onFetch: async (elements: Element[]): Promise<FilterResult | undefined> => {
+    if (config.fetchProfile.metadataQuery.isPartialFetch()) {
```
tamtamirr, line 40: This can lead to removal of the `version` of `InstalledPackage` instances we actually fetch.
This is also relevant for `fetchWithChangeDetection` (quick fetch) as a followup to other tests I ran around my DRAFT PR

packages/salesforce-adapter/src/filters/create_missing_installed_packages_instances.ts
tamtamirr: Just commented on it asynchronously. This is also relevant for `fetchWithChangeDetection` flows. (Where I combined the meaning of `isPartialFetch` to be both for Targeted fetches and Quick fetches) but it can be handled in the other PR, I'll change that to check if `InstalledPackage` is included in the fetch instead

ori-moisis: but why is this needed at all?
the next lines will effectively do this (because in a partial fetch without installed packages, the installed packages type won't be in `elements`)
even if the change itself is ok, I don't see why it needs to be in this PR with the bugfix, it doesn't seem related

tamtamirr: will remove

ori-moisis (approved):

PR #4241, opened 2023-04-16
SALTO-3997: Change typeNameOverrides validations

shir-reifenberg:
Change typeNameOverrides validations

---

duplicates across `originalName` and `newName` are supported by the infra so in this PR I changed the validation to validate there are no dups in `newName` and `originalName` separately

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/config/swagger.ts
+++ b/packages/adapter-components/src/config/swagger.ts
@@ -166,13 +166,12 @@ const validateTypeNameOverrides = (
   typeNames: string[]
 ): void => {
   const allRenamedTypes = typeNameOverrides.flat()
-  const originalNames = allRenamedTypes.map(t => t.originalName)
   const newNames = allRenamedTypes.map(t => t.newName)
   const newNameDuplicates = findDuplicates(newNames)
   if (newNameDuplicates.length > 0) {
     throw new Error(`Duplicate type names in ${apiDefinitionConfigPath}.typeNameOverrides: ${newNameDuplicates}`)
   }
-  const originalNameDuplicates = findDuplicates(originalNames)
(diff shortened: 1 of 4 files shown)


packages/adapter-components/src/config/swagger.ts
```diff
@@ -160,6 +160,31 @@ export const createSwaggerAdapterApiConfigType = ({
   })
 }
 
+const validateTypeNameOverrides = (
+  apiDefinitionConfigPath: string,
+  typeNameOverrides: TypeNameOverrideConfig[],
+  typeNames: string[]
+): void => {
+  const allRenamedTypes = typeNameOverrides.flat()
```
netama, line 168: I see this is existing code, but I think we don't really need it (based on the function signature it's already flat)

packages/adapter-components/src/config/swagger.ts
```diff
@@ -160,6 +160,31 @@ export const createSwaggerAdapterApiConfigType = ({
...
+  const newNames = allRenamedTypes.map(t => t.newName)
+  const newNameDuplicates = findDuplicates(newNames)
+  if (newNameDuplicates.length > 0) {
+    throw new Error(`Duplicate type names in ${apiDefinitionConfigPath}.typeNameOverrides: ${newNameDuplicates}`)
+  }
+  const originalNameDuplicates = findDuplicates(allRenamedTypes.map(t => t.originalName))
```
netama, line 174: nit: `allRenamedTypes.map(t => t.originalName)` is the same as `typeNameOverrides.map(t => t.originalName)` below (assuming we don't need to flatten), so if we're saving `newNames` we should probably also create this for consistency

netama (approved):

PR #4243, opened 2023-04-17
SALTO-3512: Use implemented CV

tamtamirr:
The CV was already implemented but was not in use


tamtamirr:
I've ran some sanity tests and it seems like the messages are not aligned or incorrect in some cases. I'll comment that on the Jira ticket. We should handle it in a different PR though, so I'm closing this PR. cc. @benisraeladi @amitgoldarb

PR #4253, opened 2023-04-17
SALTO-4000: fix fetch error for one category in a brand and no articles

edenhassid:
_fix fetch error for one category in a brand and no articles_

---

---
_Release Notes_: 
zendesk:
* fix fetch error for one category in a brand and no articles

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/elements/ducktype/transformer.ts
+++ b/packages/adapter-components/src/elements/ducktype/transformer.ts
@@ -179,7 +179,7 @@ const getEntriesForType = async (
       return awu(makeArray(entry[nestedFieldDetails.field.name])).map(
         (nestedEntry, nesteIndex) => {
           if (!isObjectType(nestedFieldDetails.type)) {
-            log.error('type is not objectType returning undefined')
+            log.error(`for typeName ${typeName} in adapter ${adapterName} nestedFieldDetails.type is not objectType returning undefined`)
             return undefined
           }
           return toInstance({
--- a/packages/adapter-components/src/elements/ducktype/transformer.ts
+++ b/packages/adapter-components/src/elements/ducktype/transformer.ts
@@ -177,15 +177,21 @@ const getEntriesForType = async (
   const instances = await awu(entriesValues).flatMap(async (entry, index) => {
     if (nestedFieldDetails !== undefined) {
       return awu(makeArray(entry[nestedFieldDetails.field.name])).map(
-        (nestedEntry, nesteIndex) => toInstance({
-          entry: nestedEntry,
-          type: nestedFieldDetails.type,
-          transformationConfigByType,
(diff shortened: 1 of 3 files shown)


packages/adapter-components/src/elements/ducktype/transformer.ts
```diff
@@ -177,15 +177,21 @@ const getEntriesForType = async (
...
-          type: nestedFieldDetails.type,
-          transformationConfigByType,
-          transformationDefaultConfig,
-          defaultName: `unnamed_${index}_${nesteIndex}`, // TODO improve
-          hasDynamicFields,
-          getElemIdFunc,
-        })
+        (nestedEntry, nesteIndex) => {
+          if (!isObjectType(nestedFieldDetails.type)) {
+            log.error('type is not objectType returning undefined')
```
netama, line 182: nit: this doesn't provide enough context for debugging, let's include the `typeName` (and maybe also `adapterName`)
(also - realizing now this will log for every entry if there are multiple ones - but since we don't expect to reach this flow, I think it's not critical)

packages/adapter-components/src/elements/field_finder.ts
```diff
@@ -85,7 +94,12 @@ export const findDataField: FindNestedFieldFunc = async (type, fieldsToIgnore, d
     ? await nestedFieldType.getInnerType()
     // map type currently cannot be returned from nested fields
     : nestedFieldType
-
+  if (isPrimitiveType(nestedType) && nestedType.elemID.isEqual(BuiltinTypes.UNKNOWN.elemID)) {
+    return {
```
netama, line 98: let's add a comment explaining that this can happen when a `dataField` is specified but no entries are returned

packages/adapter-components/test/elements/ducktype/transformer.test.ts
```diff
@@ -130,6 +138,62 @@ describe('ducktype_transformer', () => {
...
+          }),
+          nestedTypes: [someNested],
+        }
+      })
+      mockPaginator = mockFunction<Paginator>().mockImplementation(async function *get() {
+        yield [{
+          meta: { has_more: false },
+          links: {
+            first: 'https://salto87.zendesk.com/api/v2/help_center/categories/15828521533331/articles?include=translations&page%5Bsize%5D=100&sort_by=updated_at',
+            last: 'https://salto87.zendesk.com/api/v2/help_center/categories/15828521533331/articles?include=translations&page%5Bbefore%5D=bGFzdF9wYWdl&page%5Bsize%5D=100&sort_by=updated_at',
```
netama, line 161: nit: let's just have sample strings here instead of the zendesk example?

netama (approved):
🔥

PR #4255, opened 2023-04-17
SALTO-3988 - Handle unresolved references in handle_template_expressions

idoser96:
If an instance that is being handled by handle_template_expression contains an unresolved reference, convert it to an empty string instead of crashing

---

None

---
_Release Notes_: 
Zendesk Adapter:
* Fixed deploy crashing with a modification of an instance with removal of a reference that contains an unresolved reference

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/filters/handle_template_expressions.ts
+++ b/packages/zendesk-adapter/src/filters/handle_template_expressions.ts
@@ -287,8 +287,10 @@ const replaceFormulasWithTemplates = async (
 }
 
 export const prepRef = (part: ReferenceExpression): TemplatePart => {
-  // There are cases where part is not resolved, which means that it has no value
+  // In some cases this function may run on the .before value of a Change, which may contain unresolved references.
+  // .after values are always resolved because unresolved references are dropped by unresolved_references validator
   // This case should be handled more generic but at the moment this is a quick fix to avoid crashing (SALTO-3988)
+  // This fix is enough since the .before value is not used in the deployment process
   if (part.value instanceof UnresolvedReference) {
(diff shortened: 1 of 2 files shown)


packages/zendesk-adapter/src/filters/handle_template_expressions.ts
```diff
@@ -287,6 +287,12 @@ const replaceFormulasWithTemplates = async (
 }
 
 export const prepRef = (part: ReferenceExpression): TemplatePart => {
+  // There are cases where part is not resolved, which means that it has no value
+  // This case should be handled more generic but at the moment this is a quick fix to avoid crashing (SALTO-3988)
```
netama, line 291: I think the explanation here should be different - we expect this to only happen on the `before` values (which are not deployed to the service), since if there is an `UnresolvedReference` in `after` the change should have been dropped by the `unresolved_references` change validator.

packages/zendesk-adapter/test/filters/handle_template_expressions.test.ts
```diff
@@ -441,4 +456,47 @@ describe('handle templates filter', () => {
       expect(elementsAfterOnDeploy).toEqual(elementsAfterFetch)
     })
   })
+
+  describe('prepRef', () => {
+    it('zendeskReferenceTypes', () => {
```
netama, line 461: please use meaningful test cases (e.g. "should return empty string when ...")

netama (commented):
looks good, added a few minor comments. please add release notes

netama (approved):
LGTM, please don't forget to add release notes (which are still missing)

PR #4256, opened 2023-04-17
SALTO-3975 Deploy group schema for custom fields and  add validator for base fields

idoamit8:
1. Created endpoint for groupSchema in the config file for modification.
2. We noticed that the user cant simply delete attribute from group schema because of Okta API demands.
3. So we created a filter that assign null value when the user delete a custom field in order for Okta to understand the field need to be deleted.
4. I also created change validator that checks that base attributes didn't change + test for the validator. 

---

To test the filter manually do as follows:
1. Delete a custom attribute from groupSchema instance. 
2. Do `salto deploy` and check that the changes occurred and that the nacl has no deleted field.
3. Try to add and modify custom fields as well. 

To test the validator manually, do as follows:
1. Change bae attribute in groupSchema instance. 
2. Do 'salto deploy --dry-run'
---
_Release Notes_: 
None
---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/change_validators/group_schema_modify_base_fields.ts
+++ b/packages/okta-adapter/src/change_validators/group_schema_modify_base_fields.ts
@@ -40,7 +40,7 @@ export const groupSchemaModifyBaseValidator: ChangeValidator = async changes => 
       elemID: instance.elemID,
       severity: 'Error',
       message: `Cannot change base attributes of ${GROUP_SCHEMA_TYPE_NAME}`,
-      detailedMessage: `Cannot change ${GROUP_SCHEMA_TYPE_NAME} because properties section can not be changed for existing fields.`,
+      detailedMessage: `Cannot change ${GROUP_SCHEMA_TYPE_NAME}, because base properties section can not be changed.`,
     }))
     .toArray()
 )
--- a/packages/okta-adapter/src/change_validators/group_schema_modify_base_fields.ts
+++ b/packages/okta-adapter/src/change_validators/group_schema_modify_base_fields.ts
@@ -1,0 +1,46 @@
+/*
+*                      Copyright 2023 Salto Labs Ltd.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+import { ChangeError, ChangeValidator, InstanceElement, ModificationChange, getChangeData, isEqualValues, isInstanceChange, isModificationChange } from '@salto-io/adapter-api'
+import { collections } from '@salto-io/lowerdash'
+import { GROUP_SCHEMA_TYPE_NAME } from '../constants'
+
+const { awu } = collections.asynciterable
+
+const isBaseFieldModified = (
+  change: ModificationChange<InstanceElement>
+): boolean => {
+  const { before, after } = change.data
+  const beforeBase = before.value.definitions.base
+  const afterBase = after.value.definitions.base
+  return !isEqualValues(beforeBase, afterBase)
+}
+
+export const groupSchemaModifyBaseValidator: ChangeValidator = async changes => (
+  awu(changes)
+    .filter(isInstanceChange)
+    .filter(isModificationChange)
+    .filter(change => getChangeData(change).elemID.typeName === GROUP_SCHEMA_TYPE_NAME)
+    .filter(isBaseFieldModified)
+    .map(getChangeData)
+    .map((instance: InstanceElement): ChangeError => ({
+      elemID: instance.elemID,
+      severity: 'Error',
+      message: `Cannot change base attributes of ${GROUP_SCHEMA_TYPE_NAME}`,
+      detailedMessage: `Cannot change ${GROUP_SCHEMA_TYPE_NAME} because properties section can not be changed for existing fields.`,
+    }))
+    .toArray()
+)
--- a/packages/okta-adapter/src/change_validators/group_schema_modify_base_fields.ts
+++ b/packages/okta-adapter/src/change_validators/group_schema_modify_base_fields.ts
@@ -23,7 +23,7 @@ const isBaseFieldModified = (
   const { before, after } = change.data
   const beforeBase = before.value?.definitions.base
   const afterBase = after.value?.definitions.base
-  return !isEqualValues(beforeBase, afterBase)
+  return !isEqualValues(beforeBase, afterBase, { compareReferencesByValue: true })
 }
 
 export const groupSchemaModifyBaseValidator: ChangeValidator = async changes => (
--- a/packages/okta-adapter/src/change_validators/group_schema_modify_base_fields.ts
+++ b/packages/okta-adapter/src/change_validators/group_schema_modify_base_fields.ts
@@ -39,8 +39,8 @@ export const groupSchemaModifyBaseValidator: ChangeValidator = async changes => 
     .map((instance: InstanceElement): ChangeError => ({
       elemID: instance.elemID,
       severity: 'Error',
-      message: `Cannot change base attributes of ${GROUP_SCHEMA_TYPE_NAME}`,
-      detailedMessage: `Cannot change ${GROUP_SCHEMA_TYPE_NAME}, because base properties section can not be changed.`,
+      message: `Cannot change base properties of ${GROUP_SCHEMA_TYPE_NAME}`,
+      detailedMessage: `It is posibble to modify the custom properties section of the ${GROUP_SCHEMA_TYPE_NAME} instance.`,
     }))
     .toArray()
 )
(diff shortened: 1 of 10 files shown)


packages/okta-adapter/test/filters/group_schema_null_when_remove.test.ts
```diff
@@ -0,0 +1,82 @@
...
+          toChange({ before: groupSchemaBeforeInstanceCopy, after: groupSchemaAfterInstanceCopy })])
+        expect(groupSchemaAfterInstanceCopy.value.definitions.custom.properties.additionalProperties.field2).toBeNull()
+      })
+    })
+    describe('onDeploy', () => {
+      it('should delete fields with null', async () => {
+        const groupSchemaAfterInstanceCopy = groupSchemaAfterInstance.clone()
+        await filter.onDeploy([
+          toChange({ before: groupSchemaBeforeInstance, after: groupSchemaAfterInstanceCopy })])
+        expect(groupSchemaAfterInstanceCopy.value.definitions.custom.properties.additionalProperties.field2)
```
shir-reifenberg, line 78: Maybe I'm missing something, but it sees like `groupSchemaAfterInstanceCopy` didn't have `field2` from the first place
we would want the input for this test would be an instance with field that have a `null` value, no?

packages/okta-adapter/src/filters/group_schema_null_when_remove.ts
```diff
@@ -0,0 +1,68 @@
...
+import { Change, InstanceElement, ModificationChange, getChangeData, isInstanceChange, isModificationChange } from '@salto-io/adapter-api'
+import { resolvePath } from '@salto-io/adapter-utils'
+import { FilterCreator } from '../filter'
+import { GROUP_SCHEMA_TYPE_NAME } from '../constants'
+
+const CUSTOM_ADDITONAL_PROPERTIES_PATH = ['definitions', 'custom', 'properties', 'additionalProperties']
+
+const addNullToRemovedFields = (change: ModificationChange<InstanceElement>): void => {
+  const { before, after } = change.data
+  const customAdditionalPropertiesPath = before.elemID.createNestedID(...CUSTOM_ADDITONAL_PROPERTIES_PATH)
```
shir-reifenberg, line 26: nit: 
```suggestion
  const customPropertiesPath = before.elemID.createNestedID(...CUSTOM_ADDITONAL_PROPERTIES_PATH)
```

shir-reifenberg (commented):
Looks good ! 😃 Added a few comments but nothing major

packages/okta-adapter/src/change_validators/group_schema_modify_base_fields.ts
```diff
@@ -0,0 +1,46 @@
...
+  change: ModificationChange<InstanceElement>
+): boolean => {
+  const { before, after } = change.data
+  const beforeBase = before.value.definitions.base
+  const afterBase = after.value.definitions.base
+  return !isEqualValues(beforeBase, afterBase)
+}
+
+export const groupSchemaModifyBaseValidator: ChangeValidator = async changes => (
+  awu(changes)
```
shir-reifenberg, line 33: Do we need to use `awu` here?

packages/okta-adapter/src/change_validators/group_schema_modify_base_fields.ts
```diff
@@ -0,0 +1,46 @@
...
+import { collections } from '@salto-io/lowerdash'
+import { GROUP_SCHEMA_TYPE_NAME } from '../constants'
+
+const { awu } = collections.asynciterable
+
+const isBaseFieldModified = (
+  change: ModificationChange<InstanceElement>
+): boolean => {
+  const { before, after } = change.data
+  const beforeBase = before.value.definitions.base
```
shir-reifenberg, line 27: value is from type `any`, so it's not guaranteed `definitions` will exist. this way in case it doesn't the program will not crash 
```suggestion
  const beforeBase = before.value.definitions?.base
```

packages/okta-adapter/src/filters/group_schema_field_removal.ts
```diff
@@ -0,0 +1,74 @@
...
+
+const CUSTOM_ADDITONAL_PROPERTIES_PATH = ['definitions', 'custom', 'properties', 'additionalProperties']
+
+const addNullToRemovedProperties = (change: ModificationChange<InstanceElement>): void => {
+  const { before, after } = change.data
+  const customPropertiesPath = before.elemID.createNestedID(...CUSTOM_ADDITONAL_PROPERTIES_PATH)
+  const beforeCustomProperties = resolvePath(before, customPropertiesPath)
+  const afterCustomProperties = resolvePath(after, customPropertiesPath)
+  const afterCustomPropertiesKeys = Object.keys(afterCustomProperties)
+  const PropertiesToRemove = Object.keys(beforeCustomProperties)
```
shir-reifenberg, line 30:
```suggestion
  const propertiesToRemove = Object.keys(beforeCustomProperties)
```

packages/okta-adapter/src/filters/group_schema_field_removal.ts
```diff
@@ -0,0 +1,74 @@
...
+    .filter(key => !(afterCustomPropertiesKeys).includes(key))
+
+  PropertiesToRemove
+    .forEach(Property => {
```
shir-reifenberg, line 34:
```suggestion
    .forEach(property => {
```

packages/okta-adapter/src/filters/group_schema_field_removal.ts
```diff
@@ -0,0 +1,74 @@
...
+import { GROUP_SCHEMA_TYPE_NAME } from '../constants'
+
+const CUSTOM_ADDITONAL_PROPERTIES_PATH = ['definitions', 'custom', 'properties', 'additionalProperties']
+
+const addNullToRemovedProperties = (change: ModificationChange<InstanceElement>): void => {
+  const { before, after } = change.data
+  const customPropertiesPath = before.elemID.createNestedID(...CUSTOM_ADDITONAL_PROPERTIES_PATH)
+  const beforeCustomProperties = resolvePath(before, customPropertiesPath)
+  const afterCustomProperties = resolvePath(after, customPropertiesPath)
+  const afterCustomPropertiesKeys = Object.keys(afterCustomProperties)
```
shir-reifenberg, line 29: nit
```suggestion
  const afterCustomPropertiesKeys = new Set(Object.keys(afterCustomProperties))
```

shir-reifenberg (approved):
Looks great :) added just a few more comments but approving

packages/okta-adapter/src/filters/group_schema_field_removal.ts
```diff
@@ -0,0 +1,74 @@
...
+  const PropertiesToRemove = Object.keys(beforeCustomProperties)
+    .filter(key => !(afterCustomPropertiesKeys).includes(key))
```
netama, line 31: nit: the parentheses here seem off (did you mean `!afterCustomPropertiesKeys.includes(key)`?)
also - if there are a lot of custom properties this can be inefficient. I think we should either use a set, or check if 
```suggestion
    .filter(fieldName => Object.prototype.hasOwnProperty.call(afterCustomProperties, fieldName))
```
(probably not critical for most cases here)

packages/okta-adapter/src/change_validators/group_schema_modify_base_fields.ts
idoamit8: Nope :)

packages/okta-adapter/src/change_validators/group_schema_modify_base_fields.ts
idoamit8: I wrote the question mark here:`after.value?.definitions.base`.

packages/okta-adapter/src/change_validators/group_schema_modify_base_fields.ts
```diff
@@ -0,0 +1,46 @@
...
+    .filter(isInstanceChange)
+    .filter(isModificationChange)
+    .filter(change => getChangeData(change).elemID.typeName === GROUP_SCHEMA_TYPE_NAME)
+    .filter(isBaseFieldModified)
+    .map(getChangeData)
+    .map((instance: InstanceElement): ChangeError => ({
+      elemID: instance.elemID,
+      severity: 'Error',
+      message: `Cannot change base properties of ${GROUP_SCHEMA_TYPE_NAME}`,
+      detailedMessage: `It is posibble to modify the custom properties section of the ${GROUP_SCHEMA_TYPE_NAME} instance.`,
```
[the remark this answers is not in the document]
idoamit8: Thank you. Who can I consult with about the correct format?

packages/okta-adapter/src/filters/group_schema_field_removal.ts
idoamit8: Changed to Set

packages/okta-adapter/src/change_validators/group_schema_modify_base_fields.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const { awu } = collections.asynciterable
+
+const isBaseFieldModified = (
+  change: ModificationChange<InstanceElement>
+): boolean => {
+  const { before, after } = change.data
+  const beforeBase = before.value.definitions.base
+  const afterBase = after.value.definitions.base
+  return !isEqualValues(beforeBase, afterBase)
```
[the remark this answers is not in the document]
idoamit8: You are correct, I added `{ compareReferencesByValue: true }`

packages/okta-adapter/src/change_validators/group_schema_modify_base_fields.ts
shir-reifenberg: @idoamit8 I don't think it's enough, even if `value` is defined it might still be possible that `definitions` isn't
Also, if i'm not wrong, `value` is part of the type definition so it's not necessary

PR #4258, opened 2023-04-18
SALTO- 4008 - Move FetchDiffToWorkspace to core + add loadElementsFromFolder to the adapter api

adamro:
What's here? 

CLI - 
* Renamed `fetch-diff` to `apply-patch`
* Moved the Changes calculation logic to be a core api to extracted a lot of the logic outside of the CLI

Core - 
* Added `loadElementsFromFolder` to the adapter api (not under operations cause it doesn't require the context)
* Created a new `calculatePatch` api that gets a workspace and 2 folders with SFDX content (this is currently restricted only to salesforce) and returns a "FetchResult" to apply to the workspace

---

TODO:
- [x]  Add unit tests to the new core api
- [x]  Decide if to change the tests logic of the command
- [x]  Manually test that this doesn't break anything

---
_Release Notes_: 

*Core*:
* Add calculatePatch to the core api. It gets a workspace and two path directories with external service format (ie. SFDX) content, calculates the diff between them, translates them to Changes and returns "FetchResult" to be applied to the Workspace.

*CLI*:
* Rename the `fetch-diff` command to `apply-patch`

---
_User Notifications_: 
None

### DIFF
--- a/packages/cli/src/commands/apply_patch.ts
+++ b/packages/cli/src/commands/apply_patch.ts
@@ -97,7 +97,7 @@ const applyPatchToWorkspace = async (
   }
   if (fetchErrors.length > 0) {
     // We currently assume all fetchErrors are warnings
-    log.debug(`fetch-diff had ${fetchErrors.length} warnings`)
+    log.debug(`apply-patch had ${fetchErrors.length} warnings`)
     outputLine(
       formatFetchWarnings(fetchErrors.map(fetchError => fetchError.message)),
       output,
--- a/packages/cli/src/commands/apply_patch.ts
+++ b/packages/cli/src/commands/apply_patch.ts
@@ -14,38 +14,21 @@
 * limitations under the License.
 */
 import _ from 'lodash'
-import { DetailedChange, Element, getChangeData, isAdditionChange, isRemovalChange } from '@salto-io/adapter-api'
+import { DetailedChange, getChangeData, isAdditionChange, isRemovalChange } from '@salto-io/adapter-api'
 import { applyDetailedChanges } from '@salto-io/adapter-utils'
 import { logger } from '@salto-io/logging'
+import { State, Workspace } from '@salto-io/workspace'
 import { collections } from '@salto-io/lowerdash'
-import { expressions, elementSource, Workspace, merger, State } from '@salto-io/workspace'
-import { loadElementsFromFolder } from '@salto-io/salesforce-adapter'
-import { calcFetchChanges } from '@salto-io/core'
+import { calculatePatch } from '@salto-io/core'
 import { WorkspaceCommandAction, createWorkspaceCommand } from '../command_builder'
 import { outputLine, errorOutputLine } from '../outputer'
 import { validateWorkspace, formatWorkspaceErrors } from '../workspace/workspace'
 import { CliExitCode, CliOutput } from '../types'
 import { UpdateModeArg, UPDATE_MODE_OPTION } from './common/update_mode'
+import { formatFetchWarnings } from '../formatter'
 
 const log = logger(module)
 const { awu } = collections.asynciterable
-
-const mergeElements = async (
-  workspace: Workspace,
-  elements: Element[],
-  output: CliOutput,
-): Promise<{ elements: Element[]; success: boolean }> => {
-  const result = await merger.mergeElements(awu(elements))
-  const errors = await awu(result.errors.values()).flat().toArray()
-  if (errors.length > 0) {
-    errorOutputLine('Encountered merge errors in elements:', output)
-    errorOutputLine(await formatWorkspaceErrors(workspace, errors), output)
-  }
-  return {
-    elements: await awu(result.merged.values()).toArray(),
-    success: errors.length === 0,
-  }
-}
 
 const updateStateElements = async (
   state: State,
@@ -72,7 +55,7 @@ const updateStateElements = async (
   })
 }
 
-type FetchDiffArgs = {
+type ApplyPatchArgs = {
   fromDir: string
   toDir: string
   accountName: 'salesforce'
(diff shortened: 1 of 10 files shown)


packages/cli/src/commands/apply_patch.ts
```diff
@@ -140,6 +97,14 @@ const fetchDiffToWorkspace = async (
     errorOutputLine(`Failed to update env ${workspace.currentEnv()} because there are ${conflicts.length} conflicting changes`, output)
     return false
   }
+  if (fetchErrors.length > 0) {
+    // We currently assume all fetchErrors are warnings
+    log.debug(`fetch-diff had ${fetchErrors.length} warnings`)
```
dantal4, line 102: or `apply-patch` ?
```suggestion
    log.debug(`calculate-patch had ${fetchErrors.length} warnings`)
```

packages/cli/src/commands/index.ts
```diff
@@ -14,7 +14,7 @@
 * limitations under the License.
 */
 import fetchDef from './fetch'
-import fetchDiffDef from './fetch_diff'
+import fetchDiffDef from './apply_patch'
```
dantal4, line 17:
```suggestion
import applyPatchDef from './apply_patch'
```

packages/cli/test/commands/apply_patch.test.ts
```diff
@@ -144,21 +157,67 @@ describe('fetch-diff command', () => {
...
+      )
+      const pendingChange = detailedCompare(
+        unchangedEmployeeType,
+        toDirEmployeeType,
+      )
+      mockCalculatePatch.mockResolvedValue({
+        changes: [
+          { change: modifyTypeChanges[0], serviceChanges: [modifyTypeChanges[0]], pendingChanges: [pendingChange[0]] },
+        ],
+        mergeErrors: [],
```
dantal4, line 177: where do we get the conflict errors in this test? from `validateWorkspace(workspace)` ?

packages/core/src/api.ts
```diff
@@ -369,6 +370,91 @@ export const fetchFromWorkspace: FetchFromWorkspaceFunc = async ({
   }
 }
 
+type CalculatePatchArgs = {
+  workspace: Workspace
+  fromDir: string
+  toDir: string
+  // This currently restricts the account name to be salesforce
+  // when using this feature
+  accountName: 'salesforce'
```
dantal4, line 379: As account name can be different of "salesforce" and the service still be "salesforce" - consider making it a `string` and have the allowed services validation from `accountToServiceNameMap`.

packages/core/src/api.ts
```diff
@@ -369,6 +370,91 @@ export const fetchFromWorkspace: FetchFromWorkspaceFunc = async ({
...
+    workspace,
+    fromDir,
+    toDir,
+    accountName,
+  }: CalculatePatchArgs,
+): Promise<FetchResult> => {
+  const accountToServiceNameMap = getAccountToServiceNameMap(workspace, workspace.accounts())
+  const { loadElementsFromFolder } = adapterCreators[accountToServiceNameMap[accountName]]
+  if (loadElementsFromFolder === undefined) {
+    throw new Error(`Account ${accountName}'s adapter ${accountToServiceNameMap[accountName]} does not support fetch diff`)
```
dantal4, line 393:
```suggestion
    throw new Error(`Account ${accountName}'s adapter ${accountToServiceNameMap[accountName]} does not support calculate patch`)
```

packages/core/src/api.ts
```diff
@@ -369,6 +370,91 @@ export const fetchFromWorkspace: FetchFromWorkspaceFunc = async ({
...
+  }
+
+  const wsElements = await workspace.elements()
+  const resolvedWSElements = await expressions.resolve(
```
dantal4, line 397: 1. Why do we need to resolve all elements here?
2. There's a code in `adapters/adapters.ts` that is maybe more accurate for using here too:
```
elementsSource: createElemIDReplacedElementsSource(filterElementsSource(
    elementsSource, account
), account, accountToServiceName[account]),
```

packages/core/src/api.ts
```diff
@@ -369,6 +370,91 @@ export const fetchFromWorkspace: FetchFromWorkspaceFunc = async ({
...
+    await awu(await wsElements.getAll()).toArray(),
+    wsElements,
+  )
+  const resolvedWSElementSource = elementSource.createInMemoryElementSource(resolvedWSElements)
+
+  const { elements: beforeElements, errors: beforeLoadErrors } = await loadElementsFromFolder({
```
dantal4, line 403: can you put lines 403-417 & line 435 in a function and call it with `fromDir` & `toDir`?

packages/core/test/api.test.ts
```diff
@@ -738,6 +754,119 @@ describe('api.ts', () => {
       })
     })
   })
+
+  describe('fetchDiffToWorkspace', () => {
```
dantal4, line 758:
```suggestion
  describe('applyPatchToWorkspace', () => {
```

dantal4 (commented):
Great work!

packages/core/src/api.ts
adamro: In the CLI we decided to limit this to `salesforce` only for now and I and as first step this will also be true in the SaaS. I kinda prefer to wait for a real demand for this before doing it cause it open to more error cases.
You feel strongly bout this from day one? 🤔

packages/core/src/api.ts
adamro: 1. We need this resolved to calculate the right plan.

2. More accurate in what sense?

packages/core/test/api.test.ts
adamro: Actually `calculatePatch` but thx :D

packages/core/src/api.ts
adamro: I really understand why you would say that but I started with that and honestly it's worse than this :D 
I can still do it if you feel strongly bout it

packages/cli/test/commands/apply_patch.test.ts
adamro: the pendingChanges in the FetchResult represent "conflicts"

packages/core/src/api.ts
dantal4: Not strongly, but it would be easier to do it now, when we have the full context, and it is a pretty small change.
I don't mind leaving the limitation in the CLI, but IMO the core API should support it :)

packages/core/src/api.ts
dantal4: In regular fetch we pass the adapter the elements source that I mentioned (using `await workspace.elements()` that is unresolved).

packages/core/src/api.ts
adamro: giving it another try

packages/core/src/api.ts
adamro: 👍🏻

packages/core/src/api.ts
adamro: done

packages/core/src/api.ts
ori-moisis: in a regular fetch the adapter (at least salesforce) assumes in many places that all of the instances that were fetched are fully resolved.
the element source that is passed into the adapter on fetch is used mostly to find references and to enrich the elements that were fetched, but there is always a separation between things that were fetched (the "elements" list that all the filters get) and elements from the element source.

it would be much more difficult to find all of these assumptions in the adapter code (again, this is true at least for the salesforce adapter) then it is to pass in a fully resolved element source.

so while you are correct in your comment, unfortunately this is not really the same use case as a fetch that gets that partially resolved element source

dantal4 (approved):
💯

PR #4263, opened 2023-04-18
SALTO-3918 author information index

dantal4:
Added author information index that maps element ids to author information (changedBy & changedAt).
Added `getElementAuthorInformation` function to workspace API.

---

---
_Release Notes_: 
Workspace:
- Added author information index that maps element ids to author information (changedBy & changedAt)
- Added `getElementAuthorInformation` function to workspace API

---
_User Notifications_: 
None

### DIFF
--- a/packages/workspace/src/workspace/index_utils.ts
+++ b/packages/workspace/src/workspace/index_utils.ts
@@ -13,9 +13,13 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
-import { ReadOnlyElementsSource, Element, Change, toChange } from '@salto-io/adapter-api'
+import { logger } from '@salto-io/logging'
+import { ReadOnlyElementsSource, Element, Change, toChange, isObjectTypeChange, isAdditionOrRemovalChange, getChangeData, isAdditionChange, ObjectType, Field } from '@salto-io/adapter-api'
 import { collections } from '@salto-io/lowerdash'
+import { RemoteMap } from './remote_map'
+import { ElementsSource } from './elements_source'
 
+const log = logger(module)
 const { awu } = collections.asynciterable
 
 export const getAllElementsChanges = async (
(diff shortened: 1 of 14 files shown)


alonstern (commented):
LGTM 👍 
@ori-moisis Do you want to have a look too?

packages/workspace/src/workspace/workspace.ts
```diff
@@ -1131,20 +1160,21 @@ export const loadWorkspace = async (
...
+      id,
+      envName,
+      async envState => await envState.referenceTargets.get(id.getFullName()) ?? []
+    ),
+    getElementIncomingReferences: async (id, envName = currentEnv()) => validateBaseIdAndRunOnEnvState(
+      id,
+      envName,
+      async envState => await envState.referenceSources.get(id.getFullName()) ?? []
+    ),
+    getElementAuthorInformation: async (id, envName = currentEnv()) => validateBaseIdAndRunOnEnvState(
```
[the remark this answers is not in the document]
dantal4: The `get` call in the UI remote map implementation is debounced so it is actually a `getMany` call.

packages/workspace/src/workspace/index_utils.ts
```diff
@@ -26,3 +30,70 @@ export const getAllElementsChanges = async (
...
+const getFieldChangesFromTypeChange = (change: Change<ObjectType>): Change<Field>[] => {
+  if (isAdditionOrRemovalChange(change)) {
+    return Object.values(getChangeData(change).fields).map(field => (
+      isAdditionChange(change)
+        ? toChange({ after: field })
+        : toChange({ before: field })
+    ))
+  }
+  const { before, after } = change.data
+  const allFieldNames = Object.keys({ ...before.fields, ...after.fields })
```
ori-moisis, line 43: could crash if we have many fields?
maybe change this to
```suggestion
  const allFieldNames = new Set(Object.keys(before.fields).concat(Object.keys(after.fields)))
```

ori-moisis (approved):

packages/workspace/src/workspace/index_utils.ts
dantal4: From what I saw we got `RangeError: Maximum call stack size exceeded` only when we used spread in a function input (e.g `a.push(...b)`), but not inside an object/list creation.

PR #4265, opened 2023-04-18
SALTO-4010: Resolve fetch failures for non admin users

IdoZakk:
Fixed for automation, automationLabel, and Priority Scheme

---
_Release Notes_: 
Fixed a bug that failed fetches of non-admin users

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/automation/automation_fetch.ts
+++ b/packages/jira-adapter/src/filters/automation/automation_fetch.ts
@@ -15,12 +15,12 @@
 */
 import { ElemIdGetter, InstanceElement, isInstanceElement, ObjectType, Values } from '@salto-io/adapter-api'
 import { createSchemeGuard, naclCase, pathNaclCase } from '@salto-io/adapter-utils'
-import { elements as elementUtils } from '@salto-io/adapter-components'
+import { elements as elementUtils, client as clientUtils } from '@salto-io/adapter-components'
 import { values as lowerdashValues } from '@salto-io/lowerdash'
 import { logger } from '@salto-io/logging'
 import Joi from 'joi'
 import _ from 'lodash'
-import { AUTOMATION_TYPE, JIRA, PROJECT_TYPE } from '../../constants'
+import { AUTOMATION_TYPE, fetchFailedWarnings, JIRA, PROJECT_TYPE } from '../../constants'
 import JiraClient from '../../client/client'
 import { FilterCreator } from '../../filter'
 import { createAutomationTypes } from './types'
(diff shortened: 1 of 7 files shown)


packages/jira-adapter/src/filters/automation/automation_label/label_fetch.ts
```diff
@@ -79,21 +63,54 @@ export const filter: FilterCreator = ({ client, getElemIdFunc, config, fetchQuer
...
-
-    const automationLabelType = createAutomationLabelType()
-    automationLabels.forEach(automationLabel => elements.push(
-      createInstance(automationLabel, automationLabelType, getElemIdFunc),
-    ))
-    elements.push(automationLabelType)
+    // get labels
+    const url = client.isDataCenter
+      ? '/rest/cb-automation/latest/rule-label'
+      : `/gateway/api/automation/internal-api/jira/${await getCloudId(client)}/pro/rest/GLOBAL/rule-labels`
```
alonstern, line 77: I think the `getCloudId` should also be inside the try-catch. In that case can you add back the `getAutomationLabels` function?

packages/jira-adapter/src/filters/automation/automation_fetch.ts
```diff
@@ -148,14 +148,31 @@ const filter: FilterCreator = ({ client, getElemIdFunc, config, fetchQuery }) =>
...
+      const { automationType, subTypes } = createAutomationTypes()
+
+      automations.forEach(automation => elements.push(
+        createInstance(automation, automationType, idToProject, getElemIdFunc),
+      ))
+      elements.push(automationType, ...subTypes)
+      return undefined
+    } catch (e) {
+      if (e instanceof clientUtils.HTTPError && e.response !== undefined
+        && (e.response.status === 403
```
alonstern, line 162: Didn't we decide to do that for all errors?

packages/jira-adapter/src/filters/data_center/priority_scheme/priority_scheme_fetch.ts
```diff
@@ -64,20 +66,37 @@ const filter: FilterCreator = ({ paginator, client, fetchQuery, getElemIdFunc })
...
-      })
-
-    elements.push(type)
+      elements.push(type)
+      return undefined
+    } catch (e) {
+      if (e instanceof clientUtils.HTTPError && e.response !== undefined
+        && (e.response.status === 403
+          || e.response.status === 405)) {
+        log.error(`Received a ${e.response.status} error when fetching priority schemes. Please make sure you have the "Automation" permission enabled in Jira.`)
```
alonstern, line 88: Replace `"Automation"` in the message

alonstern (commented):
👍

packages/jira-adapter/src/filters/automation/automation_fetch.ts
IdoZakk: I saw there are tests for specific problems, and also I saw we have this protection (with 403) in the swagger framework so I though it will be better this way

packages/jira-adapter/src/filters/automation/automation_label/label_fetch.ts
IdoZakk: I took it out of the function to prevent failure in case it is not aLabelsGetResponse without complicating the catch too much ,seemed like the function increases the complication of the code instead of simplifying it this way. I will add the getCloudId into the try-catch

alonstern (approved):
👌

PR #4266, opened 2023-04-18
SALTO-3969 Avoid logging inside references in merge logs

netama:
Alternative to https://github.com/salto-io/salto/pull/4215 (that also reverts it), since it seems using `safeJsonStringify` with a replacer here is not efficient enough (`inspect` truncates after a max depth / array length).

---

there was a problem with adding similar handling for static files that requires more debugging and is not critical to the fix, so starting without it

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-api/src/element_id.ts
+++ b/packages/adapter-api/src/element_id.ts
@@ -14,6 +14,7 @@
 * limitations under the License.
 */
 import _ from 'lodash'
+import { inspect } from 'util'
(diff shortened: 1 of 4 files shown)


ori-moisis (approved):

ori-moisis:
I wonder if this is just a better solution overall now than using safe stringify in logs and we should also replace the other places that did that (in a separate, less urgent PR)

PR #4267, opened 2023-04-18
SALTO-3059 use inspect instead of safeJsonStringify with replacer

netama:
Switch to `inspect` for performance reasons.

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-utils/src/utils.ts
+++ b/packages/adapter-utils/src/utils.ts
@@ -16,6 +16,7 @@
 import os from 'os'
 import wu from 'wu'
 import _ from 'lodash'
+import { inspect, InspectOptions } from 'util'
 import safeStringify from 'fast-safe-stringify'
 import { logger } from '@salto-io/logging'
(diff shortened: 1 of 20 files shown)


ori-moisis (approved):
maybe add to "InvalidStaticFile" as well

packages/adapter-utils/src/utils.ts
```diff
@@ -983,29 +984,29 @@ export const createDefaultInstanceFromType = async (name: string, objectType: Ob
...
   replacer?: Replacer,
   space?: string | number
 ): string =>
   safeStringify(value, replacer, space)
 
+export const safeStringifyLimited = (
+  value: Value,
+  options?: InspectOptions
+): string => (
+  inspect(value, _.defaults({}, options ?? {}, { depth: 4 }))
```
ori-moisis, line 1007: we may want to add a max string length here as well?

PR #4270, opened 2023-04-19
SALTO-4019 throw on missing SuiteBundler feature in SuiteApp FileCabinet fetch

dantal4:
Throw an indicative error instead of throwing a generic "Failed to list folders" error.

---

---
_Release Notes_: 
Netsuite Adapter:
- Throw indicative error on missing SuiteBundler feature in SuiteApp FileCabinet fetch

---
_User Notifications_: 
None

### DIFF
--- a/packages/netsuite-adapter/src/client/suiteapp_client/suiteapp_client.ts
+++ b/packages/netsuite-adapter/src/client/suiteapp_client/suiteapp_client.ts
@@ -145,8 +145,10 @@ export default class SuiteAppClient {
    * ORDER BY <some unique identifier> ASC/DESC in your queries.
    * Otherwise, you might not get all the results.
    */
-  public async runSuiteQL(query: string):
-    Promise<Record<string, unknown>[] | undefined> {
+  public async runSuiteQL(
+    query: string,
+    throwOnErrors: Record<string, string> = {}
+  ): Promise<Record<string, unknown>[] | undefined> {
     log.debug('Running SuiteQL query: %s', query)
     if (!/ORDER BY .* (ASC|DESC)/.test(query)) {
       log.warn(`SuiteQL ${query} does not contain ORDER BY <unique identifier> ASC/DESC, which can cause the response to not contain all the results`)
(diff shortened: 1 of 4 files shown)


packages/netsuite-adapter/src/client/suiteapp_client/suiteapp_client.ts
```diff
@@ -163,10 +165,17 @@ export default class SuiteAppClient {
...
-        log.error('SuiteQL query error - %s', query, { error })
+        log.warn('SuiteQL query error - %s', query, { error })
         if (error instanceof InvalidSuiteAppCredentialsError) {
           throw error
         }
+        if (axios.isAxiosError(error)) {
+          const errorDetailedMessage = getAxiosErrorDetailedMessage(error)
+          const matchingErrorKey = Object.keys(throwOnErrors).find(e => errorDetailedMessage?.includes(e))
+          if (matchingErrorKey !== undefined) {
+            throw new Error(throwOnErrors[matchingErrorKey])
```
omrilit, line 176: Do we want to add a debug log here?

packages/netsuite-adapter/src/suiteapp_file_cabinet.ts
```diff
@@ -159,6 +159,11 @@ const MAX_DEPLOYABLE_FILE_SIZE = 10 * 1024 * 1024
 const DEPLOY_CHUNK_SIZE = 50
 const MAX_ITEMS_IN_WHERE_QUERY = 200
 
+export const THROW_ON_MISSING_FEATURE_ERROR = {
```
omrilit, line 162: nit: 
```suggestion
export const THROW_ON_MISSING_FEATURE_ERROR : Record<string, string> = {
```

omrilit (approved):
Thanks!

packages/netsuite-adapter/src/client/suiteapp_client/suiteapp_client.ts
dantal4: It is thrown and fails the fetch operation, so it feels redundant, isn't it?

PR #4274, opened 2023-04-19
Salto 3853 limit filecabinet size

adfineberg:
_Limit the filecabinet size. Exclude large folders according to the size defined in the configuration._

---
---
_Release Notes_: 
_NetSuite adapter_
* Limit the filecabinet size. Exclude large folders according to the size defined in the configuration.
---


packages/netsuite-adapter/config_doc.md
```diff
@@ -119,6 +120,8 @@ netsuite {
 | maxItemsInImportObjectsRequest | 40                      | Limits the max number of requested items a single import-objects request
 | sdfConcurrencyLimit            | 4                       | Limits the max number of concurrent SDF API calls. The number should not exceed the concurrency limit enforced by the upstream service.
 | installedSuiteApps             | []                      | The SuiteApps ids to deploy and fetch elements from
+| maxFileCabinetSize             | 1                       | Limits the max size in GB of the fileCabinet size
```
omrilit, line 123: What is this number based on? do we want to have a different number for when using the SuiteApp and when not?

packages/netsuite-adapter/src/client/sdf_client.ts
```diff
@@ -545,7 +548,7 @@ export default class SdfClient {
         return await this.runImportObjectsCommand(executor, ALL, ALL, suiteAppId)
       } catch (e) {
         log.warn(`Attempt to fetch all custom objects has failed with suiteApp: ${suiteAppId}`)
-        log.warn(e)
+        log.warn(e as Error)
```
omrilit, line 551: Just out of curiosity, why?

packages/netsuite-adapter/src/client/sdf_client.ts
```diff
@@ -795,6 +798,20 @@ export default class SdfClient {
...
+      const normalizedPath = (filePath: string): string => {
+        const filePathParts = filePath.split(FILE_CABINET_PATH_SEPARATOR)
+        return osPath.join(fileCabinetDirPath, ...filePathParts)
+      }
+      return Object.assign({}, ...await withLimitedConcurrency(
+        filePaths.map(filePath => async () => (
+          { [filePath]: (await stat(normalizedPath(filePath))).size }
+        )),
+        READ_CONCURRENCY
+      ))
```
omrilit, line 812: Better to avoid `...` in potentially long arrays/objects.
```suggestion
      return Object.fromEntries(await withLimitedConcurrency(
        filePaths.map(filePath => async () => [filePath, await stat(normalizedPath(filePath))).size])),
        READ_CONCURRENCY
      ))
```

omrilit (commented):
Really partial review, publishing my comments meanwhile

packages/netsuite-adapter/src/client/sdf_client.ts
adfineberg: I started with that, but I didn't manage to set up the types correctly. I'll try again

packages/netsuite-adapter/config_doc.md
adfineberg: No real basis, let me know if there's a better one - and if you prefer to split it for the account types.
Not sure I agree with about the documentation - my guess is people would prefer to shorten the fetch time when possible.

packages/netsuite-adapter/src/config.ts
```diff
@@ -547,12 +563,30 @@ export const UPDATE_SUITEAPP_TYPES_CONFIG_FORMAT = 'Some type names have been ch
 export const UPDATE_DEPLOY_CONFIG = 'All deploy\'s configuration flags are under "deploy" configuration.'
 + ' you may leave "deploy" section as undefined to set all deploy\'s configuration flags to their default value.'
 
-const createExclude = ({
-  filePaths: failedPaths = [],
-  types: failedTypes = {},
-}: Pick<NetsuiteQueryParameters, 'types' | 'filePaths'>): QueryParams =>
+const createFolderExclude = (folderPaths: NetsuiteFilePathsQueryParams): string[] =>
+  folderPaths.map(folder => {
+    const wrapWithSeparators = (): string => {
```
dantal4, line 568: this function makes it harder to understand the flow - makes me jump up and down when this is actually a simple flow.

packages/netsuite-adapter/src/config.ts
```diff
@@ -547,12 +563,30 @@ export const UPDATE_SUITEAPP_TYPES_CONFIG_FORMAT = 'Some type names have been ch
...
+      let normalizedFolder = folder
+      if (!folder.startsWith(FILE_CABINET_PATH_SEPARATOR)) {
```
dantal4, line 570: shouldn't happen

packages/netsuite-adapter/src/config.ts
```diff
@@ -547,12 +563,30 @@ export const UPDATE_SUITEAPP_TYPES_CONFIG_FORMAT = 'Some type names have been ch
...
+        normalizedFolder = `${FILE_CABINET_PATH_SEPARATOR}${normalizedFolder}`
+      }
+      if (!folder.endsWith(FILE_CABINET_PATH_SEPARATOR)) {
```
dantal4, line 573: always happen, right?

packages/netsuite-adapter/src/config.ts
```diff
@@ -547,12 +563,30 @@ export const UPDATE_SUITEAPP_TYPES_CONFIG_FORMAT = 'Some type names have been ch
...
 + ' you may leave "deploy" section as undefined to set all deploy\'s configuration flags to their default value.'
 
-const createExclude = ({
-  filePaths: failedPaths = [],
-  types: failedTypes = {},
-}: Pick<NetsuiteQueryParameters, 'types' | 'filePaths'>): QueryParams =>
+const createFolderExclude = (folderPaths: NetsuiteFilePathsQueryParams): string[] =>
+  folderPaths.map(folder => {
+    const wrapWithSeparators = (): string => {
+      let normalizedFolder = folder
```
dantal4, line 569: prefer immutable over mutable - if you can solve it with consts and not `let` it's better

packages/netsuite-adapter/src/config.ts
```diff
@@ -563,22 +597,24 @@ const toConfigSuggestions = (
 ): NetsuiteConfig => {
   const config: NetsuiteConfig = {}
 
-  if (!_.isEmpty(failedFilePaths.otherError) || !_.isEmpty(failedTypes.unexpectedError)) {
+  if (!_.isEmpty(failedFilePaths.otherError) || !_.isEmpty(failedTypes.unexpectedError)
+    || !_.isEmpty(failedFilePaths?.largeFolderError)) {
```
dantal4, line 601:
```suggestion
    || !_.isEmpty(failedFilePaths.largeFolderError)) {
```

dantal4 (commented):
Nice work!

packages/netsuite-adapter/src/client/suiteapp_client/suiteapp_file_cabinet.ts
```diff
@@ -405,6 +407,8 @@ SuiteAppFileCabinetOperations => {
       },
     }))
 
+    const fullPath = (file: { path: string[] }): string => file.path.join(FILE_CABINET_PATH_SEPARATOR)
```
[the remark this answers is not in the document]
adfineberg: I didn't want to change the way the files are saved. I'm not sure if the suiteapp fetch code knows how to fetch with or without the `FILE_CABINET_PATH_SEPARATOR` in the beginning of the path.
Does that make sense?

packages/netsuite-adapter/src/client/suiteapp_client/suiteapp_file_cabinet.ts
```diff
@@ -421,11 +425,21 @@ SuiteAppFileCabinetOperations => {
...
+      unfilteredFilesCustomizationWithoutContent,
       filesCustomizationsLinks,
     ] = _.partition(filesCustomizations, file => file.values.link === undefined)
 
+    const filesToSize = Object.fromEntries(unfilteredFilesCustomizationWithoutContent.map(
+      file => [fullPath(file), file.size]
+    ))
+    const filteredFilesCustomization = excludeLargeFolders(filesToSize, suiteAppClient.maxFileCabinetSize)
+    const filteredFilesSet = new Set(filteredFilesCustomization.listedPaths)
+    const filesCustomizationWithoutContent = unfilteredFilesCustomizationWithoutContent.filter(
```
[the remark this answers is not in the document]
adfineberg: once a folder is excluded, all files and folders under it are excluded... the `filteredFilesSet` has only the files outside of the excluded folder

packages/netsuite-adapter/src/client/file_cabinet_utils.ts
```diff
@@ -0,0 +1,120 @@
...
+type FileToSize = Record<string, number>
+
+type FolderSize = {
+  path: string
+  size: number
+  folders: FolderSize[]
+}
+
+type FolderSizeMap = Record<string, FolderSize>
+const sep = FILE_CABINET_PATH_SEPARATOR
```
dantal4, line 29: `import { FILE_CABINET_PATH_SEPARATOR as sep } from '../constants'` is nicer

packages/netsuite-adapter/src/client/file_cabinet_utils.ts
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    return [lastLargeFolder.path]
+  }
+
+  const filterMultipleFolders = (): string[] => {
+    const sortedSizes = orderBy(sizes, 'size', 'desc')
+    let selectedSize = 0
+    const selectedFolders = [] as string[]
+    while (selectedSize <= overflowSize) {
```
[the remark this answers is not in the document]
dantal4: yes, so the while condition should be `while (selectedSize < overflowSize)` , isn't it?

packages/netsuite-adapter/src/client/suiteapp_client/suiteapp_file_cabinet.ts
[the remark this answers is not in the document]
dantal4: But `foldersCustomizationInfo` isn't being filtered

packages/netsuite-adapter/src/client/file_cabinet_utils.ts
```diff
@@ -0,0 +1,122 @@
...
+type FolderSize = {
+    path: string
+    size: number
+    folders: FolderSize[]
+}
+
+type FolderSizeMap = { [path: string]: FolderSize }
+
+export const excludeLargeFolders = (files: FileToSize, maxFileCabinetSize: number):
+    { listedPaths: string[]; largeFolderError: string[] } => {
```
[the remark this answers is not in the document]
dantal4: If so it would be great to configure it IMO

packages/netsuite-adapter/src/client/file_cabinet_utils.ts
```diff
@@ -0,0 +1,122 @@
...
+  const folderSizes = (): FolderSize[] => {
+    const statFolders = (fileSizes: FileToSize): FolderSize[] => {
+      const createFlatFolderSizes = (): FolderSizeMap => {
+        const flatFolderSizes = {} as FolderSizeMap
+        Object.keys(fileSizes).forEach((path: string) => {
+          const pathParts = path.startsWith(osPath.sep)
```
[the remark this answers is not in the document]
dantal4: Didn't you build the full path from the path parts in the suiteapp case? You can add the `/` prefix there and then all places in `file_cabinet_utils.ts` & `config.ts` won't have to add it again.

packages/netsuite-adapter/src/client/file_cabinet_utils.ts
```diff
@@ -0,0 +1,122 @@
...
+            }
+            return nextPath
+          })
+        })
+        return flatFolderSizes
+      }
+      const createFolderHierarchy = (flatFolderSizes: FolderSizeMap): FolderSize[] => {
+        const folderGraph = [] as FolderSize[]
+        Object.keys(flatFolderSizes).forEach(folderName => {
+          if (folderName.indexOf(osPath.sep) === -1) { // Top level folder
```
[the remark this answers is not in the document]
dantal4: didn't you say [here](https://github.com/salto-io/salto/pull/4274/files/0cf61c31b6edb0c9eb2fa6afd3f6da85abd6356c#r1172475911) that there is the `/` in each file path?

packages/netsuite-adapter/src/client/file_cabinet_utils.ts
```diff
@@ -0,0 +1,122 @@
...
+            folderGraph.push(flatFolderSizes[folderName])
+          } else { // Sub folder
+            const parentFolder = osPath.dirname(folderName)
```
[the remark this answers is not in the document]
dantal4: I'm not sure:
```
> path.dirname('c:\\a\b\c')
'.'
```

packages/netsuite-adapter/src/client/file_cabinet_utils.ts
```diff
@@ -0,0 +1,122 @@
...
+      } else {
+        noLargeSubFolders = true
+      }
+    }
+
+    return [lastLargeFolder.path]
+  }
+
+  const filterMultipleFolders = (): string[] => {
+    const sortedSizes = orderBy(sizes, 'size', 'desc')
```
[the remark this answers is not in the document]
dantal4: Ok now I understand that "desc" is for "descending" and not "description"...
Yes I think that excluding the largest folder first is correct 👍

packages/netsuite-adapter/src/client/suiteapp_client/suiteapp_file_cabinet.ts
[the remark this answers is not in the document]
dantal4: I think that anywhere in `suiteapp_file_cabinet.ts` that takes the path parts and make a full path from it adds the `/` at the beginning.

packages/netsuite-adapter/src/client/file_cabinet_utils.ts
```diff
@@ -0,0 +1,120 @@
...
+const sep = FILE_CABINET_PATH_SEPARATOR
+const BYTES_IN_GB = 1024 ** 3
+
+export const excludeLargeFolders = (
+  files: FileToSize,
+  maxFileCabinetSize: number
+): { listedPaths: string[]; largeFolderError: string[] } => {
+  const createFlatFolderSizes = (fileSizes: FileToSize): FolderSizeMap => {
+    const flatFolderSizes = {} as FolderSizeMap
+    Object.keys(fileSizes).forEach(path => {
```
omrilit, line 38: super nit: you can use `Object.entries()`, as eventually you access `fileSizes` again in order to extract the value.

packages/netsuite-adapter/src/client/file_cabinet_utils.ts
```diff
@@ -0,0 +1,120 @@
...
+  folders: FolderSize[]
+}
+
+type FolderSizeMap = Record<string, FolderSize>
+const sep = FILE_CABINET_PATH_SEPARATOR
+const BYTES_IN_GB = 1024 ** 3
+
+export const excludeLargeFolders = (
+  files: FileToSize,
+  maxFileCabinetSize: number
```
omrilit, line 34: I think that it would be better to be more explicit with the var name here so we could easily understand the size unit here
```suggestion
  maxFileCabinetSizeInGB: number
```

packages/netsuite-adapter/src/config.ts
```diff
@@ -160,6 +164,12 @@ const clientConfigType = createMatchingObjectType<SdfClientConfig>({
     installedSuiteApps: {
       refType: new ListType(BuiltinTypes.STRING),
     },
+    maxFileCabinetSize: {
+      refType: BuiltinTypes.NUMBER,
+      annotations: {
+        [CORE_ANNOTATIONS.DEFAULT]: DEFAULT_MAX_FILE_CABINET_SIZE,
```
omrilit, line 170: IMO we should remove that row so this won't be written to the default adapter config file.

packages/netsuite-adapter/src/config.ts
```diff
@@ -160,6 +164,12 @@ const clientConfigType = createMatchingObjectType<SdfClientConfig>({
     installedSuiteApps: {
       refType: new ListType(BuiltinTypes.STRING),
     },
+    maxFileCabinetSize: {
```
omrilit, line 167: As said before, I think we should rename it so it would be more self explained
```suggestion
    maxFileCabinetSizeInGB: {
```

packages/netsuite-adapter/src/config.ts
```diff
@@ -179,6 +189,12 @@ const suiteAppClientConfigType = createMatchingObjectType<SuiteAppClientConfig>(
         }),
       },
     },
+    maxFileCabinetSize: {
```
omrilit, line 192: Why do we have this option both in clientConfigType and here?

packages/netsuite-adapter/src/config.ts
```diff
@@ -563,22 +597,24 @@ const toConfigSuggestions = (
 ): NetsuiteConfig => {
   const config: NetsuiteConfig = {}
 
-  if (!_.isEmpty(failedFilePaths.otherError) || !_.isEmpty(failedTypes.unexpectedError)) {
+  if (!_.isEmpty(failedFilePaths.otherError) || !_.isEmpty(failedTypes.unexpectedError)
+    || !_.isEmpty(failedFilePaths.largeFolderError)) {
```
omrilit, line 601: Do we have a place where we let the user know why we decided to exclude the folder?
I think that adding an indicative message to the eventLog would do the trick.

omrilit (commented):
Looks really good 👏 
Other than my comments in-code, I think that we should do some analysis to understand the current usage our customers have so when this code will be merged, we won't end up with excluding some of their configuration.
Maybe first adding code that just logs the size of the current used FileCabinet without filtering it would give us the data we need in the easiest way so we could have a data based decision here.

packages/netsuite-adapter/src/client/file_cabinet_utils.ts
```diff
@@ -0,0 +1,116 @@
...
+  log.info(`Excluding large folder(s) with total size of ${foldersToExclude.reduce((sum, folder) => sum + folder.size, 0)}`
+    + ` and name(s): ${foldersToExclude.map(folder => folder.path)}`)
+  return foldersToExclude.map(folder => `${sep}${folder.path}${sep}`)
+}
+
+export const filterFilesInFolders = (files: string[], folders: string[]): string[] =>
+  files.filter(file => !folders.some(folder => file.startsWith(folder)))
+
+export const filterFilePathsInFolders = <T extends { path: string[] }>(files: T[], folders: string[]): T[] =>
+  files.filter(file => !folders.some(folder => file.path.join(sep).startsWith(folder)))
```
dantal4, line 116: `folder` starts with `/` but `file.path.join(sep)` doesn't, right?
maybe add a test that combines a call to `largeFoldersToExclude` and then calls to `filterFilesInFolders`, `filterFilePathsInFolders`.

packages/netsuite-adapter/src/client/sdf_client.ts
```diff
@@ -795,6 +796,21 @@ export default class SdfClient {
       }
     }
 
+    const filesToSize = async (
+      filePaths: string[],
+      fileCabinetDirPath: string
+    ): Promise<{ path: string; size: number }[]> => {
```
dantal4, line 802: nit
```suggestion
    ): Promise<FileSize[]> => {
```

packages/netsuite-adapter/src/client/suiteapp_client/suiteapp_file_cabinet.ts
```diff
@@ -168,7 +170,7 @@ export const THROW_ON_MISSING_FEATURE_ERROR: Record<string, string> = {
 }
 
 export type SuiteAppFileCabinetOperations = {
-  importFileCabinet: (query: NetsuiteQuery) => Promise<ImportFileCabinetResult>
+  importFileCabinet: (query: NetsuiteQuery, maxFileCabinetSizeInGB?: number) => Promise<ImportFileCabinetResult>
```
dantal4, line 173: why is it an optional param?

packages/netsuite-adapter/src/client/suiteapp_client/suiteapp_file_cabinet.ts
```diff
@@ -349,9 +351,11 @@ SuiteAppFileCabinetOperations => {
       log.error('folder\'s parent is unknown\nfolder: %o\nidToFolder: %o', folder, idToFolder)
       throw new Error(`Failed to get absolute folder path of ${folder.name}`)
     }
-    return [...getFullPath(idToFolder[folder.parent], idToFolder), folder.name]
+    return [...fullPathParts(idToFolder[folder.parent], idToFolder), folder.name]
   }
 
+  const fullPath = (fileParts: string[]): string => fileParts.join(FILE_CABINET_PATH_SEPARATOR)
```
dantal4, line 357: it is always used with `/${fullPath(path)}` so maybe the `/` can be included in here too.

packages/netsuite-adapter/src/client/types.ts
```diff
@@ -54,6 +54,7 @@ export type GetCustomObjectsResult = {
 export type FailedFiles = {
   lockedError: NetsuiteFilePathsQueryParams
   otherError: NetsuiteFilePathsQueryParams
+  largeFolderError?: NetsuiteFilePathsQueryParams
```
dantal4, line 57: why is it optional?

packages/netsuite-adapter/src/config.ts
```diff
@@ -662,46 +677,65 @@ const updateConfigFromFailures = (
...
+    }
+
+    config.fetch = updatedFetchConfig
+    return true
   }
 
-  config.fetch = updatedFetchConfig
-  return true
+  const updateConfigFromLargeFolders = (largeFolderError: NetsuiteFilePathsQueryParams | undefined): boolean => {
+    if (largeFolderError && !_.isEmpty(largeFolderError)) {
```
dantal4, line 724: nit
```suggestion
  const updateConfigFromLargeFolders = (): boolean => {
    const { largeFolderError } = failedFilePaths
    if (largeFolderError && !_.isEmpty(largeFolderError)) {
```

dantal4 (commented):
Great work!

packages/netsuite-adapter/src/config.ts
dantal4: The `client` field in `configType` has no default so it won't be written to the nacl:
```
client: {
      refType: clientConfigType,
    },
```

dantal4 (approved):
💯 
Please fix [this comment](https://github.com/salto-io/salto/pull/4274#discussion_r1174997639) too (I see that you added a test for `filterFilesInFolders` but not `filterFilePathsInFolders`).

packages/netsuite-adapter/src/client/suiteapp_client/suiteapp_file_cabinet.ts
```diff
@@ -421,11 +426,29 @@ SuiteAppFileCabinetOperations => {
...
+    const foldersCustomizationInfo = unfilteredFoldersCustomizationInfo
+    // const filesCustomizationWithoutContent = filterFilePathsInFolders(
+    //   unfilteredFilesCustomizationWithoutContent,
+    //   largeFolders
+    // )
+    // const foldersCustomizationInfo = filterFilePathsInFolders(
+    //   unfilteredFoldersCustomizationInfo,
+    //   largeFolders
+    // )
+
```
kami-delaroz, line 451: @adfineberg I believe this was left by mistake?

PR #4276, opened 2023-04-19
SALTO-4029 correctly support async lookup functions

netama:
Fix for a theoretical bug - we were assuming all lookup functions were sync (which they currently are, but are not required to be)

---
_Release Notes_: 
None

---
_User Notifications_: 
None


shir-reifenberg (approved):
🚀 🙏

PR #4277, opened 2023-04-20
SALTO-3805: Profile attribute references

shir-reifenberg:
Add references from `ProfileEnrollmentPolicyRule` to `UserSchema`

---

`ProfileEnrollmentPolicyRule` have a list of attributes required from the user in order to login. those attributes are taken from the default user `UserSchema`. Also, user can provide attributes that doesn't exist in the schema, Okta will allow it but Okta's UI will show an error, so in this case we add a missing reference.

This PR also fixes a bug in references to custom attributes in expression_language.ts filter 

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/config.ts
+++ b/packages/okta-adapter/src/config.ts
@@ -1076,6 +1076,11 @@ const DEFAULT_TYPE_CUSTOMIZATIONS: OktaSwaggerApiConfig['types'] = {
       },
     },
   },
+  ProfileEnrollmentPolicyRuleProfileAttribute: {
+    transformation: {
+      fieldTypeOverrides: [{ fieldName: 'name', fieldType: 'unknown' }],
+    },
+  },
 }
 
 const DEFAULT_SWAGGER_CONFIG: OktaSwaggerApiConfig['swagger'] = {
--- a/packages/okta-adapter/src/config.ts
+++ b/packages/okta-adapter/src/config.ts
@@ -1078,7 +1078,7 @@ const DEFAULT_TYPE_CUSTOMIZATIONS: OktaSwaggerApiConfig['types'] = {
   },
   ProfileEnrollmentPolicyRuleProfileAttribute: {
     transformation: {
-      fieldTypeOverrides: [{ fieldName: 'name', fieldType: 'unknown' }],
+      fieldTypeOverrides: [{ fieldName: 'name', fieldType: 'UserSchemaAttribute' }],
     },
   },
 }
--- a/packages/okta-adapter/src/filters/expression_language.ts
+++ b/packages/okta-adapter/src/filters/expression_language.ts
@@ -84,10 +84,7 @@ export const resolveUserSchemaRef = (ref: ReferenceExpression): string | undefin
     (topLevelParentId.createNestedID(...USER_SCHEMA_CUSTOM_PATH).isEqual(parentId))
       || (topLevelParentId.createNestedID(...USER_SCHEMA_BASE_PATH).isEqual(parentId))
   ) {
-    const userSchemaField = ref.elemID.getFullNameParts().pop()
-    if (_.isString(userSchemaField)) {
-      return userSchemaField
-    }
+    return ref.elemID.name
   }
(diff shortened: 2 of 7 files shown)


packages/okta-adapter/src/config.ts
```diff
@@ -1076,6 +1076,11 @@ const DEFAULT_TYPE_CUSTOMIZATIONS: OktaSwaggerApiConfig['types'] = {
       },
     },
   },
+  ProfileEnrollmentPolicyRuleProfileAttribute: {
+    transformation: {
+      fieldTypeOverrides: [{ fieldName: 'name', fieldType: 'unknown' }],
```
netama, line 1081: why is the type unknown, if we only convert it to a reference? or is the goal to ensure it always exists?

packages/okta-adapter/src/filters/expression_language.ts
```diff
@@ -77,19 +77,27 @@ const getUserSchemaReference = (
...
 }
 
+export const resolveUserSchemaRef = (ref: ReferenceExpression): string | undefined => {
+  const topLevelParentId = ref.elemID.createTopLevelParentID().parent
+  const parentId = ref.elemID.createParentID()
+  if (
+    (topLevelParentId.createNestedID(...USER_SCHEMA_CUSTOM_PATH).isEqual(parentId))
+      || (topLevelParentId.createNestedID(...USER_SCHEMA_BASE_PATH).isEqual(parentId))
+  ) {
+    const userSchemaField = ref.elemID.getFullNameParts().pop()
```
netama, line 87: can use `ref.elemID.name`

packages/okta-adapter/src/filters/expression_language.ts
```diff
@@ -77,19 +77,27 @@ const getUserSchemaReference = (
...
+    if (_.isString(userSchemaField)) {
```
netama, line 88: not sure I understand what this is checking - this is a part of the elem id so it will always be a string, no? maybe we should add a test that gets to the other case to demonstrate?

packages/okta-adapter/src/filters/profile_enrollment_attributes.ts
```diff
@@ -0,0 +1,71 @@
...
+    profileEnrollmentRules.forEach(rule => {
+      const profileAttributes = resolvePath(rule, rule.elemID.createNestedID(...PROFILE_ATTRIBUTES_PATH))
+      if (!Array.isArray(profileAttributes)) {
+        log.warn(`Can not create references from instance ${rule.elemID.getFullName()} to user schema attributes`)
+        return
+      }
+      profileAttributes.forEach(att => {
+        const { name } = att
+        if (!_.isString(name)) {
+          return
```
netama, line 56: is this an expected scenario? or should we log?

packages/okta-adapter/src/filters/profile_enrollment_attributes.ts
```diff
@@ -0,0 +1,71 @@
...
+        }
+        const userSchemaRef = getUserSchemaReference(name, defaultUserSchema)
+        if (userSchemaRef !== undefined) {
+          att.name = userSchemaRef
+        } else {
+          // create missing reference
+          const missingInstance = createMissingInstance(OKTA, USER_SCHEMA_TYPE_NAME, name)
```
netama, line 63: not sure I understand this - if `getUserSchemaReference` returns a nested path, shouldn't we create a similar missing reference with a path? why the difference in the formats?
(is this also why we couldn't use a rule for this?)

netama (commented):
I might be missing something but from what I understand the "missing" instance is not aligned with the "real" references so I'm not sure we should use it this way (especially if we're soon going to switch to a custom objects model and then resolve this in a different way). might be easier to discuss IRL, there's a good chance I'm missing something in the bigger picture :)

packages/okta-adapter/src/filters/expression_language.ts
shir-reifenberg: removed

packages/okta-adapter/src/config.ts
shir-reifenberg: I got type mismatch for the field so I thought this was the solution, I'll try again

netama (approved):

PR #4278, opened 2023-04-20
SALTO-4012: alias index to include only elements with alias defined

edenhassid:
alias index to include only elements with alias defined

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/workspace/src/workspace/alias_index.ts
+++ b/packages/workspace/src/workspace/alias_index.ts
@@ -39,10 +39,11 @@ const getAlias = (element: Element): string | undefined => element.annotations[C
 
 const calculateAliasChange = (
   afterElement: Element,
-  aliasBefore: string | undefined,
-  aliasAfter: string | undefined
+  beforeElement: Element,
 )
   : Change<Element>[] => {
+  const aliasBefore = getAlias(beforeElement)
+  const aliasAfter = getAlias(afterElement)
   if (aliasBefore !== aliasAfter) {
     if (aliasAfter !== undefined) {
       return [toChange({ after: afterElement })]
@@ -66,7 +67,7 @@ const getChangedFields = (change: ModificationChange<ObjectType>): Change<Elemen
     .flatMap(afterField => {
       const field = change.data.before.fields[afterField.name]
       return isField(field)
-        ? calculateAliasChange(afterField, getAlias(field), getAlias(afterField))
+        ? calculateAliasChange(afterField, field)
         : []
     })
   return fieldAliasChanges
@@ -81,16 +82,14 @@ const getAllRelevantChanges = (changes: Change<Element>[]): Change<Element>[] =>
         const changedFields = getChangedFields(change)
         const objAliasModification = calculateAliasChange(
           change.data.after,
-          getAlias(change.data.before),
-          getAlias(change.data.after)
+          change.data.before,
         )
         return changedFields.concat(objAliasModification)
       }
       if (isInstanceChange(change)) {
         const instAliasModification = calculateAliasChange(
           change.data.after,
-          getAlias(change.data.before),
-          getAlias(change.data.after)
+          change.data.before,
         )
         return _.isEmpty(instAliasModification) ? undefined : instAliasModification
       }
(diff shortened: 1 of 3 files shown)


packages/workspace/src/workspace/alias_index.ts
```diff
@@ -25,35 +25,54 @@ import {
...
-    const field = change.data.before.fields[afterField.name]
-    return isField(field) && (getAlias(afterField) !== getAlias(field))
-  })
-  return (addedFields.map(field => toChange({ after: field })))
-    .concat(aliasModifiedFields.map(field => toChange({ after: field })))
+  const addedFields = afterFields.filter(afterField =>
+    !beforeFieldsNamesSet.has(afterField.name) && getAlias(afterField) !== undefined)
+  const removedFields = beforeFields.filter(beforeField =>
+    !afterFieldsNamesSet.has(beforeField.name) && getAlias(beforeField) !== undefined)
+  const aliasModifiedFields = afterFields
```
netama, line 67: naming nit: this contains changes, not fields anymore, right? maybe `fieldAliasChanges`?

packages/workspace/src/workspace/alias_index.ts
```diff
@@ -62,26 +81,36 @@ const getAllRelevantChanges = (changes: Change<Element>[]): Change<Element>[] =>
...
+      // take only the fields with alias
+      const addedAliasFields: Element[] = Object.values(change.data.after.fields).filter(getAlias)
+      return (addedAliasFields.map(f => toChange({ after: f })) ?? [])
+        .concat(getAlias(getChangeData(change)) ? [change] : [])
+    }
+    if (isAdditionChange(change) && isInstanceChange(change)) {
+      return getAlias(getChangeData(change)) ? change : undefined
     }
-    return change
+    return change // should only be removals
```
netama, line 113: nit: this is not really guaranteed in the current condition structure, even if it true today it might change in the future unexpectedly - I think it'd be safer to either add a check, or convert the `isAdditionChange` part to something with a fallback like we hvae for the modifications, wdyt?
```
if (isAdditionChange) {
  if (...) {
  }
  if (...) {
  }
  return undefined
}
```

packages/workspace/src/workspace/alias_index.ts
```diff
@@ -94,11 +123,16 @@ const getRemovalsAndAdditions = (changes: Change<Element>[]):
...
 
 const getInfoForIndex = (changes: Change<Element>[]):
-  { removalNames:string[]; additionsIdsToAlias: Record<string, string> } => {
+  { removalNames: string[]; additionsIdsToAlias: Record<string, string> } => {
   const { additions, removals } = getRemovalsAndAdditions(changes)
-  const additionsIdsToAlias = _.mapValues(
-    _.keyBy(additions.map(getChangeData), elem => elem.elemID.getFullName()),
-    elem => getAlias(elem)
+  const filteredAdditions = additions.map(getChangeData)
+    .filter(elem => elem.annotations[CORE_ANNOTATIONS.ALIAS] !== undefined)
```
netama, line 129: why not still use `getAlias`?

packages/workspace/src/workspace/alias_index.ts
```diff
@@ -94,11 +123,16 @@ const getRemovalsAndAdditions = (changes: Change<Element>[]):
 }
 
 const getInfoForIndex = (changes: Change<Element>[]):
-  { removalNames:string[]; additionsIdsToAlias: Record<string, string> } => {
+  { removalNames: string[]; additionsIdsToAlias: Record<string, string> } => {
   const { additions, removals } = getRemovalsAndAdditions(changes)
```
netama, line 127: this is existing code so I probably missed it last time, but why are we calling these additions when they also include "modifications" (even if they are converted into addition changes)?

netama (commented):
looks good, added small questions

packages/workspace/src/workspace/alias_index.ts
edenhassid: discussed IRL

netama: discussed IRL again :) for clarity let's call the `additions` parts `overrides` (instead of changing everything from changes to something else)

netama (approved):
pre-approving per IRL discussion

PR #4281, opened 2023-04-20
Salto-3868: reorder deploy xml only for elements not in cycle

Shulik95:
_do not list explicitly in the deploy.xml elements that have a circular dependeny_

---

_None_

---
_Release Notes_: 
_Netsuite Adapter_:
* Elements that have circular dependency will not be listed explicitly in deploy.xml to avoid circular dependency error

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/netsuite-adapter/src/client/graph_utils.ts
+++ b/packages/netsuite-adapter/src/client/graph_utils.ts
@@ -69,10 +69,10 @@ export class Graph<T> {
   private dfs(dfsParams: DFSParameters<T>): void {
     const { node, visited, resultArray, path, cycle } = dfsParams
     if (visited.has(node.value[this.key])) {
-      if (path?.includes(node)) {
+      const cycleStartIndex = path?.indexOf(node)
+      if (cycleStartIndex !== -1) {
         // node is visited & in path mean its a cycle
-        const cycleStartIndex = path.indexOf(node)
-        cycle?.push(...path.slice(cycleStartIndex))
+        cycle?.push(...(path?.slice(cycleStartIndex) ?? []))
       }
       return
     }
--- a/packages/netsuite-adapter/src/client/graph_utils.ts
+++ b/packages/netsuite-adapter/src/client/graph_utils.ts
@@ -39,6 +39,15 @@ export class GraphNode<T> {
   }
 }
 
+type DFSParameters<T>= {
+  node: GraphNode<T>
+  visited: Set<T[keyof T]>
+  resultArray: GraphNode<T>[]
+  // optional parameters for cycle detection
+  stack?: GraphNode<T>[]
+  cycle?: GraphNode<T>[]
+}
+
 export class Graph<T> {
   nodes: Map<T[keyof T], GraphNode<T>>
   key: keyof T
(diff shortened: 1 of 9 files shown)


packages/netsuite-adapter/src/client/graph_utils.ts
```diff
@@ -57,22 +66,31 @@ export class Graph<T> {
...
+        // node is visited & in path mean its a cycle
+        cycle?.push(...(path?.slice(cycleStartIndex) ?? []))
+      }
       return
     }
     visited.add(node.value[this.key])
+    path?.push(node)
     node.edges.forEach(dependency => {
-      this.dfs(dependency, visited, resultArray)
+      this.dfs({ node: dependency, visited, resultArray, path, cycle })
```
dantal4, line 82: I think that this way `path` can be used inside this function only, and `path.push`, `path.pop` can be removed
```suggestion
      this.dfs({ node: dependency, visited, resultArray, path: path.concat(node), cycle })
```

dantal4 (commented):
Nice work!

packages/netsuite-adapter/src/client/graph_utils.ts
```diff
@@ -57,22 +66,31 @@ export class Graph<T> {
...
   }
 
-  private dfs(node: GraphNode<T>, visited: Set<T[keyof T]>, resultArray: GraphNode<T>[]): void {
+  private dfs(dfsParams: DFSParameters<T>): void {
+    const { node, visited, resultArray, path, cycle } = dfsParams
     if (visited.has(node.value[this.key])) {
+      const cycleStartIndex = path?.indexOf(node)
+      if (cycleStartIndex !== -1) {
+        // node is visited & in path mean its a cycle
+        cycle?.push(...(path?.slice(cycleStartIndex) ?? []))
```
dantal4, line 75: don't we need to add all `path` items in `cycle`?

Shulik95: No, since the path includes all nodes we have visited but the cycle might start in the middle of the path and only include a part of it

packages/netsuite-adapter/test/client/graph_utils.test.ts
```diff
@@ -57,10 +57,19 @@ describe('graph utils tests', () => {
...
+    expect(cycleNodes).toEqual([testNode1, testNode2])
+  })
 
   it('should remove node from graph and edges from nodes', async () => {
     testGraph.removeNode('node1')
     expect(testGraph.nodes.get('node1')).toBeUndefined()
     expect(testNode2.edges).not.toContain(testNode1)
   })
+  it('should return an empty array if there is no cycle', async () => {
+    const cycleNodes = testGraph.findCycle()
```
dantal4, line 72: Did we get a different result here because of the "remove node" test? If it is the case, we should initiate the graph and the nodes inside a `beforeEach` section, because changing the state of the graph in one test shouldn't effect other tests - and also it is confusing that two tests that run the same function have different result.

dantal4 (commented):
Nice work!

packages/netsuite-adapter/src/client/graph_utils.ts
```diff
@@ -108,4 +126,19 @@ export class Graph<T> {
...
+    const nodesInCycle: T[keyof T][] = []
+
+    Array.from(this.nodes.values()).forEach(node => {
+      if (!visited.has(node.value[this.key])) {
+        this.dfs({ node, visited, path, cycle: nodesInCycle })
+      }
+    })
+    return nodesInCycle
+      .map(nodeKey => this.findNodeByKey(nodeKey))
+      .filter(isDefined)
```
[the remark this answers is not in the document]
Shulik95: `nodesInCycle` might have the same node more than once In case the node is contained in more than one cycle, but in `reorderDeployXml` before filtering the nodes out, `nodesInCycle` is changed to a set so any duplicates will be removed.

dantal4 (approved):
Great work!

PR #4282, opened 2023-04-20
SALTO-3987: Converts BrandTheme favicon url to TemplateExpression

shir-reifenberg:
Converts BrandTheme favicon url to TemplateExpression
Added fetch only cause `BrandTheme` deploy is not supported yet

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/filters/template_urls.ts
+++ b/packages/okta-adapter/src/filters/template_urls.ts
@@ -38,7 +38,7 @@ const filter: FilterCreator = () => ({
     if (!_.isString(subdomain)) {
       log.warn('Could not find subdomain, skipping templateUrlsFilter')
     }
-    const subdomainRegString = `://${subdomain}`
+    const subdomainRegString = `://${subdomain}.`
     const subdomainReg = new RegExp(`(${subdomainRegString})`)
 
     const brandThemeInstance = instances.find(instance => instance.elemID.typeName === 'BrandTheme')
@@ -52,7 +52,7 @@ const filter: FilterCreator = () => ({
       [subdomainReg],
       expression => {
         if (expression === `${subdomainRegString}`) {
-          return ['://', new ReferenceExpression(orgSettingsInstance.elemID.createNestedID('subdomain'), subdomain)]
+          return ['://', new ReferenceExpression(orgSettingsInstance.elemID.createNestedID('subdomain'), subdomain), '.']
         }
         return expression
       },
--- a/packages/okta-adapter/src/filters/template_urls.ts
+++ b/packages/okta-adapter/src/filters/template_urls.ts
@@ -1,0 +1,60 @@
+/*
+*                      Copyright 2023 Salto Labs Ltd.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


packages/okta-adapter/src/filters/template_urls.ts
```diff
@@ -0,0 +1,60 @@
...
+const filter: FilterCreator = () => ({
+  name: 'templateUrlsFilter',
+  onFetch: async (elements: Element[]) => {
+    const instances = elements.filter(isInstanceElement)
+    const orgSettingsInstance = instances.find(instance => instance.elemID.typeName === ORG_SETTING_TYPE_NAME)
+    if (orgSettingsInstance === undefined) {
+      log.warn(`Could not find ${ORG_SETTING_TYPE_NAME} instance, skipping templateUrlsFilter`)
+      return
+    }
+    const subdomain = orgSettingsInstance?.value?.subdomain
```
netama, line 37: 1. if it can be undefined, let's add an early return for it?
2. nit - no need for the question marks, since you already checked that it is defined (and if it is an instance it should have a `value`)

packages/okta-adapter/src/filters/template_urls.ts
```diff
@@ -0,0 +1,60 @@
...
+    if (brandThemeInstance === undefined || !_.isString(faviconUrl)) {
+      log.warn('Could not change favicon field in BrandTheme with TemplateExpression')
+      return
+    }
+    const template = extractTemplate(
+      faviconUrl,
+      [subdomainReg],
+      expression => {
+        if (expression === subdomain) {
+          return new ReferenceExpression(orgSettingsInstance.elemID.createNestedID('subdomain'), subdomain)
```
netama, line 51: this is just from reading so maybe I'm wrong here, but IIUC if we had a subdomain called `bla` we would replace all occurrences of `bla` in the text - e.g. `https://something.com/blame` would become `https://.com/${ okta.<...> }me`. I think we should be more careful here, and look for the full pattern (`https://<>.`). wdyt?

packages/okta-adapter/src/filters/template_urls.ts
```diff
@@ -0,0 +1,64 @@
...
+    const orgSettingsInstance = instances.find(instance => instance.elemID.typeName === ORG_SETTING_TYPE_NAME)
+    if (orgSettingsInstance === undefined) {
+      log.warn(`Could not find ${ORG_SETTING_TYPE_NAME} instance, skipping templateUrlsFilter`)
+      return
+    }
+    const { subdomain } = orgSettingsInstance.value
+    if (!_.isString(subdomain)) {
+      log.warn('Could not find subdomain, skipping templateUrlsFilter')
+    }
+    const subdomainRegString = `://${subdomain}`
```
netama, line 41: nit: maybe also `.` at the end?

packages/okta-adapter/test/filters/template_urls.test.ts
```diff
@@ -0,0 +1,66 @@
...
+          '.oktapreview.com/favicon.ico',
+        ],
+      })
+
+      beforeEach(() => {
+        filter = templateUrlsFilter(getFilterParams()) as typeof filter
+      })
+
+      describe('onFetch', () => {
+        it('should resolve templates in urls', async () => {
```
netama, line 52: maybe also add a test for the case that the subdomain itself exists but not as part of a URL, and ensure it doesn't get converted?

netama (approved):
LGTM with a couple last nits

packages/okta-adapter/test/filters/template_urls.test.ts
shir-reifenberg: added

PR #4283, opened 2023-04-20
SALTO-1730 - Formula parser code review

tomsellek:
---

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A


packages/salesforce-adapter/src/filters/formula_deps.ts
```diff
@@ -76,14 +73,14 @@ const referencesFromIdentifiers = async (typeInfos: FormulaIdentifierInfo[]): Pr
...
+      const typeElement = await allElements.get(typeElemId)
+      return (typeElement !== undefined) && (typeElement.fields[elemId.name] !== undefined)
     }
 
-    // field
-    const typeElemId = new ElemID(elemId.adapter, elemId.typeName)
-    const typeElement = await allElements.get(typeElemId)
-    return (typeElement !== undefined) && (typeElement.fields[elemId.name] !== undefined)
+    // 'type' or 'instance'
+    return await allElements.get(elemId) !== undefined
```
tomsellek, line 83: > In the previous code there was an assumption that any id which is not `type` or `instance` is a field but that is not true, it is not safe to make that assumption

How so? We define the typeId in `identifierTypeToElemIdType`, no?

packages/salesforce-adapter/src/filters/formula_deps.ts
```diff
@@ -103,49 +100,45 @@ const addDependenciesAnnotation = async (field: Field, allElements: ReadOnlyElem
...
     const formulaIdentifiers: string[] = log.time(
-      () => (extract(formula)),
+      () => (extractFormulaIdentifiers(formula)),
       `Parse formula '${formula.slice(0, 15)}'`
     )
 
-    const identifiersInfo = await log.time(
+    const identifiersInfo = (await log.time(
       () => Promise.all(
-        formulaIdentifiers.map(async identifier => parseFormulaIdentifier(identifier, field.parent.elemID.typeName))
```
tomsellek, line 132: >Using the typename here is not technically correct, we should be using the salesforce name this should work in a similar way to field_references, creating an index of elements by API name it may be possible to share some of the code?
this will also make some of the code below redundant

I think I'm missing something here - we don't do any lookup, do we? We straight up iterate over things. What code do you see being made redundant?

packages/salesforce-adapter/src/filters/formula_utils/grammar.ts
```diff
@@ -18,13 +18,13 @@ import { getField, getObject, parts } from './utils'
...
 
 const RELATIONSHIP_SUFFIX = '__R'
-const USER_FIELDS_REGEX = new RegExp(/^OWNER|MANAGER|CREATEDBY|LASTMODIFIEDBY$/, 'i')
-const CUSTOM_LABEL_PREFIX_REGEX = new RegExp(/^\$LABEL\./, 'i')
-const CUSTOM_SETTING_PREFIX_REGEX = new RegExp(/^\$SETUP\./, 'i')
-const OBJECT_TYPE_PREFIX_REGEX = new RegExp(/^\$OBJECTTYPE\./, 'i')
-const SELF_REFERENTIAL_PARENT_FIELD_REGEX = new RegExp(/^parentid$/, 'i')
-const SELF_REFERENTIAL_PARENT_OBJECT_REGEX = new RegExp(/^parent$/, 'i')
-const SPECIAL_PREFIXES_REGEX = new RegExp(/^\$USER|\$PROFILE|\$ORGANIZATION|\$USERROLE|\$SYSTEM/, 'i')
+const USER_FIELDS_REGEX = /^OWNER|MANAGER|CREATEDBY|LASTMODIFIEDBY$/i
```
tomsellek, line 21: > Where did this list of names come from?

Pablo :)

packages/salesforce-adapter/src/filters/formula_utils/grammar.ts
```diff
@@ -42,6 +42,11 @@ export const isParent = (value: string): boolean => SELF_REFERENTIAL_PARENT_OBJE
 export const isStandardRelationship = (value: string): boolean => (
   !value.toLocaleUpperCase().endsWith(RELATIONSHIP_SUFFIX)
 )
+
+// ORI: Seems like there are a lot of different places that mix the meaning of "." as a separator
+// this feels unsafe as we assign different meanings implicitly, this function for example
+// isn't really correct in most contexts, there are many strings where having a "." does not mean
+// it is a relationship field
```
tomsellek, line 49: Agreed. It only works in the context of `parseFormulaIdentifier` where there are a bunch of other checks before we call it. I also think this code treats any reference to a field from another object as a "relationship", parsing-wise.

packages/salesforce-adapter/src/filters/formula_utils/parse.ts
```diff
@@ -152,19 +160,22 @@ const parseRelationship = (variableName: string, originalObject: string): Formul
...
         fieldName = transformToId(fieldName)
       } else {
+        // Why can we assume that? doesn't seem like anything checks that this is the case
         // We assume the field ends with '_r'
         fieldName = fieldName.slice(0, -1).concat('c')
       }
     }
-
+    // This is unexpected, why do we have CPQ specific code here?
+    // How can CPQ do things that are not standard in salesforce?
```
tomsellek, line 169: I do not know. Presumably Pablo does.

packages/salesforce-adapter/src/filters/formula_utils/parse.ts
```diff
@@ -152,19 +160,22 @@ const parseRelationship = (variableName: string, originalObject: string): Formul
...
     if (isCPQRelationship(fieldName)) {
       fieldName = mapCPQField(fieldName, originalObject)
     }
 
     if (isUserField(fieldName)) {
       fieldName = transformToUserField(fieldName)
     }
-
+    // I am completely lost trying to understand what this is trying to achieve
+    // any comment would help, I have to believe there is a simpler way to do whatever this is doing
```
tomsellek, line 178: It essentially does a recursive semantic parse of a string of the form "A.B.C.D". It has to keep track of the "current parent" of whatever it is it's currently parsing - so A's parent object is whatever you pass as a parameter, B's parent is whatever you parsed A to be, etc...
It could probably be rewritten as a recursive function.

packages/salesforce-adapter/src/filters/formula_deps.ts
ori-moisis: this is a function that gets `ElemID`, how is that related to `identifierTypeToElemIdType`?

also, no need to read the content of the element just to check if it exists
```suggestion
    // 'type' or 'instance'
    return allElements.has(elemId)
```

packages/salesforce-adapter/src/filters/formula_deps.ts
```diff
@@ -103,49 +100,45 @@ const addDependenciesAnnotation = async (field: Field, allElements: ReadOnlyElem
     identifiersInfo: FormulaIdentifierInfo[][]
   ): void => {
     if (invalidReferences.length > 0) {
-      log.error('When parsing the formula %o in field %o, one or more of the identifiers %o was parsed to an invalid reference: ',
+      log.info('When parsing the formula %o in field %o, one or more of the identifiers %o was parsed to an invalid reference: ',
```
ori-moisis, line 103:
```suggestion
      log.info('When parsing the formula %s in field %s, one or more of the identifiers %o was parsed to an invalid reference: ',
```

PR #4284, opened 2023-04-21
Tom sellek.salto 3820.refactor remote maps location cache pool v2

tomsellek:
Replace the LRU-of-LRUs with a refcounted pool for location caches

---

This is the first step in our journey towards proper resource management in the RemoteMap implementation.
This change makes the locationCache pool no longer be LRU. If we determine this is a problem we can modify the implementation as needed.

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A


packages/core/src/local-workspace/remote_map/counters.ts
```diff
@@ -0,0 +1,86 @@
...
+type CounterType = typeof COUNTER_TYPES[number]
+
+type Counter = {
+  inc: () => void
+  value: () => number
+}
+
+export type LocationCounters = Record<CounterType, Counter> & {
+  dump: () => void
+}
```
ori-moisis, line 41: would the code be simpler if this was
```typescript
{
  dump: () => void
  counters: Record<CounterType, Counter>
}
```
?

packages/core/src/local-workspace/remote_map/counters.ts
```diff
@@ -0,0 +1,86 @@
...
+    ['dump', () => {
+      log.debug('Remote Map Stats for location \'%s\': %o',
+        location, Object.fromEntries(COUNTER_TYPES.map(counterType => [counterType, counters[counterType]])))
+    }],
+  ])
+  return counters
+}
+
+const createStatCounters = (): StatCounters => {
+  const locations: Record<string, { refCnt: number; counters: LocationCounters }> = {}
```
ori-moisis, line 68: I think we should prefer using `Map` and not `{}` where possible
it is silly, but if I ever request a counter with location `toString`, this would break in strange ways

packages/core/src/local-workspace/remote_map/location_cache.ts
```diff
@@ -0,0 +1,84 @@
...
+      if (poolEntry === undefined || poolEntry.refcnt === 0) {
+        log.warn('Returning a locationCache for an unknown location %s. poolEntry=%o', location, poolEntry)
+        return
+      }
+      poolEntry.refcnt -= 1
+
+      if (poolEntry.refcnt === 0) {
+        pool.delete(location)
+        if (pool.size === 0) {
+          log.debug('Max location cache pool size: %o', poolSizeWatermark)
```
ori-moisis, line 76: we may want to have this log not only in this scenario, we might never reach this...
I think it would be better to have the log when we increase the pool size beyond this mark instead (or in addition)

packages/core/src/local-workspace/remote_map/location_pool.ts
```diff
@@ -0,0 +1,44 @@
...
+const createRemoteMapLocationPool = (): RemoteMapLocationPool => (
+  {
+    get: location => ({
+      name: location,
+      counters: counters.get(location),
+      cache: locationCaches.get(location),
+    }),
+    return: location => {
+      locationCaches.release(location.cache)
+      counters.return(location.name)
```
ori-moisis, line 39: strange that one is `release` and the other is `return`

packages/core/src/local-workspace/remote_map/counters.ts
```diff
@@ -0,0 +1,86 @@
...
+      if (!(location in locations)) {
+        locations[location] = { refCnt: 0, counters: createLocationCounters(location) }
+      }
+      locations[location].refCnt += 1
+      return locations[location].counters
+    },
+    return: location => {
+      locations[location].refCnt -= 1
+      if (locations[location].refCnt === 0) {
+        delete locations[location]
```
ori-moisis, line 80: should the call to `dump` be here maybe?

packages/core/src/local-workspace/remote_map/remote_map.ts
```diff
@@ -443,36 +427,11 @@ ReadIterator => {
...
-    counterInc('LocationCacheCreated')
-    /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
-    locationCache = new LRU<string, any>({ max: cacheSize })
-    if (cacheSize > 0) {
-      locationCaches.set(location, locationCache)
-    }
-  }
+  const locationResources = remoteMapLocations.get(location)
+  const statCounters = locationResources.counters
+  const locationCache = locationResources.cache
```
ori-moisis, line 434:
```suggestion
  const { counters: statCounters, cache: locationCache } = remoteMapLocations.get(location)
```
just to avoid having that extra name...

ori-moisis (approved):

packages/core/src/local-workspace/remote_map/counters.ts
tomsellek: I'm not sure. It would certainly simplify the type definition a little, but I'm not seeing what usage it would simplify?

packages/core/src/local-workspace/remote_map/counters.ts
tomsellek: Ack. Though it does cost us a bit of syntactic sugar.

packages/core/src/local-workspace/remote_map/location_pool.ts
ori-moisis: I would have probably gone with `release` just because `return` is a keyword, but this is not a strong opinion

packages/core/src/local-workspace/remote_map/counters.ts
```diff
@@ -0,0 +1,88 @@
...
+  return {
+    get: location => {
+      locations.get(location).refCnt += 1
+      return locations.get(location).counters
+    },
+    return: location => {
+      locations.get(location).refCnt -= 1
+      if (locations.get(location).refCnt === 0) {
+        locations.delete(location)
+      }
```
ori-moisis, line 83: this could create a location with ref count of -1

```suggestion
      if (!locations.has(location)) {
        log.warn(...)
        return
      }
      counters = locations.get(location)
      counters.refCnt -= 1
      if (counters.refCnt === 0) {
        locations.delete(location)
      }
```

ori-moisis (approved):

PR #4285, opened 2023-04-23
SALTO-4016: Add optional prefix for config type name

shir-reifenberg:
Add optional prefix for config type name

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/config/ducktype.ts
+++ b/packages/adapter-components/src/config/ducktype.ts
@@ -43,22 +43,29 @@ export const createDucktypeAdapterApiConfigType = ({
   additionalFields,
   additionalRequestFields,
   additionalTransformationFields,
+  additionalIdPrefix,
 }: {
   adapter: string
   additionalFields?: Record<string, FieldDefinition>
   additionalRequestFields?: Record<string, FieldDefinition>
   additionalTransformationFields?: Record<string, FieldDefinition>
+  additionalIdPrefix?: string
 }): ObjectType => {
-  const transformationTypes = createTransformationConfigTypes(adapter, {
-    hasDynamicFields: { refType: BuiltinTypes.BOOLEAN },
-    sourceTypeName: { refType: BuiltinTypes.STRING },
-    ...additionalTransformationFields,
-  })
+  const transformationTypes = createTransformationConfigTypes(
+    adapter,
+    {
+      hasDynamicFields: { refType: BuiltinTypes.BOOLEAN },
+      sourceTypeName: { refType: BuiltinTypes.STRING },
+      ...additionalTransformationFields,
+    },
+    additionalIdPrefix
+  )
   return createAdapterApiConfigType({
     adapter,
     additionalRequestFields,
     transformationTypes,
     additionalFields,
+    additionalIdPrefix,
   })
 }
 
--- a/packages/adapter-components/src/config/ducktype.ts
+++ b/packages/adapter-components/src/config/ducktype.ts
@@ -43,29 +43,29 @@ export const createDucktypeAdapterApiConfigType = ({
   additionalFields,
   additionalRequestFields,
   additionalTransformationFields,
-  additionalIdPrefix,
+  elemIdPrefix = '',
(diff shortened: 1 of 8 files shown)


packages/adapter-components/src/config/ducktype.ts
```diff
@@ -43,22 +43,29 @@ export const createDucktypeAdapterApiConfigType = ({
   additionalFields,
   additionalRequestFields,
   additionalTransformationFields,
+  additionalIdPrefix,
```
netama, line 46: nit: maybe `elemIdPrefix`? (`additional` makes it sound like this only applies to some of the configuration)

packages/adapter-components/src/config/ducktype.ts
```diff
@@ -43,22 +43,29 @@ export const createDucktypeAdapterApiConfigType = ({
...
   additionalRequestFields?: Record<string, FieldDefinition>
   additionalTransformationFields?: Record<string, FieldDefinition>
+  additionalIdPrefix?: string
 }): ObjectType => {
-  const transformationTypes = createTransformationConfigTypes(adapter, {
-    hasDynamicFields: { refType: BuiltinTypes.BOOLEAN },
-    sourceTypeName: { refType: BuiltinTypes.STRING },
-    ...additionalTransformationFields,
-  })
+  const transformationTypes = createTransformationConfigTypes(
```
netama, line 54: hmm might be better to convert this function into one accepting named parameters as well? but I can also do that separately if you prefer

packages/adapter-components/src/config/request.ts
```diff
@@ -21,6 +21,10 @@ import { collections } from '@salto-io/lowerdash'
 const { findDuplicates } = collections.array
 
 export const ARG_PLACEHOLDER_MATCHER = /\{([\w_]+)\}/g
+const ID_SEPARATOR = '__'
+
+export const getConfigTypeName = (typeName: string, additionalPrefix?: string): string =>
```
netama, line 26: I don't think we should add the separator here, we can leave it up to the caller to pass it as part of the prefix if they want to. and then I think if we default to an empty string instead of undefined in the function args, then we can just assume the prefix is always passed (and then we won't need this function at all). wdyt?

packages/adapter-components/src/config/ducktype.ts
shir-reifenberg: i'll add

packages/okta-adapter/src/config.ts
```diff
@@ -1387,11 +1387,13 @@ export const configType = createMatchingObjectType<Partial<OktaConfig>>({
     [API_DEFINITIONS_CONFIG]: {
       refType: createSwaggerAdapterApiConfigType({
         adapter: OKTA,
+        elemIdPrefix: 'swagger',
```
netama, line 1390: do we need a prefix for these? I thought you said there is no risk of overlap?
(either way - we shouldn't use `swagger` for the type names, this is an implementation detail and shouldn't be that prominent for users)

packages/okta-adapter/src/config.ts
```diff
@@ -1387,11 +1387,13 @@ export const configType = createMatchingObjectType<Partial<OktaConfig>>({
...
       }),
     },
     [PRIVATE_API_DEFINITIONS_CONFIG]: {
       refType: createDucktypeAdapterApiConfigType({
         adapter: OKTA,
+        elemIdPrefix: 'ducktype',
```
netama, line 1396: I don't think "ducktype" should be user-facing - can we call these `settings_x` or something along these lines?

netama (approved):
the code LGTM, but I think we should change the prefixes for okta (also maybe we can add them in a separate PR?)

pre-approving to unblock, let's discuss before merging (I'm guessing we want to do it today?)

packages/okta-adapter/src/config.ts
shir-reifenberg: Talked IRL

PR #4291, opened 2023-04-27
SALTO-4057: Add deploy for Authenticator type

shir-reifenberg:
Support deployment of authenticators

---

- There is not endpoint for `remove`, deactivating `Authenticator` removes it from the admin console
- I added a CV that verifies `Authenticator` is not being used in MFA policy before its deactivation
- added ref

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/change_validators/enabled_authenticators.ts
+++ b/packages/okta-adapter/src/change_validators/enabled_authenticators.ts
@@ -1,0 +1,90 @@
+/*
+*                      Copyright 2023 Salto Labs Ltd.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


packages/okta-adapter/src/change_validators/enabled_authenticators.ts
```diff
@@ -0,0 +1,90 @@
...
+
+// Options are taken from: https://developer.okta.com/docs/reference/api/policy/#policy-factor-enroll-object
+const ENABLED_AUTHENTICATORS = ['OPTIONAL', 'REQUIRED']
+
+const getAutheticatorsForPolicy = (instance: InstanceElement): PolicyToAuthenticator[] => {
+  const authenticators = instance.value.settings?.authenticators
+  if (!Array.isArray(authenticators)) {
+    return []
+  }
+  return authenticators
```
edenhassid, line 39: maybe you can add a schema guard to check that the authenticator is in the right form and if it is not log a warning or something? instead of filtering the ones that are not valid. your choice :)

edenhassid (approved):
looks good! added one comment but its up to you 🌻

packages/okta-adapter/src/change_validators/enabled_authenticators.ts
shir-reifenberg: Added :)

PR #4292, opened 2023-04-27
SALTO-4050 fix more namespace errors and edit log to show full filePr…

benisraeladi:
Fix errors in namespace when API name is complex (has a `.` in the full name). Also edit log so that when the namespace prefix isn't added on purpose the log shows the full fileProps

---

---
_Release Notes_: _None_

---
_User Notifications_:  _None_

### DIFF
--- a/packages/salesforce-adapter/src/fetch.ts
+++ b/packages/salesforce-adapter/src/fetch.ts
@@ -178,7 +178,7 @@ const getFullName = (obj: FileProperties, addNamespacePrefixToFullName?: boolean
     return obj.fullName
   }
 
-  // Managed package names should exactly match obj.namespacePrefix
+  // Instances of type InstalledPackage fullNames should never include the namespace prefix
   if (obj.type === INSTALLED_PACKAGE_METADATA) {
     return obj.fullName
   }
@@ -205,7 +205,8 @@ const getFullName = (obj: FileProperties, addNamespacePrefixToFullName?: boolean
   // obj.namespacePrefix is defined. In these cases, we want to add the prefix manually
     if (obj.fullName.includes(API_NAME_SEPARATOR)) {
     // Case for API name with an object <Object>.X where it should be <Object>.<namespace>__X
-      return obj.fullName.replace(API_NAME_SEPARATOR, `${API_NAME_SEPARATOR}${namePrefix}`)
+      const [parentName, instanceName] = obj.fullName.split(API_NAME_SEPARATOR)
+      return `${parentName}${API_NAME_SEPARATOR}${namePrefix}${instanceName}`
     }
     return `${namePrefix}${obj.fullName}`
   }
--- a/packages/salesforce-adapter/src/fetch.ts
+++ b/packages/salesforce-adapter/src/fetch.ts
@@ -31,6 +31,7 @@ import {
   UNLIMITED_INSTANCES_VALUE,
   FOLDER_CONTENT_TYPE,
   API_NAME_SEPARATOR,
+  INSTALLED_PACKAGE_METADATA,
 } from './constants'
 import SalesforceClient, { ErrorFilter } from './client/client'
 import {
@@ -178,7 +179,7 @@ const getFullName = (obj: FileProperties, addNamespacePrefixToFullName?: boolean
   }
 
   // Managed package names should exactly match obj.namespacePrefix
-  if (obj.fullName === obj.namespacePrefix) {
+  if (obj.type === INSTALLED_PACKAGE_METADATA) {
     return obj.fullName
   }
 
@@ -204,8 +205,7 @@ const getFullName = (obj: FileProperties, addNamespacePrefixToFullName?: boolean
   // obj.namespacePrefix is defined. In these cases, we want to add the prefix manually
     if (obj.fullName.includes(API_NAME_SEPARATOR)) {
     // Case for API name with an object <Object>.X where it should be <Object>.<namespace>__X
-      const [nameWithoutNamespace, nameParts] = obj.fullName.split(API_NAME_SEPARATOR).reverse()
-      return `${makeArray(nameParts).reverse().join(API_NAME_SEPARATOR)}${API_NAME_SEPARATOR}${namePrefix}${nameWithoutNamespace}`
+      return obj.fullName.replace(API_NAME_SEPARATOR, `${API_NAME_SEPARATOR}${namePrefix}`)
     }
     return `${namePrefix}${obj.fullName}`
   }
--- a/packages/salesforce-adapter/src/fetch.ts
+++ b/packages/salesforce-adapter/src/fetch.ts
@@ -30,6 +30,7 @@ import {
   CUSTOM_OBJECT,
   UNLIMITED_INSTANCES_VALUE,
   FOLDER_CONTENT_TYPE,
+  API_NAME_SEPARATOR,
 } from './constants'
 import SalesforceClient, { ErrorFilter } from './client/client'
 import {
@@ -175,6 +176,12 @@ const getFullName = (obj: FileProperties, addNamespacePrefixToFullName?: boolean
   if (!obj.namespacePrefix) {
     return obj.fullName
   }
+
+  // Managed package names should exactly match obj.namespacePrefix
+  if (obj.fullName === obj.namespacePrefix) {
+    return obj.fullName
+  }
+
   const namePrefix = `${obj.namespacePrefix}${NAMESPACE_SEPARATOR}`
 
   if (obj.type === LAYOUT_TYPE_ID_METADATA_TYPE) {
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/src/fetch.ts
```diff
@@ -175,6 +176,12 @@ const getFullName = (obj: FileProperties, addNamespacePrefixToFullName?: boolean
   if (!obj.namespacePrefix) {
     return obj.fullName
   }
+
+  // Managed package names should exactly match obj.namespacePrefix
+  if (obj.fullName === obj.namespacePrefix) {
```
tamtamirr, line 181: Is this code relevant for metadata instances of type **InstalledPackage** only?
If so, wouldn't that be more accurate?
```suggestion
  if (obj.type === INSTALLED_PACKAGE_METADATA) {
```

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -188,15 +195,21 @@ const getFullName = (obj: FileProperties, addNamespacePrefixToFullName?: boolean
...
   if (addNamespacePrefixToFullName) {
-    // In some cases, obj.fullName does not contain the namespace prefix even though
-    // obj.namespacePrefix is defined. In these cases, we want to add the prefix manually
+  // In some cases, obj.fullName does not contain the namespace prefix even though
+  // obj.namespacePrefix is defined. In these cases, we want to add the prefix manually
+    if (obj.fullName.includes(API_NAME_SEPARATOR)) {
+    // Case for API name with an object <Object>.X where it should be <Object>.<namespace>__X
+      const [nameWithoutNamespace, nameParts] = obj.fullName.split(API_NAME_SEPARATOR).reverse()
+      return `${makeArray(nameParts).reverse().join(API_NAME_SEPARATOR)}${API_NAME_SEPARATOR}${namePrefix}${nameWithoutNamespace}`
+    }
```
tamtamirr, line 209: Simplify as we discussed f2f :)

packages/salesforce-adapter/test/fetch_installed_packages.test.ts
```diff
@@ -92,6 +92,32 @@ describe('Test fetching installed package metadata', () => {
       })
     })
   })
+  describe('Test fetching objects with complicated API names', () => {
+    describe('addNamespacePrefixToFullName is true', () => {
```
tamtamirr, line 96:
```suggestion
    describe('when addNamespacePrefixToFullName is true', () => {
```

packages/salesforce-adapter/test/fetch_installed_packages.test.ts
```diff
@@ -92,6 +92,32 @@ describe('Test fetching installed package metadata', () => {
...
+        const expectedFullName = 'Account.Test__TestObject'
+        const addNamespacePrefixToFullName = true
+        const fileProp = mockFileProperties({ fullName: fullNameFromList, type: 'Account', namespacePrefix: 'Test' })
+        const instance = await fetch({ fileProp, mockType: mockTypes.PermissionSet, addNamespacePrefixToFullName })
+
+        expect(instance).toBeDefined()
+        expect(instance?.value).toHaveProperty('fullName', expectedFullName)
+      })
+    })
+    describe('addNamespacePrefixToFullName is false', () => {
```
tamtamirr, line 108:
```suggestion
    describe('when addNamespacePrefixToFullName is false', () => {
```

packages/salesforce-adapter/test/fetch_installed_packages.test.ts
```diff
@@ -92,6 +92,32 @@ describe('Test fetching installed package metadata', () => {
...
+      it('should not add prefix to the object\'s name', async () => {
+        const fullNameFromList = 'TestPermissionSet'
+        const expectedFullName = 'TestPermissionSet'
+        const addNamespacePrefixToFullName = false
+        const fileProp = mockFileProperties({ fullName: fullNameFromList, type: 'PermissionSet', namespacePrefix: 'Test' })
+        const instance = await fetch({ fileProp, mockType: mockTypes.PermissionSet, addNamespacePrefixToFullName })
+
+        expect(instance).toBeDefined()
+        expect(instance?.value).toHaveProperty('fullName', expectedFullName)
```
tamtamirr, line 117: Nit: Consider the following instead
```suggestion
        expect(instance).toEqual(expect.objectContaining({
          value: expect.objectContaining({ [INSTANCE_FULL_NAME_FIELD]: expectedFullName })
        }))
```

packages/salesforce-adapter/test/fetch_installed_packages.test.ts
```diff
@@ -92,6 +92,32 @@ describe('Test fetching installed package metadata', () => {
...
+      })
+    })
+    describe('addNamespacePrefixToFullName is false', () => {
+      it('should not add prefix to the object\'s name', async () => {
+        const fullNameFromList = 'TestPermissionSet'
+        const expectedFullName = 'TestPermissionSet'
```
tamtamirr, line 111: Nit: Is this two vars really necessary? I would just inline them were they are used. I'm not sure it adds readability to the test.

packages/salesforce-adapter/src/fetch.ts
tamtamirr, line 181: Please add a UT for this scenario.

tamtamirr (commented):
Looking good 👑 Added some comments

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -175,6 +177,12 @@ const getFullName = (obj: FileProperties, addNamespacePrefixToFullName?: boolean
   if (!obj.namespacePrefix) {
     return obj.fullName
   }
+
+  // Managed package names should exactly match obj.namespacePrefix
```
tamtamirr, line 181:
```suggestion
  // Instances of type InstalledPackage fullNames should never include the namespace prefix
```

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -188,15 +196,20 @@ const getFullName = (obj: FileProperties, addNamespacePrefixToFullName?: boolean
...
   }
+
   if (obj.fullName.startsWith(namePrefix)) {
     return obj.fullName
   }
   if (addNamespacePrefixToFullName) {
-    // In some cases, obj.fullName does not contain the namespace prefix even though
-    // obj.namespacePrefix is defined. In these cases, we want to add the prefix manually
+  // In some cases, obj.fullName does not contain the namespace prefix even though
+  // obj.namespacePrefix is defined. In these cases, we want to add the prefix manually
```
tamtamirr, line 205: indentation

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -188,15 +196,20 @@ const getFullName = (obj: FileProperties, addNamespacePrefixToFullName?: boolean
...
+    if (obj.fullName.includes(API_NAME_SEPARATOR)) {
+    // Case for API name with an object <Object>.X where it should be <Object>.<namespace>__X
+      return obj.fullName.replace(API_NAME_SEPARATOR, `${API_NAME_SEPARATOR}${namePrefix}`)
```
tamtamirr, line 208: Nit:
While this is correct, this logic seems vague for me, and since this value is **EXTREMELY** important I'd prefer to be explicit here.
```suggestion
      [parentName, instanceName] = obj.fullName.split(API_NAME_SEPARATOR)
      return `${parentName}.${namePrefix}${instanceName}`
```

packages/salesforce-adapter/test/fetch_installed_packages.test.ts
```diff
@@ -92,6 +93,45 @@ describe('Test fetching installed package metadata', () => {
...
+    describe('when addNamespacePrefixToFullName is true', () => {
+      it('should add prefix to the object\'s name correctly', async () => {
+        const fullNameFromList = 'Account.TestObject'
+        const expectedFullName = 'Account.Test__TestObject'
+        const addNamespacePrefixToFullName = true
+        const fileProp = mockFileProperties({ fullName: fullNameFromList, type: 'Account', namespacePrefix: 'Test' })
+        const instance = await fetch({ fileProp, mockType: mockTypes.PermissionSet, addNamespacePrefixToFullName })
+
+        expect(instance).toBeDefined()
+        expect(instance?.value).toHaveProperty('fullName', expectedFullName)
```
tamtamirr, line 106: I've seen you replaced the other tests with my previous suggestion, probably missed this one..? It's better to be aligned in all the tests :) (Those that were part of this PR at-least, I wouldn't touch code that is unrelated to this PR)

tamtamirr (approved):
LGTM 🥳 Added some comments.
Approving to save cycles.

packages/salesforce-adapter/src/fetch.ts
ori-moisis: I really think we should not make this change "blindly"
we should stick with logging first to make sure we don't ruin other scenarios that currently work

benisraeladi: @ori-moisis There won't be any change unless the flag is off. Until then, there will only be logging on the failed scenario. Do you mean to split the flag up so that each sub-scenario is logged?

PR #4293, opened 2023-04-27
SALTO-4063: Add missing references for Okta Expression Language templates

shir-reifenberg:
Add missing references for Okta Expression Language templated

---

For now the best way to detect a missing reference to group (without changing the entire filter) is to check if the `id` starts with `00g` which is the prefix of group ids in Okta in general (from what I saw)

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/filters/expression_language.ts
+++ b/packages/okta-adapter/src/filters/expression_language.ts
@@ -16,11 +16,13 @@
 import _ from 'lodash'
 import { Change, Element, getChangeData, InstanceElement, isInstanceElement, isTemplateExpression, ReferenceExpression, TemplateExpression, TemplatePart } from '@salto-io/adapter-api'
 import { extractTemplate, replaceTemplatesWithValues, resolvePath, resolveTemplates } from '@salto-io/adapter-utils'
+import { references as referenceUtils } from '@salto-io/adapter-components'
 import { logger } from '@salto-io/logging'
 import { FilterCreator } from '../filter'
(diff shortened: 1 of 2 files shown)


packages/okta-adapter/src/filters/expression_language.ts
```diff
@@ -134,9 +136,16 @@ const stringToTemplate = (
...
-        return matchingInstance !== undefined
-          ? new ReferenceExpression(matchingInstance.elemID, matchingInstance)
-          : expression
+        if (matchingInstance !== undefined) {
+          return new ReferenceExpression(matchingInstance.elemID, matchingInstance)
+        }
+        // hack to verify this is a group id, because group ids in okta starts with '00g'
+        if (id.startsWith('00g') && id.length > 10) {
+          // create missing reference for group
+          const missingInstance = createMissingInstance(OKTA, GROUP_TYPE_NAME, id)
```
netama, line 145: looks like we don't have a `enableMissingReferences` flag in okta yet - might be worth adding it here? if we are getting into heuristics we should have an easy way out :)

netama (approved):
LGTM, added a suggestion that I think might be worth adding as part of this PR, but if you prefer we can also add separately (it will require changing a few existing places in the code as well)

packages/okta-adapter/src/filters/expression_language.ts
shir-reifenberg: Cool, I prefer adding in a different PR
Adding ticket for this

PR #4295, opened 2023-04-27
SALTO-4053: fix sort of ticket form conditions

edenhassid:
fix sort of ticket form conditions

---

---
_Release Notes_: 
zendesk:
* fix sort of ticket form conditions

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/filters/unordered_lists.ts
+++ b/packages/zendesk-adapter/src/filters/unordered_lists.ts
@@ -13,7 +13,7 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
-import _, { isArray, isString } from 'lodash'
+import _, { isArray, isBoolean, isString } from 'lodash'
 import {
   Element, InstanceElement, isInstanceElement, isReferenceExpression, ReferenceExpression,
 } from '@salto-io/adapter-api'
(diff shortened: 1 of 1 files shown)


packages/zendesk-adapter/src/filters/unordered_lists.ts
```diff
@@ -125,7 +125,8 @@ const isValidConditions = (conditions: unknown, customFieldById: Record<string,
       isReferenceExpression(condition.value)
       && (customFieldById[condition.value.elemID.getFullName()]?.value.value !== undefined)
     )
-    || (isString(condition.value)))
+    || (isString(condition.value))
+|| (isBoolean(condition.value)))
```
netama, line 129: 1. nit:
```suggestion
    || _.isString(condition.value)
    || _.isBoolean(condition.value))
```
2. do we know these are all the possible values? could it be e.g. a number as well?

packages/zendesk-adapter/src/filters/unordered_lists.ts
```diff
@@ -140,7 +141,7 @@ const sortConditions = (
     if (isValidConditions(conditions, customFieldById)) {
       form.value[conditionType] = _.sortBy(
         conditions,
-        condition => (isString(condition.value)
+        condition => (isString(condition.value) || isBoolean(condition.value)
```
netama, line 144: is it worth adding a test for this?

netama (approved):
approving to unblock

PR #4296, opened 2023-04-27
SALTO-3855: NaCls with invalid characters are ignored

alonstern:
When there is an invalid character in a NaCl (e.g., `'`), the NaCl is ignored and no error is returned 

---
_Release Notes_: 
_Core_:
- Fixed a bug where no error would be returned when there is an invalid character in the NaCl (e.g., `'`)

---
_User Notifications_: 
None

### DIFF
--- a/packages/workspace/src/parser/internal/native/parse.ts
+++ b/packages/workspace/src/parser/internal/native/parse.ts
@@ -15,6 +15,7 @@
 */
 import { Element } from '@salto-io/adapter-api'
 import { flattenElementStr } from '@salto-io/adapter-utils'
+import { logger } from '@salto-io/logging'
 import { ParseResult } from '../../types'
 import { Keywords } from '../../language'
 import { Functions } from '../../functions'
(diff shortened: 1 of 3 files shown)


packages/workspace/src/parser/internal/native/parse.ts
```diff
@@ -102,6 +103,15 @@ export async function parseBuffer(
           TOKEN_TYPES.MERGE_CONFLICT,
         ))
       }
+    } else if (e instanceof UnknownCharacter) {
```
ori-moisis, line 106: do we need another "general" catch here so we don't completely ignore errors like we did with this case before this fix?
seems like the code already did throw an error from `consumeValue` and this code would just silently suppress it.
maybe we should at least move the "unknown error" log to here, just in case the next time won't come from `consumeValue` but from somewhere else?

ori-moisis (approved):
I think the fix is good, maybe I would just move the log

PR #4299, opened 2023-04-28
SALTO-4070 - Salesforce: Replace Formulon package with a regex

tomsellek:
Formulon is currently unusable due to https://github.com/leifg/formulon/issues/1015, and we don't need a proper syntactic analysis, just the list of identifiers in the formula

---
This solution probably only sort-of-works, but:
1. We filter out invalid references anyway
2. We don't need to extract all identifiers - worst case is we don't add anything over what the Tooling API gives us.

https://xkcd.com/1171/

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/filters/formula_deps.ts
+++ b/packages/salesforce-adapter/src/filters/formula_deps.ts
@@ -14,6 +14,7 @@
 * limitations under the License.
 */
 
+import _ from 'lodash'
 import { logger } from '@salto-io/logging'
 import { collections } from '@salto-io/lowerdash'
 import { ElemID, ElemIDType, Field, isObjectType, ReadOnlyElementsSource, ReferenceExpression } from '@salto-io/adapter-api'
@@ -23,11 +24,6 @@ import { isFormulaField } from '../transformers/transformer'
 import { CUSTOM_METADATA_SUFFIX, FORMULA, SALESFORCE } from '../constants'
 import { FormulaIdentifierInfo, IdentifierType, parseFormulaIdentifier } from './formula_utils/parse'
 import { buildElementsSourceForFetch, extractFlatCustomObjectFields } from './utils'
-
-/* eslint-disable-next-line @typescript-eslint/no-var-requires */
-const formulon = require('formulon')
-
-const { extract } = formulon
 
 const log = logger(module)
 const { awu, groupByAsync } = collections.asynciterable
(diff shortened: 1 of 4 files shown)


packages/salesforce-adapter/src/filters/formula_deps.ts
```diff
@@ -74,6 +70,20 @@ const referencesFromIdentifiers = async (typeInfos: FormulaIdentifierInfo[]): Pr
     ))
 )
 
+
+const IDENTIFIER_REGEX = /(?:^|[^'"\w$]+)([^\s'"()&|+/*=\-,]+)(?:[\s),]|$)/mgi
```
amitgoldarb, line 74: This is definitely a cool concept.
I think we should probably have some UT on many different types of formulas specifically for the regex. also I think we should add some comments better explaining the regex?
Looking at the existing UT here, it's cool they didn't break :)
@ori-moisis curious to hear you're thoughts here as well.

amitgoldarb (commented):
Very cool!
Added some comments, if this is performant than I think it's better than the existing solution.

packages/salesforce-adapter/src/filters/formula_deps.ts
ori-moisis: is this ready for review?
I skipped it because it is marked as draft

tomsellek: Refactored for improved readability and marked as ready for review

packages/salesforce-adapter/src/filters/formula_deps.ts
```diff
@@ -123,7 +145,7 @@ const addDependenciesAnnotation = async (field: Field, allElements: ReadOnlyElem
 
   try {
     const formulaIdentifiers: string[] = log.time(
-      () => (extract(formula)),
+      () => (extractIdentifiers(formula)),
```
ori-moisis, line 148: is there really a reason to have this as a separate step from "parsing" the identifiers?
do we really need to time this separately?

ori-moisis (commented):
so....
not really sure what to do with this one

I think we'd all be very surprised if it was possible to properly parse the formula syntax with a single regex, so it shouldn't be very surprising that it didn't take long to find examples where this misses identifiers and examples where this returns more identifiers than it should.

example where we would return more than needed:
```
/* comments are a thing we don't handle: MyField__c */
DifferentField__c + 5
```
example where we would miss (taken from [here](https://www.salesforceben.com/salesforce-formula-examples/))
```
(IF( AnnualRevenue = 0, 0, 10)) +
(IF(ISPICKVAL(Industry, “”) , 0, 10)) +
(IF( MobilePhone = “”, 0, 10))
```
(won't return `MobilePhone`)

I am sure both of these specific issues are fixable, but I don't know if trying to fix each example and finding new ones is really the optimal path here...

PR #4301, opened 2023-04-30
SALTO-3705: schema validation for subject conditions should allow reference expression

edenhassid:
schema validation for subject conditions should allow reference expression

---

---
_Release Notes_: 
none

---
_User Notifications_: 
none


shir-reifenberg (approved):
💯

PR #4302, opened 2023-04-30
Update field validation only for custom fields

adfineberg:
_Update field scriptid validation to apply only on custom fields_

---

---
_Release Notes_: 
_NetSuite Adapter_
* Update field scriptid validation to apply only on custom fields

---
_User Notifications_:


dantal4 (approved):
Great work!

PR #4311, opened 2023-05-01
SALTO-3095: Redeploy uneditable locked fields

shoshana-houri:
Deploying data elements fails on read only locked fields. This PR solves it.

---

Deploying data elements fails on read only locked fields. The user cannot change the access level of a locked field (and make it editable), hence a change validator is irrelevant in this case. We catch the 'INSUFFICIENT_PERMISSION' error from NetSuite and redeploy the instance again without the problematic field. The result in such case is a *full success*.

---
_Release Notes_: 

---
_User Notifications_:


packages/netsuite-adapter/src/constants.ts
```diff
@@ -24,15 +24,20 @@ export const CUSTOM_RECORDS_PATH = 'CustomRecords'
 
 // Type names
 export const ADDRESS_FORM = 'addressForm'
+export const CRM_CUSTOM_FIELD = 'crmcustomfield'
+export const CRM_CUSTOM_FIELD_PREFIX = 'custevent'
```
kami-delaroz, line 28: if these prefix constants are being used only by test code, I think they should be in a test helper file

kami-delaroz (commented):
Looks good!
a few comments / questions

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/filter_uneditable_locked_field.ts
```diff
@@ -0,0 +1,103 @@
...
+  instance: InstanceElement,
+  hasElemID: HasElemIDFunc,
+): Promise<InstanceElement | undefined> => {
+  if (isWriteResponseSuccess(writeResponse)
+    || (writeResponse.status.statusDetail[0].code !== INSUFFICIENT_PERMISSION_ERROR)) {
+    return undefined
+  }
+  const fieldName = extractFieldName((writeResponse as WriteResponseError).status.statusDetail[0].message)
+  if (isDefined(fieldName) && await isLockedElement(fieldName, hasElemID)) {
+    return removeUneditableLockedField(instance, fieldName)
```
kami-delaroz, line 100: I would have added a comment here, mentioning we **suspect** this is an uneditable field, as we can't say for sure, and theoretically we might have another issue here

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/filter_uneditable_locked_field.ts
```diff
@@ -0,0 +1,103 @@
...
+  return (isDefined(fieldFullName) ? !(await hasElemID(ElemID.fromFullName(fieldFullName))) : false)
+}
+
+const removeUneditableLockedField = (
+  instance: InstanceElement,
+  fieldName: string
+): InstanceElement | undefined => {
+  const customFields = instance.value[CUSTOM_FIELD_LIST]
+  if (_.isUndefined(customFields)) {
+    return undefined
```
kami-delaroz, line 66: maybe I misunderstood something, but in this case we want to have the origin instance? as this this what will happen in the enclosing function (when calculating updatedInstances)

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
```diff
@@ -506,18 +508,102 @@ export default class SoapClient {
...
+    hasElemID: HasElemIDFunc,
+  ): Promise<({
+    redeployWriteResponseList: WriteResponse[]
+    indicesMap: Map<number, number>
+    updatedInstances: InstanceElement[]
+  }) | undefined> {
+    const modifiedInstances = await awu(writeResponseList)
+      .map((writeResponse, index) => getModifiedInstance(writeResponse, _.cloneDeep(instances[index]), hasElemID))
+      .toArray()
+    const modifiedInstancesToRedeploy = modifiedInstances.filter(isDefined)
```
dantal4, line 527: It is very confusing that `modifiedInstances` is a list of instances and undefined values. Can you have function instead of `getModifiedInstance` that returns instances in groups of "redeploy", "success", "failed" ? Then all these checks of undefined values can be removed.

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
```diff
@@ -506,18 +508,102 @@ export default class SoapClient {
...
+
+    if (modifiedInstancesToRedeploy.length === 0) {
+      log.debug('No locked fields found, cancelling redeployment.')
```
dantal4, line 530: similar to @kami-delaroz 's comment.

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/filter_uneditable_locked_field.ts
```diff
@@ -0,0 +1,103 @@
...
+    instance.value['platformCore:nullFieldList']['platformCore:name'].push(fieldName)
+  } else {
+    instance.value['platformCore:nullFieldList'] = { 'platformCore:name': [fieldName] }
+  }
+  return instance
+}
+
+export const getModifiedInstance = async (
+  writeResponse: WriteResponse,
+  instance: InstanceElement,
```
dantal4, line 91: I think that custom record instances should be treated differently. Lets f2f on it.

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
```diff
@@ -506,18 +508,102 @@ export default class SoapClient {
...
+      return undefined
+    }
+
+    log.debug('Deployment failed on \'INSUFFICIENT PERMISSION\' error for uneditable locked fields.'
+    + ' Redeploying changes without the locked fields.')
+    const redeployBody = {
```
dantal4, line 536: Can you have a function to calculate the body? It is the same code here, in `addInstances` and in `updateInstances`.

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
```diff
@@ -506,18 +508,102 @@ export default class SoapClient {
...
+  }
+
+  private async runFullDeploy(
+    originalInstances: InstanceElement[],
+    body: {
+      attributes: Record<string, string>
+      record: RecordValue[]
+    } | {
+      baseRef: object[]
+    },
```
dantal4, line 562: please extract it to a type.

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
```diff
@@ -506,18 +508,102 @@ export default class SoapClient {
...
+    action: 'updateList' | 'addList' | 'deleteList',
+    hasElemID?: HasElemIDFunc,
```
dantal4, line 564: this can be required all over right?

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
```diff
@@ -506,18 +508,102 @@ export default class SoapClient {
...
+  ): Promise<(number | Error)[]> {
+    let writeResponseList = await this.runDeployAction(body, action)
+
+    if (action !== 'deleteList' && isDefined(hasElemID)) {
+      const redeployWithRetry = async (
```
dantal4, line 569: This is very scary to see a function defined inside a condition scope, inside another function 😨 
Can you define it outside of `runFullDeploy`?

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
```diff
@@ -506,18 +508,102 @@ export default class SoapClient {
...
+        prevWriteResponseList: WriteResponse[],
+        instances: InstanceElement[]
+      ): Promise<void> => {
+        if (retriesLeft === 0) {
+          log.debug('Redeployment on locked fields exceed max retries.')
+          return
+        }
+
+        log.debug('Retrying deployment request for locked fields. Retries left: %d', retriesLeft)
+        const redeployResult = await this.redeployLockedFields(instances, prevWriteResponseList, action, hasElemID)
```
dantal4, line 580: Why do we need to count the retries here? we can run it until all locked fields are removed, right?

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
```diff
@@ -506,18 +508,102 @@ export default class SoapClient {
...
+  private async redeployLockedFields(
+    instances: InstanceElement[],
+    writeResponseList: WriteResponse[],
+    action: 'updateList' | 'addList',
+    hasElemID: HasElemIDFunc,
+  ): Promise<({
+    redeployWriteResponseList: WriteResponse[]
+    indicesMap: Map<number, number>
+    updatedInstances: InstanceElement[]
+  }) | undefined> {
```
dantal4, line 523: It doesn't feel right to return here `undefined` as a "success" code...

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/filter_uneditable_locked_field.ts
```diff
@@ -0,0 +1,103 @@
...
+import { values } from '@salto-io/lowerdash'
+import _ from 'lodash'
+import { logger } from '@salto-io/logging'
+import { CRM_CUSTOM_FIELD, ENTITY_CUSTOM_FIELD, ITEM_CUSTOM_FIELD, NETSUITE, OTHER_CUSTOM_FIELD, SOAP_SCRIPT_ID } from '../../../constants'
+import { WriteResponse, WriteResponseError, isWriteResponseSuccess } from './types'
+
+const log = logger(module)
+const { isDefined } = values
+
+export type HasElemIDFunc = (elemID: ElemID) => Promise<boolean>
```
dantal4, line 27: it's better to define it in `client/types.ts`

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
```diff
@@ -571,7 +657,7 @@ export default class SoapClient {
         })
       }).toArray(),
     }
-    return this.runDeployAction(instances, body, 'deleteList')
+    return this.runFullDeploy(instances, body, 'deleteList')
```
dantal4, line 660: you can call `runDeployAction` here and parse the response. Then you'll be able to remove the `if (action !== 'deleteList'...` condition.

dantal4 (commented):
Nice work!

packages/netsuite-adapter/src/constants.ts
shoshana-houri: Exactly, I should replace the hard coded text with them :)

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/filter_uneditable_locked_field.ts
shoshana-houri: I we can't change anything we don't want to redeploy again and again the same instance, therefore we don't want it to return as 'modified'.

packages/netsuite-adapter/src/client/client.ts
```diff
@@ -372,6 +373,7 @@ export default class NetsuiteClient {
     changes: Change[],
     groupID: string,
     additionalSdfDependencies: AdditionalDependencies,
+    hasElemID?: HasElemIDFunc,
```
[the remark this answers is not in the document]
shoshana-houri: Didn't want to create much noise in irrelevant places that don't send this argument. I see though it's only tests so I'll remove the optional and modify them :)

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/filter_uneditable_locked_field.ts
```diff
@@ -0,0 +1,103 @@
...
+const INSUFFICIENT_PERMISSION_ERROR = 'INSUFFICIENT_PERMISSION'
+const CUSTOM_FIELD_LIST = 'customFieldList'
+const fieldNameRegex = new RegExp('You do not have permissions to set a value for element [\\w]+')
+
+const extractFieldName = (errorMessage: string): string | undefined => {
+  const matchedField = errorMessage.match(fieldNameRegex)
+  if (matchedField?.length !== 1) {
+    return undefined
+  }
+  return matchedField[0].split(' ').pop() as string
```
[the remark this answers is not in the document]
shoshana-houri: Cool! thanks

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/filter_uneditable_locked_field.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export type HasElemIDFunc = (elemID: ElemID) => Promise<boolean>
+
+const INSUFFICIENT_PERMISSION_ERROR = 'INSUFFICIENT_PERMISSION'
+const CUSTOM_FIELD_LIST = 'customFieldList'
+const fieldNameRegex = new RegExp('You do not have permissions to set a value for element [\\w]+')
+
+const extractFieldName = (errorMessage: string): string | undefined => {
+  const matchedField = errorMessage.match(fieldNameRegex)
+  if (matchedField?.length !== 1) {
```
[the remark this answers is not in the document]
shoshana-houri: I didn't see any cases with more than one field, but I thought it's better to be as accurate as we can and not create unexpected behavior.

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/filter_uneditable_locked_field.ts
```diff
@@ -0,0 +1,103 @@
...
+
+const removeUneditableLockedField = (
+  instance: InstanceElement,
+  fieldName: string
+): InstanceElement | undefined => {
+  const customFields = instance.value[CUSTOM_FIELD_LIST]
+  if (_.isUndefined(customFields)) {
+    return undefined
+  }
+  const validCustomFields = customFields['platformCore:customField']
```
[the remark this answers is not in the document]
shoshana-houri: The only place I found defining this field is:
https://github.com/salto-io/salto/blob/c0ff7c31a8d41bd1a6e0b7459b8e377a62374223/packages/netsuite-adapter/src/filters/data_instances_custom_fields.ts#L75-L80
And `customFields` type is a list.

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/filter_uneditable_locked_field.ts
```diff
@@ -0,0 +1,103 @@
...
+    return undefined
+  }
+
+  log.debug('Removing locked custom field: %s from instance: %s', fieldName, instance.elemID.getFullName())
+  if (validCustomFields.length > 0) {
+    customFields['platformCore:customField'] = validCustomFields
+  } else {
+    delete instance.value[CUSTOM_FIELD_LIST]
+  }
+  if (isDefined(instance.value['platformCore:nullFieldList']?.['platformCore:name'])) {
```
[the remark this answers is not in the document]
shoshana-houri: Ohh nice, thanks for that!

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/filter_uneditable_locked_field.ts
```diff
@@ -0,0 +1,103 @@
...
+  if (fieldName.startsWith('custrecord')) {
+    dirName = OTHER_CUSTOM_FIELD
+  } else if (fieldName.startsWith('custentity')) {
+    dirName = ENTITY_CUSTOM_FIELD
+  } else if (fieldName.startsWith('custitem')) {
+    dirName = ITEM_CUSTOM_FIELD
+  } else if (fieldName.startsWith('custevent')) {
+    dirName = CRM_CUSTOM_FIELD
+  }
+  return isDefined(dirName) ? `${NETSUITE}.${dirName}.instance.${fieldName}` : undefined
```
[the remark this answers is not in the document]
shoshana-houri: > What will happen if the account name is different than `netsuite`? Try running it in an new environment with netsuite in a different account name:
> 
> ```
> salto account add --account-name <abc> netsuite
> ```

As we discussed that's fine - the elementsSource replaces it with the real account name.





> Also - why not returning `new ElemID(...)` ?

You're right, returning a new elemID would be better :)

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/filter_uneditable_locked_field.ts
```diff
@@ -0,0 +1,103 @@
...
+  log.debug('Removing locked custom field: %s from instance: %s', fieldName, instance.elemID.getFullName())
+  if (validCustomFields.length > 0) {
+    customFields['platformCore:customField'] = validCustomFields
+  } else {
+    delete instance.value[CUSTOM_FIELD_LIST]
+  }
+  if (isDefined(instance.value['platformCore:nullFieldList']?.['platformCore:name'])) {
+    instance.value['platformCore:nullFieldList']['platformCore:name'].push(fieldName)
+  } else {
+    instance.value['platformCore:nullFieldList'] = { 'platformCore:name': [fieldName] }
```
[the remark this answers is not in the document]
shoshana-houri: It never has deleted fields but only one added, perhaps I misunderstood the question.
As for sharing a function - I think the best place for it will be inside `data_instances_null_fields.ts`, but importing it creates a dependency cycle (`filter` is imported by `client`, which imports `filter_uneditable_locked_field.ts` ). Creating a new file just for this purpose seems a bit too much... WDYT?

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
```diff
@@ -506,18 +508,102 @@ export default class SoapClient {
...
+    writeResponseList: WriteResponse[],
+    action: 'updateList' | 'addList',
+    hasElemID: HasElemIDFunc,
+  ): Promise<({
+    redeployWriteResponseList: WriteResponse[]
+    indicesMap: Map<number, number>
+    updatedInstances: InstanceElement[]
+  }) | undefined> {
+    const modifiedInstances = await awu(writeResponseList)
+      .map((writeResponse, index) => getModifiedInstance(writeResponse, _.cloneDeep(instances[index]), hasElemID))
```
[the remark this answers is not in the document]
shoshana-houri: I thought the best practice will be to not modify the instances we receive, as it might create unexpected behavior in the future.

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/filter_uneditable_locked_field.ts
shoshana-houri: Ok :)
Just pointing out that the different custom fields we take care of cannot be applied to custom record instances, as we can see in `packages/netsuite-adapter/src/data_elements/custom_fields.ts`
```
const CUSTOM_FIELD_TO_TYPE: Record<string, Record<string, string[]>> = {
  entitycustomfield: {
    appliestocontact: ['Contact'],
    appliestocustomer: ['Customer'],
    appliestoemployee: ['Employee'],
    appliestopartner: ['Partner'],
    appliestovendor: ['Vendor'],
    appliestopricelist: ['PriceList'],
  },
  itemcustomfield: {
    appliestogroup: ['GroupItem'],
    appliestoinventory: ['InventoryItem'],
    appliestoitemassembly: ['AssemblyItem'],
    appliestokit: ['KitItem'],
    appliestononinventory: [
      'NonInventoryPurchaseItem',
      'NonInventorySaleItem',
      'NonInventoryResaleItem',
    ],
    appliestoothercharge: [
      'OtherChargeSaleItem',
      'OtherChargeResaleItem',
      'OtherChargePurchaseItem',
    ],
  },
  crmcustomfield: {
    appliestocampaign: ['Campaign'],
    appliestoprojecttask: ['ProjectTask'],
    appliestophonecall: ['PhoneCall'],
    appliestosolution: ['Solution'],
    appliestotask: ['Task'],
  },
}
```

If we want to be sure no unexpected behavior happens we can just filter them out using `!isCustomRecordType`, since we see they're irrelevant.

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
shoshana-houri: Yes, I'll add it in the necessary places :)

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
shoshana-houri: I implemented something kinda similar, let me know what you think of it.

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
shoshana-houri: I thought it gives a nice indication, knowing how many times we had to repeat this operation. We can use that in order to increase/decrease this number in the future.

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/filter_uneditable_locked_field.ts
```diff
@@ -0,0 +1,103 @@
...
+
+const isLockedElement = async (fieldName: string, hasElemID: HasElemIDFunc) : Promise<boolean> => {
+  const fieldFullName = getFieldFullName(fieldName)
+  return (isDefined(fieldFullName) ? !(await hasElemID(ElemID.fromFullName(fieldFullName))) : false)
+}
+
+const removeUneditableLockedField = (
+  instance: InstanceElement,
+  fieldName: string
+): InstanceElement | undefined => {
```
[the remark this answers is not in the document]
dantal4: I think that cloning here has no value, since it's all already cloned for the deploy itself. IMO a boolean would be a good solution.

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/filter_uneditable_locked_field.ts
```diff
@@ -0,0 +1,103 @@
...
+  const customFields = instance.value[CUSTOM_FIELD_LIST]
+  if (_.isUndefined(customFields)) {
```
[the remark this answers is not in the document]
dantal4: IMO whenever lodash can help you do something more simple - use it, but if it doesn't (like this case for example) I prefer the native js/ts solution.

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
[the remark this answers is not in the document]
dantal4: This is a good approach! We should definitely think about this whenever we want to change the content of an element. In this case- we clone all changes before calling the deploy, and the SOAP client is the last "consumer" of these changes, so IMO we can avoid cloning here 🙏

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/filter_uneditable_locked_field.ts
```diff
@@ -0,0 +1,139 @@
...
+  } else {
+    delete instance.value[CUSTOM_FIELD_LIST]
+  }
+
+  const nullCustomFields = Array.isArray(instance.value[PLATFORM_CORE_NULL_FIELD_LIST]?.[PLATFORM_CORE_NAME])
+    ? instance.value[PLATFORM_CORE_NULL_FIELD_LIST][PLATFORM_CORE_NAME]
+    : []
+  instance.value[PLATFORM_CORE_NULL_FIELD_LIST] = {
+    [PLATFORM_CORE_NAME]: [...nullCustomFields, fieldName],
+  }
```
dantal4, line 92: This will change the current value of the field to NULL, isn't it?
I think it's better that we'll keep it unchanged instead.

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/filter_uneditable_locked_field.ts
```diff
@@ -0,0 +1,139 @@
...
+      // removeUneditableLockedField might return undefined.
+      return removeUneditableLockedField(instance, fieldName)
+    }
+  }
+
+  return undefined
+}
+
+
+export const getModifiedInstances = async (
```
dantal4, line 115: I think that this function needs a more indicative name

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
```diff
@@ -506,7 +512,28 @@ export default class SoapClient {
...
+      attributes: {
+        'xmlns:platformCore': SOAP_CORE_URN,
+      },
+      record: await awu(instances).map(
+        async instance => this.convertToSoapRecord(instance.value, await instance.getType())
+      ).toArray(),
+    }
+  }
+
+  static parseWriteResponseList(
```
dantal4, line 531: should be `private static`

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
```diff
@@ -517,28 +544,74 @@ export default class SoapClient {
...
+      await this.getAddAndUpdateDeployBody(modifiedInstances),
+      action,
+    )
+    const writeResponseList = prevWriteResponseList.map((writeResponse, index): WriteResponse => {
+      const redeployIndex = indicesMap.get(index)
+      return isDefined(redeployIndex)
+        ? redeployWriteResponseList[redeployIndex]
+        : writeResponse
+    })
+    return this.redeployLockedFieldsWithRetry(
```
dantal4, line 577: The order of actions in is function is confusing. IMO you should first try to deploy, then calculate which instances should be deployed again, and only if there are instances to re-deploy run the recurse.

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/filter_uneditable_locked_field.ts
```diff
@@ -0,0 +1,139 @@
...
+  const allInstancesUpdated = modifiedOrUndefinedInstances
+    .map((modifiedInstance, index) => (isDefined(modifiedInstance) ? modifiedInstance : instances[index]))
+  const indicesMap = new Map(modifiedOrUndefinedInstances
+    .map((modifiedInstance, originalIdx) => (isDefined(modifiedInstance) ? originalIdx : undefined))
+    .filter(isDefined)
+    .map((originalIdx, reducedIdx) => [originalIdx, reducedIdx]))
+
+  return {
+    modifiedInstances: modifiedOrUndefinedInstances.filter(isDefined),
+    allInstancesUpdated,
```
dantal4, line 136: this should be redundant if you edit `instances` in place.

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
```diff
@@ -517,28 +544,74 @@ export default class SoapClient {
...
-    return this.runDeployAction(instances, body, 'addList')
+  private async runFullDeploy(
+    instances: InstanceElement[],
+    body: AddAndUpdateDeployBody | DeleteDeployBody,
+    action: 'updateList' | 'addList',
+    hasElemID: HasElemIDFunc,
+  ): Promise<(number | Error)[]> {
+    const writeResponseList = await this.redeployLockedFieldsWithRetry(
+      LOCKED_FIELDS_MAX_REDEPLOYS,
+      await this.runDeployAction(body, action),
```
dantal4, line 594: body can be calculated just here instead in both add/update instances functions, right?

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
```diff
@@ -517,28 +544,74 @@ export default class SoapClient {
...
+      return prevWriteResponseList
+    }
+
+    log.debug('Deployment failed on \'INSUFFICIENT PERMISSION\' error for uneditable locked fields.'
++ ' Redeploying changes without the locked fields.', 'Retries left: %d', retriesLeft)
+    const redeployWriteResponseList = await this.runDeployAction(
+      await this.getAddAndUpdateDeployBody(modifiedInstances),
+      action,
+    )
+    const writeResponseList = prevWriteResponseList.map((writeResponse, index): WriteResponse => {
```
dantal4, line 571: IMO it would be better if `redeployLockedFieldsWithRetry` won't have to know the context of the previously deployed instances. This was each iteration is responsible only for the instances that will be redeployed, and the caller will be responsible to update the results from the recursive call. It should make this function simpler, give it a try.

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/filter_uneditable_locked_field.ts
```diff
@@ -0,0 +1,139 @@
...
+  return instance
+}
+
+const getModifiedInstance = async (
+  writeResponse: WriteResponse,
+  instance: InstanceElement,
+  hasElemID: HasElemIDFunc,
+): Promise<InstanceElement | undefined> => {
+  if (isWriteResponseError(writeResponse)
+    && (writeResponse.status.statusDetail[0].code === INSUFFICIENT_PERMISSION_ERROR)) {
```
dantal4, line 102: nit- remove the brackets

packages/netsuite-adapter/src/constants.ts
```diff
@@ -91,3 +98,8 @@ export const WEB_SITE_HOSTING_FILES_FOLDER_NAME = 'Web Site Hosting Files'
 
 export const ACCOUNT_SPECIFIC_VALUE = '[ACCOUNT_SPECIFIC_VALUE]'
 export const NOT_YET_SUPPORTED_VALUE = '[NOT_YET_SUPPORTED]'
+
+// SuiteApp instance fields
+export const PLATFORM_CORE_CUSTOM_FIELD = 'platformCore:customField'
+export const PLATFORM_CORE_NULL_FIELD_LIST = 'platformCore:nullFieldList'
+export const PLATFORM_CORE_NAME = 'platformCore:name'
```
dantal4, line 105: better to put it in `client/constants.ts` IMO

dantal4 (commented):
Nice work!

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
```diff
@@ -517,28 +544,74 @@ export default class SoapClient {
...
-        'xmlns:platformCore': SOAP_CORE_URN,
-      },
-      record: await awu(instances).map(
-        async instance => this.convertToSoapRecord(instance.value, await instance.getType())
-      ).toArray(),
-    }
-    return this.runDeployAction(instances, body, 'updateList')
+  private async redeployLockedFieldsWithRetry(
+    retriesLeft: number,
+    prevWriteResponseList: WriteResponse[],
```
dantal4, line 549: Consider working everywhere with a map of `instance.elemID.getFullName()` to `WriteResponse` - I think that it will make it easier and readable to make the connection between them instead of using a list and matching the index.

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
shoshana-houri: Oh, I thought you were asking something different :)
We technically can, but that seems too risky... do we really want to redeploy again and again without any limitation?

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
shoshana-houri: That's a great idea, thanks! :)

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/filter_uneditable_locked_field.ts
shoshana-houri: From all I've tested - it doesn't change the value to null:
* Not in Netsuite - if it would have tried to change it we would have been receiving 'insufficient_permission' error again and again. Also, when we fetch the environment again - we receive the original value from Netsuite.
* Nor in the state files  - It's assumed that the user's original change have been accepted since we report a full success.

I repeated here the behavior of [data_instances_null_fields.ts](https://github.com/salto-io/salto/blob/73f2840de47b4c01d207bdb5390fe32833f42586/packages/netsuite-adapter/src/filters/data_instances_null_fields.ts#L34C1-L43), thought it'll be best to stick to a certain behavior, even if probably both methods work. WDYT?

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
dantal4: The limitation is when the error is not on a locked field, or when the locked field was already removed. Right?

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/filter_uneditable_locked_field.ts
dantal4: In data_instances_null_fields.ts we do want to change the value in Netsuite to null - in case that a field was deleted.
Anyway, If in Netsuite UI the value doesn't change it's good 👌

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
```diff
@@ -517,28 +543,87 @@ export default class SoapClient {
...
+
+    if (modifiedInstances.length > 0) {
+      await this.redeployLockedFieldsWithRetry(
+        LOCKED_FIELDS_MAX_REDEPLOYS,
+        modifiedInstances,
+        fullNameToWriteResponse,
+        action,
+        hasElemID
+      )
     }
```
dantal4, line 607: nit- this should be the same, with less duplicated code:
```
const fullNameToWriteResponse = new Map<string, WriteResponse>()
await this.redeployLockedFieldsWithRetry(
  LOCKED_FIELDS_MAX_REDEPLOYS,
  instances,
  fullNameToWriteResponse,
  action,
  hasElemID
)
```

If you choose to do it like that, I would move the log in line 558 to be in line 572.

packages/netsuite-adapter/src/client/suiteapp_client/constants.ts
```diff
@@ -15,3 +15,15 @@
...
+export const ECONN_ERROR = 'ECONN'
+export const UNEXPECTED_ERROR = 'UNEXPECTED_ERROR'
+export const INSUFFICIENT_PERMISSION_ERROR = 'INSUFFICIENT_PERMISSION'
+export const VALIDATION_ERROR = 'VALIDATION_ERROR'
+
+// SuiteApp instance fields
+export const PLATFORM_CORE_CUSTOM_FIELD = 'platformCore:customField'
+export const PLATFORM_CORE_NULL_FIELD_LIST = 'platformCore:nullFieldList'
+export const PLATFORM_CORE_NAME = 'platformCore:name'
+export const CUSTOM_FIELD = 'customField'
```
dantal4, line 29: maybe it's better to put `CUSTOM_FIELD` in `src/constants.ts` near CUSTOM_FIELD_LIST

packages/netsuite-adapter/test/client/suiteapp_client/soap_client/filter_uneditable_locked_field.test.ts
```diff
@@ -0,0 +1,266 @@
...
+          },
+        ],
+      },
+    }
+    const elementsSource = buildElementsSourceFromElements([])
+
+    const areModified = await awu(allInstances).map(instance =>
+      removeUneditableLockedField(instance, error, elementsSource.has)).toArray()
+
+    expect(areModified).toEqual([false, false, false, false])
```
dantal4, line 264: here too

dantal4 (commented):
Nice work!

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
dantal4: After talking f2f, I'm convinced that having the tries limitation can be helpful 👍

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/filter_uneditable_locked_field.ts
dantal4: @shoshana-houri  As we said f2f, I think that if it works without putting the removed fields in the NULL list, then it is preferred 👍

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/filter_uneditable_locked_field.ts
```diff
@@ -0,0 +1,110 @@
...
+  return fieldElemID !== undefined && !(await hasElemID(fieldElemID))
+}
+
+const removeCustomField = (
+  instance: InstanceElement,
+  fieldName: string
+): boolean => {
+  const customFields = instance.value[CUSTOM_FIELD_LIST]
+  if (customFields === undefined) {
+    return false
```
dantal4, line 69: maybe add a log here too? because it is weird that we'll get an INSUFFICIENT_PERMISSION_ERROR on a field, when there are no custom fields in the element.

PR #4313, opened 2023-05-01
SALTO 4086: Fix async code bug in smart values

IdoZakk:
Internal code bug that seems without influence but is risky
---

_Release Notes_: 
None
---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/automation/smart_values/smart_value_reference_filter.ts
+++ b/packages/jira-adapter/src/filters/automation/smart_values/smart_value_reference_filter.ts
@@ -165,34 +165,38 @@ const filterCreator: FilterCreator = ({ config }) => {
     },
 
     preDeploy: async (changes: Change<InstanceElement>[]) => {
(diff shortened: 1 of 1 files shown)


packages/jira-adapter/src/filters/automation/smart_values/smart_value_reference_filter.ts
```diff
@@ -165,8 +165,8 @@ const filterCreator: FilterCreator = ({ config }) => {
     },
 
     preDeploy: async (changes: Change<InstanceElement>[]) => {
-      filterAutomations(changes.map(getChangeData)).filter(isInstanceElement).flatMap(
-        async instance => getPossibleSmartValues(instance).forEach(({ obj, key }) => {
+      filterAutomations(changes.map(getChangeData)).filter(isInstanceElement).forEach(
```
alonstern, line 168: We have the same issue in onDeploy

alonstern (approved):
👌👌

PR #4314, opened 2023-05-01
SALTO-4073 Modified guide_disabled_validator

idoamit8:
__In this PR__ 
1. I modified the guide_disabled validator that validates that the user Cannot add instance when its associated brand has help center disabled or its associated brand is not include in the config. I found a great function that checks both in `utils` so I changed a lot in this CV. 
2. I added test for this CV.

---

In Order to test the code manually - do as follows:
1. Create two environments in Zendesk.
2. Check\create two brands with the same name. 
3. Make sure that one brand has help center enabled and one brand has help center disabled. 
Or make sure the one brand is part of the config and the other one isn't. 
5. clone a sub guide element from the brand in the env with help center enabled to the env with help center disabled. 
Example `salto element clone zendesk.article.instance.Sample_article__Stellar_Skyonomy_refund_policies_Announcements_General_My_company_s_uum_uuuum@sfssssuuuuuum -e prod --to-envs sandbox`
6. try to deploy the new env. 

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/change_validators/guide_disabled.ts
+++ b/packages/zendesk-adapter/src/change_validators/guide_disabled.ts
@@ -14,7 +14,7 @@
 * limitations under the License.
 */
 
-import { ChangeValidator, isInstanceChange, getChangeData, isAdditionChange, isInstanceElement, isReferenceExpression } from '@salto-io/adapter-api'
+import { ChangeValidator, isInstanceChange, getChangeData, isAdditionChange, isInstanceElement, isReferenceExpression, InstanceElement } from '@salto-io/adapter-api'
 import _ from 'lodash'
 import { collections } from '@salto-io/lowerdash'
 import { logger } from '@salto-io/logging'
@@ -24,6 +24,13 @@ import { GUIDE_TYPES_TO_HANDLE_BY_BRAND, ZendeskFetchConfig } from '../config'
 
 const log = logger(module)
 const { awu } = collections.asynciterable
+
+const isBrandWithHelpCenterFalse = (instance: InstanceElement, brandByBrandId: Record<string, InstanceElement>):
+boolean => {
+  const brandRef = instance.value.brand
+  const brand = brandByBrandId[brandRef.elemID.getFullName()]
+  return brand.value.has_help_center === false
+}
 
 export const guideDisabledValidator: (fetchConfig: ZendeskFetchConfig)
  => ChangeValidator = fetchConfig => async (changes, elementSource) => {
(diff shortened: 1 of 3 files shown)


packages/zendesk-adapter/src/change_validators/guide_disabled.ts
```diff
@@ -14,72 +14,52 @@
...
+ => ChangeValidator = fetchConfig => async (changes, elementSource) => {
+   const relevantInstances = changes
+     .filter(isInstanceChange)
+     .filter(isAdditionChange)
+     .map(getChangeData)
+     .filter(instance => GUIDE_TYPES_TO_HANDLE_BY_BRAND.includes(instance.elemID.typeName))
+   if (_.isEmpty(relevantInstances)) {
+     return []
+   }
+   if (elementSource === undefined) {
```
edenhassid, line 38: maybe move this if to be before `relevantInstances`

packages/zendesk-adapter/src/change_validators/guide_disabled.ts
```diff
@@ -14,72 +14,52 @@
...
-    .toArray())
-    .map(instance => [instance.elemID.getFullName(), instance]))
-  if (_.isEmpty(BrandsByBrandId)) {
-    return []
-  }
+   const brandsWithGuide = new Set(getBrandsForGuide(brandsInstances, fetchConfig)
+     .map(brand => brand.elemID.getFullName()))
 
-  const brandsWithoutGuideByInstanceId = getBrandsWithoutGuideByInstanceId(relevantInstances, BrandsByBrandId)
+   if (brandsWithGuide === undefined) {
```
edenhassid, line 52: I think you should also check if it is empty

edenhassid (approved):
looks good! added small comments :)

packages/zendesk-adapter/src/change_validators/guide_disabled.ts
```diff
@@ -14,72 +14,52 @@
...
-    }))
-}
+   return relevantInstances
+     .filter(instance => isReferenceExpression(instance.value.brand))
+     .filter(instance => !brandsWithGuide.has(instance.value.brand.elemID.getFullName()))
+     .map(instance => ({
+       elemID: instance.elemID,
+       severity: 'Error',
+       message: 'Cannot add this element because help center is not enabled for its associated brand, or the brand itself may not be enabled in the configuration.',
+       detailedMessage: `please enable help center for brand "${instance.value.brand.elemID.name}" or enable the brand in the configuration in order to add this element.`,
```
shir-reifenberg, line 63: nit:
```suggestion
       detailedMessage: `Please enable help center for brand "${instance.value.brand.elemID.name}" or enable the brand in the configuration in order to add this element.`,
```

also, I think we should be more clear about the reason the element is not deployable.. we can have a different message for each scenario because it requires 2 different actions from the user (enable brand in guide or change brand config in salto)

packages/zendesk-adapter/src/change_validators/guide_disabled.ts
edenhassid: make sure you do that :)

packages/zendesk-adapter/src/change_validators/guide_disabled.ts
```diff
@@ -14,72 +14,72 @@
...
-      message: 'Cannot add this element because help center is not enabled for its associated brand.',
-      detailedMessage: `please enable help center for brand "${brandsWithoutGuideByInstanceId[instance.elemID.getFullName()].elemID.name}" in order to add this element.`,
-    }))
-}
+   if (brandsWithGuide === undefined) {
+     return []
+   }
+
+   const allErrorInstances = relevantInstances
+     .filter(instance => isReferenceExpression(instance.value.brand))
```
edenhassid, line 65: I think you should add a log if the brand is not a reference expression

edenhassid (commented):
looks good! added a few comments :)

packages/zendesk-adapter/src/change_validators/guide_disabled.ts
idoamit8: This variable doesn't exist anymore. I checked it for the new one.

edenhassid (approved):

PR #4317, opened 2023-05-02
SALTO-4091: minor fix for context options

IdoZakk:
Minor fix

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/fields/context_options.ts
+++ b/packages/jira-adapter/src/filters/fields/context_options.ts
@@ -96,6 +96,12 @@ type UpdateContextOptionsParams = {
   baseUrl: string
   contextChange: ModificationChange<InstanceElement> | AdditionChange<InstanceElement>
 }
+const getFieldName = (
+  contextChange: AdditionChange<InstanceElement> | ModificationChange<InstanceElement>,
+  newOption: Value
+): Value =>
+  Object.entries(contextChange.data.after.value.options)
+    .find(([key, _value]) => contextChange.data.after.value.options[key].value === newOption)?.[0]
 
 const updateContextOptions = async ({
   addedOptions,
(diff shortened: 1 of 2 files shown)


packages/jira-adapter/src/filters/fields/context_options.ts
```diff
@@ -125,7 +131,9 @@ const updateContextOptions = async ({
           idToOption[newOption.optionId]
             .cascadingOptions[naclCase(newOption.value)].id = newOption.id
         } else {
-          contextChange.data.after.value.options[naclCase(newOption.value)].id = newOption.id
+          (contextChange.data.after.value.options[naclCase(newOption.value)]
```
alonstern, line 134: Will it work if we switch the values of two options? I wonder if we should always use `getFieldName`
Maybe we should just re-map the options like
```
const optionsMap = _(contextChange.data.after.value.options).values().keyBy(option => naclCase(option.value)).value()
```
and use `optionsMap` instead of `contextChange.data.after.value.options`

packages/jira-adapter/src/filters/fields/context_options.ts
```diff
@@ -96,6 +96,12 @@ type UpdateContextOptionsParams = {
   baseUrl: string
   contextChange: ModificationChange<InstanceElement> | AdditionChange<InstanceElement>
 }
+const getFieldName = (
+  contextChange: AdditionChange<InstanceElement> | ModificationChange<InstanceElement>,
+  newOption: Value
+): Value =>
+  Object.entries(contextChange.data.after.value.options)
+    .find(([key, _value]) => contextChange.data.after.value.options[key].value === newOption)?.[0]
```
alonstern, line 104: You use only the key so you can do `Object.keys` instead of `Object.entries`

packages/jira-adapter/src/filters/fields/context_options.ts
IdoZakk: Wow, that is a real corner case.....
The deployment works, and so are the following fetches.
We might switch ids in such a case or have an item without an id, we already have other places with problematic state after deployment, and I am not sure it is worth the time and effort to fully understand it (as the initial bug was minor as well and I only fixed it because it was easy.....)
WDYT?

alonstern: It sounds to me too like it's going to switch the ids. I agree it isn't worth much time but it sounds like doing the `optionsMap` should be a small change here, no?

IdoZakk: You are right, and it is much more elegant as well

alonstern (approved):
👌

PR #4318, opened 2023-05-02
SALTO-4083/standlone field bug

guyhamo:
fixed bug causing us to look at the standalone field type transformation and not the parent when writing the standalone instance.
causing us to ignore the parent's  "nestStandaloneInstances: false" and nesting the instance anyway.

---

I added a test to simulate a case with true default nestStandaloneInstances and false on specific type.

---
_Release Notes_: 

---
_User Notifications_:


packages/adapter-components/test/elements/swagger/instance_elements.test.ts
```diff
@@ -464,7 +540,8 @@ describe('swagger_instance_elements', () => {
             food2: { id: 'f2' },
           },
         },
-        [ADAPTER_NAME, 'Records', 'Pet', 'dog'],
+        // path has no effect on equality
+        [],
```
alonstern, line 544: Can you check separately that the path is correct?
`expect(dogO1Inst.path).toEqual(...)`

alonstern (approved):
👌

PR #4322, opened 2023-05-02
SALTO-3940: add template expressions to scriptRunner fields

IdoZakk:
Add template expression for scriptRunner's script fields.
As a first phase, we only reference custom fields, according to customers' need
This PR covers both DC and cloud, which have different structures.

---
_Release Notes_: 
Jira Adapter:
* Added support for custom field references inside ScriptRunner's scripts in cloud and DC

---
_User Notifications_: 
None


packages/jira-adapter/src/filters/script_runner/workflow_script_references.ts
```diff
@@ -0,0 +1,181 @@
...
+): void => {
+  try {
+    replaceTemplatesWithValues(
+      { values: [value], fieldName },
+      originalInstances,
+      prepRef,
+    )
+  } catch (e) {
+    log.error(`Error parsing templates in deployment during template removal for fieldName ${fieldName}`, e)
+  }
```
IdoZakk, line 47: I added the try and catch here and in the restore, as they appeared in the smart values.
Is it really needed? Do we have a suspicion it will fail sometimes?

packages/jira-adapter/src/filters/automation/smart_values/smart_value_reference_filter.ts
```diff
@@ -136,7 +136,7 @@ const replaceFormulasWithTemplates = async (instances: InstanceElement[])
   return ambiguousTokensWarnings
 }
 
-const prepRef = (part: ReferenceExpression): TemplatePart => {
+export const prepRef = (part: ReferenceExpression): TemplatePart => {
```
alonstern, line 139: This is meant to be specific for smart values,  I am not sure we would want to export it for general usage. In your case, you only have top-level references so all you need is the top-level part.

packages/jira-adapter/src/filters/script_runner/workflow_script_references.ts
```diff
@@ -0,0 +1,181 @@
...
+    return WALK_NEXT_STEP.SKIP
+  }
+  return WALK_NEXT_STEP.RECURSE
+}
+
+// This filter is used to add and remove template expressions in scriptRunner scripts
+const filter: FilterCreator = ({ config, client }) => {
+  const deployTemplateMapping: Record<string, TemplateExpression> = {}
+  return {
+    name: 'scriptRunnerWorkflowScriptReferencesFilter',
```
alonstern, line 124: I think nothing here should be specifically related to workflows and all the logic to find the scripts that is specific to workflow should be in walk on script functions. We would want this to run on any scriptrunner script and not only workflow ones

packages/jira-adapter/src/filters/script_runner/workflow_script_references.ts
```diff
@@ -0,0 +1,181 @@
...
+
+const addTemplateReferences = (fieldInstancesById: Map<string, InstanceElement>)
+  : referenceFunc => (value: Value, fieldName: string): void => {
+  value[fieldName] = referenceCustomFields(value[fieldName], fieldInstancesById)
+}
+
+const isCloudScriptRunnerItem = (value: Value): boolean =>
+  SCRIPT_RUNNER_CLOUD_TYPES.includes(value.type) && value.configuration.scriptRunner !== undefined
+
+const walkOnCloudScripts = (func: (value: string, fieldName: string) => void)
```
alonstern, line 85: nit: maybe move the `walkOnCloudScripts` and `walkOnDcScripts` to a different file?
Also maybe create a `walkOnScript` function that will wrap the two and will call the right one?

packages/jira-adapter/src/filters/script_runner/workflow_script_references.ts
```diff
@@ -0,0 +1,181 @@
...
+    },
+    preDeploy: async changes => {
+      if (!config.fetch.enableScriptRunnerAddon) {
+        return
+      }
+      changes
+        .filter(isAdditionOrModificationChange)
+        .filter(isInstanceChange)
+        .map(getChangeData)
+        .filter(instance => instance.elemID.typeName === WORKFLOW_TYPE_NAME)
```
alonstern, line 153: Can you do this inside walk on DC/Cloud scripts function?

packages/jira-adapter/src/filters/script_runner/workflow_script_references.ts
alonstern: Not sure it is needed here

alonstern (commented):
🔥

packages/jira-adapter/src/filters/script_runner/workflow_script_references.ts
IdoZakk: 90% of the code is workflow specific. I can rename the file and the header, and move the workflow logic from the filter to a dedicated function, I do not think it will make a difference for the future

packages/jira-adapter/src/filters/script_runner/workflow_script_references.ts
IdoZakk: I prefer not to, to reduce the calls to the recursive function and limit them to the relevant sections

IdoZakk:
I have split the original file into two and changed some stuff on both ends. Sorry, should have committed after the split....

packages/jira-adapter/src/filters/script_runner/workflow_script_references.ts
```diff
@@ -0,0 +1,84 @@
...
+      if (value.configuration[fieldName]?.script !== undefined) {
+        func(value.configuration[fieldName], 'script')
+      }
+    })
+    return WALK_NEXT_STEP.SKIP
+  }
+  return WALK_NEXT_STEP.RECURSE
+}
+
+export const walkOnScripts = (
```
alonstern, line 67: Couple of nits:
- The content of the file is not specific really related specifically to references
- I think when we support more types we would want this function to cover them too right? If so we would probably don't want "workflow" in the file name

So I think we should call the files something like `walk_on_scripts.ts` or `scripts_finder.ts`

alonstern (approved):
Looks great 👌

packages/jira-adapter/src/filters/script_runner/workflow_script_references.ts
IdoZakk: It is specific to workflows, the types, the structure. If we'll need others we will write a different walk. 
I will change the name to `walk_on_scripts` but I will create a workflow folder (under scriptRunner) and add it there

alonstern: The implementation is specific to workflows (since we only support workflows currently), but the interface is general. If we would write another walk function for another type we would probably end up with the same interface. Separating it to multiple walk functions will mean anyone who want to go over the scripts will have make multiple calls to each one. It sounds to me like the same scenario of `walkOnUsers` in which we created one function that covers all the types.
Anyway we can continue this discussion on the next type we support

PR #4323, opened 2023-05-03
SALTO-3953: Validate users exists in target env validator for Okta

shir-reifenberg:
Add missing users validator for Okta

---

Similar to what we have in Zendesk & Jira

---
_Release Notes_: 
None

---

_User Notifications_: 
None


edenhassid (approved):
LGTM! 💯 🔥

PR #4324, opened 2023-05-03
SALTO-4095: Add CheckDeploymentBasedOnConfigValidator to Okta

shir-reifenberg:
Add CheckDeploymentBasedOnConfigValidator to Okta

---

I merged the swagger types config with the ducktype types config so the validator would work on both

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/deployment/change_validators/check_deployment_based_on_config.ts
+++ b/packages/adapter-components/src/deployment/change_validators/check_deployment_based_on_config.ts
@@ -18,7 +18,7 @@ import { Change, ChangeError, ChangeValidator, getChangeData, Element, ElemID, i
   isReferenceExpression, ActionName } from '@salto-io/adapter-api'
 import { getParents } from '@salto-io/adapter-utils'
 import { collections } from '@salto-io/lowerdash'
-import { AdapterApiConfig, DeploymentRequestsByAction, DeployRequestConfig } from '../../config'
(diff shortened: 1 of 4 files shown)


packages/adapter-components/src/deployment/change_validators/check_deployment_based_on_config.ts
```diff
@@ -61,7 +61,7 @@ export const createCheckDeploymentBasedOnConfigValidator = ({
         return []
       }
       const getChangeErrorsByTypeName = (typeName: string): ChangeError[] => {
-        const typeConfig = apiConfig?.types?.[typeName]?.deployRequests ?? {}
+        const typeConfig = typesConfig[typeName]?.deployRequests ?? {}
```
netama, line 64: nit (for the existing implementation so not blocking this PR) - I think there might theoretically be cases where we define default deploy requests (under `typeDefaults`), so we should have merged it with default as we do in other places. but since when this is added it will be caught and lead to here, I guess we can keep it this way for now

netama (approved):
LGTM. @alonstern do we want this in jira as well? or is there a replacement?

packages/adapter-components/src/deployment/change_validators/check_deployment_based_on_config.ts
shir-reifenberg: leaving it to next time :)

PR #4325, opened 2023-05-03
SALTO-4092: Fix bug causing removals of instances with templates containing missing refs to fail

shir-reifenberg:
Fix bug causing removals of instances with templates containing missing refs to fail

---

What happened is that when trying to resolve references for templates with missing refs during pre deploy, we got an error because the reference wasn't resolved
To fix this I filtered  the changes for the `TemplateExpression` filter to be addition of modification changes only, cause there's no point to resolve templates for removals changes (Okta ignores data on `delete` method)

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/filters/expression_language.ts
+++ b/packages/okta-adapter/src/filters/expression_language.ts
@@ -14,7 +14,7 @@
 * limitations under the License.
 */
 import _ from 'lodash'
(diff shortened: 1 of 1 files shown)


netama (approved):
LGTM since it looks like for this flow in okta we really don't resolve any of the `before` values, and don't have `fullValue` references into instances with template expressions (which was the problematic scenario in zendesk). if we run into this in the future we can use the other solution

PR #4327, opened 2023-05-03
Refactor SDF dependencies graph

dantal4:
- Remove the `graph.key` attribute and use `node.id` instead.
- Use the graph as the only SOT in SDF deploy iterations (instead of using also `changesToDeploy`)

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/netsuite-adapter/src/client/client.ts
+++ b/packages/netsuite-adapter/src/client/client.ts
@@ -27,14 +27,14 @@ import SuiteAppClient from './suiteapp_client/suiteapp_client'
 import { createSuiteAppFileCabinetOperations, SuiteAppFileCabinetOperations, DeployType } from './suiteapp_client/suiteapp_file_cabinet'
 import { ConfigRecord, SavedSearchQuery, SystemInformation } from './suiteapp_client/types'
 import { CustomRecordTypeRecords, RecordValue } from './suiteapp_client/soap_client/types'
-import { FeaturesMap, GetCustomObjectsResult, getDeployableChanges, getOrTransformCustomRecordTypeToInstance, ImportFileCabinetResult, ManifestDependencies } from './types'
+import { FeaturesMap, getChangeNodeId, GetCustomObjectsResult, getDeployableChanges, getElemIdNodeId, getOrTransformCustomRecordTypeToInstance, ImportFileCabinetResult, ManifestDependencies, SDFObjectNode } from './types'
 import { toCustomizationInfo } from '../transformer'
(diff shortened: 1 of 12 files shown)


packages/netsuite-adapter/src/client/client.ts
```diff
@@ -334,23 +334,23 @@ export default class NetsuiteClient {
...
           getChangesElemIdsToRemove(error, dependencyMap, changesToDeploy),
           dependencyGraph
         )
-        elemIdsToRemove.forEach(elemIdName => dependencyGraph.removeNode(elemIdName))
-        const removedChanges = _.remove(
-          changesToDeploy,
-          change => elemIdsToRemove.has(getChangeData(change).elemID.getFullName())
-        )
-        if (removedChanges.length === 0) {
+        const numOfDeployedNodes = dependencyGraph.nodes.size
```
omrilit, line 341: I found it a bit confusing that this variable called `numOfDeployedNodes` as it felt to me like this is the number of elements that succeeded to be deployed.
Maybe better to rename to something like:
1. numOfRequestedNodesToDeploy
2. numOfAttemptedNodesToDeploy
3. any other name that indicates that it's not only the successful ones

packages/netsuite-adapter/src/client/types.ts
```diff
@@ -107,15 +107,29 @@ export class InvalidSuiteAppCredentialsError extends Error {
...
+type DeployableChange = Change<ObjectType | InstanceElement>
+
+export type SDFObjectNode = {
+  change: DeployableChange
+  serviceid: string
+  changeType: 'addition' | 'modification'
+  customizationInfo: CustomizationInfo
+}
+
+export const getElemIdNodeId = (elemId: ElemID): string => elemId.getFullName()
```
omrilit, line 119: nit: not sure if I am in favor or against it but we have lots of places in our code that use `elemId.getFullName()` as the identifier of an elemID but we chose not to have an explicit method that wraps it.
Not sure whether what's better TBH, I would prefer the implicit way, but maybe it's just because I am used to it and your way is clearer.

packages/netsuite-adapter/src/client/types.ts
```diff
@@ -107,15 +107,29 @@ export class InvalidSuiteAppCredentialsError extends Error {
...
+  getElemIdNodeId(getChangeData(change).elemID)
+
 export const getDeployableChanges = (
   changes: ReadonlyArray<Change>
-): Change<ObjectType | InstanceElement>[] =>
+): DeployableChange[] =>
   changes.filter(
     change => isInstanceChange(change) || isObjectTypeChange(change)
-  ) as Change<InstanceElement | ObjectType>[]
+  ) as DeployableChange[]
```
omrilit, line 129: super nit: you can use a type-guard here for isDeployableChange but if you don't think that it will be reused in other places, it may be redundant

omrilit (commented):
Great work!
Had mainly nits, approving to save a cycle

packages/netsuite-adapter/src/client/types.ts
dantal4: I wanted the node id creation to be transparent for `client.ts` so that's why I created this function and `getChangeNodeId`.

packages/netsuite-adapter/src/client/types.ts
dantal4: Skipping for now 🙏

omrilit (approved):
🔥

PR #4328, opened 2023-05-03
SALTO-4009: Source App reference

shir-reifenberg:
Add reference from `Group` of type `APP_GROUP` to `Application`

---

field `source` is missing from `Group` schema in swagger, so I used `cloneFrom` to create `Group__source` type

---
_Release Notes_: 
None

---

_User Notifications_: 
None


packages/okta-adapter/src/config.ts
```diff
@@ -266,8 +266,8 @@ const DEFAULT_TYPE_CUSTOMIZATIONS: OktaSwaggerApiConfig['types'] = {
   Group: {
     transformation: {
       fieldTypeOverrides: [
-        { fieldName: 'apps', fieldType: 'list<Application>' },
```
shir-reifenberg, line 269: not relevant anymore

PR #4329, opened 2023-05-03
Saas 6976 - Only allow deployment errors of type SaltoError or SaltoElementError

benisraeladi:
Alter the way we return errors from deployments, so that instead of returning one bulk error, we return multiple errors with their errors and ElemIDs (if relevant)

---

This is the beginnings of a PR to separate errors out in deployment. The way to look at this is as so:
commit 1: changes to core, adapter-api, and adapter-utils to allow this change
commit 2: Salesforce specific changes to make the adapter work

---

_Release Notes_: 
Core:
- Receive ElemIDs in deploy errors related to specific elements
Salesforce Adapter:
- Change deploy errors from `Error` type to `SaltoError` or `SaltoElementError`
---

_User Notifications_: _None_

### DIFF
--- a/packages/adapter-api/src/adapter.ts
+++ b/packages/adapter-api/src/adapter.ts
@@ -41,11 +41,11 @@ export type DeployExtraProperties = {
   groups?: Group[]
 }
 
-type SaltoDeploymentErrors = {
+type SaltoDeployErrors = {
   errors: ReadonlyArray<SaltoError | SaltoElementError>
 }
 
-type AdapterDeploymentErrors = {
+type AdapterDeployErrors = {
   errors: ReadonlyArray<SaltoError | SaltoElementError | Error>
 }
 
@@ -54,9 +54,9 @@ type BaseDeployResult = {
   extraProperties?: DeployExtraProperties
 }
 
-export type AdapterDeployResult = SaltoDeploymentErrors & BaseDeployResult
+export type AdapterDeployResult = SaltoDeployErrors & BaseDeployResult
 
-export type DeployResult = AdapterDeploymentErrors & BaseDeployResult
+export type DeployResult = AdapterDeployErrors & BaseDeployResult
 
 export type Progress = {
   message: string
--- a/packages/adapter-api/src/adapter.ts
+++ b/packages/adapter-api/src/adapter.ts
@@ -41,11 +41,22 @@ export type DeployExtraProperties = {
   groups?: Group[]
 }
 
-export type DeployResult = {
+type SaltoDeploymentErrors = {
+  errors: ReadonlyArray<SaltoError | SaltoElementError>
+}
+
+type AdapterDeploymentErrors = {
+  errors: ReadonlyArray<SaltoError | SaltoElementError | Error>
+}
+
+type BaseDeployResult = {
   appliedChanges: ReadonlyArray<Change>
-  errors: ReadonlyArray<Error>
   extraProperties?: DeployExtraProperties
 }
+
+export type AdapterDeployResult = SaltoDeploymentErrors & BaseDeployResult
+
+export type DeployResult = AdapterDeploymentErrors & BaseDeployResult
 
 export type Progress = {
   message: string
--- a/packages/cli/src/formatter.ts
+++ b/packages/cli/src/formatter.ts
@@ -27,6 +27,7 @@ import { Plan, PlanItem, FetchChange, FetchResult, LocalChange, getSupportedServ
 import { errors, SourceLocation, WorkspaceComponents, StateRecency } from '@salto-io/workspace'
 import { safeJsonStringify } from '@salto-io/adapter-utils'
 import { collections, values } from '@salto-io/lowerdash'
+import { DeployError } from '@salto-io/core/src/core/deploy'
 import Prompts from './prompts'
 
 const { awu } = collections.asynciterable
@@ -348,7 +349,21 @@ export const cancelDeployOutput = (checkOnly: boolean): string => [
   emptyLine(),
 ].join('\n')
 
-export const deployPhaseEpilogue = (numChanges: number, numErrors: number, checkOnly: boolean): string => {
+export const deployErrorsOutput = (deployErrors: DeployError[]): string => {
+  const errorOutputLines = [
+    emptyLine(),
+  ]
+  if (deployErrors.length > 0) {
+    errorOutputLines.push(error('Errors:'))
+    errorOutputLines.push(...deployErrors.map(deployError => error(`${deployError.message}`)))
+    errorOutputLines.push(emptyLine())
+    return errorOutputLines.join('\n')
+  }
+  return ''
+}
+
+export const deployPhaseEpilogue = (numChanges: number,
+  numErrors: number, checkOnly: boolean,): string => {
   const hadChanges = numChanges > 0
   const hadErrors = numErrors > 0
   if (hadChanges || hadErrors) {
--- a/packages/cli/src/formatter.ts
+++ b/packages/cli/src/formatter.ts
@@ -349,15 +349,30 @@ export const cancelDeployOutput = (checkOnly: boolean): string => [
   emptyLine(),
 ].join('\n')
 
-export const deployErrorsOutput = (deployErrors: DeployError[]): string => {
+export const deployErrorsOutput = (allErrors: DeployError[]): string => {
   const errorOutputLines = [
     emptyLine(),
   ]
-  if (deployErrors.length > 0) {
-    errorOutputLines.push(error('Errors:'))
-    errorOutputLines.push(...deployErrors.map(deployError => error(`${deployError.message}`)))
-    errorOutputLines.push(emptyLine())
-    return errorOutputLines.join('\n')
+  if (allErrors.length > 0) {
+    const warnings = allErrors.filter(warning => warning.severity === 'Warning')
+    const infos = allErrors.filter(info => info.severity === 'Info')
+    const deployErrors = allErrors.filter(err => err.severity === 'Error')
+    if (!_.isEmpty(deployErrors)) {
+      errorOutputLines.push(error('Errors:'))
+      errorOutputLines.push(...deployErrors.map(deployError => error(`${deployError.message}`)))
+      errorOutputLines.push(emptyLine())
+    }
+    if (!_.isEmpty(warnings)) {
+      errorOutputLines.push(warn('Warnings:'))
+      errorOutputLines.push(...warnings.map(warning => warn(`${warning.message}`)))
+      errorOutputLines.push(emptyLine())
+    }
+    if (!_.isEmpty(infos)) {
+      errorOutputLines.push(header('Info:'))
+      errorOutputLines.push(...infos.map(info => header(`${info.message}`)))
+      errorOutputLines.push(emptyLine())
+    }
+    return errorOutputLines.join(EOL)
   }
   return ''
 }
--- a/packages/core/src/core/deploy/deploy_actions.ts
+++ b/packages/core/src/core/deploy/deploy_actions.ts
@@ -49,10 +49,10 @@ const extractErrors = (
   result: DeployResult,
   changes: readonly Change<ChangeDataType>[]
 ): ReadonlyArray<SaltoElementError | SaltoError> =>
-  result.errors.map(error =>
+  result.errors.flatMap(error =>
     (error instanceof Error
       ? addElemIDsToError(changes, error)
-      : error)).flat()
+      : error))
 
 
 const deployOrValidate = async (
@@ -60,7 +60,7 @@ const deployOrValidate = async (
 ): Promise<AdapterDeployResult> => {
   try {
     const deployOrValidateFn = checkOnly ? adapter.validate : adapter.deploy
-    if (_.isUndefined(deployOrValidateFn)) {
+    if (deployOrValidateFn === undefined) {
       throw new Error(`${checkOnly ? 'Check-Only deployment' : 'Deployment'} is not supported in adapter ${adapterName}`)
     }
     const result = await deployOrValidateFn(opts)
@@ -182,10 +182,12 @@ export const deployActions = async (
         if (nodeError instanceof NodeSkippedError) {
           reportProgress(item, 'cancelled', deployPlan.getItem(nodeError.causingNode).groupKey)
           deployErrors.push(...[...item.changes()].map(change =>
-            ({ elemID: getChangeData(change).elemID,
+            ({
+              elemID: getChangeData(change).elemID,
               groupId: item.groupKey,
               message: `Element ${key} was not deployed, as it depends on element ${nodeError.causingNode} which failed to deploy`,
-              severity: 'Error' as SeverityLevel })))
+              severity: 'Error' as SeverityLevel,
+            })))
         } else if (nodeError instanceof WalkDeployError) {
           deployErrors.push(...nodeError.errors.map(deployError => ({ ...deployError, groupId: item.groupKey })))
         } else {
--- a/packages/core/src/core/deploy/deploy_actions.ts
+++ b/packages/core/src/core/deploy/deploy_actions.ts
@@ -17,8 +17,8 @@ import _ from 'lodash'
 import {
   AdapterOperations, getChangeData, Change,
   isAdditionOrModificationChange,
-  DeployExtraProperties, DeployOptions, Group, ElemID,
-  isSaltoElementError, SaltoElementError, SaltoError, SeverityLevel, AdapterDeployResult, DeployResult,
+  DeployExtraProperties, DeployOptions, Group,
+  isSaltoElementError, SaltoElementError, SaltoError, SeverityLevel, AdapterDeployResult, DeployResult, ChangeDataType,
 } from '@salto-io/adapter-api'
 import { detailedCompare, applyDetailedChanges } from '@salto-io/adapter-utils'
 import { WalkError, NodeSkippedError } from '@salto-io/dag'
@@ -35,38 +35,61 @@ type DeployOrValidateParams = {
   checkOnly: boolean
 }
 
-const extractErrors = (deployRes: DeployResult, opts: DeployOptions): ReadonlyArray<SaltoElementError | SaltoError> =>
-  (deployRes.errors.map(error => (
+const extractErrors = (result: DeployResult): ReadonlyArray<SaltoElementError | SaltoError> =>
+  (result.errors.map(error => (
     error instanceof Error
       ? { message: error.message,
-        severity: 'Error' as SeverityLevel,
-        group: opts.changeGroup.groupID }
+        severity: 'Error' as SeverityLevel }
       : error))
+  )
+
+const addElemIDsToError = (
+  changes: readonly Change<ChangeDataType>[], error: Error
+): ReadonlyArray<SaltoElementError> =>
+  (changes.map(change => (
+    { message: error.message,
+      severity: 'Error' as SeverityLevel,
+      elemID: getChangeData(change).elemID }
+  ))
   )
 
 const deployOrValidate = async (
   { adapter, adapterName, opts, checkOnly }: DeployOrValidateParams
 ): Promise<AdapterDeployResult> => {
   if (!checkOnly) {
-    const deployRes = await adapter.deploy(opts)
-    return {
-      appliedChanges: deployRes.appliedChanges,
-      extraProperties: deployRes.extraProperties,
-      errors: extractErrors(deployRes, opts),
+    try {
+      const deployRes = await adapter.deploy(opts)
+      return {
+        appliedChanges: deployRes.appliedChanges,
+        extraProperties: deployRes.extraProperties,
+        errors: extractErrors(deployRes),
+      }
+    } catch (error) {
+      return {
+        appliedChanges: [],
+        errors: addElemIDsToError(opts.changeGroup.changes, error as Error),
+      }
     }
   }
   if (_.isUndefined(adapter.validate)) {
     throw new Error(`Check-Only deployment is not supported in adapter ${adapterName}`)
   }
-  const validateRes = await adapter.validate(opts)
-  return {
-    appliedChanges: validateRes.appliedChanges,
-    extraProperties: validateRes.extraProperties,
-    errors: extractErrors(validateRes, opts),
+  try {
+    const validateRes = await adapter.validate(opts)
+    return {
+      appliedChanges: validateRes.appliedChanges,
+      extraProperties: validateRes.extraProperties,
+      errors: extractErrors(validateRes),
+    }
+  } catch (error) {
+    return {
+      appliedChanges: [],
+      errors: addElemIDsToError(opts.changeGroup.changes, error as Error),
+    }
   }
 }
 
-const deployAction = (
+const deployAction = async (
   planItem: PlanItem,
   adapters: Record<string, AdapterOperations>,
   checkOnly: boolean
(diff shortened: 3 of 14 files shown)


packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -121,11 +123,15 @@ export const deployActions = async (
           log.warn(
             'Failed to deploy %s, errors: %s',
             item.groupKey,
-            result.errors.map(err => err.stack ?? err.message).join('\n\n'),
-          )
-          throw new Error(
-            `Failed to ${checkOnly ? 'validate' : 'deploy'} ${item.groupKey} with errors:\n${result.errors.join('\n')}`
+            // result.errors.map(err => err.stack ?? err.message).join('\n\n'),
```
adamro, line 126: ?

packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -60,8 +60,10 @@ const deployAction = (
 }
 
 export class DeployError extends Error {
```
adamro, line 62: Is there a reason not to have Severity here and make this extend SaltoError?

packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -121,11 +123,15 @@ export const deployActions = async (
...
-            result.errors.map(err => err.stack ?? err.message).join('\n\n'),
-          )
-          throw new Error(
-            `Failed to ${checkOnly ? 'validate' : 'deploy'} ${item.groupKey} with errors:\n${result.errors.join('\n')}`
+            // result.errors.map(err => err.stack ?? err.message).join('\n\n'),
+            result.errors.map(err => err.message).join('\n\n'),
           )
+          const modifiedErrors: DeployError[] = []
+          result.errors.forEach(error => {
+            const deployError = isSaltoElementError(error) ? new DeployError(item.groupKey, error.message, error.elemID) : new DeployError(item.groupKey, error.message)
```
adamro, line 131: Why is this needed? If you made elementId optional in `DeployError` it would just work to do `new DeployError(item.groupKey, error.message, error.elemID)` in both cases, wouldn't it?

packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -121,11 +123,15 @@ export const deployActions = async (
...
             item.groupKey,
-            result.errors.map(err => err.stack ?? err.message).join('\n\n'),
-          )
-          throw new Error(
-            `Failed to ${checkOnly ? 'validate' : 'deploy'} ${item.groupKey} with errors:\n${result.errors.join('\n')}`
+            // result.errors.map(err => err.stack ?? err.message).join('\n\n'),
+            result.errors.map(err => err.message).join('\n\n'),
           )
+          const modifiedErrors: DeployError[] = []
+          result.errors.forEach(error => {
```
adamro, line 130: Why is this a forEach and not a map?

adamro (commented):
still early in the review but leaving the comments for now

packages/dag/src/nodemap.ts
```diff
@@ -71,21 +67,21 @@ export type NodeHandler = (id: NodeId) => void
...
 
-  set(nodeId: NodeId, value: Error, visited: Set<string> = new Set()): this {
+  set(nodeId: NodeId, errors: Error[], visited: Set<string> = new Set()): this {
     const idAsString = nodeId.toString()
     if (visited.has(idAsString)) {
       return this
     }
     visited.add(idAsString)
-    super.set(nodeId, value)
+    super.set(nodeId, errors)
```
adamro, line 82: Won't this override the errors even if this NodeId already had errors before? Is this really legit?

packages/adapter-api/src/error.ts
```diff
@@ -29,7 +29,7 @@ export type SaltoElementError = SaltoError & {
     elemID: ElemID
 }
 
-export const isSaltoElementError = (error: SaltoError | SaltoElementError):
+export const isSaltoElementError = (error: SaltoError | SaltoElementError | Error):
```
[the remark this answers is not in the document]
benisraeladi: This change is only temporary, the only reason `Error` is here is so that this will still compile for adapters that have not been touched yet

packages/core/src/core/deploy/deploy_actions.ts
benisraeladi: The problem with `SaltoError` is that it itself does not extend `Error`, meaning that it would be problematic throwing these errors (I tried and got an eslint error, and I tend to trust those unless proven otherwise)

packages/dag/src/nodemap.ts
```diff
@@ -71,21 +67,21 @@ export type NodeHandler = (id: NodeId) => void
...
 
-class WalkErrors<T> extends Map<NodeId, Error> {
+class WalkErrors<T> extends Map<NodeId, Error[]> {
   // eslint-disable-next-line no-use-before-define
   constructor(readonly nodeMap: DAG<T>) {
     super()
   }
 
-  set(nodeId: NodeId, value: Error, visited: Set<string> = new Set()): this {
+  set(nodeId: NodeId, errors: Error[], visited: Set<string> = new Set()): this {
```
adamro, line 76: All the calls to this function in the file still pass `Error` and not `Error[]` (cause it's in then and it's an any) :\

packages/core/src/core/deploy/deploy_actions.ts
adamro: If we always return them and not throw them, do they really need to extend Error tho?

packages/core/src/core/deploy/deploy_actions.ts
benisraeladi: We currently need to throw them because of `WalkErrors` and `WalkError` logic. Doing a deep-dive to decide on new format for error collection will also work

packages/adapter-utils/src/deploy.ts
```diff
@@ -28,7 +28,7 @@ export const deployInstance = async (
   if (!isInstanceChange(change)) {
     return {
       appliedChanges: [],
-      errors: [new Error('Only instance changes supported')],
+      errors: [{ elemID: getChangeData(change).elemID, message: 'Only instance changes supported', severity: 'Error' }],
```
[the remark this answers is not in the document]
benisraeladi: It loos like this file isn't used anywhere, I think I'll just delete it in a different PR and revert this change

.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
```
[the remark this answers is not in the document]
benisraeladi: No, should've been in `.ignore`. Will remove

packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -32,36 +34,56 @@ type DeployOrValidateParams = {
   checkOnly: boolean
 }
 
-const deployOrValidate = (
+const deployOrValidate = async (
   { adapter, adapterName, opts, checkOnly }: DeployOrValidateParams
-): Promise<DeployResult> => {
+): Promise<AdapterDeployResult> => {
```
adamro, line 39: I like this but I am not 100% this is legit :)
Let's discuss this f2f

packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -32,36 +34,56 @@ type DeployOrValidateParams = {
...
   if (!checkOnly) {
-    return adapter.deploy(opts)
+    const deployRes = await adapter.deploy(opts)
+    return {
+      appliedChanges: deployRes.appliedChanges,
+      extraProperties: deployRes.extraProperties,
+      errors: deployRes.errors.map(error => (
```
adamro, line 45: Probably best to DRY this part and write a function

packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -32,36 +34,56 @@ type DeployOrValidateParams = {
...
-): Promise<DeployResult> => {
+): Promise<AdapterDeployResult> => {
   const changes = [...planItem.changes()]
   const adapterName = getChangeData(changes[0]).elemID.adapter
   const adapter = adapters[adapterName]
   if (!adapter) {
     throw new Error(`Missing adapter for ${adapterName}`)
   }
   const opts = { changeGroup: { groupID: planItem.groupKey, changes } }
+  // TODOADI wrap deployOrValidate to make sure that the correct errors are returned
```
adamro, line 77: ?

packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -32,36 +34,56 @@ type DeployOrValidateParams = {
...
   return deployOrValidate({ adapter, adapterName, opts, checkOnly })
 }
 
 export class DeployError extends Error {
-  constructor(readonly elementId: string | string[], message: string) {
+  public readonly elemID?: ElemID
+  constructor(readonly groupId: string | string[], message: string, elementId?: ElemID) {
```
adamro, line 83: I think most of our code base has `elemID` and not `elementId`. You kept this for a reason?

Also, is there a reason for `groupId` to be typed like this and not be specific array or not?

packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -94,6 +116,14 @@ const updatePlanElement = (item: PlanItem, appliedChanges: ReadonlyArray<Change>
     })
 }
 
+class WalkDeployError extends Error {
+  public errors: ReadonlyArray<SaltoElementError | SaltoError>
+  constructor(deployErrors: ReadonlyArray<SaltoElementError | SaltoError>) {
```
adamro, line 121: I know some really hate that syntax but you can use the constructor `(public errors: ReadonlyArray<SaltoElementError | SaltoError>)` and avoid 2 lines :) (120 and 123)

packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -32,36 +34,56 @@ type DeployOrValidateParams = {
...
   return deployOrValidate({ adapter, adapterName, opts, checkOnly })
 }
 
 export class DeployError extends Error {
-  constructor(readonly elementId: string | string[], message: string) {
+  public readonly elemID?: ElemID
+  constructor(readonly groupId: string | string[], message: string, elementId?: ElemID) {
     super(message)
+    this.elemID = elementId
+    this.groupId = groupId
```
adamro, line 86: this is redundant isn't it? (can also be done for `elementId`)

packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -94,6 +116,14 @@ const updatePlanElement = (item: PlanItem, appliedChanges: ReadonlyArray<Change>
...
 
+class WalkDeployError extends Error {
+  public errors: ReadonlyArray<SaltoElementError | SaltoError>
+  constructor(deployErrors: ReadonlyArray<SaltoElementError | SaltoError>) {
+    super()
+    this.errors = deployErrors
+  }
+}
+
 export const deployActions = async (
```
adamro, line 127: I am wondering wether we want to keep `deployActions`'s return type (`DeployActionResult`) to include errors that actually extends `Error` or want a different, Salto defined type here as well
We anyway recreate everything here
This can be a `SaltoError | SaltoElementError` or we can define a `SaltoGroupError` that is the basic thing here and have ones with both Group and Element. Instead of this. not sure it adds anything or is very different, just a choice we can make cause we are changing this now and most of the Errors we return typed are `SaltoError` and not extension of `Error`

adamro (commented):
Not finished but went through most of the parts with real logic. Left comments but I think this direction is generally great!

packages/core/src/core/deploy/deploy_actions.ts
benisraeladi: That's only if we always expect `elemID` to exist, which can't happen because there are errors that don't relate to a specific elemID

packages/core/src/core/deploy/deploy_actions.ts
benisraeladi: In this part of the code originally `elementId` was actually the group ID. I changed the name for consistency, but kept the signature because I don't know how it'll impact the unaltered adapters/errors. I'll change it to `elemID` now that it's the correct type

packages/core/src/core/deploy/deploy_actions.ts
benisraeladi: oops

packages/core/src/core/deploy/deploy_actions.ts
benisraeladi: Oh, I think I misunderstood before. It's redundant and I'll remove it

packages/core/src/core/deploy/deploy_actions.ts
adamro: FYI @ori-moisis let's discuss this f2f the 3 of us

packages/core/src/core/deploy/deploy_actions.ts
benisraeladi: We decided to change the API and make it so that `DeployError` doesn't extend `Error` and instead is `SaltoError | SaltoElementError` with added `groupId`

packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -32,23 +34,42 @@ type DeployOrValidateParams = {
   checkOnly: boolean
 }
 
-const deployOrValidate = (
+const extractErrors = (deployRes: DeployResult, opts: DeployOptions): ReadonlyArray<SaltoElementError | SaltoError> =>
+  (deployRes.errors.map(error => (
+    error instanceof Error
+      ? { message: error.message,
+        severity: 'Error' as SeverityLevel,
+        group: opts.changeGroup.groupID }
```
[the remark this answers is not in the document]
adamro: We later said that the behaviour we do want is when the adapter throws an error we want to create an Error per element in the group, tied to the Element's elemID. And not allow the adapters to return Errors.

packages/core/src/core/deploy/deploy_actions.ts
adamro: Here we said we should go through with this change and return SaltoError or SaltoElementError but with groupId and not something that extends Error

packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -32,23 +34,65 @@ type DeployOrValidateParams = {
...
+  ))
+  )
+
+const deployOrValidate = async (
   { adapter, adapterName, opts, checkOnly }: DeployOrValidateParams
-): Promise<DeployResult> => {
+): Promise<AdapterDeployResult> => {
   if (!checkOnly) {
-    return adapter.deploy(opts)
+    try {
```
adamro, line 59: nit: Not completely your fault cause it was this way but the lack of DRY became a bit worse here.
Maybe have a 
`const deployOrValidateFn = checkOnly ? adapter.validate : adapter.deploy`
and then have the same code cause it's practically identical?

packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -142,16 +188,35 @@ export const deployActions = async (
         const item = deployPlan.getItem(key) as PlanItem
         if (nodeError instanceof NodeSkippedError) {
           reportProgress(item, 'cancelled', deployPlan.getItem(nodeError.causingNode).groupKey)
-          deployErrors.push(new DeployError(item.groupKey, `Element ${key} was not deployed, as it depends on element ${nodeError.causingNode} which failed to deploy`))
+          deployErrors.push({
+            groupId: item.groupKey,
+            message: `Element ${key} was not deployed, as it depends on element ${nodeError.causingNode} which failed to deploy`,
```
adamro, line 193: Shouldn't we also attach this to the Element's elemID?

packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -142,16 +188,35 @@ export const deployActions = async (
...
+            severity: 'Error' as SeverityLevel,
+          })
+        } else if (nodeError instanceof WalkDeployError) {
+          deployErrors.push(...nodeError.errors.map(
+            deployError => (isSaltoElementError(deployError)
```
adamro, line 198: Doesn't just adding the `groupId` to either work?

packages/cli/src/formatter.ts
```diff
@@ -348,7 +349,21 @@ export const cancelDeployOutput = (checkOnly: boolean): string => [
   emptyLine(),
 ].join('\n')
 
-export const deployPhaseEpilogue = (numChanges: number, numErrors: number, checkOnly: boolean): string => {
+export const deployErrorsOutput = (deployErrors: DeployError[]): string => {
+  const errorOutputLines = [
+    emptyLine(),
+  ]
+  if (deployErrors.length > 0) {
+    errorOutputLines.push(error('Errors:'))
```
adamro, line 357: We are ignoring the concept of Severity here?

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -193,7 +202,7 @@ export const retryFlow = async (
   const { retryDelay, retryableFailures } = client.dataRetry
 
   let successes: InstanceElement[] = []
-  let errMsgs: string[] = []
+  let errMsgs: (SaltoElementError | SaltoError)[] = []
```
adamro, line 205: name change to errors?

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -405,10 +414,12 @@ const deployModifyChanges = async (
...
-  const diffApiNameErrors = await awu(diffApiNameData).map(async data => new Error(`Failed to update as api name prev=${await apiName(
-    data.before
-  )} and new=${await apiName(data.after)} are different`)).toArray()
-  const errors = errorMessages.map(error => new Error(error)).concat(diffApiNameErrors)
+  const diffApiNameErrors: SaltoElementError[] = await awu(diffApiNameData).map(async data => ({
+    elemID: data.before.elemID,
+    message: `Failed to update as api name prev=${await apiName(data.before)} and new=${await apiName(data.after)} are different`,
+    severity: 'Error' as SeverityLevel,
+  })).toArray()
+  const errors: (SaltoElementError | SaltoError)[] = [...errorMessages, ...diffApiNameErrors]
```
adamro, line 422: `errorMessages` sounds like a wrong name now as well

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -66,12 +69,18 @@ ${safeJsonStringify(instance.value, undefined, 2,)}
   })
 )
 
-const getErrorMessagesFromInstAndResults = (instancesAndResults: InstanceAndResult[]): string[] =>
+const getErrorMessagesFromInstAndResults = (instancesAndResults: InstanceAndResult[]): SaltoElementError[] =>
   instancesAndResults
-    .map(({ instance, result }) => `${instance.elemID.name}:
-    \t${result.errors?.join('\n\t')}`)
+    .flatMap(({ instance, result }) => {
```
adamro, line 74: Our code base usually uses `filter` and then `map` instead of the if inside and the let

packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -195,37 +198,39 @@ const isUnFoundDelete = (message: DeployMessage, deletionsPackageName: string):
...
+      ),
+      severity: 'Error' as SeverityLevel,
+    }))
   const componentErrors = allFailureMessages
     .filter(failure => !isUnFoundDelete(failure, deletionsPackageName))
     .map(getUserFriendlyDeployMessage)
-    .map(failure => new Error(
-      `Failed to ${checkOnly ? 'validate' : 'deploy'} ${failure.fullName} with error: ${failure.problem} (${failure.problemType})`
+    .map(failure => (
+      { elemID: ElemID.fromFullName(failure.fullName), message: `${failure.problem} (${failure.problemType})`, severity: 'Error' as SeverityLevel }
```
adamro, line 222: How come we are removing the failed to XXX here?

adamro (commented):
We are very close :D

adamro (commented):
We are very close :D

packages/cli/src/formatter.ts
benisraeladi: Yes

adamro: Why? :)
I thought there's already a use-case of Warning/Info we thought of and we know we will use it. Why not differentiate between the cases in the CLI from the go?

packages/adapter-api/src/adapter.ts
```diff
@@ -41,12 +41,23 @@ export type DeployExtraProperties = {
   groups?: Group[]
 }
 
-export type DeployResult = {
+type SaltoDeployErrors = {
+  errors: ReadonlyArray<SaltoError | SaltoElementError>
+}
+
+type AdapterDeployErrors = {
```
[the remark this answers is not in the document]
benisraeladi: In the old scheme, adapters *do* return `Error` typed errors. The reason for `AdapterDeployErrors` is a gateway so that `deployOrValidate` can change all `Error` type errors to some sort of `SaltoError`. In the future, only `SaltoDeployErrors` will be used, and hopefully we will be able to delete `AdapterDeployErrors`

packages/salesforce-adapter/src/metadata_deploy.ts
benisraeladi: It felt redundant to me, but if you think it should stay I'll bring it back

packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -33,23 +35,53 @@ type DeployOrValidateParams = {
...
+      severity: 'Error' as SeverityLevel,
+      elemID: getChangeData(change).elemID }
+  ))
+  )
+
+const extractErrors = (
+  result: DeployResult,
+  changes: readonly Change<ChangeDataType>[]
+): ReadonlyArray<SaltoElementError | SaltoError> =>
+  result.errors.map(error =>
```
adamro, line 52: nit: result.errors.flatMap(...

adamro (commented):
I think we just need to decide if to keep the adapter API as is with the ability to return Error or not (not ---> Adding the code that transforms it to a SaltoElementError per Element to adapter-component/utils and have all the adapters use it instead of returning Error for now)

packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -33,23 +35,53 @@ type DeployOrValidateParams = {
...
-): Promise<DeployResult> => {
-  if (!checkOnly) {
-    return adapter.deploy(opts)
-  }
-  if (_.isUndefined(adapter.validate)) {
-    throw new Error(`Check-Only deployment is not supported in adapter ${adapterName}`)
+): Promise<AdapterDeployResult> => {
+  try {
+    const deployOrValidateFn = checkOnly ? adapter.validate : adapter.deploy
+    if (_.isUndefined(deployOrValidateFn)) {
```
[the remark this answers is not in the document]
benisraeladi: That's what was used in the function beforehand. Would the right alternative be deployOrValidateFn === undefined?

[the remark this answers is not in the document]
adamro: yep

adamro (approved):
🔥 
Great work on this! Surely not the most trivial area of our code and I think the end result is a huge jump in the right directly. Well executed!

adamro (approved):
🔥 
Great work on this! Surely not the most trivial area of our code and I think the end result is a huge jump in the right directly. Well executed!

PR #4333, opened 2023-05-04
SALTO-4107: add check for null on account id

IdoZakk:
Came from an incident, hard to verify this is the cause but it makes sense and recreates

---

None

---
_Release Notes_: 
Jira Adapter:
* Fixed a bug that caused null values to fail the fetch 

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/account_id/account_id_filter.ts
+++ b/packages/jira-adapter/src/filters/account_id/account_id_filter.ts
@@ -186,7 +186,7 @@ export const walkOnUsers = (callback: WalkOnUsersCallback, config: JiraConfig): 
         ? accountIdsScenarios(value.value, path, callback, config)
         : WALK_NEXT_STEP.EXIT
     }
-    if (value !== undefined) {
+    if (value != null) { // null or undefined
       return accountIdsScenarios(value, path, callback, config)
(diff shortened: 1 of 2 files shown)


alonstern (approved):
👌

packages/jira-adapter/src/filters/account_id/account_id_filter.ts
```diff
@@ -186,7 +186,7 @@ export const walkOnUsers = (callback: WalkOnUsersCallback, config: JiraConfig):
         ? accountIdsScenarios(value.value, path, callback, config)
         : WALK_NEXT_STEP.EXIT
     }
-    if (value !== undefined) {
+    if (value != null) { // null or undefined
```
netama, line 189: we shouldn't have `!=` in the code. can we be explicit?

netama: I see we have it in some other places as well... @ori-moisis wdyt?

ori-moisis: I don't know enough about it to say if this has any potential problems so I don't have a strong opinion either way on this

PR #4334, opened 2023-05-04
SALTO-4046 Add support for invalid static files for copying between accounts

adfineberg:
_Fix the dump of a Missing Static File which currently leads to a parse error_

---

---
_Release Notes_: 
_Core_
* Fix the dump of a Missing Static File which currently leads to a parse error

---
_User Notifications_:


ori-moisis:
in the release notes there should be a component header, should be
```
Core:
- Fixed issue where dump a Missing Static File to NaCl would create invalid syntax
```

ori-moisis (approved):

PR #4335, opened 2023-05-04
SALTO-4110: Provisioning features for apps cannot be managed

shir-reifenberg:
Add deployment annotations to notify users `features` field will not be deployed.

---

When enabling provisioning for app, `features` field is added to app instance. Okta doesn't support managing those settings and provisioning settings must be managed through the admin console. 
This doesn't fail the request so we don't need to remove it before deployment.

---
_Release Notes_: 

- Add support in Okta

_Okta adapter_:
- When trying to deploy provisioning features for apps, user will be notified the action is not supported

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/filters/app_deployment.ts
+++ b/packages/okta-adapter/src/filters/app_deployment.ts
@@ -15,7 +15,7 @@
 */
 import _ from 'lodash'
 import Joi from 'joi'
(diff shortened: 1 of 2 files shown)


netama (approved):

netama:
@shir-reifenberg I think it's time to start adding release notes for okta changes :) (and add a general one if we haven't yet)

PR #4336, opened 2023-05-04
SALTO-3911: Salesforce Data Instances Aliases

tamtamirr:
This PR adds Aliases to data instances in Salesforce.

---

Workspace Diff: https://github.com/salto-io/tamir-sf/commit/f75621e3281d311b5ac642b2d9bc9c8a9ac29c94 
Noise Suppression: https://github.com/salto-io/salto_private/pull/5596

---
_Release Notes_: 
Salesforce Adapter:
- Added Aliases to data records.

---
_User Notifications_: 
Salesforce:
- Upon the next fetch, the annotation **_alias** will be added to each data instance.


packages/salesforce-adapter/src/fetch_profile/data_management.ts
```diff
@@ -45,6 +85,7 @@ export const buildDataManagement = (params: DataManagementConfig): DataManagemen
         ?.find(override => new RegExp(`^${override.objectsRegex}$`).test(name))
       return matchedOverride?.idFields ?? params.saltoIDSettings.defaultIdFields
     },
+    getObjectAliasFields: name => ALIAS_FIELDS_BY_TYPE[name] ?? DEFAULT_ALIAS_FIELDS,
```
tamtamirr, line 88: In the future, we could easily make Aliases configurable if we'd want to.

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
```diff
@@ -226,13 +240,52 @@ const typesRecordsToInstances = async (
     return fullName
   }
 
+  const getRecordAlias = async (
```
tamtamirr, line 243: Made the calculation of the **Alias** almost identical to how we calculate the **Record Salto Name**. Notice these two small differences:
- We use different separator for Aliases (' ' instead of '___')
- We set the calculated aliases in a different Record.

tamtamirr (commented):
Added some comments. Code is ready for a brief review. Will add UTs.

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
```diff
@@ -353,33 +406,37 @@ const getParentFieldNames = (fields: Field[]): string[] =>
...
+  const idFieldsNames = dataManagement.getObjectIdsFields(typeName)
   const idFieldsWithParents = idFieldsNames.flatMap(fieldName =>
     ((fieldName === detectsParentsIndicator)
       ? getParentFieldNames(Object.values(type.fields)) : fieldName))
-  const invalidIdFieldNames = idFieldsWithParents.filter(fieldName => (
+  const aliasFieldNames = dataManagement.getObjectAliasFields(typeName)
+  const aliasFieldsWithParents = aliasFieldNames.flatMap(fieldName =>
+    ((fieldName === detectsParentsIndicator)
+      ? getParentFieldNames(Object.values(type.fields)) : fieldName))
+  const invalidIdFieldNames = idFieldsWithParents.concat(aliasFieldsWithParents).filter(fieldName => (
```
tamtamirr, line 418: Do we want to separate the handling of the invalid fields here? Does it make any sense?

tamtamirr: We probably do, considering the other-way around when an **alias field** is invalid we will lose the Instance. (It should not happen right now as the definitions of the alias fields are hard-coded so there shouldn't be user mistakes now)

Two options here:
- Add handling that is specific for **aliasFields** (meaning, adding **invalidAliasFields** to the returned object and handle them with a different massage)
- Undo line 418

netama:
didn't look at the implementation, but I think the plan was to start with a config flag for these and start with it disabled initially, right @edenhassid ?

adamro (approved):
🔥

PR #4342, opened 2023-05-07
SALTO-4121: Update the element source between deploy items

alonstern:
Update the element source between deploy items with the applied change.

This is so that when we deploy a change that depends on an addition, we will be able to access the values that are added to that addition change (such as internal id)

---

This is required for https://salto-io.atlassian.net/browse/SALTO-4067

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-api/src/elements.ts
+++ b/packages/adapter-api/src/elements.ts
@@ -623,6 +623,10 @@ export type ReadOnlyElementsSource = {
   get(id: ElemID): Promise<Value>
 }
 
+export type WritableElementsSource = ReadOnlyElementsSource & {
+  set(element: Readonly<Element>): Promise<void>
+}
+
 // This is a hack for the places we don't really need types in
 // transformElement. We need to replace this with not using transformElement.
 export const placeholderReadonlyElementsSource = {
--- a/packages/adapter-utils/src/element_source.ts
+++ b/packages/adapter-utils/src/element_source.ts
@@ -13,7 +13,7 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
-import { ReadOnlyElementsSource, Element, ElemID, Value, isElement } from '@salto-io/adapter-api'
+import { ReadOnlyElementsSource, Element, ElemID, Value, isElement, WritableElementsSource } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
 import _ from 'lodash'
 import { collections, values } from '@salto-io/lowerdash'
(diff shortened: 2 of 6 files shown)


packages/adapter-api/src/elements.ts
```diff
@@ -623,6 +623,10 @@ export type ReadOnlyElementsSource = {
   get(id: ElemID): Promise<Value>
 }
 
+export type WritableElementsSource = ReadOnlyElementsSource & {
```
ori-moisis, line 626: discussed f2f
I think we don't need this interface

packages/adapter-utils/src/element_source.ts
```diff
@@ -26,53 +26,68 @@ const { isDefined } = values
...
     }
   }
 
+  const get = async (id: ElemID): Promise<Value> => {
+    const element = elementsMap[id.getFullName()]
+    if (element !== undefined) {
+      return element
+    }
+
+    return (await awu(fallbackSources).find(async source => source.has(id)))?.get(id)
```
ori-moisis, line 80:
```suggestion
    return (await awu(fallbackSources).find(source => source.has(id)))?.get(id)
```

packages/adapter-utils/src/element_source.ts
```diff
@@ -26,53 +26,68 @@ const { isDefined } = values
...
+  }
+
+  const has = async (id: ElemID): Promise<boolean> => elementsMap[id.getFullName()] !== undefined
+    || awu(fallbackSources).some(async source => source.has(id))
```
ori-moisis, line 84:
```suggestion
  const has = async (id: ElemID): Promise<boolean> => (
    elementsMap[id.getFullName()] !== undefined
    || awu(fallbackSources).some(source => source.has(id))
  )
```

packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -109,7 +112,9 @@ export const deployActions = async (
       reportProgress(item, 'started')
       try {
         const result = await deployAction(item, adapters, checkOnly)
-        result.appliedChanges.forEach(appliedChange => appliedChanges.push(appliedChange))
+        await awu(result.appliedChanges).forEach(async appliedChange => {
+          appliedChanges.push(appliedChange)
+        })
```
ori-moisis, line 117: can revert the changes in this file?

packages/core/test/api.test.ts
```diff
@@ -514,6 +514,80 @@ describe('api.ts', () => {
...
+          ),
+        })
+
+        const actionPlan = mockPlan.createPlan([
+          [{ action: 'add', data: { after: instance.clone() } }],
+        ])
+
+        adapterOps.deploy.mockImplementationOnce(async ({ changeGroup }) => {
+          const changeInstance = getChangeData(changeGroup.changes[0]) as InstanceElement
+          changeInstance.value.id = 1
```
ori-moisis, line 565: I think in practice all adapters would create a clone of the element and not change the element from the plan in-place.
it might make the test pass accidentally

ori-moisis (approved):

PR #4343, opened 2023-05-07
SALTO-4124: deploy webhook with api_key as authentication

edenhassid:
deploy webhook with api_key as authentication

---

---
_Release Notes_: 
zendesk:
* Fix deploy of webhook when authentication is  api_key

---
_User Notifications_: 
none


netama (approved):
LGTM but let's add a unittest?

PR #4345, opened 2023-05-07
SALTO-4104 make local & remote filter creators more type safe

dantal4:
- Add `{ remote: true }` to remote filter creator in order to make local & remote filter creators more type safe.
- Remove `FilterWith` from SF adapter as it is not type safe.

---

---
_Release Notes_: 
Adapter Utils:
- Add `{ remote: true }` to remote filter creator in order to make local & remote filter creators more type safe.

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-utils/src/filter.ts
+++ b/packages/adapter-utils/src/filter.ts
@@ -53,12 +53,6 @@ export type FilterWith<
   DeployInfo = void,
 > = types.HasMember<Filter<T, DeployInfo>, M>
 
-export type FilterCreator<
-  R extends FilterResult | void,
-  T,
-  DeployInfo=void,
-> = (opts: T) => Filter<R, DeployInfo>
-
 export type LocalFilterCreator<
   R extends FilterResult | void,
   T,
@@ -70,6 +64,12 @@ export type RemoteFilterCreator<
   T,
   DeployInfo=void,
 > = (opts: T) => Filter<R, DeployInfo> & { remote: true }
+
+export type FilterCreator<
+  R extends FilterResult | void,
+  T,
+  DeployInfo=void,
+> = LocalFilterCreator<R, T, DeployInfo> | RemoteFilterCreator<R, T, DeployInfo>
 
 export type LocalFilterCreatorDefinition<
   R extends FilterResult | void,
--- a/packages/adapter-utils/src/filter.ts
+++ b/packages/adapter-utils/src/filter.ts
@@ -57,7 +57,7 @@ export type LocalFilterCreator<
   R extends FilterResult | void,
   T,
   DeployInfo=void,
-> = (opts: T) => Filter<R, DeployInfo> & { remote?: false }
+> = (opts: T) => Filter<R, DeployInfo>
 
 export type RemoteFilterCreator<
   R extends FilterResult | void,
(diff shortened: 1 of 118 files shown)


ori-moisis (commented):
not loving the side effect on the usage (specifically the tests)
any chance we can do this with the return value instead of the parameters?
meaning, this is a filter creator, right? what if we require the _local: undefined or even local: true in the return value alongside the onFetch or something?

e.g.
```
type RemoteFilterCreator = (...) => Filter<...> & { local: false }
type LocalFilterCreator = (...) => Filter<...> & { local: true }
```
so the only impact will be on the filter definition

packages/adapter-utils/src/filter.ts
```diff
@@ -59,12 +59,24 @@ export type FilterCreator<
...
+  R extends FilterResult | void,
+  T,
+  DeployInfo=void,
+> = (opts: T) => Filter<R, DeployInfo> & { remote?: false }
+
+export type RemoteFilterCreator<
+  R extends FilterResult | void,
+  T,
+  DeployInfo=void,
+> = (opts: T) => Filter<R, DeployInfo> & { remote: true }
```
ori-moisis, line 72: maybe we should reverse the roles between `LocalFilterCreator` and `FilterCreator` and make
```typescript
type FilterCreator<R,T,D> = LocalFilterCreator<R,T,D> | RemoteFilterCreator<R,T,D>
```
(with full names for the template params of course...)

I think we do want to maintain the coupling between "FilterCreator" and the local / remote variants of it, as it is they are completely decoupled

packages/salesforce-adapter/test/filters/animation_rules.test.ts
```diff
@@ -52,7 +53,7 @@ describe('animation rules filter', () => {
 
   let testElements: Element[]
 
-  const filter = filterCreator() as FilterWith<'onFetch'>
+  const filter = filterCreator({ config: {} as FilterContext }) as FilterWith<'onFetch'>
```
ori-moisis, line 56: we should not cast things like this
use `defaultFilterContext` like all the other tests do

this repeats a bunch of times in this PR, please change all of them

packages/jira-adapter/src/filters/service_url/service_url.ts
```diff
@@ -18,9 +18,19 @@ import JiraClient from '../../client/client'
 import { JiraConfig } from '../../config/config'
 import { FilterCreator, FilterResult } from '../../filter'
 
-const filter: FilterCreator = params =>
-  filters.serviceUrlFilterCreator<JiraClient, JiraConfig, FilterResult>(
+const filter: FilterCreator = params => {
```
dantal4, line 21: With the `FilterCreator` type change, TS didn't like this function's return value. Same in ZD.

ori-moisis: I think this would be fine if the definition of `LocalFilterCreator` didn't have the `& { remote?: false }` part

dantal4: `{ remote: never }` doesn't work (it is required in the return value). do you think that removing the `{ remote?: false }` is good enough?

ori-moisis (approved):

PR #4346, opened 2023-05-07
SALTO-3984 App Logo, brand Logo and Fav Icon

idoamit8:
1. In this PR I added fetch and deploy for app logos, brand logo and favorite icon.
2. Added test for all the filters. 

---

# _Additional context for reviewer_

### To test the code manually do as follows:
1. Fetch Okta env to get all instances.
2. Check that there are static files for appLogos, brandLogos and FavoriteIcon. 
3. change the path for each object type and deploy. 

### Key notes
1. app logo deploy doesn't support removal currently since there isn't API for that and it needs to return to a reset logo (Talked with Shir on it F2F). 
2. Need to add rate limit (Will do after I talk with Neta for the appropriate way to do so).

### Deploy:
Modify: 
change the source file name of appLogo, deploy and check if the appLogo changed in the website. 
`content = file("okta/AppLogo/${oldName}.png")` -> `content = file("okta/AppLogo/{newName.png")`
Add: 
Add application, appLogo and logo static file to your env and then deploy.
you can also add application and app logo and use a logo of a different app.  

---
_Release Notes_: 
_Okta adapter_:
Added option to fetch and deploy app logos, brand theme logo and brand theme icon. 

---
_User Notifications_: 
_Okta adapter_:
Added option to fetch and deploy app logos, brand theme logo and brand theme icon.


packages/okta-adapter/src/filters/app_logo.ts
```diff
@@ -0,0 +1,162 @@
...
+  const res = await httpClient.get(link, { responseType: 'arraybuffer' })
+  const content = _.isString(res.data) ? Buffer.from(res.data) : res.data
+  if (!Buffer.isBuffer(content)) {
+    log.error('Received invalid response from Okta API for attachment content')
+    return undefined
+  }
+  return content
+}
+
+const getLogoType = (contentType: string): string => {
```
shir-reifenberg, line 61: naming nit: maybe `getLogoFileType` ?

packages/okta-adapter/src/filters/app_logo.ts
```diff
@@ -0,0 +1,162 @@
...
+      change => getChangeData(change).elemID.typeName === APP_LOGO_TYPE_NAME,
+    )
+
+    const deployLogoResults = await Promise.all(appLogoChanges.map(async change => {
+      const logoInstance = getChangeData(change)
+      const fileContent = isAdditionOrModificationChange(change)
+        && isStaticFile(logoInstance.value.content)
+        ? await logoInstance.value.content.getContent()
+        : undefined
+      const deployResult = await deployAppLogo(client, logoInstance, fileContent)
```
shir-reifenberg, line 145: What happens if `fileContent` is undefined? (like in removal change)

packages/okta-adapter/src/filters/app_logo.ts
```diff
@@ -0,0 +1,162 @@
...
+  client: OktaClient,
+  logoInstance: InstanceElement,
+  fileContent: Buffer | undefined,
+): Promise<Error | void> => {
+  try {
+    const appId = getParent(logoInstance).value.id
+    const form = new FormData()
+    form.append('file', fileContent || Buffer.from(''))
+    await client.post({
+      url: `/api/v1/apps/${appId}/logo`,
```
shir-reifenberg, line 111: Can we move this part to the type's configuration ? (in config.ts)
it would look something like this:

```
    deployRequests: {
      modify: {
        url: /api/v1/apps/{appId}/logo',
        method: 'post',
        urlParamsToFields: {
          appId: '_parent.0.id',
        },
      },
     add: {..same}
    },
```
Then we can use `createUrl` to create the url from instance + config

packages/okta-adapter/src/filters/app_logo.ts
```diff
@@ -0,0 +1,162 @@
...
+      data: form,
+      headers: { ...form.getHeaders() },
+    })
+    return undefined
```
shir-reifenberg, line 115: nit: Maybe return result instead of undefined ?

shir-reifenberg (commented):
Great work :) Added some comments.
Maybe I missed it - but do we reset the logo to default on `RemovalChange` ? 
Also, we now add release notes 🎉  - so please update this please

packages/okta-adapter/src/adapter.ts
```diff
@@ -81,13 +82,18 @@ export const DEFAULT_FILTERS = [
   defaultPolicyRuleDeployment,
   policyRuleRemoval,
   schemaFieldsRemovalFilter,
+  appLogoFilter,
```
netama, line 85: nit: let's add this (and all the other filters that we can) before `fieldReferencesFilter`, @shir-reifenberg is there something that has to be after?

packages/okta-adapter/src/filters/app_logo.ts
```diff
@@ -0,0 +1,162 @@
...
+import { elements as elementsUtils } from '@salto-io/adapter-components'
+import FormData from 'form-data'
+import { getOktaError } from '../deployment'
+import { extractIdFromUrl } from '../utils'
+import { APPLICATION_TYPE_NAME, APP_LOGO_TYPE_NAME, OKTA } from '../constants'
+import { FilterCreator } from '../filter'
+import OktaClient from '../client/client'
+
+const log = logger(module)
+export const LINKS_FIELD = '_links'
```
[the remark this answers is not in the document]
netama: looks like it's already in constants

packages/okta-adapter/src/filters/app_logo.ts
```diff
@@ -0,0 +1,162 @@
...
+const { TYPES_PATH, SUBTYPES_PATH, RECORDS_PATH } = elementsUtils
+
+const APP_LOGO_TYPE = new ObjectType({
```
netama, line 34: 1. nit: it'd be better if this was a function (like `createAttachmentType` in zendesk), I see we have a hard-coded object type for brand logos in zendesk but in general it's safer to generate all types dynamically
2. there is an assumption here that this is running before one of the "common filters" which hides types, probably worth adding a note about it in the filter order in adapter.ts

packages/okta-adapter/src/filters/app_logo.ts
```diff
@@ -0,0 +1,162 @@
...
+    },
+    contentType: { refType: BuiltinTypes.STRING },
+    content: { refType: BuiltinTypes.STRING },
+  },
+  path: [OKTA, TYPES_PATH, SUBTYPES_PATH, APP_LOGO_TYPE_NAME, APP_LOGO_TYPE_NAME],
+})
+
+const getLogoContent = async (link: string,
+): Promise<Buffer | undefined> => {
+  const httpClient = axios.create({
```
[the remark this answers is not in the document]
netama: I would also like to understand :) can't we follow one of the examples from zendesk? there are some additional capabilities we get from the client (such as retries, rate limits, logging) that we shouldn't miss here

packages/okta-adapter/src/filters/app_logo.ts
```diff
@@ -0,0 +1,162 @@
...
+      refType: BuiltinTypes.STRING,
+      annotations: { [CORE_ANNOTATIONS.HIDDEN_VALUE]: true },
+    },
+    contentType: { refType: BuiltinTypes.STRING },
+    content: { refType: BuiltinTypes.STRING },
+  },
+  path: [OKTA, TYPES_PATH, SUBTYPES_PATH, APP_LOGO_TYPE_NAME, APP_LOGO_TYPE_NAME],
+})
+
+const getLogoContent = async (link: string,
```
netama, line 47: nit: why the terminology difference between `link` and `url`? can we just use `url` and avoid some renames?

packages/okta-adapter/src/filters/app_logo.ts
```diff
@@ -0,0 +1,162 @@
...
+}
+
+const getLogoType = (contentType: string): string => {
+  const contentTypeParts = contentType.split('/')
+  return contentTypeParts[contentTypeParts.length - 1]
+}
+
+const getAppLogo = async (app: InstanceElement,
+): Promise<InstanceElement | undefined> => {
+  const appLogo = app.value[LINKS_FIELD]?.logo[0]
```
netama, line 68: let's validate the structure with a schema guard (e.g. `logo` might not exist even if the links field exists)

packages/okta-adapter/src/filters/app_logo.ts
```diff
@@ -0,0 +1,162 @@
...
+  const logoLink = appLogo.href
+  if (logoLink === undefined) {
+    return undefined
+  }
+
+  const logoContent = await getLogoContent(logoLink)
+  if (logoContent === undefined) {
+    return undefined
```
netama, line 76: when would this happen?

packages/okta-adapter/src/filters/app_logo.ts
```diff
@@ -0,0 +1,162 @@
...
+  }
+  const appName = naclCase(app.value.label)
+  const pathName = pathNaclCase(appName)
+  const resourcePathName = `${normalizeFilePathPart(appName)}.${getLogoType(appLogo.type)}`
+  const logoId = extractIdFromUrl(logoLink)
+  const logo = new InstanceElement(
```
netama, line 82: generally, is there a reason this is implemented with a different function structure than the existing ones? I'm asking because I think a next step (not for this PR) would be to generalize this, so it'd be best if we can keep these as similar as possible

packages/okta-adapter/src/filters/app_logo.ts
```diff
@@ -0,0 +1,162 @@
...
+
+const appLogoFilter: FilterCreator = ({ client }) => ({
+  name: 'appLogoFilter',
+  onFetch: async elements => {
+    const appsWithLogo = elements
+      .filter(isInstanceElement)
+      .filter(e => e.elemID.typeName === APPLICATION_TYPE_NAME)
+      .filter(app => app.value[LINKS_FIELD]?.logo !== undefined)
+    elements.push(APP_LOGO_TYPE)
+    const appLogoInstances = (await Promise.all(appsWithLogo.map(async app => getAppLogo(app))))
```
netama, line 129: here for example, if we don't use the client with the rate limit then we should separate these into chunks so that we don't make too many requests at once (ideally we would just use the client)

packages/okta-adapter/src/filters/app_logo.ts
```diff
@@ -0,0 +1,162 @@
...
+      .filter(isInstanceElement)
+    appLogoInstances.forEach(logo => elements.push(logo))
```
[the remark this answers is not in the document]
netama: I think we should keep it this way actually, `...` can fail on large sets so this is one of the workarounds we've been using (even though the syntax is uglier...)

netama (commented):
cool :) added some questions that I think might result in some refactoring, so will do another pass afterwards

packages/okta-adapter/src/filters/app_logo.ts
```diff
@@ -0,0 +1,162 @@
...
+
+const getLogoType = (contentType: string): string => {
+  const contentTypeParts = contentType.split('/')
+  return contentTypeParts[contentTypeParts.length - 1]
+}
+
+const getAppLogo = async (app: InstanceElement,
+): Promise<InstanceElement | undefined> => {
+  const appLogo = app.value[LINKS_FIELD]?.logo[0]
+  const logoLink = appLogo.href
```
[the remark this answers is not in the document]
idoamit8: Yes, I was thinking that if it is undefined then `logoLink = appLogo.href` will just be undefined as well.

packages/okta-adapter/src/filters/app_logo.ts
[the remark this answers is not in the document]
idoamit8: We don't use the adapter client here because the URL is not in the shape of `https://dev-91042815.okta.com...` But rather in the shape `https://ok12static.oktacdn.com/fs/bco/4/fs09feow8dLmYUQay5d7`. 
I guess I can add retries, limits and etc if you think I should. @netama

packages/okta-adapter/src/filters/app_logo.ts
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  const logoContent = await getLogoContent(logoLink)
+  if (logoContent === undefined) {
+    return undefined
+  }
+  const appName = naclCase(app.value.label)
+  const pathName = pathNaclCase(appName)
+  const resourcePathName = `${normalizeFilePathPart(appName)}.${getLogoType(appLogo.type)}`
+  const logoId = extractIdFromUrl(logoLink)
```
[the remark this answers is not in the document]
idoamit8: @shir-reifenberg  what do you mean by fallback to the id value in this case.

packages/okta-adapter/src/filters/app_logo.ts
```diff
@@ -0,0 +1,162 @@
...
+      annotations: { [CORE_ANNOTATIONS.HIDDEN_VALUE]: true },
+    },
+    contentType: { refType: BuiltinTypes.STRING },
+    content: { refType: BuiltinTypes.STRING },
+  },
+  path: [OKTA, TYPES_PATH, SUBTYPES_PATH, APP_LOGO_TYPE_NAME, APP_LOGO_TYPE_NAME],
+})
+
+const getLogoContent = async (link: string,
+): Promise<Buffer | undefined> => {
```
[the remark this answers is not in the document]
idoamit8: @shir-reifenberg  I added the fetchErrors (SaltoErrors) in the filter itself to include all the errors. Do you think it's good?

packages/okta-adapter/src/filters/app_logo.ts
idoamit8: This filter doesn't supports removal yet (explained in the beginning of the PR) because of the fallback to the defualt logo

packages/okta-adapter/src/filters/app_logo.ts
idoamit8: @netama I'm not really sure if this can happen, I copied the implementation of this from Zendesk, where they assume that such a scenario is possible.

packages/okta-adapter/src/filters/app_logo.ts
idoamit8: I can implement getAppLogo to be more similar to getBrandLogo. The main differences are:
1. For getBrandLogo - Client is needed for getLogoContent and ןמ getAppLogo it is not needed. But it may be an optional variable.
2. Currently getBrandLogo uses a Type that is not produced by createBrandType, but you can change that of course.
I can do it in a new PR later.

packages/okta-adapter/src/adapter.ts
shir-reifenberg: can move 
```
defaultPolicyRuleDeployment,
policyRuleRemoval,
schemaFieldsRemovalFilter,
appLogoFilter,
```

to run before, i'll have to check more closely about the rest

packages/okta-adapter/src/filters/app_logo.ts
idoamit8: Talked F2F

packages/okta-adapter/src/adapter.ts
idoamit8: @shir-reifenberg Done

packages/okta-adapter/src/filters/app_logo.ts
idoamit8: Didn't do it because parent is not reference expression

shir-reifenberg (commented):
Cool 😎  Didn't review everything yet (will continue later) but a few comments for now - 
- Can you please edit the PR title to include brand logo and favicon as well + add release notes?
- not sure I got what code is relevant in the client, it's seems like there's some duplication 
- I think you can consolidate `brand_logo` and `fav_icon`, WDYT?
- We should also delete `brandThemeFilesValidator` in this PR, it wouldn't be relevant anymore 
- every exported function should be covered in tests as well, can you please add tests for `logo.ts` and for `getResource` ?

packages/okta-adapter/src/client/client.ts
```diff
@@ -98,4 +106,50 @@ export default class OktaClient extends clientUtils.AdapterHTTPClient<
...
+  //   if (url === undefined) {
+  //     log.warn('url is undefined')
+  //   }
+  //   this.resourceConn = clientUtils.createClientConnection({
+  //     retryOptions: clientUtils.createRetryOptions(
+  //       _.defaults({}, this.config?.retry, DEFAULT_RETRY_OPTS)
+  //     ),
+  //     createConnection: createResourceConnection,
+  //   })
+  // }
```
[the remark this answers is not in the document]
idoamit8: Yep, removed it.

packages/okta-adapter/src/client/client.ts
```diff
@@ -40,6 +44,10 @@ const DEFAULT_PAGE_SIZE: Required<clientUtils.ClientPageSizeConfig> = {
 export default class OktaClient extends clientUtils.AdapterHTTPClient<
   Credentials, clientUtils.ClientRateLimitConfig
 > {
+  protected isResourceApiLoggedIn = false
+  // These properties create another connection and client for Zendesk resources API
```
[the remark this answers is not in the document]
idoamit8: Done!

packages/okta-adapter/src/client/client.ts
```diff
@@ -40,6 +44,10 @@ const DEFAULT_PAGE_SIZE: Required<clientUtils.ClientPageSizeConfig> = {
 export default class OktaClient extends clientUtils.AdapterHTTPClient<
   Credentials, clientUtils.ClientRateLimitConfig
 > {
+  protected isResourceApiLoggedIn = false
```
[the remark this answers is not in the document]
idoamit8: removed it

packages/okta-adapter/src/client/client.ts
```diff
@@ -98,4 +106,50 @@ export default class OktaClient extends clientUtils.AdapterHTTPClient<
...
+    try {
+      const { url, headers, responseType } = args
+      const requestConfig = [headers, responseType].some(values.isDefined)
+        ? {
+          headers,
+          responseType,
+        }
+        : undefined
+      const { data, status } = await this.resourceClient.get(url, requestConfig)
+      log.debug('Received response for resource request %s with status %d', url, status)
```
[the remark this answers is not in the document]
idoamit8: I removed this function. I forgot to delete the client file with all the changes I tried to do

packages/okta-adapter/src/config.ts
```diff
@@ -838,6 +858,62 @@ const DEFAULT_TYPE_CUSTOMIZATIONS: OktaSwaggerApiConfig['types'] = {
...
   },
+  BrandLogo: {
+    deployRequests: {
+      add: {
+        url: '/api/v1/brands/{brandId}/themes/{themeId}/logo',
+        method: 'post',
+        urlParamsToFields: {
+          themeId: '_parent.0.id',
+          brandId: '_parent.1.id',
+        },
```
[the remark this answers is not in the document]
idoamit8: I added the links in the config because of the change validator that checks for end points. So I added it completely and not just part of it. Do you think we should remove it? @shir-reifenberg

packages/okta-adapter/src/config.ts
```diff
@@ -822,9 +840,11 @@ const DEFAULT_TYPE_CUSTOMIZATIONS: OktaSwaggerApiConfig['types'] = {
     transformation: {
       isSingleton: true,
       serviceIdField: 'id',
-      fieldsToHide: [{ fieldName: 'id' }],
-      fieldsToOmit: DEFAULT_FIELDS_TO_OMIT.concat({ fieldName: '_links' }),
-      serviceUrl: '/admin/customizations/branding',
+      fieldsToHide: [{ fieldName: 'id' }, { fieldName: '_links' }],
```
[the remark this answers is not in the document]
idoamit8: Added '_links' to fieldsToIgnore and now it works great.

netama (commented):
partial review since some of this is going to change :)

packages/okta-adapter/src/filters/app_logo.ts
```diff
@@ -0,0 +1,135 @@
...
+      .filter(isInstanceElement)
+    appLogoInstances.forEach(logo => elements.push(logo))
+  },
+  deploy: async (changes: Change<InstanceElement>[]) => {
+    const [appLogoChanges, leftoverChanges] = _.partition(
+      changes,
+      change => getChangeData(change).elemID.typeName === APP_LOGO_TYPE_NAME,
+    )
+
+    const deployLogoResults = await Promise.all(appLogoChanges.map(async change => {
```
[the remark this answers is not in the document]
idoamit8: I think its quite the same. In brand_logo in zendesk I will add 
```
await awu(
      brandLogoRemovals.concat(brandLogoAddistionsAndModifications)
    )
      .map(async change => {
        const logoInstance = getChangeData(change)
        const fileContent = isAdditionOrModificationChange(change)
          && isStaticFile(logoInstance.value.content)
          ? await logoInstance.value.content.getContent()
          : undefined
        const deployResult = await deployBrandLogo(client, logoInstance, fileContent)
        return deployResult === undefined ? change : deployResult
      })
      .toArray()
```
to the inner function (deployLogo).

packages/okta-adapter/src/config.ts
```diff
@@ -822,9 +840,11 @@ const DEFAULT_TYPE_CUSTOMIZATIONS: OktaSwaggerApiConfig['types'] = {
     transformation: {
       isSingleton: true,
       serviceIdField: 'id',
-      fieldsToHide: [{ fieldName: 'id' }],
-      fieldsToOmit: DEFAULT_FIELDS_TO_OMIT.concat({ fieldName: '_links' }),
-      serviceUrl: '/admin/customizations/branding',
```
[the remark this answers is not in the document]
idoamit8: I am not sure, I did the rebase over main after pulling

packages/okta-adapter/src/filters/app_logo.ts
```diff
@@ -0,0 +1,135 @@
...
+  }).unknown(true).required(),
+}).unknown(true).required()
+
+const isApp = createSchemeGuard<App>(EXPECTED_APP_SCHEMA, 'Received invalid response for app values')
+
+const getLogoFileType = (contentType: string): string | undefined => {
+  const contentTypeParts = contentType.split('/')
+  const fileType = contentTypeParts[contentTypeParts.length - 1]
+  if (!ALLOWED_LOGO_FILE_TYPES.has(fileType)) {
+    log.debug(`App logo content type ${contentType} is not supported`)
```
shir-reifenberg, line 63: As we skip fetching app logo in this case, I think this can be `log.warn`
```suggestion
    log.warn(`App logo content type ${contentType} is not supported`)
```

packages/okta-adapter/src/filters/brand_theme_files.ts
```diff
@@ -0,0 +1,47 @@
...
+      .find(instance => instance.elemID.typeName === BRAND_THEME_TYPE_NAME)
+
+    if (brandTheme === undefined) {
+      log.debug('No brandTheme was found')
+      return
+    }
+    // eslint-disable-next-line max-len
+    await fetchBrandThemeFiles({ client, brandTheme, elements, logoTypeNames: [BRAND_LOGO_TYPE_NAME, FAVORITE_ICON_TYPE_NAME] })
+    // await fetchBrandThemeFiles({ client, brandTheme, elements, logoTypeName: BRAND_LOGO_TYPE_NAME })
+    // await fetchBrandThemeFiles({ client, brandTheme, elements, logoTypeName: FAVORITE_ICON_TYPE_NAME })
```
shir-reifenberg, line 42: needed?

packages/okta-adapter/src/filters/logo.ts
```diff
@@ -0,0 +1,252 @@
...
+}:{
+  client: OktaClient
+  parents: InstanceElement[]
+  logoType: ObjectType
+  contentType: string
+  logoName?: string
+  link?: string
+}):
+Promise<InstanceElement | undefined> => {
+  const logoLink = link ?? parents[0].value?.[LOGO_TYPES_TO_VALUES[logoType.elemID.typeName].urlSuffix]
```
shir-reifenberg, line 156: If this is specific to `Brand` files, and we get `link` as param to the function, I think it's better (and more readable) to move `LOGO_TYPES_TO_VALUES` to `brand_theme_files.ts` filter, and pass the relevant values to this function.
Also, this way we can make `link` and `logoName` as non optional params and remove the checks for undefined.

packages/okta-adapter/src/filters/logo.ts
```diff
@@ -0,0 +1,252 @@
...
+  if (logoLink === undefined) {
+    return undefined
+  }
+  const logoContent = await getLogoContent(logoLink, client)
+  if (logoContent === undefined) {
+    return undefined
+  }
+  const name = logoName ?? LOGO_TYPES_TO_VALUES[logoType.elemID.typeName].fileName
```
shir-reifenberg, line 164: same as the comment above

packages/okta-adapter/src/filters/logo.ts
```diff
@@ -0,0 +1,252 @@
...
+  const pathName = pathNaclCase(name)
+  const resourcePathName = `${normalizeFilePathPart(name)}.${contentType}`
+  const logoId = extractIdFromUrl(logoLink)
```
shir-reifenberg, line 167: What if this is `undefined`?

packages/okta-adapter/src/filters/logo.ts
```diff
@@ -0,0 +1,252 @@
...
+}):
+Promise<InstanceElement | undefined> => {
+  const logoLink = link ?? parents[0].value?.[LOGO_TYPES_TO_VALUES[logoType.elemID.typeName].urlSuffix]
+  if (logoLink === undefined) {
+    return undefined
+  }
+  const logoContent = await getLogoContent(logoLink, client)
+  if (logoContent === undefined) {
+    return undefined
+  }
```
shir-reifenberg, line 163: I think we can try-catch here, throw inside `getLogoContent` in any error case, and return fetch warning instead of `undefined`

packages/okta-adapter/src/filters/logo.ts
```diff
@@ -0,0 +1,252 @@
...
+    },
+    [OKTA, RECORDS_PATH, logoType.elemID.typeName, pathName],
+    {
+      [CORE_ANNOTATIONS.PARENT]: refParents,
+    }
+  )
+  return logo
+}
+
+export const getBrandThemeFile = async (
```
shir-reifenberg, line 189: If this only relevant to `Brand`, let's move to `brand_theme_files.ts` :)

packages/okta-adapter/src/filters/logo.ts
```diff
@@ -0,0 +1,252 @@
...
+  const parents = getParents(brandTheme)
+  if (parents.length === 0 || parents[0].value.elemID.typeName !== BRAND_TYPE_NAME) {
+    return undefined
+  }
+  const instances = [brandTheme, parents[0].value]
+  const type = LOGO_TYPES_TO_VALUES[logoType.elemID.typeName].fileType
+  return getLogo({ client, parents: instances, logoType, contentType: type })
+}
+
+export const fetchBrandThemeFiles = async ({
```
shir-reifenberg, line 203: let's move to `brand_theme_files.ts` too

packages/okta-adapter/src/filters/brand_theme_files.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+const brandLogoFilter: FilterCreator = ({ client }) => ({
+  name: 'brandThemeFilesFilter',
+  onFetch: async elements => {
+    const brandTheme = elements
+      .filter(isInstanceElement)
+      .find(instance => instance.elemID.typeName === BRAND_THEME_TYPE_NAME)
+
+    if (brandTheme === undefined) {
+      log.debug('No brandTheme was found')
```
shir-reifenberg, line 36: nit 
```suggestion
      log.warn('No BrandTheme was found, skipping BrandLogo and BrandFavicon fetch')
```

packages/okta-adapter/src/filters/logo.ts
shir-reifenberg: After reading this function, personally I think it would be more readable to have this part directly inside the filter and not as a function

packages/okta-adapter/src/filters/logo.ts
```diff
@@ -0,0 +1,252 @@
...
+  return logo
+}
+
+export const getBrandThemeFile = async (
+  client: OktaClient,
+  brandTheme: InstanceElement,
+  logoType: ObjectType,
+): Promise<InstanceElement | undefined> => {
+  const parents = getParents(brandTheme)
+  if (parents.length === 0 || parents[0].value.elemID.typeName !== BRAND_TYPE_NAME) {
```
shir-reifenberg, line 195: can you please add a comment here? and log as well

shir-reifenberg (commented):
Great job! 👏  Sorry for the delay in review
Reviewed only fetch parts for now and will continue with deploy tomorrow. 
Few more comments - 
- `GetResource` is missing a test, can you please add one? In general, every exported function should have a test, but I think you can reduce the number of exported functions (commented inside)
- I think `logo` shouldn't be under `filters` path as this is more like utils and not a filter
- There are a lot of edge cases here, I think we should return errors where we can instead of just log + return undefined
- Can you please update release notes and user notifications ?

shir-reifenberg (commented):
Looks good 🎈 , added a few more comments but mostly nits

packages/okta-adapter/test/logo.test.ts
```diff
@@ -0,0 +1,149 @@
...
+        contentType,
+        content: new StaticFile({
+          filepath: 'okta/AppLogo/app1.png', encoding: 'binary', content,
+        }),
+      })
+      expect(logo?.annotations[CORE_ANNOTATIONS.PARENT]).toHaveLength(1)
+      expect(logo?.annotations[CORE_ANNOTATIONS.PARENT])
+        .toContainEqual(new ReferenceExpression(appInstance.elemID, appInstance))
+    })
+    it('should return error when contentis not buffer', async () => {
```
shir-reifenberg, line 103:
```suggestion
    it('should return error when content is not buffer', async () => {
```

packages/okta-adapter/test/logo.test.ts
```diff
@@ -0,0 +1,149 @@
...
+          ],
+        }
+      )
+      const appLogoChange = toChange({ after: appLogoInstance })
+      await deployLogo(appLogoChange, client)
+      expect(mockPost).toHaveBeenCalledTimes(1)
+      expect(mockPost).toHaveBeenCalledWith({
+        url: '/api/v1/apps/11/logo',
+        data: expect.anything(),
+        headers: expect.anything(),
```
shir-reifenberg, line 145: Why using `anything` ? can we validate headers too ?

packages/okta-adapter/test/filters/brand_theme_files.test.ts
```diff
@@ -0,0 +1,225 @@
...
+
+import { filterUtils } from '@salto-io/adapter-components'
+import { CORE_ANNOTATIONS, ElemID, InstanceElement, ObjectType, ReferenceExpression, StaticFile, getChangeData, isInstanceElement } from '@salto-io/adapter-api'
+import { BRAND_LOGO_TYPE_NAME, BRAND_THEME_TYPE_NAME, BRAND_TYPE_NAME, FAV_ICON_TYPE_NAME, OKTA } from '../../src/constants'
+import OktaClient from '../../src/client/client'
+import { getFilterParams, mockClient } from '../utils'
+import brandThemeFilesFilter from '../../src/filters/brand_theme_files'
+import { FilterResult } from '../../src/filter'
+
+describe('barnd files filter', () => {
```
shir-reifenberg, line 25:
```suggestion
describe('brand files filter', () => {
```

packages/okta-adapter/test/filters/brand_theme_files.test.ts
```diff
@@ -0,0 +1,225 @@
...
+        if (params.url === 'https://ok12static.oktacdn.com/bc/image/111') {
+          return {
+            status: 200,
+            data: content,
+          }
+        }
+        throw new Error('Err')
+      })
+    })
+    it('should create brandLogo type favIcon type and brnadLogo instance and favIcon instance', async () => {
```
shir-reifenberg, line 75:
```suggestion
    it('should create brandLogo type favIcon type and brandLogo instance and favIcon instance', async () => {
```

packages/okta-adapter/test/client/client.test.ts
```diff
@@ -118,4 +118,18 @@ describe('client', () => {
       expect(client.baseUrl).toEqual('http://my.okta.net')
     })
   })
+  describe('getResource ', () => {
+    let result: clientUtils.ResponseValue
+    it('sholud return the response', async () => {
```
shir-reifenberg, line 123:
```suggestion
    it('should return the response', async () => {
```

packages/okta-adapter/src/filters/app_logo.ts
```diff
@@ -0,0 +1,138 @@
...
+  id: string
+  label: string
+  _links: {
+  logo: [{
+    href: string
+    type: string
+    name: string
+    }]
+  }
+}
```
shir-reifenberg, line 44: Is this works ? shouldn't it be under `value` ?

shir-reifenberg (approved):
Yalla 😎 🎉 
Added one question and a few nits, but approving in advance.
Also, you accidentally committed test.log file, please remove it before merging.
I think it would be best to test it in staging as well (together with app / brand deployment), to make sure there's no regression.

PR #4348, opened 2023-05-07
SALTO-4120 - Change getFromPathIndex to not return multiple hints for elemIDs that do not exist in the pathIndex

adamro:
We have an issue when using `splitElementByPath` and using a pathIndex that does not have information for all the elemIDs that are being split. For example splitting an Object type and one if the fields does not exist in the pathIndex.
This is caused by the fact that `getFromPathIndex` returns the hints of the "closest" parent, even if that parent has multiple hints.
This changes the logic so that for elemIDs that exist in the pathIndex, we will still return all the relevant hints, but for those who do not, we will still use the closest parent's hints, but only one of them and not all the hints cause returning multiple will eventually cause merge errors for this element since the part that does not exist in the pathIndex will appear in multiple fragments.

---
_Release Notes_: 

*Core*:
* Fix a bug in Restore and Fetch From Workspace when used when there are parts of the Elements that exist in the Workspace/State but were not fetched yet (ie. Fields inside ObjectType that are split across multiple files that were Deployed and then Fetched From) would cause a Merge Error after the Restore/Fetch From since the parts would appear in multiple files.

---
_User Notifications_: 
None

### DIFF
--- a/packages/workspace/src/workspace/path_index.ts
+++ b/packages/workspace/src/workspace/path_index.ts
@@ -246,15 +246,21 @@ export const getFromPathIndex = async (
 ): Promise<Path[]> => {
   const idParts = elemID.getFullNameParts()
   const topLevelKey = elemID.createTopLevelParentID().parent.getFullName()
+  let isExactMatch = true
   let key: string
   do {
     key = idParts.join('.')
     // eslint-disable-next-line no-await-in-loop
     const pathHints = await index.get(key)
-    if (pathHints !== undefined) {
-      return pathHints
+    if (pathHints !== undefined && pathHints.length > 0) {
+      // If we found this elemID in the pathIndex we want to return all the hints.
+      // If this is not an exact match we want to return a single hint
+      // because otherwise, splitElementByPath will make it appear in multiple fragments
+      // and cause merge errors.
+      return isExactMatch ? pathHints : [pathHints[0]]
     }
     idParts.pop()
+    isExactMatch = false
   } while (idParts.length > 0 && key !== topLevelKey)
   return []
 }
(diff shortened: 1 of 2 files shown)


ori-moisis:
I think we can remove some of the technical details from the release notes (e.g - "pathIndex" is not a user-facing term)

@adamro  do you think I am a valid reviewer for this or should it be reviewed by someone else?

packages/workspace/src/workspace/path_index.ts
```diff
@@ -246,15 +246,21 @@ export const getFromPathIndex = async (
...
     // eslint-disable-next-line no-await-in-loop
     const pathHints = await index.get(key)
-    if (pathHints !== undefined) {
-      return pathHints
+    if (pathHints !== undefined && pathHints.length > 0) {
+      // If we found this elemID in the pathIndex we want to return all the hints.
+      // If this is not an exact match we want to return a single hint
+      // because otherwise, splitElementByPath will make it appear in multiple fragments
+      // and cause merge errors.
+      return isExactMatch ? pathHints : [pathHints[0]]
```
alonstern, line 260: nit: maybe 
```suggestion
      return key === elemID.getFullName() ? pathHints : [pathHints[0]]
```
IMO it's a bit easier to read

alonstern (approved):
LGTM 👌 
Is this the whole solution or do we plan to do one of the other suggested solutions in the ticket?

adamro:
For now this is the solution. It's not ideal but it's the simplest and solves it. When we eventually do the fragments refactor I think we will go with updating the pathIndex on deploy and then we can go with that direction.

PR #4350, opened 2023-05-07
SALTO-4127: Fix bug in missing users validator

shir-reifenberg:
Fix bug in missing users validator causing users to get errors for changes in instances that don't have users path

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/test/change_validators/users.test.ts
+++ b/packages/okta-adapter/test/change_validators/users.test.ts
@@ -107,7 +107,7 @@ describe('usersValidator', () => {
     ])
     expect(changeErrors).toHaveLength(0)
   })
-  it('should not return error if there are no users in instance', async () => {
+  it('should not return error if users path does not exist', async () => {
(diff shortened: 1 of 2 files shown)


packages/okta-adapter/test/change_validators/users.test.ts
```diff
@@ -107,4 +107,24 @@ describe('usersValidator', () => {
     ])
     expect(changeErrors).toHaveLength(0)
   })
+  it('should not return error if there are no users in instance', async () => {
```
edenhassid, line 110: don't you want to add a test if the path doesn't exist?

edenhassid (approved):
good catch! added a small comment :)

packages/okta-adapter/test/change_validators/users.test.ts
shir-reifenberg: that's what it checks :) I'll change description to make it clearer

PR #4351, opened 2023-05-07
SALTO-4096: increase page size in pagination

IdoZakk:
Increased the page size of the pagination in Jira to a max of 1000
---

It seems Jira Cloud is setting a very low limit to the max page size.
The official number is 100, though I see that in some cases it is even lower (for instance for fields it is 50).
Jira DC has a higher limit (documentation says 1000, and it confirms for users).
That is a shame as I see that a big fetch drastically reduces the fetch speed- for instance for fields there is an API call to bring all fields. Bringing 3K small fields took an average of 464ms, while bringing 50 at a time took 200ms per call. The fields were small, but even if this change is per call and does not depend on returned item size (a reasonable assumption) we save 4 seconds per 1K items.

In most cases Jira corrects the max result, so it is safe to call with a bigger number. I have set it to 1000 (besides statuses that fail when set higher than 200), the improvement for an env with 3K fields and 3K filter is around 7%

---
_Release Notes_: 
Jira Adapter:
* Increased page size in fetch to improve fetch performance

---
_User Notifications_: 
None


alonstern (commented):
Probably worth also adding it in:
priority_scheme_fetch.ts
automation_fetch.ts

Do we know if this improve the time?

IdoZakk:
I had to merge so I squashed the changes

alonstern:
@IdoZakk What about `priority_scheme_fetch.ts`?

IdoZakk:
It uses the pagination infra, did not find an easy way to add the page size, I prefer to submit it as it is now so we will have it as part of the next deploy

alonstern (approved):

PR #4353, opened 2023-05-08
SALTO-4002 - 

idoser96:
…lving orgs

---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/zendesk-adapter/src/change_validators/organization_existence.ts
+++ b/packages/zendesk-adapter/src/change_validators/organization_existence.ts
@@ -85,11 +85,12 @@ export const organizationExistenceValidator: (client: ZendeskClient, fetchConfig
 
           // If the instance includes an organization that does not exist, we won't allow a change to that instance
           if (nonExistingOrgs.size > 0) {
+            const resolveOrgsRecommendation = !orgIdsResolved ? ', We recommend you to set resolveOrganizationIDs=true in your fetch config' : ''
             return {
               elemID: entries[0].instance.elemID,
               severity: 'Error',
               message: 'Referenced organizations do not exist',
-              detailedMessage: `The following referenced organizations do not exist: ${Array.from(nonExistingOrgs).join(', ')}`,
+              detailedMessage: `The following referenced organizations do not exist: ${Array.from(nonExistingOrgs).join(', ')}${resolveOrgsRecommendation}`,
             }
           }
           return undefined


packages/zendesk-adapter/src/change_validators/organization_existence.ts
```diff
@@ -22,8 +22,9 @@ import {
 } from '@salto-io/adapter-api'
 import { client as clientUtils } from '@salto-io/adapter-components'
 import { collections, values as lowerDashValues } from '@salto-io/lowerdash'
-import _ from 'lodash'
+import _, { isNumber } from 'lodash'
```
netama, line 25: nit: please use `_.isNumber` and avoid the extra import

packages/zendesk-adapter/src/change_validators/organization_existence.ts
```diff
@@ -85,11 +86,15 @@ export const organizationExistenceValidator: (client: ZendeskClient, fetchConfig
 
           // If the instance includes an organization that does not exist, we won't allow a change to that instance
           if (nonExistingOrgs.size > 0) {
+            // Check if any of the non-existing orgs are numbers, if so, we should recommend the user to resolve them
+            const nonExistingOrgsAreIds = !orgIdsResolved || wu(nonExistingOrgs.values()).some(org => isNumber(org))
```
netama, line 90: edge-case nit, but do we know that organization names cannot be all-numbers?

packages/zendesk-adapter/src/change_validators/organization_existence.ts
```diff
@@ -85,11 +86,15 @@ export const organizationExistenceValidator: (client: ZendeskClient, fetchConfig
...
+            const resolveOrgsRecommendation = nonExistingOrgsAreIds ? '. Salto can identify organizations by their names. This requires setting the \'resolveOrganizationIDs\' to true in the zendesk configuration file of both source and target envs and fetch.\n'
+                + 'More information about Salto\'s config files can be found here' : ''
```
netama, line 92: 1. readability nit:
```suggestion
            const resolveOrgsRecommendation = nonExistingOrgsAreIds
              ? '. Salto can identify organizations by their names. This requires setting the \'resolveOrganizationIDs\' to true in the zendesk configuration file of both source and target envs and fetch.\n'
                + 'More information about Salto\'s config files can be found here'
              : ''
```
2. what is "here" supposed to signify? are we adding a link?

packages/zendesk-adapter/src/change_validators/organization_existence.ts
```diff
@@ -85,11 +86,15 @@ export const organizationExistenceValidator: (client: ZendeskClient, fetchConfig
...
             return {
               elemID: entries[0].instance.elemID,
               severity: 'Error',
-              message: 'Referenced organizations do not exist',
-              detailedMessage: `The following referenced organizations do not exist: ${Array.from(nonExistingOrgs).join(', ')}`,
+              message: `Referenced organization ${orgIdsResolved ? 'names' : 'IDs'} do not exist`,
```
netama, line 96: this might not be accurate - some organizations might be referenced by name and some by id. maybe "Some referenced organizations could not be found"?

packages/zendesk-adapter/src/change_validators/organization_existence.ts
idoser96: I actually think they can be, but I don't think we can handle it (checking if its a name of numbers or an id)
also there is a possibility that the config was turned on and a deploy started before a fetch (or it was turned on on the target env and not on the source env, and in this we need to check for an org name of a number

This is indeed an edge-case, I think this is the safer way to handle it

PR #4356, opened 2023-05-08
SALTO-4027 - Salesforce: verify another user reference

tomsellek:
Another field with a user ID we can validate

---

The logic here is similar to `FolderShare.sharedTo`, so I merged the logic of the two fields into a single function

---
_Release Notes_: 
Salesforce: The user referenced in WorkflowTask.assignedTo is now verified

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/change_validators/unknown_users.ts
+++ b/packages/salesforce-adapter/src/change_validators/unknown_users.ts
@@ -34,7 +34,7 @@ type GetUserField = (instance: InstanceElement, fieldName: string) => string[]
 
 type UserFieldGetter = {
   field: string
-  getter: GetUserField
+  getter: (instance: InstanceElement) => string[]
 }
 
 // https://stackoverflow.com/a/44154193
@@ -54,27 +54,18 @@ type MissingUser = {
   userName: string
 }
 
-const userFieldValue = (expectedFieldName: string): GetUserField => (
-  (instance, fieldName) => {
-    if (fieldName !== expectedFieldName) {
-      log.error(`Unexpected field name: ${fieldName} !== ${expectedFieldName}`)
-      return []
-    }
-    if (instance.value[fieldName] === undefined) {
-      return []
-    }
-    return [instance.value[fieldName]]
+const userFieldValue = (instance: InstanceElement, fieldName: string): string[] => {
+  if (instance.value[fieldName] === undefined) {
+    return []
   }
-)
+  return [instance.value[fieldName]]
+}
 
-const getUserDependingOnType = (typeField: string, expectedType: string, userField: string): GetUserField => (
-  (instance: InstanceElement, fieldName: string) => {
-    if (fieldName !== userField) {
-      log.warn('Expected field name %s. Got %s', userField, fieldName)
-      return []
-    }
-    if (instance.value[typeField] !== expectedType) {
-      log.trace('%s is not %s. Skipping.', typeField, expectedType)
+const getUserDependingOnType = (typeField: string): GetUserField => (
+  (instance: InstanceElement, userField: string) => {
+    const type = instance.value[typeField]
+    if (!type || type.toLocaleLowerCase() !== 'user') {
+      log.trace('%s is `%s`, not `user`. Skipping.', typeField, type)
       return []
     }
     return [instance.value[userField]]
@@ -91,11 +82,7 @@ const isEmailRecipientsValue = (recipients: Values): recipients is EmailRecipien
   && recipients.every(recipient => _.isString(recipient.type) && _.isString(recipient.recipient))
 )
 
-const getEmailRecipients: GetUserField = (instance, fieldName) => {
-  if (fieldName !== 'recipients') {
-    log.error(`Unexpected field name: ${fieldName}.`)
-    return []
-  }
+const getEmailRecipients: GetUserField = instance => {
   const { recipients } = instance.value
   if (!isEmailRecipientsValue(recipients)) {
     return []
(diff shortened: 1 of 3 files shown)


packages/salesforce-adapter/src/change_validators/unknown_users.ts
```diff
@@ -137,6 +140,12 @@ const USER_GETTERS: TypesWithUserFields = {
       getter: getEmailRecipients,
     },
   ],
+  WorkflowTask: [
+    {
+      field: 'assignedTo',
+      getter: getUserDependingOnType('assignedToType', 'user', 'assignedTo'),
```
benisraeladi, line 146: In line 134 it's `User` capitalized, while here it's `user` not capitalized. Just want to make sure it's on purpose

tomsellek: That's what the ticket says 🤷

packages/salesforce-adapter/src/change_validators/unknown_users.ts
```diff
@@ -128,7 +131,7 @@ const USER_GETTERS: TypesWithUserFields = {
   FolderShare: [
     {
       field: 'sharedTo',
-      getter: getFolderShareUser,
+      getter: getUserDependingOnType('sharedToType', 'User', 'sharedTo'),
```
yuribro, line 134: All these strings look very error prone.  Will userField always be the same as the `field` in the UserFieldGetter? 
I think they should at least be constants, but would be even better if it is written in a way where it doesn't have to be typed out twice.

packages/salesforce-adapter/src/change_validators/unknown_users.ts
```diff
@@ -78,17 +79,19 @@ const userFieldValue = (expectedFieldName: string): GetUserField => (
...
 
-const getFolderShareUser: GetUserField = (instance, fieldName) => {
-  if (fieldName !== 'sharedTo') {
-    log.error(`Unexpected field name: ${fieldName}.`)
-    return []
-  }
-  if (instance.value.sharedToType !== 'User') {
-    log.debug('sharedToType is not User. Skipping.')
-    return []
+const getUserDependingOnType = (typeField: string, expectedType: string, userField: string): GetUserField => (
```
yuribro, line 82: Doesn't this replace  getCaseSettingsOwner also?

yuribro: nit: I think this name is a little confusing :) 

This function reads to me as it gets a specific field, but verifies a different field to be specific value?

packages/salesforce-adapter/src/change_validators/unknown_users.ts
```diff
@@ -137,6 +140,12 @@ const USER_GETTERS: TypesWithUserFields = {
       getter: getEmailRecipients,
     },
   ],
+  WorkflowTask: [
```
yuribro, line 143: I think we should add a UT per new case

packages/salesforce-adapter/src/change_validators/unknown_users.ts
tomsellek: > Doesn't this replace getCaseSettingsOwner also?

Yuppers. Thanks for noticing.

> nit: I think this name is a little confusing :)

Suggestions for a better name are welcome :)

packages/salesforce-adapter/src/change_validators/unknown_users.ts
tomsellek: If I understand your comment correctly then this is the same discussion we had with Dor when this CV was created. I don't think it would be a productive use of our time to revisit this design every time we add a new field to check.

packages/salesforce-adapter/src/change_validators/unknown_users.ts
tomsellek: It is now exactly the same logic as other cases that we test - what would a new UT test?

yuribro: It will test that we didn't made a silly error in the arguments. How else would we verify once we refactor that we didn't replace `User` with `user` accidentally?

packages/salesforce-adapter/src/change_validators/unknown_users.ts
yuribro: I don't think I was part of that discussion (or maybe I did but don't remember), but that only makes the argument stronger, if it is suggested every time? 
What is the down side of making it more type safe? Will it be much less readable?

tomsellek:
@yuribro ready for a re-review

packages/salesforce-adapter/test/mock_elements.ts
```diff
@@ -29,7 +29,7 @@ import {
   CPQ_QUOTE,
   DUPLICATE_RULE_METADATA_TYPE,
   INSTALLED_PACKAGE_METADATA,
-  PATH_ASSISTANT_METADATA_TYPE,
+  PATH_ASSISTANT_METADATA_TYPE, WORKFLOW_TASK_METADATA_TYPE,
```
yuribro, line 32:
```suggestion
  PATH_ASSISTANT_METADATA_TYPE, 
  WORKFLOW_TASK_METADATA_TYPE,
```

packages/salesforce-adapter/test/change_validators/unknown_users.test.ts
```diff
@@ -19,7 +19,7 @@ import {
   ChangeValidator,
   ElemID,
   getChangeData,
-  InstanceElement,
+  InstanceElement, ObjectType,
```
yuribro, line 22:
```suggestion
  InstanceElement, 
  ObjectType,
```

packages/salesforce-adapter/src/change_validators/unknown_users.ts
```diff
@@ -53,43 +54,24 @@ type MissingUser = {
...
 
-const userFieldValue = (expectedFieldName: string): GetUserField => (
-  (instance, fieldName) => {
-    if (fieldName !== expectedFieldName) {
-      log.error(`Unexpected field name: ${fieldName} !== ${expectedFieldName}`)
+const getUserDependingOnType = (typeField: string): GetUserField => (
+  (instance: InstanceElement, userField: string) => {
+    const type = instance.value[typeField]
+    if (!type || type.toLocaleLowerCase() !== 'user') {
+      log.trace('%s is `%s`, not `user`. Skipping.', typeField, type)
```
yuribro, line 68: Are we sure that this is not very spammy? is this an exceptional case or just something which can happen a lot depending on the data?

yuribro (approved):
LGTM! approved with some small nits to save cycles

packages/salesforce-adapter/src/change_validators/unknown_users.ts
tomsellek: It can certainly happen a lot. I thought making it a trace was enough, but I'll remove it.

PR #4357, opened 2023-05-08
SALTO-3530: Change config suggestion messages format

shoshana-houri:
Change Config Suggestion Messages to be neutral in order to fit the 'auto-applied' scenario.

---
* Change config suggestion messages format to a neutral one in order to fit the 'auto-applied' scenario.
New format is discussed [here](https://www.notion.so/saltoio/SALTO-3530-modify-config-suggestions-messages-8cb4c6b1ae974c6b8ea4a5fc5adc13fa?pvs=4).
* Realign all the adapters' config suggestion messages.
* Add a CLI prefix that gives the context of the changes (as part of the interactive CLI question)
* Add context for each account's config suggestion messages by specifying the path to the config file (double win).

---
_Release Notes_: 

---
_User Notifications_:


packages/netsuite-adapter/src/config.ts
```diff
@@ -584,10 +584,10 @@ export const configType = createMatchingObjectType<NetsuiteConfig>({
 })
 
 export const STOP_MANAGING_ITEMS_MSG = 'Salto failed to fetch some items from NetSuite.'
-  + ' In order to complete the fetch operation, Salto needs to stop managing these items by modifying the configuration.'
+  + ' Failed items must be excluded from the fetch.'
 
-export const LARGE_FOLDERS_EXCLUDED_MESSAGE = 'Some File Cabinet folders were excluded from the fetch.'
- + ' To include them, increase the File Cabinet\'s size limitations and remove their exclusion rules.'
+export const LARGE_FOLDERS_EXCLUDED_MESSAGE = 'Some File Cabinet folders exceeds File Cabinet\'s size limitations.'
```
omrilit, line 589: I think
```suggestion
export const LARGE_FOLDERS_EXCLUDED_MESSAGE = 'Some File Cabinet folders exceed File Cabinet\'s size limitation.'
```

packages/netsuite-adapter/src/config.ts
```diff
@@ -584,10 +584,10 @@ export const configType = createMatchingObjectType<NetsuiteConfig>({
...
+ + ' To include them, increase the File Cabinet\'s size limitations and remove their exclusion rules from the configuration file.'
```
omrilit, line 590:
```suggestion
 + ' To include them, increase the File Cabinet\'s size limitation and remove their exclusion rules from the configuration file.'
```

omrilit (commented):
NS and general parts LGTM!

packages/core/src/core/fetch.ts
```diff
@@ -465,7 +466,14 @@ const fetchAndProcessMergeErrors = async (
...
     const updatedConfigs = fetchResults
       .map(res => res.updatedConfig)
-      .filter(values.isDefined) as UpdatedConfig[]
+      .filter(values.isDefined)
+      .map(({ config, message, accountName }) =>
+        ({
+          config,
+          message: _.isEmpty(message)
+            ? ''
+            : `Issues which triggered changes in salto.config/adapters/${accountName}:\n${message}`,
```
netama, line 475: nit: can we use `CONFIG_PATH` from `adapter_config_source.ts` in the workspace instead of hard-coding this path here? (@ori-moisis is there a better place to get it from?). I see we hard-code it in a few other places, just seems like this might hcange

packages/zendesk-adapter/src/config_change.ts
```diff
@@ -13,18 +13,17 @@
...
+import { formatConfigSuggestionsReasons } from '@salto-io/adapter-utils'
 import { configType, FETCH_CONFIG } from './config'
 
 const { makeArray } = collections.array
 
-const MESSAGE_INTRO = 'Salto failed to fetch some items from zendesk.'
-const MESSAGE_SUMMARY = 'In order to complete the fetch operation, '
-+ 'Salto needs to stop managing these items by applying the following configuration change:'
+const STOP_MANAGING_ITEMS_MSG = 'Salto failed to fetch some items from Zendesk.'
+  + ' Failed items must be excluded from the fetch.'
```
netama, line 26: nits: 
1. I'm not sure about the phrasing here, maybe something like "The configuration was updated to exclude these items in order to complete the fetch operation" (or something like that that explains the "flow" more clearly)? 
2. Did we intentionally change this to a single line?

packages/zuora-billing-adapter/src/config.ts
```diff
@@ -837,7 +837,7 @@ export type FilterContext = {
   [API_DEFINITIONS_CONFIG]: ZuoraApiConfig
 }
 
-const FIXING_SYSTEM_CONFIGURATION_INTRO = 'Fixing system configuration for the following types:'
+const FIXING_SYSTEM_CONFIGURATION_INTRO = 'Incorrect configuration format for the following types:'
```
netama, line 840: The configuration was not incorrect exactly - let's just remove this completely? this is from 2 years ago so no longer relevant :)
(I can remove it if you prefer)

netama (commented):
reviewed zendesk+zuora, added some nits on the shared logic but probably best that @adamro or @ori-moisis review it since they have more context :)

packages/core/src/core/fetch.ts
ori-moisis: that is exported from the workspace package, seems like that constant is already used once in core: search for `...acs.CONFIG_PATH`

packages/zendesk-adapter/src/config_change.ts
shoshana-houri: 1. The phrasing should be in a neutral form, meaning no past/future tense, since we want to be able to use those messages both in the CLI **before** we apply the changes and in Saas **after** we auto applied them.
The info regarding whether the configuration has or hasn't changed is moved to the intro which is added by the client, [here](https://www.notion.so/saltoio/SALTO-3530-modify-config-suggestions-messages-8cb4c6b1ae974c6b8ea4a5fc5adc13fa?pvs=4#93d842c9a26744d19b9eb55bb9cbecbe) is the full final format/s.
2. I figured both of them count as a single reason for the configuration changes, lmk if not and I'll change it :)

packages/core/src/core/fetch.ts
shoshana-houri: > can we use CONFIG_PATH

> search for ...acs.CONFIG_PATH

Thanks! didn't know it existed.

> do we really need the file name here?

The dir name was requested by @iperelman and @tomermevorach  to easily refer the user to the correct place in the fetch diff. I can move this logic outside the core, though it'll be duplicated both in the CLI and in the Saas.

tomsellek (commented):
Salesforce part LGTM. Not approving because I didn't review any of the other parts.

packages/core/src/core/fetch.ts
ori-moisis: no strong opinion, since that const is already used in core we can keep it here if you prefer

packages/zendesk-adapter/src/config_change.ts
netama: ok, so I think maybe you want @tomermevorach to review this and not me? :)

netama (commented):
zuora LGTM :), for zendesk I think @tomermevorach should probably review (the only change is the text change which IIUC he defined)

tomermevorach (approved):

PR #4359, opened 2023-05-09
SALTO-4066 Add to all salesforce CVs a more details link

Uri-Bar:
SALTO-4066 Add to all salesforce CVs a more details link
---

I add the link in all of the CV that did not include a different link already and are not in info severity.

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/salesforce-adapter/src/change_validators/animation_rule_recordtype.ts
+++ b/packages/salesforce-adapter/src/change_validators/animation_rule_recordtype.ts
@@ -31,7 +31,7 @@ const invalidRecordTypeError = (instance: InstanceElement): ChangeError => (
     elemID: instance.elemID,
     severity: 'Error',
     message: 'Invalid AnimationRule RecordType',
-    detailedMessage: `In ${instance.elemID.getFullName()}, The RecordTypeId field is missing even though RecordTypeContext requires a RecordTypeId.`,
+    detailedMessage: `In ${instance.elemID.getFullName()}, The RecordTypeId field is missing even though RecordTypeContext requires a RecordTypeId.\nfor more details: https://help.salto.io/en/articles/7793653-deployment-preview-errors`,
   }
 )
 
--- a/packages/salesforce-adapter/src/change_validators/animation_rule_recordtype.ts
+++ b/packages/salesforce-adapter/src/change_validators/animation_rule_recordtype.ts
@@ -31,7 +31,7 @@ const invalidRecordTypeError = (instance: InstanceElement): ChangeError => (
     elemID: instance.elemID,
     severity: 'Error',
     message: 'Invalid AnimationRule RecordType',
-    detailedMessage: `In ${instance.elemID.getFullName()}, The RecordTypeId field is missing even though RecordTypeContext requires a RecordTypeId.\nfor more details: https://help.salto.io/en/articles/7793653-deployment-preview-errors`,
+    detailedMessage: `In ${instance.elemID.getFullName()}, The RecordTypeId field is missing even though RecordTypeContext requires a RecordTypeId.\nfor more details on 'animationRuleRecordType': https://help.salto.io/en/articles/7793653-deployment-preview-errors`,
   }
 )
 
--- a/packages/salesforce-adapter/src/change_validators/currency_iso_codes.ts
+++ b/packages/salesforce-adapter/src/change_validators/currency_iso_codes.ts
@@ -67,7 +67,7 @@ const createOrgHasNoMultiCurrencyEnabledError = (
 ): ChangeError => ({
   elemID,
   message: 'Organization doesnt support multi currency',
-  detailedMessage: 'Cannot deploy instance with CurrencyIsoCode field to organization that does not support multi currencies',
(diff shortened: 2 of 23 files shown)


Uri-Bar:
> ## SALTO-4066 Add to all salesforce CVs a more details link
> _Additional context for reviewer_ I add the link in all of the CV that did not include a different link already and are not in info severity.
> 
> _Release Notes_: None
> 
> _User Notifications_: None

iperelman:
@Uri-Bar I think we should add this only to error CVs (since they are blocking), no need to add it to the warning ones.  As for CVs already having a link, are we not able to add a 2nd link there?  Why do you think it's weird to add this to CVs about SF API limitations?

Uri-Bar:
@iperelman no problem to change it to error only, but than it is weird that if the user has an error warning that we know that will fail the deployment (often by salesforce api limitations), we will send him to a link that tells him how to turn off the validation. how this link will help him? WDYT?

iperelman:
@Uri-Bar per https://salto-io.atlassian.net/browse/SALTO-4066 this all thing started with a bug in one of our CVs.  I agree it's an unlikely (hopefully) scenario, yet if a customer wants to try and bypass a limitation to try it out directly with the business application and see the result, this is how they can do it.  FWIW, Gearset allows users to do the same - https://docs.gearset.com/en/articles/625488-an-introduction-to-gearset-s-problem-analyzers#:~:text=Overriding%20the%20problem%20analysis

The reason to add this only to CV errors, yet not to warnings is that warnings do not lead to skipped elements, i.e. users can still experience the error directly w/o Salto blocking them from doing so.   

Does this make more sense now or am I missing anything?

Uri-Bar:
@iperelman it does make more sense. I understand the bug issue, but to add the link to all of the CVs just in case there is a bug is really excessive and  much less likely than a user that is trying to do something he can't. I gonna let @amitgoldarb take this from here and Im ok with whatever you decide.

iperelman:
@Uri-Bar honestly, I'm good either way.  We did not originally highlighted the option to bypass the CVs, since we thought it's excessive.  Following the bug, Hadar asked to add it and we figured out a relatively subtle way to add it so that we do not hurt the user experience much, yet save a little support time.  I'm 100% good either way as well :)

packages/salesforce-adapter/src/change_validators/animation_rule_recordtype.ts
```diff
@@ -31,7 +31,7 @@ const invalidRecordTypeError = (instance: InstanceElement): ChangeError => (
     elemID: instance.elemID,
     severity: 'Error',
     message: 'Invalid AnimationRule RecordType',
-    detailedMessage: `In ${instance.elemID.getFullName()}, The RecordTypeId field is missing even though RecordTypeContext requires a RecordTypeId.`,
+    detailedMessage: `In ${instance.elemID.getFullName()}, The RecordTypeId field is missing even though RecordTypeContext requires a RecordTypeId.\nfor more details: https://help.salto.io/en/articles/7793653-deployment-preview-errors`,
```
amitgoldarb, line 34: @iperelman small product issue with this text here that we thought of -
In the document there is a guide on how to turn off CV, but to turn them off we require the name of the CV itself which isn't clear from the message.
I suggest something along the following - Does this make sense?
```suggestion
for more details on <Change validator name>: <URL>
```

amitgoldarb (commented):
@iperelman WDYT?

packages/salesforce-adapter/src/change_validators/animation_rule_recordtype.ts
iperelman: @amitgoldarb yeah, it's probably better.  I'm a little bothered that we do not really share more details about the specific CV there, but rather only about how to turn it off, yet I agree it helps with understanding what CV is in play here + @pgonzaleznetwork is in the process of writing a help page for each of the CVs anyway so soon we should actually have more meaningful details about each CV.

packages/salesforce-adapter/src/change_validators/animation_rule_recordtype.ts
```diff
@@ -31,7 +31,7 @@ const invalidRecordTypeError = (instance: InstanceElement): ChangeError => (
     elemID: instance.elemID,
     severity: 'Error',
     message: 'Invalid AnimationRule RecordType',
-    detailedMessage: `In ${instance.elemID.getFullName()}, The RecordTypeId field is missing even though RecordTypeContext requires a RecordTypeId.`,
+    detailedMessage: `In ${instance.elemID.getFullName()}, The RecordTypeId field is missing even though RecordTypeContext requires a RecordTypeId.\nfor more details on 'animationRuleRecordType': https://help.salto.io/en/articles/7793653-deployment-preview-errors`,
```
amitgoldarb, line 34: We should probably pass the name to the CV in a more stable way. Possibly passing it as a "parameter" to the CV itself?
@ori-moisis do you think this would make sense in the context of adpater-api? Or should we create an extenstion type in SF?

Uri-Bar: already talked with ori about it, if the link ti the CV's article will be hard coded so the name should be as well. I dont think it worth expending the api

PR #4364, opened 2023-05-09
SALTO-2428 split NS filters to local & remote filters

dantal4:
---

---
_Release Notes_: 
Netsuite Adapter:
- Split filters to local & remote filters

---
_User Notifications_: 
None

### DIFF
--- a/packages/netsuite-adapter/src/filters/author_information/system_note.ts
+++ b/packages/netsuite-adapter/src/filters/author_information/system_note.ts
@@ -22,7 +22,7 @@ import Ajv from 'ajv'
 import moment from 'moment-timezone'
 import { TYPES_TO_INTERNAL_ID as ORIGINAL_TYPES_TO_INTERNAL_ID } from '../../data_elements/types'
 import NetsuiteClient from '../../client/client'
-import { FilterCreator, FilterWith } from '../../filter'
+import { RemoteFilterCreator } from '../../filter'
 import { getLastServerTime } from '../../server_time'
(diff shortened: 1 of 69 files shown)


packages/netsuite-adapter/src/filters/author_information/system_note.ts
```diff
@@ -239,8 +239,9 @@ const getCustomRecordsWithInternalIds = (elements: Element[]): Promise<InstanceE
     .filter(async instance => isCustomRecordType(await instance.getType()))
     .toArray()
 
-const filterCreator: FilterCreator = ({ client, config, elementsSource, elementsSourceIndex, isPartial }): FilterWith<'onFetch'> => ({
+const filterCreator: RemoteFilterCreator = ({ client, config, elementsSource, elementsSourceIndex, isPartial }) => ({
   name: 'systemNoteAuthorInformation',
+  remote: true,
```
omrilit, line 244: Why is it needed to specify that it's a remote filter?
Is there a way to enforce that all RemoteFilterCreator filters are `remote: true`?

omrilit (approved):
Looks great!
Had only one question, not blocking from merging

packages/netsuite-adapter/src/filters/author_information/system_note.ts
dantal4: It's for type-safety, that we use to make sure that local filters aren't treated as remote - see this https://github.com/salto-io/salto/pull/4345

PR #4366, opened 2023-05-10
Salto 3534 change validator incoming reference

adfineberg:
_Add a core change validation on removing an element with incoming references_

---

* Renamed an existing change validator for outgoing references

---
_Release Notes_: 
_Core_
* Add a validation for deploy plans that warns when removing an element with incoming references

---
_User Notifications_:


ori-moisis:
didn't review the code yet, but looking at the PR description:
release notes are meant to describe changes in functionality (e.g - "we no longer crash if an element changes in a specific way"), we don't need to mention implementation details and code refactoring in the release notes as they have no impact on users...

packages/adapter-components/src/deployment/change_validators/default_change_validators.ts
```diff
@@ -17,11 +17,11 @@
 import { ChangeValidator } from '@salto-io/adapter-api'
 import { values } from '@salto-io/lowerdash'
 import _ from 'lodash'
-import { createUnresolvedReferencesValidator } from './unresolved_references'
+import { createOutgoingUnresolvedReferencesValidator } from './outgoing_unresolved_references'
```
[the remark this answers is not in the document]
adfineberg: It is in its own commit, just not the first... thought of it in the middle of the work. Next time I'll rebase accordingly

packages/netsuite-adapter/src/change_validators/remove_sdf_elements.ts
```diff
@@ -57,15 +52,6 @@ const validateRemovableChange = async (
         detailedMessage: 'Can\'t remove this Custom Record Type. Try fetching and deploying again, or remove it in Netsuite UI',
       }
     }
-    // will be redundant once SALTO-3534 introduced
-    if (await hasInstances(element)) {
-      return {
-        elemID: element.elemID,
-        severity: 'Error',
```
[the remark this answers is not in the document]
adfineberg: Discussed with @omrilit - this is the desired behavior

packages/core/src/core/plan/change_validators/incoming_unresolved_references.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export const incomingUnresolvedReferencesValidator = (validationErrors: ReadonlyArray<SaltoError>)
+  : ChangeValidator => async changes => {
+  const unresolvedGroupErrors = groupUnresolvedRefsByTarget(validationErrors)
+  return changes
+    .filter(isRemovalOrModificationChange)
+    .flatMap(change => {
+      const group = unresolvedGroupErrors.find(
+        unresolvedGroupError => getChangeData(change).elemID.isEqual(unresolvedGroupError.elemID)
+      )
```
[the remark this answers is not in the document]
adfineberg: Fixed it, simplified everything. I'm sure this is what you originally meant 😨

packages/core/src/core/plan/change_validators/incoming_unresolved_references.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export const incomingUnresolvedReferencesValidator = (validationErrors: ReadonlyArray<SaltoError>)
+  : ChangeValidator => async changes => {
+  const unresolvedErrors = validationErrors.filter(isUnresolvedRefError)
+  return changes
+    .filter(isRemovalOrModificationChange)
+    .flatMap(change => {
+      const { elemID } = getChangeData(change)
+      const group = unresolvedErrors.filter(({ target }) => elemID.isEqual(target) || elemID.isParentOf(target))
+      return group.length === 0 ? [] : { change, group }
```
ori-moisis, line 34: return type should probably be consistent
```suggestion
      return group.length === 0 ? [] : [{ change, group }]
```

packages/core/src/core/plan/change_validators/incoming_unresolved_references.ts
```diff
@@ -0,0 +1,47 @@
...
+    })
+    .flatMap(({ change, group }) => {
```
ori-moisis, line 36: no need for this to be a flat map, it is just a map

packages/core/test/core/plan/change_validators/incoming_unresolved_references.test.ts
```diff
@@ -0,0 +1,70 @@
...
+  })
+
+  const secondInstance = new InstanceElement('instance2', type)
+
+  it('should not return an error when the unresolved elemID is not in the plan', async () => {
+    const errors = await incomingUnresolvedReferencesValidator([firstInstanceError])([
+      toChange({ before: secondInstance }),
+    ])
+    expect(errors).toEqual([])
+  })
```
ori-moisis, line 40: 2 test style nits:
1. We should try to avoid having shared context between tests, all these `const ... = ` in the describe block represent shared context - they create elements that are used in multiple tests, tests may modify these elements and cause context leakage between tests
2. For readability it usually helps separating the scenario from the expected result, we usually do this such that the describe block starts with `describe('when <scenario>'`, that scenario is setup in `beforeEach` (using beforeEach for initializing the scenario helps ensure we don't share context between tests) and then the tests specify the expected result (`it('should <expected result>'`).

Setting up the structure from point 2 should help with point 1, the context that needs to be shared between the setup (`beforeEach`) and the tests (`it`) is explicitly written as `let errors: ...` and these are filled separately before each test, so tests are guaranteed to have their own "clean copy" of the context

This is not a strict rule and you can see there are multiple styles of tests in our code, just pointing this out in case you were not familiar with the style above

packages/core/test/core/plan/change_validators/incoming_unresolved_references.test.ts
```diff
@@ -0,0 +1,70 @@
...
+  it('should return an error on the nested element with a reference if a parent was removed', async () => {
+    const nestedElemID = firstInstance.elemID.createNestedID('nested', 'innerInstance1')
+    const nestedInstanceError = new wsErrors.UnresolvedReferenceValidationError({
+      elemID: firstInstance.elemID, target: nestedElemID,
+    })
+    const errors = await incomingUnresolvedReferencesValidator([nestedInstanceError])([
+      toChange({ before: firstInstance }),
+    ])
+    expect(errors.length).toEqual(1)
+    expect(errors[0].elemID).toEqual(nestedElemID)
```
ori-moisis, line 67: I'd expect the error to be on the removed instance, not on the reference target

ori-moisis (approved):

PR #4368, opened 2023-05-10
SALTO-4147: Add logs for ScriptRunner failures

IdoZakk:
Add try-catch so in case we get another typescript error we will know in which element it happened 

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-utils/src/walk_element.ts
+++ b/packages/adapter-utils/src/walk_element.ts
@@ -84,7 +84,7 @@ export const walkOnValue = (
     if (e instanceof ExitWalk) {
       return
     }
-    log.error(`Failed to walk on element ${elemId.getFullName()}: ${e}`)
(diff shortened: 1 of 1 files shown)


IdoZakk:
It is in WalkOnValue, but that was a great idea :-)

packages/adapter-utils/src/walk_element.ts
```diff
@@ -81,6 +84,7 @@ export const walkOnValue = (
     if (e instanceof ExitWalk) {
       return
     }
+    log.error(`Failed to walk on element ${elemId.getFullName()}: ${e}`)
```
alonstern, line 87: Maybe warning? it's possible that in some places in the code, we expect an error to be thrown from error and an error log will alert the on-call

alonstern (approved):
👍

PR #4369, opened 2023-05-10
SALTO-4085: prevent deploy of routing attribute value to happen in parallel

edenhassid:
prevent deploy of routing attribute value to happen in parallel

---

---
_Release Notes_: 
zendesk:
* Fix issue that prevented deploying multiple values under the same routing attribute

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/deployment.ts
+++ b/packages/zendesk-adapter/src/deployment.ts
@@ -166,7 +166,7 @@ export const deployChanges = async <T extends Change<ChangeDataType>>(
   return { errors, appliedChanges }
 }
 
-export const deployChangesSynchronously = async <T extends Change<ChangeDataType>>(
+export const deployChangesSequentially = async <T extends Change<ChangeDataType>>(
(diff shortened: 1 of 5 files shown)


packages/zendesk-adapter/src/deployment.ts
```diff
@@ -166,6 +166,26 @@ export const deployChanges = async <T extends Change<ChangeDataType>>(
   return { errors, appliedChanges }
 }
 
+export const deployChangesSynchronously = async <T extends Change<ChangeDataType>>(
```
netama, line 169: tiny nit: maybe `Sequentially` instead of `Synchronously`? since this is still done async (we're not blocking the event loop until the result is returned, which is what usually happens in sync functions)

netama (approved):
LGTM with a small naming nit 🎈 
let's make the release notes slightly more informative? maybe something like "Fix issue that prevented deploying multiple values under the same routing attribute"?

PR #4371, opened 2023-05-10
SALTO-4067: Added fieldsToOmit to dummy adapter

alonstern:
Added `fieldsToOmit` to dummy so it will be easier to test deploy that omit fields from instances

---
_Release Notes_: 
None

---
_User Notifications_: 
None


netama (approved):
LGTM, discussed naming IRL

PR #4372, opened 2023-05-10
SALTO-4150: Sanitize test failure output

benisraeladi:
Salesforce test failures will not show the character `\n` and instead go down a line as required

---

before:
image

after:
image

---
_Release Notes_: 
Salesforce adapter:
- Sanitize test failures to create newline instead of showing escaped character
---
_User Notifications_: _None_


tamtamirr (approved):
LGTM 🔥

PR #4375, opened 2023-05-11
SALTO-3145: add warning when default locale is not english

edenhassid:
add fetch warning when default locale is not english

---

---
_Release Notes_: 
zendesk:
*  add fetch warning when default locale is not english

---
_User Notifications_: 
None


packages/zendesk-adapter/src/filters/locale.ts
```diff
@@ -0,0 +1,53 @@
...
+      return { errors: [] }
+    }
+
+    const warnings = defaultLocale.value.locale === 'en-US'
+      ? []
+      : getWarningsForLocale()
+    return { errors: warnings }
+  },
+})
+export default filterCreator
```
netama, line 53: don't we also need a change in adapter.ts to use this filter?

packages/zendesk-adapter/src/filters/locale.ts
```diff
@@ -0,0 +1,53 @@
...
+
+const getWarningsForLocale = (
+): SaltoError[] => [{
+  message: 'Please be aware that your Zendesk account\'s default locale is not set to English, which may impact your ability to compare environments with different default locales',
+  severity: 'Warning',
+}]
+
+/**
+ * This filter checks that the default locale is set to en-US, if not it will raise a warning. We have seen that the
+ * default local determines the language of different values and therefore may affect the elemId.
```
netama, line 32: 1. nit: `local` -> `locale`
2. I agree the main issue is that it affects the elem id, but not sure how clear it is out of the context of the ticket :) maybe elaborate a bit more - that this can cause elements to unintentionally appear as removed/added when comparing environments

packages/zendesk-adapter/src/filters/locale.ts
```diff
@@ -0,0 +1,53 @@
...
+import { logger } from '@salto-io/logging'
+import { FilterCreator } from '../filter'
+import { LOCALE_TYPE_NAME } from '../constants'
+
+const log = logger(module)
+
+
+const getWarningsForLocale = (
+): SaltoError[] => [{
+  message: 'Please be aware that your Zendesk account\'s default locale is not set to English, which may impact your ability to compare environments with different default locales',
```
netama, line 26: nit: we're checking en-US, we should mention it here as well (e.g. for UK english this is not that clear)

netama (approved):
LGTM with some nits (+ I think this may not be used yet?)

packages/zendesk-adapter/src/filters/locale.ts
edenhassid: right I missed that 😅

PR #4376, opened 2023-05-11
SALTO-4125 Update parser to support very long string contents

adfineberg:
_Update parser (lexer) to support very long string contents_

---

* The previous version of the single line string content was using a complex regex, which cause errors for extremely long strings.
This fix optimizes it by splitting the way the lexer tokenizes string parts.
---
_Release Notes_: 
_Parser_
* Update parser to support very long string contents
---
_User Notifications_:

### DIFF
--- a/packages/workspace/src/parser/internal/native/lexer.ts
+++ b/packages/workspace/src/parser/internal/native/lexer.ts
@@ -24,6 +24,7 @@ export const TOKEN_TYPES = {
   OCURLY: 'oCurly',
   CCURLY: 'cCurly',
   DOUBLE_QUOTES: 'dq',
+  ESCAPE: 'escape',
   WHITESPACE: 'whitespace',
   EQUAL: 'equal',
   NUMBER: 'number',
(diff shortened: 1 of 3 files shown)


packages/workspace/src/parser/internal/native/lexer.ts
```diff
@@ -71,7 +72,8 @@ export const rules: Record<string, moo.Rules> = {
   string: {
     [TOKEN_TYPES.REFERENCE]: { match: new RegExp(`\\$\\{[ \\t]*${WORD_PART}[ \\t]*\\}`), value: s => s.slice(2, -1).trim() },
     [TOKEN_TYPES.DOUBLE_QUOTES]: { match: '"', pop: 1 },
-    [TOKEN_TYPES.CONTENT]: { match: /(?:[^"\\\r\n]|\\.)+?(?="|\r|\n|\$\{)/, lineBreaks: false },
```
alonstern, line 74: Do you have an intuition of what made this regex slow? it's just the extra cases it needed to check in the brackets?

alonstern (approved):
👌

PR #4380, opened 2023-05-15
SALTO-2428 separate SDF parser code from SDF client

dantal4:
In order to be able to load elements from existing SDF projects.

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


omrilit (approved):
Looks great!
If you are already decoupling the code here, I would consider to separate also to sdf_transformer. WDYT?

PR #4384, opened 2023-05-15
SALTO-4156 - Salesforce: Switch the message and detailed message of the fullname_changed CV

tomsellek:
See PR title

---

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A


ori-moisis (approved):
I think "N/A" is not a conventional text in the release notes section and will therefore not be processed correctly by the release script.
please change this to `_None_` for the user notifications, and I think fixing the message can be added to the release notes actually

PR #4385, opened 2023-05-15
SALTO-4128 - Salesforce: Improve logging of email template attachment handling

tomsellek:
If there are attachment and we don't store them in a a static file, log some information about them

---

We want to understand why in cases like the one described in the ticket we end up with attachment contents inlined in the NaCl file instead of stored in a static file.

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A


packages/salesforce-adapter/src/filters/email_template_static_files.ts
```diff
@@ -65,25 +65,37 @@ const findFolderPath = (instance: InstanceElement): string =>
...
+        instance.value.attachments.forEach(attachment => {
+          attachment.content = createStaticFile(
+            // attachment.content type is a string before the creation of the static file
+            folderPath, attachment.name, attachment.content as string
+          )
+        })
+      } else if (!Array.isArray(instance.value.attachments)) {
+        log.debug('email template attachments are not an array: %s',
+          safeJsonStringify(instance.value.attachments,
+            (_key, value) => (_.isString(value) ? _.truncate(value) : value)))
```
ori-moisis, line 90: probably would be better to use `inspect` and not `safeJsonStringify` here, we've seen performance issues in other adapters.
#4267 hasn't been merged yet but some places were already changed after they "blew up" in certain scenarios

I think you don't even have to call it directly, could just replace this with
```suggestion
        log.debug('email template attachments are not an array: %o', instance.value.attachments)
```

packages/salesforce-adapter/src/filters/email_template_static_files.ts
```diff
@@ -65,25 +65,37 @@ const findFolderPath = (instance: InstanceElement): string =>
...
+      } else if (instance.value.attachments.length > 0) {
+        log.debug('email template attachments exist, but failed JOI validation: %s',
+          safeJsonStringify(instance.value.attachments,
+            (_key, value) => (_.isString(value) ? _.truncate(value) : value)))
+      }
     }
+  } catch (err) {
+    log.warn('Error when handling email template attachments: %o', err.message)
```
ori-moisis, line 98: 1. is err.message not a string? why the `%o`?
2. this is handling a specific instance, might be useful to know which one if we reach this code? (which, I must admit I am not sure how that would happen...)
3. the try catch block would catch all the handling of the instance, not just the code dealing with attachments...

packages/salesforce-adapter/src/filters/email_template_static_files.ts
```diff
@@ -65,25 +65,37 @@ const findFolderPath = (instance: InstanceElement): string =>
...
   }
```
amitgoldarb, line 99: We aren't throwing an exception here, if we were I think the fetch would be failing more violently.
I think that we can pass an errorMessage to the creation of isEmailAttachmentsArray which will give us all of the value we need.
We are failing the joi validation, what we need is the error message from the schema guard validation.
I think passing a parameter to createSchemGuard will probably give us all of the value of these debug logs. WDYT?

amitgoldarb (commented):
Added a comment. I think we might be able to reach the same debug value in a simpler way.

packages/salesforce-adapter/src/filters/email_template_static_files.ts
tomsellek: Won't this end up logging the entire contents, which may be several MB in size?

ori-moisis: exactly the opposite
inspect has a bunch of safeties in its options (max depth, max string length, max keys, etc) that `safeJsonStringify` doesn't

packages/salesforce-adapter/src/filters/email_template_static_files.ts
tomsellek: 1. Is it always? Then I'll change it to %s.
2. Ack
3. Ack

packages/salesforce-adapter/src/filters/email_template_static_files.ts
tomsellek: Makes sense. However, I tried using `inspect` with the `maxStringLength` option, but that option was added in Node.js v12.17 and we currently use `@types/node` v12.7.1 so the code fails to compile. Is it worth upgrading `@types/node` in our project for this change?

ori-moisis: you don't need to call `inspect` explicitly
`%o` does that for you and the defaults are good enough

tomsellek: Docs say the default is no limit on string length though?

> maxStringLength <integer> Specifies the maximum number of characters to include when formatting. Set to null or Infinity to show all elements. Set to 0 or negative to show no characters. Default: Infinity.

ori-moisis: huh, you are right, my bad
I think we can change our node types to at least the latest 12.x.x
hopefully soon we'll be able to bump it beyond that

tomsellek: I tried and got some type errors in jsforce-types. I'm thinking it's not worth it.

amitgoldarb (approved):
LGTM!
Thanks for this!

PR #4386, opened 2023-05-15
SALTO-4065: Remove specific duplicate project roles

IdoZakk:
Omitting duplicate a specific built-in project role that can be caused by deleting team-managed project
---

See details in the Jira ticket
This is a bandage fix for a Jira bug.
When you create a team-managed project several project-scoped roles are created by default, including atlassian-addons-project-access
When you delete a team-managed project all its project-scope roles are not really deleted.
They keep appearing in the roles API answer, but without the project scope. They are not visible in the UI
This fix will only address atlassian-addons-project-access as it is a built-in role, will always be there
and the global role will (hopefully) always have the lower id

The entire feature is under a config flag: fetch->removeDuplicateProjectRoles 

---
_Release Notes_: 
Jira Adapter:
* Provided a workaround for a Jira bug that prevented fetching a built-in project role

---
_User Notifications_: 
Jira Adapter:
* Some users might see a new project role that was previously omitted due to a Jira bug

### DIFF
--- a/packages/jira-adapter/src/filters/remove_specific_duplicate_roles.ts
+++ b/packages/jira-adapter/src/filters/remove_specific_duplicate_roles.ts
@@ -43,8 +43,7 @@ const filter: FilterCreator = ({ config }) => ({
       return
     }
     // as this is a built-in role, we assume that the one with the lowest id is the global one
-    const originalElementId = builtInProjectRoles
-      .reduce((minVal, instance) => Math.min(minVal, instance.value.id), builtInProjectRoles[0].value.id)
+    const originalElementId = _.minBy(builtInProjectRoles, instance => instance.value.id)?.value.id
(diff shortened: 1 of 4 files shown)


IdoZakk:
Let me know if you think we should also add noise reduction

packages/jira-adapter/src/filters/remove_specific_duplicate_roles.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+    const builtInProjectRoles = elements
+      .filter(isInstanceElement)
+      .filter(instance => instance.elemID.typeName === PROJECT_ROLE_TYPE
+        && instance.value.name === BUILT_IN_PROJECT_ROLE_NAME)
+    if (builtInProjectRoles.length <= 1) {
+      return
+    }
+    // as this is a built-in role, we assume that the one with the lowest id is the global one
+    const originalElementId = builtInProjectRoles
```
alonstern, line 46: You can use `_.minBy`

alonstern (approved):
👌

PR #4387, opened 2023-05-16
SALTO-4171 - Add log.time to selectElementIdsByTraversal & getEnvsDeletionsDiff

adamro:
Add log.time to selectElementIdsByTraversal & getEnvsDeletionsDiff

No logic was changed even tho git has trouble identifying it :\ 

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/core/src/core/diff.ts
+++ b/packages/core/src/core/diff.ts
@@ -189,24 +189,28 @@ export const getEnvsDeletionsDiff = async (
   sourceElemIds: ElemID[],
   envs: ReadonlyArray<string>,
   selectors: ElementSelector[]
-): Promise<Record<string, ElemID[]>> => {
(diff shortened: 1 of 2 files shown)


packages/core/src/core/diff.ts
```diff
@@ -189,24 +189,28 @@ export const getEnvsDeletionsDiff = async (
...
-        env,
-        await awu(await workspace.getElementIdsBySelectors(
-          selectors,
-          { source: 'env', envName: env },
-          true
-        )).toArray(),
-      ]
-  ).toArray())
+): Promise<Record<string, ElemID[]>> => (
+  log.time(
```
dorsev, line 193: Nit: can you `log.time` outside of the function, and not inside? I think it can improve the diff.


```typescript
export const getEnvsDeletionsDiff = log.time(async (......), 'xxx')
```

dorsev (approved):
LGTM

PR #4388, opened 2023-05-16
SALTO-3791 Support deletion in partial fetch - netsuite side

kami-delaroz:
Netsuite side - support deletion in partial fetch flow
---

None

---
_Release Notes_: 
will be able to see deletion changes in partial fetch flow

---
_User Notifications_: 
None

### DIFF
--- a/packages/netsuite-adapter/src/adapter.ts
+++ b/packages/netsuite-adapter/src/adapter.ts
@@ -16,10 +16,10 @@
 import {
   FetchResult, isInstanceElement, AdapterOperations, DeployResult, DeployOptions,
   ElemIdGetter, ReadOnlyElementsSource, ProgressReporter,
-  FetchOptions, Field, BuiltinTypes, DeployModifiers, getChangeData,
+  FetchOptions, Field, BuiltinTypes, DeployModifiers, getChangeData, ElemID,
 } from '@salto-io/adapter-api'
 import _ from 'lodash'
-import { collections, values } from '@salto-io/lowerdash'
+import { collections, values as lowerdashValues } from '@salto-io/lowerdash'
 import { logger } from '@salto-io/logging'
 import { filter } from '@salto-io/adapter-utils'
 import {
@@ -81,6 +81,7 @@ import { createCustomRecordTypes } from './custom_records/custom_record_type'
 
 const { makeArray } = collections.array
 const { awu } = collections.asynciterable
+const { isDefined } = lowerdashValues
 
 const log = logger(module)
 
@@ -257,9 +258,16 @@ export default class NetsuiteAdapter implements AdapterOperations {
     const dataElementsPromise = getDataElements(this.client, fetchQuery, this.getElemIdFunc)
     const configRecordsPromise = this.client.getConfigRecords()
 
+    // we calculate deleted elements only in partial-fetch mode
+    const currentElementsForDeletionCalculation = !isPartial ? []
+      : (await this.getCurrentElements(true)).filter(item => fetchQuery.isTypeMatch(item.elemID.typeName))
+
     const getCustomObjectsResult = this.client.getCustomObjects(
       getStandardTypesNames(),
-      updatedFetchQuery
+      updatedFetchQuery,
+      isPartial,
+      fetchQuery,
+      currentElementsForDeletionCalculation,
     )
     const importFileCabinetResult = this.client.importFileCabinetContent(
       updatedFetchQuery,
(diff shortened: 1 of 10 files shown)


kami-delaroz:
I'm still missing some unit-tests. Thinking on the right place to include them

packages/netsuite-adapter/src/adapter.ts
```diff
@@ -257,9 +258,16 @@ export default class NetsuiteAdapter implements AdapterOperations {
     const dataElementsPromise = getDataElements(this.client, fetchQuery, this.getElemIdFunc)
     const configRecordsPromise = this.client.getConfigRecords()
 
+    // we calculate deleted elements only in partial-fetch mode
```
omrilit, line 261: Can't we do that also for QuickFetch?

packages/netsuite-adapter/src/client/client.ts
```diff
@@ -132,9 +132,17 @@ export default class NetsuiteClient {
   }
 
   @NetsuiteClient.logDecorator
-  async getCustomObjects(typeNames: string[], query: NetsuiteQuery):
+  async getCustomObjects(
+    typeNames: string[],
+    query: NetsuiteQuery,
+    isPartial: boolean,
+    originFetchQuery: NetsuiteQuery,
+    currentElementsForDeletionCalculation: {elemID: ElemID; serviceID: string}[],
```
dantal4, line 140: In continuation to my previous comment - I think that it's better to pass here `ObjectID[]`.

packages/netsuite-adapter/src/client/sdf_client.ts
```diff
@@ -474,6 +486,28 @@ export default class SdfClient {
...
+      typeNames.filter(originFetchQuery.isTypeMatch),
+      suiteAppId,
+    )).filter(originFetchQuery.isObjectMatch)
+    const serviceIdsGroupByTypes = _(serviceInstancesIds)
+      .groupBy(objectId => objectId.type)
+      .mapValues(objectIds => objectIds.map(({ instanceId }) => instanceId))
+      .value()
+
+    return currentElements
+      .filter(item => !serviceIdsGroupByTypes[item.elemID.typeName].includes(item.serviceID))
```
dantal4, line 507: custom record types needs a special treat here - the type name is actually not the "NS type name" which is "customrecordtype".

packages/netsuite-adapter/test/adapter.test.ts
```diff
@@ -135,6 +135,7 @@ describe('Adapter', () => {
     client.getCustomObjects = mockFunction<NetsuiteClient['getCustomObjects']>()
       .mockResolvedValue({
         elements: [],
+        deletedElements: [],
```
dantal4, line 138: Can you add a test for the deletedElements flow (in `adapter.ts`)? To see that `currentElementsForDeletionCalculation` is calculated correctly and to see that deletedElements is returned correctly.

packages/netsuite-adapter/test/client/sdf_client.test.ts
```diff
@@ -858,14 +870,50 @@ describe('sdf client', () => {
...
+          if (context.commandName === COMMANDS.IMPORT_OBJECTS) {
+            return Promise.resolve({
+              isSuccess: () => true,
+              data: { failedImports: [] },
+            })
+          }
+          return Promise.resolve({ isSuccess: () => true })
+        })
+        const currentElementsForDeletionCalculation = [
+          { elemID: new ElemID('salto', 'addressForm', 'type'), serviceID: 'a' },
```
dantal4, line 905: only custom record types have service ids (and not other types)...

dantal4 (commented):
Great work!

dantal4:
We can also support deletions detection of data & custom record instances - because we fetch all of them in partial fetch too.

packages/netsuite-adapter/src/adapter.ts
```diff
@@ -342,9 +351,21 @@ export default class NetsuiteAdapter implements AdapterOperations {
       failedFilePaths,
       failedTypes,
       elements,
+      deletedElements,
     }
   }
 
+  private async getCurrentElements(isPartial: boolean) : Promise<{elemID: ElemID; serviceID: string}[]> {
+    const serviceIdRecords = (await createElementsSourceIndex(this.elementsSource, isPartial)
```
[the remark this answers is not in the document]
dantal4: Thinking about it again, I think that we should create the elementsSourceIndex inside  `createFiltersRunner` - and pass to it the list of removed elements, so we don't calculate them too. Otherwise we may create a reference to an element in the element references filter, but then remove the referenced element and we'll have a broken reference.

PR #4392, opened 2023-05-17
SALTO-4081/orphan contexts

guyhamo:
fixed bug causing us to try and remove an already removed context (which happens when a user removes a field but keeps the nacls of the contexts and then removes them before the next fetch.

---

---
_Release Notes_: 
jira adapter:
* fixed bug with removing field contexts after the field was already removed.

---
_User Notifications_:

### DIFF
--- a/packages/jira-adapter/src/dependency_changers/global_field_contexts.ts
+++ b/packages/jira-adapter/src/dependency_changers/global_field_contexts.ts
@@ -34,7 +34,7 @@ export const globalFieldContextsDependencyChanger: DependencyChanger = async cha
 
   const filteredGlobalContextChanges = globalContextChanges.filter(({ change }) => {
     try {
-      return getParent(getChangeData(change)).elemID.getFullName() !== undefined
+      return getParent(getChangeData(change)) !== undefined
     } catch {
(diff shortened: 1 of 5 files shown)


packages/jira-adapter/src/dependency_changers/global_field_contexts.ts
```diff
@@ -32,8 +32,16 @@ export const globalFieldContextsDependencyChanger: DependencyChanger = async cha
     .filter(({ change }) => _.isEmpty(getChangeData(change).value.projectIds)
       || _.isEmpty(getChangeData(change).value.issueTypeIds))
 
+  const filteredGlobalContextChanges = globalContextChanges.filter(({ change }) => {
```
alonstern, line 35: Maybe add to the utils file in Jira `isThereValidParent` and use it in both places?

alonstern (approved):
Coal

packages/jira-adapter/src/dependency_changers/global_field_contexts.ts
guyhamo: 👍

PR #4393, opened 2023-05-17
Salto-4054: unable to deploy statistical accounts in netsuite

Shulik95:
_fix bug preventing deployments of statistical accounts_

---

_None_

---
_Release Notes_: 

---
_User Notifications_: 
_id field will be added in the unitsType, mainAddress and state fields_

### DIFF
--- a/packages/netsuite-adapter/src/filters/data_instances_internal_id.ts
+++ b/packages/netsuite-adapter/src/filters/data_instances_internal_id.ts
@@ -65,38 +65,35 @@ const filterCreator: LocalFilterCreator = () => ({
 
       const originalInternalId = value[INTERNAL_ID]
       const fieldType = await field?.getType()
-      if (shouldUseIdField(fieldType, path) || !hasInternalIdHiddenField(fieldType)) {
+      const isInsideList = path.getFullNameParts().some(part => isNumberStr(part))
+
+      if (!(isObjectType(fieldType)
+        && (isInsideList
+          || isStandardType(fieldType)
+          || isFileCabinetType(fieldType)))) {
         value[
           shouldUseIdField(fieldType, path) ? ID_FIELD : INTERNAL_ID
         ] = ACCOUNT_SPECIFIC_VALUE
         delete value.typeId
+        return value
+      }
+      const instanceName = getSubInstanceName(path, originalInternalId)
+      if (!(instanceName in newInstancesMap)) {
+        const newInstance = new InstanceElement(
+          instanceName,
+          // If the fieldType is an SDF type we replace it with RecordRef to avoid validation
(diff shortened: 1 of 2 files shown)


omrilit (commented):
@dantal4 please review this one.
My main concern is the impact on existing WSs. Does it have footpront outside of the Types folder?

Shulik95:
> @dantal4 please review this one. My main concern is the impact on existing WSs. Does it have footprint outside of the Types folder?
@omrilit It will affect existing WS since the id field will be added in some instance but I've implemented noise suppression for these changes - https://github.com/salto-io/salto_private/pull/5678

Shulik95 (commented):
@omrilit It will affect existing WS but I've also opened a pr for noise suppression for these changes - https://github.com/salto-io/salto_private/pull/5678

omrilit:
> @omrilit It will affect existing WS but I've also opened a pr for noise suppression for these changes - [salto-io/salto_private#5678](https://github.com/salto-io/salto_private/pull/5678)

We use the noise suppression to omit changes from the monitoring notifications, to avoid alerting users on noisy changes.
The thing I am concerned about is the Audit users, which will have to commit the affected files to Git, and we should always strive to notify them in advance so they could push the uncommitted files before we deploy the new version and then fetch and push only the noisy changes in a dedicated commit.
For example you can see https://salto-io.atlassian.net/browse/PROD-76.
I think that it would be great if we could have an example of the exact changes the users will experience so we could share it with them or with CS in advance (screenshots / a commit in a git repo with the changes in one of our accounts).

packages/netsuite-adapter/src/filters/data_instances_internal_id.ts
```diff
@@ -39,10 +39,8 @@ const hasInternalIdHiddenField = (type: unknown): boolean =>
   && type.fields[INTERNAL_ID]
   && type.fields[INTERNAL_ID].annotations[CORE_ANNOTATIONS.HIDDEN_VALUE]
 
-const shouldUseIdField = (fieldType: TypeElement | undefined, path: ElemID): boolean =>
-  fieldType?.elemID.name === RECORD_REF || (
-    isCustomFieldName(path.name) && hasInternalIdHiddenField(fieldType)
-  )
+const shouldUseIdField = (fieldType: TypeElement | undefined): boolean =>
+  fieldType?.elemID.name === RECORD_REF || hasInternalIdHiddenField(fieldType)
```
dantal4, line 43: Worth checking if `netsuite.recordRef` has a hidden internal id field. If it has, we can just use `hasInternalIdHiddenField(fieldType)`

dantal4 (approved):
Great work!
Please add a test for the case that this PR now supports.

PR #4394, opened 2023-05-17
SALTO-4064 Fix circular dependencies in suitescript files deploy

dantal4:
Suitescript files generated dependencies can cause a circular dependencies error between SDF deploy group and SuiteApp FileCabinet deploy group, although there is no real dependency between them (there is a real dependency between the script instance and the file instance, so the file instance should be deployed before).

In order to solve it, we remove the dependencies that file instances have. This should have no harm (as the file instance itself don’t really depend on anything but its parent folder).

---

---
_Release Notes_: 
Netsuite Adapter:
- Fix circular dependencies in suitescript files deploy

---
_User Notifications_: 
None

### DIFF
--- a/packages/netsuite-adapter/src/dependency_changer.ts
+++ b/packages/netsuite-adapter/src/dependency_changer.ts
@@ -1,0 +1,56 @@
+/*
+*                      Copyright 2023 Salto Labs Ltd.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+import { values } from '@salto-io/lowerdash'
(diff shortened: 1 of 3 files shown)


packages/netsuite-adapter/src/dependency_changer.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+  const generatedDependencies: DetailedDependency[] = element.annotations[CORE_ANNOTATIONS.GENERATED_DEPENDENCIES] ?? []
+
+  const [parentFolder] = element.annotations[CORE_ANNOTATIONS.PARENT] ?? []
+  const parentFolderElemId = isReferenceExpression(parentFolder)
+    ? parentFolder.elemID.createBaseID().parent
+    : undefined
+
+  return generatedDependencies
+    .map(({ reference }) => reference.elemID.createBaseID().parent)
```
omrilit, line 35: In the case where reference is an UnresolvedReference, this will work right?
In addition, no one guarantees that this value is really a DetailedDependency and not just a string, in case the user edited it, but this is not interesting to address... 🤓

omrilit, line 35: Just making sure that in case of a reference to an inner value of an element it would work, e.g. to a `scriptid`, right?

dantal4: The UnresolvedReference part would be the `ref.value`, so `ref` is still a ReferenceExpression 👍

dantal4: Yes, that's why we take the base level elem id.

PR #4395, opened 2023-05-18
SALTO-4011/add display name to error

guyhamo:
add display names of missing users to the error message.

---

tested with cloud and dc
cloud has both values in this format id: "display name", ....
dc is not changed with just id.

---
_Release Notes_: 
jira adapter:
* add display name to user errors with visible id.

---
_User Notifications_:

### DIFF
--- a/packages/jira-adapter/src/change_validators/account_id.ts
+++ b/packages/jira-adapter/src/change_validators/account_id.ts
@@ -39,6 +39,9 @@ type missingUsersDetails = {
   displayName?: string
 }
 
+const formatMissingUsers = (missingUsers: Set<missingUsersDetails>): string =>
+  `${Array.from(missingUsers).map(missing => (missing.displayName ? `${missing.accountId}: "${missing.displayName}"` : missing.accountId)).join(',')}`
+
 const noDisplayNameChangeError = (
   elemId: ElemID,
 ): ChangeError => ({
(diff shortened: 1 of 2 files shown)


packages/jira-adapter/src/change_validators/account_id.ts
```diff
@@ -154,7 +162,7 @@ const checkAndAddChangeErrors = (
   }
 
   if (!Object.prototype.hasOwnProperty.call(userMap, accountId)) {
-    missingUsers.add(accountId)
+    missingUsers.add({ accountId, displayName: currentDisplayName })
```
alonstern, line 165: Objects don't work inside sets, meaning it won't remove duplications as you add them, you will need to add them to a list and then use something like `_.uniqBy`

guyhamo: 👍

packages/jira-adapter/src/change_validators/account_id.ts
```diff
@@ -196,24 +207,25 @@ const createChangeErrorsForAccountIdIssues = (
       missingDisplayNamePaths
     ), config) })
 
+  const missingUsersSet = new Set(_.uniqBy(missingUsers, formatMissingUser))
```
alonstern, line 210: no reason to work with a set now, I think you can just pass it as a list in the filter

alonstern (approved):
👌

packages/jira-adapter/src/change_validators/account_id.ts
guyhamo: 👍

PR #4398, opened 2023-05-18
SALTO-4192 log HTTP response on errors as well

netama:
Trace the HTTP responses even when there are HTTP errors

---
_Release Notes_: 
None

---
_User Notifications_: 
None


shir-reifenberg (approved):
🚀

PR #4400, opened 2023-05-18
SALTO-3751 - Salesforce: Improve modelling of track feed history and enable feeds annotations(v2)

tomsellek:
Keep a list of fields whose history/feed-history is tracked at the object level instead of keeping Salesforce's per-field annotation.

---

Salesforce enables history tracking per field + an additional "master switch" per object. This causes a lot of noise when deploying objects between environments where history tracking is enabled to environments where it's disabled (basically every single field is marked as modified). We want to make it easier to determine what the actual change is, so we centralize the list of tracked field as an annotation in the object.

---
_Release Notes_: 
Salesforce Adapter:

Improved modelling of History and Feed History Tracking on Standard and Custom Objects.

---
_User Notifications_: 
Salesforce Adapter:

### DIFF
--- a/packages/salesforce-adapter/src/filters/centralize_tracking_info.ts
+++ b/packages/salesforce-adapter/src/filters/centralize_tracking_info.ts
@@ -110,7 +110,7 @@ const createHistoryTrackingFieldChange = (
 const filter: LocalFilterCreator = () => {
   let fieldsWithSyntheticChanges: Set<string> = new Set()
   return {
-    name: 'history_tracking',
+    name: 'centralizeTrackingInfo',
     onFetch: async elements => {
       await awu(elements)
         .filter(isObjectType)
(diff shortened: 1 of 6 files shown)


packages/salesforce-adapter/src/filters/centralize_tracking_info.ts
```diff
@@ -0,0 +1,277 @@
...
+      await awu(trackedFieldsDefinitions)
+        .forEach(async trackingDef => {
+          // Added object types - set the annotations on the type and its fields
+          await distributeTrackingInfoInAddedObjectTypes(trackingDef)
+
+          // Added or modified fields - set the annotations on the fields
+          const namesOfFieldsThatChanged = new Set(await updateAnnotationsOnChangedFields(trackingDef))
+
+          // Existing object types that changed
+          const fieldChanges = await distributeTrackingInfoInModifiedObjectTypes(trackingDef, namesOfFieldsThatChanged)
```
ori-moisis, line 230: I think extracting the functions here actually makes the code less readable
the comments make less sense when the implementation isn't here and the return values are kind of strange (e.g - why does one "distribute" function return nothin, one returns names and one returns changes?)

I think the previous version was better.

ori-moisis (approved):
was a little hard to track the changes in `preDeploy` but I assume they follow the same pattern as the rest of the changes

PR #4404, opened 2023-05-21
SALTO-4042: Add missing custom fields to type definitions

shoshana-houri:
Entity custom field, item custom fields and CRM (event) custom field are not added to the type definitions of the instances they applied to. This PR solves it by modifying the  hardcoded type definitions to camel case.

---

---
_Release Notes_: 

---
_User Notifications_: 
Some custom fields (entity custom field, item custom field and crm (event) custom field) were missing from type definitions and will now appear in the relevant types.

Here are some before-after examples:
image
image
image


omrilit (commented):
Looks great!
I am not approving so it won't get merged accidentally until we decide on the deployment to production date.

In addition, could you please also share some screenshots (here & in the PROD ticket) of the diff that will occur to existing WSs to ease the external messaging.

Shulik95 (approved):

PR #4408, opened 2023-05-21
SALTO-4189 File cabinet partial fetch unit tests

kami-delaroz:
File cabinet partial fetch unit tests

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
None


dantal4 (approved):
💯

PR #4409, opened 2023-05-21
SALTO-4160/remove unknown fields on deploy

guyhamo:
made sure we dont put back field items that were not found in the element source (and deployed) after deploying field configurations.

---

there was a test literarily named according to what i need but the test did the opposite. i changed the test to actually work.
i didnt add any other tests.

---
_Release Notes_: 
jira_adapter:
* when deploying field configurations, if a field did not exist salto will remove it from the nacl on deploy.

---
_User Notifications_:


alonstern (approved):
👍

PR #4412, opened 2023-05-21
SALTO-4201: publish only built files to npm

ori-moisis:
No need to include the non-built files and the tests when publishing to npm, users of the packages only need the built .js files and the type definitions...

---

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


packages/salesforce-adapter/package.json
```diff
@@ -11,10 +11,9 @@
     "access": "public"
   },
   "files": [
-    "dist",
-    "src",
-    "e2e_test",
-    "index.ts"
+    "dist/src",
+    "dist/index.*",
+    "dist/e2e_test"
```
ori-moisis, line 16: leaving the e2e_test folder here because the jest_environment stuff is unofficially exported from there

amitizle (approved):
❤️

PR #4414, opened 2023-05-22
SALTO-4193: Suppress 404 on AppUserSchema

shir-reifenberg:
Fix bug causing built in apps to disappear from fetch as they don't have AppUserSchema which caused 404 when trying to fetch those

---
_Release Notes_: 
_Okta adapter_:
- Fix a bug in built-in applications fetch

---
_User Notifications_: 

_Okta adapter_:
- Built-in Application will be added on the next fetch

### DIFF
--- a/packages/okta-adapter/src/client/client.ts
+++ b/packages/okta-adapter/src/client/client.ts
@@ -39,6 +39,7 @@ const DEFAULT_PAGE_SIZE: Required<clientUtils.ClientPageSizeConfig> = {
   get: 50,
 }
 
+// The expression match AppUserSchema endpoint used for fetch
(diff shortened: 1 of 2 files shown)


packages/okta-adapter/src/client/client.ts
```diff
@@ -37,6 +39,8 @@ const DEFAULT_PAGE_SIZE: Required<clientUtils.ClientPageSizeConfig> = {
   get: 50,
 }
 
+const APP_USER_SCHEMA_URL = /(\/api\/v1\/meta\/schemas\/apps\/[a-zA-Z0-9]+\/default)/
```
edenhassid, line 42: nit: maybe add a comment of what this should catch

edenhassid (approved):
looks good! 🔥 added a small nit

PR #4415, opened 2023-05-23
SALTO-3817: Add postAction.showOnFailure flag

elentok-salto:
The post-deploy action messages for the salesforce adapter should be shown if the change failed but the other ones it should not.

To deal with this I've added a flag named `showOnFailure` to the `DeployAction` type (by creating a subtype called `PostDeployAction`).

When showing the post deploy messages I filter out ChangeError instances of failed changes where the `showOnFailure` flag is true.

Once this is merged the next step is to handle this flag in the SASS part.

---

---
_Release Notes_: 
CLI:
- Post-deploy actions of failed changes will no longer be shown to the user (except for Salesforce CPQ failures).

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-api/src/adapter.ts
+++ b/packages/adapter-api/src/adapter.ts
@@ -79,7 +79,7 @@ export type DeployAction = {
 }
 
 export type PostDeployAction = DeployAction & {
-  showOnFailure: boolean
+  showOnFailure?: boolean
 }
 
 export type DeployActions = {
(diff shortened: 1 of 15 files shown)


packages/adapter-api/src/adapter.ts
```diff
@@ -78,9 +78,13 @@ export type DeployAction = {
   documentationURL?: string
 }
 
+export type PostDeployAction = DeployAction & {
+  showOnFailure: boolean
```
alonstern, line 82: I suspect this addition will break the build in the SaaS. I think we should make it optional at least until we change the SaaS to support this flag and then we can change it back to required

elentok-salto: Good point! fixed.

alonstern (approved):
Looks great 👍

PR #4416, opened 2023-05-23
SALTO-4050 remove addNamespacePrefixToFullName

benisraeladi:
remove addNamespacePrefixToFullName flag and update code flow to give the output we expect

---
Removed the configuration item since it's now unnecessary, but I'm not sold on that given that there may be people using it

---
_Release Notes_: 
Salesforce Adapter:
- addNamespacePrefixToFullName fetch config flag has been removed
---
_User Notifications_: 
Salesforce Adapter:
- addNamespacePrefixToFullName has been removed


packages/salesforce-adapter/src/fetch.ts
```diff
@@ -192,29 +192,33 @@ const getFullName = (obj: FileProperties, addNamespacePrefixToFullName?: boolean
...
-    if (obj.fullName.includes(API_NAME_SEPARATOR)) {
-    // Case for API name with an object <Object>.X where it should be <Object>.<namespace>__X
-      const [parentName, instanceName] = obj.fullName.split(API_NAME_SEPARATOR)
-      return `${parentName}${API_NAME_SEPARATOR}${namePrefix}${instanceName}`
-    }
+  const fullNameParts = obj.fullName.split(API_NAME_SEPARATOR)
+  const name = fullNameParts.slice(-1)[0]
+  const parentNames = fullNameParts.slice(0, -1)
+
+  // Objects whose parents are custom objects need to check only the element's name and not the entire name
```
ori-moisis, line 208: I think it is not clear why this is related to custom objects given that we split by `API_NAME_SEPARATOR`
if we want to keep this terminology I think we need to explain it a little more in the comment

ori-moisis (commented):
I think the PR description should be a little more explicit about what is about to change.
the fact we switched the flag default is kind of a deep implementation detail, this is actually supposed to solve an issue where we wouldn't fetch certain elements.
the fact that we are now fixing that bug should be the main point in the release notes IMO

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -192,29 +192,33 @@ const getFullName = (obj: FileProperties, addNamespacePrefixToFullName?: boolean
...
+  if (name.startsWith(namePrefix)) {
+    return obj.fullName
+  }
+
+  // In some cases, obj.fullName does not contain the namespace prefix even though
```
tomsellek, line 213: nit: You could replace this part (lines 213-128) with something like
```
return [...parentNames, `${namePrefix}${name}`].join(API_NAME_SEPARATOR)
```
which would eliminate the special case of `parentNames.length===0`

tomsellek (approved):
LGTM!

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -192,29 +192,33 @@ const getFullName = (obj: FileProperties, addNamespacePrefixToFullName?: boolean
     return obj.fullName
   }
 
-  if (obj.fullName.startsWith(namePrefix)) {
+  if (!addNamespacePrefixToFullName) {
+    log.debug('obj.fullName %s is missing namespace %s. Not adding because addNamespacePrefixToFullName is false. FileProps: %o', obj.fullName, obj.namespacePrefix, obj)
```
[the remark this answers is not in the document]
benisraeladi: It's a little spammy, but together with another log helps determine if there are any other edge cases

ori-moisis (approved):

PR #4418, opened 2023-05-23
SALTO-3490: Add ticket form check to the custom statuses enabled validator

elentok-salto:
…dator

The customStatusesEnabledValidator now shows a warning message when trying to change zendesk.ticket_form instances with `custom_statuses` specified in either `agent_conditions` or `end_user_conditions`.

---

---
_Release Notes_: 
ZenDesk adapter:

- Check that custom statuses are enabled when updating zendesk.ticket_form instances that include custom_statuses under agent_conditions or end_user_conditions.

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
+++ b/packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
@@ -13,13 +13,59 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
-import { Change, ChangeDataType, ChangeError, ChangeValidator, ElemID, getChangeData, isInstanceElement, ReadOnlyElementsSource } from '@salto-io/adapter-api'
+import { Change, ChangeDataType, ChangeError, ChangeValidator, ElemID, getChangeData, InstanceElement, isInstanceElement, ReadOnlyElementsSource } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
-import { Condition, isTicketFormInstance } from '../filters/ticket_form'
+import { createSchemeGuardForInstance } from '@salto-io/adapter-utils'
+import Joi from 'joi'
 import { ACCOUNT_FEATURES_TYPE_NAME, CUSTOM_STATUS_TYPE_NAME, TICKET_FORM_TYPE_NAME, ZENDESK } from '../constants'
 
 const log = logger(module)
 const errorMsg = (reason: string): string => `Failed to run customStatusesEnabledValidator because ${reason}`
+
+type ChildField = {
+  // eslint-disable-next-line camelcase
+  required_on_statuses: {
+    type: string
+    statuses?: string[]
+    // eslint-disable-next-line camelcase
+    custom_statuses?: unknown[]
+  }
+}
+
+const CHILD_FIELD_SCHEMA = Joi.object({
+  required_on_statuses: Joi.object({
+    type: Joi.string().required(),
+    statuses: Joi.array().items(Joi.string()),
+    custom_statuses: Joi.array(),
+  }).required(),
+})
+
+type Condition = {
+  // eslint-disable-next-line camelcase
+  child_fields: ChildField[]
+}
+
+const CONDITION_SCHEMA = Joi.object({
+  child_fields: Joi.array().items(CHILD_FIELD_SCHEMA).required(),
+})
+
+type TicketFormValue = {
+  // eslint-disable-next-line camelcase
+  agent_conditions?: Condition[]
+}
+
+type TicketForm = InstanceElement & {
+  value: TicketFormValue
+}
+
+const TICKET_FORM_SCHEMA = Joi.object({
+  agent_conditions: Joi.array().items(CONDITION_SCHEMA),
+})
+
+const isTicketForm = createSchemeGuardForInstance<TicketForm>(
+  TICKET_FORM_SCHEMA,
+  'Received an invalid value for TicketForm instance'
+)
 
 /**
   * If this function fails to identify whether custom statuses are enabled
@@ -85,7 +131,7 @@ const isTicketFormWithCustomStatus = (change: Change<ChangeDataType>): boolean =
   const data = getChangeData(change)
   if (data.elemID.typeName !== TICKET_FORM_TYPE_NAME
     || !isInstanceElement(data)
-    || !isTicketFormInstance(data)) {
+    || !isTicketForm(data)) {
     return false
   }
 
--- a/packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
+++ b/packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
@@ -15,7 +15,7 @@
 */
 import { Change, ChangeDataType, ChangeError, ChangeValidator, ElemID, getChangeData, isInstanceElement, ReadOnlyElementsSource } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
-import { Condition, TicketForm } from '../filters/ticket_form'
+import { Condition, isTicketFormInstance } from '../filters/ticket_form'
 import { ACCOUNT_FEATURES_TYPE_NAME, CUSTOM_STATUS_TYPE_NAME, TICKET_FORM_TYPE_NAME, ZENDESK } from '../constants'
 
 const log = logger(module)
@@ -83,12 +83,13 @@ const hasConditionWithCustomStatuses = (conditions: Condition[]): boolean =>
 
 const isTicketFormWithCustomStatus = (change: Change<ChangeDataType>): boolean => {
   const data = getChangeData(change)
-  if (data.elemID.typeName !== TICKET_FORM_TYPE_NAME || !isInstanceElement(data)) {
+  if (data.elemID.typeName !== TICKET_FORM_TYPE_NAME
+    || !isInstanceElement(data)
+    || !isTicketFormInstance(data)) {
     return false
   }
 
-  const ticketValue = data.value as TicketForm
-  return hasConditionWithCustomStatuses(ticketValue.agent_conditions ?? [])
+  return hasConditionWithCustomStatuses(data.value.agent_conditions ?? [])
 }
 
 const createErrorsForTicketFormsWithCustomStatuses = (
--- a/packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
+++ b/packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
@@ -13,8 +13,9 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
-import { Change, ChangeDataType, ChangeError, ChangeValidator, ElemID, getChangeData, InstanceElement, isInstanceElement, ReadOnlyElementsSource, ReferenceExpression } from '@salto-io/adapter-api'
+import { Change, ChangeDataType, ChangeError, ChangeValidator, ElemID, getChangeData, InstanceElement, isInstanceElement, ReadOnlyElementsSource } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
+import { Condition, TicketForm } from '../filters/ticket_form'
 import { ACCOUNT_FEATURES_TYPE_NAME, CUSTOM_STATUS_TYPE_NAME, TICKET_FORM_TYPE_NAME, ZENDESK } from '../constants'
 
 const log = logger(module)
@@ -77,26 +78,6 @@ const createErrorsForCustomStatusTypes = (
         'Cannot deploy custom statuses when they are not enabled in the Zendesk account.',
     }))
 
-type ChildField = {
-  // eslint-disable-next-line camelcase
-  required_on_statuses: {
-    // eslint-disable-next-line camelcase
-    custom_statuses?: ReferenceExpression[]
-  }
-}
-
-type Condition = {
-  // eslint-disable-next-line camelcase
-  child_fields: ChildField[]
-}
-
-type TicketValue = {
-  // eslint-disable-next-line camelcase
-  agent_conditions?: Condition[]
-  // eslint-disable-next-line camelcase
-  end_user_conditions?: Condition[]
-}
-
 const hasConditionWithCustomStatuses = (conditions: Condition[]): boolean =>
   conditions.some((condition: Condition) =>
     condition.child_fields.some(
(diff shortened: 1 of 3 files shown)


packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
```diff
@@ -60,6 +60,75 @@ const areCustomStatusesEnabled = async (
...
+): ChangeError[] =>
+  changes
+    .filter(isTicketFormWithCustomStatus)
+    .map(getChangeData)
+    .map(instance => ({
+      elemID: instance.elemID,
+      severity: 'Warning',
+      message: 'Custom statuses are not enabled.',
+      detailedMessage:
+        'Custom statuses are not enabled which which may cause some statuses to change their context',
```
netama, line 129: I think we should be more explicit - e.g. something like
```suggestion
      message: 'Deploying ticket form with custom statuses while custom statuses are disabled',
      detailedMessage: 'It seems this ticket form originates from another account that has custom statuses enabled. Since custom statuses are disabled in the target account, this ticket form will be deployed without the custom_statuses fields',
```
@tomermevorach wdyt?

packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
```diff
@@ -60,6 +60,75 @@ const areCustomStatusesEnabled = async (
...
 }
 
+const createErrorsForCustomStatusTypes = (
+  changes: readonly Change<ChangeDataType>[]
+): ChangeError[] =>
+  changes
+    .filter(
+      change =>
+        getChangeData(change).elemID.typeName === CUSTOM_STATUS_TYPE_NAME
+    )
```
netama, line 70: 1. why not
```suggestion
    .filter(change => getChangeData(change).elemID.typeName === CUSTOM_STATUS_TYPE_NAME)
```
2. this seems like existing code, but I think we might want different handling for removals? (e.g. just silently ignore the deletions?) @edenhassid wdyt?

packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
```diff
@@ -60,6 +60,75 @@ const areCustomStatusesEnabled = async (
...
+    custom_statuses?: ReferenceExpression[]
+  }
+}
+
+type Condition = {
+  // eslint-disable-next-line camelcase
+  child_fields: ChildField[]
+}
+
+type TicketValue = {
```
netama, line 93: let's share the type definitions with the existing ones (from `ticket_form.ts`)?

packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
```diff
@@ -60,6 +60,75 @@ const areCustomStatusesEnabled = async (
...
+  // eslint-disable-next-line camelcase
+  agent_conditions?: Condition[]
+  // eslint-disable-next-line camelcase
+  end_user_conditions?: Condition[]
```
netama, line 97: @edenhassid do we need to cover these in the filter as well?

packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
```diff
@@ -60,6 +60,75 @@ const areCustomStatusesEnabled = async (
...
+
+const hasConditionWithCustomStatuses = (conditions: Condition[]): boolean =>
+  conditions.some((condition: Condition) =>
+    condition.child_fields.some(
+      field => (field.required_on_statuses.custom_statuses?.length ?? 0) > 0
+    ))
+
+const isTicketFormWithCustomStatus = (change: Change<ChangeDataType>): boolean => {
+  const data = getChangeData(change)
+  if (data.elemID.typeName !== TICKET_FORM_TYPE_NAME) return false
```
netama, line 108: please don't use shorthand in the code - 
```suggestion
  if (data.elemID.typeName !== TICKET_FORM_TYPE_NAME) {
    return false
  }
```
(and same below)

packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
```diff
@@ -60,6 +60,75 @@ const areCustomStatusesEnabled = async (
...
+  if (!isInstanceElement(data)) return false
+
+  const ticketInstance = data as InstanceElement
+  const ticketValue = ticketInstance.value as TicketValue
```
netama, line 112: please avoid casts - use typeguards instead (for the instance element it seems we just don't need it, for the second please use Joi validations like we do in other places)

packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
```diff
@@ -60,6 +60,75 @@ const areCustomStatusesEnabled = async (
...
+
+  return hasConditionWithCustomStatuses(ticketValue.agent_conditions || [])
+    || hasConditionWithCustomStatuses(ticketValue.end_user_conditions || [])
```
netama, line 115: when the value is undefined there's a preference in the code for
```suggestion
  return hasConditionWithCustomStatuses(ticketValue.agent_conditions ?? [])
    || hasConditionWithCustomStatuses(ticketValue.end_user_conditions ?? [])
```

netama (commented):
thanks for the quick fix! 🚤 
logic looks good, added some comments about code structure and reuse

packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
elentok-salto: Done 👍

packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
elentok-salto: Done 👍

packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
elentok-salto: Done 👍

packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
elentok-salto: 1. Done 👍 (I ran prettier with default config on this block which limits to 80 characters, I'll fix that in my setup)

packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
elentok-salto: Done 👍  (I'll update again if needed)

packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
elentok-salto: Cool, haven't used Joi before, do I need to create a schema for TicketValue and use its `validate()` function to check it? Does Joi also allow casting? 

Maybe I should use https://github.com/rhys-vdw/ts-auto-guard that @tomermevorach recommended? would probably be easier, I can just run it on the existing TicketValue interface, wdyt?

packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
edenhassid: I think we should, I'll check

edenhassid: @netama @elentok-salto `end_user_conditions` do not have `custom_statuses` field, they look like this:
```
end_user_conditions = [
    {
      parent_field_id = zendesk.ticket_field.instance.EdenTest3_tagger
      parent_field_type = "tagger"
      value = zendesk.ticket_field__custom_field_options.instance.EdenTest3_tagger__echeck
      child_fields = [
        {
          id = zendesk.ticket_field.instance.Customer_Priority_tagger@su
          is_required = true
        },
      ]
    },
  ]
```

packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
netama: either way - creating a typeguard manually, using Joi with `createSchemeGuardForInstance` as we do in many other validators is the most common one in zendesk (this was the "preferred" one in the past), we can also create a single typeguard with the new library (but I _think_ it might need some adjustments)

packages/zendesk-adapter/src/filters/ticket_form.ts
```diff
@@ -68,15 +72,14 @@ const invalidTicketFormChange = (change: Change<InstanceElement>): boolean =>
...
-    clonedInst.value.agent_conditions
-      .forEach(condition => condition.child_fields
-        .forEach(child => {
-          if (child.required_on_statuses.type === SOME_STATUSES
-            && (child.required_on_statuses.custom_statuses !== undefined
-              && !_.isEmpty(child.required_on_statuses.custom_statuses))) {
-            delete child.required_on_statuses.statuses
-          }
-        }))
+    clonedInst.value.agent_conditions?.forEach(condition => condition.child_fields
```
netama, line 75: I think this might conflict with @edenhassid 's work so leaving for her to review

packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
```diff
@@ -60,6 +61,55 @@ const areCustomStatusesEnabled = async (
...
+    ))
+
+const isTicketFormWithCustomStatus = (change: Change<ChangeDataType>): boolean => {
+  const data = getChangeData(change)
+  if (data.elemID.typeName !== TICKET_FORM_TYPE_NAME || !isInstanceElement(data)) {
+    return false
+  }
+
+  const ticketInstance = data as InstanceElement
+  const ticketValue = ticketInstance.value as TicketForm
```
netama, line 91: the previous comment became outdated, so just adding it again so we don't merge with the casts

packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
```diff
@@ -60,6 +61,55 @@ const areCustomStatusesEnabled = async (
...
+      severity: 'Error',
+      message: 'Custom statuses are not enabled.',
+      detailedMessage:
+        'Cannot deploy custom statuses when they are not enabled in the Zendesk account.',
+    }))
+
+const hasConditionWithCustomStatuses = (conditions: Condition[]): boolean =>
+  conditions.some((condition: Condition) =>
+    condition.child_fields.some(
+      field => (field.required_on_statuses.custom_statuses?.length ?? 0) > 0
```
netama, line 81: you can use `makeArray` from `lowerdash` -
```suggestion
      field => makeArray(field.required_on_statuses.custom_statuses).length > 0
```

netama (commented):
thanks for the fixes! added a few more, leaving this for @edenhassid to do the final review and make sure this combines well with the other fix

packages/zendesk-adapter/src/filters/ticket_form.ts
edenhassid: yes it does, I don't think you should use this file for your checks, the types guard checks something different that is relevant for this filter specifically.

packages/zendesk-adapter/src/filters/ticket_form.ts
```diff
@@ -28,8 +28,7 @@ import { TICKET_FORM_TYPE_NAME } from '../constants'
 const { awu } = collections.asynciterable
 const SOME_STATUSES = 'SOME_STATUSES'
 
-
-type Child = {
+export type Child = {
```
edenhassid, line 31: I think its wrong to use that type, this type specifies invalid forms for the filter, in the real ticket form type the `custom_statuses` can be `undefined` and the `statuses` must be

packages/zendesk-adapter/src/filters/ticket_form.ts
```diff
@@ -39,12 +38,17 @@ type Child = {
...
-  }[]
+  child_fields: Child[]
+}
+
+export type TicketForm = {
+  // eslint-disable-next-line camelcase
+  agent_conditions?: Condition[]
+
+  // eslint-disable-next-line camelcase
+  end_user_conditions?: Condition[]
```
edenhassid, line 51: I checked and we don't need this case, `end_user_conditions` can not have `custom_statuses`

packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
```diff
@@ -60,6 +61,55 @@ const areCustomStatusesEnabled = async (
...
+  const data = getChangeData(change)
+  if (data.elemID.typeName !== TICKET_FORM_TYPE_NAME || !isInstanceElement(data)) {
+    return false
+  }
+
+  const ticketInstance = data as InstanceElement
+  const ticketValue = ticketInstance.value as TicketForm
+
+  return hasConditionWithCustomStatuses(ticketValue.agent_conditions ?? [])
+    || hasConditionWithCustomStatuses(ticketValue.end_user_conditions ?? [])
```
edenhassid, line 94: don't need the `end_user_conditions`

packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
elentok-salto: Done, removed the end_user_conditions part.

packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
elentok-salto: Done.

packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
elentok-salto: Done (used `createSchemaGuardForInstance`)

packages/zendesk-adapter/src/filters/ticket_form.ts
elentok-salto: Restored ticket_form.ts to its original state and moved the types and schemas to the change validator (I would prefer to extract it to a separate file, but I don't want to delay this PR so I can do it later when we decide where to put it).

packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
elentok-salto: Done.

packages/zendesk-adapter/src/filters/ticket_form.ts
elentok-salto: Done.

packages/zendesk-adapter/src/filters/ticket_form.ts
elentok-salto: Done, restored ticket_form.ts ot its original state.

packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
elentok-salto: Done.

packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
edenhassid: I think deletion should also be included, I don't think its possible to delete a custom status if the feature is disabled

edenhassid (approved):
looks good! 💯 🔥

PR #4419, opened 2023-05-24
Support JSON files in file-cabinet

kami-delaroz:
Change Json encoding

---

None

---
_Release Notes_: 
Now we support Json files in file cabinet. fetch, deploy and preview

---
_User Notifications_: 
None


packages/netsuite-adapter/src/client/suiteapp_client/suiteapp_client.ts
```diff
@@ -61,6 +61,7 @@ const NON_BINARY_FILETYPES = new Set([
   'SMS',
   'STYLESHEET',
   'XMLDOC',
+  'JSON',
```
omrilit, line 64: should we add FSL here as well?

omrilit (approved):

PR #4421, opened 2023-05-24
SALTO-4171 - selectElementIdsByTraversal concurrent subElements calculations

adamro:
Improved performance of selectElementIdsByTraversal by using withLimitedConcurrency instead of awu.forEach

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/workspace/src/workspace/element_selector.ts
+++ b/packages/workspace/src/workspace/element_selector.ts
@@ -304,13 +304,13 @@ export const selectElementIdsByTraversal = async ({
       const currentIds = new Set(topLevelIDs.map(id => id.getFullName()))
 
       const possibleParentSelectors = subElementSelectors.map(createTopLevelSelector)
-      const possibleParentIDs = selectElementsBySelectors({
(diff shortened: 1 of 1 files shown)


packages/workspace/src/workspace/element_selector.ts
```diff
@@ -338,27 +341,30 @@ export const selectElementIdsByTraversal = async ({
...
-
-          if (isObjectType(element)) {
-            await awu(Object.values(element.fields)).forEach(async field => {
+      const addSubElementIDs = async (elemID: ElemID): Promise<void> => {
+        const element = await source.get(elemID)
+        walkOnElement({
+          element, func: selectFromSubElements,
+        })
+        if (isObjectType(element)) {
+          await withLimitedConcurrency(
```
omrilit, line 350: nit: IIUC walkOnElement already runs on all element's fields without limited concurrency, so I am not sure whether we should have limited concurrency here or just regular `map` function.

omrilit (approved):
Great catch!
had only 1 nit

ori-moisis (approved):

PR #4429, opened 2023-05-28
SALTO-4253 fix escaped unicode letters parsing

dantal4:
In case that there is an escaped Unicode letter in a string content (e.g `\u0065`) the parsing fails with `Invalid string character`.

---

---
_Release Notes_: 
Workspace:
- Fix escaped Unicode letters parsing

---
_User Notifications_: 
None


alonstern (approved):
👌

ori-moisis:
is this symmetrical with dump? if we dump a unicode char does it get escaped like this?

PR #4432, opened 2023-05-29
SALTO-3791 Support deletion in partial fetch - NetSuite side (NOT SUPPORTED YET)

kami-delaroz:
Netsuite side - support deletion in partial fetch flow

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/netsuite-adapter/src/client/sdf_client.ts
```diff
@@ -387,11 +393,15 @@ export default class SdfClient {
       [undefined, ...this.installedSuiteApps]
         .map(async suiteAppId => {
           const { executor, projectPath, authId } = await this.initProject()
+          const instancesIds = await this.listInstances(
+            executor,
+            typeNames.filter(originFetchQuery.isTypeMatch),
+            suiteAppId,
+          )
```
kami-delaroz, line 400: Pay attention I'm pulling more objects then I used to, as I ignore the info regarding changes and ask for all fetch-target objects

packages/netsuite-adapter/src/changes_detector/changes_detectors/custom_records.ts
```diff
@@ -57,5 +71,35 @@ export const getChangedCustomRecords = async (
...
+
   return changedObjects.flat()
 }
+
+export const getCustomRecordCounters = async (
+  client: NetsuiteClient,
+  { isCustomRecordTypeMatch }: Pick<NetsuiteQuery, 'isCustomRecordTypeMatch'>,
+  customRecordTypesToIgnore: Set<string>,
+): Promise<Map<string, number>> => {
+  const customRecordTypesScriptIds = await getMatchingCustomRecords(client, isCustomRecordTypeMatch)
```
kami-delaroz, line 83: we already run this query, maybe should be saved somewhere

packages/netsuite-adapter/src/changes_detector/changes_detectors/custom_records.ts
```diff
@@ -57,5 +71,35 @@ export const getChangedCustomRecords = async (
...
+      .map(async customRecordTypeScriptId => (
+      await client.runSuiteQL(`SELECT COUNT(*) as count FROM ${customRecordTypeScriptId}`)
+    )?.filter(hasCount)
+        .map(item => ({
+          typeId: customRecordTypeScriptId,
+          count: Number(item.count),
+        })))
+  )).filter(isDefined).flat()
+
+  return new Map(counters.map(counter => [counter.typeId, counter.count]))
```
kami-delaroz, line 97: maybe I could have it directly from the array manipulation above (?)

packages/netsuite-adapter/src/deletion_calculator.ts
```diff
@@ -0,0 +1,151 @@
...
+}
+
+export const getDeletedElements = async (
+  client: NetsuiteClient,
+  elementsSourceIndex: LazyElementsSourceIndexes,
+  query: NetsuiteQuery,
+  serviceInstancesIds: ObjectID[],
+  serviceTargetedCustomRecords: InstanceElement[],
+): Promise<ElemID[]> => {
+  const currentElements = await getCurrentScriptIds(elementsSourceIndex)
```
kami-delaroz, line 136: maybe I should split this list into `isTypeMatch` and `isCustomRecordTypeMatch` in one run (using groupBy?)

packages/netsuite-adapter/src/deletion_calculator.ts
```diff
@@ -0,0 +1,151 @@
...
+      const instancesScriptIds = await getCustomRecordTypeInstances(client, customRecordType.elemID.typeName)
+      return {
+        customRecordType: customRecordType.elemID.typeName,
+        instancesScriptIds: new Set(instancesScriptIds),
+      }
+    })
+  )
+  const serviceCustomRecordInstancesMap = new Map(
+    serviceCustomRecords.map(obj => [obj.customRecordType, obj.instancesScriptIds])
+  )
```
kami-delaroz, line 98: This is REALLY ugly. not sure how to handle it better though

packages/netsuite-adapter/src/deletion_calculator.ts
```diff
@@ -0,0 +1,151 @@
...
+    .map(({ elemID, serviceID }) => ({ elemID: elemID.createParentID(), serviceID }))
+}
+
+const calculateDeletedInstances = (
+  currentInstances: ElemServiceID[],
+  serviceInstancesIds: ObjectID[],
+): ElemID[] => {
+  if (currentInstances.length === 0) {
+    return []
+  }
```
kami-delaroz, line 37: You'll see a lot of such conditions, to improve performance and avoid unnecessary work

packages/netsuite-adapter/src/deletion_calculator.ts
```diff
@@ -0,0 +1,152 @@
...
+    .groupBy(objectId => objectId.type)
+    .mapValues(objectIds => objectIds.map(({ instanceId }) => instanceId))
+    .value()
+
+  return currentInstances
+    .filter(item => !serviceIdsGroupByTypes[item.elemID.typeName]?.includes(item.serviceID))
+    .map(item => item.elemID)
+}
+
+const calculateDeletedCustomRecordsFromTarget = (
```
dantal4, line 49: What "target" means? maybe there's a more common term that would make it easier to follow?

packages/netsuite-adapter/src/deletion_calculator.ts
```diff
@@ -0,0 +1,152 @@
...
+  client: NetsuiteClient,
+  currentElements: ElemServiceID[],
+  query: NetsuiteQuery,
+  serviceTargetedCustomRecords: InstanceElement[],
+): Promise<ElemID[]> => {
+  if (currentElements.length === 0) {
+    return []
+  }
+
+  const deletedCustomRecordsFromTarget = calculateDeletedCustomRecordsFromTarget(
```
dantal4, line 115: worth adding a comment that explains why do we need both `calculateDeletedCustomRecordsFromTarget` & `calculateDeletedCustomRecordsFromService`, and that there's an implicit assumption that when we fetch custom records we fetch them all (of the same custom record type)

packages/netsuite-adapter/src/deletion_calculator.ts
```diff
@@ -0,0 +1,152 @@
...
+    .map(item => item.elemID)
+}
+
+const calculateDeletedCustomRecordsFromService = async (
+  client: NetsuiteClient,
+  currentElements: ElemServiceID[],
+  query: NetsuiteQuery,
+  serviceTargetedCustomRecords: InstanceElement[],
+): Promise<ElemID[]> => {
+  const customRecordTypesToIgnore = new Set(serviceTargetedCustomRecords.map(item => item.elemID.typeName))
```
dantal4, line 72: If a custom record type didn't have any instances, we won't skip it. Maybe pass the list of custom record types too.

packages/netsuite-adapter/src/deletion_calculator.ts
dantal4: `isTypeMatch` & `isCustomRecordTypeMatch` are working on the adapter config inclusions&exclusions, without checking if the type is a custom record type or a standard type. So I think that we should first split into types/custom records types-
```suggestion
  const [currentInstances, currentCustomRecords] = _.partition(
    await getCurrentScriptIds(elementsSourceIndex),
    item => isStandardTypeName(item.elemID.typeName) || item.elemID.idType === 'type' // for custom record types
  )
```

packages/netsuite-adapter/src/deletion_calculator.ts
```diff
@@ -0,0 +1,152 @@
...
+    customRecordTypesToIgnore,
+  )
+  const currentCustomRecords = _(currentElements)
+    .filter(item => !customRecordTypesToIgnore.has(item.elemID.typeName))
+    .groupBy(item => item.elemID.typeName)
+    .value()
+
+  const customRecordTypesWithDeletions = Object.values(currentCustomRecords)
+    .filter(instances => instances.length !== serviceCustomRecordCounters.get(instances[0].serviceID))
+    .flat()
```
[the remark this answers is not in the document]
dantal4: Also - maybe we should filter only types that were queried successfully (`serviceCustomRecordCounters.has(type)`)

packages/netsuite-adapter/src/deletion_calculator.ts
```diff
@@ -0,0 +1,152 @@
...
+    return []
+  }
+
+  const serviceIdsGroupByTypes = _(serviceInstancesIds)
+    .groupBy(objectId => objectId.type)
+    .mapValues(objectIds => objectIds.map(({ instanceId }) => instanceId))
+    .value()
+
+  return currentInstances
+    .filter(item => !serviceIdsGroupByTypes[item.elemID.typeName]?.includes(item.serviceID))
```
dantal4, line 45: custom record type needs a special treat:
```suggestion
    .filter(item => !serviceIdsGroupByTypes[item.elemID.idType === 'type' ? CUSTOM_RECORD_TYPE : item.elemID.typeName]?.includes(item.serviceID))
```

packages/netsuite-adapter/src/adapter.ts
```diff
@@ -302,6 +305,16 @@ export default class NetsuiteAdapter implements AdapterOperations {
       : []
     const { elements: customRecords, largeTypesError: failedCustomRecords } = await customRecordsPromise
 
+    // we calculate deleted elements only in partial-fetch mode
+    const elementsSourceIndex = createElementsSourceIndex(this.elementsSource, isPartial)
+    const deletedElements = !isPartial ? [] : await getDeletedElements(
```
dantal4, line 310: what about data elements? do we want to support deletions of them too?

dantal4 (commented):
Great work! Very clean 👍 
Didn't go over the tests yet.

packages/netsuite-adapter/src/deletion_calculator.ts
```diff
@@ -0,0 +1,152 @@
...
+  serviceTargetedCustomRecords: InstanceElement[],
+): Promise<ElemID[]> => {
+  const currentElements = await getCurrentScriptIds(elementsSourceIndex)
+
+  const deletedInstances = calculateDeletedInstances(
+    currentElements.filter(item => query.isTypeMatch(item.elemID.typeName)),
+    serviceInstancesIds,
+  )
+
+  const deletedCustomRecords = await calculateDeletedCustomRecords(
```
dantal4, line 144: We can immediately remove all custom record instances if their custom record type was removed.

packages/netsuite-adapter/src/changes_detector/changes_detectors/custom_records.ts
```diff
@@ -57,5 +71,35 @@ export const getChangedCustomRecords = async (
...
+}
+
+export const getCustomRecordTypeInstances = async (
+  client: NetsuiteClient,
+  customRecordType: string,
+): Promise<string[]> => (
+    await client.runSuiteQL(`SELECT scriptid FROM ${customRecordType}`)
```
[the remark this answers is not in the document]
dantal4: Order by is actually needed in SuiteQL queries - see warning on `runSuiteQL` function:
```
/**
   * WARNING:
   * Due to a bug in NetSuite SuiteQL, make sure to use
   * ORDER BY <some unique identifier> ASC/DESC in your queries.
   * Otherwise, you might not get all the results.
   */
```

packages/netsuite-adapter/src/deletion_calculator.ts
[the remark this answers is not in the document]
dantal4: Deleted types can be caught separately. My concern is that we'll delete custom records that still exist. We use one API for fetching custom records and other API for querying their scriptids, and it already happened that for some reason there wasn't a SuiteQL table for a specific custom record type, although we fetched its instances successfully via SOAP.

packages/netsuite-adapter/src/deletion_calculator.ts
dantal4: `calculateDeletedCustomRecordsFromFetchResult` ?
I didn't relate it to fetch-target...

packages/netsuite-adapter/src/deletion_calculator.ts
dantal4: I think that it's [this list](https://github.com/salto-io/salto/blob/d9d20226f4c7087567fd4278673f43afe4c30e27/packages/netsuite-adapter/src/adapter.ts#LL293C24-L293C24)

packages/netsuite-adapter/src/deletion_calculator.ts
dantal4: where? not sure that I saw it...

packages/netsuite-adapter/src/adapter.ts
dantal4: Data elements - accounts, subsidiaries, customers, etc. see it [here](https://github.com/salto-io/salto/blob/d9d20226f4c7087567fd4278673f43afe4c30e27/packages/netsuite-adapter/src/adapter.ts#LL298C9-L298C9). Similar to custom records, we always fetch them all (according to the fetch target query), so by their elemIDs we can get the deleted instances.

packages/netsuite-adapter/src/adapter.ts
```diff
@@ -295,20 +301,38 @@ export default class NetsuiteAdapter implements AdapterOperations {
...
+      largeTypesError: dataTypeError,
+    } = await dataElementsPromise
     failedTypes.excludedTypes = failedTypes.excludedTypes.concat(dataTypeError)
     const suiteAppConfigElements = this.client.isSuiteAppConfigured()
       ? toConfigElements(configRecords, fetchQuery).concat(getConfigTypes())
       : []
     const { elements: customRecords, largeTypesError: failedCustomRecords } = await customRecordsPromise
 
+    // we calculate deleted elements only in partial-fetch mode
+    const deletedElements = (!isPartial || this.withPartialDeletion === false) ? [] : await getDeletedElements({
```
kami-delaroz, line 325: I explicitly check for a false value as I want to have a "default true" behavior

dantal4 (commented):
Great work 💯 
Added mostly nits.
I commented on some of the `customTypes` appearances but not all of them. Can you make sure to change them all to `customRecordTypes`? This is the term that is used all over the adapter code, and also the known netsuite term.

packages/netsuite-adapter/src/adapter.ts
```diff
@@ -295,20 +301,38 @@ export default class NetsuiteAdapter implements AdapterOperations {
...
+      client: this.client,
+      elementsSource: this.elementsSource,
+      fetchQuery,
+      serviceInstanceIds: instancesIds,
+      requestedCustomTypes: customRecordTypes,
```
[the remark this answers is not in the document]
kami-delaroz: I thought we call the instances "custom records" and the types "custom types"
am I wrong here? will have to change in a few other places

packages/netsuite-adapter/src/deletion_calculator.ts
```diff
@@ -0,0 +1,274 @@
...
+  return currentDataElements
+    .filter(elemId => !serviceDataElementsSet.has(elemId.getFullName()))
+}
+
+const calculateDeletedCustomRecordsFromFetchResult = (
+  currentCustomRecords: ElemServiceID[],
+  serviceCustomRecords: InstanceElement[],
+  requestedCustomTypes: Set<string>,
+): ElemID[] => {
+  if (currentCustomRecords.length === 0 || serviceCustomRecords.length === 0) {
```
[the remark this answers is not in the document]
kami-delaroz: right, I might have to change the logic here as now we have also the requestedCustomTypes to focus on.
will have a look

packages/netsuite-adapter/src/elements_source_index/elements_source_index.ts
```diff
@@ -123,8 +123,11 @@ const createIndexes = async (elementsSource: ReadOnlyElementsSource, isPartial:
     }
   }
 
+  const deletedElementSet = new Set(deletedElements?.map(elemId => elemId.getFullName()))
   const elements = await elementsSource.getAll()
   await awu(elements)
+    // avoid creating reference to a deleted element
+    .filter(element => !deletedElementSet.has(element.elemID.getFullName()))
```
dantal4, line 130: can you add a test to see that the deleted elements are really not processed?

packages/netsuite-adapter/src/deletion_calculator.ts
```diff
@@ -0,0 +1,274 @@
...
+    )
+
+    const deletedDataElements = calculateDeletedDataElements(
+      currentDataElements,
+      serviceDataElements,
+    )
+
+    return deletedInstances.concat(deletedDataElements).concat(await deletedCustomRecords)
+  } catch (e) {
+    log.error(`Failed calculating deleted element: ${e}, stack: ${e.stack}`)
```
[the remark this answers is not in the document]
kami-delaroz: I thought it's better to fallback to previous behavior and avoid calculating deletion instead of causing the whole fetch to fail, at least at the begging.
I'm planning to monitor this log, so this is not quite silently :)

packages/netsuite-adapter/src/deletion_calculator.ts
```diff
@@ -0,0 +1,274 @@
...
+          }
+        } else if (fetchQuery.isTypeMatch(element.elemID.typeName)) {
+          if (isStandardTypeName(element.elemID.typeName)) {
+            currentStandardInstances.push({ elemID: element.elemID, serviceID: getServiceId(element) })
+          } else if (requestedDataTypesSet.has(element.elemID.typeName)) {
+            currentDataElements.push(element.elemID)
+          }
+        }
+      } else if (isObjectType(element) && isCustomRecordType(element)) {
+        if (fetchQuery.isCustomRecordTypeMatch(element.elemID.typeName)) {
```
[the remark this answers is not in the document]
kami-delaroz: not sure I understand.
I'll get false for fetchQuery.isTypeMatch on a matching custom record type

packages/netsuite-adapter/src/elements_source_index/elements_source_index.ts
```diff
@@ -149,13 +152,14 @@ const createIndexes = async (elementsSource: ReadOnlyElementsSource, isPartial:
 
 export const createElementsSourceIndex = (
   elementsSource: ReadOnlyElementsSource,
-  isPartial: boolean
+  isPartial: boolean,
+  deletedElements?: ElemID[],
```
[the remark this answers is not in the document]
kami-delaroz: it is used by the deploy flow as well

packages/netsuite-adapter/src/changes_detector/changes_detectors/custom_records.ts
```diff
@@ -57,5 +60,31 @@ export const getChangedCustomRecords = async (
...
+  const customTypeRecords = (await Promise.all(
+    customRecordTypesScriptIds
+      .filter(customRecordTypesScriptId => !customRecordTypesToIgnore.has(customRecordTypesScriptId))
+      .map(async customRecordTypeScriptId => {
+        const scriptIds = (await client.runSuiteQL(getScriptIdsQuery({ from: customRecordTypeScriptId }))
+        )?.filter(hasScriptId).map(({ scriptid }) => (
+          scriptid.toLowerCase()
+        ))
+        return {
+          typeId: customRecordTypeScriptId,
```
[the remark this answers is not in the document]
kami-delaroz: the code is pretty difference I don't think it's a good idea

packages/netsuite-adapter/src/deletion_calculator.ts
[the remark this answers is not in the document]
omrilit: If the alternatives we have here are failing the fetch or returning as if there were no deletions, I think that the latter is preferred, but I think that it's better to have some indication in the EventLog. Is it possible to pass it to the FetchResult @kami-delaroz?

packages/netsuite-adapter/src/deletion_calculator.ts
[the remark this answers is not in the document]
dantal4: I was wrong - the query should be `fetchQuery.isObjectMatch({ type: CUSTOM_RECORD_TYPE, instanceId: element.annotations[SCRIPT_ID] })`

packages/netsuite-adapter/src/deletion_calculator.ts
```diff
@@ -0,0 +1,284 @@
...
+    )
+
+    const deletedDataElements = calculateDeletedDataElements(
+      currentDataElements,
+      serviceDataElements,
+    )
+
+    return { deletedElements: deletedInstances.concat(deletedDataElements).concat(await deletedCustomRecords) }
+  } catch (e) {
+    const errorMessage = 'Failed calculating deleted element'
```
dantal4, line 280:
```suggestion
    const errorMessage = 'Failed calculating deleted elements'
```

packages/netsuite-adapter/src/elements_source_index/elements_source_index.ts
```diff
@@ -123,8 +123,11 @@ const createIndexes = async (elementsSource: ReadOnlyElementsSource, isPartial:
     }
   }
 
+  const deletedElementSet = new Set(deletedElements?.map(elemId => elemId.getFullName()))
```
dantal4, line 126: nit
```suggestion
  const deletedElementSet = new Set(deletedElements.map(elemId => elemId.getFullName()))
```

packages/netsuite-adapter/test/deletion_calculator.test.ts
```diff
@@ -0,0 +1,225 @@
...
+      const customRecordType2 = new ObjectType({
+        elemID: new ElemID(NETSUITE, 'customrecord2'),
+        annotations: { [METADATA_TYPE]: CUSTOM_RECORD_TYPE, [SCRIPT_ID]: 'customrecord2' },
+      })
+      const customRecord = new InstanceElement(
+        'custom_record',
+        customRecordType2,
+        { [SCRIPT_ID]: 'custom_record_script_id' },
+        [],
+        { [METADATA_TYPE]: CUSTOM_RECORD_TYPE }
```
dantal4, line 86: Should be removed, as it's not reflecting the actual case and we don't want that to pass the test for some reason.

dantal4 (approved):
Great work!
Looking forward to see it takes action 👍

PR #4434, opened 2023-05-29
SAAS-7820 - Targets reference index as a TreeMap

idoser96:
Added a targets reference index that is being used as a TreeMap, and contains the references by their full path inside the element
The tree serialized and deserialized into rocksDB as key-value pairs of the full path (adapter.type.instance.instance.field....) to elemID

---

This is the OSS part of the task, the next one is to make the UI use this index and ask for the specific references by change, instead of all of the references under the element

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/workspace/test/workspace/reference_indexes.test.ts
```diff
@@ -112,13 +164,13 @@ describe('updateReferenceIndexes', () => {
       {
         key: 'test.object.field.someTemplateField',
         value: [
-          new ElemID('test', 'target2', 'type'),
```
idoser96, line 115: This does not seem to have any effect and i find it confusing

packages/workspace/test/workspace/reference_indexes.test.ts
```diff
@@ -322,15 +405,23 @@ describe('updateReferenceIndexes', () => {
 
   describe('when elements were removed', () => {
     beforeEach(async () => {
-      const changes = [toChange({ before: instance })]
```
idoser96, line 325: The test did not check if removal of an object also removes its fields

packages/workspace/src/workspace/workspace.ts
```diff
@@ -182,7 +182,8 @@ export type Workspace = {
   getSourceRanges: (elemID: ElemID) => Promise<SourceRange[]>
   getElementReferencedFiles: (id: ElemID) => Promise<string[]>
   getReferenceSourcesIndex: (envName?: string) => Promise<ReadOnlyRemoteMap<ElemID[]>>
-  getReferenceTargetsIndex: (envName?: string) => Promise<ReadOnlyRemoteMap<ElemID[]>>
+  getReferenceTargetsIndex: (envName?: string) =>
```
idoser96, line 185: This is actually not used anywhere in the code (saas included)

packages/workspace/test/workspace/reference_indexes.test.ts
```diff
@@ -322,15 +327,7 @@ describe('updateReferenceIndexes', () => {
...
-      referenceTargetsIndex.getMany.mockImplementation(async ids => ids.map(id => (
-        instance.elemID.getFullName() === id
-          ? [
-            new ElemID('test', 'target2', 'instance', 'someInstance'),
-            new ElemID('test', 'target2', 'field', 'someField', 'value'),
-          ]
-          : undefined
-      )))
+      const changes = [toChange({ before: instance }), toChange({ before: object })]
```
idoser96, line 330: The test did not check if removal of an object type also removes its fields

packages/workspace/src/workspace/reference_indexes.ts
```diff
@@ -113,30 +123,35 @@ const getReferenceTargetIndexUpdates = (
...
 ): Promise<void> => {
-  const updates = changes.flatMap(
-    change => getReferenceTargetIndexUpdates(change, changeToReferences)
-  )
+  const changesTrees = changes.flatMap(change => getReferenceTargetIndexUpdates(change, changeToReferences))
+  const [toAdd, toDelete] = _.partition(changesTrees, change => change.value.size > 0)
 
-  await updateUniqueIndex(index, updates)
+  await referenceTargetsIndex.setAll(toAdd)
+  await referenceTargetsIndex.deleteAll(toDelete.map(change => change.key))
```
ori-moisis, line 154: in `updateUniqueIndex` these were done in parallel, I assume this was to make performance better, why do we not do the same here?

packages/workspace/src/workspace/reference_indexes.ts
```diff
@@ -113,30 +123,35 @@ const getReferenceTargetIndexUpdates = (
       ...Object.values(allFields)
         .map(field => ({
           key: field.elemID.getFullName(),
-          value: baseIdToReferences[field.elemID.getFullName()] ?? [],
+          value: new collections.treeMap.TreeMap<ElemID>(
+            (baseIdToReferences[field.elemID.getFullName()] ?? []).map(ref =>
+              [ref.referenceSource.getFullName(), [ref.referenceTarget]])
+          ),
```
ori-moisis, line 129: maybe worth extracting this logic to a function just to make the code more readable.
this is borderline in terms of pure readability, but it is also duplicated so there is more motivation to extract to a simple function, I think this passes the threshold and will look better if we extract to a function.


```suggestion
          value: createReferenceTree(baseIdToReferences[field.elemID.getFullName()] ?? []),
```

packages/workspace/src/workspace/reference_indexes.ts
```diff
@@ -89,17 +101,15 @@ const updateUniqueIndex = async (
...
-    .value()
+): RemoteMapEntry<collections.treeMap.TreeMap<ElemID>>[] => {
+  const indexUpdates: RemoteMapEntry<collections.treeMap.TreeMap<ElemID>>[] = []
 
   if (isObjectTypeChange(change)) {
+    const references = changeToReferences[getChangeData(change).elemID.getFullName()]
+      .currentAndNew
+    const baseIdToReferences = _(references)
+      .groupBy(reference => reference.referenceSource.createBaseID().parent.getFullName())
+      .value()
```
ori-moisis, line 112:
```suggestion
    const baseIdToReferences = _.groupBy(
      references,
      reference => reference.referenceSource.createBaseID().parent.getFullName()
    )
```

or even
```suggestion
    const baseIdToReferences = _.groupBy(
      changeToReferences[getChangeData(change).elemID.getFullName()].currentAndNew,
      reference => reference.referenceSource.createBaseID().parent.getFullName()
    )
```

or better yet, move the `const elemId = getChangeData(change).elemID.getFullName()` line to the top and then

```suggestion
    const baseIdToReferences = _.groupBy(
      changeToReferences[elemId].currentAndNew,
      reference => reference.referenceSource.createBaseID().parent.getFullName()
    )
```

packages/lowerdash/src/collections/tree_map.ts
```diff
@@ -166,6 +166,9 @@ export class TreeMap<T> implements Map<string, T[]> {
   }
 
   valuesWithPrefix(prefix: string): IterableIterator<T[]> {
+    if (prefix === '') {
```
ori-moisis, line 169: don't we need this to be consistent with the behavior in `get` / `set` and things like that?
feels like this needs to be implemented deeper than here

ori-moisis: > get without path should give the root items and nothing else
> set without path should set the items in the root

I agree, but the current code doesn't do that.
if you set something on an empty string it will create a new child and put the values on it, not on the root.
so if we want to have consistent behavior, we need to change the code.

packages/workspace/src/workspace/reference_indexes.ts
```diff
@@ -71,37 +83,35 @@ const getReferencesFromChange = (change: Change<Element>): ChangeReferences => {
...
-  const baseIdToReferences = _(references)
-    .groupBy(reference => reference.referenceSource.createBaseID().parent.getFullName())
-    .mapValues(referencesGroup => referencesGroup.map(ref => ref.referenceTarget))
-    .value()
+): RemoteMapEntry<collections.treeMap.TreeMap<ElemID>>[] => {
+  const indexUpdates: RemoteMapEntry<collections.treeMap.TreeMap<ElemID>>[] = []
 
   if (isObjectTypeChange(change)) {
-    const type = getChangeData(change)
+    const elemId = getChangeData(change).elemID.getFullName()
```
ori-moisis, line 108: why do we have this line twice now?

packages/workspace/src/workspace/workspace.ts
```diff
@@ -1147,14 +1171,16 @@ export const loadWorkspace = async (
...
-        .referenceTargets.get(id.getFullName()) ?? []
+      const baseId = id.createBaseID().parent.getFullName()
+      const referencesTree = await (await getWorkspaceState()).states[envName].referenceTargets.get(baseId)
+
+      const idPath = id.createBaseID().path.join(ElemID.NAMESPACE_SEPARATOR)
+      return referencesTree === undefined ? []
+        : _.uniqBy(
+          Array.from(referencesTree.valuesWithPrefix(idPath)).flat(),
+          elemId => elemId.getFullName()
+        )
```
ori-moisis, line 1183: if we want to avoid the inconsistency in tree_map, we can change this to
```suggestion
      if (referencesTree === undefined) {
        return []
      }
      const idPath = id.createBaseID().path.join(ElemID.NAMESPACE_SEPARATOR)
      const references = idPath === ''
        ? referencesTree.values()
        : referencesTree.valuesWithPrefix(idPath)
      return _.uniqBy(Array.from(references).flat(), elemId => elemId.getFullName())
```

PR #4437, opened 2023-05-29
Feature/get change group ids func to adapter components

oneYAD:
Instead of using same getChangeGroupIds in Each adapter, add adapter-components the feature

---

---
_Release Notes_:
no effect

---
_User Notifications_:
no efect


netama (approved):
code looks good so pre-approving, please address the comments on the tests before merging and make sure all cases are covered.

also - please fix the PR description before merging, it does not follow the convention

PR #4438, opened 2023-05-29
SALTO-4038 Added missing references handler in Jira

idoamit8:
1. In this PR I added missing References resolver. 
2. I added enable missing references flag to the config.
3. I added Tests to test the missing references resolver. 

---

### To test the code manually
1. in the `jira.nacl` file wrote
``` 
    exclude = [
      {
      type = "Screen"
      }
    ]
``` 
2. fetch your jira env
3. Go to an instance of 'screen scheme' 
4. check that the screen reference is of missing reference type.  

### key notes
There is a flag called `enableMissingReferences`. When this flag set to false we do not convert missing references to references.

---
_Release Notes_: 
_Jira Adapter_:
When there is one object that points to another object and it is not found, instead of a reference that contains an internal id, there will be a missing reference which will emphasize that there is a missing reference problem. Each fetch will be accompanied by a warning and it will not be possible to deploy.

---
_User Notifications_: 
_Jira Adapter_:
1. When there are missing references, instead of meaningless numbers/names, a missing reference will appear whose purpose is to highlight the fact that there is a problem


packages/jira-adapter/src/adapter_creator.ts
```diff
@@ -62,6 +62,7 @@ const adapterConfigFromConfig = (
   )
   const fetch = _.defaults(
     {}, config?.value.fetch, _.pick(defaultConfig.fetch, 'hideTypes'),
+    _.pick(defaultConfig.fetch, 'enableMissingReferences'),
```
netama, line 65: why not
```suggestion
    {}, config?.value.fetch, _.pick(defaultConfig.fetch, ['hideTypes', 'enableMissingReferences']),
```

packages/jira-adapter/src/reference_mapping.ts
```diff
@@ -117,6 +131,8 @@ ReferenceContextStrategyName
     this.target = def.target
       ? { ...def.target, lookup: this.serializationStrategy.lookup }
       : undefined
+    this.missingRefStrategy = def.JiraMissingRefStrategy
+      ? JiraMissingReferenceStrategyLookup[def.JiraMissingRefStrategy] : undefined
```
netama, line 135: readability nit:
```suggestion
    this.missingRefStrategy = def.JiraMissingRefStrategy
      ? JiraMissingReferenceStrategyLookup[def.JiraMissingRefStrategy]
      : undefined
```

netama (commented):
looks great 🎈 added some nits, going to leave the actual rule review to @IdoZakk and @alonstern (I'm assuming this is covering the Screen missing references?)

for the release notes and user notifications - please add a `_Jira Adapter_:` title, so that we will know which adapter this is impacting when creating the combined notes

packages/jira-adapter/test/filters/fields/field_references.test.ts
```diff
@@ -0,0 +1,81 @@
...
+      } })
+
+    const generateElements = (): Element[] => ([
+      screenSchemeType,
+      screenType,
+      screenTypesType,
+      new InstanceElement('screen1', screenType, { id: 111 }),
+      new InstanceElement('screenScheme1', screenSchemeType, { id: 222, screens: { default: 111 } }),
+      new InstanceElement('screenScheme2', screenSchemeType, { id: 333, screens: { default: 444 } }),
+    ])
```
IdoZakk, line 50: You can use beforeEach, it is ran before every test so no need to clone or to call anything. You can see an example in most jira filters.

IdoZakk (commented):
LGTM 👍 
The User Notifications should only include the first item.
Add info about the flag (including the name) in the context for reviewer part

packages/jira-adapter/test/filters/fields/field_references.test.ts
idoamit8: Great, done!

packages/jira-adapter/test/filters/fields/field_references.test.ts
```diff
@@ -48,18 +49,19 @@ describe('fieldReferencesFilter', () => {
       new InstanceElement('screenScheme1', screenSchemeType, { id: 222, screens: { default: 111 } }),
       new InstanceElement('screenScheme2', screenSchemeType, { id: 333, screens: { default: 444 } }),
     ])
+    beforeEach(() => {
+      elements = generateElements().map(e => e.clone())
+    })
```
IdoZakk, line 54: No need to clone them :-)
If you create them inside the beforeEach they are created before each test. The purpose of the clone is not to change the original elements, but here you do not care.
The objects that do carry over are the types here. As you do not change them it is OK (and also you can create them inside the beforeEach.
Another tip: usually you can create the types without the fields unless you are doing changes to the type or using it inside the filter. More than that, in this filter (and also in most of them) you do not need the types to be part of elements, you can use just the instance elements

IdoZakk (commented):
👍 
We still need a decision about the scope and noise reduction and we can continue.

packages/jira-adapter/test/filters/fields/field_references.test.ts
shir-reifenberg: @IdoZakk  I think types are necessary here cause the reference rules are based on the nested types

packages/jira-adapter/src/reference_mapping.ts
```diff
@@ -105,6 +118,7 @@ type JiraFieldReferenceDefinition = referenceUtils.FieldReferenceDefinition<
 ReferenceContextStrategyName
 > & {
   JiraSerializationStrategy?: JiraReferenceSerializationStrategyName
+  JiraMissingRefStrategy?: referenceUtils.MissingReferenceStrategyName
```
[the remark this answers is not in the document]
netama: yes ^

packages/jira-adapter/src/adapter_creator.ts
```diff
@@ -61,7 +61,7 @@ const adapterConfigFromConfig = (
     _.omit(config?.value ?? {}, 'fetch'),
   )
   const fetch = _.defaults(
-    {}, config?.value.fetch, _.pick(defaultConfig.fetch, 'hideTypes'),
+    {}, config?.value.fetch, _.pick(defaultConfig.fetch, ['hideTypes', 'enableMissingReferences']),
```
IdoZakk, line 64: @alonstern not directly related to this, but maybe better to omit here instead of picking?

alonstern: Yeah, probably

packages/jira-adapter/src/reference_mapping.ts
```diff
@@ -110,46 +123,55 @@ const JiraReferenceSerializationStrategyLookup: Record<
...
+      ? JiraMissingReferenceStrategyLookup[def.jiraMissingRefStrategy]
+      : undefined
   }
 }
 
 export const referencesRules: JiraFieldReferenceDefinition[] = [
   {
     src: { field: 'issueTypeId', parentTypes: ['IssueTypeScreenSchemeItem', 'FieldConfigurationIssueTypeItem', SCRIPT_RUNNER_TYPE] },
     serializationStrategy: 'id',
+    // jiraMissingRefStrategy -  Field can be string
```
IdoZakk, line 150: Is this line intentional?

packages/jira-adapter/test/filters/fields/field_references.test.ts
```diff
@@ -0,0 +1,82 @@
...
+        const filter = fieldReferencesFilter(getFilterParams({})) as filterType
+        await filter.onFetch(elements)
+        const screenScheme1 = elements.find(e => e.elemID.name === 'screenScheme1') as InstanceElement
+        expect(screenScheme1.value.screens.default).toBeInstanceOf(ReferenceExpression)
+        expect(screenScheme1.value.screens.default.elemID.name).toEqual('screen1')
+      })
+      it('should create missing references if enableMissingReferences flag is enabled', async () => {
+        const configWithMissingRefs = _.cloneDeep(getDefaultConfig({ isDataCenter: false }))
+        configWithMissingRefs.fetch.enableMissingReferences = true
+        const filter = fieldReferencesFilter(getFilterParams({ config: configWithMissingRefs })) as filterType
```
IdoZakk, line 67: `enableMissingReference` is true by default, so need for it here (and you verify the default on the way)

IdoZakk (commented):
Great! 💣 
Some nits, and look at the answer of Alon, change the pick to omit

packages/jira-adapter/src/reference_mapping.ts
idoamit8: Yes, for us to know why I didn't write `jiraMissingRefStrategy` for this rules

IdoZakk: So a more descriptive phrasing will make it easier to understand, maybe `No missing references - field can be a string`?

packages/jira-adapter/src/adapter_creator.ts
idoamit8: Talked F2F with Alon and removed the `_,pick`

IdoZakk (approved):
💯

PR #4443, opened 2023-05-30
SALTO-4105 - Salesforce: Discard multienv-unfriendly fields of OrganizationSettings

tomsellek:
We only really care about account sharing settings, and some fields of OrganizationSettings are noisy on fetches (change on every fetch) or across envs.

 - [x] [Noise Suppression](https://github.com/salto-io/salto_private/pull/5796)

---

N/A

---
_Release Notes_: 
Salesforce: remove some fields of OrganizationSettings

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/adapter.ts
+++ b/packages/salesforce-adapter/src/adapter.ts
@@ -271,7 +271,6 @@ export const SYSTEM_FIELDS = [
   'LastViewedDate',
   'Name',
   'RecordTypeId',
-  'SystemModstamp',
   'OwnerId',
   'SetupOwnerId',
 ]
--- a/packages/salesforce-adapter/src/adapter.ts
+++ b/packages/salesforce-adapter/src/adapter.ts
@@ -271,6 +271,7 @@ export const SYSTEM_FIELDS = [
   'LastViewedDate',
   'Name',
   'RecordTypeId',
+  'SystemModstamp',
   'OwnerId',
   'SetupOwnerId',
 ]
--- a/packages/salesforce-adapter/src/filters/organization_wide_sharing_defaults.ts
+++ b/packages/salesforce-adapter/src/filters/organization_wide_sharing_defaults.ts
@@ -32,7 +32,36 @@ const log = logger(module)
 
 const ORGANIZATION_SETTINGS_INSTANCE_NAME = 'OrganizationSettings'
 
-const enrichTypeWithFields = async (client: SalesforceClient, type: ObjectType): Promise<void> => {
+/*
+* These fields are not multienv friendly
+* */
+const FIELDS_TO_IGNORE = [
+  'DailyWebToCaseCount',
+  'DailyWebToLeadCount',
+  'InstanceName',
+  'IsSandbox',
+  'LastWebToCaseDate',
+  'LastWebToLeadDate',
+  'MonthlyPageViewsEntitlement',
+  'MonthlyPageViewsUsed',
+  'OrganizationType',
+  'SelfServiceCaseSubmitRecordTypeId',
+  'SelfServiceEmailUserOnCaseCreation‍TemplateId',
+  'SelfServiceNewCommentTemplateId',
+  'SelfServiceNewPassTemplateId',
+  'SelfServiceNewUserTemplateId',
+  'SelfServiceSolutionCategoryStartNodeId',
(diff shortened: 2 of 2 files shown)


packages/salesforce-adapter/src/adapter.ts
```diff
@@ -266,7 +266,6 @@ export const SYSTEM_FIELDS = [
   'LastViewedDate',
   'Name',
   'RecordTypeId',
-  'SystemModstamp',
```
amitgoldarb, line 269: Why did we remove this? This makes all values on all other types non-hidden.

packages/salesforce-adapter/src/filters/organization_wide_sharing_defaults.ts
```diff
@@ -32,7 +32,36 @@ const log = logger(module)
 
 const ORGANIZATION_SETTINGS_INSTANCE_NAME = 'OrganizationSettings'
 
-const enrichTypeWithFields = async (client: SalesforceClient, type: ObjectType): Promise<void> => {
+/*
+* These fields are not multienv friendly
+* */
+const FIELDS_TO_IGNORE = [
```
amitgoldarb, line 38: This looks like quite a bit. Wondering if it's easier to do an include list? Not sure.

amitgoldarb (commented):
Looks good, added a few comments.
Should we add some UT to see if the data was omitted correctly?

packages/salesforce-adapter/src/filters/organization_wide_sharing_defaults.ts
```diff
@@ -87,12 +119,13 @@ const createOrganizationInstance = (objectType: ObjectType, fieldValues: Values)
...
+const filterCreator: RemoteFilterCreator = ({ client, config }) => ({
   name: 'organizationWideSharingDefaultsFilter',
   remote: true,
   onFetch: async elements => {
     const objectType = createOrganizationType()
-    await enrichTypeWithFields(client, objectType)
+    const fieldsToIgnore = new Set(FIELDS_TO_IGNORE.concat(config.systemFields ?? []))
+    await enrichTypeWithFields(client, objectType, fieldsToIgnore)
 
     const queryResult = await queryClient(client, ['SELECT FIELDS(ALL) FROM Organization LIMIT 200'])
```
amitgoldarb, line 130: Another thought, maybe we should just pick some specific fields here? WDYT?

packages/salesforce-adapter/src/adapter.ts
tomsellek: This is what we in the business call "an oopsie". Will fix.

packages/salesforce-adapter/src/filters/organization_wide_sharing_defaults.ts
tomsellek: Looking at the type's definition, I think an include list will end up being even longer

packages/salesforce-adapter/src/filters/organization_wide_sharing_defaults.ts
```diff
@@ -32,7 +32,35 @@ const log = logger(module)
...
+  'TrialExpirationDate',
+  'WebToCaseAssignedEmailTemplateId',
+  'WebToCaseCreatedEmailTemplateId',
+  'WebToCaseDefaultCreatorId',
+]
+
+const enrichTypeWithFields = async (
+  client: SalesforceClient,
+  type: ObjectType,
+  fieldsToIgnore: Set<string>): Promise<void> => {
```
amitgoldarb, line 63: Super Nit (style): This will create a cleaner commit next time someone adds a new parameter. One line instead of 2.
```suggestion
  fieldsToIgnore: Set<string>,
  ): Promise<void> => {
```

amitgoldarb (approved):
Looks good!
Approving.
In favor of our new SF process which requires 2 reviewers per PR. @benisraeladi / @tamtamirr can one of you possibly look here?
I think 2 should be enough.

benisraeladi (approved):

packages/salesforce-adapter/src/filters/organization_wide_sharing_defaults.ts
tomsellek: Ack

PR #4444, opened 2023-05-30
SAAS-6976 fix dummy adapter deploy due to code change in deployOrValidate

benisraeladi:
---

Because of code beautification change in `deployOrValidate`, when calling dummy adapter `this` might be undefined.

---

_Release Notes_: _None_

---

_User Notifications_: _None_

### DIFF
--- a/packages/core/src/core/deploy/deploy_actions.ts
+++ b/packages/core/src/core/deploy/deploy_actions.ts
@@ -58,22 +58,25 @@ const extractErrors = (
 const deployOrValidate = async (
   { adapter, adapterName, opts, checkOnly }: DeployOrValidateParams
 ): Promise<AdapterDeployResult> => {
-  try {
-    const deployOrValidateFn = checkOnly ? adapter.validate : adapter.deploy
-    if (deployOrValidateFn === undefined) {
-      throw new Error(`${checkOnly ? 'Check-Only deployment' : 'Deployment'} is not supported in adapter ${adapterName}`)
-    }
-    const result = await deployOrValidateFn(opts)
+  // NOTE: DRY-ing this code into something like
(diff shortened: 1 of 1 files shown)


amitgoldarb (approved):
Just a thought not blocking.
Should we add UT for this thrown error with the expected behaviour? It seems like we don't have some test that should have caught this scenario.
Approving to save cycles

packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -58,23 +58,26 @@ const extractErrors = (
...
 ): Promise<AdapterDeployResult> => {
-  try {
-    const deployOrValidateFn = checkOnly ? adapter.validate : adapter.deploy
-    if (deployOrValidateFn === undefined) {
-      throw new Error(`${checkOnly ? 'Check-Only deployment' : 'Deployment'} is not supported in adapter ${adapterName}`)
-    }
-    const result = await deployOrValidateFn(opts)
+  // NOTE: DRY-ing this code into something like
+  // const deployOrValidateFn = checkOnly ? adapter.validate : adapter.deploy
+  // can cause `this` to be undefined, causing problems within the adapters.
```
ori-moisis, line 63: when passing "instance methods" in javascript, you would need to explicitly bind the instance in order to preserve the behavior.
so, in javascript, this would be
```typescript
const deployOrValidateFn = checkOnly ? adapter.validate.bind(adapter) : adapter.deploy.bind(adapter)
```

though IMO this is not the most readable way to DRY this code anyway.
I think the cleanest option would be to move the combined logic one level above `deployOrValidate`, or use an internal function.
in this case, since `deployOrValidate` is only called in one place, I'd just add the logic to the place that calls it...

PR #4448, opened 2023-06-01
SALTO-4259: fix null values in gadget properties

alonstern:
Fixed null values in gadget properties

---
_Release Notes_: 
None

---
_User Notifications_: 
None


shayc331 (approved):
💪

PR #4449, opened 2023-06-01
SALTO-4254 Return a lexer error token and add a better catch

adfineberg:
_Currently when the parser crashes with an unexpected error, that error is logged, but the parse result doesn’t contain any error so the rest of the code doesn’t know that something failed._

---

---
_Release Notes_: 
_Workspace_
* Update the parse result with caught errors

---
_User Notifications_:

### DIFF
--- a/packages/workspace/src/parser/internal/native/errors.ts
+++ b/packages/workspace/src/parser/internal/native/errors.ts
@@ -18,7 +18,7 @@ import { ParseError } from '../../types'
 import { SourceRange } from '../types'
 
 
-export const createError = (range: SourceRange, summary: string, message?: string): ParseError => ({
+const createError = (range: SourceRange, summary: string, message?: string): ParseError => ({
   summary,
   subject: range,
   context: range,
@@ -176,6 +176,12 @@ export const contentMergeConflict = (range: SourceRange): ParseError => createEr
   range,
   'Unresolved merge conflict',
 )
+
+export const invalidSyntax = (range: SourceRange): ParseError =>
+  createError(range, 'Invalid syntax',)
+
+export const unknownParsingError = (range: SourceRange, message: string): ParseError =>
+  createError(range, message)
 
 export const invalidStringChar = (stringRange: SourceRange, errMsg: string): ParseError => {
   const errMsgPosition = Number.parseInt(_.last(errMsg.split(' ')) || '', 10)
--- a/packages/workspace/src/parser/internal/native/errors.ts
+++ b/packages/workspace/src/parser/internal/native/errors.ts
@@ -18,7 +18,7 @@ import { ParseError } from '../../types'
 import { SourceRange } from '../types'
 
 
-const createError = (range: SourceRange, summary: string, message?: string): ParseError => ({
+export const createError = (range: SourceRange, summary: string, message?: string): ParseError => ({
   summary,
   subject: range,
   context: range,
--- a/packages/workspace/src/parser/internal/native/lexer.ts
+++ b/packages/workspace/src/parser/internal/native/lexer.ts
@@ -52,8 +52,9 @@ const WORD_PART = '[a-zA-Z_][\\w.@]*'
 
 export const rules: Record<string, moo.Rules> = {
   // Regarding ERROR tokens: Each section in the state must have an error token.
-  // When there is no lexer match, the error token is returned with the rest of the buffer.
-  // If there is no error token in a section, an Error is thrown from the lexer.
+  // If there is no error token in a section, an Error is thrown from the lexer, with missing information.
+  // With an error token - when there is no lexer match, the error token is returned with the rest of the buffer.
+  // We throw our own error with the token and reflect this to the user.
   main: {
     [TOKEN_TYPES.MERGE_CONFLICT]: { match: '<<<<<<<', push: 'mergeConflict' },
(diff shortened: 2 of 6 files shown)


packages/workspace/src/parser/internal/native/lexer.ts
```diff
@@ -77,24 +79,27 @@ export const rules: Record<string, moo.Rules> = {
...
+    [TOKEN_TYPES.ERROR]: moo.error,
   },
 }
 
 export type LexerToken = Required<moo.Token>
 
 export class NoSuchElementError extends Error {
   constructor(public lastValidToken?: LexerToken) {
-    super("All of the lexer's token has already been consumed.")
+    super('All lexerֿֿֿ tokens have already been consumed.')
```
ori-moisis, line 102:
```suggestion
    super('All lexer tokens have already been consumed.')
```

packages/workspace/src/parser/internal/native/lexer.ts
```diff
@@ -67,7 +68,8 @@ export const rules: Record<string, moo.Rules> = {
     [TOKEN_TYPES.COMMENT]: /\/\//,
     [TOKEN_TYPES.WHITESPACE]: { match: /[ \t]+/ },
     [TOKEN_TYPES.NEWLINE]: { match: /[\r\n]+/, lineBreaks: true },
-    [TOKEN_TYPES.INVALID]: { match: /[^ \n]+/, error: true },
+    [TOKEN_TYPES.INVALID]: { match: /[^ \n]+/ },
```
ori-moisis, line 71: could you explain why this is no longer marked as an error?
not sure I completely understand what the purpose of this was before, but I feel like keeping the same match but saying it is not an error anymore would change the behavior in a way that I don't fully understand...

packages/workspace/src/parser/internal/native/errors.ts
```diff
@@ -18,7 +18,7 @@ import { ParseError } from '../../types'
 import { SourceRange } from '../types'
 
 
-const createError = (range: SourceRange, summary: string, message?: string): ParseError => ({
+export const createError = (range: SourceRange, summary: string, message?: string): ParseError => ({
```
ori-moisis, line 21: might be best to follow the existing pattern and export this as a more specific function?

packages/workspace/test/parser/errors.test.ts
```diff
@@ -85,8 +90,17 @@ describe('parsing errors', () => {
         expect(res.errors[3].summary)
           .toBe('Invalid block item')
       })
+
+      it('should log an error not added to the errors', () => {
+        expect(log).toHaveBeenCalledWith(
```
ori-moisis, line 95: we usually don't test logging as it is not really a part of the a function's "contract"

packages/workspace/test/parser/parse.test.ts
```diff
@@ -910,8 +910,21 @@ value
     })
   })
 
+  it('should fail gracefully with unicode line separators in multiline strings', async () => {
```
ori-moisis, line 913: it shouldn't really fail (I assume you had to change this in your next PR)

I think there is no way to test the real "unexpected error" other than mocking some internal function and making it throw an error


also, from the code, I'd expect there to be two tests since we added two different `catch` blocks.
is one of them untestable for some reason?

packages/workspace/src/parser/internal/native/lexer.ts
adfineberg: I looked a lot and couldn't find any usage of this token, which IIUC means that we never see this error.
In any case, I think that my catchall is more comprehensive - and moo doesn't allow more than 1 error token in a state.

packages/workspace/test/parser/errors.test.ts
adfineberg: I agree, but I couldn't find a "real" way to reach this code block. This is the only case I found that sends out this kind of error that isn't caught elsewhere. 
As you suggested in the next comment, I'll simply mock a throw from an inner function and remove this test.

packages/workspace/test/parser/parse.test.ts
adfineberg: Yeah, I started by writing it TDD style for the next PR and decided to leave this test in the meantime.
I think that you are right about the thrown error. That will make it possible to test both `catch` blocks

packages/workspace/src/parser/internal/native/lexer.ts
ori-moisis: in that case, it would make more sense to me if we removed that token rather than changing it so it isn't an error anymore, no?

adfineberg: Yeah, I guess so. I might be missing something, because I don't understand why this option exists (to have both a match and an error)... This is the test for it: https://github.com/no-context/moo/blob/main/test/test.js#L1008
Maybe for this "invalid" token we actually do want to leave it? If it is caught, we continue parsing until the next match. WDYT?

ori-moisis: if I understand correctly [the last time this was changed](https://salto-io.atlassian.net/browse/SALTO-1138), we need to match here anything that isn't a newline.
if that is the case, I am not sure how it would ever be possible to match on the error token (because we'd have a match on newlines followed by a match on anything that isn't a newline...)

adfineberg: You are probably right, but I prefer to be defensive here, and have a catchall for errors. If we missed some other unicode token or other Javascript quirk...

ori-moisis (approved):

PR #4454, opened 2023-06-03
SALTO-4259: Convert properties in any depths from map to list in gadgets

alonstern:
Convert properties in any depths from map to list in gadgets

---
_Release Notes_: 
_Jira adapter_:
- Properties in any depths in gadgets will be converted from map to list

---
_User Notifications_: 
_Jira adapter_:
- Properties in any depths in gadgets will be converted from map to list


IdoZakk (approved):
Much more elegant :-)
Please add a test

PR #4455, opened 2023-06-03
SALTO-4265 Limit length of value text in validation errors

kami-delaroz:
Limit length of value text in validation error messages to 20 chars
Replace the value field in the error object with this truncated value

---

None

---
_Release Notes_: 
Avoid long error messages of validation errors.
Change the message to include only the first 20 chars of the value

---
_User Notifications_: 
None

### DIFF
--- a/packages/workspace/src/validator.ts
+++ b/packages/workspace/src/validator.ts
@@ -29,13 +29,15 @@ import {
   isTemplateExpression,
   UnresolvedReference,
 } from '@salto-io/adapter-api'
-import { toObjectType } from '@salto-io/adapter-utils'
+import { safeJsonStringify, toObjectType } from '@salto-io/adapter-utils'
 import { InvalidStaticFile } from './workspace/static_files/common'
 import { CircularReference, resolve } from './expressions'
 import { IllegalReference } from './parser/parse'
 
 const log = logger(module)
 const { makeArray } = collections.array
+
+const MAX_VALUE_LENGTH = 25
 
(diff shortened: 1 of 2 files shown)


omrilit (approved):
I am approving since it looks like it works and I wouldn't like to block you, yet I am not sure that I agree with something there, please see my comment.

packages/workspace/src/validator.ts
```diff
@@ -86,8 +88,19 @@ const validateAnnotations = (
...
 
+const lengthLimiterStringify = (value: Value, length = MAX_VALUE_LENGTH): string => {
+  if (typeof value === 'string' && value.length <= length) {
+    return value
+  }
+  const safeValue = typeof value === 'string' ? value : safeJsonStringify(value)
+  if (safeValue.length > length) {
+    return `${safeValue.slice(0, length - 3)}...`
+  }
+  return `${safeValue}`
```
ori-moisis, line 99: can just `return safeValue`, there is no templating needed here

packages/workspace/src/validator.ts
```diff
@@ -86,8 +88,19 @@ const validateAnnotations = (
   return []
 }
 
+const lengthLimiterStringify = (value: Value, length = MAX_VALUE_LENGTH): string => {
+  if (typeof value === 'string' && value.length <= length) {
```
ori-moisis, line 92: this condition isn't actually doing anything...

PR #4457, opened 2023-06-04
Lowered Jira coverage threshold

alonstern:
Lowered Jira coverage threshold so the build will pass

---
_Release Notes_: 
None

---
_User Notifications_: 
None


IdoZakk (approved):
↙️

PR #4463, opened 2023-06-04
SALTO-4285: Open Aliases to Salesforce Data Records

tamtamirr:
Aliases to Data Records in Salesforce are available by default.

---

_None_

---
_Release Notes_: 
_None_

---
_User Notifications_: 
Salesforce:
- Existing Data Records `nacls` will have a new `_alias` annotation.


amitgoldarb (approved):
LGTM!
Thanks for this

PR #4464, opened 2023-06-05
SALTO-3887: Saved search objects query doesn't match date format

Shulik95:
_Saved search query matches date format, system_notes and saved_searches use date from configRecords_

---

None

---
_Release Notes_: 
_Netsuite Adapter_:
* Saved search query matches date format to user preferences
* system_notes and saved_searches use date from configRecords

---
_User Notifications_: 
None


dantal4 (approved):
Great work!

PR #4465, opened 2023-06-05
SALTO-4288 do not crash on inability to get full name for element ID and log all failures

benisraeladi:
Only return `SaltoError`s on metadata deploy component failures

---

---
_Release Notes_: _None_

---
_User Notifications_: _None_


tamtamirr (approved):
✅ Can you please update the PR description?

PR #4468, opened 2023-06-06
SALTO-4038 Added checks for undefined values in refs in Jira

idoamit8:
In this PR, I added filters and extended if statements in all the places in Jira's filters that I thought are suitable.  In order to check if reference expressions have undefined values. 

---
##### What I did to check that I didn't broke anything:
1. I ran all Jira's test and everything worked great. 
2. I fetched and depllyed Ori-testing env and everything worked great.
3. I rebased `SALTO-4038-EnableMissingReferencesInJira` over this branch and fetched everything and now it works great. 

##### followup:
After confirming the places where I wrote the checks, I will also add logs.

---
_Release Notes_: 
1. This changes won't affect Salto users, but they will be good infrastructure to add the missing reference filter. 

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-utils/test/utils.test.ts
+++ b/packages/adapter-utils/test/utils.test.ts
@@ -38,6 +38,7 @@ import {
   getPath,
   getSubtypes,
   formatConfigSuggestionsReasons,
+  isResolvedReferenceExpression,
 } from '../src/utils'
 import { buildElementsSourceFromElements } from '../src/element_source'
 
@@ -2540,6 +2541,15 @@ describe('Test utils.ts', () => {
     it('should return a formatted reasons list', () => {
       const message = formatConfigSuggestionsReasons(['reason A', 'reason B'])
       expect(message).toEqual('    * reason A\n    * reason B')
+    })
+  })
+  describe('isResolvedReferenceExpression', () => {
+    const inst = new InstanceElement('inst', new ObjectType({ elemID: new ElemID('test', 'type') }))
+    it('should return false for a reference expression with undefind value', () => {
+      expect(isResolvedReferenceExpression(new ReferenceExpression(inst.elemID))).toBeFalsy()
+    })
+    it('should return true for a resolved reference expression', () => {
+      expect(isResolvedReferenceExpression(new ReferenceExpression(inst.elemID, inst))).toBeTruthy()
     })
   })
 })
--- a/packages/adapter-utils/test/utils.test.ts
+++ b/packages/adapter-utils/test/utils.test.ts
@@ -20,7 +20,8 @@ import {
   isListType, ListType, BuiltinTypes, StaticFile, isPrimitiveType,
   Element, isReferenceExpression, isPrimitiveValue, CORE_ANNOTATIONS, FieldMap, AdditionChange,
   RemovalChange, ModificationChange, isInstanceElement, isObjectType, MapType, isMapType,
-  ContainerType, TypeReference, createRefToElmWithValue, VariableExpression, getChangeData, PlaceholderObjectType,
+  ContainerType, TypeReference, createRefToElmWithValue, VariableExpression, getChangeData,
+  PlaceholderObjectType, UnresolvedReference,
 } from '@salto-io/adapter-api'
 import { collections } from '@salto-io/lowerdash'
 import { mockFunction } from '@salto-io/test-utils'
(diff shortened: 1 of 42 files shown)


packages/jira-adapter/src/filters/field_configuration/field_configuration_dependencies.ts
```diff
@@ -23,12 +23,14 @@ import { FIELD_CONFIGURATION_ITEM_TYPE_NAME, PROJECT_TYPE } from '../../constant
 const log = logger(module)
 
 const getProjectUsedFields = (instance: InstanceElement): InstanceElement[] => {
-  if (!isReferenceExpression(instance.value.issueTypeScreenScheme)) {
+  if (!isReferenceExpression(instance.value.issueTypeScreenScheme)
```
alonstern, line 26: Maybe we can add `isResolvedReferenceExpression` in adapter-api that will check that `values` is neither `undefined` or `UnresolvedReference` so we can use it in change validators, deploy, and fetch?

packages/jira-adapter/src/filters/field_configuration/field_configuration_irrelevant_fields.ts
```diff
@@ -36,8 +36,8 @@ const filter: FilterCreator = ({ fetchQuery }) => ({
       .forEach(instance => {
         const [fields, trashedFields] = _.partition(
           instance.value.fields,
-          field => isReferenceExpression(field.id)
-            && !field.id.value.value.isLocked
+          field => isResolvedReferenceExpression(field.id)
```
alonstern, line 39: I think we need to update some of the UTs since now they are not checking the right scenario. E.g., I think we need to replace the string [here](https://github.com/salto-io/salto/blob/main/packages/jira-adapter/test/filters/field_configuration/field_configuration_irrelevant_fields.test.ts#L56) with a reference without value, since that is what the filter now gets

alonstern (commented):
I see there are many more usages of `isReferenceExpression` in the adapter, don't we need to change them too?

packages/jira-adapter/src/filters/field_configuration/field_configuration_irrelevant_fields.ts
idoamit8: I don't know if we want to change every place where it says `isReferenceExpression` to `isResolvedReferenceExpression` because there are places where we access only the ElemID of the reference and not its value. But maybe I should add the change in other places like Change validators. @netama What do you think?
@alonstern which tests do you think I should modify? Currently all the tests are passing.

alonstern: I think places that access only the ElemID will get now an invalid ID which is also something we don't want. I think we need to do this at least for change validators, and deploy.
(e.g. another place we need to do this /Users/alon/salto/packages/jira-adapter/src/filters/automation/automation_project_broken_reference.ts)

For the tests, we need to change in the tests that check that we ignore unresolved values from strings to references without value. I am not sure the exact tests, we need to go over it in the places we changed, but for example this [one](https://github.com/salto-io/salto/blob/main/packages/jira-adapter/test/filters/field_configuration/field_configuration_irrelevant_fields.test.ts#L56)

packages/adapter-utils/src/utils.ts
```diff
@@ -1163,3 +1163,8 @@ export const formatConfigSuggestionsReasons = (reasons: string[]): string => {
 
   return [...reasons.map(formatReason)].join(os.EOL)
 }
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const isResolvedReferenceExpression = (value: any): value is ReferenceExpression => (
```
alonstern, line 1168: Please add doc about this covering both fetch and deploy unresolved references

packages/jira-adapter/src/change_validators/automation_unresolved_references.ts
```diff
@@ -14,16 +14,17 @@
...
+import { isResolvedReferenceExpression } from '@salto-io/adapter-utils'
 import { AUTOMATION_TYPE } from '../constants'
 
 
 export type ProjectType = { projectId: ReferenceExpression }
 
 export const isProjectType = (element: unknown): element is ProjectType => {
   const projectId = _.get(element, 'projectId')
-  return projectId !== undefined && isReferenceExpression(projectId)
+  return projectId !== undefined && isResolvedReferenceExpression(projectId)
```
alonstern, line 27: I think this actually needs to be `isReferenceExpression` because in the validator we try to find the unresolved references, but you can replace line 31 to use `isResolvedReferenceExpression` instead of checking `instanceof UnresolvedReference`

packages/adapter-utils/src/utils.ts
alonstern, line 1168: Could you add a small test for this?

packages/adapter-utils/test/utils.test.ts
```diff
@@ -2542,4 +2543,13 @@ describe('Test utils.ts', () => {
       expect(message).toEqual('    * reason A\n    * reason B')
     })
   })
+  describe('isResolvedReferenceExpression', () => {
+    const inst = new InstanceElement('inst', new ObjectType({ elemID: new ElemID('test', 'type') }))
+    it('should return false for a reference expression with undefind value', () => {
```
alonstern, line 2548: please add a test that it return false for a reference with `UnresolvedReference` value

alonstern (approved):
🔥 🔥

PR #4469, opened 2023-06-06
SALTO-3980: allow async pagination calls

IdoZakk:
Changed the traverseRequests function of pagination to send queued pagination calls simultaneously.
It allows us to process simultaneously both different pages of the same call and also the recurse-into.
Also Added a pagination function for Jira that utilizes this behavior and after the first page returns call all other pages.

---

I feel this solution is not elegant enough code-wise, but other solutions seemed less elegant.

---
_Release Notes_: 
Jira Adapter:
* Improved fetch times (pagination requests will be called simultaneously 

---
_User Notifications_: 
None


packages/adapter-components/src/client/pagination.ts
```diff
@@ -85,10 +93,88 @@ const computeRecursiveArgs = (
...
+// from the response, and an extractPageEntries function that is responsible for extracting the
+// entries from the page.
+// the function processes all existing pages simultaneously, and yields the pages as they arrive.
+// Its primary priority is to keep the number of concurrent requests to a maximum, and deal with results only
+// when there are no requests to send
 export const traverseRequests: (
   paginationFunc: PaginationFunc,
   extractPageEntries: PageEntriesExtractor,
   customEntryExtractor?: PageEntriesExtractor,
+
```
alonstern, line 177: supernit: remove extra line

packages/adapter-components/src/client/pagination.ts
```diff
@@ -85,10 +93,88 @@ const computeRecursiveArgs = (
...
+  }):Promise<PaginationResult> => {
+  const params = { ...queryParams, ...additionalArgs }
+  const response = await client.getSinglePage({
+    url,
+    queryParams: Object.keys(params).length > 0 ? params : undefined,
+    headers,
+  })
+  if (response.status !== 200) {
+    log.warn(`error getting result for ${url}: %s %o %o`, response.status, response.statusText, response.data)
+    return Promise.reject(new PaginationError('break', objectHash(additionalArgs)))
```
alonstern, line 121: I think you can do 
```suggestion
    throw new PaginationError('break', objectHash(additionalArgs))
```

packages/adapter-components/src/client/pagination.ts
```diff
@@ -85,10 +93,88 @@ const computeRecursiveArgs = (
...
+  const entries = (
+    (!Array.isArray(response.data) && Array.isArray(response.data.items))
+      ? response.data.items
+      : makeArray(response.data)
+  ).flatMap(extractPageEntries)
+
+  // checking original entries and not the ones that passed the custom extractor, because even if all entries are
+  // filtered out we should still continue querying
+  if (entries.length === 0) {
+    return Promise.reject(new PaginationError('continue', objectHash(additionalArgs)))
```
alonstern, line 133:
```suggestion
    throw new PaginationError('continue', objectHash(additionalArgs))
```

packages/adapter-components/src/client/pagination.ts
```diff
@@ -97,61 +183,53 @@ export const traverseRequests: (
...
+        promisesQueue.delete(objectHash(result.additionalArgs))
+      } catch (error) {
+        if (!(error instanceof PaginationError)) {
+          throw error
+        }
+        // no new items returned, continue to next requests
+        promisesQueue.delete(error.hashedArgs)
+        if (error.message === 'break') {
+          // we want to clear requests that were already called but avoid calling new ones
+          requestsQueryArgsQueue.splice(0, requestsQueryArgsQueue.length)
```
alonstern, line 227: I am a bit confused by this, it means you won't yield the results of the values currently in `requestsQueryArgsQueue` but you will continue to wait for the current running promises and yield their pages. Is that what should happen?

packages/adapter-components/src/client/pagination.ts
```diff
@@ -97,61 +183,53 @@ export const traverseRequests: (
   getParams,
 }) {
   const { url, queryParams, recursiveQueryParams, headers } = getParams
-  const requestQueryArgs: Record<string, string>[] = [{}]
+  const requestsQueryArgsQueue: Record<string, string>[] = [{}]
```
alonstern, line 186: I am wondering if we can remove this queue and use only the `promisesQueue` if instead of putting the args in requestsQueryArgsQueue, we will start the promise and add it directly to `promisesQueue`. If we can I think it can simplify the code here a bit

packages/adapter-components/src/client/pagination.ts
```diff
@@ -97,61 +183,53 @@ export const traverseRequests: (
...
+        yield result.page
+        numResults += result.page.length
+        promisesQueue.delete(objectHash(result.additionalArgs))
+      } catch (error) {
+        if (!(error instanceof PaginationError)) {
+          throw error
+        }
+        // no new items returned, continue to next requests
+        promisesQueue.delete(error.hashedArgs)
+        if (error.message === 'break') {
```
alonstern, line 225: I tend to think we shouldn't do this using try-catch. Maybe we can add to `PaginationResult` something like `status` and get it from the return value?

alonstern (commented):
Looks great 👌  We should make an interview question out of this

packages/adapter-components/src/client/pagination.ts
IdoZakk: Yes.
In the original code in case of an error we did `break` under the assumption that future calls are likely to fail and it will be a waste of time to call them.
Following a discussion with Neta I realized we have to wait for the existing promises (as they can cause leaks), and processing them should not have a big time impact as we are already waiting for them to finish.

So we are waiting for the promises to finish, but emptying the request queue.

alonstern: I understand why you want to wait for the promises, but you are also going to keep yielding their pages and the new pages they add

IdoZakk: I will yield them only if they succeed. If by chance it happens then it is good. The reason to break was not to waste time on likely to failed calls.

packages/adapter-components/src/client/pagination.ts
```diff
@@ -85,6 +88,99 @@ const computeRecursiveArgs = (
     : []
 )
 
+const allSettled = async <T>(promises: IterableIterator<Promise<T>>): Promise<void> => {
+  await awu(promises)
+    .forEach(async promise => promise.catch(_e => undefined))
```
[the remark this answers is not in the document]
IdoZakk: NM, I found how to fix

packages/adapter-components/src/client/pagination.ts
```diff
@@ -85,6 +87,98 @@ const computeRecursiveArgs = (
...
+  const params = { ...queryParams, ...additionalArgs }
+
+  const response = await client.getSinglePage({
+    url,
+    queryParams: Object.keys(params).length > 0 ? params : undefined,
+    headers,
+  })
+  if (response.status !== 200) {
+    log.warn(`error getting result for ${url}: %s %o %o`, response.status, response.statusText, response.data)
+    return { page: [], additionalArgs, yieldResult: false }
```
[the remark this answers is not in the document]
IdoZakk: I was thinking about it. I was not 100% sure that empty pages should not yield. For instance, there are cases today when all the results were filtered out, but you still yield the result.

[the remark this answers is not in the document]
alonstern: @netama WDYT? is there a point in returning empty pages?

alonstern (commented):
Looks good 🔥 @netama probably should review this as well

packages/adapter-components/src/client/pagination.ts
```diff
@@ -96,64 +193,44 @@ export const traverseRequests: (
...
+    extractPageEntries,
+    customEntryExtractor,
+    pageSize,
+    usedParams,
+  }
+  pushPage(pageArgs, {})
+  while (promisesQueue.size > 0) {
+    try {
+      // eslint-disable-next-line no-await-in-loop
+      const result = await Promise.race(promisesQueue.values())
```
netama, line 213: this should be tested pretty thoroughly, I'm not sure it behaves that well on errors

netama (commented):
sorry about the delay - added a few nits, but in general I think it might be safer to put this _next_ to the existing implementation for now after all like you planned to initially, there are some small behavior changes + I am not convinced the `Promise.race()` logic is completely safe. once we see it working well for jira with some more cases we can decide what else to migrate. wdyt?

packages/adapter-components/src/client/pagination.ts
```diff
@@ -269,46 +346,38 @@ export const getWithCursorPagination = (pathChecker = defaultPathChecker): Pagin
   return nextPageCursorPages
 }
 
-export const getWithOffsetAndLimit = (): PaginationFunc => {
```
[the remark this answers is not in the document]
IdoZakk: It was removed as it was only used in Jira, the `isLast`and `values` combination seemed a bit specific. After consulting with Alon we decided to remove it. Do you think it is a global pattern?

packages/adapter-components/src/client/pagination.ts
```diff
@@ -85,6 +87,101 @@ const computeRecursiveArgs = (
...
+  const params = { ...queryParams, ...additionalArgs }
+
+  const response = await client.getSinglePage({
+    url,
+    queryParams: Object.keys(params).length > 0 ? params : undefined,
+    headers,
+  })
+  if (response.status !== 200) {
+    log.warn(`error getting result for ${url}: %s %o %o`, response.status, response.statusText, response.data)
+    return { response, page: [], additionalArgs, yieldResult: false }
```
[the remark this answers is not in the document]
IdoZakk: To some extent (given that the situation is new), we have discussed this in person (and there is a thread about it [here](https://github.com/salto-io/salto/pull/4469#discussion_r1219789520)
We used to stop completely on such an event using `break`, here we might already have some calls in progress.
Per our discussion, you mentioned that in case of an HTTP error we used to stop to prevent additional calls to most probably soon-to-fail calls.
Now, in case of a failure we will stop future calls, but any already running call will be resolved, and if any such call is successful it means there is no point in stopping.
Do you find this problematic or risky?

packages/adapter-components/src/client/pagination.ts
```diff
@@ -96,64 +193,44 @@ export const traverseRequests: (
   pageSize,
   getParams,
 }) {
-  const { url, queryParams, recursiveQueryParams, headers } = getParams
-  const requestQueryArgs: Record<string, string>[] = [{}]
   const usedParams = new Set<string>()
+  const promisesQueue: Map<string, Promise<PaginationResult>> = new Map()
```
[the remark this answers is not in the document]
IdoZakk: The way I look at it, it is a priority queue of promises that are being popped when they are fulfilled. I changed the code to reflect it better

packages/adapter-components/src/client/pagination.ts
IdoZakk: As far as I understand an error will cause an exception to be thrown, in which case the flow is pretty straightforward (waiting for all the existing promises to resolve and finishing). What flows do you think we should test here?

packages/adapter-components/src/client/pagination.ts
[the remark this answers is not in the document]
IdoZakk: @netama making sure you saw this

IdoZakk:
Pushed fixes to @netama 's code review
I made two traverseRequest functions. I have split it into 3 commits as I made structure changes and wanted to keep the review process reasonable. The second commit is just structure change

packages/adapter-components/src/client/pagination.ts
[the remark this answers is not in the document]
netama: sorry about the delay, we don't need to yield if the page is empty

packages/jira-adapter/src/adapter.ts
```diff
@@ -306,6 +306,7 @@ export default class JiraAdapter implements AdapterOperations {
       client: this.client,
       paginationFuncCreator: paginate,
       customEntryExtractor: removeScopedObjects,
+      asyncRun: true,
```
alonstern, line 309: if we made this configurable, can we add an adapter config option to control this?

packages/adapter-components/src/client/pagination.ts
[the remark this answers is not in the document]
IdoZakk: @alonstern, we need the yieldResult, as an empty page might come from the extractor and we still want to handle the next pages. So just not yielding on empty pages does not look like it is worth the change

alonstern (approved):
🔥

PR #4471, opened 2023-06-06
SALTO-3751 - Salesforce: re-enable history tracking

tomsellek:
Undo the revert and add logging to catch the case where we don't get the 'enableHistory' annotation even though history tracking is enabled.

---

Following discussion with @amitgoldarb , we believe the issue is not worth reverting the change over.
Original PR is at https://github.com/salto-io/salto/pull/4400

---
_Release Notes_: 
Salesforce Adapter:
Improved modelling of History and Feed History Tracking on Standard and Custom Objects.

---
_User Notifications_: 
Salesforce Adapter:


amitgoldarb (approved):
Thanks for the quick fix

PR #4473, opened 2023-06-07
SALTO-8004: added getAdditionalReferences method

alonstern:
added `getAdditionalReferences` method and implemented it in Salesforce

---
_Release Notes_: 
None (no effect on the user yet)

---
_User Notifications_: 
None


packages/salesforce-adapter/src/additional_references.ts
```diff
@@ -0,0 +1,51 @@
...
+
+const { awu } = collections.asynciterable
+
+
+export const getAdditionalReferences: GetAdditionalReferencesFunc = async changes => {
+  const profilesAndPermissionSets = changes
+    .filter(isInstanceChange)
+    .filter(isRemovalOrModificationChange)
+    .map(getChangeData)
+    .filter(instance => ['Profile', 'PermissionSet'].includes(instance.elemID.typeName))
```
adamro, line 31: nit: Use the `PERMISSION_SET_METADATA_TYPE` and `PROFILE_METADATA_TYPE` constants

packages/salesforce-adapter/src/additional_references.ts
```diff
@@ -0,0 +1,51 @@
...
+    .flatMap(async field => {
+      const fieldApiName = await safeApiName(field)
+      if (fieldApiName === undefined) {
+        return []
+      }
+      return profilesAndPermissionSets
+        .filter(instance => _.get(instance.value.fieldPermissions, fieldApiName) !== undefined)
+        .map(instance => ({
+          source: field.elemID,
+          target: instance.elemID.createNestedID('fieldPermissions', ...fieldApiName.split(API_NAME_SEPARATOR)),
```
adamro, line 47: Can you move the existing `const FIELD_PERMISSIONS = 'fieldPermissions'` code to the constants file and use it from there pls?

packages/salesforce-adapter/src/additional_references.ts
```diff
@@ -0,0 +1,51 @@
...
+import { API_NAME_SEPARATOR } from './constants'
+
+const { awu } = collections.asynciterable
+
+
+export const getAdditionalReferences: GetAdditionalReferencesFunc = async changes => {
+  const profilesAndPermissionSets = changes
+    .filter(isInstanceChange)
+    .filter(isRemovalOrModificationChange)
+    .map(getChangeData)
```
adamro, line 30: Use `isInstanceOfType` here instead of the next 2 lines

packages/salesforce-adapter/src/additional_references.ts
```diff
@@ -0,0 +1,51 @@
...
+import { collections } from '@salto-io/lowerdash'
+import { isFieldOfCustomObject } from './transformers/transformer'
+import { safeApiName } from './filters/utils'
+import { API_NAME_SEPARATOR } from './constants'
+
+const { awu } = collections.asynciterable
+
+
+export const getAdditionalReferences: GetAdditionalReferencesFunc = async changes => {
+  const profilesAndPermissionSets = changes
```
adamro, line 27: Maybe if there are no Field addition Changes we shouldn't do this run? Also, if we want even more optimizations just run on the changes once and filter it to -
{
    profileModifyRemove: ...
    fieldAdditions: ...
}
and then do the logic to not run on it twice

adamro (commented):
Started with the SF logic. Will look at the api levels next

packages/core/src/api.ts
```diff
@@ -670,3 +670,24 @@ export const rename = async (
 
 export const getAdapterConfigOptionsType = (adapterName: string): ObjectType | undefined =>
   adapterCreators[adapterName].configCreator?.optionsType
+
+
+export const getAdditionalReferences = async (
+  workspace: Workspace,
+  changes: Change[],
+): Promise<ReferenceMapping[]> => {
```
adamro, line 678: I think this is OK but just throwing ideas to what else we could choose to return - 
* Grouped by account
* Indexed by elemID (with or without outgoing and incoming)

adamro:
@ori-moisis @netama  - Please look at the core and adapters api changes and see if you have strong objects and/or thoughts bout the discussions points I think we can have there (or anything else you have in mind :D )

packages/core/src/api.ts
```diff
@@ -670,3 +670,24 @@ export const rename = async (
...
+  const accountToService = getAccountToServiceNameMap(workspace, workspace.accounts())
+
+  const changeGroups = _(changes)
+    .groupBy(change => getChangeData(change).elemID.adapter)
+    .values()
+    .value()
+
+  const referenceGroups = await Promise.all(
+    changeGroups.map(changeGroup => adapterCreators[accountToService[getChangeData(changeGroup[0]).elemID.adapter]]
```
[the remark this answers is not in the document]
alonstern: because the map function returns a promise so I can only flat it after the promise.all

packages/adapter-api/src/adapter.ts
```diff
@@ -163,6 +163,13 @@ export type LoadElementsFromFolderArgs = {
   elementSource: ReadOnlyElementsSource
 }
 
+export type ReferenceMapping = {
```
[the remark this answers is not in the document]
alonstern: I changed the reference_index to use this type as well. I guess we can change it in the future if there will be a difference between them 🤷

packages/core/src/api.ts
alonstern: In the SaaS code that uses it I don't need grouping since I'm just taking the reference mappings and save them to the DB as is

packages/core/src/api.ts
```diff
@@ -670,3 +670,24 @@ export const rename = async (
...
+  changes: Change[],
+): Promise<ReferenceMapping[]> => {
+  const accountToService = getAccountToServiceNameMap(workspace, workspace.accounts())
+
+  const changeGroups = _(changes)
+    .groupBy(change => getChangeData(change).elemID.adapter)
+    .values()
+    .value()
+
+  const referenceGroups = await Promise.all(
```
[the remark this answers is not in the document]
alonstern: I think it complicates the code a bit since I need to set the type of the adapter as `Adapter & { getAdditionalReferences: GetAdditionalReferencesFunc }` after the check (and there is no harm in passing undefined to Promise.all)

adamro (approved):
🔥

PR #4474, opened 2023-06-07
SALTO-4309 - Salesforce: add logs to history tracking filter

tomsellek:
Log cases where the object-level toggle is missing yet there are other indications that the object supports history tracking.

---

See the comments in https://salto-io.atlassian.net/browse/PROD-93 for background on why this situation happens. We want to get a better understanding of which types exhibit this behaviour so we can decide on a workaround.

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/filters/centralize_tracking_info.ts
+++ b/packages/salesforce-adapter/src/filters/centralize_tracking_info.ts
@@ -89,17 +89,35 @@ const deleteFieldHistoryTrackingAnnotation = (field: Field, trackingDef: Tracked
 }
 
 const centralizeHistoryTrackingAnnotations = (customObject: ObjectType, trackingDef: TrackedFieldsDefinition): void => {
+  const areAnyFieldsTracked = (obj: ObjectType): boolean => (
+    Object.values(obj.fields).some(field => field.annotations[trackingDef.fieldLevelEnable] === true)
+  )
+  const isTrackingSupported = (obj: ObjectType): boolean => (
+    obj.annotations[trackingDef.objectLevelEnable] !== undefined
+  )
+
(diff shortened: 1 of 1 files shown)


packages/salesforce-adapter/src/filters/centralize_tracking_info.ts
```diff
@@ -89,17 +89,35 @@ const deleteFieldHistoryTrackingAnnotation = (field: Field, trackingDef: Tracked
...
+        customObject.elemID.getFullName(),
+        trackingDef.objectLevelEnable,
+        customObject.annotations[trackingDef.objectLevelEnable],
+        trackingDef.recordTypeEnable)
+    }
+    if (!isTrackingSupported(customObject) && areAnyFieldsTracked(customObject)) {
+      log.debug('In object type %s, %s is %s but some fields have tracking enabled',
+        customObject.elemID.getFullName(),
+        trackingDef.objectLevelEnable,
+        customObject.annotations[trackingDef.objectLevelEnable])
```
benisraeladi, line 119: It might be interesting to log which fields are tracked. There might be a pattern here that we could use

amitgoldarb (approved):
Thanks for the quick logs!

packages/salesforce-adapter/src/filters/centralize_tracking_info.ts
tomsellek: So the reason we're doing this is to sort of double-verify the Salesforce docs that say certain types will exhibit this behaviour. We're not actually trying to come up with a theory (which would be a great reason for logging the field names!), but rather to make sure Salesforce behaves how it says it would.

PR #4476, opened 2023-06-08
SALTO-4314: invalid resolve prevents deletion

edenhassid:
invalid resolve prevents deletion

---

---
_Release Notes_: 
zendesk:
* fix error when deleting a `dynamic_content_item_variant` with missing locale 

---
_User Notifications_: 
None


netama (approved):
thanks 🚤  LGTM, we will need to follow up with some more type safeties separately

PR #4479, opened 2023-06-08
SALTO-4263: Do not override properties for system fields

tamtamirr:
For all the **System Fields** we overridden their `createable | updatable` annotation values to false with the values returned from the API.

---

Workspace Diff: https://github.com/salto-io/tamir-sf/commit/af0b2b6743d8ada7491f7c4378ab8ac99ca4d045

---
_Release Notes_: 
_Salesforce Adapter_:
- Correct createable and updateable annotation values for Salesforce System Fields.

---
_User Notifications_: 
Salesforce:
- Changes to the values of the annotations **createable** and **updateable** of System Fields in existing nacls.

### DIFF
--- a/packages/salesforce-adapter/src/filters/custom_object_instances_references.ts
+++ b/packages/salesforce-adapter/src/filters/custom_object_instances_references.ts
@@ -26,7 +26,16 @@ import {
   safeJsonStringify,
 } from '@salto-io/adapter-utils'
 import { logger } from '@salto-io/logging'
-import { Element, Values, Field, InstanceElement, ReferenceExpression, SaltoError, ElemID } from '@salto-io/adapter-api'
+import {
+  Element,
+  Values,
+  Field,
+  InstanceElement,
+  ReferenceExpression,
+  SaltoError,
+  ElemID,
+  CORE_ANNOTATIONS,
+} from '@salto-io/adapter-api'
 import { FilterResult, RemoteFilterCreator } from '../filter'
 import { apiName, isInstanceOfCustomObject, isCustomObject } from '../transformers/transformer'
 import { FIELD_ANNOTATIONS, KEY_PREFIX, KEY_PREFIX_LENGTH, SALESFORCE } from '../constants'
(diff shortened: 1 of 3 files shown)


benisraeladi (approved):

packages/salesforce-adapter/src/filters/custom_object_instances_references.ts
```diff
@@ -168,10 +179,10 @@ const replaceLookupsWithRefsAndCreateRefMap = async (
     instance: InstanceElement
   ): Promise<Values> => {
     const transformFunc: TransformFunc = async ({ value, field }) => {
-      if (!isReferenceField(field)) {
+      if (!isDefined(field) || !shouldHandleLookupRef(field)) {
```
amitgoldarb, line 182: Nit: Can we add the isDefinedField back to the shouldhandleLookupRef? It's a style nit, but I find it more readable when everything is in the same function.

packages/salesforce-adapter/src/filters/custom_object_instances_references.ts
```diff
@@ -150,8 +159,10 @@ const createWarnings = async (
   ]
 }
 
-const isReferenceField = (field?: Field): field is Field => (
-  isDefined(field) && (isLookupField(field) || isMasterDetailField(field))
+const shouldHandleLookupRef = (field: Field): boolean => (
+  (isLookupField(field) || isMasterDetailField(field))
+  && (field.annotations?.[FIELD_ANNOTATIONS.CREATABLE] || field.annotations?.[FIELD_ANNOTATIONS.UPDATEABLE])
+  && !(field.annotations?.[CORE_ANNOTATIONS.HIDDEN_VALUE] === true)
```
amitgoldarb, line 165: We will now have less logs from this filter because we validate some of the things earlier.
I think we can move this part of the validation to where it was earlier, this way we won't remove these logs that might be useful

amitgoldarb (approved):
LGTM
Added 1 comment and 1 nit. Both aren't blockers IMO.

packages/salesforce-adapter/src/filters/custom_object_instances_references.ts
tamtamirr: Typescript wise it makes less sense IMO. A method should not receive possibly undefined parameter unless it has special handing for it. (This is a better way to to force invokers sending a field and not possibly undefined field)

PR #4480, opened 2023-06-08
SALTO-4318 - Salesforce: use retrieve API for PermissionSet and CustomObject

tomsellek:
Add PermissionSet and CustomObject to the list of types we fetch using the retrieve API.

 - [x] [PROD ticket](https://salto-io.atlassian.net/browse/PROD-140)

---

We want to eventually use the retrieve API for all types because it provides better data, behaves better in regards to rate limits and is more similar to SFDX. As part of our work to improve our support for Salesforce Profiles, we need to use retrieve to fetch these two types.

[WS Diff](https://github.com/salto-io/tomsellek-sf/pull/9)

---
_Release Notes_: 
Salesforce: Use a different API to fetch custom objects.

---
_User Notifications_: 
Salesforce: Changing the API used for fetching custom objects will result in several changes across all custom object types, including additional data as well as changes to existing fields.

### DIFF
--- a/packages/salesforce-adapter/src/adapter.ts
+++ b/packages/salesforce-adapter/src/adapter.ts
@@ -230,6 +230,7 @@ const METADATA_TO_RETRIEVE = [
   'Certificate', // contains encoded zip content
   'ContentAsset', // contains encoded zip content
   'CustomMetadata', // For the XML attributes
+  'CustomObject',
   'Dashboard', // contains encoded zip content, is under a folder
   'DashboardFolder',
   'Document', // contains encoded zip content, is under a folder
@@ -239,6 +240,7 @@ const METADATA_TO_RETRIEVE = [
   'EmailTemplate', // contains encoded zip content, is under a folder
   'LightningComponentBundle', // Has several fields with base64Binary encoded content
   'NetworkBranding', // contains encoded zip content
+  'PermissionSet',
   'Report', // contains encoded zip content, is under a folder
   'ReportFolder',
   'ReportType',
--- a/packages/salesforce-adapter/src/adapter.ts
+++ b/packages/salesforce-adapter/src/adapter.ts
@@ -351,7 +351,7 @@ export default class SalesforceAdapter implements AdapterOperations {
 
     const fetchProfile = buildFetchProfile(config.fetch ?? {})
     this.fetchProfile = fetchProfile
-    if (this.fetchProfile.isFeatureEnabled('fetchCustomObjectUsingReadApi')) {
+    if (this.fetchProfile.isFeatureEnabled('fetchCustomObjectUsingRetrieveApi')) {
       _.pull(this.metadataToRetrieve, CUSTOM_OBJECT)
     }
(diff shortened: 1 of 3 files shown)


packages/salesforce-adapter/src/adapter.ts
```diff
@@ -239,6 +240,7 @@ const METADATA_TO_RETRIEVE = [
   'EmailTemplate', // contains encoded zip content, is under a folder
   'LightningComponentBundle', // Has several fields with base64Binary encoded content
   'NetworkBranding', // contains encoded zip content
+  'PermissionSet',
```
amitgoldarb, line 243: Interesting. This doesn't behave the same way as profiles, as in when you fetch it alone you receive all of it's fields?

amitgoldarb (commented):
I wonder if we can roll this out under a FF? I wonder if we should?
Let's discuss in our 1-1 later today f2f, will be easier to close the loop.

packages/salesforce-adapter/src/adapter.ts
tomsellek: Correct

amitgoldarb (approved):
LGTM :)

packages/salesforce-adapter/src/types.ts
```diff
@@ -79,6 +79,7 @@ export type OptionalFeatures = {
   skipParsingFormulas?: boolean // Negative flag because we want it disabled by default and optional features are
                                 // enabled by default
   skipAliases?: boolean
+  fetchCustomObjectUsingReadApi?: boolean
```
benisraeladi, line 82: Shouldn't this be using retrieve API?

tamtamirr: Yes, this one is rather confusing.
When an **optionalFeature** is not in the adapter config, it defaults to true.
We want a **retrieveCustomObjects** optional feature, that can be turned off explicitly with **false** value, but enabled by default.

tomsellek: @tamtamirr 
So I guess the question is do we want this enabled or disabled by default. The way I coded it it's disabled by default. I take it we want to enable it?

tomsellek: nvm, I made the change.

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -349,6 +351,9 @@ export default class SalesforceAdapter implements AdapterOperations {
 
     const fetchProfile = buildFetchProfile(config.fetch ?? {})
     this.fetchProfile = fetchProfile
+    if (this.fetchProfile.isFeatureEnabled('fetchCustomObjectUsingRetrieveApi')) {
```
tamtamirr, line 354:
```suggestion
    if (!this.fetchProfile.isFeatureEnabled('fetchCustomObjectUsingRetrieveApi')) {
```

If the feature is disabled we want to take it out, not?

tamtamirr (approved):
Please negate the **isFeatureEnabled** if. Approving to save cycles

PR #4481, opened 2023-06-11
SALTO-4180: Salesforce API version 58

tamtamirr:
Bump to API Version 58

---

Workspace Diff: https://github.com/salto-io/tamir-sf/commit/7f921046a8b3cf1f126a7c0ce3c36cbb9c455342
Prod Ticket: https://salto-io.atlassian.net/browse/PROD-137

---
_Release Notes_: 
Salesforce Adapter:
- The adapter will use API version 58.

---
_User Notifications_: 
Salesforce:
- Changes in your `nacls` due to API version bump.


tamtamirr:
There's a UT failing due to regression. Taking a look into it.

benisraeladi (approved):

packages/salesforce-adapter/test/adapter.crud.test.ts
```diff
@@ -347,7 +347,7 @@ describe('SalesforceAdapter CRUD', () => {
                     deploy: {
                       quickDeployParams: {
                         requestId: '1',
-                        hash: '5a4f3c951aeb38445b20edd27747b028',
+                        hash: 'bd17f95fb82eb6eac93a764cf0910054',
```
amitgoldarb, line 350: Can we understand why this hash changed? I don't think it should be dependent on the API version if possible.

amitgoldarb (commented):
Added a comment 🙏

packages/salesforce-adapter/test/adapter.crud.test.ts
tamtamirr: @amitgoldarb The API version we use is sent as part of the deployed zip. Nothing we can do here

benisraeladi: @amitgoldarb Regardless of whether we can change that or not, I think it's correct to not be able to quick deploy if the API version changed. Since there can be so many changes that we cannot control, even if there was a valid check before the version change it might not hold true after a version change.

amitgoldarb: I agree, @benisraeladi 's explanation makes sense.

PR #4482, opened 2023-06-11
SALTO-4288 add mapping between type and full name and ElemID

benisraeladi:
Add mapping so that during failures we will be able to know the element ID of the instance.

---

---
_Release Notes_: 
_None_

---
_User Notifications_: _None_

### DIFF
--- a/packages/salesforce-adapter/src/metadata_deploy.ts
+++ b/packages/salesforce-adapter/src/metadata_deploy.ts
@@ -197,8 +197,10 @@ const isUnFoundDelete = (message: DeployMessage, deletionsPackageName: string): 
 const processDeployResponse = (
   result: SFDeployResult,
   deletionsPackageName: string,
+  changeDeployedIds: Record<string, MetadataIdsMap>,
   checkOnly: boolean,
-): { successfulFullNames: ReadonlyArray<MetadataId>; errors: ReadonlyArray<SaltoError | SaltoElementError> } => {
+): { successfulFullNames: ReadonlyArray<MetadataId>
+  errors: ReadonlyArray<SaltoError | SaltoElementError> } => {
   const allFailureMessages = makeArray(result.details)
     .flatMap(detail => makeArray(detail.componentFailures))
 
@@ -219,6 +221,7 @@ const processDeployResponse = (
     .filter(failure => !isUnFoundDelete(failure, deletionsPackageName))
     .map(getUserFriendlyDeployMessage)
     .map(failure => ({
+      elemID: changeDeployedIds[failure.componentType]?.[failure.fullName],
       message: `Failed to ${checkOnly ? 'validate' : 'deploy'} ${failure.fullName} with error: ${failure.problem} (${failure.problemType})`,
       severity: 'Error' as SeverityLevel,
     }))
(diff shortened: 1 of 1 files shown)


packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -27,7 +27,8 @@ import { SaltoError,
   isContainerType,
   isAdditionChange,
   SaltoElementError,
-  SeverityLevel } from '@salto-io/adapter-api'
+  SeverityLevel,
+  ElemID } from '@salto-io/adapter-api'
```
amitgoldarb, line 31: super nit: This will create a cleaner commit next time with only 1 line diffs
```suggestion
  SeverityLevel,
  ElemID,
} from '@salto-io/adapter-api'
```

packages/salesforce-adapter/src/transformers/xml_transformer.ts
```diff
@@ -378,13 +378,15 @@ export type DeployPackage = {
...
   getDeletionsPackageName(): string
+  getFileNamesToElemIDsMapping(): collections.map.DefaultMap<string, ElemID[]>
 }
 
 export const createDeployPackage = (deleteBeforeUpdate?: boolean): DeployPackage => {
   const zip = new JSZip()
   const addManifest = new collections.map.DefaultMap<string, string[]>(() => [])
   const deleteManifest = new collections.map.DefaultMap<string, string[]>(() => [])
   const deletionsPackageName = deleteBeforeUpdate ? 'destructiveChanges.xml' : 'destructiveChangesPost.xml'
+  const fileNamesToElemIDs = new collections.map.DefaultMap<string, ElemID[]>(() => [])
```
amitgoldarb, line 389: Why is this an array of elementId's? Is there a case where there is more than one? I wonder if we simply choose the first like we did in this implementation, how will the error look?

packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -219,6 +221,7 @@ const processDeployResponse = (
     .filter(failure => !isUnFoundDelete(failure, deletionsPackageName))
     .map(getUserFriendlyDeployMessage)
     .map(failure => ({
+      elemID: fileNamesToElemIDs.get(failure.fileName)[0],
```
amitgoldarb, line 224: Can we possibly return the whole array? Is this supported?

amitgoldarb (commented):
Looks good.
Added some thoughts, also I think we should probably add some UT to here.

packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -379,7 +382,7 @@ export const deployMetadata = async (
   }, undefined, 2))
 
   const { errors, successfulFullNames } = processDeployResponse(
-    sfDeployRes, pkg.getDeletionsPackageName(), checkOnly ?? false
+    sfDeployRes, pkg.getDeletionsPackageName(), pkg.getFileNamesToElemIDsMapping(), checkOnly ?? false
```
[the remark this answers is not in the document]
benisraeladi: @amitgoldarb The problem is that we get this `MetadataId`, and unless we also bring all the changes into `processDeployResponse` the only way to do this is to return some `SaltoSalesforceElementError` add change it to `SaltoElementError` in the level above. It's doable, but it might not be very pretty/readable

tamtamirr (approved):
LGTM

tomsellek (approved):
LGTM! Thanks for the explanation!

PR #4484, opened 2023-06-11
SALTO-4301: update static file path

edenhassid:
update static file path.
until now, static files whose path changed but content didn't would remain in their old location, after this PR,  static files whose path changed will be created in their new location and the old files will be deleted.

---

---
_Release Notes_: 
core:
* static file path changes will be referred to as a change in fetch

---
_User Notifications_: 
* some static files might move location

### DIFF
--- a/packages/adapter-api/src/values.ts
+++ b/packages/adapter-api/src/values.ts
@@ -32,7 +32,7 @@ export interface Values {
 }
 
 export type CompareOptions = {
-  compareReferencesByValue?: boolean
+  compareByValue?: boolean
 }
 
 export const calculateStaticFileHash = (content: Buffer): string =>
@@ -250,7 +250,7 @@ const shouldCompareByValue = (
   first: Value,
   second: Value,
   options?: CompareOptions,
-): boolean => Boolean(options?.compareReferencesByValue)
+): boolean => Boolean(options?.compareByValue)
   && shouldResolve(first)
   && shouldResolve(second)
 
@@ -261,7 +261,9 @@ export const compareSpecialValues = (
 ): boolean | undefined => {
   if (isStaticFile(first) && isStaticFile(second)) {
     return first.isEqual(second)
-      && (options?.compareReferencesByValue !== true ? first.filepath === second.filepath : true)
+      && (options?.compareByValue
+        ? true
+        : first.filepath === second.filepath)
   }
   if (isReferenceExpression(first) || isReferenceExpression(second)) {
     if (shouldCompareByValue(first, second, options)) {
--- a/packages/adapter-api/src/values.ts
+++ b/packages/adapter-api/src/values.ts
@@ -14,6 +14,7 @@
 * limitations under the License.
 */
 import _ from 'lodash'
+import path from 'path'
 import { inspect } from 'util'
 import { hash as hashUtils } from '@salto-io/lowerdash'
 // import { ElementsSource } from '@salto-io/workspace'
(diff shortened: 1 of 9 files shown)


packages/adapter-api/src/values.ts
```diff
@@ -261,6 +261,7 @@ export const compareSpecialValues = (
 ): boolean | undefined => {
   if (isStaticFile(first) && isStaticFile(second)) {
     return first.isEqual(second)
+      && (options?.compareReferencesByValue !== true ? first.filepath === second.filepath : true)
```
netama, line 264: readability:
```suggestion
      && (options?.compareReferencesByValue !== true
        ? first.filepath === second.filepath
        : true)
```
or (I think slightly better)
```suggestion
      && (options?.compareReferencesByValue
        ? true
        : first.filepath === second.filepath)
```

netama (commented):
I think it's best that @ori-moisis will review the nacl-file-source part as you did this one together :) 
I think we are still missing:
1. tests
2. a decision about the flag name - if this is only used internally, maybe `compareByValue` with an explanation that this is covering reference expressions and static files?
3. maybe a bit more context in the PR description - it is technically "there" but not clear to someone not in the details, so maybe add something that explains that until now, static files whose path changed but content didn't would remain in the old location, and this is changing? + we should add a User Notifications section that says that some static files might move

packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
```diff
@@ -828,6 +828,15 @@ const buildNaclFilesSource = (
         .map(getNestedStaticFiles)
         .flat()
         .forEach(file => staticFilesSource.delete(file))
+
+      fileChanges
```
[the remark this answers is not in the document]
ori-moisis: a bit more context
there is already a bug there where `set` does not remove static files that were removed from the element being overwritten.
usually this is not an issue because we call `override` in fetch, but it is still technically a bug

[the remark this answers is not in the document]
edenhassid: discussed IRL, shouldn't do anything for now

packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
```diff
@@ -828,6 +828,18 @@ const buildNaclFilesSource = (
...
+        .filter(isModificationChange)
+        .filter(change => isStaticFile(change.data.before))
+        .filter(change => (
+          (isStaticFile(change.data.after) && change.data.before.filepath !== change.data.after.filepath)
```
netama, line 836: I'm wondering if we might need to normalize the path somehow (e.g. if we change from `a//b` to `a/b` then the created file would need to rename in the same place, @ori-moisis I have some vague recollection of needing to do this somewhere a while back, any chance you know? if not I'll check

netama (approved):
pre-approving with a couple of blockers (if no action is needed the implementation LGTM)

packages/adapter-api/src/values.ts
```diff
@@ -32,7 +32,7 @@ export interface Values {
 }
 
 export type CompareOptions = {
-  compareReferencesByValue?: boolean
+  compareByValue?: boolean
```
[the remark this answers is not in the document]
edenhassid: checked with @alonstern and he said it is not used in the SAAS

packages/adapter-api/src/values.ts
```diff
@@ -57,7 +58,7 @@ export class StaticFile {
   public readonly encoding: BufferEncoding
   private internalContent?: Buffer
   constructor(params: StaticFileParameters) {
-    this.filepath = params.filepath
+    this.filepath = path.normalize(params.filepath)
```
ori-moisis, line 61: missing tests for this functionality

edenhassid: I adden in here https://github.com/salto-io/salto/pull/4484/files#:~:text=const%20fileFunc1%20%3D%20new%20StaticFile(%7B%20filepath%3A%20%27some//path.ext%27%2C%20content%3A%20Buffer.from(%27ZOMG%27)%20%7D)

PR #4488, opened 2023-06-12
SALTO-4330: Expand verifyCredentials api to return isProduction and accountType

shoshana-houri:
Expand validateCredentials api to return accountType and isProduction.

---

We would like to be able to identify whether a new account is production or not.
Each adapter will have its own logic to determine the isProduction and account type fields which will be returned as part of the 'validateCredentials' function.

---
_Release Notes_: 
Expand verifyCredentials return type:
Before - returned { success: true, accountId: string } in case of success.
After - returns { success: true, accountId: string, accountType?: string, isProduction?: boolean }.

---
_User Notifications_:

### DIFF
--- a/packages/adapter-api/src/adapter.ts
+++ b/packages/adapter-api/src/adapter.ts
@@ -153,9 +153,9 @@ export type AdapterInstallResult = AdapterSuccessInstallResult | AdapterFailureI
 export const isAdapterSuccessInstallResult = (result: AdapterInstallResult):
   result is AdapterSuccessInstallResult => result.success
 
-export type Account = {
+export type AccountInfo = {
   accountId: string
-  accountType: string
+  accountType?: string
   isProduction?: boolean
 }
 
@@ -178,7 +178,7 @@ export type GetAdditionalReferencesFunc = (changes: Change[]) => Promise<Referen
 
 export type Adapter = {
   operations: (context: AdapterOperationsContext) => AdapterOperations
-  validateCredentials: (config: Readonly<InstanceElement>) => Promise<Account>
+  validateCredentials: (config: Readonly<InstanceElement>) => Promise<AccountInfo>
   authenticationMethods: AdapterAuthentication
   configType?: ObjectType
   configCreator?: ConfigCreator
--- a/packages/adapter-api/src/adapter.ts
+++ b/packages/adapter-api/src/adapter.ts
@@ -153,16 +153,10 @@ export type AdapterInstallResult = AdapterSuccessInstallResult | AdapterFailureI
 export const isAdapterSuccessInstallResult = (result: AdapterInstallResult):
   result is AdapterSuccessInstallResult => result.success
 
-export type AccountId = string
-
-export type AccountType = string
-
-export type IsProduction = boolean | undefined
-
 export type Account = {
-  accountId: AccountId
-  accountType: AccountType
-  isProduction: IsProduction
+  accountId: string
+  accountType: string
+  isProduction?: boolean
 }
 
 export type ConfigCreator = {
--- a/packages/adapter-api/src/adapter.ts
+++ b/packages/adapter-api/src/adapter.ts
@@ -155,6 +155,16 @@ export const isAdapterSuccessInstallResult = (result: AdapterInstallResult):
 
 export type AccountId = string
 
+export type AccountType = string
+
+export type IsProduction = boolean | undefined
+
+export type Account = {
+  accountId: AccountId
+  accountType: AccountType
+  isProduction: IsProduction
+}
+
 export type ConfigCreator = {
   optionsType: ObjectType
   getConfig: (options?: InstanceElement)
@@ -174,7 +184,7 @@ export type GetAdditionalReferencesFunc = (changes: Change[]) => Promise<Referen
 
 export type Adapter = {
   operations: (context: AdapterOperationsContext) => AdapterOperations
-  validateCredentials: (config: Readonly<InstanceElement>) => Promise<AccountId>
+  validateCredentials: (config: Readonly<InstanceElement>) => Promise<Account>
   authenticationMethods: AdapterAuthentication
   configType?: ObjectType
   configCreator?: ConfigCreator
--- a/packages/adapter-components/src/client/http_connection.ts
+++ b/packages/adapter-components/src/client/http_connection.ts
@@ -16,7 +16,7 @@
 import _ from 'lodash'
 import axios, { AxiosError, AxiosBasicCredentials, AxiosRequestConfig } from 'axios'
 import axiosRetry from 'axios-retry'
-import { AccountId, CredentialError } from '@salto-io/adapter-api'
+import { Account, CredentialError } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
 import { ClientRetryConfig } from './config'
 import { DEFAULT_RETRY_OPTS } from './constants'
@@ -49,9 +49,7 @@ export type APIConnection<T = any, S = any> = {
     => Promise<Response<T>>
 }
 
-export type AuthenticatedAPIConnection = APIConnection & {
-  accountId: AccountId
-}
+export type AuthenticatedAPIConnection = APIConnection & Account
 
 export type RetryOptions = {
   retries: number
@@ -151,10 +149,10 @@ export const createClientConnection = <TCredentials>({
 export const validateCredentials = async <TCredentials>(
   creds: TCredentials,
   createConnectionArgs: ConnectionParams<TCredentials>,
-): Promise<AccountId> => {
+): Promise<Account> => {
   const conn = createClientConnection(createConnectionArgs)
-  const { accountId } = await conn.login(creds)
-  return accountId
+  const { accountId, accountType } = await conn.login(creds)
+  return { accountId, accountType }
 }
 
 export type AuthParams = {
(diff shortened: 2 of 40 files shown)


packages/adapter-api/src/adapter.ts
```diff
@@ -152,6 +152,13 @@ export const isAdapterSuccessInstallResult = (result: AdapterInstallResult):
 
 export type AccountId = string
 
+export type AccountType = 'Production' | 'Other' | 'Unknown'
```
shoshana-houri, line 155: The actual AccountType values are not fully determined yet.

packages/core/src/api.ts
```diff
@@ -62,7 +62,7 @@ const getAdapterFromLoginConfig = (loginConfig: Readonly<InstanceElement>): Adap
 
 type VerifyCredentialsResult = {
     success: true
-    accountId: AccountId
+    account: Account
```
shoshana-houri, line 65: The current change modifies the API, meaning it's a breaking change.
Alternatively, we can just add the accountType to VerifyCredentialsResult.
What do you think will be better?

packages/core/src/api.ts
```diff
@@ -63,6 +64,7 @@ const getAdapterFromLoginConfig = (loginConfig: Readonly<InstanceElement>): Adap
 type VerifyCredentialsResult = {
     success: true
     accountId: AccountId
+    accountType: AccountType
```
shoshana-houri, line 67: I think that returning
```suggestion
    success: true
    account: Account
```
with
```
type Account = {
    accountId: string,
    accountType: AccountType
}
```

Would probably be a better modeling, but it's a breaking change so I preferred adding to the existing one.

WDYT?

ori-moisis: how about this?
```typescript
type VerifyCredentialsResult = 
  ({ success: true } & Account)
  | {
    success: false
    error: Error
  }
```
not as clean as your suggestion, but this is non-breaking

packages/jira-adapter/src/client/connection.ts
```diff
@@ -41,9 +41,10 @@ const getBaseUrl = async (
 
 export const validateCredentials = async (
   { connection }: { connection: clientUtils.APIConnection },
-): Promise<AccountId> => {
+): Promise<Account> => {
   if (await isAuthorized(connection)) {
-    return getBaseUrl(connection)
+    const accountId = await getBaseUrl(connection)
+    return { accountId, accountType: 'Unknown', isProduction: undefined } // TODO: implement actual accountType + isProduction logic
```
ori-moisis, line 47: TODO

packages/okta-adapter/src/adapter_creator.ts
```diff
@@ -128,12 +128,16 @@ export const adapter: Adapter = {
...
     }
   },
-  validateCredentials: async config => validateCredentials(
-    credentialsFromConfig(config),
-    {
-      createConnection,
-    },
-  ),
+  validateCredentials: async config => {
+    const accountId = await validateCredentials(
```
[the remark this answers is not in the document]
ori-moisis: you are correct that the logic differs between adapters, but it is already implement in each adapter through its `connection`'s `credValidateFunc`, for example, in okta this is [here](https://github.com/salto-io/salto/blob/main/packages/okta-adapter/src/client/connection.ts#L47:L47)

so, I think we will have to change the interface of `connection` and update the adapters accordingly

packages/jira-adapter/src/client/connection.ts
shoshana-houri: It's intentional, since each adapter should implement this logic separately.
I created dedicated tasks for each adapter so I can add their numbers to the TODOs. Is that ok?

packages/adapter-api/src/adapter.ts
```diff
@@ -155,6 +155,16 @@ export const isAdapterSuccessInstallResult = (result: AdapterInstallResult):
...
 
+export type AccountType = string
+
+export type IsProduction = boolean | undefined
+
+export type Account = {
+  accountId: AccountId
+  accountType: AccountType
+  isProduction: IsProduction
+}
```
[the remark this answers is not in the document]
ori-moisis: I don't think this feature is critical enough to justify forcing all adapters (current and future) to dive into the details of what "production" means at the first iteration of their implementation.
especially when we have to support the case where it is unknown anyway.

this is something that we have to consider when we add requirements of adapters, each one may seem small, but if we require each small feature to be implemented on the "first drop" of an adapter, it will drive up the cost of developing a new adapter.
this one doesn't feel important enough to justify such a strong requirement IMO.
as for adapters not returning this when we can have this info, this is a good compromise IMO, if it is a significant use case for the adapter, we will implement it. if it isn't, this leaves us the flexibility to prioritize our efforts

packages/jira-adapter/src/client/connection.ts
ori-moisis: if there are tickets, the TODO is not necessary IMO
but if you are taking responsibility of removing all of these TODO comments within the next few weeks and this is just something to help you with your development process then it is ok

packages/adapter-api/src/adapter.ts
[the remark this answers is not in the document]
shoshana-houri: Ok, got it, I'll define it as optional :)
I really appreciate the detailed explanation, thanks!!

packages/adapter-api/src/adapter.ts
```diff
@@ -153,7 +153,11 @@ export type AdapterInstallResult = AdapterSuccessInstallResult | AdapterFailureI
 export const isAdapterSuccessInstallResult = (result: AdapterInstallResult):
   result is AdapterSuccessInstallResult => result.success
 
-export type AccountId = string
+export type Account = {
```
ori-moisis, line 156: nit: maybe call this `AccountInfo` or something?
feels like `Account` is a term that means more in our system (e.g - under workspace we have methods like `addAccount` which refers to something that isn't exactly this)

packages/jira-adapter/test/adapter_creator.test.ts
```diff
@@ -41,23 +41,27 @@ describe('adapter creator', () => {
...
-        )
+        ))
       })
       it('should make an authenticated rest call', () => {
         expect(mockAxiosAdapter.history).toBeDefined()
       })
       it('should return base url as account ID', () => {
         expect(accountId).toEqual('http://my_account.net')
       })
+      it('should return Unknown as account type', () => {
```
ori-moisis, line 58: I don't think we should test for that when it is still not implemented (this relates to why I think `accountType` should be optional as well)

packages/adapter-components/src/client/http_connection.ts
```diff
@@ -151,10 +149,10 @@ export const createClientConnection = <TCredentials>({
 export const validateCredentials = async <TCredentials>(
   creds: TCredentials,
   createConnectionArgs: ConnectionParams<TCredentials>,
-): Promise<AccountId> => {
+): Promise<Account> => {
   const conn = createClientConnection(createConnectionArgs)
-  const { accountId } = await conn.login(creds)
-  return accountId
+  const { accountId, accountType } = await conn.login(creds)
+  return { accountId, accountType }
```
ori-moisis, line 155: why are we only taking these two and not the whole response?

packages/adapter-components/src/client/http_connection.ts
```diff
@@ -49,9 +49,7 @@ export type APIConnection<T = any, S = any> = {
     => Promise<Response<T>>
 }
 
-export type AuthenticatedAPIConnection = APIConnection & {
-  accountId: AccountId
-}
+export type AuthenticatedAPIConnection = APIConnection & Account
```
ori-moisis, line 52: I think here, since this is not an external interface like `validateCredentials`, we don't really have to "flatten" the account info, this can use your suggestion to have the account info under a single key
```
export type AuthenticatedAPIConnection = APIConnection & {
  accountInfo: Account
}
```
only if you think it is better though

packages/adapter-components/src/client/http_connection.ts
shoshana-houri: My bad, thanks!

packages/core/src/api.ts
```diff
@@ -62,7 +62,7 @@ const getAdapterFromLoginConfig = (loginConfig: Readonly<InstanceElement>): Adap
   adapterCreators[loginConfig.elemID.adapter]
 
 type VerifyCredentialsResult = (
-  { success: true} & Account
+  { success: true} & AccountInfo
```
ori-moisis, line 65: nit: missing space
```suggestion
  { success: true } & AccountInfo
```

packages/netsuite-adapter/src/client/sdf_client.ts
```diff
@@ -167,15 +167,16 @@ export default class SdfClient {
...
     const netsuiteClient = new SdfClient({
       credentials,
       globalLimiter: new Bottleneck(),
       instanceLimiter: () => false,
     })
     const { projectPath, authId } = await netsuiteClient.initProject()
     await netsuiteClient.projectCleanup(projectPath, authId)
-    return Promise.resolve(netsuiteClient.credentials.accountId)
+    const accountId = await Promise.resolve(netsuiteClient.credentials.accountId)
+    return { accountId }
```
ori-moisis, line 179: ummm, is this equivalent to
```suggestion
    return { accountId: netsuiteClient.credentials.accountId }
```
?

why do we need to go through Promise.resolve?

ori-moisis (approved):

packages/netsuite-adapter/src/client/sdf_client.ts
shoshana-houri: You're right. I didn't want to change the existing implementation but it is redundant. I'll change that :)

PR #4490, opened 2023-06-12
SALTO-4202: Improve SharePermissions filter's performance

IdoZakk:
Improved the performance of Share Permission Filter
---

The existing filter was one of the most time-consuming.
There was no need for a transformValue, the share permissions exist on two types and into static locations

---
_Release Notes_: 
Jira Adatper:
* Improved fetch performance

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/share_permission.ts
+++ b/packages/jira-adapter/src/filters/share_permission.ts
@@ -13,18 +13,17 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
-import { BuiltinTypes, Change, Element, ElemID, Field, InstanceElement, isInstanceChange, isInstanceElement, isObjectType, ObjectType, ReadOnlyElementsSource, Values } from '@salto-io/adapter-api'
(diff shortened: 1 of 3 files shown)


packages/jira-adapter/src/filters/share_permission.ts
```diff
@@ -67,89 +66,63 @@ const transformSharePermissionValues = (sharePermissionValues: Values): void =>
...
+const transformSharedPermissions = (instance: InstanceElement, func: (sharedPermission: Values) => void): void => {
+  SHARE_PERMISSION_FIELDS.forEach(field => {
+    if (Array.isArray(instance.value[field])) {
+      instance.value[field].forEach(func)
+    }
   })
 }
 
+const isSharePermissionType = (instance: InstanceElement): boolean =>
+  instance.elemID.typeName === DASHBOARD_TYPE
```
alonstern, line 86: nit: can you create a const for `[DASHBOARD_TYPE, FILTER_TYPE_NAME]` and use it here?

alonstern (approved):
👌👌

PR #4491, opened 2023-06-12
SALTO-4322: Data Records Circular Dependencies Workaround

tamtamirr:
Added circular dependencies workaround for additions of Records with references to Records of the same type.

---

The implementation is quite tricky, so I'll explain it in TLDR here:
1. Changed the default resolved value to be the Element itself, incase the resolved value is undefined (Which was the case for Data Instances that were added as part of this deployment)
2. Each operation now returns optionally **instancesToUpdate**, which is a list of instances that needs a further update after the initial operation.
3. The second time, we use **bulk update** to update the data instances that references to other data instances of the same type, to avoid the circular dependency.

---
_Release Notes_: 
Salesforce Adapter:
- Support deploying additions of Data Instances that reference the same type.

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/transformers/reference_mapping.ts
```diff
@@ -819,7 +819,7 @@ const getLookUpNameImpl = (defs = fieldNameToTypeMappingDefs): GetLookupNameFunc
       }
       if (isElement(ref.value)) {
         const defaultStrategy = ReferenceSerializationStrategyLookup.absoluteApiName
-        return defaultStrategy.serialize({ ref, element })
+        return await defaultStrategy.serialize({ ref, element }) ?? ref.value
```
amitgoldarb, line 822: Why this change?

packages/salesforce-adapter/src/transformers/transformer.ts
```diff
@@ -941,31 +941,52 @@ const transformCompoundValues = async (
...
-): Promise<SalesforceRecord> => {
+): Promise<{record: SalesforceRecord; needsUpdate: boolean}> => {
+  let needsUpdate = false
   const instanceType = await instance.getType()
   const valsWithNulls = {
     ..._.mapValues(instanceType.fields, () => null),
-    ...instance.value,
+    ..._.mapValues(instance.value, val => {
+      if (isInstanceElement(val)) {
+        const referencedRecordId = val.value[CUSTOM_OBJECT_ID_FIELD]
```
amitgoldarb, line 951: Are references dereferenced at this stage? Meaning shouldn't the referenced value just be some random string here?
Can you explain this if a bit, I'm not sure I follow.

amitgoldarb (commented):
Looks good!
Added some comments/questions.

packages/salesforce-adapter/src/group_changes.ts
```diff
@@ -31,7 +31,7 @@ const instanceOfCustomObjectChangeToGroupId: ChangeIdFunction = async change =>
   groupId: `${change.action}_${await apiName(await (getChangeData(change) as InstanceElement).getType())}_instances`,
   // CustomObjects instances might have references to instances of the same type (via Lookup
   // fields), and if we deploy them together the reference gets removed.
-  disjoint: isAdditionChange(change),
+  disjoint: false,
```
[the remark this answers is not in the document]
tamtamirr: I've had a F2F discussion with @ori-moisis . Generally, (and the like the comment says) The disjoint `true` was added for the use-case we tackle in this PR. I think it'd be better to even revert @yuribro code for this file, to avoid complexity.

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -229,54 +232,49 @@ export const retryFlow = async (
...
-    return []
+    return { results: [], instancesToUpdate: [] }
   }
-  const results = await client.bulkLoadOperation(
-    typeName,
-    'update',
-    await instancesToUpdateRecords(instances)
-  )
-  return groupInstancesAndResultsByIndex(results, instances)
+  const { records, instancesToUpdate } = await instancesToRecords(instances, FIELD_ANNOTATIONS.UPDATEABLE)
```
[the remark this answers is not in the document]
tamtamirr: Not sure it's a possible scenario considering it's existing instances, but there's no reason to explicitly ignore it IMO, to avoid unknown scenarios we are not aware of.

packages/salesforce-adapter/src/transformers/reference_mapping.ts
tamtamirr: Discussed F2F.
TLDR: In case the resolved value is undefined (Like in that case, Addition of **Data Record Instance**, we return the instance itself so it serves as a place-holder for the API name later.

packages/salesforce-adapter/src/transformers/transformer.ts
tamtamirr: See https://github.com/salto-io/salto/pull/4491/files#r1234917246

packages/salesforce-adapter/src/transformers/transformer.ts
```diff
@@ -941,31 +941,52 @@ const transformCompoundValues = async (
...
+      }
+      return val
+    }),
   }
   const filteredRecordValues = {
     [CUSTOM_OBJECT_ID_FIELD]: instance.value[CUSTOM_OBJECT_ID_FIELD],
     ..._.pickBy(
       valsWithNulls,
-      (_v, k) => (instanceType).fields[k]?.annotations[fieldAnnotationToFilterBy]
+      (v, k) => !isInstanceElement(v) && instanceType.fields[k]?.annotations[fieldAnnotationToFilterBy]
```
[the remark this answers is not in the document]
tamtamirr: https://github.com/salto-io/salto/pull/4491/files#r1234917246

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -186,7 +187,8 @@ type CrudFnArgs = {
   client: SalesforceClient
 }
 
-export type CrudFn = (fnArgs: CrudFnArgs) => Promise<InstanceAndResult[]>
+export type CrudFn = (fnArgs: CrudFnArgs)
+    => Promise<{results: InstanceAndResult[]; instancesToUpdate: InstanceElement[]}>
```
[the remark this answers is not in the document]
tamtamirr: **CrudFn** is supposed to be generic. **deployedInstancesResults** is specific for deploy operation only.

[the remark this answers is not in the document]
amitgoldarb: Isn't CrudFn only in the context of deployed elements?

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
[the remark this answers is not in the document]
amitgoldarb: This is creating a third call to the crudFn which I think we don't need. If we can think of no examples, I vote for removing this.

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -335,6 +333,13 @@ const deployAddInstances = async (
     const recordValuesWithoutNulls = cloneWithoutNulls(recordValues)
     return computeSaltoIdHash(recordValuesWithoutNulls)
   }
+  // Replacing self-reference field values with the resolved instances that will later contain the Record Ids
+  const instanceByElemId = _.keyBy(instances, instance => instance.elemID.getFullName())
+  await awu(instances).forEach(instance => {
+    instance.value = _.mapValues(instance.value, val => (isInstanceElement(val)
+      ? instanceByElemId[val.elemID.getFullName()] ?? val
+      : val))
+  })
```
tomsellek, line 342: Shouldn't this be recursive? Can there not be a case where the references are inside a list/map?

packages/salesforce-adapter/src/transformers/transformer.ts
```diff
@@ -941,31 +941,52 @@ const transformCompoundValues = async (
...
   instance: InstanceElement,
   fieldAnnotationToFilterBy: string,
-): Promise<SalesforceRecord> => {
+): Promise<{record: SalesforceRecord; needsUpdate: boolean}> => {
+  let needsUpdate = false
   const instanceType = await instance.getType()
   const valsWithNulls = {
     ..._.mapValues(instanceType.fields, () => null),
-    ...instance.value,
+    ..._.mapValues(instance.value, val => {
```
tomsellek, line 949: Does this need to be recursive?

packages/salesforce-adapter/test/custom_object_instances.crud.test.ts
```diff
@@ -513,56 +518,149 @@ describe('Custom Object Instances CRUD', () => {
...
+              expect(connection.bulk.load).toHaveBeenCalledWith(
+                'TestType__c', 'insert', expect.anything(), [
+                  { Id: undefined, Name: 'firstInstance', Number__c: 1, TestType__c: null },
+                  { Id: undefined, Name: 'secondInstance', Number__c: 1, TestType__c: null },
+                  { Id: undefined, Name: 'instanceWithoutRef', Number__c: 1, TestType__c: null },
+                ]
+              )
+              expect(connection.bulk.load).toHaveBeenCalledWith(
+                'TestType__c', 'update', expect.anything(), [
+                  { Id: 'newId0', Name: 'firstInstance', Number__c: 1, TestType__c: 'newId1' },
```
tomsellek, line 659: Silly question, but where did the ref from `newId0` to `newId1` come from?

tomsellek (commented):
Had a couple of questions

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
tamtamirr: No, Salesforce Records are plain, so recursive support is redundant. List of Lookups is not a valid value for Salesforce Records as-well

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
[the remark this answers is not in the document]
tamtamirr: v

packages/salesforce-adapter/test/custom_object_instances.crud.test.ts
tamtamirr: From the mock :)

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -360,15 +366,29 @@ const deployAddInstances = async (
...
-  const allSuccessInstances = [...successInsertInstances, ...successUpdateInstances]
+  const secondUpdateInstances = insertInstancesToUpdate.concat(updateInstancesToUpdate)
+  const {
+    successInstances: successSecondUpdateInstances,
+    errorInstances: errorSecondUpdateInstances,
+  } = await retryFlow(
+    updateInstances,
+    { typeName: await apiName(type), instances: secondUpdateInstances, client },
+    client.dataRetry.maxAttempts
+  )
```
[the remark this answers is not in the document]
tamtamirr: @amitgoldarb After taking a deeper look into the flow, this is actually required. I'll try to explain in TLDR but we can also take this f2f :)
- In salto, the Data Record instance is added, but when querying the service, we notice it already exists (By formatting it's Elem ID from the returned Record).
- The Record can reference another instance from the same group, and can be referenced by others. We only set it's Internal Id [here](https://github.com/salto-io/salto/blob/6048650304de2920ed60dbd388d82ef4d429a0c2/packages/salesforce-adapter/src/custom_object_instances_deploy.ts#L366). The instance/s that reference this Record will have to be updated with it's internalId in another iteration.

cc. @ori-moisis please validate me here 🙏

packages/salesforce-adapter/src/transformers/transformer.ts
```diff
@@ -946,7 +946,16 @@ const toRecord = async (
   const instanceType = await instance.getType()
   const values = {
     ...withNulls ? _.mapValues(instanceType.fields, () => null) : {},
-    ...instance.value,
+    ..._.mapValues(instance.value, val => {
+      if (isInstanceElement(val)) {
+        const referencedRecordId = val.value[CUSTOM_OBJECT_ID_FIELD]
+        if (referencedRecordId === undefined) {
+          return null
+        }
```
amitgoldarb, line 954: This seems like a change in the current behaviour, meaning in this "wierd" case we are sending null to the API instead of something else. Can we better explain when we will enter this if?

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -23,7 +23,7 @@ import {
   TypeElement,
   SaltoElementError,
   SaltoError,
-  SeverityLevel,
+  SeverityLevel, isInstanceElement,
```
amitgoldarb, line 26: Style comment - this is more consistent with current import mechanism
```suggestion
  SeverityLevel,
  isInstanceElement,
```

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -319,6 +319,15 @@ const deployAddInstances = async (
...
 ): Promise<DeployResult> => {
+  const instancesToUpdate = instances
+    .filter(instance => Object.values(instance.value).some(isInstanceElement))
+  // Replacing self-reference field values with the resolved instances that will later contain the Record Ids
+  const instanceByElemId = _.keyBy(instances, instance => instance.elemID.getFullName())
+  await awu(instances).forEach(instance => {
+    instance.value = _.mapValues(instance.value, val => (isInstanceElement(val)
+      ? instanceByElemId[val.elemID.getFullName()] ?? val
+      : val))
+  })
```
amitgoldarb, line 330: Reading this one again, I'm a bit confused by reading this, if the value is an InstanceElement, what are we replacing it with? What is the difference between the value we have  before this mapValue and after this mapValue?
These will later contain the recordId's because this is mutable?

amitgoldarb (commented):
Looking good. A few small comments just for my understanding

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
tamtamirr: Yes. We are setting the **resolvedInstances** here, instead of the unresolved instances so they later contain the internal Ids

packages/salesforce-adapter/src/transformers/transformer.ts
amitgoldarb: Nvm, I understand this now.

amitgoldarb (approved):
I think this looks good. Let's wait for another approval though before merging this.

benisraeladi (approved):

packages/salesforce-adapter/src/group_changes.ts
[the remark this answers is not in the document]
tamtamirr: I actually prefer leaving it like this, to be explicit

PR #4493, opened 2023-06-13
SALTO-4333: Added group back to e2e

alonstern:
This reverts commit 5e5826df45a4924cd47edbfce60c0c956838c112.

We removed the group from the e2e due to a Jira bug, now it seems to work

---
_Release Notes_: 
None

---
_User Notifications_: 
None


netama:
@alon we can link this and the original to SALTO-4333, right?

IdoZakk (approved):
💯

PR #4498, opened 2023-06-13
SALTO-4217: Change Jira adapter to return SaltoElementErrors

alonstern:
Change Jira adapter to return SaltoElementErrors

---
_Release Notes_: 
None

---
_User Notifications_: 
None


IdoZakk (approved):
:great:

PR #4500, opened 2023-06-14
SALTO-4338 Support adding alias to ObjectTypes

dantal4:
Today the add-alias utility in adapter-components supports only instances. It should support adding aliases to types too, because of the custom record types in netsuite.

---

---
_Release Notes_: 
Adapter Components:
- Support adding alias to ObjectTypes

---
_User Notifications_: 
None


edenhassid (approved):
looks good! 🔥 😎

PR #4504, opened 2023-06-14
SLTO-4339 - Function to check for a reference expression that is not missing

idoser96:
Added a function that casts for a reference expression only if its a non missing reference expression

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/references/missing_references.ts
+++ b/packages/adapter-components/src/references/missing_references.ts
@@ -13,7 +13,14 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
-import { Element, ElemID, InstanceElement, ObjectType } from '@salto-io/adapter-api'
+import {
+  Element,
+  ElemID,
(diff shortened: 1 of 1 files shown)


packages/adapter-components/src/references/missing_references.ts
```diff
@@ -35,3 +42,9 @@ export const createMissingInstance = (
     { [MISSING_ANNOTATION]: true },
   )
 )
+
+/** Check if a value is a reference expression that is not missing reference */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const isPresentReferenceExpression = (value: any): value is ReferenceExpression => (
```
alonstern, line 48: @idoamit8 just added a very similar function https://github.com/salto-io/salto/pull/4468/files#diff-13330c10c6dc2ead508102517c99ba806b989e913ecfe971b97ebb040e547ea9R1170

PR #4505, opened 2023-06-15
SALTO-4346 - Salesforce: Use retrieve api for profiles

tomsellek:
Use the `metadata retrieve` API to fetch profiles. The `retrieve` flow in our code behaves better for longer requests (which profiles, being fairly large, tend to generate), and the `retrieve` API returns more accurate information.

---
The Salesforce `retrieve` API behaves in a unique way for profiles - it only returns those parts of the profile that are relevant to the *other instances in the same request* (which we call **Context Types**). For this reason, we must ensure the entire list of profiles is sent with every chunk of objects we send, which leads to most of the complexity of this PR.
A couple of points of interest:
* When a single request ends up being too large, we split it into two requests, each of which has all of the profiles but only part of the Context Types. This means that we'll get partial information about the profile from every chunk it's a part of (only the part that is relevant to the Context Types that are in said chunk). The result will be multiple instances of the same profile that we must merge together.
* There can be an edge case where even after we created minuscule chunks, the profiles themselves are still too large and exceed the maximum amount of data that can be returned in a single `retrieve` request (~39MB of compressed XML). This is assumed to be very very unlikely (profiles are typically <500KB), but if we do hit this case we will fail the fetch.
* `typesToSkip` -  Certain types are fetched (and handled in a special way) by filters, and thus even though we must add them to our retrieve requests, we must discard the data we receive so that the filters can do their thing.

WS Diff is nil.

---
_Release Notes_: 
Salesforce: Profile instances now contain more accurate data and will not include data about excluded types.

---
_User Notifications_: 
Salesforce: data in Profile records may be removed if it refers to objects that are excluded from fetch.


packages/salesforce-adapter/src/fetch.ts
```diff
@@ -411,6 +434,71 @@ export const retrieveMetadataInstances = async ({
...
+    profile: FileProperties[]
+    profileContext: FileProperties[]
+    other: FileProperties[]
+  } => {
+    /* Salesforce returns the parts of the profile that apply to the instances that are requested along with the
+    profile, as described in https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_profile.htm
+    The following list of types that affect the contents of retrieved profile instances was discovered by trial and
+    error.
+    */
+    const profileContextTypes = new Set(['ApexClass',
```
amitgoldarb, line 449: Nit: Should this be a hard coded list that can be reused in the METADATA_TYPES_TO_RETRIEVE?

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -411,6 +434,71 @@ export const retrieveMetadataInstances = async ({
...
+        },
+      ),
+    }
+  }
+
+  const retrieveProfilesWithContextTypes = async (
+    profileFileProps: ReadonlyArray<FileProperties>,
+    contextFileProps: ReadonlyArray<FileProperties>,
+  ): Promise<Array<InstanceElement>> => {
+    const allInstances = await Promise.all(
```
amitgoldarb, line 482: Should we do this withLimitedConcurrency? I think we are throttled on a lower level, just a thought.

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -420,14 +508,23 @@ export const retrieveMetadataInstances = async ({
...
 
-  const instances = await Promise.all(
-    _.chunk(filesToRetrieve, maxItemsInRetrieveRequest)
+  const categorizedFileProps = categorizeFiles(filesToRetrieve)
+
+  const profileAndContextInstances = await retrieveProfilesWithContextTypes(categorizedFileProps.profile,
+    categorizedFileProps.profileContext)
+
+  const otherInstancesChunks = await Promise.all(
+    _.chunk(categorizedFileProps.other, maxItemsInRetrieveRequest)
```
amitgoldarb, line 517: This is potentially wasteful right? We might be able to add emailTemplates etc to the request above right? Also what if profiles is excluded, do we still need to chunk the requests this way? How does this affect performance?

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -556,6 +563,7 @@ export default class SalesforceAdapter implements AdapterOperations {
         metadataQuery: this.fetchProfile.metadataQuery,
         maxItemsInRetrieveRequest: this.maxItemsInRetrieveRequest,
         addNamespacePrefixToFullName: this.fetchProfile.addNamespacePrefixToFullName,
+        typesToSkip: new Set(this.metadataTypesOfInstancesFetchedInFilters),
```
amitgoldarb, line 566: Can you elaborate a bit what is this change?

amitgoldarb (commented):
Looks great!
Added a few thoughts/comments

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -420,14 +508,23 @@ export const retrieveMetadataInstances = async ({
...
       .filter(filesChunk => filesChunk.length > 0)
       .map(filesChunk => retrieveInstances(filesChunk))
   )
 
+  const allInstances = profileAndContextInstances
+    .concat(_.flatten(otherInstancesChunks))
+    .filter(instance => !typesToSkip.has(instance.elemID.typeName))
```
amitgoldarb, line 524: Can we add a comment here that we might be fetching information solely for profile purposes? I wonder if we should filter this out after the profileAndContext retreive, it might make a bit more sense.

packages/salesforce-adapter/test/fetch.test.ts
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  beforeEach(async () => {
+    ({ connection, client } = createMockClient())
+  })
+
+  afterEach(() => {
+    jest.resetAllMocks()
+    jest.restoreAllMocks()
+  })
```
tamtamirr, line 114: redundant, mocks are already cleared in the OSS by default after each test.

packages/salesforce-adapter/test/fetch.test.ts
```diff
@@ -0,0 +1,212 @@
...
+          typesToSkip: new Set(),
+        }
+      )
+      ).elements
+    })
+    it('should fetch the correct instances', () => {
+      expect(elements).toHaveLength(testData.length)
+      expect(elements).toIncludeAllPartialMembers(
+        testData
+          .map(({ type, instanceName }) => ({ elemID: type.elemID.createNestedID('instance', instanceName) }))
```
tamtamirr, line 191: I'm not a fan of this `expect`, as this is not explicit and requires more logic. This makes me think the **describe each** here is redundant and too generic. I'd prefer each of the **each cases** to be it's own explicit **describe** with explicit **expects** and not calculations.

packages/salesforce-adapter/test/fetch.test.ts
```diff
@@ -0,0 +1,212 @@
...
+      )
+      const customObjectTypes = testData
+        .filter(({ type }) => type.elemID.typeName === 'CustomObject')
```
tamtamirr, line 194:
```suggestion
        .filter(({ type }) => await safeApiName(type) === CUSTOM_OBJECT)
```

packages/salesforce-adapter/test/fetch.test.ts
```diff
@@ -0,0 +1,212 @@
...
+        .map(({ instanceName }) => instanceName)
+      elements
+        .filter(instance => instance.elemID.typeName === PROFILE_METADATA_TYPE)
```
tamtamirr, line 197:
```suggestion
        .filter(isInstanceOfType(PROFILE_METADATA_TYPE))
```

packages/salesforce-adapter/test/fetch.test.ts
```diff
@@ -0,0 +1,212 @@
...
+          expect(profileInstance.value.fieldPermissions).not.toBeEmpty()
+          profileInstance.value.fieldPermissions.forEach((perm: {field: string}) => {
+            expect(customObjectTypes).toInclude(perm.field.split('.')[0])
+          })
+
+          // ensure there are no duplications (we know we only add one field permission per object in the mock)
+          const objectsInProfile = profileInstance.value.fieldPermissions
+            .map((perm: {field: string}) => perm.field.split('.')[0])
+          expect(_.sortBy(objectsInProfile)).toEqual(_.sortBy(customObjectTypes))
+        })
```
tamtamirr, line 209: This one is complex due to the fact this test is generic for the various describes. I would prefer explicit assertion on the expected value of the FLS rather than that. We can also take that F2F :)

tamtamirr (commented):
Looking good! Added some comments

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -378,6 +389,11 @@ export const retrieveMetadataInstances = async ({
 
     if (result.errorStatusCode === RETRIEVE_SIZE_LIMIT_ERROR) {
       if (fileProps.length <= 1) {
+        if (filePropsToSendWithEveryChunk.length > 0) {
```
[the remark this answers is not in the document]
tomsellek: If the profiles themselves, with only a single context type, are too large for a single request then we give up.

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -229,16 +229,23 @@ const METADATA_TO_RETRIEVE = [
   'AuraDefinitionBundle', // Has several fields with base64Binary encoded content
   'Certificate', // contains encoded zip content
   'ContentAsset', // contains encoded zip content
+  'CustomApplication',
```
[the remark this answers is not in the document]
tomsellek: Ack.

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -359,11 +362,19 @@ export const retrieveMetadataInstances = async ({
...
+    const result = instances[0]
+    result.value = _.merge({}, ...instances.map(instance => instance.value))
+    return result
+  }
+
   const retrieveInstances = async (
     fileProps: ReadonlyArray<FileProperties>,
+    filePropsToSendWithEveryChunk: ReadonlyArray<FileProperties> = [],
   ): Promise<InstanceElement[]> => {
     // Salesforce quirk - folder instances are listed under their content's type in the manifest
```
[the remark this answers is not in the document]
tomsellek: Ack

packages/salesforce-adapter/src/fetch.ts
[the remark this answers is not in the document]
tomsellek: Not sure what you mean? The only new part here is the case where the response exceeded the max size, we can't reduce the chunk size any further and there are still profiles waiting to be fetched. In this case we're not adding to the skip list (I don't think adding `Profile` to the skip list makes sense?) but rather fail the fetch.
I had some logic here before that did second-level chunking on the profiles themselves, but @ori-moisis says it's such an unlikely edge case that we may as well not bother.

packages/salesforce-adapter/src/fetch.ts
tomsellek: I'm honestly not sure. @tamtamirr WDYT?

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -411,6 +434,71 @@ export const retrieveMetadataInstances = async ({
...
+          }
+          if (profileContextTypes.has(fileProps.type)) {
+            return 'profileContext'
+          }
+
+          return 'other'
+        },
+      ),
+    }
+  }
```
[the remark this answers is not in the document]
tomsellek: You mean grouping by fileProps.type and doing mapKeys from type to `profile`/`profileContext`/`other`? If that's what you mean, I didn't do it because `_.mapKeys` doesn't know how to handle the case where multiple original keys map to a single new key. Or did you mean something else?

packages/salesforce-adapter/src/fetch.ts
tomsellek: Yeah, I think we're already throttled by `DEFAULT_MAX_CONCURRENT_API_REQUESTS` at the client level, and I think the client is the one who should know about this sort of limits.

packages/salesforce-adapter/src/fetch.ts
tomsellek: Yes, we will have two requests instead of one. I can merge the two for sure.

packages/salesforce-adapter/src/fetch.ts
tomsellek: Added a comment next to the definition of typesToSkip.

packages/salesforce-adapter/test/fetch.test.ts
tomsellek: I see what you're saying but also having each case be a `describe` of its own is going to be *so much* redundant code. Let's do a F2F like you suggested below?

packages/salesforce-adapter/test/fetch.test.ts
tomsellek: Ack

amitgoldarb (approved):
This looks good to me!
Approved with a few small nits.

packages/salesforce-adapter/test/fetch.test.ts
tamtamirr: @tomsellek This is inconsistent with the way we treat **Salesforce Elements** in the adapter. Always use **safeApiName** to identify the unique identifier of a Salesforce Element (Wether it's an ObjectType or InstanceElement)

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -352,7 +362,12 @@ export default class SalesforceAdapter implements AdapterOperations {
     const fetchProfile = buildFetchProfile(config.fetch ?? {})
     this.fetchProfile = fetchProfile
     if (!this.fetchProfile.isFeatureEnabled('fetchCustomObjectUsingRetrieveApi')) {
-      _.pull(this.metadataToRetrieve, CUSTOM_OBJECT)
+      // We have to fetch custom objects using retrieve in order to be able to fetch the field-level permissions
+      // in profiles. If custom objects are fetched via the read API, we have to fetch profiles using that API too.
+      _.pull(this.metadataToRetrieve, CUSTOM_OBJECT, PROFILE_METADATA_TYPE)
+    }
+    if (!this.fetchProfile.isFeatureEnabled('fetchProfilesUsingRetrieveApi')) {
```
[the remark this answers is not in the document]
tomsellek: No, but I agree it's confusing: If `!fetchProfilesUsingRetrieveApi` then remove (==pull) the Profile type from the list of types we use `retrieve` to fetch.

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -359,11 +367,19 @@ export const retrieveMetadataInstances = async ({
   const typesWithMetaFile = await getTypesWithMetaFile(types)
   const typesWithContent = await getTypesWithContent(types)
 
+  const mergeProfileInstances = (instances: ReadonlyArray<InstanceElement>): InstanceElement => {
```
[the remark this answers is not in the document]
tomsellek: Ack.

packages/salesforce-adapter/test/fetch.test.ts
tamtamirr, line 1: Why did we create a new test file for fetch? We already have UTs on fetch
 
1. Move your tests to [adapter.discover.test.ts](https://github.com/salto-io/salto/blob/865f87f77b4ed93c710b3f6a3709d5508e335fef/packages/salesforce-adapter/test/adapter.discover.test.ts)
2. Test fetch using **adapter.fetch** and not calling **retrieveMetadataInstances** directly.

tamtamirr (commented):
Code looks good :) Please move your UTs as suggested

PR #4506, opened 2023-06-15
SALTO-4301 update static file path

edenhassid:
update static file path.
until now, static files whose path changed but content didn't would remain in their old location, after this PR, static files whose path changed will be created in their new location and the old files will be deleted.

---

---
_Release Notes_: 
core:
* static file path changes will be referred to as a change in fetch

---
_User Notifications_: 
* some static files might move location

### DIFF
--- a/packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
+++ b/packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
@@ -586,22 +586,20 @@ const logNaclFileUpdateErrorContext = (
 
 // Returns a list of all static files that existed in the changes 'before' and doesn't exist in the 'after'
 export const getDanglingStaticFiles = (fileChanges: DetailedChange[]): StaticFile[] => {
-  const modificationAndAdditionAfterFilesIds = new Set<string>(
+  // Using filepath is currently enough because all implementations of static files have unique file paths
+  // The only exception is 'buildHistoryStateStaticFilesSource' but it doesn't support deletion at the moment
+  const afterFilePaths = new Set<string>(
     fileChanges
       .filter(isAdditionOrModificationChange)
       .map(change => change.data.after)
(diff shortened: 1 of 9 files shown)


netama (approved):
approving based on the original reviews in https://github.com/salto-io/salto/pull/4484 and https://github.com/salto-io/salto/pull/4496

packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
```diff
@@ -559,6 +561,22 @@ const logNaclFileUpdateErrorContext = (
...
+  const modifications = fileChanges.filter(isModificationChange)
+  const modificationAfterFilesIds = new Set<string>(
+    modifications.map(change => change.data.after).map(getNestedStaticFiles).flat().map(file => file.filepath)
+  )
+  const modificationRemovedFiles = modifications.map(change => change.data.before).map(getNestedStaticFiles).flat()
+    // Using filepath is currently enough because all implementations of static files have unique file paths
+    // The only exception is 'buildHistoryStateStaticFilesSource' but it doesn't support deletion at the moment
+    .filter(file => !modificationAfterFilesIds.has(file.filepath))
+
+  return removalFiles.concat(modificationRemovedFiles)
```
ori-moisis, line 577: small issue with this - if the file moved to a different key
e.g:
before:
```
dummy.type inst {
  bla = file('my_file_path.txt')
}
```
after:
```
dummy.type inst {
  foo = file('my_file_path.txt')
}
```
we will get two separate changes, one addition and one removal.
in that case, we will delete the file but we really shouldn't

I think this is a bug in the current code as well...

we need to check all the files, both in before and in after.
maybe the alternative would be to call this before calling `updateNaclFileData`, but that would be more fragile

ori-moisis (approved):

PR #4508, opened 2023-06-15
SALTO-4275: Fix Automation Scopes

IdoZakk:
Atlassian recently changed the structure of automation scopes, removing the `projects` object. It caused all new automation to be deployed as Global

---
_Release Notes_: 
Jira Adapter:
* Fixed a bug causing all automation to be deployed with global scope

---
_User Notifications_: 
Jira Adapter:
* Recently created automations with a non-global scope will be added scope in the projects field

### DIFF
--- a/packages/jira-adapter/src/filters/automation/automation_structure.ts
+++ b/packages/jira-adapter/src/filters/automation/automation_structure.ts
@@ -18,7 +18,7 @@ import Joi from 'joi'
 import { logger } from '@salto-io/logging'
 import { InstanceElement, isInstanceElement, Values, getChangeData,
   Change, isInstanceChange } from '@salto-io/adapter-api'
-import { transformElement, applyFunctionToChangeData, resolveValues, restoreChangeElement, safeJsonStringify, restoreValues } from '@salto-io/adapter-utils'
+import { transformElement, applyFunctionToChangeData, resolveValues, restoreChangeElement, safeJsonStringify, restoreValues, createSchemeGuard } from '@salto-io/adapter-utils'
 import { elements as elementUtils } from '@salto-io/adapter-components'
 import { collections } from '@salto-io/lowerdash'
 import { AUTOMATION_TYPE, AUTOMATION_COMPONENT_TYPE, AUTOMATION_COMPONENT_VALUE_TYPE, AUTOMATION_OPERATION } from '../../constants'
@@ -57,6 +57,10 @@ type CompareFieldValueObject = {
   }
 }
 
+type RuleScope = {
+  resources: string[]
+}
+
 const LINK_TYPE_SCHEME = Joi.object({
   linkType: Joi.string().required(),
   linkTypeDirection: Joi.string().required(),
@@ -75,6 +79,10 @@ const COMPARE_FIELD_VALUE_SCHEME = Joi.object({
   }).unknown(true).required(),
 }).unknown(true).required()
 
+const RULE_SCOPE_SCHEME = Joi.object({
+  resources: Joi.array().items(Joi.string()).required(),
+}).unknown(true).required()
+
 const isLinkTypeObject = (value: unknown): value is LinkTypeObject => {
   const { error } = LINK_TYPE_SCHEME.validate(value)
   return error === undefined
@@ -90,6 +98,8 @@ const isCompareFieldValueObject = (value: unknown): value is CompareFieldValueOb
   return error === undefined
 }
 
+const isRuleScope = createSchemeGuard<RuleScope>(RULE_SCOPE_SCHEME, 'Wrong rule scope')
+
 const KEYS_TO_REMOVE = [
   'clientKey',
   'updated',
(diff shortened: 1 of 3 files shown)


packages/jira-adapter/src/filters/automation/automation_structure.ts
```diff
@@ -265,7 +288,31 @@ const revertCompareFieldValueStructure = async (instance: InstanceElement): Prom
   })).value
 }
 
-const filter: FilterCreator = () => {
+const isOldFormatScope = (projects: Values): boolean =>
+  Array.isArray(projects) && projects.length > 0
```
alonstern, line 292: I wonder if for existing automation, the `projects` is updated when the automation is edited. If it is not we should always use the rule scope

packages/jira-adapter/src/filters/automation/automation_structure.ts
```diff
@@ -265,7 +288,31 @@ const revertCompareFieldValueStructure = async (instance: InstanceElement): Prom
...
+      instance.value.projects.push({ projectId: resource.match(PROJECT_SCOPE_REGEX)?.[1] })
+    } else if (resource.match(PROJECT_TYPE_SCOPE_REGEX)) {
+      const projectTypeKey = findKeyInRecordObject(PROJECT_TYPE_TO_RESOURCE_TYPE, resource.match(PROJECT_TYPE_SCOPE_REGEX)?.[1] ?? '')
+      if (projectTypeKey === undefined) {
+        log.error(`Failed to convert automation project type: ${resource.match(PROJECT_TYPE_SCOPE_REGEX)?.[1]}`)
+      }
+      instance.value.projects.push({ projectTypeKey })
+    } else if (!resource.match(GLOBAL_SCOPE_REGEX)) {
+      log.error(`Failed to convert automation rule scope, found unknown pattern: ${resource}`)
+    }
```
alonstern, line 311: nit: extract this to a function `getProjectScope = (ruleScopeResource: string): { projected?: string, projectTypeKey? string } | undefined` and do
```
instance.value.projects = ruleScope.resources.map(getProjectScope).filter(values.isDefined)
```

packages/jira-adapter/src/filters/automation/automation_structure.ts
```diff
@@ -265,7 +288,31 @@ const revertCompareFieldValueStructure = async (instance: InstanceElement): Prom
...
+
+const convertRuleScopeToProjects = (instance: InstanceElement): void => {
+  const { ruleScope } = instance.value
+  if (!isRuleScope(ruleScope) || isOldFormatScope(instance.value.projects)) {
+    return
+  }
+  instance.value.projects = []
+  ruleScope.resources.forEach((resource: string) => {
+    if (resource.match(PROJECT_SCOPE_REGEX)) {
+      instance.value.projects.push({ projectId: resource.match(PROJECT_SCOPE_REGEX)?.[1] })
```
alonstern, line 302: I think you can avoid the `?` by doing 
```suggestion
    const projectScopeMatch = resource.match(PROJECT_SCOPE_REGEX)
    if (projectScopeMatch) {
      instance.value.projects.push({ projectId: projectScopeMatch[1] })
```

same for projectTypeKey

alonstern (commented):
Looks good 👌

packages/jira-adapter/src/filters/automation/automation_structure.ts
IdoZakk: Good catch!

packages/jira-adapter/src/filters/automation/automation_structure.ts
IdoZakk: I was thinking about it, but with the if-else it required doing all 3 matches before the if-else and it looked less clean to me

alonstern: I think if you extract it to a function like in https://github.com/salto-io/salto/pull/4508#discussion_r1230542244 you won't need the if-else and you can do
```
const match1 = ..
if (match1) {
  return ...
}
const match2 = ..
if (match2) {
  return ...
}
...
```

packages/jira-adapter/src/filters/automation/automation_structure.ts
```diff
@@ -288,28 +289,31 @@ const revertCompareFieldValueStructure = async (instance: InstanceElement): Prom
...
+const getScope = (resource: string): { projectId?: string; projectTypeKey?: string } | undefined => {
+  const projectScope = resource.match(PROJECT_SCOPE_REGEX)
+  if (projectScope) {
+    return { projectId: projectScope[1] }
+  }
+  const projectTypeScope = resource.match(PROJECT_TYPE_SCOPE_REGEX)
+  if (projectTypeScope) {
+    const projectTypeKey = findKeyInRecordObject(PROJECT_TYPE_TO_RESOURCE_TYPE, projectTypeScope[1])
+    if (projectTypeKey === undefined) {
+      log.error(`Failed to convert automation project type: ${resource.match(PROJECT_TYPE_SCOPE_REGEX)?.[1]}`)
```
alonstern, line 301: you can use `projectTypeScope` also here

alonstern, line 301: I think you should return here `undefined`

alonstern (approved):
👌

packages/jira-adapter/src/filters/automation/automation_structure.ts
IdoZakk: The problem is that empty projects means global scope, I thought it is better to have something that will prevent deployment

alonstern: What scares me is that the error the user will get will not be very intuitive and I am not sure we will remember this is the reason for it if it will happen, but whatever you think

PR #4509, opened 2023-06-15
SALTO-4319: cannot create two articles that reference each other

edenhassid:
fix bug that prevented the deploy of two articles that reference each other

---

---
_Release Notes_: 
zendesk:
* fix bug that prevented the deploy of two articles that reference each other

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/filters/article/article.ts
+++ b/packages/zendesk-adapter/src/filters/article/article.ts
@@ -305,8 +305,7 @@ const filterCreator: FilterCreator = ({ config, client, elementsSource, brandIdT
               (part: ReferenceExpression) => (
                 part.elemID.typeName === ARTICLE_ATTACHMENT_TYPE_NAME
                   ? addedArticleAttachments
-                    .find(attachment => attachment.elemID.isEqual(part.elemID))
-                    ?.value.id.toString()
+                    .find(attachment => attachment.elemID.isEqual(part.elemID))?.value.id.toString()
                   : prepRef(part)
               ),
             )
--- a/packages/zendesk-adapter/src/filters/article/article.ts
+++ b/packages/zendesk-adapter/src/filters/article/article.ts
@@ -72,6 +72,10 @@ const isAttachmentWithId = createSchemeGuard<AttachmentWithId>(
   EXPECTED_ATTACHMENT_SCHEMA, 'Received an invalid value for attachment id'
 )
 
+export const getGenericTitle = (title: string): string => `salto generic title for ${title}`
+
+export const getGenericBody = (title: string): string => `salto generic body for ${title}`
+
 const addTitleAndBodyValues = async (change: Change<InstanceElement>): Promise<void> => {
   const resolvedChange = await resolveChangeElement(change, lookupFunc)
   const currentLocale = getChangeData(resolvedChange).value.source_locale
@@ -79,8 +83,8 @@ const addTitleAndBodyValues = async (change: Change<InstanceElement>): Promise<v
     .filter(isTranslation)
     .find((tran: TranslationType) => tran.locale?.locale === currentLocale)
   if (translation !== undefined) {
-    getChangeData(change).value.title = `salto generic title for ${translation.title}`
-    getChangeData(change).value.body = 'salto generic body '
+    getChangeData(change).value.title = getGenericTitle(translation.title)
+    getChangeData(change).value.body = getGenericBody(translation.title)
   }
 }
 
--- a/packages/zendesk-adapter/src/filters/article/article.ts
+++ b/packages/zendesk-adapter/src/filters/article/article.ts
@@ -25,7 +25,6 @@ import {
 import {
   createSchemeGuard,
   getParents,
-  replaceTemplatesWithValues,
   resolveChangeElement,
 } from '@salto-io/adapter-utils'
 import Joi from 'joi'
@@ -41,7 +40,6 @@ import {
 import { addRemovalChangesId, isTranslation } from '../guide_section_and_category'
 import { lookupFunc } from '../field_references'
 import { removeTitleAndBody } from '../guide_fetch_article_section_and_category'
-import { prepRef } from './article_body'
 import ZendeskClient from '../../client/client'
 import {
   createUnassociatedAttachment,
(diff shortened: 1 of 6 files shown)


packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -74,15 +72,15 @@ const isAttachmentWithId = createSchemeGuard<AttachmentWithId>(
   EXPECTED_ATTACHMENT_SCHEMA, 'Received an invalid value for attachment id'
 )
 
-const addTranslationValues = async (change: Change<InstanceElement>): Promise<void> => {
+const addTitleAndBodyValues = async (change: Change<InstanceElement>): Promise<void> => {
```
netama, line 75: 1. nit: maybe something like `addPlaceholderTitleAndBodyValues`? 
2. I know you're planning to, but we should eventually update the text :)

packages/zendesk-adapter/src/filters/article/utils.ts
```diff
@@ -186,7 +186,7 @@ export const createUnassociatedAttachment = async (
       log.error(`Received an invalid response from Zendesk API, ${safeJsonStringify(res.data, undefined, 2).slice(0, RESULT_MAXIMUM_OUTPUT_SIZE)}. Not adding article attachments`)
       return
     }
-    const createdAttachment = [res.data.article_attachment]
+    const createdAttachment = [res.data.article_attachment] // why is this an array??
```
netama, line 189: :+1: doesn't seem like it's needed

packages/zendesk-adapter/src/filters/guide_translation.ts
```diff
@@ -34,7 +38,7 @@ const isDefaultTranslationAddition = (change: Change<InstanceElement>): boolean
   const data = getChangeData(change)
   const currentLocale = data.value.locale
   const parents = getParents(data) // the parent is not a reference expression
-  return parents.some(parent => parent.source_locale?.value.value.locale === currentLocale)
+  return parents.some(parent => parent.source_locale?.value.value?.locale === currentLocale)
```
netama, line 41: do we have a guarantee that the source_locale value will be resolved at this point? do we need to potentially complete it from the element source?

packages/zendesk-adapter/src/filters/guide_translation.ts
```diff
@@ -54,16 +58,43 @@ const needToOmit = (change: Change<InstanceElement>): boolean =>
  * this filter marks these translations as successfully deployed without actually deploying them.
  * The rest of the translations will be deployed in the default deploy filter.
  */
-const filterCreator: FilterCreator = () => ({
+const filterCreator: FilterCreator = ({ config, client }) => ({
```
netama, line 61: I think we should change the filter description?

packages/zendesk-adapter/src/filters/guide_translation.ts
```diff
@@ -54,16 +58,43 @@ const needToOmit = (change: Change<InstanceElement>): boolean =>
...
+    )
+
+    // creating modification changes from the addition of default article translations
+    const defaultArticleTranslationModificationChanges = defaultArticleTranslationAdditionChanges
+      .map(getChangeData)
+      .map(instance =>
+        toChange({ before: instance, after: instance }))
+
+    // deploy the created modification changes
+    const articleTranslationDeployResult = await deployChanges(
```
netama, line 82: is the only difference between translation modify and create in the deployment url?

netama (commented):
POC looks good! did we also test adding attachments to the mix (so - two articles with references to each other, with inline attachments from some of the translations)? just to make sure we don't need to update the change groups to include all translations together with the article+attachments (which I would think might be needed)?

edenhassid (commented):
@netama yes this works with inline attachment in one of the translations :) currently the change groups are such that we have in one group an article and its attachments and in another group all the article_translations of all the articles together

packages/zendesk-adapter/src/filters/article/article.ts
edenhassid: do you have a preference on what the text should be?

packages/zendesk-adapter/src/filters/guide_translation.ts
edenhassid: I think it will be safer to complete it from the element source

packages/zendesk-adapter/src/filters/guide_translation.ts
edenhassid: yes

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -74,15 +72,17 @@ const isAttachmentWithId = createSchemeGuard<AttachmentWithId>(
   EXPECTED_ATTACHMENT_SCHEMA, 'Received an invalid value for attachment id'
 )
 
-const addTranslationValues = async (change: Change<InstanceElement>): Promise<void> => {
+export const getGenericTitle = (title: string): string => `salto temporary title for article with title: ${title}`
```
netama, line 75: let's actually decide on this one IRL, I'm wondering if we might want to just use the title as-is after all for the low chance of something failing in the middle + the article being published (not sure if this is a valid scenario)

packages/zendesk-adapter/src/filters/guide_translation.ts
netama: should we change this check then?

packages/zendesk-adapter/src/filters/guide_translation.ts
```diff
@@ -49,21 +53,50 @@ const needToOmit = (change: Change<InstanceElement>): boolean =>
...
   deploy: async (changes: Change<InstanceElement>[]) => {
     const [translationChangesToIgnore, leftoverChanges] = _.partition(
       changes,
       needToOmit,
     )
+    // split default article translation addition changes out of changes to ignore
+    const [defaultArticleTranslationAdditionChanges, otherTranslationChanges] = _.partition(
+      translationChangesToIgnore,
+      change => getChangeData(change).elemID.typeName === ARTICLE_TRANSLATION_TYPE_NAME
+      && isDefaultTranslationAddition(change),
```
netama, line 74: readability nit - should always indent when breaking the implementation into multiple lines
```suggestion
      change => getChangeData(change).elemID.typeName === ARTICLE_TRANSLATION_TYPE_NAME
        && isDefaultTranslationAddition(change),
```

netama (approved):
🎈

netama (approved):

PR #4510, opened 2023-06-15
SALTO-4343: Account id changes for automation groups fix

IdoZakk:
Fixed a bug in which for automation with user condition we would treat groups and roles as users in some cases, causing the automations to fail deployment
---
Will also open Noise reduction

---
_Release Notes_: 
Jira Adapter: 
* Fixed a bug that prevented deployment of automations with user conditions of "is in group"/"Is in role"

---
_User Notifications_: 
Jira Adapter:
* In automations with user condition and "is in group"/"Is in role" the reference to  the group/role will be in the `value` field instead of the `id` field


alonstern (approved):
👍

PR #4514, opened 2023-06-18
SALTO-4236: Jira Deployment performance

IdoZakk:
Improved Deployment Performace
---

The deployment performance of any instance was impacted by the workflow modification filter, as it called elementSource.list() for each change, regardless of the relevance. Also, those calls lead to many calls on RocksDB which is single-threaded.

We now call the list only for relevant changes, and cache the result
---
_Release Notes_: 
Jira Adapter:
* improve deployment performance when deploying many instances
---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/workflow/workflow_modification_filter.ts
+++ b/packages/jira-adapter/src/filters/workflow/workflow_modification_filter.ts
@@ -14,7 +14,7 @@
 * limitations under the License.
 */
 import { logger } from '@salto-io/logging'
-import { AdditionChange, CORE_ANNOTATIONS, getChangeData, InstanceElement, isInstanceChange, isModificationChange, ModificationChange, ReadOnlyElementsSource, ReferenceExpression, RemovalChange, toChange } from '@salto-io/adapter-api'
+import { AdditionChange, CORE_ANNOTATIONS, ElemID, getChangeData, InstanceElement, isInstanceChange, isModificationChange, ModificationChange, ReadOnlyElementsSource, ReferenceExpression, RemovalChange, toChange } from '@salto-io/adapter-api'
 import { collections, values } from '@salto-io/lowerdash'
 import _ from 'lodash'
 import { v4 as uuidv4 } from 'uuid'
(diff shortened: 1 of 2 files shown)


packages/jira-adapter/src/filters/workflow/workflow_modification_filter.ts
```diff
@@ -181,58 +181,84 @@ const deployWorkflowModification = async ({
...
+  let workflowSchemesPromise: Promise<ElemID[]>
+  return async (): Promise<ElemID[]> => {
+    if (ids === undefined) {
+      if (workflowSchemesPromise === undefined) {
+        workflowSchemesPromise = awu(await elementsSource.list())
+          .filter(id => id.typeName === WORKFLOW_SCHEME_TYPE_NAME)
+          .filter(id => id.idType === 'instance')
+          .toArray()
+      }
+      ids = await workflowSchemesPromise
```
alonstern, line 195: I don't think the ids is needed here. I think we can do
```suggestion
const getWorkflowSchemeIds = (elementsSource: ReadOnlyElementsSource): () => Promise<ElemID[]> => {
  let workflowSchemesPromise: Promise<ElemID[]>
  return (): Promise<ElemID[]> => {
      if (workflowSchemesPromise === undefined) {
        workflowSchemesPromise = awu(await elementsSource.list())
          .filter(id => id.typeName === WORKFLOW_SCHEME_TYPE_NAME)
          .filter(id => id.idType === 'instance')
          .toArray()
      }
      return workflowSchemesPromise
```

packages/jira-adapter/src/filters/workflow/workflow_modification_filter.ts
```diff
@@ -181,58 +181,84 @@ const deployWorkflowModification = async ({
...
+        return {
+          leftoverChanges,
+          deployResult: { appliedChanges: [], errors: [] },
+        }
+      }
+      // We want to store the schemes as RocksDb is single threaded and calling list() is expensive
+      // However, elementInstances can change so the 'get' is outside the cache
+      const workflowSchemesIds = await workflowSchemesIdsCache()
+      const workflowSchemes = await awu(workflowSchemesIds)
+        .map(id => elementsSource.get(id))
```
alonstern, line 235: Rethinking about it, I think in this specific filter, the workflow scheme won't change in a way to will hurt this filter so I think we can cache the instances themselves

alonstern (approved):
LGTM 👌 added two small comments

IdoZakk:
The changes were a lot of code, I prefer that you take another short look

alonstern (approved):
👌

PR #4523, opened 2023-06-19
SALTO-4370: Don't fail fetch for 410 errors

shir-reifenberg:
Suppress 410 https errors

---

Okta returns 410 for deprecated endpoints, this shouldn't fail fetch
Didn't excluded the type that caused it because in our accounts it works.

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/client/client.ts
+++ b/packages/okta-adapter/src/client/client.ts
@@ -82,6 +82,11 @@ export default class OktaClient extends clientUtils.AdapterHTTPClient<
         log.debug('Suppressing %d error %o for AppUserSchema', status, e)
         return { data: [], status }
       }
+      // Okta returns 410 for deprecated endpoints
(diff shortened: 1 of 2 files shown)


packages/okta-adapter/src/client/client.ts
```diff
@@ -82,6 +82,11 @@ export default class OktaClient extends clientUtils.AdapterHTTPClient<
         log.debug('Suppressing %d error %o for AppUserSchema', status, e)
         return { data: [], status }
       }
+      // Okta returns 410 for deprecated endpoints
+      if (status === 410) {
+        log.warn('Suppressing %d error %o for endpoint: %s', status, e, args.url)
```
netama, line 87: let's add an alert about these so that we don't miss them if they start happening on new endpoints

netama (approved):

PR #4529, opened 2023-06-21
SALTO-1713: SBAA Approval Rules with Custom Condition deploy

tamtamirr:
Add support for addition of **ApprovalRule** instances with **Custom Condition**.
 
---

Test Scenarios:
- I've ran a big deployment of various additions (Some Rules were set with **sbaa__ConditionsMet__c = All** and some with **Custom**) to make sure we support the old form and the dedicated group.

Three groups were deployed in that process:
1. The dedicated group I've added handling to in this PR (With ApprovalRule with Custom defined and their Approval Conditions)
2. Additions of ApprovalRule instances without **Custom** conditions Met.
3. Additions of ApprovalCondition instances.

- Modification of **ApprovalRule**
- Modification of **ApprovalCondition**
- Addition of Circular Dependencies in Product2 instances (To make sure I didn't break the previous flow)

A few notes on this PR:
- I've created a dedicated group for both **sbaa__ApprovalRule__c** and**sbaa__ApprovalCondition__c** instances addition, which is the first time we have a ChangeGroup with two different Data types.
- We run three deploys in that scenario:
1. Deploy all the **Approval Rule Instances** where we replace **sbaa__ConditionsMet__c** from **Custom** to **ALL**
2. Deploy all the **Condition Instances**
3. Deploy again (modification) with all the **Approval Rule Instances** with **Custom** conditions met.
- Whoever wants to use this feature, will have to disable [this CV](https://github.com/salto-io/salto/blob/f5b79dfb0de08b28e2892f70fe4b57d409682ce8/packages/salesforce-adapter/src/change_validators/sbaa_approval_rules_custom_condition.ts#L33). We can decide to delete it as-well.
 
---
_Release Notes_: 
Salesforce Adapter:
- Support additions of **SBAA Approval Rule** instances with **Custom** conditions met.

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/adapter.ts
```diff
@@ -445,6 +523,13 @@ export default class SalesforceAdapter implements AdapterOperations {
           errors: [{ message: 'Cannot deploy CustomObject Records as part of check-only deployment', severity: 'Error' }],
         }
       }
+      if (changeGroup.groupID === ADD_APPROVAL_RULE_AND_CONDITION_GROUP) {
```
amitgoldarb, line 526: I think this should be inside the deployCustomObjectInstancesGroup. This is to high level for this specific logic IMO.
I understand why you did this, but I think maybe we should refactor customObjectInstancesGroup so that you can reuse the critical part there without having to add this so high level.

packages/salesforce-adapter/src/group_changes.ts
```diff
@@ -14,24 +14,34 @@
...
+
 type ChangeGroupDescription = {
   groupId: string
   disjoint?: boolean
 }
 
 type ChangeIdFunction = (change: Change) => Promise<ChangeGroupDescription> | ChangeGroupDescription
 
+const getGroupId = async (change: Change<InstanceElement>): Promise<string> => {
+  const typeName = await safeApiName(await getChangeData(change).getType()) ?? 'UNKNOWN'
```
amitgoldarb, line 33: Why is this different than the original? Should it be?

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -281,6 +294,71 @@ export const UNSUPPORTED_SYSTEM_FIELDS = [
   'LastViewedDate',
 ]
 
+const deployAddApprovalRuleAndCondition = async (
```
amitgoldarb, line 297: This should be in the customObjectInstances file or possibly it's own. This file is very generic IMO and this use case is too specific.

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -281,6 +294,71 @@ export const UNSUPPORTED_SYSTEM_FIELDS = [
...
+  changes: Change<InstanceElement>[],
+  client: SalesforceClient,
+  dataManagement: DataManagement | undefined
+): Promise<DeployResult> => {
+  const approvalRuleChanges = await awu(changes)
+    .filter(isInstanceOfTypeChange(SBAA_APPROVAL_RULE))
+    .toArray()
+  const approvalConditionChanges = _.pullAll(changes, approvalRuleChanges)
```
amitgoldarb, line 305: Isn't it safer to run the same filter you have run above, like for the approvalRuleChanges?

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -281,6 +294,71 @@ export const UNSUPPORTED_SYSTEM_FIELDS = [
...
+    instance => instance.elemID.getFullName()
+  )
+  await awu(approvalConditionChanges)
+    .map(getChangeData)
+    .forEach(instance => {
+      const approvalRule = instance.value[SBAA_APPROVAL_RULE]
+      if (isInstanceElement(approvalRule)) {
+        instance.value[SBAA_APPROVAL_RULE] = approvalRuleInstancesByElemId[approvalRule.elemID.getFullName()]
+      }
+    })
```
amitgoldarb, line 318: Why is this part necessary? Can't we do this straight with the result of the first deploy?

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -281,6 +294,71 @@ export const UNSUPPORTED_SYSTEM_FIELDS = [
...
+    approvalRuleChanges,
+    change => getChangeData(change).value.sbaa__ConditionsMet__c === 'Custom'
+  )
+
+  const approvalRulesNoCustomDeployResult = await deployCustomObjectInstancesGroup(
+    rulesWithoutCustomCondition.concat(await awu(rulesWithCustomCondition)
+      .map(change => applyFunctionToChangeData(change, instance => {
+        instance.value.sbaa__ConditionsMet__c = 'All'
+        return instance
+      })).toArray()),
```
amitgoldarb, line 329: Nit: Maybe we can extract this to a function for readability.

amitgoldarb (commented):
This looks good!
I added a few comments. Nothing major TBH.
Let's wait on review of @benisraeladi / @tomsellek  as well

packages/salesforce-adapter/src/adapter.ts
tamtamirr: While it might be, sadly **lodash partition** does not support async lambdas. Didn't want to add it as part of my PR as-well to avoid complexity

packages/salesforce-adapter/src/adapter.ts
tamtamirr: This is similar to the logic in https://github.com/salto-io/salto/pull/4491. We put the actually deployed instance, that will later contain the InternalId post insertion.

packages/salesforce-adapter/src/sbaa_approval_rules_and_conditions_deploy.ts
```diff
@@ -0,0 +1,121 @@
...
+    .toArray()
+  const approvalConditionChanges = _.pullAll(changes, approvalRuleChanges)
+  log.debug('Deploying ApprovalRule instances without Custom ConditionsMet')
+  approvalRuleChanges
+    .map(getChangeData)
+    .forEach(instance => {
+      instance.value[SBAA_CONDITIONS_MET] = 'All'
+    })
+  const approvalRulesNoCustomDeployResult = await deployCustomObjectInstancesGroup(
+    approvalConditionChanges,
```
adamro, line 58: How come this is `approvalConditionChanges` and not  `approvalRuleChanges`? I am super confused

adamro (commented):
Mid way but there's one critical question before I continue so publishing the comments now

packages/salesforce-adapter/src/group_changes.ts
```diff
@@ -14,44 +14,73 @@
...
+  SBAA_CONDITIONS_MET,
+} from './constants'
 
 const { awu } = collections.asynciterable
 
-type ChangeGroupDescription = {
-  groupId: string
-  disjoint?: boolean
+const getGroupId = async (change: Change): Promise<string> => {
+  if (!isInstanceChange(change) || !(await isInstanceOfCustomObjectChange(change))) {
```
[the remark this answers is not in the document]
tamtamirr: It does, this check is generally for the typeguard.

packages/salesforce-adapter/src/sbaa_approval_rules_and_conditions_deploy.ts
```diff
@@ -0,0 +1,121 @@
...
+): Promise<DeployResult> => {
+  const approvalRuleChanges = await awu(changes)
+    .filter(isInstanceOfTypeChange(SBAA_APPROVAL_RULE))
+    .toArray()
+  const approvalConditionChanges = _.pullAll(changes, approvalRuleChanges)
+  log.debug('Deploying ApprovalRule instances without Custom ConditionsMet')
+  approvalRuleChanges
+    .map(getChangeData)
+    .forEach(instance => {
+      instance.value[SBAA_CONDITIONS_MET] = 'All'
```
adamro, line 55: Can we also verify it's 'Custom' and if not do nothing & log.warn here?

packages/salesforce-adapter/src/sbaa_approval_rules_and_conditions_deploy.ts
```diff
@@ -0,0 +1,121 @@
...
+      instance.value[SBAA_APPROVAL_RULE] = await safeApiName(deployedRule)
+    })
+  const conditionsDeployResult = await deployCustomObjectInstancesGroup(
+    deployableConditionChanges,
+    client,
+    ADD_CUSTOM_APPROVAL_RULE_AND_CONDITION_GROUP,
+    dataManagement,
+  )
+
+  log.debug('Deploying ApprovalRule instances with Custom ConditionsMet')
```
adamro, line 98: This debug text is confusing to me. It Should be `Updating the ApprovalRules back to Custom`

packages/salesforce-adapter/src/sbaa_approval_rules_and_conditions_deploy.ts
```diff
@@ -0,0 +1,121 @@
...
+export const deployAddCustomApprovalRuleAndCondition = async (
+  changes: Change<InstanceElement>[],
+  client: SalesforceClient,
+  dataManagement: DataManagement | undefined
+): Promise<DeployResult> => {
+  const approvalRuleChanges = await awu(changes)
+    .filter(isInstanceOfTypeChange(SBAA_APPROVAL_RULE))
+    .toArray()
+  const approvalConditionChanges = _.pullAll(changes, approvalRuleChanges)
+  log.debug('Deploying ApprovalRule instances without Custom ConditionsMet')
```
adamro, line 51: nit: Can you specify they we deploy them with All instead of Custom in the log here please?

packages/salesforce-adapter/src/sbaa_approval_rules_and_conditions_deploy.ts
```diff
@@ -0,0 +1,121 @@
...
+  )
+  const deployedRuleByElemId = _.keyBy(
+    approvalRulesNoCustomDeployResult.appliedChanges.map(getChangeData).filter(isInstanceElement),
+    instance => instance.elemID.getFullName()
+  )
+  const deployedRuleElemIds = new Set(Object.keys(deployedRuleByElemId))
+  log.debug('Deploying ApprovalCondition instances')
+  const [deployableConditionChanges, nonDeployableConditionChanges] = _.partition(
+    approvalConditionChanges,
+    change => deployedRuleElemIds.has(getChangeData(change).value[SBAA_APPROVAL_RULE].elemID.getFullName())
```
adamro, line 71: Maybe create a Record to the (ApprovalRule change + referencedRuleElemId) and use it both for this partition and later for the reference instead of calculating it again in line 84?
(And if you need/want the verification you do in line 81 it might be a good idea to do it early and declare it non-deployable, no?)

packages/salesforce-adapter/src/sbaa_approval_rules_and_conditions_deploy.ts
```diff
@@ -0,0 +1,121 @@
...
+  await awu(deployableConditionChanges)
+    .forEach(async change => {
+      const instance = getChangeData(change)
+      const approvalRule = instance.value[SBAA_APPROVAL_RULE]
+      if (!isInstanceElement(approvalRule)) {
+        log.error('Expected ApprovalCondition with name %s to contain InstanceElement for the sbaa__ApprovalRule__c field', instance.elemID.getFullName())
+        return
+      }
+      const deployedRule = deployedRuleByElemId[approvalRule.elemID.getFullName()]
+      if (deployedRule === undefined) {
```
adamro, line 85: Didn't you check this won't happen in the partition above?

packages/salesforce-adapter/src/sbaa_approval_rules_and_conditions_deploy.ts
```diff
@@ -0,0 +1,121 @@
...
+  // Set the ApprovalRule Ids for the ApprovalCondition instances
+  // Note that the Errors here are unlikely to happen and will cause weird flows in the deploy.
+  // I've added logs for visibility just in case we ever encounter them.
+  await awu(deployableConditionChanges)
+    .forEach(async change => {
+      const instance = getChangeData(change)
+      const approvalRule = instance.value[SBAA_APPROVAL_RULE]
+      if (!isInstanceElement(approvalRule)) {
+        log.error('Expected ApprovalCondition with name %s to contain InstanceElement for the sbaa__ApprovalRule__c field', instance.elemID.getFullName())
+        return
```
adamro, line 82: Isn't this dangerous? We are gonna just deploy something with an unknown value in `SBAA_APPROVAL_RULE` here?

packages/salesforce-adapter/src/sbaa_approval_rules_and_conditions_deploy.ts
```diff
@@ -0,0 +1,121 @@
...
+import { DataManagement } from './fetch_profile/data_management'
+import { isInstanceOfTypeChange, safeApiName } from './filters/utils'
+import { ADD_CUSTOM_APPROVAL_RULE_AND_CONDITION_GROUP, SBAA_APPROVAL_RULE, SBAA_CONDITIONS_MET } from './constants'
+import { deployCustomObjectInstancesGroup } from './custom_object_instances_deploy'
+
+const { awu } = collections.asynciterable
+const log = logger(module)
+
+const createNonDeployableConditionChangeError = (change: Change): SaltoElementError => ({
+  message: `Cannot deploy ApprovalCondition instance ${getChangeData(change).elemID.getFullName()} since it depends on a non-deployable ApprovalRule instance`,
```
adamro, line 37: It's not really `non-deployable`, it's an ApprovalRule that failed to deploy, isn't it?
Should we write also the ApprovalRule that it's referencing here to be a bit more informative?

We probably want a product 👁️ here - @tomermevorach @liorn

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -445,6 +456,13 @@ export default class SalesforceAdapter implements AdapterOperations {
           errors: [{ message: 'Cannot deploy CustomObject Records as part of check-only deployment', severity: 'Error' }],
         }
       }
+      if (changeGroup.groupID === ADD_CUSTOM_APPROVAL_RULE_AND_CONDITION_GROUP) {
```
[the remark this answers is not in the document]
tamtamirr: It would be impossible to move this logic with the current restrictions of **deployCustomObjectInstancesGroup**.
The **deployCustomObjectInstancesGroup** forbids groups with more than one type (which this PR is the first use-case for it).

Moreover, we are dealing with **three groups** of CustomObject instances in this use-case, so calling **deployCustomObjectInstancesGroup** three times makes more sense IMO. (That makes me wonder, do we want to pass custom names to these groups as-well for better indication of what Group actually failed..?)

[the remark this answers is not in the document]
adamro: The restrictions can still hold if calling `deployCustomObjectInstancesGroup` has an if inside it that checks if it's this special group called  `ADD_CUSTOM_APPROVAL_RULE_AND_CONDITION_GROUP` and then runs your new code and otherwise acts "normally" with the same assumptions. The adapter.ts knowing this group is so special is what's weird to me. It's still a CustomObjectInstances group.

I didn't understand your 2nd paragraph. It's a single group that you broke the deploy of to 3 parts, it's not 3 groups.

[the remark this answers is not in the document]
tamtamirr: @adamro It can't directly call my code, since we gonna end up in infinite loop of calls to **deployCustomObjectInstancesGroup**. That can be overcome by treating each of them as a separate **Data Change Group** with dedicated names (And not just the const name I use now) (Even though it's not what we received from the adapter). Thats what I meant in my 2nd paragraph. Putting my code inside **deployCustomObjectInstancesGroup** adds complexity that does not worth it IMO.

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -18,7 +18,12 @@ import {
   ElemIdGetter, FetchResult, AdapterOperations, DeployResult, FetchOptions, DeployOptions,
   ReadOnlyElementsSource,
 } from '@salto-io/adapter-api'
-import { filter, logDuration, resolveChangeElement, restoreChangeElement } from '@salto-io/adapter-utils'
+import {
```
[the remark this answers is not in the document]
tamtamirr: v

packages/salesforce-adapter/src/group_changes.ts
```diff
@@ -14,44 +14,73 @@
...
-const instanceOfCustomObjectChangeToGroupId: ChangeIdFunction = async change => ({
-  groupId: `${change.action}_${await apiName(await (getChangeData(change) as InstanceElement).getType())}_instances`,
-  // CustomObjects instances might have references to instances of the same type (via Lookup
-  // fields), and if we deploy them together the reference gets removed.
-  disjoint: false,
-})
-
-// Everything that is not a custom object instance goes into the deploy api
-const deployableMetadataChangeGroupId: ChangeIdFunction = () => ({ groupId: 'salesforce_metadata' })
+const getCustomRuleAndConditionChangeIds = async (
```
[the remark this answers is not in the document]
tamtamirr: I've renamed and added a comment explaining what type of instances fall into this group

packages/salesforce-adapter/src/adapter.ts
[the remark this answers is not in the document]
adamro: You thinking of a specific way of doing it but it's not the only one.

This code flow with this amount of importance to this specific group does not have a place in `adapter.ts` imo.

You can have a `deployCustomObjectInstances` fn that if it's this specific calls your code and if it's not calls `deployCustomObjectInstancesGroup`, and your code will call `deployCustomObjectInstancesGroup` directly and not the new `deployCustomObjectInstances` for example.
And if you want to be more explicit then what I described as `deployCustomObjectInstances` can be called `deployCustomObjectInstancesGroup` and what is now called `deployCustomObjectInstancesGroup` renamed to `deploySingleTypeAndActionCustomObjectInstancesGroup` which indicates the restrictions it has. 


Just to not have `adapter.ts` differentiate so much between different ways of deploying CustomObjectInstances.

packages/salesforce-adapter/src/sbaa_approval_rules_and_conditions_deploy.ts
```diff
@@ -0,0 +1,121 @@
...
+    change => deployedRuleElemIds.has(getChangeData(change).value[SBAA_APPROVAL_RULE].elemID.getFullName())
+  )
+  // Set the ApprovalRule Ids for the ApprovalCondition instances
+  // Note that the Errors here are unlikely to happen and will cause weird flows in the deploy.
+  // I've added logs for visibility just in case we ever encounter them.
+  await awu(deployableConditionChanges)
+    .forEach(async change => {
+      const instance = getChangeData(change)
+      const approvalRule = instance.value[SBAA_APPROVAL_RULE]
+      if (!isInstanceElement(approvalRule)) {
```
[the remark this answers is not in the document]
tamtamirr: Is there any flow where that wouldn't be the case when we reach here though? Do we really need the extra async checks?

packages/salesforce-adapter/src/sbaa_approval_rules_and_conditions_deploy.ts
tamtamirr: If that's the case it's a critical error that I'd rather fail the whole deployment on. No? (It means we had a bug in our **group_changes** or the instance was modified in one of the **preDeploys**. I've added the following if statement in the beginning of this method:
```
if (approvalRuleChanges.map(getChangeData).some(instance => instance.value[SBAA_CONDITIONS_MET] !== 'Custom')) {
    throw new Error('Received ApprovalRule instance without Custom ConditionsMet')
  }
```

packages/salesforce-adapter/src/sbaa_approval_rules_and_conditions_deploy.ts
tamtamirr: Good idea. Combined all the logic to be within the partition code.

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -498,3 +507,105 @@ export const deployCustomObjectInstancesGroup = async (
     }
   }
 }
+
+const createNonDeployableConditionChangeError = (change: Change): SaltoElementError => ({
```
adamro, line 511: nit: I think it's still legit to have it in a different file

adamro:
Looks good!
Great job with the 2nd iteration + tests!

---

I think we need to do something about the Change Validator from the go tho. 
We probably want an adapter config for this. If it's set to true, then when deploying it uses the new code & the Change Validator should not return any Errors, and when it's false, it should work as it worked before this PR (With the Change Validator and deploying without the new grouping).

And we need to decide if the default is true or false.
cc: @amitgoldarb 

---

Also, can you describe which manual tests did you do to test this code?

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
tamtamirr: I've done it for two reasons:
- Exporting to a new file would require me to export the method `deploySingleTypeAndActionCustomObjectInstancesGroup` in order to use it.  
- This will cause circular dependencies between the new file and **custom_object_instances_deploy.ts**.

tamtamirr:
@adamro Re the CV, I think we should remove it. (Or make the severity of the CV `warning`). I feel quite confident removing it as-well though.

tamtamirr:
@adamro Updated the PR description with the test scenarios I've ran just now 🙏

adamro:
@tamtamirr Making it a Warning doesn't help cause we instruct the users there to edit. If we have confidence about this we should remove it.

adamro (approved):
Approving but please delete the Change Validator before merging

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -329,14 +329,27 @@ const deployAddInstances = async (
   client: SalesforceClient,
   groupId: string
 ): Promise<DeployResult> => {
-  const instancesToUpdate = instances
-    .filter(instance => Object.values(instance.value).some(isInstanceElement))
+  const instancesToUpdate: InstanceElement[] = []
```
adamro, line 332: We need to change the name here -
The most extreme verbose version (we need to shorten it :) )

`instancesDependentOnOtherInstancesWithTheSameType`

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -329,14 +329,27 @@ const deployAddInstances = async (
...
   // Replacing self-reference field values with the resolved instances that will later contain the Record Ids
   const instanceByElemId = _.keyBy(instances, instance => instance.elemID.getFullName())
   await awu(instances).forEach(instance => {
-    instance.value = _.mapValues(instance.value, val => (isInstanceElement(val)
-      ? instanceByElemId[val.elemID.getFullName()] ?? val
-      : val))
+    let markedToUpdate = false
+    instance.value = _.mapValues(instance.value, value => {
+      if (isInstanceElement(value)) {
+        const referencedInstance = instanceByElemId[value.elemID.getFullName()]
```
adamro, line 339: `sameTypeReferencedInstance`

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -552,13 +574,10 @@ const deployAddCustomApprovalRulesAndConditions = async (
         log.error('Expected ApprovalCondition with name %s to contain InstanceElement for the sbaa__ApprovalRule__c field', instance.elemID.getFullName())
         return false
       }
-      const deployedRule = deployedRuleByElemId[approvalRule.elemID.getFullName()]
-      if (deployedRule === undefined) {
+      if (approvalRule.value[CUSTOM_OBJECT_ID_FIELD] === undefined) {
```
adamro, line 577: Just add a comment explaining only successfully deployed instances have ids

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -329,27 +329,18 @@ const deployAddInstances = async (
   client: SalesforceClient,
   groupId: string
 ): Promise<DeployResult> => {
-  const instancesToUpdate: InstanceElement[] = []
+  const instancesWithReferenceToNonDeployedInstance = instances
```
adamro, line 332: `instancesReferencingToBeDeployedInstances`

adamro, line 332: This is a bit hackier than the previous version so maybe we need a comment here that instances with Ids are ones that were already deployed in another group and ones without will be handled inside this group

benisraeladi (approved):

PR #4530, opened 2023-06-21
SALTO-4266: delete empty accountIds in scriptRunner

IdoZakk:
_Removed empty accountIds to avoid errors during deployment_

---

_Not adding noise reduction as we do not have customers with enabled scriptrunner, except maybe Convex_

---
_Release Notes_: 
Jira Adapter:
* Fixed a bug in scriptRunner that prevented deployment of notification post functions without users

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/script_runner/workflow/empty_account_ids.ts
+++ b/packages/jira-adapter/src/filters/script_runner/workflow/empty_account_ids.ts
@@ -1,0 +1,93 @@
+/*
+*                      Copyright 2023 Salto Labs Ltd.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+import { isInstanceElement, Element, isAdditionOrModificationChange, getChangeData, isInstanceChange, Value } from '@salto-io/adapter-api'
+import { collections } from '@salto-io/lowerdash'
+import { FilterCreator } from '../../../filter'
+import { WorkflowInstance, isWorkflowInstance } from '../../workflow/types'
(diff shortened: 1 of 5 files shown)


packages/jira-adapter/src/filters/script_runner/workflow/empty_account_ids.ts
```diff
@@ -0,0 +1,93 @@
...
+import { isInstanceElement, Element, isAdditionOrModificationChange, getChangeData, isInstanceChange, Value } from '@salto-io/adapter-api'
+import { collections } from '@salto-io/lowerdash'
+import { FilterCreator } from '../../../filter'
+import { WorkflowInstance, isWorkflowInstance } from '../../workflow/types'
+import { SCRIPT_RUNNER_POST_FUNCTION_TYPE, SCRIPT_RUNNER_SEND_NOTIFICATIONS } from './workflow_cloud'
+
+const { makeArray } = collections.array
+
+const changeAccountIds = (func: (scriptRunner: Value) => Value):
+((workflowInstance: WorkflowInstance) => void) => (workflowInstance: WorkflowInstance): void => {
```
alonstern, line 26: These are pretty hard to follow, can it not be a function that returns a function and call it like `changeAccountIds(deleteEmptyAccountsId, workflowInstance)`?

packages/jira-adapter/src/filters/script_runner/workflow/empty_account_ids.ts
```diff
@@ -0,0 +1,93 @@
...
+  workflowInstance.value.transitions.forEach(transition => {
+    makeArray(transition.rules?.postFunctions).forEach(postFunction => {
+      if (postFunction.type === SCRIPT_RUNNER_POST_FUNCTION_TYPE
+          && postFunction.configuration?.scriptRunner !== undefined
+          && postFunction.configuration.scriptRunner.className === SCRIPT_RUNNER_SEND_NOTIFICATIONS) {
```
alonstern, line 31: nit:
```suggestion
          && postFunction.configuration.scriptRunner?.className === SCRIPT_RUNNER_SEND_NOTIFICATIONS) {
```

alonstern, line 31: Are we sure `SCRIPT_RUNNER_SEND_NOTIFICATIONS` is the only case of this? I think in that sense ignoring `""` in the change validator would be safer, but what ever you think

IdoZakk: The change validator is based on the general account IDs function `accountIdsScenarios`
I did not want to add a specific case just for that. And always ignoring "" is not valid as in most cases you must specify the user

PR #4531, opened 2023-06-22
SALTO-4260: Fix link type string error

IdoZakk:
_Fixed a bug with scriptrunner. FIELD_LINK_TYPE was assumed to always be of format 'type direction' but direction can be omitted

---

_According to adaptavist the direction can be omitted, in which case it is considered outwards.
I added robustness- if the string contains only one part after parsing use that part as the id.
I also added a test to reveal branch coverage and make sure I kept it_

---
_Release Notes_: 
Jira Adapter:
* Fixed a bug in scriptRunner, sometimes FIELD_LINK_TYPE was not parsed corretly

---
_User Notifications_: 
None


alonstern (approved):
👌

PR #4534, opened 2023-06-22
SALTO-3703 - Validate a user is able to deploy an action with deflection action

idoser96:
When a user attempts to deploy an instance with an action of deflection, validate that that type of action is activated in the service

---

* This removes the validator that prevents deploying deflection actions, as our check now is more specific
* The relevant files for that PR are only invalid_action.ts and deflection_action.ts, the others files were split to [a different PR](https://github.com/salto-io/salto/pull/4536/files)

---
_Release Notes_: 
Zendesk Adapter:
* It is now possible to deploy triggers, macros and automations with action 'deflection', if the feature is activated in the service

---
_User Notifications_: 
None


packages/zendesk-adapter/src/change_validators/deflection_action.ts
```diff
@@ -0,0 +1,90 @@
...
+    log.error('Failed to run deflectionActionValidator because element source is undefined')
+    return []
+  }
+
+  const accountSettings = await elementSource.get(
+    new ElemID(ZENDESK, ACCOUNT_SETTING_TYPE_NAME, 'instance', ElemID.CONFIG_NAME)
+  )
+
+  if (!isValidSettings(accountSettings)) {
+    log.error('Failed to run deflectionActionValidator because account settings instance is invalid')
```
edenhassid, line 78: I think this shouldn't be log.error, maybe log.warn, cause maybe `automatic_answers` field will not exist

edenhassid (commented):
LGTM! added small comments

edenhassid (approved):
💯

packages/zendesk-adapter/src/change_validators/deflection_action.ts
netama: @idoser96 looks like this was not updated?

packages/zendesk-adapter/src/change_validators/deflection_action.ts
```diff
@@ -0,0 +1,90 @@
...
+    return []
+  }
+
+  const isDeflectionOn = accountSettings.value.active_features.automatic_answers
+
+  return isDeflectionOn ? [] : relevantInstances.map(instance => ({
+    elemID: instance.elemID,
+    severity: 'Error',
+    message: 'Action requires turning on automatic answers',
+    detailedMessage: `To enable the configuration of the '${DEFLECTION_TYPE}' field action, which allows for ‘autoreply with articles’, please ensure that the automatic answers feature is turned on. To do so, please update the 'automatic_answers' setting to 'true' in the account_settings.`,
```
netama, line 88: I'm wondering if we might want to turn this to a warning so that the user can enable the feature and still continue with the deploy, wdyt? also @tomermevorach

PR #4536, opened 2023-06-22
SALTO-4403 - Warn the user on activation of a feature

idoser96:
When a user deploys a change to his account_setting, activating a feature, we will let him know this might cause paid features to be activated

---

Instead of researching which features are paid and which are not, we alert on all of them

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/zendesk-adapter/src/change_validator.ts
```diff
@@ -108,7 +109,6 @@ export default ({
     noDuplicateLocaleIdInDynamicContentItemValidator,
     onlyOneTicketFormDefaultValidator,
     customRoleNameValidator,
-    invalidActionsValidator,
```
edenhassid: why did you remove this CV?

packages/zendesk-adapter/src/change_validators/feature_activation.ts
```diff
@@ -0,0 +1,48 @@
...
+*/
+import {
+  ChangeValidator, getChangeData,
+  isInstanceChange, isModificationChange,
+} from '@salto-io/adapter-api'
+import { detailedCompare } from '@salto-io/adapter-utils'
+import { ACCOUNT_SETTING_TYPE_NAME } from '../filters/account_settings'
+
+/**
+ * Warns the user if he activates a feature, some features maybe cost money or be limited
```
edenhassid, line 24: nit:
```suggestion
 * Warns the user if he activates a feature, some features may cost money or be limited
```

packages/zendesk-adapter/src/change_validators/feature_activation.ts
```diff
@@ -0,0 +1,48 @@
...
+    .filter(isInstanceChange)
+    .find(change => getChangeData(change).elemID.typeName === ACCOUNT_SETTING_TYPE_NAME)
+
+  if (accountSettingsChange === undefined) {
+    return []
+  }
+
+  const activatedFeatures = detailedCompare(accountSettingsChange.data.before, accountSettingsChange.data.after)
+    .filter(isModificationChange)
+    .filter(detailedChange => detailedChange.id.getFullNameParts()[4] === 'active_features')
```
edenhassid, line 38: can you add a comment of how the name parts should look?

packages/zendesk-adapter/src/change_validators/feature_activation.ts
```diff
@@ -0,0 +1,48 @@
...
+    .filter(detailedChange => detailedChange.data.before === false && detailedChange.data.after === true)
```
edenhassid, line 39: what about features that turn from true to false? we don't want to warn about them too?

packages/zendesk-adapter/src/change_validators/feature_activation.ts
```diff
@@ -0,0 +1,48 @@
...
+    .map(detailedChange => detailedChange.id.name)
+
+  return activatedFeatures.length === 0 ? [] : [{
+    elemID: accountSettingsChange.data.after.elemID,
+    severity: 'Info',
+    message: 'Activating new features may include additional cost',
+    detailedMessage: `Features ${activatedFeatures.join(', ')} are marked for activation and may require additional cost in order to operate`,
```
edenhassid, line 46: did @tomermevorach review the message and the detailedMessage?

edenhassid (commented):
looks good! added small questions and nits

packages/zendesk-adapter/src/change_validators/feature_activation.ts
idoser96: yes

edenhassid (approved):
🔥 🔥 🔥

packages/zendesk-adapter/src/change_validators/feature_activation.ts
netama: this comparison could use some more safeties - 
1. we shouldn't compare using `nameParts` - please use `createTopLevelParentID()` and check the `path`, rather than using `getFullnameParts()`
2. we are making some assumptions about the structure here, we should use a typeguard
3. can't there be cases where these are additions / the "before" flag doesn't exist?

idoser96: 2. There isn't really anything to guard here, there is no risk of an exception or a mismatch, if path[0] is not 'active_features' it does not interest us and we won't do anything

PR #4537, opened 2023-06-22
SALTO-4402 - Salesforce: Adapter config to control profile references

tomsellek:
Fetching profiles takes a very long time. One part of that time is resolving references on the many, many fields that each profile has (basically the number of fields in a single profile = O(number of fields in all custom objects)). 
So let's not resolve references in profiles by default, unless the user explicitly asks us to.

---

N/A

---
_Release Notes_: 
Salesforce: references will no longer be generated from profile fields

---
_User Notifications_: 
Salesforce: References in fields of profiles will be replaced with the field's original content.

### DIFF
--- a/packages/salesforce-adapter/src/transformers/reference_mapping.ts
+++ b/packages/salesforce-adapter/src/transformers/reference_mapping.ts
@@ -244,7 +244,7 @@ export const defaultFieldNameToTypeMappingDefs: FieldReferenceDefinition[] = [
     target: { type: 'ApexPage' },
   },
   {
-    src: { field: 'apexClass', parentTypes: ['FlowApexPluginCall', 'FlowVariable', 'ProfileApexClassAccess', 'TransactionSecurityPolicy'] },
+    src: { field: 'apexClass', parentTypes: ['FlowApexPluginCall', 'FlowVariable', 'TransactionSecurityPolicy'] },
     target: { type: 'ApexClass' },
   },
   {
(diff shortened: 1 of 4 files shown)


packages/salesforce-adapter/src/transformers/reference_mapping.ts
```diff
@@ -756,8 +781,10 @@ export type ReferenceResolverFinder = (
  * Generates a function that filters the relevant resolvers for a given field.
  */
 export const generateReferenceResolverFinder = (
-  defs = fieldNameToTypeMappingDefs
+  defs: FieldReferenceDefinition[],
 ): ReferenceResolverFinder => {
+// TODO how to get optional features from here?
+
```
benisraeladi, line 787: todo here, I think you'll have to do the ugly thing of creating a boolean and moving it through the entire chain until getting here :(

tomsellek: Actually I think I managed to do it by changing the value of `defs` and just forgot to remove the comment. But please verify me on this.

amitgoldarb: I agree with @benisraeladi that we should add a UT here test the different config options.

packages/salesforce-adapter/src/transformers/reference_mapping.ts
```diff
@@ -693,11 +704,25 @@ const fieldPermissionEnumDisabledExtraMappingDefs: FieldReferenceDefinition[] =
...
  * 2. Resolving the resulting reference expression back returns the original value.
  */
-export const fieldNameToTypeMappingDefs: FieldReferenceDefinition[] = [
+const fieldNameToTypeMappingDefs: FieldReferenceDefinition[] = [
   ...defaultFieldNameToTypeMappingDefs,
   ...fieldPermissionEnumDisabledExtraMappingDefs,
+  ...referencesFromProfile,
 ]
 
+export const getReferenceMappingDefs = (
```
amitgoldarb, line 713: Can we add some UT to this function.

amitgoldarb (approved):
LGTM!
Approved with some small comments.

packages/salesforce-adapter/src/transformers/reference_mapping.ts
tomsellek: Added unit tests. @benisraeladi kindly re-review

benisraeladi (approved):
Awesome!

PR #4542, opened 2023-06-23
SAAS-7513 - Add some logs to workspace state initialization

tomsellek:
This issue has to do with way too many remote maps being opened. We don't know why, yet. Hopefully, adding some logs will point us in the right direction.

---

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/workspace/src/workspace/workspace.ts
+++ b/packages/workspace/src/workspace/workspace.ts
@@ -380,6 +380,7 @@ export const loadWorkspace = async (
     const initState = async (): Promise<WorkspaceState> => {
       const wsConfig = await config.getWorkspaceConfig()
       log.debug('initializing state for workspace %s/%s', wsConfig.uid, wsConfig.name)
+      log.debug('Full workspace config: %o', wsConfig)
       const states: Record<string, SingleState> = Object.fromEntries(await awu(envs())
         .map(async envName => [envName, {
           merged: new RemoteElementSource(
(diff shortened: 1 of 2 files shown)


packages/workspace/src/workspace/workspace.ts
```diff
@@ -378,6 +378,8 @@ export const loadWorkspace = async (
     validate?: boolean
   }): Promise<WorkspaceState> => {
     const initState = async (): Promise<WorkspaceState> => {
+      const wsConfig = await config.getWorkspaceConfig()
+      log.debug('initializing state for workspace %s/%s', wsConfig.uid, wsConfig.name)
```
benisraeladi, line 382: If it's not too large, it might be prudent to add the entire `wsConfig` with `%o`.

packages/workspace/src/workspace/workspace.ts
```diff
@@ -758,6 +760,8 @@ export const loadWorkspace = async (
 
   const getWorkspaceState = async (): Promise<WorkspaceState> => {
     if (_.isUndefined(workspaceState)) {
+      const wsConfig = await config.getWorkspaceConfig()
+      log.debug('No workspace state for %s/%s. Building new workspace state.', wsConfig.uid, wsConfig.name)
```
benisraeladi, line 764: add the `wsConfig` in its entirety here or where the comment above states

packages/workspace/src/workspace/workspace.ts
tomsellek: Ack

packages/workspace/src/workspace/workspace.ts
tomsellek: Ack.

benisraeladi (approved):

PR #4544, opened 2023-06-25
SALTO-4396: Support multi brands in Okta

shir-reifenberg:
Okta made a change and they now support more than one brand per org

---

Okta now supports defining more than one brand per org.
This means that on dependent types are no longer singletons.
Also, they added some references to the API response of `Domain` and `Brand`
(Still missing one more ref that I need to resolve)

---
_Release Notes_: 

_Okta_adapter_:
- Support multibrand customizations
- Support fetch of 'EmailDomain' instances

---
_User Notifications_: 

_Okta_adapter_:
- Changes to `Brand`, `BrandTheme`, `BrandLogo` and `Domain` instances due to recent changes in Okta
- 'EmailDomain' instances will be added

### DIFF
--- a/packages/okta-adapter/src/config.ts
+++ b/packages/okta-adapter/src/config.ts
@@ -729,11 +729,18 @@ const DEFAULT_TYPE_CUSTOMIZATIONS: OktaSwaggerApiConfig['types'] = {
   api__v1__brands: {
     request: {
       url: '/api/v1/brands',
-      recurseInto: [{
-        type: 'api__v1__brands___brandId___themes@uuuuuu_00123_00125uu',
-        toField: 'theme',
-        context: [{ name: 'brandId', fromField: 'id' }],
-      }],
+      recurseInto: [
+        {
+          type: 'api__v1__brands___brandId___themes@uuuuuu_00123_00125uu',
+          toField: 'theme',
+          context: [{ name: 'brandId', fromField: 'id' }],
+        },
+        {
+          type: 'api__v1__brands___brandId___templates__email@uuuuuu_00123_00125uuuu',
+          toField: 'emailTemplates',
+          context: [{ name: 'brandId', fromField: 'id' }],
+        },
+      ],
     },
     transformation: {
       dataField: '.',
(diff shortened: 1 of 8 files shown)


packages/okta-adapter/src/config.ts
```diff
@@ -838,7 +846,7 @@ const DEFAULT_TYPE_CUSTOMIZATIONS: OktaSwaggerApiConfig['types'] = {
   },
   BrandTheme: {
     transformation: {
-      isSingleton: true,
+      idFields: [],
```
netama, line 849: should we add `extendsParentId` on this (maybe also elsewhere)?

packages/okta-adapter/src/filters/brand_theme_files.ts
```diff
@@ -77,24 +82,19 @@ const toSaltoError = (err: Error): SaltoError => ({
...
-      }
-      return { errors: [saltoErr] }
-    }
-    const logoTypes = logoTypeNames.map(logoTypeName => createLogoType(logoTypeName))
-    logoTypes.forEach(logoType => elements.push(logoType))
+      .filter(instance => instance.elemID.typeName === BRAND_THEME_TYPE_NAME)
 
-    const allInstances = await Promise.all(logoTypes.map(async logoType =>
-      getBrandThemeFile(client, brandTheme, logoType)))
+    const [logoType, faviconType] = [createLogoType(BRAND_LOGO_TYPE_NAME), createLogoType(FAV_ICON_TYPE_NAME)]
```
netama, line 89: nit: should we rename the function (since not all are logos)?

packages/okta-adapter/src/reference_mapping.ts
```diff
@@ -212,6 +212,22 @@ export const referencesRules: OktaFieldReferenceDefinition[] = [
   { src: { field: 'include', parentTypes: ['DeviceCondition'] },
     serializationStrategy: 'id',
     target: { type: 'DeviceAssurance' } },
+  // TODO figure out type
```
netama, line 215: leftover?

netama (approved):
LGTM with a few questions so pre-approving. should we also update/add some tests (maybe just for the updated filters)?

packages/okta-adapter/src/config.ts
shir-reifenberg: I think `extendsParentId` is only necessary when re-creating elemIds with `referencedInstanceNames` filter. The default behavior for creating instance elemId is using the parent when it exists.

packages/okta-adapter/src/reference_mapping.ts
shir-reifenberg: Actually it's a reference for type we don't fetch today, will update PR with this type + ref

packages/okta-adapter/src/config.ts
netama: it's used only there, but when someone changes the elem id they won't necessarily know/remember they should set this, so I think it's better to set the flag up front where possible (remember that the id fields can also be customized by users)

netama (approved):

PR #4547, opened 2023-06-26
SALTO-4417: Fix Workflow Deployment bug

IdoZakk:
_Fixed a bug that prevented the deployment of a workflow if there is a workflow scheme with a reference to it and a missing reference_

---

Other solutions that I could think of were too complicated.
The best solution IMO remains to change the way we deploy workflow modifications

---
_Release Notes_: 
Jira Adapter: 
* Fixed a bug that prevented workflow modification deployment in some cases

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/reference_mapping.ts
+++ b/packages/jira-adapter/src/reference_mapping.ts
@@ -391,19 +391,16 @@ export const referencesRules: JiraFieldReferenceDefinition[] = [
   {
     src: { field: 'defaultWorkflow', parentTypes: ['WorkflowScheme'] },
     serializationStrategy: 'name',
-    jiraMissingRefStrategy: 'typeAndValue',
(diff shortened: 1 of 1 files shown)


packages/jira-adapter/src/reference_mapping.ts
```diff
@@ -391,19 +391,16 @@ export const referencesRules: JiraFieldReferenceDefinition[] = [
   {
     src: { field: 'defaultWorkflow', parentTypes: ['WorkflowScheme'] },
     serializationStrategy: 'name',
-    jiraMissingRefStrategy: 'typeAndValue',
```
alonstern, line 394: Could you doc why we can't use `jiraMissingRefStrategy` in here?

alonstern (approved):
🔥

PR #4548, opened 2023-06-26
SALTO-3967: Support deployment of circular workflow transitions

IdoZakk:
_Atlassian has added support to deploy circular workflow transitions_

---

_Just had to remove existing CR and filter_

---
_Release Notes_: 
Jira Adapter:
* Support new API to deploy circular workflow transitions 

---
_User Notifications_: 
None


alonstern (approved):
🎉 🎉

PR #4555, opened 2023-06-27
SALTO-4297 - Only create source map when required

tomsellek:
Propagate the `calcSourceMap` parameter to the callers of parseNaclFile and set the appropriate value

---

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
+++ b/packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
@@ -248,24 +248,57 @@ export const toParsedNaclFile = async (
 }
 
 async function parseNaclFile(
-  naclFile: NaclFile, functions: Functions, createSourceMap?: 'createSourceMap'
-): Promise<Required<ParseResult>>
-async function parseNaclFile(
-  naclFile: NaclFile, functions: Functions, createSourceMap: 'createSourceMap' | 'noSourceMap'
+  {
+    naclFile,
+    functions,
+    createSourceMap,
+  }:
+  {
+    naclFile: NaclFile
+    functions: Functions
+    createSourceMap?: false
+  }
 ): Promise<ParseResult>
+async function parseNaclFile(
+  {
+    naclFile,
+    functions,
+    createSourceMap,
+  }:
+  {
+    naclFile: NaclFile
+    functions: Functions
+    createSourceMap: boolean
+  }
+): Promise<Required<ParseResult>>
 
 async function parseNaclFile(
-  naclFile: NaclFile, functions: Functions, createSourceMap: 'createSourceMap' | 'noSourceMap' = 'createSourceMap'
+  {
+    naclFile,
+    functions,
+    createSourceMap = false,
+  }:
+  {
+    naclFile: NaclFile
+    functions: Functions
+    createSourceMap?: boolean
+  }
 ): Promise<ParseResult> {
-  return parse(Buffer.from(naclFile.buffer), naclFile.filename, functions, createSourceMap === 'createSourceMap')
+  return parse(Buffer.from(naclFile.buffer), naclFile.filename, functions, createSourceMap)
+}
+
+type ParseNaclFilesArgs = {
+  naclFiles: NaclFile[]
+  functions: Functions
+  createSourceMap?: boolean
 }
 
 const parseNaclFiles = async (
-  naclFiles: NaclFile[], functions: Functions, createSourceMap: 'createSourceMap' | 'noSourceMap'
+  { naclFiles, functions, createSourceMap = false } : ParseNaclFilesArgs
 ): Promise<ParsedNaclFile[]> => (
   withLimitedConcurrency(
     naclFiles.map(
-      naclFile => async () => toParsedNaclFile(naclFile, await parseNaclFile(naclFile, functions, createSourceMap))
+      naclFile => async () => toParsedNaclFile(naclFile, await parseNaclFile({ naclFile, functions, createSourceMap }))
     ),
     PARSE_CONCURRENCY,
   )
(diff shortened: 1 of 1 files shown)


packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
```diff
@@ -247,17 +247,25 @@ export const toParsedNaclFile = async (
...
+async function parseNaclFile(
+  naclFile: NaclFile, functions: Functions, createSourceMap?: 'createSourceMap'
+): Promise<Required<ParseResult>>
+async function parseNaclFile(
+  naclFile: NaclFile, functions: Functions, createSourceMap: 'createSourceMap' | 'noSourceMap'
+): Promise<ParseResult>
+
+async function parseNaclFile(
+  naclFile: NaclFile, functions: Functions, createSourceMap: 'createSourceMap' | 'noSourceMap' = 'createSourceMap'
+): Promise<ParseResult> {
```
benisraeladi, line 259: Are these all definitions for the same function that starts in 260?

packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
```diff
@@ -247,17 +247,25 @@ export const toParsedNaclFile = async (
...
-  (parse(Buffer.from(naclFile.buffer), naclFile.filename, functions))
+async function parseNaclFile(
+  naclFile: NaclFile, functions: Functions, createSourceMap?: 'createSourceMap'
+): Promise<Required<ParseResult>>
+async function parseNaclFile(
+  naclFile: NaclFile, functions: Functions, createSourceMap: 'createSourceMap' | 'noSourceMap'
+): Promise<ParseResult>
+
+async function parseNaclFile(
+  naclFile: NaclFile, functions: Functions, createSourceMap: 'createSourceMap' | 'noSourceMap' = 'createSourceMap'
```
ori-moisis, line 258: I think the default should be no source map because:
1. in most use cases we don't need the source map
2. when writing new code, if you need the source map and you got it wrong, you would figure out that you didn't get it and fix your code, but if you don't need the source map and the default generates it for you, you just wouldn't think about it (this is how we got to this state in the first place), there would be nothing to make you find your "mistake"

packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
```diff
@@ -247,17 +247,25 @@ export const toParsedNaclFile = async (
   }
 }
 
-const parseNaclFile = async (
-  naclFile: NaclFile, functions: Functions,
-): Promise<Required<ParseResult>> =>
-  (parse(Buffer.from(naclFile.buffer), naclFile.filename, functions))
+async function parseNaclFile(
+  naclFile: NaclFile, functions: Functions, createSourceMap?: 'createSourceMap'
```
ori-moisis, line 251: not loving the "enum to represent a boolean" thing
I understand wanting this to be named, so, maybe make it a named parameter instead?

either
```
type ParseNaclFileArgs = {
  naclFile: NaclFile
  functions: Functions
  createSourceMap: boolean
}
.. parseNaclFile(args: ParseNaclFileArgs) ...
```
or just the "extra options", e.g
```
type ParseNaclFileOptions = {
  createSourceMap: boolean
}
.. parseNaclFile(naclFile: NaclFile, functions: Functions, options?: ParseNaclFileOptions) ...
```

tomsellek: Sure, that makes sense.

packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
tomsellek: Ack.

packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
tomsellek: Yup. Overloading in Typescript. It's not often that I miss C++.

benisraeladi (approved):

packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
```diff
@@ -247,17 +247,58 @@ export const toParsedNaclFile = async (
...
+  }
+): Promise<ParseResult> {
+  return parse(Buffer.from(naclFile.buffer), naclFile.filename, functions, createSourceMap)
+}
+
+type ParseNaclFilesArgs = {
+  naclFiles: NaclFile[]
+  functions: Functions
+  createSourceMap?: boolean
+}
```
ori-moisis, line 294:
```suggestion
type ParseNaclFilesArgs = {
  naclFiles: NaclFile[]
  functions: Functions
  createSourceMap?: boolean
}
async function parseNaclFile(args: ParseNaclFilesArgs & { createSourceMap?: false }): Promise<ParseResult>
async function parseNaclFile(args: ParseNaclFilesArgs & { createSourceMap: true }): Promise<Required<ParseResult>>
async function parseNaclFile(
  { naclFile, functions, createSourceMap = false }: ParseNaclFilesArgs
): Promise<ParseResult> {
  return parse(Buffer.from(naclFile.buffer), naclFile.filename, functions, createSourceMap)
}
```
RYL (not quite DRY, but repeat yourself less?)

tomsellek: Unless I'm missing something, there's a difference between `ParseNaclFilesArgs` and `ParseNaclFileArgs`, which is that the `naclFiles` member is an array for `ParseNaclFilesArgs` but a single file for `ParseNaclFileArgs`. Let me see if I can take the spirit of this suggested change and build on it.

ori-moisis (approved):

PR #4558, opened 2023-06-28
SALTO-4362 - Allow to disable change validators in the config

idoser96:
Added an option to add a validator's name to a config in order to prevent it from running

---

https://salto-io.atlassian.net/jira/software/c/projects/SAAS/issues/SALTO-4468?jql=assignee%20IN%20%28currentUser%28%29%29%20AND%20statusCategory%20in%20%28%22To%20Do%22%2C%20%22In%20Progress%22%29%20ORDER%20BY%20created%20DESC

---
_Release Notes_: 
* Specific change validators can now be disabled in the config for all adapters

---
_User Notifications_: 
None


packages/okta-adapter/src/config.ts
```diff
@@ -1697,6 +1699,10 @@ export const configType = createMatchingObjectType<Partial<OktaConfig>>({
         }
       ),
     },
+    // TODO seroussi - this is wrong
+    [DEPLOY_CONFIG]: {
```
idoser96, line 1703: TODO

packages/salesforce-adapter/src/types.ts
```diff
@@ -111,7 +112,7 @@ export type ChangeValidatorName = (
   | 'dataCategoryGroup'
 )
 
-export type ChangeValidatorConfig = Partial<Record<ChangeValidatorName, boolean>>
+export type ChangeValidatorNamesConfig = Partial<Record<ChangeValidatorName, boolean>>
```
idoser96, line 115: TODO

packages/adapter-components/src/config/shared.ts
```diff
@@ -210,6 +212,7 @@ export const createUserDeployConfigType = (
     elemID: new ElemID(adapter, 'userDeployConfig'),
     fields: {
       defaultMissingUserFallback: { refType: BuiltinTypes.STRING },
+      changeValidators: { refType: createChangeValidatorsConfigType(adapter) },
```
alonstern, line 215: I think the resulting type of this is 
```
{
  changeValidators: {
    changeValidators: {
      validate: ...
      deploy: ...
    }
  }
}
```

alonstern (commented):
👌 👌

packages/adapter-components/src/config/shared.ts
```diff
@@ -68,6 +70,7 @@ export type UserFetchConfig<T extends Record<string, unknown> | undefined = Defa
 export type UserDeployConfig = {
   // Replace references for missing users during deploy with defaultMissingUserFallback value
   defaultMissingUserFallback?: string
+  changeValidators?: ChangeValidatorConfig
```
alonstern, line 73: I think also here it will be 
```
{
  changeValidators: {
    changeValidators: {
       deploy: ...
       validate: ...
    }
  }
}
```

packages/netsuite-adapter/src/change_validator.ts
```diff
@@ -146,16 +150,24 @@ const getChangeValidator: ({
...
+      const mergedValidator = createChangeValidator({
+        validators: { ...defaultChangeValidators, ...validators },
+        validatorsConfig,
+      })
       const validatorChangeErrors: ChangeError[] = _.flatten(await Promise.all([
-        ...changeValidators.map(validator => validator(changes, elementSource)),
-        ...validators.map(validator => validator(changes, deployReferencedElements, elementsSource)),
+        mergedValidator(changes, elementSource),
         warnStaleData ? safeDeployValidator(changes, fetchByQuery, deployReferencedElements) : [],
       ]))
```
dantal4, line 172: can you do this instead?
```suggestion
      const safeDeploy = warnStaleData
        ? { safeDeploy: (innerChanges: ReadonlyArray<Change>) =>
          safeDeployValidator(innerChanges, fetchByQuery, deployReferencedElements) }
        : {}
      const mergedValidator = createChangeValidator({
        validators: { ...defaultChangeValidators, ...validators, ...safeDeploy },
        validatorsConfig,
      })
      const validatorChangeErrors = await mergedValidator(changes, elementSource)
```

packages/netsuite-adapter/src/change_validators/types.ts
```diff
@@ -20,3 +20,7 @@ export type NetsuiteChangeValidator = (
     deployReferencedElements?: boolean,
     elementsSource?: ReadOnlyElementsSource,
   ) => Promise<ReadonlyArray<ChangeError>>
+
+export type ChangeValidator = (
```
dantal4, line 24: I think that it's not used

packages/netsuite-adapter/src/config.ts
```diff
@@ -527,6 +532,7 @@ const deployConfigType = createMatchingObjectType<DeployParams>({
     validate: { refType: BuiltinTypes.BOOLEAN },
     deployReferencedElements: { refType: BuiltinTypes.BOOLEAN },
     additionalDependencies: { refType: additionalDependenciesType },
+    changeValidators: { refType: createValidatorConfigType(NETSUITE) },
```
dantal4, line 535: 1. I think that it's weird that the config path is `deploy.changeValidators.deploy`. In netsuite there is no "validate" command, so I think that it's not relevant. I suggest to use `deploy.changeValidators`, `validate.changeValidators` separately, so each adapter can decide which of them to implement.
2. Maybe instead of a map type we can use a list of `{ name: "validatorName", enabled: "true" / "false" }` and `name` will be restricted to the validators names list? And also to throw in case that the user puts an unknown name? Otherwise users will have typos and they won't know why it's not working.

dantal4 (commented):
Great work!
Added nits on NS side, and a general suggestion 🙏

packages/adapter-components/src/deployment/change_validators/create_change_validator.ts
```diff
@@ -0,0 +1,74 @@
...
+* limitations under the License.
+*/
+import { BuiltinTypes, ChangeValidator, ElemID, MapType, ObjectType } from '@salto-io/adapter-api'
+import _ from 'lodash'
+import { createMatchingObjectType } from '@salto-io/adapter-utils'
+import { logger } from '@salto-io/logging'
+
+const log = logger(module)
+
+export type ValidatorConfig = Record<string, boolean | undefined>
```
netama, line 23: the type is not that clear IMO - nothing here says what the boolean means (and the type name would imply this is mapped to an object IMO)

packages/adapter-components/src/deployment/change_validators/create_change_validator.ts
```diff
@@ -0,0 +1,74 @@
...
+    deploy?: ValidatorConfig
+    validate?: ValidatorConfig
+  }
+}
+export const createChangeValidator = ({
+  validators,
+  validatorsConfig = {},
+}: {
+  validators: Record<string, ChangeValidator>
+  validatorsConfig?: ValidatorConfig
```
netama, line 35: 1. naming - it's usually bad practice to have to types where the only difference is an `s`, I know we have some places in the code where this was done anyway but here it looks confusing IMO (`ValidatorConfig` vs `ValidatorsConfig`).
2. not sure if we have to tie the change validator creation directly to the config in this case - maybe keep the original `utils` function and wrap it with something that just translates from the config to the list of disabled validators + calls the inner function?

packages/adapter-components/src/deployment/change_validators/skip_parents_of_skipped_instances.ts
```diff
@@ -16,13 +16,17 @@
...
-export const createSkipParentsOfSkippedInstancesValidator = (
-  changeValidators: ReadonlyArray<ChangeValidator>,
-  disabledValidators?: ReadonlyArray<ChangeValidator>,
-): ChangeValidator => async (changes, elementSource) => {
-  const changeValidator = createChangeValidator(changeValidators, disabledValidators)
+export const createSkipParentsOfSkippedInstancesValidator = ({
+  validators,
+  validatorsConfig = {},
+}: {
+  validators: Record<string, ChangeValidator>
```
netama, line 26: the order in which the validators run is important - if we are sure this is stable (probably, since I think @dantal4 investigated it at the time + I see @ori-moisis 's original implementation in salesforce already used a mapping so hopefully it was tested then), we should add unittests that ensure the order is maintained (if there already is and I missed it sorry)

packages/adapter-components/test/deployment/change_validators/skip_parents_of_skipped_instances.test.ts
```diff
@@ -28,17 +28,19 @@ describe('createSkipParentsOfSkippedInstsValidator', () => {
...
   )
-  const mockChangeValidator: ChangeValidator = async () => [
-    {
-      elemID: skippedInst.elemID,
-      severity: 'Error',
-      message: 'Error',
-      detailedMessage: 'detailed error',
-    },
-  ]
+  const mockChangeValidator: Record<string, ChangeValidator> = {
```
netama, line 31: the name is misleading, this is a mapping right? why is it called `mockChangeValidator`?

packages/dummy-adapter/src/change_validator.ts
```diff
@@ -14,17 +14,18 @@
...
 
+const { createChangeValidator } = deployment.changeValidators
 
 export const changeValidator = (
   config: GeneratorParams,
 ): ChangeValidator => {
-  const validators: ChangeValidator[] = [
-    fromAdapterConfig(config),
-  ]
+  const validators: Record<string, ChangeValidator> = {
```
netama, line 26: why do we need the cast?

packages/jira-adapter/src/change_validators/index.ts
```diff
@@ -57,53 +56,57 @@ import { unresolvedFieldConfigurationItemsValidator } from './unresolved_field_c
...
   deployTypesNotSupportedValidator,
+  createChangeValidator,
 } = deployment.changeValidators
 
 
 export default (
   client: JiraClient, config: JiraConfig, paginator: clientUtils.Paginator
 ): ChangeValidator => {
-  const validators: ChangeValidator[] = [
+  const validators: Record<string, ChangeValidator> = {
```
netama, line 66: same question re the cast

packages/okta-adapter/src/config.ts
```diff
@@ -44,6 +47,7 @@ export type OktaConfig = {
   [FETCH_CONFIG]: OktaFetchConfig
   [API_DEFINITIONS_CONFIG]: OktaSwaggerApiConfig
   [PRIVATE_API_DEFINITIONS_CONFIG]: OktaDuckTypeApiConfig
+  [DEPLOY_CONFIG]?: ChangeValidatorConfig
```
netama, line 50: why is this called `ChangeValidatorConfig` if this is the deploy config?

packages/sap-adapter/src/config.ts
```diff
@@ -39,6 +43,7 @@ export type SAPConfig = {
   [CLIENT_CONFIG]?: SAPClientConfig
   [FETCH_CONFIG]: SAPFetchConfig
   [API_DEFINITIONS_CONFIG]: SAPApiConfig
+  [DEPLOY_CONFIG]?: ChangeValidatorConfig
```
netama, line 46: we shouldn't add deploy config for read-only adapters, let's just assume nothing is disabled

packages/zendesk-adapter/src/change_validator.ts
```diff
@@ -93,65 +93,69 @@ export default ({
   typesDeployedViaParent: string[]
   typesWithNoDeploy: string[]
 }): ChangeValidator => {
-  const validators: ChangeValidator[] = [
+  const validators: Record<string, ChangeValidator> = {
```
netama, line 96: will go over the names later, I think we should make some adjustments here to make sure these are clear (probably in other adapters as well)

netama (commented):
cool :) added some comments on the shared implementation + read-only adapters, will do another pass on the validator names later

packages/adapter-components/src/deployment/change_validators/skip_parents_of_skipped_instances.ts
dantal4: In NS we run most of the validators in parallel, and two validators that needs to know which elements have changeErrors are running after them.
Maybe we can change it to run one after another in the given order, and pass each validator the changeErrors that were collected until then.

packages/netsuite-adapter/src/config.ts
idoser96: about 1, this might be a good idea and make the types easier
about 2 - currently this config is not user facing and there is an error log in case of a typo, in my opinion implementing this safety is currently an overkill

packages/adapter-components/src/deployment/change_validators/skip_parents_of_skipped_instances.ts
ori-moisis: for reference - in an object that only has string keys, insertion order should be maintained according to [this part of the spec](https://tc39.es/ecma262/#sec-ordinaryownpropertykeys) that deals with getting the keys of an object:
> 1. Let keys be a new empty [List](https://tc39.es/ecma262/#sec-list-and-record-specification-type).
> 2. For each own [property key](https://tc39.es/ecma262/#sec-object-type) P of O such that P is an [array index](https://tc39.es/ecma262/#array-index), in ascending numeric index order, do
a. Append P to keys.
> 3. For each own [property key](https://tc39.es/ecma262/#sec-object-type) P of O such that P [is a String](https://tc39.es/ecma262/#sec-ecmascript-language-types-string-type) and P is not an [array index](https://tc39.es/ecma262/#array-index), in ascending chronological order of property creation, do
a. Append P to keys.
> 4. For each own [property key](https://tc39.es/ecma262/#sec-object-type) P of O such that P [is a Symbol](https://tc39.es/ecma262/#sec-ecmascript-language-types-symbol-type), in ascending chronological order of property creation, do
a. Append P to keys.
> 5. Return keys.

packages/netsuite-adapter/src/config.ts
ori-moisis: I think validation (and, more importantly, auto complete) is important here given that there will be no feedback for typos.
in the SF this was not a map type, it was an object type that had all the names so it was fully validated.
is there a reason to change this?

packages/okta-adapter/src/config.ts
idoser96: To avoid creating another type that contains nothing but ChangeValidatorsConfig

packages/netsuite-adapter/src/config.ts
idoser96: currently the code error logs in case an adapter that doesn't exists is in the configuration
i can change the log to an exception that stops the run (like in SF previously)

In my opinion keeping a list of CV names (like it is on salesforce today) on each adapter is a bit messy, and the only difference is if the users sees he has a typo before the deploy or at the start of the deploy

ori-moisis: I think it is critical, if this is not done in this new mechanism, we should at least keep the salesforce implementation separate so we don't get a regression there.
I really don't want to go back to the days where we had no validation on the adapter config and every time someone made a typo when trying to disable a validator the issue became a blocker and looked like a bug

netama: I agree with @ori-moisis this will be important in all adapters

alonstern: If we change it to a list we can't override a single validator using config overrides. So if we can have the validation and keep it as a map that would be better IMO

packages/adapter-components/src/deployment/change_validators/create_change_validator.ts
```diff
@@ -0,0 +1,43 @@
...
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+import { ChangeValidator } from '@salto-io/adapter-api'
+import _ from 'lodash'
+import { logger } from '@salto-io/logging'
+
+const log = logger(module)
+
+export type ValidatorsActivationConfig = {}
```
alonstern, line 22: That's probably not the right type here. It will match any value that is not null

alonstern (approved):
Added one comment, LGTM besides that, if all the relevant people went over the code I think you can merge

packages/adapter-components/src/deployment/change_validators/create_change_validator.ts
idoser96: The right type is Record<string, boolean>, how ever this causes type issues while creating ObjectType
After talking to @ori-moisis it was decided to keep it like that as the type check also exists later in the code

I will add an explanation in the code

ori-moisis: the alternatives would be to either change `createMatchingObjectType` or to not use it in this flow.
this was the quickest fix possible

netama (commented):
zendesk LGTM

packages/salesforce-adapter/src/adapter_creator.ts
```diff
@@ -96,20 +99,17 @@ SalesforceConfig => {
...
-      const validValidatorsNames = Object.keys(changeValidators)
-      Object.entries(validators as {}).forEach(([key, value]) => {
-        if (!validValidatorsNames.includes(key)) {
-          throw new ConfigValidationError(['validators', key], `Validator ${key} does not exist, expected one of ${validValidatorsNames.join(',')}`)
-        }
-        if (!_.isBoolean(value)) {
-          throw new ConfigValidationError(['validators', key], 'Value must be true or false')
-        }
-      })
+      Object.entries(validators ?? {})
```
ori-moisis, line 107: We should not remove the validation on which validators exist

packages/salesforce-adapter/src/types.ts
```diff
@@ -714,8 +721,11 @@ export const configType = createMatchingObjectType<SalesforceConfig>({
     [CLIENT_CONFIG]: {
       refType: clientConfigType,
     },
-    validators: {
-      refType: changeValidatorConfigType,
+    [DEPLOY_CONFIG]: {
+      refType: createUserDeployConfigType(SALESFORCE, changeValidatorConfigType),
+    },
+    [VALIDATE_CONFIG]: {
+      refType: createUserDeployConfigType(SALESFORCE, changeValidatorConfigType),
```
ori-moisis, line 728: do we actually need a separate config for validate?
is there a situation where we could want to disable a validator for one of these and not the other?

packages/adapter-components/src/deployment/change_validators/create_change_validator.ts
alonstern: Where do we use this type with `createMatchingObjectType`? Isn't it always with adapter specific types that includes the adapter change validators?

packages/adapter-components/src/config/shared.ts
```diff
@@ -66,6 +67,10 @@ export type UserFetchConfig<T extends Record<string, unknown> | undefined = Defa
 }
 
 export type UserDeployConfig = {
+  changeValidators?: ValidatorsActivationConfig
+}
+
+export type DefaultMissingUserFallbackConfig = {
```
shir-reifenberg, line 73: Maybe i'm out of context, but why not add this as another optional field to `UserDeployConfig`?

shir-reifenberg (commented):
Okta LGTM, added one question :)

packages/salesforce-adapter/src/types.ts
idoser96: The default today is for 'omitData to be disabled on deploy and 'dataChange' to be disabled on validation

if this is no longer true i cant change the default config and remove 'validate' from the configuration

packages/adapter-components/src/config/shared.ts
idoser96: Because UserDeployConfig is now being used by all adapters with deploy option, and defaultMissingUserFallback is only relevant to zendesk and jira (and soon okta)

packages/salesforce-adapter/src/types.ts
ori-moisis: this is not what I asked
I know there are some validators that we need to run only on validate and not on deploy as part of the **code**.
what I asked is why we need to have separate **configuration**

packages/adapter-components/src/deployment/change_validators/create_change_validator.ts
idoser96: createUserDeployConfigType in shared.ts

but now i think the better idea is to keep 
export type ValidatorsActivationConfig = Record<string, boolean>

and do this:
// Record<string, boolean> doesn't pass type check for refType of ObjectType
      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
      // @ts-ignore
      changeValidators: { refType: changeValidatorsType },

packages/netsuite-adapter/src/change_validator.ts
```diff
@@ -46,46 +46,48 @@ import currencyUndeployableFieldsValidator from './change_validators/currency_un
...
+import { AdditionalDependencies, ChangeValidatorName } from './config'
 import { Filter } from './filter'
 import { NetsuiteChangeValidator } from './change_validators/types'
 
+type ValidatorsActivationConfig = deployment.changeValidators.ValidatorsActivationConfig
+const { createChangeValidator } = deployment.changeValidators
+
+const defaultChangeValidators = deployment.changeValidators.getDefaultChangeValidators()
+
+const netsuiteChangeValidators: Partial<Record<ChangeValidatorName, NetsuiteChangeValidator>> = {
```
dantal4, line 58: Can you split `ChangeValidatorName` to `NetsuiteValidatorName`, `NonSuiteAppValidatorName`, `OnlySuiteAppValidatorName`, so we avoid using `Partial` and casting?

dantal4 (commented):
Looks great on NS side 👍

packages/adapter-components/src/config/shared.ts
idoser96: Talked f2f

packages/adapter-components/src/deployment/change_validators/create_change_validator.ts
idoser96: Talked f2f

tamtamirr (dismissed):
Salesforce LGTM code wise. Two things that needs to be done prior to merging IMO:
1. Do we have a **PROD** ticket to communicate this change to our existing Salesforce Customers? Telling them we are going to change the way we disable Change Validators.

2. I think we should add **User Notification** for existing Salesforce Workspaces:
- We've changed the way to disable Change Validators in **salto**. If you need any help migrating your existing **Salesforce.nacl** with these changes, feel free to contact one of the maintainers.

cc. @amitgoldarb @ori-moisis @adamro @tomermevorach please keep me honest here 🙏

alonstern (approved):
logo

PR #4560, opened 2023-06-28
SALTO-3233: add second global context change validator - Jira

shayc331:
_add second global context change validator_

---

_Release Notes_: 
Jira Adapter:
* add validation that prevents the user from adding more than one global context to custom fields

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/jira-adapter/src/change_validators/field_contexts/field_contexts.ts
+++ b/packages/jira-adapter/src/change_validators/field_contexts/field_contexts.ts
@@ -26,7 +26,7 @@ import { getGlobalContextsUsedInProjectErrors } from './referenced_global_contex
 const { awu } = collections.asynciterable
 const log = logger(module)
 
-export const getFieldContexts = async (
+const getFieldContexts = async (
   field: InstanceElement,
   elementSource: ReadOnlyElementsSource,
 ): Promise<InstanceElement[]> =>
--- a/packages/jira-adapter/src/change_validators/field_contexts/field_contexts.ts
+++ b/packages/jira-adapter/src/change_validators/field_contexts/field_contexts.ts
@@ -26,7 +26,7 @@ import { getGlobalContextsUsedInProjectErrors } from './referenced_global_contex
 const { awu } = collections.asynciterable
 const log = logger(module)
 
-const getFieldContexts = async (
+export const getFieldContexts = async (
   field: InstanceElement,
   elementSource: ReadOnlyElementsSource,
 ): Promise<InstanceElement[]> =>
--- a/packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
+++ b/packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
@@ -13,57 +13,90 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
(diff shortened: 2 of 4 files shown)


packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
```diff
@@ -0,0 +1,69 @@
...
+import { getFieldContexts } from './field_contexts'
+
+
+const { awu } = collections.asynciterable
+
+export const fieldSecondGlobalContextValidator: ChangeValidator = async (changes, elementSource) => {
+  if (elementSource === undefined) {
+    return []
+  }
+  const fieldToGlobalContextsCount: Record<string, number> = {}
```
shayc331, line 29: note: to handle the scenario of adding two global contexts to the same field when the field does not have them in the `contexts` list

packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
```diff
@@ -0,0 +1,69 @@
...
+  return awu(changes)
+    .filter(isInstanceChange)
+    .filter(isAdditionOrModificationChange)
+    .map(getChangeData)
+    .filter(instance => instance.elemID.typeName === FIELD_CONTEXT_TYPE_NAME)
+    .map(instance => ({ context: instance, field: getParent(instance) }))
+    .filter(async contextAndField => {
+      const globalContexts = (await getFieldContexts(contextAndField.field, elementSource))
+        .filter(context => context.value.isGlobalContext)
+      const isNewContextIncluded = globalContexts
```
shayc331, line 54: note: 
note: to handle the scenario of adding global contexts to a field when the field does not have a reference to it in the contexts list

IdoZakk:
I think you can simplify it with the following:
1. Create a map of field elem id to context elem id
2. Itarate over the global context changes (once)
3a. If there is a global context in the element source add to map
3b. if the map contains the parent field and the context elem ID is different from the current issue a warning, else add it to the map

packages/jira-adapter/src/change_validators/field_contexts/field_contexts.ts
```diff
@@ -26,7 +26,7 @@ import { getGlobalContextsUsedInProjectErrors } from './referenced_global_contex
 const { awu } = collections.asynciterable
 const log = logger(module)
 
-const getFieldContexts = async (
+export const getFieldContexts = async (
```
alonstern, line 29: nit: this doesn't need to be exported anymore

packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
```diff
@@ -0,0 +1,84 @@
...
+  return field
+}
+
+export const fieldSecondGlobalContextValidator: ChangeValidator = async (changes, elementSource) => {
+  if (elementSource === undefined) {
+    log.error('Failed to run fieldSecondGlobalContextValidator because element source is undefined')
+    return []
+  }
+  const fieldToGlobalContextCount: Record<string, number> = {}
+  await awu(await elementSource.getAll())
```
alonstern, line 53: Performance nit: do this only if there are changes to global contexts

alonstern (approved):
👌

PR #4561, opened 2023-06-28
SALTO-4294: write workaround to a zendesk API limitation that inline article attachments are not deleted properly

edenhassid:
write workaround to a zendesk API limitation that inline article attachments are not deleted properly

---

---
_Release Notes_: 
zendesk:
* add workaround to a zendesk API limitation that inline article attachments are not deleted properly

---
_User Notifications_: 
zendesk:
* Certain attachments may be deleted. Previously, these deleted attachments continued to appear in Salto, despite being removed from the Zendesk UI due to zendesk API limitation. This issue is now fixed by adding a workaround.

### DIFF
--- a/packages/zendesk-adapter/src/filters/article/article.ts
+++ b/packages/zendesk-adapter/src/filters/article/article.ts
@@ -18,24 +18,34 @@ import { logger } from '@salto-io/logging'
 import { collections } from '@salto-io/lowerdash'
 import {
   AdditionChange,
-  Change, ElemID, getChangeData, InstanceElement, isAdditionChange, isModificationChange,
-  isAdditionOrModificationChange, isInstanceElement, isReferenceExpression, ReadOnlyElementsSource,
-  isRemovalChange, ModificationChange, ReferenceExpression, Element, CORE_ANNOTATIONS, isObjectType,
+  Change,
+  CORE_ANNOTATIONS,
+  Element,
+  ElemID,
+  getChangeData,
+  InstanceElement,
+  isAdditionChange,
+  isAdditionOrModificationChange,
+  isInstanceElement,
+  isModificationChange,
+  isObjectType,
+  isReferenceExpression,
+  isRemovalChange,
+  ModificationChange,
+  ReadOnlyElementsSource,
+  ReferenceExpression,
 } from '@salto-io/adapter-api'
-import {
-  createSchemeGuard,
-  getParents,
-  resolveChangeElement,
-} from '@salto-io/adapter-utils'
+import { createSchemeGuard, getParents, resolveChangeElement } from '@salto-io/adapter-utils'
 import Joi from 'joi'
 import { FilterCreator } from '../../filter'
 import { deployChange, deployChanges } from '../../deployment'
 import {
-  ARTICLE_TYPE_NAME,
   ARTICLE_ATTACHMENT_TYPE_NAME,
+  ARTICLE_ATTACHMENTS_FIELD, ARTICLE_TRANSLATION_TYPE_NAME,
+  ARTICLE_TYPE_NAME,
+  EVERYONE_USER_TYPE,
   USER_SEGMENT_TYPE_NAME,
   ZENDESK,
-  EVERYONE_USER_TYPE, ARTICLE_ATTACHMENTS_FIELD,
 } from '../../constants'
 import { addRemovalChangesId, isTranslation } from '../guide_section_and_category'
 import { lookupFunc } from '../field_references'
@@ -44,7 +54,8 @@ import ZendeskClient from '../../client/client'
 import {
   createUnassociatedAttachment,
   deleteArticleAttachment,
-  getArticleAttachments, isAttachments,
+  getArticleAttachments,
+  isAttachments,
   updateArticleTranslationBody,
 } from './utils'
 import { API_DEFINITIONS_CONFIG } from '../../config'
@@ -235,12 +246,12 @@ const handleArticleAttachmentsPreDeploy = async ({ changes, client, elementsSour
 }
 
 const getId = (instance: InstanceElement): number => instance.value.id
-const getName = (element: InstanceElement | ReferenceExpression): string => element.elemID.name
+const getName = (instanceOrRef: InstanceElement | ReferenceExpression): string => instanceOrRef.elemID.name
 const getFilename = (attachment: InstanceElement | undefined): string => attachment?.value.file_name
 const getContentType = (attachment: InstanceElement | undefined): string => attachment?.value.content_type
 const getInline = (attachment: InstanceElement | undefined): boolean => attachment?.value.inline
-const shouldRemoveAttachment = (attachment: Element, allRemovedAttachmentsIds: Set<number>): boolean =>
-  isInstanceElement(attachment) && allRemovedAttachmentsIds.has(attachment.value.id)
+const shouldRemoveAttachment = (attachment: Element | undefined, removedAttachmentsIds: Set<number>): boolean =>
+  isInstanceElement(attachment) && removedAttachmentsIds.has(attachment.value.id)
 
 const isRemovedAttachment = (
   attachment: unknown,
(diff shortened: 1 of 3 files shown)


packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -234,11 +235,69 @@ const handleArticleAttachmentsPreDeploy = async ({ changes, client, elementsSour
...
+const getAttachmentData = (
+  article: InstanceElement,
+  attachmentByName: Record<string, InstanceElement>
+): {inlineAttachmentsNameById: Record<number, string>; attachmentIdsFromArticleBody: Set<number | undefined>} => {
+  const attachmentsNames: string[] = article.value.attachments?.filter(isReferenceExpression).map(getName)
+  const inlineAttachmentInstances = attachmentsNames
+    .map(name => attachmentByName[name])
+    .filter(isInstanceElement)
+    .filter(attachment => attachment.value.inline)
+  const inlineAttachmentsNameById = _.mapValues(_.keyBy(inlineAttachmentInstances, getId), getName)
```
netama, line 265: to be on the safe side we should only use instances where `getId()` is defined

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -54,6 +54,7 @@ const log = logger(module)
 const { awu } = collections.asynciterable
 
 const USER_SEGMENT_ID_FIELD = 'user_segment_id'
+const ATTACHMENTS_IDS_REGEX = new RegExp(`(?<url>/${ARTICLE_ATTACHMENTS_FIELD}/)(?<id>\\d+)`, 'g')
```
netama, line 57: this is based on the `ELEMENTS_REGEXES` mapping from `article_body`, right? can we reuse it to ensure these remain aligned?

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -234,11 +235,69 @@ const handleArticleAttachmentsPreDeploy = async ({ changes, client, elementsSour
...
+
+/**
+ * returns deleted attachments ids
+ */
+const calculateAndRemoveDeletedAttachments = (
+  articleInstances: InstanceElement[],
+  attachmentByName: Record<string, InstanceElement>
+): Set<number> => {
+  const allRemovedAttachmentsIds = new Set<number>()
+  const allRemovedAttachmentsNames = new Set<string>()
```
netama, line 282: I think we can aggregate only `allRemovedAttachmentsNames` and map to `allRemovedAttachmentsIds` at the end using `attachmentByName`, right?

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -268,6 +327,16 @@ const filterCreator: FilterCreator = ({ config, client, elementsSource, brandIdT
...
       )
+      const allRemovedAttachmentsIds = calculateAndRemoveDeletedAttachments(articleInstances, attachmentByName)
+      _.remove(
+        attachments,
+        (attachment => shouldRemoveAttachment(attachment, allRemovedAttachmentsIds))
+      )
+      _.remove(
+        elements,
+        (element => element.elemID.typeName === ARTICLE_ATTACHMENT_TYPE_NAME
+          && shouldRemoveAttachment(element, allRemovedAttachmentsIds))
```
netama, line 338: 1. I think if we already made the assumption earlier that we can index the attachments by their elem id, then we can also remove them everywhere based on this, no?
2. potential future edge case here - if there will be a way to share attachments across articles, we might end up creating unresolved references here - since we remove the reference from the article inside `calculateAndRemoveDeletedAttachments`, and remove here any element that is "not referenced" at least in one article. so if articles A and B referenced an attachment and it became orphaned in A, it will be removed from the fetch even though B still references it. not sure it's worth addressing now since currently attachments have a specific article as their parent, but maybe we should add some comment about it (maybe also next to the attachment part of the `config.ts`)?

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -234,11 +235,69 @@ const handleArticleAttachmentsPreDeploy = async ({ changes, client, elementsSour
...
+  articleInstances.forEach(article => {
+    const articleRemovedAttachmentsIds = new Set<number>()
+    const attachmentData = getAttachmentData(article, attachmentByName)
+    Object.keys(attachmentData.inlineAttachmentsNameById).forEach(id => {
+      const numberId = Number(id)
+      if (!attachmentData.attachmentIdsFromArticleBody.has(numberId)) {
+        allRemovedAttachmentsIds.add(numberId)
+        articleRemovedAttachmentsIds.add(numberId)
+        allRemovedAttachmentsNames.add(attachmentData.inlineAttachmentsNameById[numberId])
+      }
```
netama, line 292: since we already modify the article body here I think it makes sense to keep the `forEach` and not switch to a `map` (which would "hide" the fact there is inline modification) - but I think it might still be slightly cleaner to extend the `allX` sets once at the end and not one-by-one, wdyt?
(hopefully as I wrote in other comments we can maintain only the `Names` one, but maybe I'm missing something)

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -234,11 +235,69 @@ const handleArticleAttachmentsPreDeploy = async ({ changes, client, elementsSour
...
+    })
+    article.value.attachments = article.value.attachments?.filter(
+      (attachment: unknown) => !isRemovedAttachment(attachment, articleRemovedAttachmentsIds, attachmentByName)
+    )
+  })
+  log.info(`the article attachments ${Array.from(allRemovedAttachmentsNames)} are going to be deleted since they are inline but do not appear in the body`)
```
netama, line 298: nit: this list can be long, so might be better to finish the sentence first - 
```suggestion
  log.info(`the following article attachments are going to be deleted, since they are inline but do not appear in the body: ${Array.from(allRemovedAttachmentsNames)} `)
```
(also - maybe "not going to be included in the fetch" rather than "deleted", just to avoid unnecessary alarm for in the future? :) )

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -234,11 +235,69 @@ const handleArticleAttachmentsPreDeploy = async ({ changes, client, elementsSour
...
+  const articleBody = article.value.body ?? ''
+  const attachmentIdsFromArticleBody = new Set(
+    Array.from(articleBody.matchAll(ATTACHMENTS_IDS_REGEX), (match: RegExpMatchArray | undefined) => match?.groups?.id)
+      .filter(id => id !== undefined).map(Number)
+  )
+  return { inlineAttachmentsNameById, attachmentIdsFromArticleBody }
+}
+
+/**
+ * returns deleted attachments ids
```
netama, line 275: since this is due to a zendesk bug, I think it's worth adding some more details here about the scenario and why this code is needed - e.g. something like
> due to an API limitation on the zendesk side, inline attachments deleted from <...> are still returned in the call to <> even though they no longer exist. we do not want to include these orphaned attachments in the fetch results, so we omit inline attachments that are not referenced from the article body

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -234,11 +235,69 @@ const handleArticleAttachmentsPreDeploy = async ({ changes, client, elementsSour
...
+  article: InstanceElement,
+  attachmentByName: Record<string, InstanceElement>
+): {inlineAttachmentsNameById: Record<number, string>; attachmentIdsFromArticleBody: Set<number | undefined>} => {
+  const attachmentsNames: string[] = article.value.attachments?.filter(isReferenceExpression).map(getName)
+  const inlineAttachmentInstances = attachmentsNames
+    .map(name => attachmentByName[name])
+    .filter(isInstanceElement)
+    .filter(attachment => attachment.value.inline)
+  const inlineAttachmentsNameById = _.mapValues(_.keyBy(inlineAttachmentInstances, getId), getName)
+  const articleBody = article.value.body ?? ''
```
netama, line 266: don't we need to look at all the translation bodies? (what happens if an attachment is referenced from a no-default translation, won't we accidentally remove it here?)

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -234,11 +235,69 @@ const handleArticleAttachmentsPreDeploy = async ({ changes, client, elementsSour
 }
 
 const getId = (instance: InstanceElement): number => instance.value.id
-const getName = (instance: InstanceElement): string => instance.elemID.name
+const getName = (element: InstanceElement | ReferenceExpression): string => element.elemID.name
 const getFilename = (attachment: InstanceElement | undefined): string => attachment?.value.file_name
 const getContentType = (attachment: InstanceElement | undefined): string => attachment?.value.content_type
 const getInline = (attachment: InstanceElement | undefined): boolean => attachment?.value.inline
+const shouldRemoveAttachment = (attachment: Element, allRemovedAttachmentsIds: Set<number>): boolean =>
+  isInstanceElement(attachment) && allRemovedAttachmentsIds.has(attachment.value.id)
```
netama, line 243: nit: I think we can probably check `isInstanceElement` outside this function in the one place that might pass a non-instance element

netama (commented):
looks good :) added some nits and a couple of "real" comments
I think it might be worth elaborating a bit more in the PR description (specifically clarifying we are working around a zendesk API limitation)

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -234,11 +235,69 @@ const handleArticleAttachmentsPreDeploy = async ({ changes, client, elementsSour
...
+  const attachmentInstance = name ? attachmentByName[name] : undefined
+  return isInstanceElement(attachmentInstance)
+    && articleRemovedAttachmentsIds.has(attachmentInstance.value.id)
+}
+
+const getAttachmentData = (
+  article: InstanceElement,
+  attachmentByName: Record<string, InstanceElement>
+): {inlineAttachmentsNameById: Record<number, string>; attachmentIdsFromArticleBody: Set<number | undefined>} => {
+  const attachmentsNames: string[] = article.value.attachments?.filter(isReferenceExpression).map(getName)
```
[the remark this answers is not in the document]
edenhassid: it doesn't work without the casting, it doesn't know that its a list of strings

packages/zendesk-adapter/src/filters/article/article.ts
edenhassid: attachment by name is per article, I didn't want to add a log for each article separately, do you think it is better?

packages/zendesk-adapter/src/filters/article/article.ts
edenhassid: the regex there is not global so it cannot be used

packages/zendesk-adapter/src/filters/article/article.ts
edenhassid: 1. I don't understand what you mean
2. 👍

netama (approved):
did a quick pass, looks good, approving to unblock

PR #4562, opened 2023-06-28
SALTO-4385 add more log.time to workspace package

benisraeladi:
Add log.time to more functions in workspace package for better granularity when trying to find performance bottlenecks

---

---
_Release Notes_: 
_None_
---
_User Notifications_: 
_None_


packages/workspace/src/workspace/workspace.ts
```diff
@@ -820,39 +820,40 @@ export const loadWorkspace = async (
...
-      } },
-      validate })
-    return {
-      naclFilesChangesCount: Object.values(workspaceChanges)
-        .map(changeSet => changeSet.changes)
-        .flat().length,
-      stateOnlyChangesCount: stateOnlyChanges.length,
-    }
-  }
+    }) : Promise<UpdateNaclFilesResult> =>
```
tomsellek, line 823: [OPINION WARNING]: I think this function is large enough that the reader may lose context and be confused by the last line of `}, 'workspace.updateNaclFiles')`. I'd move the actual function to a separate `...Impl` function

tomsellek (approved):
Seems like this would very helpful. Made one comment for your consideration.

packages/workspace/src/serializer/elements.ts
```diff
@@ -238,9 +240,9 @@ export const serialize = async <T = Element>(
   elements: T[],
   referenceSerializerMode: 'replaceRefWithValue' | 'keepRef' = 'replaceRefWithValue',
   storeStaticFile?: (file: StaticFile) => Promise<void>
-): Promise<string> => (
+): Promise<string> => log.time(async () => (
   (await awu(await serializeStream(elements, referenceSerializerMode, storeStaticFile)).toArray()).join('')
-)
+), 'workspace.serialize')
```
ori-moisis, line 245: did you run this with any real workload?

I am think this could generate a very (very) large number of logs, to the point where it would affect performance in a noticeable way (remember - JS only has one thread...)

benisraeladi: `serialize` and `deserialize` were extremely spammy, so I removed logging from them

ori-moisis:
`dumpElements` may be a problem on first fetch with a large account since it will be called on each element separately.
`parse` may be a problem when regenerating a cache (try deleting your workspace’s cache and running any command).
`getSourceRange` might be an issue if a workspace has a lot of errors, not sure how to simulate that, but some workspaces have a lot of errors

generally, I think the red flags here are functions that get called per-element or multiple times per element, I don’t think there is much to gain from logging something per-element, the aggregate is probably more interesting anyway, and you’d probably want to know which flow caused the function to be called so many times rather than seeing that each invocation took 0.0001 ms

PR #4564, opened 2023-06-29
SALTO-3380 - fix open in app

Shulik95:
_constant urls will be added to InstanceElements and not only ObjectTypes_

---

_None_

---
_Release Notes_: 
_Netsuite Adapter_:
* constant urls will be added to instances and not only ObjectTypes

---
_User Notifications_: 
_None_


omrilit (approved):
LGTM!

PR #4565, opened 2023-07-02
SALTO-3791 Support deletion in partial fetch - core side

kami-delaroz:
Support deletion in partial fetch - core side (NetSuite side was done already)

---

None

---
_Release Notes_: 
Support deletion in partial fetch (**NetSuite** only currently)
Support deletion of standard instances, custom types and custom records. **No** support for file cabinet

---
_User Notifications_: 
None

### DIFF
--- a/packages/core/src/core/fetch.ts
+++ b/packages/core/src/core/fetch.ts
@@ -67,7 +67,6 @@ export const toAddFetchChange = (elem: Element): FetchChange => {
   return { change, serviceChanges: [change], metadata: getFetchChangeMetadata(elem) }
 }
 
-
 export class StepEmitter<T = void> extends EventEmitter<StepEvents<T>> {}
 
 export type FetchProgressEvents = {
@@ -111,6 +110,10 @@ type WorkspaceDetailedChange = {
 type DetailedChangeTreeResult = {
     changesTree: collections.treeMap.TreeMap<WorkspaceDetailedChange>
     changes: DetailedChange[]
+}
+
+type PartiallyFetchedAccountData = {
+  deletedElements?: Set<string>
 }
 
 const getDetailedChangeTree = async (
@@ -320,28 +323,28 @@ const runPostFetch = async ({
   adapters,
   accountElements,
   stateElementsByAccount,
-  partiallyFetchedAccounts,
+  partiallyFetchedAccountData,
   accountToServiceNameMap,
   progressReporters,
 }: {
   adapters: Record<string, AdapterOperationsWithPostFetch>
   accountElements: Element[]
   stateElementsByAccount: Record<string, ReadonlyArray<Element>>
-  partiallyFetchedAccounts: Set<string>
+  partiallyFetchedAccountData: Map<string, PartiallyFetchedAccountData>
   accountToServiceNameMap: Record<string, string>
   progressReporters: Record<string, ProgressReporter>
 }): Promise<void> => {
   const serviceElementsByAccount = _.groupBy(accountElements, e => e.elemID.adapter)
   const getAdapterElements = (accountName: string): ReadonlyArray<Element> => {
-    if (!partiallyFetchedAccounts.has(accountName)) {
+    if (!partiallyFetchedAccountData.has(accountName)) {
       return serviceElementsByAccount[accountName] ?? stateElementsByAccount[accountName]
     }
     const fetchedIDs = new Set(
       serviceElementsByAccount[accountName].map(e => e.elemID.getFullName())
     )
-    const missingElements = stateElementsByAccount[accountName].filter(
-      e => !fetchedIDs.has(e.elemID.getFullName())
-    )
+    const missingElements = stateElementsByAccount[accountName]
+      .filter(e => !fetchedIDs.has(e.elemID.getFullName()))
+      .filter(e => !partiallyFetchedAccountData.get(accountName)?.deletedElements?.has(e.elemID.getFullName()))
     return [
       ...serviceElementsByAccount[accountName],
       ...missingElements,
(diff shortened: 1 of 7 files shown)


packages/core/src/core/fetch.ts
```diff
@@ -320,28 +323,28 @@ const runPostFetch = async ({
...
+    if (!partiallyFetchedAccountData.has(accountName)) {
       return serviceElementsByAccount[accountName] ?? stateElementsByAccount[accountName]
     }
     const fetchedIDs = new Set(
       serviceElementsByAccount[accountName].map(e => e.elemID.getFullName())
     )
-    const missingElements = stateElementsByAccount[accountName].filter(
-      e => !fetchedIDs.has(e.elemID.getFullName())
-    )
+    const missingElements = stateElementsByAccount[accountName]
```
ori-moisis, line 345: do we actually still need these elements?
I think since @idoser96 's PR we shouldn't need these in the response anymore

packages/core/src/core/fetch.ts
```diff
@@ -320,28 +323,28 @@ const runPostFetch = async ({
   adapters,
   accountElements,
   stateElementsByAccount,
-  partiallyFetchedAccounts,
+  partiallyFetchedAccountData,
   accountToServiceNameMap,
   progressReporters,
 }: {
   adapters: Record<string, AdapterOperationsWithPostFetch>
   accountElements: Element[]
```
adamro, line 331: I wonder if we want/need a bigger by account type here. 

a Type such as (let's call it `accountInput` / `accountData`)
```
{
   serviceName: string
   elements: Element[]
   stateElements: ReadonlyArray<Element>
   partiallyFetchedAccountData: PartiallyFetchedAccountData
}
```

and replace `accountElements`, `stateElementsByAccount`, `partiallyFetchedAccountData`, `accountToServiceNameMap` with 
`accountInputs: Record<string, accountInput>`

Cause basically everything here should be by account (the `accountElements` that are not grouped by account in the input are actually grouped in the first line of the function)

packages/core/src/core/fetch.ts
kami-delaroz: @ori-moisis as I understood the flow, this is the place where we generate the full element list of the env, for post-fetch. we already have the changed elements, and we add the rest from the state (except for the deletions). I'm not sure it's relevant to @idoser96 's change?

adamro (commented):
Mid way but I have a comment that might change a lot of params in a lot of functions so putting it early :D

packages/core/src/core/fetch.ts
ori-moisis: you are right, I didn't realize this is just for postFetch, this seemed like it is going out to the fetch result

packages/core/src/core/fetch.ts
```diff
@@ -113,6 +112,10 @@ type DetailedChangeTreeResult = {
     changes: DetailedChange[]
 }
 
+type PartiallyFetchedAccountData = {
+  deletedElements?: Set<string>
```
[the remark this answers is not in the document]
kami-delaroz: I'll use it only for `deletedElements.has(elemId.fullName)` so I thought it's better to generate it in the format I need as early as possible, to avoid extra processing in separated places.
But I think it's odd for the adapter to return only elem names, instead of the full elemId, as it assumes things about the way it will be handled in the other hand.

packages/core/src/core/fetch.ts
```diff
@@ -708,14 +716,14 @@ type CreateFetchChangesParams = {
   processErrorsResult: ProcessMergeErrorsResult
   currentConfigs: InstanceElement[]
   getChangesEmitter: StepEmitter
-  partiallyFetchedAccounts?: Set<string>
+  partiallyFetchedAccountData: Map<string, PartiallyFetchedAccountData>
```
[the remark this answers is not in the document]
kami-delaroz: yep

packages/core/src/core/fetch.ts
kami-delaroz: The accountElements are grouped by adapter, that's why I left the flatten we have earlier (that flat the results-by-account we received). I guess in some edge cases I'll have a different behavior now?

adamro: They are grouped by service, the `.adapter` in ElemID is a lie :(

ori-moisis: nit: (about the suggestion that includes "serviceName: string") - we don't use the term `service` anymore, it was supposed to be replaced by `account`, right?

kami-delaroz: I think serviceName stands for adapter? @adamro

adamro: Yep
Should be `adapterName` and not `serviceName` I guess :)

packages/core/src/core/fetch.ts
```diff
@@ -483,11 +486,13 @@ const fetchAndProcessMergeErrors = async (
...
-        .filter(result => result.isPartial)
-        .map(result => result.accountName)
-    )
+    const partiallyFetchedAccountData = new Map(fetchResults
+      .filter(result => result.partialFetchData?.isPartial ?? false)
+      .map(result => ({
+        accountName: result.accountName,
+        deletedElements: new Set(result.partialFetchData?.deletedElements?.map(elem => elem.getFullName())),
+      }))
+      .map(data => [data.accountName, { deletedElements: data.deletedElements }]))
```
adamro, line 495: nit: why do you need the extra map here? (you can map it to an array in the first map)

packages/core/test/core/fetch.test.ts
```diff
@@ -1986,3 +2029,37 @@ describe('fetch from workspace', () => {
     })
   })
 })
+
+// TODO: only deletion scenarios are covered here. The rest are under “fetchChanges” flows, and should be moved here
```
adamro, line 2033: Do? :)

adamro (commented):
Waiting for the next iteration with the param changes but logic wise I think we are good

packages/core/test/core/fetch.test.ts
kami-delaroz: will open a ticket

adamro (approved):
Approving but please open and attach the ticket with the params change that is supposed to happen immediately after :)

PR #4567, opened 2023-07-02
SALTO-3829: Add reference to workflow's transitions

IdoZakk:
_Stage 1: change the workflow transitions list to a map_

---

_Currently the tests do not pass as I did not touch the deployment_

---
_Release Notes_: 
Jira Adapter:
* Add references to workflow transitions in scriptrunner's relevant post functions

---
_User Notifications_: 
Jira Adapter:
* Workflow transitions will be kept in a map from now on. The key will be based on the name of the transition and the `from` field


alonstern:
Please update the descriptions and add an example of how the workflow will look after this.
Also I think we should split it into at least one PR for the workflow structure change and another PR for the deployment of the script runner references

PR #4569, opened 2023-07-02
SALTO-4415: deploy article attachments in chunks of 20

edenhassid:
deploy article attachments in chunks of 20

---

---
_Release Notes_: 
zendesk:
* fix bug where deployment of an article with more than 20 attachments did not work

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/filters/article/article.ts
+++ b/packages/zendesk-adapter/src/filters/article/article.ts
@@ -174,12 +174,17 @@ const associateAttachments = async (
   client: ZendeskClient,
   articleId: number,
   attachmentsIds: number[]
-): Promise<number> => {
-  const res = await client.post({
-    url: `/api/v2/help_center/articles/${articleId}/bulk_attachments`,
-    data: { attachment_ids: attachmentsIds },
-  })
-  return res.status
+): Promise<number[]> => {
+  const attachChunk = _.chunk(attachmentsIds, 20)
+  log.debug(`there are ${attachmentsIds?.length ?? 0} attachments to associate, associating in chunks of 20`)
+  return Promise.all(attachChunk.map(async (chunk: number[], index: number) => {
+    log.debug(`starting article attachment associate chunk ${index + 1}/${attachChunk.length}`)
+    const res = await client.post({
+      url: `/api/v2/help_center/articles/${articleId}/bulk_attachments`,
+      data: { attachment_ids: chunk },
+    })
+    return res.status
+  }))
 }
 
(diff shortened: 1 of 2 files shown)


packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -174,12 +174,17 @@ const associateAttachments = async (
...
-  const res = await client.post({
-    url: `/api/v2/help_center/articles/${articleId}/bulk_attachments`,
-    data: { attachment_ids: attachmentsIds },
-  })
-  return res.status
+): Promise<number[]> => {
+  const attachChunk = _.chunk(attachmentsIds, 20)
+  log.debug(`there are ${attachmentsIds?.length ?? 0} attachments to associate, associating in chunks of 20`)
+  return Promise.all(attachChunk.map(async (chunk: number[], index: number) => {
+    log.debug(`starting article attachment associate chunk ${index + 1}/${attachChunk.length}`)
```
netama, line 181: 1. let's also log the article? otherwise these will get mixed up in the logs
2. I think it'd be better to log the article elem id and not the internal id - IIUC it looks like all calls to this function pass `articleId: articleInstance.value.id`, so I think we can just pass in the article instance and then have both the elem id and internal id?

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -174,12 +174,17 @@ const associateAttachments = async (
   client: ZendeskClient,
   articleId: number,
   attachmentsIds: number[]
-): Promise<number> => {
-  const res = await client.post({
-    url: `/api/v2/help_center/articles/${articleId}/bulk_attachments`,
-    data: { attachment_ids: attachmentsIds },
-  })
-  return res.status
+): Promise<number[]> => {
```
netama, line 177: I think the fact these are deployed as chunks is an implementation detail that is not know externally - so IMO this should return either a single "success" boolean, or some mapping between the attachments and which one succeeded vs failed. given the current usage we can probably go with the former? (+ add some log here about the specific failed attachments)?

netama (approved):
looks great 🎈 added some nits but no blockers

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -174,12 +174,17 @@ const associateAttachments = async (
...
+  const attachChunk = _.chunk(attachmentsIds, 20)
+  log.debug(`there are ${attachmentsIds?.length ?? 0} attachments to associate, associating in chunks of 20`)
```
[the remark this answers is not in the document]
edenhassid: It shouldn't happen but It can be undefined, I will fix the function signature

PR #4571, opened 2023-07-04
SALTO-4451: Statuses pagination bug

IdoZakk:
_Due to an internal jira bug temporary fix to allow fetching all statuses_

---

_The pagination function and its tests were restored completely from a previous version of the file, they are not new_

---
_Release Notes_: 
Jira Adapter: Fixed a bug that prevented fetching more than 200 statuses

---
_User Notifications_: 
None


alonstern (approved):
👌

PR #4573, opened 2023-07-04
SALTO-4402 - undo revert and disable by default

tomsellek:
Re-introduce the change from SALTO-4402 (https://github.com/salto-io/salto/pull/4537), but make them disabled by default.

From original PR: Fetching profiles takes a very long time. One part of that time is resolving references on the many, many fields that each profile has (basically the number of fields in a single profile = O(number of fields in all custom objects)).
So let's not resolve references in profiles by default, unless the user explicitly asks us to.

---

---
_Release Notes_: 
Salesforce: add an adapter fetch configuration that controls reference generation for profiles

---
_User Notifications_: 
N/A


amitgoldarb (approved):
Thanks for this!

PR #4574, opened 2023-07-05
SALTO-4214: change zendesk errors to return saltoErrors instead of throwing them

edenhassid:
change zendesk errors to return saltoErrors instead of throwing them

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-api/src/error.ts
+++ b/packages/adapter-api/src/error.ts
@@ -33,8 +33,8 @@ export type SaltoElementError = SaltoError & {
 export const isSaltoElementError = (error: SaltoError | SaltoElementError):
     error is SaltoElementError => 'elemID' in error
 
-export const isError = (error: unknown): error is SaltoError | Error =>
-  (_.isObject(error) && ('message' in error) && ('severity' in error)) || _.isError(error)
+export const isSaltoError = (error: unknown): error is SaltoError =>
+  (_.isObject(error) && ('message' in error) && ('severity' in error))
 
 export const createSaltoElementErrorFromError = ({
   error,
--- a/packages/adapter-api/src/error.ts
+++ b/packages/adapter-api/src/error.ts
@@ -32,6 +32,26 @@ export type SaltoElementError = SaltoError & {
 export const isSaltoElementError = (error: SaltoError | SaltoElementError):
     error is SaltoElementError => 'elemID' in error
 
+export const createSaltoElementErrorFromError = ({
+  error,
+  severity,
+  elemID,
+}: {
+    error: Error
+    severity: SeverityLevel
+    elemID: ElemID
+}): SaltoElementError => ({ message: error.message, severity, elemID })
+
+export const createSaltoElementError = ({
+  message,
+  severity,
+  elemID,
+}: {
+    message: string
+    severity: SeverityLevel
+    elemID: ElemID
+}): SaltoElementError => ({ message, severity, elemID })
+
 export class CredentialError extends Error {}
 
 export const isCredentialError = (error: unknown): error is CredentialError =>
--- a/packages/adapter-api/src/error.ts
+++ b/packages/adapter-api/src/error.ts
@@ -13,6 +13,7 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
+import _ from 'lodash'
 import { ElemID } from './element_id'
 
 export type SeverityLevel = 'Error' | 'Warning' | 'Info'
@@ -31,6 +32,9 @@ export type SaltoElementError = SaltoError & {
 
 export const isSaltoElementError = (error: SaltoError | SaltoElementError):
     error is SaltoElementError => 'elemID' in error
+
+export const isError = (error: unknown): error is SaltoError | Error =>
+  (_.isObject(error) && ('message' in error) && ('severity' in error)) || _.isError(error)
 
 export const createSaltoElementErrorFromError = ({
   error,
--- a/packages/zendesk-adapter/src/adapter.ts
+++ b/packages/zendesk-adapter/src/adapter.ts
@@ -636,6 +636,51 @@ export default class ZendeskAdapter implements AdapterOperations {
     return { elements, errors: fetchErrors, updatedConfig }
   }
 
+  private async getGuideDeployResults(
+    subdomainToClient: Record<string, ZendeskClient>,
+    subdomainToGuideChanges: Record<string, Change<InstanceElement>[]>,
+    saltoErrors: SaltoError[]
+  ): Promise<DeployResult[]> {
+    try {
+      return await awu(Object.entries(subdomainToClient))
+        .map(async ([subdomain, client]) => {
+          const brandRunner = await this.createFiltersRunner({
+            filterRunnerClient: client,
+            paginator: createPaginator({
+              client,
+              paginationFuncCreator: paginate,
+            }),
+          })
+          await brandRunner.preDeploy(subdomainToGuideChanges[subdomain])
+          const { deployResult: brandDeployResults } = await brandRunner.deploy(
+            subdomainToGuideChanges[subdomain]
+          )
+          const guideChangesBeforeRestore = [...brandDeployResults.appliedChanges]
+          try {
+            await brandRunner.onDeploy(guideChangesBeforeRestore)
+          } catch (e) {
+            if (!isSaltoError(e)) {
+              throw e
+            }
+            saltoErrors.push(e)
+          }
+          return {
+            appliedChanges: guideChangesBeforeRestore,
+            errors: brandDeployResults.errors,
+          }
+        })
+        .toArray()
+    } catch (e) {
+      if (!isSaltoError(e)) {
+        throw e
+      }
+      return [{
+        appliedChanges: [],
+        errors: [e],
+      }]
+    }
+  }
+
   /**
    * Deploy configuration elements to the given account.
    */
@@ -679,23 +724,23 @@ export default class ZendeskAdapter implements AdapterOperations {
     try {
       await runner.preDeploy(supportResolvedChanges)
     } catch (e) {
-      if (isSaltoError(e)) {
-        return {
-          appliedChanges: [],
-          errors: [e],
-        }
+      if (!isSaltoError(e)) {
+        throw e
       }
-      throw e
+      return {
+        appliedChanges: [],
+        errors: [e],
+      }
     }
     const { deployResult } = await runner.deploy(supportResolvedChanges)
     const appliedChangesBeforeRestore = [...deployResult.appliedChanges]
     try {
       await runner.onDeploy(appliedChangesBeforeRestore)
     } catch (e) {
-      if (isSaltoError(e)) {
-        saltoErrors.push(e)
+      if (!isSaltoError(e)) {
+        throw e
       }
-      throw e
+      saltoErrors.push(e)
     }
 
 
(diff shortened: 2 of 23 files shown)


packages/zendesk-adapter/src/deployment.ts
```diff
@@ -148,42 +157,52 @@ export const deployChanges = async <T extends Change<ChangeDataType>>(
...
+        if (!(_.isError(err) || isSaltoElementError(err))) {
           throw err
         }
-        return err
+        errors.push({
+          message: err.message,
+          severity: 'Error',
+          elemID: getChangeData(change).elemID,
+        })
+        return undefined
```
netama, line 175: I might be missing something, but why did we need to change this part to aggregating the `errors` externally? couldn't the `Promise.all()` still return the "extended" errors, and we would just change the `partition` below to check for an extended version of `isSaltoElementError` (making it accept `any` as input)?
(I see this is similar to what we did in [jira](https://github.com/salto-io/salto/pull/4498/files#diff-2dbe5d948def2d39388a033435d59ee10d1a6f00d59e268e22953f317dc45711R103), @alonstern was there a reason we couldn't use functional style there?)

packages/zendesk-adapter/src/deployment.ts
```diff
@@ -148,42 +157,52 @@ export const deployChanges = async <T extends Change<ChangeDataType>>(
...
         throw err
       }
-      return err
+      errors.push({
+        message: err.message,
+        severity: 'Error',
+        elemID: getChangeData(change).elemID,
+      })
+      return undefined
     }
```
netama, line 202: this seems to be existing code, but let's avoid the code duplication with the "parallel" version and move this to a shared function?

packages/zendesk-adapter/src/filters/article/utils.ts
```diff
@@ -226,17 +226,21 @@ export const updateArticleTranslationBody = async ({
...
+              return attachmentInstances[attachmentIndex].value.id.toString()
+            }
+            return prepRef(part)
           }
-          return prepRef(part)
-        }
-      )
+        )
+      } catch (e) {
+        log.error(`Error serializing article translation body in Deployment for ${translationInstance.elemID.getFullName()}: ${e}, stack: ${e.stack}`)
```
netama, line 242: shouldn't we return an error here? I'm guessing the `put` request will fail, no?

packages/zendesk-adapter/src/filters/brand_logo.ts
```diff
@@ -231,19 +233,19 @@ const filterCreator: FilterCreator = ({ client }) => ({
           ? await logoInstance.value.content.getContent()
           : undefined
         const deployResult = await deployBrandLogo(client, logoInstance, fileContent)
-        return deployResult === undefined ? change : deployResult
+        if (deployResult !== undefined) { //  deployment was not successful
```
netama, line 235: same styling question (won't comment on others)

packages/zendesk-adapter/src/filters/custom_statuses.ts
```diff
@@ -132,7 +132,12 @@ const filterCreator: FilterCreator = ({ client, config }) => ({
           data: { ids: defaults },
         })
       } catch (e) {
-        error.push(new Error(e))
+        const saltoError: SaltoElementError = {
```
netama, line 135: nit: to avoid the casting (here and in several other places) and add a bit more type safety on the input, maybe add a utility function that returns a `SaltoElementError` given an error, elem id and severity level?

packages/zendesk-adapter/src/filters/macro_attachments.ts
```diff
@@ -235,9 +240,14 @@ const filterCreator: FilterCreator = ({ config, client }) => ({
           appliedChanges: [],
           errors: childrenChanges
             .map(getChangeData)
-            .map(e => new Error(
-              `Failed to update ${e.elemID.getFullName()} since it has no valid parent`
-            )),
+            .map(e => {
+              const saltoError: SaltoElementError = {
+                message: `Failed to update ${e.elemID.getFullName()} since it has no valid parent`,
```
netama, line 245: not necessarily for this PR since this is also not a small change, but I think we can omit the elem ids form the message text from now on if it is listed as the `elemID` arg

netama (commented):
🎈 
added some nits, and one potential bug (I think) in article/utils.ts

packages/zendesk-adapter/src/filters/brand_logo.ts
edenhassid: not sure I understand

packages/zendesk-adapter/src/deployment.ts
edenhassid: this function is the one that turns errors to saltoElementErrors, so a partition will not work

alonstern: `isSaltoElementError` only checks `'elemID' in error` so it's a bit more than changing to `any`

packages/adapter-api/src/error.ts
```diff
@@ -32,6 +33,29 @@ export type SaltoElementError = SaltoError & {
 export const isSaltoElementError = (error: SaltoError | SaltoElementError):
     error is SaltoElementError => 'elemID' in error
 
+export const isError = (error: unknown): error is SaltoError | Error =>
+  (_.isObject(error) && ('message' in error) && ('severity' in error)) || _.isError(error)
+
+export const createSaltoElementErrorFromError = ({
```
netama, line 39: not sure if these belong in `adapter-api` or `adapter-utils`, @ori-moisis  any preference?

packages/zendesk-adapter/src/filters/reorder/trigger.ts
```diff
@@ -59,7 +59,11 @@ const deployFunc: DeployFuncType = async (change, client, apiDefinitions) => {
   const instance = getChangeData(clonedChange)
   const { order } = instance.value
   if (!areTriggerOrderEntries(order)) {
-    throw new Error('trigger_order\' order field has an invalid format')
+    throw createSaltoElementError({
+      message: 'trigger_order\' order field has an invalid format',
```
netama, line 63: not specific for this PR, but can we fix the typo please? :) 
```suggestion
      message: 'trigger_order\'s order field has an invalid format',
```

netama (approved):
🎈

packages/adapter-api/src/error.ts
ori-moisis: as long as it is a simple "constructor" I think it is fine to be here

packages/zendesk-adapter/src/deployment.ts
```diff
@@ -144,45 +152,38 @@ export const deployChange = async (
...
-      if (!_.isError(err)) {
-        throw err
-      }
-      return err
-    }
-  }).toArray()
-
-  const [errors, appliedChanges] = _.partition(result.flat(), _.isError)
+  const result = await awu(changes).map(async change => deployChangesHelper(change, deployChangeFunc)).toArray()
+  const [errors, appliedChanges] = _.partition(result.flat(), isError)
```
benisraeladi, line 186: This is problematic. Because of the definition of `isError`, `errors` can contain something of type `Error`, which is code that we want to get rid of.
@edenhassid Please check that this will compile when the change is complete by temporarily changing the definition of `DeployResult` so that it only accepts `(SaltoError | SaltoElementError)[]`

benisraeladi (commented):
This is only a partial review

packages/zendesk-adapter/src/filters/brand_logo.ts
```diff
@@ -237,8 +237,8 @@ const filterCreator: FilterCreator = ({ client }) => ({
 
     const [deployLogoErrors, successfulChanges] = _.partition(
       deployLogoResults,
-      _.isError,
-    )
+      isError,
+    ) as [(SaltoError | Error)[], Change[]]
```
benisraeladi, line 241: Same problem as I talked about, this code will not compile after SALTO-4324 is completed

packages/zendesk-adapter/src/filters/article/utils.ts
netama: I was thinking we would catch these in `adapter.ts` and convert to errors (`preDeploy` is an implementation detail but not exposed externally), I think we are just missing this code

packages/zendesk-adapter/src/adapter.ts
```diff
@@ -703,13 +722,29 @@ export default class ZendeskAdapter implements AdapterOperations {
             paginationFuncCreator: paginate,
           }),
         })
-        await brandRunner.preDeploy(subdomainToGuideChanges[subdomain])
+        try {
+          await brandRunner.preDeploy(subdomainToGuideChanges[subdomain])
+        } catch (e) {
+          if (isSaltoError(e)) {
+            return {
+              appliedChanges: [],
```
edenhassid, line 730: not sure what I should return in the appliedChnages

benisraeladi: Does the `preDeploy` actively change anything? If not, then this should be fine because you don't want to continue to a deployment when you have something that has already failed.
But if the `preDeploy` does mutate things or send them to ZD, then this `map` is problematic and should probably be a `forEach` with a way to monitor which changes were applied

packages/zendesk-adapter/src/adapter.ts
```diff
@@ -703,25 +726,47 @@ export default class ZendeskAdapter implements AdapterOperations {
             paginationFuncCreator: paginate,
           }),
         })
-        await brandRunner.preDeploy(subdomainToGuideChanges[subdomain])
+        try {
+          await brandRunner.preDeploy(subdomainToGuideChanges[subdomain])
+        } catch (e) {
+          if (isSaltoError(e)) {
+            const appliedChanges = await awu(appliedChangesBeforeRestore)
```
netama, line 733: I might not be reading this correctly, but I think we don't need this logic here - the "aggregation" step will still be reached later, so we just need to return the result for this guide brand here, no?

... and thinking about this now, I think we should actually catch within each of the `onDeploy`s (and also the `preDeploy`s, but they are less important) rather than outside after all, so that an `onDeploy` error in one filter won't cause another filter's `onDeploy` not to run. but this is an existing problem, so we can solve it separately and not as part of this effort (and the risk should be small given the change groups)

edenhassid: should I return ```
{
          appliedChanges: [],
          errors: [e],
        }
        ```
        or add the error to the list?

packages/zendesk-adapter/src/adapter.ts
```diff
@@ -636,6 +636,51 @@ export default class ZendeskAdapter implements AdapterOperations {
     return { elements, errors: fetchErrors, updatedConfig }
   }
 
+  private async getGuideDeployResults(
+    subdomainToClient: Record<string, ZendeskClient>,
+    subdomainToGuideChanges: Record<string, Change<InstanceElement>[]>,
+    saltoErrors: SaltoError[]
```
netama, line 642: I don't think we should pass `saltoErrors` into this function, we can add an inner one per brand and use it in a similar way but right now this leaks "output" data masked as an input parameters, which is something we try to avoid in functional-style and I don't think it's necessary here

netama (approved):
LGTM pending the final change 🎈

PR #4575, opened 2023-07-05
SAAS-8289: Bump axios version to 0.26.1

ori-moisis:
---

Not really critical, but aligning our required axios with the required axios version of the openai package

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/workato-adapter/src/client/connection.ts
+++ b/packages/workato-adapter/src/client/connection.ts
@@ -14,6 +14,7 @@
 * limitations under the License.
 */
 import _ from 'lodash'
+import { AxiosRequestHeaders } from 'axios'
 import { AccountInfo, CredentialError } from '@salto-io/adapter-api'
 import { client as clientUtils } from '@salto-io/adapter-components'
 import { Credentials } from '../auth'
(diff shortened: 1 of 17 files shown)


adamro (approved):

packages/workato-adapter/src/client/connection.ts
```diff
@@ -40,10 +41,10 @@ export const createConnection: clientUtils.ConnectionCreator<Credentials> = retr
   clientUtils.axiosConnection({
     retryOptions,
     authParamsFunc: async ({ username, token }: Credentials) => ({
-      headers: _.isEmpty(username)
+      headers: username === undefined || _.isEmpty(username)
```
netama, line 44: why did we need this change?

netama (approved):

packages/workato-adapter/src/client/connection.ts
ori-moisis: `_.isEmpty` is not enough for the typescript compiler to infer that `username` is a string, and then it wouldn't let us return it because it would be `string | undefined` where headers only allow `string | number | boolean` or something like that

PR #4576, opened 2023-07-05
SALTO-4456: Fixed local dir store when a file is deleted

alonstern:
Fixed local dir store when a file is deleted

---

The local dir store `get` and `mtimestamp` function did not check if a file is in the `deleted` list which made it returns a file that was deleted from the dirstore if a flush wasn't called in between.
This fix is necessary for the "restore elements path" flag, since we remove the files and recreate them.

Fixing the dirstore created another issue that we access the content of lazy static files for the first time after they were deleted so they can't be loaded. It happened in the router when we delete a static file from common and create it in a specific env. I change the LazyStaticFile to ignore deletions that were not flushed to solve this case

---
_Release Notes_: 
None (I am not aware of this having an effect on the user so far)

---
_User Notifications_: 
None

### DIFF
--- a/packages/core/test/workspace/local/directory_store.test.ts
+++ b/packages/core/test/workspace/local/directory_store.test.ts
@@ -278,6 +278,20 @@ describe('localDirectoryStore', () => {
     it('fails to get an absolute path', () =>
       localDirectoryStore({ baseDir: 'dir', name: '', encoding }).getFiles(['/aaaa'])
         .catch(err => expect(err.message).toEqual('Filepath not contained in dir store base dir: /aaaa')))
+
+    it('Should return deleted file when requested to ignore cache', async () => {
+      mockFileExists.mockResolvedValueOnce(false)
+        .mockResolvedValueOnce(true)
+        .mockResolvedValueOnce(true)
+      mockReadFile.mockResolvedValueOnce('bla1').mockResolvedValueOnce('bla2')
+      mockStat.mockResolvedValue({ mtimeMs: 7 })
+      const dirStore = localDirectoryStore({ baseDir: '', name: '', encoding })
+      await dirStore.delete('b')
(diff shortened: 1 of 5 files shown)


ori-moisis (approved):

packages/workspace/src/workspace/static_files/source.ts
```diff
@@ -235,9 +235,14 @@ export const buildStaticFilesSource = (
...
       staticFilesCache.clone(),
     ),
-    delete: async (staticFile: StaticFile): Promise<void> => (
-      staticFilesDirStore.delete(staticFile.filepath)
-    ),
+    delete: async (staticFile: StaticFile): Promise<void> => {
+      // If the static file is lazy, and the content was not yet loaded,
+      // we need to load it before deleting the file since after we won't be able to
+      if (staticFile instanceof LazyStaticFile) {
+        await staticFile.getContent()
```
ori-moisis, line 242: this would be very inefficient in most flows as it would force reading content that we don't actually need.
I think it would be very easy for someone to delete this code when trying to optimize and the comment above does not help much because it does not give a reason why we need to read the content, it just says that we need to.

I think it would be better to handle the scenario that reads files after they were deleted and not make every flow read all this content for no reason.

if it is really impossible to fix the flow, maybe we need to provide some kind of override for lazy static files to read the content in a way that ignores unflushed deletions

packages/core/test/workspace/local/directory_store.test.ts
```diff
@@ -149,6 +180,21 @@ describe('localDirectoryStore', () => {
     })
   })
 
+  describe('mtimestamp', () => {
```
ori-moisis, line 183: we need to add tests for all the above cases when we pass in the option to ignore deletions cache

PR #4577, opened 2023-07-05
SALTO-4402 - Salesforce: Recreate profile refs as additional refs

tomsellek:
Building on top of https://github.com/salto-io/salto/pull/4573, re-add the removed profile references in the adapter's `getAdditionalReferences`. 
`getAdditionalReferences` already dealt with field-level permissions, so we add all the other profile sections.

Note: This PR is missing the references from `layoutAssignments`...`recordType`. due to the urgency of this PR, I will add them in a follow-up PR.
---

I basically factored out the implementation for fieldPermissions ~and added support for a couple of edge cases (e.g. in layoutAssignments every entry is an array with a single element for some reason)~

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/additional_references.ts
+++ b/packages/salesforce-adapter/src/additional_references.ts
@@ -13,15 +13,78 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
-import { GetAdditionalReferencesFunc, getChangeData, isAdditionChange, isFieldChange, isInstanceChange, isRemovalOrModificationChange } from '@salto-io/adapter-api'
+import {
+  Element,
+  GetAdditionalReferencesFunc,
+  getChangeData,
+  InstanceElement,
+  isAdditionChange,
+  isFieldChange,
+  isInstanceChange,
+  isRemovalOrModificationChange,
+  ReferenceMapping,
+} from '@salto-io/adapter-api'
 import _ from 'lodash'
 import { collections } from '@salto-io/lowerdash'
-import { isFieldOfCustomObject } from './transformers/transformer'
+import {
+  isCustomObject,
+  isFieldOfCustomObject,
+} from './transformers/transformer'
 import { isInstanceOfType, safeApiName } from './filters/utils'
-import { API_NAME_SEPARATOR, FIELD_PERMISSIONS, PERMISSION_SET_METADATA_TYPE, PROFILE_METADATA_TYPE } from './constants'
+import {
+  API_NAME_SEPARATOR,
+  FIELD_PERMISSIONS,
+  FLOW_METADATA_TYPE,
+  LAYOUT_TYPE_ID_METADATA_TYPE,
+  PERMISSION_SET_METADATA_TYPE,
+  PROFILE_METADATA_TYPE,
+  RECORD_TYPE_METADATA_TYPE,
+} from './constants'
 
 const { awu } = collections.asynciterable
+const { makeArray } = collections.array
 
+const CUSTOM_APPLICATION_TYPE = 'CustomApplication'
+const APEX_CLASS_TYPE = 'ApexClass'
+const APEX_PAGE_TYPE = 'ApexPage'
+const CUSTOM_APP_SECTION = 'applicationVisibilities'
+const APEX_CLASS_SECTION = 'classAccesses'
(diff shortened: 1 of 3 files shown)


packages/salesforce-adapter/src/additional_references.ts
```diff
@@ -13,15 +13,78 @@
...
+    .filter(profileOrPermissionSet => _.get(profileOrPermissionSet.value[profileSection], apiName))
+    .flatMap(profileOrPermissionSet => {
+      const sectionEntry = _.get(profileOrPermissionSet.value[profileSection], apiName)
+      // For layoutAssignments, every entry is an array
+      const shouldAddArrayIndex = Array.isArray(sectionEntry)
+      return makeArray(sectionEntry)
+        .map((_unused, index) => {
+          const refSourceField = [
+            ...apiName.split(API_NAME_SEPARATOR),
+            ...shouldAddArrayIndex ? [`[${index}]`] : [],
```
tomsellek, line 78: Reviewer note:
Is there a better way to make a reference to an array element?

packages/salesforce-adapter/src/additional_references.ts
```diff
@@ -13,15 +13,78 @@
...
+const CUSTOM_APPLICATION_TYPE = 'CustomApplication'
+const APEX_CLASS_TYPE = 'ApexClass'
+const APEX_PAGE_TYPE = 'ApexPage'
+const CUSTOM_APP_SECTION = 'applicationVisibilities'
+const APEX_CLASS_SECTION = 'classAccesses'
+const FLOW_SECTION = 'flowAccesses'
+const LAYOUTS_SECTION = 'layoutAssignments'
+const OBJECT_SECTION = 'objectPermissions'
+const APEX_PAGE_SECTION = 'pageAccesses'
+const RECORD_TYPE_SECTION = 'recordTypeVisibilities'
```
tomsellek, line 56: Reviewer note: should these be moved to constants.ts?

packages/salesforce-adapter/src/additional_references.ts
```diff
@@ -42,18 +115,37 @@ export const getAdditionalReferences: GetAdditionalReferencesFunc = async change
...
-      if (fieldApiName === undefined) {
-        return []
-      }
-      return profilesAndPermissionSets
-        .filter(instance => _.get(instance.value.fieldPermissions, fieldApiName) !== undefined)
-        .map(instance => ({
-          source: instance.elemID.createNestedID(FIELD_PERMISSIONS, ...fieldApiName.split(API_NAME_SEPARATOR)),
-          target: field.elemID,
-        }))
-    })
```
tomsellek, line 57: Reviewer note:
This was factored out into `generateRefsFromProfileOrPermissionSet`

packages/salesforce-adapter/src/additional_references.ts
amitgoldarb: @ori-moisis @tamtamirr Any idea?

packages/salesforce-adapter/src/additional_references.ts
amitgoldarb: Yes, probably will remove some of these duplications in the future. I searched for a few of these and saw some references, it could be nice in the future to DRY here.

amitgoldarb (commented):
This looks great!
Added some small comments - Also added some product questions https://salto-io.atlassian.net/browse/SALTO-4457?focusedCommentId=90744 here.

packages/salesforce-adapter/src/additional_references.ts
ori-moisis: not commenting on the actual code here (no idea what the "[]" means) - creating references to arrays is usually a red flag - usually you don't want to point to an index, e.g - I depend on the field "Bla__c", not on the 5th field in some object...
why are we creating references to arrays?

tomsellek: @ori-moisis this is the *source*, not the target of the reference though?

tomsellek: Per discussion with @alonstern we don't need the ref source to be the nested value, but rather the field that is one level up. So no need for array index shenanigans and the code can be made much simpler.

alonstern (approved):

PR #4578, opened 2023-07-06
SALTO-4227 Support loading files without their attributes in SDF integration

dantal4:
Today we need all files&folders in the SDF Project’s “FileCabinet” directory to have their metadata xmls in the .attributes folder. For SDF, it is supported to deploy a file without it’s metadata xml - if the file already exists in the account only its content is updated, and if it is a new file then it gets the default metadata (non checkbox is marked, no description, etc).

---

---
_Release Notes_: 
Netsuite Adapter:
- Support loading files without their attributes in SDF integration

---
_User Notifications_: 
None

### DIFF
--- a/packages/netsuite-adapter/src/client/sdf_parser.ts
+++ b/packages/netsuite-adapter/src/client/sdf_parser.ts
@@ -278,7 +278,7 @@ const transformFoldersWithoutAttributes = (
     .map(path => path.split(FILE_CABINET_PATH_SEPARATOR))
     .map(path => path.slice(1, -1))
     // adding all parent folders
-    .flatMap(path => path.map((p, i) => [...path.slice(0, i), p]))
+    .flatMap(path => path.map((_p, i) => path.slice(0, i + 1)))
     .map(path => path.join(FILE_CABINET_PATH_SEPARATOR))
     .uniq()
     .filter(path => !foldersFromAttributesPaths.has(path))
--- a/packages/netsuite-adapter/src/client/sdf_parser.ts
+++ b/packages/netsuite-adapter/src/client/sdf_parser.ts
@@ -14,6 +14,7 @@
 * limitations under the License.
 */
 import _ from 'lodash'
+import os from 'os'
 import osPath from 'path'
 import he from 'he'
 import xmlParser from 'fast-xml-parser'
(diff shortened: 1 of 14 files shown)


packages/netsuite-adapter/src/transformer.ts
```diff
@@ -144,6 +148,24 @@ export const createInstanceElement = async (
 
   const instanceName = getInstanceName(valuesWithTransformedAttrs)
   const instanceFileName = pathNaclCase(instanceName)
+
+  if ((isFolderCustomizationInfo(customizationInfo) || isFileCustomizationInfo(customizationInfo))
+    && customizationInfo.hadMissingAttributes) {
+    const instanceElemId = new ElemID(NETSUITE, type.elemID.name, 'instance', instanceName)
```
omrilit, line 154: Could you please add a comment explaining what we are doing here and why? I think that it worth explanation for a future reader

packages/netsuite-adapter/src/transformer.ts
```diff
@@ -144,6 +148,24 @@ export const createInstanceElement = async (
...
+    const existingInstance = await elementsSource.get(instanceElemId)
+    if (isInstanceElement(existingInstance)) {
+      const clonedInstance = existingInstance.clone()
+      clonedInstance.refType = new TypeReference(type.elemID, type)
+      if (isFileInstance(clonedInstance)) {
+        const fileContent = valuesWithTransformedAttrs[(fileContentField as Field).name]
```
omrilit, line 160: nit: maybe better to extract this disgusting value to a const? `(fileContentField as Field).name`

packages/netsuite-adapter/src/client/sdf_parser.ts
```diff
@@ -231,6 +266,31 @@ const transformFolders = (
...
+): FolderCustomizationInfo[] => {
+  const foldersFromAttributesPaths = new Set(
+    foldersFromAttributes
+      .map(folder => folder.path.join(FILE_CABINET_PATH_SEPARATOR))
+  )
+  return _(filePaths)
+    .map(path => path.split(FILE_CABINET_PATH_SEPARATOR))
+    .map(path => path.slice(1, -1))
+    // adding all parent folders
+    .flatMap(path => path.map((p, i) => [...path.slice(0, i), p]))
```
omrilit, line 281: super nit: I think it's a bit more readable, but as you prefer
```suggestion
    .flatMap(path => path.map((_, i) => path.slice(0, i + 1)))
```

omrilit (approved):
Very nice work!!

PR #4579, opened 2023-07-06
SALTO-4262 Enable to deploy more than 10K options for custom fields in jira

idoamit8:
1. Jira's API doesn't allow deploying more than 10K options to a custom field.
2. This limitation can be bypassed by using Jira's private API.
3. In this PR, I have added an option to deploy the remaining options through the private API after deploying 10K options via the public API. 

---

#### Followup Topics:
1. Maybe We need to add a notification to the user that it may take longer because it exceeds the amount supported by Jira.
2. Maybe we want to add a flag that the user can say if he wants the option to deploy more than 10K options or not. 
3. I will add tests after the correctness of the code will reviewed.

#### To test the code manually
1. Go to https://ori-salto-test.atlassian.net/ account. 
2. search for Custom field.
3. go to testSelectList.
5. Try to add options such that there will be more than 10K options ( you can remove and modify while doing it). 
6. Deploy. 
---
_Release Notes_: 
__Jira adapter__
The users will be able to deploy more than 10K options to a custom field.

---
_User Notifications_: 
None


packages/adapter-components/src/client/http_client.ts
```diff
@@ -159,6 +160,14 @@ export abstract class AdapterHTTPClient<
     return this.sendRequest('get', params)
   }
 
+  @throttle<TRateLimitConfig>({ bucketName: 'deploy', keys: ['url', 'queryParams'] })
+  @logDecorator(['url', 'queryParams'])
+  @requiresLogin()
+  public async get(params: ClientDataParams):
```
alonstern, line 166: Why not use `getSinglePage`?

packages/jira-adapter/src/filters/fields/context_options.ts
```diff
@@ -22,13 +22,14 @@ import _ from 'lodash'
...
 import { setFieldDeploymentAnnotations } from '../../utils'
+import { JSP_API_HEADERS } from '../../client/headers'
 
 const log = logger(module)
 
 const { awu } = collections.asynciterable
 
 const OPTIONS_MAXIMUM_BATCH_SIZE = 1000
-
+const OPTIONS_MAXIMUM_SIZE = 10000
```
alonstern, line 32: name nit: maybe `PUBLIC_API_OPTIONS_LIMIT`?

packages/jira-adapter/src/filters/fields/context_options.ts
```diff
@@ -108,17 +109,31 @@ const updateContextOptions = async ({
...
+    }))
+  }
 
+  if (addedOptions.length !== 0) {
+    const optionLengthBefore = isModificationChange(contextChange)
+      ? getOptionsFromContext(contextChange.data.before).length : 0
+    // eslint-disable-next-line no-nested-ternary
+    const sliceIndex = optionLengthBefore > OPTIONS_MAXIMUM_SIZE
+      ? 0 : optionLengthBefore + addedOptions.length > OPTIONS_MAXIMUM_SIZE
+        ? OPTIONS_MAXIMUM_SIZE - optionLengthBefore : addedOptions.length
```
alonstern, line 126: nit: you can do that with max, min and remove the lint disable line
```suggestion
    const numberOfPrivateApiOptions = Math.min(
      Math.max(optionLengthBefore + addedOptions.length - OPTIONS_MAXIMUM_SIZE, 0)
      addedOptions.length
    )
```

packages/jira-adapter/src/filters/fields/context_options.ts
```diff
@@ -108,17 +109,31 @@ const updateContextOptions = async ({
...
+    const firstOptions = addedOptions.slice(0, sliceIndex)
+    const secondOptions = addedOptions.slice(sliceIndex, addedOptions.length)
```
alonstern, line 128: nit: variable names:
```suggestion
    const privateApiOptions = addedOptions.slice(0, sliceIndex)
    const publicApiOptions = addedOptions.slice(sliceIndex, addedOptions.length)
```

packages/jira-adapter/src/filters/fields/context_options.ts
```diff
@@ -137,6 +152,39 @@ const updateContextOptions = async ({
...
+        resp.data.options.forEach(newOption => {
+          if (newOption.optionId !== undefined) {
+            idToOption[newOption.optionId]
+              .cascadingOptions[naclCase(newOption.value)].id = newOption.id
+          } else {
+            optionsMap[naclCase(newOption.value)].id = newOption.id
+          }
+        })
+      }
+    }
```
alonstern, line 187: Can you extract this to a different function?

packages/jira-adapter/src/filters/fields/context_options.ts
```diff
@@ -137,6 +152,39 @@ const updateContextOptions = async ({
         })
       }
     })
+    // Jira API doesn't support adding more than 10000 through the API.
+    // We need to add the rest through the private API.
+    await awu(secondOptions).forEach(async option => {
+      const body = new URLSearchParams({ addValue: option.value, fieldConfigId: contextChange.data.after.value.id })
+      const basicUrl = '/secure/admin/EditCustomFieldOptions!add.jspa'
+      await client.post({
```
alonstern, line 160: nit: you can use jspPost and remove the headers

alonstern, line 160: Does that work for cascading options?

alonstern (commented):
🔥

packages/jira-adapter/src/filters/fields/context_options.ts
alonstern: It should be 2 since it his case the number of private API options is 2, if you want to number of public API options you can do:
```suggestion
    const numberOfPublicApiOptions = Math.min(
      Math.max(OPTIONS_MAXIMUM_SIZE - optionLengthBefore  - addedOptions.length, 0)
      addedOptions.length
    )
```

idoamit8: Yep, you right. I got confused because the slice index was the number of the public API options. Thanks :)

packages/jira-adapter/src/filters/fields/context_options.ts
```diff
@@ -95,30 +97,102 @@ type UpdateContextOptionsParams = {
...
+      maxResults: '1000',
+    },
+    pageSizeArgName: 'maxResults',
+  }
+
+  const optionsPromise = toArrayAsync(paginator(
+    paginationArgs,
+    page => makeArray(page) as clientUtils.ResponseValue[]
+  ))
+  const values = (await optionsPromise).flatMap(lst => lst[0].values).filter(isOption)
```
alonstern, line 154: I think you are supposed to do 
```suggestion
  const options = await toArrayAsync(paginator(
    paginationArgs,
    page => makeArray(page.values) as clientUtils.ResponseValue[]
  ))
  const values = options.filter(isOption)
```

packages/jira-adapter/src/filters/fields/context_options.ts
```diff
@@ -192,21 +282,24 @@ const reorderContextOptions = async (
       },
     }]
   ) : optionsGroups.map(group =>
-    _.chunk(group, OPTIONS_MAXIMUM_BATCH_SIZE).map((chunk, index) =>
-      ({
+    _.chunk(group, OPTIONS_MAXIMUM_BATCH_SIZE).map((chunk, index) => {
+      const ans = ({
```
alonstern, line 286: nit: I guess this change was for debugging

alonstern (approved):
Looks great 👌 
added some nits
Also, the users notification section in the description should be empty in this case. We put there only messages about things that affect users' workspaces (e.g., nacl changes)

packages/jira-adapter/src/filters/fields/context_options.ts
idoamit8: options is still list of lists, so I think I need to have `flat()` as well. 
`const values = options.flat().filter(isOption)`

PR #4582, opened 2023-07-06
SALTO-4458 improve path index hints generation consistency

Uri-Bar:
SALTO-4458 improve path index hints generation consistency
we added for keys to the pathIndex such as .annotation .field and .attr elemId (e.g: salto.obj.field).
now when the field of an object are divided between two files `salto.obj.field`   will point to both of them.
And if the few annotation under the same file, we wont have a key of each annotation in the pathIndex. instead `salto.obj.annotation` will point at the relevant file.

---
_Release Notes_: 
none

---
_User Notifications_: 
none

### DIFF
--- a/packages/workspace/src/workspace/path_index.ts
+++ b/packages/workspace/src/workspace/path_index.ts
@@ -92,7 +92,14 @@ const getAnnotationTypesPathHints = (
 const getAnnotationPathHints = (
   fragments: Fragment<Element>[],
 ): PathHint[] => {
+  const elem = fragments[0].value
   const fragmentsWithFields = fragments.filter(f => !_.isEmpty(f.value.annotations))
+  if (isInstanceElement(elem)) {
+    return getValuePathHints(
+      fragments.map(f => ({ value: f.value.annotations, path: f.path })),
+      elem.elemID,
+    )
+  }
   if (_.isEmpty(fragmentsWithFields)) {
     return []
   }
@@ -103,10 +110,9 @@ const getAnnotationPathHints = (
   if (fragmentsWithFields.length === 1) {
     return attrTopLevelKey
   }
-  const elem = fragments[0].value
   return attrTopLevelKey.concat(getValuePathHints(
     fragments.map(f => ({ value: f.value.annotations, path: f.path })),
-    isInstanceElement(elem) ? elem.elemID : elem.elemID.createNestedID('attr'),
+    elem.elemID.createNestedID('attr'),
   ))
 }
 
@@ -309,6 +315,10 @@ export const splitElementByPath = async (
   element: Element,
   index: PathIndex
 ): Promise<Element[]> => {
+  const indexEntries = await awu(index.entries()).toArray()
+  if (!indexEntries) {
+    return [element]
+  }
   const pathHints = await getFromPathIndex(element.elemID, index)
   if (pathHints.length <= 1) {
     const clonedElement = element.clone()
--- a/packages/workspace/src/workspace/path_index.ts
+++ b/packages/workspace/src/workspace/path_index.ts
@@ -43,14 +43,14 @@ export type PathIndex = RemoteMap<Path[]>
 const getValuePathHints = (fragments: Fragment<Value>[], elemID: ElemID): PathHint[] => {
   // We only have 3 cases to handle: Object type (which can be split among files)
   // or a single list/primitive value.
-  const valueTopLevelKey = [{
-    key: elemID.getFullName(),
-    value: makeArray(fragments.map(f => f.path)),
-  }]
   if (fragments.length === 0) {
     // No fragments with this ID, so no path hints needed
     return []
   }
+  const valueTopLevelKey = [{
+    key: elemID.getFullName(),
+    value: makeArray(fragments.map(f => f.path)),
+  }]
   if (fragments.length === 1) {
     return valueTopLevelKey
   }
--- a/packages/workspace/src/workspace/path_index.ts
+++ b/packages/workspace/src/workspace/path_index.ts
@@ -21,7 +21,6 @@ import {
   Field,
   InstanceElement,
   isInstanceElement,
-  isObjectType,
   ObjectType,
   Value,
 } from '@salto-io/adapter-api'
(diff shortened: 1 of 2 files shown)


ori-moisis:
does this make any effect on users?
not sure I understand the release notes...

packages/workspace/src/workspace/path_index.ts
```diff
@@ -69,28 +71,44 @@ const getAnnotationTypesPathHints = (
   const fragmentsWithNonEmptyAnnoTypes = fragments.filter(
     f => !_.isEmpty(f.value.annotationRefTypes)
   )
-
+  if (_.isEmpty(fragmentsWithNonEmptyAnnoTypes)) {
```
ori-moisis, line 74: can we reduce the use of double negatives here by renaming `fragmentsWithNonEmptyAnnoTypes` to just `fragmentsWithAnnoTypes`

asking 'isEmpty` on something with "NonEmpty" just seems weird

packages/workspace/src/workspace/path_index.ts
```diff
@@ -154,19 +171,22 @@ const getElementPathHints = (
   const valueHints = elementFragments.every(f => isInstanceElement(f.value))
     ? getValuePathHints(
       (elementFragments as Fragment<InstanceElement>[])
-        .map(f => ({ value: f.value.value, path: f.path })),
+        .map(f => ({ value: _.merge({}, f.value.value, f.value.annotations), path: f.path })),
```
ori-moisis, line 174: similar note about needing a comment to explain what we are doing with instance elements and why.
maybe we should couple the two conditions so we only have to explain this once?

packages/workspace/src/workspace/path_index.ts
```diff
@@ -69,28 +71,44 @@ const getAnnotationTypesPathHints = (
...
 
 const getAnnotationPathHints = (
   fragments: Fragment<Element>[],
 ): PathHint[] => {
   const elem = fragments[0].value
+  const fragmentsWithAttr = fragments.filter(f => !_.isEmpty(f.value.annotations))
+  if (_.isEmpty(fragmentsWithAttr)) {
+    return []
+  }
+  if (isInstanceElement(elem)) {
```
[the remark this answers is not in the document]
Uri-Bar: ack

packages/workspace/src/workspace/path_index.ts
Uri-Bar: ack

packages/workspace/src/workspace/path_index.ts
```diff
@@ -69,28 +71,44 @@ const getAnnotationTypesPathHints = (
...
+    return []
+  }
+  if (fragmentsWithAttr.length === 1) {
```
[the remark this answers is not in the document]
Uri-Bar: ack will remove and test it

packages/workspace/src/workspace/path_index.ts
```diff
@@ -43,97 +43,99 @@ export type PathIndex = RemoteMap<Path[]>
...
   // We only have 3 cases to handle: Object type (which can be split among files)
   // or a single list/primitive value.
+  const valueTopLevelKey = [{
+    key: elemID.getFullName(),
+    value: makeArray(fragments.map(f => f.path)),
+  }]
+  if (fragments.length === 0) {
+    // No fragments with this ID, so no path hints needed
+    return []
+  }
```
ori-moisis, line 53: nit - move this the be before creating the `valueTopLevelKey`

packages/workspace/test/workspace/path_index.test.ts
```diff
@@ -536,3 +536,238 @@ describe('updatePathIndex', () => {
...
+    expect(hintsKeys).toContain('salto.obj')
+    const fieldHints = pathHints.filter(p => p.key.includes('field'))
+    expect(hintsKeys).toContain('salto.obj.field')
+    expect(fieldHints.length).toEqual(1)
+    expect(fieldHints[0].value.length).toEqual(1)
+    const annoHints = pathHints.filter(p => p.key.includes('annotation'))
+    expect(hintsKeys).toContain('salto.obj.annotation')
+    expect(annoHints.length).toEqual(1)
+    expect(annoHints[0].value.length).toEqual(1)
+    expect(hintsKeys).toContain('salto.obj.attr')
```
ori-moisis, line 589: I think it is a bit hard to follow the `expect`s here
I get lost with all the "length" checks, it also doesn't actually validate we get the correct paths.

two options I see here:
1. Turn this into an object and then check for each key and value (or all together), e.g `const pathHintObj = { ...pathHints.map(hint => ({ [hint.key]: hint.value }) }` and then `expect(pathHintObj).toHaveProperty('salto.obj', [...])`
2. Use `expect(pathHints).toContain(expect.objectMatching(...))` with the specific paths we are looking for

packages/workspace/test/workspace/path_index.test.ts
```diff
@@ -536,3 +536,227 @@ describe('updatePathIndex', () => {
...
+      annotationRefsOrTypes: {
+        anno: createRefToElmWithValue(BuiltinTypes.STRING),
+      },
+      annotations: {
+        anno: 'Bye',
+      },
+      path: ['salto', 'obj', 'annotations'],
+    })
+    const pathHints = getElementsPathHints([singleFieldObj, singleFieldObjAnnotations])
+    expect(pathHints.length).toEqual(4)
```
ori-moisis, line 579: I think we don't need this check because this should be covered by the "toEqual" check below IIUC, but for future reference, when asserting the length of an array:
```suggestion
    expect(pathHints).toHaveLength(4)
```

packages/workspace/test/workspace/path_index.test.ts
```diff
@@ -536,3 +536,227 @@ describe('updatePathIndex', () => {
...
+        ping: 'pong',
+      },
+      path: ['salto', 'obj', 'annotationsTwo'],
+    })
+    const pathHints = getElementsPathHints([
+      objFragAnnotationsOne, objFragAnnotationsTwo, objFragStdFields, objFragCustomFields,
+    ])
+    expect(pathHints.length).toEqual(10)
+    expect(pathHints).toEqual(expect.arrayContaining([
+      expect.objectContaining({ key: 'salto.obj.annotation', value: [['salto', 'obj', 'annotationsOne'], ['salto', 'obj', 'annotationsTwo']] }),
```
ori-moisis, line 638: I think extracting these paths to variables would make the test more readable

e.g.
```suggestion
      expect.objectContaining({ key: 'salto.obj.annotation', value: [annotationsOnePath, annotationsTwoPath] }),
```

ori-moisis (approved):

ori-moisis:
I think we should have more information about what this PR does in the PR description.
the description currently uses very generic words...

PR #4584, opened 2023-07-06
SALTO-3740 Add alias annotation to netsuite elements

dantal4:
Can be enabled with `netsuite { fetch = { addAlias = true } }` 

---

---
_Release Notes_: 
Netsuite Adapter:
- Add alias annotation to all elements.

---
_User Notifications_: 
None


edenhassid (commented):
alias filter looks good to me 😄 leaving final approve to @omrilit

omrilit (commented):
GREAT work!

omrilit (approved):

PR #4585, opened 2023-07-07
SALTO-4463 - Salesforce: Remove CurrencyIsoCodes field of created types

tomsellek:
If the user creates a new object, we can't keep the reference to our synthetic `CurrencyIsoCodes` type because Salesforce doesn't know about it.
We also can't just replace it with a regular value set because we get the error `You cannot define picklist, globalPicklist, or valueSet on a standard field. Use StandardValueSet instead.`.
We also can't use `StandardValueSet` because per [this doc](https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/standardvalueset_names.htm) there is no standard value set for currency codes.
So we just remove the annotation and let Salesforce figure it out. Which it does, according to my tests.

---

---
_Release Notes_: 
Salesforce: Fixed an issue with deploying new custom object types to environments where multi-currency is enabled.

---
_User Notifications_: 
N/A


packages/salesforce-adapter/src/filters/currency_iso_code.ts
```diff
@@ -103,7 +106,7 @@ const createCurrencyCodesInstance = (supportedCurrencies?: ValueSet): InstanceEl
  * Build a global list of available currency code, and a replace all the explicit ValueSets
  * with ValueSetName which points to it
  */
-const filterCreator: LocalFilterCreator = () => ({
+const filterCreator: LocalFilterCreator = (/* { config } */) => ({
```
ori-moisis, line 109: why?

packages/salesforce-adapter/src/filters/currency_iso_code.ts
```diff
@@ -118,6 +121,18 @@ const filterCreator: LocalFilterCreator = () => ({
...
+  preDeploy: async changes => {
+    const objectTypesWithIsoCodeFields = changes
+      .filter(isAdditionChange)
+      .filter(isObjectTypeChange)
+      .map(getChangeData)
+      .filter(objType => objType.fields?.CurrencyIsoCode?.annotations.valueSetName !== undefined)
+
+    objectTypesWithIsoCodeFields
+      .forEach(objType => {
+        delete objType.fields.CurrencyIsoCode?.annotations.valueSetName
```
ori-moisis, line 133: don't we need to put it back in onDeploy to avoid doing bad things to the state?

packages/salesforce-adapter/src/filters/currency_iso_code.ts
```diff
@@ -118,6 +121,18 @@ const filterCreator: LocalFilterCreator = () => ({
     elements.push(currencyCodeType, currencyCodeInstance)
     affectedElements.forEach(element => transformCurrencyIsoCodes(element, currencyCodeInstance))
   },
+  preDeploy: async changes => {
+    const objectTypesWithIsoCodeFields = changes
```
ori-moisis, line 125: do we not need to handle field changes as well?

ori-moisis (commented):
missing tests in this PR

packages/salesforce-adapter/src/filters/currency_iso_code.ts
tomsellek: Ack

packages/salesforce-adapter/src/filters/currency_iso_code.ts
ori-moisis: we can't change things like helpText / description?
is there anything preventing me from editing something in the field?

ori-moisis (approved):
in case changes to the field are really impossible I think this can be good enough.
I'd like to avoid crashing in that case as well, but if it can't happen without editing NaCl I guess we can keep going without handling this for a while

packages/salesforce-adapter/src/filters/currency_iso_code.ts
tomsellek: This is what the field looks like in the NaCl of the instance:
```
  salesforce.Picklist CurrencyIsoCode {
    apiName = "Account.CurrencyIsoCode"
    restricted = true
    label = "Account Currency"
    createable = true
    updateable = true
    queryable = true
    valueSetName = salesforce.CurrencyIsoCodes.instance
  }
```

Changing the label (which seems like the only modifiable annotation) resulted in the deploy error:
```
    Modification of standard field labels is not supported
    salesforce/Objects/Account/AccountStandardFields.nacl(line: 575)
    Error: Standard field ‘CurrencyIsoCode’ label cannot be modified
```

ori-moisis: yes, a label on a standard field is not editable, I think that is correct for all standard fields
but IIRC description / helpText is usually editable even on standard fields

PR #4587, opened 2023-07-08
SALTO-4305: Create references to custom record fields in suitescripts

Shulik95:
_Create references to custom record fields in suitescripts_

---

_The ticket: https://salto-io.atlassian.net/browse/SALTO-4305_
_Currentrly this pr contains a "softer" version which only logs the new found references and doesn't add them_

---
_Release Notes_: 
_Netsuite Adapter_:
* custom record field references in suitescripts will be logged.

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/netsuite-adapter/src/elements_source_index/elements_source_index.ts
+++ b/packages/netsuite-adapter/src/elements_source_index/elements_source_index.ts
@@ -125,8 +125,8 @@ const createIndexes = async (elementsSource: ReadOnlyElementsSource, isPartial: 
   }
 
   const updateCustomRecordFieldsIndex = (customObjectType: ObjectType): void => {
-    const serviceIdRecords = extractCustomRecordFields(customObjectType)
-    _.assign(serviceIdRecordsIndex, serviceIdRecords)
+    const serviceIdRecords = _.keyBy(extractCustomRecordFields(customObjectType), 'serviceID')
+    _.assign(customRecordFieldsServiceIdRecordsIndex, serviceIdRecords)
   }
 
   const deletedElementSet = new Set(deletedElements.map(elemId => elemId.getFullName()))
--- a/packages/netsuite-adapter/src/elements_source_index/elements_source_index.ts
+++ b/packages/netsuite-adapter/src/elements_source_index/elements_source_index.ts
@@ -14,13 +14,13 @@
 * limitations under the License.
 */
 import _ from 'lodash'
-import { CORE_ANNOTATIONS, InstanceElement, isInstanceElement, ReadOnlyElementsSource, ElemID, Element, isObjectType, Value } from '@salto-io/adapter-api'
+import { CORE_ANNOTATIONS, InstanceElement, isInstanceElement, ReadOnlyElementsSource, ElemID, Element, isObjectType, Value, ObjectType } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
 import { collections } from '@salto-io/lowerdash'
 import { getElementValueOrAnnotations, isCustomRecordType } from '../types'
 import { ElementsSourceIndexes, LazyElementsSourceIndexes, ServiceIdRecords } from './types'
 import { getFieldInstanceTypes } from '../data_elements/custom_fields'
-import { getElementServiceIdRecords } from '../filters/element_references'
+import { extractCustomRecordFields, getElementServiceIdRecords } from '../filters/element_references'
 import { CUSTOM_LIST, CUSTOM_RECORD_TYPE, INTERNAL_ID, IS_SUB_INSTANCE } from '../constants'
 import { TYPES_TO_INTERNAL_ID } from '../data_elements/types'
 
@@ -89,6 +89,7 @@ const createIndexes = async (elementsSource: ReadOnlyElementsSource, isPartial: 
   const customFieldsIndex: Record<string, InstanceElement[]> = {}
   const elemIdToChangeByIndex: Record<string, string> = {}
   const elemIdToChangeAtIndex: Record<string, string> = {}
+  const customRecordFieldsServiceIdRecordsIndex: ServiceIdRecords = {}
 
   const updateInternalIdsIndex = async (element: Element): Promise<void> => {
     await assignToInternalIdsIndex(element, internalIdsIndex, elementsSource)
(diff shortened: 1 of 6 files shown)


packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -265,12 +293,15 @@ const filterCreator: LocalFilterCreator = ({
       await generateServiceIdToElemID(elements),
       await createElementsSourceServiceIdToElemID(elementsSourceIndex, isPartial)
     )
+    // This should be unified with serviceIdToElemID once SALTO-4305 is communicated.
```
dantal4, line 296: Why should it be unified with `serviceIdToElemID`? We don't want to extract it in sdf instances, right? Maybe we can add it as generated dependencies there too.

packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -265,12 +293,15 @@ const filterCreator: LocalFilterCreator = ({
...
+    const customRecordFieldsToServiceIds = createCustomRecordFieldsToElemID(elements)
```
dantal4, line 297: We should also get the fields service ids from the `elementsSourceIndex` in case of a partial fetch.

dantal4 (commented):
Great work!

packages/netsuite-adapter/test/filters/element_references.test.ts
```diff
@@ -539,6 +540,35 @@ describe('instance_references filter', () => {
         }
       )
     })
+    // TODO: should be uncommented when SALTO-4305 is communicated and opened to all
+    // it('should add generated dependency for custom record fields referenced by field ID', async () => {
```
dantal4, line 544: I would add that test and check that the field dependency isn't created, and when we open it this should break and then to fix it.

packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -254,23 +267,56 @@ const applyValuesAndAnnotationsToElement = (element: Element, newElement: Elemen
...
+    : {})
+
 const filterCreator: LocalFilterCreator = ({
   elementsSourceIndex,
   isPartial,
   changesGroupId,
 }) => ({
   name: 'replaceElementReferences',
   onFetch: async elements => {
+    const elementsSourceIndexes = await elementsSourceIndex.getIndexes()
```
dantal4, line 304: Calling `getIndexes` is a major time consumer. We should do it only in partial fetch. Notice that once its called, calling it again returns the already-calculated object, so it's fine to have several `await elementsSourceIndex.getIndexes()` in the code, as long that it's called on a partial fetch.

packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -254,23 +267,56 @@ const applyValuesAndAnnotationsToElement = (element: Element, newElement: Elemen
...
+}
+
+const getCustomRecordFieldsServiceId = (
+  elementsSourceIndexes: ElementsSourceIndexes,
+  isPartial: boolean
+): ServiceIdRecords =>
+  (isPartial
+    ? _.pickBy(
+      elementsSourceIndexes.serviceIdRecordsIndex,
+      serviceRecord => serviceRecord.serviceID?.startsWith(CUSTOM_FIELD_SERVICEID_PREFIX)
```
dantal4, line 293: `elementsSourceIndexes.serviceIdRecordsIndex` is equivalent to `serviceIdToElemID`, why would it hold custom fields references by themself (without the custom record type scriptid prefix)?

packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -254,23 +267,56 @@ const applyValuesAndAnnotationsToElement = (element: Element, newElement: Elemen
...
+  return _.keyBy(
+    elements
+      .filter(isObjectType)
+      .filter(isCustomRecordType)
+      .flatMap(extractCustomRecordFields),
+    'serviceID'
+  )
+}
+
+const getCustomRecordFieldsServiceId = (
```
dantal4, line 286: I suggest renaming it to be similar to `createElementsSourceServiceIdToElemID`:
`createElementsSourceCustomRecordFieldsToElemID`

dantal4 (commented):
Looks good 👍

packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -254,6 +266,30 @@ const applyValuesAndAnnotationsToElement = (element: Element, newElement: Elemen
   element.annotations = newElement.annotations
 }
 
+export const extractCustomRecordFields = (customRecordType: ObjectType): ElemServiceID[] =>
+  Object.values(customRecordType.fields)
+    .map(field => ({ serviceID: getServiceId(field), elemID: field.elemID }))
```
dantal4, line 271: We usually set the references to the `scriptid` value in the element
```suggestion
    .map(field => ({ serviceID: field.annotations[SCRIPT_ID], elemID: field.elemID.createNestedID(SCRIPT_ID) }))
```

packages/netsuite-adapter/src/elements_source_index/elements_source_index.ts
```diff
@@ -123,6 +124,11 @@ const createIndexes = async (elementsSource: ReadOnlyElementsSource, isPartial:
     }
   }
 
+  const updateCustomRecordFieldsIndex = (customObjectType: ObjectType): void => {
+    const serviceIdRecords = extractCustomRecordFields(customObjectType)
+    _.assign(serviceIdRecordsIndex, serviceIdRecords)
```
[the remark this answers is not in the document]
Shulik95: Yeah, this was meant to update the customRecordFieldIndex 🤦

packages/netsuite-adapter/test/elements_source_index.test.ts
```diff
@@ -231,4 +231,26 @@ describe('createElementsSourceIndex', () => {
...
+          annotations: {
+            [SCRIPT_ID]: 'custom_field',
+          },
+        },
+      },
+      annotations: {
+        [METADATA_TYPE]: CUSTOM_RECORD_TYPE,
+      },
+    })
+    const custRecordField = new Field(custRecordType, 'custom_field', BuiltinTypes.STRING, { scriptid: 'custom_field' })
```
dantal4, line 249: nit
```suggestion
    const custRecordField = custRecordType.fields.custom_field
```

dantal4 (approved):
Great work!

PR #4483, opened 2023-06-11
SALTO-4321: Include/ Exclude list doesn't work for private API types

shir-reifenberg:
Fix a bug causing config validation error when trying to include / exclude a type defined with ducktype api definitions (=Okta private API types)

---
_Release Notes_: 

_Okta adapter_:
-  Fix a bug in `include` and `exclude` for Okta types

---
_User Notifications_: 
None

PR #4588, opened 2023-07-08
SALTO-4028: Support ScriptRunner Authentication

IdoZakk:
_Also covered ticket [2525](https://salto-io.atlassian.net/browse/SALTO-2525)- infra to allow ducktype in JIra and Authentication for scriptRunner _

---

Based on Okta's PRs [1](https://github.com/salto-io/salto/pull/4170),[2](https://github.com/salto-io/salto/pull/4483), 
Currently added Script Listeners to make the code compile and to be able to test, but the script listeners need some work before we can issue it out. 

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/jira-adapter/src/adapter.ts
```diff
@@ -392,19 +400,58 @@ export default class JiraAdapter implements AdapterOperations {
     })
   }
 
-  @logDuration('fetching account configuration')
-  async fetch({ progressReporter }: FetchOptions): Promise<FetchResult> {
+  @logDuration('generating scriptRunner instances and types from service')
+  private async getScriptRunnerElements(): Promise<elementUtils.FetchElements<Element[]>> {
```
alonstern, line 404: We probably don't want this to affect customers so either add a config flag to run this or comment the call for it out before merging

packages/jira-adapter/src/adapter.ts
```diff
@@ -392,19 +400,58 @@ export default class JiraAdapter implements AdapterOperations {
...
+    const { scriptRunnerApiDefinitions } = this.userConfig
+    if (this.scriptRunnerClient === undefined
```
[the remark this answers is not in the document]
IdoZakk: I did not want to block it to allow it in the future.
Currently the `scriptRunnerApiDefinitions` is undefined in DC so it will not run

packages/jira-adapter/src/adapter.ts
IdoZakk: Yes, I will do it, kept it inside to allow tests, will also add to context

packages/jira-adapter/src/adapter.ts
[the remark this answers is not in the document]
netama: shouldn't we clarify this with a comment (if we're not making it more explicit in the code)?

packages/jira-adapter/src/client/script_runner_connection.ts
```diff
@@ -0,0 +1,28 @@
...
+import ScriptRunnerCredentials from '../script_runner_auth'
+
+export const createScriptRunnerConnection: clientUtils.ConnectionCreator<ScriptRunnerCredentials> = retryOptions => (
+  clientUtils.axiosConnection({
+    retryOptions,
+    authParamsFunc: async _credentials => (
+      {}
+    ),
+    baseURLFunc: async credentials => credentials.getBaseUrl(),
+    credValidateFunc: async () => ({ accountId: '' }), // There is no login endpoint to call
```
netama, line 26: we usually just call some (ideally small) endpoint in these cases to make sure the credentials are correct, is that not an option here?

packages/jira-adapter/src/client/script_runner_client.ts
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  public async getSinglePage(
+    args: clientUtils.ClientBaseParams,
+  ): Promise<clientUtils.Response<clientUtils.ResponseValue | clientUtils.ResponseValue[]>> {
+    try {
+      return await super.getSinglePage({
+        ...args,
+        headers: {
+          ...await this.getAuthHeader(),
```
netama, line 129: why add it every time and not just add this to the connection?

packages/jira-adapter/src/script_runner_auth.ts
```diff
@@ -0,0 +1,70 @@
...
+  url: string
+}
+
+const TOKEN_ADDRESS_RESPONSE_SCHEME = Joi.object({
+  url: Joi.string().required(),
+}).unknown(true)
+
+const isTokenAddressResponse = createSchemeGuard<TokenAddressResponse>(TOKEN_ADDRESS_RESPONSE_SCHEME, 'Failed to get script runner token from jira service')
+
+export default class ScriptRunnerCredentials {
```
netama, line 34: in general there's a preference not to use classes for credentials (at least was in the past by @ori-moisis ). can't we add this logic in the connection instead, like we do in other places?

packages/jira-adapter/src/client/script_runner_connection.ts
IdoZakk: This is done the same way in Jira connection
Specifically for scriptRunner we get the jwt only at a later stage

packages/jira-adapter/src/client/script_runner_client.ts
```diff
@@ -0,0 +1,198 @@
...
+const isTokenResponse = createSchemeGuard<TokenResponse>(TOKEN_RESPONSE_SCHEME, 'Failed to get script runner token from scriptRunner service')
+
+const getSrTokenFromHtml = (html: string): string => {
+  const root = parse(html)
+
+  // Find the meta tag with name="sr-token"
+  const srTokenElement = root.querySelector('meta[name="sr-token"]')
+  if (srTokenElement === null) {
+    log.error('Failed to get script runner token from scriptRunner service, could not find meta tag with name="sr-token"')
+    return ''
```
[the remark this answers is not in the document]
IdoZakk: After consulting with @alonstern we do not want to fail the entire jira fetch in case we fail authenticating scriptRunner. I will add comments about it

packages/jira-adapter/src/client/script_runner_client.ts
IdoZakk: To get the token we need to access the script runner server. I preferred doing it using the scriptRunnerClient. The other alternative I could think of would be to create a temporary client inside the credentials, seems more problematic and messy to me

packages/jira-adapter/src/script_runner_auth.ts
IdoZakk: The problem is we need to do an async call to know the base URL, and the credentials are created in a sync function.
If we do it in the script_runner_connection we will have to store at that point information to the credential object (both base URL and token URL), making it mutable which sounds less good to me

Of course, I can turn the ScritRunnerCredentials class into a type with functions, but I think a class is the proper way here given the behavior.

Adding logic to the credential was my last choice, I am of course open to other suggestions

packages/jira-adapter/src/config/config.ts
```diff
@@ -248,3 +262,22 @@ export const getApiDefinitions = (config: JiraApiConfig): {
...
+export const validateJiraFetchConfig = ({
+  fetchConfig,
+  apiDefinitions,
+  scriptRunnerApiDefinitions,
+}: {
+  fetchConfig: JiraFetchConfig
+  apiDefinitions: JiraApiConfig
+  scriptRunnerApiDefinitions: JiraDuckTypeConfig
+}): void => {
+  const supportedTypes = fetchConfig.enableScriptRunnerAddon === false
```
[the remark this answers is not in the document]
IdoZakk: It is, I will change the order

packages/jira-adapter/src/script_runner_auth.ts
netama: commented above - I think you can use the `@requiresLogin` decorator for it, if you're not sure how to use it let's discuss (maybe I'm missing something)

packages/jira-adapter/src/adapter.ts
netama: should this still be addressed?

packages/jira-adapter/src/adapter.ts
```diff
@@ -392,19 +400,61 @@ export default class JiraAdapter implements AdapterOperations {
...
+
+  @logDuration('generating instances from service')
+  private async getAllJiraElements(
+    progressReporter: ProgressReporter,
+    swaggers: AdapterSwaggers
+  ): Promise<elementUtils.FetchElements<Element[]>> {
     log.debug('going to fetch jira account configuration..')
-    const swaggers = await this.generateSwaggers()
     progressReporter.reportProgress({ message: 'Fetching types' })
     const { allTypes, parsedConfigs } = await this.getAllTypes(swaggers)
```
alonstern, line 439: nit: now it's `swaggerTypes` and not all types right?

packages/jira-adapter/src/auth.ts
```diff
@@ -44,3 +44,4 @@ export const basicAuthCredentialsType = createMatchingObjectType<BasicAuthCreden
 })
 
 export type Credentials = BasicAuthCredentials
+export type ScriptRunnerCredentials = {}
```
alonstern, line 47: type `{}` will match anything that is not null, is that what you meant to do here?

packages/jira-adapter/src/client/script_runner_client.ts
```diff
@@ -0,0 +1,114 @@
...
+
+const DEFAULT_PAGE_SIZE: Required<clientUtils.ClientPageSizeConfig> = {
+  get: 1000,
+}
+
+export default class ScriptRunnerClient extends clientUtils.AdapterHTTPClient<
+  ScriptRunnerCredentials, clientUtils.ClientRateLimitConfig
+> {
+  readonly isDataCenter: boolean
+  jwTokenPromise: Promise<string> | undefined
```
alonstern, line 47: It looks like this is not used

alonstern (commented):
LGTM 👍  @netama do you want to have another look?

IdoZakk:
@alonstern I pushed fixes for your comments if you want to check them

alonstern (approved):

PR #4589, opened 2023-07-09
SALTO-4216 Return Salto Errors on NS deploy

dantal4:
Return `SaltoError` / `SaltoElementError` errors in NetSuite deploy.
Currently returning also `sdfErrors` and processing them in `client_validation` - we should reuse the same code for the deploy.

---

---
_Release Notes_: 
Netsuite Adapter:
- Return Salto Errors on deploy

---
_User Notifications_: 
None


packages/netsuite-adapter/src/adapter.ts
```diff
@@ -481,10 +481,12 @@ export default class NetsuiteAdapter implements AdapterOperations {
...
     await filtersRunner.onDeploy(appliedChanges, deployResult)
 
-    return {
-      errors: deployResult.errors,
-      appliedChanges,
-    }
+    const originalChangesElemIds = new Set(changes.map(change =>
+      getChangeData(change).elemID.getFullName()))
+    const errors = deployResult.errors.filter(error => !('elemID' in error)
+      || originalChangesElemIds.has(error.elemID.getFullName()))
```
dantal4, line 487: Add log for dropped errors.

dantal4, line 487: Add log for dropped errors.

packages/netsuite-adapter/src/suiteapp_config_elements.ts
```diff
@@ -77,26 +76,36 @@ export const toConfigDeployResult = (
...
-
-  const missingResultsError = changes.length !== results.length
-    ? [new Error('Missing deploy result for some changes')]
-    : []
+  const missingResultsError = changes
+    .map(getChangeData)
+    .filter(instance => !resultsConfigTypes.has(instance.value.configType))
+    .map(instance => ({
+      elemID: instance.elemID,
+      message: 'Missing deploy result for this instance',
```
omrilit, line 100: Is it the message that will be exposed to the user? Seems like an internal error to me.

omrilit (approved):
Looks great!

packages/netsuite-adapter/src/adapter.ts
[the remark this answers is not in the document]
dantal4: We currently set deploy errors only to base-level elem ids, but I can change it to `error.elemID.createBaseId()...` to catch also nested errors if we decide to return errors this way.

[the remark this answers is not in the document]
dantal4: I think that core&SAAS assume to receive `SaltoElementErrors` only on changes in the plan and I wouldn't want to break that assumption ATM.
The main case to filter here is a custom record type change in case that we originally wanted to deploy only a field of it - in this case we will get the errors on the fields.
In case of "auto added referenced elements", like custom segment with a custom record type - I think that the user won't be interested in an error on an element that they didn't want to deploy.

packages/netsuite-adapter/src/suiteapp_config_elements.ts
dantal4: It is an internal error. @omrilit WDYT should be the error in this case? (this is a super rare situation that I don't know if we'll ever run into it).

packages/netsuite-adapter/src/adapter.ts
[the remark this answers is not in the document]
omrilit: We agreed to return `SaltoError` for `SaltoElementErrors` that are not in `originalChangesElemIds`

packages/netsuite-adapter/src/suiteapp_config_elements.ts
omrilit: I think that in this case it would be better to just return an internal server error message, like we do in the SaaS

PR #4590, opened 2023-07-09
SALTO-3648 - Added a validator to prevent illegal changes to standard ticket fields

idoser96:
Validates that standard ticket fields are not added or removed or that their non-editable fields are not modified

---

It seems that some fields of standard ticket fields are uneditable under some conditions
I did not get into it because i did not see a failed deployment of the other fields so it might be better to wait and handle it if it pops

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/change_validators/standard_fields.ts
+++ b/packages/zendesk-adapter/src/change_validators/standard_fields.ts
@@ -1,0 +1,65 @@
+/*
+*                      Copyright 2023 Salto Labs Ltd.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


packages/zendesk-adapter/src/change_validators/standard_fields.ts
```diff
@@ -0,0 +1,69 @@
...
+    .filter(isInstanceChange)
+    .filter(change => getChangeData(change).elemID.typeName === TICKET_FIELD_TYPE_NAME)
+    .filter(change => STANDARD_FIELDS.includes(getChangeData(change).value.type))
+
+  const [modifications, additionsRemovals] = _.partition(
+    relevantChanges,
+    isModificationChange
+  )
+
+  const relevantModifications = modifications.filter(change => {
```
edenhassid, line 53: nit: instead of `relevantModifications` I would call it `invalidModifications`

edenhassid (commented):
looks good 😄   I think this will not work well with `ticket_status_custom_status.ts`. It will prevent the deployment.  we should remove it from the list, maybe there is another solution.

edenhassid (approved):
💯 🔥 🔥

PR #4592, opened 2023-07-09
[WIP] SALTO-4215 - Return SaltoErrors instead throwing (Okta)

shir-reifenberg:

---

---
_Release Notes_: 

---
_User Notifications_: 


packages/okta-adapter/src/filters/private_api_deploy.ts
```diff
@@ -43,15 +43,21 @@ const filterCreator: FilterCreator = ({ adminClient, config }) => ({
     }
     if (config[CLIENT_CONFIG]?.usePrivateAPI !== true) {
       log.debug('Skip deployment of private API types because private API is not enabled')
-      const error = new Error(`The following changes were not deployed, because usePrivateApi config option is disabled: ${relevantChanges.map(c => getChangeData(c).elemID.getFullName()).join(', ')}`)
+      const error: SaltoError = {
+        message: `The following changes were not deployed, because usePrivateApi config option is disabled: ${relevantChanges.map(c => getChangeData(c).elemID.getFullName()).join(', ')}`,
+        severity: 'Error',
+      }
```
benisraeladi, line 49: I think this should be a list of `SaltoElementError`s instead of one SaltoError

packages/okta-adapter/src/filters/private_api_deploy.ts
```diff
@@ -43,15 +43,21 @@ const filterCreator: FilterCreator = ({ adminClient, config }) => ({
...
         deployResult: { appliedChanges: [], errors: [error] },
       }
     }
     if (adminClient === undefined) {
       log.error('Skip deployment of private API types because adminClient does not exist')
-      const error = new Error(`The following changes were not deployed, due to error with the private API client: ${relevantChanges.map(c => getChangeData(c).elemID.getFullName()).join(', ')}`)
+      const error: SaltoError = {
+        message: `The following changes were not deployed, due to error with the private API client: ${relevantChanges.map(c => getChangeData(c).elemID.getFullName()).join(', ')}`,
+        severity: 'Error',
+      }
```
benisraeladi, line 60: This as well, although I'm not as sure here because this seems to indicate some error on the Okta side

packages/okta-adapter/src/logo.ts
```diff
@@ -105,9 +105,9 @@ export const deployLogo = async (
       const suffix = LOGO_TYPES_TO_VALUES[logoInstance.elemID.typeName].urlSuffix
       logoUrl = `/api/v1/brands/${brand.value.id}/themes/${brandTheme.value.id}/${suffix}`
     }
-    return await sendLogoRequest({ client, change, logoInstance, url: logoUrl, isRemoval: isRemovalChange(change) })
+    await sendLogoRequest({ client, change, logoInstance, url: logoUrl, isRemoval: isRemovalChange(change) })
   } catch (e) {
-    return getOktaError(logoInstance.elemID, e)
+    throw getOktaError(logoInstance.elemID, e)
```
benisraeladi, line 110: Shouldn't this return `getOktaError` and have the function return a `SaltoElementError`?

benisraeladi (commented):
Very nice!

packages/okta-adapter/src/filters/private_api_deploy.ts
shir-reifenberg: I prefer aggregating it to one error because there's a global config flag responsible for this

packages/okta-adapter/src/logo.ts
shir-reifenberg: I think this should throw so `deployChanges` would catch it

packages/okta-adapter/src/deployment.ts
```diff
@@ -293,25 +293,25 @@ export const deployEdges = async (
...
     changes.map(async change => {
       try {
         const res = await deployChangeFunc(change)
         return res !== undefined ? res : change
       } catch (err) {
-        if (!_.isError(err)) {
-          throw err
-        }
-        return err
+        errors.push({
```
edenhassid, line 307: at the end I didn't do it like this after the comment I got from @netama https://github.com/salto-io/salto/pull/4574#discussion_r1252970830. you can look at the PR to see how I changed the code

packages/okta-adapter/src/filters/expression_language.ts
```diff
@@ -228,9 +228,11 @@ const filter: FilterCreator = ({ config }) => {
                   createPrepRefFunc(isIdentityEngine),
                 )
               } catch (error) {
-                if (_.isError(error)) {
-                  log.error(`Error parsing templates in instance ${instance.elemID.getFullName()} before deployment: ${error.message}`)
-                  ErrorByChangeId[instance.elemID.getFullName()] = new Error(`Error parsing Okta expression language expression for instance ${instance.elemID.name} of type ${instance.elemID.typeName}`)
+                log.error(`Error parsing templates in instance ${instance.elemID.getFullName()} before deployment: ${error.message}`)
+                ErrorByChangeId[instance.elemID.getFullName()] = {
```
edenhassid, line 232: I added the function `createSaltoElementError` in my PR, maybe you can use it here once I merge the PR

edenhassid (commented):
looks good! added some comments :)

packages/okta-adapter/src/filters/private_api_deploy.ts
netama: hmm maybe we can have both? so that we also show next to each element why it wasn't deployed?
(@liorn this is another example for where grouping by error might help, not sure if we are tracking this anywhere :) )

netama (commented):
changes LGTM, I agree with @edenhassid we should probably align this to the zendesk changes + should probably decide with @liorn about whether to aggregate the private-API errors (I don't have a strong opinion)

packages/okta-adapter/src/filters/private_api_deploy.ts
benisraeladi: Discussed f2f

benisraeladi (approved):
🔥

edenhassid (approved):
LGTM 💯

PR #4595, opened 2023-07-10
Convert app_installation's app_id to its app owned (if exists)

idoser96:

---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/zendesk-adapter/test/filters/app.test.ts
+++ b/packages/zendesk-adapter/test/filters/app.test.ts
@@ -14,11 +14,11 @@
 * limitations under the License.
 */
 import {
-  ObjectType, ElemID, InstanceElement,
+  ObjectType, ElemID, InstanceElement, ReferenceExpression,
 } from '@salto-io/adapter-api'
 import { filterUtils } from '@salto-io/adapter-components'
 import ZendeskClient from '../../src/client/client'
-import { ZENDESK } from '../../src/constants'
+import { APP_OWNED_TYPE_NAME, ZENDESK } from '../../src/constants'
 import filterCreator, { APP_INSTALLATION_TYPE_NAME } from '../../src/filters/app'
 import { createFilterCreatorParams } from '../utils'
 
@@ -48,6 +48,13 @@ describe('app installation filter', () => {
       settings_objects: [{ name: 'name', value: 'My App' }, { name: 'title', value: 'My App' }],
     }
   )
+  const appOwned = new InstanceElement(
+    'appOwned',
+    new ObjectType({ elemID: new ElemID(ZENDESK, APP_OWNED_TYPE_NAME) }),
+    {
+      id: 1,
+    }
+  )
 
   beforeEach(async () => {
     jest.clearAllMocks()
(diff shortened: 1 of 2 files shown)


packages/zendesk-adapter/test/filters/app.test.ts
```diff
@@ -57,142 +64,159 @@ describe('app installation filter', () => {
...
-      change: { action: 'modify', data: { before: clonedBeforeApp, after: clonedAfterApp } },
-      client: expect.anything(),
-      endpointDetails: expect.anything(),
-      fieldsToIgnore: ['app', 'settings.title', 'settings_objects'],
+
+    it('should not modify app_id of app_installation if corresponding app_owned is not found', async () => {
+      const appInstallation = app.clone()
+      appInstallation.value.app_id = 2
+      await filter.onFetch([appInstallation, appOwned])
+      expect(appInstallation.value.app_id).toEqual(2)
```
idoser96, line 86: these two were added, everything else stayed the same

packages/zendesk-adapter/test/filters/app_installations.test.ts
```diff
@@ -0,0 +1,222 @@
...
+      const appInstallation = app.clone()
+      await filter.onFetch([appInstallation, appOwned])
+      expect(appInstallation.value.app_id).toEqual(new ReferenceExpression(appOwned.elemID, appOwned))
+    })
+
+    it('should not modify app_id of app_installation if corresponding app_owned is not found', async () => {
+      const appInstallation = app.clone()
+      appInstallation.value.app_id = 2
+      await filter.onFetch([appInstallation, appOwned])
+      expect(appInstallation.value.app_id).toEqual(2)
```
idoser96, line 86: These are new, all others stayed the same (and were put into 'fetch' and 'deploy' describe

packages/zendesk-adapter/src/config.ts
```diff
@@ -1156,8 +1156,8 @@ export const DEFAULT_TYPES: ZendeskApiConfig['types'] = {
         { fieldName: 'id', fieldType: 'number' },
       ),
       fieldsToOmit: FIELDS_TO_OMIT.concat({ fieldName: 'updated', fieldType: 'string' }),
-      idFields: ['settings.name', 'app_id'],
-      fileNameFields: ['settings.name', 'app_id'],
+      idFields: ['settings.name', 'product'],
```
idoser96, line 1159: Its ok for an app to have the same name in different products

It seems that it is possible to install some apps twice with the same name (but it will also have the same app_id, so this is a different issue that we will need to solve if it ever pops)

packages/zendesk-adapter/src/filters/app_installations.ts
```diff
@@ -89,17 +90,37 @@ Promise<void> => {
   )
 }
 
+const connectAppOwnedToInstallation = (instances: InstanceElement[]): void => {
+  const appOwnedById = _.keyBy<InstanceElement>(
+    instances
+      .filter(e => e.elemID.typeName === APP_OWNED_TYPE_NAME),
```
edenhassid, line 96: I think you should also filter ones which don't have id. (shouldn't happen but just in case)

packages/zendesk-adapter/src/filters/app_installations.ts
```diff
@@ -89,17 +90,37 @@ Promise<void> => {
...
+    e => e.value.id
+  )
+
+  instances
+    .filter(e => e.elemID.typeName === APP_INSTALLATION_TYPE_NAME)
+    .forEach(async appInstallation => {
```
edenhassid, line 102: why do you have async?

edenhassid (commented):
great!! 💯 added small comments

edenhassid (approved):
😎 🔥

PR #4597, opened 2023-07-10
Revert "SALTO-4322: Data Records Circular Dependencies Workaround"

tamtamirr:
Reverts salto-io/salto#4491 due to possible bug.
Will run more tests and re-open the PR


tamtamirr:
Rather merge a fix than revert. Closing

PR #4599, opened 2023-07-11
SALTO-4065: Remove duplicate roles, added default true

IdoZakk:
_The default was missing, added it_

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
None


IdoZakk:
Users that suffer from the bug and currently do not see this project role.
From the logs of the past week it seems relevant to 8 customers, 1 of them has change notifications. You think we should do noise reduction?

alonstern (approved):

PR #4487, opened 2023-06-12
Salto 1619/support cross service references workato jira

oneYAD:
In this PR I have added support for cross service references from Jira when fetching Workato account.

---
This PR is based on https://github.com/salto-io/salto/pull/4602 branch 

---

_Release Notes_: 
Workato adapter
- The users will be able to add jira connection to serviceConnectionNames (in workato.nacl) and get all Jira references in workaot recipes 

---
_User Notifications_: 
When using Workato. Jira connected recipe will include references to jira

PR #4600, opened 2023-07-11
workato add Joi support for recipeBlockTypes (cross-service)

oneYAD:
In this PR, I have added createBlockChecker which create recipe block checker from Joi Schema. 
In addition, I have been added 'provider' arg for the blockBase such that createBlockChecker can depend on it.
(All CrossService Blocks contanis provider anyway)

---
This PR is base branch for future SALTO-1619 PR. (we want to use Joi while validating Jira block)

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/workato-adapter/src/filters/cross_service/recipe_block_types.ts
+++ b/packages/workato-adapter/src/filters/cross_service/recipe_block_types.ts
@@ -14,7 +14,6 @@
 * limitations under the License.
 */
 import _ from 'lodash'
-// eslint-disable-next-line import/no-extraneous-dependencies
 import Joi from 'joi'
 import { createSchemeGuard } from '@salto-io/adapter-utils'
 
@@ -26,7 +25,7 @@ export type RefListItem = {
 
 export type BlockBase = {
   keyword: string
-  provider: string
+  provider?: string
 }
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
--- a/packages/workato-adapter/src/filters/cross_service/recipe_block_types.ts
+++ b/packages/workato-adapter/src/filters/cross_service/recipe_block_types.ts
@@ -14,6 +14,10 @@
 * limitations under the License.
 */
 import _ from 'lodash'
+// eslint-disable-next-line import/no-extraneous-dependencies
+import Joi from 'joi'
+import { createSchemeGuard } from '@salto-io/adapter-utils'
+
 
 export type RefListItem = {
   label: string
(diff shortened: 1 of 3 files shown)


packages/workato-adapter/src/filters/cross_service/recipe_block_types.ts
```diff
@@ -22,9 +26,21 @@ export type RefListItem = {
 
 export type BlockBase = {
   keyword: string
+  provider: string
```
alonstern, line 29: The PR looks good to me but @netama probably should verify that it's ok to add provider here

packages/workato-adapter/test/filters/recipe_block_type.test.ts
```diff
@@ -0,0 +1,92 @@
...
+      must : string
+      could? : number
+    }
+  }
+
+  let blockChecker :(value: unknown, application: string) => value is BlockType
+
+  describe('Cross service - recipe block type - create block checker', () => {
+    beforeAll(() => {
+      const BLOCK_SCHEMA = Joi.object({
```
alonstern, line 37: nit: only top level constant should be all capital letters

packages/workato-adapter/src/filters/cross_service/recipe_block_types.ts
```diff
@@ -14,6 +14,10 @@
 * limitations under the License.
 */
 import _ from 'lodash'
+// eslint-disable-next-line import/no-extraneous-dependencies
```
alonstern, line 17: We shouldn't add external libraries that are not in package json. You should add it to the package json and remove this line

packages/workato-adapter/test/filters/recipe_block_type.test.ts
```diff
@@ -0,0 +1,92 @@
...
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+// eslint-disable-next-line import/no-extraneous-dependencies
+import Joi from 'joi'
+import { BlockBase, createBlockChecker } from '../../src/filters/cross_service/recipe_block_types'
+
+/* eslint-disable camelcase */
```
alonstern, line 20: What is this for?

alonstern (approved):
👌

packages/workato-adapter/src/filters/cross_service/recipe_block_types.ts
netama: I'm not sure, let's mark as optional just in case?

netama:
we can merge this separately since it was already split, but I think it could have just fit inside https://github.com/salto-io/salto/pull/4487, as this is a helper function inside the workato code

PR #4601, opened 2023-07-11
SALTO-4306 make convertTypes filter synchronous

benisraeladi:
Shorten runtime of covertTypes filter by making it run synchronously

---

Made covertTypes filter synchronous by creating sync versions of `transformValues` and `transformPrimitive`

---

_Release Notes_:  _None_

---

_User Notifications_: _None_

### DIFF
--- a/packages/adapter-api/src/elements.ts
+++ b/packages/adapter-api/src/elements.ts
@@ -193,6 +193,14 @@ export class ListType<T extends TypeElement = TypeElement> extends Element {
     return refInnerTypeVal
   }
 
+  getInnerTypeSync(): TypeElement {
+    const refInnerTypeVal = this.refInnerType.getResolvedValueSync()
+    if (!isType(refInnerTypeVal)) {
+      throw new Error(`Cannot resolve ${this.elemID.getFullName()}'s innerType synchronously`)
+    }
+    return refInnerTypeVal
+  }
+
   setRefInnerType(innerTypeOrRefInnerType: TypeOrRef): void {
     if (innerTypeOrRefInnerType.elemID.isEqual(this.refInnerType.elemID)) {
       this.refInnerType = getRefType(innerTypeOrRefInnerType)
@@ -252,6 +260,15 @@ export class MapType<T extends TypeElement = TypeElement> extends Element {
     }
     return refInnerTypeVal
   }
+
+  getInnerTypeSync(): TypeElement {
+    const refInnerTypeVal = this.refInnerType.getResolvedValueSync()
+    if (!isType(refInnerTypeVal)) {
+      throw new Error(`Cannot resolve ${this.elemID.getFullName()}'s innerType synchronously`)
+    }
+    return refInnerTypeVal
+  }
+
 
   setRefInnerType(innerTypeOrRefInnerType: TypeOrRef): void {
     if (innerTypeOrRefInnerType.elemID.isEqual(this.refInnerType.elemID)) {
--- a/packages/adapter-api/src/elements.ts
+++ b/packages/adapter-api/src/elements.ts
@@ -302,6 +302,14 @@ export class Field extends Element {
     return type
   }
 
+  getTypeSync(): TypeElement {
+    const type = this.refType.getResolvedValueSync()
+    if (!isType(type)) {
+      throw new Error(`Element with ElemID ${this.elemID.getFullName()}'s type is resolved non-TypeElement`)
+    }
+    return type
+  }
+
   /**
    * Clones a field
    * Note that the cloned field still has the same element ID so it cannot be used in a different
@@ -476,6 +484,21 @@ export class InstanceElement extends Element {
     }
     return type
   }
+
+  getTypeSync(): ObjectType {
+    const type = this.refType.getResolvedValueSync()
+    if (!isObjectType(type)) {
+      log.warn(`Element with ElemID ${this.elemID.getFullName()}'s type is resolved non-ObjectType`)
+      if (type === undefined) {
+        throw new Error(`Element with ElemID ${this.elemID.getFullName()}'s type is undefined`)
+      }
+      return new PlaceholderObjectType({
+        elemID: this.elemID,
+      })
+    }
+    return type
+  }
+
 
   isEqual(
     other: InstanceElement,
--- a/packages/adapter-api/src/elements.ts
+++ b/packages/adapter-api/src/elements.ts
@@ -185,20 +185,11 @@ export class ListType<T extends TypeElement = TypeElement> extends Element {
   }
 
   async getInnerType(elementsSource?: ReadOnlyElementsSource): Promise<TypeElement> {
-    const refInnerTypeVal = await this.refInnerType.getResolvedValue(elementsSource)
-    // eslint-disable-next-line no-use-before-define
-    if (!isType(refInnerTypeVal)) {
-      throw new Error(`Element with ElemID ${this.elemID.getFullName()}'s innerType is resolved non-TypeElement`)
-    }
-    return refInnerTypeVal
+    return validateType(await this.refInnerType.getResolvedValue(elementsSource), this.elemID)
   }
 
   getInnerTypeSync(): TypeElement {
-    const refInnerTypeVal = this.refInnerType.getResolvedValueSync()
-    if (!isType(refInnerTypeVal)) {
-      throw new Error(`Element with ElemID ${this.elemID.getFullName()}'s innerType is resolved non-TypeElement`)
-    }
-    return refInnerTypeVal
+    return validateType(this.refInnerType.getResolvedValueSync(), this.elemID)
   }
 
   setRefInnerType(innerTypeOrRefInnerType: TypeOrRef): void {
@@ -253,20 +244,11 @@ export class MapType<T extends TypeElement = TypeElement> extends Element {
   }
 
   async getInnerType(elementsSource?: ReadOnlyElementsSource): Promise<TypeElement> {
-    const refInnerTypeVal = await this.refInnerType.getResolvedValue(elementsSource)
-    // eslint-disable-next-line no-use-before-define
-    if (!isType(refInnerTypeVal)) {
-      throw new Error(`Element with ElemID ${this.elemID.getFullName()}'s innerType is resolved non-TypeElement`)
-    }
-    return refInnerTypeVal
+    return validateType(await this.refInnerType.getResolvedValue(elementsSource), this.elemID)
   }
 
   getInnerTypeSync(): TypeElement {
-    const refInnerTypeVal = this.refInnerType.getResolvedValueSync()
-    if (!isType(refInnerTypeVal)) {
-      throw new Error(`Cannot resolve ${this.elemID.getFullName()}'s innerType synchronously as non-TypeElement`)
-    }
-    return refInnerTypeVal
+    return validateType(this.refInnerType.getResolvedValueSync(), this.elemID)
   }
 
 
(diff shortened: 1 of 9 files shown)


packages/adapter-utils/src/utils.ts
```diff
@@ -254,6 +256,156 @@ export const transformValues = async (
   return newVal
 }
 
+// TODOADI call this function only from transformValues if elementSource doesn't exist
+export const transformValuesSync = (
```
tomsellek, line 260: I dunno dude, this seems like a lot of code duplication. Is there a way to factor out the code that is common with the async variant?

packages/adapter-api/src/values.ts
```diff
@@ -84,6 +84,16 @@ type StaticFileMetadata = Pick<StaticFile, 'filepath' | 'hash'>
 export const getStaticFileUniqueName = ({ filepath, hash }: StaticFileMetadata): string =>
   `${filepath}-${hash}`
 
+const getResolvedValueSync = (
```
tomsellek, line 87: I think the name here, while accurate and logically linked to `getResolvedValue` may create some false expectations. This function never resolves the value - it either uses it as-is, or creates a placeholder. Maybe name it `createPlaceholderType` or something? It may even make sense to inline it to make things more obvious at the call site?

tomsellek (commented):
Did a sort of a high-level skim of the change and left some comments. I am a little concerned that the changes are in generic code but behave in a way that is different from the associated async variants - meaning, it would be natural for me to say "oh, cool, they added sync versions of these functions. Let me convert my code to use those!" and run into problems because I now get placeholder types with minimal content instead of the full type elements.

packages/adapter-utils/src/utils.ts
benisraeladi: I really wish there were :(
The problem with `transformValues` is that the async is so intertwined that it's almost impossible to find logic that can be factored out. If we decide to continue creating synchronous alternatives it could be a task to recreate this function so that it works in either a synchronous or asynchronous fashion, but it's such a fragile code area that I don't want to do this as part of this PR

packages/adapter-api/src/elements.ts
```diff
@@ -302,6 +302,14 @@ export class Field extends Element {
     return type
   }
 
+  getTypeSync(): TypeElement {
+    const type = this.refType.getResolvedValueSync()
+    if (!isType(type)) {
+      throw new Error(`Element with ElemID ${this.elemID.getFullName()}'s type is resolved non-TypeElement`)
```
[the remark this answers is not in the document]
benisraeladi: We should never get here. `getResolvedValueSync` will throw if `this.type` is undefined, and if it's defined then Typescript requires it to be of type `TypeElement`. It might even be redundant to have this `if` statement here but I wanted to keep the same guardrails as in the async version. Removing this check.

packages/adapter-api/src/elements.ts
```diff
@@ -477,6 +481,21 @@ export class InstanceElement extends Element {
     return type
   }
 
+  getTypeSync(): ObjectType {
+    const type = this.refType.getResolvedValueSync()
+    if (!isObjectType(type)) {
+      log.warn(`Element with ElemID ${this.elemID.getFullName()}'s type is resolved non-ObjectType`)
```
tamtamirr, line 487: Is this a possible scenario? Can you give an example when this would occur?

packages/adapter-api/src/values.ts
```diff
@@ -84,6 +84,16 @@ type StaticFileMetadata = Pick<StaticFile, 'filepath' | 'hash'>
...
+  elemID: ElemID,
+  resolvedValue?: Value
+): Value => {
+  if (resolvedValue === undefined) {
+    return new PlaceholderObjectType({ elemID })
+  }
+  return resolvedValue
```
tamtamirr, line 94:
```suggestion
  resolvedValue ?? new PlaceholderObjectType({ elemID })
```

packages/adapter-api/src/values.ts
```diff
@@ -196,6 +210,13 @@ export class TypeReference {
     return getResolvedValue(this.elemID, elementsSource, this.type)
   }
 
+  getResolvedValueSync(): TypeElement {
+    if (this.type === undefined) {
+      throw new Error(`Cannot resolve type reference ${this.elemID.getFullName()} without type`)
```
tamtamirr, line 215:
```suggestion
      throw new Error(`Type ${this.elemID.getFullName()} cannot be resolved synchronously`)
```

packages/adapter-utils/src/utils.ts
```diff
@@ -254,6 +259,160 @@ export const transformValues = async (
...
+  }: {
+    values: Value
+    type: ObjectType | TypeMap | MapType | ListType
+    transformFunc: TransformFuncSync
+    strict?: boolean
+    pathID?: ElemID
+    elementsSource?: ReadOnlyElementsSource
+    isTopLevel?: boolean
+    allowEmpty?: boolean
+  }
```
tamtamirr, line 286: We want **transformValues** and **transformValuesSync** to receive the same input. Can we make this a type and use it for both?

packages/adapter-api/src/values.ts
tamtamirr, line 87: Is this method really necessary? it's quite confusing IMO

tamtamirr (commented):
Looking good :)
Added some comments & questions

tamtamirr:
@benisraeladi I would remove the release notes here. Unless it's a major improvement (Like implementing quick fetch in SF), we shouldn't have this release notes per optimization we do in the adapter. There's probably more PRs around the same area that will be implemented soon.

packages/adapter-api/src/elements.ts
benisraeladi: It shouldn't be, but I would rather have an error thrown just in case

packages/adapter-utils/src/utils.ts
benisraeladi: We talked f2f about possibly removing `elementSource` from the sync version, and then it won't be the same input. Do you have a strong opinion in any direction?

packages/adapter-api/src/values.ts
benisraeladi: I wouldn't want to do the check to make sure that the type is defined in multiple places, so I think it's important to keep it for now at least

packages/adapter-utils/src/utils.ts
tamtamirr: I suggest the following:

1. Create type **TransformValuesInput**
2. Use the type as an input for **transformValues**
3. Use the type **Omit<TransformValuesInput, 'elementsSource'>** for **transformValuesSync**

tamtamirr (approved):
Good work 👑  Approving with [small comment](https://github.com/salto-io/salto/pull/4601#discussion_r1262384588) that needs to be addressed before merge.

packages/adapter-api/src/values.ts
```diff
@@ -84,6 +84,11 @@ type StaticFileMetadata = Pick<StaticFile, 'filepath' | 'hash'>
 export const getStaticFileUniqueName = ({ filepath, hash }: StaticFileMetadata): string =>
   `${filepath}-${hash}`
 
+const getResolvedValueSync = (
+  elemID: ElemID,
+  resolvedValue?: Value
+): Value => (resolvedValue ?? new PlaceholderObjectType({ elemID }))
```
ori-moisis: This is duplicated logic from `getResolvedValue`.
I don't think we have to duplicate this, we can have `getResolvedValue` use this function

but I think this is not actually needed at all because we don't really need to support sync resolve on reference expressions AFAICT

packages/adapter-api/src/values.ts
```diff
@@ -150,6 +155,10 @@ export class ReferenceExpression {
     return getResolvedValue(this.elemID, elementsSource, this.value)
   }
 
+  getResolvedValueSync(): Value {
```
ori-moisis: do we need this support on ReferenceExpression? why?

packages/adapter-utils/src/utils.ts
```diff
@@ -254,6 +259,160 @@ export const transformValues = async (
   return newVal
 }
 
+// IMPORTANT: this function will only work if transformFunc is synchronous.
+//  It is up to the caller to ensure that this is the case.
```
ori-moisis, line 265: but we enforce that with the type, why do we need this comment?

packages/adapter-utils/src/utils.ts
```diff
@@ -254,6 +259,160 @@ export const transformValues = async (
...
+//  they're also needed over at transformValues. The two functions are separated
+//  only because of the way async is intertwined with the logic.
+export const transformValuesSync = (
+  {
+    values,
+    type,
+    transformFunc,
+    strict = true,
+    pathID = undefined,
+    elementsSource,
```
ori-moisis: this is never actually used, why is this here?

packages/adapter-utils/src/utils.ts
ori-moisis: +1 on the general spirit of the comment, but I think we should prefer extending types over omitting stuff.
I think the type should be built from the things that are common to both functions (so, the input of the sync version) and the async version should add "elementSource" to that input instead of the sync version omitting the elementSource

packages/lowerdash/test/types.test.ts
```diff
@@ -135,4 +144,21 @@ describe('types', () => {
       expect(isNonEmptyArray([])).toEqual(false)
     })
   })
+  describe('NonPromise', () => {
```
ori-moisis, line 147: there is no way to actually write tests for types, they don't exist anymore when the test runs...
I think this block should be removed, it sets incorrect expectations on the behavior - the type should not be expected to "run the function", it is a type, it should not run anything

packages/adapter-api/src/elements.ts
```diff
@@ -477,6 +498,21 @@ export class InstanceElement extends Element {
...
+      if (type === undefined) {
+        throw new Error(`Element with ElemID ${this.elemID.getFullName()}'s type is undefined`)
+      }
```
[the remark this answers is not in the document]
ori-moisis: also, we can't actually get here, can we?
if the type is undefined, `getResolvedValueSync` would throw before we ever make it to this line

packages/adapter-api/src/elements.ts
benisraeladi: @ori-moisis Because of the asynchronicity, I'm not sure how not to duplicate this logic

packages/adapter-api/src/elements.ts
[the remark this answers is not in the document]
benisraeladi: That's right, this code is redundant now that the logic in `getResolvedValuesSync` is the same as in `getResolvedValues`

packages/adapter-api/src/elements.ts
ori-moisis: add a helper method that does the logic and call it from both the sync and async version after you get the type

or (probably worse), put the logic only the in the sync version and make the async version call the sync version (this would require allowing the sync version to get the resolved type as input, which is what I think makes this option worse)

packages/adapter-api/src/elements.ts
```diff
@@ -462,8 +460,7 @@ export class InstanceElement extends Element {
     this.refType = getRefType(typeOrRefType)
   }
 
-  async getType(elementsSource?: ReadOnlyElementsSource): Promise<ObjectType> {
-    const type = await this.refType.getResolvedValue(elementsSource)
+  objectTypeOrPlaceholder(type: TypeElement | Promise<TypeElement>): ObjectType {
```
ori-moisis, line 463: should not be public
should not accept a promise as input

packages/adapter-utils/src/utils.ts
```diff
@@ -107,6 +107,26 @@ export type TransformFuncArgs = {
...
+type TransformValuesBaseArgs = {
+  values: Value
+  type: ObjectType | TypeMap | MapType | ListType
+  strict?: boolean
+  pathID?: ElemID
+  isTopLevel?: boolean
+  allowEmpty?: boolean
+}
+
+type TransformValuesSyncArgs = TransformValuesBaseArgs & {transformFunc: TransformFuncSync}
```
ori-moisis, line 121:
```suggestion
type TransformValuesSyncArgs = TransformValuesBaseArgs & { transformFunc: TransformFuncSync }
```

ori-moisis (approved):
Assuming the work to reduce code duplication between the transformValues functions happens in a separate PR

PR #4606, opened 2023-07-12
SALTO-4336 - Zendesk - check formula is string before regexing it

idoser96:
When trying to convert a value to a template expression, make sure the the value is a string before running replace on it

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
None


netama (approved):
pre-approving based on IRL, let's refactor and log an error

idoser96:
will be covered by
https://github.com/salto-io/salto/pull/4622

PR #4608, opened 2023-07-12
SALTO-3203 return more readable error messages on NS SDF deploy and reuse in SDF validation

dantal4:
---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/netsuite-adapter/src/client/utils.ts
+++ b/packages/netsuite-adapter/src/client/utils.ts
@@ -65,7 +65,9 @@ export const sliceMessagesByRegex = (
   lookFromRegex: RegExp,
   includeMatchedRegex = true
 ): string[] => {
-  const matchedMessages = messages.map(message => lookFromRegex.test(message))
+  // remove the global flag of the regex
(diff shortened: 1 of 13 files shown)


packages/netsuite-adapter/src/client/utils.ts
```diff
@@ -60,11 +60,34 @@ export const getGroupItemFromRegex = (str: string, regex: RegExp, item: string):
     .filter(isDefined)
     .map(groups => groups[item])
 
+export const sliceMessagesByRegex = (
```
omrilit, line 63: Missing UT?

omrilit (approved):
Very nice work!!
BTW I think that this PR should be related to https://salto-io.atlassian.net/browse/SALTO-3203

packages/netsuite-adapter/src/client/utils.ts
dantal4: The usage of it is covered so I think it's good enough.

PR #4609, opened 2023-07-12
SALTO-4332 - Zendesk Adapter - Prevent automations with same conditions

idoser96:
Prevent deployment of an automation with the same conditions as another automation

---

None

---
_Release Notes_: 
Zendesk Adapter:
* Prevent deployment of automations with conditions equal to another automation 

---
_User Notifications_: 
None


packages/zendesk-adapter/src/change_validator.ts
```diff
@@ -148,6 +149,7 @@ export default ({
     featureActivation: featureActivationValidator,
     deflectionAction: deflectionActionValidator,
     standardFields: standardFieldsValidator,
+    uniqueAutomationCondition: uniqueAutomationConditionsValidator,
```
netama, line 152: tiny nit: there's an `s` in one side and not in the other (`Condition` vs `Conditions`), let's align?

packages/zendesk-adapter/src/change_validators/unique_automation_conditions.ts
```diff
@@ -0,0 +1,90 @@
...
+      if (changedAutomation.elemID.isEqual(automation.elemID)) {
+        return
+      }
+      // If the conditions are deep equal, we return an error
+      if (areConditionsEqual(changedAutomation.value.conditions, automation.value.conditions)) {
+        errors.push({
+          elemID: changedAutomation.elemID,
+          severity: 'Error',
+          message: 'Automation\'s conditions is are unique',
+          detailedMessage: `Automation '${changedAutomation.elemID.getFullName()}' has the same conditions as '${automation.elemID.getFullName()}', make sure the conditions are unique before deploying.`,
```
netama, line 83: nits:
```suggestion
          message: 'Automation conditions are not unique',
          detailedMessage: `Automation has the same conditions as '${automation.elemID.getFullName()}', make sure the conditions are unique before deploying.`,
```

packages/zendesk-adapter/src/change_validators/unique_automation_conditions.ts
```diff
@@ -0,0 +1,90 @@
...
+export const uniqueAutomationConditionsValidator: ChangeValidator = async (changes, elementSource) => {
+  const changedAutomations = changes
+    .filter(isInstanceChange)
+    .filter(isAdditionOrModificationChange)
+    .map(getChangeData)
+    .filter(instance => instance.elemID.typeName === AUTOMATION_TYPE_NAME)
+
+  const elementSourceAutomations: InstanceElement[] = []
+  if (elementSource === undefined) {
+    log.error('duplicateAutomationConditionValidator might not run correctly because element source is undefined')
```
netama, line 54: we usually just `return` here (and have a standard error message), is there a reason to use a different pattern here? (we're always supposed to have an element source and many other validators will fail if we don't)

packages/zendesk-adapter/src/change_validators/unique_automation_conditions.ts
```diff
@@ -0,0 +1,90 @@
...
+  } else {
+    const idsIterator = awu(await elementSource.list())
+    const automations: InstanceElement[] = await awu(idsIterator)
```
netama, line 57: we shouldn't cast this way, should `.filter(isInstanceElement)` instead (you can see references in other validators)

packages/zendesk-adapter/src/change_validators/unique_automation_conditions.ts
```diff
@@ -0,0 +1,90 @@
...
+      .filter(id => id.typeName === AUTOMATION_TYPE_NAME)
+      .filter(id => id.idType === 'instance')
+      .map(id => elementSource.get(id))
+      .toArray()
+    elementSourceAutomations.push(...automations)
```
netama, line 62: once we remove the `else` we won't need `elementSourceAutomations`

packages/zendesk-adapter/src/change_validators/unique_automation_conditions.ts
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  const allAutomations = _.uniqBy(
+    [...changedAutomations, ...elementSourceAutomations],
+    automation => automation.elemID.getFullName(),
+  )
+
+  const errors: ChangeError[] = []
```
netama, line 70: is there a reason why we need this here vs using `.map()` or `.flatMap()` below?

packages/zendesk-adapter/src/change_validators/unique_automation_conditions.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  changedAutomations.forEach(changedAutomation => {
+    allAutomations.forEach(automation => {
+      if (changedAutomation.elemID.isEqual(automation.elemID)) {
+        return
+      }
+      // If the conditions are deep equal, we return an error
+      if (areConditionsEqual(changedAutomation.value.conditions, automation.value.conditions)) {
```
netama, line 78: 1. this can be inefficient when deploying a lot of automations, I'm wondering if we should try to avoid the `O(n^2)` here by serializing the conditions for comparison? 
2. did we analyze the cases here and make sure everything is covered? if the `all` and `any` fields in conditions are arrays, are "reordered" arrays considered equal or not?

netama (commented):
cool :) added some questions

packages/zendesk-adapter/src/change_validators/unique_automation_conditions.ts
idoser96: We also check the conditions between changes, so something will be done if we don't return here

packages/zendesk-adapter/src/change_validators/unique_automation_conditions.ts
idoser96: This is how it is written in jira, i guess because  id.idType === 'instance' makes already sure its an instanceElement
I can add the filter but it is used only for casting

packages/zendesk-adapter/src/change_validators/unique_automation_conditions.ts
idoser96: just the regular preference, more neat than two flatMaps with filter(isDefined)

packages/zendesk-adapter/src/change_validators/unique_automation_conditions.ts
idoser96: 1. i was thinking about that but there is the problem of reference expressions unresolved, maybe i can first convert them to ElemId strings and then serialize them
2. it is possible to deploy to equal arrays with different item order (in theory the conditions are the same but it does pass zendesk's api)

packages/zendesk-adapter/src/change_validators/unique_automation_conditions.ts
netama: I think it's ok to not do anything if the element source is missing

packages/zendesk-adapter/src/change_validators/unique_automation_conditions.ts
netama: if adding the filter allows us to avoid the cast let's do it?

packages/zendesk-adapter/src/change_validators/unique_automation_conditions.ts
netama: re 1 - ok, looks like there are up to 500 active automations (btw, does the limitation take into account active automations in some direction?) - so let's benchmark how long the validator takes on 1000 new/modified automations with long conditions and see if this needs optimizing?
re 2 - cool, thanks :)

packages/zendesk-adapter/src/change_validators/unique_automation_conditions.ts
netama: I think this one is a pretty straightforward one, so let's align to the convention?

packages/zendesk-adapter/src/change_validators/unique_automation_conditions.ts
```diff
@@ -0,0 +1,94 @@
...
+import { AUTOMATION_TYPE_NAME } from '../constants'
+
+const { isDefined } = lowerDashValues
+const { awu } = collections.asynciterable
+const log = logger(module)
+
+const stringifyAutomationConditions = (automation: InstanceElement): string =>
+  safeJsonStringify(
+    automation.value.conditions,
+    // Elements in elementSource have unresolved references, we want to make sure we handle that
```
netama: nit following IRL - please elaborate in the comment about what the issue is, and mention that this is why we're not using `elementExpressionStringifyReplacer`

packages/zendesk-adapter/src/change_validators/unique_automation_conditions.ts
```diff
@@ -0,0 +1,94 @@
...
+ */
+export const uniqueAutomationConditionsValidator: ChangeValidator = async (changes, elementSource) => {
+  const changedAutomations = changes
+    .filter(isInstanceChange)
+    .filter(isAdditionOrModificationChange)
+    .map(getChangeData)
+    .filter(instance => instance.elemID.typeName === AUTOMATION_TYPE_NAME)
+    .filter(instance => instance.value.active === true)
+
+  if (elementSource === undefined) {
```
netama, line 56: nit: can be done first

packages/zendesk-adapter/src/change_validators/unique_automation_conditions.ts
```diff
@@ -0,0 +1,94 @@
...
+    log.error('Failed to run duplicateAutomationConditionValidator because element source is undefined')
+    return []
+  }
+
+  const idsIterator = awu(await elementSource.list())
+  const elementSourceAutomations = await awu(idsIterator)
```
netama, line 62: 1. why do we need two `awu` wraps? (and do we need the `idsIterator` const?)
2. nit: these are now only the active automations, maybe clarify in the name?

packages/zendesk-adapter/src/change_validators/unique_automation_conditions.ts
```diff
@@ -0,0 +1,94 @@
...
+    .filter(id => id.typeName === AUTOMATION_TYPE_NAME)
+    .filter(id => id.idType === 'instance')
+    .map(id => elementSource.get(id))
+    .filter(isInstanceElement)
+    .filter(automation => automation.value.active === true)
+    .toArray()
+
+  // We map all automations to their conditions in advance, to avoid running on the whole list every time
+  const conditionsToAutomations = _.groupBy(
```
netama, line 71: technically if there's a reference vs a value that isn't a reference we will not catch it here, but I think it's ok

packages/zendesk-adapter/src/change_validators/unique_automation_conditions.ts
```diff
@@ -0,0 +1,94 @@
...
+      const equalAutomations = conditionsToAutomations[automationConditions]
+        .filter(otherAutomation => !otherAutomation.elemID.isEqual(automation.elemID))
+        .map(equalAutomation => equalAutomation.elemID.getFullName())
+        .join(', ')
+
+      return {
+        elemID: automation.elemID,
+        severity: 'Error',
+        message: 'Automation\'s conditions is are unique',
+        detailedMessage: `Automation '${automation.elemID.getFullName()}' has the same conditions as '${equalAutomations}', make sure the conditions are unique before deploying.`,
```
netama, line 89: still needs to be fixed - see https://github.com/salto-io/salto/pull/4609#discussion_r1262197006

packages/zendesk-adapter/test/change_validators/unique_automation_conditions.test.ts
```diff
@@ -0,0 +1,146 @@
...
+import { elementSource as elementSourceUtils } from '@salto-io/workspace'
+import _ from 'lodash'
+import { AUTOMATION_TYPE_NAME, ZENDESK } from '../../src/constants'
+import { uniqueAutomationConditionsValidator } from '../../src/change_validators'
+
+const { createInMemoryElementSource } = elementSourceUtils
+
+const conditions = {
+  all: [
+    {
```
netama, line 33: let's also add a more complex condition (with a different structure) if we have one?

packages/zendesk-adapter/test/change_validators/unique_automation_conditions.test.ts
```diff
@@ -0,0 +1,146 @@
...
+    new ObjectType({ elemID: new ElemID(ZENDESK, AUTOMATION_TYPE_NAME) }),
+    {
+      active: true,
+      conditions: _.cloneDeep(conds),
+    },
+  )
+
+
+describe('duplicateAutomationConditionValidator', () => {
+  const automationInstance = createAutomationInstance('automation', conditions)
```
netama, line 60: please initialize in `beforeEach` / `beforeAll`, not globally

netama (approved):
added some nits but looks good :) so pre-approving
please don't miss the text fixes, and add release notes

packages/zendesk-adapter/test/change_validators/unique_automation_conditions.test.ts
idoser96: did not see one (except for conditions with references, but they are handled in their test)

packages/zendesk-adapter/test/change_validators/unique_automation_conditions.test.ts
idoser96: but this is used in all tests and doesn't change, so why would we initialize it every time?

idoser96: moved it outside of the describe, is that ok?

packages/zendesk-adapter/src/change_validators/unique_automation_conditions.ts
idoser96: i think two if two values are equal there is no reason for one of them to become reference expression while the other wont

PR #4611, opened 2023-07-13
SALTO-3740 add NS aliases by default

dantal4:
Can be disabled using `fetch.addAlias=false`

---

---
_Release Notes_: 
Netsuite Adapter:
- Add aliases by default

---
_User Notifications_: 
None


omrilit (changes requested):
Looks good!
Marking with "Request Changes" just for safety reasons so it won't get merged mistakenly.

omrilit (approved):

PR #4613, opened 2023-07-13
SALTO-4485: Make "field" and "attr" valid element IDs under thier type

ori-moisis:
Before this change, only "adapter.typename.annotation" was a valid ID but "field" and "attr" were not Since we these IDs do represent a logic grouping of parts of the element, it is useful to have them as valid IDs as well

This increases the nesting level of fields and attributes under types

---

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


ori-moisis:
The effect of this change is much more severe than what we want to handle right now, so closing this PR

PR #4615, opened 2023-07-13
Salto-4262 Add info CV for more than 10k options in a custom field

idoamit8:
Following to this [PR](https://github.com/salto-io/salto/pull/4579) I added a CV to notify users when they have more than 10K options in a custom field. 

---

---
_Release Notes_: 
__Jira adapter__
The users will be notified when they are deploying more than 10K options to a custom field.

---
_User Notifications_: 
None


packages/jira-adapter/src/config/config.ts
```diff
@@ -257,6 +258,7 @@ const changeValidatorConfigType = createMatchingObjectType<ChangeValidatorConfig
     permissionSchemeDeployment: { refType: BuiltinTypes.BOOLEAN },
     projectCategory: { refType: BuiltinTypes.BOOLEAN },
     unresolvedFieldConfigurationItems: { refType: BuiltinTypes.BOOLEAN },
+    customFieldsWith10KOption: { refType: BuiltinTypes.BOOLEAN },
```
IdoZakk, line 261: It seems you are not using this flag....
Also a nit: change name to `customFieldsWith10KOptions`

packages/jira-adapter/src/change_validators/field_contexts/custom_field_with_10K_options.ts
```diff
@@ -0,0 +1,34 @@
...
+import { getParents } from '@salto-io/adapter-utils'
+import { FIELD_CONTEXT_TYPE_NAME, FIELD_TYPE_NAME } from '../../filters/fields/constants'
+import { getOptionsFromContext } from '../../filters/fields/context_options'
+
+export const customFieldsWith10KOptionValidator: ChangeValidator = async changes => changes
+  .filter(isInstanceChange)
+  .filter(isAdditionOrModificationChange)
+  .map(getChangeData)
+  .filter(instance => instance.elemID.typeName === FIELD_CONTEXT_TYPE_NAME)
+  .filter(instance => getOptionsFromContext(instance).length > 10000)
```
alonstern, line 27: This will happen only if we add a new option and we have more than 10000 options right?

packages/jira-adapter/src/change_validators/field_contexts/custom_field_with_10K_options.ts
```diff
@@ -0,0 +1,34 @@
...
+  .filter(instance => getParents(instance)[0].elemID.typeName === FIELD_TYPE_NAME)
```
IdoZakk, line 28: Can the parent be something that is not a field?

packages/jira-adapter/test/change_validators/field_contexts/custom_field_with_10K_options.test.ts
```diff
@@ -0,0 +1,56 @@
...
+describe('customFieldsWith10KOptionValidator', () => {
+  const parentField = new InstanceElement('parentField', new ObjectType({ elemID: new ElemID(JIRA, 'Field') }), { id: 2 })
+  const contextInstance = new InstanceElement('context', new ObjectType({ elemID: new ElemID(JIRA, 'CustomFieldContext') }), {
+    id: 3,
+    options: [],
+  },
+  undefined,
+  {
+    [CORE_ANNOTATIONS.PARENT]: [new ReferenceExpression(parentField.elemID, parentField)],
+  })
```
alonstern, line 41: You should initialize this in a `beforeEach` section, otherwise one test might affect others

packages/jira-adapter/test/change_validators/field_contexts/custom_field_with_10K_options.test.ts
```diff
@@ -0,0 +1,56 @@
...
+  it('should return info message when context has more than 10K options', async () => {
```
alonstern, line 42: add also a test that it does not appear when we have less than 10k

packages/jira-adapter/src/change_validators/field_contexts/custom_field_with_10K_options.ts
```diff
@@ -0,0 +1,34 @@
...
+  .map(instance => ({
+    elemID: instance.elemID,
+    severity: 'Info',
+    message: 'Slow deployment due to field with more than 10K options',
+    detailedMessage: `The deployment of custom field ${getParents(instance)[0].elemID.name} will be slower because it is associated with the context ${instance.elemID.name} which has more than 10K options.`,
+  }))
```
IdoZakk, line 34: Did you verify this text with Lior?
Especially the message

packages/jira-adapter/test/change_validators/field_contexts/custom_field_with_10K_options.test.ts
```diff
@@ -0,0 +1,56 @@
...
+  Array.from({ length: count }, (_, i) => ({
+    [`p${i}`]: {
+      value: `p${i}`,
+      disabled: false,
+      position: i,
+    },
+  })).reduce((acc, option) => ({ ...acc, ...option }), {})
+
+// Set long timeout as we create instance with more than 10K options
+jest.setTimeout(1000 * 60 * 5)
```
alonstern, line 31: Is this test really take that much time? I wouldn't expect initializing 10,000 to take more than a second

packages/jira-adapter/test/change_validators/field_contexts/custom_field_with_10K_options.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    const changes = [toChange({ after: contextInstanceAfter })]
+    const changeErrors = await customFieldsWith10KOptionValidator(changes)
+    expect(changeErrors).toHaveLength(1)
+    expect(changeErrors).toEqual([{
+      elemID: contextInstance.elemID,
+      severity: 'Info',
+      message: 'Slow deployment due to field with more than 10K options',
+      detailedMessage: `The deployment of custom field ${parentField.elemID.name} will be slower because it is associated with the context ${contextInstance.elemID.name} which has more than 10K options.`,
+    }])
+  })
```
IdoZakk, line 55: I would add at least to large contexts and one regular to make sure the function does not issue errors on regular contexts, and to make sure ti handles multiple errors well

IdoZakk (commented):
👌 
LGTM, see especially the comment about the config param

packages/jira-adapter/test/change_validators/field_contexts/custom_field_with_10K_options.test.ts
```diff
@@ -0,0 +1,56 @@
...
+import { customFieldsWith10KOptionValidator } from '../../../src/change_validators/field_contexts/custom_field_with_10K_options'
+
+const generateOptions = (count: number): Values =>
+  Array.from({ length: count }, (_, i) => ({
+    [`p${i}`]: {
+      value: `p${i}`,
+      disabled: false,
+      position: i,
+    },
+  })).reduce((acc, option) => ({ ...acc, ...option }), {})
```
alonstern, line 28:
```suggestion
  Object.fromEntries(_.range(count).map(i => [[
    `p${i}`,
     {
      value: `p${i}`,
      disabled: false,
      position: i,
    },
   ]]))
```
I think this is faster since you don't copy the object on each option

alonstern (commented):
🔥

packages/jira-adapter/src/config/config.ts
idoamit8: I changed it. But I can't remove i, because build is failing. I think it has something to do with Sarrusi making the CV not mandatory.

packages/jira-adapter/src/change_validators/field_contexts/custom_field_with_10K_options.ts
idoamit8: Yes. Changed it a bit after conversation with Lior

packages/jira-adapter/src/change_validators/field_contexts/custom_field_with_10K_options.ts
idoamit8: I don't think so, but I took the safe side.

packages/jira-adapter/src/change_validators/field_contexts/custom_field_with_10K_options.ts
idoamit8: yes

packages/jira-adapter/test/change_validators/field_contexts/custom_field_with_10K_options.test.ts
idoamit8: I removed it.

packages/jira-adapter/src/change_validators/field_contexts/custom_field_with_10K_options.ts
alonstern: It looks like it will happen on any change on a field with more than 10,000 options (e.g., a description change). I think we need to check here if there are new options in the field

packages/jira-adapter/src/config/config.ts
IdoZakk: Let's try together, no reason to add it like this

IdoZakk:
There is a typo in the release notes, should be `The users will be notified`
Also this change does not influence the NaCls, so no need for User Notifications

packages/jira-adapter/src/config/config.ts
idoamit8: Can you please give me an example for a validator that is not there? I will copy the behavior. 
except of this issue, I fixed all other suggestions.

packages/jira-adapter/src/change_validators/field_contexts/custom_field_with_10K_options.ts
```diff
@@ -14,14 +14,25 @@
...
+  const { before, after } = change.data
+  const optionsBefore = getOptionsFromContext(before)
+  const optionsAfter = getOptionsFromContext(after)
+  return !isEqualValues(optionsBefore, optionsAfter)
+}
+
 export const customFieldsWith10KOptionValidator: ChangeValidator = async changes => changes
   .filter(isInstanceChange)
   .filter(isAdditionOrModificationChange)
+  .filter(hasNewOption)
```
IdoZakk, line 35: You are going to test it for every instance that has `options`, seems expensive.
You can either push the two other filters before this one (and call getChangedData inside each filter), or push this down, remove the check of isAddition and verify that before is not undefined

packages/jira-adapter/test/change_validators/field_contexts/custom_field_with_10K_options.test.ts
```diff
@@ -96,4 +96,31 @@ describe('customFieldsWith10KOptionValidator', () => {
       detailedMessage: `The deployment of custom field ${parentField.elemID.name} will be slower because it is associated with this context, which has more than 10K options.`,
     }])
   })
+  it('should not return error if context has no new options', async () => {
+    const contextInstanceBefore = contextInstance.clone()
```
IdoZakk, line 100: No need to clone it, you can use it as you recreate it on every test

packages/jira-adapter/test/change_validators/field_contexts/custom_field_with_10K_options.test.ts
```diff
@@ -96,4 +96,31 @@ describe('customFieldsWith10KOptionValidator', () => {
...
+    contextInstanceBefore.value.options = tenKOptions
+    const contextInstanceAfter = contextInstanceBefore.clone()
+    const changes = [toChange({ before: contextInstanceBefore, after: contextInstanceAfter })]
```
IdoZakk, line 103: To be on the safe side (closer to real life situation) it is better to add some change to the `after` instance

packages/jira-adapter/test/change_validators/field_contexts/custom_field_with_10K_options.test.ts
```diff
@@ -96,4 +96,31 @@ describe('customFieldsWith10KOptionValidator', () => {
...
+    const changeErrors = await customFieldsWith10KOptionValidator(changes)
+    expect(changeErrors).toHaveLength(0)
+  })
+  it('should return error for modification change', async () => {
+    const contextInstanceBefore = contextInstance.clone()
```
IdoZakk, line 108: Same here, no need for two clones

packages/jira-adapter/test/change_validators/field_contexts/custom_field_with_10K_options.test.ts
```diff
@@ -96,4 +96,31 @@ describe('customFieldsWith10KOptionValidator', () => {
...
+    const changes = [toChange({ before: contextInstanceBefore, after: contextInstanceAfter })]
+    const changeErrors = await customFieldsWith10KOptionValidator(changes)
+    expect(changeErrors).toHaveLength(1)
+    expect(changeErrors).toEqual([{
+      elemID: contextInstance.elemID,
+      severity: 'Info',
+      message: 'Slow deployment due to field with more than 10K options',
+      detailedMessage: `The deployment of custom field ${parentField.elemID.name} will be slower because it is associated with this context, which has more than 10K options.`,
+    }])
+  })
```
IdoZakk, line 125: Add another test for an addition change, and verify an error is returned

IdoZakk (commented):
Looks good, some minor comments

packages/jira-adapter/src/change_validators/field_contexts/custom_field_with_10K_options.ts
idoamit8: There is no before for additionChange so I cant make the placement if I remove the check for `isAdditionChange`. 
Changed the order of the filters.

packages/jira-adapter/test/change_validators/field_contexts/custom_field_with_10K_options.test.ts
idoamit8: Done in the first test

packages/jira-adapter/src/change_validators/field_contexts/custom_field_with_10K_options.ts
IdoZakk: With certain checks or type guards you can overcome it, for next time :-)

IdoZakk (approved):
💣

PR #4617, opened 2023-07-13
Salesforce: getAdditionalReferences doesn't return references for Profile.layoutAssignments

tomsellek:
Use fullName to map from the contents of the fields in the various Profile sections to the potential ref targets.
---

Before this change the code assumed it can create the field names from the apiName of the potential reference targets. This is not the case, and only happened to work most times for most types.

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/additional_references.ts
+++ b/packages/salesforce-adapter/src/additional_references.ts
@@ -124,10 +124,21 @@ const fieldRefsFromProfileOrPermissionSet = async (
     ))
 }
 
+type RefTargetName = {
+  typeName: string
+  refName: string
+}
+
+type RefNameGetter = (element: Value, key: string) => RefTargetName[]
+
+const refNameFromField = (typeName: string, fieldName: string): RefNameGetter => (
+  element => ([{ typeName, refName: element[fieldName] }])
+)
+
 const instanceRefsFromProfileOrPermissionSet = async (
   profilesAndPermissionSets: ModificationChange<InstanceElement>[],
   profileSection: string,
-  instanceNameGetter: (element: Value, key: string) => { typeName: string; refName: string },
+  instanceNameGetter: RefNameGetter,
   instanceIndex: multiIndex.Index<[string, string], AdditionOrModificationChange>,
 ): Promise<ReferenceMapping[]> => {
   const extractRefFromSectionEntry = async (
@@ -135,17 +146,16 @@ const instanceRefsFromProfileOrPermissionSet = async (
     sectionEntryKey: string,
     sectionEntryContents: Value,
   ): Promise<ReferenceMapping[]> => {
-    const { typeName, refName } = instanceNameGetter(sectionEntryContents, sectionEntryKey)
-    const refTarget = instanceIndex.get(typeName, refName)
-    if (refTarget === undefined) {
-      return []
-    }
-    return createReferenceMapping(
-      profileOrPermissionSetChange,
-      refTarget,
-      sectionEntryKey,
-      profileSection,
-    )
+    const refNames = instanceNameGetter(sectionEntryContents, sectionEntryKey)
+    return refNames
+      .map(({ typeName, refName }) => instanceIndex.get(typeName, refName))
+      .filter(isDefined)
+      .flatMap(refTarget => createReferenceMapping(
+        profileOrPermissionSetChange,
+        refTarget,
+        sectionEntryKey,
+        profileSection,
+      ))
   }
 
   return awu(profilesAndPermissionSets)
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/src/additional_references.ts
```diff
@@ -86,147 +103,225 @@ const createReferenceMapping = (
...
+  )
 
   // note that permission sets don't contain layout assignments, but it simplifies our code to pretend like they might
   // ref: https://ideas.salesforce.com/s/idea/a0B8W00000GdlSPUAZ/permission-sets-with-page-layout-assignment
-  const layoutRefs = awu(addedOrModifiedInstancesChangesByType[LAYOUT_TYPE_ID_METADATA_TYPE] ?? [])
-    .flatMap(async layout => refsFromProfileOrPermissionSet(profilesAndPermSetsChanges, layout, LAYOUTS_SECTION))
+  const layoutRefs = await instanceRefsFromProfileOrPermissionSet(
+    profilesAndPermSetsChanges,
+    LAYOUTS_SECTION,
+    layoutAssignment => ({ typeName: LAYOUT_TYPE_ID_METADATA_TYPE, refName: layoutAssignment[0].layout }),
```
amitgoldarb, line 262: Do we need to add recordTypes to here somehow? They exist in layoutAssignments as well
Also I think this might be safer?
```
layoutAssignment[0]?.layout
```

packages/salesforce-adapter/src/additional_references.ts
```diff
@@ -86,147 +103,225 @@ const createReferenceMapping = (
...
    * Note: if the adapter config `generateRefsInProfiles` is set then these fields will already contain the correct
    * references, in which case we will create duplicate references and drop them. We won't crash because we don't
    * actually look at the content of any field/annotation, only on the keys in the provided profile section.
+   *
+   * Note: We use the section keys (as opposed to a field that contains the apiName of the referred entity) because
+   * there is no such field in the fieldPermissions section
    */
   const apiName = await safeApiName(getChangeData(potentialTarget))
   if (apiName === undefined) {
+    log.warn('no apiName for %o', potentialTarget)
```
amitgoldarb, line 120: Might be spammy, maybe lower to debug?

packages/salesforce-adapter/src/additional_references.ts
```diff
@@ -86,147 +103,225 @@ const createReferenceMapping = (
...
-    return []
-  }
-  return layoutAssignments
-    .filter(layoutAssignment => _.isString(layoutAssignment.recordType))
-    .filter(layoutAssignment => layoutAssignment.recordType in recordTypesByApiName)
-    .flatMap(layoutAssignment => createReferenceMapping(
+const instanceRefsFromProfileOrPermissionSet = async (
+  profilesAndPermissionSets: ModificationChange<InstanceElement>[],
+  profileSection: string,
+  instanceNameGetter: (element: Value, key: string) => { typeName: string; refName: string },
```
amitgoldarb, line 136: I think if this function could return multiple values we could possibly dry up the recordType from layouts no?

packages/salesforce-adapter/src/additional_references.ts
```diff
@@ -86,147 +103,225 @@ const createReferenceMapping = (
...
-        .filter(profileOrPermissionSet => _.get(getChangeData(profileOrPermissionSet).value[LAYOUTS_SECTION], apiName))
-        .flatMap(profileOrPermissionSet => (
-          Object.values(getChangeData(profileOrPermissionSet).value[LAYOUTS_SECTION])
-            .flatMap(layoutAssignments => recordTypeRefsFromLayoutAssignments(
-              layoutAssignments,
-              recordTypesByApiName,
-              profileOrPermissionSet,
-              apiName
-            ))
+  const recordTypeRefs = profilesAndPermSetsChanges
```
amitgoldarb, line 283: Why can't we Dry this with a simple instanceRefsFromProfileOrPermissionSet like the other example. Doesn't the instanceNameGetter need to be slightly more complicated and everything else be the same?

amitgoldarb (commented):
Added a few comments.
Overall looks good.
@alonstern Can you take a look as well?

packages/salesforce-adapter/src/additional_references.ts
tomsellek: It's just a debugging leftover, I'll remove it. Why would we not have an apiName though?

packages/salesforce-adapter/src/additional_references.ts
```diff
@@ -86,147 +103,225 @@ const createReferenceMapping = (
...
     .filter(isFieldChange)
     .filter(isAdditionOrModificationChange)
     .filter(change => isFieldOfCustomObject(getChangeData(change)))
     .toArray()
 
-  const customObjectChanges = changes
+  const customObjectChanges: AwuIterable<AdditionOrModificationChange> = awu(changes)
     .filter(isAdditionOrModificationChange)
-    .filter(change => isCustomObject(getChangeData(change)))
+    .filter(isObjectTypeChange)
```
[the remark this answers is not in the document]
tomsellek: No reason. Removed.

packages/salesforce-adapter/src/additional_references.ts
```diff
@@ -86,147 +103,225 @@ const createReferenceMapping = (
...
+    .filter(async change => isCustomObject(getChangeData(change)))
 
-  const addedOrModifiedInstancesChanges = changes
+  const instanceChanges: AwuIterable<AdditionOrModificationChange> = awu(changes)
```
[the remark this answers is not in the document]
tomsellek: Nope!

packages/salesforce-adapter/src/additional_references.ts
tomsellek: You'd have to change `instanceNameGetter` to return an array, and if I'm going to do that, I may as well follow your suggestion above and extract the `recordType` refs at the same time I extract the layout refs.

packages/salesforce-adapter/src/additional_references.ts
tomsellek: Indeed.

packages/salesforce-adapter/src/additional_references.ts
tomsellek: I don't think that's possible because we create this map from a list of layoutAssignments (so I don't think we can ever have a key that points to an empty array), but it can't hurt to be too careful, I suppose.

packages/salesforce-adapter/src/additional_references.ts
```diff
@@ -86,144 +98,216 @@ const createReferenceMapping = (
   })))
 }
 
-const refsFromProfileOrPermissionSet = async (
+const fieldRefsFromProfileOrPermissionSet = async (
```
amitgoldarb, line 101: I think we might be able to dry this one with the instanceRefsFromProfileOrPermissionSet as well no?

packages/salesforce-adapter/src/additional_references.ts
amitgoldarb: We made it return an array, wouldn't it be simpler to use profilesAndPermSetsChanges with a "specific" RefNameGetter?

amitgoldarb (commented):
This PR looks great!
I think we can probably DRY all profile sections to use profilesAndPermSetsChanges don't you think? It seems like we have 3 different ways to create these profile references (profilesAndPermSetsChanges, FLS method and the recordVisibility). I might be missing something, but I'm not sure we need all of them, WDYT?

packages/salesforce-adapter/src/additional_references.ts
tomsellek: Not easily, because the field name doesn't actually appear in the field's section.

packages/salesforce-adapter/src/additional_references.ts
tomsellek: I don't think so, because recordTypeVisibilities is nested 2-deep, so the reference source behaves differently from all other cases:

```
  recordTypeVisibilities = {
    Case = {
      SomeCaseRecordType = {
        default = true
        recordType = salesforce.RecordType.instance.Case_SomeCaseRecordType
        visible = true
      }
      SomeOtherCaseRecordType = {
        default = false
        recordType = salesforce.RecordType.instance.Case_SomeOtherCaseRecordType
        visible = true
      }
    }
...
}
```

amitgoldarb: Doesn't something along the lines of the following work?
```suggestion
  (recordTypeVisibiliy, key) => Object.entries(recordTypeVisibilities).map(([someType, { recordType }]) => ({ typeName: key, refName: recordType })
```

amitgoldarb (approved):
Discussed f2f. Thanks for this!
Great work

PR #4622, opened 2023-07-16
SALTO-4295 - Zendesk Adapter - Support missing reference for dynamic_content

idoser96:
Support missing references for dynamic content
Also support array in array structure in 'actions' field of instances

---

trigger type action's are array inside array, and we covered only 1 nesting level with our code, so i made it recursive to cover all possible nesting levels (which is currently 2)

---
_Release Notes_: 
Zendesk adapter:
* References inside array inside array in 'actions' field of instances will be converted to reference expressions
* Dynamic content references will now be converted to reference expressions

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-utils/src/template.ts
+++ b/packages/adapter-utils/src/template.ts
@@ -65,14 +65,15 @@ export const replaceTemplatesWithValues = (
     deployTemplateMapping[templateUsingIdField.value] = template
     return templateUsingIdField.value
   }
-  const replaceIfTemplate = (value: unknown): unknown =>
-    (isTemplateExpression(value) ? handleTemplateValue(value) : value)
-  container.values.forEach(value => {
-    if (Array.isArray(value[fieldName])) {
-      value[fieldName] = value[fieldName].map(replaceIfTemplate)
-    } else {
-      value[fieldName] = replaceIfTemplate(value[fieldName])
+  const replaceIfTemplate = (value: unknown): unknown => {
+    if (Array.isArray(value)) {
+      return value.map(replaceIfTemplate)
     }
+    return (isTemplateExpression(value) ? handleTemplateValue(value) : value)
+  }
+
+  container.values.forEach(value => {
+    value[fieldName] = replaceIfTemplate(value[fieldName])
   })
 }
 
@@ -83,10 +84,17 @@ export const resolveTemplates = (
   const resolveTemplate = (value: string): TemplateExpression | string =>
     deployTemplateMapping[value] ?? value
 
+  const resolveTemplateValue = (value: unknown): unknown => {
+    if (Array.isArray(value)) {
+      return value.map(resolveTemplateValue)
+    }
+    return _.isString(value) ? resolveTemplate(value) : value
+  }
+
   const { fieldName } = container
   container.values.forEach(value => {
     const val = value[fieldName]
-    value[fieldName] = Array.isArray(val) ? val.map(resolveTemplate) : resolveTemplate(val)
+    value[fieldName] = Array.isArray(val) ? val.map(resolveTemplateValue) : resolveTemplate(val)
   })
 }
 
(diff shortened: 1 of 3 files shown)


packages/adapter-utils/src/template.ts
```diff
@@ -65,14 +65,15 @@ export const replaceTemplatesWithValues = (
...
     return templateUsingIdField.value
   }
-  const replaceIfTemplate = (value: unknown): unknown =>
-    (isTemplateExpression(value) ? handleTemplateValue(value) : value)
-  container.values.forEach(value => {
-    if (Array.isArray(value[fieldName])) {
-      value[fieldName] = value[fieldName].map(replaceIfTemplate)
-    } else {
-      value[fieldName] = replaceIfTemplate(value[fieldName])
+  const replaceIfTemplate = (value: unknown): unknown => {
```
edenhassid, line 68: did you check that this does not affect other adapters, incase we need a noise reduction for them too

edenhassid: can you add tests for this change?

packages/adapter-utils/src/template.ts
```diff
@@ -83,10 +84,17 @@ export const resolveTemplates = (
...
+      return value.map(resolveTemplateValue)
+    }
+    return _.isString(value) ? resolveTemplate(value) : value
+  }
+
   const { fieldName } = container
   container.values.forEach(value => {
     const val = value[fieldName]
-    value[fieldName] = Array.isArray(val) ? val.map(resolveTemplate) : resolveTemplate(val)
+    value[fieldName] = Array.isArray(val) ? val.map(resolveTemplateValue) : resolveTemplate(val)
```
edenhassid, line 97: can you add tests for this change?

packages/zendesk-adapter/src/filters/handle_template_expressions.ts
```diff
@@ -229,6 +229,15 @@ const formulaToTemplate = (
     if (elem) {
       return new ReferenceExpression(elem.elemID, elem)
     }
+    if (!_.isEmpty(dcPlaceholder) && enableMissingReferences) {
+      const missingInstance = createMissingInstance(
+        ZENDESK,
+        DYNAMIC_CONTENT_ITEM_TYPE_NAME,
+        dcPlaceholder.slice(3) // slice dc. from the placeholder
```
edenhassid, line 236: maybe check before that it starts with dc. , in case it changes in the future and we will miss it

packages/zendesk-adapter/src/filters/handle_template_expressions.ts
```diff
@@ -265,20 +274,20 @@ const getContainers = async (instances: InstanceElement[]): Promise<TemplateCont
...
       const { fieldName } = container
-      const instancesByType = _.groupBy(instances, i => i.elemID.typeName)
       container.values.forEach(value => {
-        if (Array.isArray(value[fieldName])) {
-          value[fieldName] = value[fieldName].map((innerValue: unknown) =>
-            (_.isString(innerValue)
-              ? formulaToTemplate(innerValue, instancesByType, enableMissingReferences)
-              : innerValue))
-        } else if (value[fieldName]) {
-          value[fieldName] = formulaToTemplate(value[fieldName], instancesByType,
```
edenhassid, line 279: discussed IRL

edenhassid (commented):
looks good 😄  added a few comments

packages/adapter-utils/src/template.ts
idoser96: already covered, talked IRL

packages/adapter-utils/src/template.ts
idoser96: already covered, talked IRL

packages/adapter-utils/src/template.ts
idoser96: This also affects Okta and Jira
Okta doesn't have any customers with monitoring so its ok
Jira instances doesn't seem to have a structure that will be a affected by this change (checked with @IdoZakk )

edenhassid (approved):
Fantastic job! ⭐

PR #4623, opened 2023-07-16
SALTO-4465: retrieve of unmanaged instances from namespace

tamtamirr:
This is an alternate solution to https://github.com/salto-io/salto/pull/4594 which had a regression for some Layout instances from certain Installed Packages.

---

This PR takes into consideration the namespace of the current org and won't apply the **getFullName** logic on instances
that come from this namespace.

- Workspace Diff (Only additions 🥳): https://github.com/salto-io/tamir-sf/commit/962bb44ecde8fb691190a998eb3229ca2f5186ab

---
_Release Notes_: 
Salesforce Adapter:
- Support retrieve of unmanaged instances for orgs with namespace defined.

---
_User Notifications_: 
Salesforce:
- Existing Workspaces of Salesforce Orgs that have a namespace defined, may encounter addition of instances that were not retrieved before.


packages/salesforce-adapter/src/client/client.ts
```diff
@@ -542,6 +543,8 @@ export default class SalesforceClient {
   @requiresLogin()
   public async listMetadataTypes(): Promise<MetadataObject[]> {
     const describeResult = await this.retryOnBadResponse(() => this.conn.metadata.describe())
+    this.orgNamespace = describeResult.organizationNamespace
+    log.debug('describeResult: %o', _.omit(describeResult, 'metadataObjects'))
```
tamtamirr, line 547: This log is for visibility about the describe result of the org. It contains some useful information, like if **rollbackOnError** is allowed for this org, and if tests should run here upon deploy. (We could probably create tickets on top of this info later, but for now this is good for our visibility as we don't parse all of the **describeResult** in our code)

amitgoldarb: This feels like it's unnecessarily risky. If this is a large object (if the types are lying somehow) than we will log something huge. I feel it's better to be explicit than implicit in this case. Can we possibly pick and choose what we want to log?

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -357,9 +365,12 @@ export const retrieveMetadataInstances = async ({
       ? await listMetadataObjectsWithinFolders(client, metadataQuery, typeName, folderType)
       : await listMetadataObjects(client, typeName)
     configChanges.push(...listObjectsConfigChanges)
+    if (client.orgNamespace === undefined) {
+      log.warn('Expected orgNamespace to be defined when listing metadata instances')
```
amitgoldarb, line 369: We will have this log many times per fetch right? I think this log might be a bit noisy.

amitgoldarb (commented):
Added a few small comments. Other than that I think this is safe. Nice catch :)

packages/salesforce-adapter/src/client/client.ts
tamtamirr: Why would the type lie? **jsforce** explicitly populate these values from the response. Note that this is not raw response from the **Metadata API**. Unless we bump the version of our **js-force** this type should never change.

amitgoldarb: It's a very cryptic place to remember if the type changes due to any reason (js-force bump/type mis-alignment).
Can't we only log what we think is important?

tamtamirr: I've changed the log to only print out the namespace

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -215,8 +215,16 @@ const getFullName = (obj: FileProperties, addNamespacePrefixToFullName: boolean)
...
-const getPropsWithFullName = (obj: FileProperties, addNamespacePrefixToFullName: boolean): FileProperties => {
-  const correctFullName = getFullName(obj, addNamespacePrefixToFullName)
+const getPropsWithFullName = (
+  obj: FileProperties,
+  addNamespacePrefixToFullName: boolean,
+  orgNamespace?: string
+): FileProperties => {
+  // Do not run getFullName logic if the namespace of the instance is the current org namespace.
+  // If we couldn't determine the namespace of the org, we will run the logic for all the instances.
+  const correctFullName = orgNamespace === undefined || obj.namespacePrefix !== orgNamespace
```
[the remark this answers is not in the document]
tamtamirr: Salesforce forbids installing packages with the same namespace of the org (Meaning packages that were published from this org, as **namespace** must be unique across orgs. Here's an example of an attempt:
Screenshot 2023-07-17 at 11 44 53

benisraeladi (approved):
Looks good!

amitgoldarb (approved):
LGTM

PR #4624, opened 2023-07-17
correct spelling

ShloYo:
Fix the spelling of "recieved" to "received" - it was killing me.  Besides, this is a simple change that does not need the time/effort of a developer so I grabbed it


omrilit (approved):
Haha thanks @ShloYo!

PR #4625, opened 2023-07-17
SALTO-4280 - Zendesk Adapter - allow omitInactive in typeDefaults

idoser96:
Allow to put omitInactive in typeDefault to omit inactive instances from all types, being able to override that default in the type specific config

---

None

---
_Release Notes_: 
Zendesk Adapter:
* Added the option to default omit all inactive instances using typeDefaults config

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/test/filters/omit_inactive.test.ts
+++ b/packages/zendesk-adapter/test/filters/omit_inactive.test.ts
@@ -22,8 +22,17 @@ import filterCreator from '../../src/filters/omit_inactive'
 import { FilterResult } from '../../src/filter'
 import { createFilterCreatorParams } from '../utils'
 
+type FilterType = filterUtils.FilterWith<'onFetch', FilterResult>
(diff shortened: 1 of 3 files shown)


packages/zendesk-adapter/test/filters/omit_inactive.test.ts
```diff
@@ -84,6 +85,20 @@ describe('omit inactive', () => {
           webhook3.elemID.getFullName(),
         ])
     })
+    it('should omit inactive instances if omitInactive in typeDefaults is true', async () => {
```
netama, line 88: let's also add the "negative" tests that ensure that the behavior doesn't change with the new default - that ticket_form instances / instances without the active field are not omitted

netama (approved):
nice, LGTM (added one comment about additional tests).
let's clarify in the release notes what the caveats are (the `ticket_form` instance and the additional checks)

PR #4626, opened 2023-07-17
SALTO-4495: Add Salesforce Deploy logs

tamtamirr:
Added logs to Salesforce Adapter Deploy flow.

---

_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/adapter.ts
```diff
@@ -451,13 +457,15 @@ export default class SalesforceAdapter implements AdapterOperations {
     { changeGroup }: DeployOptions,
     checkOnly: boolean
   ): Promise<DeployResult> {
+    log.debug(`about to ${checkOnly ? 'validate' : 'deploy'} group ${changeGroup.groupID} with scope ${safeJsonStringify(changeGroup.changes.map(getChangeData).map(e => e.elemID.getFullName()))}`)
```
amitgoldarb, line 460: This can be very long no? It could be thousands of elements and also this is saved in the DB. Will this have debug value?

amitgoldarb (commented):
Looks good!
Added a small concern

packages/salesforce-adapter/src/adapter.ts
tamtamirr: Do we save in the DB the scope of each deploy group? Isn't it the scope of all of the deployment? Even if we do, it'd be easier to see that in Datadog without the need to go and search for it in the DB. I personally don't think there's any concern here, and it's worth logging it.

amitgoldarb (approved):
Approved with a single comment. 
Approving to save cycles

PR #4629, opened 2023-07-17
SALTO-4445 - Zendesk Adapter - Handle article attachments requests parallel

idoser96:
Changed how article attachments are handles so they can run parallel, and their rate limit is defined by the client

---

No release notes because this doesn't actually changes any behaviour

---
_Release Notes_: 
None 

---
_User Notifications_: 
None


netama:
I don't think this should be merged separately from https://github.com/salto-io/salto/pull/4631 , will review there directly (this solution is risky because it will make it impossible to reduce the rate limit to 1 even with the config)

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -201,8 +201,8 @@ const handleArticleAttachmentsPreDeploy = async ({ changes, client, elementsSour
   const attachmentChanges = changes
     .filter(isAdditionOrModificationChange)
     .filter(change => getChangeData(change).elemID.typeName === ARTICLE_ATTACHMENT_TYPE_NAME)
-  await awu(attachmentChanges)
-    .forEach(async attachmentChange => {
+  const attachmentsPromises = attachmentChanges
```
netama, line 204: 1. why not just 
```suggestion
  await Promise.all(...)
```
here?

2. how did we test / benchmark this, and how much of the flow was covered? I'm not sure we should parallelize everything since the `associateAttachments` call might have some race conditions we're not aware of. we can probably parallelize the `createUnassociatedAttachment` part but that requires some refactoring

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -201,47 +203,69 @@ const handleArticleAttachmentsPreDeploy = async ({ changes, client, elementsSour
...
-        if (!res) {
-          log.error(`Association of attachment ${instanceBeforeResolve.elemID.name} with id ${attachmentInstance.value.id} has failed `)
+        const parentArticleRef = getAttachmentArticleRef(instanceBeforeResolve)
+        if (parentArticleRef === undefined) {
+          log.error(`Couldn't find attachment ${instanceBeforeResolve.elemID.name} article parent instance.`)
           await deleteArticleAttachment(client, attachmentInstance)
           return
         }
-        await deleteArticleAttachment(client, attachmentChange.data.before)
-        return
```
idoser96, line 238: Only this was removed

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -201,47 +203,69 @@ const handleArticleAttachmentsPreDeploy = async ({ changes, client, elementsSour
...
+          const articleInstance = await parentArticleRef.getResolvedValue(elementsSource)
+          if (articleInstance === undefined || !isInstanceElement(articleInstance)) {
+            log.error(`Couldn't get article ${parentArticleRef} in the elementsSource`)
+            await deleteArticleAttachment(client, attachmentInstance)
+            return
+          }
+          const articleName = articleInstance.elemID.getFullName()
+          // We want to associate all attachments in bulks, so we fill it up here handle it later
+          articleToModifiedAttachments[articleName] = articleToModifiedAttachments[articleName] ?? [articleInstance, []]
+          articleToModifiedAttachments[articleName][1].push(attachmentChange)
```
idoser96, line 239: Only this was added

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -201,47 +203,69 @@ const handleArticleAttachmentsPreDeploy = async ({ changes, client, elementsSour
...
+        ({ ids }) => ids.forEach(id => deleteArticleAttachment(client, attachmentChangesById[id].data.before))
+      ),
+      ...failures.map(({ ids }) => ids.forEach(id => {
+        const afterAttachment = attachmentChangesById[id].data.after
+        log.error(`Association of attachment ${afterAttachment.elemID.name} with id ${afterAttachment.value.id} has failed `)
+        return deleteArticleAttachment(client, afterAttachment)
+      })),
+    ])
+  })
+
```
idoser96, line 268: And this was added

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -201,47 +220,69 @@ const handleArticleAttachmentsPreDeploy = async ({ changes, client, elementsSour
...
-        if (!res) {
-          log.error(`Association of attachment ${instanceBeforeResolve.elemID.name} with id ${attachmentInstance.value.id} has failed `)
+        const parentArticleRef = getAttachmentArticleRef(instanceBeforeResolve)
+        if (parentArticleRef === undefined) {
+          log.error(`Couldn't find attachment ${instanceBeforeResolve.elemID.name} article parent instance.`)
           await deleteArticleAttachment(client, attachmentInstance)
           return
         }
-        await deleteArticleAttachment(client, attachmentChange.data.before)
-        return
```
idoser96, line 238: Only this was removed here

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -170,26 +172,43 @@ const getAttachmentArticleRef = (
...
+      const res = await client.post({
+        url: `/api/v2/help_center/articles/${articleId}/bulk_attachments`,
+        data: { attachment_ids: chunk },
+      })
+      if (res.status !== SUCCESS_STATUS_CODE) {
+        log.warn(createErrorMsg(res.data, res.status))
+      }
+      return { status: res.status, ids: chunk }
+    } catch (e) {
+      const error = e.reponse?.data ?? e
```
netama, line 204: what non-http error are we expecting here?

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -201,47 +220,69 @@ const handleArticleAttachmentsPreDeploy = async ({ changes, client, elementsSour
...
-      const parentArticleName = parentArticleRef.elemID.name
-      articleNameToAttachments[parentArticleName] = (
-        articleNameToAttachments[parentArticleName] || []
-      ).concat(attachmentInstance.value.id)
-    })
+        // We can't really modify article attachments in Zendesk
+        // To do so we're going to delete the existing attachment and create a new one instead
+        if (isModificationChange(attachmentChange)) {
+          const articleInstance = await parentArticleRef.getResolvedValue(elementsSource)
+          if (articleInstance === undefined || !isInstanceElement(articleInstance)) {
```
netama, line 248: what case is the 2nd condition covering?

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -201,47 +220,69 @@ const handleArticleAttachmentsPreDeploy = async ({ changes, client, elementsSour
...
-        return
-      }
-      // We can't really modify article attachments in Zendesk
-      // To do so we're going to delete the existing attachment and create a new one instead
-      if (isModificationChange(attachmentChange)) {
-        const articleInstance = await parentArticleRef.getResolvedValue(elementsSource)
-        if (articleInstance === undefined) {
-          log.error(`Couldn't get article ${parentArticleRef} in the elementsSource`)
+
+  const articleToModifiedAttachments: Record<string, [InstanceElement, ModificationChange<InstanceElement>[]]> = {}
```
netama, line 224: it's hard to understand what the values are and I don't think there's any advantage to using an array here, can we make them an object with meaningful field names?

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -201,47 +220,69 @@ const handleArticleAttachmentsPreDeploy = async ({ changes, client, elementsSour
...
+        // We can't really modify article attachments in Zendesk
+        // To do so we're going to delete the existing attachment and create a new one instead
+        if (isModificationChange(attachmentChange)) {
+          const articleInstance = await parentArticleRef.getResolvedValue(elementsSource)
+          if (articleInstance === undefined || !isInstanceElement(articleInstance)) {
+            log.error(`Couldn't get article ${parentArticleRef} in the elementsSource`)
+            await deleteArticleAttachment(client, attachmentInstance)
+            return
+          }
+          const articleName = articleInstance.elemID.getFullName()
```
netama, line 253: tiny nit: this is the elem id, let's either use something like `articleElemID` or shorten and use just the last part

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -201,47 +220,69 @@ const handleArticleAttachmentsPreDeploy = async ({ changes, client, elementsSour
...
+        }
+        const parentArticleName = parentArticleRef.elemID.name
+        articleNameToAttachments[parentArticleName] = (
+          articleNameToAttachments[parentArticleName] || []
+        ).concat(attachmentInstance.value.id)
+      })
+  )
+
+  Object.values(articleToModifiedAttachments).forEach(async ([article, modificationChanges]) => {
+    const attachmentChangesById = _.keyBy<ModificationChange<InstanceElement>>(
```
netama, line 266: let's try to avoid the cast? (should be possible after changing the structure as suggested above)

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -201,47 +220,69 @@ const handleArticleAttachmentsPreDeploy = async ({ changes, client, elementsSour
...
+      modificationChanges,
+      attachment => getChangeData(attachment).value.id
```
netama, line 268: we shouldn't assume the id exists without verifying

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -201,47 +220,69 @@ const handleArticleAttachmentsPreDeploy = async ({ changes, client, elementsSour
...
+    const [successes, failures] = _.partition(results, result => result.status === SUCCESS_STATUS_CODE)
+    // On success, delete the old attachment
+    // On failure, alert and delete the new attachment
+    await Promise.all([
+      ...successes.map(
+        ({ ids }) => ids.forEach(id => deleteArticleAttachment(client, attachmentChangesById[id].data.before))
+      ),
+      ...failures.map(({ ids }) => ids.forEach(id => {
+        const afterAttachment = attachmentChangesById[id].data.after
+        log.error(`Association of attachment ${afterAttachment.elemID.name} with id ${afterAttachment.value.id} has failed `)
```
netama, line 280: nit: maybe clarify it will be removed?

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -201,47 +220,69 @@ const handleArticleAttachmentsPreDeploy = async ({ changes, client, elementsSour
...
+        ).concat(attachmentInstance.value.id)
+      })
+  )
+
+  Object.values(articleToModifiedAttachments).forEach(async ([article, modificationChanges]) => {
+    const attachmentChangesById = _.keyBy<ModificationChange<InstanceElement>>(
+      modificationChanges,
+      attachment => getChangeData(attachment).value.id
+    )
+    const results = await associateAttachments(client, article, Object.keys(attachmentChangesById).map(parseInt))
```
netama, line 270: nit: why use `Object.keys(attachmentChangesById).map(parseInt)` and not reuse the logic from above (`modificationChanges.map(change => getChangeData(attachment).value.id)` up to the comments) which doesn't require the extra conversion?

netama (commented):
looks good :) added some questions. also, let's make sure there are UTs for the new logic (something checking bulk success with the right call, partial failure, and "full" failure)?

packages/zendesk-adapter/src/filters/article/article.ts
idoser96: I don't think we are expecting any, but I still prefer to be safer for free

packages/zendesk-adapter/src/filters/article/article.ts
idoser96: I don't know, i copied this code, i do think however that the first condition is pointless as it is covered by the second

packages/zendesk-adapter/src/filters/article/article.ts
idoser96: it still requires it, something about the key function (attachment => getChangeData(attachment).value.id)
the solution is either casting or doing parseInt() on the key function

its not the first time im facing this, so if you know how to handle it i'll be happy for help

idoser96: Still required, i think keyBy gets confused when it's key is not typed

packages/zendesk-adapter/src/filters/article/article.ts
idoser96: did not understand this comment, i need to convert it to int because associateAttachments receives a list of numbers

idoser96:
Closing because this moved to https://github.com/salto-io/salto/pull/4753

PR #4630, opened 2023-07-17
SALTO-4457 - Salesforce: Fine-tune profile refs in getAdditionalReferences

tomsellek:
Only create references that are meaningful to the user

---

See the ticket for the definition of "meaningful to the user". Basically we only want to create refs where not creating them would make a difference to what we deploy.

~This PR builds on top of https://github.com/salto-io/salto/pull/4617~

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/additional_references.ts
+++ b/packages/salesforce-adapter/src/additional_references.ts
@@ -18,6 +18,7 @@ import {
   ChangeDataType,
   ElemID,
   GetAdditionalReferencesFunc,
+  getAllChangeData,
   getChangeData,
   InstanceElement,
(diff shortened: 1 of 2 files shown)


amitgoldarb (approved):
This looks great!
I think we should probably add some UT for each profile section though. I think we might be missing some -  I noticed recordTypeVisibility, might be more?
Approving to save cycles

tomsellek:
> I think we should probably add some UT for each profile section though. I think we might be missing some - I noticed recordTypeVisibility, might be more?

Ack. Added UTs for `recordTypeVisibility` and `objectPermissions`

PR #4602, opened 2023-07-11
Workato jira cross service split project issuetype

oneYAD:
In this pull request, I have divided 'project_issuetype' and 'sample_project_issuetype' into two separate entities: 'project_key' and 'issuetype'.

---

This PR is based on https://github.com/salto-io/salto/pull/4600  branch  and is base branch for future SALTO-1619 https://github.com/salto-io/salto/pull/4487   PR. (we want to use Joi while validating Jira block) 

---
Release Notes:

Workato Adapter -

- change jira-connected input arguments. splitting project_issuetype and sample_project_issuetype.

---
User Notifications:

With each fetch using the Workato adapter, any Jira-Connected recipe code will now retrieve 'projectKey' and 'issueType' as separate entities rather than the combined 'project_issue_type'. This change will be reflected similarly in 'sample_project_issuetype'

PR #4632, opened 2023-07-18
fit to workato ids model, new 2 tests

oneYAD:
In  Workato Adapter - Jira recipe. allow split project_issuetype even if dynamicPickListSelction not contained.

---

This is fix for the PR here - https://github.com/salto-io/salto/pull/4602
needed because Workato change interface

---
_Release Notes_: 
Workato adapter
- jira-connected input arguments. now splitting project_issuetype and sample_project_issuetype even if dynamicPickLiseSelection isn't exist.
 

---
_User Notifications_: 
- From now, you will see the projectKey and IssueType instead of project_issuetype even if dynamicPickListSelction is not exist


alonstern (approved):
👌

PR #4633, opened 2023-07-18
SALTO-4311 - Zendesk Adapter - Fix type of missing instances in article body

idoser96:
Use the type name instead of the type field

---

None

---
_Release Notes_: 
Zendesk Adapter:
* Missing instances in article body will have the correct type

---
_User Notifications_: 
Zendesk Adapters:
* article_translations with reference expressions in their body will change, the typename will change from plural to singular (eg articles -> article)


edenhassid (approved):
looks great! 🚀 added a small nit

netama:
this will change the nacls, please add details in the User Notifications section

PR #4634, opened 2023-07-18
Salto 4145 salesforce metadata instances aliases

tamtamirr:

---

---
_Release Notes_: 

---
_User Notifications_: 


amitgoldarb:
@tomermevorach can you look at the aliases here and approve them as well. @tamtamirr Added a link to a workspace with the relevant new aliases

packages/salesforce-adapter/src/filters/metadata_instances_aliases.ts
```diff
@@ -0,0 +1,91 @@
...
+    .split(NAMESPACE_SEPARATOR)) ?? instanceFullName
+}
+
+const getParentAlias = async (
+  instance: MetadataInstanceElement,
+  elementsSource: ReadOnlyElementsSource
+): Promise<string | undefined> => {
+  const [parent] = getParents(instance)
+  if (!isReferenceExpression(parent)) {
+    log.debug('parent is not a reference expression. %o', parent)
```
amitgoldarb, line 50: This is logging an entire element. Maybe just print the elementId?

packages/salesforce-adapter/src/filters/metadata_instances_aliases.ts
```diff
@@ -0,0 +1,91 @@
...
+  const namespace = await getNamespace(instance)
+  const parentAlias = await getParentAlias(instance, elementsSource)
+  const alias = [
+    parentAlias ? `${parentAlias}:` : undefined,
+    getAliasFromFullName(instance.value.fullName),
+    namespace ? `(${namespace})` : undefined,
+  ].filter(isDefined)
+    .join(' ')
+    .replace(/_/g, ' ') // replace all underscores with spaces
+  if (alias !== instance.value.fullName) {
```
amitgoldarb, line 72: Why do we need this if? When can this happen? Also isn't it better to place the alias always?

packages/salesforce-adapter/src/filters/metadata_instances_aliases.ts
```diff
@@ -0,0 +1,91 @@
...
+    ? resolvedParent.annotations[CORE_ANNOTATIONS.ALIAS]
+    : undefined
+}
+
+const setInstanceAlias = async (
+  instance: MetadataInstanceElement,
+  elementsSource: ReadOnlyElementsSource
+): Promise<void> => {
+  const namespace = await getNamespace(instance)
+  const parentAlias = await getParentAlias(instance, elementsSource)
```
amitgoldarb, line 64: I'm really not sure, but can we do the parent after the child? Would this create inconsistent results between the first result and the second?

amitgoldarb (commented):
Looks great!
Added a few comments/thoughts.

packages/salesforce-adapter/src/filters/metadata_instances_aliases.ts
```diff
@@ -0,0 +1,91 @@
...
+  }
+}
+
+const filterCreator: LocalFilterCreator = ({ config }) => ({
+  name: 'metadataInstancesAliases',
+  onFetch: async (elements: Element[]): Promise<void> => {
+    if (config.fetchProfile.isFeatureEnabled('skipAliases')) {
+      log.debug('not adding aliases to metadata instances.')
+      return
+    }
```
tomsellek, line 83: You pretty much get this for free in `ensureSafeFilterFetch`. Any reason not to use it for this filter?

packages/salesforce-adapter/src/filters/metadata_instances_aliases.ts
```diff
@@ -0,0 +1,91 @@
...
+  instance: MetadataInstanceElement,
+  elementsSource: ReadOnlyElementsSource
+): Promise<string | undefined> => {
+  const [parent] = getParents(instance)
+  if (!isReferenceExpression(parent)) {
+    log.debug('parent is not a reference expression. %o', parent)
+    return undefined
+  }
+  const resolvedParent = await parent.getResolvedValue(elementsSource)
+  return isElement(resolvedParent)
```
tomsellek, line 54: Can the parent of a metadata instance be another metadata instance?

packages/salesforce-adapter/src/filters/metadata_instances_aliases.ts
tamtamirr: This is not an Element though :) It's unknown. I've enhanced the check to ignore **undefined** though.

packages/salesforce-adapter/src/filters/metadata_instances_aliases.ts
tamtamirr: This if basically make sure we don't create unnecessary aliases, and it happens quite often when the **Metadata Instance** does not have any `special` use-case. Since the UI for Elements without alias is basically the reverted nacl case of the Element name, we will get the **Instance Full Name** as result. Having a lot of aliases will have impact on the performance and IMO we shouldn't create them when unnecessary. (As they will be indexed both in the RemoteMap and our DB in the SAAS later on). cc. @edenhassid @netama WDYT?

I'll add a comment above the `if` though

packages/salesforce-adapter/src/filters/metadata_instances_aliases.ts
tamtamirr: Nope :) Parents are always **CustomObject** types.

packages/salesforce-adapter/src/filters/metadata_instances_aliases.ts
tamtamirr: Oops I've actually messed up the order of the **namespace** and **parent**. Will fix

packages/salesforce-adapter/src/filters/metadata_instances_aliases.ts
```diff
@@ -0,0 +1,91 @@
...
+const { awu } = collections.asynciterable
+const { isDefined } = values
+const log = logger(module)
+
+const getAliasFromFullName = (instanceFullName: string): string => {
+  const nameWithoutParent = _.last(instanceFullName.split(/[.-]/)) ?? instanceFullName
+  return _.last(nameWithoutParent
+    .replace(ENDS_WITH_CUSTOM_SUFFIX_REGEX, '')
+    .split(NAMESPACE_SEPARATOR)) ?? instanceFullName
+}
```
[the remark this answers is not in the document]
tamtamirr: Exported a new function **apiNamePartsFromApiName** and reusing it within **apiNameParts** since I'm dealing with a string and not Element here

packages/salesforce-adapter/src/filters/metadata_instances_aliases.ts
tamtamirr: This is impossible since **skipAliases** is set to **true** by default, which means we don't want to run this filter by default. No?

packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -206,7 +206,7 @@ export const addDefaults = async (element: ChangeDataType): Promise<void> => {
   }
 }
 
-const ENDS_WITH_CUSTOM_SUFFIX_REGEX = new RegExp(`__(${INSTANCE_SUFFIXES.join('|')})$`)
+export const ENDS_WITH_CUSTOM_SUFFIX_REGEX = new RegExp(`__(${INSTANCE_SUFFIXES.join('|')})$`)
```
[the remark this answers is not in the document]
tamtamirr: Not really internal as this is used in various use-cases:
1. Omitting the custom suffix from apiName
2. Testing if Element is Custom.

I do agree it can stay in the scope of the **utils.ts** module so I've exported a new method **removeCustomSuffix** that uses the regex instead

packages/salesforce-adapter/src/filters/metadata_instances_aliases.ts
amitgoldarb: This if has a lot of assumptions underneath doesn't it? It assumes lots of implementation details of having/not having aliases.
I am not aware of performance impacts on aliases. @edenhassid is that correct?

packages/salesforce-adapter/src/filters/metadata_instances_aliases.ts
amitgoldarb: I'm not sure if you misunderstood my comment, but what happens if the parent alias isn't resolved at this time? Does order matter here somehow? Does this need to be recursive somehow?
Is it even possible? If not why not?

packages/salesforce-adapter/src/filters/metadata_instances_aliases.ts
```diff
@@ -0,0 +1,100 @@
...
+
+
+const { awu } = collections.asynciterable
+const { isDefined } = values
+const log = logger(module)
+
+const getAliasFromFullName = (instanceFullName: string): string => {
+  const nameWithoutParent = _.last(namePartsFromApiName(instanceFullName)) ?? instanceFullName
+  return _.last(removeCustomSuffix(nameWithoutParent)
+    .split(NAMESPACE_SEPARATOR)) ?? instanceFullName
```
amitgoldarb, line 45: I think it's better to use a util for this if it exists rather than using NAMESPACE_SEPERATOR const if available

amitgoldarb (commented):
Looking great!
Added a few more small comments

packages/salesforce-adapter/src/filters/metadata_instances_aliases.ts
edenhassid: it will make the index larger but I am not sure how much it will affect performance. I don't have a strong opinion on having or not having an alias in this case

tamtamirr: Please also refer to @ori-moisis 's comment [here](https://salto-io.atlassian.net/browse/SALTO-4145?focusedCommentId=82225). I don't have a strong opinion though :) I'm just not sure it's worth the performance risk considering SF will be the first adapter to possibly produce thousands of aliases.

packages/salesforce-adapter/src/filters/metadata_instances_aliases.ts
tamtamirr: 1. The parent alias is added as soon as we create the **ObjectType** from the Metadata Instances of **CustomObject**. It will always be resolve here. (Especially because this alias filter should run last)
2. By order matters, do you mean the order of the alias calculation? Right now it's [namespace, instanceFullName, parent] where **namespace** and **parent** are optional.
3. It does not need to be recursive for the time being and probably never, unless we introduce a Parent to **CustomObjects** which is **VERY** unlikely.

@amitgoldarb Hope that covers all your questions

packages/salesforce-adapter/src/filters/metadata_instances_aliases.ts
tamtamirr: There's no such utility and this method is only used in the context of this filter. Wouldn't go and export it, as the split here is solely implementation detail

amitgoldarb (approved):
LGTM!
Thanks for this PR 🙏 
Let's get @tomermevorach 's ack here on the aliases before merging though

PR #4638, opened 2023-07-18
SALTO-4374: Salesforce return isProduction and accountType from validateCredentials

tamtamirr:
Implemented isProduction and accountType in Salesforce validateCredentials.

---

_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/client/client.ts
```diff
@@ -441,21 +456,36 @@ export const getConnectionDetails = async (
...
   if (remainingDailyRequests < minApiRequestsRemaining) {
     throw new ApiLimitsTooLowError(
       `Remaining limits: ${remainingDailyRequests}, needed: ${minApiRequestsRemaining}`
     )
   }
-  return { accountId: orgId }
+  return {
+    accountId: orgId,
+    accountType,
+    isProduction: creds.isSandbox
```
amitgoldarb, line 485: I'm not sure we should use the creds.isSandbox here. Probably we should query this from the API? It will make results more consistent no? One of the options (the chosen one) is user input which can be wrong whilst the other is consistent in the SF realm.
Also if the query failed, how can we be certain of anything really? Shouldn't we return undefined/null?

packages/salesforce-adapter/src/client/client.ts
```diff
@@ -426,10 +426,25 @@ export const loginFromCredentialsAndReturnOrgId = async (
   return identityInfo.organization_id
 }
 
+const queryAccountType = async (conn: Connection, orgId: string): Promise<string | undefined> => {
+  try {
+    const result = await conn.query(`SELECT OrganizationType FROM Organization WHERE Id = '${orgId}'`)
+    const [organizationRecord] = result.records as SalesforceRecord[]
+    log.debug('organization record: %o', organizationRecord)
```
amitgoldarb, line 433: Isn't this record very large? I don't think we should log it as a whole. Let's log only what we need.

packages/salesforce-adapter/src/client/client.ts
```diff
@@ -426,10 +426,25 @@ export const loginFromCredentialsAndReturnOrgId = async (
   return identityInfo.organization_id
 }
 
+const queryAccountType = async (conn: Connection, orgId: string): Promise<string | undefined> => {
```
amitgoldarb, line 429: Do we need to do another query? Is there any chance this data is already returned from loginFromCredentialsAndReturnOrgId?

amitgoldarb (commented):
Looks good. Added a few comments/thoughts

packages/salesforce-adapter/src/client/client.ts
tamtamirr: No, this Record will only contain the **OrganizationType** field that we query here. I've tested it.

packages/salesforce-adapter/src/client/client.ts
tamtamirr: I initially thought that the **IsSandbox** field  on the Organization Object will be true only for **Sandbox Orgs** and not **Scratch Orgs**. The value for both should be **true** so we can rely on that instead of the user input.

packages/salesforce-adapter/src/client/client.ts
tamtamirr: We sadly do. **connection.login** returns `orgId | userId | orgUrl`

packages/salesforce-adapter/src/client/client.ts
tamtamirr: Also returning undefined for both **accountType** and **isProduction** when we fail

amitgoldarb (approved):
LGTM!
THanks for this

tomsellek (approved):

PR #4642, opened 2023-07-19
Bump to version 0.3.42

netama:
# Salto Version 0.3.42

## New features and bug fixes
_Core_
* Fixed issue where static files would sometimes cause elements to be in the deploy plan when only their path changed

_Zendesk adapter_:
* Fixed bug where deployment of an article with more than 20 attachments did not work
* Prevent illegal changes to standard fields
* `app_installation` instances will now have references to the corresponding `app_owned` instances
* Prevent deployment of automations with conditions equal to another automation
* On removal of an automation that is default in Zendesk, the user will be warned that it is not possible on a production environment
* Support missing references to dynamic content items
* Added the option to default omit all inactive instances using the `typeDefaults` config

_Jira adapter_:
* Fixed a bug that prevented fetching more than 200 statuses
* Added validation that prevents the user from adding more than one global context to custom fields
* Support deploying custom fields with more than 10K options

_Netsuite adapter_:
* Fixed open-in-app for some instances
* Support loading files without their attributes in SDF integration
* Added alias annotation to all elements
* Bundle field will be added to elements which are part of a bundle
* sysInfo now contains the env types (PRODUCTION, SANDBOX, BETA, INTERNAL)
* Return Salto Errors on deploy
* Timeout added to soap requests
* Support deletion in partial fetch
* Support deletion of standard instances, custom types and custom records. No support for file cabinet

_Salesforce adapter_:
* Added an adapter fetch configuration that controls reference generation for profiles
* Profile instances now contain more accurate data and will not include data about excluded types
* Improved stability when fetching certain types
* The adapter will use API version 58
* Support additions of *SBAA Approval Rule* instances with *Custom* conditions met
* Support retrieve of unmanaged instances for orgs with namespace defined
* Can now manage history tracking of all standard objects
* Support retrieve of unmanaged instances for orgs with namespace defined

_Workato adapter_:
* Add support for cross-service references to jira

## Expected changes to existing workspaces
_Workato adapter_:
* In `recipe__code` instances with Jira blocks, `projectKey` and `IssueType` will appear instead of `project_issuetype` even if `dynamicPickListSelction` does not exist

_Salesforce adapter_:
* Data in Profile records may be removed if it refers to objects that are excluded from fetch
* Existing Workspaces of Salesforce Orgs that have a namespace defined may encounter addition of instances that were not retrieved before
* The `historyTrackedFields` annotation may be added to some standard object types

_Zendesk adapter_:
* References in nested arrays under the `actions` field of instances will be converted to reference expressions
* Added references to missing dynamic content items


ori-moisis:
> Improved splitElementByPath and filterByID performance

was reverted so it is not in this release yet AFAIK

ori-moisis (approved):

PR #4645, opened 2023-07-20
SALTO-4456: Add flag to CLI restore to also restore paths

alonstern:
Added flag to CLI restore to also restore paths

---

The way to restore paths currently is to delete and re-add all the elements in the workspace.
I added another API in core to do that.

---
_Release Notes_: 
CLI:
- Added `-r` flag to restore to also restore elements to their original paths

---
_User Notifications_: 
None

### DIFF
--- a/packages/cli/src/commands/restore.ts
+++ b/packages/cli/src/commands/restore.ts
@@ -72,7 +72,7 @@ type RestoreArgs = {
     dryRun: boolean
     detailedPlan: boolean
     listPlannedChanges: boolean
-    restorePaths?: boolean
+    reorganizeDirStructure?: boolean
 } & AccountsArg & EnvArg & UpdateModeArg
 
 const applyLocalChangesToWorkspace = async (
@@ -153,9 +153,9 @@ export const action: WorkspaceCommandAction<RestoreArgs> = async ({
 }): Promise<CliExitCode> => {
   const {
     elementSelectors = [], force, dryRun,
-    detailedPlan, listPlannedChanges, accounts, mode, restorePaths,
+    detailedPlan, listPlannedChanges, accounts, mode, reorganizeDirStructure,
   } = input
-  if (elementSelectors.length > 0 && restorePaths) {
+  if (elementSelectors.length > 0 && reorganizeDirStructure) {
     errorOutputLine(Prompts.RESTORE_PATHS_WITH_ELEMENT_SELECTORS, output)
     return CliExitCode.UserInputError
   }
@@ -201,7 +201,9 @@ export const action: WorkspaceCommandAction<RestoreArgs> = async ({
     return CliExitCode.Success
   }
 
-  if (!restorePaths && _.isEmpty(await getRestoreChanges())) {
+  // getRestoreChanges() returns only semantic changes so
+  // if reorganizeDirStructure is passed we need to proceed
+  if (!reorganizeDirStructure && _.isEmpty(await getRestoreChanges())) {
     outputLine(EOL, output)
     outputLine(Prompts.FETCH_NO_CHANGES, output)
     return CliExitCode.Success
@@ -212,7 +214,10 @@ export const action: WorkspaceCommandAction<RestoreArgs> = async ({
     return CliExitCode.Success
   }
 
-  const changesToApply = restorePaths
+  // The restorePathsCore changes also include the elements'
+  // value changes so we when reorganizeDirStructure is passed we need to apply
+  // only those and not the changes from restorePathsCore
+  const changesToApply = reorganizeDirStructure
     ? await restorePathsCore(workspace, accounts)
     : await getRestoreChanges()
 
@@ -272,9 +277,9 @@ const restoreDef = createWorkspaceCommand({
         type: 'boolean',
       },
       {
-        name: 'restorePaths',
+        name: 'reorganizeDirStructure',
         alias: 'r',
-        description: 'Restore the elements to their original path',
+        description: 'Reorganize directory structure and move files to their default locations',
         required: false,
         type: 'boolean',
       },
--- a/packages/cli/src/commands/restore.ts
+++ b/packages/cli/src/commands/restore.ts
@@ -156,7 +156,7 @@ export const action: WorkspaceCommandAction<RestoreArgs> = async ({
     detailedPlan, listPlannedChanges, accounts, mode, reorganizeDirStructure,
   } = input
   if (elementSelectors.length > 0 && reorganizeDirStructure) {
-    errorOutputLine(Prompts.RESTORE_PATHS_WITH_ELEMENT_SELECTORS, output)
+    errorOutputLine(Prompts.REORGANIZE_DIR_STRUCTURE_WITH_ELEMENT_SELECTORS, output)
     return CliExitCode.UserInputError
   }
 
--- a/packages/cli/src/commands/restore.ts
+++ b/packages/cli/src/commands/restore.ts
@@ -17,7 +17,7 @@ import _ from 'lodash'
 import { Workspace, nacl, createElementSelectors } from '@salto-io/workspace'
 import { EOL } from 'os'
 import { logger } from '@salto-io/logging'
-import { CommandConfig, LocalChange, restore } from '@salto-io/core'
+import { CommandConfig, LocalChange, restore, restorePaths as restorePathsCore } from '@salto-io/core'
 import { getChangeData, isRemovalChange } from '@salto-io/adapter-api'
 import { collections, promises, values } from '@salto-io/lowerdash'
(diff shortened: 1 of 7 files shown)


packages/cli/src/commands/restore.ts
```diff
@@ -254,6 +271,13 @@ const restoreDef = createWorkspaceCommand({
         description: 'Print a summary of the planned changes',
         type: 'boolean',
       },
+      {
+        name: 'restorePaths',
+        alias: 'r',
+        description: 'Restore the elements to their original path',
```
adamro, line 277: Reading this as I user I am puzzled by what this does and what happens when this flag is not on.
Regular restore doesn't put it in the "original" path? What is an original path? What is a PATH?! :D  (For the user it's the file maybe, not the path? 🤔 )

Any chance to try and explain the behaviour here in English and we probably want @tomermevorach 's help here with the text

Trying to explain to myself - 
1. Elements that exist in the state and not in the NaCLs, will appear in the NaCLs in the last location the adapter intends them to be in
2. Elements that exist both in the state and in the NaCL, might have parts moving to different files according to the last location the adapter intends them to be in

(And the difference from a normal Restore is that a normal one only does #1, right?)

That's right?

packages/cli/src/commands/restore.ts
```diff
@@ -254,6 +271,13 @@ const restoreDef = createWorkspaceCommand({
         description: 'Print a summary of the planned changes',
         type: 'boolean',
       },
+      {
+        name: 'restorePaths',
```
adamro, line 275: this is a restore command, it's weird that a flag has `restore` in it

packages/cli/src/commands/restore.ts
```diff
@@ -254,6 +271,13 @@ const restoreDef = createWorkspaceCommand({
...
+        alias: 'r',
```
adamro, line 276: `r` seems so random :D

packages/core/src/core/restore.ts
```diff
@@ -102,3 +102,30 @@ export async function createRestoreChanges(
     .flatMap(change => splitDetailedChangeByPath(change, index))
     .toArray()
 }
+
+export const createRestorePathChanges = async (
```
adamro, line 106: It's more `createAlignPathsWithIndexChanges`, no?

packages/core/src/core/restore.ts
```diff
@@ -102,3 +102,30 @@ export async function createRestoreChanges(
...
+  workspaceElements: elementSource.ElementsSource,
```
adamro, line 107: Q/nit: Is there actually a reason to get ElementsSource and not Element[] here? 🤔

packages/core/src/core/restore.ts
```diff
@@ -102,3 +102,30 @@ export async function createRestoreChanges(
...
+  index: remoteMap.RemoteMap<pathIndex.Path[]>,
+  accounts?: string[],
+): Promise<DetailedChange[]> => {
+  const elements = await awu(await (workspaceElements).getAll())
+    .filter(element => accounts === undefined || accounts.includes(element.elemID.adapter))
```
adamro, line 112: They are probably not critical/relevant for this flow but we also have "GLOBAL_ADAPTER" Elements (ie. Builtin/Primitives) that will behave differently if accounts is passed here.
I *think* they are all hidden anyway so we probably don't need to worry about them, and we have full control of their placing even if they aren't. So I think we can ignore this.
I am just writing this as a reminder (also to myself :D) that they exist when we write this kind of code.

packages/core/src/api.ts
```diff
@@ -521,6 +521,15 @@ export async function restore(
   return detailedChanges.map(change => ({ change, serviceChanges: [change] }))
 }
 
+export const restorePaths = async (
```
adamro, line 524: It's more `getAlignPathsWithIndexChanges`, no?

packages/cli/src/commands/restore.ts
```diff
@@ -177,9 +183,16 @@ export const action: WorkspaceCommandAction<RestoreArgs> = async ({
   outputLine(EOL, output)
   outputLine(formatStepStart(Prompts.RESTORE_CALC_DIFF_START), output)
 
-  const changes = await restore(workspace, activeAccounts, validSelectors)
+  let changes: LocalChange[] | undefined
```
adamro, line 186: The code flow here confuses me. 
Why do we need the `getRestoreChanges` function instead of just calculating "restoreChanges" once and working with it?

packages/cli/src/commands/restore.ts
```diff
@@ -199,8 +212,12 @@ export const action: WorkspaceCommandAction<RestoreArgs> = async ({
     return CliExitCode.Success
   }
 
+  const changesToApply = restorePaths
```
adamro, line 215: Can you write a comment here explaining that the "paths" changes also include the "restore" changes so if the flag is on we only apply those?

packages/cli/src/commands/restore.ts
```diff
@@ -188,7 +201,7 @@ export const action: WorkspaceCommandAction<RestoreArgs> = async ({
     return CliExitCode.Success
   }
 
-  if (_.isEmpty(changes)) {
+  if (!restorePaths && _.isEmpty(await getRestoreChanges())) {
```
adamro, line 204: We might need a comment here explaining why with the flag even if there are no "restoreChanges" we still need to continue the flow, for the path changes.

packages/cli/test/commands/restore.test.ts
```diff
@@ -540,4 +542,50 @@ describe('restore command', () => {
       expect(result).toBe(CliExitCode.Success)
     })
   })
+
+  describe('restore paths', () => {
```
adamro, line 546: Shouldn't we also test that `applyLocalChangesToWorkspace` was called with the right changes if the flag is on?

packages/cli/src/commands/restore.ts
alonstern: That's correct. @tomermevorach do you have an idea for the name, alias and description of this flag?

packages/core/src/core/restore.ts
alonstern: Isn't the word `index` exposing implementation here? We can call it `createAlignPathsWithStateChanges` but I tend to think it's less good because every restore is restore with state and align is just a different word for restore in this case no?

adamro: Convinced me :)

packages/core/src/core/restore.ts
alonstern: Not really, changing

packages/cli/src/commands/restore.ts
alonstern: Because there cases where we don't need to calculate it at all, when `restorePaths` is true and `listPlannedChanges || dryRun` is false

packages/cli/src/commands/restore.ts
tomermevorach: Flag:
--reorganize-dir-structure, -r 
Desc:
Reorganize directory structure and move files to their default locations.

packages/cli/src/commands/restore.ts
adamro: Oh right, that random behaviour that is different from deploy :\

So I understand why you did this. 
To me it's still harder to read than alternatives I think
(For example, extracting the "apply" phase to a function and calling it early if we are in the case we don't need to calculate the restoreChanges in and only in the other cases continue, after calculating it)
But that's pure style, you can leave it as is.

alonstern: I tried extracting the apply phase but IMO it makes to code a bit more complicated. It means I need to kind of cut the function in the middle (`outputLine(formatStepCompleted(Prompts.RESTORE_CALC_DIFF_FINISH), output)` will be the first line of the function), and I need to do multiple checks on `listPlannedChanges`, `dryRun` and `restorePaths`, so if it's ok with you I think I would rather the leave it like this

packages/cli/src/prompts.ts
```diff
@@ -216,6 +216,7 @@ ${Prompts.ACCOUNT_ADD_HELP}`
   public static readonly RESTORE_SUCCESS_FINISHED = 'Done! Your NaCL files are now updated with the latest changes.'
   public static readonly RESTORE_UPDATE_WORKSPACE_FAIL = 'Failed to apply changes to your NaCL files.'
   public static readonly SHOULD_EXECUTE_RESTORE = 'Do you want to restore?'
+  public static readonly RESTORE_PATHS_WITH_ELEMENT_SELECTORS = 'Restore paths flag with element selectors is not supported'
```
adamro, line 219: Change the text according to the new name

adamro (approved):

PR #4648, opened 2023-07-20
SALTO-4486 deal with case where internal ID is an empty string, and don't return it in getTypeIdenitifier

benisraeladi:
Changes to retrieve API caused internal ID to have empty string. 

---

---
_Release Notes_: 
Salesforce:
- Fix broken go-to-service URLs

---
_User Notifications_: _None_

### DIFF
--- a/packages/salesforce-adapter/src/fetch.ts
+++ b/packages/salesforce-adapter/src/fetch.ts
@@ -236,7 +236,7 @@ const getPropsWithFullName = (
 
 const getInstanceFromMetadataInformation = (metadata: MetadataInfo,
   filePropertiesMap: Record<string, FileProperties>, metadataType: ObjectType): InstanceElement => {
-  const newMetadata = filePropertiesMap[metadata.fullName]?.id
+  const newMetadata = filePropertiesMap[metadata.fullName]?.id !== undefined && filePropertiesMap[metadata.fullName]?.id !== ''
     ? { ...metadata, [INTERNAL_ID_FIELD]: filePropertiesMap[metadata.fullName]?.id } : metadata
   return createInstanceElement(newMetadata, metadataType,
     filePropertiesMap[newMetadata.fullName]?.namespacePrefix,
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/src/elements_url_retreiver/lightining_url_resolvers.ts
```diff
@@ -66,8 +66,13 @@ const METADATA_TYPE_TO_URI: Record<string, string> = {
...
 }
 
-const getTypeIdentifier = async (element?: Element): Promise<string | undefined> =>
-  (element === undefined ? undefined : (getInternalId(element) ?? apiName(element)))
+const getTypeIdentifier = async (element?: Element): Promise<string | undefined> => {
+  if (element === undefined) {
+    return undefined
+  }
+  const internalId = getInternalId(element)
+  return internalId !== null && internalId !== undefined && internalId !== '' ? internalId : apiName(element)
```
amitgoldarb, line 74: I think this is the same. But need to check every use case.
Also I think this should have a UT as well.
```suggestion
  return internalId ? internalId : apiName(element)
```

amitgoldarb (approved):
Approved with a very small comment.
Also I think this might be needing a test.
Approving to save cycles.

packages/salesforce-adapter/src/elements_url_retreiver/lightining_url_resolvers.ts
benisraeladi: This would convert into the original code using `??`, which doesn't apply to empty string

tomsellek: But `!internalId` would cover all the cases, no?

tomsellek (approved):

tamtamirr (commented):
LGTM :)

tamtamirr (approved):

PR #4649, opened 2023-07-20
SALTO-4504: make parallel pagination keep its order

shayc331:
_make parallel pagination keep its order_

---

* solving race condition that we found in this [ticket](https://salto-io.atlassian.net/browse/SALTO-4504) 

performance tests:
I fetched a large env couple of times with and without the new implementation:
old implementation time (in seconds): 
* 2:02.21
* 3:04.13
* 2:19.06
* 2:35.83
* 1:32.10

new implementation time (in seconds): 
* 2:34.74
* 3:16.73
* 1:51.87
* 1:34.27
* 1:46.61

---
_Release Notes_: 
Jira Adapter:
* fix a bug of inconsistent order in elements that contain a list with many objects

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/adapter-components/src/client/pagination/pagination_async.ts
+++ b/packages/adapter-components/src/client/pagination/pagination_async.ts
@@ -31,18 +31,40 @@ const allSettled = async <T>(promises: IterableIterator<Promise<T>>): Promise<vo
   await Promise.all(Array.from(promises).map(p => p.catch(() => undefined)))
 }
 
-class PromisesQueue<T> {
-  private readonly promises: Promise<T>[]
+type PaginationResult = {
+  page: ResponseValue[]
+  response: Response<ResponseValue| ResponseValue[]>
+  additionalArgs: Record<string, string>
+  yieldResult: boolean
+}
+
+type GetPageArgs = {
+  // eslint-disable-next-line no-use-before-define
+  promisesQueue: PromisesQueue
+  paginationFunc: PaginationFunc
+  client: HTTPReadClientInterface
+  extractPageEntries: PageEntriesExtractor
+  customEntryExtractor?: PageEntriesExtractor
+  getParams: ClientGetWithPaginationParams
+  pageSize: number
+  usedParams: Set<string>
+}
+
+class PromisesQueue {
+  private readonly promises: Promise<PaginationResult>[]
+  private shouldContinue: boolean
   constructor() {
     this.promises = []
+    this.shouldContinue = true
   }
 
-  enqueue(promise: Promise<T>): void {
+  enqueue(pageArgs: GetPageArgs, additionalArgs: Record<string, string>): void {
+    const promise = this.singlePagePagination(pageArgs, additionalArgs)
     this.promises.push(promise)
     log.debug(`Added promise to pagination queue. Queue size: ${this.promises.length}`)
   }
 
-  async dequeue(): Promise<T> {
+  async dequeue(): Promise<PaginationResult> {
     const settledPromise = await this.promises.shift()
     if (settledPromise === undefined) {
       // this may occur if the queue is empty which we checked it before calling dequeue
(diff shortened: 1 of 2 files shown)


packages/adapter-components/src/client/pagination/pagination_async.ts
```diff
@@ -33,27 +33,35 @@ const allSettled = async <T>(promises: IterableIterator<Promise<T>>): Promise<vo
 
 type KeyFromPromiseResult<T> = (result: T) => string
 class PromisesQueue<T> {
-  private readonly promises: Map<string, Promise<T>>
+  private readonly promises: Promise<T>[]
+  private readonly promisesKeys: Set<string>
```
alonstern, line 37: It looks like you only write to it but never read

packages/adapter-components/src/client/pagination/pagination_async.ts
```diff
@@ -33,27 +33,35 @@ const allSettled = async <T>(promises: IterableIterator<Promise<T>>): Promise<vo
...
+    this.promisesKeys.add(key)
+    this.promises.push(promise)
+    log.debug(`Added promise to pagination queue. Queue size: ${this.promises.length}`)
   }
 
   async dequeue(): Promise<T> {
-    const settledPromise = await Promise.race(this.promises.values())
-    this.promises.delete(this.hashFunc(settledPromise))
-    log.debug(`Removed promise from pagination queue. Queue size: ${this.promises.size}`)
+    const settledPromise = await this.promises.shift()
```
alonstern, line 52: You still need to change in https://github.com/salto-io/salto/pull/4649/files#diff-82e6e9bae39969f98ccae225bb8008294cabd5136accb81e707176b2043a8e70R200 to do the
```
addNextPages({
          pageArgs,
          page: result.page,
          response: result.response,
          additionalArgs: result.additionalArgs,
        })
```
when any request finishes and not only the next one

alonstern (commented):
👌

packages/adapter-components/src/client/pagination/pagination_async.ts
shayc331: deleted it

packages/adapter-components/test/client/pagination.test.ts
```diff
@@ -475,36 +475,6 @@ describe('client_pagination', () => {
       paginationFunc.mockReturnValueOnce([{ page: '2' }, { page: '4' }])
         .mockReturnValueOnce([{ page: '6' }]).mockReturnValueOnce([{ page: '8' }])
     })
-    it('should not call additional calls before asked', async () => {
```
shayc331, line 478: note: 
Those tests are not relevant for the new implementation. 
In the previous implementation we waited for the user to yield the page and only then added the new pages to the queue, and now we will add the new pages to the queue once a promise is resolved.

packages/adapter-components/src/client/pagination/pagination_async.ts
```diff
@@ -189,12 +195,6 @@ export const traverseRequestsAsync: (
...
         numResults += result.page.length
-        addNextPages({
-          pageArgs,
-          page: result.page,
-          response: result.response,
-          additionalArgs: result.additionalArgs,
-        })
       }
     } catch (e) {
       // avoid leaking promises
```
alonstern, line 200: We have an issue here that now when there is an error we will still create new promises and wait for them whereas before we waited only for the existing promises. We probably set some variable here that will make the promises to stop creating promises

alonstern (commented):
LGTM 🔥 @IdoZakk should probably also have a look.
Also this fixes a bug so could you update the release notes?

packages/adapter-components/test/client/pagination.test.ts
IdoZakk: It makes the all yield mechanism here absolute if the calls keep going after you have made the first call, but I guess we are past this point

IdoZakk (approved):
👍 
Not too happy with the solution, but I guess that is what we have for now

PR #4650, opened 2023-07-20
SALTO-4281: EmailTemplate with attachments deploy bugfix

tamtamirr:
Support deploying EmailTemplate instances with attachments.

---

We did not convert the **attachments content** back to **base64** upon deploy.

---
_Release Notes_: 
Salesforce Adapter:
- Support deploying EmailTemplate instances with attachments.

---
_User Notifications_: 
_None_


packages/salesforce-adapter/test/filters/email_template_static_files.test.ts
```diff
@@ -37,19 +45,15 @@ describe('emailTemplate static files filter', () => {
   })
 
   describe('on fetch', () => {
-    type FilterType = FilterWith<'onFetch'>
-    let filter: FilterType
-
     describe('attachment as an object', () => {
-      beforeAll(async () => {
+      beforeEach(async () => {
```
tamtamirr, line 49: **beforeAll** executes before the parent **beforeEach**, so I've changed it. Also, it's more correct.

packages/salesforce-adapter/src/filters/email_template_static_files.ts
```diff
@@ -110,6 +125,24 @@ const filter: LocalFilterCreator = () => ({
...
+  // Convert EmailTemplate attachments to base64 string
+  preDeploy: async changes => {
+    (await getAttachmentsFromChanges(changes))
+      .forEach(attachment => {
+        if (_.isBuffer(attachment.content)) {
+          attachment.content = attachment.content.toString('base64')
+        }
+      })
+  },
+  // Convert EmailTemplate attachments back to binary files
```
benisraeladi, line 137: Why do we need to convert back on `onDeploy`? Shouldn't that only happen on `onFetch`?

tamtamirr: If you don't convert back changes performed in **preDeploy** we will have unwanted applied changes to the **nacls**

packages/salesforce-adapter/src/filters/email_template_static_files.ts
```diff
@@ -110,6 +125,24 @@ const filter: LocalFilterCreator = () => ({
...
+  onDeploy: async changes => {
+    (await getAttachmentsFromChanges(changes))
+      .forEach(attachment => {
+        if (_.isString(attachment.content)) {
+          attachment.content = Buffer.from(attachment.content, 'base64')
```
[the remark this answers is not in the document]
tomsellek: I think keeping things as stateless as possible is a good goal to have

tomsellek (approved):
LGTM FWIW LMAO

packages/salesforce-adapter/src/filters/email_template_static_files.ts
```diff
@@ -110,6 +125,24 @@ const filter: LocalFilterCreator = () => ({
       .filter(isInstanceOfType(EMAIL_TEMPLATE_METADATA_TYPE))
       .forEach(organizeStaticFiles)
   },
+  // Convert EmailTemplate attachments to base64 string
```
Uri-Bar, line 128:
```suggestion
  // Convert EmailTemplate attachments content from buffer to base64 string
```

packages/salesforce-adapter/src/filters/email_template_static_files.ts
```diff
@@ -110,6 +125,24 @@ const filter: LocalFilterCreator = () => ({
...
+  preDeploy: async changes => {
+    (await getAttachmentsFromChanges(changes))
+      .forEach(attachment => {
+        if (_.isBuffer(attachment.content)) {
```
Uri-Bar, line 132: after we discussed f2f I think you should add a log when you don't receive a buffer

packages/salesforce-adapter/src/filters/email_template_static_files.ts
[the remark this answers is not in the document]
Uri-Bar: I agree, this is much cleaner

packages/salesforce-adapter/test/filters/email_template_static_files.test.ts
```diff
@@ -88,4 +91,30 @@ describe('emailTemplate static files filter', () => {
...
+
+    let deployedInstance: InstanceElement
+    beforeEach(async () => {
+      deployedInstance = createInstanceElement({ fullName: 'unfiled$public/emailTemplate',
+        content: staticContent,
+        attachments: [{ name: ATTACHMENT_NAME, content: ATTACHMENT_AS_BUFFER }] },
+      mockTypes.EmailTemplate)
+    })
+
+    it('should encode the attachment to base64 string on preDeploy and revert back to binary buffer on onDeploy', async () => {
```
Uri-Bar, line 106: I think you should dived it into two tests, the preDeploy and the OnDeploy should be checked separately

packages/salesforce-adapter/src/filters/email_template_static_files.ts
```diff
@@ -99,6 +103,17 @@ const organizeStaticFiles = async (instance: InstanceElement): Promise<void> =>
   }
 }
 
+const getAttachmentsFromChanges = async (changes: Change[]): Promise<Attachment[]> => (
```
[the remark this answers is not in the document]
tamtamirr: **isInstanceOfType** is async :)

packages/salesforce-adapter/test/filters/email_template_static_files.test.ts
tamtamirr: I think it's more correct to test **preDeploy** and **onDeploy** together with the same changes as that's how the **adapter** behaves.

Uri-Bar: I am pretty sure that this is the convention in the OSS that you have a test for pre deploy and one for on deploy, and basically whenever you can you should want these two different funcs to be checked on their own.
If you want you can add a test thats checks the whole flow but I dont think it should be instead

tamtamirr: We can take this F2F but I disagree with this approach. While we say **preDeploy** and **onDeploy** should be stateless it's not precisely true, as we run on the same instance of **filter**. Separating them to two different tests will execute each method independently on a separate instance of filter. This will not work in cases where we decide to store the original changes before modifying them in **preDeploy** (We already have various filters that do that in Salesforce, see [minifyDeployFilter](https://github.com/salto-io/salto/blob/bc92deeeb1263ba88b00d87383350996f68237a6/packages/salesforce-adapter/src/filters/minify_deploy.ts#L101-L100) as an example). I'm also not a fan of this disconnection considering the actual deploy flow of the Salesforce Adapter is executing the **onDeploy** right after the **preDeploy** with the applied changes. cc. @ori-moisis @adamro WDYT?

packages/salesforce-adapter/src/filters/email_template_static_files.ts
```diff
@@ -33,23 +42,18 @@ const log = logger(module)
...
 type Attachment = {
   name: string
-  content: string | StaticFile
+  content: string | Buffer | StaticFile
 }
 
-type EmailAttachmentsArray = {
-  attachments: Attachment[]
+const isEmailTemplateAttachment = (attachment: unknown): attachment is Attachment => {
+  const name = _.get(attachment, 'name')
```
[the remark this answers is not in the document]
tamtamirr: attachment is of type **unknown**. Moreover, even if we chose to treat **atachment** as **any** (a.k.a Values in Salto 🔫 ) attachment.name on a non **PlainObject** type will actually throw runtime error. Using lodash get is safe in that sense.

packages/salesforce-adapter/test/filters/email_template_static_files.test.ts
ori-moisis: you are both right to some extent
the overall aim should be to test things separately when possible, in the case of preDeploy + onDeploy specifically they sometimes need to be tested together because they maintain state between calls.
In the case where `preDeploy` and `onDeploy` are stateless (as I understand this is the case in this example) both options are open and could be valid.
on the one hand the "pure" test is available in this specific case and we could say it should be preferred. on the other hand it creates a potentially unhealthy coupling of the test and the current implementation (e.g - if the implementation changes to be stateful, it would require updating tests even though the behavior didn't change).

Uri-Bar (approved):

PR #4651, opened 2023-07-20
SALTO-4513 remove redundant logs

netama:
We have generic logs for filters and change validators, removing redundant ones that were added manually

---
_Release Notes_: 
None

---
_User Notifications_: 
None


IdoZakk (approved):
💯

PR #4652, opened 2023-07-20
SALTO-4520 - Salesforce: Allow different default values for adapter configs

tomsellek:
Allow a custom default value for adapter configs. This PR also changes the default value of `fetchProfilesUsingReadApi` - both to show how it's done and because we've been wanting to do it anyway.

---

If we want to use adapter configs as feature flags we need to be able to change their default values without renaming them. This means the "default is always true" logic no longer holds. So let's change it slightly to "default is true, unless explicitly overridden".

- [x] PROD-156

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/fetch_profile/fetch_profile.ts
+++ b/packages/salesforce-adapter/src/fetch_profile/fetch_profile.ts
@@ -33,6 +33,14 @@ export type FetchProfile = {
   readonly addNamespacePrefixToFullName: boolean
 }
 
+type OptionalFeaturesDefaultValues = {
+  [FeatureName in keyof OptionalFeatures]?: boolean
+}
+
+const optionalFeaturesDefaultValues: OptionalFeaturesDefaultValues = {
+  fetchProfilesUsingReadApi: false,
+}
+
 export const buildFetchProfile = ({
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/src/fetch_profile/fetch_profile.ts
```diff
@@ -33,6 +33,14 @@ export type FetchProfile = {
   readonly addNamespacePrefixToFullName: boolean
 }
 
+type OptionalFeaturesDefaultValues = {
+  [FeatureName in keyof OptionalFeatures]?: boolean
+}
+
+const optionalFeaturesDefaultValues: OptionalFeaturesDefaultValues = {
+  fetchProfilesUsingReadApi: false,
+}
```
benisraeladi, line 42: Make sure to update `salesforce/config_doc.md`

tomsellek: Gah! Thanks for noticing!

amitgoldarb (approved):
Thanks for this one!

benisraeladi (approved):

PR #4653, opened 2023-07-21
SALTO-2781: Add filter and CV for classTranslationList changes on data elements

Shulik95:
_Deploying data elements with changes to classTranslationList field fails, Added filter to remove it in pre-deploy and CV to let the user know_

---

_The ticket - https://salto-io.atlassian.net/browse/SALTO-2781_

---
_Release Notes_: 
_Netsuite Adapter_:
* changes to the classTranslationList field in data elements are not deployable.

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/netsuite-adapter/src/config.ts
+++ b/packages/netsuite-adapter/src/config.ts
@@ -549,6 +549,7 @@ export type NetsuiteValidatorName = (
   | 'undeployableConfigFeatures'
   | 'extraReferenceDependencies'
   | 'rolePermission'
+  | 'dataElementsOmitFields'
 )
 
 export type NonSuiteAppValidatorName = (
--- a/packages/netsuite-adapter/src/filters/data_elements_omit_fields.ts
+++ b/packages/netsuite-adapter/src/filters/data_elements_omit_fields.ts
@@ -1,0 +1,58 @@
+/*
+*                      Copyright 2023 Salto Labs Ltd.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+import { isAdditionOrModificationChange, isInstanceChange, isModificationChange, isAdditionChange, Change, ModificationChange, getChangeData, ChangeDataType } from '@salto-io/adapter-api'
+import _ from 'lodash'
+import { LocalFilterCreator } from '../filter'
+import { isDataElementChange } from '../change_validators/inactive_parent'
+import { getElementValueOrAnnotations } from '../types'
+
+export const CLASS_TRANSLATION_LIST = 'classTranslationList'
+const FIELDS_TO_OMIT = [CLASS_TRANSLATION_LIST]
+
+const fieldIsModified = (
+  change: ModificationChange<ChangeDataType>,
+  fieldToOmit: string,
+): boolean => {
+  const { before, after } = change.data
+  return !_.isEqual(
+    getElementValueOrAnnotations(before)[fieldToOmit],
+    getElementValueOrAnnotations(after)[fieldToOmit]
+  )
+}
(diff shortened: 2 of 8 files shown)


packages/netsuite-adapter/src/filters/data_elements_omit_fields.ts
```diff
@@ -0,0 +1,61 @@
...
+import { isDataElementChange } from '../change_validators/inactive_parent'
+import { getElementValueOrAnnotations } from '../types'
+
+const { awu } = collections.asynciterable
+
+export const CLASS_TRANSLATION_LIST = 'classTranslationList'
+const FIELDS_TO_OMIT = [CLASS_TRANSLATION_LIST]
+
+const fieldIsModified = (
+  change: ModificationChange<ChangeDataType>,
```
dantal4, line 30: nit - change to `ModificationChange<InstanceElement>`, and in line 35&36 you'll be able to use `instance.value` directly.

packages/netsuite-adapter/src/filters/data_elements_omit_fields.ts
```diff
@@ -0,0 +1,61 @@
...
+  fieldToOmit: string,
+): boolean => {
+  const { before, after } = change.data
+  return !_.isEqual(
```
dantal4, line 34: There's a function called `isEqualValues` (in adapter-api) that takes care of special values (like references and static files) which would give a more accurate result.

packages/netsuite-adapter/src/filters/data_elements_omit_fields.ts
```diff
@@ -0,0 +1,61 @@
...
+    getElementValueOrAnnotations(before)[fieldToOmit],
+    getElementValueOrAnnotations(after)[fieldToOmit]
+  )
+}
+
+export const shouldOmitField = (
+  change: Change,
+  fieldToOmit: string
+): boolean => (isAdditionChange(change) || (isModificationChange(change) && fieldIsModified(change, fieldToOmit)))
```
dantal4, line 43: in case of addition change, we'll return true even when the field doesn't exist, which I don't think is the right behavior.

packages/netsuite-adapter/src/filters/data_elements_omit_fields.ts
```diff
@@ -0,0 +1,61 @@
...
+  name: 'dataElementsOmitFields',
+  preDeploy: async changes => {
+    await awu(changes)
+      .filter(isAdditionOrModificationChange)
+      .filter(isInstanceChange)
+      .filter(isDataElementChange)
+      .filter(change => shouldOmitField(change, CLASS_TRANSLATION_LIST))
+      .map(getChangeData)
+      .forEach(instanceElement => {
+        instanceElement.value = _.omit(instanceElement.value, FIELDS_TO_OMIT)
```
dantal4, line 56: 1. In line 53 we check if `CLASS_TRANSLATION_LIST` should be omitted, but here we omit a different list. I know that currently they are equal, but that can easily become a bug in the future.
2. We probably should omit the value in both before & after, otherwise in `data_instances_null_fields` we'll mark that field as "should be deleted" instead of just ignoring it (please check what happens in NS in both cases - when there's a value in that field in NS UI).

packages/netsuite-adapter/test/change_validators/data_elements_omit_fields.test.ts
```diff
@@ -0,0 +1,93 @@
...
+      subsidiaryType,
+      {
+        name: 'subsidiary_name',
+        state: 'AZ',
+        classTranslationList,
+      }
+    )
+    after = subsidiaryInstance.clone()
+  })
+  it('should have a warning on addition change', async () => {
```
dantal4, line 58: in continuation to my previous comment - please test that there's no warning when the field doesn't exist.

packages/netsuite-adapter/src/config.ts
```diff
@@ -549,6 +549,7 @@ export type NetsuiteValidatorName = (
   | 'undeployableConfigFeatures'
   | 'extraReferenceDependencies'
   | 'rolePermission'
+  | 'dataElementsOmitFields'
```
dantal4, line 552: We should add it a field in `changeValidatorConfigType`.
I would expect it to break in this case - can you define `ChangeValidatorConfig` this way? -
```
type ChangeValidatorConfig = Record<
  NetsuiteValidatorName | NonSuiteAppValidatorName | OnlySuiteAppValidatorName,
  boolean | undefined
>
```

dantal4 (commented):
Great Work!

PR #4654, opened 2023-07-23
SALTO-4467: add dependency between attachment to article

edenhassid:
add dependency between attachment to article

---

---
_Release Notes_: 
* deploy of addition of attachments to an article with change validator errors will fail.

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/dependency_changers/article_attachment_change.ts
+++ b/packages/zendesk-adapter/src/dependency_changers/article_attachment_change.ts
@@ -1,0 +1,65 @@
+/*
+*                      Copyright 2023 Salto Labs Ltd.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


packages/zendesk-adapter/src/dependency_changers/article_attachment_change.ts
```diff
@@ -0,0 +1,65 @@
...
+  : DependencyChange[] => {
+  const articleModificationChanges = changes.filter(change =>
+    (getChangeData(change.change).elemID.typeName === ARTICLE_TYPE_NAME && isModificationChange(change.change)))
+  const articleAttachmentAdditionChanges = changes.filter(change =>
+    (getChangeData(change.change).elemID.typeName === ARTICLE_ATTACHMENT_TYPE_NAME && isAdditionChange(change.change)))
+
+  const articleElemIdToChange = _.keyBy(articleModificationChanges, getNameFromChange)
+  return articleAttachmentAdditionChanges.map(change => {
+    const parent = getParents(getChangeData(change.change))[0]
+    const parentElemId = parent ? parent.elemID.getFullName() : ''
```
netama, line 43: 1. we should just exit if there is no parent, no? what do we get from using an empty string?
2. please be explicit - `parent !== undefined`

netama (commented):
cool 🎉 added some minor comments

netama (approved):

PR #4655, opened 2023-07-23
Keep bundle inner-types relations

dantal4:
We should create the bundle type and its inner types once, so the field types of the bundle type will point to the same inner type objects that we return.

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


omrilit (approved):
LGTM

PR #4657, opened 2023-07-24
SALTO-3829:  refactor workflow transitions list to map

IdoZakk:
_Changing the workflow transitions list to a map to support transition links_

---
The transitions will have a key with the following logic:
{Name}::From: {list of from names, comma separated}::{transition type}
Initial status will have 'none' in the from, and global statuses will have 'any status'
I added an additional type named Circular for global circular transitions, as it helps to differentiate same name transitions

Here is an example of the transition name:
```
Close__From__Open__Directed@fffsff = {
      name = "Close"
      description = ""
      from = [
        {
          id = jira.Status.instance.open
          sourceAngle = 78.11134196037203
          targetAngle = -70.01689347810002
        },
      ]
      to = jira.Status.instance.closed
      type = "directed"
}
```

For duplicate names, I currently add a number to the name (both the transition name and the key), it should be further discussed. A change validator will be added based on the discussion's decision

I will also add noise reduction

---
_Release Notes_: 
Jira Adapter:
* Changed workflow transitions from a list to a map

---
_User Notifications_: 
Jira Adapter:
* Workflow transitions will be kept in a map from now on. The key will be based on the name of the transition and the from field

### DIFF
--- a/packages/jira-adapter/src/filters/workflow/workflow_structure_filter.ts
+++ b/packages/jira-adapter/src/filters/workflow/workflow_structure_filter.ts
@@ -36,15 +36,15 @@ const FETCHED_ONLY_INITIAL_POST_FUNCTION = [
   'IssueStoreFunction',
 ]
 
-const PARTS_SEPARATOR = '::'
-
-const TYPE_TO_FROM_MAP = new Map([
-  ['Initial', 'none'],
-  ['Global', 'any status'],
-  ['Circular', 'any status'],
-])
+export const TRANSITION_PARTS_SEPARATOR = '::'
 
 type TransitionType = 'Directed' | 'Initial' | 'Global' | 'Circular'
+
+const TYPE_TO_FROM_MAP: Record<Exclude<TransitionType, 'Directed'>, string> = {
+  Initial: 'none',
+  Global: 'any status',
+  Circular: 'any status',
+}
 
 export const createStatusMap = (statuses: Status[]): Map<string, string> => new Map(statuses
   .filter((status): status is {id: string; name: string} => typeof status.id === 'string' && status.name !== undefined)
@@ -78,9 +78,9 @@ export const getTransitionKey = (transition: Transition, statusesMap: Map<string
       .map(from => statusesMap.get(from) ?? from)
       .sort()
       .join(',')
-    : TYPE_TO_FROM_MAP.get(type) ?? ''
+    : TYPE_TO_FROM_MAP[type]
 
-  return naclCase([transition.name ?? '', `From: ${fromSorted}`, type].join(PARTS_SEPARATOR))
+  return naclCase([transition.name ?? '', `From: ${fromSorted}`, type].join(TRANSITION_PARTS_SEPARATOR))
 }
 
 const transformProperties = (item: Status | Transition): void => {
@@ -188,11 +188,7 @@ const transformRules = (rules: Rules, transitionType?: string): void => {
 
 const transformTransitions = (value: Value): SaltoError[] => {
   const statusesMap = createStatusMap(value.statuses ?? [])
-  const maxCounts: Record<string, number> = {}
-  value.transitions.forEach((transition: Transition) => {
-    const key = getTransitionKey(transition, statusesMap)
-    maxCounts[key] = (maxCounts[key] ?? 0) + 1
-  })
+  const maxCounts = _(value.transitions).map(transition => getTransitionKey(transition, statusesMap)).countBy().value()
 
   const counts: Record<string, number> = {}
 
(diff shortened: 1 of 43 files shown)


packages/jira-adapter/src/filters/workflow/workflow_structure_filter.ts
```diff
@@ -36,6 +36,53 @@ const FETCHED_ONLY_INITIAL_POST_FUNCTION = [
   'IssueStoreFunction',
 ]
 
+const PARTS_SEPARATOR = '::'
```
alonstern, line 39: Can we use `_` instead of `:`?  if there aren't any other special character in the other names, there won't be @ with a bunch of letters in the end of the string, which is a little more readable

packages/jira-adapter/src/filters/workflow/workflow_structure_filter.ts
```diff
@@ -36,6 +36,53 @@ const FETCHED_ONLY_INITIAL_POST_FUNCTION = [
...
+// The types can be Directed, Initial, Global, and Circular (which is global circular)
+export const getTransitionKey = (transition: Transition, statusesMap: Map<string, string>): string => {
+  const type = getTransitionType(transition)
+  const fromSorted = type === 'Directed'
+    ? (transition.from?.map(from => (
+      typeof from === 'string' ? from : from.id ?? ''
+    )) ?? [])
+      .map(from => statusesMap.get(from) ?? from)
+      .sort()
+      .join(',')
```
alonstern, line 80: same as with `:`, can we use `_` instead of `,`?

packages/jira-adapter/src/filters/workflow/workflow_structure_filter.ts
```diff
@@ -36,6 +36,53 @@ const FETCHED_ONLY_INITIAL_POST_FUNCTION = [
   'IssueStoreFunction',
 ]
 
+const PARTS_SEPARATOR = '::'
+
+const TYPE_TO_FROM_MAP = new Map([
+  ['Initial', 'none'],
+  ['Global', 'any status'],
+  ['Circular', 'any status'],
+])
```
alonstern, line 45: since the keys here is a closed list, you can use an object and have type safety at compile time
```suggestion
const TYPE_TO_FROM_MAP: Record<Exclude<TransitionType, 'directed>>, string> = {
  Initial: 'none',
  Global: 'any status',
  Circular: 'any status'
}
```
Also, you can remove the `?? ''` in https://github.com/salto-io/salto/pull/4657/files#diff-22c6fad8ca1e0ddd5a61c7c55d3f89c87887e3ead79fb03579f367cec76f8208R81 after this

packages/jira-adapter/src/filters/workflow/workflow_structure_filter.ts
```diff
@@ -36,6 +36,53 @@ const FETCHED_ONLY_INITIAL_POST_FUNCTION = [
...
+  const fromSorted = type === 'Directed'
+    ? (transition.from?.map(from => (
+      typeof from === 'string' ? from : from.id ?? ''
+    )) ?? [])
+      .map(from => statusesMap.get(from) ?? from)
+      .sort()
+      .join(',')
+    : TYPE_TO_FROM_MAP.get(type) ?? ''
+
+  return naclCase([transition.name ?? '', `From: ${fromSorted}`, type].join(PARTS_SEPARATOR))
```
alonstern, line 83: `: ` to `_`

packages/jira-adapter/src/filters/workflow/workflow_structure_filter.ts
```diff
@@ -139,29 +186,71 @@ const transformRules = (rules: Rules, transitionType?: string): void => {
...
 }
 
-const transformWorkflowInstance = (workflowValues: Workflow): void => {
+const transformTransitions = (value: Value): SaltoError[] => {
+  const statusesMap = createStatusMap(value.statuses ?? [])
+  const maxCounts: Record<string, number> = {}
+  value.transitions.forEach((transition: Transition) => {
+    const key = getTransitionKey(transition, statusesMap)
+    maxCounts[key] = (maxCounts[key] ?? 0) + 1
+  })
```
alonstern, line 195: suggestion
```suggestion
  const maxCounts = _(value.transitions).map(transition => getTransitionKey(transition, statusesMap)).countBy().value()
```

packages/jira-adapter/src/filters/workflow/workflow_structure_filter.ts
```diff
@@ -139,29 +186,71 @@ const transformRules = (rules: Rules, transitionType?: string): void => {
...
+  workflowValues.transitions.forEach(transformProperties)
   workflowValues.transitions
-    ?.filter(transition => transition.rules !== undefined)
+    .filter(transition => transition.rules !== undefined)
     .forEach(transition => {
       transformRules(transition.rules as Rules, transition.type)
     })
+
+  // The type is changed after transform Transitions, so should be last
+  return transformTransitions(workflowValues)
```
alonstern, line 237: Can we extract this and all the relevant code to a different file? This file already does many things

packages/jira-adapter/src/filters/workflow/workflow_structure_filter.ts
```diff
@@ -139,29 +186,71 @@ const transformRules = (rules: Rules, transitionType?: string): void => {
   rules.validators?.forEach(transformValidator)
 }
 
-const transformWorkflowInstance = (workflowValues: Workflow): void => {
+const transformTransitions = (value: Value): SaltoError[] => {
```
alonstern, line 189: Can this get `WorkflowResponse` instead of `Value`?

alonstern (commented):
👌 🔥

packages/jira-adapter/src/filters/workflow/workflow_structure_filter.ts
IdoZakk: There are also spaces..... I am open to better suggestions, I think Lior asked not to use underscore

packages/jira-adapter/src/filters/workflow/workflow_structure_filter.ts
IdoZakk: The important part is that it will be readable in the UFN, so _ won't be great

packages/jira-adapter/src/filters/workflow/workflow_structure_filter.ts
IdoZakk: It was done in the second PR

packages/jira-adapter/src/filters/workflow/workflow_structure_filter.ts
IdoZakk: As I mentioned in person I cannot as I change the type in the function

alonstern (approved):
🔥  Approving in advance

PR #4659, opened 2023-07-24
SALTO-4239: Support deploying Salesforce Hierarchy CustomFields

tamtamirr:
Support deploying Salesforce Hierarchy CustomFields

---

Two issues upon transformation back to **CustomField** in Salesforce for this type:
1. The **relationshipName** was not send to the service.
2. Unlike other relationship fields (e.g. Lookup and MasterDetail), we can't specify **referenceTo** upon deploy for
Hierarchy fields. See the comment in the new code for more reference.
---
_Release Notes_: 
Salesforce Adapter:
- Support deploying Salesforce Hierarchy CustomFields 

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/constants.ts
```diff
@@ -66,18 +66,6 @@ export enum FIELD_TYPE_NAMES {
   FILE = 'File',
 }
 
-const RELATIONSHIP_FIELD_NAMES = [
```
tamtamirr, line 69: Moved to **types.ts** as this is the only place to consume it.

packages/salesforce-adapter/src/client/types.ts
```diff
@@ -16,7 +16,18 @@
...
+const RELATIONSHIP_FIELD_NAMES = [
+  FIELD_TYPE_NAMES.METADATA_RELATIONSHIP,
+  FIELD_TYPE_NAMES.LOOKUP,
+  FIELD_TYPE_NAMES.MASTER_DETAIL,
+  FIELD_TYPE_NAMES.HIERARCHY,
+]
+
+const isRelationshipFieldName = (fieldName: string): boolean => (
+  (RELATIONSHIP_FIELD_NAMES as ReadonlyArray<string>).includes(fieldName)
+)
```
benisraeladi, line 30: If you removed the as const from RELATIONSHIP_FIELD_NAMES then there are other changes you can do. For example, why not define RELATIONSHIP_FIELD_NAMES as ReadonlyArray instead of casting?

tamtamirr: The casting is not needed anymore, i'll remove it. Thanks :)

benisraeladi (approved):

PR #4660, opened 2023-07-24
SALTO-4525: add alias to dummy adapter

edenhassid:
add alias to dummy adapter

---

added only for instances and custom objects

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/dummy-adapter/src/adapter.ts
+++ b/packages/dummy-adapter/src/adapter.ts
@@ -15,11 +15,25 @@
 */
 import {
   FetchResult, AdapterOperations, DeployResult, FetchOptions,
-  DeployOptions, DeployModifiers, getChangeData, isInstanceElement,
+  DeployOptions, DeployModifiers, getChangeData, isInstanceElement, Element, isObjectType, CORE_ANNOTATIONS,
 } from '@salto-io/adapter-api'
 import _ from 'lodash'
 import { generateElements, GeneratorParams } from './generator'
 import { changeValidator } from './change_validator'
+
+// will add alias only to instances under records and custom objects under objects
+export const addAlias = (element: Element): void => {
+  if (isInstanceElement(element) && element.path && element.path[1] === 'Records') {
+    element.annotations[CORE_ANNOTATIONS.ALIAS] = `${element.elemID.name}_alias`
+  }
+  if (isObjectType(element) && element.path && element.path[1] === 'Objects') {
+    // for when element is split to annotations and fields files
+    if (_.last(element.path)?.endsWith('Fields')) {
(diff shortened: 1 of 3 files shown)


packages/dummy-adapter/src/adapter.ts
```diff
@@ -30,8 +44,10 @@ export default class DummyAdapter implements AdapterOperations {
    * Account credentials were given in the constructor.
    */
   public async fetch({ progressReporter }: FetchOptions): Promise<FetchResult> {
+    const elements = await generateElements(this.genParams, progressReporter)
+    elements.forEach(addAlias)
```
adamro, line 48: Why did you choose do it in retrospect after everything was generated and not put the alias logic inside the generate?
You could add it to for example `generateRecords` in `generate.ts` for records instead of running on all the elements and "guess" it's a record based on the path, no?

edenhassid: I thought about the case that we would want to turn the alias off using some kind of a flag in the config, and that if the addition of the alias will be in retrospect it will be easier.  Do you think it is a requirement?

adamro: I just think it's weird that this is so special in the code. All the other values/annotations on the Elements comes from generate and here we do a post-processing in the main flow (not even a filter like in other adapters). I wouldn't want any new addition to Elements to be pushed up here from now on or have a random decision if it's inside generate or here per new thing.
I think it makes more sense inside generate, yea.
And if we need a config we can pass the adapter config to generate and add logic there.

edenhassid: ok :)

adamro (approved):
🔥

packages/dummy-adapter/test/adapter.test.ts
```diff
@@ -93,6 +101,25 @@ describe('dummy adapter', () => {
         message: 'Generation done',
       })
     })
+    it('should add alias to instances and custom objects', async () => {
+      const mockReporter = { reportProgress: jest.fn() }
+      const fetchResult = await adapter.fetch({ progressReporter: mockReporter })
+      expect(fetchResult.elements.every(elem => {
```
adamro, line 107: nit: If/when this test fails it will be very un-informative for debugging. Cause it will just be "expected: true, got: false" with no hint of what failed.
You could just do filter on the elements and then forEach that has expect for each one with the actual expected alias. Which will lead to a failure like "expected: "myName_alias", got: undefined" which will direct you to the issue.
If you are not worried bout this you can leave it tho :)

PR #4666, opened 2023-07-25
SALTO-4533: State static files are flushed twice for each workspace flush

alonstern:
Fixed files in s3_dir_store being flushed twice on each flush of the workspace

---
_Release Notes_: 
- Fixed files in s3_dir_store being flushed twice on each flush of the workspace

---
_User Notifications_: 
None

### DIFF
--- a/packages/core/src/local-workspace/s3_dir_store.ts
+++ b/packages/core/src/local-workspace/s3_dir_store.ts
@@ -41,7 +41,7 @@ export const buildS3DirectoryStore = (
     concurrencyLimit?: number
   }
 ): staticFiles.StateStaticFilesStore => {
-  const updated: Record<string, dirStore.File<Buffer>> = {}
+  let updated: Record<string, dirStore.File<Buffer>> = {}
   const s3 = S3Client ?? createS3Client()
   const bottleneck = new Bottleneck({ maxConcurrent: concurrencyLimit })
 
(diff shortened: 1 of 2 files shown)


packages/core/src/local-workspace/s3_dir_store.ts
```diff
@@ -134,6 +134,7 @@ export const buildS3DirectoryStore = (
 
   const flush = async (): Promise<void> => {
     await Promise.all(Object.values(updated).map(f => writeFile(f)))
+    updated = {}
```
dantal4, line 137: If a static file was added during the `writeFile`s in the line above, we will ignore it, right?
So maybe do this?
```
const entries = Object.entries(updated)
await Promise.all(entries.map(([key, f]) => {
  writeFile(f)
  delete updated[key]
}))
```

alonstern: nice catch

ori-moisis:
```typescript
const entries = Object.entries(updated)
updated = {}
await Promise.all(entries.map(f => writeFile(f)))
```
once we get a copy of the entires we can clear the updated map.
in the suggestion above (assuming it is fixed to `await writeFile(f)` we allow an async "break" in the code which would allow a similar issue (instead of missing a file, it may write incorrect data in the very unlikely scenario where we keep writing files while flushing

alonstern: Let's make an interview question out of this

alonstern:
@ori-moisis @dantal4 Can we merge this or is there anything left to do?

ori-moisis (approved):

PR #4670, opened 2023-07-26
SALTO-4536 add alias to settings instances & bundles (NS)

dantal4:
---

---
_Release Notes_: 
Netsuite Adapter:
- Add alias to settings instances & bundles

---
_User Notifications_: 
Netsuite Adapter:
- Add alias to settings instances & bundles


omrilit (approved):
Looks good, thanks!

Do we need to inform some customers that are using aliases that they are about to have changes in their next fetch?
Please make sure we don't miss it if needed.

PR #4671, opened 2023-07-26
SALTO-4538: Salesforce Metadata Deploy with Element Values Bugfix

tamtamirr:
Removing un-deployable values of type Element in toDeployableInstance.

---

When a reference lookup is resolved to undefined, we return the Element value as a fallback. This caused some regression in flows where it happens.

Another instance of this we recently solved was https://github.com/salto-io/salto/pull/4585. We should probably revert it soon.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/transformers/transformer.ts
```diff
@@ -1363,14 +1396,19 @@ export const getSObjectFieldElement = (
...
+  const removeNonDeployableValues: TransformFunc = ({ value, field }) => {
+    if (isLocalOnly(field)) {
+      return undefined
+    }
+    if (isElement(value)) {
+      log.warn('The value of the field %s is Element with Id %s in toDeployableInstance', field?.elemID.getFullName(), value.elemID.getFullName())
+      return undefined
+    }
+    return value
+  }
```
tomsellek, line 1408: Some day some body is going to be really confused about why we even have non deployable values in the first place. Can we add a comment or a link to a place where the whole thing is explained (i.e. the workaround that necessitated this function in the first place)?

tomsellek (approved):

amitgoldarb (commented):
I saw the bug happening only on layouts in incidents, yet the UT here is around heigharchy's.
Is this the same underlying issue that is linked to the ticket?

packages/salesforce-adapter/src/transformers/transformer.ts
tamtamirr: Added a comment :)

tamtamirr:
@amitgoldarb Oops It's accidentally rebased on top of the **Hierarchy bugfix** PR. I'll undo. these changes and ping you

packages/salesforce-adapter/src/transformers/transformer.ts
```diff
@@ -1363,14 +1396,21 @@ export const getSObjectFieldElement = (
...
-  const removeLocalOnly: TransformFunc = ({ value, field }) => (
-    (isLocalOnly(field))
-      ? undefined
-      : value
-  )
+  const removeNonDeployableValues: TransformFunc = ({ value, field }) => {
+    if (isLocalOnly(field)) {
+      return undefined
+    }
+    // When we have a reference that resolves to undefined, we return the Element as a PlaceHolder.
```
benisraeladi, line 1403: I don't understand this comment. Where is the reference resolved to undefined?

packages/salesforce-adapter/src/transformers/transformer.ts
```diff
@@ -1363,14 +1396,21 @@ export const getSObjectFieldElement = (
...
+    // This value is not deployable and should not be parsed to XML.
+    if (isElement(value)) {
```
amitgoldarb, line 1405: I'm not sure I understand this. We decide if something is undeployable by the value? Shouldn't the field be "local" in this case? In which case is a field deployable and in another not deployable.

tamtamirr: See https://github.com/salto-io/salto/blob/c507c286bad61258a2544057e051cee876ca138c/packages/salesforce-adapter/src/transformers/reference_mapping.ts#L847

packages/salesforce-adapter/src/transformers/transformer.ts
tamtamirr: See https://github.com/salto-io/salto/blob/c507c286bad61258a2544057e051cee876ca138c/packages/salesforce-adapter/src/transformers/reference_mapping.ts#L847

benisraeladi: I know why this happens, but connecting between these two areas when coming back to the code even a month from now will be very difficult. My point was to explain that this is due to the reference mapper, why it's done this way, and why it needs to be mitigated here

PR #4672, opened 2023-07-27
SALTO-4549 do not fetch personal macros

netama:
After a recent zendesk change, users get personal macros by default, and by default these are less interesting and add more potential conflicts. 
there is an alternative solution where we add the restrictions to the id field instead and still fetch the macros - so will wait with merging until we decide.

TBD whether noise reduction is needed

---
_Release Notes_: 
_Zendesk Adapter_:
* Personal macros will not be fetched by default

---
_User Notifications_: 
_Zendesk Adapter_:
* Personal macros will be removed


edenhassid (approved):
🔥 🔥

PR #4673, opened 2023-07-27
SALTO-4553 - Fix TOCTOU issue when creating workspace state

tomsellek:
Double-check whether workspace state exists before creating it.

---

There is a call to an async function between the time we check whether the workspace state exists and the time we initialize it if it doesn't. 
This `await` can cause Node.js to switch to another context that also called this function, which would also see that the workspace has no state and try to create it.
Best case scenario we crash because we open too many files when we parse the NaCl files in the workspace multiple times concurrently. Worst case scenario we overwrite the state with another state, which may or may not be benign but we'll never know we did it.

---
_Release Notes_: 
Improve stability when loading a workspace

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/workspace/src/workspace/workspace.ts
+++ b/packages/workspace/src/workspace/workspace.ts
@@ -499,11 +499,13 @@ export const loadWorkspace = async (
      * This means you can't put an `await` between the start of `buildWorkspaceState` and the following lines.
      */
     const initBuild = workspaceState === undefined
-    const stateToBuild = workspaceState !== undefined
-      ? await workspaceState
-      : await initState()
 
     const wsChanges = await workspaceChanges
+
+    const stateToBuild = (!initBuild
+      ? await workspaceState
+      : await initState()) as WorkspaceState
+
 
     const getWorkspaceChangesForEnv = (envName: string): ChangeSet<Change> | undefined => wsChanges[envName]
 
@@ -1285,7 +1287,9 @@ export const loadWorkspace = async (
       return loadWorkspace(config, adaptersConfig, credentials, envSources, remoteMapCreator)
     },
     clear: async (args: ClearFlags) => {
+      log.info('Starting workspace.clear()')
       const currentWSState = await getWorkspaceState()
+      log.info('After first getWorkspaceState()')
       if (args.cache || args.nacl || args.staticResources) {
         if (args.staticResources && !(args.state && args.cache && args.nacl)) {
           throw new Error('Cannot clear static resources without clearing the state, cache and nacls')
@@ -1303,7 +1307,9 @@ export const loadWorkspace = async (
         await promises.array.series(envs().map(e => (() => credentials.delete(e))))
       }
       workspaceState = undefined
+      log.info('Before second getWorkspaceState()')
       await getWorkspaceState()
+      log.info('After second getWorkspaceState()')
     },
     addAccount,
     addService: addAccount,
--- a/packages/workspace/src/workspace/workspace.ts
+++ b/packages/workspace/src/workspace/workspace.ts
@@ -368,8 +368,6 @@ export const loadWorkspace = async (
     mergedRecoveryMode
   )
   let workspaceState: Promise<WorkspaceState> | undefined
-
-
   const buildWorkspaceState = async ({
     workspaceChanges,
     stateOnlyChanges = {},
@@ -496,20 +494,27 @@ export const loadWorkspace = async (
      * `getWorkspaceState`). `buildWorkspaceState` will return a promise the first time it executes an `await`.
      * This means you can't put an `await` between the start of `buildWorkspaceState` and the following lines.
      *
-     * Dragon 2: initState() uses naclFilesSource. One can't use naclFilesSource until load() was called on it. Hence
-     * setting wsChanges must happen after we set initBuild and previousState, but before we call initState(). Don't
-     * move the initialization of wsChanges.
+     * Dragon 2: initState() uses naclFilesSource. One can't use naclFilesSource until load() was called on it. Hence,
+     * setting wsChanges must happen after we set previousState, but before we call initState().
+     * Don't move the initialization of wsChanges.
      */
     const previousState = workspaceState
-    const initBuild = previousState === undefined
 
-    const wsChanges = (workspaceChanges !== undefined)
-      ? workspaceChanges
+    if (workspaceChanges === undefined && previousState !== undefined) {
+      // We assume that we only want to get the changes from naclFilesSource if the workspace is 'clean' (i.e. there's
+      // no previous state to override.
+      throw new Error('workspaceChanges can be undefined only if the workspace is clean')
+    }
+
+    const wsChanges = (previousState !== undefined)
+      ? (workspaceChanges ?? {})
       : await naclFilesSource.load({ ignoreFileChanges })
 
-    const stateToBuild = (previousState !== undefined
+    const stateToBuild = (previousState !== undefined)
       ? await previousState
-      : await initState()) as WorkspaceState
+      : await initState()
+
+    const initBuild = previousState === undefined
 
     if (ignoreFileChanges) {
       // Skip all updates to the state since this flag means we are operating under the assumption
@@ -779,7 +784,7 @@ export const loadWorkspace = async (
 
   const getWorkspaceState = async (): Promise<WorkspaceState> => {
     if (_.isUndefined(workspaceState)) {
-      workspaceState = buildWorkspaceState({ })
+      workspaceState = buildWorkspaceState({})
     }
     return workspaceState
   }
--- a/packages/workspace/src/workspace/workspace.ts
+++ b/packages/workspace/src/workspace/workspace.ts
@@ -500,14 +500,8 @@ export const loadWorkspace = async (
      */
     const previousState = workspaceState
 
-    if (workspaceChanges === undefined && previousState !== undefined) {
-      // We assume that we only want to get the changes from naclFilesSource if the workspace is 'clean' (i.e. there's
-      // no previous state to override.
-      throw new Error('workspaceChanges can be undefined only if the workspace is clean')
-    }
-
-    const wsChanges = (previousState !== undefined)
-      ? (workspaceChanges ?? {})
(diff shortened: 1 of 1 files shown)


ori-moisis:
this is not the fix
it will still create the 800 separate workspace states
the `await naclFilesSource.load` needs to move somewhere into `buildWorkspaceState`

packages/workspace/src/workspace/workspace.ts
```diff
@@ -368,15 +368,32 @@ export const loadWorkspace = async (
...
-    validate?: boolean
-  }): Promise<WorkspaceState> => {
+  }: BuildWorkspaceStateArgs): Promise<WorkspaceState> => {
+    let wsChanges: Record<string, ChangeSet<Change>>
+    if (workspaceChanges !== undefined) {
+      wsChanges = workspaceChanges
+    } else if (workspaceFileSource !== undefined) {
+      wsChanges = await workspaceFileSource.load({ ignoreFileChanges })
+    }
+
```
ori-moisis, line 396: if we wanted minimal changes, we can could just allow the workspace changes to be a promise
```typescript
const buildWorkspaceState = async ({
    workspaceChanges = {},
    stateOnlyChanges = {},
    validate = true,
  }: {
    workspaceChanges?: Record<string, ChangeSet<Change>> | Promise<Record<string, ChangeSet<Change>>>
    stateOnlyChanges?: Record<string, ChangeSet<Change>>
    validate?: boolean
  }): Promise<WorkspaceState> => {
  const wsChanges = await workspaceChanges
```
then we just need to change
```typescript
const workspaceChanges = await naclFilesSource.load({ ignoreFileChanges })
```
to
```typescript
const workspaceChanges = naclFilesSource.load({ ignoreFileChanges })
```
and that would be it...


if you want to reduce the repetition in `Record<string, ChangeSet<Change>> | Promise<Record<string, ChangeSet<Change>>>` we can maybe export the `Thenable<T>` type from lowerdash and use it here as
```typescript
workspaceChanges?: Thenable<Record<string, ChangeSet<Change>>>
```

packages/workspace/src/workspace/workspace.ts
```diff
@@ -368,15 +368,19 @@ export const loadWorkspace = async (
     mergedRecoveryMode
   )
   let workspaceState: Promise<WorkspaceState> | undefined
+
+  type WorkspaceChanges = Record<string, ChangeSet<Change>>
+  type BuildWorkspaceStateArgs = {
+    workspaceChanges: Promise<WorkspaceChanges> | WorkspaceChanges
+    stateOnlyChanges?: Record<string, ChangeSet<Change>>
```
ori-moisis, line 375: nit: this is the same type as `WorkspaceChanges`, can use the same type here (maybe rename it to `ChangesByEnv` or something like that?)

packages/workspace/src/workspace/workspace.ts
```diff
@@ -488,11 +492,21 @@ export const loadWorkspace = async (
...
+     * This means you can't put an `await` between the start of `buildWorkspaceState` and the following lines.
+     */
     const initBuild = workspaceState === undefined
     const stateToBuild = workspaceState !== undefined
       ? await workspaceState
       : await initState()
 
+    const wsChanges = await workspaceChanges
+
+    const getWorkspaceChangesForEnv = (envName: string): ChangeSet<Change> | undefined => wsChanges[envName]
```
ori-moisis, line 508: why did you add this function?

packages/workspace/src/workspace/workspace.ts
```diff
@@ -488,11 +492,21 @@ export const loadWorkspace = async (
       return initializedState
     }
 
+    /** HERE BE DRAGONS!
+     *
+     * `workspaceState` will be `undefined` until `buildWorkspaceState` returns a promise (see the code in
+     * `getWorkspaceState`). `buildWorkspaceState` will return a promise the first time it executes an `await`.
+     * This means you can't put an `await` between the start of `buildWorkspaceState` and the following lines.
```
ori-moisis, line 499: instead of doing this, it would be easier and less error prone to just pass this in as an argument if this is the initial state or not

packages/workspace/src/workspace/workspace.ts
```diff
@@ -760,13 +774,18 @@ export const loadWorkspace = async (
   }
 
   const getWorkspaceState = async (): Promise<WorkspaceState> => {
+    const wsConfig = await config.getWorkspaceConfig()
```
ori-moisis, line 777: do we still need to do this?
feels like a lot of effort for just a log that doesn't really need this info...

packages/workspace/src/workspace/workspace.ts
```diff
@@ -368,8 +368,10 @@ export const loadWorkspace = async (
     mergedRecoveryMode
   )
   let workspaceState: Promise<WorkspaceState> | undefined
+
+
```
ori-moisis, line 372: no need to add these?

packages/workspace/src/workspace/workspace.ts
```diff
@@ -488,10 +490,26 @@ export const loadWorkspace = async (
...
+     * Dragon 1: `workspaceState` will be `undefined` until `buildWorkspaceState` returns a promise (see the code in
+     * `getWorkspaceState`). `buildWorkspaceState` will return a promise the first time it executes an `await`.
+     * This means you can't put an `await` between the start of `buildWorkspaceState` and the following lines.
+     *
+     * Dragon 2: initState() uses naclFilesSource. One can't use naclFilesSource until load() was called on it. Hence
+     * setting wsChanges must happen after we set initBuild and previousState, but before we call initState(). Don't
+     * move the initialization of wsChanges.
+     */
+    const previousState = workspaceState
+    const initBuild = previousState === undefined
```
ori-moisis, line 504: I wouldn't mix in `initBuild` into this mess, it is irrelevant once we set `previousState`
I think we should move it back to where it was and not mention it in the comment

packages/workspace/src/workspace/workspace.ts
```diff
@@ -488,10 +490,26 @@ export const loadWorkspace = async (
...
+
+    const wsChanges = (workspaceChanges !== undefined)
+      ? workspaceChanges
+      : await naclFilesSource.load({ ignoreFileChanges })
```
ori-moisis, line 508: why change the logic here?
we only want to load underlying changes if this is a "clean" load of the workspace, this has been defined so far by the previous state not existing.
I think it would be a mistake to change that logic and make it depend on whether or not we passed in changes

packages/workspace/src/workspace/workspace.ts
```diff
@@ -488,10 +490,26 @@ export const loadWorkspace = async (
...
+
+    const stateToBuild = (previousState !== undefined
+      ? await previousState
+      : await initState()) as WorkspaceState
```
ori-moisis, line 512: we should not have to cast here

packages/workspace/src/workspace/workspace.ts
```diff
@@ -761,12 +779,7 @@ export const loadWorkspace = async (
 
   const getWorkspaceState = async (): Promise<WorkspaceState> => {
     if (_.isUndefined(workspaceState)) {
-      const wsConfig = await config.getWorkspaceConfig()
-      log.debug('No workspace state for %s/%s. Building new workspace state.', wsConfig.uid, wsConfig.name)
-      const workspaceChanges = await naclFilesSource.load({ ignoreFileChanges })
-      workspaceState = buildWorkspaceState({
-        workspaceChanges,
-      })
+      workspaceState = buildWorkspaceState({ })
```
ori-moisis, line 782:
```suggestion
      workspaceState = buildWorkspaceState({})
```

packages/workspace/src/workspace/workspace.ts
tomsellek: Ack

packages/workspace/src/workspace/workspace.ts
tomsellek: Ack.

packages/workspace/src/workspace/workspace.ts
tomsellek: I get what you're saying, and I will make the change, but I think it would make things even less obvious - why is the way we get the workspace changes related to whether or not there's a previous state? What happens if `changes === undefined` but there's an existing state? etc...
At the very least I want to add a check for when `changes` is `undefined` and a previous state exists.

packages/workspace/src/workspace/workspace.ts
tomsellek: Ack.

packages/workspace/src/workspace/workspace.ts
tomsellek: Ack.

packages/workspace/src/workspace/workspace.ts
```diff
@@ -488,11 +488,34 @@ export const loadWorkspace = async (
...
+      throw new Error('workspaceChanges can be undefined only if the workspace is clean')
+    }
+
+    const wsChanges = (previousState !== undefined)
+      ? (workspaceChanges ?? {})
+      : await naclFilesSource.load({ ignoreFileChanges })
+
+    const stateToBuild = (previousState !== undefined)
+      ? await previousState
       : await initState()
```
tomsellek, line 515: @ori-moisis would it make sense to change these lines to one big if-else where we handle the different cases of "clean workspace" vs. "dirty workspace"? I think it would require changing wsChanges and stateToBuild from `const` to `let`.

ori-moisis: I think the previous code was easier, lets go back to it

tomsellek: @ori-moisis sorry, what previous code?

ori-moisis: https://github.com/salto-io/salto/pull/4673/commits/215a916734e29002e743f899bcbd237bda79bf38

tomsellek: Gotcha

ori-moisis (approved):

PR #4674, opened 2023-07-27
SALTO-4554: exclude built-in names from automations' account id filter

IdoZakk:
_Fixed a bug where automations could not be deployed due to a certain usage of a built-in user type_
[Noise reduction](https://github.com/salto-io/salto_private/pull/6332)

---

To recreate add a create issue action for an automation. 
In 'Choose fields to set' select assignee or reporter.
At the bottom of the page find the field, on the right side press the ... and change to `copy` 
Fetch, make a copy and try to deploy.
Note- the custom fields with users does not use the 'ID' type, they use 'name' so they are no a concern here

---
_Release Notes_: 
Jira Adapter: 
* Fixed a bug that prevented the deployment of automations in specific conditions

---
_User Notifications_: 
Jira Adapter:
* On very specific cases there will be a change in autmations of the structure of assignee or reporter fields

### DIFF
--- a/packages/jira-adapter/src/filters/account_id/account_id_filter.ts
+++ b/packages/jira-adapter/src/filters/account_id/account_id_filter.ts
@@ -84,9 +84,14 @@ const callbackValueOrValues = (
   }
 }
 
-const walkOnAutomationValue = (regexPath: string, callback: WalkOnUsersCallback)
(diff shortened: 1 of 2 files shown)


alonstern (commented):
Can you add a test for that?

alonstern (approved):
👌

PR #4677, opened 2023-07-30
SALTO-4518: add logs for jira and zendesk to indicate of elem id discrepancy

edenhassid:
add logs to indicate of elem id discrepancy

---

---
_Release Notes_: 

---
_User Notifications_: 


packages/adapter-utils/index.ts
```diff
@@ -26,3 +26,4 @@ export * from './src/template'
 export * from './src/walk_element'
 export * from './src/collisions'
 export * as references from './src/references'
+export * from './src/elemId_discrepancy'
```
alonstern, line 29: nit: `elem_id_discrepancy`

packages/adapter-utils/src/elemId_discrepancy.ts
```diff
@@ -0,0 +1,46 @@
...
+* limitations under the License.
+*/
+import { ElemID, ElemIdGetter, ServiceIds } from '@salto-io/adapter-api'
+import { logger } from '@salto-io/logging'
+
+
+const log = logger(module)
+
+export const getElemIdFuncWrapper = (func: ElemIdGetter): {getElemIdFunc: ElemIdGetter; logIdsFunc: () => void} => {
+  const nonMatchingIdsMap: Record<string, string> = {}
```
alonstern, line 23: This probably should be a `new Map`. Using an object here might cause unexpected behavior if we pass it string like `toString`, `valueOf`, etc... (it will return a function instead of returning `string | undefined`)

packages/adapter-utils/src/elemId_discrepancy.ts
```diff
@@ -0,0 +1,46 @@
...
+  const logIdsFunc = (): void => {
+    if (Object.keys(nonMatchingIdsMap).length === 0) {
+      return
+    }
+    const diffsListString = Object.keys(nonMatchingIdsMap).map(current => {
+      const calculated = nonMatchingIdsMap[current]
+      return `current id: ${current} --- calculated id: ${calculated}`
+    })
+    log.warn(`The following elements have differences between current elemId and calculated elemId:\n${diffsListString.join('\n')}`)
```
alonstern, line 32: Let's maybe limit it to the first 100?

alonstern (approved):
👌

PR #4678, opened 2023-07-30
SALTO-4486 deal with case where internal ID is an empty string, and d…

benisraeladi:
…on't add it to nacl

---

This change is in retrieve API

---
_Release Notes_: 
Salesforce:
- Fixed some invalid go-to-service URLs that were caused by API regression

---
_User Notifications_: _None_


tamtamirr (approved):
Approved with small comment to save cycles 👑 Also, I'd rephrase the release notes to be less broad, considering there are still known gaps in **Go To Service**:
`Fixed some Go To Service URLs that lead to invalid URLs due to regression in the API`

PR #4679, opened 2023-07-30
SALTO-4560: Salesforce data deploy of existing instances bug-fix

tamtamirr:
fixed Data deploy bug when added instance is marked for update.

---

This flow occurs when we recognize an Instance that already exists in the target org with the same **saltoIdFields** values.
In that case, the adapter will attempt to **update** the Record insert of **inserting** it, which will require us to pass the mandatory field **OwnerId** along side the mandatory **Id** field.
 
---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/adapter.ts
```diff
@@ -457,7 +458,7 @@ export default class SalesforceAdapter implements AdapterOperations {
     { changeGroup }: DeployOptions,
     checkOnly: boolean
   ): Promise<DeployResult> {
-    log.debug(`about to ${checkOnly ? 'validate' : 'deploy'} group ${changeGroup.groupID} with scope (first 100): ${safeJsonStringify(changeGroup.changes.slice(100).map(getChangeData).map(e => e.elemID.getFullName()))}`)
+    log.debug(`about to ${checkOnly ? 'validate' : 'deploy'} group ${changeGroup.groupID} with scope (first 100): ${safeJsonStringify(changeGroup.changes.slice(0, 100).map(getChangeData).map(e => e.elemID.getFullName()))}`)
```
tamtamirr, line 461: This fix for the log is irrelevant for the PR but I wanna squeeze it in, feel free to make me do it in a different branch 😿

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -155,15 +157,15 @@ const getRecordsBySaltoIds = async (
     return r
   }))
 
-  // Should always query Id together with the SaltoIdFields to match it to instances
-  const saltoIdFieldsWithIdField = (saltoIdFields
-    .find(field => field.name === CUSTOM_OBJECT_ID_FIELD) === undefined)
-    ? [type.fields[CUSTOM_OBJECT_ID_FIELD], ...saltoIdFields] : saltoIdFields
-
-  const fieldNames = await awu(saltoIdFieldsWithIdField).flatMap(getFieldNamesForQuery).toArray()
```
tamtamirr, line 163: This logic was excessive and redundant. Didn't wanna copy it for **OwnerId** as-well as we only care about the **fieldNames** at the end.

tamtamirr (commented):
Added some small comments for clarification 🤓

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -378,9 +380,12 @@ const deployAddInstances = async (
     client.dataRetry.maxAttempts
   )
   existingInstances.forEach(instance => {
-    instance.value[
-      CUSTOM_OBJECT_ID_FIELD
-    ] = existingRecordsLookup[computeSaltoIdHash(instance.value)][CUSTOM_OBJECT_ID_FIELD]
+    const salesforceRecordLookup = existingRecordsLookup[computeSaltoIdHash(instance.value)]
```
adamro, line 386: I think the name of the const here just adds confusion

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -122,6 +122,8 @@ const isCompoundFieldType = (type: TypeElement): type is ObjectType => (
   && Object.values(Types.compoundDataTypes).some(compoundType => compoundType.isEqual(type))
 )
 
+const MANDATORY_FIELDS_FOR_UPDATE = [CUSTOM_OBJECT_ID_FIELD, OWNER_ID]
```
adamro, line 125: Is `OwnerId` actually mandatory on all CustomObjects?

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -378,9 +380,12 @@ const deployAddInstances = async (
...
   )
   existingInstances.forEach(instance => {
-    instance.value[
-      CUSTOM_OBJECT_ID_FIELD
-    ] = existingRecordsLookup[computeSaltoIdHash(instance.value)][CUSTOM_OBJECT_ID_FIELD]
+    const salesforceRecordLookup = existingRecordsLookup[computeSaltoIdHash(instance.value)]
+    MANDATORY_FIELDS_FOR_UPDATE.forEach(mandatoryField => {
+      if (instance.value[mandatoryField] === undefined) {
+        instance.value[mandatoryField] = salesforceRecordLookup[mandatoryField]
+      }
```
benisraeladi: The code used to always do this for ID fields, and it does it only if the value is undefined. Are there any cases where the ID field in the instance is different from that in `salesforceRecordLookup`?

adamro (approved):
I am approving this as a workaround but -
1. Are we sure OwnerId is actually mandatory in all CustomObjects?
2. An alternative solution to this can be to just remove OwnerId from being hidden as we planned, no?

Also, this is not the first time the code reached this flow of update instead of insert and we never encountered this before. Is this a regression? Do we understand what's the cause?

benisraeladi (commented):
Overall looks great! One question below

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
tamtamirr: **CustomObjects** yes, **CustomSettings** no. I've commited a fix with comment explaining it. This was caught by our E2Es actually 🤯

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
tamtamirr: Since this is flow of **addition of Instance**, basically any hidden value upon clone will be **undefined** here.
Alternatively, if the user specified any of these values explicitly (For example edit before deploy) we should prefer his value. Moreover, when we choose to implement [this](https://salto-io.atlassian.net/browse/SALTO-4442) ticket we may cause a bug, if we prefer the value from the Service and not the nacl.

tamtamirr:
I've discussed F2F with @adamro re his questions and we are aligned on this workaround for now.

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
tamtamirr: I've tested the second edge-case here, and indeed the **OwnerId** field does not exist on the **CustomObject** in case it's the detail side of a master-detail relationship. See my [Test2__c](https://github.com/salto-io/tamir-sf/blob/main/envs/env1/salesforce/Objects/Test2__c/Test2__cStandardFields.nacl) Object for example.

PR #4680, opened 2023-07-30
SALTO-4481: omit fields in preDeploy

Shulik95:
_omit_fields filter now removes fields on preDeploy according to hard coded fields and fields from config_

---

_This pr covers two tickets_:
* https://salto-io.atlassian.net/browse/SALTO-2781
* https://salto-io.atlassian.net/browse/SALTO-4481

---
_Release Notes_: 
_Netsuite Adapter_:
* 'classTranslationList' field in data elements and 'currency' field in inventory items will be omitted in preDeploy
* fieldToOmit added to config.deploy similar to the existing logic in config.fetch

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/netsuite-adapter/src/change_validators/omit_fields.ts
+++ b/packages/netsuite-adapter/src/change_validators/omit_fields.ts
@@ -14,7 +14,7 @@
 * limitations under the License.
 */
 
-import { Change, ChangeDataType, ChangeError, getChangeData, isModificationChange, Values } from '@salto-io/adapter-api'
+import { Change, ChangeDataType, ChangeError, getChangeData, isAdditionOrModificationChange, isModificationChange, Values } from '@salto-io/adapter-api'
 import { values } from '@salto-io/lowerdash'
 import { NetsuiteChangeValidator } from './types'
 import { FIELDS_TO_OMIT_PRE_DEPLOY, getFieldsToOmitByType } from '../filters/omit_fields'
@@ -40,7 +40,7 @@ const getChangeError = (
     return {
       elemID: element.elemID,
       severity: 'Error',
-      message: 'This element will be removed from deployment',
+      message: 'This element contains an undeployable change',
       detailedMessage: `This element will be removed from deployment because it only contains changes to the undeployable field '${fieldsToOmitByType[element.elemID.typeName].join(', ')}'.`,
     }
   }
@@ -58,6 +58,7 @@ const changeValidator: NetsuiteChangeValidator = async changes => {
   const fieldsToOmitByType = getFieldsToOmitByType(typeNames, FIELDS_TO_OMIT_PRE_DEPLOY)
 
   return changes
+    .filter(isAdditionOrModificationChange)
     .map(change => {
       const element = getChangeData(change)
       const { typeName } = element.elemID
--- a/packages/netsuite-adapter/src/change_validators/omit_fields.ts
+++ b/packages/netsuite-adapter/src/change_validators/omit_fields.ts
@@ -1,0 +1,73 @@
+/*
+*                      Copyright 2023 Salto Labs Ltd.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+import { Change, ChangeDataType, ChangeError, getChangeData, isModificationChange, Values } from '@salto-io/adapter-api'
+import { values } from '@salto-io/lowerdash'
+import { NetsuiteChangeValidator } from './types'
+import { FIELDS_TO_OMIT_PRE_DEPLOY, getFieldsToOmitByType } from '../filters/omit_fields'
+import { getDifferentKeys } from '../filters/data_instances_diff'
+import { getElementValueOrAnnotations } from '../types'
+
+const { isDefined } = values
+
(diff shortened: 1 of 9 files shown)


packages/netsuite-adapter/src/change_validators/omit_fields.ts
```diff
@@ -0,0 +1,74 @@
...
+  typeName: string,
+  fieldsToOmitByType: Record<string, string[]>
+): string[] =>
+  fieldsToOmitByType[typeName]
+    .filter(fieldToOmit => fieldToOmit in value)
+
+const getChangeError = (
+  change: Change,
+  fieldsToOmitByType: Record<string, string[]>,
+  element: ChangeDataType,
```
dantal4, line 37: passing `change` and `element` is redundant

packages/netsuite-adapter/src/change_validators/omit_fields.ts
```diff
@@ -0,0 +1,74 @@
...
+  const fieldsToOmitByType = getFieldsToOmitByType(typeNames, FIELDS_TO_OMIT_PRE_DEPLOY)
+
+  return changes
+    .filter(isAdditionOrModificationChange)
+    .map(change => {
+      const element = getChangeData(change)
+      const { typeName } = element.elemID
+      if (typeName in fieldsToOmitByType
+        && getFieldsToOmit(getElementValueOrAnnotations(element), typeName, fieldsToOmitByType).length > 0) {
+        return getChangeError(change, fieldsToOmitByType, element)
```
dantal4, line 67: nit- this is a little bit more readable IMO
```suggestion
      const elementValues = getElementValueOrAnnotations(element)
      if (fieldsToOmitByType[typeName]?.some(fieldToOmit => elementValues[fieldToOmit] !== undefined)) {
        return getChangeError(change, fieldsToOmitByType[typeName])
```

packages/netsuite-adapter/src/change_validators/omit_fields.ts
```diff
@@ -0,0 +1,74 @@
...
+): ChangeError => {
+  if (isModificationChange(change) && getDifferentKeys(change).size === 1) {
+    return {
+      elemID: element.elemID,
+      severity: 'Error',
+      message: 'This element contains an undeployable change',
+      detailedMessage: `This element will be removed from deployment because it only contains changes to the undeployable field '${fieldsToOmitByType[element.elemID.typeName].join(', ')}'.`,
+    }
+  }
+  const wrappedFieldNames = fieldsToOmitByType[element.elemID.typeName].map(field => `'${field}'`).join(', ')
```
dantal4, line 47: Is it possible to take only the modified fields that we're going to omit?

packages/netsuite-adapter/src/change_validators/omit_fields.ts
```diff
@@ -0,0 +1,74 @@
...
+  return {
+    elemID: element.elemID,
+    severity: 'Warning',
+    message: `This element will be deployed without the following fields: ${wrappedFieldNames}`,
+    detailedMessage: `This element will be deployed without the following fields: ${wrappedFieldNames}, as NetSuite does not support deploying them.`,
+  }
+}
+
+const changeValidator: NetsuiteChangeValidator = async changes => {
+  const typeNames = FIELDS_TO_OMIT_PRE_DEPLOY.map(fieldToOmit => fieldToOmit.type)
```
dantal4, line 57: can we also take the user configured fieldToOmit ?

packages/netsuite-adapter/src/filters/omit_fields.ts
```diff
@@ -14,96 +14,140 @@
...
-      if (!isObjectType(fieldType)) {
-        return value
-      }
-      return omitValues(value, fieldType.elemID.name)
+    const typesForDeepTransformation = getTypesForDeepTransformation(typeNames, fieldsToOmit)
+
+    await omitFieldsFromElements(elements, fieldsToOmitByType, typesForDeepTransformation)
+  },
+  preDeploy: async changes => {
+    const elements = changes.map(getChangeData)
```
dantal4, line 135: In modifications we should go over the before & after (otherwise we will get wrong fields to the null list - `data_instances_null_fields.ts`). We can do the following:
```suggestion
    const elements = changes.filter(isAdditionOrModificationChange).flatMap(change => Object.values(change.data))
```

packages/netsuite-adapter/src/filters/omit_fields.ts
```diff
@@ -14,96 +14,140 @@
...
-        if (typesForDeepTransformation.has(instance.elemID.typeName)) {
-          instance.value = await transformValues({
-            values: instance.value,
-            type: await instance.getType(),
-            transformFunc,
-            strict: false,
-          }) ?? {}
-        }
-      })
+    const typeNames = await awu(await elementsSource.list()).map(elemId => elemId.name).toArray()
```
dantal4, line 142: we should filter only `elemId.idType === 'type'`

packages/netsuite-adapter/test/change_validators/omit_fields.test.ts
```diff
@@ -0,0 +1,68 @@
...
+      },
+    })
+    instance = new InstanceElement('test', type, {
+      currency: 'Shekel',
+      field2: true,
+    })
+    after = instance.clone()
+  })
+  it('should have warning on elements that contain fields that will be omitted', async () => {
+    const changeError = await omitFieldsValidation([toChange({ after: instance })])
```
dantal4, line 41: can you add a test for a modification that returns a warning?

dantal4 (commented):
Great work!

dantal4:
A suggestion for the flow-
1. Clone after
2. Run omitFields filter on the cloned after
3. Run recursively on after and get all the keys that are undefined in the cloned after- if there are non- return no error
4. Clone before and run omitFields filter on it
5. Compare the cloned before and after- if they are different (means that other fields were modified) return a warning with the list from step 3. If they are equal- return an error.

dantal4 (commented):
Great work!

dantal4 (approved):
💯

PR #4683, opened 2023-07-30
SALTO-4450 - Zendesk Adapter - Prevent addition of duplicate routing attribute value

idoser96:
Prevent deployment of routing attribute value with the same name as another value under that routing attribute

---

None

---
_Release Notes_: 
Zendesk Adapter:
* Prevent deployment of addition of duplicate routing attribute value

---
_User Notifications_: 
None


edenhassid (commented):
looks good! there is just this case where we add two values with the same name, and a test for that

edenhassid (commented):
can you add a test to check this case?

edenhassid (approved):
looks great! 💯 🌻

PR #4692, opened 2023-08-01
SALTO-4566 - Zendesk Adapter - Order view's custom_fields by id

idoser96:
Order view's order by id, to keep a consistent order between environments

---

custom_fields order is irrelevant and cannot be changed in the service, the real order is defined by 'columns' field
It is possible to have two custom_fields with the same name and type, but this will also cause a Salto conflict

---
_Release Notes_: 
Zendesk Adapter:
* order of custom_fields in views is now by title and type fields

---
_User Notifications_: 
None


packages/zendesk-adapter/src/filters/unordered_lists.ts
```diff
@@ -200,6 +201,17 @@ const orderFormCondition = (instances: InstanceElement[]): void => {
   sortChildFields(formInstances, ticketFieldById)
 }
 
+// The order is irrelevant and cannot be changed
+// We need to make it constant between environments
+const orderViewCustomFields = (instances: InstanceElement[]): void => {
+  instances.filter(e => e.elemID.typeName === VIEW_TYPE_NAME).forEach(view => {
+    const customFields = view.value.execution?.custom_fields
+    if (isArray(customFields)) {
```
edenhassid, line 209: nit: why not use _.isArray like this?

edenhassid: can you add a log.debug if it is not an array?

edenhassid (approved):
looks great! added some very small comments, pre-approving 🤩

PR #4693, opened 2023-08-01
SALTO-4517: Pass ResolvedTypesElementsSource to the adapters from core.

tamtamirr:
passing **ResolvedTypesElementsSource** to the adapters from core. 

---

Taking advantage of the **expressions.resolve** logic exported from Workspace, and some tweaks and added options to this method, adapters will receive elements source with types that are resolved recursively.

Motives for this PR:
1. This is a blocker for **fetchWithChangesDetection** in Salesforce. (https://salto-io.atlassian.net/browse/SALTO-3985)
2. This will allow eliminating a lot of async code in the adapters & core, assuming all the Elements and types are resolved recursively if we received them through the **ElementsSource**.
3. In Salesforce, we already [use additional implementation](https://github.com/salto-io/salto/blob/c49366eb500865b033792e3af56234c59fe90359/packages/salesforce-adapter/src/change_validator.ts#L112) of ElementsSource with **resolveTypeShallow**. We can eliminate that and use this elements source instead. (And also remove the other implementation)

# Fetch Times on my Salesforce Account (ms):
  - Regular Fetch: 123897
   - Partial Fetch (Role): 17427 
   
With old elementsSource  :

- Regular Fetch: 116839
- Partial Fetch (Role): 17549

From CPU profiling & Heap Snapshots I didn't notice any noticeable spikes in execution times and memory.

---

_Release Notes_: 
_None_

---

_User Notifications_: 
_None_


packages/workspace/test/core/expressions.test.ts
```diff
@@ -165,303 +165,321 @@ describe('Test Salto Expressions', () => {
...
-      const element = findResolved<InstanceElement>(falsyVarRefInst)
-      expect(isVariableExpression(element.value.test)).toBe(true)
-      expect(element.value.test.value).toEqual(false)
-    })
-
-    it('should resolve a variable value which is a reference', () => {
-      const element = findResolved<Variable>(varRef)
-      expect(isReferenceExpression(element.value)).toBe(true)
-      expect(element.value.value).toEqual('simple')
+    describe('when shouldResolveReferences is false', () => {
```
tamtamirr, line 173: This is a new test

packages/core/test/core/adapters/adapters.test.ts
```diff
@@ -283,4 +303,98 @@ describe('adapters.ts', () => {
...
+            await elementsSource.get(type.elemID)
+            expect(jest.isMockFunction(expressions.resolve)).toBeTrue()
+            expect(resolveSpy).toHaveBeenCalledOnce()
+          })
+        })
+      })
+      describe('when element is InstanceElement', () => {
+        it('should return instance with resolved type', async () => {
+          const resolvedInstance = await elementsSource.get(instance.elemID) as InstanceElement
+          expect(resolvedInstance.refType.type).toBeInstanceOf(ObjectType)
```
ori-moisis, line 360: this only checks for shallow resolve, we should check all types are resolved deeply

packages/core/test/core/adapters/adapters.test.ts
```diff
@@ -283,4 +303,98 @@ describe('adapters.ts', () => {
...
+          expect(resolveSpy).toHaveBeenCalledOnce()
```
ori-moisis, line 361: I don't think this is part of the interface, why is it part of the test?

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -412,7 +414,8 @@ export default class SalesforceAdapter implements AdapterOperations {
    */
   @logDuration('fetching account configuration')
   async fetch({ progressReporter }: FetchOptions): Promise<FetchResult> {
-    log.debug('going to fetch salesforce account configuration..')
+    const allElements = await awu(await this.elementsSource.getAll()).toArray()
+    log.debug('going to fetch salesforce account configuration.. %d', allElements.length)
```
ori-moisis, line 418: should not be here?
all of the changes in this file actually?

packages/workspace/test/core/expressions.test.ts
```diff
@@ -165,303 +165,321 @@ describe('Test Salto Expressions', () => {
 
     let resolved: Element[]
 
-    beforeAll(async () => {
```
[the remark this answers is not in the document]
tamtamirr: Actually no noticeable effect. Tests ran fast

packages/salesforce-adapter/src/adapter.ts
tamtamirr: Yes oops, my bad. This was for testing purposes will undo

packages/workspace/src/expressions.ts
```diff
@@ -279,12 +279,23 @@ const getClonedElements = (elements: Element[]): Element[] => {
...
+  shouldCloneElements?: boolean
+  shouldResolveReferences?: boolean
+}
+
 export const resolve = (
   elements: Element[],
-  elementsSource: ReadOnlyElementsSource
+  elementsSource: ReadOnlyElementsSource,
+  {
+    shouldCloneElements = true,
```
ori-moisis, line 291: this option is not needed anymore?

packages/core/src/core/adapters/adapters.ts
```diff
@@ -181,6 +193,76 @@ const filterElementsSource = (
...
+    }
+    // TypeElements
+    const alreadyResolvedType = resolvedTypes.get(id.getFullName())
+    if (alreadyResolvedType !== undefined) {
+      return alreadyResolvedType
+    }
+    const value = await elementsSource.get(id)
+    // Instances & Fields
+    if (isInstanceElement(value) || isField(value)) {
+      const element = isField(value) ? value : value.clone()
```
ori-moisis, line 228: if we got here with a field we need to go get it from the type that we already resolved.
this is not handling fields correctly...

the easiest way to make this work is to separate getting the top level from resolving whatever is in the top level.
you can look at how `RemoteElementSource` does it

packages/core/src/core/adapters/adapters.ts
```diff
@@ -181,6 +193,76 @@ const filterElementsSource = (
...
+      }
+    )
+    resolved.filter(isType)
+      .forEach(typeElement => {
+        resolvedTypes.set(typeElement.elemID.getFullName(), typeElement)
+      })
+  }
+
+  const getResolved = async (id: ElemID): Promise<Element> => {
+    if (resolvedTypes.size === 0) {
```
ori-moisis, line 217: won't this break on a first fetch? (i.e - it will repeatedly try to get all elements)
I think we need a better condition here (e.g. have an explicit boolean of whether we initialized for example)

packages/core/src/core/adapters/adapters.ts
```diff
@@ -181,6 +193,76 @@ const filterElementsSource = (
...
+      if (elementResolvedValue === undefined) {
+        log.warn('Expected type of Element %s to be resolved. Type elemID: %s. Returning Element with non fully resolved type.', element.elemID.getFullName(), element.refType.elemID.getFullName())
+        return element
+      }
+      // If the type of the Element is resolved, simply set the type on the instance
+      element.refType.type = elementResolvedValue
+      return element
+    }
+    // Container types
+    if (isContainerType(value)) {
```
ori-moisis, line 240: this doesn't handle nested contained types (e.g - map of map)

we have some utility functions that can help with this, maybe check how the code in expressions.resolve handles this

packages/core/src/core/adapters/adapters.ts
```diff
@@ -181,6 +193,76 @@ const filterElementsSource = (
...
+  return {
+    get: getResolved,
+    getAll: async () => {
+      if (resolvedTypes.size === 0) {
+        await resolveTypes()
+      }
+      return awu(await elementsSource.list())
+        .filter(id => id.idType !== 'type')
+        .map(getResolved)
+        .concat(resolvedTypes.values())
```
ori-moisis, line 259:
```suggestion
      return awu(await elementsSource.list()).map(getResolved)
```

tamtamirr:
@ori-moisis Ran tests on fetch of NS account as we discussed (since NS invokes **getAll** on the **elementsSource** every fetch). Times were almost identical

1. Second fetch using the current version in main:
`2023-08-15T08:42:40.169Z info cli/cli ran "fetch -v" in 67095 ms`
2. Third fetch using the new elements source:
`2023-08-15T08:46:27.274Z info cli/cli ran "fetch -v" in 68611 ms`

No elements were changed in the workspace using the new **elementsSource**

packages/core/src/core/adapters/adapters.ts
```diff
@@ -214,36 +222,45 @@ export const createResolvedTypesElementsSource = (
   }
 
   const getResolved = async (id: ElemID): Promise<Element> => {
-    if (resolvedTypes.size === 0) {
+    if (!typesWereResolved) {
       await resolveTypes()
     }
```
tomsellek, line 227: Ever since https://github.com/salto-io/salto/pull/4673 I am... wary of mixing these kinds of checks with async functions. Is there a less fragile way to achieve the same result?

tomsellek (commented):
I can't claim to fully understand the repercussions here, but are we sure loading all data up front won't cause memory consumption issues on large workspaces?

packages/core/src/core/adapters/adapters.ts
```diff
@@ -181,6 +195,91 @@ const filterElementsSource = (
...
+export const createResolvedTypesElementsSource = (
+  elementsSource: ReadOnlyElementsSource
+): ReadOnlyElementsSource => {
+  const resolvedTypes: Map<string, TypeElement> = new Map()
+  let typesWereResolved = false
+  const resolveTypes = async (): Promise<void> => {
+    typesWereResolved = true
+    const typeElements = await awu(await elementsSource.getAll()).filter(isType).toArray()
+    // Relevant for first fetch
+    if (typeElements === []) {
```
ori-moisis, line 207: this condition is always false, no?
```
➜  salto_repo git:(main) node
Welcome to Node.js v14.15.3.
Type ".help" for more information.
> x = []
[]
> x === []
false
```

probably doesn't matter given that `expressions.resolve` should handle an empty input if it gets there, we can probably delete this `if`

packages/core/src/core/adapters/adapters.ts
```diff
@@ -181,6 +195,91 @@ const filterElementsSource = (
...
+    // Fields
+    if (isField(value)) {
+      const { parent } = id.createTopLevelParentID()
+      const topLevel = resolvedTypes.get(parent.getFullName())
+      if (topLevel === undefined) {
+        log.warn('Expected parent of Field %s to be resolved. Returning field with non fully resolved type.', id.getFullName())
+        return value
+      }
+      return resolvePath(topLevel, id)
+    }
```
ori-moisis, line 251: could be more efficient if we check the ID first (we don't need to go the to the element source at all when requesting anything inside a field or a type)

packages/core/src/core/adapters/adapters.ts
```diff
@@ -181,6 +195,91 @@ const filterElementsSource = (
...
+    // Instances
+    if (isInstanceElement(value)) {
+      const instance = value.clone()
+      const elementResolvedValue = resolvedTypes.get(instance.refType.elemID.getFullName())
```
ori-moisis, line 255: this name doesn't really say anything
can we call this something more meaningful?
like, this is the type, right? so, maybe call this `instanceType` or `resolvedInstanceType` or just `resolvedType`?

packages/core/src/core/adapters/adapters.ts
ori-moisis: this comment still applies, we don't need to make this code more complicated that it has to be
there is no reason for it to be aware of the implementation details of `getResolved`

packages/core/src/core/adapters/adapters.ts
```diff
@@ -181,6 +195,91 @@ const filterElementsSource = (
...
+      // If the type of the Element is resolved, simply set the type on the instance
+      instance.refType.type = elementResolvedValue
+      return instance
+    }
+    return value
+  }
+  return {
+    get: getResolved,
+    getAll: async () => {
+      if (resolvedTypes.size === 0) {
```
ori-moisis, line 270: this is checking the wrong condition
also, not sure it needs to check any condition, everything will be handled in `getResolved` anyway, no?

packages/workspace/test/core/expressions.test.ts
```diff
@@ -661,20 +679,25 @@ describe('Test Salto Expressions', () => {
...
         const mapInner = (resolved[0].fields.mapObj.refType.type as MapType).refInnerType.type
         expect(listInner).toBe(mapInner)
       })
-      it('should handle multi-level containers', () => {
+      it('should handle multi-level containers', async () => {
+        resolved = await resolve(
+          [outerObjType],
+          createInMemoryElementSource([innerObjType]),
+        ) as typeof resolved
+        resolvedInnerType = (resolved[0].fields.listObj.refType.type as ListType)?.refInnerType.type
```
ori-moisis, line 700: why did you take the resolve out of the before each?

and if there is a reason to do that, why are these (`resolved`, `resolvedInnerType`) still shared variables and not `const` in each test?


was anything actually changed here at all?

packages/workspace/test/core/expressions.test.ts
```diff
@@ -165,303 +165,321 @@ describe('Test Salto Expressions', () => {
...
+          await resolve(
+            elements,
+            createInMemoryElementSource(elements),
+            { shouldResolveReferences: false },
+          )
+        ).toArray()
+      })
+      it('should not resolve references', () => {
+        const element = findResolved<InstanceElement>(simpleRefInst)
+        expect(isReferenceExpression(element.value.test)).toEqual(true)
```
ori-moisis, line 185: this doesn't actually test any behavior
this will be a reference expression regardless of the flag you pass into `shouldResolveReferences`

packages/workspace/test/core/expressions.test.ts
tamtamirr: I took it out since I create different **ElementSource** as inputs.  I'll make the **resolved** and **resolvedInnerType** consts though :)

ori-moisis: what is different exactly?

packages/core/src/core/adapters/adapters.ts
```diff
@@ -181,6 +194,79 @@ const filterElementsSource = (
...
+      const resolvedType = resolvedTypes.get(instance.refType.elemID.getFullName())
+      // The type of the Element must be resolved here, otherwise we have a bug
+      if (resolvedType === undefined) {
+        log.warn('Expected type of Instance %s to be resolved. Type elemID: %s. Returning Instance with non fully resolved type.', instance.elemID.getFullName(), instance.refType.elemID.getFullName())
+        return instance
+      }
+      // If the type of the Element is resolved, simply set the type on the instance
+      instance.refType.type = resolvedType
+      return instance
+    }
```
ori-moisis, line 259: We can handle all cases of non-top-level IDs as a single case.
in the code before if the ID is a specific attribute in a type for example, we will bypass the cache and get the type again.
Handling it like this also handles fields because field IDs are nested under their type

```suggestion
  const getResolved = async (id: ElemID): Promise<Element> => {
    if (!typesWereResolved) {
      await resolveTypes()
    }
    // Container types
    const containerInfo = id.getContainerPrefixAndInnerType()
    if (containerInfo !== undefined) {
      const resolvedInner = await getResolved(ElemID.fromFullName(containerInfo.innerTypeName))
      if (isType(resolvedInner)) {
        return buildContainerType(containerInfo.prefix, resolvedInner)
      }
    }
    const { parent } = id.createTopLevelParentID()
    const alreadyResolvedType = resolvedTypes.get(parent.getFullName())
    if (alreadyResolvedType !== undefined) {
      // resolvePath here to handle cases where the input ID was for a field or attribute inside the type
      return resolvePath(alreadyResolvedType, id)
    }
    const value = await elementsSource.get(id)
    // Instances
    if (isInstanceElement(value)) {
      const instance = value.clone()
      const resolvedType = resolvedTypes.get(instance.refType.elemID.getFullName())
      // The type of the Element must be resolved here, otherwise we have a bug
      if (resolvedType === undefined) {
        log.warn('Expected type of Instance %s to be resolved. Type elemID: %s. Returning Instance with non fully resolved type.', instance.elemID.getFullName(), instance.refType.elemID.getFullName())
        return instance
      }
      // If the type of the Element is resolved, simply set the type on the instance
      instance.refType.type = resolvedType
      return instance
    }
```

ori-moisis (approved):
suggested a minor improvement for cases where a non-top-level ID is requested which should be fairly rare anyway

PR #4694, opened 2023-08-02
SALTO-4138: enforce uniqueness of workflow and scripts custom fields

SeggevHaimovich:
_Added workflows and scripts to the unique fields change validator_

---

_ticket: https://salto-io.atlassian.net/browse/SALTO-4138 first ticket about this change validator https://salto-io.atlassian.net/browse/SALTO-2915
Changed some logics in the existing code for it to match the new type of fields (inner fields)_

---
_Release Notes_: 
_In case users give the same scriptid for the custom fields, the change validator will cache this and cancel the deployment instead of getting exception_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/netsuite-adapter/src/change_validators/unique_fields.ts
+++ b/packages/netsuite-adapter/src/change_validators/unique_fields.ts
@@ -206,11 +206,7 @@ const validateDuplication = (
   return awu(changesData)
     .map(change => ({
       elemID: change.elemID,
-      fields: Array.from(
-        new Set(
-          getters.getChangeRestrictedFields(change).filter(field => uniqueFieldToID[field] > 1)
-        )
-      ),
+      fields: _.uniq(getters.getChangeRestrictedFields(change).filter(field => uniqueFieldToID[field] > 1)),
     }))
     .filter(({ fields }) => fields.length > 0)
     .map(({ elemID, fields }): ChangeError => ({
--- a/packages/netsuite-adapter/src/change_validators/unique_fields.ts
+++ b/packages/netsuite-adapter/src/change_validators/unique_fields.ts
@@ -68,13 +68,18 @@ const getCustomRecordRestrictedData = async ({ elemID, elementsSource }: GetterP
     .map(field => element.fields[field].annotations[SCRIPT_ID])
 }
 
-const workflowSourceGetter = async ({ elemID, elementsSource }: GetterParams): Promise<string[]> =>
-  _.values((await elementsSource.get(elemID)).value.workflowcustomfields?.workflowcustomfield ?? {})
-    .map(val => val.scriptid)
+const getChangeNestedScriptIDField = (
+  change: ChangeDataType,
+  nestPath: string[],
+): string[] =>
+  _.values(getChangeNestedField(change, nestPath)).map(val => val[SCRIPT_ID])
 
-const scriptSourceGetter = async ({ elemID, elementsSource }: GetterParams): Promise<string[]> =>
-  _.values((await elementsSource.get(elemID)).value.scriptcustomfields?.scriptcustomfield ?? {})
-    .map(val => val.scriptid)
+const getNestedScriptIDField = async (
+  { elemID, elementsSource }: GetterParams,
+  nestPath: string[],
+): Promise<string[]> =>
+  _.values(await elementsSource.get(elemID.createNestedID(...nestPath)))
+    .map(val => val[SCRIPT_ID])
 
 const savedSearchGetters: RestrictedTypeGetters = {
   getChangeRestrictedFields: change => [getChangeNestedField(change, [FIELD_DEFAULT_NAME])],
@@ -101,18 +106,16 @@ const customRecordGetters: RestrictedTypeGetters = {
 }
 
 const workflowGetters: RestrictedTypeGetters = {
-  getChangeRestrictedFields: change =>
-    _.values(getChangeNestedField(change, WORKFLOW_RESTRICTED_PATH)).map(val => val.scriptid),
-  getSourceRestrictedFields: workflowSourceGetter,
+  getChangeRestrictedFields: change => getChangeNestedScriptIDField(change, WORKFLOW_RESTRICTED_PATH),
+  getSourceRestrictedFields: params => getNestedScriptIDField(params, WORKFLOW_RESTRICTED_PATH),
   getMessage: () => 'Workflow contains custom fields with non-unique IDs',
   getDetailedMessage: scriptids => `Can't deploy this workflow as it contains custom fields with IDs that are not unique within this environment: "${scriptids.join(', ')}".`
   + ' To deploy it, change their IDs to unique ones.',
 }
 
 const scriptGetters: RestrictedTypeGetters = {
-  getChangeRestrictedFields: change =>
-    _.values(getChangeNestedField(change, SCRIPT_RESTRICTED_PATH)).map(val => val.scriptid),
-  getSourceRestrictedFields: scriptSourceGetter,
+  getChangeRestrictedFields: change => getChangeNestedScriptIDField(change, SCRIPT_RESTRICTED_PATH),
+  getSourceRestrictedFields: params => getNestedScriptIDField(params, SCRIPT_RESTRICTED_PATH),
   getMessage: () => 'Script contains parameters with non-unique IDs',
   getDetailedMessage: scriptids => `Can't deploy this script as it contains parameters ("scriptcustomfields") with IDs that are not unique within this environment: "${scriptids.join(', ')}".`
   + ' To deploy it, change their IDs to unique ones.',
--- a/packages/netsuite-adapter/src/change_validators/unique_fields.ts
+++ b/packages/netsuite-adapter/src/change_validators/unique_fields.ts
@@ -14,12 +14,12 @@
 * limitations under the License.
 */
 import { ElemID, getChangeData, isAdditionOrModificationChange, ChangeError,
-  ReadOnlyElementsSource, ChangeDataType, isObjectType } from '@salto-io/adapter-api'
+  ReadOnlyElementsSource, ChangeDataType, isObjectType, isInstanceElement } from '@salto-io/adapter-api'
 import { values, collections, promises } from '@salto-io/lowerdash'
 import _ from 'lodash'
 import { resolvePath } from '@salto-io/adapter-utils'
-import { isCustomFieldName, isCustomRecordType } from '../types'
-import { NAME_FIELD, FINANCIAL_LAYOUT, SAVED_SEARCH, SCRIPT_ID, CUSTOM_RECORD_TYPE_NAME_PREFIX } from '../constants'
+import { isCustomFieldName, isCustomRecordType, SCRIPT_TYPES } from '../types'
+import { NAME_FIELD, FINANCIAL_LAYOUT, SAVED_SEARCH, SCRIPT_ID, CUSTOM_RECORD_TYPE_NAME_PREFIX, WORKFLOW } from '../constants'
 import { NetsuiteChangeValidator } from './types'
 
 const { awu } = collections.asynciterable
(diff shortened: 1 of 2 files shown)


SeggevHaimovich:
As you can see, in the last commit, I changed 2 things:
1. Instead of checking if the change is an instance I assumed it is instance. As I see it we use those functions only on elements that we previously filtered using the "getRestrictedType" function, where we check they are instances.
2. I removed the undefined check. I don't think that there is a case we can get undefined there because the uniqueFieldToID is a Record of the restricted field of all elements of this type in the element source. There is no option that our element have a field that isn't in this Record.

I would like the reviewers to double-check I didn't miss any edge cases in these 2 decisions.

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -66,35 +66,68 @@ const getCustomRecordRestrictedData = async ({ elemID, elementsSource }: GetterP
     .map(field => element.fields[field].annotations[SCRIPT_ID])
 }
 
+const getWorkflowChangeRestrictedData = (change: ChangeDataType): string[] =>
+  // The change nust be an instance, we call this function only on instances of workflows we filtered from the changes
+  _.values((change as InstanceElement).value.workflowcustomfields?.workflowcustomfield ?? {}).map(val => val.scriptid)
```
dantal4, line 71: 1. Casting shouldn't be used any where (except  for places that we're 110% sure of the type). In here- the assumption in the code that we're iterating only instances is too far in the code to be taken as a fact, and if it will be change we might not catch the bug here.
2. We should use consts here- especially that they are used in more than one place.
3. Consider using `resolvePath` and `elemId.createNestedID(...WORKFLOW_CUSTOM_FIELDS)`. You can also change `getChangeNestedField` to return `Value` and then to use it here.

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -66,35 +66,68 @@ const getCustomRecordRestrictedData = async ({ elemID, elementsSource }: GetterP
...
+
+const workflowSourceGetter = async ({ elemID, elementsSource }: GetterParams): Promise<string[]> =>
+  _.values((await elementsSource.get(elemID)).value.workflowcustomfields?.workflowcustomfield ?? {})
+    .map(val => val.scriptid)
+
+const getscriptChangeRestrictedData = (change: ChangeDataType): string[] =>
```
dantal4, line 77:
```suggestion
const getScriptChangeRestrictedData = (change: ChangeDataType): string[] =>
```

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -66,35 +66,68 @@ const getCustomRecordRestrictedData = async ({ elemID, elementsSource }: GetterP
...
     + ' To deploy it, change its title to a unique one.',
 }
 
 const financialLayoutGetters: RestrictedTypeGetters = {
-  getChangeRestrictedField: change => getChangeNestedField(change, NAME_FIELD),
+  getChangeRestrictedField: change => [getChangeNestedField(change, NAME_FIELD)],
   getSourceRestrictedFields: params => getNestedField(params, NAME_FIELD),
   getMessage: () => 'A Financial Layout with that name already exists',
-  getDetailedMessage: name => `Can't deploy this Financial Layout because there is already a Financial Layout with the name "${name}" in the target account.`
+  getDetailedMessage: name => `Can't deploy this Financial Layout because there is already a Financial Layout with the name "${name[0]}" in the target account.`
```
dantal4, line 97: nit
```suggestion
  getDetailedMessage: ([name]) => `Can't deploy this Financial Layout because there is already a Financial Layout with the name "${name}" in the target account.`
```

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -29,21 +29,21 @@ const { mapValuesAsync } = promises.object
...
   elemID: ElemID
   elementsSource: ReadOnlyElementsSource
 }
 
 type ElementGetter = (params: GetterParams) => Promise<string[]>
-type ChangeGetter = (change: ChangeDataType) => string
+type ChangeGetter = (change: ChangeDataType) => string[]
 
 type RestrictedTypeGetters = {
   getChangeRestrictedField: ChangeGetter
```
dantal4, line 43: This should be changed to plural, right?

```suggestion
  getChangeRestrictedFields: ChangeGetter
```

dantal4 (commented):
Great work!
Just to make sure, workflow parameters and script parameters must be unique accross the accout, right?

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -164,14 +207,18 @@ const validateDuplication = (
   return awu(changesData)
     .map(change => ({
       elemID: change.elemID,
-      field: getters.getChangeRestrictedField(change),
+      fields: Array.from(
+        new Set(
+          getters.getChangeRestrictedField(change).filter(field => (uniqueFieldToID[field]) > 1)
```
dantal4, line 212: nit
```suggestion
          getters.getChangeRestrictedField(change).filter(field => uniqueFieldToID[field] > 1)
```

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -66,35 +68,62 @@ const getCustomRecordRestrictedData = async ({ elemID, elementsSource }: GetterP
     .map(field => element.fields[field].annotations[SCRIPT_ID])
 }
 
+const workflowSourceGetter = async ({ elemID, elementsSource }: GetterParams): Promise<string[]> =>
+  _.values((await elementsSource.get(elemID)).value.workflowcustomfields?.workflowcustomfield ?? {})
```
dantal4, line 72: `elementsSource.get` can be used with a nested elemID, and then you'll be able to use the `WORKFLOW_RESTRICTED_PATH` here too.

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -66,35 +68,62 @@ const getCustomRecordRestrictedData = async ({ elemID, elementsSource }: GetterP
...
+    .map(val => val.scriptid)
+
+const scriptSourceGetter = async ({ elemID, elementsSource }: GetterParams): Promise<string[]> =>
```
dantal4, line 75: except for the value path in the element, this function is identical to `workflowSourceGetter` - please share a code here.

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -66,35 +68,62 @@ const getCustomRecordRestrictedData = async ({ elemID, elementsSource }: GetterP
...
   getSourceRestrictedFields: getCustomRecordRestrictedData,
   getMessage: () => 'A Custom Record Type Field with that ID already exists',
-  getDetailedMessage: scriptID => `Can't deploy this Custom Record Type Field because there is already a Custom Record Type Field with the ID "${scriptID}" in the target account.`
+  getDetailedMessage: ([scriptID]) => `Can't deploy this Custom Record Type Field because there is already a Custom Record Type Field with the ID "${scriptID}" in the target account.`
     + ' To deploy it, change its ID to a unique one.',
 }
 
+const workflowGetters: RestrictedTypeGetters = {
+  getChangeRestrictedFields: change =>
+    _.values(getChangeNestedField(change, WORKFLOW_RESTRICTED_PATH)).map(val => val.scriptid),
```
dantal4, line 105: please use the existing `SCRIPT_ID` const (also in other places)

packages/netsuite-adapter/test/change_validators/unique_fields.test.ts
```diff
@@ -59,6 +60,22 @@ describe('unique fields validator', () => {
       [elem.elemID.createNestedID(nestedField).getFullName(),
         (elem as InstanceElement).value[nestedField]]))
 
+  const getIDToValWorkflow = (
```
dantal4, line 63: how this function different from `getIDToVal`?

packages/netsuite-adapter/test/change_validators/unique_fields.test.ts
```diff
@@ -59,6 +60,22 @@ describe('unique fields validator', () => {
...
+    testInstances: TestElements,
+    nestedField: string
+  ): Map<string, string> => new Map(Object.values(testInstances)
+    .map(elem =>
+      [elem.elemID.createNestedID(nestedField).getFullName(),
+        _.get((elem as InstanceElement).value, nestedField)]))
+
+  const getIDToValScript = (
```
dantal4, line 71: same question

packages/netsuite-adapter/test/change_validators/unique_fields.test.ts
```diff
@@ -82,6 +99,36 @@ describe('unique fields validator', () => {
...
 
+  const getWorkflowElements = (): TestElements => getTestElements(WORKFLOW,
+    (name: string, elemID: ObjectType, uniqueField: string) => {
+      const custworkflow1 = {
+        scriptid: uniqueField,
+      }
+      const workflowcustomfield = { custworkflow1 }
+      const workflowCustomFields = { workflowcustomfield }
+      return new InstanceElement(name, elemID, { workflowcustomfields: workflowCustomFields })
+    })
```
dantal4, line 110: nit
```suggestion
    (name: string, elemID: ObjectType, uniqueField: string) =>
      new InstanceElement(name, elemID, {
        workflowcustomfields: {
          workflowcustomfield: {
            custworkflow1: {
              scriptid: uniqueField,
            },
          },
        },
      })
    )
```

packages/netsuite-adapter/test/change_validators/unique_fields.test.ts
```diff
@@ -82,6 +99,36 @@ describe('unique fields validator', () => {
...
+
+  const getScriptSuiteletElements = (): TestElements => getTestElements('suitelet',
+    (name: string, elemID: ObjectType, uniqueField: string) => {
+      const custscript1 = {
+        scriptid: uniqueField,
+      }
+      const scriptcustomfield = { custscript1 }
+      const scriptCustomFields = { scriptcustomfield }
+      return new InstanceElement(name, elemID, { scriptcustomfields: scriptCustomFields })
```
dantal4, line 119: same nit

packages/netsuite-adapter/test/change_validators/unique_fields.test.ts
```diff
@@ -82,6 +99,36 @@ describe('unique fields validator', () => {
...
+    })
+
+  const getScriptRestletElements = (): TestElements => getTestElements('restlet',
+    (name: string, elemID: ObjectType, uniqueField: string) => {
+      const custscript1 = {
+        scriptid: uniqueField,
+      }
+      const scriptcustomfield = { custscript1 }
+      const scriptCustomFields = { scriptcustomfield }
+      return new InstanceElement(name, elemID, { scriptcustomfields: scriptCustomFields })
```
dantal4, line 129: same nit

dantal4 (commented):
Great work!

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -164,14 +206,18 @@ const validateDuplication = (
   return awu(changesData)
     .map(change => ({
       elemID: change.elemID,
-      field: getters.getChangeRestrictedField(change),
+      fields: Array.from(
+        new Set(
+          getters.getChangeRestrictedFields(change).filter(field => uniqueFieldToID[field] > 1)
+        )
+      ),
```
dantal4, line 213: Another way to do it- decide which one you prefer
```suggestion
      fields: _.uniq(getters.getChangeRestrictedFields(change).filter(field => uniqueFieldToID[field] > 1)),
```

dantal4 (approved):
💯

PR #4696, opened 2023-08-02
SALTO-4293 Update validateCredentials to identify whether the account is production or not

idoamit8:
As of today, Jira does not have an API that checks if an account is production or sandbox.
After consultations with @omrilit  and @liorn, the solution implemented is:
1. Check if the URL contains `-sandbox-`.
2. Check if the account has paid apps.
If the URL contains the word sandbox or all the apps in the account are in a free plan, then the account will not be defined as Production.
---
None

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/client/connection.ts
+++ b/packages/jira-adapter/src/client/connection.ts
@@ -13,10 +13,14 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
+import { logger } from '@salto-io/logging'
 import { AccountInfo, CredentialError } from '@salto-io/adapter-api'
 import { client as clientUtils } from '@salto-io/adapter-components'
+import { safeJsonStringify } from '@salto-io/adapter-utils'
 import { Credentials } from '../auth'
 import { FORCE_ACCEPT_LANGUAGE_HEADERS } from './headers'
+
+const log = logger(module)
 
 type appInfo = {
   id: string
@@ -57,7 +61,8 @@ export const validateCredentials = async (
       return { accountId, isProduction: false, accountType: 'Sandbox' }
     }
     const response = await connection.get('/rest/api/3/instance/license')
-    const hasPaidApp = response.data.applications.some((app: appInfo) => app.plan === 'PAID') ?? false
+    log.info(`Jira application's info: ${safeJsonStringify(response.data.applications)}`)
+    const hasPaidApp = response.data.applications.some((app: appInfo) => app.plan === 'PAID')
     const isProduction = hasPaidApp ? undefined : false
     return { accountId, isProduction }
   }
--- a/packages/jira-adapter/src/client/connection.ts
+++ b/packages/jira-adapter/src/client/connection.ts
@@ -52,7 +52,7 @@ const isSandBox = async (connection: clientUtils.APIConnection): Promise<boolean
 const isProd = async (connection: clientUtils.APIConnection): Promise<boolean | undefined> => {
   const isSandBoxAccount = await isSandBox(connection)
   const response = await connection.get('/rest/api/3/instance/license')
-  const hasPaidApp = response.data?.applications.some((app: appInfo) => app.plan === 'PAID') ?? false
+  const hasPaidApp = response.data.applications.some((app: appInfo) => app.plan === 'PAID') ?? false
   return !isSandBoxAccount && hasPaidApp ? undefined : false
 }
 
--- a/packages/jira-adapter/src/client/connection.ts
+++ b/packages/jira-adapter/src/client/connection.ts
@@ -44,27 +44,21 @@ const getBaseUrl = async (
   return response.data.baseUrl
 }
 
-const isSandBox = async (connection: clientUtils.APIConnection): Promise<boolean> => {
-  const baseURL = await getBaseUrl(connection)
-  return baseURL.includes('-sandbox-')
-}
-
-const isProd = async (connection: clientUtils.APIConnection): Promise<boolean | undefined> => {
-  const isSandBoxAccount = await isSandBox(connection)
-  const response = await connection.get('/rest/api/3/instance/license')
-  const hasPaidApp = response.data.applications.some((app: appInfo) => app.plan === 'PAID') ?? false
-  return !isSandBoxAccount && hasPaidApp ? undefined : false
-}
-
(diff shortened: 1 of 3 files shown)


packages/jira-adapter/src/client/connection.ts
```diff
@@ -39,12 +44,20 @@ const getBaseUrl = async (
   return response.data.baseUrl
 }
 
+const isProd = async (connection: clientUtils.APIConnection): Promise<boolean> => {
```
omrilit, line 47: IMO it's better to return undefined `isProduction` when we are not sure about it rather then guessing it is true.
We have 2 heuristic that can determine whether the account is free/SB. In those cases let's return `isProduction: false`.
In all other cases lets return `undefined`.
WDYT @shoshana-houri @idoamit8 ?

BTW, when you know the answer, let's populate also the `accountType` field of `AccountInfo`.

idoamit8: Great. I added `accountType = Sandbox` when the URL includes -sandbox-.

packages/jira-adapter/src/client/connection.ts
```diff
@@ -39,12 +44,28 @@ const getBaseUrl = async (
   return response.data.baseUrl
 }
 
+const isSandBox = async (connection: clientUtils.APIConnection): Promise<boolean> => {
+  const baseURL = await getBaseUrl(connection)
+  return baseURL.includes('-sandbox-')
+}
+
+const isProd = async (connection: clientUtils.APIConnection): Promise<boolean | undefined> => {
+  const isSandBoxAccount = await isSandBox(connection)
```
omrilit, line 53: nit
```suggestion
  const isSandboxAccount = await isSandbox(connection)
```

packages/jira-adapter/src/client/connection.ts
```diff
@@ -39,12 +44,28 @@ const getBaseUrl = async (
   return response.data.baseUrl
 }
 
+const isSandBox = async (connection: clientUtils.APIConnection): Promise<boolean> => {
+  const baseURL = await getBaseUrl(connection)
+  return baseURL.includes('-sandbox-')
+}
+
+const isProd = async (connection: clientUtils.APIConnection): Promise<boolean | undefined> => {
```
omrilit, line 52: I would add a comment here explaining that based on the current data we extract we cannot be sure that the account isProduction but sometimes we can be sure that it is no a Production account
```suggestion
const isProd = async (connection: clientUtils.APIConnection): Promise<false | undefined> => {
```

packages/jira-adapter/src/client/connection.ts
```diff
@@ -39,12 +44,28 @@ const getBaseUrl = async (
...
+}
+
 export const validateCredentials = async (
   { connection }: { connection: clientUtils.APIConnection },
 ): Promise<AccountInfo> => {
   if (await isAuthorized(connection)) {
     const accountId = await getBaseUrl(connection)
-    return { accountId }
+    const isProduction = await isProd(connection)
+    if (await isSandBox(connection)) {
```
omrilit, line 65: I think that there is no reason to always ask whether the account isSandbox when we already know whether it's production or not.
In addition I see in row 63 you already call `getBaseUrl` which is called also in isSandbox.
Maybe it would be better to extract all of this code to a single function called `getAccountInfo` and will extract all related data?

packages/jira-adapter/src/client/connection.ts
```diff
@@ -39,12 +44,28 @@ const getBaseUrl = async (
   return response.data.baseUrl
 }
 
+const isSandBox = async (connection: clientUtils.APIConnection): Promise<boolean> => {
+  const baseURL = await getBaseUrl(connection)
+  return baseURL.includes('-sandbox-')
```
alonstern, line 49: Can we be more specific than that? e.g., if it always ends with `-sandbox-` and then a number verify with a regex that it ends with that?

packages/jira-adapter/src/client/connection.ts
idoamit8: I thought that the account can be neither production or sandbox. Am I wrong?
If it not production, you want to me to write `accountType: 'Sandbox'`?

packages/jira-adapter/src/client/connection.ts
idoamit8: I thought about it, but I couldn't find any certainty on the internet. Indeed, all the examples ended with 3 digits after the sandbox, but they did not commit to this anywhere.

packages/jira-adapter/src/client/connection.ts
```diff
@@ -39,12 +44,22 @@ const getBaseUrl = async (
...
   { connection }: { connection: clientUtils.APIConnection },
 ): Promise<AccountInfo> => {
   if (await isAuthorized(connection)) {
     const accountId = await getBaseUrl(connection)
-    return { accountId }
+    if (accountId.includes('-sandbox-')) {
+      return { accountId, isProduction: false, accountType: 'Sandbox' }
+    }
+    const response = await connection.get('/rest/api/3/instance/license')
+    const hasPaidApp = response.data.applications.some((app: appInfo) => app.plan === 'PAID') ?? false
```
omrilit, line 60:
```suggestion
    const hasPaidApp = response.data.applications.some((app: appInfo) => app.plan === 'PAID')
```

packages/jira-adapter/src/client/connection.ts
```diff
@@ -39,12 +44,22 @@ const getBaseUrl = async (
...
 export const validateCredentials = async (
   { connection }: { connection: clientUtils.APIConnection },
 ): Promise<AccountInfo> => {
   if (await isAuthorized(connection)) {
     const accountId = await getBaseUrl(connection)
-    return { accountId }
+    if (accountId.includes('-sandbox-')) {
+      return { accountId, isProduction: false, accountType: 'Sandbox' }
+    }
+    const response = await connection.get('/rest/api/3/instance/license')
```
omrilit, line 59: If you think that it would be valuable for the future so we could better understand if we can analyze the data here, we can consider to log `response.data.applications`

omrilit (approved):
Thanks @idoamit8 looks good!
Approving to save another review cycle.

PR #4697, opened 2023-08-02
SALTO-4377: Return account type for Okta

shir-reifenberg:
Return account type for Okta

---

identify if the account is sandbox or production based on base url

---
_Release Notes_: 
None

---
_User Notifications_: 
None


edenhassid (approved):
looks good!! 🔥

PR #4698, opened 2023-08-02
SALTO-4499 - Handle errors on article attachment content requests

idoser96:
When failing to get content of attachment, don't fail the fetch, don't put any content and alert the user about the failure

---

None

---
_Release Notes_: 
Zendesk Adapter:
* Failure of requesting article attachment content will no longer fail the fetch

---
_User Notifications_: 
None


edenhassid (approved):
great!! 💯

PR #4700, opened 2023-08-06
SALTO-4374: update validateCredentials to identify production accounts

edenhassid:
update validateCredentials to identify production accounts

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/client/connection.ts
+++ b/packages/zendesk-adapter/src/client/connection.ts
@@ -18,9 +18,17 @@ import axiosRetry from 'axios-retry'
 import { AccountInfo } from '@salto-io/adapter-api'
 import { client as clientUtils } from '@salto-io/adapter-components'
 import { logger } from '@salto-io/logging'
+import _ from 'lodash'
 import { Credentials, isOauthAccessTokenCredentials, OauthAccessTokenCredentials, UsernamePasswordCredentials } from '../auth'
 
 const log = logger(module)
+type AccountRes = {
+  data: {
+    account: {
+      sandbox: boolean
+    }
+  }
+}
 
 export const instanceUrl = (subdomain: string, domain?: string): string => (
   domain === undefined ? `https://${subdomain}.zendesk.com` : `https://${subdomain}.${domain}`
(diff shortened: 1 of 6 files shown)


packages/zendesk-adapter/src/client/connection.ts
```diff
@@ -50,6 +68,16 @@ export const validateCredentials = async ({ credentials, connection }: {
     throw new clientUtils.UnauthorizedError(e)
   }
   const accountId = instanceUrl(credentials.subdomain, credentials.domain)
+  try {
+    const res = await connection.get('/api/v2/account')
```
shir-reifenberg, line 72: Any reason not to replace the call in line 65 to this one? Seems like we're not using the info in `/api/v2/account/settings` anyway

packages/zendesk-adapter/src/client/connection.ts
```diff
@@ -50,6 +68,16 @@ export const validateCredentials = async ({ credentials, connection }: {
...
+    if (isValidAccountRes(res)) {
+      const isProduction = !res.data.account.sandbox
+      return { accountId, isProduction }
```
shir-reifenberg, line 75: Do we have account types in Zendesk? if we know what are the types maybe it's a good point to add this as well

packages/zendesk-adapter/src/client/connection.ts
```diff
@@ -39,6 +47,16 @@ export const APP_MARKETPLACE_HEADERS = {
   'X-Zendesk-Marketplace-App-Id': MARKETPLACE_APP_ID,
 }
 
+const isValidAccountRes = (res: unknown): res is AccountRes => (
```
shir-reifenberg, line 50: Why not `joi` ?

packages/zendesk-adapter/src/client/connection.ts
edenhassid: we don't have account types its either sandbox or production. I didn't want to add the account type as it will be duplication of information form `isProduction`

packages/zendesk-adapter/src/client/connection.ts
```diff
@@ -39,18 +56,25 @@ export const APP_MARKETPLACE_HEADERS = {
...
   try {
-    await connection.get('/api/v2/account/settings')
+    const res = await connection.get('/api/v2/account')
+    const accountId = instanceUrl(credentials.subdomain, credentials.domain)
+    if (isValidAccountRes(res)) {
+      const isProduction = !res.data.account.sandbox
+      return { accountId, isProduction }
+    }
+    log.warn('res is not valid for /api/v2/account, could not find if account is production')
+    return { accountId }
```
shir-reifenberg, line 73: @shoshana-houri is it valid not to return isProduction? Should we throw in this case / provide default value ?

shir-reifenberg (commented):
LGTM 😎 
I'm not sure what the behavior should be if we couldn't find out if it's production, let's wait to @shoshana-houri's answer

shir-reifenberg (approved):

PR #4701, opened 2023-08-06
SALTO-4534 - Zendesk Adapter - Prevent deployment of instances with the same id fields values

idoser96:
On addition, on specific types, make sure there are no duplicate id fields values between instances

---

* This can happen in case of a change of the field of idField, which makes it look like there is a new instance (but it is the old one before the rename)

---
_Release Notes_: 
Zendesk Adapter:
* Prevent deployment of groups that have the same values in their Id fields 

---
_User Notifications_: 
None


packages/zendesk-adapter/src/change_validators/duplicate_id_field_values.ts
```diff
@@ -0,0 +1,80 @@
...
+      .map(getChangeData)
+      .filter(instance => TYPES_TO_CHECK.includes(instance.elemID.typeName)),
+    change => change.elemID.typeName
+  )
+
+  const errors = await Promise.all(Object.entries(changedInstancesByType).map(async ([typeName, instances]) => {
+    const typeInstances = await getInstancesFromElementSource(elementSource, typeName)
+    const instancesByIdFields = _.groupBy(
+      typeInstances,
+      instance => generateInstanceNameFromConfig(instance.value, typeName, apiConfig)
```
edenhassid, line 60: you get here undefined right? this will cause unexpected behavior

packages/zendesk-adapter/src/change_validators/duplicate_id_field_values.ts
```diff
@@ -0,0 +1,80 @@
...
+        return undefined
+      }
+      const instancesWithSameIdFields = instancesByIdFields[instanceName]
+        .filter(i => i.elemID.getFullName() !== instance.elemID.getFullName())
+      return instancesWithSameIdFields.length > 0
+        ? {
+          elemID: instance.elemID,
+          severity: 'Error',
+          message: 'Duplicate unique field values',
+          detailedMessage: `This element has the same unique fields as '${instancesWithSameIdFields.map(i => i.elemID.name).join(', ')}', deploying it will cause Salto collisions, please make sure this is not an existing modified element`,
```
edenhassid, line 74: don't forget to approve the message, maybe we can make the message specific to the type

edenhassid (commented):
logic looks good! added a few comments

packages/zendesk-adapter/src/change_validators/duplicate_id_field_values.ts
idoser96: waiting for Tomer's approval

packages/zendesk-adapter/src/change_validators/duplicate_id_field_values.ts
edenhassid: discussed IRL

edenhassid (approved):
looks good! don't forget to get @tomermevorach approval as well

tomermevorach:
> looks good! don't forget to get @tomermevorach approval as well

@netama @edenhassid @idoser96  How about this?

Message: 
{TypeName}s duplication detected

Detailed Message:
This {TypeName} cannot be deployed as it is a duplicate of '${instancesWithSameIdFields.map(i => i.elemID.name).join(', ')}'. This likely indicates a misalignment of Salto IDs. To address this, please execute a fetch on both the source and target environments. Ensure you select the 'Regenerate Salto IDs' option in the advanced settings.

packages/zendesk-adapter/src/change_validators/duplicate_id_field_values.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  const errors = await Promise.all(Object.entries(changedInstancesByType).map(async ([typeName, instances]) => {
+    const typeInstances = await getInstancesFromElementSource(elementSource, [typeName])
+    const instancesByIdFields: Record<string, InstanceElement[]> = {}
+    // generated name can be undefined, so we can't use _.groupBy
+    typeInstances.forEach(instance => {
+      const instanceName = generateInstanceNameFromConfig(instance.value, typeName, apiConfig)
+      if (instanceName) {
+        instancesByIdFields[instanceName] = (instancesByIdFields[instanceName] ?? []).concat(instance)
+      }
```
netama, line 64: 1. please initialize these with functional-style
2. I don't think we should rely on the exact id fields, we should define a separate mapping for each type that depends on what zendesk itself requires. id fields can be customized in different ways and shouldn't impact this change validator

packages/zendesk-adapter/test/change_validators/duplicate_id_field_values.test.ts
```diff
@@ -0,0 +1,128 @@
...
+  })
+  it('should return an error for instances with the same zendesk id and same id fields values', async () => {
+    const duplicateGroup1 = createGroupInstance('duplicate1', 'name1')
+    const duplicateGroup2 = createGroupInstance('duplicate2', 'name2')
+
+    const changes = [toChange({ after: duplicateGroup1 })]
+    const elementSource = createInMemoryElementSource([duplicateGroup1, duplicateGroup2])
+
+    // Currently there are on types that have the different id fields in zendesk and salto, so we need to mock it
+    toZendeskIdMock.mockReturnValue('duplicate')
```
edenhassid, line 116: I don't understand why you had to use this mock, why not do
```
const duplicateGroup1 = createGroupInstance('duplicate1', 'name1')
    const duplicateGroup2 = createGroupInstance('duplicate1', 'name1')
    ```
    this will result in the same error right?

File: packages/zendesk-adapter/src/change_validators/duplicate_id_field_values.ts [hunk 5]
```diff
@@ -0,0 +1,106 @@
...
+const { isDefined } = lowerDashValues
+
+const log = logger(module)
+
+// Type to field names that together create a unique id in zendesk
+const ZENDESK_ID_FIELDS: Record<string, string[]> = {
+  group: ['name'],
+}
+
+export const toZendeskId = (typeName: string, instance: InstanceElement): string =>
```
edenhassid, line 39: why is this exported?

edenhassid (commented):
looks good! 😄  added a few comments. @netama can you also take a look?

PR #4702, opened 2023-08-06
SALTO-4578 - add getInstancesFromElementSource

idoser96:
Added getInstancesFromElementSource to get instance by type from elementSource

---

Also converted all usages
except for contexts_project_filter, as it is a bit more complicated there (and I don't want to overdo it)

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/adapter-utils/src/utils.ts
```diff
@@ -1323,3 +1323,11 @@ export const formatConfigSuggestionsReasons = (reasons: string[]): string => {
 export const isResolvedReferenceExpression = (value: unknown): value is ReferenceExpression => (
   isReferenceExpression(value) && !(value.value instanceof UnresolvedReference) && value.value !== undefined
 )
+
+export const getInstancesFromElementSource = async (elementSource: ReadOnlyElementsSource, typeNames: string[])
+  : Promise<InstanceElement[]> =>
+  awu(await elementSource.list())
```
alonstern, line 1329: @ori-moisis WDYT? do you have an idea if this is better than doing `getAll` and then filter?

alonstern (commented):
thanks 👌

packages/adapter-utils/src/utils.ts
ori-moisis: did you check the performance?
IIRC last time I checked filtering `getAll` works noticeably faster than many separate `get` calls

packages/adapter-utils/src/utils.ts
```diff
@@ -1323,3 +1323,10 @@ export const formatConfigSuggestionsReasons = (reasons: string[]): string => {
 export const isResolvedReferenceExpression = (value: unknown): value is ReferenceExpression => (
   isReferenceExpression(value) && !(value.value instanceof UnresolvedReference) && value.value !== undefined
 )
+
+export const getInstancesFromElementSource = async (elementSource: ReadOnlyElementsSource, typeNames: string[])
```
netama, line 1327: please don't forget to add a unittest if this is shared logic

PR #4704, opened 2023-08-06
SALTO-4576: Attempt to fix Salesforce Deploy regression

tamtamirr:
Every-time the default strategy would resolve a reference to undefined, we would return as a fallback the Element itself.
While this is valid to specific flows in the adapter, it shouldn't be valid to others which caused **regression** in various deploy flows for users.

---

This PR enhances the **getLookupName** implementation to be based on **ChangeGroup**. We will only apply to placeholder logic if the unresolved Reference exists in the current deploy group. Otherwise, we log it and omit the value from the deployed Element (That was basically the old behavior in the adapter)

---

_Release Notes_: 
Salesforce Adapter:
- Fixed regression that caused some deploys to fail in various scenarios

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/transformers/reference_mapping.ts
```diff
@@ -849,7 +865,18 @@ const getLookUpNameImpl = (defs = fieldNameToTypeMappingDefs): GetLookupNameFunc
       }
       if (isElement(ref.value)) {
         const defaultStrategy = ReferenceSerializationStrategyLookup.absoluteApiName
-        return await defaultStrategy.serialize({ ref, element }) ?? ref.value
+        const resolvedValue = await defaultStrategy.serialize({ ref, element })
+        if (resolvedValue !== undefined) {
+          return resolvedValue
+        }
+        if (groupElemIds !== undefined && groupElemIds.has(ref.value.elemID.getFullName())) {
+          // We want to return the referenced Element in that case, which will be handled later in the deployment flow.
```
benisraeladi, line 873: I think it would be easier to read if `ref.value.elemID.getFullName()` was a variable with a more indicative name.

ori-moisis:
discussed f2f - seems like the decision on which serialization strategy to fallback to can depend on just if this is a group of data records or metadata (data records can fallback to the instance, metadata should not)

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -461,8 +461,12 @@ export default class SalesforceAdapter implements AdapterOperations {
     checkOnly: boolean
   ): Promise<DeployResult> {
     log.debug(`about to ${checkOnly ? 'validate' : 'deploy'} group ${changeGroup.groupID} with scope (first 100): ${safeJsonStringify(changeGroup.changes.slice(0, 100).map(getChangeData).map(e => e.elemID.getFullName()))}`)
+    const isDataDeployGroup = await isCustomObjectInstanceChanges(changeGroup.changes as Change[])
```
ori-moisis, line 464: you can change `isCustomObjectInstanceChanges` to get a readonly array, it doesn't really need to change it

packages/salesforce-adapter/src/transformers/reference_mapping.ts
```diff
@@ -809,7 +849,10 @@ export const generateReferenceResolverFinder = (
   ]).filter(resolver => resolver.match(field, element)).toArray())
 }
 
-const getLookUpNameImpl = (defs = fieldNameToTypeMappingDefs): GetLookupNameFunc => {
+const getLookUpNameImpl = (
+  defs = fieldNameToTypeMappingDefs,
+  resolveToElementFallback: boolean,
+): GetLookupNameFunc => {
```
ori-moisis, line 855: nit: I think we should change this to named parameters

packages/salesforce-adapter/src/transformers/transformer.ts
```diff
@@ -988,14 +988,19 @@ const toRecord = async (
         }
         return referencedRecordId
       }
+      // Can happen when we failed to resolve a reference to Metadata Element.
+      if (isElement(val)) {
+        log.warn('Unexpected non deployable value of Element with elemID %s in toRecord on Instance with elemID %s. Omitting value', val.elemID.getFullName(), instance.elemID.getFullName())
+        return undefined
```
ori-moisis, line 994: looks like the other case above here returns `null` and not `undefined`
is there a reason this should behave differently?

ori-moisis (approved):

packages/salesforce-adapter/src/transformers/transformer.ts
tamtamirr: Yes and no. Rethinking it when implementing this code, it'd be better to return undefined in the other flow as-well, since undefined will not omit any existing values while null will omit values upon update. Didn't wanna couple the two, as I don't want to test data deploy flows all over again

PR #4705, opened 2023-08-07
SALTO-4503: A reference expression to a parent value enters an infinite loop in deployment

alonstern:
Fixed a bug where referencing the parent value in the same instance would cause an infinite recursion

---

In Jira we are going to have a case like
```
jira.Workflow workflow {
   ...
   transitions = {
      transitionName = {
         properties = [
           {
             key = "relevant transition"
             value = jira.Workflow.workflow.transitions.transitionName
           }
         ]
      }
   }
}
```
When deploying, `jira.Workflow.workflow.transitions.transitionName` will be converted to the id of the transition.

This case currently causing an infinite recursion in the validation part on fetch and in the compareSpecialValues part on deploy.

---
_Release Notes_: 
None (no effect on the user currently)

---
_User Notifications_: 
None


packages/adapter-api/src/values.ts
```diff
@@ -273,13 +274,23 @@ export const compareSpecialValues = (
   }
   if (isReferenceExpression(first) || isReferenceExpression(second)) {
     if (shouldCompareByValue(first, second, options)) {
+      if (isReferenceExpression(first) && isReferenceExpression(second)) {
+        const key = `${first.elemID.getFullName()}-${second.elemID.getFullName()}`
+        if (comparedReferences.has(key)) {
+          // We assume two circular references are equal (sames as _.isEqual with two circular objects)
```
ori-moisis, line 280: this is implemented differently from `_.isEqual` though.
`_.isEqual` will stop earlier on loops, this key will iterate more cycles and return a different result (`_.isEqual` will return false on things that this will return `true` for)

not sure if this is important for your use case...

ori-moisis (commented):
I think this is ok, not sure...

packages/workspace/src/validator.ts
```diff
@@ -491,11 +492,16 @@ const validateValue = (
...
       ) ?? []
     }
-    return validateValue(elemID, value, new ListType(type))
+    return validateValue(elemID, value, new ListType(type), validatedReferences)
   }
 
   if (isReferenceExpression(value)) {
-    return isElement(value.value) ? [] : validateValue(elemID, value.value, type)
+    const key = `${value.elemID.getFullName()}-${type.elemID.getFullName()}`
+    if (!isElement(value.value) && !(validatedReferences.has(key) && value.elemID.isParentOf(elemID))) {
```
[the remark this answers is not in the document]
alonstern: Rethinking about it, I don't think we need to type elemId in the key. Added docs

packages/workspace/src/validator.ts
```diff
@@ -491,11 +492,19 @@ const validateValue = (
...
     }
-    return validateValue(elemID, value, new ListType(type))
+    return validateValue(elemID, value, new ListType(type), validatedReferences)
   }
 
   if (isReferenceExpression(value)) {
-    return isElement(value.value) ? [] : validateValue(elemID, value.value, type)
+    if (!isElement(value.value)
+      // We check for both validatedReferences and isParentOf because if we got to
+      // the same reference twice but not through a cycle we do want to validate it
```
ori-moisis, line 501: won't this mean that if we get to a cycle between two elements we'd end up in an infinite loop again?

packages/workspace/test/validator.test.ts
```diff
@@ -2364,6 +2364,29 @@ describe('Elements validation', () => {
       expect(errors).toHaveLength(1)
       expect(errors[0].message).toBe('Error validating "instance.notExists.instance.name": type notExists of instance name does not exist')
     })
+
+    it('should handle circular references in the same instance', async () => {
```
ori-moisis, line 2368: I think we should also test for circular references between two instances

packages/adapter-api/test/values.test.ts
```diff
@@ -234,6 +234,21 @@ describe('Values', () => {
         const ref1 = new ReferenceExpression(new ElemID('adapter', 'type', 'instance', 'inst', 'val1'), 1)
         expect(isEqualValues(ref1, 1, { compareByValue: true })).toBeTruthy()
       })
+
+      it('Comparing circular dependencies should return true', () => {
```
ori-moisis, line 238: is there no case where we would want circular dependencies to not compare equal?
e.g - if there are two cycles that are completely detached, do we still want them to be considered equal?

packages/adapter-api/test/values.test.ts
```diff
@@ -234,6 +234,21 @@ describe('Values', () => {
...
+          new ObjectType({ elemID: new ElemID('adapter', 'type') }),
+        )
+
+        instance.value.a = {
+          ref: new ReferenceExpression(instance.elemID.createNestedID('a')),
+        }
+
+        instance.value.a.ref.value = instance.value.a
+
+        expect(isEqualValues(instance, instance, { compareByValue: true })).toBeTruthy()
```
ori-moisis, line 250: I don't think comparing an object to itself is a valid test, for example `_.isEqual` may not recurse into it at all because `instance === instance` will be true

packages/adapter-api/src/values.ts
ori-moisis: if you are sure this doesn't matter for our use case, can we at least change the comment to not claim this is like `_.isEqual`?
I might be the only one that this will bother, but knowing how `_.isEqual` works, this would be a problem if we ever need to debug this (might cause us to assume this works like `_.isEqual` when in fact it doesn't)

packages/workspace/src/validator.ts
alonstern: Why? I added a test for two instances

packages/adapter-api/src/values.ts
alonstern: I meant it's only like isEqual with the assumption about the circular references and not the implementation (based on the comment [here](https://github.com/lodash/lodash/blob/master/.internal/equalObjects.js#L40))

packages/adapter-api/src/comparison.ts
```diff
@@ -0,0 +1,110 @@
...
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+import _ from 'lodash'
+// eslint-disable-next-line import/no-cycle
+import { areReferencesEqual, shouldResolve } from './reference_comparison'
+// eslint-disable-next-line import/no-cycle
```
alonstern, line 19: There is a circular dependency between values.ts and element.ts. Since I exported logic from values.ts to a new file it is still part of the cycle so I have to add this eslit-disable

ori-moisis: lets make it one new file instead of two to lower the impact of this a little

packages/workspace/src/validator.ts
```diff
@@ -491,11 +492,20 @@ const validateValue = (
...
       ) ?? []
     }
-    return validateValue(elemID, value, new ListType(type))
+    return validateValue(elemID, value, new ListType(type), validatedReferences)
   }
 
   if (isReferenceExpression(value)) {
-    return isElement(value.value) ? [] : validateValue(elemID, value.value, type)
+    if (!isElement(value.value)
+      // We check for both validatedReferences and isParentOf because if we got to
```
ori-moisis, line 500: the comment doesn't match the code

packages/adapter-api/src/comparison.ts
```diff
@@ -0,0 +1,110 @@
...
+): boolean => Boolean(options?.compareByValue)
+  && shouldResolve(first)
+  && shouldResolve(second)
+
+
+const compareSpecialValuesWithCircularRefs = (
+  first: Value,
+  second: Value,
+  sourceComparedReferences: Set<string>,
+  targetComparedReferences: Set<string>,
```
ori-moisis, line 45: I think we should keep the alignment of the names, if we keep the values named `first` and `second`, we should call these sets `first...` and `second...` accordingly

would probably be best to align with the existing names from the other function
```
  firstVisitedReferences,
  secondVisitedReferences,
```
or replace both to something else that still has `first` and `second`

or align the other way and replace all occurrences of `first` and `second` with `source` and `target`

ori-moisis (approved):

PR #4707, opened 2023-08-07
WIP Salto 4567/support jira condition references

oneYAD:

---

---
_Release Notes_: 

---
_User Notifications_: 


packages/workato-adapter/src/filters/cross_service/recipe_conditions.ts
```diff
@@ -0,0 +1,107 @@
...
+  lhsToDependencies: dependencyFinder,
+): FormulaReferenceFinder => {
+  const addConditionMatchReferences = (
+    condition: (lhs: string, rhs: string) => boolean,
+    conditionBlock: ConditionBlock,
+    path: ElemID,
+    direction: DependencyDirection | undefined,
+  ): MappedReference[] => {
+    const a = lhsToDependencies(conditionBlock.lhs)
+      .filter(({ key }) => condition(key, conditionBlock.rhs)) // TODO what to do in case of rhs-lhs replcement
```
alonstern, line 71: I am not sure I understand this condition, what are you comparing here? e.g., if the lhs is 
`"#{_('data.jira.f7c7c7f5.fields.project.key')}"` and the rhs is `PROJ` and the operand is `equals_to` ?

packages/workato-adapter/src/filters/cross_service/recipe_conditions.ts
```diff
@@ -0,0 +1,107 @@
...
+  operand: Joi.string().valid(...supportedOperands).required(), // TODO check more options
+  lhs: Joi.string().required(),
+  rhs: Joi.string(), // The rhs is not required for operand like 'is present'
+}).unknown(true).required()
+
+
+const CONDITIONS_FULL_BLOCK_SCHEMA = Joi.object({
+  keyword: Joi.string().valid('if').required(), // TODO check more options
+  input: Joi.object({
+    operand: Joi.string().valid('and', 'or').required(),
```
alonstern, line 50: The operan in the input must be `'and'` or `'or'`? Can't it start with `equals_to`?

packages/workato-adapter/src/filters/cross_service/recipe_conditions.ts
oneYAD: you should look for all the lhsToDependencies. for every schema lhs. For your example, lhsToDependencies will return all projects keyd by project.key. so, for that example I will check the equality of each project key with PROJ.

packages/workato-adapter/src/filters/cross_service/recipe_conditions.ts
oneYAD: there is operand of the full conditions block and operand of each condition struct. the first one is 'and' or 'or'. the second one could be ~15 options

packages/workato-adapter/src/filters/cross_service/recipe_conditions.ts
oneYAD: but how could we do it otherwise? we want to know which project/issueType met the condition.

alonstern: For equals_to we can index the projects, for contains we probably can, but I am wondering if we should support that. (e.g., if there is an empty string, we are going to create a reference to every project in the account)

alonstern: Workato doesn't check for all the fields/projects in Jira. For each run of the recipe, it gets one project and checks it matches the condition. We are going to go over all the conditions in all the recipes and compare each one of them to all the projects/fields/statuses. So I think we should try to optimize that if we can

oneYAD: how could I check customer recipes? someone showed me this - https://www.notion.so/saltoio/In-Review-Guide-to-copy-client-workspacesFolders-to-staging-fe74a06091894b6d86e4ea2d03ae826b once. Is this the best way to do it?

alonstern: I think so

oneYAD: OK, thanks

PR #4709, opened 2023-08-07
SALTO-4177 Support fetch Issue Layout in Jira 

idoamit8:
This PR supports issueLayout fetch via a private API

---
#### To check the code manually:
1. Go to your jira's account.
2. Pick a project. 
3. go to setting -> issues -> Layout.
4. add\delete an attribute from the issue layout .
5. fetch the env. 

#### Picture of the result:
image

#### More info:
Support deploy in a following PR. 

---
_Release Notes_: 
Now support fetch of issue Layout. 

---
_User Notifications_: 
Now support fetch of issue Layout.

### DIFF
--- a/packages/jira-adapter/src/client/client.ts
+++ b/packages/jira-adapter/src/client/client.ts
@@ -15,6 +15,8 @@
 */
 import { client as clientUtils } from '@salto-io/adapter-components'
 import { logger } from '@salto-io/logging'
+import Joi from 'joi'
+import { createSchemeGuard } from '@salto-io/adapter-utils'
 import { handleDeploymentErrors } from '../deployment/deployment_error_handling'
 import { createConnection } from './connection'
 import { JIRA } from '../constants'
@@ -37,6 +39,18 @@ const DEFAULT_MAX_CONCURRENT_API_REQUESTS: Required<clientUtils.ClientRateLimitC
 const DEFAULT_PAGE_SIZE: Required<clientUtils.ClientPageSizeConfig> = {
   get: 1000,
 }
+
+export type graphQLResponseType = {
+  data: unknown
+  errors?: unknown
+}
+
+const GRAPHQL_RESPONSE_SCHEME = Joi.object({
+  data: Joi.object().required(),
+  errors: Joi.object().optional(),
+}).unknown(true).required()
+
+const isGraphQLResponse = createSchemeGuard<graphQLResponseType>(GRAPHQL_RESPONSE_SCHEME, 'Failed to get graphql response')
 
 export default class JiraClient extends clientUtils.AdapterHTTPClient<
   Credentials, clientUtils.ClientRateLimitConfig
@@ -119,8 +133,8 @@ export default class JiraClient extends clientUtils.AdapterHTTPClient<
   // Sends a post request to Jira with GQL body
   public async gqlPost(
     args: {url: string; query: string; variables?: Record<string, unknown> },
-  ): Promise<clientUtils.Response<clientUtils.ResponseValue | clientUtils.ResponseValue[]>> {
-    return this.post({
+  ): Promise<graphQLResponseType> {
+    const response = await this.post({
       url: args.url,
       data: {
         query: args.query,
@@ -128,6 +142,10 @@ export default class JiraClient extends clientUtils.AdapterHTTPClient<
       },
       headers: PRIVATE_API_HEADERS,
     })
+    if (isGraphQLResponse(response.data)) {
+      return response.data
+    }
+    throw new Error('Failed to get issue layout response')
   }
 
   public async getPrivate(
--- a/packages/jira-adapter/src/client/client.ts
+++ b/packages/jira-adapter/src/client/client.ts
@@ -20,7 +20,7 @@ import { createConnection } from './connection'
 import { JIRA } from '../constants'
 import { Credentials } from '../auth'
 import { getProductSettings } from '../product_settings'
-import { JSP_API_HEADERS, PRIVATE_API_HEADERS } from './headers'
+import { JIRA_GQL_HEADERS, JSP_API_HEADERS, PRIVATE_API_HEADERS } from './headers'
 
 const log = logger(module)
 
@@ -112,6 +112,22 @@ export default class JiraClient extends clientUtils.AdapterHTTPClient<
       headers: {
         ...JSP_API_HEADERS,
         ...(args.headers ?? {}),
+      },
+    })
+  }
+
+  // Sends a post request to Jira with GQL body
+  public async gqlPost(
+    args: {url: string; query: string; variables?: Record<string, unknown> },
+  ): Promise<clientUtils.Response<clientUtils.ResponseValue | clientUtils.ResponseValue[]>> {
+    return this.post({
+      ...args,
+      data: {
+        query: args.query,
+        variables: args.variables,
+      },
+      headers: {
+        ...JIRA_GQL_HEADERS,
       },
     })
   }
--- a/packages/jira-adapter/src/client/client.ts
+++ b/packages/jira-adapter/src/client/client.ts
@@ -20,7 +20,7 @@ import { createConnection } from './connection'
 import { JIRA } from '../constants'
 import { Credentials } from '../auth'
 import { getProductSettings } from '../product_settings'
-import { JIRA_GQL_HEADERS, JSP_API_HEADERS, PRIVATE_API_HEADERS } from './headers'
+import { JSP_API_HEADERS, PRIVATE_API_HEADERS } from './headers'
 
 const log = logger(module)
 
@@ -121,14 +121,12 @@ export default class JiraClient extends clientUtils.AdapterHTTPClient<
     args: {url: string; query: string; variables?: Record<string, unknown> },
   ): Promise<clientUtils.Response<clientUtils.ResponseValue | clientUtils.ResponseValue[]>> {
     return this.post({
-      ...args,
+      url: args.url,
       data: {
         query: args.query,
         variables: args.variables,
       },
-      headers: {
-        ...JIRA_GQL_HEADERS,
-      },
+      headers: PRIVATE_API_HEADERS,
     })
   }
 
--- a/packages/jira-adapter/src/client/client.ts
+++ b/packages/jira-adapter/src/client/client.ts
@@ -46,8 +46,8 @@ export type graphQLResponseType = {
 }
 
 const GRAPHQL_RESPONSE_SCHEME = Joi.object({
-  data: Joi.object().required(),
-  errors: Joi.object().optional(),
+  data: Joi.required(),
+  errors: Joi.optional(),
 }).unknown(true).required()
 
 const isGraphQLResponse = createSchemeGuard<graphQLResponseType>(GRAPHQL_RESPONSE_SCHEME, 'Failed to get graphql response')
--- a/packages/jira-adapter/src/filters/issue_layout/issue_layout.ts
+++ b/packages/jira-adapter/src/filters/issue_layout/issue_layout.ts
@@ -23,7 +23,7 @@ import JiraClient from '../../client/client'
 import { ISSUE_LAYOUT_TYPE, JIRA, PROJECT_TYPE } from '../../constants'
 import { FilterCreator } from '../../filter'
 import { QUERY } from './issue_layout_query'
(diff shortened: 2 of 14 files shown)


alonstern:
We need to add references to the NaCl, from the image you sent it looks like
projected -> reference to a project
extraDefinerId -> reference to a screen (IIRC)
owner -> list of references to IssueTypes
key -> reference to a field

This can be either here or in a separated PR, whatever you prefer

packages/jira-adapter/src/filters/issue_layout/issue_layout_types.ts
```diff
@@ -0,0 +1,121 @@
...
+})
+
+export const issueLayoutConfigType = new ObjectType({
+  elemID: new ElemID(JIRA, 'issueLayoutConfig'),
+  fields: {
+    items: { refType: new ListType(issueLayoutItemsType) },
+  },
+})
+
+export const ISSUE_LAYOUT_SUB_TYPES = [
```
alonstern, line 55: Generally it's better to not create mutable objects at the module level since it makes it harder to test them (one test that edits them can affect another). You can check how we do it for automation https://github.com/salto-io/salto/blob/main/packages/jira-adapter/src/filters/automation/types.ts

packages/jira-adapter/src/filters/issue_layout/issue_layout.ts
```diff
@@ -0,0 +1,216 @@
...
+          const issueLayout = new InstanceElement(
+            naclCase(name),
+            issueLayoutType,
+            {
+              projectId,
+              extraDefinerId: screenId,
+              owners: fromResponseLayoutOwnersToLayoutOwners(issueLayoutResult.usageInfo.edges[0].node.layoutOwners),
+              issueLayoutConfig: fromIssueLayoutConfigRespToIssueLayoutConfig(containers),
+            },
+            [JIRA, adapterElements.RECORDS_PATH, ISSUE_LAYOUT_TYPE, pathNaclCase(name)],
```
alonstern, line 204: Let's put it under the project folder under a "layouts" folder

packages/jira-adapter/src/filters/issue_layout/issue_layout.ts
```diff
@@ -0,0 +1,216 @@
...
+    key: node.fieldItemId || node.panelItemId || '',
+  })))
+  return { items }
+}
+const createReferences = async (config: JiraConfig, elements: Element[]): Promise<void> => {
+  const fixedDefs = referencesRules
+    .map(def => (
+      config.fetch.enableMissingReferences ? def : _.omit(def, 'jiraMissingRefStrategy')
+    ))
+  await referenceUtils.addReferences({
```
alonstern, line 142: This is problematic performance-wise. I think we would want to scan for references only the issuelayout instances, but pass all the elements as context for the target of the references. Let's talk F2F about it

alonstern (commented):
🔥

packages/jira-adapter/src/client/client.ts
```diff
@@ -38,6 +40,18 @@ const DEFAULT_PAGE_SIZE: Required<clientUtils.ClientPageSizeConfig> = {
...
 }
 
+export type graphQLResponseType = {
+  data: unknown
+  errors?: unknown
+}
+
+const GRAPHQL_RESPONSE_SCHEME = Joi.object({
+  data: Joi.object().required(),
+  errors: Joi.object().optional(),
```
alonstern, line 50: since it is not necessarily an object
```suggestion
  data: Joi.required(),
  errors: Joi.optional(),
```

packages/jira-adapter/src/filters/issue_layout/issue_layout.ts
```diff
@@ -0,0 +1,166 @@
...
+
+const fromIssueLayoutConfigRespToIssueLayoutConfig = (
+  containers: containerIssueLayoutResponse[]
+):
+IssueLayoutConfig => {
+  const items = containers.flatMap(container => container.items.nodes.map(node => ({
+    type: 'FIELD',
+    sectionType: container.containerType,
+    key: node.fieldItemId,
+  }))).filter(isDefined).filter(isIssueLayoutConfigItem)
```
alonstern, line 73: Won't it always be defined and according to the schema? since you build this type right before

packages/jira-adapter/src/filters/issue_layout/issue_layout.ts
```diff
@@ -0,0 +1,166 @@
...
+const createReferences = async (
+  config: JiraConfig, elements: Element[], contextElements: Element[]): Promise<void> => {
+  const fixedDefs = referencesRules
+    .map(def => (
+      config.fetch.enableMissingReferences ? def : _.omit(def, 'jiraMissingRefStrategy')
+    ))
+  await referenceUtils.addReferences({
+    elements,
+    contextElements,
+    fieldsToGroupBy: ['id', 'name', 'originalName', 'groupId', 'key'],
```
alonstern, line 87: I think you don't need here anything besides 'id' right?

packages/jira-adapter/src/filters/issue_layout/issue_layout.ts
```diff
@@ -0,0 +1,166 @@
...
+        })))
+        .filter(isDefined)
+    )
+    const { issueLayoutType, subTypes } = createIssueLayoutType()
+    elements.push(issueLayoutType)
+    subTypes.forEach(type => elements.push(type))
+
+    await Promise.all(Object.entries(projectToScreenId)
+      .flatMap(([projectId, screenIds]) => screenIds.map(async screenId => {
+        const response = await getIssueLayout({
```
alonstern, line 134: Let's add a try-catch on this to make sure a fail here won't fail the entire fetch

packages/jira-adapter/src/filters/issue_layout/issue_layout_query.ts
```diff
@@ -0,0 +1,125 @@
...
+  fragment FieldItemBaseFragment on JiraIssueLayoutFieldItemConfiguration {
+    fieldItemId
+    key
+    name
+    type
+  }
+  
+    fragment FieldItemFragment on JiraIssueLayoutFieldItemConfiguration {
+      ...FieldItemBaseFragment
+      properties(keys: $fieldPropertyKeys)
```
alonstern, line 93: fieldPropertyKeys is an empty array so can we remove the properties here and the `fieldPropertyKeys` variable?

packages/jira-adapter/src/filters/issue_layout/issue_layout_query.ts
```diff
@@ -0,0 +1,125 @@
...
+    }
+  }
+  
+  fragment CMPJSWLayoutConfigurationFragment on Query {
+    issueLayoutConfiguration(issueLayoutKey: {projectId: $projectId, extraDefinerId: $extraDefinerId}, type: ISSUE_VIEW) {
+      ... on JiraIssueLayoutConfigurationResult {
+        issueLayoutResult {
+          id
+          name
+          usageInfo {
```
alonstern, line 111: Do you need the `usageInfo` and `containers`?

packages/jira-adapter/src/filters/issue_layout/issue_layout.ts
```diff
@@ -0,0 +1,166 @@
...
+        const screens = screenSchemes.map(screenScheme => screenScheme.value.screens.default.value.value.id)
+        return [project.value.id, screens]
+      })))
+  )
+  return projectToScreenId
+}
+
+const filter: FilterCreator = ({ client, config }) => ({
+  name: 'issueLayoutFilter',
+  onFetch: async elements => {
```
alonstern, line 114: You should also check here that the type was not excluded in the config like we do for automations https://github.com/salto-io/salto/blob/main/packages/jira-adapter/src/filters/automation/automation_fetch.ts#L136

Please test that it's possible to exclude it in the config.
I think until the deploy is finished and merged with would want it to not be fetched by default

alonstern (commented):
🔥

packages/jira-adapter/src/filters/issue_layout/issue_layout.ts
idoamit8: It will always be defined. But I need to keep `.filter(isIssueLayoutConfigItem)` so I can return it (otherwise the type is 'any' and not 'IssueLayoutConfig'

packages/jira-adapter/src/filters/issue_layout/issue_layout_query.ts
idoamit8: Yes, both of them.

packages/jira-adapter/src/filters/issue_layout/issue_layout.ts
idoamit8: I added this check,  and I am planning to merge the deploy and the fetch together,

packages/jira-adapter/src/filters/issue_layout/issue_layout_query.ts
alonstern: What about `metadata`? I don't see it is used

packages/jira-adapter/src/filters/issue_layout/issue_layout.ts
alonstern: It will make the deploy a bit harder to review. I think you can just comment out the usage of the filter in adapter.ts before merging. BTW did you check that excluding this type in the configuration works? I remember adding the type to some list in the api_config is needed but not sure

packages/jira-adapter/src/filters/issue_layout/issue_layout.ts
```diff
@@ -0,0 +1,170 @@
...
+          projectId: Number(projectId),
+          screenId,
+          client,
+        })
+        if (!Array.isArray(response.data) && isIssueLayoutResponse(response.data)) {
+          const { issueLayoutResult } = response.data.issueLayoutConfiguration
+          const { containers } = issueLayoutResult
+          const name = `${projectIdToProjectName[projectId]}_${issueLayoutResult.name}`
+          const issueLayout = new InstanceElement(
+            naclCase(name),
```
alonstern, line 148: Another thing I forgot. We need to create the instance name using `ElemIdGetter` like we do in https://github.com/salto-io/salto/blob/main/packages/jira-adapter/src/filters/automation/automation_fetch.ts/#L100

packages/jira-adapter/src/filters/issue_layout/issue_layout_query.ts
```diff
@@ -0,0 +1,124 @@
...
+* limitations under the License.
+*/
+export const QUERY = `query SwiftJswCmpInitial($projectId: Long!, $extraDefinerId: Long!, $requestOwnerPropertyKeys: [String!] = []) {
+    ...CMPJSWLayoutConfigurationFragment
+  }
+  
+  
+  
+  fragment JiraIssueLayoutOwnerFragment on JiraIssueLayoutOwner {
+    ... on JiraIssueLayoutIssueTypeOwner {
```
alonstern, line 23: Do we need to pull all the extra properties besides id? (name description etc?) (relevant to some other places in this file)

alonstern (commented):
👌 Added some last comment and I think I will approved afterward

packages/jira-adapter/src/filters/issue_layout/issue_layout_query.ts
idoamit8: Now that I have references, I do not need the metadata :)

packages/jira-adapter/src/filters/issue_layout/issue_layout_query.ts
idoamit8: I removed everything that isn't used.

packages/jira-adapter/src/filters/issue_layout/issue_layout.ts
idoamit8: We can exclude this type if I add this to the exclude part in jira.nacl file. This is what you asked?

packages/jira-adapter/src/filters/issue_layout/issue_layout_query.ts
```diff
@@ -0,0 +1,74 @@
...
+    ...CMPJSWLayoutConfigurationFragment
+  }
+  
+  fragment JiraIssueLayoutOwnerFragment on JiraIssueLayoutOwner {
+    ... on JiraIssueLayoutIssueTypeOwner {
+      id
+    }
+    ... on JiraIssueLayoutRequestTypeOwner {
+      id
+      properties(keys: $requestOwnerPropertyKeys)
```
alonstern, line 26: Can we also remove the properties and `requestOwnerPropertyKeys`?

alonstern (approved):
👌 
Added one small comment
make sure to comment out the filter that fetches before you merge

PR #4711, opened 2023-08-08
SALTO-4462 change post fetch params into a single map

kami-delaroz:
SALTO-4462 change post fetch params into a single map

---

[_Additional context for reviewer_](https://github.com/salto-io/salto/pull/4565)

---
_Release Notes_: 
None

---
_User Notifications_: 
None


adamro:
Did you look at all the functions in the fetch flow that have similar input to see if they can also benefit from something similar (and possibly need a different type to support all)? If not, can you please do? :)

kami-delaroz:
@adamro fetchChanges receives accountsToAdapters and accountToServiceNameMap. But I'm not sure it worth the change

packages/core/src/core/fetch.ts
```diff
@@ -319,48 +319,46 @@ const isAdapterOperationsWithPostFetch = (
...
-  const elementsByAccount = Object.fromEntries(
-    [...new Set([
-      ...Object.keys(stateElementsByAccount),
-      ...Object.keys(serviceElementsByAccount),
-    ])].map(accountName => [accountName, getAdapterElements(accountName)])
-  )
+  const elementsByAccount = Object.fromEntries(Object.keys(accountInputs)
+    .map(accountName => [accountName, getAdapterElements(accountName)]))
   // only modifies elements in-place, done sequentially to avoid race conditions
+  const accountToServiceNameMap = Object.fromEntries(Object.entries(accountInputs).map(([k, v]) => [k, v.adapterName]))
```
netama, line 357: if we need to pass this further I think we should just pass it in rather than reconstruct IMO. especially since nothing in this function looks at the `adapterName`, we can just keep it completely separate AFAICT

packages/core/src/core/fetch.ts
```diff
@@ -497,16 +495,25 @@ const fetchAndProcessMergeErrors = async (
       await stateElements.getAll(),
       elem => elem.elemID.adapter
     )
+    const serviceElementsByAccount = _.groupBy(accountElements, e => e.elemID.adapter)
+    const accountInputs = Object.keys(accountToServiceNameMap)
+      .reduce((acc: Record<string, AccountInput>, key: string) => {
```
netama, line 500: nit: is there a need to use `reduce` here? can we use something "simpler" instead like `_.mapValues`?

packages/core/src/core/fetch.ts
```diff
@@ -319,48 +319,46 @@ const isAdapterOperationsWithPostFetch = (
...
-  partiallyFetchedAccountData: Map<string, PartiallyFetchedAccountData>
-  accountToServiceNameMap: Record<string, string>
+  accountInputs: Record<string, AccountInput>
   progressReporters: Record<string, ProgressReporter>
 }): Promise<void> => {
-  const serviceElementsByAccount = _.groupBy(accountElements, e => e.elemID.adapter)
   const getAdapterElements = (accountName: string): ReadonlyArray<Element> => {
-    if (!partiallyFetchedAccountData.has(accountName)) {
-      return serviceElementsByAccount[accountName] ?? stateElementsByAccount[accountName]
+    const accountInput = accountInputs[accountName]
```
netama, line 339: nit: if you destructure the values here (`const { stateElements, partiallyFetchedAccountData ...} = accountInputs[accountName]` ) you can save some `?.` checks later (e.g. since we have an early return for `partiallyFetchedAccountData === undefined`, we know it is defined if we continue)

netama (commented):
cool :) added some comments, but up to you and @adamro , I don't have a strong opinion on this

PR #4712, opened 2023-08-08
SALTO-4582: Update permissions table

Shulik95:
_Updated hardcoded permission table to contain missing permission and simplified script_

---

_None_

---
_Release Notes_: 
_Netsuite Adapter_:
* Permission change validator covers permissions which do not appear in NS documentation

---
_User Notifications_: 
_None_


omrilit (approved):
Thanks!

PR #4715, opened 2023-08-09
SALTO-4138: Added a check for unique scriptid of custom fields in states within workflows

SeggevHaimovich:
_Added states of workflows to the unique fields change validator_

---

_Incident: https://salto-io.atlassian.net/browse/INCIDENT-7262?focusedCommentId=95306&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-95306_

---
_Release Notes_: 
_Users will get an error when trying to deploy a workflow which contains a state with a custom field that has a non-unique scriptid_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/netsuite-adapter/src/change_validators/unique_fields.ts
+++ b/packages/netsuite-adapter/src/change_validators/unique_fields.ts
@@ -83,9 +83,9 @@ const getNestedScriptIDField = async (
   _.values(await elementsSource.get(elemID.createNestedID(...nestPath)))
     .map(val => val[SCRIPT_ID])
 
-const getWorkflowFields = (elem: Value): string[] => {
-  const customFieldsScriptid = _.values(_.get(elem.value, WORKFLOW_CUSTOM_FIELDS_PATH)).map(val => val[SCRIPT_ID])
-  const stateFieldsScriptid = _.values(_.get(elem.value, WORKFLOW_STATES_PATH))
+const getWorkflowFields = (elem: ChangeDataType): string[] => {
+  const customFieldsScriptid = getChangeNestedScriptIDField(elem, WORKFLOW_CUSTOM_FIELDS_PATH)
+  const stateFieldsScriptid = _.values(getChangeNestedField(elem, WORKFLOW_STATES_PATH))
     .flatMap(state => _.values(_.get(state, WORKFLOW_STATES_CUSTOM_FIELDS_PATH)))
     .map(val => val[SCRIPT_ID])
   return customFieldsScriptid.concat(stateFieldsScriptid)
--- a/packages/netsuite-adapter/src/change_validators/unique_fields.ts
+++ b/packages/netsuite-adapter/src/change_validators/unique_fields.ts
@@ -28,7 +28,9 @@ const { mapValuesAsync } = promises.object
 
 
 const FIELD_DEFAULT_NAME = 'FIELD_DEFAULT_NAME'
-const WORKFLOW_RESTRICTED_PATH = ['workflowcustomfields', 'workflowcustomfield']
+const WORKFLOW_CUSTOM_FIELDS_PATH = ['workflowcustomfields', 'workflowcustomfield']
+const WORKFLOW_STATES_PATH = ['workflowstates', 'workflowstate']
+const WORKFLOW_STATES_CUSTOM_FIELDS_PATH = ['workflowstatecustomfields', 'workflowstatecustomfield']
 const SCRIPT_RESTRICTED_PATH = ['scriptcustomfields', 'scriptcustomfield']
 
 type RestrictedType = 'savedSearch' | 'financialLayout' | 'customRecordField' | 'workflow' | 'script'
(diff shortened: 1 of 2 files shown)


packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -81,6 +83,28 @@ const getNestedScriptIDField = async (
...
+  [..._.values(getChangeNestedField(change, WORKFLOW_CUSTOM_FIELDS_PATH)).map(val => val[SCRIPT_ID]),
+    ..._.values(getChangeNestedField(change, WORKFLOW_STATES_PATH))
+      .flatMap(state => _.values(_.get(state, WORKFLOW_STATES_CUSTOM_FIELDS_PATH)))
+      .map(custfield => custfield[SCRIPT_ID])]
+
+const getSourceWorkflowFields = async (
+  { elemID, elementsSource }: GetterParams
+): Promise<string[]> =>
+  // custom fields and states' custom fields must have a different prefix: "custworkflow" and "custwfstate"
+  _.values(await elementsSource.get(elemID))
```
dantal4, line 99: `elemID` is a top-level instance id, so `elementsSource.get(elemID)` returns the whole instance. don't we need to take `instance.value` instead of `_.values(instance)` ?

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -81,6 +83,28 @@ const getNestedScriptIDField = async (
...
+    .map(elem => ({
+      customFieldsScriptid: _.values(_.get(elem, WORKFLOW_CUSTOM_FIELDS_PATH)).map(val => val[SCRIPT_ID]),
+      stateFieldsScriptid: _.values(_.get(elem, WORKFLOW_STATES_PATH))
+        .flatMap(state => _.values(_.get(state, WORKFLOW_STATES_CUSTOM_FIELDS_PATH)))
+        .map(val => val[SCRIPT_ID]),
+    }))
+    .flatMap(({ customFieldsScriptid, stateFieldsScriptid }) => [...customFieldsScriptid, ...stateFieldsScriptid])
```
dantal4, line 106: I think that it would be more readable to have one function like this:
```
const customFieldsScriptid = ....
const stateFieldsScriptid = ...
...
return customFieldsScriptid.concat(stateFieldsScriptid)
```

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -81,6 +83,28 @@ const getNestedScriptIDField = async (
...
+  // custom fields and states' custom fields must have a different prefix: "custworkflow" and "custwfstate"
+  [..._.values(getChangeNestedField(change, WORKFLOW_CUSTOM_FIELDS_PATH)).map(val => val[SCRIPT_ID]),
+    ..._.values(getChangeNestedField(change, WORKFLOW_STATES_PATH))
+      .flatMap(state => _.values(_.get(state, WORKFLOW_STATES_CUSTOM_FIELDS_PATH)))
+      .map(custfield => custfield[SCRIPT_ID])]
+
+const getSourceWorkflowFields = async (
+  { elemID, elementsSource }: GetterParams
+): Promise<string[]> =>
+  // custom fields and states' custom fields must have a different prefix: "custworkflow" and "custwfstate"
```
dantal4, line 98: I think that this comment should be written once, maybe in `workflowGetters`, and with more context (something like "it is fine to have one list for custom/state fields because...")

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -81,6 +83,28 @@ const getNestedScriptIDField = async (
...
     .map(val => val[SCRIPT_ID])
 
+const getChangeWorkflowFields = (
+  change: ChangeDataType,
+): string[] =>
+  // custom fields and states' custom fields must have a different prefix: "custworkflow" and "custwfstate"
+  [..._.values(getChangeNestedField(change, WORKFLOW_CUSTOM_FIELDS_PATH)).map(val => val[SCRIPT_ID]),
+    ..._.values(getChangeNestedField(change, WORKFLOW_STATES_PATH))
+      .flatMap(state => _.values(_.get(state, WORKFLOW_STATES_CUSTOM_FIELDS_PATH)))
+      .map(custfield => custfield[SCRIPT_ID])]
```
dantal4, line 93: can you prettify this function? it's hard to understand what do we get here...

packages/netsuite-adapter/test/change_validators/unique_fields.test.ts
```diff
@@ -95,6 +116,24 @@ describe('unique fields validator', () => {
...
+  const getWorkflowStatesElements = (): TestElements => getOtherTestElements(WORKFLOW,
+    (name: string, elemID: ObjectType, uniqueField: string) =>
+      new InstanceElement(name, elemID, {
+        workflowstates: {
+          workflowstate: {
+            workflowstate1: {
+              workflowstatecustomfields: {
+                workflowstatecustomfield: {
+                  custwfstate1: {
+                    scriptid: uniqueField,
```
dantal4, line 128: nit - instead of creating different instances for workflow with state fields, have both custom fields and state fields in the same instances - with `custworkflow${uniqueField}` for custom fields and `custwfstate${uniqueField}` for state fields.

dantal4 (commented):
Great work!

packages/netsuite-adapter/test/change_validators/unique_fields.test.ts
SeggevHaimovich: This will make the tests check both of the fields without the ability to check each one separately.
Should I change it despite this?

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -81,6 +83,14 @@ const getNestedScriptIDField = async (
   _.values(await elementsSource.get(elemID.createNestedID(...nestPath)))
     .map(val => val[SCRIPT_ID])
 
+const getWorkflowFields = (elem: Value): string[] => {
```
dantal4, line 86: `Value` is not type-safe. why can't this be `ChangeDataType`?

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -81,6 +83,14 @@ const getNestedScriptIDField = async (
...
+  const customFieldsScriptid = _.values(_.get(elem.value, WORKFLOW_CUSTOM_FIELDS_PATH)).map(val => val[SCRIPT_ID])
```
dantal4, line 87: can be replaced with `getChangeNestedScriptIDField` ?

packages/netsuite-adapter/src/change_validators/unique_fields.ts
```diff
@@ -81,6 +83,14 @@ const getNestedScriptIDField = async (
...
+  const stateFieldsScriptid = _.values(_.get(elem.value, WORKFLOW_STATES_PATH))
```
dantal4, line 88: can be replaced with `getChangeNestedField` ?

dantal4 (commented):
Good work! Added some comments about code sharing

packages/netsuite-adapter/test/change_validators/unique_fields.test.ts
dantal4: I think that a workflow with both custom fields & state fields is a common case that should be also tested. I think that we'll gain the same "code guard" as testing a workflow with only one kind, and we'll gain another case that separate tests won't give us.

packages/netsuite-adapter/test/change_validators/unique_fields.test.ts
```diff
@@ -331,45 +344,42 @@ describe('unique fields validator', () => {
...
   })
 
-  describe('Workflow custom field unique `scriptid` field validator', () => {
+  describe("Workflow and states' custom field unique `scriptid` field validator", () => {
     const testElements = getWorkflowElements()
-    const idToVal = getIDToVal(testElements, 'workflowcustomfields', 'workflowcustomfield')
-    const buildElementsSource = (elements: readonly Element[]): ReadOnlyElementsSource =>
-      buildMockElementsSource(buildElementsSourceFromElements(elements), idToVal)
-
+    // const testElementsStates = getWorkflowStatesElements()
```
dantal4, line 349: leftover?

packages/netsuite-adapter/test/change_validators/unique_fields.test.ts
```diff
@@ -539,14 +554,14 @@ describe('unique fields validator', () => {
           toChange({ after: savedSearchTestInstances.basic }),
           toChange({ after: financialLayoutTestInstances.basic }),
           toChange({ after: customRecordTestChanges.basic }),
-          toChange({ after: workflowTestInstances.basic }),
+          toChange({ after: workflowCustomFieldTestInstances.basic }),
```
dantal4, line 557: not sure if that rename is necessary

dantal4 (approved):
💯

PR #4716, opened 2023-08-09
SAAS-8197 define dependency type for errors

benisraeladi:
Add 'dependency' to new `SaltoErrorType` to determine whether an error is a dependency error

---

---
_Release Notes_: _None_

---
_User Notifications_: _None_

### DIFF
--- a/packages/adapter-api/src/error.ts
+++ b/packages/adapter-api/src/error.ts
@@ -18,12 +18,12 @@ import { ElemID } from './element_id'
 
 export type SeverityLevel = 'Error' | 'Warning' | 'Info'
 
-export type SaltoErrorSource = 'config'
+export type SaltoErrorType = 'config' | 'dependency'
 
 export type SaltoError = {
     message: string
     severity: SeverityLevel
-    source?: SaltoErrorSource
+    type?: SaltoErrorType
 }
 
 export type SaltoElementError = SaltoError & {
--- a/packages/core/src/core/deploy/deploy_actions.ts
+++ b/packages/core/src/core/deploy/deploy_actions.ts
@@ -18,7 +18,7 @@ import {
   AdapterOperations, getChangeData, Change,
   isAdditionOrModificationChange,
   DeployExtraProperties, DeployOptions, Group,
-  SaltoElementError, SaltoError, SeverityLevel, DeployResult, ChangeDataType,
+  SaltoElementError, SaltoError, SeverityLevel, DeployResult, ChangeDataType, SaltoErrorType,
 } from '@salto-io/adapter-api'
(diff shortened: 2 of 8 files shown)


packages/adapter-api/src/error.ts
```diff
@@ -18,12 +18,12 @@ import { ElemID } from './element_id'
...
 export type SeverityLevel = 'Error' | 'Warning' | 'Info'
 
-export type SaltoErrorSource = 'config'
+export type SaltoErrorType = 'config' | 'dependency'
 
 export type SaltoError = {
     message: string
     severity: SeverityLevel
-    source?: SaltoErrorSource
+    type?: SaltoErrorType
```
amitgoldarb, line 26: This is a breaking change right? This means that all indexed workspaces with the earlier values will receive a "wrong" value post migration.
@ori-moisis Do you think it's better to do this in 2 phases?

ori-moisis: we could add a bit of migration code that would read `source` into `type` on deserialize

ori-moisis: actually, it seems like `source` was never actually serialized to rocksdb...
it was only ever added [here](https://github.com/salto-io/salto/pull/4716/files#diff-53d2bb7830fad95fe7aa1b03dab5634f9f597a0933fecec316e486ea67dc064eL181)

benisraeladi: @ori-moisis so do you think it's important to do in two stages?

ori-moisis: what would be the stages and why would we need them? did I miss somewhere that the old value is saved?

benisraeladi: I don't see a need since the value is only used to decide on the error type and hasn't been saved. I just didn't understand from your answer whether you still had any doubts

amitgoldarb (approved):
LGTM!
Thanks for this PR.
Let's wait for another approval as well from @adamro / @ori-moisis, I think it's benificial we get another eye on a core api change.

packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -171,8 +171,9 @@ export const deployActions = async (
             ({
               elemID: getChangeData(change).elemID,
               groupId: item.groupKey,
-              message: `Element ${key} was not deployed, as it depends on element ${nodeError.causingNode} which failed to deploy`,
+              message: `Element was not deployed, as it depends on element ${nodeError.causingNode} which failed to deploy`,
```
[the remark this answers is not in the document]
benisraeladi: True, must have missed it before

adamro (approved):

PR #4717, opened 2023-08-09
SALTO-4594 fix the restore flow for modify instance change

Uri-Bar:
SALTO-4594 fix the restore flow for modify instance change

before this fix modify changes in instances return without data as it was excluded in the new `filterByPathHint` func.
the tests did not caught this bug as it did not checked the data of the changes.

---

_Release Notes_: 
None

---
_User Notifications_: 
None

---

### DIFF
--- a/packages/core/src/core/restore.ts
+++ b/packages/core/src/core/restore.ts
@@ -31,6 +31,12 @@ const splitDetailedChangeByPath = async (
   if (_.isEmpty(changeHints) || isRemovalChange(change)) {
     return [change]
   }
+  if (changeHints.length === 1) {
+    return [{
+      ...change,
+      path: changeHints[0],
+    }]
+  }
   return Promise.all(changeHints.map(async hint => {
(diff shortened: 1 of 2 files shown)


packages/core/src/core/restore.ts
```diff
@@ -31,6 +31,12 @@ const splitDetailedChangeByPath = async (
   if (_.isEmpty(changeHints) || isRemovalChange(change)) {
     return [change]
   }
+  if (changeHints.length === 1) {
```
dantal4, line 34: IIUC, all nested changes should get into this condition-
1. when the nested value is in one file (which is the only used case)
2. when the nested value is over several files - `pathIndex.getFromPathIndex` will still return only one path hint (https://github.com/salto-io/salto/commit/c0305166a1500fef37a0dab495447b5b5dd60bc1)

dantal4 (approved):
LGTM

PR #4718, opened 2023-08-09
update sysInfo enum and schema

Shulik95:
_update sysInfo enum and schema_

---

_None_

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/netsuite-adapter/src/client/suiteapp_client/types.ts
+++ b/packages/netsuite-adapter/src/client/suiteapp_client/types.ts
@@ -123,7 +123,7 @@ export const SYSTEM_INFORMATION_SCHEME = {
       properties: {
         ...BASIC_SYSTEM_INFO_SCHEME_PROPERTIES,
         envType: {
-          type: 'number',
+          type: 'string',
           enum: [EnvType.PRODUCTION, EnvType.SANDBOX, EnvType.BETA, EnvType.INTERNAL],
         },
       },
@@ -135,11 +135,7 @@ export const SYSTEM_INFORMATION_SCHEME = {
         ...BASIC_SYSTEM_INFO_SCHEME_PROPERTIES,
       },
       required: ['time', 'appVersion'],
-      additionalProperties: {
-        not: {
-          required: ['envType'], // Ensure that envType is not present in this case
-        },
-      },
+      additionalProperties: false,
     },
   ],
   additionalProperties: true,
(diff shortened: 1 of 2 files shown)


packages/netsuite-adapter/src/client/suiteapp_client/types.ts
```diff
@@ -135,6 +135,7 @@ export const SYSTEM_INFORMATION_SCHEME = {
         ...BASIC_SYSTEM_INFO_SCHEME_PROPERTIES,
       },
       required: ['time', 'appVersion'],
+      additionalProperties: false,
```
shoshana-houri, line 138: We probably should add it to the other options as well.

shoshana-houri (commented):
Looks good 🥳 
Had 1 comment. I think the `additionalProperties: false,` is great but also a bit risky, so I would doube-check to validate we don't receive any other properties.

packages/netsuite-adapter/test/client/suiteapp_client/suiteapp_client.test.ts
```diff
@@ -54,7 +54,7 @@ describe('SuiteAppClient', () => {
       mockAxiosAdapter.onPost().replyOnce(200, {
         status: 'success',
         results: {
-          appVersion: [0, 1, 7],
+          appVersion: [0, 1, 9],
```
dantal4, line 57: Is there a reason for that? Does the suiteapp version going to change?

dantal4:
@Shulik95 can yoi explain why do we need to update the enum? What happened until now?

packages/netsuite-adapter/test/client/suiteapp_client/suiteapp_client.test.ts
Shulik95: Yes, we've previously deprecated version 0.1.7 since we had some bugs in it. I uploaded our suiteapp again as version 0.1.8 (since NetSuite doesn't allow reusing 0.1.7) and had to upload it again to choose the minimal NetSuite version compatible so we could upgrade as many users as possible to our newest suiteapp which leads us to version 0.1.9

Shulik95:
> @Shulik95 can you explain why do we need to update the enum? What happened until now?

The envType we got back from our suiteApp was a string and not a number as I thought in the beginning and the validation passed because the result with the wrong envType value fell under the second case in the `SYSTEM_INFORMATION_SCHEME`. So I've updated the enums and schema to match what we got from our suiteApp.

packages/netsuite-adapter/test/client/suiteapp_client/suiteapp_client.test.ts
dantal4: Ok, thanks. Can you open a PR in salto_suiteapp that upgrades the version there?

PR #4719, opened 2023-08-09
SALTO-4025: match files referenced in suiteScripts that don't start with path prefix

Shulik95:
_match files referenced in suiteScripts that don't start with path prefix_

---

_The ticket - https://salto-io.atlassian.net/browse/SALTO-4025
This PR might cause audit noise, need communication before a noisy deployment._

---
_Release Notes_: 
_Netsuite Adapter_:
* file references in suiteScripts that dont start with a path prefix will be added to generated dependencies

---
_User Notifications_: 
*_file references in suiteScripts that dont start with a path prefix will be added to generated dependencies_

### DIFF
--- a/packages/netsuite-adapter/src/filters/element_references.ts
+++ b/packages/netsuite-adapter/src/filters/element_references.ts
@@ -127,30 +127,23 @@ const generateServiceIdToElemID = async (
 const resolveRelativePath = (absolutePath: string, relativePath: string): string =>
   osPath.resolve(osPath.dirname(absolutePath), relativePath)
 
-const getFileNameFromPath = (path: string):string =>
-  path.split(FILE_CABINET_PATH_SEPARATOR)
-    .slice(-1)[0]
-    .split('.')[0]
-
 const getServiceElemIDsFromPaths = (
   foundReferences: string[],
   serviceIdToElemID: ServiceIdRecords,
   customRecordFieldsToServiceIds: ServiceIdRecords,
   element: InstanceElement,
-  fileNames: Set<string>,
+  // fileNames: Set<string>,
 ): ElemID[] =>
   foundReferences
     .flatMap(ref => {
-      if (fileNames.has(getFileNameFromPath(ref))) {
-        const prefixedRef = pathPrefixRegex.test(ref) ? ref : FILE_CABINET_PATH_SEPARATOR.concat(ref)
-        const absolutePath = resolveRelativePath(element.value[PATH], prefixedRef)
-        return [absolutePath].concat(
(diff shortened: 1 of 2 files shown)


packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -127,16 +127,23 @@ const generateServiceIdToElemID = async (
...
   element: InstanceElement,
+  fileNames: Set<string>,
 ): ElemID[] =>
   foundReferences
     .flatMap(ref => {
-      if (pathPrefixRegex.test(ref)) {
-        const absolutePath = resolveRelativePath(element.value[PATH], ref)
+      if (fileNames.has(getFileNameFromPath(ref))) {
+        const prefixedRef = pathPrefixRegex.test(ref) ? ref : FILE_CABINET_PATH_SEPARATOR.concat(ref)
+        const absolutePath = resolveRelativePath(element.value[PATH], prefixedRef)
```
dantal4, line 146: we don't need to resolve the relative path in the second case
```suggestion
        const absolutePath = pathPrefixRegex.test(ref) ? resolveRelativePath(element.value[PATH], ref) : FILE_CABINET_PATH_SEPARATOR.concat(ref)
```

packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -127,16 +127,23 @@ const generateServiceIdToElemID = async (
...
   serviceIdToElemID: ServiceIdRecords,
   customRecordFieldsToServiceIds: ServiceIdRecords,
   element: InstanceElement,
+  fileNames: Set<string>,
 ): ElemID[] =>
   foundReferences
     .flatMap(ref => {
-      if (pathPrefixRegex.test(ref)) {
-        const absolutePath = resolveRelativePath(element.value[PATH], ref)
+      if (fileNames.has(getFileNameFromPath(ref))) {
```
dantal4, line 144: I'm not sure that I understand why do we need that test... maybe we can always return `[ref, absolutePath, ubsolutePathWithExtension]` because we are looking later in `serviceIdToElemID` and if it's not there we just skip it.

dantal4 (commented):
looks great! had just one question ⬇️

packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -132,18 +132,18 @@ const getServiceElemIDsFromPaths = (
   serviceIdToElemID: ServiceIdRecords,
   customRecordFieldsToServiceIds: ServiceIdRecords,
   element: InstanceElement,
+  // fileNames: Set<string>,
```
dantal4, line 135: Leftover?

dantal4 (approved):
💯

PR #4721, opened 2023-08-10
SALTO-4025: log paths without path prefix instead of adding them

Shulik95:
_log paths without path prefix instead of adding them until communicated_

---

_we should log the references instead of adding them until PROD-125 is communicated_

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/netsuite-adapter/src/filters/element_references.ts
+++ b/packages/netsuite-adapter/src/filters/element_references.ts
@@ -135,9 +135,11 @@ const getServiceElemIDsFromPaths = (
 ): ElemID[] =>
   foundReferences
     .flatMap(ref => {
-      const absolutePath = pathPrefixRegex.test(ref)
-        ? resolveRelativePath(element.value[PATH], ref)
-        : FILE_CABINET_PATH_SEPARATOR.concat(ref)
+      const absolutePath = resolveRelativePath(element.value[PATH], ref)
+      // TODO: The log should be removed when SALTO-4025 is communicated.
+      if (!pathPrefixRegex.test(ref)) {
+        log.debug('Found a possible file reference without a path prefix: %s', ref)
+      }
       return [ref, absolutePath].concat(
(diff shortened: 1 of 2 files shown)


packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -135,9 +135,11 @@ const getServiceElemIDsFromPaths = (
 ): ElemID[] =>
   foundReferences
     .flatMap(ref => {
-      const absolutePath = pathPrefixRegex.test(ref)
-        ? resolveRelativePath(element.value[PATH], ref)
-        : FILE_CABINET_PATH_SEPARATOR.concat(ref)
+      const absolutePath = resolveRelativePath(element.value[PATH], ref)
+      // TODO: The log should be removed when SALTO-4025 is communicated.
+      if (!pathPrefixRegex.test(ref)) {
+        log.debug('Found a possible file reference without a path prefix: %s', ref)
```
dantal4, line 141: to keep the original behavior before this [commit](https://github.com/salto-io/salto/commit/a05213074a6569dc4ebe217abfdd5e3094d69be8), and to log only found references:
```suggestion
        const maybeServiceIdRecord = serviceIdToElemID[FILE_CABINET_PATH_SEPARATOR.concat(ref)]
        if (_.isPlainObject(maybeServiceIdRecord)) {
          log.debug('Found a possible file reference without a path prefix: %s', ref)
        }
        return [ref]
```

packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -135,9 +135,11 @@ const getServiceElemIDsFromPaths = (
...
+      }
       return [ref, absolutePath].concat(
```
dantal4, line 143: also for now we should keep the original behavior from [here](https://github.com/salto-io/salto/commit/a05213074a6569dc4ebe217abfdd5e3094d69be8)-
```suggestion
      return [absolutePath].concat(
```

dantal4 (approved):

PR #4722, opened 2023-08-10
SALTO-4604: Improve remote map performance when iterating over entries

ori-moisis:
Prefer using the cache over deserializing as it tends to be faster

---

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


tomsellek (approved):
Love the reimplementation of the `xxxImpl` functions in terms of lower-level building blocks. 
Code LGTM, but I can't claim to understand what the side effects of the added utilization of the location cache will end up being. Nor do I grok if this will result in a difference after the remote map has had data written to it.

PR #4727, opened 2023-08-13
SALTO-4606: Fetch Group memberships in Okta

shir-reifenberg:
Fetch Group memberships in Okta

---

Added a flag called `includeGroupMemberships` which will be false by default.
When enabled, we will fetch group memberships data, creating a `GroupMembership` instance for each group

---
_Release Notes_: 

_Okta adapter_:
* Added the option to fetch group memberships. To enable this feature, use 'includeGroupMemberships = true' to the fetch config in adapter config file.

---
_User Notifications_: 
None


packages/okta-adapter/src/filters/group_members.ts
```diff
@@ -0,0 +1,101 @@
...
+    )
+    : undefined
+}
+
+/**
+ * Create a single group-memberships instance per group.
+ */
+const groupMembersFilter: FilterCreator = ({ config, paginator }) => ({
+  name: 'groupMembersFilter',
+  onFetch: async (elements: Element[]): Promise<void> => log.time(async () => {
```
netama, line 81: nit: I don't think we need the `log.time()`, we now have wrapper logs on each filter action (including `onFetch`) so you'll see these with the filter name

netama (approved):
LGTM 🎈

PR #4729, opened 2023-08-14
SALTO-4006: Making the fetch.include and fetch.exclude required fields in the Netsuite config file

SeggevHaimovich:
_Made fetch & fetch.include & fetch.exclude must have fields.
Changed the types definitions and changed all the places handling an undefined field_

---

_Salto ticket: https://salto-io.atlassian.net/browse/SALTO-4006
Incident: https://salto-io.atlassian.net/browse/INCIDENT-4449?focusedCommentId=77053&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-77053_

---
_Release Notes_: 
_Users won't be able to delete these sections in the config file (they shouldn't do this)_

---
_User Notifications_: 
_None_


packages/netsuite-adapter/src/config.ts
```diff
@@ -894,11 +907,16 @@ export const combineQueryParams = (
   }
 }
 
-const emptyQueryParams = (): QueryParams => combineQueryParams(undefined, undefined)
+// const emptyQueryParams = (): QueryParams => combineQueryParams(undefined, undefined)
```
Shulik95, line 910: Should this be removed?

Shulik95 (approved):
Great Work! 💯 
Approving, please fix the small comment before merging.

packages/netsuite-adapter/src/config.ts
```diff
@@ -487,8 +487,18 @@ const fieldsToOmitConfig = createMatchingObjectType<FieldToOmitParams>({
 const fetchConfigType = createMatchingObjectType<FetchParams>({
   elemID: new ElemID(NETSUITE, 'fetchConfig'),
   fields: {
-    include: { refType: queryParamsConfigType },
-    exclude: { refType: queryParamsConfigType },
+    include: {
+      refType: queryParamsConfigType,
+      annotations: {
+        _required: true,
```
omrilit, line 493:
```suggestion
       [CORE_ANNOTATIONS.REQUIRED]: true,
```

packages/netsuite-adapter/src/query.ts
```diff
@@ -353,6 +369,15 @@ export function validatePlainObject(
...
 }
 
+export function validateDefined(
+  value: unknown,
+  configPath: string | string[]
+): asserts value is Record<string, unknown> {
+  if (value === undefined) {
+    throw new Error(`${makeArray(configPath).join('.')} should be defined`)
+  }
+}
```
omrilit, line 379: This API feels a bit weird to me.
Maybe something in the sense of the below will be clearer?
```suggestion
export function validateDefined(
  value: T | undefined,
  configPath: string | string[]
): asserts value is T {
  if (value === undefined) {
    throw new Error(`${makeArray(configPath).join('.')} should be defined`)
  }
}
```

packages/netsuite-adapter/src/config.ts
```diff
@@ -894,11 +907,16 @@ export const combineQueryParams = (
...
+// const emptyQueryParams = (): QueryParams => combineQueryParams(undefined, undefined)
 
 const updateConfigFromFailedFetch = (config: NetsuiteConfig, failures: FetchByQueryFailures): boolean => {
   const suggestions = toConfigSuggestions(failures)
-  if (_.isEmpty(suggestions)) {
+  if (_.isEqual(suggestions, {
+    fetch: {
+      include: fullQueryParams,
+      exclude: emptyQueryParams,
+    },
```
omrilit, line 918: maybe we can define this as a var and reuse it in other places?

packages/netsuite-adapter/src/config.ts
```diff
@@ -894,11 +907,16 @@ export const combineQueryParams = (
   }
 }
 
-const emptyQueryParams = (): QueryParams => combineQueryParams(undefined, undefined)
+// const emptyQueryParams = (): QueryParams => combineQueryParams(undefined, undefined)
 
 const updateConfigFromFailedFetch = (config: NetsuiteConfig, failures: FetchByQueryFailures): boolean => {
   const suggestions = toConfigSuggestions(failures)
-  if (_.isEmpty(suggestions)) {
+  if (_.isEqual(suggestions, {
```
omrilit, line 914: could you please explain the logic here?

omrilit (commented):
looks good!

packages/netsuite-adapter/src/config.ts
SeggevHaimovich: I see that every other place in the file used _required as I did.
when I try to use the suggested change I get this problem:
image

packages/netsuite-adapter/src/config.ts
SeggevHaimovich: Before my changes, "toConfigSuggestions" function initiates an empty config, adds relevant fields in case of any suggestions and returns the config. Now the function initiates a config with a full fetch and adds the other fields. So now if there were no suggestions the outcome will be a config with a full fetch.

packages/netsuite-adapter/src/adapter.ts
```diff
@@ -300,7 +300,7 @@ export default class NetsuiteAdapter implements AdapterOperations {
     const elementsToCreate = [
       ...customObjects,
       ...fileCabinetContent,
-      ...(this.userConfig?.fetch?.addBundles ? bundlesCustomInfo : []),
+      ...(this.userConfig?.fetch.addBundles ? bundlesCustomInfo : []),
```
dantal4, line 303: nit
```suggestion
      ...(this.userConfig.fetch.addBundles ? bundlesCustomInfo : []),
```

packages/netsuite-adapter/src/adapter.ts
```diff
@@ -395,9 +395,9 @@ export default class NetsuiteAdapter implements AdapterOperations {
       filePaths: this.filePathRegexSkipList.map(reg => `.*${reg}.*`),
     }))
     const fetchQuery = [
-      this.fetchInclude && buildNetsuiteQuery(this.fetchInclude),
+      buildNetsuiteQuery(this.fetchInclude),
       this.fetchTarget && buildNetsuiteQuery(convertToQueryParams(this.fetchTarget)),
-      this.fetchExclude && notQuery(buildNetsuiteQuery(this.fetchExclude)),
+      notQuery(buildNetsuiteQuery(this.fetchExclude)),
       this.lockedElements && notQuery(buildNetsuiteQuery(this.lockedElements)),
```
dantal4, line 401: nit
```suggestion
      notQuery(buildNetsuiteQuery(this.lockedElements)),
```

packages/netsuite-adapter/src/config.ts
```diff
@@ -686,14 +696,14 @@ const validateAdditionalDependencies = (
...
 }: Record<keyof FetchParams, unknown>): void => {
-  if (include !== undefined) {
-    validatePlainObject(include, [CONFIG.fetch, FETCH_PARAMS.include])
-    validateFetchParameters(include)
-  }
-  if (exclude !== undefined) {
-    validatePlainObject(exclude, [CONFIG.fetch, FETCH_PARAMS.exclude])
-    validateFetchParameters(exclude)
-  }
+  validateDefined(include, [CONFIG.fetch, FETCH_PARAMS.include])
```
dantal4, line 699: why do we need `validateDefined`?
`validatePlainObject` validates that it's an object, which is always defined...

packages/netsuite-adapter/src/query.ts
```diff
@@ -47,24 +47,40 @@ export type QueryParams = {
...
+  fileCabinet: ['.*'],
+  customRecords: [{ name: '.*' }],
+}
+
+export const emptyQueryParams: QueryParams = {
+  types: [],
+  fileCabinet: [],
+}
+
+export type SystemConfig = {
```
dantal4, line 61: maybe `LockedElementsConfig` ?

packages/netsuite-adapter/src/query.ts
```diff
@@ -47,24 +47,40 @@ export type QueryParams = {
   customRecords?: FetchTypeQueryParams[]
 }
 
+export const fullQueryParams: QueryParams = {
```
dantal4, line 50: this should be a function, to prevent cases that modifying it in one place will effect other places
```suggestion
export const fullQueryParams = (): QueryParams => ({
```

packages/netsuite-adapter/src/query.ts
```diff
@@ -47,24 +47,40 @@ export type QueryParams = {
...
+  types: [{ name: '.*' }],
+  fileCabinet: ['.*'],
+  customRecords: [{ name: '.*' }],
+}
+
+export const emptyQueryParams: QueryParams = {
```
dantal4, line 56: same comment like https://github.com/salto-io/salto/pull/4729/files#r1299392532

packages/netsuite-adapter/src/config.ts
```diff
@@ -1044,6 +1064,11 @@ function validateConfig(config: Record<string, unknown>): asserts config is Nets
     suiteAppClient,
   } = _.pick(config, Object.values(CONFIG))
 
+  validatePlainObject(config, 'config')
```
dantal4, line 1067: this is unnecessary, because `config` defined as `Record<string, unknown>`

packages/netsuite-adapter/src/config.ts
```diff
@@ -1090,21 +1110,27 @@ export const netsuiteConfigFromConfig = (
 ): NetsuiteConfig => {
   try {
     if (!configInstance) {
-      return {}
+      return {
```
dantal4, line 1113: can you add a log here of "using full fetch query" ?

packages/netsuite-adapter/src/config.ts
```diff
@@ -1090,21 +1110,27 @@ export const netsuiteConfigFromConfig = (
...
+
+    return {
+      ..._.pickBy(config, (_value, key) => {
+        if (key in CONFIG) {
+          return true
+        }
+        log.debug('Unknown config property was found: %s', key)
+        return false
+      }),
+      fetch: config.fetch,
```
dantal4, line 1132: is this change needed? it is a bit confusing

packages/netsuite-adapter/test/config.test.ts
```diff
@@ -188,10 +190,50 @@ describe('config', () => {
     expect(configChange?.message)
       .toBe(formatConfigSuggestionsReasons([STOP_MANAGING_ITEMS_MSG, LARGE_FOLDERS_EXCLUDED_MESSAGE]))
   })
+  describe('Should throw an error if the config fetch is non-valid', () => {
```
dantal4, line 193: Usually "Should... " is used for the `it` test, and describe is more of a title - like "fetch config"

dantal4 (changes requested):
Great work! (marking as "request changes" since there are some things that I think are important here....)

packages/netsuite-adapter/src/config.ts
SeggevHaimovich: Don't we want to throw a different error?
In the validatePlainObject func the message is "... should be an object", I think that it would be confusing if the user won't have the fetch field and will get this message.
We can also change the message of validatePlainObject to be something like "... should be defined as an object"
WDYT?

packages/netsuite-adapter/src/config.ts
SeggevHaimovich: The problem is that the return value must contain fetch field and TS does't know that the result of the pickBy must contain fetch field...
Do you have a better solution?

packages/netsuite-adapter/src/config.ts
dantal4: maybe inside `validatePlainObject` throw "...should be defined" if the input param is not defined

packages/netsuite-adapter/src/query.ts
```diff
@@ -47,24 +47,46 @@ export type QueryParams = {
...
+  fileCabinet: ['.*'],
+  customRecords: [{ name: '.*' }],
+})
+
+export const emptyQueryParams = (): QueryParams => ({
+  types: [],
+  fileCabinet: [],
+})
+
+export const fullFetch = (): FetchParams => ({
```
dantal4, line 86: naming nit
```suggestion
export const fullFetchConfig = (): FetchParams => ({
```

dantal4 (approved):
💯

PR #4730, opened 2023-08-14
SALTO-3958: generalize broken reference filter and CV

IdoZakk:
Make the automation broken references mechanism general to support ScriptRunner types

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/jira-adapter/src/change_validators/broken_references.ts
```diff
@@ -0,0 +1,66 @@
...
+            .filter(typeInfo.filter)
+            .map((referencedValue: Value) => _.get(referencedValue, typeInfo.namePath)),
+        }))
+        .filter(({ brokenReferencesNames }) => brokenReferencesNames.length > 0)
+        .map(({ instance, brokenReferencesNames }) => {
+          const brokenReferencesNamesString = brokenReferencesNames.join(', ')
+          const prettyTypeName = prettifyName(typeName).toLowerCase()
+          const referencesName = typeInfo.referencesTypeName
+          if (brokenReferencesNames.length === instance.value.projects.length) {
+            return {
```
alonstern, line 49: Is it impossible to create script runner instances that are not attached to any project?
I think we are a bit overgeneralizing with trying to create one change validator for all cases. IMO the BrokenReferenceInfo can contain only location and filter and we will have a function that omits these references so the change validators will only check it and return the error with the messages

IdoZakk: It is impossible to create them without a project or an issue type.
I considered what you are suggesting, but in that case I would do 4-5 copy pastes of almost the exact same code

alonstern: I was able to deploy a scripted field without projects, maybe only in the UI it can't be done? I tend to think even so if we can support it we should but I am not sure

IdoZakk: Yes, sorry, I meant impossible in the UI
If we support what? you think we should allow to deploy a scripted field without a project/issue type even if the UI does not allow it?

alonstern: Yes, if that is possible I think the user would prefer that over a change validator error, but we can ask in jira-dev

alonstern (approved):
I am not a big fan of trying the generalize multiple cases with many variables, but your call. (Anyway when the weak references will be ready most of this code will be changed)

PR #4733, opened 2023-08-14
SALTO-4609: Added core change validators to Salesforce Adapter.

tamtamirr:
Added core change validators to Salesforce Adapter.

---

This caused flows where users where not blocked when the deploy scope did not contain required dependencies.

---
_Release Notes_: 
Salesforce Adapter:
- Deploys that miss required missing dependencies will be blocked.

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/src/change_validator.ts
+++ b/packages/salesforce-adapter/src/change_validator.ts
@@ -48,7 +48,7 @@ import dataCategoryGroupValidator from './change_validators/data_category_group'
 import SalesforceClient from './client/client'
 import { ChangeValidatorName, DEPLOY_CONFIG, SalesforceConfig } from './types'
 
-const { createChangeValidator } = deployment.changeValidators
+const { createChangeValidator, getDefaultChangeValidators } = deployment.changeValidators
 
 type ChangeValidatorCreator = (config: SalesforceConfig,
                                isSandbox: boolean,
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/test/change_validator.test.ts
```diff
@@ -155,4 +155,7 @@ describe('createSalesforceChangeValidator', () => {
       })
     })
   })
+  it('should include the default change validators from core', () => {
+    expect(changeValidators).toContainKeys(Object.keys(deployment.changeValidators.getDefaultChangeValidators()))
+  })
```
tamtamirr, line 160: Was too lazy to add this to all the other adapters. Should I @alonstern ?

alonstern: I'm ok with doing it only in Salesforce for now

alonstern (approved):
👌

packages/salesforce-adapter/src/change_validator.ts
```diff
@@ -89,6 +89,7 @@ export const changeValidators: Record<ChangeValidatorName, ChangeValidatorCreato
   unknownPicklistValues: () => unknownPicklistValues,
   installedPackages: () => installedPackages,
   dataCategoryGroup: () => dataCategoryGroupValidator,
+  ...getDefaultChangeValidators(),
```
idoser96, line 92: nit. i would have prefer it to be on top

idoser96 (approved):

packages/salesforce-adapter/src/change_validator.ts
tamtamirr: If it would be on top that means the adapter will be able to override those.
I prefer the core validators to be at higher priority than the adapter.

PR #4736, opened 2023-08-15
SALTO-4612 - Remove CPU intensive log from SF adapter

amitgoldarb:
Remove CPU intensive log from SF adapter

---

This would get stuck in some cases where we needed to serialize profiles.

---
_Release Notes_: 
Resolve some rare cases where deploy would get stuck.

---
_User Notifications_: 
Resolve some rare cases where deploy would get stuck.

### DIFF
--- a/packages/salesforce-adapter/src/custom_object_instances_deploy.ts
+++ b/packages/salesforce-adapter/src/custom_object_instances_deploy.ts
@@ -66,11 +66,7 @@ type InstanceAndResult = {
 const logErroredInstances = (instancesAndResults: InstanceAndResult[]): void => (
   instancesAndResults.forEach(({ instance, result }) => {
     if (result.errors !== undefined) {
-      log.error(`Instance ${instance.elemID.getFullName()} had deploy errors - ${['', ...result.errors].join('\n\t')}
-
-and values -
-${safeJsonStringify(instance.value, undefined, 2,)}
-`)
+      log.error(`Instance ${instance.elemID.getFullName()} had deploy errors - ${['', ...result.errors].join('\n\t')}`)
     }
   })
 )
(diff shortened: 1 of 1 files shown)


packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -66,11 +66,7 @@ type InstanceAndResult = {
 const logErroredInstances = (instancesAndResults: InstanceAndResult[]): void => (
   instancesAndResults.forEach(({ instance, result }) => {
     if (result.errors !== undefined) {
-      log.error(`Instance ${instance.elemID.getFullName()} had deploy errors - ${['', ...result.errors].join('\n\t')}
-
-and values -
-${safeJsonStringify(instance.value, undefined, 2,)}
-`)
+      log.error(`Instance ${instance.elemID.getFullName()} had deploy errors - ${['', ...result.errors].join('\n\t')}`)
```
adamro, line 69: Is this a temp solution? Cause this information is really useful when investigating issues in this area..

ori-moisis, line 69: do we not need this information?
I think we should probably do something more along the lines of what @netama did last time with [this PR](https://github.com/salto-io/salto/pull/4266)

amitgoldarb: Don't we have the values in various other places? Why do we need them in DD as well?

amitgoldarb: Don't we have the values in various other places? Why do we need them in DD as well?

adamro: These are the values at the time of the deploy. What we have is other information that could be a heuristic but sometimes the issues can be caused from the actual values we are trying to deploy.

amitgoldarb: Switched to inspect as @ori-moisis suggested.

amitgoldarb: Nvm, switched to inspect.

ori-moisis (approved):

PR #4740, opened 2023-08-15
SALTO-4626: do not treat CredentialError differently in validateCredentials and fetch

omrilit:
There is an implicit assumption that whenever the adapter fails due to invalid credentials in fetch or validateCredentials flows, an CredentialError will be thrown from the adapter.
We found that there are cases that this doesn't really happen and we wouldn't like to count on it and behave differently in such cases.

---
_Release Notes_: 
Core:
SALTO-4626: identify credentials error in fetch, deploy and validateCredentials flows

---
_User Notifications_: 

### DIFF
--- a/packages/core/src/api.ts
+++ b/packages/core/src/api.ts
@@ -30,7 +30,6 @@ import {
   SaltoError,
   Element,
   DetailedChange,
-  isCredentialError,
   DeployExtraProperties,
   ReferenceMapping,
   AccountInfo,
@@ -96,13 +95,10 @@ export const verifyCredentials = async (
       const account = await adapterCreator.validateCredentials(loginConfig)
       return { success: true, ...account }
     } catch (error) {
-      if (isCredentialError(error)) {
-        return {
-          success: false,
-          error,
-        }
+      return {
+        success: false,
+        error,
       }
-      throw error
     }
   }
   throw new Error(`unknown adapter: ${loginConfig.elemID.adapter}`)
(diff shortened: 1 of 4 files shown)


packages/core/src/api.ts
```diff
@@ -306,16 +302,13 @@ export const fetch: FetchFunc = async (
...
   } catch (error) {
-    if (isCredentialError(error)) {
-      return {
-        changes: [],
-        fetchErrors: [{ message: error.message, severity: 'Error' }],
-        mergeErrors: [],
-        success: false,
-        updatedConfig: {},
-      }
+    return {
```
adamro, line 305: Are we sure this part is legit? Don't we have an assumption that anything other than CredentialsError will be thrown and not returned? (otherwise why did we throw anything that's not a CredentialsError)

omrilit: TL;DR - I checked these flows and it's legit.

In https://github.com/salto-io/salto_private/blob/94c08567c169e18636f93aa2af7b84fb776e1eb1/packages/ws-operations/src/wslc/operations/fetch/executor/index.ts#L83 we call `coreFetchWithUserVisibleExceptions` and mark every thrown error as user visible error.
The thrown error is eventually caught in https://github.com/salto-io/salto_private/blob/94c08567c169e18636f93aa2af7b84fb776e1eb1/packages/ws-operations/src/wslc/operation-runner.ts#L185 and handled in `handleOperationFailure`.

If we don't throw an error but return it in the `FetchResult.fetchErrors`, a `FetchOperationError` (which is a user visible error) will be thrown in https://github.com/salto-io/salto_private/blob/94c08567c169e18636f93aa2af7b84fb776e1eb1/packages/ws-operations/src/wslc/operations/fetch/executor/index.ts#L318 and will include the error message from the OSS.

adamro: And in the CLI? 🤔

omrilit: Seems like there is some weird behavior in the CLI TBH.
Currently (before my code changes) when having a credentials error, the error is outputted to the screen as part of the fetch but we proceed with applying the changes although there is no change that was reported.

It means that this will be the behavior for every operation from now on and not only for credentials.
For some reason we explicitly commented [in the code](https://github.com/salto-io/salto/blob/b6fb2eda72a99ee7d5b27d6ee23abe5581c46768/packages/cli/src/commands/fetch.ts#L221) that "We currently assume all fetchErrors are warnings".

image

adamro: Cause that was the assumption when that code was written. That Warnings are passed that way and Errors are thrown. I guess we already broke it with the CredentialErrors thing and didn't follow up on this properly.
You want to open a ticket about the CLI behaviour or decide on a legit behaviour and do it now cause we are changing the behaviour for non-Creds one to a non-intuitive one.

adamro (approved):

PR #4744, opened 2023-08-16
SALTO-4597 add log to determine which instances are causing convertMa…

benisraeladi:
…ps to decide to convert inner value to list

---

In LwcResources.LwcResource, there are times where we determine the type as `Map<salesforce.LwcResource>` and others where we determine the type as `Map<List<salesforce.LwcResource>>`. This can cause issues when cloning elements from one environment to another since we don't copy the type. Since the root cause is unknown, updating the log that might allow us to determine what is going on.

---
_Release Notes_:  _None_
---
_User Notifications_: _None_

### DIFF
--- a/packages/salesforce-adapter/src/filters/convert_maps.ts
+++ b/packages/salesforce-adapter/src/filters/convert_maps.ts
@@ -244,13 +244,15 @@ const convertInstanceFieldsToMaps = async (
   instanceMapFieldDef: Record<string, MapDef>,
 ): Promise<string[]> => {
   const nonUniqueMapFields = _.uniq(instancesToConvert.flatMap(
-    instance => convertArraysToMaps(instance, instanceMapFieldDef)
+    instance => {
+      const nonUniqueFields = convertArraysToMaps(instance, instanceMapFieldDef)
+      if (nonUniqueFields.length > 0) {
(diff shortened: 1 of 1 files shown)


packages/salesforce-adapter/src/filters/convert_maps.ts
```diff
@@ -248,8 +248,8 @@ const convertInstanceFieldsToMaps = async (
   ))
   if (nonUniqueMapFields.length > 0) {
     log.info(`Converting the following fields to non-unique maps: ${nonUniqueMapFields},
-     instances types are: ${
-  await awu(instancesToConvert).map(inst => metadataType(inst)).toArray()
+     instances are: ${
+  await awu(instancesToConvert).map(inst => inst.elemID.getFullName()).toArray()
 }`)
```
amitgoldarb, line 253: I think this log should be in the flatMap in line 247, otherwise we are just getting a list of all of the instances to convert right? We need the instance with the non-unique field itself

benisraeladi: You're right, fixing

amitgoldarb: Nit: we can probably remove the second log (the older one)

amitgoldarb (approved):
Approved with a small nit.
Thanks for this!

PR #4745, opened 2023-08-16
SALTO-4627: Improved Metadata aliases

tamtamirr:
Changed the aliases as for the plan described on the ticket.

---

Here are examples per different use-cases:
1. Instance with label: https://github.com/salto-io/tamir-sf/blob/alias/envs/env1/salesforce/Records/CMT2__mdt/CMT2_Record1.nacl
2. Instance with label from a namespace: https://github.com/salto-io/tamir-sf/blob/alias/envs/env1/salesforce/InstalledPackages/sbaa/Records/PermissionSet/sbaa__AdvancedApprovalsUser.nacl 
3. Custom Settings: https://github.com/salto-io/tamir-sf/blob/alias/envs/env1/salesforce/Objects/MyCustomSettings__c/MyCustomSettings__cAnnotations.nacl 
4. Custom Object from a namespace: https://github.com/salto-io/tamir-sf/blob/alias/envs/env1/salesforce/InstalledPackages/SBQQ/Objects/SBQQ__AttributeSet__c/SBQQ__AttributeSet__cAnnotations.nacl
5. Layout from a Namespace: https://github.com/salto-io/tamir-sf/blob/18d592e69d6f769de6d6711079dc9014a9e41039/envs/env1/salesforce/InstalledPackages/SBQQ/Objects/SBQQ__AttributeSet__c/Layout/SBQQ__Attribute_Set_Layout.nacl#L80C13-L80C34

- Added the option to use **fullName** instead of label (By providing **useLabelAsAlias = false** in the config)

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/src/filters/metadata_instances_aliases.ts
+++ b/packages/salesforce-adapter/src/filters/metadata_instances_aliases.ts
@@ -13,70 +13,16 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
-import {
-  CORE_ANNOTATIONS,
-  Element,
-  isElement,
-  isReferenceExpression,
-  ReadOnlyElementsSource,
-} from '@salto-io/adapter-api'
-import { getParents } from '@salto-io/adapter-utils'
-import _ from 'lodash'
-import { collections, values } from '@salto-io/lowerdash'
+import { CORE_ANNOTATIONS, Element } from '@salto-io/adapter-api'
+import { collections } from '@salto-io/lowerdash'
 import { logger } from '@salto-io/logging'
 import { LocalFilterCreator } from '../filter'
-import {
-  namePartsFromApiName,
-  buildElementsSourceForFetch,
-  removeCustomSuffix,
-  getNamespace,
-} from './utils'
-import { NAMESPACE_SEPARATOR } from '../constants'
+import { getInstanceAlias } from './utils'
 import { isMetadataInstanceElement, MetadataInstanceElement } from '../transformers/transformer'
 
 
 const { awu } = collections.asynciterable
-const { isDefined } = values
 const log = logger(module)
-
-const getAliasFromFullName = (instanceFullName: string): string => {
-  const nameWithoutParent = _.last(namePartsFromApiName(instanceFullName)) ?? instanceFullName
-  return _.last(removeCustomSuffix(nameWithoutParent)
-    .split(NAMESPACE_SEPARATOR)) ?? instanceFullName
-}
-
-const getParentAlias = async (
-  instance: MetadataInstanceElement,
-  elementsSource: ReadOnlyElementsSource
-): Promise<string | undefined> => {
-  const [parent] = getParents(instance)
-  if (parent === undefined) {
-    return undefined
-  }
-  if (!isReferenceExpression(parent)) {
-    log.debug('parent is not a reference expression. parent: %o', parent)
-    return undefined
-  }
-  const resolvedParent = await parent.getResolvedValue(elementsSource)
-  return isElement(resolvedParent)
-    ? resolvedParent.annotations[CORE_ANNOTATIONS.ALIAS]
-    : undefined
-}
-
-const setInstanceAlias = async (
-  instance: MetadataInstanceElement,
-  elementsSource: ReadOnlyElementsSource
-): Promise<void> => {
-  const namespace = await getNamespace(instance)
-  const parentAlias = await getParentAlias(instance, elementsSource)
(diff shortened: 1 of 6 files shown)


packages/salesforce-adapter/src/filters/metadata_instances_aliases.ts
```diff
@@ -13,82 +13,32 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
-import {
```
tamtamirr, line 16: Don't read into all the changes in the file, just re-review it as the implementation completely changed

packages/salesforce-adapter/test/filters/metadata_instances_aliases.test.ts
```diff
@@ -105,10 +94,26 @@ describe('metadataInstancesAliases filter', () => {
...
+        filter = filterCreator({
+          config: {
+            ...defaultFilterContext,
+            fetchProfile: buildFetchProfile({
+              optionalFeatures: { skipAliases: false, useLabelAsAlias: false },
+            }),
+          },
+        }) as typeof filter
+      })
+      it('should add correct aliases', async () => {
```
benisraeladi, line 111: `should add correct aliases` -> `should not add aliases`

packages/salesforce-adapter/test/filters/metadata_instances_aliases.test.ts
```diff
@@ -36,43 +38,30 @@ describe('metadataInstancesAliases filter', () => {
       mockTypes.CustomLabel,
       { fullName: 'TestInstance' }
     )
-    instanceWithCustomSuffix = new InstanceElement(
-      'TestValueSet__gvs',
-      mockTypes.GlobalValueSet,
-      { fullName: 'TestValueSet__gvs' }
-    )
```
benisraeladi, line 43: Why is something like this not necessary anymore?

benisraeladi (commented):
Looks much simpler than the previous implementation! Small questions still left

packages/salesforce-adapter/test/filters/metadata_instances_aliases.test.ts
tamtamirr: Not sure I understand the suggestion..? We still add aliases :)

packages/salesforce-adapter/test/filters/metadata_instances_aliases.test.ts
tamtamirr: Because we simply use the **Instance fullName** now, and we won't remove **custom instance suffixes** :)

packages/salesforce-adapter/test/filters/metadata_instances_aliases.test.ts
benisraeladi: The name is always added as an alias then? Why is adding it as an annotation on top of having the fullName necessary?

packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -442,3 +442,17 @@ export const isStandardObject = async (objectType: ObjectType): Promise<boolean>
...
   && !ENDS_WITH_CUSTOM_SUFFIX_REGEX.test(await safeApiName(objectType) ?? '')
 )
+
+export const getInstanceAlias = async (
+  instance: MetadataInstanceElement,
+  useLabelAsAlias: boolean
+): Promise<string> => {
+  const label = instance.value[LABEL]
+  if (!useLabelAsAlias || label === undefined) {
+    return instance.value[INSTANCE_FULL_NAME_FIELD]
```
amitgoldarb, line 452: Don't we want the namespace in parenthesees here as well?

amitgoldarb (commented):
Can we add an example of a layout from a managed package to a custom object from a managed package as well?

packages/salesforce-adapter/src/filters/utils.ts
tamtamirr: Nope, since the namespace is already incldued in the **fullName** and that would be excessive and confusing.

packages/salesforce-adapter/test/filters/metadata_instances_aliases.test.ts
tamtamirr: Me and @pgonzaleznetwork did not like the **Prettier** of the UI default output when no alias is defined.
Moreover, this will allow to search elements by their fullNames (We support search by alias).  This is an improvement from **elemID** searches which are the **naclCase** format of the **fullName**

amitgoldarb (approved):
LGTM Thanks!

PR #4748, opened 2023-08-16
SALTO-3985: Salesforce fetch with changes detection

tamtamirr:
Initial support in **fetch with changes detection** for Salesforce Adapter

---

**Do not review the files under core** as they come from this PR: https://github.com/salto-io/salto/pull/4693

This PR tackles two optimizations:
1. Retrieve instances that were updated since the last fetch. (No deletion support)
2. Do not create **Metadata Types** in **fetch with changes detection** mode. 

---
_Release Notes_: 
Salesforce Adapter:
- Initial support in **Fetch With Changes Detection** mode, which allows fetching only instances that were updated since the last fetch and decrease the fetch time significantly. 

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/adapter.ts
```diff
@@ -307,6 +313,15 @@ export const UNSUPPORTED_SYSTEM_FIELDS = [
   'LastViewedDate',
 ]
 
+const getMetadataTypesFromElementsSource = async (
```
tamtamirr, line 316: We use this for **fetchWithChangesDetection** to avoid creating wrong types (From filters that change types according to values of instances like **convert_lists** and **convert_maps**)

packages/salesforce-adapter/src/adapter_creator.ts
```diff
@@ -174,21 +174,39 @@ In Addition, ${configFromFetch.message}`,
...
     const credentials = credentialsFromConfig(context.credentials)
     const client = new SalesforceClient({ credentials, config: config[CLIENT_CONFIG] })
-    const salesforceAdapter = new SalesforceAdapter({
-      client,
-      config,
-      getElemIdFunc: context.getElemIdFunc,
-      elementsSource: context.elementsSource,
-    })
+
+    const createSalesforceAdapter = async ({
```
tamtamirr, line 188: Created this since getting the **ChangedAtSingleton** from the **ElementsSource** is async, and cannot be done in the constructor of the class **SalesforceAdapter**

packages/salesforce-adapter/src/adapter_creator.ts
```diff
@@ -200,8 +218,14 @@ export const adapter: Adapter = {
         return fetchResults
       },
 
-      deploy: salesforceAdapter.deploy.bind(salesforceAdapter),
-      validate: salesforceAdapter.validate.bind(salesforceAdapter),
+      deploy: async opts => {
+        const salesforceAdapter = await createSalesforceAdapter()
```
tamtamirr, line 222: Deploy & Validate don't need the **ChangedAtSingleton**

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -287,8 +288,12 @@ export const fetchMetadataInstances = async ({
       metadataType: metadataTypeName,
       name: prop.fullName,
       isFolderType: isDefined(metadataType.annotations[FOLDER_CONTENT_TYPE]),
+      changedAt: prop.lastModifiedDate,
     }))
 
+  if (filePropsToRead.length === 0) {
```
tamtamirr, line 294: Avoid sending empty requets for types that had no **instances** that were changed from the previous fetch. Do we need a comment here?

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -480,6 +485,11 @@ export const retrieveMetadataInstances = async ({
       .map(listFilesOfType)
   )).filter(props => notInSkipList(metadataQuery, props, false))
 
+  if (filesToRetrieve.length === 0) {
```
tamtamirr, line 488: Same logic that was added to **read**

packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
```diff
@@ -115,6 +119,15 @@ export const buildMetadataQuery = (
     }
     return false
   }
+  const wasUpdated = (instance: MetadataInstance): boolean => {
```
tamtamirr, line 122: This is where the magic happens 🪄 

- When this is not **fetchWithChangesDetection** it will return true.
- When the instance does not support the optimization (does not have **_changed_at** value) it will return true.
- When the instance support the optimization, it will compare the current **_changed_at** time with the one from the previous fetch.

packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -442,3 +443,10 @@ export const isStandardObject = async (objectType: ObjectType): Promise<boolean>
   await isCustomObject(objectType)
   && !ENDS_WITH_CUSTOM_SUFFIX_REGEX.test(await safeApiName(objectType) ?? '')
 )
+
+export const getChangedAtSingleton = async (
```
tamtamirr, line 447: Helper method to get the singleton. Used in both the **changed_at_singleton** filter and the **adapter creator**.

packages/salesforce-adapter/test/adapter.discover.test.ts
```diff
@@ -335,7 +335,8 @@ describe('SalesforceAdapter fetch', () => {
         + 1 /* treat blank as */
         + 1 /* value set */
         + 2 /* field dependency & value settings */
-        + 7 /* range restrictions */)
+        + 7 /* range restrictions */
+        + 2 /* ChangedAtSingleton type & instance */)
```
tamtamirr, line 339: Will be included in every fetch

packages/salesforce-adapter/test/adapter.discover.test.ts
```diff
@@ -1436,8 +1437,10 @@ public class LargeClass${index} {
       } as unknown as ObjectType
       const metadataQuery = {
         isTypeMatch: jest.fn(),
-        isInstanceMatch: jest.fn(),
-        isPartialFetch: jest.fn(),
+        isInstanceMatch: () => true,
+        isPartialFetch: () => false,
```
tamtamirr, line 1441: The previous mock values were a bug that was exposed by my changes.

tamtamirr (commented):
Added some comments for clarification 🤓 
This is the current flows I'll be testing and post their results:
1. No modifications.
2. Addition of **CustomField** to all profiles
3. Addition of **CustomField** to one profile only

packages/salesforce-adapter/src/fetch.ts
amitgoldarb: I think yes 🙏

packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
```diff
@@ -131,9 +144,14 @@ export const buildMetadataQuery = (
...
       && !fullExcludeList.some(params => isInstanceMatchQueryParams(instance, params))
+      && wasUpdated(instance)
     ),
 
-    isPartialFetch: () => target !== undefined,
+    isTargetedFetch: () => target !== undefined,
+
+    isFetchWithChangesDetection: () => changedAtSingleton !== undefined,
+
+    isPartialFetch: () => target !== undefined || changedAtSingleton !== undefined,
```
amitgoldarb, line 154: Why is partial fetch relevant to the changedAtSingleton?

tamtamirr: This PR changes the terminology of **Partial Fetch**. The old **Partial Fetch** is now **Targeted Fetched**, and now we added another mode **Fetch With Changes Detection**. Both are partial fetch, but different modes that can run independently or all together.
cc. @ori-moisis @tomermevorach @adamro we may want to sync on this :)

packages/salesforce-adapter/src/filters/changed_at_singleton.ts
```diff
@@ -0,0 +1,87 @@
...
+  const changedAtSingleton = elementsSource !== undefined
+    ? await getChangedAtSingleton(elementsSource)
+    : undefined
+  return changedAtSingleton ?? createEmptyChangedAtSingletonInstance()
+}
+
+const filterCreator: LocalFilterCreator = ({ config }) => ({
+  name: 'changedAtSingletonFilter',
+  onFetch: async (elements: Element[]) => {
+    const metadataInstancesByType = await groupByAsync(
```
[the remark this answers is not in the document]
tamtamirr: Why? Notice that is not the **apiName** of the instance, but the **apiName** of it's type, which is basically the type identifier

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -307,6 +313,15 @@ export const UNSUPPORTED_SYSTEM_FIELDS = [
...
+  elementsSource: ReadOnlyElementsSource
+): Promise<MetadataObjectType[]> => (
+  awu(await elementsSource.getAll())
```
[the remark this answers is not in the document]
tamtamirr: Not really, we simply return an **AsyncIterable** that is not awaited :)
(The resolved type of **ElementsSource.getAll** is `AsyncIterable<Element>`)

packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
```diff
@@ -74,6 +77,7 @@ const isFolderMetadataTypeNameMatch = ({ name: instanceName }: MetadataInstance,
 
 export const buildMetadataQuery = (
   { include = [{}], exclude = [] }: MetadataParams,
+  changedAtSingleton?: InstanceElement,
   target?: string[],
```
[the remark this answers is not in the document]
tamtamirr: The reason I've decided to add a single property is to have an alignment with the existing mode **Targeted Fetch**. If that's the logic, you'd probably want to pass **isTargetedFetch** boolean along with the **target** array.. No?
I think this many arguments will actually cause some confusion, but I don't have a strong opinion here

amitgoldarb (commented):
Discussed f2f and added a comment.

packages/salesforce-adapter/src/adapter_creator.ts
```diff
@@ -174,21 +174,39 @@ In Addition, ${configFromFetch.message}`,
...
-    })
+
+    const createSalesforceAdapter = async ({
+      isFetchWithChangesDetection = false,
+    }: CreateSalesforceAdapterParams = {}
+    ): Promise<SalesforceAdapter> => {
+      const { elementsSource, getElemIdFunc } = context
+      const changedAtSingleton = isFetchWithChangesDetection
+        ? await getChangedAtSingleton(elementsSource)
+        : undefined
```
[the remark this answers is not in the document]
tamtamirr: Been trying to implement the approach of using **prepare** and passing elementsSource, but couldn't avoid making **isInstanceMatch** async. I think it's better to stick with the current implementation and pass the **ChangedAtSingleton** directly.

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -232,6 +236,8 @@ export interface SalesforceAdapterParams {
   config: SalesforceConfig
 
   elementsSource: ReadOnlyElementsSource
+
+  changedAtSingleton?: InstanceElement
```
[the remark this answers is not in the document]
tamtamirr: discussed F2F

packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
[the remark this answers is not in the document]
tamtamirr: Discussed f2f and agreed to pass the **isFetchWithChangesDetection** param as-well

amitgoldarb (commented):
Looks good!
I'm curious to see how the code will change a bit in the adapter/adapter-creator after we do the chance we discussed. Everything else looks great. I'd like to see how they turn out before approving 🙏

packages/salesforce-adapter/src/fetch_profile/fetch_profile.ts
```diff
@@ -42,26 +43,44 @@ const optionalFeaturesDefaultValues: OptionalFeaturesDefaultValues = {
   generateRefsInProfiles: false,
 }
 
+type BuildFetchProfileParams = {
+  fetchParams: FetchParameters
+  isFetchWithChangesDetection: boolean
```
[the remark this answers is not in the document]
tamtamirr: I prefer explicit over implicit in such cases to avoid weird bugs and have a more clear code.

packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
```diff
@@ -115,6 +144,18 @@ export const buildMetadataQuery = (
     }
     return false
   }
+  const isIncludedInFetchWithChangesDetection = (instance: MetadataInstance): boolean => {
+    if (UNSUPPORTED_FETCH_WITH_CHANGES_DETECTION_TYPES.includes(instance.metadataType)) {
+      return false
+    }
+    if (changedAtSingleton === undefined || instance.changedAt === undefined) {
```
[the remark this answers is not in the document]
tamtamirr: Yes. That means any improvement will slightly improve quick-fetch.
We don't have many known gaps other than https://github.com/salto-io/salto/pull/4828 and **CustomObjects** which we excluded by default.

packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
```diff
@@ -53,6 +65,13 @@ const DEFAULT_NAMESPACE_MATCH_ALL_TYPE_LIST = [
 ]
 
 
+// Instances of this type won't be fetched in fetchWithChangesDetection mode
+const UNSUPPORTED_FETCH_WITH_CHANGES_DETECTION_TYPES = [
+  FLOW_METADATA_TYPE,
```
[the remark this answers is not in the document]
tamtamirr: I keep confusing in my brain **Flows** with **Profiles** for some odd reason 😢

packages/salesforce-adapter/src/adapter_creator.ts
```diff
@@ -174,21 +173,36 @@ In Addition, ${configFromFetch.message}`,
...
     const credentials = credentialsFromConfig(context.credentials)
     const client = new SalesforceClient({ credentials, config: config[CLIENT_CONFIG] })
-    const salesforceAdapter = new SalesforceAdapter({
-      client,
-      config,
-      getElemIdFunc: context.getElemIdFunc,
-      elementsSource: context.elementsSource,
-    })
+
+    const createSalesforceAdapter = async ({
```
amitgoldarb, line 187: @tamtamirr This function isn't async anymore and much of this change can be reverted right?

amitgoldarb (approved):
Approved with a single comment.
Great work here! Clean work.

packages/salesforce-adapter/src/filters/settings_type.ts
```diff
@@ -68,6 +68,10 @@ const filterCreator: RemoteFilterCreator = ({ client, config }) => ({
    * @param elements
    */
   onFetch: async (elements: Element[]): Promise<FilterResult> => {
+    // This filter shouldn't run in quick fetch as it takes a relatively long time
```
amitgoldarb, line 71: Can we link here to the ticket where we re-add this filter?

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -307,6 +318,17 @@ export const UNSUPPORTED_SYSTEM_FIELDS = [
   'LastViewedDate',
 ]
 
+const getMetadataTypesFromElementsSource = async (
+  elementsSource: ReadOnlyElementsSource
+): Promise<MetadataObjectType[]> => (
+  awu(await elementsSource.getAll())
+    .filter(isMetadataObjectType)
+    .filter(metadataType => !isCustomObjectSync(metadataType))
+    // Custom types shouldn't be caught here (CustomMetadata / CustomObject / CustomSettings)
```
amitgoldarb, line 327: This should also have a comment related to the ticket number where we remove this.

amitgoldarb (approved):
LGTM with a few small comments. Approving to save more cycles.

packages/salesforce-adapter/src/adapter.ts
tamtamirr: Note that this should never be removed as this method only returns the MetadataTypes from the Elements Source and not Custom Objects

PR #4755, opened 2023-08-17
SALTO-4637: add ability to create missing orgs in deployment

edenhassid:
add ability to create missing orgs in deployment

---

---
_Release Notes_: 
zendesk:
* when the deploy flag `defaultMissingOrgFallback` is true, missing referenced organizations will be created on deploy

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/change_validators/organization_existence.ts
+++ b/packages/zendesk-adapter/src/change_validators/organization_existence.ts
@@ -28,7 +28,7 @@ import wu from 'wu'
 import {
   createOrganizationPathEntries,
   getOrganizationsByIds,
-  getOrganizationsByNames, Organization,
+  getOrCreateOrganizationsByNames, Organization,
   TYPE_NAME_TO_REPLACER,
 } from '../filters/organizations'
 import ZendeskClient from '../client/client'
@@ -40,8 +40,8 @@ const { isDefined } = lowerDashValues
 const log = logger(module)
 
 /**
- * Validates the existence of organizations that are referenced in added or modified elements, if resolveOrganizationIDs
- * is true and defaultMissingOrgFallback than we will warn the user instead of error
+ * Validates the existence of organizations that are referenced in added or modified elements.
+ * If resolveOrganizationIDs is true and createMissingOrganizations then we will warn the user instead of error
  */
 export const organizationExistenceValidator: (
   client: ZendeskClient,
@@ -52,7 +52,7 @@ export const organizationExistenceValidator: (
   // If the user change this config between a fetch and a deploy, this validator will fail
   // This is a known issue and is ok because handling it is not trivial and the use case shouldn't be common
   const orgIdsResolved = fetchConfig.resolveOrganizationIDs === true
-  const defaultMissingOrgFallback = deployConfig?.defaultMissingOrgFallback === true
+  const createMissingOrganizations = deployConfig?.createMissingOrganizations === true
 
   const relevantChanges = changes.filter(isAdditionOrModificationChange).filter(isInstanceChange).map(getChangeData)
     .filter(instance => Object.keys(TYPE_NAME_TO_REPLACER).includes(instance.elemID.typeName))
@@ -70,7 +70,7 @@ export const organizationExistenceValidator: (
   let existingOrgs: Organization[]
   try {
     existingOrgs = orgIdsResolved
-      ? await getOrganizationsByNames({
+      ? await getOrCreateOrganizationsByNames({
         organizationNames: orgIdentifiers,
         paginator,
       })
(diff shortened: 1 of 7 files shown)


packages/zendesk-adapter/src/change_validators/organization_existence.ts
```diff
@@ -33,73 +33,89 @@ import {
...
+import { ZedneskDeployConfig, ZendeskFetchConfig } from '../config'
 
 const { awu } = collections.asynciterable
 const { isDefined } = lowerDashValues
 const log = logger(module)
 
 /**
- * Validates the existence of organizations that are referenced in added or modified elements
+ * Validates the existence of organizations that are referenced in added or modified elements, if resolveOrganizationIDs
+ * is true and defaultMissingOrgFallback than we will warn the user instead of error
```
netama, line 44: nits:
```suggestion
 * Validates the existence of organizations that are referenced in added or modified elements.
 * If resolveOrganizationIDs is true and defaultMissingOrgFallback then we will warn the user instead of error
```

packages/zendesk-adapter/src/change_validators/organization_existence.ts
```diff
@@ -33,73 +33,89 @@ import {
...
+        const nonExistingOrgsAreIds = !orgIdsResolved || wu(nonExistingOrgs.values()).some(org => _.isNumber(org))
+        const resolveOrgsRecommendation = nonExistingOrgsAreIds ? '. Salto can identify organizations by their names. This requires setting the \'resolveOrganizationIDs\' to true in the zendesk configuration file of both source and target envs and fetch.\n'
                 + 'More information about Salto\'s config files can be found here: \'https://help.salto.io/en/articles/7439324-salto-configuration-file\'' : ''
-            return {
-              elemID: entries[0].instance.elemID,
-              severity: 'Error',
-              message: 'Referenced organizations do not exist',
-              detailedMessage: `The following referenced organizations do not exist in the target environment: ${Array.from(nonExistingOrgs).join(', ')}${resolveOrgsRecommendation}`,
-            }
+        if (orgIdsResolved && defaultMissingOrgFallback) {
```
netama, line 101: nit (and not a blocker): if it is not valid to have the flag when `resolveOrganizationIDs` is false, can we add a config error for that case when validating the config (not here)?

packages/zendesk-adapter/src/filters/organizations.ts
```diff
@@ -142,6 +161,20 @@ export const getOrganizationsByNames = async (
         const organization = res.find(org => org.name === organizationName)
         if (organization === undefined) {
           log.error(`could not find any organization with name ${organizationName}`)
+          if (defaultMissingOrgFallback === true && client !== undefined) {
```
netama, line 164: nit: maybe log an error if the client is undefined?

packages/zendesk-adapter/src/filters/organizations.ts
```diff
@@ -142,6 +161,20 @@ export const getOrganizationsByNames = async (
         const organization = res.find(org => org.name === organizationName)
         if (organization === undefined) {
           log.error(`could not find any organization with name ${organizationName}`)
```
netama, line 163: I think this should be info/debug if we're going to create the org (+ maybe also clarify in the text?)

packages/zendesk-adapter/src/filters/organizations.ts
```diff
@@ -142,6 +161,20 @@ export const getOrganizationsByNames = async (
...
+          if (defaultMissingOrgFallback === true && client !== undefined) {
+            try {
+              const postRes = (await client.post({
+                url: '/api/v2/organizations',
+                data: { organization: { name: organizationName } },
```
netama, line 168: I think it's right to start with this - maybe add a comment mentioning the bulk-create alternative (https://developer.zendesk.com/api-reference/ticketing/organizations/organizations/#create-many-organizations) in case we run into issues? (can probably also just add it on the jira ticket so it's clear this was an intentional decision :) )

packages/zendesk-adapter/src/filters/organizations.ts
```diff
@@ -142,6 +161,20 @@ export const getOrganizationsByNames = async (
...
+              })).data
+              if (isSingleOrganizationResponse(postRes)) {
+                return postRes.organization
+              }
+              log.error('invalid organization creation response')
```
netama, line 173: nit: can we include the org name here as well?

packages/zendesk-adapter/src/filters/organizations.ts
```diff
@@ -117,9 +128,17 @@ export const getOrganizationsByIds = async (
   return results.flatMap(orgResponse => orgResponse.organizations)
 }
 
-export const getOrganizationsByNames = async (
-  organizationNames: string[],
-  paginator: clientUtils.Paginator,
+export const getOrganizationsByNames = async ({
```
netama, line 131: nit: maybe `getOrCreate` to clarify this may not be "read-only"?

packages/zendesk-adapter/src/filters/organizations.ts
```diff
@@ -117,9 +128,17 @@ export const getOrganizationsByIds = async (
...
+  organizationNames,
+  paginator,
+  defaultMissingOrgFallback,
+  client,
+}: {
+ organizationNames: string[]
+ paginator: clientUtils.Paginator
+ defaultMissingOrgFallback?: boolean
+ client?: ZendeskClient
+                                                }
```
netama, line 141:
```suggestion
}: {
  organizationNames: string[]
  paginator: clientUtils.Paginator
  defaultMissingOrgFallback?: boolean
  client?: ZendeskClient
}
```

netama (approved):
great ⭐ I think we should use a different name for the flag, the rest are nits so pre-approving

packages/zendesk-adapter/src/change_validators/organization_existence.ts
edenhassid: discussed IRL

packages/zendesk-adapter/src/filters/organizations.ts
edenhassid: I don't think we should log an error, client can be undefined if we don't pass it to the function like we do in the change validator

netama: so maybe debug? just some thing to help us understand when debugging why an org wasn't created

netama (approved):

PR #4757, opened 2023-08-17
Improve NS change errors

dantal4:
1. Make some lists unique.
2. Do not return dependency errors on validation (core does it already).
3. Add known error messages.

---

---
_Release Notes_: 
Netsuite Adapter:
- Improve NS change errors.

---
_User Notifications_: 
None


omrilit (approved):
Looks great! 💯

PR #4758, opened 2023-08-17
SALTO-4642: Display ElemId in deploy error messages

shir-reifenberg:
Display ElemId in deploy error messages & some readability improvements 

---

before:
image

after:
image

---
_Release Notes_: 

_CLI_
- Improve readability of deploy errors in the CLI

---
_User Notifications_: 
None

### DIFF
--- a/packages/cli/src/formatter.ts
+++ b/packages/cli/src/formatter.ts
@@ -21,7 +21,7 @@ import {
   Element, isInstanceElement, Values, Change, Value, getChangeData, ElemID,
   isObjectType, isField, isPrimitiveType, Field, PrimitiveTypes, ReferenceExpression,
   ActionName, ChangeError, SaltoError, isElement, TypeMap, DetailedChange, ChangeDataType,
-  isStaticFile, Group,
+  isStaticFile, Group, isSaltoElementError,
 } from '@salto-io/adapter-api'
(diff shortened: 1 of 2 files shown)


packages/cli/src/formatter.ts
```diff
@@ -359,17 +367,17 @@ export const deployErrorsOutput = (allErrors: DeployError[]): string => {
     const deployErrors = allErrors.filter(err => err.severity === 'Error')
     if (!_.isEmpty(deployErrors)) {
       errorOutputLines.push(error('Errors:'))
-      errorOutputLines.push(...deployErrors.map(deployError => error(`${deployError.message}`)))
+      deployErrors.forEach(deployError => errorOutputLines.push(error(getErrorMessage(deployError)), emptyLine()))
```
netama, line 370: leaving the styling review up to @ori-moisis as I don't have a strong opinion on this (could also keep the more functional-style variant with `flatMap` instead of `map`)

netama (commented):
cool :) let's add a unittest with some examples?

packages/cli/src/formatter.ts
ori-moisis: I think this is good, avoids the `...` which we don't like, and this was an array was mutated anyway before this change

ori-moisis (approved):
+1 on the unittest (I think we just need one that makes sure the element ID is in the output when we have an element error)

other than that, looks good

PR #4763, opened 2023-08-20
SALTO-4636 SALTO-4624: Append original error to user facing Errors & Map Metadata deploy limit error

tamtamirr:
Append original error to user facing Errors & Map Metadata deploy limit error

---

To improve the experience with **user facing errors**, it's better to include the original error returned from **Salesforce** to give more context on the error.

---
_Release Notes_: 
Salesforce Adapter:
- Custom Salto errors will include the original error from Salesforce. 
- Improved Salesforce Metadata Deploy size limit error.

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/src/client/user_facing_errors.ts
+++ b/packages/salesforce-adapter/src/client/user_facing_errors.ts
@@ -61,10 +61,14 @@ export type ErrorMappers = {
   [ENOTFOUND]: ErrorMapper<DNSException>
 }
 
+const withOriginalError = (error: Error, saltoErrorMessage: string): string => (
+  `${saltoErrorMessage}\n\nOriginal error: ${error.message}`
+)
+
 export const ERROR_MAPPERS: ErrorMappers = {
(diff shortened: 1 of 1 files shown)


packages/salesforce-adapter/src/client/user_facing_errors.ts
```diff
@@ -61,42 +61,48 @@ export type ErrorMappers = {
   [ENOTFOUND]: ErrorMapper<DNSException>
 }
 
+const withOriginalError = (salesforceError: string, saltoErrorMessage: string): string => (
+  `${saltoErrorMessage}\n\nOriginal error: ${salesforceError}`
```
amitgoldarb, line 65: @tomermevorach I wonder if customers will understand this "Original error text" phrase. WDYT?

tomermevorach: Lets change this to 'Underlying Error' please

amitgoldarb (approved):
Thanks for this one!

PR #4764, opened 2023-08-21
SALTO-4515: Projects are not added to automation names

idoamit8:
Jira changed Their API so project names are supported through `ruleScope`. This PR supports this change.

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/jira-adapter/src/filters/automation/automation_fetch.ts
```diff
@@ -89,13 +90,11 @@ const createInstance = (
   getElemIdFunc?: ElemIdGetter,
 ): InstanceElement => {
   const serviceIds = elementUtils.createServiceIds(values, 'id', type.elemID)
-
   const defaultName = naclCase([
     values.name,
-    ...values.projects
```
alonstern, line 95: Isn't changing `values.projects` to `(convertRuleScopeValueToProjects(values) ?? [])` enough? Why do we need to rest of the changes here?

alonstern (approved):
👌

PR #4768, opened 2023-08-22
SALTO-4657: Jira DC authentication fails

alonstern:
We added a check for the license in validate creds, this check works only in Cloud and fails validate creds on DC so I changed to run it only in cloud

---
_Release Notes_: 
__Jira Adapter__:
- Fixed bug with authentication to Jira

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/test/client/connection.test.ts
+++ b/packages/jira-adapter/test/client/connection.test.ts
@@ -42,6 +42,7 @@ describe('connection', () => {
       beforeEach(async () => {
         ({ accountId } = await validateCredentials({
           connection,
+          isDataCenter: false,
         }))
       })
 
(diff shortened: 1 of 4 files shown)


packages/jira-adapter/test/client/connection.test.ts
```diff
@@ -166,9 +171,24 @@ describe('connection', () => {
...
       expect(accountType).toEqual(undefined)
       expect(isProduction).toEqual(false)
     })
+
+    it('should return isProduction undefined and accountType = undefined', async () => {
+      connection = await createConnection({ retries: 1 }).login(
+        { baseUrl: 'https://test-sandbox-999.atlassian.net', user: 'me', token: 'tok', isDataCenter: true }
+      )
+      mockAxios.onGet('/rest/api/3/serverInfo').reply(200, { baseUrl: 'https://test.atlassian.net' })
+      mockAxios.onGet('/rest/api/3/instance/license').reply(200, { applications: [{ plan: 'FREE' }] })
```
IdoZakk, line 185: If I understand correctly the point was to avoid this call, I would check that indeed it was not called in the test

IdoZakk (approved):
It looks fine as a patch, don't we want to perform the correct call in case of DC?

PR #4770, opened 2023-08-22
SALTO-4597 change mapping function for lwcResources.lwcResource to ma…

benisraeladi:
…ke sure that it doesn't become a list

---

There is an edge case where, in some managed packages (specifically found this in pse) there is a `__mocks__` folder inside the lwc resource folder. This caused Salto to erroneously believe that the type needs to be `Map<List<lwcResource>>`

---

_Release Notes_: 
Salesforce Adapter:
- Solve LwcResources.LwcResource type discrepancy in LightningComponentBundle

---
_User Notifications_: 
Salesforce Adapter:
- It is possible that LightningComponentBundle records might be modified


tomsellek (approved):
LGTM :)

packages/salesforce-adapter/src/filters/convert_maps.ts
```diff
@@ -56,6 +56,10 @@ export const defaultMapper = (val: string): string[] => (
   val.split(API_NAME_SEPARATOR).map(v => naclCase(v))
 )
 
+const lwcMapper = (val: string): string[] => (
```
tamtamirr, line 59: I'm not a fan of this name, what does this method really does? Can we add a comment explaining this logic?

tamtamirr (commented):
Added small comment

packages/salesforce-adapter/src/filters/convert_maps.ts
```diff
@@ -56,6 +56,21 @@ export const defaultMapper = (val: string): string[] => (
...
+ * Convert a string value of a file path to the map index key.
+ * In this case, we want to use the last part of the path as the key, and it's
+ * unknown how many levels the path has. If there are less than two levels, take only the last.
+ */
+const filePathMapper = (val: string): string[] => {
+  const splitPath = val.split('/')
+  if (splitPath.length >= 3) {
+    return [naclCase(splitPath.slice(2).join('/'))]
+  }
+  log.warn(`Path ${val} has less than two levels, using only the last part as the key`)
```
amitgoldarb, line 69: Nit: This might be noisy, and it really isn't a warning, I think it can be a debug.

amitgoldarb (approved):
LGTM!
Thanks for this!

packages/salesforce-adapter/src/filters/convert_maps.ts
benisraeladi: I think this shouldn't happen at all, since we know the path is `lwc/<bundle_name>/<file_name>`. I would want a warning if we get to this because it means SF side has changed and we want to change the logic accordingly.

PR #4771, opened 2023-08-22
SALTO-3960: ScriptRunner Settings

IdoZakk:
Added support for Script Runner Settings

---

There were a few non-trivial items here:
* Adaptavist is not calling this API for get, they are calling a regular Jira API (which we cannot call as it is a connect call)
* The fields returned here have `-` in them. Ducktype is doing `naclCase` for the fields' names. We need to reverse it in order to deploy, but we need to do it after we resolve the references, so I added it in the standard deploy

---
_Release Notes_: 
Jira Adapter:
* Added support for ScriptRunner Settings

---
_User Notifications_: 
Jira Adapter:
* Customers with ScriptRunner enabled will now see the ScriptRunner Settings NaCl


packages/jira-adapter/src/config/api_config.ts
```diff
@@ -1764,6 +1764,26 @@ const DUCKTYPE_TYPES: JiraDuckTypeConfig['types'] = {
       },
     },
   },
+  ScriptRunnerSettings: {
+    request: {
+      url: 'sr-dispatcher/jira/admin/token/settings/features',
```
alonstern, line 1769: IDK if it really matters but should it start with `/` ?

alonstern (approved):
👌

PR #4772, opened 2023-08-22
SALTO-4226 Support subdirs of Objects folder in SDF integration

dantal4:
Also supports a case where the "Objects" folder isn't exists (treated as empty folder)

---

---
_Release Notes_: 
Netsuite Adapter:
- Support subdirs of Objects folder in SDF integration

---
_User Notifications_: 
None


omrilit (approved):
Looks great!

PR #4773, opened 2023-08-22
SALTO-4659: Creating new field configuration does not work in the Jira DC e2e account

alonstern:
Add field configuration back to Jira DC E2E

---

It doesn't reproduce anymore so for now I will add the test back

---
_Release Notes_: 
None

---
_User Notifications_: 
None


IdoZakk (approved):
:suspect:

PR #4775, opened 2023-08-22
SALTO-4641: Fix NS crash in data_instances_attributes trying to access "attributes" on null

omrilit:

---

---
_Release Notes_: 
Netsuite Adapter:
SALTO-4641: Fix crash in data_instances_attributes trying to access "attributes" on null
---
_User Notifications_: 


dantal4 (approved):
💯

PR #4778, opened 2023-08-23
SALTO-4645 allow excluding tag instances from fetch

netama:
It will now be possible to exclude tags, so that users are not required to work with tag instances (which are not really a zendesk concept and were added for impact analysis purposes due to past requests).
The deploy behavior remains identical.

No noise reduction is needed since the default behavior doesn't change.

---
_Release Notes_: 
_Zendesk adapter_:
* It is now possible to exclude `tag` instances in the fetch from config. When tags are excluded, references to them will not be created, but the referencing instances (e.g. automations and macros) will continue behaving as usual.

---
_User Notifications_: 
_Zendesk adapter_:
* It is now possible to exclude `tag` instances from the fetch config. Therefore, users who only included specific types explicitly might see the tag instances removed and the references converted to strings. To revert to the old behavior, make sure to include `tag` in the fetch config. (no impact on functionality is expected, since tags are never deployed)


edenhassid (approved):
looks good! 🚀 😃

PR #4779, opened 2023-08-23
SALTO-4350 - Okta Adapter - limit get calls to avoid reaching the rate limit

idoser96:
Limit the GET calls in order to not reach the limit, not allowing any GET calls when there are the same number of calls as the current limit, until the limit resets

---

* It is possible to receive 429s for the first call of each url, we have a buffer to prevent that, but it is still possible
* Because javascript is single threaded, unless the code runs into an await, it is atomic, which means we don't need to worry about race conditions and apply an actual mutex

---
_Release Notes_: 
Okta Adapter:
* GET calls will not pass the rate limit and will wait for the rate limit to reset before continuing

---
_User Notifications_: 
None


packages/okta-adapter/src/client/client.ts
```diff
@@ -45,9 +45,59 @@ const DEFAULT_PAGE_SIZE: Required<clientUtils.ClientPageSizeConfig> = {
...
+  new RegExp('/api/v1/events.*'),
+  new RegExp('/oauth2/v1/clients.*'),
+  new RegExp('/api/v1/org/email/bounces/remove-list'),
+  new RegExp('/api/v1/users'),
+  new RegExp('/api/v1/certificateAuthorities.*'),
+  new RegExp('/api/v1/devices.*'),
+  new RegExp('/api/v1.+'), // Has to be last
+]
+
+const updateRateLimits = (
```
shir-reifenberg, line 71: Can we reuse / consolidate with `getRetryDelayFromHeaders` ?

packages/okta-adapter/src/client/client.ts
```diff
@@ -70,11 +120,31 @@ export default class OktaClient extends clientUtils.AdapterHTTPClient<
...
+    const rateLimits = this.rateLimits.find(rateLimit => args.url.match(rateLimit.url))?.limits ?? this.defaultRateLimit
+    if (rateLimits.currentlyRunning > 0 && rateLimits.rateLimitRemaining - rateLimits.currentlyRunning <= 1) {
+      return this.waitAndReGet(rateLimits.rateLimitReset, args)
+    }
     try {
-      return await super.getSinglePage(args)
+      rateLimits.currentlyRunning += 1
+      const res = await super.getSinglePage(args)
+      if (res.headers) {
+        updateRateLimits(rateLimits, res.headers)
```
shir-reifenberg, line 145: shouldn't we call `updateRateLimits` even when there's an error?

idoser96: It seems that if there is an error there are no headers, i guess the remaining count doesn't change on failures

shir-reifenberg: cool :)

packages/okta-adapter/src/client/client.ts
idoser96: I was thinking about it, the problem is that we need to save the header instead of calculate from it
The only way i can think of using it is to safe the header itself and run getRetryDelayFromHeaders instead of
''' const currentTime = Date.now()
    const rateLimitResetTime = rateLimitReset * 1000 - currentTime ''', which seems a bit of overkill

packages/okta-adapter/src/client/client.ts
```diff
@@ -70,11 +124,37 @@ export default class OktaClient extends clientUtils.AdapterHTTPClient<
     return this.credentials.baseUrl
   }
 
+  private async waitAndReGet(
```
shir-reifenberg, line 127: I think it would be clearer if this method will just wait and we'll do the recursive call outside

packages/okta-adapter/src/client/client.ts
```diff
@@ -32,10 +34,14 @@ const log = logger(module)
...
   total: RATE_LIMIT_UNLIMITED_MAX_CONCURRENT_REQUESTS,
-  // TODO SALTO-2649: add better handling for rate limits
-  get: 2,
+  // There is a concurrent rate limit of minimum 15, depending on the plan, we take a buffer
+  get: 10,
   deploy: 2,
 }
+
+// We have a buffer to prevent reaching the rate limit with the initial request on parallel or consecutive fetches
+export const RATE_LIMIT_BUFFER = 6
```
shir-reifenberg, line 43: why do we need this? should it be part of the client's configuration? why 6?

packages/okta-adapter/src/client/client.ts
```diff
@@ -70,11 +124,37 @@ export default class OktaClient extends clientUtils.AdapterHTTPClient<
...
+    args: clientUtils.ClientBaseParams
+  ): Promise<clientUtils.Response<clientUtils.ResponseValue | clientUtils.ResponseValue[]>> {
+    // Calculate the time to wait until the rate limit is reset, and then recall the function
+    const currentTime = Date.now()
+    const rateLimitResetTime = Math.max(rateLimitReset - currentTime, 100)
+    await sleep(rateLimitResetTime)
+    return this.getSinglePage(args)
+  }
+
+  // This getSinglePage implements a rate limit logic, for more information: https://salto-io.atlassian.net/browse/SALTO-4350
```
shir-reifenberg, line 138: Can you please remove the jira project url and leave just the ticket numbe ? 😅 (personally I think it is better to include a short explanation here rather than a ticket, but your call)

packages/okta-adapter/src/client/client.ts
```diff
@@ -45,9 +51,59 @@ const DEFAULT_PAGE_SIZE: Required<clientUtils.ClientPageSizeConfig> = {
...
+]
+
+const updateRateLimits = (
+  rateLimits: { rateLimitRemaining: number; rateLimitReset: number },
+  headers: Record<string, string>
+): void => {
+  const updatedRateLimitRemaining = Number(headers['x-rate-limit-remaining'])
+  const updatedRateLimitReset = Number(headers['x-rate-limit-reset'])
+  if (!_.isNumber(updatedRateLimitRemaining) || !_.isNumber(updatedRateLimitReset)) {
+    log.error(`Invalid getSinglePage response headers: ${safeJsonStringify(headers)}`)
```
shir-reifenberg, line 84: nit: maybe log only the relevant headers ?

packages/okta-adapter/src/client/client.ts
```diff
@@ -70,11 +124,37 @@ export default class OktaClient extends clientUtils.AdapterHTTPClient<
...
   }
 
+  private async waitAndReGet(
+    rateLimitReset: number,
+    args: clientUtils.ClientBaseParams
+  ): Promise<clientUtils.Response<clientUtils.ResponseValue | clientUtils.ResponseValue[]>> {
+    // Calculate the time to wait until the rate limit is reset, and then recall the function
+    const currentTime = Date.now()
+    const rateLimitResetTime = Math.max(rateLimitReset - currentTime, 100)
+    await sleep(rateLimitResetTime)
```
shir-reifenberg, line 134: Does this `sleep` makes everything wait? or other requests will still run ?

shir-reifenberg (commented):
Very cool ! 😎 
Added a few comments inside, but my main concern is that we are still operating under the config's rate limits limitations. for example number of calls of `super.getSinglePage` will not exceed `DEFAULT_MAX_REQUESTS_PER_MINUTE` which defined in the config which is problematic in 2 ways IMO:
- We probably can do more if we only use the dynamic rate limits
- The rate limit logic is split to two mechanisms which makes it hard to follow
WDYT? we can talk IRL
Another question - does the info in `x-rate-limit-remaining` takes into account concurrency limits ? 

Also, can you please add some fetch time measurements with the new mechanism?

packages/okta-adapter/src/client/client.ts
```diff
@@ -45,9 +51,59 @@ const DEFAULT_PAGE_SIZE: Required<clientUtils.ClientPageSizeConfig> = {
...
+  rateLimitReset: 0,
+  currentlyRunning: 0,
+}
+
+
+const rateLimitUrls = [
+  new RegExp('/api/v1/apps/.+/(?:logo|groups)'), // has to be before /api/v1/apps.*
+  new RegExp('/api/v1/apps.*'),
+  new RegExp('/api/v1/groups/.+/roles'), // has to be before /api/v1/groups.*
+  new RegExp('/api/v1/groups.*'),
```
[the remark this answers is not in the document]
idoser96: yes

packages/okta-adapter/src/client/client.ts
```diff
@@ -45,9 +51,59 @@ const DEFAULT_PAGE_SIZE: Required<clientUtils.ClientPageSizeConfig> = {
...
+  new RegExp('/api/v1/users.*'),
+  new RegExp('/api/v1/logs.*'),
+  new RegExp('/api/v1/events.*'),
+  new RegExp('/oauth2/v1/clients.*'),
+  new RegExp('/api/v1/org/email/bounces/remove-list'),
```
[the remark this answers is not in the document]
idoser96: talked irl

packages/okta-adapter/src/client/client.ts
idoser96: Talked IRL - will also move to configuration

packages/okta-adapter/src/client/client.ts
idoser96: other requests will run

packages/okta-adapter/src/client/client.ts
idoser96: There are headers, we just ignored them in http_client, added them and am using them now

packages/okta-adapter/src/client/client.ts
```diff
@@ -19,24 +19,35 @@ import { Values } from '@salto-io/adapter-api'
...
+type OktaRateLimitConfig = {
+  rateLimitBuffer?: number
+}
+
 const DEFAULT_MAX_CONCURRENT_API_REQUESTS: Required<clientUtils.ClientRateLimitConfig> = {
   total: RATE_LIMIT_UNLIMITED_MAX_CONCURRENT_REQUESTS,
-  // TODO SALTO-2649: add better handling for rate limits
-  get: 2,
+  // There is a concurrent rate limit of minimum 15, depending on the plan, we take a buffer
+  get: 10,
```
netama, line 42: not a blocker if this was agreed with Shir, but did we see that we need this buffer?

packages/okta-adapter/src/client/client.ts
```diff
@@ -45,11 +56,78 @@ const DEFAULT_PAGE_SIZE: Required<clientUtils.ClientPageSizeConfig> = {
...
+  const currentTime = Date.now()
+  const rateLimitResetTime = Math.max(rateLimitReset - currentTime, 100)
+  await sleep(rateLimitResetTime)
+}
+
+const updateRateLimits = (
+  rateLimits: { rateLimitRemaining: number; rateLimitReset: number },
+  headers: Record<string, string>
+): void => {
+  const updatedRateLimitRemaining = Number(headers['x-rate-limit-remaining'])
```
netama, line 101: we should add safeties here + make sure it is possible to disable the new logic in case it's causing issues

packages/okta-adapter/src/client/client.ts
```diff
@@ -45,11 +56,78 @@ const DEFAULT_PAGE_SIZE: Required<clientUtils.ClientPageSizeConfig> = {
...
+  rateLimitRemaining: 0,
+  rateLimitReset: 0,
+  currentlyRunning: 0,
+}
+
+// according to okta, the way to tell if it is an id is that includes both letters and numbers
+// this regex makes sure there is at least 1 letter and at least 1 number, and no slashes
+const idRegex = '(?=.*[a-zA-Z])(?=.*\\d)(?!.*\\/).+'
+// According to https://developer.okta.com/docs/reference/rl-global-mgmt/
+const rateLimitUrls = [
```
netama, line 69: nits: 
1. these are "buckets", not urls, right? maybe add some explanation in the comment?
2. please use uppercase for constants

packages/okta-adapter/src/client/client.ts
```diff
@@ -45,11 +56,78 @@ const DEFAULT_PAGE_SIZE: Required<clientUtils.ClientPageSizeConfig> = {
 // The expression match AppUserSchema endpoint used for fetch
 const APP_USER_SCHEMA_URL = /(\/api\/v1\/meta\/schemas\/apps\/[a-zA-Z0-9]+\/default)/
 
+const initRateLimits = {
```
netama, line 59: nit: please use uppercase for consts

packages/okta-adapter/src/client/client.ts
```diff
@@ -58,24 +136,45 @@ export default class OktaClient extends clientUtils.AdapterHTTPClient<
...
+  // Then it resumes when the rate limit resets.
+  // For more information: SALTO-4350
   public async getSinglePage(
     args: clientUtils.ClientBaseParams,
   ): Promise<clientUtils.Response<clientUtils.ResponseValue | clientUtils.ResponseValue[]>> {
+    const rateLimits = this.rateLimits.find(({ url }) => args.url.match(url))?.limits ?? this.defaultRateLimits
+    // If this is the initial request, don't wait
+    if (!_.isEqual(rateLimits, initRateLimits)
+      // If nothing is running, and we passed the rate limit reset time, don't wait
+      && !(rateLimits.currentlyRunning === 0 && rateLimits.rateLimitReset < Date.now())
```
netama, line 161: not sure I understand this one - why do we need to check that nothing is running?

also (and related), I think the negation is complicating things here - IMO this would be easier to understand as
```suggestion
      && (rateLimits.currentlyRunning !== 0 || rateLimits.rateLimitReset >= Date.now())
```
we might need some other handling for this, but I think recursing and sleeping every time there's a concurrent call is not ideal (if we could avoid recursing it might be ok since the `sleep` function has an early exit if the sleep time is non-positive)

packages/okta-adapter/src/client/client.ts
```diff
@@ -45,11 +56,78 @@ const DEFAULT_PAGE_SIZE: Required<clientUtils.ClientPageSizeConfig> = {
...
+  new RegExp('^/api/v1/certificateAuthorities.*'),
+  new RegExp('^/api/v1/devices.*'),
+  new RegExp('^/api/v1/org/email/bounces/remove-list'),
+  new RegExp('^/oauth2/v1/clients.*'),
+  /* these endpoints are not documented, but have their own rate limit */
+  new RegExp('^/api/v1/iam.*'),
+  new RegExp('^/api/v1/domains'),
+  new RegExp('^/api/v1/device-assurances'),
+  /**/
+  new RegExp('^/api/v1.*'), // Has to be last
```
netama, line 87: if these are indeed buckets, we should add a fallback for a more general API as well, it's not guaranteed that every API call we make will be under the `/api/v1` prefix

netama (commented):
did a first pass, looks cool :) added some questions and comments

packages/okta-adapter/src/client/client.ts
idoser96: It is the recommended number according to their documentation, but i can play with it and see if it is safe to raise it without receiving 429

packages/okta-adapter/src/client/client.ts
idoser96: There is a buffer on the sleep to avoid extreme recursing, but i change it into a while loop and it should be equal
also removed the currentlyRunning === 0, it does seem like the other condition covers it

packages/okta-adapter/src/client/client.ts
idoser96: changed the check later to isFinite which covers safety
also made rateLimitBuffer = -1 to disable the feature

packages/okta-adapter/src/adapter_creator.ts
```diff
@@ -132,6 +133,7 @@ export const adapter: Adapter = {
       client: new OktaClient({
         credentials,
         config: config.client,
+        rateLimitBuffer: config[FETCH_CONFIG].rateLimitBuffer,
```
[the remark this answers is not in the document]
netama: I think it should still be there, this is related to the client calls and we can call it `clientRateLimitBuffer` if we want. also, are you sure it's not relevant for deploy?

packages/okta-adapter/src/client/client.ts
```diff
@@ -45,11 +56,105 @@ const DEFAULT_PAGE_SIZE: Required<clientUtils.ClientPageSizeConfig> = {
...
+]
+
+export const waitForRateLimit = async (rateLimitReset: number): Promise<void> => {
```
netama, line 98: please add a debug log in this function, we need to be able to benchmark when we're actually running the get calls

netama (commented):
added a few more comments, I think other than these we're good

packages/okta-adapter/src/client/client.ts
```diff
@@ -46,8 +46,8 @@ const DEFAULT_MAX_CONCURRENT_API_REQUESTS: Required<clientUtils.ClientRateLimitC
 // We have a buffer to prevent reaching the rate limit with the initial request on parallel or consecutive fetches
 export const DEFAULT_RATE_LIMIT_BUFFER = 6
 
-// Unlimited because the rate max requests is calculated dynamically
-const DEFAULT_MAX_REQUESTS_PER_MINUTE = -1
+const DEFAULT_MAX_REQUESTS_PER_MINUTE = 700
```
netama, line 49: what is the number based on?

packages/adapter-components/src/client/http_client.ts
```diff
@@ -256,7 +256,7 @@ export abstract class AdapterHTTPClient<
       }
       if (e.response !== undefined) {
         logResponse(e.response)
-        throw new HTTPError(`Failed to ${method} ${url} with error: ${e}`, _.pick(e.response, ['status', 'data']))
+        throw new HTTPError(`Failed to ${method} ${url} with error: ${e}`, { ..._.pick(e.response, ['status', 'data']), headers: this.extractHeaders(e.response.headers) })
```
netama, line 259: nit: not sure it makes sense to use `_.pick()` anymore - 
```suggestion
        throw new HTTPError(`Failed to ${method} ${url} with error: ${e}`, {
          status: e.response.status,
          data: e.response.data,
          headers: this.extractHeaders(e.response.headers),
        })
```

packages/okta-adapter/src/adapter_creator.ts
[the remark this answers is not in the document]
idoser96: It is related to deploy, but from my understanding currently we are only limiting fetch, but generally yes you are right

packages/okta-adapter/src/client/client.ts
idoser96: I don't know, but i think 600 is better because its the most common rate limit of non-free users
https://developer.okta.com/docs/reference/rl-global-mgmt/

netama (approved):

PR #4780, opened 2023-08-23
SALTO-4674: add trace logs for http client

edenhassid:
add trace logs for http client

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/client/http_client.ts
+++ b/packages/adapter-components/src/client/http_client.ts
@@ -15,7 +15,7 @@
 */
 import _ from 'lodash'
 import { ResponseType } from 'axios'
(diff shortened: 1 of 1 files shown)


packages/adapter-components/src/client/http_client.ts
```diff
@@ -231,6 +231,9 @@ export abstract class AdapterHTTPClient<
         }
         : undefined
 
+      if (isMethodWithData(params)) {
+        log.trace('going to send %s request with data %s', method.toUpperCase(), inspectValue(params.data))
```
netama, line 235: let's add the url as well? (technically we might also need the query params, but usually they're not used in the vast majority of cases so can probably start without them)

PR #4782, opened 2023-08-24
SALTO-4177: Fix issue layout owner order

alonstern:
Fixed issue layout owner order which seems to change on each fetch

---
_Release Notes_: 
__Jira Adapter__:
- Fixed issue layout owners order to not change on each fetch

---
_User Notifications_: 
None


edenhassid (approved):

packages/jira-adapter/src/filters/sort_lists.ts
```diff
@@ -81,6 +81,9 @@ const VALUES_TO_SORT: Record<string, Record<string, string[]>> = {
   [WORKFLOW_STATUS_TYPE_NAME]: {
     properties: ['key'],
   },
+  [ISSUE_LAYOUT_TYPE]: {
+    owners: ['elemID.name'],
```
netama, line 85: I see we did this in some other places, but is this really safe? what if the referenced element is not included in the fetch, won't the sort order change?

netama:
@idoamit8 fyi for when you're back :)

PR #4785, opened 2023-08-24
SALTO-4649: add important attribute core annotation

edenhassid:
add important attribute core annotation
the annotations are: `_important_values` and `_self_important_values`

---

---
_Release Notes_: 
core:
* add important attribute core annotation `_important_values` and `_self_important_values`

---
_User Notifications_: 
None

### DIFF
--- a/docs/syntax.md
+++ b/docs/syntax.md
@@ -439,17 +439,23 @@ _alias = "Example Instance"
 
 #### _important_values / _self_important_values
 
-This annotation is used to define a user-friendly alias for the element. The alias can be used in Salto enabled editors to display a shorter, clearer element name to users. Unlike the element ID, it does not have to be unique.
-
-Type: `string` Default: `undefined` Applicable to: Types, Instances, Fields
+This annotation is used to define what are these element's most important values. This list will be used in Salto enabled editors to highlight more important parts of the element, and index some of them for easy searchability. The list should include objects with the key of the element's important values and if it is indexed.
+Type: `string` Default: `undefined` Applicable to: Types
 Example:
 ```HCL
-type salto.example_type_long_id {
-_alias = "Example Type"
-}
-
-salto.example_type_long_id example_instance_id_with_long_prefix {
-_alias = "Example Instance"
+type salto.example {
+  _important_values = [
+    {
+      value = "active"
+      indexed = true
+    },
+  ]
+  _self_important_values = [
+    {
+      value = "name"
+      indexed = false
+    },
+  ]
 }
 ```
 
--- a/docs/syntax.md
+++ b/docs/syntax.md
@@ -440,7 +440,11 @@ _alias = "Example Instance"
 #### `_important_values` / `_self_important_values`
 
 This annotation is used to define what are these element's most important values. This list will be used in Salto enabled editors to highlight more important parts of the element, and index some of them for easy searchability. The list should include objects with the key of the element's important values and if it is indexed.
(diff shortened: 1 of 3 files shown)


docs/syntax.md
```diff
@@ -436,6 +437,22 @@ _alias = "Example Instance"
 }
 ```
 
+#### _important_values / _self_important_values
+
+This annotation is used to define a user-friendly alias for the element. The alias can be used in Salto enabled editors to display a shorter, clearer element name to users. Unlike the element ID, it does not have to be unique.
```
edenhassid, line 442: need to change to be relevant for impartant values

edenhassid: @liorn

docs/syntax.md
```diff
@@ -293,6 +293,7 @@ Currently the following core annotations are supported:
 - [_updatable](#_changed_by)
 - [_deletable](#_changed_at)
 - [_additional_properties](#_additional_properties)
+- [_important_values / _self_important_values](#_important_values-_self_important_values)
```
ori-moisis, line 296: does this link work? I think this is not the correct value to put there as the link location.
IIUC it should be
`#_important_values--_self_important_values`

seems like the other links with `/` in them also don't work. we should probably fix them as well

docs/syntax.md
```diff
@@ -436,6 +437,28 @@ _alias = "Example Instance"
 }
 ```
 
+#### _important_values / _self_important_values
```
ori-moisis, line 440: seems like some of the annotations have ` `` ` around them and others don't, may be worth fixing that and making sure everything looks in the same style
```suggestion
#### `_important_values` / `_self_important_values`
```

packages/adapter-api/src/builtins.ts
```diff
@@ -103,6 +103,25 @@ const restrictionType = new ObjectType({
   },
 })
 
+const importantValueType = new ObjectType({
+  elemID: new ElemID(GLOBAL_ADAPTER, BUILTIN_TYPE_NAMES.IMPORTANT_VALUE),
+  fields: {
+    // eslint-disable-next-line camelcase
```
ori-moisis, line 109: why is this disable needed?

docs/syntax.md
```diff
@@ -436,6 +437,28 @@ _alias = "Example Instance"
...
+
+This annotation is used to define what are these element's most important values. This list will be used in Salto enabled editors to highlight more important parts of the element, and index some of them for easy searchability. The list should include objects with the key of the element's important values and if it is indexed.
+Type: `string` Default: `undefined` Applicable to: Types
```
ori-moisis, line 443: "Type: \`string`" is incorrect here

edenhassid: what should the type be? is it ok to write `{value: string; indexed: boolean}[]` ?

ori-moisis: you can see examples form other complex annotations like `_restriction`

ori-moisis (approved):

packages/adapter-api/src/constants.ts
```diff
@@ -48,4 +50,5 @@ export const BUILTIN_TYPE_NAMES = {
   DEPENDENCY: 'dependency',
   HIDDEN_STRING: 'hidden_string',
   HIDDEN_BOOLEAN: 'hidden_boolean',
+  IMPORTANT_VALUE: 'importantValue',
```
netama, line 53: shouldn't this have been `important_value`?

PR #4790, opened 2023-08-27
SALTO-4656: Add getCustomReferences adapter API

alonstern:
Add getCustomReferences function to the adapter interface to allow him return references that are not in the NaCl and mark references as weak.

Changed the refernceTargetIndex structure to have a reference type for each reference

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/core/test/workspace/local/workspace.test.ts
+++ b/packages/core/test/workspace/local/workspace.test.ts
@@ -555,5 +555,13 @@ describe('local workspace', () => {
       const references = await getCustomReferences([instance], { test2: 'test' })
       expect(references).toEqual([])
     })
+
+    it('Should return empty array if adapter getCustomReferences throws an error', async () => {
+      adapterCreators.test = {
+        getCustomReferences: mockFunction<GetCustomReferencesFunc>().mockRejectedValue(new Error('aaa')),
+      } as unknown as Adapter
+      const references = await getCustomReferences([instance], { test2: 'test' })
+      expect(references).toEqual([])
+    })
   })
 })
--- a/packages/core/test/workspace/local/workspace.test.ts
+++ b/packages/core/test/workspace/local/workspace.test.ts
@@ -238,6 +238,8 @@ describe('local workspace', () => {
             expect.anything(),
             expect.anything(),
             expect.anything(),
+            undefined,
+            expect.anything(),
           )
         })
       })
@@ -255,6 +257,8 @@ describe('local workspace', () => {
             expect.anything(),
             expect.anything(),
             false,
+            undefined,
+            expect.anything(),
           )
         })
       })
--- a/packages/core/test/workspace/local/workspace.test.ts
+++ b/packages/core/test/workspace/local/workspace.test.ts
@@ -14,7 +14,7 @@
 * limitations under the License.
 */
 import path from 'path'
-import { Value } from '@salto-io/adapter-api'
+import { Adapter, ElemID, GetCustomReferencesFunc, InstanceElement, ObjectType, Value } from '@salto-io/adapter-api'
 import * as ws from '@salto-io/workspace'
 import * as file from '@salto-io/file'
 import { EnvironmentsSources, configSource as cs } from '@salto-io/workspace'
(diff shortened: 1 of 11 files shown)


packages/workspace/src/workspace/workspace.ts
```diff
@@ -328,9 +345,10 @@ export const loadWorkspace = async (
   credentials: ConfigSource,
   environmentsSources: EnvironmentsSources,
   remoteMapCreator: RemoteMapCreator,
+  getCustomReferences: (elements: Element[], accountToServiceName: Record<string, string>) => Promise<ReferenceInfo[]>,
```
ori-moisis, line 348: 1. lets give this function type a name `type ... = (elements: Element[], accountToServiceName: Record<string, string>) => Promise<ReferenceInfo[]>`
2. to avoid breaking interfaces, I suggest we add a default (or add the ability to use named arguments like we did in `loadLocalWorkspace`)

packages/workspace/src/workspace/reference_indexes.ts
```diff
@@ -39,37 +41,58 @@ export const REFERENCE_INDEXES_VERSION = 4
...
         })
       }
       return WALK_NEXT_STEP.RECURSE
     },
   })
-  return references
+
+  customReferences.forEach(refInfo => {
+    references[getReferenceDetailsIdentifier(refInfo)] = refInfo
+  })
```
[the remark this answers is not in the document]
alonstern: We want the custom references to also override existing references (to change a reference from a strong to weak) so a concat won't do that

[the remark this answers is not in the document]
ori-moisis: I think it will because eventually you put it in a tree map

packages/core/test/workspace/local/workspace.test.ts
```diff
@@ -493,4 +498,62 @@ describe('local workspace', () => {
       expect(file.rm).not.toHaveBeenCalled()
     })
   })
+
+  describe('getCustomReferences', () => {
```
ori-moisis, line 502: nit: test when the adapter throws

packages/workspace/src/workspace/reference_indexes.ts
```diff
@@ -234,13 +257,30 @@ const updateReferenceSourcesIndex = async (
...
+  const customReferencesAfter = await getCustomReferences(
+    changes.filter(isAdditionOrModificationChange).map(change => change.data.after)
+  )
+
+  const customReferencesBefore = await getCustomReferences(
+    changes.filter(isRemovalOrModificationChange).map(change => change.data.before)
+  )
+
+  return {
+    before: _.groupBy(customReferencesBefore, ref => ref.source.createTopLevelParentID().parent.getFullName()),
```
ori-moisis, line 271: should this be top level or base?
we access these later with the ID from the change, is the change always on a top level?

ori-moisis (commented):
do we need to bump the index version given that we changed its structure?
I guess we could wait with this until we have actual implementation maybe?

but this is something we need to think about, what happens when the code changes and we need to update references on elements that didn't change, do we need a new mechanism? do we migrate by fully regenerating the cache?

packages/workspace/src/workspace/reference_indexes.ts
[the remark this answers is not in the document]
alonstern: The results of this function are going to multiple places (also for calculating the incoming references). I am not sure we should rely on that each place will handle it right

packages/workspace/src/workspace/reference_indexes.ts
alonstern: The changes we get here are always top level

alonstern:
> do we need to bump the index version given that we changed its structure? I guess we could wait with this until we have actual implementation maybe?
> 
> but this is something we need to think about, what happens when the code changes and we need to update references on elements that didn't change, do we need a new mechanism? do we migrate by fully regenerating the cache?

Yeah I thought about bumping when the implementation will be ready.

I thought about regenerating the indexes when we change the code changes, I don't really see other options

ori-moisis (approved):

PR #4791, opened 2023-08-27
SALTO-4410: Enable CV on additions/removals of Standard Fields and Objects

tamtamirr:
Enable CV on additions/removals of Standard Fields and Objects

---

This is a previously implemented CV that wasn't enabled by mistake.

---
_Release Notes_: 
Salesforce Adapter:
- Add a Change Validator that forbids Additions and Deletions of Standard Fields and Objects.

---
_User Notifications_: 
_None_


benisraeladi (approved):
🙏

PR #4792, opened 2023-08-27
SALTO-4588 fix SF errors not always saving element IDs

benisraeladi:
Fix issue where element IDs were not saved on failed deployments
Also, make deploy message more concise, since all the information is now saved elsewhere

---

---
_Release Notes_: _None_

---
_User Notifications_: _None_

### DIFF
--- a/packages/salesforce-adapter/src/metadata_deploy.ts
+++ b/packages/salesforce-adapter/src/metadata_deploy.ts
@@ -27,7 +27,8 @@ import { SaltoError,
   isContainerType,
   isAdditionChange,
   SaltoElementError,
-  SeverityLevel } from '@salto-io/adapter-api'
+  SeverityLevel,
+  ElemID } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
 
 
@@ -55,6 +56,8 @@ export const DEPLOY_WRAPPER_INSTANCE_MARKER = '_magic_constant_that_means_this_i
 
 // Mapping of metadata type to fullNames
 type MetadataIdsMap = Record<string, Set<string>>
+
+type NameToElemIDMap = Record<string, ElemID>
 
 export type NestedMetadataTypeInfo = {
   nestedInstanceFields: string[]
@@ -177,8 +180,6 @@ type MetadataId = {
   fullName: string
 }
 
-type FailureMetadata = MetadataId & { deployMessage: DeployMessage }
-
 const getUnFoundDeleteName = (
   message: DeployMessage,
   deletionsPackageName: string
@@ -199,17 +200,21 @@ const isUnFoundDelete = (message: DeployMessage, deletionsPackageName: string): 
 const processDeployResponse = (
   result: SFDeployResult,
   deletionsPackageName: string,
+  typeAndNameToElemId: Record<string, NameToElemIDMap>
+
 ): { successfulFullNames: ReadonlyArray<MetadataId>
-  nonComponentErrors: ReadonlyArray<SaltoError | SaltoElementError>
-  componentErrorsFullNames: ReadonlyArray<FailureMetadata> } => {
+  errors: ReadonlyArray<SaltoError | SaltoElementError> } => {
   const allFailureMessages = makeArray(result.details)
     .flatMap(detail => makeArray(detail.componentFailures))
 
-  const componentErrorsFullNames = allFailureMessages
-    .map(failure => (
-      { type: failure.componentType,
-        fullName: failure.fullName,
-        deployMessage: getUserFriendlyDeployMessage(failure) }))
+  const componentErrors = allFailureMessages
+    .filter(failure => !isUnFoundDelete(failure, deletionsPackageName))
+    .map(getUserFriendlyDeployMessage)
+    .map(failure => ({
+      elemID: typeAndNameToElemId[failure.componentType]?.[failure.fullName],
+      message: failure.problem,
+      severity: 'Error' as SeverityLevel,
+    }))
 
   const testFailures = makeArray(result.details)
     .flatMap(detail => makeArray((detail.runTestResult as RunTestsResult)?.failures))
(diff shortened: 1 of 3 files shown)


packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -394,10 +393,36 @@ export const deployMetadata = async (
     )
   }
 
+  const isComponentFailureChange = (change: Change<MetadataInstanceElement>): boolean => {
+    const changeElem = getChangeData(change)
+    const changeDeployedIds = changeToDeployedIds[changeElem.elemID.getFullName()]
+    // TODO - this logic is not perfect, it might produce false positives when there are
```
amitgoldarb, line 399: I really do not now how severe this is. @ori-moisis @tamtamirr any idea?

packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -394,10 +393,36 @@ export const deployMetadata = async (
...
+    // child xml instances (because we pass in everything with a single change)
+    return (componentErrorsFullNames.some(
+      failureId => changeDeployedIds[failureId.type]?.has(failureId.fullName)
+    ))
+  }
+
+  const componentErrorsValidChanges = validChanges.filter(isComponentFailureChange)
```
amitgoldarb, line 406: What are we essentially filtering out here? Are we removing errors from customers views? This seems like something that either shouldn't happen and we should add a log.warn here or it will happen all the time and we are removing errors.

packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -394,10 +393,36 @@ export const deployMetadata = async (
...
+
+  const componentErrors = componentErrorsValidChanges
+    .flatMap(change => {
+      const changeElem = getChangeData(change)
+      return componentErrorsFullNames.filter(error =>
+        !isUnFoundDelete(error.deployMessage, pkg.getDeletionsPackageName())
+        && error.type === changeElem.elemID.typeName
+        && error.fullName === changeElem.elemID.name)
```
amitgoldarb, line 414: I'm not positive about this specific assumption. @ori-moisis @tamtamirr are you aware of possible edge cases?

amitgoldarb (commented):
Added a few comments. This might have some edge cases I am not aware of, I'm really not sure.

packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -217,30 +224,22 @@ const processDeployResponse = (
       ),
       severity: 'Error' as SeverityLevel,
     }))
-  const componentErrors = allFailureMessages
-    .filter(failure => !isUnFoundDelete(failure, deletionsPackageName))
-    .map(getUserFriendlyDeployMessage)
-    .map(failure => ({
-      elemID: changeDeployedIds[failure.componentType]?.[failure.fullName],
-      message: `Failed to ${checkOnly ? 'validate' : 'deploy'} ${failure.fullName} with error: ${failure.problem} (${failure.problemType})`,
```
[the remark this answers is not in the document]
benisraeladi: I don't think we fully solved the issue of what happens if we pass a `SaltoElementError` that is not of severity `Error`, so that could be problematic until then. If it's solved, then it's a great idea, otherwise I would open a ticket about it instead

packages/salesforce-adapter/src/metadata_deploy.ts
benisraeladi: We are removing all changes that were not failures

`validChanges` are all the changes that were sent. `componentFailureFullNames` are the names of all elements that failed

packages/salesforce-adapter/src/metadata_deploy.ts
tamtamirr: This should not work when the **naclCase(fullName)** is different from **fullName**.
Since this is **MetadataInstance**, we should use **safeApiName** to compare here.

tamtamirr (commented):
Looking good :) Added some comments.
I think we are missing UTs coverage for this changes.

packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -197,13 +199,18 @@ const isUnFoundDelete = (message: DeployMessage, deletionsPackageName: string):
...
-  changeDeployedIds: Record<string, MetadataIdsMap>,
-  checkOnly: boolean,
 ): { successfulFullNames: ReadonlyArray<MetadataId>
-  errors: ReadonlyArray<SaltoError | SaltoElementError> } => {
+  nonComponentErrors: ReadonlyArray<SaltoError | SaltoElementError>
+  componentErrorsFullNames: ReadonlyArray<FailureMetadata> } => {
   const allFailureMessages = makeArray(result.details)
     .flatMap(detail => makeArray(detail.componentFailures))
 
+  const componentErrorsFullNames = allFailureMessages
```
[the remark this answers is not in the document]
benisraeladi: I was trying to follow the same naming as `successfulFullNames`, since we don't return the failures now, only the names of the elements that failed

packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -56,6 +57,8 @@ export const DEPLOY_WRAPPER_INSTANCE_MARKER = '_magic_constant_that_means_this_i
 // Mapping of metadata type to fullNames
 type MetadataIdsMap = Record<string, Set<string>>
 
+type NameToElemIDMap = Record<string, ElemID>
```
tamtamirr, line 60: Why is this type needed? I find it quite confusing. I would honestly inline it.

packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -197,13 +200,22 @@ const isUnFoundDelete = (message: DeployMessage, deletionsPackageName: string):
 const processDeployResponse = (
   result: SFDeployResult,
   deletionsPackageName: string,
-  changeDeployedIds: Record<string, MetadataIdsMap>,
-  checkOnly: boolean,
+  typeAndNameToElemId: Record<string, NameToElemIDMap>
```
tamtamirr, line 203:
```suggestion
  deployedIdsByType: Record<string, NameToElemIDMap>
```
I find this name quite confusing

packages/salesforce-adapter/src/metadata_deploy.ts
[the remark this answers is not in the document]
tamtamirr: This change makes me think we lack a UT here. If we say all the severities should be **Error** for now, we should have a UT that validates that, until the implementation changes to support warnings as-well).

tamtamirr (commented):
Looking Good. Added some comments.

My biggest issue with this PR is the fact no tests were added / modified.
Since this PR adds more coverage for the **ElemIDs** mapping it either means the previous tests weren't occuratte (As none actually broke) or did not exist.
Also, let's link tickets for the things we know are not supported for now (e.g. **CustomObjects** sub-instances and **Workflow** sub-instances).

We need a Product decision regarding the sub-instances of **CustomObjects**, as IIUC errors on them will be linked to their parent **CustomObject**. @tomermevorach @amitgoldarb Are we okay with that?

packages/salesforce-adapter/test/adapter.crud.test.ts
```diff
@@ -150,30 +151,75 @@ describe('SalesforceAdapter CRUD', () => {
...
+                { action: 'add', data: { after: workflowFieldUpdate } }],
             },
           })
         })
 
-        it('should return an error', () => {
-          expect(result.errors).toHaveLength(1)
-          expect(result.errors[0].message).toContain('Some error')
+        it('should return errors', () => {
+          console.log(result.errors)
```
tamtamirr, line 200: remove

tamtamirr (approved):
LGTM 👑

PR #4799, opened 2023-08-28
SALTO-4693: fix deletion of static file when fetch is run with regenerate salto ids

edenhassid:
fix deletion of static file when fetch is run with regenerate salto ids

---

---
_Release Notes_: 
core:
* fix bug where regenerate salto ids would cause deletion of static file

---
_User Notifications_: 
None


alonstern (commented):
Can we add a test for it?
Besides that LGTM. @ori-moisis Can you have a look if you know any performance issues with this change?

ori-moisis (approved):

PR #4801, opened 2023-08-28
SALTO-4694: summarizeDeployChanges runs on modified requested changes

dorsev:
---

`deploy` method, changes the `actionPlan` is receives, and thus, hurts the `summarizeDeployChanges` which needs the original and un-modified "requestedChanges".

This PR "saves" them by copying them before the plan, and reusing them afterwards.

---

_Release Notes_: 
Bug fix: Post-deploy actions of failed changes was not shown in salesforce CPQ use-case.

---

_User Notifications_: 
None


dorsev:
Thank you for the review @ori-moisis

As I'm not entirely familiar with the code & it's context(and this is taking me a bit longer then I intended), I'm closing this PR, and will let someone else take this to the finish line.

PR #4802, opened 2023-08-28
SALTO-4527 combine transform values and transform values sync logic

benisraeladi:
Combine logic between transformValues and transformValuesSync so that they share more code.

---

The difficulty in combining the logic is the fact that we have to by synchronicity-agnostic, which is something Typescript is not very good at. With @ori-moisis's help, the logic that does not depend on synchronicity was refactored out into `recurseIntoValue`, and any part that still required synchronicity logic was kept inside the separate base functions `transformValues` and `transformValuesSync`

---
_Release Notes_: _None_

---
_User Notifications_: _None_

### DIFF
--- a/packages/adapter-utils/src/utils.ts
+++ b/packages/adapter-utils/src/utils.ts
@@ -15,7 +15,7 @@
 */
 import os from 'os'
 import wu from 'wu'
-import _, { mapValues } from 'lodash'
+import _ from 'lodash'
 import { inspect, InspectOptions } from 'util'
 import safeStringify from 'fast-safe-stringify'
 import { logger } from '@salto-io/logging'
@@ -156,7 +156,6 @@ const recurseIntoValue = (
 
   if (field && isListType(fieldType)) {
     const transformListInnerValue = (item: Value, index?: number): Value => (
-      // eslint-disable-next-line no-use-before-define
       transformFunc(
         item,
         !_.isUndefined(index) ? keyPathID?.createNestedID(String(index)) : keyPathID,
@@ -208,17 +207,17 @@ const recurseIntoValue = (
   return newVal
 }
 
-const depromise = async (x: Value): Promise<Value> => {
-  if (x instanceof Promise) {
-    return x
+const depromise = async (value: Value): Promise<Value> => {
+  if (value instanceof Promise) {
+    return value
   }
-  if (Array.isArray(x) && x.some(item => item instanceof Promise)) {
-    return Promise.all(x)
+  if (Array.isArray(value) && value.some(item => item instanceof Promise)) {
+    return Promise.all(value)
   }
-  if (_.isPlainObject(x)) {
-    return mapValuesAsync(x, async val => val)
+  if (_.isPlainObject(value)) {
+    return mapValuesAsync(value, async val => depromise(val))
   }
-  return x
+  return value
 }
 
 export const transformValues = async (
@@ -320,7 +319,7 @@ export const transformValuesSync = (
   const newVal = isTopLevel ? transformFunc({ value: values, path: pathID }) : values
   if (_.isPlainObject(newVal)) {
     const result = _.omitBy(
-      mapValues(
+      _.mapValues(
         newVal ?? {},
         (value, key) => transformValue(value, pathID?.createNestedID(key), fieldMapper(key))
       ),
@@ -329,15 +328,13 @@ export const transformValuesSync = (
     return _.isEmpty(result) && !allowEmpty ? undefined : result
   }
   if (_.isArray(newVal)) {
-    const result = wu(newVal)
-      .enumerate()
-      .map(([value, index]) => transformValue(
+    const result = newVal
+      .map((value, index) => transformValue(
         value,
         pathID?.createNestedID(String(index)),
         fieldMapper(String(index))
       ))
       .filter(value => !_.isUndefined(value))
-      .toArray()
     return result.length === 0 && !allowEmpty ? undefined : result
   }
   return newVal
--- a/packages/adapter-utils/src/utils.ts
+++ b/packages/adapter-utils/src/utils.ts
@@ -245,7 +245,7 @@ export const transformValues = async (
     const newVal = await transformFunc({ value, path: keyPathID, field })
     const recursed = recurseIntoValue(
       newVal,
-      transformFunc,
+      transformValue,
       strict,
       allowEmpty,
       keyPathID,
--- a/packages/adapter-utils/src/utils.ts
+++ b/packages/adapter-utils/src/utils.ts
@@ -15,7 +15,7 @@
 */
 import os from 'os'
 import wu from 'wu'
-import _ from 'lodash'
+import _, { mapValues } from 'lodash'
 import { inspect, InspectOptions } from 'util'
 import safeStringify from 'fast-safe-stringify'
 import { logger } from '@salto-io/logging'
@@ -282,9 +282,6 @@ export const transformValues = async (
   return newVal
 }
 
-// NOTE: Any changes that are made to this function need to take into account whether
-//  they're also needed over at transformValues. The two functions are separated
-//  only because of the way async is intertwined with the logic.
 export const transformValuesSync = (
   {
     values,
@@ -294,12 +291,9 @@ export const transformValuesSync = (
     pathID = undefined,
     isTopLevel = true,
     allowEmpty = false,
-  }: TransformValuesSyncArgs,
-): Values | undefined => {
-  const transformValueSync = (
-    value: Value,
-    keyPathID?: ElemID, field?: Field
-  ): Value => {
+  }: TransformValuesSyncArgs
+): lowerDashTypes.NonPromise<Value> | undefined => {
+  const transformValue = (value: Value, keyPathID?: ElemID, field?: Field): lowerDashTypes.NonPromise<Value> => {
     if (field === undefined && strict) {
       return undefined
     }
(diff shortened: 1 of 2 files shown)


packages/adapter-utils/src/utils.ts
```diff
@@ -101,33 +125,102 @@ const fieldMapperGenerator = (
...
-  pathID?: ElemID
-  isTopLevel?: boolean
-  allowEmpty?: boolean
+  if (isReferenceExpression(newVal)) {
+    return newVal
+  }
+
+  if (field && isListType(fieldType)) {
+    const transformListInnerValue = (item: Value, index?: number): Value => (
+      // eslint-disable-next-line no-use-before-define
```
ori-moisis, line 159: do we need this lint disable?

packages/adapter-utils/src/utils.ts
```diff
@@ -101,33 +125,102 @@ const fieldMapperGenerator = (
...
+const depromise = async (x: Value): Promise<Value> => {
+  if (x instanceof Promise) {
+    return x
+  }
+  if (Array.isArray(x) && x.some(item => item instanceof Promise)) {
+    return Promise.all(x)
+  }
+  if (_.isPlainObject(x)) {
+    return mapValuesAsync(x, async val => val)
+  }
```
ori-moisis, line 220: do we not need to recurse here? i.e - can we not get a list of lists or an object that contains objects?

packages/adapter-utils/src/utils.ts
```diff
@@ -293,115 +303,41 @@ export const transformValuesSync = (
...
+    return removeEmptyParts(recursed, allowEmpty)
   }
 
   const fieldMapper = fieldMapperGenerator(type, values)
 
   const newVal = isTopLevel ? transformFunc({ value: values, path: pathID }) : values
   if (_.isPlainObject(newVal)) {
     const result = _.omitBy(
-      _.mapValues(
+      mapValues(
```
ori-moisis, line 323: nit: consistency
```suggestion
      _.mapValues(
```

packages/adapter-utils/src/utils.ts
```diff
@@ -293,115 +303,41 @@ export const transformValuesSync = (
...
+        (value, key) => transformValue(value, pathID?.createNestedID(key), fieldMapper(key))
       ),
       _.isUndefined
     )
     return _.isEmpty(result) && !allowEmpty ? undefined : result
   }
   if (_.isArray(newVal)) {
-    const result = newVal
-      .map((value, index) => transformValueSync(
+    const result = wu(newVal)
```
ori-moisis, line 332: why change to `wu`?

packages/adapter-utils/src/utils.ts
```diff
@@ -101,33 +125,102 @@ const fieldMapperGenerator = (
...
+    )
+  }
+  return newVal
 }
 
-type TransformValuesSyncArgs = TransformValuesBaseArgs & { transformFunc: TransformFuncSync }
-type TransformValuesArgs = TransformValuesBaseArgs & {
-  transformFunc: TransformFunc
-  elementsSource?: ReadOnlyElementsSource
+const depromise = async (x: Value): Promise<Value> => {
```
ori-moisis, line 211: shouldn't call things "x"
```suggestion
const depromise = async (value: Value): Promise<Value> => {
```

ori-moisis (commented):
that failing unit test seems concerning
if this is really caused by this change, maybe something changed in the traversal order, which could be pretty scary as it might cause unexpected side effects in many places

packages/adapter-utils/src/utils.ts
```diff
@@ -101,33 +125,112 @@ const fieldMapperGenerator = (
...
+    return newVal
+  }
+
+  const listMapFunc = isAsync
+    ? (list: Iterable<Value>,
+      mapFunc: (t: Value, index: number) => Promise<unknown>) => toArrayAsync(mapAsync(list, mapFunc))
+    : _.map
+  const objMapFunc = isAsync
+    ? (obj: Iterable<Value>, mapFunc: (val: Value, key: string) => Promise<unknown>) => mapValuesAsync(obj, mapFunc)
+    : _.mapValues
```
ori-moisis, line 164: 1. I think this would be easier to read with a helper type
2. the input for map values is an object, not an iterable
```suggestion
  type AsyncMapFunc<K> = (t: Value, key: K) => Promise<unknown>
  const listMapFunc = isAsync
    ? (list: Iterable<Value>, mapFunc: AsyncMapFunc<number>) => toArrayAsync(mapAsync(list, mapFunc))
    : _.map
  const objMapFunc = isAsync
    ? (obj: Record<string, Value>, mapFunc: AsyncMapFunc<string>) => mapValuesAsync(obj, mapFunc)
    : _.mapValues
```

packages/adapter-utils/src/utils.ts
```diff
@@ -101,33 +125,112 @@ const fieldMapperGenerator = (
...
+        log.debug(`Array value and isListType mis-match for field - ${field.name}. Got non-array for ListType.`)
+      }
+      return transformListInnerValue(newVal)
+    }
+    return listMapFunc(newVal, (v: Value, i: number) => transformListInnerValue(v, i))
+  }
+  if (_.isArray(newVal)) {
+    // Even fields that are not defined as ListType can have array values
+    return listMapFunc(
+      newVal, (item: Value, index: number) => transformFunc(item, keyPathID?.createNestedID(String(index)), field)
```
ori-moisis, line 190: 1. types and not needed in the callback, they can be inferred from the `listMapFunc`
2. we should be consistent with the argument names
```suggestion
    return listMapFunc(newVal, (item, index) => transformListInnerValue(item, index))
  }
  if (_.isArray(newVal)) {
    // Even fields that are not defined as ListType can have array values
    return listMapFunc(
      newVal, (item, index) => transformFunc(item, keyPathID?.createNestedID(String(index)), field)
```

packages/adapter-utils/src/utils.ts
```diff
@@ -101,33 +125,112 @@ const fieldMapperGenerator = (
...
+      // but we do not want to omit those, we only want to omit empty
+      // objects, arrays and strings. we don't need to check for objects here
+      // because we cannot get here with an object
+      const valueIsEmpty = (Array.isArray(newVal) || _.isString(newVal)) && _.isEmpty(newVal)
+      return (valueIsEmpty && !allowEmpty) ? undefined : newVal
+    }
+    const fieldMapper = fieldMapperGenerator(fieldType, newVal)
+    return objMapFunc(
+      newVal,
+      (value: Value, key: string) => transformFunc(value, keyPathID?.createNestedID(key), fieldMapper(key))
```
ori-moisis, line 209: similar note, types can be inferred
```suggestion
      (value, key) => transformFunc(value, keyPathID?.createNestedID(key), fieldMapper(key))
```

packages/adapter-utils/src/utils.ts
```diff
@@ -152,91 +253,17 @@ export const transformValues = async (
...
+      newVal,
+      transformValue,
+      strict,
+      allowEmpty,
+      true,
+      keyPathID,
+      field,
+      await field?.getType(elementsSource)
+    )
+    return removeEmptyParts(await depromise(recursed), allowEmpty)
```
ori-moisis, line 266: with the new mapper functions, we will always get back a single promise from `recurseIntoValue`, we don't need the `depromise` logic anymore, we can remove that function and simply change this into
```suggestion
    const recursed = await recurseIntoValue(
      newVal,
      transformValue,
      strict,
      allowEmpty,
      true,
      keyPathID,
      field,
      await field?.getType(elementsSource)
    )
    return removeEmptyParts(recursed, allowEmpty)
```

packages/adapter-utils/src/utils.ts
```diff
@@ -101,33 +125,112 @@ const fieldMapperGenerator = (
...
+const recurseIntoValue = (
+  newVal: Value,
+  transformFunc: (value: Value, keyPathID?: ElemID, field?: Field) => Value,
+  strict: boolean,
+  allowEmpty: boolean,
+  isAsync: boolean,
+  keyPathID?: ElemID,
+  field?: Field,
+  fieldType?: TypeElement,
+): Value => {
```
ori-moisis, line 149: these are a lot of parameters for a function, they should be named and not positional so we'd be able to keep track of the meaning
specifically booleans are a problem because now we get calls like `recurseIntoValue(newVal, transformValue, strict, allowEmpty, true)` and it is really hard to know what `true` means

packages/adapter-utils/src/utils.ts
benisraeladi: I'm getting errors when removing the types :(
```
../adapter-utils/src/utils.ts:185:33 - error TS7006: Parameter 'item' implicitly has an 'any' type.

185     return listMapFunc(newVal, (item, index) => transformListInnerValue(item, index))
                                    ~~~~

../adapter-utils/src/utils.ts:185:39 - error TS7006: Parameter 'index' implicitly has an 'any' type.

185     return listMapFunc(newVal, (item, index) => transformListInnerValue(item, index))
```

packages/adapter-utils/src/utils.ts
benisraeladi: same here, getting build failures

ori-moisis: newer TS versions would be able to handle it but I guess our old one doesn't...
never mind then

packages/adapter-utils/src/utils.ts
```diff
@@ -152,91 +251,19 @@ export const transformValues = async (
...
-        if (strict) {
-          log.debug(`Value mis-match for field ${field?.name} - value is not an object`)
-        }
-        // _.isEmpty returns true for primitive values (boolean, number)
-        // but we do not want to omit those, we only want to omit empty
-        // objects, arrays and strings. we don't need to check for objects here
-        // because we cannot get here with an object
-        const valueIsEmpty = (Array.isArray(newVal) || _.isString(newVal)) && _.isEmpty(newVal)
-        return (valueIsEmpty && !allowEmpty) ? undefined : newVal
+    const recursed = await recurseIntoValue(
```
ori-moisis, line 254: style nit
```typescript
    const recursed = await recurseIntoValue({
      newVal,
      transformFunc: transformValue,
      strict,
      allowEmpty,
      isAsync: true,
      keyPathID,
      field,
      fieldType: await field?.getType(elementsSource),
    })

ori-moisis (approved):
do we maybe want to add a test that explicitly checks the order of recursion?
if we had that I assume it would have been easier to understand the issue that blocked this refactor for a while.
having such a test would also prevent us from accidentally changing it even if that one random test in workato is changed...

PR #4807, opened 2023-08-29
SAAS-8955: Add visibility into summarizeDeployChanges function

dorsev:
Today, it's pretty hard to debug when `summarizeDeployChanges` returns "failure" or "partial-failure" on a detailed change.

I'm adding some logs to help us identify the root cause quicker.

---

_Release Notes_: 
Add logs to all error cases in `summarizeDeployChanges` logic

---

_User Notifications_: 
None


packages/core/src/core/deploy/deploy_summary.ts
```diff
@@ -38,10 +38,14 @@ export type DeploySummaryResult = 'success' | 'failure' | 'partial-success'
 export type DetailedChangeId = string
 export type DetailedChangeDeploySummaryResult = [DetailedChangeId, DeploySummaryResult]
 
+const toTrimmedChangeValue = (value: any): string => inspectValue(value).slice(0, 10000)
```
dorsev, line 41: Note: the 10k char limit, is just a hard-coded value, added to make sure we don't print super big changes into the log file. 
The 10k is just a magic number, no real reason why I chose this specific value

dorsev:
Also: Not sure if this deserves a release notes part or not, but I've added one in any case. LMK your thoughts

packages/core/src/core/deploy/deploy_summary.ts
dorsev: Since I'm printing this for each change, maybe we can reduce this to 1k? 🤔

packages/core/src/core/deploy/deploy_summary.ts
```diff
@@ -38,10 +39,14 @@ export type DeploySummaryResult = 'success' | 'failure' | 'partial-success'
 export type DetailedChangeId = string
 export type DetailedChangeDeploySummaryResult = [DetailedChangeId, DeploySummaryResult]
 
+const toTrimmedChangeValue = (value: Value): string => inspectValue(value).slice(0, 1000)
```
ori-moisis, line 42: I think we should not do this here
if we wanted log trimming logic we'd put it in the log implementation (we have that for JSON format logs where we split them to multiple logs).
but tbh I don't see a reason why we should trim this at all, this isn't supposed to be bigger than other logs we have where we log values

packages/core/src/core/deploy/deploy_summary.ts
```diff
@@ -71,6 +76,12 @@ const summarizeAdditionChange = (
     const deployResult = diffBetweenRequestedAndApplied.every(isAdditionChange)
       ? 'success'
       : 'partial-success'
+    if (deployResult === 'partial-success') {
+      log.debug('Addition Summary for change id %s returned partial-success result', {
+        requestedChangeName,
+        diffs: toTrimmedChangeValue(diffBetweenRequestedAndApplied),
+      })
```
ori-moisis, line 83: this is not the style of logs we use in the OSS, extra params are not logged in text format (and here specifically also it won't replace the `%s` in the string which I think you wanted to replace)

this should probably be
```suggestion
      log.debug('Addition Summary for change id %s returned partial-success result, diff between requested and applied: %o',
        requestedChangeName,
        diffBetweenRequestedAndApplied,
      )
```
I think we don't have to explicitly call `inspect` here, this is the syntax of `%o` anyway IIRC, might be good to verify that though...

packages/core/src/core/deploy/deploy_summary.ts
```diff
@@ -90,9 +101,14 @@ const summarizeRemovalChange = (
       return 'success'
     }
     if (isModificationChange(appliedChange)) {
+      log.debug('Removal Summary for change id %s returned partial-success result', {
+        requestedChangeName,
+        requestedChange: toTrimmedChangeValue(requestedChange),
```
ori-moisis, line 106: I think we don't need to log the requested change, it doesn't really hold useful information beyond the ID that we wanted to remove

packages/core/src/core/deploy/deploy_summary.ts
dorsev: 👍 removed

packages/core/src/core/deploy/deploy_summary.ts
dorsev: 👍 removed

packages/core/src/core/deploy/deploy_summary.ts
dorsev: I saw other places do that already, for example https://github.com/salto-io/salto/blob/main/packages/adapter-utils/src/utils.ts#L1268

Any objections to keeping the same style? `%s, inspectValue(x)`

dorsev: Since I know printing `Values` to log is sensitive, I prefer to call `inspectValue` explicitly.

dorsev:
Thanks for the quick review @ori-moisis , I think I'm ready for another round

dorsev:
my e2e are failing with 

```
  ● Okta adapter E2E › fetch and deploy › should fetch the regular instances and types

    expect(received).toHaveLength(expected)

    Expected length: 0
    Received length: 1
    Received array:  [{"elemID": {"adapter": "okta", "fullName": "okta.Application.instance.TestSAMLApp62066706", "idType": "instance", "nameParts": ["TestSAMLApp62066706"], "typeName": "Application"}, "message": "Api validation failed: generalSettings. More info: Cannot create application instance. Maximum number of instances has been reached. (status code: 400)", "severity": "Error"}]

```

opened SALTO-4702 to keep track

ori-moisis (approved):

dorsev:
Thank you for the review  @ori-moisis , much appreciated 🙏

PR #4810, opened 2023-08-29
SALTO-4704: First Instance of ScriptRunner

IdoZakk:
Handle deployment of first instance of ScriptRunner- if there are no items we have an empty type and references will not be resolved

---
None
---
_Release Notes_: 
Jira Adapter:
* Fixed a bug that prevented deployment of first instance of a ScriptRunner type
---
_User Notifications_: 
None


alonstern (approved):
🔥

PR #4813, opened 2023-08-30
SALTO-4697: Add fixElements adapter API

alonstern:
Added fixElements adapter API

---

- Added fixElements API to adapter and core
- Added CLI command to run it
- Added fixElements function to dummy

---
_Release Notes_: 
__CLI__:
- Added `salto element fix` command

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-api/src/adapter.ts
+++ b/packages/adapter-api/src/adapter.ts
@@ -124,6 +124,11 @@ export type DeployModifiers = {
 
 export type ValidationModifiers = Pick<DeployModifiers, 'changeValidator'>
 
+export type FixElementsFunc = (elements: Element[]) => Promise<{
+  fixedElements: Element[]
+  errors: ChangeError[]
+}>
+
 export type AdapterOperations = {
   fetch: (opts: FetchOptions) => Promise<FetchResult>
   deploy: (opts: DeployOptions) => Promise<DeployResult>
@@ -131,6 +136,7 @@ export type AdapterOperations = {
   postFetch?: (opts: PostFetchOptions) => Promise<void>
   deployModifiers?: DeployModifiers
   validationModifiers?: ValidationModifiers
+  fixElements?: FixElementsFunc
 }
 
 export type AdapterOperationName = keyof AdapterOperations
@@ -192,8 +198,6 @@ export type ReferenceInfo = {
 
 export type GetCustomReferencesFunc = (elements: Element[]) => Promise<ReferenceInfo[]>
 
-export type FixElementsFunc = (elements: Element[]) => Promise<ChangeError[]>
-
 export type Adapter = {
   operations: (context: AdapterOperationsContext) => AdapterOperations
   validateCredentials: (config: Readonly<InstanceElement>) => Promise<AccountInfo>
@@ -204,7 +208,6 @@ export type Adapter = {
   loadElementsFromFolder?: (args: LoadElementsFromFolderArgs) => Promise<FetchResult>
   getAdditionalReferences?: GetAdditionalReferencesFunc
   getCustomReferences?: GetCustomReferencesFunc
-  fixElements?: FixElementsFunc
 }
 
 export const OBJECT_SERVICE_ID = 'object_service_id'
--- a/packages/adapter-api/src/adapter.ts
+++ b/packages/adapter-api/src/adapter.ts
@@ -192,12 +192,7 @@ export type ReferenceInfo = {
 
 export type GetCustomReferencesFunc = (elements: Element[]) => Promise<ReferenceInfo[]>
 
-export type ElementFix = Pick<ChangeError, 'message' | 'detailedMessage'> & {
-  severity: 'Info' | 'Warning'
-  fixedElement: Element
-}
-
-export type FixElementsFunc = (elements: Element[]) => Promise<ElementFix[]>
+export type FixElementsFunc = (elements: Element[]) => Promise<ChangeError[]>
 
 export type Adapter = {
   operations: (context: AdapterOperationsContext) => AdapterOperations
--- a/packages/adapter-api/src/adapter.ts
+++ b/packages/adapter-api/src/adapter.ts
@@ -192,6 +192,13 @@ export type ReferenceInfo = {
 
 export type GetCustomReferencesFunc = (elements: Element[]) => Promise<ReferenceInfo[]>
 
+export type ElementFix = Pick<ChangeError, 'message' | 'detailedMessage'> & {
+  severity: 'Info' | 'Warning'
+  fixedElement: Element
+}
+
+export type FixElementsFunc = (elements: Element[]) => Promise<ElementFix[]>
+
 export type Adapter = {
   operations: (context: AdapterOperationsContext) => AdapterOperations
   validateCredentials: (config: Readonly<InstanceElement>) => Promise<AccountInfo>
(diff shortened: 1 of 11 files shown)


packages/adapter-api/src/adapter.ts
```diff
@@ -192,6 +192,13 @@ export type ReferenceInfo = {
 
 export type GetCustomReferencesFunc = (elements: Element[]) => Promise<ReferenceInfo[]>
 
+export type ElementFix = Pick<ChangeError, 'message' | 'detailedMessage'> & {
+  severity: 'Info' | 'Warning'
+  fixedElement: Element
+}
```
ori-moisis, line 198: I don't understand this type
is it supposed to be an error message?
if so, it doesn't really make sense to couple it with the fixed element (we could reasonably have multiple messages and multiple fixes on a single element)

and why is it so detached from the (already horribly complex) SaltoError type hierarchy?

ori-moisis (commented):
not reviewing the code yet, lets agree on an interface and design first

packages/cli/src/commands/element.ts
```diff
@@ -775,6 +776,93 @@ const printElementDef = createWorkspaceCommand({
...
+      element
+    ))
+
+  const errors = await promises.array.withLimitedConcurrency(
+    fixes
+      .map(fix => () => workspace.transformToWorkspaceError(fix)),
+    20
+  )
+
+  outputLine(formatChangeErrors(errors), output)
```
ori-moisis, line 839: what is the point of calculating and printing the locations of the errors if the next line (`updateNaclFiles`) is going to change them anyway?
also, these are printed as if they are just errors, but in the case of `element fix` these are being fixed, the output should have some indication that the command is fixing these issues and not just that these are issues

packages/cli/test/commands/element.test.ts
```diff
@@ -1583,4 +1590,102 @@ Moving the specified elements to common.
       })
     })
   })
+
+  describe('fix command', () => {
+    const commandName = 'fix'
+    let cliArgs: mocks.MockCliArgs
+    let workspace: MockWorkspace
+    let fixElementsMock: jest.Mock
```
ori-moisis, line 1598: we should used typed functions
```suggestion
    let fixElementsMock: jest.MockedFunction<typeof fixElements>
```

packages/core/src/api.ts
```diff
@@ -76,6 +77,8 @@ const log = logger(module)
 
 const { mapValuesAsync } = promises.object
 
+const MAX_FIX_RUNS = 10
```
ori-moisis, line 80: should not be hard coded in core
this should be a parameter

packages/core/src/api.ts
```diff
@@ -692,3 +695,59 @@ export const getAdditionalReferences = async (
...
+export const fixElements = async (
+  workspace: Workspace,
+  selectors: ElementSelector[],
+): Promise<{ fixes: ChangeError[]; fixedElements: Element[] }> => {
+  const relevantIds = await selectElementIdsByTraversal({
+    selectors,
+    source: await workspace.elements(),
+    referenceSourcesIndex: await workspace.getReferenceSourcesIndex(),
+  })
+  const elements = await awu(relevantIds).map(id => workspace.getValue(id)).toArray()
```
ori-moisis, line 748: if we assert that all IDs are top level, it would be more efficient to get the elements with
```typescript
const elements = await (await workspace.elements()).getMany(...)
```
right?

packages/core/src/api.ts
```diff
@@ -692,3 +695,59 @@ export const getAdditionalReferences = async (
...
+
+  const elementGroups = _.groupBy(elements, element => element.elemID.adapter)
+
+  const elementFixGroups = await Promise.all(
+    Object.entries(elementGroups).map(async ([account, elementGroup]) => {
+      try {
+        // eslint-disable-next-line @typescript-eslint/return-await
+        return await adapterCreators[accountToService[account]]?.fixElements?.(elementGroup)
+      } catch (e) {
+        log.error(`Failed to fix elements for ${account} adapter: ${e}`)
```
ori-moisis, line 716: one of the many problems of designing an API where you expect the input to be modified in place:
when an error is thrown, you don't know if the input elements are still valid or not.

the API should return elements, the cloning part can / should be done by the adapter *if needed*

packages/core/src/api.ts
```diff
@@ -692,3 +695,59 @@ export const getAdditionalReferences = async (
...
+    .filter(values.isDefined)
+
+  if (elementFixes.length > 0 && runsLeft > 0) {
+    return elementFixes.concat(
+      await fixElementsContinuously(workspace, elements, runsLeft - 1)
+    )
+  }
+
+  if (elementFixes.length > 0) {
+    log.warn('Max element fix runs reached')
```
ori-moisis, line 733: this information should probably be in the return value somehow.
if the caller expects everything to be fixed, it would be good to tell them that not everything is fixed

ori-moisis (commented):
I'd be interested to see how this can actually be used to fix things like missing weak references.
I can't figure out how the adapter can do that without more information...

packages/cli/src/commands/element.ts
alonstern: For cases where we aren't deleting the value but changing it (isn't that the reason we return elemIDs in addition to the fixed element?)
Re the second point, the plan currently is that the message itself will be "Removed missing references from ..." or "Replaced missing users with user.." so it should be shown in the message (same in the SaaS where there isn't going to be a different between change validator errors to these messages)

packages/core/src/api.ts
alonstern: There isn't a `getMany` function in `ElementsSource`

packages/core/src/api.ts
alonstern: Why do you think that? do you expect the CLI and the SaaS to have different values?

packages/core/src/api.ts
alonstern: I am not sure we should give this case too much weight. This exists for if we have a bug so we won't enter an infinite loop. I can change it to throw an error instead so nothing will be returned if you think that's better.

packages/core/src/api.ts
```diff
@@ -692,3 +697,114 @@ export const getAdditionalReferences = async (
...
+): Promise<{ errors: ChangeError[]; fixedElements: Element[] }> => {
+  log.debug(`Fixing elements. ${runsLeft} runs left.`)
+
+  const elementGroups = _.groupBy(elements, element => element.elemID.adapter)
+
+  const elementFixGroups = await Promise.all(
+    Object.entries(elementGroups).map(async ([account, elementGroup]) => {
+      try {
+        // eslint-disable-next-line @typescript-eslint/return-await
+        return await adapters[account]?.fixElements?.(elementGroup)
```
ori-moisis, line 726: account name isn't always the adapter name.
grouping by account is probably the right thing to do, but accessing the `adapters` map using the account name could fail when it shouldn't

packages/dummy-adapter/src/adapter.ts
```diff
@@ -63,4 +63,28 @@ export default class DummyAdapter implements AdapterOperations {
...
+    }
+
+    const clonedInstance = fullInst1.clone()
+    delete clonedInstance.value.strField
+
+    return {
+      fixedElements: [clonedInstance],
+      errors: [{
+        message: 'Fixed FullInst1.strField',
+        detailedMessage: 'Removed FullInst1.strField',
```
ori-moisis, line 84: is this a good example for the pattern people should follow?
I know that in errors we don't usually want the element ID to be in the message as it breaks the ability to group similar errors together for example

alonstern: Removed the elemIDs from the messages, but yeah that's the general idea

PR #4815, opened 2023-08-30
SALTO-4177: Remove layouts

IdoZakk:
The previous try was not enough
---

None

---
_Release Notes_: 
Jira Adapter:
* Removing Issue Layouts

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/config/api_config.ts
+++ b/packages/jira-adapter/src/config/api_config.ts
@@ -1862,7 +1862,7 @@ const SUPPORTED_TYPES = {
   Automation: [],
   Webhook: [],
   [AUTOMATION_LABEL_TYPE]: [],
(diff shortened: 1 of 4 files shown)


alonstern:
Can we check if it is in the supported types in the beginning of the filter instead so we will be able to add it for specific users?

IdoZakk:
We removed it from the supported types, so we cannot add it now (it became un-recognized...)

IdoZakk:
I added it as a config flag now (default off), so we will be able to turn it on for customers

alonstern (approved):

alonstern:
> We removed it from the supported types, so we cannot add it now (it became un-recognized...)

I meant that we will be able to add it back by overriding the supported types in the user's adapter config, but this will work too

PR #4816, opened 2023-08-31
SALTO-4705 Support serializing and deserializing values

dantal4:
Support serializing and deserializing non element values - as primitives, reference expressions, static files etc.

This PR also fixes SALTO-4706 (static files are not well serialized on their own).

---

---
_Release Notes_: 
Workspace:
None

---
_User Notifications_: 
None


ori-moisis (approved):

ori-moisis:
I don't think this needs to be in the release notes as it is a completely internal implementation detail

PR #4819, opened 2023-08-31
SALTO-4717: Add hidden fields to types

IdoZakk:
Add hidden fields to types in case they do not exist

---

See the description in Jira

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/elements/type_elements.ts
+++ b/packages/adapter-components/src/elements/type_elements.ts
@@ -33,22 +33,26 @@ const log = logger(module)
  */
 export const hideFields = (
   fieldsToHide: FieldToHideType[],
-  typeFields: Record<string, FieldDefinition | Field>,
+  type: ObjectType,
   typeName: string,
 ): void => {
+  const typeFields = type.fields
   fieldsToHide.forEach(({ fieldName, fieldType }) => {
-    const field = Object.prototype.hasOwnProperty.call(typeFields, fieldName)
-      ? typeFields[fieldName]
-      : undefined
-    if (field === undefined) {
+    if (fieldType !== undefined && fieldType !== typeFields[fieldName]?.refType.elemID.name) {
       return
     }
-    if (fieldType === undefined || fieldType === field.refType.elemID.name) {
-      log.debug('Hiding values for field %s.%s', typeName, fieldName)
-      field.annotations = {
-        ...(field.annotations ?? {}),
-        [CORE_ANNOTATIONS.HIDDEN_VALUE]: true,
-      }
+    if (!Object.prototype.hasOwnProperty.call(typeFields, fieldName)) {
+      typeFields[fieldName] = new Field(
+        type,
+        fieldName,
+        BuiltinTypes.UNKNOWN,
+      )
+    }
+    const field = typeFields[fieldName]
+    log.debug('Hiding values for field %s.%s', typeName, fieldName)
+    field.annotations = {
+      ...(field.annotations ?? {}),
+      [CORE_ANNOTATIONS.HIDDEN_VALUE]: true,
     }
   })
 }
--- a/packages/adapter-components/src/elements/type_elements.ts
+++ b/packages/adapter-components/src/elements/type_elements.ts
@@ -30,6 +30,7 @@ const log = logger(module)
 
 /**
  * Annotate fields with _hidden_value=true if they match the specified configuration.
+ * Also add fields with _hidden_value=true if they are not defined in the type.
  */
 export const hideFields = (
   fieldsToHide: FieldToHideType[],
--- a/packages/adapter-components/src/elements/type_elements.ts
+++ b/packages/adapter-components/src/elements/type_elements.ts
@@ -34,14 +34,16 @@ const log = logger(module)
 export const hideFields = (
   fieldsToHide: FieldToHideType[],
   type: ObjectType,
-  typeName: string,
 ): void => {
   const typeFields = type.fields
   fieldsToHide.forEach(({ fieldName, fieldType }) => {
-    if (fieldType !== undefined && fieldType !== typeFields[fieldName]?.refType.elemID.name) {
+    if (fieldType !== undefined
+      && Object.prototype.hasOwnProperty.call(typeFields, fieldName)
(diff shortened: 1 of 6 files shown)


packages/adapter-components/src/elements/type_elements.ts
```diff
@@ -33,22 +33,26 @@ const log = logger(module)
...
+  type: ObjectType,
   typeName: string,
 ): void => {
+  const typeFields = type.fields
   fieldsToHide.forEach(({ fieldName, fieldType }) => {
-    const field = Object.prototype.hasOwnProperty.call(typeFields, fieldName)
-      ? typeFields[fieldName]
-      : undefined
-    if (field === undefined) {
+    if (fieldType !== undefined && fieldType !== typeFields[fieldName]?.refType.elemID.name) {
```
netama, line 41: we use `hasOwnProperty` because some builtin functions will "exist" under `typeFields[filedName]` even if no field exists. so this is implicitly going to behave differently for them and not create the field in these specific cases (which can be hard to debug). 
I think we should check based on `hasOwnProperty` here as well (and add a test case for the new functionality for one of the builtin functions, e.g. `toString`)

packages/adapter-components/src/elements/type_elements.ts
```diff
@@ -33,22 +33,26 @@ const log = logger(module)
...
       return
     }
-    if (fieldType === undefined || fieldType === field.refType.elemID.name) {
-      log.debug('Hiding values for field %s.%s', typeName, fieldName)
-      field.annotations = {
-        ...(field.annotations ?? {}),
-        [CORE_ANNOTATIONS.HIDDEN_VALUE]: true,
-      }
+    if (!Object.prototype.hasOwnProperty.call(typeFields, fieldName)) {
+      typeFields[fieldName] = new Field(
```
netama, line 45: let's add a debug log when implicitly creating fields

packages/adapter-components/test/elements/type_elements.test.ts
```diff
@@ -31,60 +31,67 @@ describe('type_elements', () => {
...
       // eslint-disable-next-line no-underscore-dangle
-      expect(fields.num.annotations?._hidden_value).toBeTruthy()
+      expect(myParentCustomType.fields.num.annotations?._hidden_value).toBeTruthy()
       // eslint-disable-next-line no-underscore-dangle
-      expect(fields.custom.annotations?._hidden_value).toBeTruthy()
-      expect(fields.str.annotations).toBeUndefined()
+      expect(myParentCustomType.fields.custom.annotations?._hidden_value).toBeTruthy()
+      expect(myParentCustomType.fields.str.annotations).toEqual({})
     })
     it('should not hide values for fields that do not have the right type', () => {
```
netama, line 61: please add an explicit test case for the new behavior

netama (commented):
LGTM with one issue. please also document the new behavior in the function docstring

packages/adapter-components/src/elements/type_elements.ts
IdoZakk: I did not change the usage of `hasOwnProperty`, I just moved it down the code.
But if the record contains `toString`, we will crash.
I made a change assuming we want to add `toString` as a hidden field in such a case

packages/adapter-components/test/elements/type_elements.test.ts
IdoZakk: I did... look at the test starting at line 76 (new file): I converted an older test to fit the new logic.
I will add a test for the `toString` case

packages/adapter-components/src/elements/type_elements.ts
```diff
@@ -30,25 +30,32 @@ const log = logger(module)
...
 ): void => {
+  const typeFields = type.fields
   fieldsToHide.forEach(({ fieldName, fieldType }) => {
-    const field = Object.prototype.hasOwnProperty.call(typeFields, fieldName)
-      ? typeFields[fieldName]
-      : undefined
-    if (field === undefined) {
+    if (fieldType !== undefined
+      && Object.prototype.hasOwnProperty.call(typeFields, fieldName)
+      && fieldType !== typeFields[fieldName].refType.elemID.name) {
```
netama, line 43: nit: this is also an error flow now (unexpected field type), let's log this one as well as a warning?

packages/adapter-components/src/elements/type_elements.ts
```diff
@@ -30,25 +30,32 @@ const log = logger(module)
...
-      field.annotations = {
-        ...(field.annotations ?? {}),
-        [CORE_ANNOTATIONS.HIDDEN_VALUE]: true,
-      }
+    if (!Object.prototype.hasOwnProperty.call(typeFields, fieldName)) {
+      log.debug(`Creating hidden field ${type.elemID.name}.${fieldName} with type ${fieldType}`)
+      typeFields[fieldName] = new Field(
+        type,
+        fieldName,
+        BuiltinTypes.UNKNOWN,
```
netama, line 51: shouldn't this be the field type (or alternatively the conditions and log should be different)? (let's make sure this is covered in a unittest)

packages/adapter-components/test/elements/type_elements.test.ts
```diff
@@ -31,60 +31,79 @@ describe('type_elements', () => {
...
       hideFields([
         { fieldName: 'missing' },
-      ], myCustomType.fields, 'bla')
-      expect(Object.keys(myCustomType.fields)).toHaveLength(2)
+      ], myCustomType)
+      expect(Object.keys(myCustomType.fields)).toHaveLength(3)
       expect(myCustomType.fields.str.annotations).toEqual({})
       expect(myCustomType.fields.num.annotations).toEqual({})
+      // eslint-disable-next-line no-underscore-dangle
+      expect(myCustomType.fields.missing.annotations._hidden_value).toBeTruthy()
```
netama, line 84: continuing from above - we should check the field types in the unittests as well and cover the relevant cases

netama (approved):
sorry about the delay, added some comments, pre-approving to unblock but please address them before merging

PR #4820, opened 2023-08-31
SALTO-4714: add onDeploy filter for ticket form

edenhassid:
add onDeploy filter for ticket form.
when there is a ticket status field in the env and it does not have an id, the onDeploy for `ticket_forms` will add the `ticket_field_ids` to the form from the `elmentSource` 

---

---
_Release Notes_: 
zendesk:
* prevent the failure of deployment of `ticket_form` when ticket status `ticket_field` is added to the env

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/filters/ticket_form.ts
+++ b/packages/zendesk-adapter/src/filters/ticket_form.ts
@@ -194,7 +194,9 @@ const filterCreator: FilterCreator = ({ config, client, elementsSource }) => ({
     }
     return { deployResult, leftoverChanges }
   },
-  // this onDeploy filter should be temporary and deleted once SALTO-4726 is implemented
+  // this onDeploy filter should be temporary and deleted once SALTO-4726 is implemented. This is because when
+  // custom_status ticket_field does not exist in the env and is referenced by a form, it is removed from the form and
+  // the deploy fails due to summarizeDeployChanges. This onDeploy filter restores ticket_field_ids to its initial state
   onDeploy: async (changes: Change<InstanceElement>[]) => {
     const ticketFormChanges = changes
       .filter(isAdditionOrModificationChange)
@@ -224,7 +226,6 @@ const filterCreator: FilterCreator = ({ config, client, elementsSource }) => ({
       .map(async id => {
         const form = await elementsSource.get(id)
         if (!isInstanceElement(form)) {
-          log.error(`could not find in the elementsSource a form with name ${id.name} `)
           return [id.getFullName(), undefined]
         }
         const ticketFieldIds = form.value.ticket_field_ids // the references are unresolved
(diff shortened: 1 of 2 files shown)


packages/zendesk-adapter/src/filters/ticket_form.ts
```diff
@@ -159,5 +165,53 @@ const filterCreator: FilterCreator = ({ config, client, elementsSource }) => ({
     }
     return { deployResult, leftoverChanges }
   },
+  // this onDeploy filter should be temporary and deleted once SALTO-4726 is implemented
+  onDeploy: async (changes: Change<InstanceElement>[]) => {
```
shir-reifenberg, line 169: May add a comment explaining why do we need this ?

packages/zendesk-adapter/src/filters/ticket_form.ts
```diff
@@ -159,5 +165,53 @@ const filterCreator: FilterCreator = ({ config, client, elementsSource }) => ({
...
+    if (_.isEmpty(ticketFormChanges)) {
+      return
+    }
+
+    const ticketFields = await awu(await elementsSource.list())
+      .filter(id => id.typeName === TICKET_FIELD_TYPE_NAME)
+      .map(id => elementsSource.get(id))
+      .filter(isInstanceElement)
+      .toArray()
+    const ticketStatusTicketField = ticketFields.find(field => field.value.type === 'custom_status')
```
shir-reifenberg, line 182: Do we have one `custom_status` ticket field ? if so, why do we need to get all `ticketFields` from elementSource ? don't we know its elemId and can get it specifically ?

packages/zendesk-adapter/src/filters/ticket_form.ts
```diff
@@ -159,5 +165,53 @@ const filterCreator: FilterCreator = ({ config, client, elementsSource }) => ({
...
+        const ticketFieldIds = form.value.ticket_field_ids // the references are unresolved
+        return [id.getFullName(), ticketFieldIds]
+      })
+      .toArray())
+
+    ticketFormChanges.forEach(change => {
+      const form = getChangeData(change)
+      const ticketFieldIds = ticketFormsToFieldIds[form.elemID.getFullName()]
+      if (ticketFieldIds === undefined) {
+        log.error(`could not find ticketFieldIds for form ${form.elemID.name}`)
```
shir-reifenberg, line 210: nit: I think this / the log in 198 is redundant cause we're logging twice due to the same reason

shir-reifenberg (approved):
😎 
added a few small comments, pre approving

packages/zendesk-adapter/src/filters/ticket_form.ts
edenhassid: the elemId can change according to the config right? so it may be irrelevant and then the only way to find this field is using its type

PR #4821, opened 2023-09-03
SALTO-4712: Add order filter in Okta

shir-reifenberg:
We noticed differences in `GroupRule` target groups between envs. The order has no meaning so we should sort the list to avoid this 

---

---
_Release Notes_: 

_Okta_Adapter_:
- Fix a bug casing unwanted differences in GroupRule lists

---
_User Notifications_: 
_Okta_Adapter_:
- In GroupRule instances, target group list order might change


edenhassid (approved):
LGTM! 🔥

PR #4823, opened 2023-09-03
SALTO-4386 Changed isFreeLicense in Jira

idoamit8:
In this PR, I've split isFreeLicense into two functions:
1. `isAllFreeLicense`, which checks if there are any non-free apps for the account.
2. `isJiraSoftwareFreeLicense`, which checks if the account has a non-free Jira Software app.

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/test/filters/permission_scheme/deploy_permission_scheme.test.ts
+++ b/packages/jira-adapter/test/filters/permission_scheme/deploy_permission_scheme.test.ts
@@ -52,7 +52,7 @@ describe('deploy permission scheme', () => {
     const elementsSource = getLicenseElementSource(true)
     filter = deployPermissionScheme(getFilterParams({ elementsSource })) as typeof filter
   })
-  it('should mock deploy of addition schemes on free license', async () => {
+  it('should deploy of addition schemes on free license', async () => {
     const { deployResult, leftoverChanges } = await filter.deploy(changes)
     expect(deployResult).toEqual({
       appliedChanges: [changes[0], changes[2]],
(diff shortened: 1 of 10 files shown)


IdoZakk (commented):
Great! Please add coverage for the new function? (I know there is no coverage currently for the old one, but there should be....)
Also, and maybe more importantly, add tests for both using fetches to verify the business logic is not hurt if there is a change in those functions

packages/jira-adapter/test/utils.test.ts
```diff
@@ -0,0 +1,119 @@
...
+              },
+            ],
+          },
+        }
+      )
+      elementsSource = buildElementsSourceFromElements([accountInfo])
+      isFree = await isJiraSoftwareFreeLicense(elementsSource)
+      expect(isFree).toBeFalsy()
+    })
+  })
```
IdoZakk, line 118: Add tests for no account info and no applications in the account info

IdoZakk (commented):
Nice!
Please add also a test for `deploy_permission_shceme_filter` (like you added for hierarchy filter)

packages/jira-adapter/test/filters/permission_scheme/deploy_permission_scheme.test.ts
```diff
@@ -77,6 +78,44 @@ describe('deploy permission scheme', () => {
     })
     expect(leftoverChanges).toEqual(changes)
   })
+  it('should not mock deploy on paid license due to account info with some paid app', async () => {
```
IdoZakk, line 81: Title should be should not deploy..., I see it is mock deploy in other tests but there is no reason for that.....

IdoZakk (approved):
💣  see one small comment

PR #4824, opened 2023-09-03
SALTO 2602: update bundle generation script and type

Shulik95:
_Bundle generation script updated, bundleType changed_

---

_The ticket - https://salto-io.atlassian.net/browse/SALTO-2602
**background** - Because relying on the NetSuite bundle search doesn't necessarily give the correct result, The script is changed to directly access bundles using its url. The log for installed bundles is updated to return the bundle_id, fetched_from and publisher_id which are required to access bundle's url directly. Also, this saves us some overhead when running the script.
The bundle type and schema are updated for better cover and now differentiate between the type returned from our suiteapp and the full type we late create._

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


dantal4 (approved):
LGTM 👍

PR #4825, opened 2023-09-03
SALTO-4633: fix removal of a field and conditions from a form at the same time

edenhassid:
fix removal of a field and conditions from a form at the same time

---

---
_Release Notes_: 
zendesk:
* Fix deployment error where removal of a field and relevant conditions from a form at the same time would return an error
---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/filters/ticket_form.ts
+++ b/packages/zendesk-adapter/src/filters/ticket_form.ts
@@ -130,12 +130,15 @@ const returnValidInstance = (inst: InstanceElement): InstanceElement => {
 
 // this function returns an instance that contains the removed field. This is because zendesk does not allow removing
 // field and condition at the same time
-const getChangeWithoutRemovedFields = (change: ModificationChange<InstanceElement>): InstanceElement => {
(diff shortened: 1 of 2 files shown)


packages/zendesk-adapter/src/filters/ticket_form.ts
```diff
@@ -121,6 +128,19 @@ const returnValidInstance = (inst: InstanceElement): InstanceElement => {
...
 }
 
+// this function returns an instance that contains the removed field. This is because zendesk does not allow removing
+// field and condition at the same time
+const getChangeWithoutRemovedFields = (change: ModificationChange<InstanceElement>): InstanceElement => {
+  const { before } = change.data
+  const { after } = change.data
+  const beforeFields: number[] = before.value.ticket_field_ids ?? []
+  const afterFields = new Set(after.value.ticket_field_ids ?? [])
+  const removedFields = beforeFields.filter(field => !afterFields.has(field))
```
shir-reifenberg, line 138: I think we can avoid the extra API call if there are no removedFields, but up to you

shir-reifenberg (approved):
Added one suggestion, besides this LGTM 😎

PR #4826, opened 2023-09-03
SALTO-4722: Allow silent DC plugin upgrades

IdoZakk:
Do not raise fetch errors if dc plugin version is newer to allow updates

---

We need to update the supported versions of Jira DC. For that we need to issue a new plugin version.
Currently, we have a problem- we issue warnings on plugin versions which are higher or lower than expected. It means we need to issue both a new SAAS version and a plugin version at the same time, and we cannot do it.

The solution is to remove the info warning if the plugin version is higher than expected, as the warning is for CLI users and I think the amount of possible users for CLI against DC is small, and the loss is minimal.

---
_Release Notes_: 
None

---
_User Notifications_: 
None


alonstern (approved):
👍

PR #4828, opened 2023-09-03
SALTO-4607: Add Author Information on Salesforce Sub-Instances

tamtamirr:
Add Author Information on Salesforce Sub-Instances.

---

Sub-instances in Salesforce were missing **Author Information** since we'd only call **client.list** on their containing type but not on their own type.

You can see all the **Author Information** that were added on my Environment in this file: [added.log](https://github.com/salto-io/salto/files/12506342/added.log)

---
_Release Notes_: 
Salesforce Adapter:
- Add author information on **Sub-Instances**.

---
_User Notifications_: 
_None_


packages/salesforce-adapter/test/filters/author_information/validation_rules.test.ts
tamtamirr, line 1: Filter was removed

packages/salesforce-adapter/src/filters/author_information/validation_rules.ts
tamtamirr, line 1: This filter is covered by the new **nestedInstancesAuthorInformationFilter**

packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -105,6 +105,47 @@ export const safeApiName = async (elem: Readonly<Element>, relative = false): Pr
...
+export const apiNameSync = (elem: Readonly<Element>, relative = false): string | undefined => {
+  const name = fullApiNameSync(elem)
+  return name && relative ? relativeApiName(name) : name
+}
+
+export const isMetadataInstanceElementSync = (elem: Element): elem is MetadataInstanceElement => (
+  isInstanceElement(elem)
+  && isMetadataObjectType(elem.getTypeSync())
+  && elem.value[INSTANCE_FULL_NAME_FIELD] !== undefined
+)
```
tamtamirr, line 146: I've made sync versions to existing methods to avoid unnecessary async code. I can also mark the old methods as `@deprecated` as part of this PR. WDYT?

tamtamirr (commented):
Added some comments

packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -68,12 +68,13 @@ import { JSONBool, SalesforceRecord } from '../client/types'
...
   apiName,
   defaultApiName,
-  isCustomObject,
+  isCustomObject, isMetadataObjectType,
   isNameField, MetadataInstanceElement,
   metadataType,
   MetadataValues,
   Types,
 } from '../transformers/transformer'
+import * as transformer from '../transformers/transformer'
```
benisraeladi, line 77: Why is this necessary? Why can't you change `relativeApiName` to be imported like the rest of the methods being used?

packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -105,6 +106,47 @@ export const safeApiName = async (elem: Readonly<Element>, relative = false): Pr
...
+
+const fullApiNameSync = (elem: Readonly<Element>): string | undefined => {
+  if (isInstanceElement(elem)) {
+    return (isCustomObjectSync(elem.getTypeSync()))
+      ? elem.value[CUSTOM_OBJECT_ID_FIELD] : elem.value[INSTANCE_FULL_NAME_FIELD]
+  }
+  return elem.annotations[API_NAME] ?? elem.annotations[METADATA_TYPE]
+}
+
+export const apiNameSync = (elem: Readonly<Element>, relative = false): string | undefined => {
```
amitgoldarb, line 138: This is a code dup of the original implementation. I think we should probably dry the 2 implementations. That's probably correct for a few other implementaions here.

packages/salesforce-adapter/src/filters/author_information/nested_instances.ts
```diff
@@ -0,0 +1,99 @@
...
+  client: SalesforceClient
+  typeName: NestedInstanceMetadataType
+  instances: MetadataInstanceElement[]
+}
+
+const setAuthorInformationForInstancesOfType = async (
+  {
+    client, typeName, instances,
+  }: SetAuthorInformationForTypeParams): Promise<void> => {
+  const { result: filesProps } = await client.listMetadataObjects([{ type: typeName }])
```
amitgoldarb, line 56: We are doing these lists serially. Should we do them in parellel?

amitgoldarb (commented):
Looks great! Added a few small comments

packages/salesforce-adapter/src/filters/utils.ts
tamtamirr: This is temp code, the other method should be deprecated. More over **fullApiName** shouldn't be exported like it it from the **transformer**. I'll mark the other methods as deprecated, as part of this PR

packages/salesforce-adapter/src/filters/utils.ts
tamtamirr: There's a property with name **relativeApiName** in some of the utils in this file. I use the implementation
of `relativeApiName` from the transformer module

packages/salesforce-adapter/src/filters/author_information/nested_instances.ts
tamtamirr: using Promise.all instead

benisraeladi (approved):
LGTM

PR #4830, opened 2023-09-04
SALTO-4544 add internal ID to custom fields

benisraeladi:
Custom fields of instances coming from managed installed packages did not save internal IDs. This fix should solve this issue.

---

---

_Release Notes_: 
Salesforce Adapter:
- Add internal ID to custom fields inside managed package if exists

---

_User Notifications_: 
Salesforce Adapter:
- Internal ID might be added to some custom fields


tamtamirr (approved):
LGTM 👑

PR #4832, opened 2023-09-04
SALTO-4718 - missing references in ticket_form

idoser96:
add missing references to ticket_form

---

has noise reduction

---
_Release Notes_: 
Zendesk Adapter:
* ticket_form will have now missing references for ticket_fields

---
_User Notifications_: 
None


edenhassid (approved):
🔥 🔥

PR #4834, opened 2023-09-04
SALTO-4688: add a function to get important attributes

edenhassid:
add a function to get important attributes form element

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/adapter-utils/src/important_values.ts
```diff
@@ -0,0 +1,77 @@
...
+      (isInstanceElement(element) ? element.value : element.annotations),
+      obj.value,
+      null
+    ),
+  )
+  return _.omitBy(finalImportantValues, _.isNull)
+}
+
+
+export const getImportantValues = async ({
```
netama, line 58: please add a docstring (that also explains the flag - continuing from above, I think it should probably be something like `indexedOnly`)

packages/adapter-utils/src/important_values.ts
```diff
@@ -0,0 +1,77 @@
...
+  indexed?: boolean
+}): Values => {
+  if (_.isEmpty(importantValues)) {
+    return {}
+  }
+  const relevantImportantValues = getRelevantIA(importantValues, indexed)
+  const finalImportantValues = _.mapValues(
+    _.keyBy(relevantImportantValues, obj => obj.value),
+    obj => _.get(
+      (isInstanceElement(element) ? element.value : element.annotations),
```
netama, line 49: 1. nit: maybe move this to before the `_.mapValues()` so that the condition is evaluated only once?
2. can also update later once we see a use case, but I'm wondering if for instances we don't want to _also_ support extracting annotation values?

packages/adapter-utils/test/important_values.test.ts
```diff
@@ -0,0 +1,177 @@
...
+    ],
+    [CORE_ANNOTATIONS.SELF_IMPORTANT_VALUES]: [
+      {
+        value: 'name',
+        indexed: false,
+      },
+      {
+        value: 'apiName',
+        indexed: true,
+      },
```
netama, line 80: maybe also add one that doesn't exist and cover with a test?

packages/adapter-utils/src/important_values.ts
```diff
@@ -0,0 +1,77 @@
...
+      obj.value,
+      null
+    ),
+  )
+  return _.omitBy(finalImportantValues, _.isNull)
```
netama, line 54: this might be ok, but how are we planning to distinguish between non-existing important attributes and between other values? based on the type annotations?

netama (approved):
added some nits but LGTM so pre-approving

packages/adapter-utils/src/important_values.ts
edenhassid: discussed IRL

PR #4835, opened 2023-09-04
SALTO-4731 - Remove content size limitation on client post requests

idoser96:
default limitation is 10MB, there is no reason to limit us at the client level

---

None

---
_Release Notes_: 
* Deploying now supports objects larger than 10MB

---
_User Notifications_: 
None


netama (approved):
LGTM assuming we've done sanity tests on a few cases, thanks!

PR #4837, opened 2023-09-05
SALTO-4732: "Used by" profile dependencies for new objects do not include field permissions, just object permissions

alonstern:
Added references between new objects to field permissions

---
_Release Notes_: 
_Salesforce Adapter_:
- Added references between new objects to field permissions

---
_User Notifications_: 
None


tomsellek (approved):
LGTM!

PR #4838, opened 2023-09-05
SALTO-4705 fix static files deserialization

dantal4:
Use the static-files-reviver when deserializing a single static file value (not in an element).

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/workspace/src/serializer/elements.ts
+++ b/packages/workspace/src/serializer/elements.ts
@@ -530,7 +530,7 @@ const generalDeserializeParsed = async <T>(
   if (!Array.isArray(parsed)) {
     throw new Error('got non-array JSON data')
   }
-  const elements = parsed.map(restoreClasses)
+  const elements = restoreClasses(parsed)
   if (staticFiles.length > 0) {
     await Promise.all(staticFiles.map(
       async ({ obj, key }) => {
--- a/packages/workspace/test/serializer/serializer.test.ts
+++ b/packages/workspace/test/serializer/serializer.test.ts
@@ -341,6 +341,15 @@ describe('State/cache serialization', () => {
       ])
       expect(mockStoreStaticFile).toHaveBeenCalledWith(staticFile)
     })
+    it('should deserialize static files - with a reviver', async () => {
+      const staticFile = new StaticFile({ filepath: 'abc', content: Buffer.from('hello world!') })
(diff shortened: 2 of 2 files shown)


packages/workspace/src/serializer/elements.ts
```diff
@@ -530,7 +533,7 @@ const generalDeserializeParsed = async <T>(
   if (!Array.isArray(parsed)) {
```
ori-moisis, line 533: if the plan is the support any kind of value in `serialize` / `deserialize`, don't we need to remove this check as well and support deserializing single values?

packages/workspace/src/serializer/elements.ts
```diff
@@ -530,7 +533,7 @@ const generalDeserializeParsed = async <T>(
...
     throw new Error('got non-array JSON data')
   }
-  const elements = parsed.map(restoreClasses)
+  const elements = await Promise.all(parsed.map(restoreClasses))
```
ori-moisis, line 536: I am a bit worried about always adding an `await` here in terms of performance since
1. in almost all cases there is no need to do that
2. we know async breaks can have a significant effect on performance
3. we call `deserialize` A LOT in basic and time-sensitive flows (especially compare / getPlan)

is there a way to maybe use the `staticFiles` list here somehow? mark the "top level" with a special parent or something?

packages/workspace/test/serializer/serializer.test.ts
```diff
@@ -336,9 +336,10 @@ describe('State/cache serialization', () => {
       const serialized = await serialize([staticFile], undefined, mockStoreStaticFile)
       expect(serialized).toContain('StaticFile')
       expect(serialized).not.toContain('internalContent')
-      expect(await deserializeValues(serialized)).toEqual([
-        new StaticFile({ filepath: staticFile.filepath, hash: staticFile.hash }),
-      ])
+      expect(await deserializeValues(serialized, async sf => (
+        sf.filepath === staticFile.filepath && sf.hash === staticFile.hash
+          ? staticFile : sf
+      ))).toEqual([staticFile])
```
ori-moisis, line 342: the reviver is a bit unclear, not sure I understand what changed in the test and why this is the correct thing to test

ori-moisis (approved):

PR #4842, opened 2023-09-05
SALTO-4720: fix bug in jql references

IdoZakk:
Fixed a bug in jql references

---
Verfifed on the service. recreation: Edit a reference in the jql part of an automation and deploy. The NaCl after the deployment will contain the value before the change.

---
_Release Notes_: 
Jira Adapter:
* Fixed a bug with deployments of jqls

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/jql/jql_references.ts
+++ b/packages/jira-adapter/src/filters/jql/jql_references.ts
@@ -13,8 +13,8 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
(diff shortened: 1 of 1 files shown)


packages/jira-adapter/src/filters/jql/jql_references.ts
```diff
@@ -150,54 +150,44 @@ const filter: FilterCreator = ({ config }) => {
     preDeploy: async changes => {
       await awu(changes)
```
alonstern, line 151: do we still need to use awu?

alonstern (approved):
Can we add a UT the reproduce this?

PR #4854, opened 2023-09-10
SALTO-4749 fix author information bug

benisraeladi:
Only try to add author information for objects that receive it

---

---
_Release Notes_: 

---
_User Notifications_: _None_


tamtamirr (approved):
LGTM

PR #4855, opened 2023-09-10
SALTO-4755: Fetch fails on invalid cascading options

IdoZakk:
Seen on a customer's env.
The test recreates a similar case

---

None

---
_Release Notes_: 
Jira Adapter:
* Fixed a bug that caused fetch to fail for custom fields with parentless cascading options

---
_User Notifications_: 
None


packages/jira-adapter/src/filters/fields/field_structure_filter.ts
```diff
@@ -115,6 +115,17 @@ const addCascadingOptionsToOptions = (instance: InstanceElement): void => {
     .forEach((context: Values) => {
       const idToOption = _.keyBy(context.options, option => option.id)
 
+      const invalidCascadeOptionIds = context.options
```
alonstern, line 118: Performance nit: let's make this a set

alonstern (approved):
👌

PR #4856, opened 2023-09-10
INCIDENT-8074: Deleting issue types from the default issue type scheme fails

alonstern:
Fixed deleting an issue type and removing it from the issue type scheme

---

---
_Release Notes_: 
__Jira Adapter__:
- Fixed deleting an issue type and removing it from the issue type scheme

---
_User Notifications_: 
None


IdoZakk (approved):
😮

PR #4859, opened 2023-09-11
SALTO-4751: Set automation project references to be weak references

alonstern:
Set automation project references to be weak references

---

- Added in adapter components functions to help implement fixElements and getCustomReferneces in adapters
- Implemented fixElements and getCustomReferences to make automation projects weak references

---
_Release Notes_: 
__Jira Adapter__:
- `salto element fix` will now clean invalid projects from automation

---
_User Notifications_: 
None


packages/adapter-components/src/custom_references.ts
```diff
@@ -0,0 +1,33 @@
...
+
+const { awu } = collections.asynciterable
+
+const getReferenceInfoIdentifier = (ref: ReferenceInfo): string =>
+  `${ref.source.getFullName()}-${ref.target.getFullName()}`
+
+export const combineCustomReferenceGetters = (customReferenceGetters: GetCustomReferencesFunc[])
+: GetCustomReferencesFunc => async elements => {
+  const idToRef: Record<string, ReferenceInfo> = {}
+  await awu(customReferenceGetters).forEach(async customReferenceGetter => {
```
netama, line 28: this might rely on previous discussions, but is there a reason not to parallelize here?

packages/adapter-components/src/custom_references.ts
```diff
@@ -0,0 +1,33 @@
...
+    const refs = await customReferenceGetter(elements)
+    _.assign(idToRef, _.keyBy(refs, getReferenceInfoIdentifier))
```
netama, line 30: not sure I follow this one - is there something we get from maintaining `idToRef` that we wouldn't get from using `_.uniqBy`?

edit: also not sure about the logic for prioritizing, added a question on the tests

packages/adapter-components/src/custom_references.ts
```diff
@@ -0,0 +1,33 @@
+/*
```
netama, line 1: nit: let's add both of these files under the `references` folder (or a new one if you prefer)? seems they should be tied together and don't belong at the root of the package IMO

packages/adapter-components/src/element_fixers.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  const elementFixesByElemID = _.keyBy(
+    fixedElements,
+    element => element.elemID.getFullName()
+  )
+  return elements.map(element => elementFixesByElemID[element.elemID.getFullName()] ?? element)
+    .concat(fixedElements.filter(element => !elementIds.has(element.elemID.getFullName())))
+}
+
+export const combineElementFixers = (fixers: FixElementsFunc[]): FixElementsFunc => async elements =>
```
netama, line 36: I have a feeling there's some context here about what we aggregate and how, so would be better if @ori-moisis reviewed this part

packages/adapter-components/test/custom_references.test.ts
```diff
@@ -0,0 +1,58 @@
...
+      ]),
+    ])
+
+    const refs = await getCustomReferencesFunc([])
+
+    expect(refs).toEqual([
+      {
+        source: new ElemID('adapter', 'type1', 'instance', 'inst1'),
+        target: new ElemID('adapter', 'type1', 'instance', 'inst2'),
+        type: 'weak',
```
netama, line 49: looking now at the implementation, it seems this is a bit flaky - meaning, this will be chosen based on the order and not based on the "strength". is this really what we want? (and if we do define by strength - do we definitely want weak to win over strong?)
(not sure what the right answer is, and I'm guessing maybe you already discussed this with @ori-moisis - I get why we need to be able to fix these so weak should win, but it also seems a bit counter-intuitive, so if this is the desired behavior I think we should document it clearly)

packages/jira-adapter/src/weak_references/automation_projects.ts
```diff
@@ -0,0 +1,98 @@
...
+  return awu(automationProjects)
+    .map(async (proj, index) => (
+      isReferenceExpression(proj.projectId)
+        ? { source: instance.elemID.createNestedID(index.toString(), 'projectId'), target: proj.projectId.elemID, type: 'weak' as const }
+        : undefined))
+    .filter(values.isDefined)
+    .toArray()
+}
+
+export const getAutomationProjectsReferences: GetCustomReferencesFunc = async elements =>
```
netama, line 54: again this may be something you already discussed with @ori-moisis , but I thought we were going to export something that combines both together so that they're "hard to miss"? e.g. (could look different, just one way) - have from this file something exported like `{ finder: <>, fixer: <> }` where `finder` is optional, and then combine them by taking all the objects and extracting the finders in one and the fixers in the other

packages/jira-adapter/src/weak_references/automation_projects.ts
```diff
@@ -0,0 +1,98 @@
...
+      const automationProjects = instance.value.projects
+      if (!isAutomationProjects(automationProjects)) {
+        return undefined
+      }
+
+      const fixedInstance = instance.clone()
+      fixedInstance.value.projects = await awu(automationProjects)
+        .filter(async proj => (proj.projectId === undefined
+          || (
+            isReferenceExpression(proj.projectId)
```
netama, line 76: we don't need to check that this is really a weak reference somehow? (related to my previous question about what happens if a reference is marked both as weak and strong)

netama (commented):
very cool :) added some questions, I have a feeling a lot of this was already discussed with @ori-moisis so probably best that he at least goes over the mentions tomorrow.

nit re the release notes - can you please add the "Jira adapter" title so it's associated correctly?

packages/adapter-components/src/custom_references.ts
alonstern: Nope, no reason

packages/adapter-components/src/custom_references.ts
alonstern: `_.uniqBy` will get the first one, here we want last one with the same id to override its previous ones. I will add a comment

packages/adapter-components/test/custom_references.test.ts
alonstern: What I thought was that the last one should override the previous ones. I will add a comment. LMK if you disagree

packages/jira-adapter/src/weak_references/automation_projects.ts
netama: thanks for the explanation :)
if we go with the suggested `WeakReferenceHandler` , what kind of custom args do we need to pass? either way they will need to somehow be passed in from `getElementFixer`, no? is there an example of something that should only be passed to one fixer that is specific to it and not relevant for others? (I think anything "generic" like the element source / client / basically anything we pass to filters today can be passed to all of them, but I might be missing something)

I think what you did in the aggregation functions is good, was thinking of just adding some wrappers, for example

1. in the "handler" code, instead of exporting each function individually, exporting 
```
export const handler = {
  findWeakReferences: getAutomationProjectsReferences,
  removeWeakReferences: getElementFixer,
}
```

2. in the code that's currently under `index.ts`, adding
```
const allReferenceHandlers = [
  automationProjects,
  ...
]
```
and then the args for the functions become slightly more generic -

```
export const getCustomReferences = combineCustomReferenceGetters(allReferenceHandlers.map(handler => handler.findWeakReferences).filter(isDefined))

export const getElementFixer = (elementsSource: ReadOnlyElementsSource): FixElementsFunc => combineElementFixers(allReferenceHandlers.map(handler => handler.findWeakReferences(elementsSource))
```
(and can potentially also move to adapter-components as well if we generalize the args like you suggested)

what am I missing?

packages/adapter-components/src/references/element_fixers.ts
```diff
@@ -0,0 +1,50 @@
...
+  return elements.map(element => elementFixesByElemID[element.elemID.getFullName()] ?? element)
+    .concat(fixedElements.filter(element => !elementIds.has(element.elemID.getFullName())))
+}
+
+/**
+ * Combine several fixElements functions into one that will run all of them.
+ */
+export const combineElementFixers = (fixers: FixElementsFunc[]): FixElementsFunc => async elements =>
+  awu(fixers).reduce(
+    async (fixes, fixer) => {
```
netama, line 41: nit: usually it's preferred to have args/variables with names that are different by more than one word to avoid confusion :) any chance we can replace `fixes` with something else / longer?

packages/jira-adapter/src/weak_references/automation_projects.ts
netama: just in the context of validating assumptions / aligning with what we inform the user about, but I guess we can add it later as well

netama (approved):
very cool :)

PR #4860, opened 2023-09-11
Salto-4756 - Zendesk Adapter - Fix ticket_form schema guard and remove error login if it doesn't pass

idoser96:
* isTicketForm now allow unknown in order to not fail on every ticket form
* also added 'end_user_conditions' to the possible conditions
* no longer error log in case the schema doesn't pass (as it is not necessarily an error)

---

None

---
_Release Notes_: 
Zendesk Adapter:
* Fix warning on ticket_form deployments if the ticket form has custom ticket status, and the feature it not enabled

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
+++ b/packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
@@ -136,7 +136,9 @@ const isTicketFormWithCustomStatus = (change: Change<ChangeDataType>): boolean =
     return false
   }
 
-  return hasConditionWithCustomStatuses([...data.value.agent_conditions ?? [], ...data.value.end_user_conditions ?? []])
+  return hasConditionWithCustomStatuses(
+    (data.value.agent_conditions ?? []).concat(data.value.end_user_conditions ?? [])
+  )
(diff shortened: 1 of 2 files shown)


packages/zendesk-adapter/src/change_validators/custom_statuses_enabled.ts
```diff
@@ -134,11 +132,11 @@ const isTicketFormWithCustomStatus = (change: Change<ChangeDataType>): boolean =
   const data = getChangeData(change)
   if (data.elemID.typeName !== TICKET_FORM_TYPE_NAME
     || !isInstanceElement(data)
-    || !isTicketForm(data)) {
+    || !isTicketFormWithConditions(data)) {
     return false
   }
 
-  return hasConditionWithCustomStatuses(data.value.agent_conditions ?? [])
+  return hasConditionWithCustomStatuses([...data.value.agent_conditions ?? [], ...data.value.end_user_conditions ?? []])
```
edenhassid, line 139: can you use concat?

edenhassid (approved):
looks good! 🔥 😎  added one comment

PR #4861, opened 2023-09-11
SALTO-4757 add mergeDiffs util to lowerdash

dantal4:
Add a 3 way diff merger.

The logic in this file is taken from https://github.com/bhousel/node-diff3/blob/v3.1.2/index.mjs , I only transferred it to TS and refactored it.

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/lowerdash/jest.config.js
+++ b/packages/lowerdash/jest.config.js
@@ -26,10 +26,10 @@ module.exports = deepMerge(
     ],
     coverageThreshold: {
       global: {
-        branches: 95.7,
-        functions: 98.9,
-        lines: 99.3,
-        statements: 99.4,
+        branches: 99,
+        functions: 100,
+        lines: 100,
+        statements: 100,
       },
     },
   }
--- a/packages/lowerdash/jest.config.js
+++ b/packages/lowerdash/jest.config.js
@@ -28,7 +28,7 @@ module.exports = deepMerge(
       global: {
         branches: 95.7,
         functions: 98.9,
-        lines: 99.4,
+        lines: 99.3,
         statements: 99.4,
       },
     },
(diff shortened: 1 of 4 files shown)


packages/lowerdash/jest.config.js
```diff
@@ -26,10 +26,10 @@ module.exports = deepMerge(
...
     coverageThreshold: {
       global: {
-        branches: 99,
-        functions: 100,
-        lines: 100,
-        statements: 100,
+        branches: 95.7,
+        functions: 98.9,
+        lines: 99.3,
+        statements: 99.4,
```
tamtamirr, line 32: This is a drastic drop in coverage. I assume it's because a lot of the imported code from `node-diff3` wasn't tested. Another reason not to import it directly into our project. 🤔

tamtamirr (commented):
Assuming this code is taken from a pretty well contributed package, why we manually import all of it's code and not put it in our dependencies with a wrapping interface?

I'd expect the util **mergedDiffs** to invoke code from `node-diff3` as an implementation, but I don't see any reason to directly copy code from there. This means we un-necessarily maintain more code, which also seems quite complex as-well.

dantal4:
@tamtamirr the current version of node-diff3 isn't support in node 14, which is currently used in salto repo. the latest version that is supported in node 14 is from 2 years ago. IMO having that code in salto repo, written in TS and in the salto code style will help us be aware and follow on what it does, and adjust it for our needs.

tamtamirr:
@dantal4 
Re maintaining the code in salto if we choose to modify it later: It feels like a premature optimization for me for the time being, but I don't have a strong opinion here.
However, if we choose to put this code in **lowerdash** we probably want more coverage here, and some UTs are missing. The coverage downgrade is pointing on branches that weren't covered in the new code.

packages/lowerdash/jest.config.js
dantal4: raised it to the current coverage 👍

PR #4865, opened 2023-09-12
SALTO-4631: Extra Dependencies for current namespace only bugfix

tamtamirr:
The wrong field was queried which caused the queries to fail when running with the optional feature `toolingDepsOfCurrentNamespace`

---

_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


benisraeladi (commented):
👍

benisraeladi (approved):

PR #4867, opened 2023-09-12
SALTO-4724 - Zendesk Adapter - prevent conditional ticket field if it is not in a ticket field list

idoser96:
Prevent adding a conditional ticket field to a form if the ticket field doesn't exists in the ticket fields list

---

Message still in review with PM

---
_Release Notes_: 
Zendesk Adapter:
* Prevent adding a conditional ticket field to a form if the ticket field doesn't exists in the ticket fields list

---
_User Notifications_: 
None


edenhassid (commented):
good work! 🌵  added small comments

edenhassid (approved):
LGTM! 🚀

PR #4868, opened 2023-09-12
SALTO-2349: Added references mapping to supported JSM types

idoamit8:
In this PR I added reference mapping for the supported JSM types

---
Please review only from commit a6fa09fa8ab2825d5702abafaca5dc7f9b7c6eee since commit 1 is reviewed in the previous PR. 

---
_Release Notes_: 
_Jira_:
Supported JSM types will contain reference mapping.

---
_User Notifications_: 
_Jira Adapter_:
Supported JSM types will contain reference mapping.


packages/jira-adapter/src/reference_mapping.ts
```diff
@@ -959,6 +959,48 @@ export const referencesRules: JiraFieldReferenceDefinition[] = [
...
+    jiraSerializationStrategy: 'key',
+    jiraMissingRefStrategy: 'typeAndValue',
+    target: { type: PROJECT_TYPE },
+  },
+  {
+    src: { field: 'projectKey', parentTypes: [REQUEST_TYPE_NAME] },
+    jiraSerializationStrategy: 'key',
+    jiraMissingRefStrategy: 'typeAndValue',
+    target: { type: PROJECT_TYPE },
+  },
```
netama, line 979: pretty sure you can combine these (plus the two below)
```suggestion
  {
    src: { field: 'projectKey', parentTypes: [CUSTOMER_PERMISSIONS_TYPE, QUEUE_TYPE, REQUEST_TYPE_NAME, REQUEST_TYPE_GROUP_TYPE, CALENDAR_TYPE] },
    jiraSerializationStrategy: 'key',
    jiraMissingRefStrategy: 'typeAndValue',
    target: { type: PROJECT_TYPE },
  },
```

netama (commented):
looks good :) added a suggestion (and btw this doesn't have to be rebased since there are no code dependencies, I think you can create this as a "flat" PR above main if you want)

packages/jira-adapter/src/reference_mapping.ts
```diff
@@ -959,6 +959,24 @@ export const referencesRules: JiraFieldReferenceDefinition[] = [
...
+  {
+    src: { field: 'issueTypeId', parentTypes: [REQUEST_TYPE_NAME] },
+    serializationStrategy: 'id',
+    jiraMissingRefStrategy: 'typeAndValue',
+    target: { type: ISSUE_TYPE_NAME },
+  },
```
IdoZakk, line 973: Can be merged with the previous `issueTypeId` (when adding a reference look if the field already exists with the relevant parameters)
Note that we do not use missing ref strategy as the issue type ID can be a string

IdoZakk (commented):
🎸 
Not approving yet as it contains the rest of the code

IdoZakk (approved):
Make sure to first merge the initial fetch, rebase and then commit this one

PR #4869, opened 2023-09-12
SALTO-4652 - Salesforce: add a fetch warning for custom fields that are not queryable

tomsellek:
When the user fetches data records that have some non-queryable fields, the result may be fear, uncertainty and confusion during deploy - the fields will be deployed as deleted.
Let's warn the user about fetching data records with inaccessible fields. But not if they're hidden, since the user can't see them anyway.

---

---
_Release Notes_: 
Salesforce: Warn when fetching data records with non-queryable fields.

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/config_doc.md
+++ b/packages/salesforce-adapter/config_doc.md
@@ -135,16 +135,17 @@ salesforce {
 
 ## Fetch configuration options
 
-| Name                                           | Default when undefined  | Description                                                                                                                                                                                                           |
-|------------------------------------------------|-------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| [metadata](#metadata-configuration-options)    | Fetch all metdata       | Specified the metadata fetch                                                                                                                                                                                          |
+| Name                                      | Default when undefined  | Description                                                                                                                                                                                                           |
+|-------------------------------------------|-------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| [metadata](#metadata-configuration-options) | Fetch all metdata       | Specified the metadata fetch                                                                                                                                                                                          |
 | [data](#data-management-configuration-options) | {} (do not manage data) | Data management configuration object names will not be fetched in case they are matched in includeObjects                                                                                                             |
-| fetchAllCustomSettings                         | true                    | Whether to fetch all the custom settings instances. When false, it is still possible to choose specific custom settings instances via the `data` option                                                               |
-| [optionalFeatures](#optional-features)         | {} (all enabled)        | Granular control over which features are enabled in the adapter, by default all features are enabled in order to get the most information. can be used to turn off features that cause problems until they are solved |
-| maxInstancesPerType                            | 5000                    | Do not fetch metadataTypes and CustomObjects with more instances than this number, and add those to the exclude lists                                                                                                 |
-| preferActiveFlowVersions                       | false                   | When set to false, flows' latest version will be fetched. Otherwise, flows' active version will be fetched if exists                                                                                                  |
-| addNamespacePrefixToFullName                   | true                   | When set to true, namespace prefix will be added to instances in a namespace whose fullName does not begin with the namespace. Otherwise, there will be no change to fullName                                               |
-## Metadata configuration options
+| fetchAllCustomSettings                    | true                    | Whether to fetch all the custom settings instances. When false, it is still possible to choose specific custom settings instances via the `data` option                                                               |
+| [optionalFeatures](#optional-features)    | {} (all enabled)        | Granular control over which features are enabled in the adapter, by default all features are enabled in order to get the most information. can be used to turn off features that cause problems until they are solved |
+| maxInstancesPerType                       | 5000                    | Do not fetch metadataTypes and CustomObjects with more instances than this number, and add those to the exclude lists                                                                                                 |
+| preferActiveFlowVersions                  | false                   | When set to false, flows' latest version will be fetched. Otherwise, flows' active version will be fetched if exists                                                                                                  |
+| addNamespacePrefixToFullName              | true                    | When set to true, namespace prefix will be added to instances in a namespace whose fullName does not begin with the namespace. Otherwise, there will be no change to fullName                                         |
+| [warningSettings](#warning-settings)      | {}                      | Enable/disable specific warnings                                                                                                                                                                                      |
+### Metadata configuration options
 
 | Name                           | Default when undefined       | Description                                                                                                                                                     |
 |--------------------------------|------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------|
@@ -152,14 +153,14 @@ salesforce {
 | [exclude](#metadata-query)     | [] (Exclude nothing)         | Specified the metadata to not fetch. Metadata that matches any of the exclude criteria will not be fetched even if it also matches some of the include criteria |
 | objectsToSeperateFieldsToFiles | [] (Don't split any objects) | Specified a list of objects which will be stored as one field per nacl file                                                                                     |
 
-## Metadata Query
+#### Metadata Query
 | Name         | Default when undefined | Description                                                    |
 |--------------|------------------------|----------------------------------------------------------------|
 | namespace    | ".*" (All namespaces)  | A regular expression of a namespace to query with              |
 | metadataType | ".*" (All types)       | A regular expression of a metadata type to query with          |
 | name         | ".*" (All names)       | A regular expression of a metadata instance name to query with |
 
-## Optional Features
+### Optional Features
 
 | Name                              | Default when undefined | Description                                                                                                                                                           |
 |-----------------------------------|------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------|
@@ -175,6 +176,14 @@ salesforce {
 | generateRefsInProfiles            | false                  | Generate references from profiles. This will have a significant performance impact, but will provide references from profiles to fields and elements.                 |
 | skipAliases                       | false                  | Do not create aliases for Metadata Elements                                                                                                                           |
 
+
+### Warning settings
+
+| Name                | Default when undefined | Description                                                       |
+|---------------------|------------------------|-------------------------------------------------------------------|
+| nonQueryableFields  | true                   | Warn when fetching records of an object with non-queryable fields |
+
+
 ### Data management configuration options
 
 | Name                                                                          | Default when undefined                           | Description                                                                                               |
@@ -185,9 +194,13 @@ salesforce {
 | [saltoIDSettings](#salto-id-settings-configuration-options)                   | N/A (required when dataManagement is configured) | Configuration for cross environments data record ids management                                           |
 | [saltoAliasSettings](#salto-alias-settings-configuration-options)             | N/A                                              | Configuration for data record aliases                                                                     |
 | [saltoManagementFieldSettings](#salto-management-field-configuration-options) | {}                                               | Configuration for managed-by-Salto field                                                                  |
+<<<<<<< HEAD
 | [brokenOutgoingReferencesSettings](#broken-outgoing-references-settings)      | {}                                               | Configuration for handling broken references                                                              |
 | omittedFields                                                                 | []                                               | List of API names of fields to discard when fetching data records.                                        |
 | [warningSettings](#data-warning-settings)                                     | {}                                               | Enabling/disabling specific warnings                                                                      | 
+=======
+| [brokenOutgoingReferencesSettings](#broken-outgoing-references-settings)      | {}                                               | Configuration for handling broken references                                                              | 
+>>>>>>> 3efcfb758 (Move config to fetchProfile)
 
 #### Salto ID settings configuration options
 
@@ -223,14 +236,6 @@ salesforce {
 | "ExcludeInstance" | Do not fetch instances that contain a reference whose target was not fetched                      |
 | "BrokenReference" | Fetch the instance and create Salto references to non-existant targets.                           |
 | "InternalId"      | Fetch the instance and keep the existing field value (the internal ID of the referenced instance) |
-
-
-#### Data warning settings
-
-| Name                | Default when undefined | Description                                                       |
-|---------------------|------------------------|-------------------------------------------------------------------|
-| nonQueryableFields  | true                   | Warn when fetching records of an object with non-queryable fields |
-
 
 #### Object ID settings configuration options
 
--- a/packages/salesforce-adapter/config_doc.md
+++ b/packages/salesforce-adapter/config_doc.md
@@ -187,7 +187,7 @@ salesforce {
 | [saltoManagementFieldSettings](#salto-management-field-configuration-options) | {}                                               | Configuration for managed-by-Salto field                                                                  |
 | [brokenOutgoingReferencesSettings](#broken-outgoing-references-settings)      | {}                                               | Configuration for handling broken references                                                              |
 | omittedFields                                                                 | []                                               | List of API names of fields to discard when fetching data records.                                        |
-
+| [warningSettings](#data-warning-settings)                                     | {}                                               | Enabling/disabling specific warnings                                                                      | 
 
 #### Salto ID settings configuration options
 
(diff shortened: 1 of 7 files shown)


packages/salesforce-adapter/src/filters/custom_objects_instances.ts
```diff
@@ -534,6 +541,14 @@ const createInvalidManagedBySaltoFieldFetchWarning = async (
   severity: 'Warning',
 })
 
+const createInaccessibleFieldsFetchWarning = async (
+  objectType: ObjectType,
+  inaccessibleFields: string[],
+): Promise<SaltoError> => ({
+  message: `The following fields are not accessible and will not appear in records of type ${await apiName(objectType)}: ${inaccessibleFields.join(',')}'`,
```
adamro, line 548: Product questions honestly but I will shoot here :) 

Are we sure we want this in the field level and not aggregated by type?

Did we do a check to see how many of this will appear in current user's Fetches with legit permissions? I am worried we are going to create a lot of noise here

Also, till now, a Warning on Fetch meant we didn't Fetch something. This is not it. I would argue this more of an "Info" level thing.

cc: @tomermevorach

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
```diff
@@ -520,6 +520,13 @@ const filterTypesWithManyInstances = async (
   }
 }
 
+const getInaccessibleCustomFields = (objectType: ObjectType): string[] => (
+  Object.entries(objectType.fields)
+    .filter(([, field]) => !isQueryableField(field))
+    .filter(([, field]) => !field.annotations[CORE_ANNOTATIONS.HIDDEN_VALUE])
```
adamro, line 526: I am actually not sure, but are the hidden definition of CustomField actually set at this stage of the filters?

tamtamirr: No, FYI @tomsellek we have a logic that sets fields that are not **createable and updateable** as hidden in the hideReadOnlyValuesFilter that runs after [this](https://github.com/salto-io/salto/blob/ca1d501cd346958228ba116b930627dda520d24c/packages/salesforce-adapter/src/adapter.ts#L166) one.

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
```diff
@@ -520,6 +520,13 @@ const filterTypesWithManyInstances = async (
...
+    .map(([name]) => name)
```
tamtamirr, line 527:
```suggestion
  Object.values(objectType.fields)
    .filter(field) => !isQueryableField(field))
    // No reason to notify users on fields that their value won't be visible
    .filter((field) => !field.annotations[CORE_ANNOTATIONS.HIDDEN_VALUE])
    .map(apiNameSync)
    .filter(isDefined)
```
1. While they're probably the same, it's better to use the **apiName** of a field than the field name in the **ObjectType**.
2. Added a small comment why we filter out hidden value fields

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
tamtamirr: +1 on changing the severity to **info**

tamtamirr (commented):
Added some small comments. Looking good :)

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
tomsellek: @tomermevorach agrees with **info**. I will make the change.

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
tomsellek: Ack. Updated the code.

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
tomsellek: Makes sense

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
tomsellek: Also, this already aggregates by type?

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
```diff
@@ -517,6 +532,14 @@ const filterTypesWithManyInstances = async (
   }
 }
 
+const getInaccessibleCustomFields = (objectType: ObjectType): string[] => (
+  Object.values(objectType.fields)
+    .filter(field => !isQueryableField(field))
+    .filter(field => !isSystemField(field)) // the value of system fields won't be visible anyway
```
tamtamirr, line 538: The terminology here is not very accurate. Some **System Fields** are both modifiable and are not hidden (e.g. **Name** and **CurrenyIsoCode**). I suggest we remove the wrapping method and inline it with

```suggestion
    // At the end of the fetch, the values of these fields won't be visible as we later hide read-only fields
    .filter(field => !isHiddenField(field) && !isReadOnlyField(field))
```

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
```diff
@@ -531,6 +554,14 @@ const createInvalidManagedBySaltoFieldFetchWarning = async (
   severity: 'Warning',
 })
 
+const createInaccessibleFieldsFetchWarning = async (
+  objectType: ObjectType,
+  inaccessibleFields: string[],
+): Promise<SaltoError> => ({
+  message: `The following fields are not accessible and will not be fetched for records of type ${await apiName(objectType)}: ${inaccessibleFields.join(',')}. Values of these fields will appear as 'deleted' if these data records are ever deployed`,
```
tamtamirr, line 561:
```suggestion
  message: `The following fields are not accessible and will not be fetched for records of type ${apiNameSync(objectType)}: ${inaccessibleFields.join(',')}. Values of these fields will appear as 'deleted' if these data records are ever deployed`,
```

We can make this sync now 🥳

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
```diff
@@ -584,6 +615,24 @@ const filterCreator: RemoteFilterCreator = ({ client, config }) => ({
       .filter(setting => setting.invalidManagedBySaltoField !== undefined)
       .map(createInvalidManagedBySaltoFieldFetchWarning)
 
+    const typesOfFetchedInstances = new Set(
+      elements
+        .filter(isInstanceElement)
+        .map(instance => instance.getTypeSync())
+    )
```
tamtamirr, line 622:
```suggestion
    const typesOfFetchedDataInstances = new Set(
      elements
        .filter(isInstanceElement)
        .map(instance => instance.getTypeSync())
        .filter(objectType => isCustomObjectSync(objectType))
    )
```
Isn't this logic relevant for Data instances only?

packages/salesforce-adapter/config_doc.md
```diff
@@ -220,6 +221,13 @@ salesforce {
 | "InternalId"      | Fetch the instance and keep the existing field value (the internal ID of the referenced instance) |
 
 
+#### Data warning settings
+
+| Name                | Default when undefined | Description                                                       |
+|---------------------|------------------------|-------------------------------------------------------------------|
+| nonQueryableFields  | true                   | Warn when fetching records of an object with non-queryable fields |
+
+
```
tamtamirr, line 230: Do we really need a whole new section in the config for warnings? (instead of setting this as optional feature)
And if we do, isn't it better if it sits down under the **fetchConfig** directly instead of **dataManagement**? We can have this mechanism for Metadata fetches as-well 🤔 @tomermevorach @pgonzaleznetwork WDYT?

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
```diff
@@ -584,6 +615,24 @@ const filterCreator: RemoteFilterCreator = ({ client, config }) => ({
...
+
+    if (config.fetchProfile.dataManagement?.isWarningEnabled('nonQueryableFields') ?? false) {
+      invalidPermissionsWarnings = awu(customObjectFetchSetting)
+        .map(fetchSettings => fetchSettings.objectType)
+        .filter(isCustomObject)
+        .map(objectType => ({ type: objectType, fields: getInaccessibleCustomFields(objectType) }))
+        .filter(({ fields }) => fields.length > 0)
+        .filter(({ type }) => typesOfFetchedInstances.has(type))
+        .map(({ type, fields }) => createInaccessibleFieldsFetchWarning(type, fields))
+    }
```
tamtamirr, line 634: This whole code can be sync now :)

packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -183,7 +183,7 @@ export const isHiddenField = (field: Field): boolean => (
 )
 
 export const isReadOnlyField = (field: Field): boolean => (
-  !field.annotations[FIELD_ANNOTATIONS.CREATABLE] && !field.annotations[FIELD_ANNOTATIONS.UPDATEABLE]
+  field.annotations[FIELD_ANNOTATIONS.CREATABLE] === false && field.annotations[FIELD_ANNOTATIONS.UPDATEABLE] === false
```
tamtamirr, line 186: 💘

packages/salesforce-adapter/test/filters/custom_objects_instances.test.ts
```diff
@@ -1616,7 +1662,7 @@ describe('Custom Object Instances filter', () => {
 
     it('Should warn if the field is not queryable', () => {
       expect(filterResult.errors ?? []).not.toBeEmpty()
-      expect(filterResult.errors).toEqual([{
+      expect(filterResult.errors).toIncludeAllPartialMembers([{
```
tamtamirr, line 1665: Why did we change to a less strict matcher here?

packages/salesforce-adapter/test/filters/custom_objects_instances.test.ts
```diff
@@ -587,15 +588,60 @@ describe('Custom Object Instances filter', () => {
...
+            message: expect.stringContaining(includeObjectName) && expect.stringContaining('NonQueryable'),
+            severity: 'Info',
+          }])
+          expect(fetchResult.errors?.[0].message).not.toInclude('HiddenNonQueryable')
+        })
+        it('should not issue a warning if there are no instances of the object', () => {
+          expect(fetchResult.errors).not.toIncludeAllPartialMembers([{
+            message: expect.stringContaining(refToObjectName),
+          }])
+        })
```
tamtamirr, line 644: We lack a UT that validates the warning is not created when it's disabled in the config

tamtamirr (commented):
Code looks clean and readable. Good work 👑 
Added some comments.

Can we attach example of the warning in the PR description?

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
tomsellek: Ack

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
tomsellek: I dunno, is it?

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
tomsellek: Yaaaaasssss

packages/salesforce-adapter/test/filters/custom_objects_instances.test.ts
tomsellek: This test validates another warning - when the "managed by salto" field is non-queryable. Before this PR, this condition would result in a single warning. Now there will be two warnings about the non-queryable field, and the test only validates one of them.

packages/salesforce-adapter/test/filters/custom_objects_instances.test.ts
tomsellek: That would require us to duplicate the entire beforeEach logic just for this one test case. Do you think it's worth it?

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
tamtamirr: It is :)

packages/salesforce-adapter/test/filters/custom_objects_instances.test.ts
tamtamirr: Yes. This branching is important

packages/salesforce-adapter/test/filters/custom_objects_instances.test.ts
tamtamirr: If this code yields two SaltoErrors, we should do one of the following:
1. Test them both in the same test.
2. Have different setup.

Strict equality is important, especially around **filter result**

pgonzaleznetwork:
I discussed this with @tomsellek and I think we need to make the error as clear as possible to the user, here's what I recommend

```
There are inaccessibleFields.length  fields in the ${apiNameSync(objectType)} object that the fetch user does not have access to. These are the fields: ${inaccessibleFields.join(',')}

If ${apiNameSync(objectType)} records are deployed from this environment, values of these fields will appear as deletions.
```

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
```diff
@@ -552,6 +571,14 @@ const createInvalidManagedBySaltoFieldFetchWarning = async (
   severity: 'Warning',
 })
 
+const createInaccessibleFieldsFetchWarning = (
+  objectType: ObjectType,
+  inaccessibleFields: string[],
+): SaltoError => ({
+  message: `The following fields are not accessible and will not be fetched for records of type ${apiNameSync(objectType)}: ${inaccessibleFields.join(',')}. Values of these fields will appear as 'deleted' if these data records are ever deployed`,
```
tomsellek, line 578: @tomermevorach your opinion is sought on the error message - either the one in the PR, or the (IMO better) one that @pgonzaleznetwork suggested:
```
There are ${inaccessibleFields.length} fields in the ${apiNameSync(objectType)} object that the fetch user does not have access to. These are the fields: ${inaccessibleFields.join(',')}

If ${apiNameSync(objectType)} records are deployed from this environment, values of these fields will appear as deletions.
```

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
tomsellek: Ack

tamtamirr (approved):
LGTM code-wise.

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
tomsellek: (approved by Tomer offline)

packages/salesforce-adapter/test/filters/custom_objects_instances.test.ts
tomsellek: OK, so I combined this comment with https://github.com/salto-io/salto/pull/4869#discussion_r1347449698 and ended up disabling the warning for the "managed by Salto" and verifying we don't get the warning in that case. Does this work?

packages/salesforce-adapter/test/filters/custom_objects_instances.test.ts
```diff
@@ -586,15 +591,62 @@ describe('Custom Object Instances filter', () => {
...
+              severity: 'Info',
+            }])
+            expect(fetchResult.errors?.[0].message).not.toInclude('HiddenNonQueryable')
+          })
+          it('should not issue a warning if there are no instances of the object', () => {
+            expect(fetchResult.errors).not.toIncludeAllPartialMembers([{
+              message: expect.stringContaining(refToObjectName),
+            }])
+          })
+        })
```
tamtamirr, line 649: The terminology here uses warnings while the severity is **Info**..?
Also, we still don't cover the case where the **warning is disabled**.

tamtamirr (commented):
small comment on the test

packages/salesforce-adapter/test/filters/custom_objects_instances.test.ts
tomsellek: As I mentioned in https://github.com/salto-io/salto/pull/4869#discussion_r1370367811, the case of the warning being disabled is covered in the test `Managed by Salto.Should warn if the field is not queryable` where the warning is disabled and we verify that it indeed does not appear in the filter result.
Ack on the terminology bit.

PR #4872, opened 2023-09-13
SALTO-4757 Auto-merge conflicts in value

dantal4:
except for auto-merging conflicts in the semantic level (when different fields were changed), we should also auto-merge conflicts in value - in static files / multi-line strings.

This is a 3 way diff merge when the base is the state version (or PR’s before), and the two other versions are the new service value and the pending value in the nacl.

---

---
_Release Notes_: 
Core:
- Auto-merge conflicts in static-files / multi-line strings

---
_User Notifications_: 
- Pending changes in static-files / multi-line strings will be merge with service changes, if there are no conflicts.


packages/core/test/core/fetch.test.ts
```diff
@@ -818,6 +822,126 @@ describe('fetch', () => {
...
+          const result = await fetchChanges(
+            mockAdapters,
+            createElementSource([workspaceInstance]),
+            createElementSource([stateInstance]),
+            { [testID.adapter]: 'dummy' },
+            [],
+          )
+          changes = [...result.changes]
+        })
+        it('should merge content if possible', async () => {
```
ori-moisis, line 870: 1. too much copy-pasting IMO, I think we can extract most of the code here to be the same with a boolean to just determine if the content is in a static file or a string
2. missing tests for fairly trivial things like - there should be a test that checks we don't merge when there is a conflict on the same line. I am sure if we think about it a bit more we can find more scenarios that need to be covered

packages/core/src/core/fetch.ts
```diff
@@ -15,6 +15,7 @@
 */
 import wu from 'wu'
 import _ from 'lodash'
+import * as diff3 from 'node-diff3'
```
ori-moisis, line 18: seems like we only use `mergeDiff3` from here, why do we need to import `*`?

packages/core/src/core/fetch.ts
```diff
@@ -170,6 +173,68 @@ const addFetchChangeMetadata = (
...
+  pendingChanges: [],
+})
+
+const resolveConflict: ChangeTransformFunction = async change => {
+  const { serviceChanges, pendingChanges = [] } = change
+  if (serviceChanges.length !== 1 || pendingChanges.length !== 1
+  || !change.change.id.isEqual(serviceChanges[0].id) || !change.change.id.isEqual(pendingChanges[0].id)
+  || !isModificationChange(serviceChanges[0]) || !isModificationChange(pendingChanges[0])) {
+    return [change]
+  }
```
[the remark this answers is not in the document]
dantal4: validate that we're dealing with a flat change, with one service change and one pending, and they are all modifications, because otherwise we won't have 3 versions, right?

packages/core/src/core/fetch.ts
```diff
@@ -170,6 +173,68 @@ const addFetchChangeMetadata = (
...
+  const versions = [
+    // current
+    serviceChanges[0].data.after,
+    // base
+    serviceChanges[0].data.before,
+    // incoming
+    pendingChanges[0].data.after,
```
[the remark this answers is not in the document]
dantal4: I looked at it as rebase - where the current version is the new code from master... I can switch them.

packages/core/src/core/fetch.ts
dantal4: I preferred the `diff3.mergeDiff3(...)` usage style, to make it clear that we're using an external util. IDM changing it.

packages/core/test/core/fetch.test.ts
```diff
@@ -818,6 +822,126 @@ describe('fetch', () => {
...
+          expect(getChangeData(unmergeableChange.change)).toEqual(serviceStaticFile)
+          expect(otherFileChange.pendingChanges).toHaveLength(1)
+          expect(otherFileChange.change.id).toEqual(testID.createNestedID('instance', instanceName, 'otherFile'))
+          expect(getChangeData(otherFileChange.change)).toEqual(serviceStaticFile)
+        })
+      })
+      describe('when the working copy has some mergeable changes in multiline strings', () => {
+        const instanceName = 'name'
+        const stateString = 'hello world!\nmy name is:\nNaCls'
+        const serviceString = 'Hello World!\nmy name is:\nNaCls'
```
tamtamirr, line 890: nit: Why are these consts in camelCase and not UPPER_CASE?

packages/core/test/core/fetch.test.ts
```diff
@@ -818,6 +822,126 @@ describe('fetch', () => {
...
+            [],
+          )
+          changes = [...result.changes]
+        })
+        it('should merge content if possible', async () => {
+          expect(changes).toHaveLength(3)
+          const [mergeableChange, unmergeableChange, otherFieldChange] = changes
+          expect(mergeableChange.pendingChanges).toHaveLength(0)
+          expect(mergeableChange.change.id).toEqual(testID.createNestedID('instance', instanceName, 'mergeableContent'))
+          expect(getChangeData(mergeableChange.change)).toEqual('Hello World!\nmy name is:\nNaCls the great!')
```
tamtamirr, line 936:
```suggestion
          expect(mergeableChange).toEqual(expect.objectContaining({
            pendingChanges: [],
            change: {
              id: testID.createNestedID('instance', instanceName, 'mergeableContent'),
              data: expect.objectContaining({
                after: 'Hello World!\nmy name is:\nNaCls the great!'
              })
            },
          }))
```
Nit: I find having too many expects on the same Change Object in this test(s) confusing. Consider the following instead.

tamtamirr (commented):
Added some questions/nits :)

packages/core/src/core/fetch.ts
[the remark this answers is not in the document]
dantal4: `array.every(...)` provides a type guard which was useful later

packages/core/test/core/fetch.test.ts
dantal4: AFAIK upper case style is used for global consts (in the module level) and camel case for any variables that are inside scopes, even if it's a const.

packages/core/src/core/fetch.ts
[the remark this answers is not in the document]
tamtamirr: I see. We could also implement this simple typegaurd for that use-case (totally up to you though, I just find working with this array quite confusing): 

```
const everyRecordValues = (record: Record<string, unknown>, typeGuard: (value: unknown) => value is T): record is Record<string, T> => Object.values(record).every(typeGuard)
```

packages/core/src/core/fetch.ts
```diff
@@ -170,6 +176,100 @@ const addFetchChangeMetadata = (
...
+  change.serviceChanges.length === 1
+  && change.pendingChanges?.length === 1
+  && change.change.id.isEqual(change.serviceChanges[0].id)
+  && change.change.id.isEqual(change.pendingChanges[0].id)
+  && isModificationChange(change.serviceChanges[0])
+  && isModificationChange(change.pendingChanges[0])
+
+const toMergedTextChange = (
+  change: FetchChange,
+  contents: string[],
```
ori-moisis, line 193: this interface allows passing any number of strings, but we really only want to support 3
maybe we should get this explicitly as 3 arguments?

packages/core/src/core/fetch.ts
```diff
@@ -170,6 +176,100 @@ const addFetchChangeMetadata = (
...
+    const { filepath, encoding } = versions[0]
+    if (!versions.every(file => file.filepath === filepath && file.encoding === encoding)) {
+      log.warn(
+        'skipping merge of %s since static files filepath & encoding does not match',
+        change.change.id.getFullName()
+      )
+      return [change]
+    }
+    const contents = await awu(versions)
+      .map(async file => await file.getContent() ?? Buffer.from(''))
```
ori-moisis, line 256: is there a case where this fallback to `Buffer.from('')` valid?

packages/core/src/core/fetch.ts
```diff
@@ -170,6 +176,100 @@ const addFetchChangeMetadata = (
...
+      .map(buffer => buffer.toString(encoding))
+      .toArray()
+    return [
+      toMergedTextChange(
+        change,
+        contents,
+        merged => new StaticFile({ filepath, content: Buffer.from(merged, encoding), encoding })
+      ),
+    ]
+  }
```
ori-moisis, line 266: I think we should avoid trying to merge binary files, feels like it would break things if we tried it
lets add a test for that as well

packages/core/src/core/fetch.ts
```diff
@@ -170,6 +176,100 @@ const addFetchChangeMetadata = (
...
+  if (!isThreeWayDiffChange(change)) {
+    return [change]
+  }
+  const versions = [
+    // current
+    change.serviceChanges[0].data.after,
+    // base
+    change.serviceChanges[0].data.before,
+    // incoming
+    change.pendingChanges[0].data.after,
```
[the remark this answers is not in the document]
dantal4: I commented [here](https://github.com/salto-io/salto/pull/4872#discussion_r1325589300). I'll change it 👍

packages/core/src/core/fetch.ts
```diff
@@ -170,6 +176,100 @@ const addFetchChangeMetadata = (
...
+  ]
+  if (versions.every(isStaticFile)) {
+    const { filepath, encoding } = versions[0]
+    if (!versions.every(file => file.filepath === filepath && file.encoding === encoding)) {
+      log.warn(
```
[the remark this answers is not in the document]
dantal4: It is just an unlikely behavior AFAIK, which I think should catch the eye when investogating. Only saw it happens in dummy.

packages/core/src/core/fetch.ts
dantal4: Waiting for your reply [here](https://github.com/salto-io/salto/pull/4872#discussion_r1328367254)

packages/core/src/core/fetch.ts
dantal4: Maybe if we fail to get the 'before' from S3. Do you think we should just return in that case?

packages/core/src/core/fetch.ts
dantal4: we won't be able to "enjoy" `array.every(versions)` as a type-guard if `versions` will be a tuple and not an array.

packages/core/src/core/fetch.ts
```diff
@@ -170,6 +176,100 @@ const addFetchChangeMetadata = (
...
+  serviceChanges: [ModificationChange<Value>]
+  pendingChanges: [ModificationChange<Value>]
+}
+const isThreeWayDiffChange = (change: FetchChange): change is ThreeWayDiffChange =>
+  change.serviceChanges.length === 1
+  && change.pendingChanges?.length === 1
+  && change.change.id.isEqual(change.serviceChanges[0].id)
+  && change.change.id.isEqual(change.pendingChanges[0].id)
+  && isModificationChange(change.serviceChanges[0])
+  && isModificationChange(change.pendingChanges[0])
```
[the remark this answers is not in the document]
dantal4: We can support it, but notice it won't be a 3 way diff merge. `node-diff3` has a function for that too.
@adamro I would like to hear your opinion here too 🙏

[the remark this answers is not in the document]
adamro: My intuition is that we should support it

packages/core/src/core/fetch.ts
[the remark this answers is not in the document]
ori-moisis: warnings are for cases where something fails but can be recovered, this doesn't seem like a failure case, this seems like something that could legitimately happen in a bunch of use cases and it is covered correctly by the logic here, it has no problematic effect.
this really doesn't seem like it should be a warning

packages/core/src/core/fetch.ts
ori-moisis: do you think it makes sense to just continue as if nothing happened in that case?
seems to me like that would create completely incorrect results, no?

packages/core/src/core/fetch.ts
```diff
@@ -170,6 +173,68 @@ const addFetchChangeMetadata = (
...
+  }
+  if (versions.every(_.isString)) {
+    const [current, base, incoming] = versions
+    log.debug('trying to merge strings - path: %s', change.change.id.getFullName())
+    const merged = diff3.mergeDiff3(current, base, incoming, options)
+    if (merged.conflict) {
+      log.debug('conflict found')
+      return [change]
+    }
+    log.debug('merged successfully')
```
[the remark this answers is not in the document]
ori-moisis: we can use that

what was the test on large files?
I assume if they were similar and had no conflicts it would be fast and the time would increase with the amount of differences - the interesting question is whether that growth is linear in file size or something else

packages/core/src/core/fetch.ts
ori-moisis: why not?
nothing is preventing you from putting it in an array inside the function

packages/core/src/core/fetch.ts
dantal4: I understand. I'll return in this case.

packages/core/test/core/fetch.test.ts
dantal4: 1. 👍 
2. there's a test for conflict on the same line - otherwise there won't be a conflict ("should not merge content on %s conflict").

packages/core/src/core/fetch.ts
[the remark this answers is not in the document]
dantal4: the results are very disappointing... I tested it on "very different" files, and for 512 KB files, it took 20 seconds to calculate the diff. I thought about adding a timeout but it's the function is synchronous so there's nothing to do.... 
Maybe we should have our logic to decide if it's worth trying to calculate the diff - maybe if no more than 1000 lines are different?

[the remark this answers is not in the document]
dantal4: Ok, I tried it again, and saw that the problem was because of the amount of **identical lines in each file** - and not the different lines -  there were ~2000 empty lines in each file, and it makes the algorithm look for things like crazy. when I removed the empty lines it took 20 ms for 10 MB files.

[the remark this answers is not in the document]
ori-moisis: that sounds pretty bad
means we don't have a good metric for when this is going to be bad and we have no way of stopping it if it starts being bad.

I don't think this is something we can just live with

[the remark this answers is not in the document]
dantal4: added a timeout option to node-diff3 functions (forked) - https://github.com/salto-io/node-diff3/pull/1

packages/core/src/core/fetch.ts
```diff
@@ -170,6 +180,172 @@ const addFetchChangeMetadata = (
...
+    log.debug('merging %s reached timeout', changeId)
+    return undefined
+  }
+  if (conflict) {
+    log.debug('conflict found in %s', changeId)
+    return undefined
+  }
+  log.debug('merged %s successfully', changeId)
+  return result.join(options.stringSeparator)
+}
```
ori-moisis, line 260: Moved the `log.time` to the function definition instead of the function call (more conventional in this repo)
```suggestion
const mergeStrings = (
  changeId: string,
  { current, base, incoming }: {
    current: string
    base: string | undefined
    incoming: string
  },
): string | undefined => log.time(
  () => {
    const contents = [current, base ?? '', incoming]
    log.debug(
      'trying to merge contents of %s - sizes: %s',
      changeId,
      contents.map(item => humanFileSize(item.length)).join(', ')
    )
    if (contents.some(item => item.length > MAX_MERGE_CONTENT_SIZE)) {
      log.warn(
        'skipping merge since contents size has reached the limit of %s',
        humanFileSize(MAX_MERGE_CONTENT_SIZE),
      )
      return undefined
    }
    const options = { excludeFalseConflicts: true, stringSeparator: '\n' }
    try {
      const { conflict, result } = base !== undefined
        ? diff3.mergeDiff3(current, base, incoming, options, MERGE_TIMEOUT)
        : merge2Strings(current, incoming, options, MERGE_TIMEOUT)
      if (conflict) {
        log.debug('conflict found in %s', changeId)
      } else {
        log.debug('merged %s successfully', changeId)
        return result.join(options.stringSeparator)
      }
    } catch (e) {
      if (e instanceof diff3.TimeoutError) {
        log.warn('merging %s reached timeout', changeId)
      } else {
        log.warn('merging %s failed with error %o', changeId, e)
      }
    }
    return undefined
  },
  'mergeStrings for %s', changeId
)
```

packages/core/src/core/fetch.ts
```diff
@@ -170,6 +180,172 @@ const addFetchChangeMetadata = (
...
+  return merged !== undefined
+    ? new StaticFile({ filepath, encoding, content: Buffer.from(merged, encoding) })
+    : undefined
+}
+
+const toMergedTextChange = (change: FetchChange, after: string | StaticFile): FetchChange => ({
+  ...change,
+  change: {
+    ...change.change,
+    ...toChange({
```
ori-moisis, line 319: why do we need the `toChange` logic here?

packages/lowerdash/src/types.ts
```diff
@@ -100,6 +100,11 @@ export const isNonEmptyArray = <T> (array: T[]): array is NonEmptyArray<T> => (
   array.length > 0
 )
 
+export const isTypeOfOrUndefined = <S>(
+  value: unknown,
+  typeGuard: TypeGuard<unknown, S>
+): value is S | undefined => value === undefined || typeGuard(value)
```
ori-moisis, line 106: shouldn't have to force the input type to be `unknown`
```suggestion
export const isTypeOfOrUndefined = <T, S extends T>(
  value: T,
  typeGuard: TypeGuard<T, S>
): value is S | undefined => value === undefined || typeGuard(value)
```

packages/core/test/core/fetch.test.ts
```diff
@@ -47,6 +49,10 @@ jest.mock('@salto-io/adapter-utils', () => ({
   ...jest.requireActual<{}>('@salto-io/adapter-utils'),
   applyInstancesDefaults: jest.fn().mockImplementation(e => mockAwu(e)),
 }))
+jest.mock('@salto-io/node-diff3', () => ({
```
ori-moisis, line 52: I think it would be better to test without mocking this library.
lets create examples with real conflicts, it would make more sense

ori-moisis (commented):
I think the amount of code here justifies putting the text merging logic in a separate file.

I think it will also make the tests more sensible, we'd only need one simple test in the fetch tests and we could take care of coverage and edge cases (e.g - timeout, binary files by filename vs by content, etc.) in a separate test suite

packages/core/src/core/fetch.ts
dantal4: ts fails on this syntax:
```javascript
{
  ...change.change,
  data: {
    ...change.change.data,
    after,
  },
},
```
with:
`Property 'before' is missing in type '{ after: string | StaticFile; }' but required in type '{ before: any; after: any; }'.`

packages/core/test/core/fetch.test.ts
dantal4: its only for the timeout test. for other test it works regularly.

packages/lowerdash/src/types.ts
dantal4: ts fails with an error:
```
A type predicate's type must be assignable to its parameter's type.
  Type 'S | undefined' is not assignable to type 'T'.
    'T' could be instantiated with an arbitrary type which could be unrelated to 'S | undefined'.ts(2677)
```

ori-moisis:
```typescript
export const isTypeOfOrUndefined = <T, S extends T>(
  value: T | undefined,
  typeGuard: TypeGuard<T, S>
): value is S | undefined => value === undefined || typeGuard(value)
```

ori-moisis (approved):

ori-moisis:
For extra safety, lets add a "kill switch" env var
e.g - `SALTO_FETCH_AUTO_MERGE` and if the value of that is `false` we don't call this new logic

PR #4873, opened 2023-09-13
Bump to version 0.3.47

kami-delaroz:
release 0.3.47

---

---
_Release Notes_: 
	# Salto Version v0.3.47
	
	## New features and bug fixes
	_Core_
	- Deploying now supports objects larger than 10MB
	- Display the total number of workSpace errors from diagnostics
	- Raise total number of ws errors that returns from getDiagnostics
	- Add logs to all error casess in summarizeDeployChanges logic
	- Add important attribute core annotation _important_values and _self_important_values
	- Fix issue where "fetch from" would omit fields that were not in the path index (fields that were recently deployed)
	- Fixed a bug where deploy error reportProgress wouldn't be reported for Errors with empty string
	- Fix bug where regenerate salto ids would cause deletion of static file
	
	_CLI_
	- Added command: salto element fix
	
	_Netsuite adapter_:
	- Add aliases by default
	- Create references in SuiteScripts for paths that don't start with a path prefix and to custom record fields
	- Fix crash in data_instances_attributes trying to access "attributes" on null

	_Salesforce adapter_:
	- Add a Change Validator that forbids Additions and Deletions of Standard Fields and Objects.
	- Solve LwcResources.LwcResource type discrepancy in LightningComponentBundle
	- Add the ability to create extra dependencies from the tooling API only for Elements from the current organization namespace.
	- Add internal ID to custom fields inside managed package if exists
	- Added references between new objects to field permissions
	- Add alias to Metadata Instances and CustomObject types by default.

	_Jira adapter_:
	- Added support for ScriptRunner's scripted fragments
	- Fixed a bug that prevented deployment of permission schemes for some customers
	- Issue layouts will not be fetched due to a performance issue for some cusomters
	- Fixed a bug that prevented deployment of first instance of a ScriptRunner type
	- Removing Issue Layouts
	- Fixed a bug with deployments of jqls
	- Fixed a bug that caused fetch to fail for custom fields with parentless cascading options
	- Fixed deleting an issue type and removing it from the issue type scheme

	_Zendesk adapter_:
	- Duplicate id values CV will point to a doc with an explanation
	- On begining of fetch, we will attempt to get the account subscription data and log it, only works on Zendesk production environment
	- Fix deployment error where removal of a field and relevant conditions from a form at the same time would return an error
	- Prevent the failure of deployment of ticket_form when ticket status ticket_field is added to the env
	- View restrictions field is not ordered on fetch, removing order changes from being shown on compare
	- 'field' fields that are supposed to have ticket_field can now have missing references
	- ticket_form will have now missing references for ticket_fields
	- Fix warning on ticket_form deployments if the ticket form has custom ticket status, and the feature it not enabled

	_Okta adapter_:
	- Improve fetch times
	- RoleAssignment instances will be excluded by default, this can be changed by editing the adapter config file
	- Fix a bug casing unwanted differences in GroupRule lists

---
_User Notifications_: 
None


tamtamirr (approved):
LGTM 👑

PR #4875, opened 2023-09-14
SALTO-4770: make guideTranslationFilter deploy run only when necessary

edenhassid:
make guideTranslationFilter deploy run only when necessary

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


netama (approved):
LGTM, let's also add an earlier check that guide is enabled in the config

edenhassid:
> LGTM, let's also add an earlier check that guide is enabled in the config

discussed IRL that we should add a separate filter for that

PR #4877, opened 2023-09-14
SALTO-4703 - Escape multi line quotes inside multi line

idoser96:
Escape the string ''' inside multi line when dumping a nacl, and unescape it when parsing the nacl

---

None

---
_Release Notes_: 
Core:
* Fixed issue where multi line quotes inside a multi line value were not properly escaped in NaCl

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/workspace/src/parser/internal/native/consumers/values.ts
+++ b/packages/workspace/src/parser/internal/native/consumers/values.ts
@@ -260,6 +260,8 @@ const consumeArrayItems = (
   return items
 }
 
+const unescapeMultilineMarker = (prim: string): string => prim.replace(/\\'''/g, "'''")
+
 const consumeMultilineString: Consumer<string | TemplateExpression> = context => {
   // Getting the position of the start marker
   const start = positionAtStart(context.lexer.next())
@@ -277,7 +279,7 @@ const consumeMultilineString: Consumer<string | TemplateExpression> = context =>
     context,
     tokens,
     true,
-    (_c, t) => unescapeTemplateMarker(t.text)
+    (_c, t) => unescapeMultilineMarker(unescapeTemplateMarker(t.text))
   )
   return {
     value,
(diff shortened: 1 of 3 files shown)


ori-moisis:
removing the user notification section since this will not actually make a different in existing workspaces

packages/workspace/src/parser/internal/native/consumers/values.ts
```diff
@@ -277,7 +279,7 @@ const consumeMultilineString: Consumer<string | TemplateExpression> = context =>
     context,
     tokens,
     true,
-    (_c, t) => unescapeTemplateMarker(t.text)
+    (_c, t) => unescapeMultilineMarker(unescapeTemplateMarker(t.text))
```
adfineberg, line 282: nit: Maybe wrap this with a `unescapeMultilineString` with both functions called?

adfineberg (commented):
Looks good, one small comment - do you think we should have a larger test to cover the original bug? Maybe add something that will get to `can be parsed back` in `dump.test`

packages/workspace/test/parser/dump.test.ts
```diff
@@ -416,8 +417,16 @@ describe('Salto Dump', () => {
...
+    it('should serialize strings and escape correctly', async () => {
+      expect(await dumpValues('"aaa"', functions)).toEqual('"\\"aaa\\""\n')
+    })
+    it('should serialize multi line string and escape correctly', async () => {
+      // eslint-disable-next-line no-template-curly-in-string
+      expect(await dumpValues("a${a}a\n'''aaa", functions)).toEqual("'''\na\\${a}a\n\\'''aaa\n'''\n")
+    })
+    it('should unescape multi line string correctly', async () => {
+      // eslint-disable-next-line no-template-curly-in-string
+      expect(unescapeMultilineString(await dumpValues("a${a}a\n'''aaa", functions))).toEqual("'''\na${a}a\n'''aaa\n'''\n")
```
ori-moisis, line 429: I think it would be better to keep this test on the `dumpValues` / `parse` level

we are missing a test that checks we parse back this format correctly

idoser96: just changed

adfineberg (approved):

packages/workspace/test/parser/dump.test.ts
ori-moisis: I might be missing something but where is the test that checks that parsing this back works as expected?
doesn't seem like any of the new tests try to parse anything

PR #4879, opened 2023-09-14
SALTO-4773 - Increase data depth in deploy instance trace log

idoser96:
Inspect data with bigger depth than the default, to log complicated instances correctly

---

None

---
_Release Notes_: 
Core
* trace log of instance deployment have bigger depth for the data

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/deployment/deployment.ts
+++ b/packages/adapter-components/src/deployment/deployment.ts
@@ -132,7 +132,7 @@ export const deployChange = async ({
   if (_.isEmpty(valuesToDeploy) && isAdditionOrModificationChange(change)) {
     return undefined
   }
-  log.trace(`deploying instance ${instance.elemID.getFullName()} with params ${inspectValue({ method: endpoint.method, url, queryParams, data: inspect(data, { depth: 6 }) }, { compact: true })}`)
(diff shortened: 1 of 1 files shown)


packages/adapter-components/src/deployment/deployment.ts
```diff
@@ -131,7 +132,7 @@ export const deployChange = async ({
   if (_.isEmpty(valuesToDeploy) && isAdditionOrModificationChange(change)) {
     return undefined
   }
-  log.trace(`deploying instance ${instance.elemID.getFullName()} with params ${inspectValue({ method: endpoint.method, url, data, queryParams }, { compact: true })}`)
+  log.trace(`deploying instance ${instance.elemID.getFullName()} with params ${inspectValue({ method: endpoint.method, url, queryParams, data: inspect(data, { depth: 6 }) }, { compact: true })}`)
```
netama, line 135: why not 
```suggestion
  log.trace(`deploying instance ${instance.elemID.getFullName()} with params ${inspectValue({ method: endpoint.method, url, queryParams, data, { compact: true, depth: 6 })}`)
```
?

netama (approved):
pre-approving to unblock

packages/adapter-components/src/deployment/deployment.ts
idoser96: yea, i got confused with _.defaults, but this is what i was looking for

PR #4881, opened 2023-09-14
SALTO-4761: Support fetch of Workflow on Custom Objects

tamtamirr:
Support fetch of Workflow (it's sub-instances) on Custom Objects

---

Workspace diff with the optional feature enabled: https://github.com/salto-io/tamir-sf/commit/9a8d247550eef24f62a1be9f762247438e2b8ad1

The fileName that returned for a Workflow on a Custom Object are incorrect. See this as an example:
Screen Shot 2023-09-14 at 11 04 56

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/src/adapter.ts
+++ b/packages/salesforce-adapter/src/adapter.ts
@@ -599,7 +599,7 @@ export default class SalesforceAdapter implements AdapterOperations {
         types: metadataTypesToRetrieve,
         metadataQuery: this.fetchProfile.metadataQuery,
         maxItemsInRetrieveRequest: this.maxItemsInRetrieveRequest,
-        addNamespacePrefixToFullName: this.fetchProfile.addNamespacePrefixToFullName,
+        fetchProfile: this.fetchProfile,
         typesToSkip: new Set(this.metadataTypesOfInstancesFetchedInFilters),
       }),
       readInstances(metadataTypesToRead),
--- a/packages/salesforce-adapter/src/transformers/xml_transformer.ts
+++ b/packages/salesforce-adapter/src/transformers/xml_transformer.ts
@@ -21,7 +21,7 @@ import JSZip from 'jszip'
 import { collections, values as lowerDashValues } from '@salto-io/lowerdash'
 import { Values, StaticFile, InstanceElement } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
-import { MapKeyFunc, mapKeysRecursive, TransformFunc, transformValues } from '@salto-io/adapter-utils'
+import { MapKeyFunc, mapKeysRecursive, safeJsonStringify, TransformFunc, transformValues } from '@salto-io/adapter-utils'
 import { API_VERSION } from '../client/client'
 import {
   INSTANCE_FULL_NAME_FIELD, IS_ATTRIBUTE, METADATA_CONTENT_FIELD, SALESFORCE, XML_ATTRIBUTE_PREFIX,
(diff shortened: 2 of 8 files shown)


packages/salesforce-adapter/src/transformers/xml_transformer.ts
```diff
@@ -268,12 +291,18 @@ export const fromRetrieveResult = async (
...
-      typesWithMetaFile.has(file.type) || isComplexType(file.type), isComplexType(file.type),
-      file.namespacePrefix)
+    const fileNameToValuesBuffer = await extractFileNameToData({
+      zip,
+      fileName: file.fileName,
+      withMetadataSuffix: typesWithMetaFile.has(file.type) || isComplexType(file.type),
+      complexType: isComplexType(file.type),
+      namespacePrefix: file.namespacePrefix,
+      fixRetrieveFilePaths,
+    })
```
tamtamirr, line 305: changed to name parameters

packages/salesforce-adapter/src/transformers/xml_transformer.ts
```diff
@@ -288,8 +317,14 @@ export const fromRetrieveResult = async (
...
-      const fileNameToContent = await extractFileNameToData(zip, file.fileName, false,
-        isComplexType(file.type), file.namespacePrefix)
+      const fileNameToContent = await extractFileNameToData({
+        zip,
+        fileName: file.fileName,
+        withMetadataSuffix: false,
+        complexType: isComplexType(file.type),
+        namespacePrefix: file.namespacePrefix,
+        fixRetrieveFilePaths,
+      })
```
tamtamirr, line 327: change to name parameters

packages/salesforce-adapter/src/transformers/xml_transformer.ts
```diff
@@ -312,6 +347,7 @@ export const fromRetrieveResult = async (
   }
 
   const zip = await new JSZip().loadAsync(Buffer.from(result.zipFile, 'base64'))
+  log.debug(`retrieved zip contains the following files: ${safeJsonStringify(Object.keys(zip.files))}`)
```
tamtamirr, line 350: Added this log for better visibility when we can't extract a file from the retrieve zip

tamtamirr (commented):

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -599,7 +599,7 @@ export default class SalesforceAdapter implements AdapterOperations {
         types: metadataTypesToRetrieve,
         metadataQuery: this.fetchProfile.metadataQuery,
         maxItemsInRetrieveRequest: this.maxItemsInRetrieveRequest,
-        addNamespacePrefixToFullName: this.fetchProfile.addNamespacePrefixToFullName,
+        fetchProfile: this.fetchProfile,
```
tamtamirr, line 602: Since we need two things from the **fetchProfile**, it's better to pass it as a parameter instead.

packages/salesforce-adapter/src/transformers/xml_transformer.ts
```diff
@@ -248,11 +248,34 @@ export const xmlToValues = (xmlAsString: string): { values: Values; typeName: st
...
+  withMetadataSuffix: boolean
+  complexType: boolean
+  namespacePrefix?: string
+  fixRetrieveFilePaths: boolean
+}
+
+
+const fixPath = (path: string): string => (
+  // The fileName of Workflow instances on Custom Objects return as "Workflow/WorkflowName.workflow"
+  // while the actual file in the zip is "workflows/WorkflowName.workflow"
```
tomsellek, line 263: :barf:

packages/salesforce-adapter/src/transformers/xml_transformer.ts
tomsellek: This could get pretty spammy, no? How many Zip files do we have in a typical fetch?

tomsellek (approved):

tomsellek (approved):
LGTM!

packages/salesforce-adapter/src/transformers/xml_transformer.ts
tamtamirr: It depends on the amount of files (instances in retrieve) so it's usually 1 and at most just a few.

benisraeladi (approved):

PR #4882, opened 2023-09-14
SALTO-2349: Added additional request context

idoamit8:
Following this [PR](https://github.com/salto-io/salto/pull/4857) I splited adapter components from jsm initial fetch.

---

None

---
_Release Notes_: 
None
---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/elements/ducktype/transformer.ts
+++ b/packages/adapter-components/src/elements/ducktype/transformer.ts
@@ -143,7 +143,8 @@ const getEntriesForType = async (
   const requestWithDefaults = getConfigWithDefault(request, typeDefaultConfig.request ?? {})
 
   const getEntries = async (context: Values | undefined): Promise<Values[]> => {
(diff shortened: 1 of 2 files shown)


packages/adapter-components/src/elements/ducktype/transformer.ts
```diff
@@ -284,6 +285,7 @@ export const getTypeAndInstances = async ({
   getEntriesResponseValuesFunc,
   reversedSupportedTypes,
   customInstanceFilter,
+  requestContext,
```
netama, line 288: should also rename to `additional...`, no?

netama (commented):
looks good, added one small comment - but we should also add unittests for this before merging

packages/adapter-components/src/elements/ducktype/transformer.ts
idoamit8: yes :)

netama (approved):

PR #4883, opened 2023-09-14
SALTO-4775: change timeout for workflow scheme

IdoZakk:
Switched to new configuration (which also means double the time)

---

None

---
_Release Notes_: 
Jira Adapter:
* Timeout for workflow scheme publish was doubled and added to the configuration

---
_User Notifications_: 
None


alonstern (approved):
👍

PR #4886, opened 2023-09-18
SALTO-4747: add handleIdenticalAttachmentConflicts flag

edenhassid:
add handleIdenticalAttachmentConflicts flag

---

---
_Release Notes_: 
zendesk:
* add `handleIdenticalAttachmentConflicts` flag. when it is true it will look at all attachments with the same elemId and hash and will keep only one.

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/filters/handle_identical_attachment_conflicts.ts
+++ b/packages/zendesk-adapter/src/filters/handle_identical_attachment_conflicts.ts
@@ -23,6 +23,10 @@ import { ARTICLE_ATTACHMENT_TYPE_NAME } from '../constants'
 
 const log = logger(module)
 
+/**
+ * when handleIdenticalAttachmentConflicts flag is true, this filter will omit article attachments with the same elemId
+ * and hash, and keep only one attachment of this type in the elements.
+ */
 const filterCreator: FilterCreator = ({ config }) => ({
   name: 'handleIdenticalAttachmentConflictsFilter',
   onFetch: async elements => {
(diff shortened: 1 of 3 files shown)


packages/zendesk-adapter/src/filters/handle_identical_attachment_conflicts.ts
```diff
@@ -0,0 +1,59 @@
...
+import _ from 'lodash'
+import { isInstanceElement, isStaticFile } from '@salto-io/adapter-api'
+import { FilterCreator } from '../filter'
+import { FETCH_CONFIG } from '../config'
+import { ARTICLE_ATTACHMENT_TYPE_NAME } from '../constants'
+
+
+const log = logger(module)
+
+const filterCreator: FilterCreator = ({ config }) => ({
```
netama, line 26: please add a docstring :)

packages/zendesk-adapter/src/filters/handle_identical_attachment_conflicts.ts
```diff
@@ -0,0 +1,59 @@
...
+        && isStaticFile(attachment.value.content)
+        && attachment.value.content.hash !== undefined)
+    const attachmentsByElemId = _.groupBy(articleAttachmentInstances, elem => elem.elemID.getFullName())
+    const idsToRemove = new Set(Object.values(attachmentsByElemId)
+      .filter(attachments => attachments.length > 1)
+      .flatMap(attachments => {
+        const hashSet = new Set(attachments.map(attachment => attachment.value.content.hash))
+        // check if there are different files with the same name
+        if (hashSet.size !== 1) {
+          return []
```
netama, line 47: we should probably at least log here that we're not doing anything, but maybe even group by hash and still add the handling? (could log how many we have per hash or something like that)

packages/zendesk-adapter/src/filters/handle_identical_attachment_conflicts.ts
```diff
@@ -0,0 +1,59 @@
...
+        }
+        // all files have the same hash, so we can remove the duplicates and keep only one
+        return attachments.slice(1).map(att => att.value.id)
+      }))
+    _.remove(elements, elem =>
```
netama, line 52: let's log a summary of which elem ids are impacted by this? (whatever would be helpful for debugging)

netama (approved):
cool :) pre-approving but added some suggested logic changes, I can do another pass afterwards if you want

PR #4887, opened 2023-09-18
SALTO-4784: handle undefined apiNames in transformer.isCustom in SF

tamtamirr:
handle undefined apiNames in transformer.isCustom in Salesforce Adapter.

---

Upon modification of picklist standard fields, when the **apiName** is undefined the preview flow yielded runtime error.
See the Jira ticket for more context.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/src/transformers/transformer.ts
+++ b/packages/salesforce-adapter/src/transformers/transformer.ts
@@ -124,8 +124,8 @@ export const isFieldOfCustomObject = async (field: Field): Promise<boolean> =>
 export const isInstanceOfCustomObject = async (element: Readonly<Element>): Promise<boolean> =>
   isInstanceElement(element) && isCustomObject(await element.getType())
 
-export const isCustom = (fullName: string): boolean =>
-  fullName.endsWith(SALESFORCE_CUSTOM_SUFFIX)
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/src/transformers/transformer.ts
```diff
@@ -124,8 +124,8 @@ export const isFieldOfCustomObject = async (field: Field): Promise<boolean> =>
 export const isInstanceOfCustomObject = async (element: Readonly<Element>): Promise<boolean> =>
   isInstanceElement(element) && isCustomObject(await element.getType())
 
-export const isCustom = (fullName: string): boolean =>
-  fullName.endsWith(SALESFORCE_CUSTOM_SUFFIX)
+export const isCustom = (fullName: string | undefined): boolean =>
```
benisraeladi, line 127: Did you take a look at other uses of `isCustom`? Can this change affect any of them?

tamtamirr: This change didn't break existing code, as it only made it safer. In case this method was invoked with undefined (For example calling the deprecated **apiName** with the incorrect return type of **Promise<string>**)

PR #4889, opened 2023-09-19
SALTO-4794:Empty Validators

IdoZakk:
The Empty workflow validators change validators contained some validators that should be empty

---
_Release Notes_: 
Jira Adapter:
* Fixed a bug that prevented the deployment of some workflow validators

---
_User Notifications_: 
None


edenhassid (approved):
🔥

PR #4890, opened 2023-09-19
SALTO-4476 extract string from FieldDef for detailedMessage

benisraeladi:
Extract string so that error message does not show `[Object object]` anymore

---

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


tamtamirr (approved):
Thanks!

PR #4891, opened 2023-09-19
SALTO-4747: handle identical attachments conflicts fixes

edenhassid:
handle identical attachments conflicts fixes
fixes of https://github.com/salto-io/salto/pull/4886

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/config_doc.md
+++ b/packages/zendesk-adapter/config_doc.md
@@ -77,7 +77,7 @@ zendesk {
 | [guide]                            | undefined (Guide will be disabled)| Configuration for defining which brands will be included in Zendesk Guide fetch
 | [resolveOrganizationIDs]           | false                             | When enabled, organization IDs will be replaced with organization names
 | includeAuditDetails                | false                             | When enabled, changed_at and changed_by information will be added to instances
-| handleIdenticalAttachmentConflicts | false                             | When enabled, one attachment will be kept from each set of duplicate article attachments
+| handleIdenticalAttachmentConflicts | false                             | When enabled, one attachment will be kept from each set of identical attachments (having the same hash) associated with the same article
 
 
 ## Fetch entry options
--- a/packages/zendesk-adapter/config_doc.md
+++ b/packages/zendesk-adapter/config_doc.md
@@ -30,6 +30,7 @@ zendesk {
       brands = [".*"]
     }
     includeAuditDetails = true
+    handleIdenticalAttachmentConflicts = true
   }
   deploy = {
     defaultMissingUserFallback = "##DEPLOYER##"
@@ -69,13 +70,14 @@ zendesk {
 
 ## Fetch configuration options
 
-| Name                            | Default when undefined            | Description
-|---------------------------------|-----------------------------------|------------
-| [include](#fetch-entry-options) | [{ type = ".*" }]                 | List of entries to determine what instances to include in the fetch
-| [exclude](#fetch-entry-options) | []                                | List of entries to determine what instances to exclude in the fetch
-| [guide]                         | undefined (Guide will be disabled)| Configuration for defining which brands will be included in Zendesk Guide fetch
-| [resolveOrganizationIDs]        | false                             | When enabled, organization IDs will be replaced with organization names
-| includeAuditDetails             | false                             | When enabled, changed_at and changed_by information will be added to instances
+| Name                               | Default when undefined            | Description
+|------------------------------------|-----------------------------------|------------
+| [include](#fetch-entry-options)    | [{ type = ".*" }]                 | List of entries to determine what instances to include in the fetch
+| [exclude](#fetch-entry-options)    | []                                | List of entries to determine what instances to exclude in the fetch
+| [guide]                            | undefined (Guide will be disabled)| Configuration for defining which brands will be included in Zendesk Guide fetch
+| [resolveOrganizationIDs]           | false                             | When enabled, organization IDs will be replaced with organization names
+| includeAuditDetails                | false                             | When enabled, changed_at and changed_by information will be added to instances
+| handleIdenticalAttachmentConflicts | false                             | When enabled, one attachment will be kept from each set of duplicate article attachments
 
 
 ## Fetch entry options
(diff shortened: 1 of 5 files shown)


packages/zendesk-adapter/config_doc.md
```diff
@@ -69,13 +70,14 @@ zendesk {
...
+| Name                               | Default when undefined            | Description
+|------------------------------------|-----------------------------------|------------
+| [include](#fetch-entry-options)    | [{ type = ".*" }]                 | List of entries to determine what instances to include in the fetch
+| [exclude](#fetch-entry-options)    | []                                | List of entries to determine what instances to exclude in the fetch
+| [guide]                            | undefined (Guide will be disabled)| Configuration for defining which brands will be included in Zendesk Guide fetch
+| [resolveOrganizationIDs]           | false                             | When enabled, organization IDs will be replaced with organization names
+| includeAuditDetails                | false                             | When enabled, changed_at and changed_by information will be added to instances
+| handleIdenticalAttachmentConflicts | false                             | When enabled, one attachment will be kept from each set of duplicate article attachments
```
netama, line 80: nit: maybe `... each set of identical attachments (having the same hash) associated with the same article`?

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -266,24 +266,23 @@ const getInline = (attachment: InstanceElement | undefined): boolean => attachme
 const isRemovedAttachment = (
   attachment: unknown,
   articleRemovedAttachmentsIds: Set<number>,
-  attachmentByName: Record<string, InstanceElement>
 ): boolean => {
-  const name = isReferenceExpression(attachment) ? getName(attachment) : undefined
-  const attachmentInstance = name ? attachmentByName[name] : undefined
-  return isInstanceElement(attachmentInstance) && articleRemovedAttachmentsIds.has(attachmentInstance.value.id)
+  const id = isReferenceExpression(attachment) ? getId(attachment.value) : undefined
```
netama, line 270: this seems to be a problem wit hthe existing code as well, but can we confirm that the referenced value is an instance element? (this seems to be a problem in a few places, so might be best to add this check inside `getId`)

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -306,28 +305,34 @@ const getAttachmentData = (
  * we do not want to include these orphaned attachments in the fetch results, so we omit inline attachments that are not
  * referenced from the article body.
  */
-const calculateAndRemoveDeletedAttachments = (
-  articleInstances: InstanceElement[],
-  attachmentByName: Record<string, InstanceElement>,
-  translationByName: Record<string, InstanceElement>
-): Set<string> => {
+const calculateAndRemoveDeletedAttachments = ({
```
netama, line 308: 🎉

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -358,28 +363,34 @@ const filterCreator: FilterCreator = ({ config, client, elementsSource, brandIdT
           .filter(attachment => getName(attachment) !== undefined),
         getName,
       )
+      const attachmentById: Record<string, InstanceElement> = _.keyBy(
```
netama, line 366: nit: doesn't matter much if the code works, but in some places we seem to have `Record<number,` and in others `Record<string,` when indexing by id, may be worth aligning?

packages/zendesk-adapter/src/filters/handle_identical_attachment_conflicts.ts
```diff
@@ -15,14 +15,39 @@
...
 import { FilterCreator } from '../filter'
 import { FETCH_CONFIG } from '../config'
-import { ARTICLE_ATTACHMENT_TYPE_NAME } from '../constants'
+import { ARTICLE_ATTACHMENT_TYPE_NAME, ARTICLE_TYPE_NAME } from '../constants'
 
 
 const log = logger(module)
 
+
+const uniqueAttachmentField = (
```
netama, line 33: 1. let's add a docstring explaining what this does and why it's needed?
2. naming nit: maybe `makeAttachmentFieldUnique` to clarify this is modifying in-place?

packages/zendesk-adapter/src/filters/handle_identical_attachment_conflicts.ts
```diff
@@ -53,14 +78,20 @@ const filterCreator: FilterCreator = ({ config }) => ({
           return []
         }
         // all files have the same hash, so we can remove the duplicates and keep only one
+        duplicateElemIds.add(attachments[0].elemID.getFullName())
         return attachments.slice(1).map(att => att.value.id)
       }))
-    log.debug(`going to remove duplicates for attachments: ${duplicateElemIds.join(',')}`)
-    log.debug(`ids removed: ${Array.from(idsToRemove).join(',')}`)
+    log.debug(`going to remove duplicates for attachments: ${Array.from(duplicateElemIds).join(',')}`)
```
netama, line 84: nit: for debug-level logs (and in general below-error), there's a preference for using arguments and not string templates so that we don't calculate the combined string if it won't be logged - in this case something like
```suggestion
    log.debug('going to remove duplicates for attachments: %s', Array.from(duplicateElemIds).join(','))
```

packages/zendesk-adapter/src/filters/handle_identical_attachment_conflicts.ts
```diff
@@ -53,14 +78,20 @@ const filterCreator: FilterCreator = ({ config }) => ({
...
+
+    // remove duplicates from attachments field in article
+    uniqueAttachmentField(elements, idsToRemove)
+
     _.remove(elements, elem =>
       isInstanceElement(elem)
       && elem.elemID.typeName === ARTICLE_ATTACHMENT_TYPE_NAME
       && idsToRemove.has(elem.value.id))
+
+    log.debug(`ids removed: ${Array.from(idsToRemove).join(',')}`)
```
netama, line 94: same nit (though we're not that consistent with it in the logs, Ori thinks it can make a difference)

netama (approved):
looks great 🎈 added some nits but pre-approving.
I think it is worth including release notes for this - saying what new flag was added and what it does

packages/zendesk-adapter/src/filters/article/article.ts
edenhassid: most of the times I use `getId` its on an Instance element, I added test where needed

packages/zendesk-adapter/src/filters/article/article.ts
edenhassid: it should be number there, I fixed it :)

PR #4892, opened 2023-09-19
SALTO-1872: support analytics (NETSUITE)

SeggevHaimovich:
Support analytics (workbooks and datasets).

Added new types - workbook, dataset.

Filters:
Created filter to match the objects to the new type on fetch and back to the old one before deploy.
Created filter to add workbooks that must be added to deploy the wanted datasets.

Change Validators:
created change validator to make sure the datasets have a workbook referencing them.
created a change validator to inform the user that the deployment of analytics may have a silent failure.
created a change validator to check if the workbooks are supported (any workbook that contains pivots/charts/datasetLinks is unsupported)

---

Jira ticket: https://salto-io.atlassian.net/browse/SALTO-1872

---
_Release Notes_: 
Users can fetch and deploy analytics with a few limitations:
1. Worbooks that contains pivots/charts/datasetLinks cannot be deployed
2. In some cases, deployments may fail without any warning (silent failure): SDF approves the deployment and indicates success, yet nothing actually happens. We added a warning to check the outcomes in the UI

---
_User Notifications_: 
_Users that fetched analytics before have a definition field (a long string). After that change when they fetch again, this field will disappear and they will have all the information in a convenient form._

### DIFF
--- a/packages/netsuite-adapter/src/adapter.ts
+++ b/packages/netsuite-adapter/src/adapter.ts
@@ -103,11 +103,11 @@ export const allFilters: (LocalFilterCreatorDefinition | RemoteFilterCreatorDefi
   { creator: dataInstancesDiff },
   // addParentFolder must run before replaceInstanceReferencesFilter
   { creator: addParentFolder },
-  { creator: parseReportTypes },
-  // analyticsDefinitionHandle must run before translationConverter and replaceElementReferences
-  // and after parseReportTypes
-  { creator: analyticsDefinitionHandle },
   { creator: convertLists },
+  { creator: parseReportTypes },
+  // analyticsDefinitionHandle must run after convertLists
+  // and before translationConverter and replaceElementReferences
+  { creator: analyticsDefinitionHandle },
   { creator: consistentValues },
   // excludeInstances should run after parseReportTypes, analyticsDefinitionHandle & consistentValues,
   // so users will be able to exclude elements based on parsed values.
--- a/packages/netsuite-adapter/src/adapter.ts
+++ b/packages/netsuite-adapter/src/adapter.ts
@@ -44,6 +44,8 @@ import dataInstancesAttributes from './filters/data_instances_attributes'
 import dataInstancesNullFields from './filters/data_instances_null_fields'
 import dataInstancesDiff from './filters/data_instances_diff'
 import dataInstancesIdentifiers from './filters/data_instances_identifiers'
+import addReferencingWorkbooks from './filters/add_referencing_workbooks'
+import analyticsDefinitionHandle from './filters/analytics_definition_handle'
 import suiteAppInternalIds from './filters/internal_ids/suite_app_internal_ids'
 import SDFInternalIds from './filters/internal_ids/sdf_internal_ids'
 import accountSpecificValues from './filters/account_specific_values'
@@ -102,9 +104,12 @@ export const allFilters: (LocalFilterCreatorDefinition | RemoteFilterCreatorDefi
   // addParentFolder must run before replaceInstanceReferencesFilter
   { creator: addParentFolder },
   { creator: parseReportTypes },
+  // analyticsDefinitionHandle must run before translationConverter and replaceElementReferences
+  // and after parseReportTypes
+  { creator: analyticsDefinitionHandle },
   { creator: convertLists },
   { creator: consistentValues },
-  // excludeInstances should run after parseReportTypes & consistentValues,
+  // excludeInstances should run after parseReportTypes, analyticsDefinitionHandle & consistentValues,
   // so users will be able to exclude elements based on parsed values.
   { creator: excludeInstances },
   // convertListsToMaps must run after convertLists and consistentValues
@@ -142,8 +147,11 @@ export const allFilters: (LocalFilterCreatorDefinition | RemoteFilterCreatorDefi
   { creator: addBundleReferences },
   // omitFieldsFilter should be the last onFetch filter to run
   { creator: omitFieldsFilter },
-  // additionalChanges should be the first preDeploy filter to run
+  // additionalChanges should be right after addReferencingWorkbooks
+  // (adds required referenced elements to the deployent)
   { creator: additionalChanges },
+  // addReferencingWorkbooks should be the first preDeploy filter to run (adds workbooks to the deployment)
+  { creator: addReferencingWorkbooks },
 ]
 
 // By default we run all filters and provide a client
--- a/packages/netsuite-adapter/src/autogen/types/standard_types/workbook.ts
+++ b/packages/netsuite-adapter/src/autogen/types/standard_types/workbook.ts
@@ -72,7 +72,7 @@ export const workbookType = (): TypeAndInnerTypes => {
     },
     fields: {
       dependency: {
-        refType: createRefToElmWithValue(BuiltinTypes.STRING /* Original type was single-select list */),
+        refType: createRefToElmWithValue(new ListType(BuiltinTypes.STRING)),
         annotations: {
           [CORE_ANNOTATIONS.REQUIRED]: true,
         },
@@ -185,7 +185,7 @@ export const workbookType = (): TypeAndInnerTypes => {
         annotations: {
           [CORE_ANNOTATIONS.REQUIRED]: true,
         },
-      },
+      }, /* Original description:   */
       charts: {
         refType: createRefToElmWithValue(workbook_charts),
         annotations: {
--- a/packages/netsuite-adapter/src/change_validators/check_referenced_datasets.ts
+++ b/packages/netsuite-adapter/src/change_validators/check_referenced_datasets.ts
@@ -1,0 +1,114 @@
+/*
+*                      Copyright 2024 Salto Labs Ltd.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+import { ChangeError, InstanceElement, ReadOnlyElementsSource, ReferenceExpression, getChangeData, isAdditionOrModificationChange, isInstanceChange, isReferenceExpression } from '@salto-io/adapter-api'
+import { collections } from '@salto-io/lowerdash'
+import { NetsuiteChangeValidator } from './types'
+import { DATASET, SCRIPT_ID, WORKBOOK } from '../constants'
+
+const { awu } = collections.asynciterable
+
+type UnreferencedDataset = {
+  forInfos: InstanceElement[]
+  forErrors: InstanceElement[]
+}
+
+export const getInstanceReferenceDependencies = (instance: InstanceElement): ReferenceExpression[] =>
+  collections.array.makeArray(instance.value.dependencies?.dependency)
+    .filter(dep => isReferenceExpression(dep))
+
+export const getAllWorkbooksFromElementsSource = async (
+  elementsSource: ReadOnlyElementsSource,
+): Promise<InstanceElement[]> =>
+  awu(await elementsSource.list())
+    .filter(elemID => elemID.typeName === WORKBOOK && elemID.idType === 'instance')
+    .map(async elemId => elementsSource.get(elemId))
+    .toArray()
+
+const getUnreferencedDatasetsDeploymentAndEnvironment = async (
+  changedDatasets: InstanceElement[],
+  changedWorkbooks: InstanceElement[],
+  elementsSource: ReadOnlyElementsSource,
+): Promise<UnreferencedDataset> => {
+  const unreferencedDatasetsWithinDeployment = new Map<string, InstanceElement>()
+  const unreferencedDatasetsWithinEnvironment = new Map<string, InstanceElement>(
+    changedDatasets
+      .map(dataset => [dataset.elemID.createNestedID(SCRIPT_ID).getFullName(), dataset])
+  )
+  changedWorkbooks
+    .flatMap(workbook => getInstanceReferenceDependencies(workbook))
+    .map(dep => dep.elemID.getFullName())
+    .forEach(fullName => unreferencedDatasetsWithinEnvironment.delete(fullName))
+  if (unreferencedDatasetsWithinEnvironment.size > 0) {
+    const allWorkbooks = await getAllWorkbooksFromElementsSource(elementsSource)
+    allWorkbooks
+      .flatMap(workbook => getInstanceReferenceDependencies(workbook))
+      .map(dep => dep.elemID.getFullName())
+      .forEach(fullName => {
+        const dataset = unreferencedDatasetsWithinEnvironment.get(fullName)
+        if (dataset !== undefined) {
+          unreferencedDatasetsWithinDeployment.set(fullName, dataset)
+          unreferencedDatasetsWithinEnvironment.delete(fullName)
+        }
+      })
+  }
+
+  return {
+    forInfos: [...unreferencedDatasetsWithinDeployment.values()],
+    forErrors: [...unreferencedDatasetsWithinEnvironment.values()],
+  }
+}
+
+const changeValidator: NetsuiteChangeValidator = async (changes, _deployReferencedElements, elementsSource) => {
+  const instancesFromAdditionOrModificationChanges = changes
+    .filter(isAdditionOrModificationChange)
+    .filter(isInstanceChange)
+    .map(getChangeData)
+
+  const changedDatasets = instancesFromAdditionOrModificationChanges
+    .filter(inst => inst.elemID.typeName === DATASET)
+
+  if (elementsSource === undefined || changedDatasets.length === 0) {
+    return []
+  }
+
+  const changedWorkbooks = instancesFromAdditionOrModificationChanges
+    .filter(inst => inst.elemID.typeName === WORKBOOK)
+
+  const {
+    forInfos: unreferencedInDeployment,
+    forErrors: unreferencedInEnvironment,
+  } = await getUnreferencedDatasetsDeploymentAndEnvironment(changedDatasets, changedWorkbooks, elementsSource)
+
+  const errors = unreferencedInDeployment
+    .map(({ elemID }): ChangeError => ({
+      elemID,
+      severity: 'Info',
+      message: 'To deploy this, an existing workbook will be re-deployed',
+      detailedMessage: 'A dataset must be deployed alongside a workbook which was connected to it. '
+        + 'Salto will automatically add to the SDF deployment an existing workbook (and the translation collection instances related to that workbook) which was already connected to this dataset. This should not affect the target environment.',
+    })).concat(unreferencedInEnvironment
+      .map(({ elemID }): ChangeError => ({
+        elemID,
+        severity: 'Error',
+        message: 'This dataset cannot be deployed without any connected workbooks',
+        detailedMessage: 'A dataset must be deployed alongside a workbook which was connected to it. '
+          + 'To deploy this dataset, add at least one workbook to your deployment which has this dataset connected.',
+      })))
+  return errors
+}
+
+export default changeValidator
--- a/packages/netsuite-adapter/src/filters/add_referencing_workbooks.ts
+++ b/packages/netsuite-adapter/src/filters/add_referencing_workbooks.ts
@@ -1,0 +1,153 @@
+/*
+*                      Copyright 2024 Salto Labs Ltd.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+import { createRefToElmWithValue, Element, getChangeData, InstanceElement, isAdditionOrModificationChange, isElement, isInstanceChange, isInstanceElement, isReferenceExpression, ReadOnlyElementsSource, toChange } from '@salto-io/adapter-api'
+import { transformElement, TransformFunc } from '@salto-io/adapter-utils'
+import { DATASET, SCRIPT_ID, TRANSLATION_COLLECTION, WORKBOOK } from '../constants'
+import { LocalFilterCreator } from '../filter'
+import { getAllWorkbooksFromElementsSource, getInstanceReferenceDependencies } from '../change_validators/check_referenced_datasets'
+import { parsedWorkbookType } from '../type_parsers/analytics_parsers/parsed_workbook'
+
+const resolveTypes = async (
+  elem: Element,
+  elementsSource: ReadOnlyElementsSource,
+): Promise<void> => {
+  if (isInstanceElement(elem)) {
+    const type = await elem.getType(elementsSource)
+    elem.refType = createRefToElmWithValue(type)
+  }
+
+  const transFunc:TransformFunc = async ({ value, field }) => {
+    if (field === undefined) {
+      return value
+    }
+    const fieldType = await field.getType(elementsSource)
+    field.refType = createRefToElmWithValue(fieldType)
+    return value
+  }
+
+  await transformElement({
+    element: elem,
+    transformFunc: transFunc,
+    strict: false,
+    elementsSource,
+  })
+}
+
+const addInnerReferencesTopLevelParent = async (
+  elem: InstanceElement,
+  elementsSource: ReadOnlyElementsSource,
+): Promise<void> => {
+  const transFunc: TransformFunc = async ({ value }) => {
+    if (isReferenceExpression(value) && value.topLevelParent === undefined) {
+      const topLevelParent = (await elementsSource.get(value.elemID.createTopLevelParentID().parent))
+      if (isElement(topLevelParent)) {
+        await resolveTypes(topLevelParent, elementsSource)
+        value.topLevelParent = topLevelParent.clone()
+      }
+      if (value.elemID.typeName === TRANSLATION_COLLECTION
+        && isInstanceElement(value.topLevelParent)
+        && isReferenceExpression(value.topLevelParent.value.name)) {
+        if (value.elemID.isEqual(value.topLevelParent.value.name.elemID)) {
+          value.topLevelParent.value.name.topLevelParent = value.topLevelParent
+        } else {
+          const refTopLevelParent = await elementsSource.get(
+            value.topLevelParent.value.name.elemID.createTopLevelParentID().parent
+          )
+          if (isElement(refTopLevelParent)) {
+            await resolveTypes(refTopLevelParent, elementsSource)
+            value.topLevelParent.value.name.topLevelParent = refTopLevelParent.clone()
+          }
+        }
+      }
+    }
+    return value
+  }
+
+  await transformElement({
+    element: elem,
+    transformFunc: transFunc,
+    strict: false,
+  })
+}
+
+const getUnreferencedDatasets = (
+  datasets: InstanceElement[],
+  workbooks: InstanceElement[],
+): Set<string> => {
+  const datasetElemIdFullNameSet = new Set<string>(
+    datasets
+      .map(dataset => dataset.elemID.createNestedID(SCRIPT_ID).getFullName())
+  )
+
+  workbooks
+    .flatMap(workbook => getInstanceReferenceDependencies(workbook))
+    .map(dep => dep.elemID.getFullName())
+    .forEach(fullName => datasetElemIdFullNameSet.delete(fullName))
+  return datasetElemIdFullNameSet
+}
+
+const getWorkbooksToPush = async (
+  datasets: InstanceElement[],
+  changedWorkbooks: InstanceElement[],
+  elementsSource: ReadOnlyElementsSource
+): Promise<InstanceElement[]> => {
+  const datasetFullNameSet = getUnreferencedDatasets(datasets, changedWorkbooks)
+
+  if (datasetFullNameSet.size === 0) {
+    return []
+  }
+
+  const allWorkbooks = await getAllWorkbooksFromElementsSource(elementsSource)
+
+  return allWorkbooks.reduce(async (acc, workbook) => {
+    if (datasetFullNameSet.size === 0) {
+      return acc
+    }
+    const dependencies = getInstanceReferenceDependencies(workbook)
+    const datasetDependencies = dependencies
+      .map(dep => dep.elemID.getFullName())
+      .filter(dep => datasetFullNameSet.has(dep))
+    if (datasetDependencies.length > 0) {
+      datasetDependencies.forEach(dep => datasetFullNameSet.delete(dep))
+      const clonedWorkbook = workbook.clone()
+      clonedWorkbook.refType = createRefToElmWithValue(parsedWorkbookType().type)
+      await addInnerReferencesTopLevelParent(clonedWorkbook, elementsSource)
+      const workbooks = await acc
+      return workbooks.concat(clonedWorkbook)
+    }
+    return acc
+  }, Promise.resolve([] as InstanceElement[]))
+}
+
+const filterCreator: LocalFilterCreator = ({ elementsSource }) => ({
+  name: 'pushReferencingWorkbooks',
+  preDeploy: async changes => {
+    const instancesFromAdditionOrModificationChanges = changes
+      .filter(isAdditionOrModificationChange)
+      .filter(isInstanceChange)
+      .map(getChangeData)
+
+    const workbooksToPush = await getWorkbooksToPush(
+      instancesFromAdditionOrModificationChanges.filter(instance => instance.elemID.typeName === DATASET),
+      instancesFromAdditionOrModificationChanges.filter(instance => instance.elemID.typeName === WORKBOOK),
+      elementsSource,
+    )
+    changes.push(...workbooksToPush.map(workbook => toChange({ before: workbook, after: workbook })))
+  },
+})
+
+export default filterCreator
--- a/packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
+++ b/packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
@@ -1,0 +1,361 @@
+/*
+*                      Copyright 2024 Salto Labs Ltd.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+import { BuiltinTypes, Change, createRefToElmWithValue, Element, getChangeData, InstanceElement, isAdditionOrModificationChange, isInstanceChange, isInstanceElement, isListType, isObjectType, ObjectType, Value, Values } from '@salto-io/adapter-api'
+import _ from 'lodash'
+import { TransformFuncArgs, transformValues, WALK_NEXT_STEP, WalkOnFunc, walkOnValue } from '@salto-io/adapter-utils'
+import { parse, j2xParser } from 'fast-xml-parser'
+import { decode, encode } from 'he'
+import { collections, strings } from '@salto-io/lowerdash'
+import { logger } from '@salto-io/logging'
+import { APPLICATION_ID, DATASET, REAL_VALUE_KEY, SCRIPT_ID, SOAP_SCRIPT_ID, WORKBOOK } from '../constants'
+import { LocalFilterCreator } from '../filter'
+import { ATTRIBUTE_PREFIX, CDATA_TAG_NAME } from '../client/constants'
+import { parsedWorkbookType } from '../type_parsers/analytics_parsers/parsed_workbook'
+import { parsedDatasetType } from '../type_parsers/analytics_parsers/parsed_dataset'
+import { TypeAndInnerTypes } from '../types/object_types'
+import { AnalyticOriginalFields, CHART_IDS, DATA_VIEW_IDS, DEFAULT_VALUE, DEFAULT_XML_TYPE, DEFINITION, DEPENDENCIES, DO_NOT_ADD, EmptyObject, EXPRESSION_VALUE_VALUE_REGEX, FALSE, FIELD_DEFINITION, FIELD_TYPE, fieldsToOmitFromDefinition, fieldsToOmitFromOriginal, FieldWithType, ITEM, NAME, originalFields, OriginalWorkbookArrays, PIVOT_IDS, ROOT, StringToStringRecord, TRANSLATION_SCRIPT_ID, TRUE, TValuesToIgnore, TYPE, XML_TYPE, xmlType } from '../type_parsers/analytics_parsers/analytics_constants'
+import { captureServiceIdInfo } from '../service_id_info'
+import { workbookType } from '../autogen/types/standard_types/workbook'
+import { datasetType } from '../autogen/types/standard_types/dataset'
+
+const log = logger(module)
+const { awu } = collections.asynciterable
+
+const isNotAtrophiedScriptIdField = (name: string, val: Value): boolean =>
+  (name.endsWith(SOAP_SCRIPT_ID) ? !_.isEqual(val, { [TYPE]: xmlType.null }) : true)
+
+const fetchTransformFunc = async (
+  { value, field, path }: TransformFuncArgs,
+  type: ObjectType,
+): Promise<Value> => {
+  const fieldType = path?.isTopLevel() ? type : await field?.getType()
+  if (
+    fieldType === undefined
+    && path !== undefined
+    && isNotAtrophiedScriptIdField(path.getFullName(), value)
+  ) {
+    log.debug('unexpected path in analytics type. Path: %s', path?.getFullName())
+  }
+  if (
+    fieldType?.elemID.isEqual(BuiltinTypes.UNKNOWN.elemID)
+    && !(_.isPlainObject(value) && value[TYPE] === xmlType.null)
+    && path !== undefined
+    && !EXPRESSION_VALUE_VALUE_REGEX.test(path.getFullName())
+  ) {
+    log.debug('Found a value with an unkown type. path: %s, value: %o', path?.getFullName(), value)
+  }
+  if (_.isPlainObject(value)) {
+    if (value[TYPE] === xmlType.null) {
+      return undefined
+    }
+    if (value[TYPE] === xmlType.array) {
+      return collections.array.makeArray(value[ITEM])
+    }
+    if (value[TYPE] === xmlType.boolean) {
+      return value[REAL_VALUE_KEY]
+    }
+    if (value[TYPE] === xmlType.string) {
+      return String(value[REAL_VALUE_KEY])
+    }
+    if (FIELD_DEFINITION in value) {
+      if (!isObjectType(fieldType) || !(XML_TYPE in fieldType.annotations)) {
+        log.debug('unexpected _T_ field in analytic instance. Path: %s', path?.getFullName())
+        return {
+          [FIELD_TYPE]: value[FIELD_DEFINITION],
+          ..._.omit(value, FIELD_DEFINITION),
+        }
+      }
+      return TValuesToIgnore.has(value[FIELD_DEFINITION]) ? _.omit(value, FIELD_DEFINITION) : {
+        [value[FIELD_DEFINITION]]: _.omit(value, FIELD_DEFINITION),
+      }
+    }
+    return value
+  }
+  if (fieldType?.elemID.isEqual(BuiltinTypes.STRING.elemID)) {
+    return String(value)
+  }
+  return value
+}
+
+const createAnalyticsInstance = async (
+  instance: InstanceElement,
+  analyticsType: ObjectType,
+): Promise<void> => {
+  const definitionValues = _.omit(parse(instance.value[DEFINITION], {
+    attributeNamePrefix: ATTRIBUTE_PREFIX,
+    ignoreAttributes: false,
+    tagValueProcessor: val => decode(val),
+  })[ROOT], fieldsToOmitFromDefinition)
+
+  const updatedValues = await transformValues({
+    values: definitionValues,
+    type: analyticsType,
+    transformFunc: args => fetchTransformFunc(args, analyticsType),
+    strict: false,
+    pathID: instance.elemID,
+  })
+
+  if (updatedValues !== undefined) {
+    instance.value = {
+      ..._.omit(instance.value, fieldsToOmitFromOriginal),
+      ...updatedValues,
+    }
+  }
+}
+
+const createEmptyObjectOfType = async (
+  fieldType: ObjectType,
+  key: string,
+): Promise<EmptyObject> => {
+  if (DEFAULT_VALUE in fieldType.fields[key].annotations) {
+    return fieldType.fields[key].annotations[DEFAULT_VALUE]
+  }
+  const keyType = await fieldType.fields[key].getType()
+
+  if (isListType(keyType)) {
+    return {
+      [TYPE]: xmlType.array,
+    }
+  }
+
+  return {
+    [TYPE]: xmlType.null,
+  }
+}
+
+const handleTypeField = (value: Value): FieldWithType => {
+  if (Array.isArray(value)) {
+    return {
+      [TYPE]: xmlType.array,
+      [ITEM]: value,
+    }
+  }
+  if (_.isBoolean(value)) {
+    return {
+      [TYPE]: xmlType.boolean,
+      [REAL_VALUE_KEY]: value ? TRUE : FALSE,
+    }
+  }
+  if (_.isString(value) && strings.isNumberStr(value)) {
+    return {
+      [TYPE]: xmlType.string,
+      [REAL_VALUE_KEY]: value,
+    }
+  }
+  return value
+}
+
+const handleFieldDefinition = (value: Value): Value => {
+  if (_.isPlainObject(value)) {
+    const innerKeys = Object.keys(value)
+    if (innerKeys.includes(FIELD_TYPE)) {
+      value[FIELD_DEFINITION] = value[FIELD_TYPE]
+      delete value[FIELD_TYPE]
+    } else if (innerKeys.includes(FIELD_DEFINITION) && !TValuesToIgnore.has(value[FIELD_DEFINITION])) {
+      // We assume that we have exactly 2 keys because the only option for an object
+      // to include the FIELD_DEFINITION key is if we added it in addMissingFields and we do so in 2 options:
+      // 1. The field DEFAULT_XML_TYPE exists
+      // 2. The field DEFAULT_XML_TYPE does not exist (not in the TValuesToIgnore) and there is only one other key
+      if (innerKeys.length !== 2) {
+        log.warn('The number of keys in an object with FIELD_DEFINITION is not 2')
+        return value
+      }
+      const otherKey = innerKeys.filter(innerKey => innerKey !== FIELD_DEFINITION)[0]
+      return {
+        [FIELD_DEFINITION]: value[FIELD_DEFINITION],
+        ...value[otherKey],
+      }
+    }
+  }
+  return value
+}
+
+const matchToXmlObjectForm = (instance: InstanceElement, definitionValues: Values): void => {
+  const deployWalkFunc: WalkOnFunc = ({ value }) => {
+    if (_.isPlainObject(value)) {
+      const keys = Object.keys(value)
+
+      keys.forEach(key => {
+        value[key] = handleFieldDefinition(value[key])
+      })
+
+      // to prevent endless recursion in children arrays
+      if (!(TYPE in value)) {
+        keys.forEach(key => {
+          value[key] = handleTypeField(value[key])
+        })
+      }
+    } else if (Array.isArray(value)) {
+      value.forEach((_val, index) => {
+        value[index] = handleFieldDefinition(value[index])
+        value[index] = handleTypeField(value[index])
+      })
+    }
+    return WALK_NEXT_STEP.RECURSE
+  }
+
+  walkOnValue({
+    elemId: instance.elemID,
+    value: definitionValues,
+    func: deployWalkFunc,
+  })
+}
+
+const addMissingFields = async (
+  instance: InstanceElement,
+  definitionValues: Values,
+  analyticsType: ObjectType,
+): Promise<void> => {
+  const deployTransformFunc = async (
+    { value, field, path }: TransformFuncArgs,
+  ): Promise<Value> => {
+    const fieldType = path?.isTopLevel() ? analyticsType : await field?.getType()
+    if (isObjectType(fieldType) && _.isPlainObject(value) && !(TYPE in value)) {
+      if (XML_TYPE in fieldType.annotations
+        && Object.keys(value).length === 1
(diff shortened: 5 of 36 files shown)


dantal4 (commented):
Great work!
I didn't review the parsers & parsed types yet.
Just verifying - all CV messages are approved by a PM?

packages/netsuite-adapter/src/filters/add_referencing_workbooks.ts
```diff
@@ -0,0 +1,85 @@
...
+        .map(dataset => dataset.elemID.createNestedID('scriptid'))
+        .map(elemId => elemId.getFullName())
+    )
+    const validWorkbooks = await awu(await elementsSource.list())
+      .filter(elemID => elemID.typeName === WORKBOOK)
+      .map(async elemId => elementsSource.get(elemId))
+      .filter(isInstanceElement)
+      .filter(checkWorkbookValidity)
+      .toArray()
+    for (const workbook of validWorkbooks) {
```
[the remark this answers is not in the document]
SeggevHaimovich: why should I use reduce and not use forEach?
```
validWorkbooks.forEach(async workbook => {
      if (datasetFullNameSet.size !== 0) {
        const dependencies = getInstanceReferenceDependencies(workbook)
        const datasetDependencies = dependencies
          .map(dep => dep.elemID.getFullName())
          .filter(dep => datasetFullNameSet.has(dep))
        if (datasetDependencies.length > 0) {
          datasetDependencies.forEach(dep => datasetFullNameSet.delete(dep))
          await addInnerReferencesTopLevelParent(workbook, elementsSource)
          workbooksToPush.push(workbook)
        }
      }
    })
```

packages/netsuite-adapter/src/reference_dependencies.ts
```diff
@@ -61,6 +78,9 @@ export const findDependingElementsFromRefs = async (
         && !visitedIdToElement.has(elementFullName(topLevelParent))
         && elemID.adapter === NETSUITE
         && await isRefToServiceId(topLevelParent, elemID)) {
+        if (isInstanceElement(topLevelParent) && elementsSource !== undefined) {
+          topLevelParent.value = await addInnerReferencesTopLevelParent(topLevelParent.value, elementsSource)
```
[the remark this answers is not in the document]
SeggevHaimovich: Here I add the topLevelParent to references in the added elements.
If I don't do it the references in the translation collection don't have topLevelParent (because they also came from elementsSource.get) and it makes some problems in the deployment.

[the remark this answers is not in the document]
dantal4: if it's only for the added translation instances, it's better to call it on the final list of them (`referencedTranslationCollectionInstances`), and avoid running it on all elements.

packages/netsuite-adapter/src/type_parsers/dataset_parsing/parsed_dataset.ts
```diff
@@ -0,0 +1,539 @@
...
+        refType: datasetCriteria,
+        annotations: {
+        },
+      },
+      description: { refType: datasetTranslation },
+      formulas: { refType: new ListType(datasetFieldOrFormula) },
+      id: { refType: BuiltinTypes.UNKNOWN },
+      ownerId: { refType: BuiltinTypes.NUMBER },
+      version: { refType: BuiltinTypes.STRING },
+      scriptId: { refType: BuiltinTypes.UNKNOWN },
```
dantal4, line 534: why are there some `UNKNOWN` fields ?

packages/netsuite-adapter/src/type_parsers/workbook_parsing/parsed_workbook.ts
```diff
@@ -0,0 +1,877 @@
...
+
+// import { logger } from '@salto-io/logging'
+import { BuiltinTypes, CORE_ANNOTATIONS, ElemID, ListType, ObjectType, createRefToElmWithValue, createRestriction } from '@salto-io/adapter-api'
+import { createMatchingObjectType } from '@salto-io/adapter-utils'
+import { TypeAndInnerTypes } from '../../types/object_types'
+import * as constants from '../../constants'
+import { ApplicationId, Audience, BaseRecord, Condition, ConditionOrFilter, DEFAULT_VALUE, DO_NOT_ADD, DefinitionId, DefinitionScriptId, Dependencies, Expression, ExpressionValue, FieldOrFormula, FieldReference, Filter, Formula, FormulaFormula, IGNORE_T_VALUE, Join, JoinTrail, Meta, Operator, TranslationType, XML_TYPE } from '../dataset_parsing/parsed_dataset'
+import { fieldTypes } from '../../types/field_types'
+
+// const log = logger(module)
```
dantal4, line 25: leftover?

packages/netsuite-adapter/src/type_parsers/dataset_parsing/parsed_dataset.ts
```diff
@@ -0,0 +1,539 @@
...
+export type ParsedDataset = {
+  scriptid: string
+  name: string
+  dependencies?: {
+    dependency?: string[]
+  }
+  definition?: string
+} & DatasetDefinitionType
+
+export const ParsedDatasetType = (): TypeAndInnerTypes => {
```
dantal4, line 175: function name should be in camel case

packages/netsuite-adapter/src/type_parsers/workbook_parsing/parsed_workbook.ts
```diff
@@ -0,0 +1,877 @@
...
+type TargetFieldContextName = typeof targetFieldContextNameList[number]
+type FieldStateName = typeof fieldStateNameList[number]
+
+type TargetFieldContext = {
+  name?: TargetFieldContextName
+}
+
+type Sorting = {
+  caseSensitive?: boolean
+  direction?: 'ASCENDING' | 'DESCENDING'
```
dantal4, line 39: can you reuse it for the restriction like you did in `FieldStateName` ?

packages/netsuite-adapter/src/type_parsers/workbook_parsing/parsed_workbook.ts
```diff
@@ -0,0 +1,877 @@
...
+  Workbook?: InnerWorkbook
+}
+
+export type Workbook = {
+  scriptid: string
+  name: string
+  dependencies?: {
+    dependency?: string[]
+  }
+  definition?: string
```
dantal4, line 170: can be shared with dataset's original fields?

packages/netsuite-adapter/src/type_parsers/workbook_parsing/parsed_workbook.ts
```diff
@@ -0,0 +1,877 @@
...
+    fields: {
+      id: { refType: BuiltinTypes.UNKNOWN },
+      scriptId: { refType: BuiltinTypes.UNKNOWN },
+      applicationId: { refType: BuiltinTypes.UNKNOWN },
+      version: { refType: BuiltinTypes.STRING },
+      name: { refType: workbookTranslation },
+      workbook: { refType: BuiltinTypes.STRING },
+      datasets: { refType: new ListType(BuiltinTypes.STRING) },
+      mapping: {
+        refType: createRefToElmWithValue(fieldTypes.cdata),
```
dantal4, line 716: is there a difference between that being `cdata` and not `string` ?

dantal4 (commented):
Great work!

packages/netsuite-adapter/src/autogen/types/standard_types/workbook.ts
```diff
@@ -72,7 +72,7 @@ export const workbookType = (): TypeAndInnerTypes => {
     },
     fields: {
       dependency: {
-        refType: createRefToElmWithValue(BuiltinTypes.STRING /* Original type was single-select list */),
+        refType: createRefToElmWithValue(new ListType(BuiltinTypes.STRING) /* Original type was single-select list */),
```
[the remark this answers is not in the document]
omrilit: As part of this PR, environments that already manage Workbooks will have a lot of noise in the Workbook structure.
This particular change will cause noise for all WSs in the Types folder but this is negligible and we shouldn't do something about it.
We will have to manage the communication of this feature to customers that already use workbooks.

packages/netsuite-adapter/src/change_validators/check_referenced_datasets.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  const {
+    withinTheDeployment: unreferencedInDeployment,
+    withinTheEnvironment: unreferencedInEnvironment,
+  } = await getUnreferencedDatasetsDeploymentAndEnvironment(changedDatasets, changedWorkbooks, elementsSource)
+
+  const errors = unreferencedInDeployment
+    .map(({ elemID }): ChangeError => ({
+      elemID,
+      severity: 'Info',
```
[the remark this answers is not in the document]
SeggevHaimovich: We discussed it with Lior and that is the severity that was decided

packages/netsuite-adapter/src/autogen/types/standard_types/workbook.ts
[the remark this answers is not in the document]
SeggevHaimovich: We discussed this matter. After this PR, all the environments fetching workbooks/datasets will experience noise, as all the workbooks and datasets are going to be changed, regardless of this change.

packages/netsuite-adapter/src/filters/add_referencing_workbooks.ts
```diff
@@ -0,0 +1,114 @@
...
+  preDeploy: async changes => {
+    const instanceElems = changes
+      .filter(isInstanceChange)
+      .map(getChangeData)
+
+    await getWorkbooksToPush(
+      instanceElems.filter(instance => instance.elemID.typeName === DATASET),
+      instanceElems.filter(instance => instance.elemID.typeName === WORKBOOK),
+      elementsSource,
+    ).then(arr => arr.forEach(workbook => changes.push(toChange({ before: workbook, after: workbook }))))
```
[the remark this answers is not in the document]
SeggevHaimovich: The only change in the elements is adding top level parent to the inner references.
What problem could it make?

packages/netsuite-adapter/src/type_parsers/workbook_parsing/parsed_workbook.ts
SeggevHaimovich: I changed it when I thought the problem with the pivots' definition was in the code.
I don't really know what is the difference but I guess it is more accurate this way.

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,409 @@
...
+    if (TYPE in value) {
+      if (value[TYPE] === 'null') {
+        return {}
+      } if (value[TYPE] === 'array') {
+        if (value[ITEM] !== undefined) {
+          return (Array.isArray(value[ITEM])) ? value[ITEM] : [value[ITEM]]
+        }
+        return []
+      } if (value[TYPE] === 'boolean') {
+        return value[TEXT]
```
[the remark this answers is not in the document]
SeggevHaimovich: it would be true/false (boolean not text)

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,409 @@
...
+      } if (value[TYPE] === 'string') {
```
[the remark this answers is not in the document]
SeggevHaimovich: if the value is number and it considered as number than there won't be a TYPE field at all.
for example:
number: <value>18</value>
string of number: <version type="string">0.1</version>

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,409 @@
...
+    if (T in value) {
+      if (!isObjectType(fieldType) || !(XML_TYPE in fieldType.annotations)) {
+        log.debug('unexpected _T_ field in analytic instance. Path: %o', path)
+        return {
+          xmlType: value[T],
+          ..._.omit(value, T),
+        }
+      }
+      return TValuesToIgnore.has(value[T]) ? _.omit(value, T) : {
+        [value[T]]: _.omit(value, T),
```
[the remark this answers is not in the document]
SeggevHaimovich: `<_T_>dataSetFormula</_T_>
       <formula>
             <_T_>formula</_T_>
`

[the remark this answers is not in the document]
SeggevHaimovich: image
notice that under formula there is a `<_T_>formula<_T_>`

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,409 @@
...
+      }
+    }
+    return value
+  }
+  if (isString(value) && value.startsWith('custcollectiontranslations')) {
+    const nameParts = value.split('.')
+    if (nameParts.length === 2) {
+      const instanceElemId = new ElemID(NETSUITE, 'translationcollection', 'instance', nameParts[0])
+      const instance = custCollectionTranslationsMap.get(instanceElemId.getFullName())
+      if (instance !== undefined && instance.value.strings?.string?.[nameParts[1]]?.scriptid !== undefined) {
```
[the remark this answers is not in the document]
SeggevHaimovich: I split the if to 2 different ones and added a log.
If Instance is undefined it could happen when someone doesn't fetch translation collection, this is not a weird behavior.

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,409 @@
...
+  const updatedValues = await transformValues({
+    values: definitionValues,
+    type: analyticsType,
+    transformFunc: args => fetchTransformFunc(args, analyticsType, custCollectionTranslationsMap),
+    strict: false,
+    pathID: instance.elemID,
+    allowEmpty: false,
+  })
+
+  if (updatedValues !== undefined) {
```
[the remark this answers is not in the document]
SeggevHaimovich: but then if something would go wrong, I will omit the 'definition' field from the value and the user will get dataset/workbook with nothing.

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,409 @@
...
+      await awu(Object.keys(fieldType.fields))
+        .filter(key => !(key in value) && fieldType.fields[key].annotations[DO_NOT_ADD] !== true)
+        .forEach(async key => {
+          value[key] = await createEmptyObjectOfType(fieldType, key)
+        })
+    }
+  }
+  return value
+}
+const checkReference = (key: string, value: Value): Value => {
```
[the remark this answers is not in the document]
SeggevHaimovich: I check that the key is equal 'translationScriptId'.
I can do this outside the function but I think it is cleaner that way

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,409 @@
...
+    return {
+      [TYPE]: 'array',
+      [ITEM]: value[key],
+    }
+  }
+  if (isBoolean(value[key])) {
+    // eslint-disable-next-line no-param-reassign
+    return {
+      [TYPE]: 'boolean',
+      [TEXT]: String(value[key]),
```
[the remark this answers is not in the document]
SeggevHaimovich: this was a left-over, I kept it as I wrote and deleted the "eslint-disable-next-line" line

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,409 @@
...
+}
+
+const checkT = (key: string | number, value: Value): Value => {
+  const innerVal = value[key]
+  if (isPlainObject(innerVal)) {
+    const innerkeys = Object.keys(innerVal)
+    if (innerkeys.includes('xmlType')) {
+      innerVal[T] = innerVal.xmlType
+      delete innerVal.xmlType
+    } else if (innerkeys.length === 2 && innerkeys.includes(T)) {
```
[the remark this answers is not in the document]
SeggevHaimovich: there can be 2 scenarios where the innerkeys include _T_ 
1. there is a default value for _T_ (like formula)
2. there isn't a default value and I created the _T_ based on the single value that were in this object beforehand.

I changed the condition to be:
`innerKeys.includes(FIELD_DEFINITION) && !TValuesToIgnore.has(value[FIELD_DEFINITION])`
and in that case there has to be exactly 2 keys

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,409 @@
...
+      const otherKey = innerkeys.filter(innerKey => innerKey !== T)[0]
+      return {
+        [T]: innerVal[T],
+        ...innerVal[otherKey],
```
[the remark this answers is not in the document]
SeggevHaimovich: I don't think that this does what I want it to do.
I'll give an example:
The original:
`x = {
  _T_: 1,
  1: {
    a: 2,
    b: 3,
  },
}`
I want to make it:
`y = {
  _T_: 1,
  a: 2,
  b: 3,
}`

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,409 @@
...
+    // to prevent endless recursion in children arrays
+    if (!(TYPE in value)) {
+      keys.forEach(key => {
+        value[key] = checkTypeField(key, value)
+      })
+    }
+  } else if (Array.isArray(value)) {
+    value.forEach((_val, index) => {
+      value[index] = checkT(index, value)
+      value[index] = checkTypeField(index, value)
```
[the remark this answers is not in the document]
SeggevHaimovich: because this is 2 different actions, I think that I should separate between methods to make it more readable

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,409 @@
...
+  definitionValues.name = instance.value.name
+  if (isReferenceExpression(instance.value.name?.[TEXT])) {
+    const name = instance.value.name[TEXT].elemID.getFullName()
+    if (name !== undefined && checkReferenceToTranslation(name)) {
+      const nameList = name.split('.')
+      definitionValues.name = {
+        translationScriptId: nameList[3].concat('.', nameList[6]),
+      }
+    }
+  }
```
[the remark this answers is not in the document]
SeggevHaimovich: I added a function that changes the definition value so there is no return value.
In any case I can't know what type is the instance.value.name (the user can put anything) so I would have to write Value in the return type.

packages/netsuite-adapter/src/type_parsers/dataset_parsing/parsed_dataset.ts
```diff
@@ -0,0 +1,539 @@
...
+export const TYPE = '@_type'
+
+type AudienceItem = Value
+
+
+type ExpressionSubType = Value
+type ExpressionUiData = Value
+
+// maybe we should ignore these values
+export type ApplicationId = Value
```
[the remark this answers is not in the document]
SeggevHaimovich: I did it because I thought it would be good that in the future when users will use datasets we will know these types and we could update this.
This way I make sure we won't forget to update it in the objectType.

packages/netsuite-adapter/src/type_parsers/dataset_parsing/parsed_dataset.ts
SeggevHaimovich: Because I didn't find any dataset that has values in those fields.
these are values that came from the definition and they are always null

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,409 @@
...
+  if (isContainerType(keyType)) {
+    // we only have lists in the types
+    return arrayObject
+  }
+
+  return nullObject
+}
+
+const checkReferenceToTranslation = (name: string): boolean => {
+  const regex = /^netsuite.translationcollection.instance.\w+.strings.string.\w+.scriptid$/
```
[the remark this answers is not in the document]
SeggevHaimovich: I changed it so the creation and the validation use the same 'translationPathList', a shared list of strings that represent the path to the scriptid

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,409 @@
...
+
+const deployTransformFunc = async (
+  { value, field, path }: TransformFuncArgs,
+  analyticsType: ObjectType,
+): Promise<Value> => {
+  const fieldType = (path?.getFullName().split('.').length === 4) ? analyticsType : await field?.getType()
+  if (isObjectType(fieldType) && isPlainObject(value) && !(TYPE in value)) {
+    if (XML_TYPE in fieldType.annotations
+      && Object.keys(value).length === 1
+      && !isString(fieldType.annotations[XML_TYPE])) {
```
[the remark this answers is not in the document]
SeggevHaimovich: what I did earlier was:
I added for every field in the xml that has the _T_ value, the XML_TYPE annotations. the fields with the determined value in _T_ (formula, workbook and dataset) had the relevant string as the value of the annotation and other fields with _T_ had the value true in the annotation.

I split it to 2 different annotations:
XML_TYPE - fields with _T_ in the xml will have this annotation with true value.
DEFAULT_XML_TYPE - the fields formula, workbook and dataset will have this annotation with the value formula|workbook|dataset accordingly

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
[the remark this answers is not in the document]
dantal4: you're right

packages/netsuite-adapter/src/type_parsers/dataset_parsing/parsed_dataset.ts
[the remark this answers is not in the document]
dantal4: it still should be `unknown` and not `Value`. using `Value` will prevent any type-safety for these fields.

packages/netsuite-adapter/src/type_parsers/workbook_parsing/parsed_workbook.ts
dantal4: how does this field value looks like? (BTW, you can remove the `createRefToElmWithValue`)

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,415 @@
...
+  if (Array.isArray(value)) {
+    return {
+      [TYPE]: xmlType.array,
+      [ITEM]: value,
+    }
+  }
+  if (_.isBoolean(value)) {
+    return {
+      [TYPE]: xmlType.boolean,
+      [REAL_VALUE_KEY]: String(value),
```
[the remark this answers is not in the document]
dantal4: maybe the `String(...)` here is unnecessary?

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,415 @@
...
+      if (XML_TYPE in fieldType.annotations
+        && Object.keys(value).length === 1
+        && fieldType.annotations[DEFAULT_XML_TYPE] === undefined) {
+        const [key] = Object.keys(value)
+        value[FIELD_DEFINITION] = key
+      } else {
+        if (fieldType.annotations[DEFAULT_XML_TYPE] !== undefined) {
+          value[FIELD_DEFINITION] = fieldType.annotations[DEFAULT_XML_TYPE]
+        }
+        await awu(Object.keys(fieldType.fields))
```
dantal4, line 264: is this suppose to be inside the "else" here?

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,415 @@
...
+  arrName: string,
+  value: Values,
+): StringToStringRecord[] =>
+  collections.array.makeArray(value[arrName])
+    .filter(_.isPlainObject)
+    .map((val: Values) => val[arrName.slice(0, -1)])
+    .flatMap((val: Values) => {
+      const { scriptId } = val
+      return _.isString(scriptId) ? { [SCRIPT_ID]: scriptId } : []
+    })
```
dantal4, line 293: safer:
```suggestion
    .flatMap((val: Values) => (
      _.isPlainObject(val) && _.isString(val[ALSO_SCRIPT_ID])
        ? { [SCRIPT_ID]: val[ALSO_SCRIPT_ID] } : []
    ))
```

packages/netsuite-adapter/src/type_parsers/analytics_parsers/analytics_constants.ts
```diff
@@ -0,0 +1,230 @@
...
+  dependencies?: {
+    dependency?: string[]
+  }
+  definition?: string
+  // charts?: workbookListsElement[]
+  // pivots?: workbookListsElement[]
+  // tables?: workbookListsElement[]
+}
+
+export type StringToStringRecord = { [key: string]: string }
```
dantal4, line 189: why isn't it `{ scriptid: string }` ? I think that it is only used this way

packages/netsuite-adapter/src/type_parsers/analytics_parsers/analytics_constants.ts
```diff
@@ -0,0 +1,230 @@
...
+export type OriginalWorkbookArrays = { [key: string]: { [innerKey: string]: StringToStringRecord[] } }
```
dantal4, line 190: nit
```suggestion
export type OriginalWorkbookArrays = Record<string, Record<string, StringToStringRecord[]>>
```

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,415 @@
...
+  tables: {
+    table: createOriginalArray(DATA_VIEWS, value),
+  },
+})
+
+const addNameToDefinition = (instance: InstanceElement, definitionValues: Values): void => {
+  definitionValues[NAME] = instance.value[NAME]
+  // REAL_VALUE_KEY is a value added in translation_converter preDeploy filter
+  // it represents the value of the field and the type of the field is given in another innerField
+  if (isReferenceExpression(instance.value[NAME]?.[REAL_VALUE_KEY])) {
```
dantal4, line 311: how is it different than `convertReference` ?

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,415 @@
...
+        [TRANSLATION_SCRIPT_ID]: nameList[3].concat('.', nameList[6]),
+      }
+    }
+  }
+}
+
+const returnToOriginalShape = async (
+  instance: InstanceElement,
+  analyticsType: ObjectType,
+): Promise<AnalyticOriginalFields> => { // TODO how does it work? I add arrays and they are not in the type
```
dantal4, line 325: leftover?

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,415 @@
...
+    cdataTagName: CDATA_TAG_NAME,
+    tagValueProcessor: val => encode(val.toString()),
+  }).parse({ [ROOT]: definitionValues })
+
+  const arrays = (analyticsType.elemID.typeName === WORKBOOK) ? createOriginalArrays(instance.value) : []
+
+  return {
+    [NAME]: instance.value[NAME],
+    [SCRIPT_ID]: instance.value[SCRIPT_ID],
+    [DEPENDENCIES]: instance.value[DEPENDENCIES],
```
dantal4, line 357: APPLICATION_ID is missing here

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,415 @@
...
+      }
+    }
+  }
+}
+
+const returnToOriginalShape = async (
+  instance: InstanceElement,
+  analyticsType: ObjectType,
+): Promise<AnalyticOriginalFields> => { // TODO how does it work? I add arrays and they are not in the type
+  const definitionValues: Values = _.omit(instance.clone().value, Object.values(originalFields))
```
dantal4, line 326: why are you cloning the instance?

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,415 @@
...
+      .filter(isInstanceChange)
+      .map(getChangeData)
+    const typeMap: { [key: string]: ObjectType} = {
+      [DATASET]: parsedDatasetType().type,
+      [WORKBOOK]: parsedWorkbookType().type,
+    }
+    await awu(instanceElemsFromAdditionOrModification)
+      .filter(instance => instance.elemID.typeName === DATASET || instance.elemID.typeName === WORKBOOK)
+      .forEach(async instance => {
+        instance.value = await returnToOriginalShape(instance, typeMap[instance.elemID.typeName])
```
dantal4, line 410: why aren't you using `instance.getType()`?

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,415 @@
...
+
+    await changeType(parsedDatasetType(), elements, custCollectionTranslationsMap)
+  },
+
+  preDeploy: async (changes: Change[]) => {
+    const instanceElemsFromAdditionOrModification = changes
+      .filter(isAdditionOrModificationChange)
+      .filter(isInstanceChange)
+      .map(getChangeData)
+    const typeMap: { [key: string]: ObjectType} = {
```
dantal4, line 403: nit
```suggestion
    const typeMap: Record<string, ObjectType> = {
```

packages/netsuite-adapter/src/type_parsers/analytics_parsers/analytics_constants.ts
```diff
@@ -0,0 +1,230 @@
...
+export const NAME = 'name'
+export const TRANSLATION_SCRIPT_ID = 'translationScriptId'
+export const CHARTS = 'charts'
+export const PIVOTS = 'pivots'
+export const TABLES = 'tables'
+export const DATA_VIEWS = 'dataViews'
+
+// types
+
+type AudienceItem = Value
```
dantal4, line 51: please see [this](https://github.com/salto-io/salto/pull/4892#discussion_r1378940166) comment

packages/netsuite-adapter/src/type_parsers/dataset_parsing/parsed_dataset.ts
```diff
@@ -0,0 +1,409 @@
...
+      definition: { refType: createRefToElmWithValue(fieldTypes.cdata) },
+      applicationId: { refType: BuiltinTypes.UNKNOWN },
+      audience: { refType: datasetAudience },
+      baseRecord: { refType: datasetBaseRecord },
+      columns: { refType: new ListType(datasetColumn) },
+      criteria: {
+        refType: datasetCriteria,
+        annotations: {
+        },
+      },
```
dantal4, line 398: nit
```suggestion
      criteria: { refType: datasetCriteria },
```

packages/netsuite-adapter/src/type_parsers/workbook_parsing/parsed_workbook.ts
```diff
@@ -0,0 +1,867 @@
...
+*/
+
+import { BuiltinTypes, CORE_ANNOTATIONS, ElemID, ListType, ObjectType, createRefToElmWithValue, createRestriction } from '@salto-io/adapter-api'
+import { createMatchingObjectType } from '@salto-io/adapter-utils'
+import { TypeAndInnerTypes } from '../../types/object_types'
+import * as constants from '../../constants'
+import { fieldTypes } from '../../types/field_types'
+import { AnalyticOriginalFields, ApplicationId, Audience, BaseRecord, Condition, ConditionOrFilter, DEFAULT_VALUE, DEFAULT_XML_TYPE, DO_NOT_ADD, DefinitionId, Dependencies, Expression, ExpressionValue, FieldOrFormula, FieldReference, Filter, Formula, FormulaFormula, IGNORE_T_VALUE, Join, JoinTrail, Meta, Operator, TranslationType, XML_TYPE } from '../analytics_parsers/analytics_constants'
+
+// const targetFieldContextNameList = ['DEFAULT', 'DISPLAY', 'HIERARCHY', 'IDENTIFIER', 'HIERARCHY_IDENTIFIER'] as const
```
dantal4, line 24: some leftovers in this file

packages/netsuite-adapter/src/type_parsers/workbook_parsing/parsed_workbook.ts
```diff
@@ -0,0 +1,867 @@
...
+  conditionalFormatFilter?: ConditionalFormatFilter
+}
+
+type RgbColor = {
+  blue?: number
+  green?: number
+  red?: number
+}
+
+type BackgroundColor = {
```
dantal4, line 60: looks like `Color` is defined the same

packages/netsuite-adapter/src/type_parsers/workbook_parsing/parsed_workbook.ts
```diff
@@ -0,0 +1,867 @@
...
+    annotations: {
+    },
+    fields: {
+      translationScriptId: { refType: BuiltinTypes.STRING },
+    },
+    path: [constants.NETSUITE, constants.TYPES_PATH, constants.WORKBOOK],
+  })
+  innerTypes.workbookTranslation = workbookTranslation
+
+  const workbookBaseRecordElemID = new ElemID(constants.NETSUITE, 'workbook_baseRecord')
```
dantal4, line 436: why did you use underscore in some type names and camel case in the others?

packages/netsuite-adapter/src/type_parsers/workbook_parsing/parsed_workbook.ts
```diff
@@ -0,0 +1,867 @@
...
+    fields: {
+      id: { refType: BuiltinTypes.UNKNOWN },
+      scriptId: { refType: BuiltinTypes.STRING },
+      applicationId: { refType: BuiltinTypes.UNKNOWN },
+      version: { refType: BuiltinTypes.STRING },
+      name: { refType: workbookTranslation },
+      workbook: { refType: BuiltinTypes.STRING },
+      datasets: { refType: new ListType(BuiltinTypes.STRING) },
+      mapping: {
+        refType: createRefToElmWithValue(fieldTypes.cdata),
```
dantal4, line 709: please answer [this](https://github.com/salto-io/salto/pull/4892/files#r1378945078) question

packages/netsuite-adapter/src/type_parsers/workbook_parsing/parsed_workbook.ts
```diff
@@ -0,0 +1,867 @@
...
+          [CORE_ANNOTATIONS.RESTRICTION]: createRestriction({ regex: '^custworkbook[0-9a-z_]+' }),
+        },
+      },
+      name: {
+        refType: BuiltinTypes.STRING,
+        annotations: {
+          _required: true,
+        },
+      },
+      dependencies: { refType: createRefToElmWithValue(workbookDependencies) },
```
dantal4, line 856: I think that `createRefToElmWithValue` can be removed all over here

packages/netsuite-adapter/test/type_parsers/analytics_parsers/parsed_dataset.test.ts
```diff
@@ -0,0 +1,85 @@
...
+
+import { getSubtypes } from '@salto-io/adapter-utils'
+import { datasetType } from '../../../src/autogen/types/standard_types/dataset'
+import { parsedDatasetType } from '../../../src/type_parsers/dataset_parsing/parsed_dataset'
+import { parsedWorkbookType } from '../../../src/type_parsers/workbook_parsing/parsed_workbook'
+
+describe('analytics parsed types tests', () => {
+  describe('parsed dataset tests', () => {
+    const { type: dataset, innerTypes } = parsedDatasetType()
+    test('each inner type has unique elemID', () => {
```
dantal4, line 25: please use `it` instead

packages/netsuite-adapter/test/type_parsers/analytics_parsers/parsed_dataset.test.ts
```diff
@@ -0,0 +1,85 @@
...
+        .filter(usedType => !innerTypesNames.has(usedType.elemID.getFullName()))
+      expect(typesNotFoundInInnerTypes.length).toEqual(0)
+    })
+    test('all of the fields in the original type should appear in the new type', () => {
+      const newFields = new Set<string>(Object.keys(dataset))
+      const unusedFields = Object.keys(datasetType().type).filter(fieldName => !newFields.has(fieldName))
+      expect(unusedFields.length).toEqual(0)
+    })
+  })
+  describe('parsed workbook tests', () => {
```
dantal4, line 54: please put in a different file, or rename this file to `netsuite-adapter/test/type_parsers/parsed_analytics_types.test.ts` (and maybe DRY this code)

dantal4 (commented):
Great work!
I think that we're close to finish this one 👍

packages/netsuite-adapter/src/adapter.ts
```diff
@@ -110,6 +112,8 @@ export const allFilters: (LocalFilterCreatorDefinition | RemoteFilterCreatorDefi
   { creator: dataTypesCustomFields },
   { creator: dataInstancesCustomFields },
   { creator: dataInstancesNullFields },
+  // analyticsDefinitionHandle must run before translationConverter
+  { creator: analyticsDefinitionHandle },
```
dantal4, line 116: Can you move the filter to run after `parseReportTypes` ? The filters after that should run on the parsed elements, and soon we'll add a filter that exclude elements by a criteria, and it should run on the parsed elements, but before we find references (in replaceElementReferences)

dantal4:
@SeggevHaimovich  can you add dataset & workbook to [here](https://github.com/salto-io/salto/blob/863b6a777aadbd750e2dfb1ad06c24faa9ffe585/packages/netsuite-adapter/src/filters/convert_lists.ts#L64) too?

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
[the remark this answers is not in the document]
SeggevHaimovich: Ok, got it

packages/netsuite-adapter/src/type_parsers/workbook_parsing/parsed_workbook.ts
SeggevHaimovich: In the XML it looks like this:
<mapping>&lt;root>&lt;_T_>reportDataSourceExpressionMapping&lt;/_T_>&lt;expressions type="array">&lt;_ITEM_ type="array">&lt;_ITEM_>&lt;function>field&lt;/function>&lt;args>&lt;type>&lt;_T_>recordKeyType&lt;/_T_>&lt;valueType>&lt;_T_>integerType&lt;/_T_>&lt;/valueType>&lt;/type>&lt;dataSourceId>custdataset27&lt;/dataSourceId>&lt;fieldId>id&lt;/fieldId>&lt;/args>&lt;/_ITEM_>&lt;_ITEM_>&lt;function>field&lt;/function>&lt;args>&lt;type>&lt;_T_>recordKeyType&lt;/_T_>&lt;valueType>&lt;_T_>integerType&lt;/_T_>&lt;/valueType>&lt;/type>&lt;dataSourceId>custdataset31&lt;/dataSourceId>&lt;fieldId>id&lt;/fieldId>&lt;/args>&lt;/_ITEM_>&lt;/_ITEM_>&lt;/expressions>&lt;/root></mapping>

in the nacl it looks like this:

image

packages/netsuite-adapter/src/filters/add_referencing_workbooks.ts
```diff
@@ -0,0 +1,113 @@
...
+    const instancesFromAdditionOrModificationChanges = changes
+      .filter(isAdditionOrModificationChange)
+      .filter(isInstanceChange)
+      .map(getChangeData)
+
+    await getWorkbooksToPush(
+      instancesFromAdditionOrModificationChanges.filter(instance => instance.elemID.typeName === DATASET),
+      instancesFromAdditionOrModificationChanges.filter(instance => instance.elemID.typeName === WORKBOOK),
+      elementsSource,
+    ).then(arr => arr.forEach(workbook => changes.push(toChange({ before: workbook, after: workbook }))))
```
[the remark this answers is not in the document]
SeggevHaimovich: I thought it would be good to avoid creating another const if I don't have to.
I changed it to what you wrote

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,415 @@
...
+  type: ObjectType,
+  custCollectionTranslationsMap: Map<string, InstanceElement>,
+): Promise<Value> => {
+  const fieldType = path?.isTopLevel() ? type : await field?.getType()
+  if (fieldType === undefined) {
+    log.debug('unexpected path in analytics type. Path: %s', path?.getFullName())
+  }
+  if (fieldType?.elemID.isEqual(BuiltinTypes.UNKNOWN.elemID)
+    && !(_.isPlainObject(value) && value[TYPE] === xmlType.null)) {
+    log.debug('Founded a value with an unkown type. path: %s, value-type: %s, value: %o', path?.getFullName(), typeof value, value)
```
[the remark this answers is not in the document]
SeggevHaimovich: why? won't it make it easier to understand the type of the field?

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,415 @@
...
+}
+
+const convertReference = (key: string, value: Value): Value => {
+  if (isReferenceExpression(value[key])) {
+    const name = value[key].elemID.getFullName()
+    if (key === TRANSLATION_SCRIPT_ID && checkReferenceToTranslation(name)) {
+      const parts = value[key].elemID.getFullNameParts()
+      return `${parts[3]}.${parts[6]}`
+    }
+    return name
```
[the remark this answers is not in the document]
SeggevHaimovich: We talked about it on zoom

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,415 @@
...
+          ..._.omit(value, FIELD_DEFINITION),
+        }
+      }
+      return TValuesToIgnore.has(value[FIELD_DEFINITION]) ? _.omit(value, FIELD_DEFINITION) : {
+        [value[FIELD_DEFINITION]]: _.omit(value, FIELD_DEFINITION),
+      }
+    }
+    return value
+  }
+  if (_.isString(value) && value.startsWith(CUST_COLLECTION)) {
```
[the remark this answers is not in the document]
SeggevHaimovich: We talked about it on zoom

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
SeggevHaimovich: No, there is a situation where we want to add the default_xml_type and to add the missing keys
For instance: the field_definition "formula", that we ignore because it is under the object "formula". I want to add the field_definition and to add other missing keys.

packages/netsuite-adapter/src/type_parsers/workbook_parsing/parsed_workbook.ts
dantal4: `cdata` value starts with `<![CDATA[`, and the `cdata` field type is used to help us transform a nacl back to sdf XML, with the `<![CDATA[` initial (that is not represented in the nacl). So IMO it should be here a `string`.

packages/netsuite-adapter/src/filters/add_referencing_workbooks.ts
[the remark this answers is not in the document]
dantal4: constants usually gives more readability to the code, so don't hesitate using more constants 👍

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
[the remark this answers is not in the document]
dantal4: when you use `%o` the formatted value matches its type:
```
> console.log('test: %o', "1")
test: '1'
> console.log('test: %o', 1)
test: 1
> console.log('test: %o', {a:1})
test: { a: 1 }
```

packages/netsuite-adapter/src/filters/translation_converter.ts
```diff
@@ -36,7 +36,7 @@ const FIELDS_TO_CONVERT = [
   new ElemID(NETSUITE, CENTER_TAB, 'field', 'label'),
   new ElemID(NETSUITE, SUBTAB, 'field', 'title'),
 ]
-const REAL_VALUE_KEY = '#text'
+export const REAL_VALUE_KEY = '#text'
```
dantal4, line 39: nit - move to `packages/netsuite-adapter/src/client/constants.ts` - there are more XML related constants there

packages/netsuite-adapter/src/type_parsers/analytics_parsers/analytics_constants.ts
```diff
@@ -0,0 +1,222 @@
...
+export const DO_NOT_ADD = 'do_not_add'
+export const IGNORE_T_VALUE = 'ignore_t_value'
+
+// xml fields
+export const FIELD_DEFINITION = '_T_'
+export const TYPE = '@_type'
+export const FIELD_TYPE = 'fieldType'
+
+// const strings
+export const CUST_COLLECTION = 'custcollection'
```
dantal4, line 34: not used anymore?

packages/netsuite-adapter/src/type_parsers/analytics_parsers/analytics_constants.ts
```diff
@@ -0,0 +1,222 @@
...
+]
+export const originalFields: types.TypeKeysEnum<AnalyticOriginalFields> = {
+  [SCRIPT_ID]: SCRIPT_ID,
+  [NAME]: NAME,
+  [DEPENDENCIES]: DEPENDENCIES,
+  [DEFINITION]: DEFINITION,
+  [APPLICATION_ID]: APPLICATION_ID,
+}
+
+export const translationPathList = {
```
dantal4, line 215: not used anymore?

packages/netsuite-adapter/src/type_parsers/dataset_parsing/parsed_dataset.ts
```diff
@@ -0,0 +1,402 @@
...
+      audience: { refType: datasetAudience },
+      baseRecord: { refType: datasetBaseRecord },
+      columns: { refType: new ListType(datasetColumn) },
+      criteria: { refType: datasetCriteria },
+      description: { refType: datasetTranslation },
+      formulas: { refType: new ListType(datasetFieldOrFormula) },
+      id: { refType: BuiltinTypes.UNKNOWN },
+      ownerId: { refType: BuiltinTypes.NUMBER },
+      version: { refType: BuiltinTypes.STRING },
+      scriptId: { refType: BuiltinTypes.STRING },
```
dantal4, line 396: we shouldn't have both `scriptid` & `scriptId`. the first is the one we're using in all other types, so let's remove the second field (and value from the instance)

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,352 @@
...
+const createOriginalArray = (
+  arrName: string,
+  value: Values,
+): StringToStringRecord[] =>
+  collections.array.makeArray(value[arrName])
+    .filter(_.isPlainObject)
+    .map((val: Values) => val[arrName.slice(0, -1)])
+    .flatMap((val: Values) => (
+      _.isPlainObject(val) && _.isString(val[SOAP_SCRIPT_ID])
+        ? { [SCRIPT_ID]: val[SOAP_SCRIPT_ID] } : []
```
dantal4, line 252: can we change `SOAP_SCRIPT_ID` to be `SCRIPT_ID` in the fetch? (similar to my previous comment)

packages/netsuite-adapter/src/type_parsers/workbook_parsing/parsed_workbook.ts
```diff
@@ -0,0 +1,811 @@
...
+import * as constants from '../../constants'
+import { fieldTypes } from '../../types/field_types'
+import { AnalyticOriginalFields, ApplicationId, Audience, BaseRecord, Condition, ConditionOrFilter, DEFAULT_VALUE, DEFAULT_XML_TYPE, DO_NOT_ADD, DefinitionId, Dependencies, Expression, ExpressionValue, FieldOrFormula, FieldReference, Filter, Formula, FormulaFormula, IGNORE_T_VALUE, Join, JoinTrail, Meta, Operator, TranslationType, XML_TYPE } from '../analytics_parsers/analytics_constants'
+
+const sortDirectionList = ['ASCENDING', 'DESCENDING'] as const
+
+type sortDirection = typeof sortDirectionList[number]
+
+type TargetFieldContext = {
+  // name?: TargetFieldContextName
```
dantal4, line 29: leftover?

packages/netsuite-adapter/src/type_parsers/workbook_parsing/parsed_workbook.ts
```diff
@@ -0,0 +1,811 @@
...
+  cellConditionalFormat?: CellConditionalFormat
+}
+
+type Column = {
+  conditionalFormat?: ConditionalFormat[]
+  criterion?: Filter
+  customLabel?: TranslationType
+  dataSetColumnId?: number
+  datasetScriptId?: string
+  // fieldStateName?: FieldStateName
```
dantal4, line 95: leftover?

dantal4 (commented):
We're almost there!

dantal4 (commented):
more comments on things that I saw when I ran it locally on my NS workspace

packages/netsuite-adapter/src/change_validators/check_referenced_datasets.ts
```diff
@@ -0,0 +1,114 @@
...
+    forErrors: unreferencedInEnvironment,
+  } = await getUnreferencedDatasetsDeploymentAndEnvironment(changedDatasets, changedWorkbooks, elementsSource)
+
+  const errors = unreferencedInDeployment
+    .map(({ elemID }): ChangeError => ({
+      elemID,
+      severity: 'Info',
+      message: 'To deploy this, an existing workbook will be re-deployed',
+      detailedMessage: 'A dataset must be deployed alongside a workbook which was connected to it. '
+        + 'Salto will automatically add an existing workbook which was already connected to this dataset to the SDF deployment. This should not affect the target environment.',
```
dantal4, line 102: we should probably add here "and the translation collection instances related to that workbook"
@omrilit  FYI

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "cSpell.words": [
```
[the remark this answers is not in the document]
SeggevHaimovich: No, I'll delete it

packages/netsuite-adapter/src/change_validators/unsupported_workbooks.ts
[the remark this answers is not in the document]
SeggevHaimovich: Yes, I wanted to make sure it works and then delete it, I'll delete it now

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,352 @@
...
+  })
+}
+
+const createOriginalArray = (
+  arrName: string,
+  value: Values,
+): StringToStringRecord[] =>
+  collections.array.makeArray(value[arrName])
+    .filter(_.isPlainObject)
+    .map((val: Values) => val[arrName.slice(0, -1)])
```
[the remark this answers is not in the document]
SeggevHaimovich: It is supposed to look like this"
image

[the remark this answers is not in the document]
SeggevHaimovich: It is supposed to look like this:
image

I changed it to use the list of scriptids under the workbook title:
image

Now it is simpler: 
`const createOriginalArray = (
  arrName: string,
  value: Values,
): StringToStringRecord[] =>
  collections.array.makeArray(value.Workbook?.[arrName])
    .flatMap((val: string) => ({ [SCRIPT_ID]: val }))`

packages/netsuite-adapter/src/type_parsers/analytics_parsers/analytics_constants.ts
SeggevHaimovich: I used `CUST_COLLECTION` in the references code, now that I deleted it I don't use `CUST_COLLECTION`

packages/netsuite-adapter/src/type_parsers/analytics_parsers/analytics_constants.ts
SeggevHaimovich: same as last comment

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,352 @@
...
+    ...arrays,
+  }
+}
+
+const changeType = async (
+  { type: analyticType, innerTypes }: TypeAndInnerTypes,
+  elements: Element[],
+): Promise<void> => {
+  _.remove(elements, e => isObjectType(e) && e.elemID.typeName === analyticType.elemID.name)
+  _.remove(elements, e => isObjectType(e) && e.elemID.name.startsWith(analyticType.elemID.name))
```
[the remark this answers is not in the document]
SeggevHaimovich: Ok, I'll change it.
We do the same thing in `parse_report_types.td`, should I change it there too?

packages/netsuite-adapter/src/type_parsers/dataset_parsing/parsed_dataset.ts
SeggevHaimovich: There is a field named scriptId in the definition and there is a field named scriptid outside it.
I haven't seen any scriptId value different than null.
We can erase this field (and the deployment will still work), but than we will have a missing field. should I delete it?

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -0,0 +1,352 @@
...
+        && Object.keys(value).length === 1
+        && fieldType.annotations[DEFAULT_XML_TYPE] === undefined) {
+        const [key] = Object.keys(value)
+        value[FIELD_DEFINITION] = key
+      } else {
+        if (fieldType.annotations[DEFAULT_XML_TYPE] !== undefined) {
+          value[FIELD_DEFINITION] = fieldType.annotations[DEFAULT_XML_TYPE]
+        }
+        await awu(Object.keys(fieldType.fields))
+          .filter(key => !(key in value) && fieldType.fields[key].annotations[DO_NOT_ADD] !== true)
```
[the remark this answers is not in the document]
SeggevHaimovich: I think that I should add the DO_NOT_ADD as you said earlier.
This way we don't treat this as a special case, in my opinion this is better...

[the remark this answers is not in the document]
dantal4: so please make sure that you didn't miss any of those fields 🙏

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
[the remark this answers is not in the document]
dantal4: sounds good 👍

packages/netsuite-adapter/src/type_parsers/dataset_parsing/parsed_dataset.ts
dantal4: Let's remove the field only from the type, and in the future if we'll see a case where `scriptId` is not null, then decide what to do with it.

packages/netsuite-adapter/src/type_parsers/analytics_parsers/analytics_constants.ts
dantal4: so let's delete it too 👍

dantal4 (approved):
Really great work 💯 💯 💯 💯  👍 👍

Before you merge this PR, we should make sure to communicate this change to all customers.
After this PR is merged, please do the following:
1. fix types_generator so it will automatically set the right field type for workbook dependencies.
2. add fetch + deploy of both types to NS e2e

PR #4893, opened 2023-09-19
SALTO-4669 SALTO-4523 add metadata types to default exclude list

benisraeladi:
Add problematic metadata types to default exclude list

---

---
_Release Notes_: 
Salesforce Adapter:
- Add EclairGeoData, OmniUiCard, OmniDataTransform, OmniIntegrationProcedure, OmniInteractionAccessConfig, OmniInteractionConfig, and OmniScript to default exclude list
---
_User Notifications_:  _None_


tomsellek (approved):
Approved. We should probably comment somewhere that (per the comments in SALTO-4523) if any customer has the "new" (whatever that means) version of OmniStudio this change will come back to bite us.

PR #4894, opened 2023-09-19
SALTO-4792

benisraeladi:
Wrap organizationWideSharingDefaultsFilter with ensureSafeFilterFetch

---
---

_Release Notes_: 
_None_

---

_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/src/filters/organization_wide_sharing_defaults.ts
+++ b/packages/salesforce-adapter/src/filters/organization_wide_sharing_defaults.ts
@@ -18,7 +18,7 @@ import _ from 'lodash'
 import { CORE_ANNOTATIONS, ElemID, InstanceElement, ObjectType, Values } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
 import { RemoteFilterCreator } from '../filter'
-import { queryClient } from './utils'
+import { ensureSafeFilterFetch, queryClient } from './utils'
 import { apiName, getSObjectFieldElement, getTypePath } from '../transformers/transformer'
 import {
   API_NAME,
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/src/filters/organization_wide_sharing_defaults.ts
```diff
@@ -119,24 +119,31 @@ const createOrganizationInstance = (objectType: ObjectType, fieldValues: Values)
   )
 )
 
+export const WARNING_MESSAGE = 'Failed to fetch OrganizationSettings. Please make sure you have the correct permissions to the Organization object.'
```
benisraeladi, line 122: @tomermevorach Can you take a look at this warning message?

tomermevorach: I assume we need read permissions right, so why not say 'read permissions' instead of 'correct permissions'?
Also, we should probably specify how to reach the org permissions screen in SF

benisraeladi: I'm removing the second sentence altogether.

OrganizationSettings are not something that can be seen through the UI, because these are permissions. Not only would it be difficult to ask user to navigate to profiles/permission sets, but in the ticket attached we don't know the root cause.

tamtamirr (approved):
LGTM. Thanks! :)

packages/salesforce-adapter/src/filters/organization_wide_sharing_defaults.ts
```diff
@@ -119,24 +119,31 @@ const createOrganizationInstance = (objectType: ObjectType, fieldValues: Values)
...
-    }
-
-    const organizationInstance = createOrganizationInstance(objectType, queryResult[0])
-
-    elements.push(objectType, organizationInstance)
-  },
+  onFetch: ensureSafeFilterFetch({
+    warningMessage: WARNING_MESSAGE,
+    config,
+    filterName: 'authorInformation',
```
amitgoldarb, line 130: Edit this name

PR #4897, opened 2023-09-19
SALTO-4738 - Salesforce: update CPQ README

tomsellek:
Make the CPQ example config be the same as the default CPQ config we actually generate

---

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/cpq-sbaa-config-doc.md
+++ b/packages/salesforce-adapter/cpq-sbaa-config-doc.md
@@ -123,187 +123,46 @@ salesforce {
       ]
     }
     data = {
-        includeObjects = [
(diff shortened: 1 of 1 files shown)


adamro:
Are we sure about replacing and not having both in our OSS docs?

tomsellek:
> Are we sure about replacing and not having both in our OSS docs?

@amitgoldarb ?

amitgoldarb:
@adamro @tomsellek we can have both. I think it makes sense to place the newer internalId based one on top though and make clear this is the suggested configuration.

tomsellek:
Closing. @amitgoldarb will handle this.

PR #4898, opened 2023-09-20
SAAS-4710 Remove duplicate path from getFromPathIndex response

Uri-Bar:
SAAS-4710 Remove duplicate path from getFromPathIndex response

---
_Release Notes_: 
Core:
- fixed bug where restore/fetchFrom on element that has multiple fragments in the same file would create merge errors in nacl

---
_User Notifications_: 
None

### DIFF
--- a/packages/workspace/src/workspace/path_index.ts
+++ b/packages/workspace/src/workspace/path_index.ts
@@ -157,23 +157,39 @@ const getElementPathHints = (
 export const getElementsPathHints = (unmergedElements: Element[]):
 RemoteMapEntry<Path[]>[] => {
   const elementsByID = _.groupBy(unmergedElements, e => e.elemID.getFullName())
-  return Object.values(elementsByID)
+  const pathHints = Object.values(elementsByID)
     .flatMap(elementFragments => getElementPathHints(
       elementFragments
         .filter(element => values.isDefined(element.path))
         .map(element => ({ value: element, path: element.path as Path }))
     ))
+  const uniqueEntriesToSet = pathHints.map(entry => {
+    const uniquePaths = _.uniqWith(entry.value, _.isEqual)
+    return {
+      key: entry.key,
+      value: uniquePaths,
+    }
+  })
+  return uniqueEntriesToSet
 }
 
 export const getTopLevelPathHints = (unmergedElements: Element[]): PathHint[] => {
   const topLevelElementsWithPath = unmergedElements
     .filter(e => e.path !== undefined)
   const elementsByID = _.groupBy(topLevelElementsWithPath, e => e.elemID.getFullName())
-  return Object.entries(elementsByID)
+  const pathHints = Object.entries(elementsByID)
     .map(([key, value]) => ({
       key,
       value: value.map(e => e.path as Path),
     }))
+  const uniqueEntriesToSet = pathHints.map(entry => {
+    const uniquePaths = _.uniqWith(entry.value, _.isEqual)
+    return {
+      key: entry.key,
+      value: uniquePaths,
+    }
+  })
+  return uniqueEntriesToSet
 }
 
 export type PathIndexArgs = {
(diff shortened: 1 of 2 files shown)


packages/workspace/src/workspace/path_index.ts
```diff
@@ -157,23 +157,39 @@ const getElementPathHints = (
...
         .map(element => ({ value: element, path: element.path as Path }))
     ))
+  const uniqueEntriesToSet = pathHints.map(entry => {
+    const uniquePaths = _.uniqWith(entry.value, _.isEqual)
+    return {
+      key: entry.key,
+      value: uniquePaths,
+    }
+  })
+  return uniqueEntriesToSet
```
ori-moisis, line 173: why the separate iteration?
why not just run `uniqWith` in line 164 in the first place?

packages/workspace/src/workspace/path_index.ts
```diff
@@ -157,23 +157,39 @@ const getElementPathHints = (
...
   const topLevelElementsWithPath = unmergedElements
     .filter(e => e.path !== undefined)
   const elementsByID = _.groupBy(topLevelElementsWithPath, e => e.elemID.getFullName())
-  return Object.entries(elementsByID)
+  const pathHints = Object.entries(elementsByID)
     .map(([key, value]) => ({
       key,
       value: value.map(e => e.path as Path),
     }))
+  const uniqueEntriesToSet = pathHints.map(entry => {
```
ori-moisis, line 185: same comment as above

packages/workspace/src/workspace/path_index.ts
```diff
@@ -246,7 +262,8 @@ export const getFromPathIndex = async (
       // If this is not an exact match we want to return a single hint
       // because otherwise, splitElementByPath will make it appear in multiple fragments
       // and cause merge errors.
-      return isExactMatch ? pathHints : [pathHints[0]]
+      const uniquePathHints = _.uniqWith(pathHints, _.isEqual)
```
ori-moisis, line 265: the comment above is not related to this line, if we add this here, it should be above the comment, and maybe with a comment of its own explaining why we do this

packages/workspace/src/workspace/path_index.ts
Uri-Bar: I can, I thought it was more readable this way

packages/workspace/src/workspace/path_index.ts
```diff
@@ -173,7 +173,7 @@ export const getTopLevelPathHints = (unmergedElements: Element[]): PathHint[] =>
     .map(([key, value]) => ({
       key,
       value: value.map(e => e.path as Path),
-    }))
+    })).map(entry => ({ key: entry.key, value: _.uniqWith(entry.value, _.isEqual) }))
```
ori-moisis, line 176: can we change the line above to
```typescript
value: _.uniqWith(value.map(e => e.path as Path), _.isEqual)
```
instead of adding a separate `.map`?

packages/workspace/src/workspace/path_index.ts
```diff
@@ -162,7 +162,7 @@ RemoteMapEntry<Path[]>[] => {
       elementFragments
         .filter(element => values.isDefined(element.path))
         .map(element => ({ value: element, path: element.path as Path }))
-    ))
+    )).map(entry => ({ key: entry.key, value: _.uniqWith(entry.value, _.isEqual) }))
```
ori-moisis, line 165: My intuition here was to avoid the problem at the source rather than trying to clean it up after it happened.
in this case that would mean putting the  `uniqWith` in `getValuePathHints` (line 52) and in `getTypePathHints` (line 121) as these are the only places that could possibly put more than one path in the value.

this solution also works so I'll approve either way, just something to consider

ori-moisis (approved):

packages/workspace/src/workspace/path_index.ts
Uri-Bar: ack

PR #4905, opened 2023-09-21
SALTO-4581 - Salesforce: Fix validation of broken ref config

tomsellek:
Follow-up to https://github.com/salto-io/salto/pull/4853 where we move the validation of the nested fields to a function, since using restrictions doesn't work as expected.
---

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/types.ts
+++ b/packages/salesforce-adapter/src/types.ts
@@ -148,7 +148,7 @@ export type SaltoManagementFieldSettings = {
   defaultFieldName: string
 }
 
-const outgoingReferenceBehaviors = ['ExcludeInstance', 'BrokenReference', 'InternalId'] as const
+export const outgoingReferenceBehaviors = ['ExcludeInstance', 'BrokenReference', 'InternalId'] as const
 export type OutgoingReferenceBehavior = typeof outgoingReferenceBehaviors[number]
 
 export type BrokenOutgoingReferencesSettings = {
@@ -257,11 +257,6 @@ const brokenOutgoingReferencesSettingsType = new ObjectType({
     },
     perTargetTypeOverrides: {
       refType: new MapType(BuiltinTypes.STRING),
-      annotations: {
-        [CORE_ANNOTATIONS.RESTRICTION]: createRestriction({
-          values: outgoingReferenceBehaviors,
-        }),
-      },
     },
   },
   annotations: {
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/src/types.ts
```diff
@@ -148,7 +148,7 @@ export type SaltoManagementFieldSettings = {
   defaultFieldName: string
 }
 
-const outgoingReferenceBehaviors = ['ExcludeInstance', 'BrokenReference', 'InternalId'] as const
+export const outgoingReferenceBehaviors = ['ExcludeInstance', 'BrokenReference', 'InternalId'] as const
```
adamro, line 151: nit: If it's not used here and is used else-where wouldn't the more fitting place for it be the file named `constants`? 🤔

packages/salesforce-adapter/src/types.ts
```diff
@@ -257,11 +257,6 @@ const brokenOutgoingReferencesSettingsType = new ObjectType({
     },
     perTargetTypeOverrides: {
       refType: new MapType(BuiltinTypes.STRING),
-      annotations: {
```
adamro, line 260: How come we are deleting this? What did it do not as expected?

adamro (approved):
Approving with a question of why are we doing this :)
If you are sure this is the right thing to do you can merge

packages/salesforce-adapter/src/types.ts
tomsellek: When I set up the following config:
```
      brokenOutgoingReferencesSettings = {
        defaultBehavior = "BrokenReference"
        perTargetTypeOverrides = {
            User = "InternalId"
        }
      }
```
I got the error
```
Error validating "salesforce._config.instance._config.fetch.data.brokenOutgoingReferencesSettings.perTargetTypeOverrides": Value "{"User":"InternalId"}" is not valid for field perTargetTypeOverrides expected one of: "ExcludeInstance", "BrokenReference", "InternalId"
```

so looks like the constraint acted on the whole field, not on the values in the map.

packages/salesforce-adapter/src/types.ts
tomsellek: It's used in the line below, though

adamro: 👍🏻 my bad

packages/salesforce-adapter/src/types.ts
adamro: I wonder how it works [here](https://github.com/salto-io/salto/blob/191d8de111a8aa2f71109155367ccb3b55a69bd0/packages/salesforce-adapter/src/types.ts#L572) and not in this case
Anyway you can merge this as a fix and we can try to check this out after

PR #4908, opened 2023-09-22
SALTO-4596 - Salesforce: Add a config option to omit certain data instance fields

tomsellek:
Add a configuration value to allow the user to omit specific fields.

---
See question in the PRs body

There are several places where the omission can take place. I chose to remove the omitted fields from the query we send to Salesforce. I could have modified `getQueryableFields` (the earliest possible moment) or deleted the fields after we convert the data records to `InstanceElement`s. Please let me know if you think the logic is better placed somewhere else.

---
_Release Notes_: 
Salesforce: added the 'omittedFields' data configuration.

---
_User Notifications_: 
N/A


packages/salesforce-adapter/src/fetch_profile/data_management.ts
```diff
@@ -36,6 +36,7 @@ export type DataManagement = {
   getObjectAliasFields: (name: string) => types.NonEmptyArray<string>
   showReadOnlyValues?: boolean
   managedBySaltoFieldForType: (objType: ObjectType) => string | undefined
+  omittedFieldsForType: (name: string | undefined) => string[]
```
tamtamirr, line 39: Why is the name possibly undefined here? This is also not aligned with the rest of the methods (e.g. getObjecgtAliasFields)

packages/salesforce-adapter/src/fetch_profile/data_management.ts
```diff
@@ -163,6 +164,15 @@ export const buildDataManagement = (params: DataManagementConfig): DataManagemen
...
     showReadOnlyValues: params.showReadOnlyValues,
+    omittedFieldsForType: name => {
+      if (name === undefined) {
+        return []
+      }
+      const typePrefix = `${name}.`
+      return (params.omittedFields ?? [])
+        .filter(fieldName => fieldName.startsWith(typePrefix))
+        .map(fieldName => fieldName.slice(typePrefix.length))
+    },
```
tamtamirr, line 175: I think we can simplify the logic here. I suggest we **group** the omitted fields by type once and then simply return `makeArray(omittedFieldsByType[typeName])`

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
```diff
@@ -111,11 +113,13 @@ const getRecords = async (
...
   ids?: string[],
   managedBySaltoField?: string,
+  omittedFields: string[] = [],
 ): Promise<RecordById> => {
   const queryableFields = getQueryableFields(type)
+    .filter(field => !omittedFields.includes(apiNameSync(field) ?? ''))
   const typeName = await apiName(type)
   if (_.isEmpty(queryableFields)) {
-    log.debug(`Type ${typeName} had no queryable fields`)
+    log.debug('Type %s had no queryable fields (%d fields omitted)', typeName, omittedFields.length)
```
tamtamirr, line 122:
```suggestion
    log.debug('Type %s had no queryable fields or all of its queryable fields were omitted. %o', typeName, {
      queryableFields: queryableFieldNames,
      omittedFields,
    })
```
I think we miss critical info here that will help us debug edge-cases.

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
```diff
@@ -383,7 +387,10 @@ const getReferencedRecords = async (
     const typeToMissingIds = getMissingReferencedIds(currentLevelRecords)
     const newReferencedRecords = await mapValuesAsync(
       typeToMissingIds,
-      (ids, typeName) => getRecords(client, customObjectFetchSetting[typeName].objectType, ids)
+      (ids, typeName) => {
+        const fetchSettings = customObjectFetchSetting[typeName]
+        return getRecords(client, fetchSettings.objectType, ids, undefined, fetchSettings.omittedFields)
```
tamtamirr, line 392: Two things I suggest here (some we discussed already):
1. Make getRecords receive a single object of params.
2. Instead of receiving both `fetchSettings.objectType` and `fetchSettings.omittedFields` receive the `fetchSettings` themselves.

packages/salesforce-adapter/test/filters/custom_objects_instances.test.ts
```diff
@@ -252,6 +253,7 @@ describe('Custom Object Instances filter', () => {
         saltoManagementFieldSettings: {
           defaultFieldName: MANAGED_BY_SALTO_FIELD_NAME,
         },
+        ...(omittedFields ? { omittedFields } : {}),
```
tamtamirr, line 256:
```suggestion
       omittedFields,
```
Shouldn't it be like this..? I'm rather confused.

packages/salesforce-adapter/test/filters/custom_objects_instances.test.ts
```diff
@@ -234,6 +234,7 @@ describe('Custom Object Instances filter', () => {
   }
   const buildTestFetchProfile = (
     types: TestFetchProfileParams[],
+    omittedFields: string[] = [],
   ): FetchProfile => {
```
tamtamirr, line 238: nit: Can we make this method receive a params object? This will only do good for future changes, and make the code clearer.

tamtamirr (commented):
Looking good! Added some comments :) Also, let's update the docs here explaining the feature and the expected format of the fields (their `apiName`)

packages/salesforce-adapter/src/types.ts
```diff
@@ -526,6 +527,9 @@ const dataManagementType = new ObjectType({
     brokenOutgoingReferencesSettings: {
       refType: brokenOutgoingReferencesSettingsType,
     },
+    omittedFields: {
```
tamtamirr, line 530: Can we add a config validation that all the strings here are in the correct format of a field `apiName`?

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
tomsellek: Ack

packages/salesforce-adapter/src/types.ts
tomsellek: Ack

packages/salesforce-adapter/test/filters/custom_objects_instances.test.ts
tomsellek: If the array `omittedFields` is not empty then we want an object with a single entry with the key `"omittedFields"` and the value of the array `omittedFields`. But if the array is empty we don't want the object at all.

In other words, we want either
```
{
  ...
  omittedFields: [ ... ],
}
```
or
```
{
  ...
}
```
but never
```
{
  ...
  omittedFields: [],
}
```

packages/salesforce-adapter/src/fetch_profile/data_management.ts
tamtamirr: I would expect we handle undefined **apiNames** outside and not within this method. This smells IMO

packages/salesforce-adapter/test/filters/custom_objects_instances.test.ts
tamtamirr: Why can't we have an empty array of **omittedFields** though?

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
```diff
@@ -103,18 +105,33 @@ const buildQueryStrings = async (
   return buildSelectQueries(typeName, fieldNames, queryConditions)
 }
 
+type GetRecordsParams = {
+  client: SalesforceClient
+  typeAndFetchSettings: CustomObjectFetchSetting
```
tamtamirr, line 110:
```suggestion
  settings: CustomObjectFetchSetting
```
I'd simplify the name as it's quite confusing

packages/salesforce-adapter/src/fetch_profile/data_management.ts
```diff
@@ -204,4 +211,9 @@ export const validateDataManagementConfig = (
       }
     )
   }
+  const invalidOmittedFieldNames = makeArray(dataManagementConfig.omittedFields)
+    .filter(omittedFieldName => !omittedFieldName.match(/\w+(?:\.\w+)/))
+  if (invalidOmittedFieldNames.length > 0) {
+    throw new ConfigValidationError([...fieldPath, 'omittedFields'], `The following entries are not valid Salesforce API names: ${invalidOmittedFieldNames.join(',')}`)
```
tamtamirr, line 217:
```suggestion
    throw new ConfigValidationError([...fieldPath, 'omittedFields'], `The following omitted fields API names are invalid: ${invalidOmittedFieldNames.join(',')}`)
```

tamtamirr (commented):
Looking good! Added some small comments

packages/salesforce-adapter/src/fetch_profile/data_management.ts
tomsellek: Ugh. Yeah, but everywhere else just uses `await apiName` and ignores the possibility of the API name being undefined.

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
tomsellek: The reason I chose this name is because it seemed weird that the function doesn't take the type to fetch as a parameter. So I wanted to make that more explicit. Do you have an idea on how I can make it clear from the function's signature that it receives the type as a parameter?

packages/salesforce-adapter/test/filters/custom_objects_instances.test.ts
tomsellek: I just didn't want it in the test

tamtamirr (approved):
LGTM 👑

PR #4909, opened 2023-09-26
SALTO-2349: Moved change_config.ts to adapterComponents

idoamit8:
In order for Jira and zendesk to use the same code that chagne their config by the suggest config changes. I moved change_config to adapter components. 

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/adapter-components/src/config/config_change.ts
```diff
@@ -14,26 +14,27 @@
...
+import { ConfigChangeSuggestion } from './shared'
 
-const { makeArray } = collections.array
 
-const STOP_MANAGING_ITEMS_MSG = 'Salto failed to fetch some items from Zendesk.'
-  + ' Failed items must be excluded from the fetch.'
+const { makeArray } = collections.array
+const FETCH_CONFIG = 'fetch'
 
 export const getConfigFromConfigChanges = (
```
netama, line 26: 1. nit: let's use named parameters? since the number of arguments grew
2. this only supports a specific type of config change, let's clarify in the function name? e.g. something like `getConfigWithExcludeFromConfigChanges`?
3. please add a docstring :) explaining when this is relevant and what it doesn't cover

packages/adapter-components/src/config/shared.ts
```diff
@@ -26,6 +26,10 @@ export const DEPLOYER_FALLBACK_VALUE = '##DEPLOYER##'
 
 export class InvalidSingletonType extends Error {}
 
+export type ConfigChangeSuggestion = {
```
netama, line 29: why did this need to move? technically it doesn't fit here, since it is not related to the user-facing config. so if this moves IMO it should be to the new `config_change.ts` file

packages/adapter-components/test/config/config_change.test.ts
```diff
@@ -0,0 +1,55 @@
...
+            { type: 'Type1' },
+          ],
+        },
+      }
+    )
+  })
+  it('should return undefined when no changes are suggested', () => {
+    expect(getConfigFromConfigChanges([], config, configType, 'Defualt')).toBeUndefined()
+  })
+  it('should return new config when changes are suggested amd a message', () => {
```
netama, line 43: nit: `amd` -> `and`

netama (approved):
added some comments, pre-approving for once they're addressed

packages/adapter-components/src/config/shared.ts
idoamit8: It moved in order to not create circular dependencies between config and elements. Will move to config_change.

PR #4912, opened 2023-09-26
SALTO-4813 - Salesforce: only show relevant instances when warning about broken refs

tomsellek:
The warning message about dropped instances due to broken refs is aggregated by type - we display a separate warning for each type that contains broken refs. But when we list the actual instances that were dropped we list all of them, not just the instances of the specific type we're warning about.

---

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A


adamro (approved):

adamro:
Can we add a test that would fail before this fix please?

tomsellek:
> Can we add a test that would fail before this fix please?

So, I tried to do this, but it turned into a huge mess - the test would involve an uncomfortable amount of string manipulation and regex wrangling but even worse we'll need to add a whole new type + instances to the existing tests. It's going to be a significant increase in complexity just to test an error message.
I will keep the WIP test code around in case you feel it's worth the added complexity, but for now I'll merge without a UT.

PR #4913, opened 2023-09-26
SALTO-4814 - Salesforce: Consider metadata instances as potential lookup targets

tomsellek:
In the package `Field Service Lightning` we have custom objects instances that reference instances of metadata types. Our current code complains of invalid references, because we only look up reference targets among instances of custom objects.
So let's make a distinction between potential reference sources (only custom object instances) and potential reference targets (all instances).

---

~This relies on instances of metadata types having an `Id` field. I haven't tested this yet.~
Please note the code that gets the internal ID for both custom objects and metadata instances.

Tests:
 - Fetch and observe a reference created in the `SkillId` field of an instance of `ServiceResourceSkill`
 - Deploy an instance that contains a reference to a metadata instance

---
_Release Notes_: 
Salesforce: Lookup fields in records of custom objects that reference instances of metadata types are now handled correctly.

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/filters/custom_object_instances_references.ts
+++ b/packages/salesforce-adapter/src/filters/custom_object_instances_references.ts
@@ -37,6 +37,7 @@ import {
   CORE_ANNOTATIONS,
   isObjectType,
   ObjectType,
+  isInstanceElement,
 } from '@salto-io/adapter-api'
 import { FilterResult, RemoteFilterCreator } from '../filter'
 import {
@@ -329,9 +330,9 @@ const filter: RemoteFilterCreator = ({ client, config }) => ({
     if (dataManagement === undefined) {
       return {}
     }
-    const customObjectInstances = await awu(elements).filter(isInstanceOfCustomObject)
-      .toArray() as InstanceElement[]
-    const internalToInstance = await keyByAsync(customObjectInstances, serializeInstanceInternalID)
+    const allInstances = elements.filter(isInstanceElement)
+    const customObjectInstances = await awu(allInstances).filter(isInstanceOfCustomObject).toArray()
+    const internalToInstance = await keyByAsync(allInstances, serializeInstanceInternalID)
     const internalIdPrefixToType = await buildCustomObjectPrefixKeyMap(elements)
     const { reverseReferencesMap, missingRefs } = await replaceLookupsWithRefsAndCreateRefMap(
       customObjectInstances,
--- a/packages/salesforce-adapter/src/filters/custom_object_instances_references.ts
+++ b/packages/salesforce-adapter/src/filters/custom_object_instances_references.ts
@@ -45,12 +45,15 @@ import {
   isInstanceOfCustomObject,
 } from '../transformers/transformer'
 import {
+  CUSTOM_OBJECT_ID_FIELD,
   FIELD_ANNOTATIONS,
   KEY_PREFIX,
   KEY_PREFIX_LENGTH,
   SALESFORCE,
 } from '../constants'
 import {
+  getInternalId,
+  isCustomObjectSync,
   isReadOnlyField,
   isReferenceField,
   referenceFieldTargetTypes,
@@ -79,8 +82,12 @@ const MAX_BREAKDOWN_ELEMENTS = 10
 const serializeInternalID = (typeName: string, id: string): string =>
   (`${typeName}${INTERNAL_ID_SEPARATOR}${id}`)
 
+const instanceInternalId = (instance: InstanceElement): string => (
+  isCustomObjectSync(instance.getTypeSync()) ? instance.value[CUSTOM_OBJECT_ID_FIELD] : getInternalId(instance)
+)
+
 const serializeInstanceInternalID = async (instance: InstanceElement): Promise<string> => (
-  serializeInternalID(await apiName(await instance.getType(), true), await apiName(instance))
+  serializeInternalID(await apiName(await instance.getType(), true), instanceInternalId(instance))
 )
 
 const deserializeInternalID = (internalID: string): RefOrigin => {
--- a/packages/salesforce-adapter/src/filters/custom_object_instances_references.ts
+++ b/packages/salesforce-adapter/src/filters/custom_object_instances_references.ts
@@ -45,15 +45,13 @@ import {
   isInstanceOfCustomObject,
 } from '../transformers/transformer'
 import {
-  CUSTOM_OBJECT_ID_FIELD,
   FIELD_ANNOTATIONS,
   KEY_PREFIX,
   KEY_PREFIX_LENGTH,
   SALESFORCE,
 } from '../constants'
 import {
-  getInternalId,
-  isCustomObjectSync,
+  instanceInternalId,
   isReadOnlyField,
   isReferenceField,
   referenceFieldTargetTypes,
@@ -81,10 +79,6 @@ const MAX_BREAKDOWN_ELEMENTS = 10
 
 const serializeInternalID = (typeName: string, id: string): string =>
   (`${typeName}${INTERNAL_ID_SEPARATOR}${id}`)
-
-const instanceInternalId = (instance: InstanceElement): string => (
-  isCustomObjectSync(instance.getTypeSync()) ? instance.value[CUSTOM_OBJECT_ID_FIELD] : getInternalId(instance)
-)
 
 const serializeInstanceInternalID = async (instance: InstanceElement): Promise<string> => (
   serializeInternalID(await apiName(await instance.getType(), true), instanceInternalId(instance))
--- a/packages/salesforce-adapter/src/transformers/reference_mapping.ts
+++ b/packages/salesforce-adapter/src/transformers/reference_mapping.ts
@@ -98,7 +98,7 @@ const safeApiName = ({ ref, path, relative }: {
 }
 
 type ReferenceSerializationStrategyName = 'absoluteApiName' | 'relativeApiName' | 'configurationAttributeMapping' | 'lookupQueryMapping' | 'scheduleConstraintFieldMapping'
- | 'mapKey' | 'customLabel' | 'referenceToInstance'
+ | 'mapKey' | 'customLabel' | 'fromDataInstance'
 const ReferenceSerializationStrategyLookup: Record<
   ReferenceSerializationStrategyName, ReferenceSerializationStrategy
 > = {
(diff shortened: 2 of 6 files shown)


packages/salesforce-adapter/src/filters/custom_object_instances_references.ts
```diff
@@ -333,9 +334,9 @@ const filter: RemoteFilterCreator = ({ client, config }) => ({
     if (dataManagement === undefined) {
       return {}
     }
-    const customObjectInstances = await awu(elements).filter(isInstanceOfCustomObject)
-      .toArray() as InstanceElement[]
-    const internalToInstance = await keyByAsync(customObjectInstances, serializeInstanceInternalID)
+    const allInstances = elements.filter(isInstanceElement)
+    const customObjectInstances = await awu(allInstances).filter(isInstanceOfCustomObject).toArray()
+    const internalToInstance = await keyByAsync(allInstances, serializeInstanceInternalID)
```
adamro, line 339: Your assumption about the `Id` field doesn't hold for other Instances.

What are the actual values in these Instances that are supposed to be references to Metadata Instances?

My guess is that it's the `internalId` that some of them have. If that's true, we need to use that and not `Id` for them and I guess make sure it actually exists etc.? (the `internalID` name is misleading, it actually uses apiName that for CustomObject Instances is the Id value, we need to use `InternalId` for others I guess in this case)

Also, did you make sure the whole `internalToTypeName` code is still legit with this? (cause it's specific for CustomObjects obviously)

tomsellek: Ugh. You're right. We use `apiName` as a proxy for the ID, but `apiName` returns `fullName` for metadata instances, which is not what's in the referring field.

adamro (commented):
Can you please add a test for this case?
Also, we might wanna cover this in the e2e as well (not a blocker for this PR but maybe open a ticket or write a comment in a relevant one)
Another question - did you test this manually e2e, including deploy etc?

packages/salesforce-adapter/src/transformers/reference_mapping.ts
```diff
@@ -880,6 +888,16 @@ const getLookUpNameImpl = ({ defs, resolveToElementFallback }: {
...
+  const determineDefaultLookupStrategy = ({ ref }: GetLookupNameFuncArgs): ReferenceSerializationStrategy => {
+    if (isInstanceElement(ref.value) && !isCustomObjectSync(ref.value.getTypeSync())) {
+      // Unlike custom object instances, the apiName of metadata instances is not their Internal ID but rather their
+      // actual API name. But the references to metadata instances use their internal ID - so when we serialize a
+      // ReferenceExpression back into a field value we need to use a different serialization strategy.
+      return ReferenceSerializationStrategyLookup.internalId
+    }
+    return ReferenceSerializationStrategyLookup.absoluteApiName
+  }
+
```
tomsellek, line 900: I'm not sure if that's the right way to do this, but I can't think of a better way. Any thoughts?

tomsellek:
> Can you please add a test for this case? Also, we might wanna cover this in the e2e as well (not a blocker for this PR but maybe open a ticket or write a comment in a relevant one) Another question - did you test this manually e2e, including deploy etc?

1. https://salto-io.atlassian.net/browse/SALTO-4893
2. Ugh^2. When I tested deploy I discovered we need an additional change in how we serialize the reference expression during deploy. Please review.

packages/salesforce-adapter/src/transformers/reference_mapping.ts
```diff
@@ -880,6 +888,17 @@ const getLookUpNameImpl = ({ defs, resolveToElementFallback }: {
...
+    if (isInstanceElement(element) && isCustomObjectSync(element.getTypeSync())
+      && isInstanceElement(ref.value) && !isCustomObjectSync(ref.value.getTypeSync())) {
+      // Unlike custom object instances, the apiName of metadata instances is not their Internal ID but rather their
+      // actual API name. But the references to metadata instances use their internal ID - so when we serialize a
+      // ReferenceExpression back into a field value we need to use a different serialization strategy.
+      return ReferenceSerializationStrategyLookup.internalId
+    }
+    return ReferenceSerializationStrategyLookup.absoluteApiName
+  }
+
```
tomsellek, line 901: I'm not sure if that's the right way to do this, but I can't think of a better way. Any thoughts?

adamro: Something here feels wrong to me. Maybe even bigger than just this.
It will probably be easier to discuss f2d than in writing tho.

tomsellek: As discussed offline, I will add a `preDeploy`+`onDeploy' to the custom objects references filter to handle the serialization of the references it creates rather than have it rely on the reference mapping code.

packages/salesforce-adapter/src/transformers/reference_mapping.ts
```diff
@@ -913,8 +919,13 @@ const getLookUpNameImpl = ({ defs, resolveToElementFallback }: {
...
-export const getLookupNameWithFallbackToElement = getLookUpNameImpl({
+export const getLookUpName = getLookUpNameImpl({
+  defs: fieldNameToTypeMappingDefs,
+  resolveToElementFallback: false,
+  defaultStrategyName: 'absoluteApiName',
+})
+export const getLookupNameForDataInstances = getLookUpNameImpl({
   defs: fieldNameToTypeMappingDefs,
   resolveToElementFallback: true,
+  defaultStrategyName: 'internalId',
```
adamro, line 930: Are you sure this is right as a default for 100% of the references going out of Data Instances? If we had cases that fell to the default from Data Instances, before, they were serialized with `absoluteApiName`, which works both for the cases the target is a Data Instance and if it's a CustomObject/Field. With the new default it won't work for the 2nd case. The 2nd case doesn't exist now? 🤔

packages/salesforce-adapter/src/filters/custom_object_instances_references.ts
adamro, line 1: Did you end up confirming that all the references that can exist from Data Instances to Metadata Instances will indeed do it by InternalId and not a different method (ie. apiName etc.). Cause if those cases can also exist this code won't work for them. 
If we can't confirm it through documentation or another method. We might wanna at least add a log for the cases of "misses" which gives us as much information as possible of when it happens

tomsellek: I have not been able to even create a lookup relationship to a metadata instance. The Salesforce UI seems to only allow lookups to custom object instances.

packages/salesforce-adapter/src/transformers/reference_mapping.ts
tomsellek: I'm not sure I understand the question, but I'll still try to answer :)
1. Any reference we create in `custom_object_instances_references` can only be via the internal ID (either via the `internalID` or the `ID` field), so these references should be handled correctly
2. If the reference is defined explicitly in the reference mapping, then whatever serialization strategy is defined will be used and this change won't make any difference
3. If the reference is to a custom object, then the API name is the same as the ID, so there's no difference

So I'm not sure what the potential issue is. Can you maybe give an example?

packages/salesforce-adapter/src/filters/custom_object_instances_references.ts
adamro: So how are you testing this? With the Installed Package that had this?

Anyway, even if you don't know if this case exists. I think we need to handle the case the values doesn't fit any of the InternalId of the instances of that Type and log it to have info of it.

adamro, line 1: Another unrelated thought. We need to take into account Partial/Quick fetch here. 
When it was just CustomObject Instances to CustomObject Instances, we assumed we always fetch all of them cause that's the config we have for Partial Fetch. Now that you can reference other stuff you need to handle the case these Instances exist in the ElementsSource and not in the Elements.

packages/salesforce-adapter/src/transformers/reference_mapping.ts
adamro: The references defined in `custom_object_instances_references` are not the only outgoing references defined in filter from CustomObject Instances.
For example - `filters/cpq/lookup_fields.ts`

There, you have definition of references to CustomObject / CustomFields (which are ObjectType/Field)

So, first thing I am worried about, is that this new strategy you defined as `internalId` only handles Instances properly. 
And possibly can cause a degradation in the default handling of the cases from the filter above.
If the references to CustomObject / CustomField were serialized, with the default that existed before, using the `absoluteApiName`, that works for both CustomObject Instances and for CustomObject/Fields (cause references to them is by apiName). With your code, it will still work for references to CustomObject Instances (cause Id==apiName there) but it won't work for CustomObject/Fields that need an apiName based strategy,

Makes sense?

packages/salesforce-adapter/src/filters/custom_object_instances_references.ts
tomsellek: Yes

packages/salesforce-adapter/src/transformers/reference_mapping.ts
tomsellek: @adamro @tamtamirr makes sense. But this seems like another reason to have each filter do/undo their own reference mapping?

tomsellek: Per our offline discussion I will create a dedicated serialization strategy

packages/salesforce-adapter/src/transformers/reference_mapping.ts
```diff
@@ -155,6 +156,14 @@ const ReferenceSerializationStrategyLookup: Record<
       return val
     },
   },
+  referenceToInstance: {
```
adamro, line 159: Why are you calling this `toInstance`? 🤔 
It can also be `toCustomObject`/`toField`

Why not `fromDataInstance`? `fromInstanceOfCustomObject` or anything else that describes to from, which is what is really the logic that leads to using this atm

tomsellek: 🤷

adamro (approved):

packages/salesforce-adapter/src/filters/custom_object_instances_references.ts
adamro: > Another unrelated thought. We need to take into account Partial/Quick fetch here. When it was just CustomObject Instances to CustomObject Instances, we assumed we always fetch all of them cause that's the config we have for Partial Fetch. Now that you can reference other stuff you need to handle the case these Instances exist in the ElementsSource and not in the Elements.

@tomsellek I forgot we never followed up on this. This is still an issue.

tomsellek: https://salto-io.atlassian.net/browse/SALTO-5072

PR #4915, opened 2023-09-27
SALTO-4767: add highlighted in annotation of important values

edenhassid:
add highlighted in annotation of important values
fix `getImportantValues` to return an array and to test values that are returned
add only last value of path when indexed

---

---
_Release Notes_: 
core:
* add highlighted field in annotation of important values

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-api/src/constants.ts
+++ b/packages/adapter-api/src/constants.ts
@@ -50,5 +50,5 @@ export const BUILTIN_TYPE_NAMES = {
   DEPENDENCY: 'dependency',
   HIDDEN_STRING: 'hidden_string',
   HIDDEN_BOOLEAN: 'hidden_boolean',
-  IMPORTANT_VALUE: 'importantValue',
+  IMPORTANT_VALUE: 'important_value',
 }
--- a/packages/adapter-utils/src/important_values.ts
+++ b/packages/adapter-utils/src/important_values.ts
@@ -19,7 +19,7 @@ import {
   Element,
   isField,
   isInstanceElement,
-  isObjectType,
+  isObjectType, isReferenceExpression,
   ReadOnlyElementsSource, Values,
 } from '@salto-io/adapter-api'
 import { values } from '@salto-io/lowerdash'
@@ -39,7 +39,11 @@ const isValidIndexedValueData = (ImportantValue: ImportantValue, valueData: unkn
     return valueData.every(part => _.isString(part) || _.isNumber(part))
   }
 
-  return _.isString(valueData) || _.isNumber(valueData) || _.isBoolean(valueData) || valueData === undefined
+  return _.isString(valueData)
+    || _.isNumber(valueData)
+    || _.isBoolean(valueData)
+    || valueData === undefined
+    || isReferenceExpression(valueData)
 }
 
 
--- a/packages/adapter-utils/src/important_values.ts
+++ b/packages/adapter-utils/src/important_values.ts
@@ -22,12 +22,25 @@ import {
   isObjectType,
   ReadOnlyElementsSource, Values,
 } from '@salto-io/adapter-api'
+import { values } from '@salto-io/lowerdash'
 import _ from 'lodash'
 
 const log = logger(module)
+const { isDefined } = values
 
 type ImportantValue = { value: string; indexed: boolean; highlighted: boolean }
 type ImportantValues = ImportantValue[]
+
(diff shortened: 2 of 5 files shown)


packages/adapter-api/src/constants.ts
```diff
@@ -50,5 +50,5 @@ export const BUILTIN_TYPE_NAMES = {
   DEPENDENCY: 'dependency',
   HIDDEN_STRING: 'hidden_string',
   HIDDEN_BOOLEAN: 'hidden_boolean',
-  IMPORTANT_VALUE: 'importantValue',
+  IMPORTANT_VALUE: 'important_value',
```
netama, line 53: :+1:

packages/adapter-utils/test/important_values.test.ts
```diff
@@ -59,28 +68,34 @@ const obj = new ObjectType({
...
     [CORE_ANNOTATIONS.SELF_IMPORTANT_VALUES]: [
       {
         value: 'name',
         indexed: false,
+        highlighted: true,
       },
       {
         value: 'apiName',
         indexed: true,
+        highlighted: false,
```
netama, line 93: seems both booleans are optional (and also `value` - I think it should be required?) - if this is intentional, let's also cover cases where they're not specified (and I assume treated as falsy)?

packages/adapter-utils/src/important_values.ts
```diff
@@ -13,42 +13,90 @@
...
     return {}
   }
-  const relevantImportantValues = getRelevantImportantValues(importantValues, indexedOnly)
+  const relevantImportantValues = getRelevantImportantValues(importantValues, indexedOnly, highlightedOnly)
   const getFrom = isInstanceElement(element) ? element.value : element.annotations
-  const finalImportantValues = _.mapValues(
-    _.keyBy(relevantImportantValues, obj => obj.value),
-    obj => _.get(getFrom, obj.value, undefined),
-  )
+  const finalImportantValues = relevantImportantValues.map(ImportantValue => {
```
netama, line 87: nit: 
```suggestion
  const finalImportantValues = relevantImportantValues.map(importantValue => {
```

packages/adapter-utils/src/important_values.ts
```diff
@@ -13,42 +13,90 @@
...
+    const { value } = ImportantValue
+    const valueData = _.get(getFrom, value, undefined)
+    if (!isValidIndexedValueData(ImportantValue, valueData)) {
+      log.warn(`${ImportantValue.value} for element ${element.elemID.getFullName()} is not a primitive value,
+      we do not support non primitive values as indexed important values`)
+      return undefined
+    }
+    const valueSplit = value.split('.')
+    const finalValue = indexedOnly === true ? valueSplit[valueSplit.length - 1] : value
```
netama, line 96: nit: can use `valueSplit.pop()`

packages/adapter-utils/src/important_values.ts
```diff
@@ -13,42 +13,90 @@
...
-type ImportantValue = { value: string; indexed: boolean }
+const log = logger(module)
+const { isDefined } = values
+
+type ImportantValue = { value: string; indexed: boolean; highlighted: boolean }
 type ImportantValues = ImportantValue[]
 
-const getRelevantImportantValues = (importantValues: ImportantValues, indexedOnly?: boolean): ImportantValues => (
-  indexedOnly === true
+const isValidIndexedValueData = (ImportantValue: ImportantValue, valueData: unknown): boolean => {
```
netama, line 34: nit: 
```suggestion
const isValidIndexedValueData = (importantValue: ImportantValue, valueData: unknown): boolean => {
```

packages/adapter-utils/src/important_values.ts
```diff
@@ -13,42 +13,90 @@
...
+  if (ImportantValue.indexed !== true) {
+    return true
+  }
+  if (_.isArray(valueData)) {
+    return valueData.every(part => _.isString(part) || _.isNumber(part))
```
netama, line 39: 1. do we intentionally not support booleans or reference expressions in arrays?
2. can we reuse more of the checks between the array and the single value? seems the main difference is the `undefined` case, which we can probably either handle separately or just allow (probably won't cause issues)?
3. nit: I see we have some existing functions such as `isPrimitiveValue`, can we reuse them for the cases they cover?

netama (approved):
cool :) added some comments but looks good so pre-approving

PR #4916, opened 2023-09-27
SALTO-4817: Return reference to project category

IdoZakk:
We used to return them based on id, but we do not have an id on addition.
Changed to key, added tests and some coverage

---

---
_Release Notes_: 
Jira Adapter:
* Fixed a bug that caused project category not to be translated into a reference after project addition
---
_User Notifications_: 
None


shayc331 (approved):
LGTM

PR #4922, opened 2023-09-28
SALTO-2349: Fetch and deploy Forms 

idoamit8:
In this PR I started to support fetch of forms. 
I also added support for deploy to check that the fetch and deploy is multienv friendly. 

---

Please Review from [initial forms fetch support](https://github.com/salto-io/salto/commit/81a7768780bb2beaea8ead9f3a99ff950c17b558) and any sub commit. 

---
_Release Notes_: 
_Jira adapter_:
Now supporting fetch  and deploy operations for Form.

---
_User Notifications_: 
_JIra Adapter_:
You can now fetch and deploy Forms in JSM.

### DIFF
--- a/packages/jira-adapter/src/filters/forms/forms.ts
+++ b/packages/jira-adapter/src/filters/forms/forms.ts
@@ -17,17 +17,18 @@
 import { InstanceElement, isInstanceElement } from '@salto-io/adapter-api'
 import { values as lowerDashValues } from '@salto-io/lowerdash'
 import { elements as adapterElements } from '@salto-io/adapter-components'
-import { createSchemeGuard, naclCase, pathNaclCase } from '@salto-io/adapter-utils'
+import { naclCase, pathNaclCase } from '@salto-io/adapter-utils'
 import { FilterCreator } from '../../filter'
-import { JIRA, PROJECT_TYPE } from '../../constants'
+import { JIRA, PROJECT_TYPE, SERVICE_DESK } from '../../constants'
 import { getCloudId } from '../automation/cloud_id'
-import { DETAILED_FORM_RESPONSE_SCHEME, FORMS_RESPONSE_SCHEME, createFormType, detailedFormResponse, formsResponse } from './forms_types'
+import { createFormType, isDetailedFormsResponse, isFormsResponse } from './forms_types'
 
 const { isDefined } = lowerDashValues
 
-const isFormsResponse = createSchemeGuard<formsResponse>(FORMS_RESPONSE_SCHEME)
-const isDetailedFormsResponse = createSchemeGuard<detailedFormResponse>(DETAILED_FORM_RESPONSE_SCHEME)
-
+/*
+This filter fetches all forms from Jira Service Management and creates an instance element for each form.
+We use filter because we need to use cloudId which is not available in the infrastructure.
+*/
 const filter: FilterCreator = ({ config, client, getElemIdFunc }) => ({
   name: 'formsFilter',
   onFetch: async elements => {
@@ -39,19 +40,25 @@ const filter: FilterCreator = ({ config, client, getElemIdFunc }) => ({
     elements.push(formType)
     subTypes.forEach(subType => { elements.push(subType) })
 
-    const jsmProject = elements
+    const jsmProjects = elements
       .filter(isInstanceElement)
-      .filter(e => e.elemID.typeName === PROJECT_TYPE)
-      .filter(e => e.value.projectTypeKey === 'service_desk')
+      .filter(instnace => instnace.elemID.typeName === PROJECT_TYPE)
+      .filter(project => project.value.projectTypeKey === SERVICE_DESK)
 
-    const forms = (await Promise.all(jsmProject.flatMap(async project => {
-      const url = `/gateway/api/proforma/cloudid/${cloudId}/api/1/projects/${project.value.id}/forms`
-      const res = await client.getSinglePage({ url })
-      if (isFormsResponse(res)) {
-        return Promise.all(res.data.map(async formResponse => {
-          const detailedUrl = `/gateway/api/proforma/cloudid/${cloudId}/api/2/projects/${project.value.id}/forms/${formResponse.id}`
-          const detailedRes = await client.getSinglePage({ url: detailedUrl })
-          if (isDetailedFormsResponse(detailedRes)) {
+    const forms = (await Promise.all(jsmProjects
+      .flatMap(async project => {
+        const url = `/gateway/api/proforma/cloudid/${cloudId}/api/1/projects/${project.value.id}/forms`
+        const res = await client.getSinglePage({ url })
+        if (!isFormsResponse(res)) {
+          return undefined
+        }
+        return Promise.all(res.data
+          .map(async formResponse => {
+            const detailedUrl = `/gateway/api/proforma/cloudid/${cloudId}/api/2/projects/${project.value.id}/forms/${formResponse.id}`
+            const detailedRes = await client.getSinglePage({ url: detailedUrl })
+            if (!isDetailedFormsResponse(detailedRes)) {
+              return undefined
+            }
             const name = `${project.value.name}_${formResponse.name}`
             delete detailedRes.data?.publish
             delete detailedRes.data.design.settings.templateFormUuid
@@ -70,12 +77,10 @@ const filter: FilterCreator = ({ config, client, getElemIdFunc }) => ({
               formValue,
               [...prefixPath.slice(0, -1), 'forms', pathNaclCase(instanceName)],
             )
-          }
-          return undefined
-        }))
-      }
-      return undefined
-    }))).flat().filter(isDefined)
+          }))
+      })))
+      .flat()
+      .filter(isDefined)
     forms.forEach(form => elements.push(form))
   },
 })
--- a/packages/jira-adapter/src/filters/forms/forms.ts
+++ b/packages/jira-adapter/src/filters/forms/forms.ts
@@ -49,7 +49,7 @@ const deployForms = async (
         },
       })
       if (!isCreateFormResponse(resp.data)) {
-        return
+        throw new Error('Failed to create form')
       }
       form.value.id = resp.data.id
       form.value.design.settings.templateId = resp.data.id
@@ -125,6 +125,18 @@ const filter: FilterCreator = ({ config, client, getElemIdFunc }) => ({
       .filter(isDefined)
     forms.forEach(form => elements.push(form))
   },
+  preDeploy: async changes => {
+    const fieldNames = ['questions', 'sections', 'conditions']
+    changes
+      .filter(isInstanceChange)
+      .map(change => getChangeData(change))
+      .filter(instance => instance.elemID.typeName === FORM_TYPE)
+      .forEach(instance => {
+        fieldNames.forEach(fieldName => {
+          instance.value.design[fieldName] = instance.value.design[fieldName] ?? {}
+        })
+      })
+  },
   deploy: async changes => {
     if (!config.fetch.enableJSM || !config.fetch.enableJsmExperimental) {
       return {
@@ -144,6 +156,20 @@ const filter: FilterCreator = ({ config, client, getElemIdFunc }) => ({
       leftoverChanges,
       deployResult,
     }
+  },
+  onDeploy: async changes => {
+    const fieldNames = ['questions', 'sections', 'conditions']
+    changes
+      .filter(isInstanceChange)
+      .map(change => getChangeData(change))
+      .filter(instance => instance.elemID.typeName === FORM_TYPE)
+      .forEach(instance => {
+        fieldNames.forEach(fieldName => {
+          if (_.isEmpty(instance.value.design[fieldName])) {
+            delete instance.value.design[fieldName]
+          }
+        })
+      })
   },
 })
 export default filter
--- a/packages/jira-adapter/src/filters/forms/forms.ts
+++ b/packages/jira-adapter/src/filters/forms/forms.ts
@@ -26,8 +26,8 @@ import { createFormType, isDetailedFormsResponse, isFormsResponse } from './form
 const { isDefined } = lowerDashValues
 
 /*
-This filter fetches all forms from Jira Service Management and creates an instance element for each form.
-We use filter because we need to use cloudId which is not available in the infrastructure.
+* This filter fetches all forms from Jira Service Management and creates an instance element for each form.
+* We use filter because we need to use cloudId which is not available in the infrastructure.
 */
 const filter: FilterCreator = ({ config, client, getElemIdFunc }) => ({
   name: 'formsFilter',
@@ -42,7 +42,7 @@ const filter: FilterCreator = ({ config, client, getElemIdFunc }) => ({
 
     const jsmProjects = elements
       .filter(isInstanceElement)
-      .filter(instnace => instnace.elemID.typeName === PROJECT_TYPE)
+      .filter(instance => instance.elemID.typeName === PROJECT_TYPE)
       .filter(project => project.value.projectTypeKey === SERVICE_DESK)
 
     const forms = (await Promise.all(jsmProjects
@@ -60,13 +60,7 @@ const filter: FilterCreator = ({ config, client, getElemIdFunc }) => ({
               return undefined
             }
             const name = `${project.value.name}_${formResponse.name}`
-            delete detailedRes.data?.publish
-            delete detailedRes.data.design.settings.templateFormUuid
-            detailedRes.data.design.questions = Object.values(detailedRes.data.design.questions)
-            detailedRes.data.design.sections = Object.values(detailedRes.data.design.sections)
-            detailedRes.data.design.conditions = Object.values(detailedRes.data.design.conditions)
             const formValue = detailedRes.data
-            formValue.id = formResponse.id
             const serviceIds = adapterElements.createServiceIds(formValue, 'uuid', formType.elemID)
             const instanceName = getElemIdFunc ? getElemIdFunc(JIRA, serviceIds, naclCase(name)).name
               : naclCase(name)
--- a/packages/jira-adapter/src/filters/forms/forms.ts
+++ b/packages/jira-adapter/src/filters/forms/forms.ts
@@ -115,7 +115,6 @@ const filter: FilterCreator = ({ config, client, getElemIdFunc }) => ({
               transformationConfigByType: getTransformationConfigByType(jsmDuckTypeApiDefinitions.types),
               transformationDefaultConfig: jsmDuckTypeApiDefinitions.typeDefaults.transformation,
               parent: project,
-              // what ever default name you'd like
               defaultName: name,
               getElemIdFunc,
             })
--- a/packages/jira-adapter/src/filters/forms/forms.ts
+++ b/packages/jira-adapter/src/filters/forms/forms.ts
@@ -35,8 +35,9 @@ const filter: FilterCreator = ({ config, client, getElemIdFunc }) => ({
       return
     }
     const cloudId = await getCloudId(client)
-    const { formType } = createFormType()
+    const { formType, subTypes } = createFormType()
     elements.push(formType)
+    subTypes.forEach(subType => { elements.push(subType) })
 
     const jsmProject = elements
       .filter(isInstanceElement)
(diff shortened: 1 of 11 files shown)


packages/jira-adapter/test/filters/forms/forms.test.ts
```diff
@@ -0,0 +1,166 @@
...
+                  ],
+                },
+              ],
+              conditions: {},
+              sections: {},
+              questions: {},
+            },
+          },
+        })
+        elements = [projectInstance, projectType]
```
IdoZakk, line 113: nit: no need for the project type here

packages/jira-adapter/test/filters/forms/forms.test.ts
```diff
@@ -0,0 +1,166 @@
...
+      })
+      it('should add forms to elements when enableJSM is true', async () => {
+        await filter.onFetch(elements)
+        const instances = elements.filter(isInstanceElement)
+        const formInstance = instances.find(e => e.elemID.typeName === FORM_TYPE)
```
IdoZakk, line 118: You can do it combined, no need to store instances

packages/jira-adapter/src/filters/forms/forms.ts
```diff
@@ -0,0 +1,87 @@
...
+import { JIRA, PROJECT_TYPE, SERVICE_DESK } from '../../constants'
+import { getCloudId } from '../automation/cloud_id'
+import { createFormType, isDetailedFormsResponse, isFormsResponse } from './forms_types'
+
+const { isDefined } = lowerDashValues
+
+/*
+This filter fetches all forms from Jira Service Management and creates an instance element for each form.
+We use filter because we need to use cloudId which is not available in the infrastructure.
+*/
```
netama, line 31: nit:
```suggestion
/*
 This filter fetches all forms from Jira Service Management and creates an instance element for each form.
 We use filter because we need to use cloudId which is not available in the infrastructure.
*/
```

packages/jira-adapter/src/filters/forms/forms.ts
```diff
@@ -0,0 +1,87 @@
...
+const filter: FilterCreator = ({ config, client, getElemIdFunc }) => ({
+  name: 'formsFilter',
+  onFetch: async elements => {
+    if (!config.fetch.enableJSM) {
```
netama, line 35: nit: maybe we should consider adding a `jsm` folder under `filters` and putting these under it? will probably make navigation a little easier

packages/jira-adapter/src/filters/forms/forms.ts
```diff
@@ -0,0 +1,87 @@
...
+      return
+    }
+    const cloudId = await getCloudId(client)
+    const { formType, subTypes } = createFormType()
+    elements.push(formType)
+    subTypes.forEach(subType => { elements.push(subType) })
+
+    const jsmProjects = elements
+      .filter(isInstanceElement)
+      .filter(instnace => instnace.elemID.typeName === PROJECT_TYPE)
```
netama, line 45: typo
```suggestion
      .filter(instance => instance.elemID.typeName === PROJECT_TYPE)
```

packages/jira-adapter/src/filters/forms/forms.ts
```diff
@@ -0,0 +1,87 @@
...
+          return undefined
+        }
+        return Promise.all(res.data
+          .map(async formResponse => {
+            const detailedUrl = `/gateway/api/proforma/cloudid/${cloudId}/api/2/projects/${project.value.id}/forms/${formResponse.id}`
+            const detailedRes = await client.getSinglePage({ url: detailedUrl })
+            if (!isDetailedFormsResponse(detailedRes)) {
+              return undefined
+            }
+            const name = `${project.value.name}_${formResponse.name}`
```
netama, line 62: are these guaranteed to be unique? in general I think where we can reuse the `idFields` it'd be better, but doesn't have to be a blocker.
also - do we support explicitly mentioning this type in the include/exclude config?

packages/jira-adapter/src/filters/forms/forms.ts
```diff
@@ -0,0 +1,78 @@
...
+      const url = `/gateway/api/proforma/cloudid/${cloudId}/api/1/projects/${project.value.id}/forms`
+      const res = await client.getSinglePage({ url })
+      if (isFormsResponse(res)) {
+        return Promise.all(res.data.map(async formResponse => {
+          const detailedUrl = `/gateway/api/proforma/cloudid/${cloudId}/api/2/projects/${project.value.id}/forms/${formResponse.id}`
+          const detailedRes = await client.getSinglePage({ url: detailedUrl })
+          if (isDetailedFormsResponse(detailedRes)) {
+            const name = `${project.value.name}_${formResponse.name}`
+            delete detailedRes.data?.publish
+            delete detailedRes.data.design.settings.templateFormUuid
```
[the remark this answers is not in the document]
netama: 1. ^ was this discussed? (I don't have context about the motivation for this)
2. I might be missing something, but why do we need the `data?.` part in the prev line but not here and below? is this now covered in a type guard (which means we can remove the one from the prev line)?

packages/jira-adapter/src/filters/forms/forms_types.ts
```diff
@@ -0,0 +1,201 @@
...
+* limitations under the License.
+*/
+
+import { ObjectType, ElemID, BuiltinTypes, CORE_ANNOTATIONS, ListType } from '@salto-io/adapter-api'
+import Joi from 'joi'
+import { elements as adapterElements } from '@salto-io/adapter-components'
+import { createSchemeGuard } from '@salto-io/adapter-utils'
+import { JIRA, FORM_TYPE } from '../../constants'
+
+export type detailedFormResponse = {
```
netama, line 23: type names should start with a capital letter (also applies for all the other places)

packages/jira-adapter/src/filters/forms/forms.ts
```diff
@@ -0,0 +1,87 @@
...
+          .map(async formResponse => {
+            const detailedUrl = `/gateway/api/proforma/cloudid/${cloudId}/api/2/projects/${project.value.id}/forms/${formResponse.id}`
+            const detailedRes = await client.getSinglePage({ url: detailedUrl })
+            if (!isDetailedFormsResponse(detailedRes)) {
+              return undefined
+            }
+            const name = `${project.value.name}_${formResponse.name}`
+            delete detailedRes.data?.publish
+            delete detailedRes.data.design.settings.templateFormUuid
+            detailedRes.data.design.questions = Object.values(detailedRes.data.design.questions)
```
netama, line 65: why are we changing this? is this for the response format?

packages/jira-adapter/src/filters/forms/forms.ts
```diff
@@ -0,0 +1,87 @@
...
+import { FilterCreator } from '../../filter'
+import { JIRA, PROJECT_TYPE, SERVICE_DESK } from '../../constants'
+import { getCloudId } from '../automation/cloud_id'
+import { createFormType, isDetailedFormsResponse, isFormsResponse } from './forms_types'
+
+const { isDefined } = lowerDashValues
+
+/*
+This filter fetches all forms from Jira Service Management and creates an instance element for each form.
+We use filter because we need to use cloudId which is not available in the infrastructure.
```
netama, line 30: can you please open a ticket for it and cc me? I think we can adjust it soon and then get rid of the filter (since it is re-implementing some logic that we'd like to have shared and easy-to-control using the infra)

packages/jira-adapter/src/filters/forms/forms_types.ts
```diff
@@ -0,0 +1,201 @@
...
+    publish?: string
+    id: number
+    design: {
+      settings: {
+        templateFormUuid?: string
+      }
+      questions: {}
+      sections: {}
+      conditions: {}
+    }
```
netama, line 35: where are all of these covered in the scheme guard? they should be 1-1, otherwise things can crash
(also applies to the other guards, I sampled this one randomly :) )

packages/jira-adapter/src/filters/forms/forms_types.ts
```diff
@@ -0,0 +1,201 @@
...
+export const DETAILED_FORM_RESPONSE_SCHEME = Joi.object({
+  data: Joi.object({
+    uuid: Joi.string().required(),
+    design: Joi.object({
+    }).unknown(true).required(),
+  }).unknown(true).required(),
+}).unknown(true).required()
+
+
+export const createFormType = (): {
```
netama, line 62: this was already reviewed so let's leave it this way for now, but in general I think we should try to reduce the amount of types we define manually here

packages/jira-adapter/src/filters/forms/forms_types.ts
```diff
@@ -0,0 +1,201 @@
...
+      },
+      questionKey: {
+        refType: BuiltinTypes.STRING,
+      },
+      jiraField: {
+        refType: BuiltinTypes.STRING,
+      },
+    },
+  })
+  const FormDesignType = new ObjectType({
```
netama, line 153: nit: 
```suggestion
  const formDesignType = new ObjectType({
```

packages/jira-adapter/src/filters/forms/forms_types.ts
```diff
@@ -0,0 +1,201 @@
...
+      localId: {
+        refType: new ListType(BuiltinTypes.STRING),
+        annotations: { [CORE_ANNOTATIONS.HIDDEN_VALUE]: true },
+      },
+      attrs: {
+        refType: attributeContnetLayoutType,
+      },
+    },
+  })
+  const FormLayoutItemType = new ObjectType({
```
netama, line 119: nit: 
```suggestion
  const formLayoutItemType = new ObjectType({
```
(also in a few others)

packages/jira-adapter/src/filters/forms/forms_types.ts
```diff
@@ -0,0 +1,201 @@
...
+      submit: {
+        refType: FormSubmitType,
+      },
+      templateFormUuid: {
+        refType: BuiltinTypes.STRING,
+      },
+    },
+  })
+  const attributeContnetLayoutType = new ObjectType({
+    elemID: new ElemID(JIRA, 'attributeContnetLayout'),
```
netama, line 96: typo
```suggestion
  const attributeContentLayoutType = new ObjectType({
    elemID: new ElemID(JIRA, 'attributeContentLayout'),
```

packages/jira-adapter/src/filters/forms/forms_types.ts
```diff
@@ -0,0 +1,201 @@
...
+        refType: new ListType(BuiltinTypes.STRING),
+        annotations: { [CORE_ANNOTATIONS.HIDDEN_VALUE]: true },
+      },
+      attrs: {
+        refType: attributeContnetLayoutType,
+      },
+    },
+  })
+  const FormLayoutItemType = new ObjectType({
+    elemID: new ElemID(JIRA, 'layoutForm'),
```
netama, line 120: forgot to update? it should log a warning for duplicates IIRC, probably best to look for these before merging to see if there are others

packages/jira-adapter/src/filters/forms/forms_types.ts
```diff
@@ -0,0 +1,201 @@
...
+    fields: {
+      versoin: {
```
netama, line 122:
```suggestion
      version: {
```
(I'm assuming?)
in general, this is part of why we shouldn't define these manually - this is very error-prone, might be worth discussing IRL what the process is here and seeing if we can do somthing safer and more efficient :)

packages/jira-adapter/src/filters/forms/forms_types.ts
```diff
@@ -0,0 +1,201 @@
...
+    },
+  })
+  const FormDesignType = new ObjectType({
+    elemID: new ElemID(JIRA, 'FormDesign'),
+    fields: {
+      settings: {
+        refType: FormSettingsType,
+      },
+      layout: {
+        refType: new ListType(FormLayoutItemType),
```
netama, line 160: is this intentionally layout item and not layout?

packages/jira-adapter/src/filters/forms/forms_types.ts
```diff
@@ -0,0 +1,201 @@
...
+      },
+      conditions: {
+        refType: BuiltinTypes.UNKNOWN,
```
netama, line 163: why are these unknown? is this an OR situtation that is not supported? (let's also discuss IRL, and also if you can add a comment next to this explaining what support is missing, hopefully we can address it eventually)

packages/jira-adapter/src/reference_mapping.ts
```diff
@@ -983,6 +983,12 @@ export const referencesRules: JiraFieldReferenceDefinition[] = [
     jiraMissingRefStrategy: 'typeAndValue',
     target: { type: CALENDAR_TYPE },
   },
+  {
+    src: { field: 'jiraField', parentTypes: ['Question'] },
+    serializationStrategy: 'id',
+    jiraMissingRefStrategy: 'typeAndValue',
+    target: { type: 'Field' },
```
netama, line 990:
```suggestion
    target: { type: FIELD_TYPE_NAME },
```

netama (commented):
looks good overall, added some questions - my main concern is how we work with types here (both typescript and object types), probably best to discuss IRL
also - if this is adding more fetch types, I think it might be worth adding a secondary fetch config flag until we are done reviewing and testing and can open this up for everyone? it can be something like `enableJsmExperimental` or something else along these lines, and once the next batch of types are stable we can add them to the main flag and delete it / use for the next milestone

packages/jira-adapter/src/filters/forms/forms.ts
idoamit8: will do it in a different PR after all the current PRS will merge to avoid merging conflicts.

packages/jira-adapter/src/filters/forms/forms.ts
idoamit8: No it's not unique. There could be two forms with the same name. Not quite sure what to put here that can be unique - All I can think is uuid

packages/jira-adapter/src/filters/forms/forms.ts
idoamit8: No, Its because the response returns an object that its keys are numbers and the values are as shown

packages/jira-adapter/src/filters/forms/forms_types.ts
idoamit8: Will talk with you IRL

packages/jira-adapter/src/filters/forms/forms_types.ts
idoamit8: yes just becuase layout is list of layoutItems

packages/jira-adapter/src/filters/forms/forms_types.ts
```diff
@@ -0,0 +1,227 @@
...
+    conditions: Joi.object().unknown(true).required(),
+  }).unknown(true).required(),
+}).unknown(true).required()
+
+
+export const createFormType = (): {
+    formType: ObjectType
+    subTypes: ObjectType[]
+  } => {
+  const FormSubmitType = new ObjectType({
```
shir-reifenberg, line 83: All ObjectType variables here should start with lowercase as this is a variable and not a typescript type 
(also relevant for all other Object types in this function) 
```suggestion
  const formSubmitType = new ObjectType({
```

packages/jira-adapter/src/filters/forms/forms.ts
```diff
@@ -0,0 +1,147 @@
...
+const { isDefined } = lowerDashValues
+
+const deployForms = async (
+  change: Change<InstanceElement>,
+  client: JiraClient,
+): Promise<void> => {
+  const form = getChangeData(change)
+  const project = getParent(form)
+  if (form.value.design?.settings?.name === undefined) {
+    return
```
shir-reifenberg, line 39: What is this case? is it valid (if so let's add a comment)?and if not, maybe `log.error` or return deploy error?

packages/jira-adapter/test/filters/forms/forms.test.ts
```diff
@@ -0,0 +1,177 @@
...
+import { safeJsonStringify } from '@salto-io/adapter-utils'
+import { getDefaultConfig } from '../../../src/config/config'
+import formsFilter from '../../../src/filters/forms/forms'
+import { createEmptyType, getFilterParams, mockClient } from '../../utils'
+import { FORM_TYPE, JIRA, PROJECT_TYPE } from '../../../src/constants'
+import JiraClient from '../../../src/client/client'
+import { CLOUD_RESOURCE_FIELD } from '../../../src/filters/automation/cloud_id'
+
+
+describe('forms filter', () => {
```
shir-reifenberg, line 30: Can you please add tests for deploy as well ? 🙏

packages/jira-adapter/src/filters/forms/forms_types.ts
```diff
@@ -0,0 +1,227 @@
...
+* limitations under the License.
+*/
+
+import { ObjectType, ElemID, BuiltinTypes, CORE_ANNOTATIONS, ListType } from '@salto-io/adapter-api'
+import Joi from 'joi'
+import { elements as adapterElements } from '@salto-io/adapter-components'
+import { createSchemeGuard } from '@salto-io/adapter-utils'
+import { JIRA, FORM_TYPE } from '../../constants'
+
+export type DetailedFormDataResponse = {
```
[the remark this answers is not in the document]
idoamit8: Nope

packages/jira-adapter/src/filters/forms/forms.ts
```diff
@@ -0,0 +1,147 @@
...
+            )
+          }))
+      })))
+      .flat()
+      .filter(isDefined)
+    forms.forEach(form => elements.push(form))
+  },
+  deploy: async changes => {
+    if (!config.fetch.enableJSM || !config.fetch.enableJsmExperimental) {
+      return {
```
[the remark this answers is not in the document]
idoamit8: it will not make a lot of errors? In most cases people doesn't have those turned on

packages/jira-adapter/src/filters/forms/forms.ts
idoamit8: added error

packages/jira-adapter/src/filters/forms/forms.ts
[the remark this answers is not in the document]
shir-reifenberg: If we only log when there are changes of JSM types && flag is off, it shouldn't be noisy. and we should be able to explain users why the deployment didn't succeed. 
(For example if in envA JSM is enabled, and envB JSM is off, user might select JSM changes and try to deploy it...)

packages/jira-adapter/src/filters/forms/forms.ts
```diff
@@ -104,19 +105,20 @@ const filter: FilterCreator = ({ config, client, getElemIdFunc }) => ({
...
-              formValue,
-              [...prefixPath.slice(0, -1), 'forms', pathNaclCase(instanceName)],
-              {
-                [CORE_ANNOTATIONS.PARENT]: [new ReferenceExpression(project.elemID, project)],
-              }
-            )
+            const jsmDuckTypeApiDefinitions = config[JSM_DUCKTYPE_API_DEFINITIONS]
+            if (jsmDuckTypeApiDefinitions === undefined) {
+              return undefined
+            }
```
shir-reifenberg, line 111: can it happen ? 🤔

packages/jira-adapter/src/filters/forms/forms.ts
```diff
@@ -0,0 +1,149 @@
...
+            return toBasicInstance({
+              entry: formValue,
+              type: formType,
+              transformationConfigByType: getTransformationConfigByType(jsmDuckTypeApiDefinitions.types),
+              transformationDefaultConfig: jsmDuckTypeApiDefinitions.typeDefaults.transformation,
+              parent: project,
+              // what ever default name you'd like
```
shir-reifenberg, line 118: please remove this 😅

packages/jira-adapter/src/filters/forms/forms.ts
```diff
@@ -0,0 +1,149 @@
...
+              defaultName: name,
```
shir-reifenberg, line 119: Sorry if my previous suggestion wasn't clear, but if we're creating instance with infra, IMO we should also use the config instead of creating the name in here (for many reasons - it can be easily configured even without changing code, shared logic, easier to maintain..)
 
in this case I _think_ that by having those values in the config you'll get the desired result (please validate):

```
Transformation: {
  idFields: ['name'],
  extendsParentId: true,
}
```

packages/jira-adapter/src/filters/forms/forms_types.ts
idoamit8: oh, I miss understand this. Thanks

shir-reifenberg (approved):
Looks good overall 😎 , added a few comments
approving to unblock, don't forget to add tests for deployment 🤞

PR #4924, opened 2023-09-29
SALTO-4752: Add generic fetch error exception in ducktype

IdoZakk:
This is done to enable fetch errors for ScriptRunner login

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/adapter-components/src/config/shared.ts
```diff
@@ -26,6 +25,12 @@ export const DEPLOYER_FALLBACK_VALUE = '##DEPLOYER##'
 
 export class InvalidSingletonType extends Error {}
 
+export class FetchError extends Error {
```
netama, line 28: @ori-moisis making sure you're ok with the name? (this is not exposed externally for now, Ido wanted to add it to to differentiate the error type inside the adapter-components code)

netama (approved):
LGTM but might be worth waiting for Ori

packages/adapter-components/src/config/shared.ts
netama: let's just rename to `AdapterFetchError` so no one mixes it up with something external-facing, and merge?

PR #4925, opened 2023-09-29
SALTO-4752: script runner survey

IdoZakk:
Change the login failures in ScriptRunner to cause fetch errors.
Also add a special message when the cause of failure is the survey screen
(And changed an error I made in the previous PR that exported the AdapterFetchError under the wrong name)

---

It was tested against a real-life scenario

---
_Release Notes_: 
Jira Adapter:
* In case a fetch that includes ScriptRunner fails the fetch will end with a fetch error. Also, in case the reason is the ScriptRunner's Survey, the user will be provided with the relevant instructions

---
_User Notifications_: 
None


shir-reifenberg (approved):
LGTM 😎

PR #4930, opened 2023-10-01
SALTO-4727: Instead of removing Salesforce MetadataTypes, hide them (Quick Fetch step)

tamtamirr:
Instead of removing Salesforce MetadataTypes, hide them
 
---

This is required since in quick fetch we rely on the **MetadataTypes** in the Workspace instead of fetching them.
 
---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/filters/custom_objects_to_object_type.ts
```diff
@@ -324,12 +324,12 @@ const createNestedMetadataInstances = (instance: InstanceElement,
...
           nestedInstanceValues[INSTANCE_FULL_NAME_FIELD] = fullName
           const path = [
-            ...(objPath as string[]).slice(0, -1),
+            ...(objectType.path as string[]).slice(0, -1),
             typeFolderName,
             instanceFileName,
           ]
           return new InstanceElement(instanceName, type, nestedInstanceValues,
-            path, { [CORE_ANNOTATIONS.PARENT]: [new ReferenceExpression(objElemID)] })
+            path, { [CORE_ANNOTATIONS.PARENT]: [new ReferenceExpression(objectType.elemID, objectType)] })
```
tamtamirr, line 332: This is a small change, since that caused the values fo the references to be unresolved for no reason. Unrelated to hiding the types though.

tamtamirr (commented):
Added a small comment for clarification

tomsellek (approved):
The code looks fine to me, but I would prefer it was reviewed by someone who understands the implications better than me.

packages/salesforce-adapter/src/filters/custom_objects_to_object_type.ts
```diff
@@ -852,7 +847,14 @@ const filterCreator: LocalFilterCreator = ({ config }) => {
         .filter(elem => !existingElementIDs.has(elem.elemID.getFullName()))
         .forEach(newElem => elements.push(newElem))
 
-      await removeIrrelevantElements(elements)
+      // Remove instances and set MetadataType as hidden
+      await removeAsync(elements, isInstanceOfType(CUSTOM_OBJECT))
+      const customObjectMetadataType = elements
+        .filter(isObjectType)
+        .find(objectType => metadataTypeSync(objectType) === CUSTOM_OBJECT)
```
amitgoldarb, line 854: I'm slightly confused. The implementations of metaTypeSync and apiName are different? Will this result in the same thing.
I would expect maybe to use apiNameSync in this case if we had it. Changing from async to sync probably doesn't have significant performance implications, probably better to keep the same?
Also we are using find because we are confident there is one? Are we leaving something here leftover?

packages/salesforce-adapter/src/filters/topics_for_objects.ts
```diff
@@ -84,10 +84,17 @@ const filterCreator: LocalFilterCreator = () => ({
       }
     })
 
-    // Remove TopicsForObjects Instances & Type to avoid information duplication
-    await removeAsync(
+    // Remove TopicsForObjects Instances & and set the MetadataType as hidden to avoid information duplication
+    const topicsForObjectType = elements
+      .filter(isObjectType)
+      .find(objectType => apiNameSync(objectType) === TOPICS_FOR_OBJECTS_METADATA_TYPE)
```
amitgoldarb, line 90: Same comment here with the find. Also in this case we are using apiNameSync. Also looking at the implementation (comment for a different PR), it would probably be benificial to DRY the two implementaitons.

amitgoldarb (commented):
Looks good.
Added a few comments/questions

packages/salesforce-adapter/src/filters/custom_objects_to_object_type.ts
tamtamirr: **metadataTypeSync** will return **"CustomObject"** for both the **Data Records** and the type itself. Since we filter for **ObjectTypes** as-well this is safe. However, it's better to use **apiNameSync** like in the other filter. I'll change that.

Re async -> sync: No reason to keep async code when not needed, especially when implementing new code / refactoring a specific logic. (Not saying we should fix everything right now, but the small changes we make should also be sync implementations)

Re find, I don't understand the concern..? It will return undefined if the type does not exist.

PR #4933, opened 2023-10-02
SALTO-4727: Aggregated CustomObject author information

tamtamirr:
This PR improves the author information on **CustomObjects**.
 
---

Using the **Author Information** of the last changed **sub-instance** /  **CustomField**  is more accurate than using the **Author Information** on the **CustomObject** itself which never really change other than creation (of a **Custom Object**) or don't exist for standard ones (they do exist, with **unix time zero** timestamp which we later omit)

- This is also a required step for **Quick Fetch** on **CustomObjects**

---
_Release Notes_: 
Salesforce Adapter:
- More accurate Author Information on **Custom and Standard Object**.

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/filters/custom_objects_to_object_type.ts
```diff
@@ -324,12 +324,12 @@ const createNestedMetadataInstances = (instance: InstanceElement,
...
           nestedInstanceValues[INSTANCE_FULL_NAME_FIELD] = fullName
           const path = [
-            ...(objPath as string[]).slice(0, -1),
+            ...(objectType.path as string[]).slice(0, -1),
             typeFolderName,
             instanceFileName,
           ]
           return new InstanceElement(instanceName, type, nestedInstanceValues,
-            path, { [CORE_ANNOTATIONS.PARENT]: [new ReferenceExpression(objElemID)] })
+            path, { [CORE_ANNOTATIONS.PARENT]: [new ReferenceExpression(objectType.elemID, objectType)] })
```
tamtamirr, line 332: Not passing value to the **ReferenceExpression** here will require us to resolve it later for no good reason. You will see some small changes in UTs that were affected by it :) It's a good improvement overall as-well

packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -94,15 +94,15 @@ export const isMetadataValues = createSchemeGuard<MetadataValues>(METADATA_VALUE
 // This function checks whether an element is an instance of a certain metadata type
 // note that for instances of custom objects this will check the specific type (i.e Lead)
 // if you want instances of all custom objects use isInstanceOfCustomObject
-export const isInstanceOfType = (...types: string[]) => (
+export const isInstanceOfType = (...typeNames: string[]) => (
```
tamtamirr, line 97: The rename here was needed because I'm importing **lowerdash types**

packages/salesforce-adapter/src/filters/author_information/custom_objects.ts
```diff
@@ -114,12 +158,29 @@ const filterCreator: RemoteFilterCreator = ({ client, config }) => ({
...
-            customFieldsFilePropertiesMap,
-            object)
+      const instancesByParent = _.groupBy(
+        elements
+          .filter(isMetadataInstanceElementSync)
+          .filter(isElementWithResolvedParent),
+        instance => {
+          // SALTO-4824
+          // eslint-disable-next-line no-underscore-dangle
+          const [parent] = instance.annotations._parent
```
tamtamirr, line 168: While this might seem odd/ugly this is really safe. https://salto-io.atlassian.net/browse/SALTO-4824 explains this limitation

tamtamirr (commented):

tamtamirr:
Closing due to refactor

packages/salesforce-adapter/src/filters/author_information/custom_objects.ts
```diff
@@ -67,38 +81,68 @@ const getCustomFieldFileProperties = async (client: SalesforceClient):
...
+
+const setObjectAuthorInformation = (
+  {
+    object,
+    typeFileProperties,
+    customFieldsFileProperties,
+    nestedInstances,
+  }: ObjectAuthorInformationSupplierArgs
+): void => {
+  // Set author information on the CustomObject's fields
```
tomsellek, line 103: In general I think these types of comments that describe the intention behind a block of code are an indication that the block of code should be extracted to a named function

packages/salesforce-adapter/src/filters/author_information/custom_objects.ts
```diff
@@ -67,38 +81,68 @@ const getCustomFieldFileProperties = async (client: SalesforceClient):
...
+  )
+  if (mostRecentAuthorInformation) {
+    object.annotations[CORE_ANNOTATIONS.CHANGED_BY] = mostRecentAuthorInformation.changedBy
+    object.annotations[CORE_ANNOTATIONS.CHANGED_AT] = mostRecentAuthorInformation.changedAt
+    // This info should always come from the FileProperties of the CustomObject.
+    // Standard Objects won't have values here
+    if (typeFileProperties) {
+      object.annotations[CORE_ANNOTATIONS.CREATED_BY] = typeFileProperties.createdByName
+      object.annotations[CORE_ANNOTATIONS.CREATED_AT] = typeFileProperties.createdDate
+    }
```
tomsellek, line 134: This may result in a case where there are `CHANGED_AT` and `CHANGED_BY` values but no `CREATED_BY` or `CREATED_AT`. This is probably a non-issue, unless someone somewhere checks for `CREATED_BY` as a proxy for whether the object has any author info.

tomsellek (approved):
LGTM. Added a couple of questions/nits.

packages/salesforce-adapter/src/filters/author_information/custom_objects.ts
tamtamirr: I don't think it's an actual concern. Assuming all the information exist if one of the properties does is wrong IMO

packages/salesforce-adapter/src/filters/author_information/custom_objects.ts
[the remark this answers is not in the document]
tamtamirr: While I tend to agree, in that case exporting to a method makes the code more complex IMO

PR #4935, opened 2023-10-02
SALTO-4838: Support aliases in Okta

shir-reifenberg:
Add alias filter
There's already a noise reduction in saas

---

---
_Release Notes_: 

_Okta adapter_:
- add alias to Okta elements

---
_User Notifications_: 
None


edenhassid (approved):
looks good!! 💯

PR #4937, opened 2023-10-02
SALTO-4727: Do not correct CustomObject sub-instances types if they are already correct

tamtamirr:
Added code that asserts we don't correct already correct types of fields on CustomObject sub-instances Metadata Types. This is a required step for quick fetch on CustomObjects.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -525,3 +525,7 @@ export const getChangedAtSingleton = async (
 export const isCustomType = (element: Element): element is ObjectType => (
   isObjectType(element) && ENDS_WITH_CUSTOM_SUFFIX_REGEX.test(apiNameSync(element) ?? '')
 )
+
+export const asListType = (type: TypeElement): ListType => (
```
amitgoldarb, line 529: super nit: I find this new naming suggestion consistent with the lodash _.toArray which _feels_ slightly better IMO.
```suggestion
export const toListType = (type: TypeElement): ListType => (
```

amitgoldarb (approved):
LGTM! Also I like this PR, it is very concise.

PR #4938, opened 2023-10-02
SALTO-4846 - Salesforce: make all references in REFERENCE_TO resolved

tomsellek:
There does not seem to be a reason to use unresolved references, and it will help us avoid making the utility in https://github.com/salto-io/salto/pull/4914 async

---

This was not an issue until now because the `reference_annotation` filter runs very late and so we never tried to access the references it creates. But if we want to create generic utilities to access the contents of the REFERENCE_TO annotation, we need them to be resolved.

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/filters/reference_annotations.ts
+++ b/packages/salesforce-adapter/src/filters/reference_annotations.ts
@@ -14,7 +14,7 @@
 * limitations under the License.
 */
 import {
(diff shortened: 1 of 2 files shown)


tamtamirr (approved):
LGTM

PR #4942, opened 2023-10-03
SALTO-4826 Remove deleted files before creating new ones

netama:
We ran into a bug with static files, where due to a folder structure change, we removed file `a` and instead created a nested file under the folder `a`. This caused a `ENOTDIR` error, since `a` was still a file when we tried to create a nested file under it.

When flushing the dir store, there are `deleted` and `updated` files. 
Switching to first delete the deleted files, and then create the new files, in order to support this scenario.

---

The fix applies to all dir stores, but the main realistic scenario where this can happen is with static files with no extension.

---
_Release Notes_: 

_Core_:
* Fix bug where if file `x` was removed and a new file was created under a (new) folder called `x`, we would fail on `ENOTDIR` (since the original file `x` was still there)

---
_User Notifications_: 
None

### DIFF
--- a/packages/core/src/local-workspace/dir_store.ts
+++ b/packages/core/src/local-workspace/dir_store.ts
@@ -168,12 +168,14 @@ const buildLocalDirectoryStore = <T extends dirStore.ContentType>(
   )
 
   const flush = async (): Promise<void> => {
+    // first delete the files, so that we can create files nested under deleted-file paths and vice versa
+    // (e.g., removing a and creating a/b)
+    await withLimitedConcurrency(Array.from(deleted).map(f => () => deleteFile(f, true)), DELETE_CONCURRENCY)
     await withLimitedConcurrency(
       Object.values(updated).map(f => () => writeFile(f)), WRITE_CONCURRENCY
     )
-    await withLimitedConcurrency(Array.from(deleted).map(f => () => deleteFile(f, true)), DELETE_CONCURRENCY)
-    updated = {}
     deleted = new Set()
+    updated = {}
   }
 
   const deleteAllEmptyDirectories = async (): Promise<void> => {
--- a/packages/core/test/workspace/local/directory_store.test.ts
+++ b/packages/core/test/workspace/local/directory_store.test.ts
@@ -19,6 +19,7 @@ import {
   stat, exists, readFile, replaceContents, mkdirp, rm, isEmptyDir,
   rename, notFoundAsUndefined,
 } from '@salto-io/file'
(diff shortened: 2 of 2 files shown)


packages/core/src/local-workspace/dir_store.ts
```diff
@@ -168,12 +168,14 @@ const buildLocalDirectoryStore = <T extends dirStore.ContentType>(
...
+    // first delete the files, so that we can create files nested under deleted-file paths and vice versa
+    // (e.g., removing a and creating a/b)
+    await withLimitedConcurrency(Array.from(deleted).map(f => () => deleteFile(f, true)), DELETE_CONCURRENCY)
     await withLimitedConcurrency(
       Object.values(updated).map(f => () => writeFile(f)), WRITE_CONCURRENCY
     )
-    await withLimitedConcurrency(Array.from(deleted).map(f => () => deleteFile(f, true)), DELETE_CONCURRENCY)
-    updated = {}
     deleted = new Set()
+    updated = {}
```
ori-moisis, line 178: nit - this isn't new but since you moved the `update =...` thing, to make sure we don't miss deletions / updates while running async stuff, I think we need to do this "reset" before running async code here, i.e
```typescript
const deletesToHandle = deleted
delete = new Set()
const updatesToHandle = updated
updated = {}
await withLimitedConcurrency(...
```

ori-moisis (approved):

packages/core/src/local-workspace/dir_store.ts
```diff
@@ -168,12 +168,19 @@ const buildLocalDirectoryStore = <T extends dirStore.ContentType>(
...
+    // (e.g., removing a and creating a/b)
+    await withLimitedConcurrency(Array.from(deletesToHandle).map(f => () => deleteFile(f, true)), DELETE_CONCURRENCY)
     await withLimitedConcurrency(
-      Object.values(updated).map(f => () => writeFile(f)), WRITE_CONCURRENCY
+      Object.values(updatesToHandle).map(f => () => writeFile(f)), WRITE_CONCURRENCY
     )
-    await withLimitedConcurrency(Array.from(deleted).map(f => () => deleteFile(f, true)), DELETE_CONCURRENCY)
-    updated = {}
-    deleted = new Set()
+    if (deleted.size > 0 || Object.keys(updated).length > 0) {
```
netama, line 181: @ori-moisis I made the change you suggested, also adding logs for now since this could potentially change the existing behavior - lmk if you think we should do something else (e.g. continue in a loop until these are empty, which I think is not ideal either)

PR #4944, opened 2023-10-03
SALTO-4769: Add properties & push status to ProfileMapping instances (Okta)

shir-reifenberg:
Add push status info for ProfileMapping instances

---

changes:
1. Add deployment for ProfileMapping instances (only modifications)
2. Fetch properties for profile mapping which also contains info re push status of each property

---

_Release Notes_: 

_Okta_adapter_:
- Fetch properties push status in ProfileMapping instances
- Support deployments of ProfileMapping

---

_User Notifications_: 

_Okta_adapter_:
- properties push status will be added to ProfileMapping instances

### DIFF
--- a/packages/okta-adapter/src/config.ts
+++ b/packages/okta-adapter/src/config.ts
@@ -44,6 +44,7 @@ export type OktaFetchConfig = configUtils.UserFetchConfig & {
   convertUsersIds?: boolean
   enableMissingReferences?: boolean
   includeGroupMemberships?: boolean
+  includeProfileMappingProperties?: boolean
 }
 
 export type OktaSwaggerApiConfig = configUtils.AdapterSwaggerApiConfig<OktaActionName>
@@ -1224,6 +1225,16 @@ const DEFAULT_TYPE_CUSTOMIZATIONS: OktaSwaggerApiConfig['types'] = {
       fieldsToOmit: DEFAULT_FIELDS_TO_OMIT.concat({ fieldName: '_links' }),
       fieldsToHide: [{ fieldName: 'id' }],
     },
+    deployRequests: {
+      // only modificaiton is supported
+      modify: {
+        url: '/api/v1/mappings/{mappingId}',
+        method: 'post',
+        urlParamsToFields: {
+          mappingId: 'id',
+        },
+      },
+    },
   },
   ProfileMappingSource: {
     transformation: {
@@ -1465,7 +1476,6 @@ export const SUPPORTED_TYPES = {
     'api__v1__idps',
   ],
   InlineHook: ['api__v1__inlineHooks'],
-  // TODO SALTO-2734 returns 401
   ProfileMapping: ['api__v1__mappings'],
   LinkedObjectDefinitions: ['api__v1__meta__schemas__user__linkedObjects'],
   GroupSchema: ['GroupSchema'],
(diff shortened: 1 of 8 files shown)


packages/okta-adapter/src/config.ts
```diff
@@ -1224,6 +1225,16 @@ const DEFAULT_TYPE_CUSTOMIZATIONS: OktaSwaggerApiConfig['types'] = {
       fieldsToOmit: DEFAULT_FIELDS_TO_OMIT.concat({ fieldName: '_links' }),
       fieldsToHide: [{ fieldName: 'id' }],
     },
+    deployRequests: {
+      // only modificaiton is supported
```
netama, line 1229: 1. typo nit
```suggestion
      // only modification is supported
```
2. we can still support additions if we first find the correct mapping id, no? (in a deploy that creates a new app won't we also want to support customizing the mapping?)

packages/okta-adapter/src/filters/profile_mapping_properties.ts
```diff
@@ -0,0 +1,58 @@
...
+import { FETCH_CONFIG } from '../config'
+
+const log = logger(module)
+
+const getProfileMapping = async (
+  mappingId: string,
+  client: OktaClient
+): Promise<Values> => (await client.getSinglePage({
+  url: `/api/v1/mappings/${mappingId}`,
+})).data as Values[]
```
netama, line 30: hmm is it `Values` or `Values[]`? (and should we verify something about the structure?)

netama (commented):
looks good :) added some questions

packages/okta-adapter/src/config.ts
shir-reifenberg: yes thats a good point, I'll add it

packages/okta-adapter/src/config.ts
```diff
@@ -1219,11 +1220,27 @@ const DEFAULT_TYPE_CUSTOMIZATIONS: OktaSwaggerApiConfig['types'] = {
   },
   ProfileMapping: {
     transformation: {
-      idFields: ['source.name', 'target.name'],
+      idFields: ['&source.id', '&target.id'],
```
shir-reifenberg, line 1223: Not super related to this PR, but we need to change id fields. name might repeat per app (for example if there are two salesforce apps connected), might do this in a separate PR as we still have some missing references there that might cause internal ids in elemIds

packages/okta-adapter/test/filters/add_alias.test.ts
```diff
@@ -25,8 +24,6 @@ describe('addAliasFilter', () => {
   const groupType = new ObjectType({ elemID: new ElemID(OKTA, GROUP_TYPE_NAME) })
   const appType = new ObjectType({ elemID: new ElemID(OKTA, APPLICATION_TYPE_NAME) })
   const profileMappingType = new ObjectType({ elemID: new ElemID(OKTA, PROFILE_MAPPING_TYPE_NAME) })
-  const config = { ...DEFAULT_CONFIG }
-  config[FETCH_CONFIG].includeGroupMemberships = true
```
shir-reifenberg, line 29: not related to this PR but I noticed its redundant

packages/okta-adapter/src/filters/profile_mapping_addition.ts
```diff
@@ -0,0 +1,88 @@
...
+        && isAdditionChange(change)
+        && getChangeData(change).elemID.typeName === PROFILE_MAPPING_TYPE_NAME
+    )
+
+    const deployResult = await deployChanges(
+      relevantChanges.filter(isInstanceChange),
+      async change => deployProfileMappingAddition(change, client, config[API_DEFINITIONS_CONFIG])
+    )
+
+    return {
```
netama, line 81: from IRL - let's add a TODO for the deletions

netama (approved):

PR #4951, opened 2023-10-04
SALTO-4863 create path index for element

dantal4:
extract into a workspace API the code that we use in fetch-from-workspace for creating a path index for elements that aren’t in the state.

NOTE: this API can be removed when we do the fragment refactor.

---

---
_Release Notes_: 
Workspace:
- create path index for element API

---
_User Notifications_: 
None

### DIFF
--- a/packages/core/src/core/fetch.ts
+++ b/packages/core/src/core/fetch.ts
@@ -30,7 +30,7 @@ import {
 } from '@salto-io/adapter-api'
 import { applyInstancesDefaults, resolvePath, flattenElementStr, buildElementsSourceFromElements, safeJsonStringify, walkOnElement, WalkOnFunc, WALK_NEXT_STEP, setPath, walkOnValue } from '@salto-io/adapter-utils'
 import { logger } from '@salto-io/logging'
-import { merger, elementSource, expressions, Workspace, pathIndex, updateElementsWithAlternativeAccount, createAdapterReplacedID, remoteMap, adaptersConfigSource as acs } from '@salto-io/workspace'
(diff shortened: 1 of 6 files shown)


packages/workspace/src/workspace/workspace.ts
```diff
@@ -193,6 +194,8 @@ export type Workspace = {
   getAllChangedByAuthors: (envName?: string) => Promise<Author[]>
   getChangedElementsByAuthors: (authors: Author[], envName?: string) => Promise<ElemID[]>
   getElementNaclFiles: (id: ElemID) => Promise<string[]>
+  // this API can be removed when we do the fragment refactor (SALTO-2217)
+  createPathIndexForElement: (id: ElemID) => Promise<PathIndex>
```
ori-moisis, line 198: I wonder if this is just a helper function, does it really need to be defined in the main workspace interface?
maybe it would be easier to put it somewhere on the side and export it separately?

also, is this really all the logic we want to share?
IIUC in all the use cases where this is relevant we want to get the path index for an element, if it already exists in the state we want to use it, and generating it from the files is just a fallback for when it isn't in the state.
I expected that we'd share the full logic that also deals with the case where the element's path index exists in the state, not just this part...

dantal4: I moved it to path_index_fallbacks.ts (couldn't put it in path_index.ts, because of dependency cycle).

re sharing more logic here - I'm not sure that it will give us the result we want to in [here](https://github.com/salto-io/salto_private/pull/6660) -  because we're there interested in all path index entries for an element, including the entries for nested ids in it, and we can't get them without first reading the whole path index (which is not the original purpose of the index == reading only necessary entries, by their elemId.

ori-moisis: I think there are ways to share that logic but this is not important enough, if you think this is sufficient logic sharing then this is good enough and we'll just deal with it when we hopefully someday finally implement the fragment refactor

ori-moisis (approved):

PR #4956, opened 2023-10-05
SALTO-4658: add important values to zendesk types

edenhassid:
add important values to zendesk types

---

---
_Release Notes_: 
zendesk:
* add important values annotation to types

---
_User Notifications_: 
None


packages/zendesk-adapter/src/filters/add_important_values.ts
```diff
@@ -0,0 +1,571 @@
...
+      value: 'sorting',
+      highlighted: false,
+      indexed: true,
+    },
+  ],
+}
+
+const filterCreator: FilterCreator = () => ({
+  name: 'addImportantValues',
+  onFetch: async (elements: Element[]): Promise<void> => {
```
netama, line 559: what's the plan around generalizing? can we later move this filter to adapter-components?

packages/zendesk-adapter/src/filters/add_important_values.ts
```diff
@@ -0,0 +1,571 @@
...
+    const objectTypes = elements.filter(isObjectType)
+    objectTypes.forEach(obj => {
+      const { typeName } = obj.elemID
+      const importantValuesArray = importantValuesMap[typeName]
+      if (_.isArray(importantValuesArray)) {
```
netama, line 564: nit: we prefer builtins when available (`Array.isArray`)

netama (approved):
didn't compare the cases to the table, but looks good code-wise :)

packages/zendesk-adapter/src/filters/add_important_values.ts
edenhassid: discussed IRL, we will consider in the future if we want to move it to adapter-components

PR #4957, opened 2023-10-05
SALTO-4727: Removed the metadataQuery preparation code

tamtamirr:
After discussing with @ori-moisis, I've realised it' be better to remove the preparation logic for **fetchWithChangesDetection** mode that was previously implemented within the **metadataQuery**.

---

**Please ignore the changes in tests.**
This is a refactor. No tests were broken.
I've created a different **fetchProfile** creator that is specific for **fetchWithChangesDetection** mode.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/adapter.ts
```diff
@@ -444,8 +439,16 @@ export default class SalesforceAdapter implements AdapterOperations {
...
   async fetch({ progressReporter, withChangesDetection = false }: FetchOptions): Promise<FetchResult> {
+    const fetchProfile = await this.fetchProfilePromise
+    if (!fetchProfile.isFeatureEnabled('fetchCustomObjectUsingRetrieveApi')) {
+      // We have to fetch custom objects using retrieve in order to be able to fetch the field-level permissions
+      // in profiles. If custom objects are fetched via the read API, we have to fetch profiles using that API too.
+      _.pull(this.metadataToRetrieve, CUSTOM_OBJECT, PROFILE_METADATA_TYPE)
+    }
+    if (fetchProfile.isFeatureEnabled('fetchProfilesUsingReadApi')) {
+      _.pull(this.metadataToRetrieve, PROFILE_METADATA_TYPE)
+    }
```
tamtamirr, line 450: was moved from the constructor. Any objections @tomsellek ?

tomsellek: Hmmm... I am concerned about having `this.metadataToRetrieve` be almost-but-not-quite-correct between the the time the object is constructed and `fetch` being called. What are the benefits of moving this code?

amitgoldarb (approved):
LGTM.
Let's wait for @tomsellek 's approval as well.

packages/salesforce-adapter/src/adapter.ts
tamtamirr: SInce building the `fetchProfile` is async now (as we receive a promise) using it requires execution of async code. It can't really sit under the constructor here.

packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
```diff
@@ -99,21 +97,19 @@ const isFolderMetadataTypeNameMatch = ({ name: instanceName }: MetadataInstance,
 type BuildMetadataQueryParams = {
   metadataParams: MetadataParams
   target?: string[]
-  elementsSource: ReadOnlyElementsSource
   isFetchWithChangesDetection: boolean
+  changedAtSingleton?: InstanceElement
 }
 
 export const buildMetadataQuery = ({
```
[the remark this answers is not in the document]
tamtamirr: In the case of `MetadataQuery` I think this separation will create unnecessary complexity (they will have to share most of the code). Moreover, note we already have three fetch modes that all affect the fetch (targeted fetch, regular and quick). IMO they must all sit in the same `MetadataQuery` as they can run independently or together (for example you can run a **quick targeted fetch**). Would you imagine having three different `MetadataQueries`?

packages/salesforce-adapter/src/adapter.ts
tomsellek: Makes sense. 
I think one thing we can do, since nobody really passes a non-default value to the `metadataToRetrieve` c'tor param is to move it to be a parameter to `fetch`. This would enforce the requirement that this information should only be used in `fetch`.
I really hate the idea of having a member variable that has a value but that value is almost-correct until a specific method is called. There's a real risk that somebody will end up using this variable somewhere and introduce a really subtle bug.

packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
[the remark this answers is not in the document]
tomsellek: Fair point :)

PR #4959, opened 2023-10-10
SALTO-4869 ignore hidden values on calculatePatch

dantal4:
In the `calculatePatch` & `calculatePatchFromChanges` APIs we currently return removal changes on hidden values-

If the patch includes an addition of an element (without hidden values in it, since it’s coming from PR/SAAS change) and it is already exists in the workspace, we’ll have removal changes on the hidden fields/annotations of the element- like `internalId`, `_service_url` etc.

The reason is that we’re calling `calcFetchChanges` with `workspace.elements()` for the `workspaceElements` param- which includes the hidden values from the state in it. The problem is that for `stateElements`&`accountElements` we’re passing the PR’s elements, that doesn’t have hidden values in them- and the result is having removal changes on the hidden values.

IMO we should call `calcFetchChanges` with `workspace.elements({ includeHidden: false })` for the `workspaceElements`, as we’re actually interested in calculating the changes between the source PR and the visible elements in the workspace.

---

---
_Release Notes_: 
Core:
- Ignore hidden values on calculatePatch

---
_User Notifications_: 
None

### DIFF
--- a/packages/core/src/api.ts
+++ b/packages/core/src/api.ts
@@ -390,7 +390,7 @@ export const calculatePatchFromChanges = async (
     unmergedAfterElements,
     afterElements,
     elementSource.createInMemoryElementSource(beforeElements),
-    await workspace.elements(),
+    await workspace.elements(false),
     partialFetchData,
     new Set(accounts),
   )
@@ -487,7 +487,7 @@ export const calculatePatch = async (
     afterElements,
     mergedAfterElements,
     elementSource.createInMemoryElementSource(mergedBeforeElements),
-    adapterContext.elementsSource,
+    await workspace.elements(false),
     new Map([[accountName, {}]]),
     new Set([accountName]),
   )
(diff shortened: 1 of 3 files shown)


packages/core/src/api.ts
```diff
@@ -487,7 +487,7 @@ export const calculatePatch = async (
     afterElements,
     mergedAfterElements,
     elementSource.createInMemoryElementSource(mergedBeforeElements),
-    adapterContext.elementsSource,
```
ori-moisis, line 490: this feels wrong to use a different element source here than the one that the adapters use to create their response.
are you sure this change is needed? what is the justification for using different elements here?

I don't have a specific use case of why this would break something, just a general intuition that this would create conflicting definitions, so if you feel strongly that this change is correct you can move forward with this

ori-moisis (approved):
maybe add a test or two to have some example of the behavior you want?

packages/core/src/api.ts
dantal4: The adapters needs the a fully resolved elements source with their hidden values too, but here we should get it without hidden values. IDK if there's another way to do it (except for removing the hidden values from each element in runtime which will be much slower).

PR #4961, opened 2023-10-12
SALTO-4881 - Salesforce: Support more complex SOQL queries

tomsellek:
Make it possible to run range-based queries in addition to exact-match queries. If the query is too long and has to be broken into multiple queries, make sure the range-based queries are appended to every query that is created, to maintain the expected behavior.

---

**PR Chain**:
 - https://github.com/salto-io/salto/pull/4961 👈 YOU ARE HERE 👈 
 - https://github.com/salto-io/salto/pull/4964
 - https://github.com/salto-io/salto/pull/4965

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/filters/utils.ts
+++ b/packages/salesforce-adapter/src/filters/utils.ts
@@ -420,23 +420,37 @@ export const extractFlatCustomObjectFields = async (elem: Element): Promise<Elem
     : [elem]
 )
 
-export const getWhereConditions = (
-  conditionSets: Record<string, string>[],
-  maxLen: number
+type QueryOperator = '>' | '<'
+type ComplexQuery = [string, QueryOperator, string]
+
+const getWhereConditions = (
+  exactConditionSets: Record<string, string>[],
+  complexConditionSets: ReadonlyArray<ComplexQuery>,
+  maxLen: number,
 ): string[] => {
-  const keys = _.uniq(conditionSets.flatMap(Object.keys))
+  const keys = _.uniq(exactConditionSets.flatMap(Object.keys))
   const constConditionPartLen = (
     _.sumBy(keys, key => `${key} IN ()`.length)
     + (' AND '.length * (keys.length - 1))
   )
 
+  const complexConditionQuery = complexConditionSets
+    .map(([field, op, value]) => `${field} ${op} ${value}`)
+    .join(' AND ')
+  const complexConditionsLength = complexConditionQuery.length
+    + ((Object.keys(exactConditionSets).length > 0 && complexConditionQuery.length > 0) ? ' AND '.length : 0)
+
+  if (complexConditionsLength > maxLen) {
+    throw new Error('Complex conditions need to be included in every query, thus can`t be longer than the maximum query length')
+  }
+
   const conditionChunks = weightedChunks(
-    conditionSets,
+    exactConditionSets,
     maxLen - constConditionPartLen,
     // Note - this calculates the condition length as if all values are added to the query.
     // the actual query might end up being shorter if some of the values are not unique.
     // this can be optimized in the future if needed
-    condition => _.sumBy(Object.values(condition), val => `${val},`.length),
+    condition => _.sumBy(Object.values(condition), val => `${val},`.length) + complexConditionsLength,
   )
   const r = conditionChunks.map(conditionChunk => {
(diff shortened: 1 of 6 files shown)


packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -375,23 +375,37 @@ export const extractFlatCustomObjectFields = async (elem: Element): Promise<Elem
...
   )
 
+  const complexConditionQuery = complexConditionSets
+    .map(([field, op, value]) => `${field} ${op} ${value}`)
+    .join(' AND ')
+  const complexConditionsLength = complexConditionQuery.length
+    + ((Object.keys(exactConditionSets).length > 0 && complexConditionQuery.length > 0) ? ' AND '.length : 0)
+
+  if (complexConditionsLength > maxLen) {
+    throw new Error('Complex conditions need to be included in every query, thus can`t be longer than the maximum query length')
```
tamtamirr, line 399:
```suggestion
    log.error('the WHERE part of the built query exceeds the max query length. getWhereConditions params: %o', { exactConditionSets, complexConditionSets, maxLen}')
    throw new Error('Built SOQL query too long')
```
If we reached this flow, I agree this should be an unrecoverable error. However, we need to make sure that the visible error to the users does not contain implementation details and needs to be understandable.
On our end, we need info on how we reached this flow as this is most likely coding error, as it's not affected by user input (at-least for now)

packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -404,13 +418,27 @@ export const getWhereConditions = (
...
 }
 
-export const conditionQueries = (query: string, conditionSets: Record<string,
-  string>[], maxQueryLen = MAX_QUERY_LENGTH): string[] => {
+export const conditionQueries = (
+  query: string,
+  exactConditionSets: Record<string, string>[],
+  complexConditionSets: ReadonlyArray<ComplexQuery> = [],
+  maxQueryLen = MAX_QUERY_LENGTH,
+): string[] => {
```
tamtamirr, line 432: Consider refactoring to a single object param. (Shouldn't be too hard considering it's only used in tests and one filter)

tamtamirr (commented):
Added some key comments around the design and the visible errors.

packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -375,23 +375,37 @@ export const extractFlatCustomObjectFields = async (elem: Element): Promise<Elem
...
 
-export const getWhereConditions = (
-  conditionSets: Record<string, string>[],
-  maxLen: number
+type QueryOperator = '>' | '<'
+type ComplexQuery = [string, QueryOperator, string]
+
+const getWhereConditions = (
+  exactConditionSets: Record<string, string>[],
+  complexConditionSets: ReadonlyArray<ComplexQuery>,
```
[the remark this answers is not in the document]
tomsellek: "exact" and "complex" conditions behave differently when we have to break the query apart (if it's too long). We can break exact conditions apart (the results of `IN (1,2,3)` are the same as the concatenation of the results of `IN (1)` + `IN (2)` + `IN (3)`, but this is not the case for so-called "complex" conditions.
Perhaps it would be better to name "complex" conditions to "limiting" conditions? or "range" conditions?

packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -375,23 +375,37 @@ export const extractFlatCustomObjectFields = async (elem: Element): Promise<Elem
     : [elem]
 )
 
-export const getWhereConditions = (
-  conditionSets: Record<string, string>[],
-  maxLen: number
+type QueryOperator = '>' | '<'
+type ComplexQuery = [string, QueryOperator, string]
```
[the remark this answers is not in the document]
tomsellek: Yeah, makes sense to me.

packages/salesforce-adapter/src/filters/utils.ts
tomsellek: I agree it wouldn't be too hard, but I think the way the current code (including the upcoming PRs) use this function, we won't gain much by the added verbosity at the call site.

tamtamirr (approved):

PR #4964, opened 2023-10-13
SALTO-4839 - Salesforce: Aggregate changed-at for instances

tomsellek:
Add a new "section" to the changed-at Singleton where we store the most recent "changed-at" among all instances of a given type

---

 - I chose to create a new "section" as the least-disruptive way of storing the data. Feedback welcome!
 - I added the `isInstanceOfCustomObjectSync` utility. Let me know if it's better added in a separate PR.
 - I added a couple of utility functions for use with the changed-at Singleton: `getTypeChangedAt` and `getTypeInstancesChangedAt`.

**PR Chain**:
 - https://github.com/salto-io/salto/pull/5071
 - https://github.com/salto-io/salto/pull/4964 👈 YOU ARE HERE 👈 
 - https://github.com/salto-io/salto/pull/4965

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/constants.ts
+++ b/packages/salesforce-adapter/src/constants.ts
@@ -428,7 +428,7 @@ export const KEY_PREFIX_LENGTH = 3
 
 // Magics
 export const DETECTS_PARENTS_INDICATOR = '##allMasterDetailFields##'
-export const LAST_MODIFIED_INSTANCES = '__Instances__'
+export const DATA_INSTANCES_CHANGED_AT_MAGIC = '__DataInstances__'
 
 // CPQ CustomObjects
 export const CPQ_NAMESPACE = 'SBQQ'
--- a/packages/salesforce-adapter/src/constants.ts
+++ b/packages/salesforce-adapter/src/constants.ts
@@ -428,6 +428,7 @@ export const KEY_PREFIX_LENGTH = 3
 
 // Magics
 export const DETECTS_PARENTS_INDICATOR = '##allMasterDetailFields##'
+export const LAST_MODIFIED_INSTANCES = '__Instances__'
 
 // CPQ CustomObjects
 export const CPQ_NAMESPACE = 'SBQQ'
--- a/packages/salesforce-adapter/src/filters/changed_at_singleton.ts
+++ b/packages/salesforce-adapter/src/filters/changed_at_singleton.ts
@@ -25,7 +25,6 @@ import _ from 'lodash'
 import { LocalFilterCreator } from '../filter'
 import {
   ArtificialTypes,
-  CUSTOM_OBJECT,
   DATA_INSTANCES_CHANGED_AT_MAGIC,
 } from '../constants'
 import {
@@ -99,7 +98,7 @@ const filterCreator: LocalFilterCreator = ({ config }) => ({
     instanceLastChangedByCustomObjectType
       .entries()
       .forEach(([typeName, dateString]) => {
-        _.set(changedAtInstance.value, [DATA_INSTANCES_CHANGED_AT_MAGIC, CUSTOM_OBJECT, typeName], dateString)
+        _.set(changedAtInstance.value, [DATA_INSTANCES_CHANGED_AT_MAGIC, typeName], dateString)
       })
   },
 })
--- a/packages/salesforce-adapter/src/filters/changed_at_singleton.ts
+++ b/packages/salesforce-adapter/src/filters/changed_at_singleton.ts
@@ -63,11 +63,11 @@ const getChangedAtSingletonInstance = async (
   return changedAtSingleton ?? createEmptyChangedAtSingletonInstance()
 }
 
-const dateStringOfMostRecentlyChangedInstance = (instances: InstanceElement[]): string => (
+const dateStringOfMostRecentlyChangedInstance = (instances: InstanceElement[]): string | undefined => (
   _(instances)
     .map(instance => instance.annotations[CORE_ANNOTATIONS.CHANGED_AT])
-    .filter(changedAt => changedAt !== undefined)
-    .maxBy((changedAt: string) => new Date(changedAt).getTime())
+    .filter(_.isString)
+    .maxBy(changedAt => new Date(changedAt).getTime())
 )
 
 const filterCreator: LocalFilterCreator = ({ config }) => ({
@@ -97,8 +97,9 @@ const filterCreator: LocalFilterCreator = ({ config }) => ({
 
     instanceLastChangedByCustomObjectType
       .entries()
-      .forEach(([typeName, dateString]) => {
-        _.set(changedAtInstance.value, [DATA_INSTANCES_CHANGED_AT_MAGIC, typeName], dateString)
+      .filter(([, mostRecentChangedAt]) => mostRecentChangedAt !== undefined)
+      .forEach(([typeName, mostRecentChangedAt]) => {
+        _.set(changedAtInstance.value, [DATA_INSTANCES_CHANGED_AT_MAGIC, typeName], mostRecentChangedAt)
       })
   },
 })
--- a/packages/salesforce-adapter/src/filters/changed_at_singleton.ts
+++ b/packages/salesforce-adapter/src/filters/changed_at_singleton.ts
@@ -23,12 +23,17 @@ import {
 } from '@salto-io/adapter-api'
 import _ from 'lodash'
 import { LocalFilterCreator } from '../filter'
-import { ArtificialTypes } from '../constants'
+import {
+  ArtificialTypes,
+  CUSTOM_OBJECT,
+  LAST_MODIFIED_INSTANCES,
+} from '../constants'
 import {
   apiNameSync,
   getChangedAtSingleton,
   isCustomObjectSync,
   isMetadataInstanceElementSync,
+  isInstanceOfCustomObjectSync,
   metadataTypeSync,
 } from './utils'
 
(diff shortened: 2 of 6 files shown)


packages/salesforce-adapter/src/filters/changed_at_singleton.ts
```diff
@@ -59,6 +64,12 @@ const getChangedAtSingletonInstance = async (
   return changedAtSingleton ?? createEmptyChangedAtSingletonInstance()
 }
 
+const dateStringOfMostRecentlyChangedInstance = (instances: InstanceElement[]): string => (
+  _(instances)
+    .map(instance => instance.annotations[CORE_ANNOTATIONS.CHANGED_AT])
+    .maxBy((changedAt: string) => new Date(changedAt).getTime())
```
tamtamirr, line 70:
```suggestion
  _(instances)
    .map(instance => instance.annotations[CORE_ANNOTATIONS.CHANGED_AT])
    .filter(changedAt => changedAt !== undefined && changedAt !== '')
    .maxBy((changedAt: string) => new Date(changedAt).getTime())
```
Let's filter empty or undefined values, to avoid problem with creating a `javascript Date` object from them.

packages/salesforce-adapter/src/constants.ts
```diff
@@ -406,6 +406,7 @@ export const KEY_PREFIX_LENGTH = 3
 
 // Magics
 export const DETECTS_PARENTS_INDICATOR = '##allMasterDetailFields##'
+export const LAST_MODIFIED_INSTANCES = '__Instances__'
```
tamtamirr, line 409:
```suggestion
export const DATA_INSTANCES_MAGIC = '__DataInstances__'
```

packages/salesforce-adapter/src/filters/changed_at_singleton.ts
```diff
@@ -78,6 +89,17 @@ const filterCreator: LocalFilterCreator = ({ config }) => ({
       createChangedAtSingletonInstanceValues(elementsByType),
       changedAtInstance.value,
     )
+
+    const instanceLastUpdateByType = _(elements)
```
tamtamirr, line 93:
```suggestion
    const lastChangedByCustomObjectType = _(elements)
```

tamtamirr (commented):
LGTM! Added some small comments and suggestions :)

packages/salesforce-adapter/src/filters/changed_at_singleton.ts
tomsellek: Ack

packages/salesforce-adapter/src/constants.ts
tomsellek: Ack-ish.

packages/salesforce-adapter/src/filters/changed_at_singleton.ts
tomsellek: Ack-ish. I wanted to still make it clear we're dealing with instances.

tamtamirr (commented):
Great work! Added some comments :)

packages/salesforce-adapter/src/filters/changed_at_singleton.ts
```diff
@@ -78,6 +90,17 @@ const filterCreator: LocalFilterCreator = ({ config }) => ({
...
+
+    const instanceLastChangedByCustomObjectType = _(elements)
+      .filter(isInstanceOfCustomObjectSync)
+      .groupBy(instance => apiNameSync(instance.getTypeSync()))
+      .mapValues(dateStringOfMostRecentlyChangedInstance)
+
+    instanceLastChangedByCustomObjectType
+      .entries()
+      .forEach(([typeName, dateString]) => {
+        _.set(changedAtInstance.value, [DATA_INSTANCES_CHANGED_AT_MAGIC, CUSTOM_OBJECT, typeName], dateString)
```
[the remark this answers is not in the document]
tomsellek: Ack.

packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -526,6 +566,23 @@ export const getChangedAtSingleton = async (
   return isInstanceElement(element) ? element : undefined
 }
 
+export const getTypeChangedAt = (
+  changedAtSingleton: InstanceElement,
+  metadataTypeName: string,
+  typeName: string,
+): string | undefined => (
+  _.get(changedAtSingleton.value, [metadataTypeName, typeName])
+)
```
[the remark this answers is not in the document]
tomsellek: nvm, I just deleted it.

packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -134,6 +135,9 @@ export const isCustomObjectSync = (element: Readonly<Element>): element is Objec
   return res
 }
 
+export const isInstanceOfCustomObjectSync = (element: Readonly<Element>): element is InstanceElement => (
```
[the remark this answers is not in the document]
tomsellek: 1. I actually think it makes sense to have Sync/regular versions of the same function because the Sync version can only be used in specific situations (when types are resolved, like in fetch) while the regular versions are not limited - so I'd want to make it easy to find the unrestricted version from the restricted version and vice versa.
2. Ack on the deprecation

packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -526,6 +566,23 @@ export const getChangedAtSingleton = async (
...
+  _.get(changedAtSingleton.value, [metadataTypeName, typeName])
+)
+
+export const getTypeInstancesChangedAt = (
+  changedAtSingleton: InstanceElement,
+  metadataTypeName: string,
+  typeName: string,
+): string | undefined => (
+  _.get(changedAtSingleton.value, [DATA_INSTANCES_CHANGED_AT_MAGIC, metadataTypeName, typeName])
+)
```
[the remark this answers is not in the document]
tomsellek: removed

packages/salesforce-adapter/src/transformers/transformer.ts
```diff
@@ -121,6 +121,9 @@ export const isFieldOfCustomObject = async (field: Field): Promise<boolean> =>
 // for instances of Lead, but it will not be true for Lead itself when it is still an instance
 // (before the custom objects filter turns it into a type).
 // To filter for instances like the Lead definition, use isInstanceOfType(CUSTOM_OBJECT) instead
+/**
+ * @deprecated use {@link isInstanceOfCustomObjectSync} in fetch flows.
+*/
```
tamtamirr, line 126: Types are also resolved during deploy no? At-least shallowly. (Otherwise we'd have to call `apiName` with elements source which is not the case)

tamtamirr (commented):
LGTM! Added a small comment

packages/salesforce-adapter/src/transformers/transformer.ts
tomsellek: Fair.

packages/salesforce-adapter/src/filters/changed_at_singleton.ts
```diff
@@ -59,6 +63,13 @@ const getChangedAtSingletonInstance = async (
   return changedAtSingleton ?? createEmptyChangedAtSingletonInstance()
 }
 
+const dateStringOfMostRecentlyChangedInstance = (instances: InstanceElement[]): string => (
+  _(instances)
+    .map(instance => instance.annotations[CORE_ANNOTATIONS.CHANGED_AT])
+    .filter(changedAt => changedAt !== undefined)
+    .maxBy((changedAt: string) => new Date(changedAt).getTime())
+)
```
tamtamirr, line 71: What if no element has `_changed_at` value? The return type is incorrect and since we iterate any values from `L68` TS didn't catch that.

```suggestion
const dateStringOfMostRecentlyChangedInstance = (instances: InstanceElement[]): string | undefined => (
  _(instances)
    .map(instance => instance.annotations[CORE_ANNOTATIONS.CHANGED_AT])
    .filter(_.isString)
    .filter(changedAt => changedAt !== '')
    .maxBy(changedAt => new Date(changedAt).getTime())
)
```

Please also note we should make sure we don't deal with empty string values (which in some flows of the adapter are what Salesforce returns, specifically for Metadata Instances, which is not the use-case here, but since the received argument is instances we should be bulletproof there)

Consider exporting to a utility `isValidChangedAt(value: unknown): value is string`

packages/salesforce-adapter/src/filters/changed_at_singleton.ts
```diff
@@ -78,6 +89,17 @@ const filterCreator: LocalFilterCreator = ({ config }) => ({
...
     )
+
+    const instanceLastChangedByCustomObjectType = _(elements)
+      .filter(isInstanceOfCustomObjectSync)
+      .groupBy(instance => apiNameSync(instance.getTypeSync()))
+      .mapValues(dateStringOfMostRecentlyChangedInstance)
+
+    instanceLastChangedByCustomObjectType
+      .entries()
+      .forEach(([typeName, dateString]) => {
```
tamtamirr, line 100:
```suggestion
      .forEach(([typeName, mostRecentChangedAt]) => {
```
nit

packages/salesforce-adapter/src/filters/changed_at_singleton.ts
```diff
@@ -78,6 +89,17 @@ const filterCreator: LocalFilterCreator = ({ config }) => ({
       createChangedAtSingletonInstanceValues(elementsByType),
       changedAtInstance.value,
     )
+
+    const instanceLastChangedByCustomObjectType = _(elements)
+      .filter(isInstanceOfCustomObjectSync)
+      .groupBy(instance => apiNameSync(instance.getTypeSync()))
+      .mapValues(dateStringOfMostRecentlyChangedInstance)
+
+    instanceLastChangedByCustomObjectType
```
tamtamirr, line 98: We should filter out entries where `most recent changed at` is undefined (even though it's probably not a use-case for data instances)

tamtamirr (approved):
A single issue due to `any` values 🙏 approving to save cycles

packages/salesforce-adapter/src/filters/changed_at_singleton.ts
tomsellek: Ack

packages/salesforce-adapter/src/filters/changed_at_singleton.ts
tomsellek: Ack

PR #4967, opened 2023-10-15
SALTO-4852: ProfileMapping instances improvements in Okta

shir-reifenberg:

---

Changes:
- Change ProfileMapping idFields to be dependent on source and target ids instead of name, as name is not unique
- Add missing references from ProfileMapping to IdentityProvider Instances
- Omit ProfileMapping instances that maps from Okta_Authenticator app to user profiles. Okta_Autheticator is an internal Okta app that isn't visible in Okta's UI and cannot be configured

---
_Release Notes_: 

_Okta Adapter_:
- Improvements to ProfileMapping instances

---
_User Notifications_: 

_Okta Adapter_:
- References will be added to ProfileMapping instances
- Omit ProfileMapping instances that maps to internal Okta apps


packages/okta-adapter/src/filters/omit_authenticator_mapping.ts
```diff
@@ -0,0 +1,44 @@
...
+* Okta_Authenticator is an internal Okta app and its mapping cannot be managed
+*/
+const filterCreator: FilterCreator = () => ({
+  name: 'omitAuthenticatorMappingFilter',
+  onFetch: async (elements: Element[]) => {
+    const instancesToOmit = elements
+      .filter(isInstanceElement)
+      .filter(instance => instance.elemID.typeName === PROFILE_MAPPING_TYPE_NAME)
+      .filter(isMappingToAuthenticatorApp)
+    _.pullAll(elements, instancesToOmit)
```
netama, line 40: nit: do we want to log the omitted elem ids?

netama (approved):
🎈

packages/okta-adapter/src/filters/omit_authenticator_mapping.ts
shir-reifenberg: yes why not

PR #4968, opened 2023-10-15
SALTO-4872: workflow transition changes are not deployed 

alonstern:
Added missing annotation to workflow

---
_Release Notes_: 
__Jira Adapter__:
- Fixed a bug where workflow transition configuration would not be deployed

---
_User Notifications_: 
None


shir-reifenberg (approved):
LGTM

PR #4969, opened 2023-10-15
SALTO-4850: Support deploy of request Type 

idoamit8:
In this PR I support deploy of request type. 

---

Request types have several endpoints:

1. For request forms, which also include request type name, description, and help text.
2. For the issue view.
3. For workflow statuses.
4. For groups.

Please review only from commit number 3 - 28b73c4efd87e8cbfe0d2371dc0df208a9c37892. 

---
_Release Notes_: 
_Jira Adapter:_
We are now providing support for the deployment of request types.

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
+++ b/packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
@@ -36,6 +36,7 @@ type layoutTypeDetails = {
     pathParam: string
     query: string
     layoutType?: string
+    fieldName?: string
 }
 
 type QueryVariables = {
@@ -44,12 +45,13 @@ type QueryVariables = {
     layoutType?: string
   }
 
-type LayoutTypeName = 'RequestForm' | 'IssueView' | 'IssueLayout'
-const layoutTypeNameToDetails: Record<LayoutTypeName, layoutTypeDetails> = {
+export type LayoutTypeName = 'RequestForm' | 'IssueView' | 'IssueLayout'
+export const layoutTypeNameToDetails: Record<LayoutTypeName, layoutTypeDetails> = {
   [REQUEST_FORM_TYPE]: {
     layoutType: 'REQUEST_FORM',
     pathParam: 'RequestForm',
     query: QUERY_JSM,
+    fieldName: 'requestForm',
   },
   [ISSUE_LAYOUT_TYPE]: {
     pathParam: 'layouts',
@@ -59,10 +61,11 @@ const layoutTypeNameToDetails: Record<LayoutTypeName, layoutTypeDetails> = {
     layoutType: 'ISSUE_VIEW',
     pathParam: 'IssueView',
     query: QUERY_JSM,
+    fieldName: 'issueView',
   },
 }
 
-const isIssueLayoutResponse = createSchemeGuard<IssueLayoutResponse>(ISSUE_LAYOUT_RESPONSE_SCHEME)
+export const isIssueLayoutResponse = createSchemeGuard<IssueLayoutResponse>(ISSUE_LAYOUT_RESPONSE_SCHEME)
 const isLayoutConfigItem = createSchemeGuard<layoutConfigItem>(ISSUE_LAYOUT_CONFIG_ITEM_SCHEME)
 
 function isLayoutTypeName(typeName: string): typeName is LayoutTypeName {
--- a/packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
+++ b/packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
@@ -32,11 +32,11 @@ import { ISSUE_LAYOUT_TYPE, ISSUE_VIEW_TYPE, JIRA, REQUEST_FORM_TYPE, REQUEST_TY
 const log = logger(module)
 const { isDefined } = lowerDashValues
 
-type layoutTypeDetails = {
+type LayoutTypeDetails = {
     pathParam: string
     query: string
     layoutType?: string
-    fieldName?: string
+    fieldName: string
 }
 
 type QueryVariables = {
@@ -46,7 +46,7 @@ type QueryVariables = {
   }
 
 export type LayoutTypeName = 'RequestForm' | 'IssueView' | 'IssueLayout'
-export const layoutTypeNameToDetails: Record<LayoutTypeName, layoutTypeDetails> = {
+export const LAYOUT_TYPE_NAME_TO_DETAILS: Record<LayoutTypeName, LayoutTypeDetails> = {
   [REQUEST_FORM_TYPE]: {
     layoutType: 'REQUEST_FORM',
     pathParam: 'RequestForm',
@@ -56,6 +56,7 @@ export const layoutTypeNameToDetails: Record<LayoutTypeName, layoutTypeDetails> 
   [ISSUE_LAYOUT_TYPE]: {
     pathParam: 'layouts',
     query: QUERY,
+    fieldName: 'issueLayout',
   },
   [ISSUE_VIEW_TYPE]: {
     layoutType: 'ISSUE_VIEW',
@@ -69,7 +70,7 @@ export const isIssueLayoutResponse = createSchemeGuard<IssueLayoutResponse>(ISSU
 const isLayoutConfigItem = createSchemeGuard<layoutConfigItem>(ISSUE_LAYOUT_CONFIG_ITEM_SCHEME)
 
 function isLayoutTypeName(typeName: string): typeName is LayoutTypeName {
-  return Object.keys(layoutTypeNameToDetails).includes(typeName)
+  return Object.keys(LAYOUT_TYPE_NAME_TO_DETAILS).includes(typeName)
 }
 
 export const getLayoutResponse = async ({
@@ -83,7 +84,7 @@ export const getLayoutResponse = async ({
     }): Promise<graphQLResponseType> => {
   const baseUrl = '/rest/gira/1'
   try {
-    const query = layoutTypeNameToDetails[typeName]?.query
+    const query = LAYOUT_TYPE_NAME_TO_DETAILS[typeName]?.query
     if (query === undefined) {
       log.error(`Failed to get issue layout for project ${variables.projectId} and screen ${variables.extraDefinerId}: query is undefined`)
     }
@@ -145,7 +146,7 @@ export const getLayout = async ({
       instanceName,
       layoutType,
       value,
-      [...instance.path.slice(0, -1), layoutTypeNameToDetails[typeName].pathParam, pathNaclCase(instanceName)],
+      [...instance.path.slice(0, -1), LAYOUT_TYPE_NAME_TO_DETAILS[typeName].pathParam, pathNaclCase(instanceName)],
     )
   }
   return undefined
@@ -271,7 +272,7 @@ export const fetchRequestTypeDetails = async ({
       const variables = {
         projectId: projectInstance.value.id,
         extraDefinerId: requestTypeId,
-        layoutType: layoutTypeNameToDetails[typeName].layoutType,
+        layoutType: LAYOUT_TYPE_NAME_TO_DETAILS[typeName].layoutType,
       }
       const response = await getLayoutResponse({
         variables,
--- a/packages/jira-adapter/src/filters/layouts/layout_types.ts
+++ b/packages/jira-adapter/src/filters/layouts/layout_types.ts
@@ -114,7 +114,7 @@ export type layoutConfigItem = {
 
 export const ISSUE_LAYOUT_CONFIG_ITEM_SCHEME = Joi.object({
   type: Joi.string().required(),
-  sectionType: Joi.string().invalid('HIDDEN_ITEMS', 'REQUEST_PORTAL').required(),
+  sectionType: Joi.string().valid('PRIMARY', 'SECONDARY', 'CONTENT', 'REQUEST').required(),
   key: Joi.string().required(),
 }).unknown(true).required()
 
--- a/packages/jira-adapter/src/filters/layouts/layout_types.ts
+++ b/packages/jira-adapter/src/filters/layouts/layout_types.ts
@@ -108,13 +108,13 @@ export type IssueLayoutResponse = {
 
 export type layoutConfigItem = {
   type: string
-  sectionType: 'PRIMARY' | 'SECONDARY' | 'CONTENT' | 'REQUEST' | 'REQUEST_PORTAL'
+  sectionType: 'PRIMARY' | 'SECONDARY' | 'CONTENT' | 'REQUEST'
   key: string
 }
 
 export const ISSUE_LAYOUT_CONFIG_ITEM_SCHEME = Joi.object({
   type: Joi.string().required(),
-  sectionType: Joi.string().invalid('HIDDEN_ITEMS').required(),
+  sectionType: Joi.string().invalid('HIDDEN_ITEMS', 'REQUEST_PORTAL').required(),
   key: Joi.string().required(),
 }).unknown(true).required()
 
--- a/packages/jira-adapter/src/filters/layouts/request_types_layouts_to_values.ts
+++ b/packages/jira-adapter/src/filters/layouts/request_types_layouts_to_values.ts
@@ -58,6 +58,7 @@ const filter: FilterCreator = ({ config }) => ({
       } else {
         requestType.value.issueView = layout.value
       }
+      requestType.value.avatarId = requestType.value.icon.id
     })
   },
 })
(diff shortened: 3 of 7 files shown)


packages/jira-adapter/src/filters/request_type.ts
```diff
@@ -0,0 +1,180 @@
...
+
+const deployWorkflowStatuses = async (
+  change: ModificationChange<InstanceElement> | AdditionChange<InstanceElement>,
+  client: JiraClient): Promise<void> => {
+  if ((isAdditionChange(change))
+    || (isModificationChange(change)
+    && !isEqualValues(change.data.before.value.workflowStatuses, change.data.after.value.workflowStatuses))) {
+    const instance = getChangeData(change)
+    const parent = getParent(instance)
+    const workflowStatuses = instance.value.workflowStatuses.map((status: statusType) => ({
```
netama, line 53: probably worth adding a type guard for `workflowStatuses`?

packages/jira-adapter/src/filters/request_type.ts
```diff
@@ -0,0 +1,180 @@
...
+  }
+}
+
+const isRequesTypeDetailsChange = (
+  change: ModificationChange<InstanceElement>,
+  fieldName: string,
+): boolean => fieldName === 'requestForm'
+  && (change.data.before.value.name !== change.data.after.value.name
+    || change.data.before.value.description !== change.data.after.value.description
+    || change.data.before.value.helpText !== change.data.after.value.helpText)
```
netama, line 71: nit: maybe 
```suggestion
  && ['name', 'description', 'helpText'].some(additionalFieldName => change.data.before.value[additionalFieldName] !==  change.data.after.value[additionalFieldName])
```
(I see you already did something similar below :) )

packages/jira-adapter/src/filters/request_type.ts
```diff
@@ -0,0 +1,180 @@
...
+): Promise<void> => {
+  const requestType = getChangeData(change)
+  const { fieldName } = layoutTypeNameToDetails[typeName]
+  if (fieldName === undefined) {
+    return undefined
+  }
+  if ((isAdditionChange(change))
+  || (isModificationChange(change)
+  && (!isEqualValues(change.data.before.value[fieldName], change.data.after.value[fieldName])
+  || isRequesTypeDetailsChange(change, fieldName)))) {
```
netama, line 87: readability nit - especially in complex conditions with and+or at different levels, indentation can really help :) 
```suggestion
  if (
    isAdditionChange(change)
    || (isModificationChange(change)
      && (
        !isEqualValues(change.data.before.value[fieldName], change.data.after.value[fieldName])
        || isRequesTypeDetailsChange(change, fieldName)
      ))
  ) {
```
(or some other alternative, but clarifying where a compound condition begins and where it ends)

packages/jira-adapter/src/filters/request_type.ts
```diff
@@ -0,0 +1,180 @@
...
+          type: item.type,
+          sectionType: item.sectionType.toLocaleLowerCase(),
+          key,
+          data: {
+            name: item.key.value.value.name,
+            type: item.key.value.value.type ?? item.key.value.value.schema.system,
+          },
+        }
+      }
+      return undefined
```
netama, line 102: 1. nit: better to reverse the conditions for clarity + less nested code
```suggestion
      if (!isResolvedReferenceExpression(item.key)) {
        return undefined
      }
      const key = item.key.value.value.id
      return {
        type: item.type,
        sectionType: item.sectionType.toLocaleLowerCase(),
        key,
        data: {
          name: item.key.value.value.name,
          type: item.key.value.value.type ?? item.key.value.value.schema.system,
        },
      }
```
2. when do we expect the reference to not be resolved? if it's an error flow let's add an error log (or warning/info/debug if this is a valid scenario)
3. is there a valid scenario where some items are valid and some are not? is there any risk in deploying only some of the items the user requested?

packages/jira-adapter/src/filters/request_type.ts
```diff
@@ -0,0 +1,180 @@
...
+    }).filter(isDefined)
+
+    const data = {
+      projectId: getParent(requestType).value.id,
+      extraDefinerId: requestType.value.id,
+      issueLayoutType: layoutTypeNameToDetails[typeName].layoutType,
```
netama, line 108: seems this is re-implemented below? can we reuse / pick the relevant values?
(logically it might make more sense to first intiialize the "variables" (which maybe we can call something like `context`)?) - and then only create the full request when we're planning to use it, and include them in it?)

packages/jira-adapter/src/filters/request_type.ts
```diff
@@ -0,0 +1,180 @@
...
+    }
+    const url = `/rest/internal/1.0/issueLayouts/${layout.id}`
+    await client.put({ url, data })
+  }
+
+  return undefined
+}
+
+
+const filter: FilterCreator = ({ config, client }) => ({
```
netama, line 145: please add a docstring

netama (commented):
looks good overall, added some comments and questions

packages/jira-adapter/src/filters/layouts/layout_types.ts
```diff
@@ -108,13 +108,13 @@ export type IssueLayoutResponse = {
 
 export type layoutConfigItem = {
   type: string
-  sectionType: 'PRIMARY' | 'SECONDARY' | 'CONTENT' | 'REQUEST' | 'REQUEST_PORTAL'
+  sectionType: 'PRIMARY' | 'SECONDARY' | 'CONTENT' | 'REQUEST'
```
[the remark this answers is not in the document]
idoamit8: 1. I Will make it consistent. 
2. This was changed because I found out there is a duplication. There are some items that are in the request form and issueView points at them and their section type called `request_portal`

packages/jira-adapter/src/filters/layouts/request_types_layouts_to_values.ts
```diff
@@ -58,6 +58,7 @@ const filter: FilterCreator = ({ config }) => ({
       } else {
         requestType.value.issueView = layout.value
       }
+      requestType.value.avatarId = requestType.value.icon.id
```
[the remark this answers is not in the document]
idoamit8: you can't deploy modification change without it. Yes it is multienv friendly.

packages/jira-adapter/src/filters/request_type.ts
idoamit8: Its not suppose to happen. It just in case. I will add a error log

packages/jira-adapter/src/filters/request_type.ts
idoamit8: not quite sure I understand

packages/jira-adapter/src/filters/request_type.ts
```diff
@@ -0,0 +1,209 @@
...
+        getChangeData(change).elemID.typeName === REQUEST_TYPE_NAME
+        && isInstanceChange(change)
+    )
+
+    const deployResult = await deployChanges(
+      relevantChanges,
+      async change => {
+        if (isAdditionOrModificationChange(change)) {
+          if (isAdditionChange(change)) {
+            await defaultDeployChange({
```
netama, line 186: we never need this part on modifications?

packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
```diff
@@ -36,6 +36,7 @@ type layoutTypeDetails = {
     pathParam: string
```
netama, line 36: existing code, but should capitalize the type name to `LayoutTypeDetails`

packages/jira-adapter/src/filters/request_type.ts
```diff
@@ -0,0 +1,209 @@
...
+
+const deployRequestTypeLayout = async (
+  change: Change<InstanceElement>,
+  client: JiraClient,
+  typeName: LayoutTypeName,
+): Promise<void> => {
+  const requestType = getChangeData(change)
+  const { fieldName } = layoutTypeNameToDetails[typeName]
+  if (fieldName === undefined) {
+    return undefined
```
netama, line 93: is this expected? if not let's log an error?

packages/jira-adapter/src/filters/request_type.ts
```diff
@@ -0,0 +1,209 @@
...
+    || (isModificationChange(change)
+      && (
+        !isEqualValues(change.data.before.value[fieldName], change.data.after.value[fieldName])
+        || isRequesTypeDetailsChange(change, fieldName)
+      ))
+  ) {
+    const layout = requestType.value[fieldName]
+    const items = layout.issueLayoutConfig.items.map((item: layoutConfigItem) => {
+      if (!isResolvedReferenceExpression(item.key)) {
+        log.error('Failed to deploy request type layout due to bad reference expression')
```
netama, line 106: 1. this error string doesn't have any details about the specific change, so it will be hard to debug based on it. let's add the change's elem id + field name + item key path?
2. should we turn this into an error we show the user?

packages/jira-adapter/src/filters/request_type.ts
```diff
@@ -0,0 +1,209 @@
...
+        return undefined
+      }
+      const key = item.key.value.value.id
```
netama, line 109: will the id exist on addition?

packages/jira-adapter/src/filters/request_type.ts
```diff
@@ -0,0 +1,209 @@
...
+      return {
+        type: item.type,
+        sectionType: item.sectionType.toLocaleLowerCase(),
```
netama, line 112: why do we need the locale-specific function here? won't this always be in english?

packages/jira-adapter/src/filters/request_type.ts
netama: lines 142-146 have the same args as this one, I'm wondering if we can consolidate

packages/jira-adapter/src/filters/request_type.ts
```diff
@@ -0,0 +1,209 @@
...
+    change.data.before.value[additionalFieldName] !== change.data.after.value[additionalFieldName])
+
+
+const deployRequestTypeLayout = async (
+  change: Change<InstanceElement>,
+  client: JiraClient,
+  typeName: LayoutTypeName,
+): Promise<void> => {
+  const requestType = getChangeData(change)
+  const { fieldName } = layoutTypeNameToDetails[typeName]
```
netama, line 91: there's nothing guaranteeing that `layoutTypeNameToDetails[typeName]` itself won't be undefined - maybe fallback to `?? {}` and add an early return? (might be nice to also extract all the values we need here so that we don't need to call `layoutTypeNameToDetails[typeName]` again later in this function, but if there's this safety then it should hopefully not break later)

packages/jira-adapter/src/filters/request_type.ts
```diff
@@ -0,0 +1,209 @@
...
+        if (isAdditionOrModificationChange(change)) {
+          if (isAdditionChange(change)) {
+            await defaultDeployChange({
+              change,
+              client,
+              apiDefinitions: jsmApiDefinitions,
+              fieldsToIgnore: FIELDS_TO_IGNORE,
+            })
+          }
+          await deployRequestTypeLayout(change, client, REQUEST_FORM_TYPE)
```
netama, line 193: the function can throw, we should catch the error and turn it into a change error on this change, otherwise it will fail all the other changes as well. we should also decide if we should continue to the next step if it fails on one of the fields, or if we should stop as soon as we run into an issue

netama (commented):
sorry about the delay, added some questions and suggested safeties. I think we should add test cases for failures as well. a good indication can be the coverage, if it goes down then there are code lines/branches we didn't cover in the tests

packages/jira-adapter/src/filters/request_type.ts
idoamit8: Didn't understand the question

packages/jira-adapter/src/filters/request_type.ts
idoamit8: It can't happen currently but I added a log

packages/jira-adapter/src/filters/request_type.ts
idoamit8: I guess we can show it to the user? should I just throw the error? and the defaultDeployChange will return it to the user?

packages/jira-adapter/src/filters/request_type.ts
idoamit8: yep. I just needed it lower case.

packages/jira-adapter/src/filters/request_type.ts
idoamit8: typeName type is `LayoutTypeName` and `layoutTypeNameToDetails` is `Record<LayoutTypeName, LayoutTypeDetails> so I can't see how it will be undefined

packages/jira-adapter/src/filters/request_type.ts
idoamit8: defaultDeployChange handles this. I added more throws.

packages/jira-adapter/src/filters/request_type.ts
```diff
@@ -0,0 +1,207 @@
...
+      issueLayoutConfig: {
+        items,
+      },
+    }
+    if (isAdditionChange(change)) {
+      const variables = {
+        projectId: getParent(requestType).value.id,
+        extraDefinerId: requestType.value.id,
+        layoutType: LAYOUT_TYPE_NAME_TO_DETAILS[typeName].layoutType,
+      }
```
netama, line 143: I think this can be `_.pick()` from `data`?

packages/jira-adapter/src/filters/request_type.ts
```diff
@@ -0,0 +1,207 @@
...
+  const requestType = getChangeData(change)
+  const { fieldName } = LAYOUT_TYPE_NAME_TO_DETAILS[typeName]
+  if (
+    isAdditionChange(change)
+    || (isModificationChange(change)
+      && (
+        !isEqualValues(change.data.before.value[fieldName], change.data.after.value[fieldName])
+        || isRequestTypeDetailsChange(change, fieldName)
+      ))
+  ) {
```
netama, line 99: nit: can reverse the condition and add an early return

packages/jira-adapter/src/filters/request_type.ts
```diff
@@ -0,0 +1,207 @@
...
+  && ['name', 'description', 'helpText'].some(additionalFieldName =>
+    change.data.before.value[additionalFieldName] !== change.data.after.value[additionalFieldName])
+
+
+const deployRequestTypeLayout = async (
+  change: Change<InstanceElement>,
+  client: JiraClient,
+  typeName: LayoutTypeName,
+): Promise<void> => {
+  const requestType = getChangeData(change)
```
netama, line 90: you are making changes to the change directly here, right? what happens to the nacl? it shouldn't be modified by the deploy operation, so if these changes are just for deploy we should revert them in `onDeploy` - either by transforming back, or by keeping a backup of the original change and restoring it

packages/jira-adapter/src/filters/request_type.ts
netama: the `defaultDeployChange` is run either on addition or on deletion IIUC, but not on modification. just making sure it's intentional

packages/jira-adapter/src/filters/request_type.ts
netama: oops somehow missed the wrapping...

netama (commented):
looks good :) added a few more questions/comments but otherwise LGTM

netama (approved):
pre-approving in case the change modifications are not an issue

packages/jira-adapter/src/filters/request_type.ts
idoamit8: Yep. In modification we need different endpoint

packages/jira-adapter/src/filters/request_type.ts
idoamit8: the key here is different

packages/jira-adapter/src/filters/request_type.ts
idoamit8: I am not changing anything, and the nacl doesn't change as well during deploy

packages/jira-adapter/src/filters/request_type.ts
shir-reifenberg: @idoamit8 What I was talking about was `SaltoElementError` that we return as errors in `DeployResults`, when using log.error we should add some details about the change :)

PR #4971, opened 2023-10-16
SALTO-4884: Fetch and Deploy Calendar

idoamit8:
In this Deploy I support fetch and deploy of calendars. 

---

None

---
_Release Notes_: 
_Jira Adapter:_
Now we support deploying of JSM calendars. 

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/config/api_config.ts
+++ b/packages/jira-adapter/src/config/api_config.ts
@@ -1917,7 +1917,6 @@ const JSM_DUCKTYPE_SUPPORTED_TYPES = {
   Queue: ['Queue'],
   PortalGroup: ['PortalGroup'],
   Calendar: ['Calendar'],
-  Calendar__holidays: ['Calendar__holidays'],
 }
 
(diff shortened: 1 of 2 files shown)


alonstern:
Looks great!
Small comment:
Can we remove the `date` field from the holidays array? It looks like it is just a dup of `iso8601Date` and we don't need it for deploy

packages/jira-adapter/src/config/api_config.ts
```diff
@@ -1910,6 +1917,7 @@ const JSM_DUCKTYPE_SUPPORTED_TYPES = {
   Queue: ['Queue'],
   PortalGroup: ['PortalGroup'],
   Calendar: ['Calendar'],
+  Calendar__holidays: ['Calendar__holidays'],
```
alonstern, line 1920: I think that shouldn't be here. Only types of top level instanecs should be in this list

alonstern (approved):
👌

PR #4973, opened 2023-10-16
SALTO-4886: Creating workflow fails with the workflow already exist error

alonstern:
Added a workaround for a bug in Jira that when creating a workflow we get an error that it already exist although it wasn't.

Since although we get this error the workflow is created, the workaround is to get the id from the service in this case and continue the deployment

---
_Release Notes_: 
__Jira Adapter__:
- Added a workaround for a Jira bug that when creating a workflow we somtimes receive an error that the workflow name is already used although it wasn't

---
_User Notifications_: 
None


netama (approved):
🔥

PR #4974, opened 2023-10-16
SALTO-4887: Increase page size for group rules and apps in Okta

shir-reifenberg:
Default page size was 20 for apps and 50 for rules, changed both to 200 (which is the maximum value supported)

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/config.ts
+++ b/packages/okta-adapter/src/config.ts
@@ -981,6 +981,12 @@ const DEFAULT_TYPE_CUSTOMIZATIONS: OktaSwaggerApiConfig['types'] = {
       serviceUrl: '/admin/workflow/eventhooks',
     },
   },
+  api__v1__groups__rules: {
+    request: {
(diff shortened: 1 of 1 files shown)


packages/okta-adapter/src/config.ts
```diff
@@ -327,6 +327,7 @@ const DEFAULT_TYPE_CUSTOMIZATIONS: OktaSwaggerApiConfig['types'] = {
   api__v1__apps: {
     request: {
       url: '/api/v1/apps',
+      queryParams: { limit: '200' },
```
edenhassid, line 330: nit: wdyt about saving the 200 in a constant?

edenhassid (approved):
looks good! added one small nit :)

packages/okta-adapter/src/config.ts
shir-reifenberg: I think It's more readable inline cause the file is pretty long 😅

PR #4979, opened 2023-10-18
SALTO-4190: Support Group Push (Okta)

shir-reifenberg:
Support fetch and deploy of group push in Okta using private API

---

Changes 
- Add filter to fetch 2 new types - `GroupPush` and `GroupPushRule`
- pre deploy and on deploy apply to removal of those types, in this case we ignore all instance fields and only send the action we want (in the future i'll change it to be configurable)
- Add references, alias, service url to the added types

Still missing UT that I'll add soon

---
_Release Notes_: 

_Okta adapter_:
- Support fetch and deploy of Group Push 

---
_User Notifications_: 

_Okta adapter_:
- Instances of Group Push and Group Push rules will be added


packages/okta-adapter/src/filters/group_push.ts
```diff
@@ -0,0 +1,291 @@
...
+
+    const paginator = createPaginator({
+      client: adminClient,
+      // Pagination method is different from the rest of Okta's API
+      paginationFuncCreator: () => getWithCursorPagination(),
+    })
+
+    const instances = (await Promise.all(appsWithGroupPush
+      .map(async appInstance => {
+        const groupPushEntries = await getGroupPushForApp(paginator, appInstance.value.id)
```
edenhassid, line 230: can id be undefined?

edenhassid (commented):
LGTM! 💯 🔥  added some small comments, don't forget to add tests :)

packages/okta-adapter/src/filters/add_alias.ts
```diff
@@ -90,6 +91,18 @@ const aliasMap: Record<string, AliasData> = {
     ],
     separator: ' : ',
   },
+  [GROUP_PUSH_TYPE_NAME]: {
```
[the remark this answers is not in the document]
shir-reifenberg: 🙏

packages/okta-adapter/src/filters/group_push.ts
shir-reifenberg: don't think so cause it's on fetch

edenhassid (approved):
looks good!! 💯 🏆 🔥

PR #4988, opened 2023-10-19
SALTO-4905: fix a bug in screenTab fields with missingReference - Jira

shayc331:
_fix a bug in screenTab fields with missingReference_

---

* currently we have a change validator that prevents missing references in the `after` instance
* if the change is a modification, the screen filter uses the `before` instance fields, so we got an error because it is a missing reference.
* I added a hidden value that stores the original fields ids and a `preDeploy` operation to the screen filter that uses the original fields ids in order to prevent this error.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


alonstern (approved):
🔥

PR #4989, opened 2023-10-21
SALTO-4905: add onDeploy to screen filter - Jira

shayc331:
_add onDeploy to screen filter_

---

* fix a bug in https://github.com/salto-io/salto/pull/4988

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/jira-adapter/src/filters/screen/screen.ts
+++ b/packages/jira-adapter/src/filters/screen/screen.ts
@@ -184,12 +184,16 @@ const filter: FilterCreator = ({ config, client }) => ({
       .forEach(instance => {
         Object.values(instance.value.tabs).forEach((tab: Value): void => {
           if (tab.fields) {
-            tab.originalFieldsIds.ids = tab.fields
-              .map((refOrFieldId: ReferenceExpression | string) => (
-                isReferenceExpression(refOrFieldId) ? refOrFieldId.value.value.id : refOrFieldId
-              ))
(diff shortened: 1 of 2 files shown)


packages/jira-adapter/src/filters/screen/screen.ts
```diff
@@ -174,6 +174,26 @@ const filter: FilterCreator = ({ config, client }) => ({
...
+    changes
+      .filter(isInstanceChange)
+      .filter(isAdditionOrModificationChange)
+      .map(change => getChangeData(change))
+      .filter(instance => instance.elemID.typeName === SCREEN_TYPE_NAME)
+      .filter(instance => !_.isEmpty(instance.value.tabs))
+      .forEach(instance => {
+        Object.values(instance.value.tabs).forEach((tab: Value): void => {
+          if (tab.fields) {
+            tab.originalFieldsIds.ids = tab.fields
```
alonstern, line 187: I think on addition `originalFieldsIds` will be undefined so we probably should do
```
tab.originalFieldsIds = {
  ids: tab.fields...
}
```
both here and in 192

alonstern (approved):
Alt Text

packages/jira-adapter/src/filters/screen/screen.ts
shayc331: nice catch, thanks!

PR #4997, opened 2023-10-23
SALTO-4915 fix jira scriptedFields e2e flakiness

netama:
fix `Full error from the logs: Search term is already being used by another scripted field` by randomizing the search term as well

---
_Release Notes_: 
None

---
_User Notifications_: 
None


netama:
seems to do the trick so far

shir-reifenberg (approved):
👑

PR #4998, opened 2023-10-24
SALTO-4808 - Convert JSON pairs of 'id' to reference expression

idoser96:
In free text fields, if the value is a JSON, and there is a key named 'id' with a number value - attempt to convert it to reference expression

---

No noise reduction needed because this is behind a config

---
_Release Notes_: 
Zendesk Adapter:
* Added `convertJsonIdsToReferences` config, that allows convert of JSON pairs of 'id' to reference expression

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/filters/handle_template_expressions.ts
+++ b/packages/zendesk-adapter/src/filters/handle_template_expressions.ts
@@ -422,9 +422,9 @@ const replaceFormulasWithTemplates = ({
           return expression
         }
 
-        const expressionWithId = expression.replace(idRegex, '')
+        const expressionWithoutId = expression.replace(idRegex, '')
         const idInstance = instance ?? createMissingInstance(ZENDESK, 'unknown', id)
-        return [expressionWithId, new ReferenceExpression(idInstance.elemID, idInstance)]
+        return [expressionWithoutId, new ReferenceExpression(idInstance.elemID, idInstance)]
       }
     )
   }
--- a/packages/zendesk-adapter/src/filters/handle_template_expressions.ts
+++ b/packages/zendesk-adapter/src/filters/handle_template_expressions.ts
@@ -449,6 +449,7 @@ const replaceFormulasWithTemplates = ({
     getContainers(instances).forEach(container => {
       const { fieldName } = container
       container.values.forEach(value => {
+        // Has to be first because it needs to receive the whole string, not a template expression
         if (convertJsonIdsToReferences) {
           value[fieldName] = convertIdsInJson(value[fieldName])
         }
--- a/packages/zendesk-adapter/src/filters/handle_template_expressions.ts
+++ b/packages/zendesk-adapter/src/filters/handle_template_expressions.ts
@@ -47,6 +47,7 @@ const TICKET_FIELD_SPLIT = '(?:(ticket.ticket_field|ticket.ticket_field_option_t
 const KEY_SPLIT = '(?:([^ ]+\\.custom_fields)\\.)'
 const TITLE_SPLIT = '(?:([^ ]+)\\.(title))'
 const SPLIT_REGEX = `${TICKET_FIELD_SPLIT}|${KEY_SPLIT}|${TITLE_SPLIT}`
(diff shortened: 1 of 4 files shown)


packages/zendesk-adapter/src/filters/handle_template_expressions.ts
```diff
@@ -326,7 +327,7 @@ const formulaToTemplate = ({
       if (extractReferencesFromFreeText) {
         // Check if the expression is a link to a zendesk page without a subdomain
         // href="/hc/en-us/../articles/123123
-        const isZendeskLink = new RegExp(`"/hc/\\S*${expression}`).test(formula)
+        const isZendeskLink = new RegExp(`"/hc/\\S*${_.escapeRegExp(expression)}`).test(formula)
```
idoser96, line 330: bigfix

packages/zendesk-adapter/test/filters/handle_template_expressions.test.ts
```diff
@@ -38,6 +38,12 @@ import { DEFAULT_CONFIG, FETCH_CONFIG } from '../../src/config'
 
 const { createMissingInstance } = referencesUtils
 
+let id = 0
+const newId = (): number => {
+  id += 1
+  return id
+}
+
```
idoser96, line 46: we are depended on unique ids now

packages/zendesk-adapter/src/filters/handle_template_expressions.ts
```diff
@@ -382,6 +389,46 @@ const replaceFormulasWithTemplates = (
...
+          return expression
+        }
+        const instance = instancesById[id]
+        if (instance === undefined && !enableMissingReferences) {
+          return expression
+        }
+
+        const expressionWithId = expression.replace(idRegex, '')
+        const idInstance = instance ?? createMissingInstance(ZENDESK, 'unknown', id)
+        return [expressionWithId, new ReferenceExpression(idInstance.elemID, idInstance)]
```
edenhassid, line 427: this does not support the case where we have something after the id, such as `,`

packages/zendesk-adapter/src/filters/handle_template_expressions.ts
```diff
@@ -47,6 +47,7 @@ const TICKET_FIELD_SPLIT = '(?:(ticket.ticket_field|ticket.ticket_field_option_t
 const KEY_SPLIT = '(?:([^ ]+\\.custom_fields)\\.)'
 const TITLE_SPLIT = '(?:([^ ]+)\\.(title))'
 const SPLIT_REGEX = `${TICKET_FIELD_SPLIT}|${KEY_SPLIT}|${TITLE_SPLIT}`
+const ID_KEY_IN_JSON_REGEX = /("id"\s*:\s*\d+)/
```
edenhassid, line 50: does this regex also catches other characters after the id? such as `,`?

packages/zendesk-adapter/src/filters/handle_template_expressions.ts
```diff
@@ -382,6 +389,46 @@ const replaceFormulasWithTemplates = (
...
+        if (id === undefined) {
+          log.error(`Error parsing id in expression: ${expression}`)
+          return expression
+        }
+        const instance = instancesById[id]
+        if (instance === undefined && !enableMissingReferences) {
+          return expression
+        }
+
+        const expressionWithId = expression.replace(idRegex, '')
```
edenhassid, line 425:
```suggestion
        const expressionWithoutId = expression.replace(idRegex, '')
```

edenhassid (commented):
looks good! added some comments 🌻

packages/zendesk-adapter/src/filters/handle_template_expressions.ts
idoser96: There can't be anything after the id, because the extractTemplate only gives the regex itself (which ends with the number) - this is tested

packages/zendesk-adapter/src/filters/handle_template_expressions.ts
idoser96: it doesn't and it is not needed to

PR #4999, opened 2023-10-24
SALTO-4923: Added flag for enableJsmExperimental

idoamit8:
Added enableJsmExperimental flag

---

We are planning to release JSM to customers. 
For not expose them to JSM elements that are being developed, I added another flag called `enableJsmExperimental`.
The customers will be able to use jsm when `enableJSM = true`. And after each type will be approve it will change from `enableJsmExperimental=true` to `enableJSM=true`

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/config/config.ts
+++ b/packages/jira-adapter/src/config/config.ts
@@ -67,6 +67,7 @@ type JiraFetchConfig = configUtils.UserFetchConfig<JiraFetchFilters> & {
   parseTemplateExpressions?: boolean
   enableScriptRunnerAddon?: boolean
   enableJSM?: boolean
+  enableJsmExperimental?: boolean
   removeDuplicateProjectRoles?: boolean
   addAlias?: boolean
   splitFieldConfiguration?: boolean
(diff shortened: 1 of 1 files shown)


packages/jira-adapter/src/config/config.ts
```diff
@@ -67,6 +67,7 @@ type JiraFetchConfig = configUtils.UserFetchConfig<JiraFetchFilters> & {
   parseTemplateExpressions?: boolean
   enableScriptRunnerAddon?: boolean
   enableJSM?: boolean
+  enableJsmExperimental?: boolean
```
netama, line 70: shouldn't we also omit this from the default config so that it doesn't show up when initializing the jira nacl? (I'd expect the same to be needed for the other flags, but maybe that was intentional - this one is a temporary flag so should not be user-visible)

idoamit8: We should include it. I omitted it from the default config. I'm not sure if `SCRIPT_RUNNER_API_DEFINITIONS` and `JSM_DUCKTYPE_API_DEFINITIONS` that are on the list mean that they are also omitted?

netama (approved):

PR #5000, opened 2023-10-24
SALTO-4896: expand secondGlobalContext CV for handling invalid removals of contextFields from a project - Jira

shayc331:
_expand secondGlobalContext CV for handling invalid removals of contextFields from a project_

---

* we already have a change validator for second global context (which is forbidden in Jira)
* the scenario: when we remove a fieldContext from a project it implicitly becomes a global context, if there is another global context the service fails this change.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
+++ b/packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
@@ -37,8 +37,7 @@ const createProjectErrorMessage = (elemID: ElemID, fieldContextName: string): Ch
   detailedMessage: `Can't remove the field context ${fieldContextName} from this project as it will result in more than a single global context.`,
 })
 
-const getParentElemID = (instance: InstanceElement)
-: ElemID => {
+const getParentElemID = (instance: InstanceElement): ElemID => {
   const parent = getParents(instance)[0]
   if (!isReferenceExpression(parent)) {
     throw new Error(`Expected ${instance.elemID.getFullName()} parent to be a reference expression`)
@@ -58,7 +57,6 @@ export const fieldSecondGlobalContextValidator: ChangeValidator = async (changes
     return []
   }
   const fieldToGlobalContextCount: Record<string, number> = {}
-  const fieldContextToProjectChangeData = new Map<string, ProjectChangeData>()
   const fillFieldToGlobalContextCount = async (): Promise<void> => (
     awu(await elementSource.getAll())
       .filter(elem => elem.elemID.typeName === FIELD_CONTEXT_TYPE_NAME)
@@ -74,8 +72,9 @@ export const fieldSecondGlobalContextValidator: ChangeValidator = async (changes
         }
       })
   )
-  const fillFieldContextToProjectChangeData = async (): Promise<void> => (
-    awu(changes).filter(isInstanceChange)
+  const fillFieldContextToProjectChangeData = async (): Promise<Map<string, ProjectChangeData>> => {
+    const fieldContextToProjectChangeData = new Map<string, ProjectChangeData>()
+    await awu(changes).filter(isInstanceChange)
       .filter(change => getChangeData(change).elemID.typeName === PROJECT_TYPE)
       .filter(isRemovalOrModificationChange)
       .forEach(change => {
@@ -89,32 +88,38 @@ export const fieldSecondGlobalContextValidator: ChangeValidator = async (changes
             )
           })
       })
-  )
+    if (fieldContextToProjectChangeData.size === 0) {
+      return fieldContextToProjectChangeData
+    }
+    const projectInstances = await getInstancesFromElementSource(elementSource, [PROJECT_TYPE])
+    // if there is a project that using a field context that is not global context
+    projectInstances.flatMap(instance => instance.value.fieldContexts)
+      .filter(isReferenceExpression)
+      .forEach((context: ReferenceExpression) => {
+        fieldContextToProjectChangeData.delete(context.elemID.getFullName())
+      })
+    return fieldContextToProjectChangeData
+  }
+
   const globalContextList = await awu(changes).filter(isInstanceChange)
     .filter(isAdditionOrModificationChange)
     .map(getChangeData)
     .filter(instance => instance.elemID.typeName === FIELD_CONTEXT_TYPE_NAME)
     .filter(instance => instance.value.isGlobalContext)
     .toArray()
+  const globalContextElemIdsSet = new Set(globalContextList.map(instance => instance.elemID.getFullName()))
 
   await fillFieldToGlobalContextCount()
-  await fillFieldContextToProjectChangeData()
+  const fieldContextToProjectChangeData = await fillFieldContextToProjectChangeData()
 
-  const projectInstances = fieldContextToProjectChangeData.size > 0 // add ut
-    ? await getInstancesFromElementSource(elementSource, [PROJECT_TYPE])
-    : []
-
-  // if there is a project that using a field context that is not global context
-  projectInstances.flatMap(instance => instance.value.fieldContexts)
-    .filter(isReferenceExpression)
-    .forEach((context: ReferenceExpression) => {
-      fieldContextToProjectChangeData.delete(context.elemID.getFullName())
-    })
-  const errorMessages = Array.from(fieldContextToProjectChangeData.values())
-    .filter(({ fieldElemId }) => (
+  const errorMessages = Array.from(fieldContextToProjectChangeData.entries())
+    .filter(([fieldContextName]) => (!globalContextElemIdsSet.has(fieldContextName)))
+    .filter(([, { fieldElemId }]) => (
       fieldToGlobalContextCount[fieldElemId.getFullName()] >= 1
     ))
-    .map(({ change, fieldContextName }) => createProjectErrorMessage(getChangeData(change).elemID, fieldContextName))
+    .map(([, { change, fieldContextName }]) => createProjectErrorMessage(
+      getChangeData(change).elemID, fieldContextName
+    ))
 
   if (globalContextList.length > 0) {
     const secondGlobalContextErrorMessages = globalContextList
--- a/packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
+++ b/packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
@@ -45,8 +45,8 @@ const getParentElemID = (instance: InstanceElement): ElemID => {
   return parent.elemID
 }
 
-type ProjectChangeData = {
-  change: Change<InstanceElement>
+type ProjectChangesData = {
+  changes: Change<InstanceElement>[]
   fieldElemId: ElemID
   fieldContextName: string
 }
@@ -57,6 +57,7 @@ export const fieldSecondGlobalContextValidator: ChangeValidator = async (changes
     return []
   }
   const fieldToGlobalContextCount: Record<string, number> = {}
+  const fieldToImplicitGlobalContextCount: Record<string, number> = {}
   const fillFieldToGlobalContextCount = async (): Promise<void> => (
     awu(await elementSource.getAll())
       .filter(elem => elem.elemID.typeName === FIELD_CONTEXT_TYPE_NAME)
@@ -72,21 +73,25 @@ export const fieldSecondGlobalContextValidator: ChangeValidator = async (changes
         }
       })
   )
-  const fillFieldContextToProjectChangeData = async (): Promise<Map<string, ProjectChangeData>> => {
-    const fieldContextToProjectChangeData = new Map<string, ProjectChangeData>()
+  const fillFieldContextToProjectChangeData = async (): Promise<Map<string, ProjectChangesData>> => {
+    const fieldContextToProjectChangeData = new Map<string, ProjectChangesData>()
     await awu(changes).filter(isInstanceChange)
       .filter(change => getChangeData(change).elemID.typeName === PROJECT_TYPE)
       .filter(isRemovalOrModificationChange)
-      .forEach(change => {
-        change.data.before.value.fieldContexts
+      .forEach(async change => {
+        const promises = change.data.before.value.fieldContexts
           .filter(isReferenceExpression)
-          .forEach(async (context: ReferenceExpression) => {
-            const fieldElemId = getParentElemID(await context.getResolvedValue(elementSource))
-            fieldContextToProjectChangeData.set(
-              context.elemID.getFullName(),
-              { change, fieldElemId, fieldContextName: context.elemID.name }
-            )
+          .map(async (context: ReferenceExpression) => {
+            if (fieldContextToProjectChangeData.get(context.elemID.getFullName()) === undefined) {
+              const fieldElemId = getParentElemID(await context.getResolvedValue(elementSource))
+              fieldContextToProjectChangeData.set(
+                context.elemID.getFullName(),
+                { fieldElemId, fieldContextName: context.elemID.name, changes: [] }
+              )
+            }
+            fieldContextToProjectChangeData.get(context.elemID.getFullName())?.changes.push(change)
           })
+        await Promise.all(promises)
       })
     if (fieldContextToProjectChangeData.size === 0) {
       return fieldContextToProjectChangeData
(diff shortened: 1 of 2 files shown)


netama (commented):
started reviewing the code and added some comments on the code, but I _think_ logically we might not be covering everything - IIUC there are other cases we should cover besides project modifications, e.g.:
1. if a project is removed can't it also make a field context "orphaned"?
2. if a project is added and references a field context, can't it "cancel out" the removed referenced of a modified project with a removed reference to a field context? (also in this case - do we need to do anything about the API call order to make sure this succeeds?)
etc

if the above is incorrect, let's discuss IRL :)
if it is - I think both for code clarity and for correctness, it might be better to calculate the references from projects similarly to how we calculate the `fieldToGlobalContextCount` today - based on the element source (which we're already reading, and I assume the number of projects will usually be significantly lower than the number of parent fields), by reading all project instances and calculating the number of references per elem id / internal id. that way we won't have to consider all the edge cases. wdyt?

also (for the existing code) - any idea why we need to access the element source for `getParentWithElementSource`, if we're only using it to extract its elem id? I think we might be able to save all these calls?

packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
```diff
@@ -70,10 +73,32 @@ export const fieldSecondGlobalContextValidator: ChangeValidator = async (changes
     .filter(instance => instance.elemID.typeName === FIELD_CONTEXT_TYPE_NAME)
     .filter(instance => instance.value.isGlobalContext)
     .toArray()
+  await fillFieldToGlobalContextCount()
+  const isInvalidProjectFieldContextsModification = async (change: ModificationChange<InstanceElement>)
+    : Promise<boolean> => {
+    const afterContexts = new Set((change.data.after.value.fieldContexts ?? [])
+      .map((context: ReferenceExpression) => context.elemID.getFullName()))
+    const removedContexts = change.data.before.value.fieldContexts
```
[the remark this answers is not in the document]
shayc331: im not sure I understand when `elemId` is not good enough

packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
```diff
@@ -71,9 +71,43 @@ export const fieldSecondGlobalContextValidator: ChangeValidator = async (changes
...
+  const errorMessages = Array.from(fieldContextToChange.keys())
+    .filter(context => fieldContextToChange.get(context) !== undefined)
+    .filter(context => {
+      const changeAndFieldElemId = fieldContextToChange.get(context)
+      if (changeAndFieldElemId !== undefined) {
+        const { fieldElemId } = changeAndFieldElemId
+        // if there is another global context for the field
+        return fieldToGlobalContextCount[fieldElemId.getFullName()] === 1
+      }
+      return false
```
netama, line 103: styling nit: early return (+ log) for unexpected cases
```suggestion
      if (changeAndFieldElemId === undefined) {
        log.error(...)
        return false
      }
      const { fieldElemId } = changeAndFieldElemId
      // if there is another global context for the field
      return fieldToGlobalContextCount[fieldElemId.getFullName()] === 1
      }
```

packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
```diff
@@ -71,9 +71,43 @@ export const fieldSecondGlobalContextValidator: ChangeValidator = async (changes
...
+    .forEach((context: ReferenceExpression) => {
+      fieldContextToChange.delete(context.elemID.getFullName())
+    })
+  const errorMessages = Array.from(fieldContextToChange.keys())
+    .filter(context => fieldContextToChange.get(context) !== undefined)
+    .filter(context => {
+      const changeAndFieldElemId = fieldContextToChange.get(context)
+      if (changeAndFieldElemId !== undefined) {
+        const { fieldElemId } = changeAndFieldElemId
+        // if there is another global context for the field
```
netama, line 100: not sure I understand the comment

packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
```diff
@@ -71,9 +71,43 @@ export const fieldSecondGlobalContextValidator: ChangeValidator = async (changes
...
+    })
+  const projectInstances = await getInstancesFromElementSource(elementSource, [PROJECT_TYPE])
+
+  // if there is a project that using a field context that is not global context
+  projectInstances.flatMap(instance => instance.value.fieldContexts)
+    .filter(isReferenceExpression)
+    .forEach((context: ReferenceExpression) => {
+      fieldContextToChange.delete(context.elemID.getFullName())
+    })
+  const errorMessages = Array.from(fieldContextToChange.keys())
```
netama, line 94: maybe just `fieldContextToChange.entries()` and then you don't need to check for undefined? (wrote this after some of the comments below)

packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
```diff
@@ -71,9 +71,43 @@ export const fieldSecondGlobalContextValidator: ChangeValidator = async (changes
...
+    .filter(context => fieldContextToChange.get(context) !== undefined)
+    .filter(context => {
+      const changeAndFieldElemId = fieldContextToChange.get(context)
+      if (changeAndFieldElemId !== undefined) {
+        const { fieldElemId } = changeAndFieldElemId
+        // if there is another global context for the field
+        return fieldToGlobalContextCount[fieldElemId.getFullName()] === 1
```
netama, line 101: why `=== 1` and not `>== 1`?

packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
```diff
@@ -71,9 +71,43 @@ export const fieldSecondGlobalContextValidator: ChangeValidator = async (changes
...
+      }
+      return false
+    })
+    .map(context => fieldContextToChange.get(context)?.change)
+    .filter(values.isDefined)
+    .map(change => createErrorMessage(getChangeData(change).elemID))
```
netama, line 107: continuing from above - IIUC there can be more than one relevant change

packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
```diff
@@ -85,6 +119,7 @@ export const fieldSecondGlobalContextValidator: ChangeValidator = async (changes
         return undefined
       })
       .filter(values.isDefined)
+    errorMessages.push(...secondGlobalContextErrorMessages)
```
netama, line 122: continuing from earlier - I'm not sure there's much value in concatenating the errors at this level, I think in practice these are pretty much two separate cases (though we might need to remove overlaps?) - so might be cleaner and easier to maintain if we use two separate functions here and only combine at the end. wdyt?

netama (commented):
looks good overall, added some comments about potential edge cases (I might be missing something though), and some nits around code structure

packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
```diff
@@ -71,9 +75,43 @@ export const fieldSecondGlobalContextValidator: ChangeValidator = async (changes
...
     .toArray()
 
+  await fillFieldToGlobalContextCount()
+  await awu(changes).filter(isInstanceChange)
+    .filter(change => getChangeData(change).elemID.typeName === PROJECT_TYPE)
+    .filter(isRemovalOrModificationChange)
+    .forEach(change => {
+      change.data.before.value.fieldContexts
+        .filter(isReferenceExpression)
+        .forEach(async (context: ReferenceExpression) => {
```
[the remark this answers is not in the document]
shayc331: If the context becomes global it does not matter which old project used it. If there are some projects that removed this context, only one of them is not allowed. We can (and IMHO should) deploy the rest.

packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
shayc331: I agree to separate the cases but I dont think there is an overlap. The first is about project's changes and the second is fieldContext's changes

packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
```diff
@@ -49,13 +52,14 @@ export const fieldSecondGlobalContextValidator: ChangeValidator = async (changes
     return []
   }
   const fieldToGlobalContextCount: Record<string, number> = {}
+  const fieldContextToChange = new Map<string, {change: Change<InstanceElement>; field: InstanceElement}>()
```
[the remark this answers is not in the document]
shayc331: I prefer `Map`, I used `size` and `values` which Im not sure what are the equivalent in `Record`

packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
```diff
@@ -71,9 +71,43 @@ export const fieldSecondGlobalContextValidator: ChangeValidator = async (changes
...
 
+  await fillFieldToGlobalContextCount()
+  await awu(changes).filter(isInstanceChange)
+    .filter(change => getChangeData(change).elemID.typeName === PROJECT_TYPE)
+    .filter(isRemovalOrModificationChange)
+    .forEach(change => {
+      change.data.before.value.fieldContexts
+        .filter(isReferenceExpression)
+        .forEach(async (context: ReferenceExpression) => {
+          const fieldElemId = await getParentElemID(await context.getResolvedValue())
```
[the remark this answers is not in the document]
shayc331: I added the `elementSource` to the calls for safety

packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
```diff
@@ -13,34 +13,43 @@
...
+  severity: 'Error' as SeverityLevel,
+  message: 'A field can only have a single global context',
+  detailedMessage: `Can't remove the field context ${fieldContextName} from this project as it will result in more than a single global context.`,
+})
 
-const getParenWithElementSource = async (instance: InstanceElement, elementSource: ReadOnlyElementsSource)
-: Promise<InstanceElement> => {
-  const parent = instance.annotations[CORE_ANNOTATIONS.PARENT][0]
+const getParentElemID = (instance: InstanceElement)
+: ElemID => {
```
netama, line 41: tiny nit: no need for newline

packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
```diff
@@ -49,42 +58,78 @@ export const fieldSecondGlobalContextValidator: ChangeValidator = async (changes
...
     .filter(isAdditionOrModificationChange)
     .map(getChangeData)
     .filter(instance => instance.elemID.typeName === FIELD_CONTEXT_TYPE_NAME)
     .filter(instance => instance.value.isGlobalContext)
     .toArray()
 
+  await fillFieldToGlobalContextCount()
+  await fillFieldContextToProjectChangeData()
+
+  const projectInstances = fieldContextToProjectChangeData.size > 0 // add ut
```
netama, line 103: leftover comment?

packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
```diff
@@ -49,42 +58,78 @@ export const fieldSecondGlobalContextValidator: ChangeValidator = async (changes
...
+    ? await getInstancesFromElementSource(elementSource, [PROJECT_TYPE])
+    : []
+
+  // if there is a project that using a field context that is not global context
+  projectInstances.flatMap(instance => instance.value.fieldContexts)
+    .filter(isReferenceExpression)
+    .forEach((context: ReferenceExpression) => {
+      fieldContextToProjectChangeData.delete(context.elemID.getFullName())
+    })
```
netama, line 112: nit: can't this part be inside the `fillFieldContextToProjectChangeData` function as well? (and then you can also return early when there are no project instances + return the map from the function instead of initializing it outside)

packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
netama: can't there be a case where a field context is both marked as global with `isGlobalContext` and disconnected from all projects in the same deploy?

packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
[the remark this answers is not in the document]
netama: you can use `Object.values(x)` and get the length of an array, but no objections to using the map (I actually thought I removed this one earlier after reading the usage, sorry)

packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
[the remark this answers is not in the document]
netama: ok - but how do we choose which one we won't deploy? I think the behavior should be deterministic? (might be worth consulting with @liorn on this one)

packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
```diff
@@ -49,42 +58,78 @@ export const fieldSecondGlobalContextValidator: ChangeValidator = async (changes
...
+      .filter(change => getChangeData(change).elemID.typeName === PROJECT_TYPE)
+      .filter(isRemovalOrModificationChange)
+      .forEach(change => {
+        change.data.before.value.fieldContexts
+          .filter(isReferenceExpression)
+          .forEach(async (context: ReferenceExpression) => {
+            const fieldElemId = getParentElemID(await context.getResolvedValue(elementSource))
+            fieldContextToProjectChangeData.set(
+              context.elemID.getFullName(),
+              { change, fieldElemId, fieldContextName: context.elemID.name }
```
netama, line 88: no strong opinion if you prefer to write it this way, but I think we can avoid the duplication of using the context both as a key and a value (will require using below `.entries()` instead of `.values()`)

packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
```diff
@@ -13,34 +13,43 @@
...
   elemID,
   severity: 'Error' as SeverityLevel,
   message: 'A field can only have a single global context',
   detailedMessage: `Can't deploy this global context because the deployment will result in more than a single global context for field ${fieldName}.`,
 })
+const createProjectErrorMessage = (elemID: ElemID, fieldContextName: string): ChangeError => ({
+  elemID,
+  severity: 'Error' as SeverityLevel,
+  message: 'A field can only have a single global context',
+  detailedMessage: `Can't remove the field context ${fieldContextName} from this project as it will result in more than a single global context.`,
```
netama, line 37: nit (not a blocker): I think it _might_ be worth being slightly more explicit and explaining that the field context will not be connected to any project and therefore be global?
(also IMO adding a different `message` as well to distinguish between the cases, but I understand there are also some advantages to keeping it)

netama (approved):
looks good, added a few more questions and some nits - pre-approving to unblock, but if there are changes it might be worth doing another short round of reviews (can be by someone else just on the changes)

also (and not a blocker technically for merging the PR, just a case we should consider and potentially handle as well) - I think we should test the case where a project is removed and another is added - if we first remove and then add, the deploy might fail on the removal even though it won't be caught by the validator? (didn't run this, just speculating - lmk if you want to discuss IRL)

packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
shayc331: it could have happened, I added a check for it [here](https://github.com/salto-io/salto/pull/5000/files#diff-f502943dbb0c442d69e5c5435c7578b8f644477d15238cd5c87c8fe1bf0e137eR116)

packages/jira-adapter/src/change_validators/field_contexts/second_global_context.ts
```diff
@@ -72,21 +73,25 @@ export const fieldSecondGlobalContextValidator: ChangeValidator = async (changes
...
+        const promises = change.data.before.value.fieldContexts
           .filter(isReferenceExpression)
-          .forEach(async (context: ReferenceExpression) => {
-            const fieldElemId = getParentElemID(await context.getResolvedValue(elementSource))
-            fieldContextToProjectChangeData.set(
-              context.elemID.getFullName(),
-              { change, fieldElemId, fieldContextName: context.elemID.name }
-            )
+          .map(async (context: ReferenceExpression) => {
+            if (fieldContextToProjectChangeData.get(context.elemID.getFullName()) === undefined) {
```
netama, line 85: I think there can still be a race condition here if these are run in parallel with async, no? (a case where there are two changes but the map will end up having only one since both will get into the `if`)
I think there are a few ways to resolve this:
1. instead of using `Promise.all()` here use `awu()`, so that no two changes run in parallel - that's probably the simplest and it won't cause a real optimization hit, since there is only a single async call anyway for all field contexts of a project
2. turn `fieldElemId` into a promise and leave everything else sync - and then `await` its result only when you need it
3. have two separate mappings, one for `fieldElemId` and one for the changes, and initialize the `fieldElemId` map in a blocking way

shayc331: followed your first suggestion, thanks!

netama (approved):

PR #5001, opened 2023-10-24
SALTO-4922: Mark dev accounts as non production in Okta

shir-reifenberg:
Added another account type - dev account
all preview accounts (regular and dev) will be marked as preview, dev accounts that are not preview will be marked as dev, and the others as production.

---
_Release Notes_: 
None

---
_User Notifications_: 
None


omrilit (approved):
Looks good!

PR #5002, opened 2023-10-24
Saved searches author information query is skipped in full fetch

dantal4:
We were passing the timeZone&format to filtersRunner only in case of `useChangesDetection=true`, but we should always pass it. It caused the author_information/saved_searches filter to skip the query.

---

---
_Release Notes_: 
Netsuite Adapter:
- Fix bug - saved searches author information query is skipped in full fetch

---
_User Notifications_: 
None


omrilit (approved):
Looks good! and nice catch!

PR #5003, opened 2023-10-24
SALTO-4717 reduce log level to debug in some cases, relax zendesk restrictions

netama:
Reduce the noise in the logs - 
1. reduce level to `debug` if field is not defined
2. relax restrictions in zendesk to not expect a specific type for the default-hidden fields, since there is sometimes inconsistency between string/number that causes the field type to be unknown and we should still hide it in this case
3. assign a few more field types with `fieldTypeOverrides` since they don't always exist

---
_Release Notes_: 
None

---
_User Notifications_: 
None


edenhassid (approved):
LGTM 🔥

PR #5004, opened 2023-10-25
SALTO-4919 Fix SDF deploy dependencies when referencing an added inner object/field

dantal4:
We are currently considering a referenced SDF object as a dependency only in case that the referenced object was added. We should also consider it a dependency when the reference is to an inner object/field (`customworkflow123.customstate123` / `customrecord123.custfield123`), that was added (although the whole object was modified).

In order to do that:
- in `additional_changes` filter - set the `before` of an added fields parent to be without the added fields.
- move logic of extracting inner objects from `element_references` filter to `service_id_info`.
- use the above logic to find the added objects in a modified SDF object.

---

---
_Release Notes_: 
Netsuite Adapter:
- Fix SDF deploy dependencies when referencing an added inner object/field

---
_User Notifications_: 
None

### DIFF
--- a/packages/netsuite-adapter/src/client/client.ts
+++ b/packages/netsuite-adapter/src/client/client.ts
@@ -14,9 +14,7 @@
 * limitations under the License.
 */
 
-import { Change, getChangeData, InstanceElement, isInstanceChange, isModificationChange, CredentialError,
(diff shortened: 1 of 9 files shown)


omrilit (approved):
Great work!

PR #5009, opened 2023-10-25
SAAS-9299: NetSuite credentials not propagated as credentials issue

dorsev:
A similar issue to https://github.com/salto-io/salto/pull/4995 but in the NetSuite adapter.
we create multiple promises, without awaiting all of them, which causes a "crash" while all rejects, but just the first is caught. 

More info inside the attached PR.

---

---

_Release Notes_: 
None

---

User Notifications_: 

None

### DIFF
--- a/packages/netsuite-adapter/src/adapter.ts
+++ b/packages/netsuite-adapter/src/adapter.ts
@@ -266,7 +266,11 @@ export default class NetsuiteAdapter implements AdapterOperations {
       )
       : []
 
-    const dataElementsPromise = getDataElements(this.client, fetchQuery, this.getElemIdFunc)
+    const {
+      elements: dataElements,
+      requestedTypes: requestedDataTypes,
+      largeTypesError: dataTypeError,
+    } = await getDataElements(this.client, fetchQuery, this.getElemIdFunc)
 
     const getCustomObjectsResult = this.client.getCustomObjects(
       getStandardTypesNames(),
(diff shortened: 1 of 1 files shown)


packages/netsuite-adapter/src/adapter.ts
```diff
@@ -266,7 +266,11 @@ export default class NetsuiteAdapter implements AdapterOperations {
       )
       : []
 
-    const dataElementsPromise = getDataElements(this.client, fetchQuery, this.getElemIdFunc)
+    const {
```
dorsev, line 269: I've of course, tested this, and made sure this works as expected.
However, I would have expected `getConfigRecords` to fail earlier and not allow the code to pass all the way through here.

I also think we should probably change the following code to

https://github.com/salto-io/salto/blob/main/packages/netsuite-adapter/src/client/suiteapp_client/suiteapp_client.ts#L348

```
async getConfigRecords() { 
       try { ...code } 
        catch (e) {
            log.error('getConfigRecords failed. received error: %s', e.message);
            // or e isInstanceOf InvalidSuiteAppCredentialsError
            if (e.message === 'Invalid SuiteApp credentials') {
                log.error('getConfigRecords failed with invalid creds error');
                throw e
            }
            return [];
        }

```


thoughts? 
I think probably doing both is legit, LMK your thoughts
(would love some directions here, as it's my first time touching the NS adapter code)

dantal4:
@dorsev I opened https://github.com/salto-io/salto/pull/5010 to take care of it without harming the ability to run several APIs at the same time. It is important for the fetch time.

packages/netsuite-adapter/src/adapter.ts
omrilit: Nice catch @dorsev and thanks for taking it!
The reason we didn't have await here in first place is that we wanted to parallelize the calls to `getDataElements` (which perform "SuiteApp" calls) and `getCustomObjects` which performs SDF calls.

I wonder how we can make sure that we don't lose the advantage of parallelism while failing fast on invalid credentials.
The naive option would be to explicitly call `validateCredentials` at the beginning of the adapter's fetch, but it takes few seconds until it's done so if possible it's better to avoid that.
@dantal4 WDYT?

omrilit: I see now that Daniel opened https://github.com/salto-io/salto/pull/5010/files, I will review it

PR #5010, opened 2023-10-25
SAAS-9299: NetSuite credentials not propagated as credentials issue

dantal4:
Based on @dorsev's PR (https://github.com/salto-io/salto/pull/5009) with making more promises go in parallel (using `Promise.all`), and removing any variable of a promise (so we don't have unhandled promises).

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/netsuite-adapter/src/adapter.ts
+++ b/packages/netsuite-adapter/src/adapter.ts
@@ -257,44 +257,43 @@ export default class NetsuiteAdapter implements AdapterOperations {
     const updatedFetchQuery = changedObjectsQuery !== undefined
       ? andQuery(changedObjectsQuery, fetchQuery)
       : fetchQuery
+    const queries = { updatedFetchQuery, originFetchQuery: fetchQuery }
 
-    const serverTimeElements = serverTime !== undefined
-      ? await getOrCreateServerTimeElements(
-        serverTime,
-        this.elementsSource,
-        isPartial,
-      )
-      : []
-
-    const dataElementsPromise = getDataElements(this.client, fetchQuery, this.getElemIdFunc)
-
-    const getCustomObjectsResult = this.client.getCustomObjects(
-      getStandardTypesNames(),
-      { updatedFetchQuery, originFetchQuery: fetchQuery }
-    )
-    const importFileCabinetResult = this.client.importFileCabinetContent(
-      updatedFetchQuery,
-      this.userConfig.client?.maxFileCabinetSizeInGB ?? DEFAULT_MAX_FILE_CABINET_SIZE_IN_GB
-    )
     progressReporter.reportProgress({ message: 'Fetching file cabinet items' })
-
-    const bundlesCustomInfo = (await this.client.getInstalledBundles())
-      .map(bundle => ({ typeName: BUNDLE, values: { ...bundle, id: bundle.id.toString() } }))
-    // TODO: remove this log when SALTO-2602 is open for all customers
-    log.debug('The following bundle ids are missing in the bundle record: %o', bundlesCustomInfo.map(bundle => [bundle.values.id, bundle.values.installedFrom?.toUpperCase(), bundle.values.publisher]))
     const {
       elements: fileCabinetContent,
       failedPaths: failedFilePaths,
-    } = await importFileCabinetResult
+    } = await this.client.importFileCabinetContent(
+      updatedFetchQuery,
+      this.userConfig.client?.maxFileCabinetSizeInGB ?? DEFAULT_MAX_FILE_CABINET_SIZE_IN_GB
+    )
     log.debug('importFileCabinetContent: fetched %d fileCabinet elements', fileCabinetContent.length)
 
     progressReporter.reportProgress({ message: 'Fetching instances' })
-    const {
-      elements: customObjects,
-      instancesIds,
-      failedToFetchAllAtOnce,
-      failedTypes,
-    } = await getCustomObjectsResult
+
+    const [
+      { elements: customObjects, instancesIds, failedToFetchAllAtOnce, failedTypes },
+      { elements: dataElements, requestedTypes: requestedDataTypes, largeTypesError: dataTypeError },
+      bundlesList,
+    ] = await Promise.all([
+      this.client.getCustomObjects(getStandardTypesNames(), queries),
+      getDataElements(this.client, fetchQuery, this.getElemIdFunc),
+      this.client.getInstalledBundles(),
+    ])
+
+    const bundlesCustomInfo = bundlesList.map(bundle => ({
+      typeName: BUNDLE,
+      values: { ...bundle, id: bundle.id.toString() },
+    }))
+    // TODO: remove this log when SALTO-2602 is open for all customers
+    log.debug(
+      'The following bundle ids are missing in the bundle record: %o',
+      bundlesCustomInfo.map(bundle => [
+        bundle.values.id,
+        bundle.values.installedFrom?.toUpperCase(),
+        bundle.values.publisher,
+      ])
+    )
 
     progressReporter.reportProgress({ message: 'Running filters for additional information' })
     const elementsToCreate = [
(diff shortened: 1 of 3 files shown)


packages/netsuite-adapter/src/adapter.ts
```diff
@@ -257,44 +257,43 @@ export default class NetsuiteAdapter implements AdapterOperations {
     const updatedFetchQuery = changedObjectsQuery !== undefined
       ? andQuery(changedObjectsQuery, fetchQuery)
       : fetchQuery
+    const queries = { updatedFetchQuery, originFetchQuery: fetchQuery }
```
dorsev, line 260: I like this PR 🙏 
One main comment is about `getConfigRecords`

I think it should fail fast on (invalid creds), and that does not happen today
See my comment here - https://github.com/salto-io/salto/pull/5009#discussion_r1371843042

Maybe we should address this here as well?
thoughts?

omrilit, line 260: nit: I find this variable confusing. maybe it's only a naming issue, but I am not sure that we have value to define it when it's used only in a single place and it's hard to find a proper name

dorsev: will let @omrilit approve this ofc. but from my point of view. LGTM 👍

packages/netsuite-adapter/src/adapter.ts
```diff
@@ -257,44 +257,43 @@ export default class NetsuiteAdapter implements AdapterOperations {
...
-
-    const bundlesCustomInfo = (await this.client.getInstalledBundles())
-      .map(bundle => ({ typeName: BUNDLE, values: { ...bundle, id: bundle.id.toString() } }))
-    // TODO: remove this log when SALTO-2602 is open for all customers
-    log.debug('The following bundle ids are missing in the bundle record: %o', bundlesCustomInfo.map(bundle => [bundle.values.id, bundle.values.installedFrom?.toUpperCase(), bundle.values.publisher]))
     const {
       elements: fileCabinetContent,
       failedPaths: failedFilePaths,
-    } = await importFileCabinetResult
+    } = await this.client.importFileCabinetContent(
```
omrilit, line 266: why did you decide to not parallelize importFileCabinetContent as well?
Is it only in order to be able to report correct progress? If so, I don't think that it's a good enough reason, yet it would be a challenge to report progress.

omrilit (commented):
Looks good, yet I think that the suggested solution still might have performance implications, unless I am missing anything here..

packages/netsuite-adapter/src/adapter.ts
omrilit: @dantal4 do you remember why we return an empty array when failing to fetch the config records? is it because it may happen consistently to certain customers?

packages/netsuite-adapter/src/adapter.ts
dantal4: indeed the progress reporting is the reason I did it. WDYT we should do?

omrilit: I **think** that calling validateCredentials explicitly would be a better solution in that case since overall I think that it'll provide better performance than the current suggestion, especially for large account where not parallelizing the fileCabinet fetching with the others may have big impact.

Other than that, theoretically we can try to perform a small SuiteApp & SDF queries (e.g. list objects & getServerTime and use the listObjects result in importObjects) before fetching everything but it would mess the code and I also think that it's better to validate explicitly in that case.

I have another idea which is super weird (and will not get implemented eventually, yet I thought that the concept could be interesting)
We can make some calls to return  Promise<ValueAndPromise>
```
type ValueAndPromise <T> = {
  isValidCredentials: boolean
  value: Promise<T>
}
```
and then we could return only after a simple code that ran and did some stuff that will approve whether the creds are valid and then will async run the other parts of the code.

omrilit: Maybe in SDF we could reuse the created SDF project from validateCredentials to the importObjects call.
It's also ugly I know..

packages/netsuite-adapter/src/adapter.ts
dantal4: We saw accounts that weren't able to fetch specific config types, but that is a different case. I think that we just never gave a deep thought on it.

packages/netsuite-adapter/src/adapter.ts
dantal4: We can add `importFileCabinetContent` to all other promises and pass it `reportOnDone: () => progressReporter.reportProgress({ message: 'Fetching instances' })`
This way we're keeping it parallel and the progress bar behavior will remain the same (still not accurate, because the instances fetch may finish before the fileCabinet)

dantal4: I changed it to run everything that we can in parallel

omrilit (approved):
Looks great, thanks!

packages/netsuite-adapter/src/adapter.ts
```diff
@@ -258,75 +259,95 @@ export default class NetsuiteAdapter implements AdapterOperations {
...
-    const getCustomObjectsResult = this.client.getCustomObjects(
-      getStandardTypesNames(),
-      { updatedFetchQuery, originFetchQuery: fetchQuery }
-    )
-    const importFileCabinetResult = this.client.importFileCabinetContent(
-      updatedFetchQuery,
-      this.userConfig.client?.maxFileCabinetSizeInGB ?? DEFAULT_MAX_FILE_CABINET_SIZE_IN_GB
-    )
-    progressReporter.reportProgress({ message: 'Fetching file cabinet items' })
+      progressReporter.reportProgress({ message: 'Fetching instances' })
```
dorsev, line 268: Nit: shouldn't this progressReporter, be submitted before `getCustomObjects` call is called?

Why is this reported here?

packages/netsuite-adapter/src/adapter.ts
```diff
@@ -258,75 +259,95 @@ export default class NetsuiteAdapter implements AdapterOperations {
...
-    } = await importFileCabinetResult
-    log.debug('importFileCabinetContent: fetched %d fileCabinet elements', fileCabinetContent.length)
+    const getStandardAndCustomElements = async (): Promise<{
+      standardInstances: InstanceElement[]
+      standardTypes: TypeElement[]
+      customRecordTypes: ObjectType[]
+      customRecords: InstanceElement[]
+      instancesIds: ObjectID[]
+      failures: FetchByQueryFailures
+    }> => {
```
dorsev, line 279: Nit: Consider extracting this type outside of this function? to improve readability. ?

packages/netsuite-adapter/src/adapter.ts
```diff
@@ -258,75 +259,95 @@ export default class NetsuiteAdapter implements AdapterOperations {
...
+          updatedFetchQuery,
+          originFetchQuery: fetchQuery,
+        }),
+        this.client.getInstalledBundles(),
+      ])
+      const bundlesCustomInfo = bundlesList.map(bundle => ({
+        typeName: BUNDLE,
+        values: { ...bundle, id: bundle.id.toString() },
+      }))
+      // TODO: remove this log when SALTO-2602 is open for all customers
```
dorsev, line 296: Nit: SALTO-2602 is "IN REVIEW". 
Is this log still needed?

packages/netsuite-adapter/src/adapter.ts
```diff
@@ -258,75 +259,95 @@ export default class NetsuiteAdapter implements AdapterOperations {
...
+        instancesIds,
+        failures,
+      },
+      { elements: dataElements, requestedTypes: requestedDataTypes, largeTypesError: dataTypeError },
+    ] = await Promise.all([
+      getStandardAndCustomElements(),
+      getDataElements(this.client, fetchQuery, this.getElemIdFunc),
+    ])
 
     progressReporter.reportProgress({ message: 'Running filters for additional information' })
```
dorsev, line 345: Nit: shouldn't this reportProgress be sent directly before we call `await this.createFiltersRunner(
      { isPartial, fetchTime: serverTime, timeZoneAndFormat, deletedElements }
    ).onFetch(elements)` ?

packages/netsuite-adapter/src/client/client.ts
```diff
@@ -545,6 +544,9 @@ export default class NetsuiteClient {
     try {
       return await this.suiteAppClient?.getSystemInformation()
     } catch (error) {
+      if (error instanceof InvalidSuiteAppCredentialsError) {
```
dorsev, line 547: Nit: shouldn't we add this "handling" to `getConfigRecords` as well?

dorsev (commented):
already merged, but reviewed as well.
all my comments are nits @dantal4

PR #5013, opened 2023-10-26
SALTO-4931: Allow sending DELETE requests with data

shir-reifenberg:
Until now we were not sending any payload on DELETE requests. 
To send data those request, we need to pass as part of the request config.

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/client/http_client.ts
+++ b/packages/adapter-components/src/client/http_client.ts
@@ -44,7 +44,7 @@ export type ClientBaseParams = {
 }
 
 export type ClientDataParams = ClientBaseParams & {
-  data: unknown
+  data?: unknown
 }
 
 export type ClientParams = ClientBaseParams | ClientDataParams
@@ -57,7 +57,7 @@ export interface HTTPReadClientInterface {
 export interface HTTPWriteClientInterface {
   post(params: ClientDataParams): Promise<Response<ResponseValue | ResponseValue[]>>
   put(params: ClientDataParams): Promise<Response<ResponseValue | ResponseValue[]>>
-  delete(params: ClientBaseParams): Promise<Response<ResponseValue | ResponseValue[]>>
+  delete(params: ClientDataParams): Promise<Response<ResponseValue | ResponseValue[]>>
   patch(params: ClientDataParams): Promise<Response<ResponseValue | ResponseValue[]>>
 }
 
@@ -66,8 +66,10 @@ export type HttpMethodToClientParams = {
   post: ClientDataParams
   put: ClientDataParams
   patch: ClientDataParams
-  delete: ClientBaseParams
+  delete: ClientDataParams
 }
+
+type MethodsWithDataParam = 'put' | 'post' | 'patch'
 
 export class HTTPError extends Error {
   constructor(message: string, readonly response: Response<ResponseValue>) {
@@ -80,6 +82,11 @@ export class TimeoutError extends Error {
 
 const isMethodWithData = (params: ClientParams): params is ClientDataParams =>
   'data' in params
+
+// Determines if the given HTTP method uses 'data' as the second parameter, based on APIConnection
+const isMethodWithDataParam = <T extends keyof HttpMethodToClientParams>(
+  method: T
+): method is T & MethodsWithDataParam => ['put', 'post', 'patch'].includes(method)
 
 export abstract class AdapterHTTPClient<
   TCredentials,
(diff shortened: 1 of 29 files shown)


packages/adapter-components/src/client/http_client.ts
```diff
@@ -44,7 +44,7 @@ export type ClientBaseParams = {
 }
 
 export type ClientDataParams = ClientBaseParams & {
-  data: unknown
+  data?: unknown
```
shir-reifenberg, line 47: I made this optional in case some API endpoints will block getting data on delete requests. @netama WDYT ?

packages/adapter-components/src/client/http_client.ts
```diff
@@ -231,15 +238,15 @@ export abstract class AdapterHTTPClient<
...
         ? await this.apiClient[method](
           url,
-          params.data,
+          isMethodWithData(params) ? params.data : undefined,
           requestConfig,
         )
         : await this.apiClient[method](
           url,
-          requestConfig,
+          { ...requestConfig, data: isMethodWithData(params) ? params.data : undefined },
```
netama, line 249: nit: maybe 
```suggestion
          { ...requestConfig, ..._.pick(params, 'data') },
```

packages/adapter-components/src/client/http_client.ts
```diff
@@ -231,15 +238,15 @@ export abstract class AdapterHTTPClient<
         }
         : undefined
 
-      const res = isMethodWithData(params)
+      const res = isMethodWithDataParam(method)
         ? await this.apiClient[method](
           url,
-          params.data,
+          isMethodWithData(params) ? params.data : undefined,
```
netama, line 244: nit: can't we just call 
```suggestion
          params.data,
```
? if it doesn't exist it will be undefined anyway, no?

netama (commented):
🔥

packages/adapter-components/src/client/http_client.ts
shir-reifenberg: typescript doesn't allow it as params might be `ClientBaseParams` that doesn't contain `data`

packages/jira-adapter/test/filters/data_center/priority_scheme/priority_scheme_fetch.test.ts
```diff
@@ -114,7 +114,7 @@ describe('prioritySchemeFetchFilter', () => {
 
       expect(connection.get).toHaveBeenCalledWith(
         '/rest/api/2/priorityschemes',
-        undefined,
+        { data: undefined },
```
netama, line 117: nit: this seems a bit odd for `GET` requests. not a major issue, but I think it might be worth playing a bit with the input in the client (even if it makes it slightly less clean) so that these are either `undefined` or an empty object - just because there is no reason for `data` to be singled out 
(one option is to pick only the keys that have defined values)

packages/adapter-components/src/client/http_client.ts
```diff
@@ -231,22 +238,15 @@ export abstract class AdapterHTTPClient<
...
-        ? await this.apiClient[method](
-          url,
-          params.data,
-          requestConfig,
-        )
-        : await this.apiClient[method](
-          url,
-          requestConfig,
-        )
+      const data = isMethodWithData(params) ? params.data : undefined
```
netama, line 241: let's add unittests for delete with/without a body?

packages/adapter-components/src/deployment/deployment.ts
```diff
@@ -132,6 +132,10 @@ export const deployChange = async ({
     return undefined
   }
   log.trace(`deploying instance ${instance.elemID.getFullName()} with params ${inspectValue({ method: endpoint.method, url, queryParams, data }, { compact: true, depth: 6 })}`)
-  const response = await client[endpoint.method]({ url, data, queryParams })
+  const response = await client[endpoint.method]({
+    url,
+    data: endpoint.omitRequestBody ? undefined : data,
```
netama, line 137: let's cover this with a unittest?

netama (approved):
looks great :) added some final nits (but good to merge otherwise)

packages/jira-adapter/test/filters/data_center/priority_scheme/priority_scheme_fetch.test.ts
shir-reifenberg: agree, changed 👍

PR #5017, opened 2023-10-26
SALTO-4727: StandardValueSets references Support for fetch with changes detection mode

tamtamirr:
Modified the SVS filter to support the upcoming support in CustomObjects fetchWithChangesDetection.

---

As part of testing **CustomObjects** quick fetch, I've noticed that we lose references on **Standard Picklist Fields**, on objects that were updated since the last fetch. I've changed the filter's implementation to support that.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/filters/standard_value_sets.ts
```diff
@@ -258,17 +261,13 @@ export const makeFilter = (
...
-              const clone = inst.clone()
-              await resolveTypeShallow(clone, config.elementsSource)
-              return clone
-            })
-            .filter(isInstanceOfType(STANDARD_VALUE_SET))
+        const svsInstances = config.fetchProfile.metadataQuery.isFetchWithChangesDetection()
+          // In fetch with changes detection mode, we need all the StandardValueSets
+          // in order not to lose existing references on fetched CustomObjects.
+          ? await awu(await elementsSource.getAll())
+            .filter(isInstanceOfTypeSync(STANDARD_VALUE_SET))
```
tomsellek, line 268: 🔥 🔥 🔥

packages/salesforce-adapter/src/filters/standard_value_sets.ts
```diff
@@ -258,17 +261,13 @@ export const makeFilter = (
...
             .toArray()
+          : fetchedSVSInstances
```
tomsellek, line 270: Is this a performance optimization or is there a reason not to just use the element source in all cases?

tomsellek (approved):
LGTM!

packages/salesforce-adapter/src/filters/standard_value_sets.ts
tamtamirr: Breaking regular fetch is more concerning, that's why I try to keep the separation  between the modes as strict as possible. However, once we have enough confidence in **Fetch With Changes Detection**, we can safely align the logics in many places.

PR #5018, opened 2023-10-26
SALTO-4874 - Zendesk Adapter - Make business_hours_schedule_holiday elemIDs more unique

idoser96:
Add start_year and end_year hidden fields to business_hour_schedule_holiday
use the start_year in the elemID to prevent duplications

---

None

---
_Release Notes_: 
Zendesk:
* Add the start year of an holiday to the elemID

---
_User Notifications_: 
None


packages/zendesk-adapter/src/config.ts
```diff
@@ -1588,10 +1588,18 @@ export const DEFAULT_TYPES: ZendeskApiConfig['types'] = {
       },
     },
     transformation: {
+      idFields: ['name', '&start_year'],
```
edenhassid, line 1591: @netama do we want to change the elemId for all users?

edenhassid (approved):
looks good! pre-approving. we should figure out if we want to change the elemId for all users

PR #5021, opened 2023-10-27
SALTO-4939 - Salesforce: validate managed-by-Salto configuration

tomsellek:
Catch some cases of misconfiguration and issue a warning about them

---

The most pressing issue is that when the field exists but is of the wrong type we get a user-visible cryptic SOQL error.

Tests:
 - Fetched with the management field in the `omittedFields` list and got a configuration error
 - Fetched with a type where the field exists but is a number and got a warning for the type and no records fetched
 - Fetched with only a single type that had the wrong type for the field and got a warning and no records fetched

image

image

image

---
_Release Notes_: 
Salesforce: improved validation of saltoManagementSettings configuration

---
_User Notifications_: 
N/A


packages/salesforce-adapter/src/filters/custom_objects_instances.ts
```diff
@@ -561,10 +565,21 @@ const createInvalidAliasFieldFetchWarning = async (
 
 const createInvalidManagedBySaltoFieldFetchWarning = async (
   { objectType, invalidManagedBySaltoField }: CustomObjectFetchSetting
```
tamtamirr, line 567: L569 comment hinted on a code smell that can be avoided by enforcing it on the received type.
This can be done in many ways, for example: 
```suggestion
const createInvalidManagedBySaltoFieldFetchWarning = async (
  { objectType, invalidManagedBySaltoField }: CustomObjectFetchSettings & Required<Pick<CustomObjectFetchSettings, 'invalidManagedBySaltoField'>>
```

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
```diff
@@ -561,10 +565,21 @@ const createInvalidAliasFieldFetchWarning = async (
...
-): Promise<SaltoError> => ({
-  message: `The field ${await apiName(objectType)}${API_NAME_SEPARATOR}${invalidManagedBySaltoField} is configured as the filter field in the saltoManagementFieldSettings.defaultFieldName section of the Salto environment configuration. However, the user configured for fetch does not have read access to this field. Records of type ${await apiName(objectType)} will not be fetched.`,
-  severity: 'Warning',
-})
+): Promise<SaltoError> => {
+  const fieldName = invalidManagedBySaltoField as string // if we got here the field name can't be undefined
+  const errorPreamble = `The field ${await apiName(objectType)}${API_NAME_SEPARATOR}${invalidManagedBySaltoField} is configured as the filter field in the saltoManagementFieldSettings.defaultFieldName section of the Salto environment configuration.`
+  if ((objectType.fields[fieldName].annotations[FIELD_ANNOTATIONS.QUERYABLE] ?? true) === false) {
```
tamtamirr, line 571: I think this whole method should be simplified. There's not reason to receive anything but the actual **managedBySaltoField**. Also, it seems to be using deprecated code (e.g. apiName), and access a Field without testing whether it exists on the type (objectType.fields[fieldName].annotations)
Changing the input here to **managedBySaltoField: Field** will greatly simplify the code and avoid these issues.

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
```diff
@@ -591,6 +606,20 @@ const filterCreator: RemoteFilterCreator = ({ client, config }) => ({
...
       setting => setting.invalidIdFields.length === 0 && setting.invalidManagedBySaltoField === undefined
     )
+
+    if (validFetchSettings.length === 0
+    && invalidFetchSettings.length > 0
+    && invalidFetchSettings.every(setting => setting.invalidManagedBySaltoField !== undefined)) {
+      return {
+        errors: [
+          {
+            message: 'The Salto environment is configured to use a Salto Management field but the field is missing or is of the wrong type for all data records. No data records will be fetched. Please verify the field name in saltoManagementFieldSettings.defaultFieldName in the environment configuration file.',
```
tamtamirr, line 616:
```suggestion
            message: 'The Salto environment is configured to use the Salto Management field "${fieldName}" which seem to be invalid on all of the Custom Objects. No data records will be fetched. Please verify this field exists and is of supported type.',
```
I think we can improve the message here. Also, @pgonzaleznetwork @tomermevorach will we have an article explaining this? Perhaps we can include it later as part of this warning with more detailed explanation.

packages/salesforce-adapter/test/filters/custom_objects_instances.test.ts
```diff
@@ -1641,63 +1641,159 @@ describe('Custom Object Instances filter', () => {
...
-      setMockQueryResults(testRecords)
+        testRecords.forEach(record => { record[MANAGED_BY_SALTO_FIELD_NAME] = true })
+        setMockQueryResults(testRecords)
+
+        filterResult = await filter.onFetch(elements) as FilterResult
+      })
 
-      filterResult = await filter.onFetch(elements) as FilterResult
+      it('Should warn', () => {
+        expect(filterResult.errors ?? []).not.toBeEmpty()
```
tamtamirr, line 1712: Nit: This expect is redundant IMO

packages/salesforce-adapter/test/filters/custom_objects_instances.test.ts
```diff
@@ -1641,63 +1641,159 @@ describe('Custom Object Instances filter', () => {
...
+        expect(filterResult.errors).toEqual([{
+          message: expect.stringContaining('TestType') && expect.stringContaining(MANAGED_BY_SALTO_FIELD_NAME),
+          severity: 'Warning',
+        }])
+        expect(elements).toHaveLength(2)
```
tamtamirr, line 1717: What are we asserting here?

packages/salesforce-adapter/test/filters/custom_objects_instances.test.ts
```diff
@@ -1641,63 +1641,159 @@ describe('Custom Object Instances filter', () => {
...
+        expect(basicQueryImplementation).not.toHaveBeenCalledWith(expect.stringContaining('TestType'))
+      })
     })
+    describe('when all the types have inqccessible fields', () => {
```
tamtamirr, line 1721:
```suggestion
    describe('when all the types have invalid fields', () => {
```

tamtamirr (commented):
Looking good. Added some comments

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
```diff
@@ -487,7 +488,10 @@ export const getCustomObjectsFetchSettings = async (
     if (managedBySaltoFieldName === undefined) {
       return false
     }
+    const fieldType = type.fields[managedBySaltoFieldName].getTypeSync()
```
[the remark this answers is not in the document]
tomsellek: If the field does not exist on the type then `dataManagement.managedBySaltoFieldForType(type)` will return `undefined` and we won't reach here. 
I will add a UT, though.

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
```diff
@@ -487,7 +488,10 @@ export const getCustomObjectsFetchSettings = async (
...
     return (type.fields[managedBySaltoFieldName].annotations[FIELD_ANNOTATIONS.QUERYABLE] ?? true) === false
+    || !isPrimitiveType(fieldType)
+    || !fieldType.isEqual(Types.primitiveDataTypes.Checkbox)
```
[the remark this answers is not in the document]
tomsellek: I changed the name to make it hopefully more self-documenting

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
tomsellek: Ack

packages/salesforce-adapter/test/filters/custom_objects_instances.test.ts
tomsellek: Ack

packages/salesforce-adapter/test/filters/custom_objects_instances.test.ts
tomsellek: That the filter did not add any instances

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
tomsellek: @tomermevorach ? @pgonzaleznetwork ?

tamtamirr (approved):
LGTM code-wise

PR #5022, opened 2023-10-29
SALTO-4925: Make okta config options visible in the adapter's nacl

shir-reifenberg:
Make okta config options visible in the adapter's nacl

---

We recently added many config options to Okta. Making those visible will make it easier to CE to follow.
Only changed config options that I believe can be useful in first fetch

The new adapter config default - 

```
okta {
  fetch = {
    include = [
      {
        type = ".*"
      },
    ]
    exclude = [
    ]
    convertUsersIds = true
    includeGroupMemberships = false
    includeProfileMappingProperties = true
  }
}
```

---
_Release Notes_: 

---
_User Notifications_: 


netama (approved):
we should probably keep track somewhere of the envs that don't have these visible, so that we don't accidentally assume everyone has it in the future

PR #5024, opened 2023-10-29
SALTO-4727: TopicsForObjects support for Fetch With Changes Detection 

tamtamirr:
This PR modifies the **topicsForObjects** filter to support fetch with changes detection, as a required step for quick fetch for **CustomObjects**.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


packages/salesforce-adapter/test/filters/topics_for_objects.test.ts
```diff
@@ -40,48 +52,77 @@ describe('Topics for objects filter', () => {
...
-      [constants.INSTANCE_FULL_NAME_FIELD]: 'Test__c',
-    })
-
-  let filter: FilterWith<'onFetch' | 'preDeploy' | 'onDeploy'>
+    let topicsForObjectsMetadataType: ObjectType
+    let topicsForObjectsType: ObjectType
+    let topicsForObjectsInstance: InstanceElement
+    let elements: Element[]
 
-  describe('onFetch', () => {
```
tamtamirr, line 68: Some refactor done here. Quite simple:
1. The previous **onFetch** test, validated that the MetadataType **TopicsForObjects** was removed. Changed the implementation to test that the type is set to hidden and not removed.
2. Sharing logic and preparation between the UTs simplified it's code.

packages/salesforce-adapter/src/filters/topics_for_objects.ts
```diff
@@ -58,20 +67,37 @@ const createTopicsForObjectsInstance = (values: TopicsForObjectsInfo): InstanceE
...
     if (_.isEmpty(customObjectTypes)) {
       return
     }
 
     const topicsForObjectsInstances = await getInstancesOfMetadataType(elements,
       TOPICS_FOR_OBJECTS_METADATA_TYPE)
-    if (_.isEmpty(topicsForObjectsInstances)) {
-      return
-    }
-
```
tamtamirr, line 74: Removed this statement, as this is incorrect for `fetch with changes detection`. L85 should achieve the same.

packages/salesforce-adapter/src/filters/topics_for_objects.ts
```diff
@@ -58,36 +67,94 @@ const createTopicsForObjectsInstance = (values: TopicsForObjectsInfo): InstanceE
...
+    iter: await elementsSource.getAll(),
+    filter: isCustomObjectWithTopics,
+    key: obj => [apiNameSync(obj) ?? ''],
+    map: obj => obj.annotations.topicsForObjects.enableTopics,
+  })
+  customObjects.forEach(customObject => {
+    const typeApiName = apiNameSync(customObject) ?? 'unknown'
+    const isTopicsEnabled = isTopicsEnabledByType[typeApiName]
+      ? isTopicsEnabledByType[typeApiName]
+      : isTopicsEnabledForObjectFromSource.get(typeApiName)
```
tomsellek, line 106: What if the change to the object was that topics were disabled? Won't this re-enable them?

tomsellek (commented):
Overall looks good, but I did have one question.

packages/salesforce-adapter/src/filters/topics_for_objects.ts
tamtamirr: The value of the annotation comes from the **TopicsForObjects** instance.
If there was a change in the instance (Or for new CustomObject, addition of corresponding **TopicsForObjects** instance as-well) it will be fetched, and we will get the value from the instance and not from the elements source. Otherwise we get the value from the elements source.

tomsellek: I'm not sure I follow, perhaps we should discuss this F2F? Isn't `isTopicsEnabledByType` the value you get from the fetch? If it is, and it's false, then won't you end up using the value from `isTopicsEnabledForObjectFromSource`?

tamtamirr: gotcha

packages/salesforce-adapter/src/filters/topics_for_objects.ts
```diff
@@ -58,37 +73,93 @@ const createTopicsForObjectsInstance = (values: TopicsForObjectsInfo): InstanceE
...
+    log.debug('isTopicsEnabledByType in fetchWithChangesDetection: %o', isTopicsEnabledByType)
+  }
+  const isTopicsEnabledForObjectFromSource = await multiIndex.keyByAsync({
+    iter: await elementsSource.getAll(),
+    filter: isCustomObjectWithTopics,
+    key: obj => [apiNameSync(obj) ?? ''],
+    map: obj => obj.annotations.topicsForObjects.enableTopics,
+  })
+  customObjects.forEach(customObject => {
+    const typeApiName = apiNameSync(customObject) ?? 'unknown'
```
tomsellek, line 109: Why not bail out if `customObject` has no API name?

packages/salesforce-adapter/src/filters/topics_for_objects.ts
```diff
@@ -58,37 +73,93 @@ const createTopicsForObjectsInstance = (values: TopicsForObjectsInfo): InstanceE
...
+    setTopicsForObjects(customObject, isTopicsEnabled)
+  })
+}
+
+const filterCreator: LocalFilterCreator = ({ config }) => ({
   name: 'topicsForObjectsFilter',
   onFetch: async (elements: Element[]): Promise<void> => {
+    if (!config.fetchProfile.metadataQuery.isTypeMatch(TOPICS_FOR_OBJECTS_METADATA_TYPE)) {
+      log.debug('skipping topicsForObjectsFilter since the MetadataType TopicsForObjects is excluded')
+    }
```
tomsellek, line 126: shouldn't you `return` here?

tomsellek (commented):
Overall looks good but I had a couple of questions

tomsellek (approved):
LGTM!

PR #5030, opened 2023-10-29
SALTO-4727: fetch with changes detection on custom objects

tamtamirr:
This PR adds support for **CustomObjects**, their fields and their sub-instances in fetch with changes detection mode.

---

This PR is rebased on top https://github.com/salto-io/salto/pull/5024 so do not review the changes of the `topics_for_objects.ts` filter and it's test file.

This PR is also rebased on top this small refactor: https://github.com/salto-io/salto/pull/5053

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

PR #5031, opened 2023-10-30
SALTO-4727: Remove creation of the fetchProfile from the SalesforceAdapter constructor  

tamtamirr:
Moved the creation of the fetchProfile from the SalesforceAdapter constructor  to the operations themselves.

---

This is a necessary refactor, since in https://github.com/salto-io/salto/pull/5030 the **fetchProfile** receives the promise of **relatedPropsByMetadataType** that should only start executing in fetch, and not in the creation of the adapter. (As we invoke client calls there).
We should never leave hanging promises that won't be necessarily awaited in our code.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


tomsellek (approved):
Love this!

PR #5035, opened 2023-10-30
SALTO-4957: change return value for getImportantValues func

edenhassid:
change return value for getImportantValues func

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/adapter-utils/src/important_values.ts
```diff
@@ -30,6 +30,7 @@ const { isDefined } = values
 
 type ImportantValue = { value: string; indexed: boolean; highlighted: boolean }
 export type ImportantValues = ImportantValue[]
+type formattedImportantValueData = {key: string; value: Value}
```
netama, line 33:
```suggestion
type FormattedImportantValueData = { key: string; value: Value }
```

netama (approved):
LGTM with one nit

PR #5040, opened 2023-11-01
SALTO-4961 Zendesk - return early in irrelevant filters, avoid too many concurrent calls to element source

netama:
Adding early returns in a few filters to avoid doing unneeded work.

The motivation for this was an OOM we saw for macro deployments, that seems to at least be partially caused by attempting to initialize the element source many times in parallel, likely for the `ticketFormDeploy`'s `isCustomStatusesEnabled` call - but using this to remove some other cases as well (focusing for now on the ones that do not require a bigger refactor).

(besides this, we should also investigate the issue directly in the element source, but starting with a mitigation)

---
_Release Notes_: 
_Zendesk adapter_:
* Fix potential crash on large deployments due to memory consumption

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/filters/article/article.ts
+++ b/packages/zendesk-adapter/src/filters/article/article.ts
@@ -482,7 +482,7 @@ const filterCreator: FilterCreator = ({ config, client, elementsSource, brandIdT
       }
 
       const everyoneUserSegmentElemID = new ElemID(ZENDESK, USER_SEGMENT_TYPE_NAME, 'instance', EVERYONE_USER_TYPE)
-      const everyoneUserSegmentInstance = await elementsSource.get(everyoneUserSegmentElemID) // ?
+      const everyoneUserSegmentInstance = await elementsSource.get(everyoneUserSegmentElemID)
       relevantChanges
         .map(getChangeData)
         .forEach(articleInstance => {
--- a/packages/zendesk-adapter/src/filters/article/article.ts
+++ b/packages/zendesk-adapter/src/filters/article/article.ts
@@ -58,7 +58,7 @@ import {
   isAttachments,
   updateArticleTranslationBody,
 } from './utils'
-import { API_DEFINITIONS_CONFIG } from '../../config'
+import { API_DEFINITIONS_CONFIG, FETCH_CONFIG, isGuideEnabled } from '../../config'
 
 
 const log = logger(module)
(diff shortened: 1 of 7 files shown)


packages/zendesk-adapter/src/filters/ticket_form.ts
```diff
@@ -151,14 +151,23 @@ const getChangeWithoutRemovedFields = (change: ModificationChange<InstanceElemen
 const filterCreator: FilterCreator = ({ config, client, elementsSource }) => ({
   name: 'ticketFormDeploy',
   deploy: async (changes: Change<InstanceElement>[]) => {
-    const [TicketFormChanges, leftoverChanges] = _.partition(
+    const [ticketFormChanges, leftoverChanges] = _.partition(
       changes,
       change => TICKET_FORM_TYPE_NAME === getChangeData(change).elemID.typeName,
     )
+    if (ticketFormChanges.length === 0) {
```
netama, line 158: this is the important case - returning early to avoid calling `isCustomStatusesEnabled` for every change group

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -468,10 +476,14 @@ const filterCreator: FilterCreator = ({ config, client, elementsSource, brandIdT
...
+      const relevantChanges = changes.filter(change => getChangeData(change).elemID.typeName === ARTICLE_TYPE_NAME)
+      if (relevantChanges.length === 0) {
+        return
+      }
+
       const everyoneUserSegmentElemID = new ElemID(ZENDESK, USER_SEGMENT_TYPE_NAME, 'instance', EVERYONE_USER_TYPE)
-      const everyoneUserSegmentInstance = await elementsSource.get(everyoneUserSegmentElemID)
-      changes
-        .filter(change => getChangeData(change).elemID.typeName === ARTICLE_TYPE_NAME)
+      const everyoneUserSegmentInstance = await elementsSource.get(everyoneUserSegmentElemID) // ?
```
edenhassid, line 485: nit:
```suggestion
      const everyoneUserSegmentInstance = await elementsSource.get(everyoneUserSegmentElemID)
```

edenhassid (approved):
looks good! 😄 added a small nit

packages/zendesk-adapter/src/filters/article/article.ts
netama: oops

PR #5041, opened 2023-11-01
SALTO-4965: Better Data Instances Unknown exception handling

tamtamirr:
- Add log on thrown unknown error in Salesforce Data Deploy group.
- Do not expose unknown runtime errors.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/src/custom_object_instances_deploy.ts
+++ b/packages/salesforce-adapter/src/custom_object_instances_deploy.ts
@@ -487,6 +487,12 @@ const isModificationChangeList = <T>(
     changes.every(isModificationChange)
   )
 
+
+const customObjectInstancesDeployError = (message: string): DeployResult => ({
+  appliedChanges: [],
+  errors: [{ message, severity: 'Error' }],
+})
+
 const deploySingleTypeAndActionCustomObjectInstancesGroup = async (
   changes: ReadonlyArray<Change<InstanceElement>>,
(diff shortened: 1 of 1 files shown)


packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -498,19 +504,19 @@ const deploySingleTypeAndActionCustomObjectInstancesGroup = async (
     const instanceTypes = [...new Set(await awu(instances)
       .map(async inst => apiName(await inst.getType())).toArray())]
     if (instanceTypes.length > 1) {
-      throw new Error(`Custom Object Instances change group should have a single type but got: ${instanceTypes}`)
+      return customObjectInstancesDeployError(`Custom Object Instances change group should have a single type but got: ${instanceTypes}`)
```
amitgoldarb, line 507: Aren't these redundant if we are wrapping this in the catch?

amitgoldarb (approved):
Added a small comment. Approving to save cycles. Thanks for this

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
tamtamirr: No. I've changed the user-visible error within the catch. This catch clause now should only catch unhandled errors.

PR #5042, opened 2023-11-01
SALTO-4964: Fail to deploy global automation in Jira DC

shayc331:
_Fail to deploy global automation in Jira DC_

---

* it was caused by this [change](https://github.com/salto-io/salto/pull/4885) that affected only cloud users.
* I added a simple change that removes the project list when it's empty to align with the cloud behavior.

---
_Release Notes_: 
Jira Adapter: 
* fix a bug in deploying global Automation using Jira DC

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/jira-adapter/src/filters/automation/automation_structure.ts
+++ b/packages/jira-adapter/src/filters/automation/automation_structure.ts
@@ -335,6 +335,12 @@ const convertRuleScopeToProjects = (instance: InstanceElement): void => {
   instance.value.projects = convertRuleScopeValueToProjects(instance.value)
 }
 
+const handleDCProjects = (instance: InstanceElement): void => {
+  if (instance.value.projects.length === 0) {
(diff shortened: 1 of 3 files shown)


packages/jira-adapter/src/filters/automation/automation_structure.ts
```diff
@@ -335,6 +335,12 @@ const convertRuleScopeToProjects = (instance: InstanceElement): void => {
   instance.value.projects = convertRuleScopeValueToProjects(instance.value)
 }
 
+const handleDCProjects = (instance: InstanceElement): void => {
```
omrilit, line 338: Maybe we can think of a more self explained naming here? something in the sense of `removeProjectsForGlobalDCAutomation` WDYT?

omrilit (approved):
Looks great! thanks for that.
Had only 1 naming comment

PR #5044, opened 2023-11-02
Revert "added missing annotations to workflow (#4968)"

netama:
This reverts commit 6725c81861e618b04f7179eb537e0f627b866bb0.

Not all cases are supported by the old workflow APIs - reverting for now until this is tested more thoroughly

---

---
_Release Notes_: 

_Jira adapter_:
* Temporarily restricting deployment of some parts of workflows that are not fully supported using the current APIs

---
_User Notifications_: 
None


omrilit (approved):
thanks for that!

PR #5045, opened 2023-11-02
SALTO-4780 - Salesforce: convert profile and permission sets to custom references

tomsellek:
Create references from profiles and permission sets in `getCustomReferences`

---

1. Note how I create the ElemID name for the references. Given the fullName, the correct thing to do would be to look up the element with this full name and use its ElemID. However, in `getCustomReferences` we are supposed to deal with elements one at a time with no context, so that's not an option. I basically replicated how we create ElemID from fullName where we create instances.
2. ~Ideally, this would be an optional feature, but I'm not sure how to get access to the fetch profile from this callback~

** PR Chain **
 - https://github.com/salto-io/salto/pull/5150 
 - https://github.com/salto-io/salto/pull/5045 👈 YOU ARE HERE 👈
---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/custom_references.ts
+++ b/packages/salesforce-adapter/src/custom_references.ts
@@ -1,5 +1,5 @@
 /*
-*                      Copyright 2023 Salto Labs Ltd.
+*                      Copyright 2024 Salto Labs Ltd.
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with
--- a/packages/salesforce-adapter/src/custom_references.ts
+++ b/packages/salesforce-adapter/src/custom_references.ts
@@ -13,6 +13,7 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
+import _ from 'lodash'
 import { collections } from '@salto-io/lowerdash'
 import { logger } from '@salto-io/logging'
 import {
@@ -40,7 +41,7 @@ const RECORD_TYPE_SECTION = 'recordTypeVisibilities'
 
 const FIELD_NO_ACCESS = 'NoAccess'
 
-const fullNameToElemIdName = (fullName: string): string => (
+const getMetadataElementName = (fullName: string): string => (
   Types.getElemId(
     fullName.replace(API_NAME_SEPARATOR, '_'),
     true,
@@ -54,16 +55,27 @@ type ReferenceInSection = {
 
 type RefTargetsGetter = (sectionEntry: Values, sectionEntryKey: string) => ReferenceInSection[]
 
-const referencesFromSection = (
-  profile: InstanceElement,
-  sectionName: string,
-  filter: (sectionEntry: Values) => boolean,
-  targetsGetter: RefTargetsGetter,
-): ReferenceInfo[] => {
-  if (!profile.value[sectionName]) {
+type ReferenceFromSectionParams = {
+  profile: InstanceElement
+  sectionName: string
+  filter?:(sectionEntry: Values) => boolean
+  targetsGetter:RefTargetsGetter
+}
+
+const referencesFromSection = ({
+  profile,
+  sectionName,
+  filter = () => true,
+  targetsGetter,
+}: ReferenceFromSectionParams): ReferenceInfo[] => {
+  const sectionValue = profile.value[sectionName]
+  if (!_.isPlainObject(sectionValue)) {
+    if (sectionValue !== undefined) {
+      log.warn('Section %s of %s is not an object. References will not be extracted.', sectionName, profile.elemID)
+    }
     return []
   }
-  return Object.entries(profile.value[sectionName] as Values)
+  return Object.entries(sectionValue as Values)
     .filter(([, sectionEntry]) => filter(sectionEntry))
     .flatMap(([sectionEntryKey, sectionEntry]): ReferenceInfo[] => {
       const refTargets = targetsGetter(sectionEntry, sectionEntryKey)
@@ -76,16 +88,18 @@ const referencesFromSection = (
 }
 
 const isEnabled = (sectionEntry: Values): boolean => (
-  sectionEntry.enabled
+  sectionEntry.enabled === true
 )
 
 const isAnyAccessEnabledForObject = (objectAccessSectionEntry: Values): boolean => (
-  objectAccessSectionEntry.allowCreate
-  || objectAccessSectionEntry.allowDelete
-  || objectAccessSectionEntry.allowEdit
-  || objectAccessSectionEntry.allowRead
-  || objectAccessSectionEntry.modifyAllRecords
-  || objectAccessSectionEntry.viewAllRecords
+  [
+    objectAccessSectionEntry.allowCreate,
+    objectAccessSectionEntry.allowDelete,
+    objectAccessSectionEntry.allowEdit,
+    objectAccessSectionEntry.allowRead,
+    objectAccessSectionEntry.modifyAllRecords,
+    objectAccessSectionEntry.viewAllRecords,
+  ].some(permission => permission === true)
 )
 
 const isAnyAccessEnabledForField = (fieldPermissionsSectionEntry: Values): boolean => (
@@ -94,7 +108,7 @@ const isAnyAccessEnabledForField = (fieldPermissionsSectionEntry: Values): boole
 
 const referenceToInstance = (fieldName: string, targetType: string): RefTargetsGetter => (
   sectionEntry => {
-    const elemIdName = fullNameToElemIdName(sectionEntry[fieldName])
+    const elemIdName = getMetadataElementName(sectionEntry[fieldName])
     return [{
       target: Types.getElemId(targetType, false).createNestedID('instance', elemIdName),
     }]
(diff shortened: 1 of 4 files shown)


packages/salesforce-adapter/src/custom_references.ts
```diff
@@ -0,0 +1,214 @@
...
+const referenceToType = (fieldName: string): RefTargetsGetter => (
+  sectionEntry => [{
+    target: Types.getElemId(sectionEntry[fieldName], true),
+  }]
+)
+
+const referencesToFields: RefTargetsGetter = (sectionEntry, sectionEntryKey) => {
+  const typeElemId = Types.getElemId(sectionEntryKey, true)
+  return Object.entries(sectionEntry)
+    .filter(([, fieldAccess]) => fieldAccess !== FIELD_NO_ACCESS)
```
amitgoldarb, line 110: We've already filtered these out with isAnyAccessEnabledForField no?

amitgoldarb (commented):
This looks very good!
I have a few concerns here:
1. This will make the rocksDB index potentially a lot larger due to lots of new references we are adding. I think we should test how this affects performance of other workspace functionality.
2. I'm sure that whatever testing we do for phase 1 locally, will be challenged by our clients. We should have a core flag to toggle this behaviour.
Thanks for this PR. I'll wait for approval until we have an answer to both of my comments above.

packages/salesforce-adapter/src/custom_references.ts
```diff
@@ -0,0 +1,214 @@
...
+    .concat(fieldPermissionsRefs)
+    .concat(layoutAssignmentRefs)
+    .concat(recordTypeRefs)
+}
+
+export const getCustomReferences: GetCustomReferencesFunc = async (elements: Element[]): Promise<ReferenceInfo[]> => (
+  log.time(() => (elements
+    .filter(isInstanceOfTypeSync(PROFILE_METADATA_TYPE, PERMISSION_SET_METADATA_TYPE))
+    .flatMap(referencesFromProfile)),
+  'Generating references from profiles')
```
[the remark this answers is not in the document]
tomsellek: Ack

packages/salesforce-adapter/src/custom_references.ts
tomsellek: yuppers

packages/salesforce-adapter/src/custom_references.ts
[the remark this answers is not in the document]
tomsellek: image

tomsellek:
@amitgoldarb @tamtamirr This PR is now ready for review.

packages/salesforce-adapter/src/types.ts
```diff
@@ -291,6 +302,7 @@ export type DataManagementConfig = {
   saltoManagementFieldSettings?: SaltoManagementFieldSettings
   brokenOutgoingReferencesSettings?: BrokenOutgoingReferencesSettings
   omittedFields?: string[]
+  [CUSTOM_REFS_CONFIG]?: CustomReferencesSettings
```
amitgoldarb, line 305: Can we possibly have a default for this? Similar to what you implemented in the past for optionalFeatures? I think we would benifit from this, like we did for optionalFeatures.

amitgoldarb (approved):
Approving to save cycles. Added a small question

packages/salesforce-adapter/src/custom_references.ts
```diff
@@ -0,0 +1,237 @@
...
+      target: new ElemID(SALESFORCE, RECORD_TYPE_METADATA_TYPE, 'instance', fullNameToElemIdName(recordTypeVisibility.recordType)),
+      sourceField: recordTypeVisibilityKey,
+    }))
+)
+
+const referencesFromProfile = (profile: InstanceElement): ReferenceInfo[] => {
+  const appVisibilityRefs = referencesFromSection(
+    profile,
+    APP_VISIBILITY_SECTION,
+    appVisibilityEntry => appVisibilityEntry.default || appVisibilityEntry.visible,
```
tamtamirr, line 151:
```suggestion
    appVisibilityEntry => appVisibilityEntry.default === true || appVisibilityEntry.visible === true, 
```

packages/salesforce-adapter/src/custom_references.ts
```diff
@@ -0,0 +1,237 @@
...
+  )
+
+  return appVisibilityRefs
+    .concat(apexClassRefs)
+    .concat(flowRefs)
+    .concat(apexPageRefs)
+    .concat(objectRefs)
+    .concat(fieldPermissionsRefs)
+    .concat(layoutAssignmentRefs)
+    .concat(recordTypeRefs)
```
tamtamirr, line 209: Can we add a log with how many references we created per type?
I think this is good for visibility.
I would include the total amount of refs here as-well and remove the log in L222

packages/salesforce-adapter/test/custom_references.test.ts
```diff
@@ -0,0 +1,603 @@
...
+* limitations under the License.
+*/
+import { InstanceElement, ReferenceInfo, Values, ObjectType, ElemID } from '@salto-io/adapter-api'
+import { INSTANCE_FULL_NAME_FIELD } from '../src/constants'
+import { mockTypes } from './mock_elements'
+import { createCustomObjectType, createMetadataTypeElement } from './utils'
+import { getCustomReferences } from '../src/custom_references'
+import { CUSTOM_REFS_CONFIG, DATA_CONFIGURATION, FETCH_CONFIG } from '../src/types'
+
+describe('getCustomReferences', () => {
```
tamtamirr, line 23: Can you add another nested `describe('profiles')` ? So when we add other `custom references` features in the future they will have their dedicated suite.

packages/salesforce-adapter/src/custom_references.ts
```diff
@@ -0,0 +1,237 @@
...
+    profile,
+    FIELD_PERMISSIONS,
+    isAnyAccessEnabledForField,
+    referencesToFields,
+  )
+
+  const layoutAssignmentRefs = referencesFromSection(
+    profile,
+    LAYOUTS_SECTION,
+    () => true,
```
tamtamirr, line 191: This one smells. Can we make the `filter` param optional instead?

packages/salesforce-adapter/src/custom_references.ts
```diff
@@ -0,0 +1,237 @@
...
+  const recordTypeRefs = sectionEntry
+    .filter((layoutAssignment: Values) => layoutAssignment.recordType !== undefined)
+    .map((layoutAssignment: Values) => ({
+      target: new ElemID(SALESFORCE, RECORD_TYPE_METADATA_TYPE, 'instance', fullNameToElemIdName(layoutAssignment.recordType)),
+    }))
+
+  return [
+    ...makeArray(layoutRef),
+    ...recordTypeRefs,
+  ]
```
tamtamirr, line 135:
```suggestion
    return [layoutRef].concat(recordTypeRefs)
```
nit

packages/salesforce-adapter/src/types.ts
```diff
@@ -268,6 +274,11 @@ const brokenOutgoingReferencesSettingsType = new ObjectType({
   },
 })
 
+const customReferencesSettingsType = new ObjectType({
+  elemID: new ElemID(constants.SALESFORCE, 'saltoCustomReferencesSettings'),
+  fields: Object.fromEntries(customReferencesTypeNames.map(name => [name, { refType: BuiltinTypes.BOOLEAN }])),
+})
```
tamtamirr, line 280: Please document this in the `Config Doc`

tamtamirr (commented):
Good work!
One thing I'm concerned about is that many parts of the code reference `any` type with no type assertion or type guards. I've commented what I've caught but please re-validate me as much code here works with `Values` 🙏

packages/salesforce-adapter/src/custom_references.ts
```diff
@@ -0,0 +1,237 @@
...
+const APEX_CLASS_SECTION = 'classAccesses'
+const FLOW_SECTION = 'flowAccesses'
+const LAYOUTS_SECTION = 'layoutAssignments'
+const OBJECT_SECTION = 'objectPermissions'
+const APEX_PAGE_SECTION = 'pageAccesses'
+const RECORD_TYPE_SECTION = 'recordTypeVisibilities'
+
+const FIELD_NO_ACCESS = 'NoAccess'
+
+const fullNameToElemIdName = (fullName: string): string => (
```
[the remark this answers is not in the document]
tomsellek: Ack

packages/salesforce-adapter/src/custom_references.ts
```diff
@@ -0,0 +1,237 @@
...
+  Types.getElemId(
+    fullName.replace(API_NAME_SEPARATOR, '_'),
+    true,
+  ).name
+)
+
+type ReferenceInSection = {
+  sourceField?: string
```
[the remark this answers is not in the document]
tomsellek: What would be a better name? The way this code works is we look at each entry in each profile section and use a callback to extract all the refs from it. The `sourceField` is the path inside the section where the ref came from.

e.g. for FLS, we have:
```
  fieldPermissions = {
    Seller = {
      ActiveFromDate = "ReadWrite"
      ActiveToDate = "ReadWrite"
      SalesAmount = "ReadWrite"
      SellerTier = "ReadWrite"
      SellerType = "ReadWrite"
    }
```

So `fieldPermissions` is the section, `Seller` is the entry in the section, and e.g. `ActiveFromDate` is the `sourceField` that is returned from the callback.

packages/salesforce-adapter/src/custom_references.ts
```diff
@@ -0,0 +1,237 @@
...
+}
+
+type RefTargetsGetter = (sectionEntry: Values, sectionEntryKey: string) => ReferenceInSection[]
+
+const referencesFromSection = (
+  profile: InstanceElement,
+  sectionName: string,
+  filter: (sectionEntry: Values) => boolean,
+  targetsGetter: RefTargetsGetter,
+): ReferenceInfo[] => {
```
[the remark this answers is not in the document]
tomsellek: Ack

packages/salesforce-adapter/src/custom_references.ts
```diff
@@ -0,0 +1,237 @@
...
+  if (!profile.value[sectionName]) {
+    return []
+  }
+  return Object.entries(profile.value[sectionName] as Values)
```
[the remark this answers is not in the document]
tomsellek: Sadly `_.isPlainObject` is not a type assertion, but I implemented the rest of the change. I also added a log in case it's not an object, because that would mean something is *very* wrong.

packages/salesforce-adapter/src/custom_references.ts
```diff
@@ -0,0 +1,237 @@
...
+    .filter(([, sectionEntry]) => filter(sectionEntry))
+    .flatMap(([sectionEntryKey, sectionEntry]): ReferenceInfo[] => {
+      const refTargets = targetsGetter(sectionEntry, sectionEntryKey)
+      return refTargets.map(({ target, sourceField }) => ({
+        source: profile.elemID.createNestedID(sectionName, sectionEntryKey, ...makeArray(sourceField)),
```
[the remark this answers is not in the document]
tomsellek: The section entry

packages/salesforce-adapter/src/custom_references.ts
```diff
@@ -0,0 +1,237 @@
...
+        target,
+        type: 'weak',
+      }))
+    })
+}
+
+const isEnabled = (sectionEntry: Values): boolean => (
+  sectionEntry.enabled
+)
```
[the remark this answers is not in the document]
tomsellek: Ack

packages/salesforce-adapter/src/custom_references.ts
```diff
@@ -0,0 +1,237 @@
...
+
+const isAnyAccessEnabledForObject = (objectAccessSectionEntry: Values): boolean => (
+  objectAccessSectionEntry.allowCreate
+  || objectAccessSectionEntry.allowDelete
+  || objectAccessSectionEntry.allowEdit
+  || objectAccessSectionEntry.allowRead
+  || objectAccessSectionEntry.modifyAllRecords
+  || objectAccessSectionEntry.viewAllRecords
```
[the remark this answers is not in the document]
tomsellek: Ack. Made a change that is sort of a middle ground.

packages/salesforce-adapter/src/custom_references.ts
```diff
@@ -0,0 +1,237 @@
...
+}
+
+const layoutReferences: RefTargetsGetter = sectionEntry => {
+  const layoutElemIdName = fullNameToElemIdName(sectionEntry[0].layout)
+  const layoutRef = {
+    target: new ElemID(SALESFORCE, LAYOUT_TYPE_ID_METADATA_TYPE, 'instance', layoutElemIdName),
+  }
+
+  const recordTypeRefs = sectionEntry
+    .filter((layoutAssignment: Values) => layoutAssignment.recordType !== undefined)
```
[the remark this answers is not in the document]
tomsellek: Ack

packages/salesforce-adapter/src/custom_references.ts
tomsellek: Ack

packages/salesforce-adapter/src/custom_references.ts
tomsellek: Ack

packages/salesforce-adapter/src/custom_references.ts
tomsellek: Ack

packages/salesforce-adapter/src/custom_references.ts
tomsellek: I dunno man, this would result in a log line for every profile we deal with. Seems like a lot?

packages/salesforce-adapter/src/types.ts
tomsellek: So, this raises an interesting question - do we want users to use this config or is this just to be used as a feature flag? Because I don't think users know (or should care about) what custom references are.
@amitgoldarb ?

amitgoldarb: I tend to agree with @tomsellek .I'm not sure that we want to document this, I'm also not sure how we explain custom references to our users.
We should probably open a jira ticket to discuss this, we can add documentaiton later if we need it.

tamtamirr (approved):
LGTM

PR #5048, opened 2023-11-05
SALTO-4981: Pass resolvedTypesElementsSource only for Salesforce accounts

tamtamirr:
Passing resolvedTypesElementsSource only for Salesforce accounts from core.

---

We should open it to other adapters gradually, after they've been fully tested with the new elements source, as we saw memory issues.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/core/src/core/adapters/adapters.ts
+++ b/packages/core/src/core/adapters/adapters.ts
@@ -200,6 +200,7 @@ export const createResolvedTypesElementsSource = (
   const resolvedTypes: Map<string, TypeElement> = new Map()
   let typesWereResolved = false
   const resolveTypes = async (): Promise<void> => {
+    log.debug('Resolving types within createResolvedTypesElementsSource')
     typesWereResolved = true
     const typeElements = await awu(await elementsSource.getAll()).filter(isType).toArray()
     // Resolve all the types together for better performance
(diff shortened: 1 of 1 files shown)


packages/core/src/core/adapters/adapters.ts
```diff
@@ -274,14 +280,17 @@ export const getAdaptersCreatorConfigs = async (
...
+      ), account, accountToServiceName[account])
       return [
         account,
         {
           credentials: credentials[account],
           config: await getConfig(account, defaultConfig),
-          elementsSource: createResolvedTypesElementsSource(createElemIDReplacedElementsSource(filterElementsSource(
-            elementsSource, account
-          ), account, accountToServiceName[account])),
+          elementsSource: RESOLVED_TYPES_ELEMENTS_SOURCE_ACCOUNTS.includes(account)
```
ori-moisis, line 291: assuming we need to use the resolved types elements source when the adapter is salesforce, this needs to use the `accountToServiceName` mapping here

tamtamirr: Got confused here. Service === Adapter ?

tamtamirr:
Closing as this fix shouldn't solve the OOM issue.

PR #5051, opened 2023-11-05
SALTO-4721: Removed the valueSetFilter in Salesforce

tamtamirr:
Removed the **onDeploy** valueSetFilter as it caused different **applied changes** than the actual received changes. This is a bad behavior which caused weird flows.

---

- removed the valueSetFilter.
- Moved exported utils from the removed filter to the `utils.ts` module. No extra UTs added since this utils are quite deprecated and already covered by the corresponding UTs on the **Change Validators** that use them.

---
_Release Notes_: 
Salesforce Adapter:
- The **valueSetFilter** which used to set deleted values of **Picklist field value sets** as "inactive" upon successful deployment was removed.

---
_User Notifications_: 
_None_


dorsev (approved):
LGTM

PR #5052, opened 2023-11-05
SALTO-4986: Support getting workspaceId and fetching assests schema

idoamit8:
In this PR I support fetching assets schema using the infrastructure. 

---

In order to fetch assets schema I need to have `workspaceId`. 
Only users who purchased Assets will have workspaceId. 
 
---
_Release Notes_: 
_JIra Adapter:_
Now support fetching of assets schemas. 

---
_User Notifications_: 
_JIra Adapter:_
Now support fetching of assets schemas.

### DIFF
--- a/packages/jira-adapter/src/adapter.ts
+++ b/packages/jira-adapter/src/adapter.ts
@@ -512,7 +512,8 @@ export default class JiraAdapter implements AdapterOperations {
     // jsmApiDefinitions is currently undefined for DC
     if (this.client === undefined
       || jsmApiDefinitions === undefined
-      || !this.userConfig.fetch.enableJSM) {
+      || !this.userConfig.fetch.enableJSM
+      || !this.userConfig.fetch.enableJsmExperimental) {
       return { elements: [] }
     }
 
--- a/packages/jira-adapter/src/adapter.ts
+++ b/packages/jira-adapter/src/adapter.ts
@@ -160,6 +160,8 @@ import portalGroupsFilter from './filters/portal_groups'
 import ScriptRunnerClient from './client/script_runner_client'
 import { weakReferenceHandlers } from './weak_references'
 import { jiraJSMEntriesFunc } from './jsm_utils'
+import { getWorkspaceId } from './workspace_id'
+import { JSM_ASSESTS_DUCKTYPE_SUPPORTED_TYPES } from './config/api_config'
 
 const { getAllElements } = elementUtils.ducktype
 const { findDataField, computeGetArgs } = elementUtils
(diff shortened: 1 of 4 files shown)


packages/jira-adapter/src/adapter.ts
```diff
@@ -503,6 +505,38 @@ export default class JiraAdapter implements AdapterOperations {
...
+    // jsmApiDefinitions is currently undefined for DC
+    if (this.client === undefined
+      || jsmApiDefinitions === undefined
+      || !this.userConfig.fetch.enableJSM
+      || !this.userConfig.fetch.enableJsmExperimental) {
+      return { elements: [] }
+    }
+
+    const workspaceId = await getWorkspaceId(this.client)
+    if (workspaceId === undefined) {
```
shir-reifenberg, line 521: maybe add a log (even .trace) here ?

packages/jira-adapter/src/adapter.ts
```diff
@@ -503,6 +505,38 @@ export default class JiraAdapter implements AdapterOperations {
     })
   }
 
+  @logDuration('generating JSM assets instances and types from service')
+  private async getJSMAssetsElements():
+  Promise<elementUtils.FetchElements<Element[]>> {
+    const { jsmApiDefinitions } = this.userConfig
+    // jsmApiDefinitions is currently undefined for DC
+    if (this.client === undefined
```
shir-reifenberg, line 513: This logic repeats in `getJSMElements`, so I think it's best to combine those.

packages/jira-adapter/src/adapter.ts
```diff
@@ -503,6 +505,38 @@ export default class JiraAdapter implements AdapterOperations {
...
+      shouldAddRemainingTypes: false,
+      supportedTypes: JSM_ASSETS_DUCKTYPE_SUPPORTED_TYPES,
+      fetchQuery: this.fetchQuery,
+      paginator: this.paginator,
+      nestedFieldFinder: findDataField,
+      computeGetArgs,
+      typeDefaults: jsmApiDefinitions.typeDefaults,
+      additionalRequestContext: workspaceContext,
+      getElemIdFunc: this.getElemIdFunc,
+    })
```
shir-reifenberg, line 537: We might have talked about it so sorry if we did 🥺  - do we need another call to`getAllElements` ? can't you get `workspaceId` in advance, and add it to `serviceDeskProjRecord` and then combine the 2 api configs ?

shir-reifenberg (commented):
Nice! Added some questions inside :)

packages/adapter-components/src/elements/ducktype/standalone_field_extractor.ts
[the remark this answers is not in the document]
idoamit8: of course :)

packages/jira-adapter/src/adapter.ts
```diff
@@ -513,13 +547,14 @@ export default class JiraAdapter implements AdapterOperations {
...
     }
+
     const serviceDeskProjects = swaggerResponseElements
       .filter(project => project.elemID.typeName === PROJECT_TYPE)
       .filter(isInstanceElement)
       .filter(project => project.value.projectTypeKey === SERVICE_DESK)
 
     const fetchResultWithDuplicateTypes = await Promise.all(serviceDeskProjects.map(async projectInstance => {
-      const serviceDeskProjRecord: Record<string, string> = {
+      const serviceDeskProjRecord: Record<string, unknown> = {
```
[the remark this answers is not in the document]
idoamit8: I changed it back to string. In the begining of this PR I put an object as value but than I decided to put the string instead and I forgot to change it back.

packages/jira-adapter/src/adapter.ts
idoamit8: Ok. It will happen a lot for users that have JSM but don't have assets.

packages/jira-adapter/src/adapter.ts
idoamit8: its not exactly the same.

packages/jira-adapter/src/adapter.ts
idoamit8: The problem is that if workspaceId is undefined I don't want to call all the assets types

packages/jira-adapter/src/adapter.ts
shir-reifenberg: got it, so whatever you think :)

packages/jira-adapter/src/adapter.ts
```diff
@@ -584,17 +619,19 @@ export default class JiraAdapter implements AdapterOperations {
     ])
 
     const jsmElements = await this.getJSMElements(swaggerResponse.elements)
-
+    const jsmAssetsElements = await this.getJSMAssetsElements()
```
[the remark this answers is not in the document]
shir-reifenberg: another suggestion, if JSM assets doesn't rely on `swaggerResponse` / `scriptRunnerElements` we can add it to the `promise.all`

shir-reifenberg (approved):
LGTM 😎

PR #5056, opened 2023-11-06
SALTO-4991: Fix account/adapter name replace for container types

ori-moisis:
The code did not handle the case where there was an unresolved type referece to a container type, it only worked for resolved type references.

---

This can cause an infinite loop when trying to resolve an element because when resolving types we would try to get the element using its original name and the element source would replace it, so we'd never get the type we need in order to resolve (see INCIDENT-8902 for more details)

---
_Release Notes_: 
Core:
- Fixed issue where fetching with a non-default account name could get stuck

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/workspace/src/workspace/elements_source.ts
+++ b/packages/workspace/src/workspace/elements_source.ts
@@ -13,7 +13,7 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
-import { Element, ElemID, Value, BuiltinTypesByFullName, ListType, MapType, isContainerType, ReadOnlyElementsSource, ContainerTypeName, TypeElement } from '@salto-io/adapter-api'
+import { Element, ElemID, Value, BuiltinTypesByFullName, ListType, MapType, isContainerType, ReadOnlyElementsSource, ContainerTypeName, TypeElement, TypeReference } from '@salto-io/adapter-api'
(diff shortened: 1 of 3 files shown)


packages/workspace/src/workspace/elements_source.ts
```diff
@@ -60,6 +60,17 @@ export function buildContainerType(
   return new typeCtors[prefix](innerType)
 }
 
+export const buildContainerTypeId = (
```
adamro, line 63: nit: Wouldn't this make more sense in adapter_api's `element_id.ts`? 🤔

adamro, line 63: Should we add a unit test to this?

adamro (approved):

packages/workspace/src/workspace/elements_source.ts
ori-moisis: oops
good catch

ori-moisis: I think I can't put it in `element_id.ts` while using `ListType.createElemID` (cyclic dependency)
this is indeed an area of sub-optimal code though as ElemID already has some kind that is tightly coupled with List/Map types.
might be good to tackle this, if I have the time I'll try to do something in a follow up PR

PR #5059, opened 2023-11-06
Bump to version 0.3.48

edenhassid:
# Salto Version v0.3.48
	
## New features and bug fixes
_Core_
- Fixed issue where fetching with a non-default account name could get stuck
- Omit conflicts on no-conflict core annotations.
- Ignore hidden values on calculatePatch
- Add calculatePatchFromChanges API
- Fix bug where if file `x` was removed and a new file was created under a (new) folder called `x`, we would fail on `ENOTDIR` (since the original file `x` was still there)
- Auto-merge conflicts in static-files / multi-line strings
- Add highlighted field in annotation of important values
- Fixed bug where restore/fetchFrom on element that has multiple fragments in the same file would create merge errors in nacl
- Trace log of instance deployment have bigger depth for the data
- Fixed issue where multi line quotes inside a multi line value were not properly escaped in NaCl

_Workspace_
- create path index for element API
- fix issue where unresolved references inside template expressions did not show as workspace warnings

_Netsuite adapter_:
- Fix SDF deploy dependencies when referencing an added inner object/field
- Fix CV to find invalid references in parent of field changes
- Better message in NS SOAP search error
- Support fetching `bin` instances. To start fetching `bin` instances in your environment, go to your environment configuration, remove `{ name = "bin" }` from the `fetch.exclude.types` section, then add it to the `fetch.include.types` section.

_Salesforce adapter_:
- Warn when fetching data records with non-queryable fields.
- Hierarchy fields in data records are now treated as references
- The account ID of Salesforce Sandboxes will be the org URL, which is unique.
- The adapter will now use API version 59.0.
- validateCredentials on Sandboxes will now return the organization url as a stable accountId and will include extraProperties with the Organization Id
- More accurate Author Information on **Custom and Standard Object**.
- Add Support in deactivating Flows.
- Fixed a crash if fields with certain names were created in instances whose types did not include said fields.
- Added the 'omittedFields' data configuration.
- will now warn when deploying instances of types for which the fetching user did not have permission to read certain fields
- Bugfix: Retrieve Workflow instances on **Custom Objects**.
- Initial support in **Fetch With Changes Detection** mode, which allows fetching only instances that were updated since the last fetch and decrease the fetch time significantly.
- Introduced the 'brokenOutgoingReferencesSettings' data configuration section. Use it to control what happens when Salto encounters an invalid reference in a data record.
- Add EclairGeoData, OmniUiCard, OmniDataTransform, OmniIntegrationProcedure, OmniInteractionAccessConfig, OmniInteractionConfig, and OmniScript to default exclude list
- Support in fetching Workflow instances of **Custom Objects**

_jira adapter_:
- fix a bug in deploying global Automation using Jira DC
- Now support Fetch of issueView and RequestForm as part of Request Type instances.
- Now support deploy of portal settings.
- Now supporting deploy of SLA.
- For those who enable JSM, SLA nacls will be visible upon your next fetch
- Now support fetch of portalSettings fetch.
- Now supports deployment of portal groups.
- Now supporting fetch and deploy operations for queue.
- Now supports Workflow statuses in JSM.
- Added service Url to issue layouts.
- Now we support deploying of JSM calendars.
- Fixing bug in issue layouts that causes to duplicate elemIDs.
- Added a workaround for a Jira bug that when creating a workflow we somtimes receive an error that the workflow name is already used although it wasn't
- Fixed a bug where workflow transition configuration would not be deployed
- Add important values annotation to types
- Jsm types will be in subfolder under their related project.
- Fixed a bug that caused deployment of Issue Type Schemes to fail in certain conditions
- Now supporting fetch and deploy operations for customerPermissions.
- Automations with several projects will now have a consistent elemID (projects names in the ID are now sorted)
- Supported JSM types will contain reference mapping.
- supporting fetch of RequestType in jira's JSM.
- Fixed a bug that caused project category not to be translated into a reference after project addition
- Fixed a bug that prevented the deployment of some workflow validators
- Changed automation structure to not have projects in case it is a global automation
- Timeout for workflow scheme publish was doubled and added to the configuration
- We now able to fetch issue layouts in jira.
- `salto element fix` will now clean invalid projects from automation

 
_Zendesk adapter_:
- Fix potential crash on large deployments due to memory consumption
- Add the start year of an holiday to the elemID
- Deployments will no longer fail on simultaneously addition and removal of elements with equal idFields values
- Added missing reference to missing organizations in trigger, automation, sla_policy and ticket_field
- Prevent deployment of actions with custom_status_id if the feature is not enabled
- Added gaps to elements positions of triggers, automations and views
- No longer prevents deployment of  a new default ticket form if the current default is not changing
- Added `convertJsonIdsToReferences` config, that allows convert of JSON pairs of 'id' to reference expression
- Fix crash on deployment of a guide instance without a brand
- In elements, in string fields, it is now possible to change ids in links with reference expressions
- Fixed a race condition that caused us to query the users multiple times on deploy
- Added support to fetch custom_object instances
- When using `&hash` in the `article_attachment` idFields, if the hash field does not exist it will add to the `elemId` the string `INVALID_HASH`
- Fix bug when using the OAuth credentials flow that required providing a domain argument
- Change the path of article attachment static file, the new path will end with `/article_attachment/fileName/<hash>_fileName.png` instead of `/article_attachment/fileName.png`
- Add the possibility to use `&hash` in `idFields` of `article_attachment`
- On deployment of custom_field_options - put the value of 'raw_name' inside 'name', because 'name' is actually the parsed value of 'raw_name'
- Hide 'name' fields as they are no longer relevant
- Fix an issue with macro attachment and organization custom_field_options, where preDeploy changes to these types were ignored
- Prevent removal of dynamic_content that is being used by types that blocks its deletion
- Add `handleIdenticalAttachmentConflicts` flag. when it is true it will look at all attachments with the same elemId and hash and will keep only one.
- Prevent adding a conditional ticket field to a form if the ticket field doesn't exists in the ticket fields list
- Mismatches between environments of the elemIds of groups will be prevented

_Okta adapter_:
- Fix issue with fetch when environment contains inactive apps with Group Push enabled
- Support fetch and deploy of Group Push 
- Improvements to ProfileMapping instances
- Fetch properties push status in ProfileMapping instances
- Support deployments of ProfileMapping
- add alias to Okta elements
- GET calls will not pass the rate limit and will wait for the rate limit to reset before continuing


adamro (approved):
Thanks 🔥

PR #5060, opened 2023-11-06
SALTO-4991: Avoid infinite loop in resolve elem source has a bug

ori-moisis:
In the previous fix of this issue we hopefully took care of the bug in the element source. But even so, and just in case there is another such issue somewhere else or in the future adding another layer of protection against cases that should never happen

---

no release notes here since this should not solve any issue, the issue was supposed to be fixed by the previous commit

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/workspace/src/expressions.ts
+++ b/packages/workspace/src/expressions.ts
@@ -101,6 +101,15 @@ const getElementCloneFromSource = async (
     }
     return undefined
   }
+  if (!elem.elemID.isEqual(id)) {
+    // This should really never happen, but if it does, we might find ourselves in an infinite loop
+    // so we try to protect against this case anyway
+    log.warn(
+      'resolve expected element with ID %s but found element with ID %s, returning undefine to avoid loop',
+      elem.elemID.getFullName(), id.getFullName(),
+    )
+    return undefined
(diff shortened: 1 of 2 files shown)


adamro (approved):

packages/workspace/src/expressions.ts
```diff
@@ -101,6 +101,15 @@ const getElementCloneFromSource = async (
     }
     return undefined
   }
+  if (!elem.elemID.isEqual(id)) {
```
adamro, line 104: You were a bit worried about performance in resolve. Is there a legit easy way to check the effect of this? 🤔

ori-moisis: yes, thanks for reminding me
I am going to go with "this is fine" because:
1. `ElemID.isEqual` is just running `===` between two strings that are already in memory, so it should be fairly fast
2. the next line after this one runs `elem.clone()` which should be many orders of magnitude slower, and also, it means the number of times we reach this code should be limited by the number of elements in the environment so this should be a fairly small number compared to the higher scale parts of this process (e.g - number of references between elements)

PR #5061, opened 2023-11-06
SALTO-4986: Add assets statuses

idoamit8:
In this PR I added support for fetching assets statuses. 

---
Please review only the last commit "assets statuses". Only added a type in api_config.ts

---
_Release Notes_: 
_Jira Adapter:_
Now supporting fetch of assets statuses. 

---
_User Notifications_: 
_Jira Adapter:_
Now supporting fetch of assets statuses.


shir-reifenberg (approved):
🚀

PR #5063, opened 2023-11-06
SALTO-4980:  remove child elements if their parent is omitted due to id collision

edenhassid:
remove child elements if their parent is omitted due to id collision

---

---
_Release Notes_: 
zendesk:
- child elements will be omitted when their parent is omitted due to elemId collision

---
_User Notifications_: 

### DIFF
--- a/packages/zendesk-adapter/src/filters/collision_errors.ts
+++ b/packages/zendesk-adapter/src/filters/collision_errors.ts
@@ -13,12 +13,17 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
-import { isInstanceElement, Element } from '@salto-io/adapter-api'
-import { config as configUtils } from '@salto-io/adapter-components'
-import { getAndLogCollisionWarnings, getInstancesWithCollidingElemID } from '@salto-io/adapter-utils'
+import { isInstanceElement, Element, isReferenceExpression } from '@salto-io/adapter-api'
+import { logger } from '@salto-io/logging'
+import { config as configUtils, filters } from '@salto-io/adapter-components'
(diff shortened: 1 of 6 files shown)


packages/zendesk-adapter/src/filters/collision_errors.ts
```diff
@@ -13,24 +13,55 @@
...
+          log.error(`${removedInst.elemID.getFullName()} does not have a parent`)
+          return false
+        }
+        return true
+      }),
+      removedInst => getParents(removedInst)[0].elemID.getFullName()
+    )
+    const childWarning = Object.keys(removedChildsByParent)
+      .map(parent =>
+        `removed ${removedChildsByParent[parent].map(element => element.elemID.getFullName()).join(' , ')} as its parent ${parent} was removed due to collision of elemId`)
```
netama, line 59: not sure this part is worth adding if it's just for logging, I think we can just say they were omitted due to their parent being omitted

edenhassid: do you think we should add it as a fetch warning to the user?

netama: hmm I guess usually they will be more interested in the "parent" instances - maybe we can change the message a bit to clarify that there are the elements that were omitted "directly", and then their children were also omitted. maybe something like
1. "Omitted ${numInstances} instances of ${type} and <> child instances due to Salto ID collisions"
2. "Breakdown per colliding Salto ID (note that additionally there may have been child instances that were also omitted):" (but might be worth phrasing with @tomermevorach ). I think it can be good if the counts also show 

wdyt?

PR #5066, opened 2023-11-06
Rephrase data elements ACCOUNT_SPECIFIC_VALUE CV

dantal4:
To be more accurate.

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/netsuite-adapter/src/change_validators/data_account_specific_values.ts
+++ b/packages/netsuite-adapter/src/change_validators/data_account_specific_values.ts
@@ -16,6 +16,7 @@
 import { collections } from '@salto-io/lowerdash'
 import {
   ChangeError,
+  ElemID,
   getChangeData,
   InstanceElement,
   isAdditionChange,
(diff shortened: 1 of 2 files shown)


packages/netsuite-adapter/src/change_validators/data_account_specific_values.ts
```diff
@@ -64,14 +64,15 @@ const changeValidator: NetsuiteChangeValidator = async changes => (
       return modificationChange
     })
     .map(getChangeData)
-    .filter(hasUnresolvedAccountSpecificValue)
-    .map((instance): ChangeError => ({
-      elemID: instance.elemID,
+    .flatMap(getPathsWithUnresolvedAccountSpecificValue)
+    .map((elemID): ChangeError => ({
+      elemID,
```
omrilit, line 69: You modified the CV to point on the inner elemID as the one with the ChangeError instead of doing it on the topLevel instance's elemID.
Is it intentional? What is the reasoning for of doing it?

omrilit (approved):
Looks great, thanks!
Approving to save cycle

packages/netsuite-adapter/src/change_validators/data_account_specific_values.ts
dantal4: It was intentional, because I thought that there can be several errors in the same element, and it would be better to see the errors near the field itself.
Now I see that the UI not always show the errors well when they aren't on the base level, so I'll change it.

PR #5067, opened 2023-11-07
SALTO-4894 Weak reference handlers rename

kami-delaroz:
rename weak references to a more generic name, as this is a general fix element type

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/adapter-components/src/references/element_fixer_handler.ts
```diff
@@ -16,7 +16,7 @@
 import { FixElementsFunc, GetCustomReferencesFunc } from '@salto-io/adapter-api'
 
 
-export type WeakReferencesHandler<T> = {
+export type ElementFixersHandler<T> = {
   findWeakReferences: GetCustomReferencesFunc
```
kami-delaroz, line 20: @netama looks like this one indeed deals with references only. not sure whether we want to take it out or not, leaving it like this for now, will see once adding the new elementFix what's best

PR #5073, opened 2023-11-12
SALTO-4984 Remove field restrictions

adfineberg:
Remove field restrictions for user and organization fields.

---

---
_Release Notes_: 
__Zendesk Adapter__
* Remove validation for user and organization fields.

---
_User Notifications_: 


edenhassid (approved):
🔥 🚀

PR #5074, opened 2023-11-12
SALTO-5019: Support nestedPath in ducktypes

idoamit8:
Added support for nestedPath in duckType when we recurse tInto.

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/elements/ducktype/standalone_field_extractor.ts
+++ b/packages/adapter-components/src/elements/ducktype/standalone_field_extractor.ts
@@ -68,7 +68,6 @@ const addFieldTypeAndInstances = async ({
   transformationConfigByType,
   transformationDefaultConfig,
   getElemIdFunc,
-  nestedPath,
 }: {
   adapterName: string
   typeName: string
@@ -78,7 +77,6 @@ const addFieldTypeAndInstances = async ({
   transformationConfigByType: Record<string, TransformationConfig>
   transformationDefaultConfig: TransformationDefaultConfig
   getElemIdFunc?: ElemIdGetter
-  nestedPath?: string[]
 }): Promise<Element[]> => {
   if (type.fields[fieldName] === undefined) {
     log.info('type %s field %s does not exist (maybe it is not populated by any of the instances), not extracting field', type.elemID.name, fieldName)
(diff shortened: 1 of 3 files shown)


packages/adapter-components/src/elements/ducktype/transformer.ts
```diff
@@ -287,6 +287,7 @@ export const getTypeAndInstances = async ({
   reversedSupportedTypes,
   customInstanceFilter,
   additionalRequestContext,
+  nestedPath,
```
shir-reifenberg, line 290: Not sure why we need this here

packages/adapter-components/src/elements/ducktype/standalone_field_extractor.ts
```diff
@@ -131,6 +133,7 @@ const addFieldTypeAndInstances = async ({
         transformationConfigByType,
         transformationDefaultConfig,
         getElemIdFunc,
+        nestedPath: [...(nestedPath ?? [type.elemID.name]), inst.elemID.name, fieldName],
```
shir-reifenberg, line 136: I might be missing something, but we're passing `nestedPath` all the way and I'm not sure in what cases it will be defined?

shir-reifenberg (commented):
Looks good overall 🚀 , added one question cause I think we might be able to simplify this

shir-reifenberg (approved):
🚀

packages/adapter-components/src/elements/ducktype/standalone_field_extractor.ts
```diff
@@ -131,6 +131,7 @@ const addFieldTypeAndInstances = async ({
         transformationConfigByType,
         transformationDefaultConfig,
         getElemIdFunc,
+        nestedPath: [type.elemID.name, inst.elemID.name, fieldName],
```
netama, line 134: this looks good :) but does this work recursively when there's a nested resource under a nested resource? I think we might need to base it more closely on the instance's path and not the type path? it'd be a little ugly though, something like
```suggestion
nestedPath: [
          ...inst.path?.slice(2, inst.path?.length - 1) ?? [],
          fieldName,
        ],
```
which is not ideal.
(this is still correct for the "basic" most common case, so can also merge and figure it out later)

idoamit8: Great :)

PR #5078, opened 2023-11-13
SALTO-5025: Improve Retrieve Polling time out retry mechanism in Salesforce

tamtamirr:
This PR improves the retry mechanism of `Polling time out` upon retrieve of instances.

---

As more and more types have moved to retrieve, we may face issues with the times retrieve requests can take.
We see increasing amount of fetches that fail on **Polling time out** upon retrieve. I've used the retry mechanism of sending smaller chunks that was previously implemented by @yuribro for this scenario as-well.

As for the tests:
You can see what the **it.each** I've added translates to: 
Screen Shot 2023-11-13 at 17 35 02

---
_Release Notes_: 
Salesforce Adapter:
- Improved the retry mechanism for Fetches that encounter Polling time out.

---
_User Notifications_: 
_None_


tamtamirr:
Closing until this comes more frequent.

PR #5085, opened 2023-11-15
SALTO-5030: fix template expressions in zendesk

edenhassid:
fix side conversation group template expression

---

---
_Release Notes_: 
zendesk:
* remove empty strings from side conversation group template expressions

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/filters/side_conversation.ts
+++ b/packages/zendesk-adapter/src/filters/side_conversation.ts
@@ -18,13 +18,13 @@ import {
   InstanceElement,
   isInstanceElement,
   ReferenceExpression,
-  TemplateExpression,
   Value,
 } from '@salto-io/adapter-api'
 import { references as referencesUtils } from '@salto-io/adapter-components'
 import _ from 'lodash'
 import joi from 'joi'
 import { logger } from '@salto-io/logging'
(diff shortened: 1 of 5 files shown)


packages/zendesk-adapter/src/filters/side_conversation.ts
```diff
@@ -76,12 +76,12 @@ export const sideConversationsOnFetch = (elements: Element[], config: ZendeskCon
...
           missingInstance.value.id = groupId
           // Replace the group part with a missing reference
-          action.value[2] = new TemplateExpression({
+          action.value[2] = createTemplateExpression({
             parts: [
               prefix,
               new ReferenceExpression(missingInstance.elemID, missingInstance),
               suffix,
-            ],
+            ].filter(part => !_.isEmpty(part)),
```
ori-moisis, line 84: filtering out empty parts happens in `createTemplateExpression` no?
do we need to do it in both places?

edenhassid: no, I was supposed to remove it, will do it now

ori-moisis (approved):
going over other places that create template expressions directly, should we also look at the code that creates templates in `article_body.ts` which might have the same issue in a different scenario?

PR #5088, opened 2023-11-15
SALTO-4517: only resolve types in fetch, not in deploy

ori-moisis:
The original implementation was meant to only [affect fetch operations](https://salto-io.atlassian.net/browse/SALTO-4517?focusedCommentId=93101) Since it uses a cache, it must not be used in deploy where the underlying element source changes

---

We haven't seen a bug directly from the original reason why this shouldn't be used in deploy, but we did see a possibly related bug in zendesk deployments.
we should still get to the bottom of the the possibly related bug probably, but changing the implementation here because regardless of the other bug, caching elements in deploy could lead to incorrect behavior when deploying ObjectType elements

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


adamro (approved):
Do we have any way to test this properly without a too big of an effort? 🤔

ori-moisis:
I couldn't think of something that would be simple enough to test

PR #5089, opened 2023-11-15
SALTO-4933 more logs

netama:
Add logs, remove uneeded uniq

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/adapter.ts
+++ b/packages/zendesk-adapter/src/adapter.ts
@@ -334,13 +334,13 @@ const zendeskGuideEntriesFunc = (
 
 const getBrandsFromElementsSource = async (
   elementsSource: ReadOnlyElementsSource
-): Promise<InstanceElement[]> => (
+): Promise<InstanceElement[]> => log.time(async () => (
   awu(await elementsSource.list())
     .filter(id => id.typeName === BRAND_TYPE_NAME && id.idType === 'instance')
     .map(id => elementsSource.get(id))
     .filter(isInstanceElement)
     .toArray()
-)
+), 'get brands from element source')
 
 /**
  * Fetch Guide (help_center) elements for the given brands.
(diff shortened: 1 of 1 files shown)


packages/zendesk-adapter/src/adapter.ts
```diff
@@ -334,13 +334,13 @@ const zendeskGuideEntriesFunc = (
 
 const getBrandsFromElementsSource = async (
   elementsSource: ReadOnlyElementsSource
-): Promise<InstanceElement[]> => (
+): Promise<InstanceElement[]> => log.time(async () => (
```
ori-moisis, line 337: pretty sure this won't count the right thing since this returns a promise
it will count the time it takes to create that promise (which will be zero) but it won't count the time it takes to consume it (which is the interesting time)

I think it would be better to remove this log since it will be misleading

packages/zendesk-adapter/src/adapter.ts
```diff
@@ -795,8 +795,7 @@ export default class ZendeskAdapter implements AdapterOperations {
       saltoErrors.push(e)
     }
 
-
-    const brandsList = _.uniq(await getBrandsFromElementsSource(this.elementsSource))
+    const brandsList = await getBrandsFromElementsSource(this.elementsSource)
```
ori-moisis, line 798: I think we should add the log after this line

on a separate subject but same place, there is no reason to run this at all in most deployments since most deployments won't have Guide related changes

netama: I agree but it requires more changes that I didn't want to get to today, I thought the log was the important thing for now

packages/zendesk-adapter/src/adapter.ts
netama: it seems to work for me, what's the concern exactly?

packages/zendesk-adapter/src/adapter.ts
ori-moisis: the log isn't really important since the scenario where that log could have helped has been removed anyway
thinking about this again I think we don't need this log.
I do think we could use a different log though, I'll open a different PR for that

PR #3042, opened 2022-05-17
Salto 2294: sf polling time out

tamtamirr:
We encountered some users that received timeout during fetch. Timeout occurs after 90 minutes by default.

---

Decided to go for a simple solution since there was already a retry flow on Client::retrieve on specific error messages. Added the polling timeout to the list.

---
_Release Notes_: 
Salesforce Adapter:
- Retry on fetch timeout.

---
_User Notifications_: 
- None (Bugfix)

PR #5090, opened 2023-11-15
SALTO-5013 - Salesforce: Report progress during deploy

tomsellek:
Allow adapters to report progress during deploy

- [ ] ~Review of the progress message from @tomermevorach and @pgonzaleznetwork~

---

1. Allow adapters to accept a `ProgressReporter` during deploy
2. In the Salesforce adapter, Instead of waiting for the deploy to complete, check its status once in a while and report the progress

Tests:
 - Deploy and check the logs to see the expected data is printed out
 - Change the `client.polling.interval` and `client.polling.deployTimeout` configs to 1 second and confirm the deploy times out with the correct error message.

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/adapter-api/src/adapter.ts
+++ b/packages/adapter-api/src/adapter.ts
@@ -75,6 +75,7 @@ export type FetchOptions = {
 }
 
 export type DeployOptions = {
+  progressReporter?: ProgressReporter
   changeGroup: ChangeGroup
 }
 
--- a/packages/adapter-api/src/adapter.ts
+++ b/packages/adapter-api/src/adapter.ts
@@ -75,7 +75,7 @@ export type FetchOptions = {
 }
 
 export type DeployOptions = {
-  progressReporter?: ProgressReporter
+  progressReporter: ProgressReporter
   changeGroup: ChangeGroup
 }
 
--- a/packages/core/src/core/deploy/deploy_actions.ts
+++ b/packages/core/src/core/deploy/deploy_actions.ts
@@ -18,7 +18,8 @@ import {
   AdapterOperations, getChangeData, Change,
   isAdditionOrModificationChange,
   DeployExtraProperties, DeployOptions, Group,
-  SaltoElementError, SaltoError, SeverityLevel, DeployResult, ChangeDataType, SaltoErrorType,
+  SaltoElementError, SaltoError, SeverityLevel, DeployResult, ChangeDataType, SaltoErrorType, ProgressReporter,
+  Progress,
 } from '@salto-io/adapter-api'
 import { detailedCompare, applyDetailedChanges } from '@salto-io/adapter-utils'
 import { WalkError, NodeSkippedError } from '@salto-io/dag'
@@ -66,7 +67,8 @@ const deployOrValidate = async (
 const deployAction = async (
   planItem: PlanItem,
   adapters: Record<string, AdapterOperations>,
-  checkOnly: boolean
+  checkOnly: boolean,
+  progressReporter: ProgressReporter,
 ): Promise<DeployResult> => {
   const changes = [...planItem.changes()]
   const adapterName = getChangeData(changes[0]).elemID.adapter
@@ -74,7 +76,7 @@ const deployAction = async (
   if (!adapter) {
     throw new Error(`Missing adapter for ${adapterName}`)
   }
-  const opts = { changeGroup: { groupID: planItem.groupKey, changes } }
+  const opts = { changeGroup: { groupID: planItem.groupKey, changes }, progressReporter }
   return deployOrValidate({ adapter, adapterName, opts, checkOnly })
 }
 
@@ -135,7 +137,10 @@ export const deployActions = async (
       })
       reportProgress(item, 'started')
       try {
-        const result = await deployAction(item, adapters, checkOnly)
+        const progressReporter = {
+          reportProgress: (progress: Progress) => reportProgress(item, 'started', progress.message),
+        }
+        const result = await deployAction(item, adapters, checkOnly, progressReporter)
         result.appliedChanges.forEach(appliedChange => appliedChanges.push(appliedChange))
         if (result.extraProperties?.groups !== undefined) {
           groups.push(...result.extraProperties.groups)
--- a/packages/salesforce-adapter/src/adapter.ts
+++ b/packages/salesforce-adapter/src/adapter.ts
@@ -16,7 +16,7 @@
 import {
   TypeElement, ObjectType, InstanceElement, isAdditionChange, getChangeData, Change,
   ElemIdGetter, FetchResult, AdapterOperations, DeployResult, FetchOptions, DeployOptions,
-  ReadOnlyElementsSource, ElemID, PartialFetchData, Element,
+  ReadOnlyElementsSource, ElemID, PartialFetchData, Element, ProgressReporter,
 } from '@salto-io/adapter-api'
 import {
   filter,
@@ -530,7 +530,7 @@ export default class SalesforceAdapter implements AdapterOperations {
   }
 
   private async deployOrValidate(
-    { changeGroup }: DeployOptions,
+    { changeGroup, progressReporter }: DeployOptions,
     checkOnly: boolean
   ): Promise<DeployResult> {
     const fetchParams = this.userConfig.fetch ?? {}
@@ -564,9 +564,14 @@ export default class SalesforceAdapter implements AdapterOperations {
         fetchProfile.dataManagement,
       )
     } else {
+      const nullProgressReporter: ProgressReporter = {
+        // eslint-disable-next-line @typescript-eslint/no-empty-function
+        reportProgress: () => {},
+      }
       deployResult = await deployMetadata(resolvedChanges, this.client, changeGroup.groupID,
-        this.nestedMetadataTypes, this.userConfig.client?.deploy?.deleteBeforeUpdate, checkOnly,
-          this.userConfig.client?.deploy?.quickDeployParams)
+        this.nestedMetadataTypes, progressReporter ?? nullProgressReporter,
+        this.userConfig.client?.deploy?.deleteBeforeUpdate, checkOnly,
+        this.userConfig.client?.deploy?.quickDeployParams)
     }
     log.debug(`received deployResult for group ${changeGroup.groupID}`)
     // onDeploy can change the change list in place, so we need to give it a list it can modify
--- a/packages/salesforce-adapter/src/client/client.ts
+++ b/packages/salesforce-adapter/src/client/client.ts
@@ -24,6 +24,7 @@ import {
   Connection as RealConnection,
   DeployOptions as JSForceDeployOptions,
   DeployResult,
+  DeployResultLocator,
   DescribeGlobalSObjectResult,
   DescribeSObjectResult,
   DescribeValueTypeResult,
@@ -799,6 +800,35 @@ export default class SalesforceClient {
     )
   }
 
+  private async reportDeployProgressUntilComplete(
+    deployStatus: DeployResultLocator<DeployResult>,
+    progressCallback: (inProgressResult: DeployResult) => void
+  ): Promise<DeployResult> {
+    const progressCallbackWrapper = async (): Promise<void> => {
+      const partialResult = await deployStatus.check()
+      try {
+        const detailedResult = await this.conn.metadata.checkDeployStatus(partialResult.id, true)
+        progressCallback(detailedResult)
+      } catch (e) {
+        log.warn('checkDeployStatus API call failed. Progress update will not take place. Error: %s', e.message)
+      }
+    }
+    const pollingInterval = setInterval(progressCallbackWrapper, this.conn.metadata.pollInterval)
+
+    const clearPollingInterval = (result: DeployResult): DeployResult => {
+      clearInterval(pollingInterval)
+      return result
+    }
+    const clearPollingIntervalOnError = (error: Error): DeployResult => {
+      clearInterval(pollingInterval)
+      throw error
+    }
+    // We can't use finally() because, despite what the type definition for jsforce says,
+    // DeployResultLocator.complete() actually returns an AsyncResultLocator<T> and not a Promise<T>.
+    // ref. https://github.com/jsforce/jsforce/blob/c04515846e91f84affa4eb87a7b2adb1f58bf04d/lib/api/metadata.js#L830
+    return deployStatus.complete(true).then(clearPollingInterval, clearPollingIntervalOnError)
+  }
+
   /**
    * Updates salesforce metadata with the Deploy API
    * @param zip The package zip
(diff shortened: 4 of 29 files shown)


packages/salesforce-adapter/src/client/client.ts
```diff
@@ -520,6 +522,47 @@ export const validateCredentials = async (
...
+    const status = await this.locator.check()
+    if (status.done) {
+      return status
+    }
+
+    const detailedStatus = await this.connection.metadata.checkDeployStatus(status.id, true)
+    await sleep(interval).then(() => callback(detailedStatus))
+
+    return detailedStatus
+  }
```
tomsellek, line 550: I am not crazy about this. It can probably be done better using something like `Promise.race` but I'm not sure how.

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -491,7 +491,7 @@ export default class SalesforceAdapter implements AdapterOperations {
   }
 
   private async deployOrValidate(
-    { changeGroup }: DeployOptions,
+    { changeGroup, progressReporter }: DeployOptions,
```
amitgoldarb, line 494: Why is this optional? Isn't it a must from the core level? Shouldn't it always be defined?

packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -54,6 +53,8 @@ const log = logger(module)
 // and is not meant to actually be deployed
 export const DEPLOY_WRAPPER_INSTANCE_MARKER = '_magic_constant_that_means_this_is_a_wrapper_instance'
 
+const SALESFORCE_PROGRESS_UPDATE_INTERVAL = 500
```
amitgoldarb, line 56: I think this should probably be a deploy parameter in the adapter.config
Nit: I also think the 500ms is too little, my intuition is somewhere along the lines of 5s. Is this new polling API throttled?

packages/salesforce-adapter/src/client/client.ts
```diff
@@ -772,29 +815,27 @@ export default class SalesforceClient {
...
     const optionsToSend = ['rollbackOnError', 'ignoreWarnings', 'purgeOnDelete', 'testLevel', 'runTests']
-    const deployResult = flatValues(
-      await this.conn.metadata.deploy(
-        zip,
-        {
-          ...defaultDeployOptions,
-          ..._.pick(this.config?.deploy, optionsToSend),
-          checkOnly,
-        },
-      ).complete(true)
```
amitgoldarb, line 794: Does this change in functionality somehow affect the deployTimeout? Now that we have moved to polling? Do we need to reimplement it?

packages/salesforce-adapter/src/client/client.ts
```diff
@@ -520,6 +522,47 @@ export const validateCredentials = async (
...
 }
+
+export class DeployStatus {
+  constructor(
+    private locator: DeployResultLocator<DeployResult>,
+    private connection: Connection,
+  ) {
+  }
+
+  public async complete(): Promise<DeployResult> {
```
amitgoldarb, line 533: Do we still need this API? Shouldn't this only have one public API being pollUntilComplete with a callback like you gave it?

packages/salesforce-adapter/src/client/client.ts
```diff
@@ -520,6 +522,47 @@ export const validateCredentials = async (
...
+  private async waitIfNotDone(
+    interval: number,
+    callback: (inProgressResult: DeployResult) => Promise<void>,
+  ): Promise<DeployResult> {
+    const status = await this.locator.check()
+    if (status.done) {
+      return status
+    }
+
+    const detailedStatus = await this.connection.metadata.checkDeployStatus(status.id, true)
```
amitgoldarb, line 546: What do we do when this API fails? We will fail rather agressively here I think. Should we add some sort of retry?

packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -383,9 +397,31 @@ export const deployMetadata = async (
...
+  const sfDeployStatus = await quickDeployOrDeploy(client, pkgData, checkOnly, quickDeployParams)
+
+  await sfDeployStatus.poll(SALESFORCE_PROGRESS_UPDATE_INTERVAL, 5000, async result => {
+    if (progressReporter) {
+      progressReporter.reportProgress({
+        message: await deployProgressMessage(client, result),
+      })
+    }
+    log.debug('in-progress deploy result: %s', safeJsonStringify({
+      ...result,
```
amitgoldarb, line 409: Nit: we should extract this if we use it twice

packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -383,9 +397,31 @@ export const deployMetadata = async (
     }
   }
 
-  const sfDeployRes = await quickDeployOrDeploy(client, pkgData, checkOnly, quickDeployParams)
+  const sfDeployStatus = await quickDeployOrDeploy(client, pkgData, checkOnly, quickDeployParams)
+
+  await sfDeployStatus.poll(SALESFORCE_PROGRESS_UPDATE_INTERVAL, 5000, async result => {
```
amitgoldarb, line 402: Nit: I think we should extract this function for readability

packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -339,11 +340,24 @@ const quickDeployOrDeploy = async (
 const isQuickDeployable = (deployRes: SFDeployResult): boolean =>
   deployRes.id !== undefined && deployRes.checkOnly && deployRes.success && deployRes.numberTestsCompleted >= 1
 
+const deployProgressMessage = async (
+  client: SalesforceClient,
+  deployResult: SFDeployResult,
+): Promise<string> => {
+  const url = await getDeployStatusUrl(deployResult, client)
+  const testStatus = `${deployResult.numberTestsCompleted}/${deployResult.numberComponentsTotal} (${deployResult.numberTestErrors} errors)`
```
amitgoldarb, line 348: @tomsellek I think it would be a bit easier to review this if we have a CLI output example, could you possibly add one?

amitgoldarb (commented):
Looks great! Added a few comments

packages/salesforce-adapter/src/client/client.ts
tamtamirr: consider creating a javascript interval and removing it when the promise is resolved:

```typescript
  const intervalId = setInterval(() => {
    progressReporter.reportProgress(...)
  }, 500)
  const deployResult = flatValues(
    await this.conn.metadata.deploy(
      zip,
      {
        ...defaultDeployOptions,
        ..._.pick(this.config?.deploy, optionsToSend),
        checkOnly,
      },
    ).complete(true)
      .finally(() => {
        clearInterval(intervalId)
      })
  )
```

packages/adapter-api/src/adapter.ts
```diff
@@ -75,6 +75,7 @@ export type FetchOptions = {
 }
 
 export type DeployOptions = {
+  progressReporter?: ProgressReporter
```
tamtamirr, line 78: Why is this optional?

tamtamirr (commented):
Good work! I think this PR can be simplified. See my suggestion on your comment 🙏

packages/adapter-api/src/adapter.ts
tomsellek: So I don't have to worry about reporting progress in other adapters

packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -135,7 +137,10 @@ export const deployActions = async (
       })
       reportProgress(item, 'started')
       try {
-        const result = await deployAction(item, adapters, checkOnly)
+        const progressReporter = {
```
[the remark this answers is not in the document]
tomsellek: The CLI just prints "Still changing" on every update and discards the actual message. See https://github.com/salto-io/salto/blob/36c7bb5b197f682bb6c066f292865e25592228c8/packages/cli/src/commands/deploy.ts#L147C13-L147C13 .

packages/salesforce-adapter/src/client/client.ts
```diff
@@ -772,29 +777,51 @@ export default class SalesforceClient {
...
+        const detailedResult = await this.conn.metadata.checkDeployStatus(partialResult.id, true)
+        progressCallback(detailedResult)
+      }
+      const pollingInterval = setInterval(progressCallbackWrapper, DEPLOY_STATUS_POLLING_INTERVAL_MS)
+
+      const clearPollingInterval = (result: DeployResult): DeployResult => {
+        clearInterval(pollingInterval)
+        return result
+      }
+      deployResult = await deployStatus.complete(true).then(clearPollingInterval, clearPollingInterval)
```
tomsellek, line 818: The promise that is returned from `this.conn.metadata.deploy()` doesn't have a `finally()` method for some reason?!

packages/salesforce-adapter/src/client/client.ts
tomsellek: Updated per @tamtamirr 's suggestion

packages/salesforce-adapter/src/client/client.ts
tamtamirr: Please note that conn.metadata.deploy returns a **DeployResultLocator**. The Promise is only created when you invoke **complete()** in L818. **complete()** returns Promise<DeployResult>

packages/salesforce-adapter/src/adapter.ts
tomsellek: https://github.com/salto-io/salto/pull/5090/files#r1399230447

packages/salesforce-adapter/src/client/client.ts
tomsellek: Given the new implementation, we will call the progress callback periodically until the deploy completes. I don't think this will have any effect on any timeouts.

packages/adapter-api/src/adapter.ts
amitgoldarb: I think this is the other way around no? Making it non-optional doesn't mean the adapter needs to implement something with it, right? It could easily ignore the input. The API now means we aren't sure if core passes to the adpater the progressReporter which isn't the intention. I tried this and the only thing that breaks are some e2e tests to the adapter (which makes sense), it's a bit cumbersome but I tend to think we should make this non-optional.

packages/core/src/core/deploy/deploy_actions.ts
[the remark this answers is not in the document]
amitgoldarb: Can we print a CLI output as an example? Just so that we see this "feels nice"?

packages/salesforce-adapter/src/adapter.ts
amitgoldarb: Commented there 🙏

packages/salesforce-adapter/src/client/client.ts
amitgoldarb: Did we test this? I think we can test this locally rather naively by lowering the timeout to something very low and seeing this still functions as expected.

packages/salesforce-adapter/src/client/client.ts
```diff
@@ -772,29 +777,54 @@ export default class SalesforceClient {
...
+        ...defaultDeployOptions,
+        ..._.pick(this.config?.deploy, optionsToSend),
+        checkOnly,
+      },
     )
+    let deployResult: DeployResult
+    if (progressCallback) {
+      const progressCallbackWrapper = async (): Promise<void> => {
+        const partialResult = await deployStatus.check()
+        const detailedResult = await this.conn.metadata.checkDeployStatus(partialResult.id, true)
```
amitgoldarb, line 809: What happens if this call fails? Should we add some retry logic to this.

packages/salesforce-adapter/src/client/client.ts
```diff
@@ -93,6 +93,8 @@ const MAX_ITEMS_IN_READ_METADATA_REQUEST = 10
 //  https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_listmetadata.htm?search_text=listmetadata
 const MAX_ITEMS_IN_LIST_METADATA_REQUEST = 3
 
+const DEPLOY_STATUS_POLLING_INTERVAL_MS = 500
```
amitgoldarb, line 96: I wonder if this is too frequent. Not sure, this should probably be configurable via the adapter.config though so we won't need a deploy to fix this if this causes some issues for a customer.

amitgoldarb (commented):
Looking good!
Added a few comments.

packages/salesforce-adapter/src/metadata_deploy.ts
tomsellek: > Is this new polling API throttled?

I don't know. I can't find anything specifically about it beyond the general metadata API limits. @tamtamirr any reason to believe the `checkDeployStatus` is covered by anything other than the general metadata API rate limits?

>I also think the 500ms is too little, my intuition is somewhere along the lines of 5s

Sure, I'll go with whatever interval you decide

> I think this should probably be a deploy parameter in the adapter.config
Hmmm.... My gut reaction is that this is too low-res to be a config knob though. Should we consult our Product friends?

packages/salesforce-adapter/src/client/client.ts
tomsellek: I'll just use the existing `client.polling.interval` config

packages/salesforce-adapter/src/client/client.ts
tomsellek: Tested. There were a couple of bugs but it works OK now.

packages/salesforce-adapter/src/client/client.ts
```diff
@@ -800,29 +803,63 @@ export default class SalesforceClient {
...
+      zip,
+      {
+        ...defaultDeployOptions,
+        ..._.pick(this.config?.deploy, optionsToSend),
+        checkOnly,
+      },
     )
-    this.setFetchPollingTimeout()
-    return deployResult
+    this.setDeployPollingTimeout()
```
tomsellek, line 831: The existing code was `this.setFetchPollingTimeout()` which is either a bug or something done for a reason I completely missed.

amitgoldarb: https://github.com/salto-io/salto/pull/3042/files#r879435492 - Please look at this comment. I think this is intentional. It's meant to revert to the original timeout afterwards

amitgoldarb (approved):
Approving to save cycles.
Before merging I think we should receive @tomermevorach 's approval on how this looks in the CLI progress, and how this works in the SAAS.
Other than that, no more concerns from me.

packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -321,21 +324,44 @@ const getDeployStatusUrl = async (
...
 }
 
+const deployProgressMessage = async (
+  client: SalesforceClient,
+  deployResult: SFDeployResult,
+): Promise<string> => {
+  const url = await getDeployStatusUrl(deployResult, client)
+  const testStatus = `${deployResult.numberTestsCompleted}/${deployResult.numberComponentsTotal} (${deployResult.numberTestErrors} errors)`
+  const componentStatus = `${deployResult.numberComponentsDeployed}/${deployResult.numberComponentsTotal} (${deployResult.numberComponentErrors} errors)`
+  const logLine = `Tests: ${testStatus} Components: ${componentStatus} URL: ${url}`
```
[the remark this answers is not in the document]
tomsellek: Ack

packages/salesforce-adapter/src/client/client.ts
```diff
@@ -800,29 +803,63 @@ export default class SalesforceClient {
...
+
+    let deployResult: DeployResult
```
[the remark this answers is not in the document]
tomsellek: Sure

tomsellek:
@tomermevorach @pgonzaleznetwork I'm still trying to figure out how to display this progress info in the SaaS UI, but here's what it looks like in the tooltip. Note the last parts: "Tests xxx/yyy, Components: zzz/aaa, URL: uuu".
image

PR #5093, opened 2023-11-16
SALTO-5041: Using jsforce and jsforce-types from salto-io npm

tamtamirr:
I've uploaded both **jsforce** and **jsforce-types** to npm, so we can use a fixed version from npm instead of a github commit hash.


packages/cli/tsconfig.json
```diff
@@ -7,7 +7,10 @@
         "resolveJsonModule": true,
         "paths": {
             "jsforce": [
-                "../../node_modules/jsforce-types"
+                "../../node_modules/@salto-io/jsforce-types"
+            ],
+            "jsforce-types": [
+              "../../node_modules/@salto-io/jsforce-types"
```
tamtamirr, line 13: I've done this mapping so the current imports won't break.

packages/vscode/tsconfig.json
tamtamirr, line 1: Unrelated to this PR, but are we aware we have several packages importing either **jsforce / jsforce-types** even though none of them has it listed in their **package.json**? cc. @ori-moisis @amitizle

eshedg (approved):
🚀

PR #5098, opened 2023-11-19
SALTO-5050: fix jira e2e flakiness

shayc331:
_fix jira e2e flakiness_

---

* add a header to delete automation request in order to solve the 415 issue

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/jira-adapter/src/client/client.ts
+++ b/packages/jira-adapter/src/client/client.ts
@@ -164,7 +164,7 @@ export default class JiraClient extends clientUtils.AdapterHTTPClient<
   }
 
   public async deletePrivate(
-    args: clientUtils.ClientBaseParams,
+    args: clientUtils.ClientDataParams,
   ): Promise<clientUtils.Response<clientUtils.ResponseValue | clientUtils.ResponseValue[]>> {
     return this.delete({
       ...args,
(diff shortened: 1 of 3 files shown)


omrilit (approved):

packages/jira-adapter/src/filters/automation/automation_deployment.ts
```diff
@@ -237,6 +237,7 @@ const removeAutomation = async (
 ): Promise<void> => {
   await client.deletePrivate({
     url: `${getUrlPrefix(cloudId)}/GLOBAL/rule/${instance.value.id}`,
+    headers: { 'Content-Type': 'application/json' },
```
netama, line 240: do we ever not want to send this header? might be safer to always add it in the client instead? seems this is its only use

packages/jira-adapter/src/client/client.ts
```diff
@@ -164,7 +164,7 @@ export default class JiraClient extends clientUtils.AdapterHTTPClient<
   }
 
   public async deletePrivate(
-    args: clientUtils.ClientBaseParams,
+    args: clientUtils.ClientDataParams,
```
netama, line 167: the fact we needed to change this makes me suspect we are sending a body in the delete where we shouldn't, @shir-reifenberg wdyt? should we add the `omitRequestBody` flag instead?

shir-reifenberg: I also suspected at first, but I don't think it's related because in automation we don't send body anyway

PR #5102, opened 2023-11-20
SALTO-5054: Fix forms bug

idoamit8:
In this PR I added a fix to forms instance in JSM. 

---

Currently, We cannot deploy forms with empty values in JSM. 
This PR fixes this issue.

---
_Release Notes_: 
_Jira Adapter:_
Fix that allows deployment of forms with empty values. 

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/forms/forms.ts
+++ b/packages/jira-adapter/src/filters/forms/forms.ts
@@ -14,11 +14,10 @@
 * limitations under the License.
 */
 
-import { CORE_ANNOTATIONS, Change, InstanceElement, getChangeData, isAdditionChange, isAdditionOrModificationChange, isInstanceChange, isInstanceElement } from '@salto-io/adapter-api'
+import { CORE_ANNOTATIONS, Change, InstanceElement, ReferenceExpression, getChangeData, isAdditionChange, isAdditionOrModificationChange, isInstanceChange, isInstanceElement } from '@salto-io/adapter-api'
 import { values as lowerDashValues } from '@salto-io/lowerdash'
-import { getParent } from '@salto-io/adapter-utils'
+import { getParent, pathNaclCase } from '@salto-io/adapter-utils'
 import _ from 'lodash'
-import { elements as elementUtils, config as configUtils } from '@salto-io/adapter-components'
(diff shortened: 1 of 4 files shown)


packages/jira-adapter/src/filters/forms/forms.ts
```diff
@@ -105,35 +102,33 @@ const filter: FilterCreator = ({ config, client, getElemIdFunc }) => ({
...
-              entry: formValue,
-              type: formType,
-              transformationConfigByType: getTransformationConfigByType(jsmDuckTypeApiDefinitions.types),
-              transformationDefaultConfig: jsmDuckTypeApiDefinitions.typeDefaults.transformation,
-              parent: project,
-              defaultName: name,
-              getElemIdFunc,
-            })
+            return new InstanceElement(
+              name,
```
shir-reifenberg, line 111: nit: can we use `getInstanceNaclName` & `getInstanceFilePath` for creating name & path? 
if not, let's remove the `Form` from config so it will be clear that we're not using this for instance creation

packages/jira-adapter/src/filters/forms/forms.ts
```diff
@@ -105,35 +102,33 @@ const filter: FilterCreator = ({ config, client, getElemIdFunc }) => ({
...
-    const fieldNames = ['questions', 'sections', 'conditions']
     changes
       .filter(isInstanceChange)
       .map(change => getChangeData(change))
       .filter(instance => instance.elemID.typeName === FORM_TYPE)
       .forEach(instance => {
-        fieldNames.forEach(fieldName => {
-          instance.value.design[fieldName] = instance.value.design[fieldName] ?? {}
-        })
+        instance.value.updated = new Date().toISOString()
```
shir-reifenberg, line 131: it must have this field in order for deploy to work? are we putting here a random date?

shir-reifenberg (approved):
added a few comments but approving to unblock :)

packages/jira-adapter/src/filters/forms/forms.ts
idoamit8: Yes. It must have that field. It's not random, its the current time and date

PR #5103, opened 2023-11-20
SALTO-5052: Added Forms to E2E

idoamit8:
In this PR I added Forms to E2E
Currently we compare the time in the ms level so the test if flaky. Change it to the minute time. 

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/test/filters/forms/forms.test.ts
+++ b/packages/jira-adapter/test/filters/forms/forms.test.ts
@@ -485,7 +485,7 @@ describe('forms filter', () => {
       })
       it('should add the current updated time', async () => {
         await filter.preDeploy([{ action: 'add', data: { after: formInstance } }])
-        expect(formInstance.value.updated).toEqual(new Date().toISOString())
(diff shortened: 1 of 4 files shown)


packages/jira-adapter/test/filters/forms/forms.test.ts
```diff
@@ -485,7 +485,7 @@ describe('forms filter', () => {
       })
       it('should add the current updated time', async () => {
         await filter.preDeploy([{ action: 'add', data: { after: formInstance } }])
-        expect(formInstance.value.updated).toEqual(new Date().toISOString())
+        expect(formInstance.value.updated.slice(0, -8)).toEqual(new Date().toISOString().slice(0, -8))
```
shayc331, line 488: you can maybe use `moment` library like [here](https://stackoverflow.com/questions/35758963/remove-seconds-milliseconds-from-date-convert-to-iso-string)

shayc331 (approved):
🏅

PR #5109, opened 2023-11-21
SALTO-4970: Salesforce Quick fetch support for types with nested instances

tamtamirr:
This PR adds support in quick fetch of nested instances.

---

---
_Release Notes_: 

---
_User Notifications_: 


tamtamirr:
Old work. Will recreate PR

PR #5112, opened 2023-11-22
SALTO-4765 Change default excluding in NS adapter config

dantal4:
add the following default exclusions to NS config:

1. exclude inactive elements
2. exclude private saved searches
3. exclude data files

---

---
_Release Notes_: 
Netsuite Adapter:
- Change default excluding in the adapter config

---
_User Notifications_: 
None

### DIFF
--- a/packages/netsuite-adapter/src/config.ts
+++ b/packages/netsuite-adapter/src/config.ts
@@ -463,6 +463,12 @@ export const fetchDefault: FetchParams = {
           .join('|'),
       }, // may be a lot of data that takes a lot of time to fetch
       ...inactiveElementsCriteria,
+      {
+        name: SAVED_SEARCH,
+        criteria: {
+          FLAG_PUBLIC: false,
+        },
+      },
     ],
     fileCabinet: [
       '^/Templates/Letter Templates/Mail Merge Folder.*',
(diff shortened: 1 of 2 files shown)


packages/netsuite-adapter/src/config.ts
```diff
@@ -451,9 +462,12 @@ export const fetchDefault: FetchParams = {
           .filter(itemTypeName => !['giftCertificateItem', 'downloadItem'].includes(itemTypeName))
           .join('|'),
       }, // may be a lot of data that takes a lot of time to fetch
+      ...inactiveElementsCriteria,
     ],
     fileCabinet: [
       '^/Templates/Letter Templates/Mail Merge Folder.*',
+      FILES_TO_EXCLUDE_REGEX,
```
omrilit, line 469: The fact that we don't have here all of the files that we exclude, e.g. `^/Templates/Letter Templates/Mail Merge Folder.*`, make me think that the naming should nbe more specific.
Maybe something like FILE_SUFFIXES_TO_EXCLUDE_REGEX, FILE_TYPES_TO_EXCLUDE_REGEX?

omrilit (approved):
Looks great!
Approving, but please make sure not to merge unintentionally before it's communicated properly

PR #5114, opened 2023-11-22
SALTO-5070: Move Salesforce retrieve settings into the Fetch Profile

tamtamirr:
Required step for fetch with changes detection on Profiles. 

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


tomsellek (approved):
🤷

PR #5115, opened 2023-11-22
SALTO-4894 Jira adapter - remove missing field configuration items in fix element

kami-delaroz:
SALTO-4894 remove missing field configuration items in fix element

---

None

---
_Release Notes_: 

_Jira adapter_:
* Fix bug of deployment partial success when missing fields where removed because they do not exist in the target env

---
_User Notifications_: 
None


packages/jira-adapter/src/adapter.ts
```diff
@@ -294,8 +293,6 @@ export const DEFAULT_FILTERS = [
   fieldConfigurationIrrelevantFields,
   // Must run after fieldConfigurationIrrelevantFields
   fieldConfigurationSplitFilter,
-  // Must run after fieldReferencesFilter
-  replaceFieldConfigurationReferences,
```
netama, line 298: how come this is no longer needed? did we decide we no longer handle `splitFieldConfiguration` being falsy? or is it covered somewhere else regardless of this change?

packages/jira-adapter/src/weak_references/field_configuration_items.ts
```diff
@@ -0,0 +1,114 @@
...
+      return fixedInstance
+    })
+    .filter(values.isDefined)
+    .toArray()
+
+  const errors = fixedElements.map(instance => ({
+    elemID: instance.elemID.createNestedID('fields'),
+    severity: 'Info' as const,
+    message: 'Deploying field configuration without all attached field configuration items',
+    detailedMessage: 'This field configuration is attached to some field configuration items that do not exist in the target environment. It will be deployed without referencing these field configuration items.',
```
netama, line 106: were these coordinated with @liorn ? (just wondering about the `attached` terminology + maybe replacing `without all` to `without some` and the explanation that was omitted + no longer listing the items that are omitted, and I think the previous message was defined by him - as long as you're aligned on the changes no objections from my side :) )

packages/jira-adapter/src/weak_references/field_configuration_items.ts
```diff
@@ -0,0 +1,114 @@
...
+  }))
+  return { fixedElements, errors }
+}
+
+export const fieldConfigurationsHandler: WeakReferencesHandler = {
```
netama, line 111: please add a docstring explaining what this handler does in each part
also, if we don't have one yet, maybe add `log.time()` surrounding these functions so we know how long they take (like we have in filters and change validators)? can also do just for these for now if you don't want to deal with the shared mechanism and we can do a follow-up in the infra if you prefer, but since we're going to have this running on some very large envs it could be useful

packages/jira-adapter/src/weak_references/field_configuration_items.ts
```diff
@@ -0,0 +1,114 @@
...
+
+const getFieldConfigurationItemsReferences: GetCustomReferencesFunc = async elements =>
+  awu(elements)
+    .filter(isInstanceElement)
+    .filter(instance => instance.elemID.typeName === FIELD_CONFIGURATION_TYPE_NAME)
+    .flatMap(instance => getFieldReferences(instance))
+    .toArray()
+
+const removeMissingFields: WeakReferencesHandler['removeWeakReferences'] = ({ elementsSource })
+  : FixElementsFunc => async elements => {
```
netama, line 73: nit: I think no need for `: FixElementsFunc`

packages/jira-adapter/src/weak_references/field_configuration_items.ts
```diff
@@ -0,0 +1,114 @@
...
+import { WeakReferencesHandler } from './weak_references_handler'
+
+const { awu } = collections.asynciterable
+
+
+type FieldConfigurationItems = {
+  id: unknown
+  description: string
+  isHidden: boolean
+  isRequired: boolean
```
netama, line 37: these don't seem aligned with the `FIELD_CONFIGURATION_ITEMS_SCHEME` Joi scheme?

packages/jira-adapter/src/weak_references/field_configuration_items.ts
```diff
@@ -0,0 +1,114 @@
...
+): Promise<ReferenceInfo[]> => {
+  const fieldConfigurationItems = instance.value.fields
+  if (fieldConfigurationItems === undefined || !isFieldConfigurationItems(fieldConfigurationItems)) {
+    return []
+  }
+
+  return awu(fieldConfigurationItems)
+    .map(async (field, index) => (
+      isReferenceExpression(field.id)
+        ? { source: instance.elemID.createNestedID(index.toString(), 'fieldId'), target: field.id.elemID, type: 'weak' as const }
```
netama, line 59: (for all of this code) if we're not removing the option to have `splitFieldConfiguration` off the source could be a map, no? (seems the old change validator actually assumed it is a map)

packages/jira-adapter/src/weak_references/field_configuration_items.ts
```diff
@@ -0,0 +1,114 @@
...
+
+const removeMissingFields: WeakReferencesHandler['removeWeakReferences'] = ({ elementsSource })
+  : FixElementsFunc => async elements => {
+  const fixedElements = await awu(elements)
+    .filter(isInstanceElement)
+    .filter(instance => instance.elemID.typeName === FIELD_CONFIGURATION_TYPE_NAME)
+    .map(async instance => {
+      const fieldConfigurationItems = instance.value.fields
+      if (fieldConfigurationItems === undefined || !isFieldConfigurationItems(fieldConfigurationItems)) {
+        return undefined
```
netama, line 80: nit: if this is an unexpected scenario maybe add a more informative error log (with the instance's details as well)?

packages/jira-adapter/src/config/config.ts
```diff
@@ -281,7 +280,6 @@ const changeValidatorConfigType = createMatchingObjectType<ChangeValidatorConfig
     workflowTransitionDuplicateName: { refType: BuiltinTypes.BOOLEAN },
     permissionSchemeDeployment: { refType: BuiltinTypes.BOOLEAN },
     projectCategory: { refType: BuiltinTypes.BOOLEAN },
-    unresolvedFieldConfigurationItems: { refType: BuiltinTypes.BOOLEAN },
```
netama, line 284: not for this PR, but if we don't have a mechanism for disabling the fixElement code we should probably add it eventually as well. opened a ticket and linked to this one

packages/jira-adapter/src/change_validators/unresolved_field_configuration_items.ts
```diff
@@ -1,51 +0,0 @@
...
-const { awu } = collections.asynciterable
-
-const getUnresolvedFieldConfigurationItems = async (
-  instance: InstanceElement,
-  elementsSource: ReadOnlyElementsSource
-): Promise<string[]> => awu(Object.keys(instance.value.fields ?? {}))
-  .filter(async fieldName => !(await elementsSource.has(new ElemID(JIRA, FIELD_TYPE_NAME, 'instance', fieldName))))
-  .toArray()
-
-export const unresolvedFieldConfigurationItemsValidator: ChangeValidator = async (changes, elementsSource) => {
```
netama, line 30: seems we removed the change validator but didn't change the filter that handles the deploy silently - I'm wondering if we should add some log in the deploy filter now when filtering out unresolved references, now that it is supposed to be an unreachable case?

packages/jira-adapter/src/weak_references/field_configuration_items.ts
```diff
@@ -0,0 +1,114 @@
...
+      }
+
+      const fixedInstance = instance.clone()
+      fixedInstance.value.fields = await awu(fieldConfigurationItems)
+        .filter(async field => (field.id === undefined
+          || (
+            isReferenceExpression(field.id)
+            // eslint-disable-next-line no-return-await
+            && await elementsSource.has(field.id.elemID)
+          )))
```
netama, line 90: might be worth benchmarking this on a large env to make sure this is efficient?

netama (commented):
new logic looks good for the main case :) added some minor comments + a question about handling the `splitFieldConfiguration` config flag

packages/jira-adapter/src/adapter.ts
kami-delaroz: I wanted to ask you about it.
Before I removed this filter, the fields value was a dictionary with no references for the fields, only the field name as a key in the dictionary.
I removed it, as I understood this was there because we couldn't mark them as **weak** references.
I saw there was the splitFieldConfiguration filter, and that the condition was "take this or that". but I'm not sure I need any of them. I'll give it a look tomorrow, but I guess we'll need to zoom about it on Monday

packages/jira-adapter/src/filters/field_configuration/replace_field_configuration_references.ts
```diff
@@ -33,33 +29,21 @@ const replaceToMap = (instance: InstanceElement): void => {
     .filter((field: Values) => isResolvedReferenceExpression(field.id))
     .map((field: Values) => [
       field.id.elemID.name,
-      _.omit(field, 'id'),
+      field,
```
netama, line 32: tiny nit: can now use `_.keyBy` and simplify a bit e.g.
```
  instance.value.fields = _.keyBy(
    instance.value.fields.filter((field: Values) => isResolvedReferenceExpression(field.id)),
    field => field.id.elemID.name,
  )

```

packages/jira-adapter/src/filters/field_configuration/replace_field_configuration_references.ts
```diff
@@ -33,33 +29,21 @@ const replaceToMap = (instance: InstanceElement): void => {
...
-        log.debug(`Omitting element id ${elemId.getFullName()} from ${instance.elemID.getFullName()} since it does not exist in the account`)
-        return undefined
-      }
-      return {
-        id: new ReferenceExpression(elemId, fieldInstance),
-        ...config,
-      }
-    })
+  instance.value.fields = await awu(Object.values(instance.value.fields))
+    .map(async (config: Value) => config)
```
netama, line 41: does this do anything? (and do we still need `awu` here?)

packages/jira-adapter/src/filters/field_configuration/replace_field_configuration_references.ts
```diff
@@ -98,7 +82,7 @@ const filter: FilterCreator = ({ config, elementsSource }) => ({
       .filter(isInstanceElement)
       .filter(instance => instance.elemID.typeName === FIELD_CONFIGURATION_TYPE_NAME)
       .filter(instance => instance.value.fields !== undefined)
-      .forEach(async instance => replaceFromMap(instance, elementsSource))
+      .forEach(async instance => replaceFromMap(instance))
```
netama, line 85: continuing from above - 
1. can this become sync?
2. can this just be
```suggestion
      .forEach(replaceFromMap)
```
(and can we remove `awu` here as well)?

packages/jira-adapter/src/weak_references/field_configuration_items.ts
```diff
@@ -0,0 +1,125 @@
...
+    render: string
+  }
+}
+
+const FIELD_CONFIGURATION_ITEM_SCHEME = Joi.object({
+  id: Joi.optional(),
+  description: Joi.optional(),
+  isHidden: Joi.optional(),
+  isRequired: Joi.optional(),
+  renderer: Joi.optional(),
```
netama, line 43: why not define the field types correctly?

packages/jira-adapter/src/weak_references/field_configuration_items.ts
```diff
@@ -0,0 +1,125 @@
...
+ * Remove invalid fields (not references or missing references) from field configuration.
+ */
+const removeMissingFields: WeakReferencesHandler['removeWeakReferences'] = ({ elementsSource }) => async elements => log.time(async () => {
+  const fixedElements = await awu(elements)
+    .filter(isInstanceElement)
+    .filter(instance => instance.elemID.typeName === FIELD_CONFIGURATION_TYPE_NAME)
+    .map(async instance => {
+      const fieldConfigurationItems = instance.value.fields
+      if (fieldConfigurationItems === undefined || !isFieldConfigurationItems(fieldConfigurationItems)) {
+        log.warn(`fields value is corrupted in instance ${instance.elemID.getFullName()}`)
```
netama, line 88: thanks for adding the log - I think it might be worth explaining the impact (this is the main purpose of the log) - e.g. "..., not omitting missing fields"

packages/jira-adapter/src/weak_references/field_configuration_items.ts
```diff
@@ -0,0 +1,125 @@
...
+      }
+
+      const fixedInstance = instance.clone()
+      fixedInstance.value.fields = await awu(Object.entries(fieldConfigurationItems))
+        .filter(async ([_id, field]) =>
+          (isReferenceExpression(field.id)
+            // eslint-disable-next-line no-return-await
+            && await elementsSource.has(field.id.elemID)
+          ))
+        .reduce((obj, [id, field]) => {
```
netama, line 99: in general we try to avoid `.reduce()` for readability - any chance to avoid it here? e.g. with `_.pickBy()` / `_.omitBy()`? can also shorten the code and help avoid the cast

packages/jira-adapter/src/weak_references/field_configuration_items.ts
netama: did this happen?

netama (approved):
added some minor comments but other than that looks good 🎈 , so pre-approving

packages/jira-adapter/src/weak_references/field_configuration_items.ts
liorn: WDYT about:
message: Deploying field configuration without all of its fields
detailedMessage: This field configuration references some fields that do not exist in the target environment. It will be deployed without them.

PR #5117, opened 2023-11-23
SAAS-9617: Make referenceSourcesIndex optional

dorsev:
It can only be used when elementSelectors are provided, and since elementSelectors is optional, this should be as well.

---

---

_Release Notes_: 
None

---

_User Notifications_: 
None

### DIFF
--- a/packages/core/src/core/diff.ts
+++ b/packages/core/src/core/diff.ts
@@ -136,12 +136,12 @@ export function createDiffChanges(
 export async function createDiffChanges(
   toElementsSrc: elementSource.ElementsSource,
   fromElementsSrc: elementSource.ElementsSource,
-  referenceSourcesIndex: remoteMap.ReadOnlyRemoteMap<ElemID[]> | undefined,
+  referenceSourcesIndex: remoteMap.ReadOnlyRemoteMap<ElemID[]> = new remoteMap.InMemoryRemoteMap<ElemID[]>(),
   elementSelectors: ElementSelector[] = [],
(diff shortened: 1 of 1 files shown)


packages/core/src/core/diff.ts
```diff
@@ -120,28 +120,28 @@ const createMatchers = async (
...
-  referenceSourcesIndex: remoteMap.ReadOnlyRemoteMap<ElemID[]>,
+  referenceSourcesIndex?: remoteMap.ReadOnlyRemoteMap<ElemID[]>,
   elementSelectors?: ElementSelector[],
   topLevelFilters?: IDFilter[],
   resultType?: 'detailedChanges'
 ): Promise<DetailedChange[]>
 export async function createDiffChanges(
   toElementsSrc: elementSource.ElementsSource,
   fromElementsSrc: elementSource.ElementsSource,
-  referenceSourcesIndex: remoteMap.ReadOnlyRemoteMap<ElemID[]>,
```
dantal4, line 139: In theory, you can have selectors without referencedBy in them, and then you don't need referenceSourcesIndex as well. Maybe we can set a default, and remove your change in line 144:
```
referenceSourcesIndex = new remoteMap.ReadOnlyRemoteMap<ElemID[]>(),
```

dantal4 (approved):
Very helpful, thanks!

packages/core/src/core/diff.ts
dorsev: good suggestion

PR #5119, opened 2023-11-23
SALTO-4840: Salesforce Profiles Fetch With Changes Detection

tamtamirr:
This PR  adds support for Profiles in fetch with changes detection mode.

---

Fetch times from testing:

Full Fetch: Ranging between 140 - 160 seconds
Fetch with changes detection where an Apex Class was added, and it's settings were added to all Profiles: 25 - 30 seconds
Fetch with changes detection where 3 Profiles were modified: 45 seconds - 55 seconds
Fetch with changes detection where 3 Profiles were modified alongside addition of Apex Class: 70-75 seconds

Will add UTs as soon as the design is approved.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/filters/merge_profiles_with_source_values.ts
```diff
@@ -0,0 +1,59 @@
...
+    if (profileInstances.length === 0) {
+      return
+    }
+    const profileValuesFromSourceByFullName: Record<string, Values> = {}
+    await awu(await config.elementsSource.getAll())
+      .filter(isInstanceOfTypeSync(PROFILE_METADATA_TYPE))
+      .forEach(instanceFromSource => {
+        const fullName = apiNameSync(instanceFromSource) ?? ''
+        profileValuesFromSourceByFullName[fullName] = instanceFromSource.value
+      })
```
tomsellek, line 47: Can this be done via some combination of `groupBy` and `mapValues` instead of using a side effect?

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -489,3 +496,50 @@ export const retrieveMetadataInstances = async ({
...
+    return { profileProps, nonProfileProps }
+  }
+
+  const retrievePartialProfileInstances = retrieveMetadataInstances({
+    ...params,
+    getFilesToRetrieveFunc: allProps => {
+      const { profileProps, nonProfileProps } = getPartitions(allProps)
+      if (nonProfileProps.length === 0) {
+        log.debug('No profile related props were changed')
+        return []
```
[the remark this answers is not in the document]
tamtamirr: Yes, because the current implementation of Profiles retrieve does not pick specific file props of the related types. If filesToRetrieve.length === 0 there's no reason to retrieve the partial profiles

packages/salesforce-adapter/src/filters/merge_profiles_with_source_values.ts
tamtamirr: Changed my implementation to use reduce

packages/salesforce-adapter/src/fetch.ts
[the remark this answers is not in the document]
tomsellek: I'm not sure that is the case. There are some sections of the profiles that are retrieved even if we don't send anything else in the retrieve request, and those sections could be the reason the profile is marked as modified, no?

ref. https://www.notion.so/saltoio/c057cf5b0ff94f33b165dde46081cd06?v=85e26eac95594b80a86fe1049035f14d&pvs=4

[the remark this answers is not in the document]
tomsellek: Fair. In that case perhaps change the log to something like `The types that would affect the contents of the profile did not change.` ?

tomsellek (approved):
Approved when UTs are added.

packages/salesforce-adapter/src/filters/merge_profiles_with_source_values.ts
```diff
@@ -0,0 +1,69 @@
...
+    log.debug('about to merge the following profiles with their source values: %s', safeJsonStringify(profileInstances.map(instance => apiNameSync(instance))))
+    if (profileInstances.length === 0) {
+      return
+    }
+    const profileValuesFromSourceByFullName = await awu(await config.elementsSource.getAll())
+      .filter(isInstanceOfTypeSync(PROFILE_METADATA_TYPE))
+      .reduce<Record<string, Values>>((acc, instanceFromSource) => {
+        const fullName = apiNameSync(instanceFromSource)
+        if (fullName === undefined) {
+          log.warn('profile instance from source %s does not have fullName', instanceFromSource.elemID.getFullName())
```
tomsellek, line 52: nit: won't this log always print 'undefined' as the fullName?

tamtamirr: Please note we log the `elemID fullName` and not `instance.value.fullName`

PR #5121, opened 2023-11-23
SALTO-4968: Cache Salesforce Client listMetadataObjects results

tamtamirr:
This PR adds caching on the results of **listMetadataObjects**.

---

Main reason for this caching is the fact we have some MetadataTypes we list more than once, especially in **fetch with changes detection mode**. Listing types can be time consuming, depending on the amount of **Instances** per type. It's widely noticeable around **CustomFields** and **CustomObjects**.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/src/client/client.ts
+++ b/packages/salesforce-adapter/src/client/client.ts
@@ -637,6 +637,22 @@ export default class SalesforceClient {
     listMetadataQuery: ListMetadataQuery | ListMetadataQuery[],
     isUnhandledError: ErrorFilter = isSFDCUnhandledException,
   ): Promise<SendChunkedResult<ListMetadataQuery, FileProperties>> {
+    const sendChunkedList = async (
+      input: typeof listMetadataQuery
+    ): Promise<SendChunkedResult<ListMetadataQuery, FileProperties>> => (
+      sendChunked({
+        operationInfo: 'listMetadataObjects',
+        input,
+        sendChunk: chunk => this.retryOnBadResponse(() => this.conn.metadata.list(chunk)),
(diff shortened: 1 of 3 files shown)


packages/salesforce-adapter/src/client/client.ts
```diff
@@ -635,12 +637,38 @@ export default class SalesforceClient {
...
+    }
+    const [cachedQueries, nonCachedQueries] = _.partition(
+      makeArray(listMetadataQuery),
+      query => Object.keys(this.filePropsByType).includes(query.type)
+    )
+    const cachedProps = cachedQueries.flatMap(query => makeArray(this.filePropsByType[query.type]))
+    if (nonCachedQueries.length === 0) {
+      log.debug('returning cached listMetadataObjects for %s', safeJsonStringify(listMetadataQuery))
+      return { result: cachedProps, errors: [] }
+    }
```
tomsellek, line 664: If you taught `sendChunkedList` to handle empty arrays as its parameter this `if` would not be necessary, right?

tomsellek (approved):
LGTM. 1 question on a possible simplification.

packages/salesforce-adapter/src/client/client.ts
tamtamirr: This if is relevant when all of the queried typed were already queried before. So we return them from the cache :)

tamtamirr: It just makes more sense IMO to execute this `if` before invoking `sendChunkedList`.  L661-664 will have to be either here or within the private method **sendChunkedList** so I don't see how it simplifies the code TBH 🤔

PR #5123, opened 2023-11-23
SALTO-4726 Move fallback users replacement to be FixElements

adfineberg:
Move fallback users replacement to be FixElements

---

---
_Release Notes_: 
__Zendesk Adapter__
* Move the replacement of fallback user emails to happen before the deploy in order to reduce incorrect user errors (partial successes).

---
_User Notifications_: 

### DIFF
--- a/packages/zendesk-adapter/src/fallback_user/user.ts
+++ b/packages/zendesk-adapter/src/fallback_user/user.ts
@@ -1,0 +1,140 @@
+/*
+*                      Copyright 2023 Salto Labs Ltd.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+import {
+  ChangeError,
+  ElemID,
+  FixElementsFunc,
+  InstanceElement,
+  isInstanceElement,
+} from '@salto-io/adapter-api'
+import { client as clientUtils } from '@salto-io/adapter-components'
+import { resolvePath, setPath } from '@salto-io/adapter-utils'
+import { logger } from '@salto-io/logging'
+import { values } from '@salto-io/lowerdash'
+import _ from 'lodash'
+import { paginate } from '../client/pagination'
+import { MISSING_USERS_DOC_LINK, MISSING_USERS_ERROR_MSG, TYPE_NAME_TO_REPLACER, VALID_USER_VALUES, getUserFallbackValue, getUsers } from '../user_utils'
+import { FallBackUserHandler } from './fallback_user_handler'
+
+const log = logger(module)
+const { createPaginator } = clientUtils
+
+const getFallbackUserIsMissingError = (
+  instance: InstanceElement,
+  missingUsers: string[],
+  userFallbackValue: string
+): ChangeError => ({
+  elemID: instance.elemID,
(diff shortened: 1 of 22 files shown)


packages/zendesk-adapter/src/fallback_user/user.ts
```diff
@@ -0,0 +1,140 @@
...
+  client,
+  config
+): FixElementsFunc => async elements => {
+  const paginator = createPaginator({
+    client,
+    paginationFuncCreator: paginate,
+  })
+  const users = await getUsers(paginator)
+  const { defaultMissingUserFallback } = config
+  if (_.isEmpty(users) || defaultMissingUserFallback === undefined) {
```
edenhassid, line 109: in case someone uses it and sends false instead of {}
```suggestion
  if (_.isEmpty(users) || defaultMissingUserFallback === undefined || defaultMissingUserFallback === false) {
```

packages/zendesk-adapter/src/fallback_user/user.ts
```diff
@@ -0,0 +1,140 @@
...
+  return userPaths.map(path => {
+    const user = resolvePath(instance, path)
+    if (!VALID_USER_VALUES.includes(user) && !users.has(user)) {
+      return { user, path }
+    }
+    return undefined
+  }).filter(values.isDefined)
+}
+
+const getFallbackUserIsMissingUsers = (users: Set<string>) => (
```
edenhassid, line 72: the name `getFallbackUserIsMissingUsers` is confusing, can you think of. something else?

packages/zendesk-adapter/src/fallback_user/user.ts
```diff
@@ -0,0 +1,140 @@
...
+  const fallbackValue = await getUserFallbackValue(
+    defaultMissingUserFallback,
+    userEmails,
+    client
+  )
+  if (fallbackValue === undefined) {
+    log.error('Error while trying to get defaultMissingUserFallback value')
+    const errors = elements.filter(isInstanceElement)
+      .map(getFallbackUserIsMissingUsers(userEmails))
+      .filter(values.isDefined)
```
edenhassid, line 123: why do you need this? can `getFallbackUserIsMissingUsers` return undefined? if so can you change the return value?

packages/zendesk-adapter/src/fallback_user/user.ts
```diff
@@ -0,0 +1,140 @@
...
+  ({ instance, users: _.uniq(getMissingUserPaths(users, instance).map(({ user }) => user)) })
+
+const replaceMissingUsers = (
+  users: Set<string>,
+  fallbackUser: string
+) => (instance: InstanceElement): undefined |
+{ fixedInstance: InstanceElement; missingUsers: string[] } => {
+  const missingUserPaths = getMissingUserPaths(users, instance)
+
+  if (!missingUserPaths) {
```
edenhassid, line 84:
```suggestion
  if (_.isEmpty(missingUserPaths)) {
```

packages/zendesk-adapter/test/fallback_user/user.test.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  describe('fallback user not provided', () => {
+    it('should not replace missing users and should not report errors', async () => {
+      const instances = [macroInstance, articleInstance].map(e => e.clone())
+      const { fixedElements, errors } = await usersHandler.missingUsersToFallback(
+        client,
+        { defaultMissingUserFallback: undefined },
+      )(instances)
+      const fallbackMacro = macroInstance.clone()
+      fallbackMacro.value.actions[1].value = 'fallback@.com'
```
edenhassid, line 142: why do you do this? you don't use it anywhere

edenhassid (commented):
looks good! 💯 🤩

packages/zendesk-adapter/src/fallback_user/user.ts
edenhassid: you are right, my bad :)

edenhassid (approved):
LGTM! 🔥

PR #5127, opened 2023-11-26
SALTO-5086: remove old state migration code

ori-moisis:
Cleaning up the code before working on new features in the area

This also contains a couple of unit test fixes in other packages

---

---
_Release Notes_: 
Core:
- Removed code that would migrate very old state files to the "new" format - this would only affect workspaces that were not used for over 10 months

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/cli/test/command_builder.test.ts
+++ b/packages/cli/test/command_builder.test.ts
@@ -13,7 +13,7 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
-import { logger } from '@salto-io/logging'
+import { logger, LogLevel } from '@salto-io/logging'
 import { loadLocalWorkspace } from '@salto-io/core'
 import { mockFunction } from '@salto-io/test-utils'
 import * as mocks from './mocks'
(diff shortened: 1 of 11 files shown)


packages/cli/test/command_builder.test.ts
```diff
@@ -117,23 +117,38 @@ describe('Command builder', () => {
...
-          config,
-          workspacePath: '.',
-          telemetry,
-          spinnerCreator,
-          commanderInput: [
-            'str',
-            ['str2', 'str3'],
-            { bool: true, choices: 'a', verbose: true },
-          ],
+      describe('when log level is above debug', () => {
```
ori-moisis, line 120: unrelated - this was just annoying since it would fail if the test was run from a terminal where `SALTO_LOG_LEVEL=debug` was defined

packages/workspace/test/workspace/workspace.test.ts
```diff
@@ -2007,13 +2007,18 @@ describe('workspace', () => {
...
     const durationAfterLastModificationMs = 1000 * 60 * durationAfterLastModificationMinutes
     beforeEach(async () => {
       now = Date.now()
-      jest.spyOn(Date, 'now').mockImplementation(() => now)
+      mockDateNow = jest.spyOn(Date, 'now')
+      mockDateNow.mockImplementation(() => now)
       modificationDate = new Date(now - durationAfterLastModificationMs)
     })
+    afterEach(() => {
+      mockDateNow.mockRestore()
```
ori-moisis, line 2020: unrelated fix - this would mock `Date.now` and not restore it, causing other tests to fail randomly

packages/workspace/test/workspace/workspace.test.ts
```diff
@@ -2007,13 +2007,18 @@ describe('workspace', () => {
   describe('getStateRecency', () => {
     let now: number
     let modificationDate: Date
+    let mockDateNow: jest.SpiedFunction<typeof Date.now>
     const durationAfterLastModificationMinutes = 7
     const durationAfterLastModificationMs = 1000 * 60 * durationAfterLastModificationMinutes
     beforeEach(async () => {
       now = Date.now()
-      jest.spyOn(Date, 'now').mockImplementation(() => now)
+      mockDateNow = jest.spyOn(Date, 'now')
```
ori-moisis, line 2015: unrelated fix - this would mock `Date.now` and not restore it, causing other tests to fail randomly

packages/workspace/test/workspace/workspace.test.ts
```diff
@@ -2964,10 +2969,13 @@ describe('workspace', () => {
   describe('accountConfig', () => {
     let workspace: Workspace
     let adaptersConfig: MockInterface<AdaptersConfigSource>
+    let resolveMock: jest.SpiedFunction<typeof resolve>
     const configElemId = new ElemID('dummy', 'new')
     const configObjectType = new ObjectType({ elemID: configElemId })
     const configInstanceElement = new InstanceElement('aaa', configObjectType)
-    const resolveMock = jest.spyOn(expressionsModule, 'resolve')
```
ori-moisis, line 2970: unrelated fix - this would mock the `resolve` function even if this test suite wasn't supposed to run.
this caused other tests to fail if they ran before this suite, or if this suite didn't run (e.g - when debugging a single test in this file)

packages/file/src/gzip.ts
```diff
@@ -37,30 +31,3 @@ export const createGZipWriteStream = (
...
   createGzip(),
 ])
-
-
-//
-// the rest of this file contains backward-compatibility with old state file -
-// should be removed once all workspaces have been upgraded (see SALTO-3149)
-//
-
-export const isOldFormatStateZipFile = async (zipFilename: string): Promise<boolean> => {
```
yuribro, line 47: Since these are exported from the package, do we need to do anything special? @deprecate them first? or add it to the release notes?

yuribro (commented):
looks good to me, are we waiting for @netama to approve?

packages/file/src/gzip.ts
ori-moisis: well, technically all of our packages are version 0 (which means API breakage can happen).
I think we should be fine with this kind of breakage of these specific functions, I only worry about breakage with functions and APIs that could conceivably be used by external code, I don't think these qualify

packages/core/src/local-workspace/state.ts
```diff
@@ -193,18 +160,9 @@ export const localState = (
...
-      return currentFilePaths
-    }
-    const oldStateFilePath = `${filePrefix}${ZIPPED_STATE_EXTENSION}`
-    if (await exists(oldStateFilePath)) {
-      pathToClean = oldStateFilePath
-      return [oldStateFilePath]
-    }
-    return []
-  }
+  const getRelevantStateFiles = async (): Promise<string[]> => (
```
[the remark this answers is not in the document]
ori-moisis: probably not

netama (approved):

PR #5129, opened 2023-11-27
SALTO-5087: Salesforce Metadata Partial Fetch Deletions Support

tamtamirr:
This PR implements deletion support for Salesforce Partial Fetch

---

Will add UTs once the PR design is approved.

Design key notes:
- Added the map `listedInstancesByType` that stores the listed fullNames per Metadata Type.
- Added a Proxy on the **SalesforceClient.listMetadataObjects** that populates the `listedInstancesByType`.
- Added logic that compares the **Element IDs** from the Elements Source to **Element IDs** I've created from the populated `listedInstancesByType` to detect deletions
 
---
_Release Notes_: 
Salesforce Adapter:
- Salesforce partial fetch will now support deletions.

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/adapter.ts
```diff
@@ -395,11 +404,32 @@ export default class SalesforceAdapter implements AdapterOperations {
...
+                  const listedInstances = this.listedInstancesByType.get(typeName)
+                  props.forEach(({ fullName }) => listedInstances.add(fullName))
+                })
+              return listMetadataObjectsResult
+            })
+          return proxyListMetadataObjects
+        }
+        return target[propertyKey as keyof SalesforceClient]
+      },
+    })
```
tamtamirr, line 428: I personally think this is the correct place to create the Proxy on the client, even when taking into considering that the `listedInstancesByType` is only used in partial fetch for the time being.

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -481,11 +511,21 @@ export default class SalesforceAdapter implements AdapterOperations {
...
+      if (deletedElemIds.length > 0) {
+        log.debug('The following elemIDs were deleted in partial fetch: %s', safeJsonStringify(deletedElemIds.map(e => e.getFullName())))
+      }
+      return { isPartial: true, deletedElements: deletedElemIds }
+    }
     return {
       elements,
       errors: onFetchFilterResult.errors ?? [],
       updatedConfig,
-      partialFetchData: setPartialFetchData(fetchProfile.metadataQuery.isPartialFetch()),
```
tamtamirr, line 488: Using `setPartialFetchData` didn't make sense anymore

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -668,4 +708,44 @@ export default class SalesforceAdapter implements AdapterOperations {
...
+  }): Promise<Required<PartialFetchData>['deletedElements']> {
+    const metadataTypesByName = _.keyBy(
+      fetchElements.filter(isMetadataObjectType),
+      type => apiNameSync(type) ?? 'unknown'
+    )
+    const elemIdsByTypeFromSource = await this.getElemIdsByTypeFromSource()
+    const deletedElemIds = new Set<ElemID>()
+    Object.entries(elemIdsByTypeFromSource)
+      // We only want to check types that are included. This is especially relevant for targeted fetch
+      .filter(([typeName]) => fetchProfile.metadataQuery.isTypeMatch(typeName))
```
tamtamirr, line 733: Here is where the magic happens for targeted fetch.
Alternatively, could be filtered in `getElemIdsByTypeFromSource` but I think it makes more sense to do the filtering here.

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -481,11 +510,21 @@ export default class SalesforceAdapter implements AdapterOperations {
       configChangeSuggestions,
       this.userConfig,
     )
+    const getPartialFetchData = async (): Promise<PartialFetchData | undefined> => {
+      if (!fetchProfile.metadataQuery.isPartialFetch()) {
+        return undefined
+      }
+      const deletedElemIds = await this.getDeletedElemIdsForPartialFetch({ fetchProfile, fetchElements: elements })
+      if (deletedElemIds.length > 0) {
+        log.debug('The following elemIDs were deleted in partial fetch: %s', safeJsonStringify(deletedElemIds.map(e => e.getFullName())))
```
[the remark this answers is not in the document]
tamtamirr: It shouldn't be too long :)

tomsellek (approved):
I'm not crazy about the whole "spooky action at a distance" aspect of how this PR uses a proxy, but I frankly can't think of a better way to accomplish this. Approved assuming UTs are added.

PR #5130, opened 2023-11-28
SALTO-4726 Move custom object field fallback to FixElements

adfineberg:
Move custom object field fallback to FixElements

---

---
_Release Notes_: 
__Zendesk Adapter__
* Move the replacement of fallback user emails also for custom object fields to happen before the deploy in order to reduce incorrect user errors (partial successes).

---
_User Notifications_: 


edenhassid (approved):
LGTM! 🤩 💯

PR #5133, opened 2023-11-28
SAAS-9634 dummy adapter to support conflicts

omrifarkash:
Adding support on creating conflicts in dummy adapter. for developing and testing purposes

---

The idea:
Create 3 different versions of new elements under ConflictedStuff directory (similar to what we did with 'Extra' and 'EnvStuff')

Flow for creating conflicts of all kinds:
   - Fetch envA with configuration conflictedElementsVersion = 'a'
   - Fetch envB with configuration conflictedElementsVersion = 'b'
   - Create deployment between envA and envB, select all elements inside ConflictedStuff, calculate plan
   - Fetch envB with configuration conflictedElementsVersion = 'c'
   - Sync target in the deployment 
   
   
This flow will create all kinds of conflicts:
detailed conflicts:
- (string, number, inner object, static file) all on complicatedInst
base level conflicts:
- (modify by user while deleted in the service) on simpleInstDeletedInC
- (deleted by user while modified in the service) on simpleInstDeletedInA
- (added by user while added in the service with different values) on simpleInstDeletedInB
   

---

_Release Notes_: 

Dummy adapter support on create conflicts

---

_User Notifications_: 
None


packages/dummy-adapter/src/generator.ts
```diff
@@ -789,6 +790,100 @@ export const generateElements = async (
...
+        strField: {
+          refType: BuiltinTypes.STRING,
+        },
+      },
+      path: [DUMMY_ADAPTER, 'ConflictedStuff', 'SimpleObject'],
+      annotations: {
+        [CORE_ANNOTATIONS.ALIAS]: 'SimpleObject',
+      },
+    })
+    const simpleInstDeletedInA = new InstanceElement('simpleInstDeletedInA', simpleObject, {
```
adamro, line 852: We should also do a similar case for Fields inside Objects (additions/deletions per version)

packages/dummy-adapter/src/generator.ts
```diff
@@ -789,6 +790,100 @@ export const generateElements = async (
...
+      annotations: {
+        [CORE_ANNOTATIONS.ALIAS]: 'ComplicatedObject',
+      },
+    })
+    const complicatedInst = new InstanceElement('complicatedInst', complicatedObject, {
+      strField: version,
+      numField: version.charCodeAt(0),
+      listField: ['mockValue', `version: ${version}`],
+      staticFileField: new StaticFile({
+        content: Buffer.from(`Version is: ${version}`),
```
adamro, line 827: Do we want a more complex static file value maybe? One that is both a conflict and auto-merged in other lines?

packages/dummy-adapter/src/generator.ts
```diff
@@ -789,6 +790,100 @@ export const generateElements = async (
...
+    const simpleObject = new ObjectType({
+      elemID: new ElemID(DUMMY_ADAPTER, 'SimpleObject'),
+      fields: {
+        strField: {
+          refType: BuiltinTypes.STRING,
+        },
+      },
+      path: [DUMMY_ADAPTER, 'ConflictedStuff', 'SimpleObject'],
+      annotations: {
+        [CORE_ANNOTATIONS.ALIAS]: 'SimpleObject',
```
adamro, line 849: add `_alias` where you don't have it

adamro (approved):
Approving to save a cycle

PR #5144, opened 2023-11-30
SALTO-5100: add more logs to dummy adapter

edenhassid:
add more logs to dummy adapter

---

---
_Release Notes_: 

---
_User Notifications_: 


amitgoldarb (approved):
LGTM!

PR #5150, opened 2023-11-30
SALTO-5104 - Adapter config for custom references

tomsellek:
Enable/disable specific custom reference getters based on the adapters config

---
The flow is as follows: 
1. Pass the adapter config through the `loadWorkspace` flow all the way down to the `getCustomReferences` callback
2. The callback itself is assumed to be built by the adapter using `combineCustomReferenceGetters`, so `combineCustomReferenceGetters` now takes an additional parameter that specifies how to extract the relevant configuration from the adapter config
3. The actual function that's returned from `combineCustomReferenceGetters` will now use the provided parameters (the adapter config and the function that extracts the custom refs config from it) to retrieve the appropriate configuration values and only run the custom reference getters that are enabled.

** PR Chain **
 - https://github.com/salto-io/salto/pull/5150 👈 YOU ARE HERE 👈
 - https://github.com/salto-io/salto/pull/5045
---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A


tamtamirr (commented):
EDIT: Pretty sure initAdapters is unrelated to this flow. ignore

I think we can implement this within [initAdapters](https://github.com/tamtamirr/salto/blob/01ad2c523c02cb82385c5cf8a6316f0df355c6b6/packages/core/src/core/adapters/adapters.ts#L60) instead. This implementation feels way to deep within core. Consider the following implementation:

```typescript
const areCustomReferencesEnabled = context.config?.value?.customReferences === true ?? true
if (!areCustomReferencesEnabled) {
      log.debug('disabling custom references for account %s', account)
      creator.getCustomReferences = undefined
}
```

tomsellek:
> core has zero knowledge about the structure of the adapter config, this is a very basic decoupling in the architecture of the code. this is not something we should just change in a PR without at least having a discussion first
> 
> please **DO NOT MERGE THIS** without having that discussion first

Ack.

tomsellek:
> > core has zero knowledge about the structure of the adapter config, this is a very basic decoupling in the architecture of the code. this is not something we should just change in a PR without at least having a discussion first
> > please **DO NOT MERGE THIS** without having that discussion first
> 
> Ack.

@ori-moisis reworked the PR so that only the adapter knows about the contents of the adapter config. Can you please take another look?

packages/adapter-components/src/references/custom_references.ts
```diff
@@ -23,10 +24,18 @@ const getReferenceInfoIdentifier = (ref: ReferenceInfo): string =>
  * Combine several getCustomReferences functions into one that will run all of them.
  * When a reference is returned twice with different types the last one will override the previous ones.
  */
-export const combineCustomReferenceGetters = (customReferenceGetters: GetCustomReferencesFunc[])
-: GetCustomReferencesFunc => async elements => {
+export const combineCustomReferenceGetters = (
+  namedCustomReferenceGetters: Record<string, GetCustomReferencesFunc>,
+  getCustomRefsConfig: (adapterConfig: InstanceElement) => Record<string, boolean> = () => ({}),
```
ori-moisis, line 29: why does this need to be a callback?
feels like this is too deep of a place to be aware of the entire adapter config
is there a reason we can't get just the relevant part of the config as a parameter here?

also, is it just me or is this dead code?
seems like nothing is actually using this option...

packages/core/src/local-workspace/workspace.ts
```diff
@@ -241,17 +241,24 @@ export const getAdapterConfigsPerAccount = async (envs: EnvConfig[]): Promise<Ob
...
   accountToServiceName: Record<string, string>,
+  adaptersConfig: adaptersConfigSource.AdaptersConfigSource,
 ): Promise<ReferenceInfo[]> => {
-  const accountToElements = _.groupBy(elements.filter(e => e.elemID.adapter !== GLOBAL_ADAPTER), e => e.elemID.adapter)
-  return (await Promise.all(Object.entries(accountToElements).map(async ([account, accountElements]) => {
+  const accountElementsToRefs = async ([account, accountElements]: [string, Element[]]): Promise<ReferenceInfo[]> => {
     const serviceName = accountToServiceName[account] ?? account
     try {
-      return await adapterCreators[serviceName]?.getCustomReferences?.(accountElements) ?? []
+      const adapterConfig = await adaptersConfig.getAdapter(account)
```
ori-moisis, line 249: we should not read the adapter config if there is no custom references function

packages/core/src/local-workspace/workspace.ts
```diff
@@ -241,17 +241,24 @@ export const getAdapterConfigsPerAccount = async (envs: EnvConfig[]): Promise<Ob
...
+      const refFunc = adapterCreators[serviceName]?.getCustomReferences
+      if (adapterConfig !== undefined && refFunc !== undefined) {
```
ori-moisis, line 251: it is technically valid for an adapter to not have a configuration, this should not prevent the custom references code from running

packages/jira-adapter/test/weak_references/automation_projects.test.ts
```diff
@@ -22,6 +22,10 @@ describe('automation_projects', () => {
   let projectInstance: InstanceElement
   let instance: InstanceElement
   let elementsSource: ReadOnlyElementsSource
+  const AdapterConfigType = new ObjectType({
+    elemID: new ElemID('adapter', 'AdapterConfig'),
+  })
+  const adapterConfig = new InstanceElement('settings', AdapterConfigType)
```
ori-moisis, line 28: adapter config is a singleton, the instance name should not be `settings`
Ideally this should use the real config type, but if it not possible, lets at least make it similar to a real configuration

```suggestion
  const AdapterConfigType = new ObjectType({
    elemID: new ElemID('adapter'),
    isSettings: true
  })
  const adapterConfig = new InstanceElement(ElemID.CONFIG_NAME, AdapterConfigType)
```

packages/workspace/src/workspace/reference_indexes.ts
```diff
@@ -35,11 +35,17 @@ import { collections } from '@salto-io/lowerdash'
...
+import { AdaptersConfigSource } from './adapters_config_source'
 
 const log = logger(module)
+
+const { awu } = collections.asynciterable
+
 export const REFERENCE_INDEXES_VERSION = 4
 export const REFERENCE_INDEXES_KEY = 'reference_indexes'
 
+// const CUSTOM_REFERENCES_CONFIG = 'customReferences'
```
ori-moisis, line 47: remove commented out code

packages/workspace/src/workspace/reference_indexes.ts
```diff
@@ -273,6 +284,13 @@ const getIdToCustomReferences = async (getCustomReferences: GetCustomReferencesF
   }
 }
 
+// const areCustomReferencesEnabled = async (
+//   adaptersConfig: AdaptersConfigSource,
+//   accountName: string,
+// ): Promise<boolean> => (
+//   (await adaptersConfig.getAdapter(accountName))?.value[CUSTOM_REFERENCES_CONFIG] ?? true
+// )
```
ori-moisis, line 292: remove more commented out code

packages/workspace/src/workspace/reference_indexes.ts
```diff
@@ -302,7 +321,27 @@ export const updateReferenceIndexes = async (
...
+  const customReferences = await awu(changesForCustomRefs)
+    .reduce(async (total: CustomRefs, currentValue: {adapterName: string; adapterChanges: Change<Element>[]}) => {
+      const adapterConfig = await adaptersConfig.getAdapter(currentValue.adapterName)
+      if (adapterConfig === undefined) {
+        return total
+      }
+      const addedRefs = await getIdToCustomReferencesForAdapter(getCustomReferences, currentValue.adapterChanges,
+        adapterConfig)
+      return _.merge(total, addedRefs)
+    }, { before: {}, after: {} })
```
ori-moisis, line 344: why is this here?
it seems the code in core's `getCustomReferences` ([here](https://github.com/salto-io/salto/pull/5150/files#diff-b6837d9d607b68ba50569e242180effc3d590e0023dd0082c73a63563be0e0bdR241)) expects to get the adapter config source and handles reading the relevant configs on its own, we shouldn't have to do this twice

packages/workspace/src/workspace/workspace.ts
```diff
@@ -758,7 +759,8 @@ export const loadWorkspace = async (
         stateToBuild.states[envName].mapVersions,
         stateToBuild.states[envName].merged,
         changeResult.cacheValid,
-        element => getCustomReferences(element, currentEnvConf().accountToServiceName ?? {}),
+        element => getCustomReferences(element, currentEnvConf().accountToServiceName ?? {}, adaptersConfig),
```
ori-moisis, line 762: nit, because this is confusing when it is actually an array
```suggestion
        elements => getCustomReferences(elements, currentEnvConf().accountToServiceName ?? {}, adaptersConfig),
```

packages/workspace/src/workspace/reference_indexes.ts
```diff
@@ -281,6 +299,7 @@ export const updateReferenceIndexes = async (
   elementsSource: ElementsSource,
   isCacheValid: boolean,
   getCustomReferences: GetCustomReferencesFunc,
+  adaptersConfig: AdaptersConfigSource,
```
ori-moisis, line 302: the extra parameter is not needed here and the type of the function is incorrect
the function being passed into here from [here](https://github.com/salto-io/salto/pull/5150/files#diff-3e73f2742b5a2b53ef4f7b71696009c5566b930ed6be759f13151ffd3650a771R762) still has the same signature as it did before this PR - it only gets the elements

packages/workspace/test/workspace/reference_indexes.test.ts
```diff
@@ -524,6 +534,11 @@ describe('updateReferenceIndexes', () => {
           val3: 'string',
         }
       )
+      const AdapterConfigType = new ObjectType({
+        elemID: new ElemID('adapter', 'AdapterConfig'),
+      })
+      const adapterConfig = new InstanceElement('settings', AdapterConfigType)
```
ori-moisis, line 540: same comment as the other test

also, doesn't seem like anything is actually testing that this works, looks like we set the adapter config, but we don't actually check that it does something

packages/adapter-components/test/references/custom_references.test.ts
```diff
@@ -13,13 +13,13 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
-import { ElemID } from '@salto-io/adapter-api'
+import { ElemID, InstanceElement, ObjectType } from '@salto-io/adapter-api'
 import { combineCustomReferenceGetters } from '../../src/references/custom_references'
 
 describe('combineCustomReferenceGetters', () => {
   it('should run all the custom reference getters', async () => {
```
ori-moisis, line 20: what about tests for when it shouldn't run all of them?
I didn't go over all the test scenarios, but it seems like this PR is generally under-tested

packages/adapter-components/src/references/custom_references.ts
tomsellek: > why does this need to be a callback?

So - `combineCustomReferenceGetters` returns something that can be used as a `getCustomReferences` callback. And the core code that calls `getCustomReferences` only has the entire adapter config as an `InstanceElement`. So the returned function must accept an `InstanceElement`, no?

> also, is it just me or is this dead code?

Yes. I figured we can add actual handling of the config in separate per-adapter PRs to keep the size of this PR manageable.

packages/core/src/local-workspace/workspace.ts
tomsellek: So is there a way for us to tell the difference between an adapter with a config that couldn't be parsed and an adapter with no config? Or should we just run all custom refs getters if the config is invalid?

packages/workspace/src/workspace/reference_indexes.ts
```diff
@@ -273,14 +276,15 @@ const getIdToCustomReferences = async (getCustomReferences: GetCustomReferencesF
...
+
 export const updateReferenceIndexes = async (
   changes: Change<Element>[],
   referenceTargetsIndex: RemoteMap<ReferenceTargetIndexValue>,
   referenceSourcesIndex: RemoteMap<ElemID[]>,
   mapVersions: RemoteMap<number>,
   elementsSource: ElementsSource,
   isCacheValid: boolean,
-  getCustomReferences: GetCustomReferencesFunc,
+  getCustomReferences: (elements: Element[]) => Promise<ReferenceInfo[]>,
```
tomsellek, line 287: This can no longer be a `GetCustomReferencesFunc` because:
1. `GetCustomReferencesFunc` now receives an adapter config, which function doesn't need to take
2. There's was no real reason to couple this type and the type of the adapter callback in the first place? I think?

packages/adapter-components/src/references/custom_references.ts
ori-moisis: is there really a big difference in the amount of code here if we also include the usage of the configuration?
if there is, it could be a signal that we are doing something wrong with the interface here

ori-moisis:
I see the request for review, but it doesn't seem like this PR is actually done so not sure what to do with this one
let me know when the PR is actually ready for review

packages/adapter-components/src/references/custom_references.ts
tomsellek: No big difference, but I would need to add a new config to two different adapters I don't usually deal with.

packages/workspace/test/workspace/reference_indexes.test.ts
tomsellek: With the latest implementation (where all we do is pass the config to the callback) we don't really expect `updateReferenceIndexes` to do much, no?

tomsellek:
> I see the request for review, but it doesn't seem like this PR is actually done so not sure what to do with this one let me know when the PR is actually ready for review

@ori-moisis should be ready now

packages/adapter-components/test/references/custom_references.test.ts
```diff
@@ -13,46 +13,155 @@
...
+import { ElemID, GetCustomReferencesFunc, InstanceElement, ObjectType } from '@salto-io/adapter-api'
 import { combineCustomReferenceGetters } from '../../src/references/custom_references'
 
 describe('combineCustomReferenceGetters', () => {
-  it('should run all the custom reference getters', async () => {
-    const getCustomReferencesFunc = combineCustomReferenceGetters([
-      async () => ([
+  const adapterConfigType = new ObjectType({
+    elemID: new ElemID('adapter', 'AdapterConfig'),
+  })
```
ori-moisis, line 22:
```suggestion
  const adapterConfigType = new ObjectType({
    elemID: new ElemID('adapter'),
  })
```

packages/workspace/src/workspace/reference_indexes.ts
```diff
@@ -257,14 +258,16 @@ const updateReferenceSourcesIndex = async (
   ])
 }
 
-const getIdToCustomReferences = async (getCustomReferences: GetCustomReferencesFunc, changes: Change<Element>[])
-: Promise<{ before: Record<string, ReferenceInfo[]>; after: Record<string, ReferenceInfo[]> }> => {
+const getIdToCustomReferencesForAdapter = async (
+  getCustomReferences: (elements: Element[]) => Promise<ReferenceInfo[]>,
```
ori-moisis, line 262: nit - define this function type once instead of repeating it in two functions

packages/workspace/src/workspace/reference_indexes.ts
```diff
@@ -302,7 +306,10 @@ export const updateReferenceIndexes = async (
     initialIndex = true
   }
 
-  const customReferences = await getIdToCustomReferences(getCustomReferences, changes)
+  const customReferences = await getIdToCustomReferencesForAdapter(
```
ori-moisis, line 309: nit - why did you rename the function? I think the new name isn't actually better...

packages/workspace/test/workspace/reference_indexes.test.ts
```diff
@@ -524,6 +528,12 @@ describe('updateReferenceIndexes', () => {
           val3: 'string',
         }
       )
+      const AdapterConfigType = new ObjectType({
+        elemID: new ElemID('adapter'),
+        isSettings: true,
+      })
+      const adapterConfig = new InstanceElement(ElemID.CONFIG_NAME, AdapterConfigType)
+      await adaptersConfig.setAdapter('test', 'test', adapterConfig)
```
ori-moisis, line 536: is this used anywhere that I am missing?

ori-moisis (approved):

PR #5157, opened 2023-12-04
SAAS 9634 dummy generate extra elements to support real file path

omrifarkash:
- Make generate extra elements in dummy adapter, support real static file content by putting real files inside the same folder where the nacl.mock exists

- Reverting old api of `conflictedElementsVersion` on dummy adapter

- Make generate extra elements to support more than a single path input

---

---

_Release Notes_: 
None

---

_User Notifications_: 
None


packages/dummy-adapter/src/generator.ts
```diff
@@ -666,20 +664,29 @@ export const generateElements = async (
...
         file: {
-          parse: async funcParams => new StaticFile({
-            content: Buffer.from('THIS IS STATIC FILE'),
-            filepath: funcParams[0],
-          }),
+          parse: async funcParams => {
+            const [filepath] = funcParams
+            let fileContent: Buffer
+            try {
+              fileContent = fs.readFileSync(`${file.fullPath.replace(file.basename, '')}${filepath}`)
```
ori-moisis, line 681: nit: this is an async function anyway, why `readFileSync` and not just `readFile`?

ori-moisis (approved):

packages/dummy-adapter/src/generator.ts
omrifarkash: No specific reason, will change it

omrifarkash: Just because on line 673 we're using this function to read content since forever :)

PR #5158, opened 2023-12-04
SALTO-5071 Support axios timeouts, with lastRetryNoTimeout

adfineberg:
In order to cap very long requests, we want to add an option to timeout Axios requests. 
In addition, because we don't want to lose these requests, we want the last retry for each request lifecycle to have no timeout.

---

---
_Release Notes_: 

---
_User Notifications_: 


packages/adapter-components/src/client/http_client.ts
```diff
@@ -115,6 +118,7 @@ export abstract class AdapterHTTPClient<
     this.conn = createClientConnection({
       connection,
       retryOptions: createRetryOptions(_.defaults({}, this.config?.retry, defaults.retry)),
+      timeout: this.config?.timeout || defaults.timeout || NO_TIMEOUT,
```
netama, line 121: I think you might want `??` and not `||` here, so that even if the default has a defined timeout users will be able to override to no timeout? alternatively you can set `NO_TIMEOUT` to `-1` (like we do in `RATE_LIMIT_UNLIMITED_MAX_CONCURRENT_REQUESTS`)

packages/adapter-components/src/client/config.ts
```diff
@@ -32,13 +32,16 @@ export type ClientRetryConfig = Partial<{
   retryDelay: number
   // This is not included in clientRetryConfigType because currently we don't want to allow the user to change it
   additionalStatusCodesToRetry: number[]
+  shouldResetTimeout: boolean
```
netama, line 35: can you please add an explanation about what this does / rename to something like `resetTimeoutBetweenCalls`?

packages/adapter-components/src/client/config.ts
```diff
@@ -32,13 +32,16 @@ export type ClientRetryConfig = Partial<{
...
+  lastRetryNoTimeout: boolean
 }>
 
 export type ClientBaseConfig<RateLimitConfig extends ClientRateLimitConfig> = Partial<{
   retry: ClientRetryConfig
   rateLimit: RateLimitConfig
   maxRequestsPerMinute: number
   pageSize: ClientPageSizeConfig
+  timeout: number
```
netama, line 44: 1. nit: I'm wondering if it might fit better to have all 3 args under a `timeout` object here? I agree that the two booleans also fit under `retry`, but these 3 args work together (I was looking for the actual number next to the booleans)
2. I think at least the timeout should also be exposed to users? if so, please document it in the relevant adapters' `config_doc.md` (and add in the relevant object types - `createClientConfigType`)

netama (commented):
looks good :) added some minor comments

packages/adapter-components/src/client/http_client.ts
adfineberg: According to Axios, when the timeout is 0, there is no timeout. If this doesn't answer, then I don't understand what you meant

netama: I meant - if `defaults.timeout` is non-zero, even if you set `this.config?.timeout` to 0 you won't be able to override it. because `0 || 100` is 100

packages/adapter-components/src/client/config.ts
netama: 1. I agree, I was suggesting moving the other two to be under a shared `timeout` object here (instead of under `retry`)
2. I think that will change the zendesk default, but after this PR anyone will be able to define this in the config for all adapters, no?

packages/adapter-components/src/client/config.ts
```diff
@@ -41,7 +46,7 @@ export type ClientBaseConfig<RateLimitConfig extends ClientRateLimitConfig> = Pa
   rateLimit: RateLimitConfig
   maxRequestsPerMinute: number
   pageSize: ClientPageSizeConfig
-  timeout: number
+  timeoutOptions: ClientTimeoutOptions
```
netama, line 49: nit: I think all of the other ones are also options (they're inside the adapter config), can we keep this as `timeout` (and maybe call the inner value `maxDuration` or something like that)?

packages/adapter-components/src/client/config.ts
```diff
@@ -32,6 +33,10 @@ export type ClientRetryConfig = Partial<{
   retryDelay: number
   // This is not included in clientRetryConfigType because currently we don't want to allow the user to change it
   additionalStatusCodesToRetry: number[]
+}>
+
+export type ClientTimeoutOptions = Partial<{
+  timeout: number
```
netama, line 39: nit: is there a point to having this object if we don't set `timeout`? maybe make it required and the others optional (and then do the same for the `clientTimeoutOptionsType` below)?

packages/adapter-components/test/client/config.test.ts
```diff
@@ -20,7 +20,7 @@ describe('client_config', () => {
   describe('createClientConfigType', () => {
     it('should return default type when no custom buckets were added', async () => {
       const type = createClientConfigType('myAdapter')
-      expect(Object.keys(type.fields)).toHaveLength(4)
+      expect(Object.keys(type.fields)).toHaveLength(5)
```
netama, line 23: can we also test the value of the new field (here and below)?

netama (approved):
added a few nits but LGTM otherwise so approving :) (might still be worth updating the config docs? can also leave for the next one if you want, just please update all adapters and not only zendesk, just write the default value as unlimited)

packages/adapter-components/src/client/config.ts
adfineberg: I think it's a bit more complicated than that - the only constraint is that if one of the others is set, then we require the timeout. But the timeout itself is not really required as part of the config.
Not sure it's worth writing out that specific logic...

packages/adapter-components/src/client/http_connection.ts
```diff
@@ -130,21 +129,32 @@ export const createRetryOptions = (retryOptions: Required<ClientRetryConfig>): R
   },
   retryCondition: err => axiosRetry.isNetworkOrIdempotentRequestError(err)
     || shouldRetryStatusCode(err.response?.status, retryOptions.additionalStatusCodesToRetry),
+  onRetry: (retryCount, _err, requestConfig) => {
+    if (timeoutOptions?.lastRetryNoTimeout && retryCount === retryOptions.maxAttempts) {
+      requestConfig.timeout = 0
```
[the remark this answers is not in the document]
adfineberg: So 2 things:
* The `requestConfig` is carried through the different retries. In the current implementation that's not a problem - because it's changed only in the last retry.
* The config doesn't transfer between requests. So the next GET/POST/Whatever will have a regular config.

PR #5172, opened 2023-12-07
SALTO-5123-Fix forms elemID

idoamit8:
In this PR I fixed forms elemID and modified the e2e 
---

None

---
_Release Notes_: 
_Jira Adpater:_
Change forms elemID

---
_User Notifications_: 
_Jira Adapter:_
To change the current elemID, customers must perform a fetch --regenerate command. This will change the forms elemID, which may cause some disruptions


shir-reifenberg (approved):
👌 
suggestion - maybe we should use config (and specifically `idFields`) in the generation o the form's name so it can be adjusted without code changes

PR #5173, opened 2023-12-07
SALTO-5108 Add resolveUserIDs option, continue with no users

adfineberg:
Enable fetching without users, either defined in config, or if the query is forbidden. In both cases, complete the fetch (if forbidden, then with errors).

---

---
_Release Notes_: 
__Zendesk adapter__
* Allow fetch without users, either by configuration of `resolveUserIDs = false`, or if the users query is forbidden

---
_User Notifications_: 


packages/zendesk-adapter/src/fix_elements/fallback_user.ts
```diff
@@ -104,6 +104,18 @@ const replaceMissingUsers = (
   return { fixedInstance, missingUsers: _.uniq(missingUserPaths.map(({ user }) => user)) }
 }
 
+const noRelevantUsers = (
```
adfineberg, line 107: this is yucky, but I couldn't make it less yucky - any suggestions?

edenhassid (approved):
LGTM! 🔥 💪

PR #5175, opened 2023-12-07
SALTO-4744: add important values to all object types in dummy

edenhassid:
add important values to all object types in dummy

---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/dummy-adapter/src/generator.ts
+++ b/packages/dummy-adapter/src/generator.ts
@@ -165,7 +165,7 @@ export type GeneratorParams = {
     generateEnvName? : string
     fieldsToOmitOnDeploy?: string[]
     elementsToExclude?: string[]
-    importantValuesFreq: number
+    importantValuesFreq?: number
 }
 
 export const defaultParams: Omit<GeneratorParams, 'extraNaclPaths'> = {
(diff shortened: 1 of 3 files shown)


packages/dummy-adapter/src/generator.ts
```diff
@@ -387,6 +389,30 @@ export const generateElements = async (
   const generateFileContent = (): Buffer => Buffer.from(getMultiLine(
     normalRandom(params.staticFileLinesMean, params.staticFileLinesStd)
   ))
+
+  const generateImportantValues = (fieldNames: string[]): ImportantValues | undefined => {
+    // the  important values should be only a small portion of the fields
+    const halfLength = Math.floor((fieldNames.length / 2) + 1)
```
netama, line 395: I think it might be good to control this in a config arg (e.g. `importantValuesFreq` to match the current pattern), so that it can be turned off if someone doesn't want to have them?
(also - maybe cover this in one of the unittests? if it's easy)

packages/dummy-adapter/src/generator.ts
```diff
@@ -387,6 +389,30 @@ export const generateElements = async (
...
+    const numberOfImportantValues = Math.floor(randomGen() * halfLength)
+    const fieldSet = new Set<string>()
+    const importantValuesDef = Array.from({ length: numberOfImportantValues }).map(() => {
+      const value = weightedRandomSelect(fieldNames)
+      if (fieldSet.has(value) || value.startsWith('_')) {
```
netama, line 400: nit: can filter out the `_` prefix upfront?
(would also have been nice if we had `weightedRandomPick` so you wouldn't have to deal with collisions, but I don't think it's worth the extra work now)

netama (approved):
LGTM with a couple of nit suggestions

netama (approved):

PR #5182, opened 2023-12-10
SALTO-5131: Salesforce Tooling extra dependencies V2

tamtamirr:
Added an alternate implementation of the **extra_dependencies** filter with better coverage, as the old implementation reached the limit of max tooling records using Bulk API v1 (2000).

---

Added this under the optional feature `extraDependenciesV2` that is turned off by default.

---
_Release Notes_: 
Salesforce Adapter:
- Improved the coverage of resolving extra dependencies between Metadata components. This is a beta feature that requires the optional feature `extraDependenciesV2`

---
_User Notifications_: 
Salesforce:
- Using the beta feature `extraDependenciesV2`, may result in `_extra_dependencies` changes on existing nacls.


packages/salesforce-adapter/src/filters/extra_dependencies.ts
```diff
@@ -203,11 +305,12 @@ const addExtraReferences = async (
...
     })
 
-    addGeneratedDependencies(
+    return addGeneratedDependencies(
       elem,
       dependencies.map(item => item.elemId).filter(isDefined),
     )
-  })
+  }).toArray().then(references => references.flat())
+  log.debug('Added %d extra dependencies', addedDeps.length)
```
tamtamirr, line 314: Added this log to better understand how many total generated dependencies this filter created. This may help us track performance issues on larger accounts. This is unrelated to this PR.

packages/salesforce-adapter/src/filters/extra_dependencies.ts
```diff
@@ -118,16 +144,92 @@ const getDependencies = async (
   }))
 }
 
+const resolveDeps = async ({ client, elements }: GetDependenciesParams): Promise<Dependency[]> => {
+  const elementsByMetadataComponentId = _.keyBy(elements.filter(hasInternalId), getInternalId)
+  // The MetadataComponentId of standard objects is the object name
+  elements.filter(isStandardObjectSync).forEach(standardObject => {
```
tamtamirr, line 150: That's how the API behaves since Standard Objects don't have an internal id. See [this doc ](https://developer.salesforce.com/docs/atlas.en-us.api_tooling.meta/api_tooling/tooling_api_objects_metadatacomponentdependency.htm#:~:text=Details-,MetadataComponentId,-Type)for more context.

tamtamirr (commented):

packages/salesforce-adapter/src/filters/extra_dependencies.ts
```diff
@@ -118,16 +144,92 @@ const getDependencies = async (
...
+          to: {
+            type: rec.RefMetadataComponentType,
+            id: rec.RefMetadataComponentId,
+            name: rec.RefMetadataComponentName,
+          },
+        }))
+      })).then(res => res.flat())
+  }
+  return chunkedQuery(Object.keys(elementsByMetadataComponentId), INITIAL_QUERY_CHUNK_SIZE)
+    .then(deps => {
```
amitgoldarb, line 198: Why 'then'? I think this might be a bit more readable with an await.

packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -601,11 +601,19 @@ export const ensureSafeFilterFetch = ({
...
+ */
 export const isStandardObject = async (objectType: ObjectType): Promise<boolean> => (
   await isCustomObject(objectType)
   && !ENDS_WITH_CUSTOM_SUFFIX_REGEX.test(await safeApiName(objectType) ?? '')
 )
 
+export const isStandardObjectSync = (element: Element): element is ObjectType => (
+  isCustomObjectSync(element)
+  && !ENDS_WITH_CUSTOM_SUFFIX_REGEX.test(apiNameSync(element) ?? '')
+)
```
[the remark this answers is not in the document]
tamtamirr: Please note that `isStandardObject` is marked as deprecated.
Also there's no logic to share as the implementation of `isStandardObject` uses `isCustomObject` and not `isCustomObjectSync`. Nevertheless I've planned adding UTs to `isStandardObjectSync` 🙏

packages/salesforce-adapter/src/filters/extra_dependencies.ts
```diff
@@ -118,16 +144,92 @@ const getDependencies = async (
   }))
 }
 
+const resolveDeps = async ({ client, elements }: GetDependenciesParams): Promise<Dependency[]> => {
```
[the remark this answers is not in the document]
tamtamirr: Been thinking about it yesterday on my way home, I'll rename to `queryDeps` as we simply query `MetadataComponentDependency` records here.

packages/salesforce-adapter/src/filters/extra_dependencies.ts
```diff
@@ -118,16 +144,92 @@ const getDependencies = async (
...
+    log.debug('No elements to query dependencies for')
+    return []
+  }
+  let queriesExecuted = 0
+  const errorIds = new Set<string>()
+  const chunkedQuery = async (ids: string[], chunkSize: number): Promise<Dependency[]> => {
+    if (chunkSize === 1) {
+      errorIds.add(ids[0])
+      return []
+    }
```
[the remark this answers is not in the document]
tamtamirr: This code is actually unreachable and needs to be cleaned-up as I'm validating the `chunkSize` before reaching here in my current implementation. Will remove

packages/salesforce-adapter/src/filters/extra_dependencies.ts
tamtamirr: IMO if a method returns a Promise it's better to avoid unnecessary awaits and let the caller await the promise where used. Also in this specific scenario, I prefer using `then` as it gives a sense of scope as this is a callback to the operation. This is personal preference, I don't have a strong opinion here though

amitgoldarb: I think using await and .then should be exactly the same computational-wise. They are both placed on the event-loop exactly at the same time. The only difference is the syntax.

packages/salesforce-adapter/src/filters/extra_dependencies.ts
```diff
@@ -203,11 +303,12 @@ const addExtraReferences = async (
         elem.elemID.getFullName()}`)
     })
 
-    addGeneratedDependencies(
+    return addGeneratedDependencies(
       elem,
       dependencies.map(item => item.elemId).filter(isDefined),
     )
-  })
+  }).toArray().then(references => references.flat())
```
amitgoldarb, line 310: Again why here then? Why not await?

PR #5184, opened 2023-12-11
SALTO-5126: Fix parsing error In form maps

idoamit8:
This PR is following this incident - [Parsing error when applying addition detailed change to a list in a map](https://salto-io.atlassian.net/browse/SALTO-5115?focusedCommentId=114806)
1. Fetch - change each numeric key to nackCase. 
2. deploy - cahgne each naclCase key to numeric. 

---

The bug as reproduce with Jira account and got fixed
image

---
_Release Notes_: 
_Jira Adapter:_
There will be noise since the names of the keys are going to change. 

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/forms/forms.ts
+++ b/packages/jira-adapter/src/filters/forms/forms.ts
@@ -16,7 +16,7 @@
 
 import { CORE_ANNOTATIONS, Change, InstanceElement, ReferenceExpression, getChangeData, isAdditionChange, isAdditionOrModificationChange, isInstanceChange, isInstanceElement } from '@salto-io/adapter-api'
 import { values as lowerDashValues } from '@salto-io/lowerdash'
-import { WALK_NEXT_STEP, getParent, invertNaclCase, naclCase, pathNaclCase, walkOnElement } from '@salto-io/adapter-utils'
+import { getParent, invertNaclCase, mapKeysRecursive, naclCase, pathNaclCase } from '@salto-io/adapter-utils'
 import _ from 'lodash'
 import { FilterCreator } from '../../filter'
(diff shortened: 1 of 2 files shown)


omrilit (commented):
Looks great!
I think that the Release Notes are not indicative enough on the change that is about to occur, please rephrase 🙏 
This requires noise reduction, and consider also external communication, let's f2f it and decide.

packages/jira-adapter/src/filters/forms/forms.ts
```diff
@@ -120,7 +143,19 @@ const filter: FilterCreator = ({ config, client, fetchQuery }) => ({
...
-    forms.forEach(form => elements.push(form))
+    forms.forEach(form => {
+      fixFormsKeysName({
+        instance: form,
+        transformKeyFunc: ((key, value) => {
+          if (naclCase(key) !== key) {
+            const newKey = naclCase(key)
+            value[newKey] = value[key]
+            delete value[key]
+          }
```
omrilit, line 154: I think that it's better to use the calculated value and not recalculate it.
Same goes for the `invertNaclCase`
```suggestion
         const naclCaseKey = naclCase(key)
          if (naclCaseKey !== key) {
            value[naclCaseKey] = value[key]
            delete value[key]
          }
```

omrilit (approved):

ori-moisis:
@omrilit, wrote a utility function called `mapKeysRecursive` that does kind of exactly this, might reduce the amount of code here.
could also make it simpler since it would be just
```typescript
mapKeysRecursive(instance.value, ({ key }) => naclCase(key))

mapKeysRecursive(instance.value, ({ key }) => invertNaclCase(key))
```

PR #5187, opened 2023-12-12
SALTO-5088: Deploy Artifacts

tamtamirr:
This PR adds the capability to add artifacts upon deploy.

---

Will add UTs once the design is approved.

- Added the deploy extra property `artifacts`.
- Added the deploy CLI option `artifactsDir` that when specified will write the artifacts to.
- We write artifacts for checkOnly deployments as-well.

Usage example:
`salto deploy -a /tmp/deploy-artifacts`

The artifact(s) path is:
`${artifactsDir}/${accountName}/${artifact.name}`

---
_Release Notes_: 
Core:
- Added the optional capability to write deploy artifacts.

---
_User Notifications_: 
_None_


packages/adapter-api/src/adapter.ts
```diff
@@ -46,8 +46,15 @@ export type Group = {
...
 
+export type Artifact = {
+  name: string
+  content: Buffer
+  account?: string
+}
+
 export type DeployExtraProperties = {
   groups?: Group[]
+  artifacts? : Artifact[]
```
adamro, line 57: Probably discussed in a different forum, but how come the artifacts are not under a group? With this implementation, is there relation between the other of the groups and the artifacts? Is there any relation between them?

tamtamirr: The reason I didn't want to put the artifacts under the **groups** is because the current SAAS implementation puts `extraProperties.groups` directly in the DB. I think it's cleaner & safer this way.
If you have a strong opinion that it should sit under the group, we will need to modify the code in the BE to omit `extraProperties.groups.artifacts`, which does not exist on the type right now 🤔

packages/salesforce-adapter/src/constants.ts
```diff
@@ -541,3 +541,9 @@ export const isSalesforceError = (error: Error): error is SalesforceError => {
   const errorCode = _.get(error, ERROR_PROPERTIES.ERROR_CODE)
   return _.isString(errorCode) && (Object.values(SALESFORCE_ERRORS) as ReadonlyArray<string>).includes(errorCode)
 }
+
+
+// Artifacts
+export const SalesforceArtifacts = {
```
tamtamirr, line 547: We probably want to expose this type so we can later use it in the SAAS query. WDYT?

amitgoldarb: It makes sense IMO.

packages/core/src/core/plan/plan_item.ts
```diff
@@ -42,11 +43,16 @@ const getGroupAction = (group: Group<Change>): ActionName => {
     : 'modify'
 }
 
+const getGroupAccount = (group: Group<Change>): string => (
+  getChangeData(wu(group.items.values()).toArray()[0]).elemID.adapter
```
amitgoldarb, line 47: Is this what @tomsellek described where the adapter is the accountName?

amitgoldarb (approved):
This PR looks good to me. But there were a few iterations of planning after I was involved. I think this would probably benifit from @ori-moisis 's approval as well.

packages/adapter-api/src/adapter.ts
adamro: I am not sure SaaS and what we save in the DB in the SaaS considerations should lead our api.
What logically makes more sense in your opinion, under groups or not? What's the connections between these? (is there a chance we eventually write code that matches the group and it's artifact? cause we can think of a reason we will I think it should be inside)

tamtamirr:
I'm converting to DRAFT after discussing this with @ori-moisis yesterday. Will ping to re-iterate when I finish.
cc. @adamro @amitgoldarb

packages/adapter-api/src/adapter.ts
tamtamirr: Agreed, converted the PR back to draft as I'm changing the interface 🙏

packages/core/src/core/plan/plan_item.ts
tamtamirr: yes

packages/cli/src/commands/deploy.ts
```diff
@@ -191,6 +195,28 @@ const deployPlan = async (
...
+  group.artifacts !== undefined && group.accountName !== undefined
+)
+
+const writeArtifacts = (
+  groups: types.NonEmptyArray<GroupWithArtifacts>,
+  artifactsDir: string,
+): void => {
+  log.debug('writing deploy artifacts to %s', artifactsDir)
+  groups.forEach(({ accountName, artifacts }) => {
+    const artifactDir = `${artifactsDir}/${accountName ?? 'unknown'}`
```
dantal4, line 210: isn't `accountName` a required string here?
```suggestion
    const artifactDir = `${artifactsDir}/${accountName}`
```

packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -69,7 +80,7 @@ const deployAction = async (
   checkOnly: boolean
 ): Promise<DeployResult> => {
   const changes = [...planItem.changes()]
-  const adapterName = getChangeData(changes[0]).elemID.adapter
+  const adapterName = planItem.account
   const adapter = adapters[adapterName]
```
dantal4, line 84: isn't a different account name will make `adapter` to be undefined?

packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -137,9 +148,9 @@ export const deployActions = async (
       try {
         const result = await deployAction(item, adapters, checkOnly)
         result.appliedChanges.forEach(appliedChange => appliedChanges.push(appliedChange))
-        if (result.extraProperties?.groups !== undefined) {
-          groups.push(...result.extraProperties.groups)
-        }
+        makeArray(result.extraProperties?.groups)
+          .map(group => Object.assign(group, { accountName: item.account, id: item.groupKey }))
+          .map(group => allGroups.push(group))
```
dantal4, line 153:
```suggestion
        makeArray(result.extraProperties?.groups).forEach(group => {
          allGroups.push({ ...group, accountName: item.account, id: item.groupKey })
        })
```

packages/salesforce-adapter/src/constants.ts
dantal4: shouldn't we avoid exposing adapter-specific functions/constants/types in the adapters packages? @netama @ori-moisis

dantal4 (commented):
Looks great!

packages/adapter-api/src/adapter.ts
```diff
@@ -39,11 +39,18 @@ export interface FetchResult {
...
 }
 
+export type Artifact = {
+  name: string
+  content: Buffer
+}
+
 export type Group = {
   id?: string
+  accountName?: string
```
ori-moisis, line 49: IIUC `accountName` is filled by `core` and should never actually be `undefined`, no?
same with `id`

packages/cli/src/commands/deploy.ts
```diff
@@ -191,6 +195,28 @@ const deployPlan = async (
...
+    fs.mkdirSync(artifactDir, { recursive: true })
+    artifacts.forEach(artifact => {
+      const artifactPath = `${artifactDir}/${artifact.name}`
+      fs.writeFileSync(artifactPath, artifact.content)
```
ori-moisis, line 214: notice that none of the code in core / cli uses `fs` directly - it goes through `@salto-io/file`

also, why is this `sync`? we should generally not use `sync` file operations

packages/core/src/core/plan/plan_item.ts
ori-moisis: doesn't this kind of conflict with what we are doing in #5196?
it seems like in both places there is a need to add a property of "which account does this belong to" to change groups and it is done in different ways in each PR.
I think we should not have two different ways to implement this...

packages/core/src/core/deploy/deploy_actions.ts
ori-moisis: +1 on the suggestion, mainly because the current code changes the value of `group`

packages/salesforce-adapter/src/constants.ts
ori-moisis: if this is information which is unique to salesforce, then the salesforce adapter is the only place that can expose it, no?
I am not sure if there is a real need to expose this information at all (personally I don't think there is such a need), but if @tamtamirr and @amitgoldarb think they need to expose this, I think they will need to maintain it as part of the specific adapter's interface since this is definitely not a part of the core interface....

tamtamirr: @dantal4 @ori-moisis I don't see any reason not to expose it from the adapter TBH.
If the adapter yields such artifacts, it should be exposed and I'd expect any other adapter to do the same. Can't be exposed from core.

packages/adapter-api/src/adapter.ts
tamtamirr: Yes, the biggest issue with the current implementation is the fact CoreDeployResult and AdapterDeployResult are of the same type `DeployResult` sadly

Until we have separate types, this must be undefined since the adapter shouldn't fill this values. Not great, but attempting to separate these types resulted in a huge break... WDYT? @ori-moisis

packages/core/src/core/deploy/deploy_actions.ts
tamtamirr: Not a fan of using spread, especially when dealing with possibly big objects (e.g. artifacts buffers)
I'll use your suggestion with `allGroups.push(Object.assign({}, group, { accountName: item.account, id: item.groupKey })` to not modify the original group from the adapter

packages/core/src/core/deploy/deploy_actions.ts
tamtamirr: This one is confusing so I might rename the old code. The **adapters** record keys is actually the **accountNames**. So it's basically a record between **accountName** and **adapter**

packages/cli/src/commands/deploy.ts
tamtamirr: While it never should be undefined here, this is a tech debt due to the fact the adapter & core share the `DeployResult` interface. See my other comment on the interface itself 🙏

packages/core/src/core/plan/plan_item.ts
tamtamirr: IMO the fact core groupId contains the accountName doesn't mean the `accountName` shouldn't be part of the group interface, as this is only implementation detail 🤔 If you have a strong opinion against it, we'd need to expose a utility that gets the accountName from the `groupId` which I think kinda smells TBH.

packages/cli/src/commands/deploy.ts
tamtamirr: My bad, literally forgot I added this typeguard there 😆 👍

packages/core/src/types.ts
tamtamirr, line 1: I had to separate some types to fix dependency cycles

packages/cli/src/commands/deploy.ts
```diff
@@ -160,7 +164,7 @@ const deployPlan = async (
   }
   const executingDeploy = (force || await shouldDeploy(actionPlan, checkOnly))
   await printStartDeploy(output, executingDeploy, checkOnly)
-  const result = executingDeploy
+  const result: DeployResult = executingDeploy
```
ori-moisis, line 167: is this type not inferred?

packages/cli/src/commands/deploy.ts
```diff
@@ -224,7 +261,7 @@ export const action: WorkspaceCommandAction<DeployArgs> = async ({
   const actionPlan = await preview(workspace, actualAccounts, checkOnly)
   await printPlan(actionPlan, output, workspace, detailedPlan)
 
-  const result = dryRun ? { success: true, errors: [] } : await deployPlan(
+  const result: DeployResult = dryRun ? { success: true, errors: [] } : await deployPlan(
```
ori-moisis, line 264: same as before nit - is the type not inferred?

packages/cli/test/commands/deploy.test.ts
```diff
@@ -99,6 +108,34 @@ describe('deploy command', () => {
     })
   })
 
+  describe('when artifactsDir param is provided', () => {
```
ori-moisis, line 111: there are more conditions that the code is supposed to handle and are not tested (e.g - have artifacts but not dir, have dir but no artifacts, artifacts from different accounts, etc...)

packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -164,7 +169,7 @@ export const deployActions = async (
         throw error
       }
     })
-    return { errors: [], appliedChanges, extraProperties: { groups } }
+    return { errors: [], appliedChanges, extraProperties: { groups: allGroups } }
```
ori-moisis, line 172: nit - renaming the variable from `groups` to `allGroups` I think makes the code a bit more cumbersome and doesn't really help with readability (the word `all` doesn't really add information here AFAICT)

packages/salesforce-adapter/src/constants.ts
ori-moisis: I think this isn't actually helpful and personally I don't like committing to external APIs when they are not needed, but this is up to you if you want to maintain this as part of the adapter interface you can

packages/salesforce-adapter/src/transformers/xml_transformer.ts
```diff
@@ -413,6 +413,7 @@ export type DeployPackage = {
   addToManifest(type: MetadataObjectType, name: string): void
   delete(type: MetadataObjectType, name: string): void
   getZip(): Promise<Buffer>
+  getPackageXmlContent(): Buffer
```
ori-moisis, line 416: kind of a nit, but I don't think this should return a Buffer
I understand why you'd want an `artifact` to be a buffer, but here this is a getter for the content of a textual format, there is no reason for it to be a buffer...
I think this should return a string and the conversion of the string to a buffer should be done by the code that decides to put this string in an artifact

packages/cli/src/commands/deploy.ts
```diff
@@ -191,6 +195,39 @@ const deployPlan = async (
...
+      return
+    }
+    const artifactsByAccountName = _(groupsWithArtifacts)
+      .groupBy(group => group.accountName)
+      .mapValues(groups => groups.flatMap(group => group.artifacts))
+      .value()
+    await awu(Object.entries(artifactsByAccountName)).forEach(async ([accountName, artifacts]) => {
+      const artifactDir = `${artifactsDir}/${accountName}`
+      await mkdirp(artifactDir)
+      await awu(artifacts).forEach(async artifact => {
```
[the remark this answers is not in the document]
tamtamirr: Using **Promise.all** instead of awu to allow concurrency

packages/adapter-api/src/adapter.ts
```diff
@@ -39,15 +40,20 @@ export interface FetchResult {
...
-export type Group = {
-  id?: string
+export type Artifact = {
+  name: string
+  content: Buffer
+}
+
+export type AdapterGroupProperties = {
+  url?: string
+  artifacts?: types.NonEmptyArray<Artifact>
```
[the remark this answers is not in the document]
tamtamirr: Using NonEmptyArray forces the adapter to return a non empty array of artifacts.
If there are no artifacts, the property should remain undefined. We can further discuss this F2F if you disagree 🙏

packages/cli/src/commands/deploy.ts
tamtamirr: You right, I guess at some part of me coding it wasn't

packages/adapter-api/src/adapter.ts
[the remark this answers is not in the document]
ori-moisis: I understand what it means, all I am saying is that in terms of a public interface, this is more complex than it has to be, and I don't think we should have overly complex public interfaces

I realize that having two ways to declare no artifacts (undefined or empty array) is redundant, but I'd rather have the code deal with it than have the interface use non trivial typescript tricks

PR #5189, opened 2023-12-13
SALTO-5141: Allow metadata in requestForm to be null

idoamit8:
In this PR I added the option to fetch and deploy when metadata is null. 

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
+++ b/packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
@@ -107,7 +107,7 @@ const fromLayoutConfigRespToLayoutConfig = (
   layoutConfig: IssueLayoutConfiguration
 ): issueLayoutConfig => {
   const { containers } = layoutConfig.issueLayoutResult
-  const fieldItemIdToMetaData = Object.fromEntries(layoutConfig.metadata.configuration.items.nodes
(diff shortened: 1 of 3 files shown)


packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
```diff
@@ -107,7 +107,7 @@ const fromLayoutConfigRespToLayoutConfig = (
   layoutConfig: IssueLayoutConfiguration
 ): issueLayoutConfig => {
   const { containers } = layoutConfig.issueLayoutResult
-  const fieldItemIdToMetaData = Object.fromEntries(layoutConfig.metadata.configuration.items.nodes
+  const fieldItemIdToMetaData = Object.fromEntries((layoutConfig.metadata?.configuration.items.nodes ?? [])
```
omrilit, line 110: Should we be more defensive here, not assuming that `configuration.items.nodes` are all defined, and place more `?`?
Anyway, I think that `IssueLayoutConfiguration` type should reflect that `metadata` is nullable as well, right?

omrilit (commented):
Looks good!

packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
idoamit8: if the metadata is not null, but the other object (configuration.items.nodes) are not exist, it will not pass the schema guard.

PR #5190, opened 2023-12-13
SALTO-5140: Allow state static files to be stored in S3

ori-moisis:
This allows the workspace configuration to determine the location of the static file source for the state
Currently using the same storage provider as the state content

---

---
_Release Notes_: 
Core:
- Added support for storing state static files in S3

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/core/src/local-workspace/state/content_providers/file_content_provider.ts
+++ b/packages/core/src/local-workspace/state/content_providers/file_content_provider.ts
@@ -20,13 +20,17 @@ import { createHash } from 'crypto'
 import getStream from 'get-stream'
 
 import { collections } from '@salto-io/lowerdash'
+import { state } from '@salto-io/workspace'
 import { exists, rm, rename, replaceContents, createReadStream } from '@salto-io/file'
 import { StateContentProvider, getHashFromHashes } from './common'
+import { localDirectoryStore } from '../../dir_store'
 
 const { awu } = collections.asynciterable
 const glob = promisify(origGlob)
 
-export const createFileStateContentProvider = (): StateContentProvider => {
+export const STATE_STATIC_FILES_LOCAL_DIR = 'static-resources'
+
+export const createFileStateContentProvider = (localStorageDir: string): StateContentProvider => {
   const buildLocalStateFileName = (prefix: string, account: string): string => `${prefix}.${account}.jsonl.zip`
   const findStateFiles = (prefix: string): Promise<string[]> => glob(buildLocalStateFileName(prefix, '*([!.])'))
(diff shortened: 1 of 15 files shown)


packages/core/src/local-workspace/state/content_providers/file_content_provider.ts
```diff
@@ -20,13 +20,17 @@ import { createHash } from 'crypto'
...
+import { state } from '@salto-io/workspace'
 import { exists, rm, rename, replaceContents, createReadStream } from '@salto-io/file'
 import { StateContentProvider, getHashFromHashes } from './common'
+import { localDirectoryStore } from '../../dir_store'
 
 const { awu } = collections.asynciterable
 const glob = promisify(origGlob)
 
-export const createFileStateContentProvider = (): StateContentProvider => {
+export const STATE_STATIC_FILES_LOCAL_DIR = 'static-resources'
```
ori-moisis, line 31: this used a different constant in the past (the same constant as the NaCl's static resource folder name)
I felt ok with using a different constant because this really represents something else (the state's static resource folder name)
using the same value for backwards compatibility

packages/core/src/local-workspace/state/state.ts
```diff
@@ -98,14 +99,15 @@ export const getStateContentProvider = (
 ): StateContentProvider => {
   switch (stateConfig.provider) {
     case 'file': {
-      return createFileStateContentProvider()
+      const localStorageDir = stateConfig.options?.file?.localStorageDir ?? getLocalStoragePath(workspaceId)
```
yuribro, line 102: Doesn't this obsoletes all the BC in getLocalStorage in packages/core/src/local-workspace/workspace_config.ts?

And we assume that any implementation which wants to change the localStorageDir provides a custom WorkspaceConfigSource?

ori-moisis: yes, the code in `getLocalStorage` is for a migration that happened a very very long time ago, there is no real point in keeping it anymore (I can open a separate PR to remove it from there as well)

as for changing the localStorageDir, it seems like this option isn't really exposed today anyway (only three places use the variation of `workspaceConfigSource` the has a localStorageDir, two pass undefined and the other passes the value of getLocalStoragePath).
so in practice, there isn't really an option today to set a different local storage dir beyond setting the `SALTO_HOME` env var.
anyway, with the new code, if someone wants to configure the directory of the file content provider, that is what the option (`stateConfig.options.file.localStorageDir`) is there for

yuribro (approved):

PR #5192, opened 2023-12-13
SALTO-5110 NS important values filter

dantal4:
Adding general important values to all types - only those that exist in the type (in that order):
1. name
2. label
3. description
4. scriptid
5. isinactive / inactive / isInactive

This filter will run with the adapter config value `addImportantValues=true`

---

TODO:
- [x] add tests

---
_Release Notes_: 
Netsuite Adapter:
- Add important values (with adapter config `addImportantValues=true`)

---
_User Notifications_: 
Netsuite Adapter:
- The `_important_values` annotation will be added to all types (custom record types will also have `_self_important_values`)

### DIFF
--- a/packages/netsuite-adapter/src/filters/add_important_values.ts
+++ b/packages/netsuite-adapter/src/filters/add_important_values.ts
@@ -16,7 +16,7 @@
 import _ from 'lodash'
 import { logger } from '@salto-io/logging'
 import { CORE_ANNOTATIONS, ObjectType, isObjectType } from '@salto-io/adapter-api'
-import { ImportantValues } from '@salto-io/adapter-utils'
+import { ImportantValues, toImportantValues } from '@salto-io/adapter-utils'
 import { LocalFilterCreator } from '../filter'
 import { isCustomRecordType, netsuiteSupportedTypes } from '../types'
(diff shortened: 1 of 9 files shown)


packages/netsuite-adapter/src/filters/add_important_values.ts
```diff
@@ -0,0 +1,104 @@
...
+    indexed: false,
+  },
+  {
+    value: INACTIVE_FIELDS.isInactive,
+    highlighted: true,
+    indexed: true,
+  },
+]
+
+const toHighlightedImportantValues = (
```
omrilit, line 48: Don't we have it already written somewhere so we can DRY? (same goes for other parts in this file)

omrilit (approved):
Looks great!
I think that we can use this already for NS demos if needed, until we'll have the full mapping.
Do you think that it worth mentioning it in #ask-netsuite once in production?

packages/netsuite-adapter/src/filters/add_important_values.ts
dantal4: I'll move it to adapter-utils/important_values 👍

PR #5193, opened 2023-12-14
SALTO-5146: Fixing Form Fetch

idoamit8:
In this PR I fixed the form schemaGuard to allow a form to be untitled when I get the list of forms in a project. 
Their is additional  schemaGuard that blocks the untitled form to be fetched. 

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
None


shayc331 (approved):
🏆

PR #5200, opened 2023-12-17
SALTO-5071 Zendesk request timeout

adfineberg:
Add request timeout documentation for the adapters, add default of 5 seconds for Zendesk with retries.

---

---
_Release Notes_: 
__Zendesk adapter:__
* Add default timeout for requests

---
_User Notifications_: 

### DIFF
--- a/packages/adapter-components/src/client/config.ts
+++ b/packages/adapter-components/src/client/config.ts
@@ -37,7 +37,7 @@ export type ClientRetryConfig = Partial<{
 
 export type ClientTimeoutConfig = Partial<{
   maxDuration: number
-  resetTimeoutBetweenAttempts: boolean
+  retryOnTimeout: boolean
   lastRetryNoTimeout: boolean
 }>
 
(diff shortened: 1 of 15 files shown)


packages/adapter-components/src/client/config.ts
```diff
@@ -37,7 +37,7 @@ export type ClientRetryConfig = Partial<{
 
 export type ClientTimeoutConfig = Partial<{
   maxDuration: number
-  resetTimeoutBetweenAttempts: boolean
+  retryOnTimeout: boolean
```
netama, line 40: this doesn't look like just documentation?

adfineberg: I wasn't sure if we want to expose all the options in the documentation. What is the norm for this?

packages/jira-adapter/config_doc.md
```diff
@@ -57,6 +57,13 @@ jira {
 | [rateLimit](#rate-limit-configuration-options)                | `{}` (no overrides)      | Limits on the number of concurrent requests of different types
 | [maxRequestsPerMinute]                                        | unlimited                | Limits on the number of requests per minute
 | usePrivateAPI                                                 | true                     | Whether to use Jira Private API when fetching and deploying changes
+| [timeout](#client-timeout-configuration-options)              | `{}` (no overrides)      | Configuration for setting request timeouts
+
+#### Client timeout configuration options
+
+| Name           | Default when undefined | Description
+|----------------|------------------------|------------
+| [maxDuration]  | `0` (unlimited)     | Set a timeout (in milliseconds) on requests
```
netama, line 66: 1. nit:
```suggestion
| [maxDuration]  | `0` (unlimited)        | Set a timeout (in milliseconds) on requests
```
(also in other places)
2. is there a reason not to document the other fields?

PR #5201, opened 2023-12-17
SALTO-5157: Add serviceDeskId in JSM project addition 

idoamit8:
In this PR I added support for adding serviceDeskId in JSM project addition 

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
None


omrilit (approved):
LGTM!
Is there anything that we should take care for existing JSM projects that were created using Salto?

idoamit8:
@omrilit If a project is added to a new environment using Salto and there hasn't been a fetch since, the next deploy of JSM type that needs the project serviceDeskId will fail because the added project lacks a serviceDeskId. However, fetching the environment will retrieve the project.

PR #5207, opened 2023-12-18
SALTO-5170: Add jsm user count

idoamit8:
Added log to count users in jsm accounts

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
None


shayc331 (approved):
👑

PR #5209, opened 2023-12-18
SALTO-5154: Disable usePrivateAPI when connected with OAuth

shir-reifenberg:
private API types are not supported when account is added with OAuth login

---

When connecting with OAuth, private APIs can't be fetched. the flag is on by default, so each user connecting with OAuth will experience this.
We create fetch warning + add config change to change `usePrivateAPI` to false

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/config/config_change.ts
+++ b/packages/adapter-components/src/config/config_change.ts
@@ -13,36 +13,59 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
-import { collections } from '@salto-io/lowerdash'
+import { values as lowerdashValues } from '@salto-io/lowerdash'
 import { InstanceElement, ElemID, ObjectType } from '@salto-io/adapter-api'
-
 import { formatConfigSuggestionsReasons } from '@salto-io/adapter-utils'
 
-const { makeArray } = collections.array
+const { isDefined } = lowerdashValues
 const FETCH_CONFIG = 'fetch'
+const CLIENT_CONFIG = 'client'
+
+export const TYPE_TO_EXCLUDE = 'typeToExclude'
+export const DISABLE_PRIVATE_API = 'disablePrivateAPI'
+type ConfigSuggestionType = 'typeToExclude' | 'disablePrivateAPI'
 
 export type ConfigChangeSuggestion = {
-  typeToExclude: string
+  type: ConfigSuggestionType
+  value?: string
+  reason: string
 }
 
-// When some items cannot be fetched, they are excluded from the fetch process, and the user is notified.
-export const getConfigWithExcludeFromConfigChanges = ({
+/**
+ * Update config with types to exclude or disabling private API according to config changes
+ */
+export const getUpdatedCofigFromConfigChanges = ({
   configChanges,
   currentConfig,
   configType,
-  adapterName,
 }: {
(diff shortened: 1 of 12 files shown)


packages/adapter-components/src/elements/ducktype/transformer.ts
```diff
@@ -413,7 +413,11 @@ export const getAllElements = async ({
           && (reversedSupportedTypes[args.typeName] !== undefined)) {
           const typesToExclude = reversedSupportedTypes[args.typeName]
           typesToExclude.forEach(type => {
-            configSuggestions.push({ typeToExclude: type })
+            configSuggestions.push({
+              type: TYPE_TO_EXCLUDE,
+              value: type,
+              reason: `Salto failed to fetch type ${type}`,
```
omrilit, line 419: Maybe worth consulting with a PM about the message here?
IMO it's better to use "Salto failed to fetch ${type} type".

packages/adapter-components/src/config/config_change.ts
```diff
@@ -13,52 +13,70 @@
...
+  const shouldDisablePrivateApi = configChanges.find(configChange => configChange.type === DISABLE_PRIVATE_API)
+
+  const updatedFetchConfig = typesToExclude.length > 0
+    ? {
+      ...currentConfig.value[FETCH_CONFIG],
+      exclude: [
+        ...currentConfig.value[FETCH_CONFIG].exclude,
+        ...typesToExclude,
+      ],
+    }
```
omrilit, line 63: nit:
I think that styling wise this would be clearer, but you can definitely ignore it if you prefer your version
```suggestion
  const typesToExclude = configChanges
    .filter(configChange => configChange.type === TYPE_TO_EXCLUDE && isDefined(configChange.value))
    .map(configChange => configChange.value)

  const shouldDisablePrivateApi = configChanges.find(configChange => configChange.type === DISABLE_PRIVATE_API)

  const updatedFetchConfig = typesToExclude.length > 0
    ? {
      ...currentConfig.value[FETCH_CONFIG],
      exclude: [
        ...currentConfig.value[FETCH_CONFIG].exclude,
        ...typesToExclude.map(typeToExclude => ({ type: typeToExclude })),
      ],
    }
```

packages/okta-adapter/src/filters/group_push.ts
```diff
@@ -230,11 +230,7 @@ const groupPushFilter: FilterCreator = ({ config, adminClient, getElemIdFunc })
 
     if (adminClient === undefined) {
       log.error('Admin client is undefined, could not fetch group push instances')
-      const error: SaltoError = {
```
omrilit, line 233: Should we emit an error when isOAuthLogin is true?
Is it a valid use case that we will want to log an error but not return such indication to the user using FetchError?

omrilit (commented):
Great work!
Approving to save cycle

omrilit (approved):

packages/okta-adapter/src/filters/group_push.ts
shir-reifenberg: I changed the logic now and I don't create `adminClient` when connecting with OAuth, so it wouldn't be an error
I'll adjust log level accordingly

PR #5212, opened 2023-12-21
SALTO-5185: Fixed Duplication of recursed into instances. 

idoamit8:
Modified the implementation to ensure that the parent instance being pointed to and the newly created instance are the same.

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/adapter-components/src/elements/swagger/instance_elements.ts
```diff
@@ -132,13 +132,14 @@ const extractStandaloneFields = async (
...
-    element: inst,
+  inst.value = await transformValues({
     transformFunc: extractFields,
+    type: await inst.getType(),
+    values: inst.value,
     strict: false,
-  })
-  return [updatedInst, ...additionalInstances]
+  }) ?? {}
+  additionalInstances.push(inst)
```
netama, line 141: I think it'd be better if `inst` was first (then you also won't need to change the tests), maybe rename this to `allInstances` and push `inst` first?

idoamit8: I changed the tests already, but I can re do the change :)

packages/adapter-components/src/elements/swagger/instance_elements.ts
```diff
@@ -132,13 +132,14 @@ const extractStandaloneFields = async (
...
-  const updatedInst = await transformElement({
-    element: inst,
+  inst.value = await transformValues({
     transformFunc: extractFields,
+    type: await inst.getType(),
+    values: inst.value,
     strict: false,
-  })
-  return [updatedInst, ...additionalInstances]
+  }) ?? {}
```
netama, line 140: maybe fallback to `inst.value` (or even log an error, though not sure it's worth it)?

packages/adapter-components/src/elements/swagger/instance_elements.ts
idoamit8: I was thinking that doing deconstruction to an array can lead to memory errors

netama: I think it'd be better for the parent instance to be first (you can still push it to the array). it's better not to deconstruct (this is pretty old code), I think at this scale we're usually ok but it's probably better - we should just keep the parent first if possible

netama (approved):
LGTM with a nit. good catch! 💪

PR #5219, opened 2023-12-25
SALTO-5196: Salesforce internalIds of managed Elements bugfix

tamtamirr:
Fixed a bug where we failed to resolve internal ids of some managed Elements.

---

The previous code calculated a wrong `fullName`, so we failed to link the correct FileProperties to the Element.
- I've added a log as part of this PR to monitor bugs in the `add_missing_ids` filter. 

---
_Release Notes_: 
Salesforce Adapter:
- Fixed a bug where we failed to resolve some internal ids of managed Elements. 

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/filters/add_missing_ids.ts
```diff
@@ -14,18 +14,51 @@
...
+  'salesforce.RecordType.instance.Idea_InternalIdeasIdeaRecordType',
+])
+
+const TYPES_WITH_NO_INTERNAL_IDS = [
+  TOPICS_FOR_OBJECTS_METADATA_TYPE,
+  GLOBAL_VALUE_SET_TRANSLATION_METADATA_TYPE,
+]
+
+// Used for logging
+const shouldHaveInternalId = (element: Element): boolean => {
```
amitgoldarb, line 49: I think that this part of the code isn't great. We don't really need this persistently in our code. I think we should strive to have this in our code for at most a week. It will probably give us sufficient understanding of all of the edge cases. @tamtamirr Does that make sense?

amitgoldarb (approved):
LGTM!
Added a small comment.

packages/salesforce-adapter/src/filters/add_missing_ids.ts
tamtamirr: I would want this log to remain to monitor regressions here as they did happen a couple times before.
Considering the adapter as logics that rely on the `internal ids` I think it's important.
While this code is very specific, it serves a needed purpose IMO. It's also safe and used for logging only

PR #5220, opened 2023-12-25
SALTO-4650: Salesforce InstalledPackage Instances ElemID with version

tamtamirr:
This PR adds an optional feature that changes the ElemID of `InstalledPakcage` instances to contain the Package version as-well.

---

Workspace diff: https://github.com/salto-io/tamir-sf/commit/fc8d6c481741d439f2ebf9eafe848a46ddb11227

Please only review the changes in `transformer.ts` and the test suite `createInstanceElement` I've added to `utils.test.ts`, as all of the other changes are related to the refactor of the params of `createInstanceElement`

This will force alignment between the installed versions in the source and the target, since the `Installeed Package` instances will have different **ElemIDs** 

---
_Release Notes_: 
Salesforce Adapter:
- Add the optional feature `installedPackageWithVersion` that changes the Element ID of `InstalledPackage` instances to contain the package version as-well. 

---
_User Notifications_: 
Salesforce:
When using the feature `installedPackageWithVersion`, expect workspace changes to all the Elements coming from `Installed Package` and the `InstalledPackage` metadata instances

### DIFF
--- a/packages/salesforce-adapter/e2e_test/adapter.test.ts
+++ b/packages/salesforce-adapter/e2e_test/adapter.test.ts
@@ -46,7 +46,7 @@ import {
 import realAdapter from './adapter'
 import {
   findElements, findStandardFieldsObject, findAnnotationsObject, findCustomFieldsObject,
-  findFullCustomObject,
+  findFullCustomObject, defaultFilterContext,
 } from '../test/utils'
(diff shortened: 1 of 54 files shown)


tamtamirr:
decided not to do so for the time being. Closing due to lack of attention

PR #5221, opened 2023-12-25
SALTO-3432 add GetSelectValue NetSuite SOAP API

dantal4:
The [GetSelectValue](https://docs.oracle.com/en/cloud/saas/netsuite/ns-online-help/section_N3504236.html#getSelectValue) API can be used to get the correct `internalId` of missing references (ASV) in data instances, as it return a list of the possible options for a specific field of a specific type.
The `filterBy` field should be used for fields that are dependent on the value of another field - for example:
```typescript
getSelectValue('account', 'unitstype')
getSelectValue('account', 'unit', [{ field: 'unitstype', internalId: '2' }])
```

This API should be used in deploy, in order to get missing internalIds in data instances.

---

TODO:
- [x] add tests

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/netsuite-adapter/src/client/client.ts
+++ b/packages/netsuite-adapter/src/client/client.ts
@@ -586,4 +586,15 @@ export default class NetsuiteClient {
     }
     return this.suiteAppClient.getCustomRecords(customRecordTypes)
   }
+
+  public async getSelectValue(
+    type: string,
+    field: string,
+    filterBy: { field: string; internalId: string }[] = [],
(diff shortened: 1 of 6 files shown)


omrilit (approved):
The code looks good so I am approving.
Could you please elaborate when and how we will use it? will it be done in Modification changes only?

dantal4:
We need to use this API whenever data element is deployed and has ASV in it, in additions/modifications - so we'll get the correct/updated internalIds to put instead of ASV.

PR #5223, opened 2023-12-26
SALTO-5150: handle socket hang up errors

edenhassid:
handle socket hang up errors in zendesk

---

---
_Release Notes_: 
zendesk:
* handle socket hang up errors

---
_User Notifications_: 

### DIFF
--- a/packages/zendesk-adapter/src/filters/article/article.ts
+++ b/packages/zendesk-adapter/src/filters/article/article.ts
@@ -15,7 +15,7 @@
 */
 import _ from 'lodash'
 import { logger } from '@salto-io/logging'
-import { collections, strings } from '@salto-io/lowerdash'
+import { collections, strings, promises } from '@salto-io/lowerdash'
 import {
   AdditionChange,
   Change,
@@ -58,13 +58,14 @@ import {
   isAttachments,
   updateArticleTranslationBody,
 } from './utils'
-import { API_DEFINITIONS_CONFIG, FETCH_CONFIG, isGuideEnabled } from '../../config'
+import { API_DEFINITIONS_CONFIG, CLIENT_CONFIG, FETCH_CONFIG, isGuideEnabled, ZendeskConfig } from '../../config'
 
 
 const log = logger(module)
 const { awu } = collections.asynciterable
 const { makeArray } = collections.array
 const { matchAll } = strings
+const { sleep } = promises.timeout
 
 const USER_SEGMENT_ID_FIELD = 'user_segment_id'
 const ATTACHMENTS_IDS_REGEX = new RegExp(`(?<url>/${ARTICLE_ATTACHMENTS_FIELD}/)(?<id>\\d+)`, 'g')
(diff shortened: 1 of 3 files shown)


packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -205,46 +264,21 @@ const handleArticleAttachmentsPreDeploy = async ({ changes, client, elementsSour
...
-          return
-        }
-        await deleteArticleAttachment(client, attachmentChange.data.before)
-        return
-      }
-      const parentArticleName = parentArticleRef.elemID.name
-      articleNameToAttachments[parentArticleName] = (
-        articleNameToAttachments[parentArticleName] || []
-      ).concat(attachmentInstance.value.id)
+  const rateLimit = config[CLIENT_CONFIG]?.rateLimit?.get ?? 50
```
netama, line 267: technically we have a default in `DEFAULT_MAX_CONCURRENT_API_REQUESTS`, is there a reason not to reuse it? (if so - maybe this should be in a new property?)

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -205,46 +264,21 @@ const handleArticleAttachmentsPreDeploy = async ({ changes, client, elementsSour
...
+  log.debug(`there are ${attachmentChanges.length} attachment changes, going to handle them in chunks of ${rateLimit}`)
+  const attachChangesChunks = _.chunk(attachmentChanges, rateLimit)
+  await awu(attachChangesChunks)
+    .forEach(async (
+      attachmentChangesChunk: (AdditionChange<InstanceElement> | ModificationChange<InstanceElement>)[],
+      index: number) => {
```
netama, line 273: tiny nit:
```suggestion
      index: number,
    ) => {
```

packages/zendesk-adapter/src/filters/article/article.ts
```diff
@@ -205,46 +264,21 @@ const handleArticleAttachmentsPreDeploy = async ({ changes, client, elementsSour
...
+      log.debug(`starting article attachment change chunk ${index + 1}/${attachChangesChunks.length}`)
+      await awu(attachmentChangesChunk).forEach(attachmentChange => createUnassociatedAttachmentFromChange({
```
netama, line 275: 1. is there a reason not to use `Promise.all()` here instead of `awu().forEach()`?
2. (not blocking) any chance we can generalize the code from `getArticleAttachments` to get the function to run as a callback (here it will call `createUnassociatedAttachmentFromChange` and there it will call the function wrapping `getAttachmentContent`), and then reuse the surrounding code that separates into chunks and adds the sleep? I have a feeling we might run into this in more places 🙃

netama (commented):
looks good :) added some minor questions. do we want to add unittests for these? or just making sure the existing code won't break?

edenhassid:
about the tests, I was going with the "just making sure the existing code won't break" approach

packages/zendesk-adapter/src/filters/article/article.ts
edenhassid: do you mean a new property inside of the `rateLimit` in the config?

packages/zendesk-adapter/src/filters/article/article.ts
edenhassid: 1. yes, we think that sending too many requests at the same time will cause the socket hang up errors
2. they are not exactly the same and I think for now it will make it more complicated to generalize it

netama (approved):

PR #5225, opened 2023-12-26
SALTO-5204: Link Salesforce Data Deploy Errors to Elements

tamtamirr:
Link Data Deploy Errors to Elements.

---

This PR simply creates `SaltoElementError` per data instance upon failure.

---
_Release Notes_: 
Salesforce Adapter:
- Data Deploy Errors will be linked to the failed Elements.

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -507,18 +507,20 @@ const isModificationChangeList = <T>(
...
   dataManagement?: DataManagement,
 ): Promise<DeployResult> => {
+  const customObjectInstancesDeployError = (message: string): DeployResult => ({
+    appliedChanges: [],
+    errors: changes.map(change => ({
+      message,
+      severity: 'Error',
+      elemID: getChangeData(change).elemID,
+    })),
+  })
```
tomsellek, line 523: This will create an error for the entire change group, not just for the failing elements, no? I get that the whole change group will fail to deploy if one element fails, but won't it be confusing for the user to see seemingly unrelated errors associated with elements that have nothing wrong with them?

tomsellek (approved):
Approving to save cycles, but I did have a question about whether we are being over-eager with the errors.

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
tamtamirr: This is the expected behavior here since we perform a batch operation on all of the Elements of this group in data deploy.

PR #5227, opened 2023-12-26
SALTO-5183: Solved serviceid validator for queue

idoamit8:
In this PR I solved serviceId validator for queue type

---

There is an odd bug in standard_deployment relates to queues that brings back workspace warning Error validating `[jira.Queue.instance.Unassigned_issues10_MYJ8@su.id](mailto:jira.Queue.instance.Unassigned_issues10_MYJ8@su.id)": Invalid value type for serviceid` because in the fetch (Get) the id of the queue is a string and in the addition (POST) the id is a number.

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/deployment/standard_deployment.ts
+++ b/packages/jira-adapter/src/deployment/standard_deployment.ts
@@ -31,11 +31,11 @@ type DeployChangeParam = {
   fieldsToIgnore?: string[] | ((path: ElemID) => boolean)
   additionalUrlVars?: Record<string, string>
   elementsSource?: ReadOnlyElementsSource
-  modifyChangeFunc?: (
+  serviceIdSetter?: (
     instance: InstanceElement,
     serviceIdField: string,
-    response: clientUtils.ResponseValue)
-    => void
+    response: clientUtils.ResponseValue
+    ) => void
 }
 
 const invertKeysNames = (instance: Record<string, unknown>): void => {
@@ -45,6 +45,14 @@ const invertKeysNames = (instance: Record<string, unknown>): void => {
       instance[invertNaclCase(key)] = instance[key]
       delete instance[key]
     })
+}
+
+const defaultServiceIdSetter = (
+  instance: InstanceElement,
+  serviceIdField: string,
+  response: clientUtils.ResponseValue
+): void => {
+  instance.value[serviceIdField] = response[serviceIdField]
 }
 
 /**
@@ -57,9 +65,7 @@ export const defaultDeployChange = async ({
   fieldsToIgnore = [],
   additionalUrlVars,
   elementsSource,
-  modifyChangeFunc = (instance, serviceIdField, response) => {
-    instance.value[serviceIdField] = response[serviceIdField]
-  },
+  serviceIdSetter = defaultServiceIdSetter,
 }: DeployChangeParam): Promise<
   clientUtils.ResponseValue | clientUtils.ResponseValue[] | undefined
 > => {
@@ -102,7 +108,7 @@ export const defaultDeployChange = async ({
     if (!Array.isArray(response)) {
       const serviceIdField = apiDefinitions.types[getChangeData(change).elemID.typeName]?.transformation?.serviceIdField ?? 'id'
       if (response?.[serviceIdField] !== undefined) {
-        modifyChangeFunc(change.data.after, serviceIdField, response)
+        serviceIdSetter(change.data.after, serviceIdField, response)
       }
     } else {
       log.warn('Received unexpected response from deployChange: %o', response)
--- a/packages/jira-adapter/src/deployment/standard_deployment.ts
+++ b/packages/jira-adapter/src/deployment/standard_deployment.ts
@@ -35,7 +35,7 @@ type DeployChangeParam = {
     instance: InstanceElement,
     serviceIdField: string,
     response: clientUtils.ResponseValue
-    ) => void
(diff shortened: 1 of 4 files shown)


packages/jira-adapter/src/deployment/standard_deployment.ts
```diff
@@ -31,6 +31,11 @@ type DeployChangeParam = {
   fieldsToIgnore?: string[] | ((path: ElemID) => boolean)
   additionalUrlVars?: Record<string, string>
   elementsSource?: ReadOnlyElementsSource
+  modifyChangeFunc?: (
+    instance: InstanceElement,
+    serviceIdField: string,
+    response: clientUtils.ResponseValue)
+    => void
```
omrilit, line 38: Styling nit
```suggestion
  modifyChangeFunc?: (
    instance: InstanceElement,
    serviceIdField: string,
    response: clientUtils.ResponseValue
  ) => void
```

packages/jira-adapter/src/deployment/standard_deployment.ts
```diff
@@ -52,6 +57,9 @@ export const defaultDeployChange = async ({
   fieldsToIgnore = [],
   additionalUrlVars,
   elementsSource,
+  modifyChangeFunc = (instance, serviceIdField, response) => {
+    instance.value[serviceIdField] = response[serviceIdField]
+  },
```
omrilit, line 62: please extract this function to a variable and use it as the default arg.

packages/jira-adapter/src/deployment/standard_deployment.ts
```diff
@@ -31,6 +31,11 @@ type DeployChangeParam = {
   fieldsToIgnore?: string[] | ((path: ElemID) => boolean)
   additionalUrlVars?: Record<string, string>
   elementsSource?: ReadOnlyElementsSource
+  modifyChangeFunc?: (
```
omrilit, line 34: I think that if the argument is called serviceIdField, we should name this function serviceIdSetter.
Otherwise the naming here is too implicit and unclear IMO

packages/jira-adapter/src/filters/jsm_types_deploy_filter.ts
```diff
@@ -71,6 +71,12 @@ const filterCreator: FilterCreator = ({ config, client }) => ({
           apiDefinitions: jsmApiDefinitions,
           fieldsToIgnore,
           additionalUrlVars,
+          modifyChangeFunc: (instance, serviceIdField, response) => {
```
omrilit, line 74: I think that the intention was that the modifyChangeFunc func will be passed only for QUEUE_TYPE, here you override the default impl of modifyChangeFunc and do not take care for all the other types.
Am I right?

omrilit (commented):
Looks good! had few comments :)

packages/jira-adapter/src/filters/jsm_types_deploy_filter.ts
idoamit8: Yep, Fixed it now :)

packages/jira-adapter/src/deployment/standard_deployment.ts
```diff
@@ -31,6 +31,11 @@ type DeployChangeParam = {
   fieldsToIgnore?: string[] | ((path: ElemID) => boolean)
   additionalUrlVars?: Record<string, string>
   elementsSource?: ReadOnlyElementsSource
+  serviceIdSetter?: (
+    instance: InstanceElement,
+    serviceIdField: string,
+    response: clientUtils.ResponseValue
+    ) => void
```
omrilit, line 38:
```suggestion
  serviceIdSetter?: (
    instance: InstanceElement,
    serviceIdField: string,
    response: clientUtils.ResponseValue
  ) => void
```

packages/jira-adapter/src/filters/jsm_types_deploy_filter.ts
```diff
@@ -71,6 +71,13 @@ const filterCreator: FilterCreator = ({ config, client }) => ({
           apiDefinitions: jsmApiDefinitions,
           fieldsToIgnore,
           additionalUrlVars,
+          serviceIdSetter: (instance, serviceIdField, response) => {
+            const serviceFieldValue = response?.[serviceIdField]
+            if (instance.elemID.typeName === QUEUE_TYPE && _.isNumber(serviceFieldValue)) {
+              instance.value[serviceIdField] = serviceFieldValue.toString()
+            }
+            instance.value[serviceIdField] = serviceFieldValue
+          },
```
omrilit, line 80: The only thing I can think of is that actually you'd like to override the default serviceIdSetter behavior in case of Queue type.
In the other cases it's better not to pass the serviceIdSetter func or to pass `defaultServiceIdSetter` explicitly (it should be exported in order to allow it...).
The advantage of this approach is that in case defaultServiceIdSetter will change in the future you won't have to remember to update the implicit duplication of it here.
WDYT?

omrilit (approved):
Looks great!
Had an idea for 1 last small improvement, yet it's not mandatory, your call. Approving in case you think it's redundant

packages/jira-adapter/src/filters/jsm_types_deploy_filter.ts
idoamit8: i agree with you and the code has been changed :)

PR #5228, opened 2023-12-26
SALTO-4902 Validate group is not already mapped before adding a GroupPush instance

omrifarkash:
New change validator that detects if user tries to add groupPush instance with userGroupId and a parent (application)
that already exists in the target env, and return change error with severity `Error`

---

---
_Release Notes_: 
_Okta Adapter_:
Validation error when user tries to add group push with groupId and application relation that already exists in the target environment

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/change_validators/group_push_to_application_uniqueness.ts
+++ b/packages/okta-adapter/src/change_validators/group_push_to_application_uniqueness.ts
@@ -29,31 +29,46 @@ export const groupPushToApplicationUniquenessValidator: ChangeValidator = async 
     log.warn('elementsSource was not provided to groupPushToApplicationUniqueness, skipping validator')
     return []
   }
-
-  const existingGroupToApplication = Object.fromEntries(await (awu(await elementsSource.getAll())
-    .filter(isInstanceElement)
-    .filter(instance => instance.elemID.typeName === GROUP_PUSH_TYPE_NAME)
-    .map(groupPush =>
-      ([
-        groupPush.value?.userGroupId?.elemID?.getFullName(),
-        groupPush.annotations[CORE_ANNOTATIONS.PARENT]?.[0]?.elemID?.getFullName(),
-      ])))
-    .toArray())
-
-  return changes
+  const groupPushAdditionChanges = changes
     .filter(isInstanceChange)
     .filter(isAdditionChange)
     .map(change => getChangeData(change))
     .filter(instance => instance.elemID.typeName === GROUP_PUSH_TYPE_NAME)
+  if (groupPushAdditionChanges.length === 0) {
+    return []
+  }
+  const existingGroupToApplication = await awu(await elementsSource.getAll())
+    .filter(isInstanceElement)
+    .filter(instance => instance.elemID.typeName === GROUP_PUSH_TYPE_NAME)
+    .reduce<Record<string, { application: string; groupPushAlias: string }[]>>((record, currentGroupPush) => {
+      const currentGroup = currentGroupPush.value?.userGroupId?.elemID?.getFullName() as string
+      const currentApplication = currentGroupPush
+        .annotations[CORE_ANNOTATIONS.PARENT]?.[0]?.elemID?.getFullName() as string
+      if (!currentGroup || !currentApplication) {
+        return record
+      }
+      return {
+        ...record,
+        [currentGroup]: (record[currentGroup] || [])
+          .concat({
+            application: currentApplication,
+            groupPushAlias: currentGroupPush.annotations[CORE_ANNOTATIONS.ALIAS],
+          }),
+      }
+    }, {})
+
+  return groupPushAdditionChanges
     .flatMap((instance): ChangeError[] => {
       const group = instance.value?.userGroupId?.elemID?.getFullName()
       const application = instance.annotations[CORE_ANNOTATIONS.PARENT]?.[0]?.elemID?.getFullName()
-      if (group && application && existingGroupToApplication[group] === application) {
+      const existingRecord = (existingGroupToApplication[group] || [])
+        .find(record => record.application === application)
+      if (existingRecord) {
         return [{
           elemID: instance.elemID,
           severity: 'Error',
-          message: `${group} to ${application} can only be defined on a single groupPush instance`,
-          detailedMessage: `${group} to ${application} can only be defined on a single groupPush instance`,
+          message: `Group ${group} is already mapped to ${application}`,
+          detailedMessage: `GroupPush ${existingRecord.groupPushAlias} already maps group ${group} to application ${application}`,
         }]
       }
       return []
--- a/packages/okta-adapter/src/change_validators/group_push_to_application_uniqueness.ts
+++ b/packages/okta-adapter/src/change_validators/group_push_to_application_uniqueness.ts
@@ -1,0 +1,61 @@
+/*
+*                      Copyright 2023 Salto Labs Ltd.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+import { CORE_ANNOTATIONS, ChangeError, ChangeValidator, getChangeData, isAdditionChange, isInstanceChange, isInstanceElement } from '@salto-io/adapter-api'
+import { collections } from '@salto-io/lowerdash'
+import { logger } from '@salto-io/logging'
+import { GROUP_PUSH_TYPE_NAME } from '../constants'
+
+const { awu } = collections.asynciterable
+const log = logger(module)
+
+/**
+ * Validate uniqueness of group push to application
+ */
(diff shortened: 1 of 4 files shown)


packages/okta-adapter/src/change_validators/group_push_to_application_uniqueness.ts
```diff
@@ -0,0 +1,61 @@
...
+    .filter(instance => instance.elemID.typeName === GROUP_PUSH_TYPE_NAME)
+    .flatMap((instance): ChangeError[] => {
+      const group = instance.value?.userGroupId?.elemID?.getFullName()
+      const application = instance.annotations[CORE_ANNOTATIONS.PARENT]?.[0]?.elemID?.getFullName()
+      if (group && application && existingGroupToApplication[group] === application) {
+        return [{
+          elemID: instance.elemID,
+          severity: 'Error',
+          message: `${group} to ${application} can only be defined on a single groupPush instance`,
+          detailedMessage: `${group} to ${application} can only be defined on a single groupPush instance`,
```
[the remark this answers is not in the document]
shir-reifenberg: Another thing - as this errors are user visible, I think we should not include `.getFullName()` rather but `.name` or even alias

packages/okta-adapter/src/change_validators/group_push_to_application_uniqueness.ts
```diff
@@ -0,0 +1,61 @@
...
+    return []
+  }
+
+  const existingGroupToApplication = Object.fromEntries(await (awu(await elementsSource.getAll())
+    .filter(isInstanceElement)
+    .filter(instance => instance.elemID.typeName === GROUP_PUSH_TYPE_NAME)
+    .map(groupPush =>
+      ([
+        groupPush.value?.userGroupId?.elemID?.getFullName(),
+        groupPush.annotations[CORE_ANNOTATIONS.PARENT]?.[0]?.elemID?.getFullName(),
```
shir-reifenberg, line 39: I think we should verify we have a reference here in both cases. also, please you can use `getParent`.

we can do something like
```
.filter(groupPush => isReferenceExpression(groupPush.value.userGroupId) && isReferenceExpression(getParent(groupPush)))
.map(groupPush => ....)
```

packages/okta-adapter/src/change_validators/group_push_to_application_uniqueness.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  return changes
+    .filter(isInstanceChange)
+    .filter(isAdditionChange)
+    .map(change => getChangeData(change))
+    .filter(instance => instance.elemID.typeName === GROUP_PUSH_TYPE_NAME)
+    .flatMap((instance): ChangeError[] => {
+      const group = instance.value?.userGroupId?.elemID?.getFullName()
+      const application = instance.annotations[CORE_ANNOTATIONS.PARENT]?.[0]?.elemID?.getFullName()
+      if (group && application && existingGroupToApplication[group] === application) {
```
shir-reifenberg, line 51: I might be missing something, but isn't the current `GroupPush`'s `Group` and `Application` will be in `existingGroupToApplication` and we'll get true here even if no other `GroupPush` maps this `Group` ?

shir-reifenberg (commented):
Super fast 🚀 Added some comment and a question :)

packages/okta-adapter/src/change_validators/group_push_to_application_uniqueness.ts
omrifarkash: I am not sure, what is inside sourceElement? I thought that we have there only the existing elements.
If so, the current `GroupPush` shouldn't be included there since the change is `addition`

packages/okta-adapter/src/change_validators/group_push_to_application_uniqueness.ts
omrifarkash: I will do it if you think it is better code, but it will have no effect.
any record from type `undefined: someting | somthing: undefined | undefined: undefined`
Will have no effect since I am verifying that the change `group` and `application are defined 
`if (group && application && existingGroupToApplication[group] === application)` (line 58? index might changed)

packages/okta-adapter/src/change_validators/group_push_to_application_uniqueness.ts
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Validate uniqueness of group push to application
+ */
+export const groupPushToApplicationUniquenessValidator: ChangeValidator = async (changes, elementsSource) => {
+  if (!elementsSource) {
+    log.warn('elementsSource was not provided to groupPushToApplicationUniqueness, skipping validator')
+    return []
+  }
+
+  const existingGroupToApplication = Object.fromEntries(await (awu(await elementsSource.getAll())
```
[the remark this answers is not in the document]
omrifarkash: Nice

packages/okta-adapter/src/change_validators/group_push_to_application_uniqueness.ts
omrifarkash: Also `getParent` supposed to return and `InstanceElement` and throws error in this case

packages/okta-adapter/src/change_validators/group_push_to_application_uniqueness.ts
shir-reifenberg: If I remember right, elementSource also include pending changes, and if so there will be unwanted behaviour as you said :)

packages/okta-adapter/src/change_validators/group_push_to_application_uniqueness.ts
shir-reifenberg: 1. you can use `getParents` if you prefer (it does not throw), but we usually access parent annotation with getParent or getParents.
2. I think it's better practice to verify `userGroupId` is a reference expression (and also more readable)

packages/okta-adapter/src/change_validators/group_push_to_application_uniqueness.ts
```diff
@@ -0,0 +1,78 @@
...
+    return []
+  }
+  const existingGroupToApplication = await awu(await elementsSource.getAll())
+    .filter(isInstanceElement)
+    .filter(instance => instance.elemID.typeName === GROUP_PUSH_TYPE_NAME)
+    .reduce<Record<string, { application: string; groupPushInstance: InstanceElement }[]>>(
+      (record, currentGroupPush) => {
+        const currentGroup = currentGroupPush.value?.userGroupId?.elemID?.getFullName() as string
+        const currentApplication = currentGroupPush
+          .annotations[CORE_ANNOTATIONS.PARENT]?.[0]?.elemID?.getFullName() as string
```
shir-reifenberg, line 47: please avoid casting when possible 🙏

packages/okta-adapter/src/change_validators/group_push_to_application_uniqueness.ts
```diff
@@ -0,0 +1,78 @@
...
+            }),
+        }
+      }, {}
+    )
+
+  return groupPushAdditionChanges
+    .flatMap((instance): ChangeError[] => {
+      const group = instance.value?.userGroupId?.elemID?.getFullName()
+      const application = instance.annotations[CORE_ANNOTATIONS.PARENT]?.[0]?.elemID?.getFullName()
+      const existingRecord = (existingGroupToApplication[group] || [])
```
shir-reifenberg, line 66: nit 
```suggestion
      const existingRecord = (existingGroupToApplication[group] ?? [])
```

packages/okta-adapter/src/change_validators/group_push_to_application_uniqueness.ts
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * Validate uniqueness of group push to application
+ */
+export const groupPushToApplicationUniquenessValidator: ChangeValidator = async (changes, elementsSource) => {
+  if (!elementsSource) {
+    log.warn('elementsSource was not provided to groupPushToApplicationUniqueness, skipping validator')
+    return []
+  }
+  const groupPushAdditionChanges = changes
```
shir-reifenberg, line 32: naming nit - it's no longer changes but instances
so maybe `groupPushAdditions` / `addedGroupPushInstances`

packages/okta-adapter/src/change_validators/group_push_to_application_uniqueness.ts
```diff
@@ -0,0 +1,78 @@
...
+    .filter(isAdditionChange)
+    .map(change => getChangeData(change))
+    .filter(instance => instance.elemID.typeName === GROUP_PUSH_TYPE_NAME)
+  if (groupPushAdditionChanges.length === 0) {
+    return []
+  }
+  const existingGroupToApplication = await awu(await elementsSource.getAll())
+    .filter(isInstanceElement)
+    .filter(instance => instance.elemID.typeName === GROUP_PUSH_TYPE_NAME)
+    .reduce<Record<string, { application: string; groupPushInstance: InstanceElement }[]>>(
```
shir-reifenberg, line 43: personally I don't like reduce cause I think it's harder to read. so I have a suggestion that I think might simplify code here a little (but feel free to ignore)

you can use _.groupBy to group `GroupPush` by parent application, and then when iterating on GroupPush changes, check if group already exist

something like 

```
  const groupPushInstances = await awu(await elementsSource.getAll())...
  const groupPushByApp = _.groupBy(groupPushInstances, groupPush => getParent(groupPush).elemID.getFullName()
  
  return groupPushAdditionChanges.map(groupPush => {
     // check if groupPushByApp[getParent(groupPush)] already has a groupPush that maps the same group and has a different name
  })
```

shir-reifenberg (approved):
I think we should verify things are references instead of using `?` , but rather then that LGTM
also added one suggestion to simplify code IMO but feel free to ignore

PR #5229, opened 2023-12-26
SALTO-5203: Handle deletion of Project with IssueLayout

shir-reifenberg:
resolve several issues with deployment of IssueLayout

---

1. resolved circular dependency between `Project` and `IssueLayout` caused by reference from `IssueLayout` to project and a dependency changer.
2. handle deployment of `Project` with `IssueLayout` - if project already removed we mark `IssueLayout` as deployed. Other cases of deletion are not yet supported, added a comment with TODO ticket

---
_Release Notes_: 

_Jira adapter_:
- Fix bug caused when trying to remove Project with IssueLayout

---
_User Notifications_: 

_Jira adapter_:
- `projectId` field will be removed from IssueLayout instances


idoamit8 (approved):
Great Work ! 💯 
Thanks for the fix

 we could consider updating all issue layouts by project in one group. This way, we'd only need to make a single call instead of doing it for each change. But I'm open to your suggestions too!

omrilit (commented):
Should this modification be communicated to customers in advance? worth coordinating with @kami-delaroz  and @liorn as there are more noisy changes that are about to happen in Jira.

shir-reifenberg:
> Should this modification be communicated to customers in advance? worth coordinating with @kami-delaroz and @liorn as there are more noisy changes that are about to happen in Jira.

Added noise reduction, so no noise is expected

omrilit:
> > Should this modification be communicated to customers in advance? worth coordinating with @kami-delaroz and @liorn as there are more noisy changes that are about to happen in Jira.
> 
> @omrilit Added noise reduction, so no noise is expected

I mean for noisy changes as part of the changes that will appear after fetch.
For example, assume an EnvCompare between 2 envs that have a similar IssueLayout and the fetch ran only in one of them after your code changes. There will be changes that will be hard to understand for the user unless we communicated it to them.
It's relevant also for audit where users push the changes that occurred in the env over time, yet probably less relevant for Okta.
I think that we should consider whether we should couple those changes together to have a single communication message covering both, unless @liorn thinks that we can just ship it as is without proper communication.

PR #5230, opened 2023-12-26
SALTO-5208: Add missing reference strategy in Okta

shir-reifenberg:
We had some places where we didn't create missing references but allowed it in Okta

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


edenhassid (approved):
🔥 💯

PR #5231, opened 2023-12-27
SALTO-4978: workflow new API - add transition references - Jira

shayc331:
_workflow new API - add transition references_

---

in this pr I added: 
* filter that transforms conditions or validators parameters from "a,b,c" to ["a", "b", "c"] to create references from them
* reference mapping rules for conditions, validators and post functions

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


packages/jira-adapter/src/filters/workflowV2/transition_reference_filter.ts
```diff
@@ -0,0 +1,77 @@
...
+  if (parameters === undefined) {
+    return
+  }
+  Object.entries(parameters)
+    .filter(([, value]) => !_.isEmpty(value))
+    .forEach(([key, value]) => {
+      if (value === undefined || !listFields.has(key)) {
+        return
+      }
+      parameters[key as keyof (typeof parameters)] = convertIdsStringToList(value)
```
dantal4, line 46: I think that there are better options than this casting-
1. create a type guard for the key
2. replace the type of parameters to be `Values` or `Record<string, unknown>` / `Record<string, string | string[]`

packages/jira-adapter/src/filters/workflowV2/transition_reference_filter.ts
```diff
@@ -0,0 +1,77 @@
...
+}
+
+/*
+* This filter converts workflow transition validators and conditions parameters from a concatenated string
+* to a list of strings to create references
+*/
+const filter: FilterCreator = ({ config, fetchQuery }) => ({
+  name: 'workflowTransitionReferenceFilter',
+  onFetch: async (elements: Element[]) => {
+    if (!config.fetch.enableNewWorkflowAPI || !fetchQuery.isTypeMatch(JIRA_WORKFLOW_TYPE)) {
```
dantal4, line 57: I think that `fetchQuery.isTypeMatch(JIRA_WORKFLOW_TYPE)` is redundant - you already used it in `workflow_filter` to decide if the workflow instances should be fetched (which is correct), so here you can just use the `typeName` filter that you have in line 61

packages/jira-adapter/src/filters/workflowV2/transition_reference_filter.ts
```diff
@@ -0,0 +1,77 @@
...
+      return
+    }
+    elements
+      .filter(element => element.elemID.typeName === JIRA_WORKFLOW_TYPE)
+      .filter(isInstanceElement)
+      .forEach(instance => {
+        instance.value.transitions.forEach((transition: Transition) => {
```
dantal4, line 64: can `transitions` be undefined?

packages/jira-adapter/src/filters/workflowV2/transition_reference_filter.ts
```diff
@@ -0,0 +1,77 @@
...
+          transition.conditions?.conditions
+            ?.forEach(condition => {
```
dantal4, line 66: styling nit
```suggestion
          transition.conditions?.conditions?.forEach(condition => {
```

packages/jira-adapter/src/filters/workflowV2/types.ts
```diff
@@ -47,6 +47,41 @@ type WorkflowResponse = {
...
+  id?: string
+}
+
+
+type TransitionConditions = {
+  operation: string
+  conditions?: Condition[]
+}
+
+export type Transition = {
```
dantal4, line 80: I'm not sure what is the value of having all of those types, why not just dealing with the instance value as `Values`? And if we're using the types - should we run a type-guard to validate it? Are you planning to use them in other places too?

packages/jira-adapter/src/filters/workflowV2/transition_reference_filter.ts
```diff
@@ -0,0 +1,77 @@
...
+      parameters[key as keyof (typeof parameters)] = convertIdsStringToList(value)
+    })
+}
+
+/*
+* This filter converts workflow transition validators and conditions parameters from a concatenated string
+* to a list of strings to create references
+*/
+const filter: FilterCreator = ({ config, fetchQuery }) => ({
+  name: 'workflowTransitionReferenceFilter',
```
dantal4, line 55: maybe we should have a more specific name here, as this function doesn't really deal with references.... maybe "convert workflow parameters to lists" or something like that? (relevant for the file name too)

dantal4 (commented):
Great work!

packages/jira-adapter/src/filters/workflowV2/transition_reference_filter.ts
```diff
@@ -0,0 +1,77 @@
...
+  parameters: ConditionParameters | ValidatorParameters | undefined,
+  listFields: Set<string>
+): void => {
+  if (parameters === undefined) {
+    return
+  }
+  Object.entries(parameters)
+    .filter(([, value]) => !_.isEmpty(value))
+    .forEach(([key, value]) => {
+      if (value === undefined || !listFields.has(key)) {
```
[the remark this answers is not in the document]
shayc331: you are right about isEmpty, but typescript does not understand it

packages/jira-adapter/src/filters/workflowV2/transition_reference_filter.ts
shayc331: no, there is an initial transition for every workflow that cannot be deleted
Screen Shot 2023-12-28 at 11 44 42

packages/jira-adapter/src/filters/workflowV2/transition_reference_filter.ts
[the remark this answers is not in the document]
dantal4: so maybe get rid of `.filter()` and do all in this condition?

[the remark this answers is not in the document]
shayc331: your first suggestion works after the type changes, thanks!

packages/jira-adapter/src/filters/workflowV2/transition_reference_filter.ts
shayc331: changed it to `Record<string, string[]`, thanks!

packages/jira-adapter/src/filters/workflowV2/types.ts
shayc331: followed your suggestion, its `Values` now

packages/jira-adapter/src/filters/workflowV2/transition_reference_filter.ts
shayc331: renamed it to `workflowTransitionParametersFilter`

packages/jira-adapter/src/filters/workflowV2/transition_parameters_filter.ts
```diff
@@ -0,0 +1,72 @@
...
+import { Element, Values, isInstanceElement } from '@salto-io/adapter-api'
+import { FilterCreator } from '../../filter'
+import { JIRA_WORKFLOW_TYPE } from '../../constants'
+
+
+const VALIDATOR_LIST_FIELDS = new Set(['statusIds', 'groupsExemptFromValidation', 'fieldsRequired'])
+const CONDITION_LIST_FIELDS = new Set(['roleIds', 'groupIds', 'statusIds'])
+
+const convertIdsStringToList = (idOrIdList: string | string[]): string[] => {
+  if (_.isArray(idOrIdList)) {
```
dantal4, line 26: is that a valid value from the service? can it be returned as list? if it is, please add a test for that too

dantal4 (approved):
Great work 💯 
I have one last question 👇 🙏

packages/jira-adapter/src/filters/workflowV2/transition_parameters_filter.ts
shayc331: actually its not, the value returned from the service as a string and I convert it to a string[].
The check is there just to calm down typescript

dantal4: so it this case I think that `idOrIdList` should be only `string`, and `parameters` should be `Record<string, string>`. does it makes sense?

shayc331: I agree it is cleaner, but then I cant execute this [line](https://github.com/salto-io/salto/pull/5231/files#diff-bdb278ba22b3aa289b9d520cff39ad8ba56c9f6adb0eca4c4da144fef0f28bfdR42)

`parameters[key] = convertIdsStringToList(value)`
because parameters type would be `Record<string, string>`

dantal4: So we have a problem because we have different input and output types. I think that it makes more sense to set `parameters` as `Values` in this case. WDYT?

packages/jira-adapter/src/filters/workflowV2/transition_parameters_filter.ts
```diff
@@ -0,0 +1,67 @@
...
+
+const convertParameters = (
+  parameters: Values,
+  listFields: Set<string>
+): void => {
+  if (parameters === undefined) {
+    return
+  }
+  Object.entries(parameters)
+    .filter(([key, value]) => !_.isEmpty(value) && listFields.has(key))
```
dantal4, line 35: nit - add `_.isString(value)` (can be instead of `isEmpty`) to make it bug proof in case of an unexpected value.

dantal4 (approved):
💯

PR #5234, opened 2023-12-27
SALTO-5190 - Salesforce: Exponential backoff for data instances deploy

tomsellek:
Instead of always delaying by the same amount of time, let's have a pluggable strategy. If the user provides an explicit config, use the old constant-interval strategy with the provided value; if there is no explicit config, use an exponential-backoff strategy.

---

See the comment about how the `lowerdash` retry strategies are instantiated.

---
_Release Notes_: 
Salesforce: If no explicit configuration is provided, Salto will now use exponential backoff when retrying deployment of custom object instances.

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/client/client.ts
+++ b/packages/salesforce-adapter/src/client/client.ts
@@ -17,7 +17,7 @@ import _ from 'lodash'
 import { EOL } from 'os'
 import requestretry, { RequestRetryOptions, RetryStrategies, RetryStrategy } from 'requestretry'
 import Bottleneck from 'bottleneck'
-import { collections, decorators, hash } from '@salto-io/lowerdash'
+import { collections, decorators, hash, retry as lowerDashRetry } from '@salto-io/lowerdash'
 import {
   BatchResultInfo,
   BulkLoadOperation,
@@ -45,7 +45,6 @@ import { Options, RequestCallback } from 'request'
 import { AccountInfo, CredentialError, Value } from '@salto-io/adapter-api'
 import {
   CUSTOM_OBJECT_ID_FIELD,
-  DEFAULT_CUSTOM_OBJECTS_DEFAULT_RETRY_OPTIONS,
   DEFAULT_MAX_CONCURRENT_API_REQUESTS,
   SALESFORCE,
 } from '../constants'
@@ -66,6 +65,7 @@ import { mapToUserFriendlyErrorMessages } from './user_facing_errors'
 import { HANDLED_ERROR_PREDICATES } from '../config_change'
 
 const { makeArray } = collections.array
+const { exponentialBackoff, intervals: constantDelay } = lowerDashRetry.retryStrategies
 const { toMD5 } = hash
 
 const log = logger(module)
@@ -524,6 +524,20 @@ export const validateCredentials = async (
 
 type DeployProgressCallback = (inProgressResult: DeployResult) => void
 
+export type ClientDataRetry = Omit<CustomObjectsDeployRetryConfig, 'retryDelay'> & {
+  retryDelayStrategy: lowerDashRetry.RetryStrategy
+}
+
+export const DEFAULT_CUSTOM_OBJECTS_DEFAULT_RETRY_OPTIONS: ClientDataRetry = {
+  maxAttempts: 5,
+  retryDelayStrategy: exponentialBackoff({ initial: 1000 })(),
+  retryableFailures: [
+    'FIELD_CUSTOM_VALIDATION_EXCEPTION',
+    'UNABLE_TO_LOCK_ROW',
+  ],
+}
+
+
 export default class SalesforceClient {
   private readonly retryOptions: RequestRetryOptions
   private readonly conn: Connection
@@ -534,7 +548,7 @@ export default class SalesforceClient {
   private readonly setFetchPollingTimeout: () => void
   private readonly setDeployPollingTimeout: () => void
   readonly rateLimiters: Record<RateLimitBucketName, Bottleneck>
-  readonly dataRetry: CustomObjectsDeployRetryConfig
+  readonly dataRetry: ClientDataRetry
   readonly clientName: string
   readonly readMetadataChunkSize: Required<ReadMetadataChunkSizeConfig>
   private readonly filePropsByType: Record<string, FileProperties[]>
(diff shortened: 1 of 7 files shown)


packages/salesforce-adapter/src/constants.ts
```diff
@@ -311,14 +311,7 @@ export const DEFAULT_MAX_INSTANCES_PER_TYPE = 5000
 export const MINIMUM_MAX_ITEMS_IN_RETRIEVE_REQUEST = 500
 export const MAXIMUM_MAX_ITEMS_IN_RETRIEVE_REQUEST = 10000
 export const DEFAULT_ENUM_FIELD_PERMISSIONS = true
-export const DEFAULT_CUSTOM_OBJECTS_DEFAULT_RETRY_OPTIONS = {
```
tomsellek, line 314: This was getting a bit too complicated to keep in `constants.ts`, so I moved it to the only place it's used.

packages/salesforce-adapter/src/client/client.ts
```diff
@@ -524,6 +524,20 @@ export const validateCredentials = async (
 
 type DeployProgressCallback = (inProgressResult: DeployResult) => void
 
+export type ClientDataRetry = Omit<CustomObjectsDeployRetryConfig, 'retryDelay'> & {
+  retryDelayStrategy: lowerDashRetry.RetryStrategy
+}
+
+export const DEFAULT_CUSTOM_OBJECTS_DEFAULT_RETRY_OPTIONS: ClientDataRetry = {
+  maxAttempts: 5,
+  retryDelayStrategy: exponentialBackoff({ initial: 1000 })(),
```
tomsellek, line 533: Why do I need the extra call (`()`) at the end? Best I can figure the c`tor (which is just a call to `defaultOpts`) should return the `RetryStrategy` (which is a function) but looks like it actually returns a function that returns a `RetryStrategy`?

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -256,7 +256,13 @@ export const retryFlow = async (
     }
   }
 
-  await sleep(retryDelay)
+  const retryDelay = retryDelayStrategy()
```
amitgoldarb, line 259: I'm slightly confused by this. Why is this a function? Should it receive the retryAttempts as a parameter? How can this work?

packages/salesforce-adapter/src/client/client.ts
amitgoldarb: This seems a little bit complicated. I wonder if all we need to add to our config is multiplier and initialTimeout, instead of this more complex version.

amitgoldarb (commented):
Added a few comments. Let me know what you think 🙏

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
tomsellek: I'm using the existing infrastructure from `lowerdash`, so I'm kinda forced to use this interface. It increments the retry count on every call to the function (it's stored in the closure).
Funnily enough, my first implementation was hand-rolled and did take the retry count as a parameter.
Do you think it would be better to have a bespoke implementation here?

packages/salesforce-adapter/src/client/client.ts
tomsellek: Won't it break existing configs?

amitgoldarb: @tomsellek Why would it? It's okay IMO to change the system policy around our retry's, what we need to continue to do is respect existing persisited user preferences. Adding more configuration options should respect anything the user already has predefined.

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
amitgoldarb: Interestingly enough, I think that lowerdash is making things more complex here.. Maybe we can utilize it a bit differently so it's more readable

tomsellek:
@amitgoldarb can you please re-review?

packages/salesforce-adapter/src/client/client.ts
```diff
@@ -44,8 +44,7 @@ import { logger } from '@salto-io/logging'
 import { Options, RequestCallback } from 'request'
 import { AccountInfo, CredentialError, Value } from '@salto-io/adapter-api'
 import {
-  CUSTOM_OBJECT_ID_FIELD,
-  DEFAULT_CUSTOM_OBJECTS_DEFAULT_RETRY_OPTIONS,
+  CUSTOM_OBJECT_ID_FIELD, DEFAULT_CUSTOM_OBJECTS_DEFAULT_RETRY_OPTIONS,
```
amitgoldarb, line 47: I think we can probably revert both changes on this file

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -226,13 +228,21 @@ const isRetryableErr = (retryableFailures: string[]) =>
...
+    initial: client.dataRetry.retryDelay ?? DEFAULT_CUSTOM_OBJECT_DEPLOY_RETRY_DELAY,
+    multiplier: client.dataRetry.retryDelayMultiplier ?? DEFAULT_CUSTOM_OBJECT_DEPLOY_RETRY_DELAY_MULTIPLIER,
+  })()
+)
+
 export const retryFlow = async (
   crudFn: CrudFn,
   crudFnArgs: CrudFnArgs,
   retriesLeft: number,
+  retryDelayStrategy: retry.RetryStrategy = retryDelayStrategyFromConfig(crudFnArgs.client),
```
amitgoldarb, line 242: Does this recreate the parameter once per invocation or once overall? If it's option 2 then it should be inlined into the function.

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -256,7 +266,13 @@ export const retryFlow = async (
     }
   }
 
-  await sleep(retryDelay)
+  const retryDelay = retryDelayStrategy()
+  if (_.isNumber(retryDelay)) {
+    await sleep(retryDelay)
+  } else {
+    log.warn('Invalid delay %s', retryDelay)
+    await sleep(1000)
```
amitgoldarb, line 274: Nit: I wonder if in this weird case we should pick something from the config?

amitgoldarb (approved):
Looks good. Added a few comments
Approving to save cycles. I wonder if we should add UT to this?

packages/salesforce-adapter/src/client/client.ts
tomsellek: Ack

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
tomsellek: Ack

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
tomsellek: I will move it into the function just to make sure.

PR #5235, opened 2023-12-28
SALTO-3080 Fetch guide themes

adfineberg:
Add support Zendesk Guide Themes for fetch only, opt in with config flag.

---

---
_Release Notes_: 
__Zendesk Adapter:__
* Add Zendesk Guide Themes for fetch only, opt in by adding the brands to fetch the themes for in the config flag `themesForBrands` under `fetch -> guide`. So your config under guide might look like this `{ fetch: { guide: { brands: ['.*'], themesForBrands: ['.*'] } } }`. This allows users to differentiate between regular guide elements and the guide themes to fetch.

---
_User Notifications_: 


packages/zendesk-adapter/src/config.ts
```diff
@@ -2728,6 +2749,7 @@ export const GUIDE_TYPES_TO_HANDLE_BY_BRAND = [
   CATEGORY_ORDER_TYPE_NAME,
   SECTION_ORDER_TYPE_NAME,
   ARTICLE_ORDER_TYPE_NAME,
+  GUIDE_THEME_TYPE_NAME,
```
edenhassid, line 2752: didn't we say we are not adding this here because of the fetch?

edenhassid (commented):
looks good!! 🤩

packages/zendesk-adapter/src/filters/guide_theme.ts
```diff
@@ -0,0 +1,133 @@
...
+
+const unzipFolderToElements = async (buffer: Buffer, brandName: string, name: string): Promise<ThemeDirectory> => {
+  const zip = new JSZip()
+  const unzippedContents = await zip.loadAsync(buffer)
+
+  const elements: ThemeDirectory = {}
+
+  unzippedContents.forEach(async (relativePath, file) => {
+    if (!file.dir) {
+      const content = await file.async('nodebuffer')
```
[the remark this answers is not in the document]
adfineberg: Apparently the `Buffer` that nodejs has: https://stuk.github.io/jszip/documentation/api_zipobject/async.html#type-option

edenhassid (approved):
LGTM! maybe wait for @netama to also take a look before you merge :)

packages/zendesk-adapter/src/adapter.ts
```diff
@@ -534,15 +538,23 @@ export default class ZendeskAdapter implements AdapterOperations {
     }))
   }
 
+  private filterSupportedTypes(): Record<string, string[]> {
+    const isGuideEnabledInConfig = isGuideEnabled(this.userConfig[FETCH_CONFIG])
+    const isGuideThemeEnabledInConfig = isGuideThemeEnabled(this.userConfig[FETCH_CONFIG])
```
netama, line 543: nit: shouldn't this be Themes (and not Theme), at least in the config (which seems ok) and here? this is the name of the product, I think it should only be singular when looking at a specific theme

packages/zendesk-adapter/src/change_validators/guide_theme_readonly.ts
```diff
@@ -0,0 +1,37 @@
...
+    .filter(instance => instance.elemID.typeName === GUIDE_THEME_TYPE_NAME)
+
+  if (_.isEmpty(themeInstances)) {
+    return []
+  }
+
+  return themeInstances.map(instance => ({
+    elemID: instance.elemID,
+    severity: 'Error',
+    message: 'Cannot add this element because Guide Themes is currently under development.',
```
netama, line 34: nit: `add` is incorrect (this could be any type of change), maybe `deploy`? (or use the correct verb based on the change action)?

packages/zendesk-adapter/src/config.ts
```diff
@@ -77,32 +77,33 @@ export type IdLocator = {
 
 export type Guide = {
   brands: string[]
+  themes?: boolean
```
netama, line 80: 1. (commenting here but this is a general comment about the conditions)
I think it's worth discussing if Themes should or shouldn't require the "main" Guide elements to be fetched. 
from the current code it looks like it requires it - but there might be two separate use cases here (some times users may want to fetch only the "content" or only the theme, especially since both are heavy).
I think it might be worth having separate flags for fetching the guide content (but with a better name :) ), and themes - we could have the "content" flag on by default when missing to save some migration pains, since I assume it is more common. but there should be an option for users to disable it.
if this was discussed and the decision was that users should "exclude" the guide element then ok - though I have a feeling this could become more complex later.
also @edenhassid @amitgoldarb @tomermevorach 



2. please make sure to update `config_doc.md` and also `system_config_doc.md` with the changes

packages/zendesk-adapter/src/config.ts
```diff
@@ -77,32 +77,33 @@ export type IdLocator = {
...
-  handleIdenticalAttachmentConflicts?: boolean
-  greedyAppReferences?: boolean
-  appReferenceLocators?: IdLocator[]
-  guide?: Guide
-  resolveOrganizationIDs?: boolean
-  resolveUserIDs?: boolean
-  extractReferencesFromFreeText?: boolean
-  convertJsonIdsToReferences?: boolean
-}
+    enableMissingReferences?: boolean
```
netama, line 88: why change the indentation? the old one seems correct

packages/zendesk-adapter/src/filters/guide_arrange_paths.ts
```diff
@@ -142,10 +144,9 @@ const pathForBrandSpecificRootElements = (
   instance: InstanceElement,
   brandName: string | undefined,
   needTypeDirectory: boolean
-)
-: readonly string[] => {
+): readonly string[] => {
   if (brandName === undefined) {
-    log.error('brandName was not found for instance %s.', instance.elemID.getFullName())
+    log.warn('brandName was not found for instance %s.', instance.elemID.getFullName())
```
netama, line 149: is this a valid scenario now?

packages/zendesk-adapter/src/filters/guide_arrange_paths.ts
```diff
@@ -240,17 +241,18 @@ const filterCreator: FilterCreator = () => ({
...
       })
 
-    // category, settings, language_settings, category_order
+    // category, settings, language_settings, category_order, themes
     BRAND_SECOND_LEVEL
       .flatMap(type => guideGrouped[type])
       .filter(instance => instance !== undefined)
       .forEach(instance => {
-        const brandElemId = instance.value.brand?.elemID.getFullName()
+        const brandElemId = (instance.value.brand || instance.value.brand_id)?.elemID.getFullName()
```
netama, line 249: why not align and call it `brand` in the new instance as well?

packages/zendesk-adapter/src/filters/guide_locale.ts
```diff
@@ -78,7 +78,7 @@ const filterCreator: FilterCreator = ({ config }) => ({
         }
       }
     })
-    logsSet.forEach(message => log.error(message))
+    logsSet.forEach(message => log.warn(message))
```
netama, line 81: same question

packages/zendesk-adapter/src/filters/guide_theme.ts
```diff
@@ -0,0 +1,133 @@
...
+} from '../constants'
+import { FilterCreator } from '../filter'
+import { UNSORTED } from './guide_arrange_paths'
+import { download } from './guide_themes/download'
+
+const log = logger(module)
+const { awu } = collections.asynciterable
+
+type ThemeFile = { filename: string; content: StaticFile }
+type ThemeDirectory = { [key: string]: ThemeFile | ThemeDirectory }
```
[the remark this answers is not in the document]
netama: :+1:

packages/zendesk-adapter/src/filters/guide_theme.ts
```diff
@@ -0,0 +1,139 @@
...
+
+const addFileToDirectory = (root: ThemeDirectory, relativeFilename: string, file: ThemeFile): void => {
+  const pathSegments = relativeFilename.split('/')
+  const fileSegment = pathSegments.pop() as string // Remove and store the file segment, the array is never empty
```
netama, line 38: can we avoid the cast?

packages/zendesk-adapter/src/filters/guide_theme.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  // Use reduce to traverse and/or build the directory structure
```
netama, line 40: I think we do this in a few places already, @ori-moisis wdyt?

packages/zendesk-adapter/src/filters/guide_theme.ts
```diff
@@ -0,0 +1,139 @@
...
+ */
+const filterCreator: FilterCreator = ({ config, client }) => ({
+  name: 'guideThemesFilter',
+  onFetch: async elements => {
+    if (!isGuideEnabled(config[FETCH_CONFIG]) || !isGuideThemeEnabled(config[FETCH_CONFIG])) {
+      return undefined
+    }
+
+    const instances = elements.filter(isInstanceElement)
+    const guideThemes = instances.filter(i => i.elemID.typeName === GUIDE_THEME_TYPE_NAME)
```
netama, line 95: nit: 
```suggestion
    const guideThemes = instances.filter(inst => inst.elemID.typeName === GUIDE_THEME_TYPE_NAME)
```
(we try to avoid single-letter / unclear acronyms)

packages/zendesk-adapter/src/filters/guide_theme.ts
```diff
@@ -0,0 +1,139 @@
...
+      const brandName = fullNameByNameBrand[brandElemId]
+      if (brandName === undefined) {
+        log.info('brandName was not found for instance %s.', theme.elemID.getFullName())
+        return undefined
+      }
+      return brandName
+    }
+
+    const errors: SaltoError[] = []
+    await awu(guideThemes).forEach(async theme => {
```
netama, line 112: is there a reason to run these sequentially and not in parallel?

packages/zendesk-adapter/src/filters/guide_theme.ts
```diff
@@ -0,0 +1,139 @@
...
+      if (themeZip === undefined) {
+        errors.push(...addDownloadErrors(theme, downloadErrors))
+        remove(elements, element => element.elemID.isEqual(theme.elemID))
+        return
+      }
+      try {
+        const brandName = getBrandName(theme)
+        if (brandName === undefined) {
+          remove(elements, element => element.elemID.isEqual(theme.elemID))
+          return
```
netama, line 123: not sure I understand when this can happen, but if it can, shouldn't we check this before fetching the theme?

packages/zendesk-adapter/src/filters/guide_theme.ts
```diff
@@ -0,0 +1,139 @@
...
+        }
+        const themeElements = await unzipFolderToElements(themeZip, brandName, theme.value.name)
+        theme.value.files = themeElements
+      } catch (e) {
+        errors.push({
+          message: `Error fetching theme id ${theme.value.id}, ${(e as Error).message}`,
```
netama, line 129: can we avoid the cast? you can use a type guard

packages/zendesk-adapter/src/filters/guide_theme.ts
```diff
@@ -0,0 +1,139 @@
+/*
```
netama, line 1: nit: if this (and only this) is using the helper files from the `guide_themes` folder, why not nest it under it?

packages/zendesk-adapter/src/filters/guide_themes/api/createThemeExportJob.ts
```diff
@@ -0,0 +1,44 @@
...
+import { logger } from '@salto-io/logging'
+import ZendeskClient from '../../../client/client'
+import { DownloadJobData, PendingJob, isPendingJobResponse } from '../types'
+
+const log = logger(module)
+
+export const createThemeExportJob = async (
+  themeId: string, client: ZendeskClient
+): Promise<{ job: PendingJob<DownloadJobData> | undefined; errors: string[] }> => {
+  log.trace('Creating theme export job')
```
netama, line 26: probably worth adding some identifying info here (theme id / brand?)

packages/zendesk-adapter/src/filters/guide_themes/api/pollJobStatus.ts
```diff
@@ -0,0 +1,68 @@
...
+  return job.status === 'completed'
+}
+
+export const pollJobStatus = async (
+  jobId: string, client: ZendeskClient, interval = 5000, retries = 5
+): Promise<{ success: boolean; errors: string[] }> => {
+  log.trace('Polling job status')
+  try {
+    return {
+      success: await withRetry(
```
netama, line 54: not a blocker, but can we consolidate this with other polling usages (at least the other zendesk pone in `app_installations.ts`)?

netama (commented):
looks cool, added some comments and questions.

re the release notes - the suggested config is not accurate (since the user is likely to also want to have the `brands` configured in order to fetch Guide - and with the current implementation this is actually required so the suggested config cannot be used), probably worth rephrasing?

packages/zendesk-adapter/src/filters/guide_theme.ts
ori-moisis: not really sure what you mean when you say "this"

looking at only this part, very possible I am missing something, but it seems like we can just replace the whole thing with a single
```typescript
_.set(root, pathSegments, file)
```

packages/zendesk-adapter/src/filters/guide_theme.ts
adfineberg: Yes! Bad miss of mine! It happens because we bring all of the themes in the list API, so this is where we filter them

packages/zendesk-adapter/src/config.ts
adfineberg: I'm not sure I understand the separation, can we discuss f2f?

packages/zendesk-adapter/src/filters/guide_arrange_paths.ts
netama: the link is to workato so not relevant here (the brand itself is listed by its id field, and the `brand_id` in the input is coming from a workato recipe)
but anyway, up to you and @edenhassid

packages/zendesk-adapter/src/filters/guide_theme.ts
netama: one option is
```
const fileSegment = pathSegments.pop() ?? ''
```

packages/zendesk-adapter/src/filters/guide_theme.ts
netama: well, if it is possible for it to be something else then you should probably check for it explicitly and log the error and not push (or decide on a different error to push, or throw). and if it's an impossible scenario then you can also do the same thing (and should also log probably since it means something else broke). in practice I'd assume you want to return it without the more detailed part, if it doesn't contain a `.message`? and log including the call stack

packages/zendesk-adapter/src/filters/guide_arrange_paths.ts
adfineberg: Sorry, wrong link: https://github.com/salto-io/salto/blob/7da370b9cb5816dd5a88ee0bc0ba6a817ff5c929/packages/zendesk-adapter/src/filters/field_references.ts#L60

edenhassid: I think we should keep the `brand_id`

packages/zendesk-adapter/src/filters/guide_theme.ts
```diff
@@ -0,0 +1,121 @@
...
+    if (!isGuideEnabled(config[FETCH_CONFIG]) || !isGuideThemesEnabled(config[FETCH_CONFIG])) {
+      return undefined
+    }
+
+    const instances = elements.filter(isInstanceElement)
+    const guideThemes = instances.filter(instance => instance.elemID.typeName === GUIDE_THEME_TYPE_NAME)
+    const brands = getBrandsForGuideThemes(instances, config[FETCH_CONFIG])
+    const fullNameByNameBrand = _.mapValues(_.keyBy(brands, getFullName), 'value.name')
+    const getBrandName = (theme: InstanceElement): string | undefined => {
+      const brandElemId = theme.value.brand_id?.elemID.getFullName()
```
edenhassid, line 80: brand_id can be a number and not a referenceExpression

edenhassid (approved):

PR #5238, opened 2023-12-28
SALTO-5213 - Salesforce: Improve UX of deploy failure reason in case of metadata deploy and rollback on error

tomsellek:
We deploy with the `rollbackOnError` option, which means if any part of the deploy group fails, the entire group doesn't deploy. Let's make this more obvious to the user.

 - [x] @tomermevorach to review the error message
 - [ ] ~@tomermevorach to provide input on whether we want to keep the successful components if there's an error in the deploy group~
 - [x] Unit tests

---

The current code creates warnings for every successful component in a group where at least one component failed. It also does not return any successful components. We can do either one instead of both if that makes more sense. The current code implements both.
I will add UTs once we decide on what we want to do.

---
_Release Notes_: 
Salesforce: will now be more explicit about metadata types that did not deploy because they were in a deploy group along with types that failed to deploy because of an error.

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/metadata_deploy.ts
+++ b/packages/salesforce-adapter/src/metadata_deploy.ts
@@ -276,7 +276,7 @@ const processDeployResponse = (
     .map(message => getUnFoundDeleteName(message, deletionsPackageName))
     .filter(isDefined)
 
-  const successfulFullNames = anyErrors ? [] : allSuccessMessages
+  const successfulFullNames = (result.rollbackOnError && anyErrors) ? [] : allSuccessMessages
     .map(success => ({ type: success.componentType, fullName: success.fullName }))
     .concat(unFoundDeleteNames)
 
--- a/packages/salesforce-adapter/src/metadata_deploy.ts
+++ b/packages/salesforce-adapter/src/metadata_deploy.ts
@@ -279,7 +279,7 @@ const processDeployResponse = (
     .map(message => getUnFoundDeleteName(message, deletionsPackageName))
     .filter(isDefined)
 
-  const successfulFullNames = (!isCheckOnly && result.rollbackOnError && anyErrors) ? [] : allSuccessMessages
+  const successfulFullNames = allSuccessMessages
     .map(success => ({ type: success.componentType, fullName: success.fullName }))
     .concat(unFoundDeleteNames)
 
--- a/packages/salesforce-adapter/src/metadata_deploy.ts
+++ b/packages/salesforce-adapter/src/metadata_deploy.ts
@@ -202,7 +202,8 @@ const isUnFoundDelete = (message: DeployMessage, deletionsPackageName: string): 
 const processDeployResponse = (
   result: SFDeployResult,
   deletionsPackageName: string,
-  typeAndNameToElemId: Record<string, NameToElemIDMap>
+  typeAndNameToElemId: Record<string, NameToElemIDMap>,
+  isCheckOnly: boolean,
 ): { successfulFullNames: ReadonlyArray<MetadataId>
   errors: ReadonlyArray<SaltoError | SaltoElementError> } => {
   const allFailureMessages = makeArray(result.details)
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -244,6 +243,24 @@ const processDeployResponse = (
...
+    || testErrors.length > 0
+  if (result.rollbackOnError !== false
+    && anyErrors) {
+    // If we deployed with 'rollbackOnError' (the default) and any component in the group fails to deploy, then every
+    // component in the group will not deploy. Let's create an explicit error for the components that did not have
+    // errors to make it clear that they didn't deploy either.
+    makeArray(result.details)
+      .flatMap(detail => makeArray(detail.componentSuccesses))
+      .map(component => ({
+        elemID: typeAndNameToElemId[component.componentType]?.[component.fullName],
```
amitgoldarb, line 257: This part here is very gentle. I might add this at the end, and try to use the same function as isSuccessfulChange somehow.

packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -244,6 +243,24 @@ const processDeployResponse = (
...
+        message: 'This metadata type did not deploy because other types in its deployment group encountered failures',
```
amitgoldarb, line 258: @tomermevorach Can you please suggest some text here?
My suggestion:
```suggestion
        message: 'This element has failed to deploy due to other errors in salesforce metadata deployment and rollback on error setting turned on',
```

packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -244,6 +243,24 @@ const processDeployResponse = (
...
+        severity: 'Warning' as const,
```
amitgoldarb, line 259: I think this should be an error.
Also we should add type: 'DependencyError'

packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -260,7 +277,7 @@ const processDeployResponse = (
     .map(message => getUnFoundDeleteName(message, deletionsPackageName))
     .filter(isDefined)
 
-  const successfulFullNames = allSuccessMessages
+  const successfulFullNames = (result.rollbackOnError && anyErrors) ? [] : allSuccessMessages
```
amitgoldarb, line 280: I'm really not sure about this.
@tomermevorach I think we need to have a brief discussion here.

amitgoldarb (commented):
Looking great!
Added a few small comments

packages/salesforce-adapter/src/metadata_deploy.ts
tomsellek: Ack on the error type.
As to it being an error, won't it create a lot of noise that will make it difficult for the user to find the error that actually caused the issue?

packages/salesforce-adapter/src/metadata_deploy.ts
tomsellek: It's the same exact code we use when we build `componentErrors` - do you expect this to be more fragile in this context?

packages/salesforce-adapter/src/metadata_deploy.ts
tomsellek: @tomermevorach the only thing is that the user is not the one who sets the `rollbackOnError` option - it's something we do when we make API calls. I don't think the user necessarily knows what `rollbackOnError` means. Just let me know if you're still OK with mentioning `rollbackOnError` in the message.

packages/salesforce-adapter/src/metadata_deploy.ts
amitgoldarb: Warning and dependency looks good 👍

amitgoldarb (approved):
Approving this PR, please add UT and waiting for alignment on @tomermevorach 's exact message.

packages/salesforce-adapter/src/metadata_deploy.ts
tomermevorach: > @tomermevorach the only thing is that the user is not the one who sets the `rollbackOnError` option - it's something we do when we make API calls. I don't think the user necessarily knows what `rollbackOnError` means. Just let me know if you're still OK with mentioning `rollbackOnError` in the message.

You've got a point but I tend to think that mentioning it hints the user this default behavior can be turned off.
We should also make sure to mention this option [here](https://help.salto.io/en/articles/6927208-salto-for-salesforce-overview) so the user finds this article when searching rollbackOnError in our docs

tomsellek: @tomermevorach That's a Texas-size 10-4 big daddy

PR #5239, opened 2023-12-28
SALTO-5206 - Salesforce: Remove fields the user doesn't have permissions to deploy from applied changes

tomsellek:
If the user doesn't have access to a field (i.e. it's not creatable for add or updateable for modify), remove it from the instance during deploy.
~UTs still pending.~

---

1. Since we are now making the deploy code remove inaccessible fields, we could go all the way and remove the handling of inaccessible fields from `toRecord`. Would that make sense?
2. I had to use `forEach` instead of `map` when removing the fields because the same instance object is referenced from other instances - thus we can't clone+modify, we have to make the change in-place.

---
_Release Notes_: 
Salesforce: When deploying data instances with fields that are not creatable (for addition) or updateable (for modification) by the fetching user, the fields will now be deleted from the workspace.

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/e2e_test/utils.ts
+++ b/packages/salesforce-adapter/e2e_test/utils.ts
@@ -18,7 +18,7 @@ import {
   Value, ObjectType, ElemID, InstanceElement, Element, isObjectType, ChangeGroup, getChangeData, DeployResult,
   ProgressReporter,
 } from '@salto-io/adapter-api'
-import { filter, findElement } from '@salto-io/adapter-utils'
+import { filter, findElement, safeJsonStringify } from '@salto-io/adapter-utils'
 import { collections, values } from '@salto-io/lowerdash'
 import { MetadataInfo } from '@salto-io/jsforce'
 import { SalesforceRecord } from '../src/client/types'
@@ -191,7 +191,7 @@ export const createElement = async <T extends InstanceElement | ObjectType>(
   const result = await adapter.deploy({ changeGroup, progressReporter: nullProgressReporter })
   if (verify && result.errors.length > 0) {
     if (result.errors.length === 1) throw result.errors[0]
-    throw new Error(`Failed adding element ${element.elemID.getFullName()} with errors: ${result.errors}`)
+    throw new Error(`Failed adding element ${element.elemID.getFullName()} with errors: ${result.errors.map(error => safeJsonStringify(error))}`)
   }
   if (verify && result.appliedChanges.length === 0) {
     throw new Error(`Failed adding element ${element.elemID.getFullName()}: no applied changes`)
--- a/packages/salesforce-adapter/e2e_test/utils.ts
+++ b/packages/salesforce-adapter/e2e_test/utils.ts
@@ -189,8 +189,9 @@ export const createElement = async <T extends InstanceElement | ObjectType>(
     changes: [{ action: 'add', data: { after: element } }],
   }
   const result = await adapter.deploy({ changeGroup, progressReporter: nullProgressReporter })
-  if (verify && result.errors.length > 0) {
-    if (result.errors.length === 1) throw result.errors[0]
+  const errors = result.errors.filter(error => error.severity !== 'Warning')
+  if (verify && errors.length > 0) {
+    if (errors.length === 1) throw result.errors[0]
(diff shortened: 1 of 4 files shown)


packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -272,6 +272,21 @@ export const retryFlow = async (
   }
 }
 
+const removeFieldsWithNoPermission = async (
```
amitgoldarb, line 275: I think if we have this part, we don't need the logic in toRecord. There is no justification for it.

amitgoldarb (commented):
Looks good to me!
Just making sure we tested this out with real data instances where some fields were creatable/updatable and others were not and we receieved the output we expected.

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
amitgoldarb: cc @tamtamirr Do you agree?

amitgoldarb (commented):
Another thought, is that we should probably add a warning per element that was partially deployed explaining why this element was partially deployed. Otherwise this may be confusing to our users.

amitgoldarb:
@tomermevorach Can you help us with some text for the warning.

tomsellek:
> Another thought, is that we should probably add a warning per element that was partially deployed explaining why this element was partially deployed. Otherwise this may be confusing to our users.

@amitgoldarb Creating a warning is actually going to require a more significant change because the existing interfaces only pass around a message, and eventually we just assume the message represents an error (see https://github.com/salto-io/salto/blob/5a4eb0115e7a783ddb86029e94ab547abef10d69/packages/salesforce-adapter/src/custom_object_instances_deploy.ts#L81 )

tomermevorach:
> @tomermevorach Can you help us with some text for the warning.

How about:
Due to a lack of permissions, certain field values will be omitted from the deployment. Affected fields: [Field Names].

amitgoldarb:
> > Another thought, is that we should probably add a warning per element that was partially deployed explaining why this element was partially deployed. Otherwise this may be confusing to our users.
> 
> @amitgoldarb Creating a warning is actually going to require a more significant change because the existing interfaces only pass around a message, and eventually we just assume the message represents an error (see
> 
> https://github.com/salto-io/salto/blob/5a4eb0115e7a783ddb86029e94ab547abef10d69/packages/salesforce-adapter/src/custom_object_instances_deploy.ts#L81
> 
> )

I think we should change the plumbing to support this flow

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
amitgoldarb: We didn't remove the part from toRecord right? It's redundant no? It will never actually do anything.

amitgoldarb (approved):
Looks good. Approving to save cycles.
I think we can safely remove the logic from toRecord now that we have it somewhere else.

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
tomsellek: We didn't because we haven't heard @tamtamirr 's opinion yet.

packages/salesforce-adapter/e2e_test/utils.ts
```diff
@@ -189,9 +189,10 @@ export const createElement = async <T extends InstanceElement | ObjectType>(
     changes: [{ action: 'add', data: { after: element } }],
   }
   const result = await adapter.deploy({ changeGroup, progressReporter: nullProgressReporter })
-  if (verify && result.errors.length > 0) {
-    if (result.errors.length === 1) throw result.errors[0]
-    throw new Error(`Failed adding element ${element.elemID.getFullName()} with errors: ${result.errors}`)
+  const errors = result.errors.filter(error => error.severity !== 'Warning')
```
tamtamirr, line 192:
```suggestion
  const errors = result.errors.filter(error => error.severity === 'Error')
```
Otherwise we will face the same issue for `SaltoError` with severity `Info`

tomsellek: Ack

PR #5244, opened 2023-12-31
SALTO-5207: Salesforce Data Instance Unknown Fields Values Change Validator

tamtamirr:
Salesforce Data Instance Unknown Fields Values Change Validator

---

The added implementation creates a ChangeError on Data Instance Elements that has values of unknown fields to avoid partial deploy issue.

Here is the CV output for various types:

```
Please review the following messages; changes with errors will not be deployed.
    Error: Data instance has values of unknown fields in the following elements:
      salesforce.Product2.instance.01t8d000003FmGUAA0
        The Product "GC3020 GenWatt Propane 100kW" has values to the following unknown fields: [UnknownField__c]. Please include them in your deployment or remove these values from the instance.
      salesforce.Product2.instance.01t8d000003FmGaAAK
        The Product "GC5020 GenWatt Gasoline 300kWs" has values to the following unknown fields: [UnknownField1__c, UnknownField2__c]. Please include them in your deployment or remove these values from the instance.
      salesforce.SBQQ__CustomAction__c.instance.a0B8d000008r49tEAA
        The Custom Action (SBQQ) "View Discount Schedule" has values to the following unknown fields: [UnknownField1__c, UnknownField2__c]. Please include them in your deployment or remove these values from the instance.
```
 
---
_Release Notes_: 
Salesforce Adapter:
- Add Change Validator on data instances values of unknown fields.

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/change_validators/data_change.ts
```diff
@@ -13,30 +13,56 @@
...
- * Creates a ChangeError of type Info when one of the changes is on a data instance.
- */
+const createUnknownFieldValuesChangeError = (instance: InstanceElement): ChangeError | undefined => {
+  const typeFields = new Set(Object.keys(instance.getTypeSync().fields))
+  const unknownFields = Object.keys(instance.value)
+    .filter(fieldName => !typeFields.has(fieldName))
+  return unknownFields.length > 0
+    ? {
+      elemID: instance.elemID,
+      severity: 'Error',
```
amitgoldarb, line 43: I think we want this to be a warning. O/W this will block the deployment of such an instance which is far from the desired behaviour.

packages/salesforce-adapter/src/change_validators/data_change.ts
```diff
@@ -13,30 +13,56 @@
...
+      message: 'Data instance has values of unknown fields',
+      detailedMessage: `The ${aliasOrElemID(instance.getTypeSync())} "${aliasOrElemID(instance)}" has values to the following unknown fields: [${unknownFields.join(', ')}]. Please include them in your deployment or remove these values from the instance.`,
```
amitgoldarb, line 45: @tomermevorach WDYT?

amitgoldarb (commented):
Looks good! Added a few small comments

packages/salesforce-adapter/src/change_validators/data_change.ts
tamtamirr: Why? I think it's better to block this weird flow in first place. Users already proved to be ignoring CV warnings before. Is partial deploy the desired behavior?

amitgoldarb: You are right, I think it should be partial deploy as well (we don't have to couple the two to the same PR). We might also be able to ask @tomsellek who is working in a similar area to help us with this part.
Blocking the deploy literally means the customer can't deploy with salto, which is something we should be extremely careful when doing.

packages/salesforce-adapter/src/change_validators/data_change.ts
tomermevorach: I think this should be a warning and not error.
I will provide a message soon

tomermevorach: How about this message:
Some instance fields do not exist in the target environment, therefore their values will be omitted from the deployment. Missing fields: [Field Names].

packages/salesforce-adapter/src/change_validators/data_change.ts
```diff
@@ -13,30 +13,56 @@
...
+  if (dataChanges.length === 0) {
+    return []
+  }
+  changeErrors.push(createDataDeploymentChangeInfo(getChangeData(dataChanges[0]).elemID))
+  dataChanges
+    // Deletions are supported in this use-case
+    .filter(isAdditionOrModificationChange)
+    .map(getChangeData)
+    .map(createMissingFieldsValuesChangeError)
+    .filter(isDefined)
```
tomsellek, line 63: Don't we usually separate the detection of error from the creation of the error message?

packages/salesforce-adapter/src/change_validators/data_change.ts
```diff
@@ -13,30 +13,56 @@
...
   severity: 'Info',
   message: 'Data instances will be changed in deployment.',
   detailedMessage: '',
 })
 
-/**
- * Creates a ChangeError of type Info when one of the changes is on a data instance.
- */
+const createMissingFieldsValuesChangeError = (instance: InstanceElement): ChangeError | undefined => {
+  const typeFields = new Set(Object.keys(instance.getTypeSync().fields))
```
tomsellek, line 37: This will crash if the type has no fields. Is that a possibility?

tamtamirr: ObjectType always has the `fields` property

packages/salesforce-adapter/src/change_validators/data_change.ts
```diff
@@ -13,30 +13,56 @@
...
 
-  return isDefined(dataChange)
-    ? [createChangeError(getChangeData(dataChange).elemID)]
-    : []
+  if (dataChanges.length === 0) {
+    return []
+  }
+  changeErrors.push(createDataDeploymentChangeInfo(getChangeData(dataChanges[0]).elemID))
+  dataChanges
+    // Deletions are supported in this use-case
```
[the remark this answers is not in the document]
tamtamirr: Meant that deletions still work even if the instance contains values of non existing fields. I've updated my comment to avoid confusion

tomsellek (approved):
LGTM!

PR #5246, opened 2024-01-01
SALTO-5219: fix removal of missing ticket in ticket_field_ids in ticket_form

edenhassid:
fix removal of missing ticket in ticket_field_ids in ticket_form

---

---
_Release Notes_: 
zendesk:
* fix bug where the removal of a missing `ticket_field` by editing from `ticket_field_ids` list in `ticket_form` would fail the deploy

---
_User Notifications_: 

### DIFF
--- a/packages/zendesk-adapter/src/filters/ticket_form.ts
+++ b/packages/zendesk-adapter/src/filters/ticket_form.ts
@@ -138,8 +138,11 @@ const getChangeWithoutRemovedFields = (change: ModificationChange<InstanceElemen
   const removedFields = beforeFields.filter(field => !afterFields.has(field))
   // filtering out anything that is a reference since we can have a missing reference in the before that could have
   // been removed in the after and we don't want it to be added to the list again
-  const [finalRemovedFields, referenceFields] = _.partition(removedFields, field => !isReferenceExpression(field))
-  log.trace(`these reference fields are in the before and not in the after. their elemIds are: ${
+  const [finalRemovedFields, referenceFields] = _.partition(
+    removedFields,
+    field => _.isString(field) || _.isNumber(field),
+  )
+  log.debug(`these reference fields are in the before and not in the after. their elemIds are: ${
     referenceFields.filter(isReferenceExpression).map(ref => ref.elemID.getFullName())}`)
   if (_.isEmpty(finalRemovedFields)) {
     return undefined
--- a/packages/zendesk-adapter/src/filters/ticket_form.ts
+++ b/packages/zendesk-adapter/src/filters/ticket_form.ts
@@ -22,12 +22,12 @@ import {
   isAdditionOrModificationChange,
   isInstanceChange,
   isInstanceElement,
-  isModificationChange, isReferenceExpression, ModificationChange,
+  isModificationChange, ModificationChange,
   ReadOnlyElementsSource, ReferenceExpression, toChange,
 } from '@salto-io/adapter-api'
 import _ from 'lodash'
 import { logger } from '@salto-io/logging'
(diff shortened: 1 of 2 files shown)


packages/zendesk-adapter/src/filters/ticket_form.ts
```diff
@@ -135,7 +135,7 @@ const getChangeWithoutRemovedFields = (change: ModificationChange<InstanceElemen
   const { after } = change.data
   const beforeFields: number[] = before.value.ticket_field_ids ?? []
   const afterFields = new Set(after.value.ticket_field_ids ?? [])
-  const removedFields = beforeFields.filter(field => !afterFields.has(field))
+  const removedFields = beforeFields.filter(field => !afterFields.has(field)).filter(field => _.isNumber(field))
```
netama, line 138: 1. probably worth adding a comment (and maybe even a debug log) here about why this is done?
2. if the goal is to not send `undefined` / `null`, I think it might be safer to filter-out these rather than filter-in numbers, because in some cases we've seen ids written as strings (though not sure in this specific scenario), wdyt?
3. cool that this works for this case - based on the function's documentation it looks like this was not supposed to work in some scenarios (because "zendesk does not allow removing field and condition at the same time") - at least in the past and for "valid" ids. are we sure this is always ok? might there be scenarios where we should still send the "missing" id in order for the API to work? (you made the previous change as well so you probably remember this better than me :) just asking because of the comment)

netama (approved):
added a couple of nits and a question, if this works for all scenarios then LGTM :)

packages/zendesk-adapter/src/filters/ticket_form.ts
edenhassid: 3. this is right before the request, and after the resolve, I don't think that its ok to send missing at all. If the missing is removed from the ids list but there is still a condition linked to it then this condition will have a missing reference and the CV will fail the deploy. when we get to this code we should not have any missing references in the change.

edenhassid: 2. the goal is not to send a missing reference, and in this case the api allows to send only numbers

netama: 2. (sorry edited with missing before I saw this comment) if there is a number as a string, we can convert it to a number and still send it, right? I think the concern here is that we can accidentally remove something the user defined due to quirks in the API (that we've seen in the past)
3. internally in Salto I agree - but this code seems to work around limitations of the API and making changes in two steps (SALTO-4633). so things blocked in the CV are less relevant here I think. might be easier IRL :)

packages/zendesk-adapter/src/filters/ticket_form.ts
```diff
@@ -135,7 +135,15 @@ const getChangeWithoutRemovedFields = (change: ModificationChange<InstanceElemen
   const { after } = change.data
   const beforeFields: number[] = before.value.ticket_field_ids ?? []
   const afterFields = new Set(after.value.ticket_field_ids ?? [])
-  const removedFields = beforeFields.filter(field => !afterFields.has(field))
+  // filtering out anything that is not a number since we can have a missing reference in the before that could have
+  // been removed in the after and we don't want it to be added to the list again
+  const removedFields = beforeFields.filter(field => !afterFields.has(field)).filter(field => {
+    if (!_.isNumber(field)) {
+      log.trace(`the field is in the before and not in the after and is not a number: ${inspectValue(field)}`)
```
netama, line 142: might be better to have one log for all the values combined? (can use `.partition` maybe)

PR #5251, opened 2024-01-01
SALTO-5223: replace workflowScheme statusMigration type - Jira

shayc331:
_replace workflowScheme statusMigration type_

---

While working on [upgrading Workflow API](https://salto-io.atlassian.net/browse/SALTO-4978) Im using the type statusMigration from the Swagger.
This typeName collides with another type that we create manually [here](https://github.com/salto-io/salto/blob/main/packages/jira-adapter/src/filters/workflow_scheme.ts#L302). 
In this PR I am replacing the type that we created with an identical type from the Swagger for solving the collision.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/jira-adapter/src/config/api_config.ts
+++ b/packages/jira-adapter/src/config/api_config.ts
@@ -1397,6 +1397,9 @@ const DEFAULT_TYPE_CUSTOMIZATIONS: JiraApiConfig['types'] = {
 
   WorkflowScheme: {
     transformation: {
+      fieldTypeOverrides: [
+        { fieldName: 'statusMigrations', fieldType: 'List<StatusMapping>' },
+      ],
       fieldsToHide: [
         {
           fieldName: 'id',
--- a/packages/jira-adapter/src/filters/workflow_scheme.ts
+++ b/packages/jira-adapter/src/filters/workflow_scheme.ts
@@ -371,4 +371,3 @@ const filter: FilterCreator = ({ config, client, paginator, elementsSource }) =>
 })
 
 export default filter
-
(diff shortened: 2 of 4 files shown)


packages/jira-adapter/src/config/api_config.ts
```diff
@@ -1397,6 +1397,9 @@ const DEFAULT_TYPE_CUSTOMIZATIONS: JiraApiConfig['types'] = {
 
   WorkflowScheme: {
     transformation: {
+      fieldTypeOverrides: [
+        { fieldName: 'statusMigrations', fieldType: 'List<StatusMapping>' },
```
dantal4, line 1401: why do we need this mapping if the filter set the field manually?

dantal4 (commented):
LGTM 👍 💯 🥇 
We don't need noise suppression here because all types are hidden in jira, right?

packages/jira-adapter/src/config/api_config.ts
shayc331: without this, the core won't bring the type (StatusMapping) because its not a part of a GET response

dantal4 (approved):
💯

packages/jira-adapter/src/config/api_config.ts
shir-reifenberg: maybe I missed something, but why do we need the code in the filter if we already use `fieldTypeOverrides` ?

shayc331: Its not the renaming, I had issues with the `statusMigrations` deploy annotations which now is fixed, I removed the unnecessary code from the filter, please take a look

shir-reifenberg (approved):
🚀 🎸

PR #5253, opened 2024-01-02
SALTO-5225: Avoid changing hash when migrating state content

ori-moisis:
The previous implementation of the state content provider migration had the chance to change the content hash This was because "getContentAndHash" always uses the current salto version, which could cause the "migrate" to actually update the content

---

---
_Release Notes_: 
Core:
- Fixed issue in "salto workspace set-state-provider" where the state content hash could be modified in some cases

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/core/src/local-workspace/state/state.ts
+++ b/packages/core/src/local-workspace/state/state.ts
@@ -320,16 +320,26 @@ export const localState = (
       setDirty()
     },
     updateConfig: async args => {
+      const currentStateFiles = await currentContentProvider.findStateFiles(currentFilePrefix)
+      const contents = await awu(currentContentProvider.readContents(currentStateFiles))
+        .map(async ({ name, stream }): Promise<ContentAndHash> => {
+          const content = await getStream.buffer(stream)
+          return { account: path.basename(name), content, contentHash: toMD5(content.toString()) }
+        })
+        .toArray()
+
       const newProvider = getStateContentProvider(args.workspaceId, args.stateConfig)
-      const contents = await getContentAndHash()
 
-      const tempPrefix = path.join(path.dirname(currentFilePrefix), `.tmp_${path.basename(currentFilePrefix)}`)
-      await newProvider.writeContents(tempPrefix, contents)
+      if (contents.length > 0) {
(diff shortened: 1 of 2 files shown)


packages/core/test/workspace/local/state/state.test.ts
```diff
@@ -432,20 +441,28 @@ describe('localState', () => {
...
+    })
     it('should use the overriding files source and not the one from the content provider', async () => {
       await state.set(instanceWithStaticFile)
       await state.flush()
       expect(overridingStateFilesSource.flush).toHaveBeenCalledTimes(1)
       expect(contentProvider.staticFilesSource.flush).not.toHaveBeenCalled()
     })
     it('should keep the overriding static file source after config update', async () => {
+      const newContentProvider = mockContentProvider({})
+      getStateContentProviderSpy.mockReturnValueOnce(newContentProvider)
```
ori-moisis, line 464: unrelated fix, without this, the test would flush a real implementation of a `fileContentProvider` and create a file on the disk, which we don't want unit tests to do

packages/core/src/local-workspace/state/state.ts
```diff
@@ -320,16 +320,26 @@ export const localState = (
       setDirty()
     },
     updateConfig: async args => {
+      const currentStateFiles = await currentContentProvider.findStateFiles(currentFilePrefix)
```
yuribro, line 323: Don't we want to share this code with `syncQuickAccessStateData`? (Here we just calculate the hash, and there we continue to parse the content)

packages/core/src/local-workspace/state/state.ts
```diff
@@ -320,16 +320,26 @@ export const localState = (
...
+      const contents = await awu(currentContentProvider.readContents(currentStateFiles))
+        .map(async ({ name, stream }): Promise<ContentAndHash> => {
+          const content = await getStream.buffer(stream)
+          return { account: path.basename(name), content, contentHash: toMD5(content.toString()) }
```
yuribro, line 327: And maybe share this logic with `calculateContentAndHash`? These has to match, including the legacy `toString()`

PR #5255, opened 2024-01-02
SALTO-5152: Add Aliases to JSM

idoamit8:
In this PR I added aliases to JSM Types 

---

---
_Release Notes_: 
_Jira Adapter:_
JSM instances now have _alias fields. 

---
_User Notifications_: 
_Jira Adapter:_
JSM instances now have _alias fields.

### DIFF
--- a/packages/jira-adapter/src/filters/add_alias.ts
+++ b/packages/jira-adapter/src/filters/add_alias.ts
@@ -21,6 +21,7 @@ import {
 import { addAliasToElements, AliasData } from '@salto-io/adapter-components'
 import { logger } from '@salto-io/logging'
 import { FilterCreator } from '../filter'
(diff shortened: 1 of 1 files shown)


packages/jira-adapter/src/filters/add_alias.ts
```diff
@@ -79,6 +80,122 @@ const aliasMap: Record<string, AliasData> = {
...
+      {
+        fieldName: 'name',
+      }],
+    separator: ':',
+  },
+  [OBJECT_TYPE_TYPE]: {
+    aliasComponents: [
+      {
+        fieldName: 'parentObjectTypeId',
+        referenceFieldName: '_alias',
```
shir-reifenberg, line 181: IIUC if you count on other instance `alias`, then `OBJECT_TYPE_TYPE` should be part of `SECOND_ITERATION_TYPES`, (same  for `OBJECT_TYPE_ATTRIBUTE_TYPE`)

PR #5258, opened 2024-01-03
3 examples of CPQ configuration

pgonzaleznetwork:
_3 new examples on how to configure Salto for Salesforce CPQ with different mechanisms for creating the Salto ID_

---
_Release Notes_: 
_No effect, just better documentation._

---


amitgoldarb (approved):
Looks good!
Approving to save cycles, added a few smaller comments

PR #5260, opened 2024-01-03
SALTO-5230: Omit values from users response before logging

shir-reifenberg:
We fetch users data in order to convert userIds to emails. To avoid logging users data, we should omit it before writing to log.

---

`profile` object might include properties that may vary between account. we only use `login`, so the rest is omitted

---
_Release Notes_: 
None

---
_User Notifications_: 
None


omrilit (approved):
🏅

PR #5261, opened 2024-01-03
SALTO-4943: initiate getUsers early in fetch

shir-reifenberg:
This is part of an effort to improve fetch times in Okta

---

Until now, we fetched users data in filter. as responses are paginated (max page size 200), in envs with many users we need to make many requests to fetch this data, which we did one-by-one

when initiating getUsers call earlier in fetch flow, we can make API call while we fetch this.

measured with 2000 users and total fetch time improve in 30~ seconds

I only made this change in fetch, LMK if you think we should do the same for deploy

_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/user_utils.ts
+++ b/packages/okta-adapter/src/user_utils.ts
@@ -1,0 +1,62 @@
+/*
+*                      Copyright 2024 Salto Labs Ltd.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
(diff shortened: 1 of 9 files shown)


packages/okta-adapter/src/user_utils.ts
shir-reifenberg, line 1: copies as is from `user.ts`

omrilit (approved):
LGTM!
What is the suggestion to tackle it in deploy?
Anyway I think that this can be shipped without depending on improving the deploy.

shir-reifenberg:
@omrilit we can do it together with caching the results so we won't need to make the call for each change group, there's already a ticket

PR #5262, opened 2024-01-03
SALTO-5232: Improve performance of filtered plan

ori-moisis:
When calculating a plan that is filiterd by top level ID, perform the filtering before reading and deserializing the element

---

---
_Release Notes_: 
Core:
- Improved performance for calculating plan / env diff when the diff is filtered

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/cli/src/commands/env.ts
+++ b/packages/cli/src/commands/env.ts
@@ -115,7 +115,7 @@ export const diffAction: WorkspaceCommandAction<EnvDiffArgs> = async ({
     toEnv,
     hidden,
     state,
-    actualAccounts,
+    accounts === undefined ? undefined : actualAccounts,
     validSelectors,
   )
   outputLine(await formatEnvDiff(changes, detailedPlan, toEnv, fromEnv), output)
--- a/packages/core/src/core/plan/plan.ts
+++ b/packages/core/src/core/plan/plan.ts
@@ -314,12 +314,16 @@ const addDifferentElements = (
     }
     return elementPair
   }
-  const getFilteredElements = async (source: ReadOnlyElementsSource):
-    Promise<AsyncIterable<ChangeDataType>> =>
-    (awu(await source.getAll()).filter(async elem =>
-      _.every(await Promise.all(
-        topLevelFilters.map(filter => filter(elem.elemID))
-      )))) as AsyncIterable<ChangeDataType>
+  const getFilteredElements = async (
+    source: ReadOnlyElementsSource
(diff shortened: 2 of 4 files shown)


packages/core/src/core/plan/plan.ts
```diff
@@ -314,12 +314,17 @@ const addDifferentElements = (
...
-    (awu(await source.getAll()).filter(async elem =>
-      _.every(await Promise.all(
-        topLevelFilters.map(filter => filter(elem.elemID))
-      )))) as AsyncIterable<ChangeDataType>
+  log.debug('get plan filters len %d', topLevelFilters.length)
+  const getFilteredElements = async (
+    source: ReadOnlyElementsSource
+  ): Promise<AsyncIterable<ChangeDataType>> => (
+    topLevelFilters.length === 0
+      ? await source.getAll()
```
ori-moisis, line 322: when there are no filters and the remote map cache is empty, iterating over all elements with `.getAll` is slightly faster, so keeping this option here

packages/cli/src/commands/env.ts
```diff
@@ -115,7 +115,7 @@ export const diffAction: WorkspaceCommandAction<EnvDiffArgs> = async ({
     toEnv,
     hidden,
     state,
-    actualAccounts,
+    accounts === undefined ? undefined : actualAccounts,
```
ori-moisis, line 118: in order to take advantage of the quicker `getAll` code path we need to not pass in any filter, so if accounts were not specified, we don't pass any accounts anymore

adamro (approved):

packages/core/src/core/plan/plan.ts
```diff
@@ -314,12 +314,16 @@ const addDifferentElements = (
...
-      )))) as AsyncIterable<ChangeDataType>
+  const getFilteredElements = async (
+    source: ReadOnlyElementsSource
+  ): Promise<AsyncIterable<ChangeDataType>> => (
+    topLevelFilters.length === 0
+      ? await source.getAll()
+      : awu(await source.list())
+        .filter(async id => _.every(await Promise.all(topLevelFilters.map(filter => filter(id)))))
+        .map(id => source.get(id))
+  ) as AsyncIterable<ChangeDataType>
```
adamro, line 325: @ori-moisis do we have cases where doing source.get(id) for an id that is returned from list() returns undefined? Cause that would explain the `Cannot read properties of undefined (reading 'elemID')`, right?
Should we anyway filter undefined out? :\

PR #5265, opened 2024-01-04
SALTO-5121: Fix Exclude boards when no jira software projects 

idoamit8:
In this PR I support exclusion of type 'Board' 

---

Jira `Board` type is associated to `software project` only. 
Now we check if there are  `software projects`, and if there isn't - we exclude the boards. 

---
_Release Notes_: 
_Jira Adapter:_
The 'Board' type is now excluded when interfacing with no jira software projects. 

---
_User Notifications_: 
None


packages/jira-adapter/src/jsm_utils.ts
```diff
@@ -112,27 +112,38 @@ elementUtils.ducktype.EntriesRequester => async ({ paginator, args, typeName, ty
...
 
-export const getServerInfoTitle = async (client: JiraClient):
-Promise<string> => {
+const isProjectDataResponse = createSchemeGuard<ProjectDataResponse>(PROJECT_DATA_RESPONSE_SCHEMA)
+
+/*
+* Checks if the Jira service has a software project. The default is to assume that it does.
+* We are going to exclude Borads if we sure that there is no software project.
+*/
+export const hasSoftwareProject = async (client: JiraClient):
```
omrilit, line 135: Should this function be placed in JSM related code?

packages/jira-adapter/src/jsm_utils.ts
```diff
@@ -112,27 +112,38 @@ elementUtils.ducktype.EntriesRequester => async ({ paginator, args, typeName, ty
...
     const response = await client.getSinglePage({
-      url: '/rest/api/3/serverInfo',
+      url: '/rest/api/3/project/search',
     })
-    if (isServerInfo(response.data)) {
-      return response.data.serverTitle
+    if (!isProjectDataResponse(response.data)) {
+      return true
     }
+    return response.data.values.some(project => project.projectTypeKey === 'software')
```
omrilit, line 144: do we have a const for `software`?

packages/jira-adapter/src/adapter.ts
```diff
@@ -644,7 +644,7 @@ export default class JiraAdapter implements AdapterOperations {
     progressReporter.reportProgress({ message: 'Fetching types' })
     const { allTypes: swaggerTypes, parsedConfigs } = await this.getAllTypes(swaggers)
     const userConfigSupportedTypes = this.userConfig.apiDefinitions.supportedTypes
-    const shuldModifySupportedTypes = await getServerInfoTitle(this.client) === JIRA_SERVICE_DESK_FIELD
+    const shuldModifySupportedTypes = !(await hasSoftwareProject(this.client))
```
omrilit, line 647: nit: I think that this naming would be clearer
```suggestion
    const shouldOmitBoardSupportedType = !(await hasSoftwareProject(this.client))
```

omrilit (approved):
Looks great!
Approving to save cycle

packages/jira-adapter/src/jsm_utils.ts
idoamit8: Added :)

packages/jira-adapter/src/jsm_utils.ts
idoamit8: removed to utils.ts

PR #5270, opened 2024-01-05
SALTO-5233: Fetch referenced users only (Okta)

shir-reifenberg:
default `getUsers` in Okta will now fetch specific users instead of getting all users

---

1. Add config option to to determine how we fetch users - `allUsers` or `searchQuery`. in the future we might want to support `oneByOne` strategy (which will use a different endpoint without query params)
2. `searchQuery` will be the default strategy, and will only fetch user ids that appear in nacls (in both fetch and deploy)
3. left the option to use `allUsers` strategy, cause it might be more efficient in certain envs

for my env with (2000 users total, and 300 that appears in nacls), using the both strategies gave almost the same results for fetch times (with a little advantage to  `allUsers`)

---
_Release Notes_: 

_Okta_adapter_:
- Add config option to enable fetching only users referenced by other elements. This will be enabled by default and is more efficient for bigger accounts.

---
_User Notifications_: 

### DIFF
--- a/packages/okta-adapter/src/adapter.ts
+++ b/packages/okta-adapter/src/adapter.ts
@@ -14,7 +14,7 @@
 * limitations under the License.
 */
 import _ from 'lodash'
-import { Element, FetchResult, AdapterOperations, DeployResult, InstanceElement, TypeMap, isObjectType, FetchOptions, DeployOptions, Change, isInstanceChange, ElemIdGetter, ReadOnlyElementsSource, getChangeData, ProgressReporter } from '@salto-io/adapter-api'
+import { Element, FetchResult, AdapterOperations, DeployResult, InstanceElement, TypeMap, isObjectType, FetchOptions, DeployOptions, Change, isInstanceChange, ElemIdGetter, ReadOnlyElementsSource, getChangeData, ProgressReporter, isInstanceElement } from '@salto-io/adapter-api'
 import { config as configUtils, elements as elementUtils, client as clientUtils } from '@salto-io/adapter-components'
 import { applyFunctionToChangeData, logDuration, resolveChangeElement, restoreChangeElement } from '@salto-io/adapter-utils'
 import { logger } from '@salto-io/logging'
@@ -57,7 +57,7 @@ import groupPushFilter from './filters/group_push'
 import addImportantValues from './filters/add_important_values'
 import { APP_LOGO_TYPE_NAME, BRAND_LOGO_TYPE_NAME, FAV_ICON_TYPE_NAME, OKTA } from './constants'
 import { getLookUpName } from './reference_mapping'
-import { User, getUsers } from './user_utils'
+import { User, getUsers, getUsersFromInstances } from './user_utils'
 
 const { awu } = collections.asynciterable
 
@@ -277,8 +277,13 @@ export default class OktaAdapter implements AdapterOperations {
   @logDuration('fetching account configuration')
   async fetch({ progressReporter }: FetchOptions): Promise<FetchResult> {
     log.debug('going to fetch okta account configuration..')
-    const usersPromise = this.userConfig[FETCH_CONFIG]?.convertUsersIds ? getUsers(this.paginator) : undefined
+    const { convertUsersIds, getUsersStrategy } = this.userConfig[FETCH_CONFIG]
     const { elements, errors, configChanges } = await this.getAllElements(progressReporter)
+
+    const usersToFetch = getUsersFromInstances(elements.filter(isInstanceElement))
+    const usersPromise = convertUsersIds
+      ? getUsers(this.paginator, getUsersStrategy === 'searchQuery' ? { userIds: usersToFetch, property: 'id' } : undefined)
+      : undefined
 
     log.debug('going to run filters on %d fetched elements', elements.length)
     progressReporter.reportProgress({ message: 'Running filters for additional information' })
--- a/packages/okta-adapter/src/adapter.ts
+++ b/packages/okta-adapter/src/adapter.ts
@@ -280,9 +280,13 @@ export default class OktaAdapter implements AdapterOperations {
     const { convertUsersIds, getUsersStrategy } = this.userConfig[FETCH_CONFIG]
     const { elements, errors, configChanges } = await this.getAllElements(progressReporter)
 
-    const usersToFetch = getUsersFromInstances(elements.filter(isInstanceElement))
     const usersPromise = convertUsersIds
-      ? getUsers(this.paginator, getUsersStrategy === 'searchQuery' ? { userIds: usersToFetch, property: 'id' } : undefined)
+      ? getUsers(
+        this.paginator,
+        getUsersStrategy === 'searchQuery'
+          ? { userIds: getUsersFromInstances(elements.filter(isInstanceElement)), property: 'id' }
+          : undefined
+      )
       : undefined
 
     log.debug('going to run filters on %d fetched elements', elements.length)
--- a/packages/okta-adapter/src/change_validators/user.ts
+++ b/packages/okta-adapter/src/change_validators/user.ts
@@ -22,7 +22,7 @@ import { values } from '@salto-io/lowerdash'
 import { paginate } from '../client/pagination'
 import OktaClient from '../client/client'
 import { OktaConfig, FETCH_CONFIG } from '../config'
-import { getUsers, getUsersFromInstances, USER_MAPPING } from '../user_utils'
+import { DEFAULT_CONVERT_USERS_IDS_VALUE, getUsers, getUsersFromInstances, USER_MAPPING } from '../user_utils'
 
 const { isDefined } = values
 const { createPaginator } = clientUtils
@@ -34,7 +34,7 @@ const log = logger(module)
 export const usersValidator: (client: OktaClient, config: OktaConfig) =>
     ChangeValidator = (client, config) => async changes => {
       const { convertUsersIds, getUsersStrategy } = config[FETCH_CONFIG]
-      if (!(convertUsersIds ?? true)) {
+      if (!(convertUsersIds ?? DEFAULT_CONVERT_USERS_IDS_VALUE)) {
         log.debug('Skipped usersValidator because convertUsersIds config flag is disabled')
         return []
       }
--- a/packages/okta-adapter/src/change_validators/user.ts
+++ b/packages/okta-adapter/src/change_validators/user.ts
@@ -22,8 +22,7 @@ import { values } from '@salto-io/lowerdash'
 import { paginate } from '../client/pagination'
 import OktaClient from '../client/client'
 import { OktaConfig, FETCH_CONFIG } from '../config'
-import { USER_MAPPING } from '../filters/user'
-import { getUsers } from '../user_utils'
+import { getUsers, getUsersFromInstances, USER_MAPPING } from '../user_utils'
 
 const { isDefined } = values
 const { createPaginator } = clientUtils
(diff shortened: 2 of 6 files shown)


packages/okta-adapter/src/adapter.ts
```diff
@@ -277,9 +277,14 @@ export default class OktaAdapter implements AdapterOperations {
   @logDuration('fetching account configuration')
   async fetch({ progressReporter }: FetchOptions): Promise<FetchResult> {
     log.debug('going to fetch okta account configuration..')
-    const usersPromise = this.userConfig[FETCH_CONFIG]?.convertUsersIds ? getUsers(this.paginator) : undefined
+    const { convertUsersIds, getUsersStrategy } = this.userConfig[FETCH_CONFIG]
     const { elements, errors, configChanges } = await this.getAllElements(progressReporter)
 
+    const usersToFetch = getUsersFromInstances(elements.filter(isInstanceElement))
```
omrilit, line 283: `usersToFetch` can be calculated only when `convertUsersIds` is truthy & `getUsersStrategy === 'searchQuery'`, we can avoid calculating it otherwise.

packages/okta-adapter/src/change_validators/user.ts
```diff
@@ -34,7 +33,8 @@ const log = logger(module)
 */
 export const usersValidator: (client: OktaClient, config: OktaConfig) =>
     ChangeValidator = (client, config) => async changes => {
-      if (!(config[FETCH_CONFIG]?.convertUsersIds ?? true)) {
+      const { convertUsersIds, getUsersStrategy } = config[FETCH_CONFIG]
+      if (!(convertUsersIds ?? true)) {
```
omrilit, line 37: WDYT about extracting `true` to a const? e.g. `DEFAULT_CONVERT_USERS_IDS_VALUE`.
If we have other places that assume that the default value is true, I think that we should strive to use that const as well.
The main reason for my suggestion is that if in the future we'd like to modify the default we will have to do it in a single place and will not miss other places.

packages/okta-adapter/src/change_validators/user.ts
```diff
@@ -43,11 +43,10 @@ export const usersValidator: (client: OktaClient, config: OktaConfig) =>
         .filter(isInstanceChange)
         .map(getChangeData)
         .filter(instance => Object.keys(USER_MAPPING).includes(instance.elemID.typeName))
-        .filter(instance =>
-          USER_MAPPING[instance.elemID.typeName]
-            .some(path => resolvePath(instance, instance.elemID.createNestedID(...path)) !== undefined))
 
-      if (_.isEmpty(relevantInstances)) {
+      const usersToFetch = getUsersFromInstances(relevantInstances)
```
omrilit, line 47: same as my comment in `adapter.ts`, please calculate `usersToFetch` only when it's really in use.

packages/okta-adapter/src/change_validators/user.ts
```diff
@@ -56,7 +55,10 @@ export const usersValidator: (client: OktaClient, config: OktaConfig) =>
         paginationFuncCreator: paginate,
       })
 
-      const users = await getUsers(paginator)
+      const users = await getUsers(
+        paginator,
+        getUsersStrategy === 'searchQuery' ? { userIds: usersToFetch, property: 'profile.login' } : undefined
```
omrilit, line 60: Why does the fetch's `getUsersStrategy` affect the strategy in deploy?
I assume that in deploy the amount of users that were referenced in the to-be-deployed changes will always be much lower than all the users in the target account.
Don't you think that we should always fetch here only the referenced ones?
In addition, the fact that the `getUsersStrategy` is configured in the **fetch** config, implies that it shouldn't affect our strategy in deploy.

packages/okta-adapter/src/config.ts
```diff
@@ -1766,6 +1768,7 @@ export const DEFAULT_CONFIG: OktaConfig = {
     enableMissingReferences: true,
     includeGroupMemberships: false,
     includeProfileMappingProperties: true,
+    getUsersStrategy: 'searchQuery',
```
omrilit, line 1771: Is this default assumed somewhere else in the code? If so, it's the same comment as I wrote before, please extract to a const and reuse that value.

packages/okta-adapter/src/config.ts
```diff
@@ -1846,6 +1849,7 @@ export const configType = createMatchingObjectType<Partial<OktaConfig>>({
           enableMissingReferences: { refType: BuiltinTypes.BOOLEAN },
           includeGroupMemberships: { refType: BuiltinTypes.BOOLEAN },
           includeProfileMappingProperties: { refType: BuiltinTypes.BOOLEAN },
+          getUsersStrategy: { refType: BuiltinTypes.STRING },
```
omrilit, line 1852: should it be with `_restriction` to have only `'searchQuery'` or `'allUsers'` values?

packages/okta-adapter/src/filters/user.ts
```diff
@@ -104,16 +90,20 @@ const filterCreator: FilterCreator = ({ paginator, config, usersPromise }) => {
       })
     },
     preDeploy: async (changes: Change<InstanceElement>[]) => {
-      if (config[FETCH_CONFIG].convertUsersIds === false) {
+      const { convertUsersIds, getUsersStrategy } = config[FETCH_CONFIG]
+      if (convertUsersIds === false) {
```
omrilit, line 94: This implicitly assumes that the default value of `convertUsersIds` is `true`.
We can maybe be more explicit here.
Having said that, when writing this code it feels like overkill to me, so your call :)
```suggestion
      if (convertUsersIds === false || (convertUsersIds === undefined && DEFAULT_CONVERT_USERS_IDS_VALUE === false) {
```

packages/okta-adapter/src/filters/user.ts
```diff
@@ -104,16 +90,20 @@ const filterCreator: FilterCreator = ({ paginator, config, usersPromise }) => {
...
+      const usersToReplace = getUsersFromInstances(changes.map(getChangeData))
+
+      if (_.isEmpty(usersToReplace)) {
         return
       }
-      const users = await getUsers(paginator)
+      const users = await getUsers(
+        paginator,
+        getUsersStrategy === 'searchQuery' ? { userIds: usersToReplace, property: 'profile.login' } : undefined
+      )
```
omrilit, line 106: 1. same comments as in the CV (calculate `usersToReplace` only when needed and do not use the `getUsersStrategy` from the fetch config)
2. seeing that it's 99% same code as in the CV, maybe we can reuse the code?

packages/okta-adapter/src/user_utils.ts
```diff
@@ -13,23 +13,35 @@
...
+import { InstanceElement } from '@salto-io/adapter-api'
 import { collections } from '@salto-io/lowerdash'
 import { client as clientUtils } from '@salto-io/adapter-components'
+import { ACCESS_POLICY_RULE_TYPE_NAME, AUTHORIZATION_POLICY_RULE, GROUP_RULE_TYPE_NAME, MFA_RULE_TYPE_NAME, PASSWORD_RULE_TYPE_NAME, SIGN_ON_RULE_TYPE_NAME } from './constants'
 
 const log = logger(module)
 const { toArrayAsync } = collections.asynciterable
 const { makeArray } = collections.array
 
+const USER_CHUNK_SIZE = 200
```
omrilit, line 29: Do we want to allow the user to control it using a value in the adapter config?

packages/okta-adapter/src/user_utils.ts
```diff
@@ -43,20 +55,62 @@ export const areUsers = createSchemeGuard<User[]>(
...
+  [PASSWORD_RULE_TYPE_NAME]: [EXCLUDE_USERS_PATH],
+  [SIGN_ON_RULE_TYPE_NAME]: [EXCLUDE_USERS_PATH],
+  [MFA_RULE_TYPE_NAME]: [EXCLUDE_USERS_PATH],
+  [AUTHORIZATION_POLICY_RULE]: [INCLUDE_USERS_PATH],
+  EndUserSupport: [['technicalContactId']],
+}
+
+export const getUsersFromInstances = (instances: InstanceElement[]): string[] => _.uniq(
+  instances
+    .filter(instance => Object.keys(USER_MAPPING).includes(instance.elemID.typeName))
```
omrilit, line 73: Because this row is called for every instance, I wonder whether using a Set that is declared before the loop here would have some efficiency impact.
WDYT?

packages/okta-adapter/src/user_utils.ts
```diff
@@ -43,20 +55,62 @@ export const areUsers = createSchemeGuard<User[]>(
...
-      paginator(paginationArgs, page => makeArray(page) as clientUtils.ResponseValue[])
-    )).flat(),
-  'getUsers with pagination')
-  if (!areUsers(users)) {
-    return []
+
+  if (searchUsersParams) {
+    log.debug('getUsers called with searchQuery strategy')
+    const userChunks = _.chunk(searchUsersParams.userIds, USER_CHUNK_SIZE)
+    const allUsers = (await Promise.all(
```
omrilit, line 98: Are we fine with having all Promises opened together?
In case there are 1M users, it means that we can reach 50K open promises in parallel which will affect the performance.
I know that this is unlikely to have such amount of referenced users tho.
Maybe better to use `withLimitedConcurrency' to be on the safe side here?

omrilit (commented):
Great work!!

packages/okta-adapter/src/user_utils.ts
shir-reifenberg: I don't think so.. 200 is the maximal page size, any value above it will require more calls for the same chunk

packages/okta-adapter/src/user_utils.ts
shir-reifenberg: AFAIK we're already limited by the [client's rate limits](https://github.com/salto-io/salto/blob/main/packages/okta-adapter/src/client/client.ts#L165)

packages/okta-adapter/src/adapter.ts
shir-reifenberg: good point :) changed

packages/okta-adapter/src/change_validators/user.ts
shir-reifenberg: I think in this cause i'd like to check if `usersToFetch` is not empty before calling `getUsers` (no matter the strategy)

packages/okta-adapter/src/config.ts
shir-reifenberg: Currently it's not assumed but added a const for future use

packages/okta-adapter/src/filters/user.ts
shir-reifenberg: re 1 - see https://github.com/salto-io/salto/pull/5270#discussion_r1443961010
re 2 - as this part of code includes exiting the flow (cv or filter) when there are no relevant instances, I think it we consolidating this would make it less readable 😅

packages/okta-adapter/src/change_validators/user.ts
shir-reifenberg: mmm I agree that during deploy we suppose to have less referenced users, but I think in small envs there's no reason to use `searchQuery` at all (it also slows down the request it self). I prefer it to be consistent with fetch but let's talk IRL

packages/okta-adapter/src/change_validators/user.ts
omrilit: Not sure I get your point here let's f2f it

packages/okta-adapter/src/user_utils.ts
omrilit: I know that we will limit the calls that are actually performing against Okta, yet we will have multiple promises in parallel.
I don't think that it's really an issue in this scale, so you can ignore it

omrilit (approved):
Looks great!!

PR #5271, opened 2024-01-05
SALTO-5217 refactor NS config & query files

dantal4:
Split `config.ts` & `query.ts` to several files under `src/config/`.
In addition, remove the config suggestion of "bin" type exclusion (it's been used for 3 months).

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


omrilit (approved):
Assuming that the only changes that were made are really what is described in the ticket's summary I am approving.
It's hard to follow thousands of lines changes :)
Please LMK if there is something specific that I should take a look at..

PR #5276, opened 2024-01-07
SAAS-9985: Add TopLevelFilter to preview API

dorsev:
Add TopLevelFilters to Preview API. this allows consumers to improve the performance of Plan calculation to consider only the elements we expect to have changes on and not the whole workspace.

---

---

_Release Notes_: 
None

---

_User Notifications_: 
None

### DIFF
--- a/packages/core/src/api.ts
+++ b/packages/core/src/api.ts
@@ -150,6 +150,7 @@ export const preview = async (
   accounts = workspace.accounts(),
   checkOnly = false,
   skipValidations = false,
+  topLevelFilters = [shouldElementBeIncluded(accounts)],
 ): Promise<Plan> => {
   const stateElements = workspace.state()
   const adapters = await getAdapters(
@@ -166,7 +167,7 @@ export const preview = async (
       ? {} : getChangeValidators(adapters, checkOnly, await workspace.errors()),
     dependencyChangers: defaultDependencyChangers.concat(getAdapterDependencyChangers(adapters)),
     customGroupIdFunctions: getAdapterChangeGroupIdFunctions(adapters),
-    topLevelFilters: [shouldElementBeIncluded(accounts)],
+    topLevelFilters,
     compareOptions: { compareByValue: true },
   })
 }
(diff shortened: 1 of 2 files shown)


packages/core/src/api.ts
```diff
@@ -150,6 +150,7 @@ export const preview = async (
   accounts = workspace.accounts(),
   checkOnly = false,
   skipValidations = false,
+  topLevelFilters = [shouldElementBeIncluded(accounts)],
```
ori-moisis, line 153: this means that if someone passes non-default top level filters, the `accounts` argument becomes irrelevant
I think that is confusing API, I think we should not do it like that

if we keep `accounts` as a separate argument (which, I assume we should, even if just for backwards compatibility), it has to have a meaning even if we pass in other filters

unless the long term plan here is go remove the accounts argument completely and force the external users to draft their own top level filter to implement account filtering...

dorsev: Good comment. I agree and I've refactored the code accordingly.

LMK if you had something else in mind

packages/core/src/api.ts
```diff
@@ -166,7 +167,7 @@ export const preview = async (
       ? {} : getChangeValidators(adapters, checkOnly, await workspace.errors()),
     dependencyChangers: defaultDependencyChangers.concat(getAdapterDependencyChangers(adapters)),
     customGroupIdFunctions: getAdapterChangeGroupIdFunctions(adapters),
-    topLevelFilters: [shouldElementBeIncluded(accounts)],
+    topLevelFilters: [...(topLevelFilters ?? []), shouldElementBeIncluded(accounts)],
```
ori-moisis, line 170: I think I'd put the accounts filter first, we know its implementation is simple and quick to run, whereas for the filters we get as input here we don't know how complex their logic would be.
so it would probably be better to let the "quick" filters run first so the subsequent filters don't have to be calculated on everything if not needed

ori-moisis (approved):

dorsev:
Thank you for the review @ori-moisis 🙏

PR #5277, opened 2024-01-07
SALTO-5243: Create Deploy Warning when attempting to deploy data instance with missing fields

tamtamirr:
Create Deploy Warning when attempting to deploy data instance with missing fields.

---

When users attempt to deploy data instances with values to fields that do not exist in the `CustomObject`, we warn the users that the instance was most likely partially deployed in Salesforce.

- The implementation does not intercept the applied changes, as we don't really know what was actually deployed to the service.

Screen Shot 2024-01-07 at 11 02 18

---
_Release Notes_: 
Salesforce Adapter:
- Add Deploy Warning when attempting to deploy Data Instance with fields that do not exist on the CustomObject type.

---
_User Notifications_: 
_None_


tamtamirr:
It seems like this PR broke e2es. Moving to Draft until the issues are resolved.

PR #5282, opened 2024-01-08
SALTO-5217 NS simple config

dantal4:
Add simple fields to NS adapter config - fields that should be used to easily set customized configuration.

- All simple fields are in the config type's `_important_values` annotation, ordered by their presentation order.
- The field's title is in the `_alias` annotation
- The field's description is in the `_description` annotation
  - Links are in a markdown format (`[Learn more](https://url...)`)
- The field's type should be inferred by the field's `refType`
- The field's suggested values are in the `_restriction.values` annotation
  - When `_restriction.enforce_value = false`, it means that the suggested values list is partial, and the adapter will receive also other values
  - There are some special values that the adapter defines: "All" means *include all* and translates to `.*`, "FileCabinet" translated to `['file', 'folder']`

---

TODO:
- [x] add validations
- [x] add tests

Next step will be to migrate the existing config by adding the simple fields, with values that match the config (e.g set `includeCustomRecords = ["All"]` if `fetch.include.customRecords = [{ name = ".*" }]`).

---
_Release Notes_: 
Netsuite Adapter:
- Add simple config

---
_User Notifications_: 

### DIFF
--- a/packages/netsuite-adapter/src/config/config_creator.ts
+++ b/packages/netsuite-adapter/src/config/config_creator.ts
@@ -79,7 +79,10 @@ const updatedFetchTarget = (
 }
 
 const includeFileCabinetFolders = (config: NetsuiteConfig): string[] =>
-  config.includeFileCabinetFolders?.map(folderName => `^${folderName}/.*`) ?? []
+  config.includeFileCabinetFolders?.map(
+    folderName =>
+      `^${folderName.startsWith('/') ? '' : '/'}${folderName}${folderName.endsWith('/') ? '' : '/'}.*`
+  ) ?? []
 
 const includeCustomRecords = (config: NetsuiteConfig): FetchTypeQueryParams[] => {
   if (config.includeCustomRecords === undefined || config.includeCustomRecords.length === 0) {
--- a/packages/netsuite-adapter/src/config/config_creator.ts
+++ b/packages/netsuite-adapter/src/config/config_creator.ts
@@ -21,7 +21,7 @@ import { CUSTOM_RECORD_TYPE, CUSTOM_SEGMENT, INACTIVE_FIELDS } from '../constant
 import { removeCustomRecordTypePrefix } from '../types'
 import { fileCabinetTypesNames } from '../types/file_cabinet_types'
 import { ClientConfig, CriteriaQuery, FETCH_PARAMS, FetchParams, FetchTypeQueryParams, InstanceLimiterFunc, NetsuiteConfig, NetsuiteQueryParameters, QueryParams } from './types'
-import { ALL_TYPES_REGEX, DATA_FILE_TYPES, DEFAULT_MAX_INSTANCES_PER_TYPE, DEFAULT_MAX_INSTANCES_VALUE, FILE_CABINET, INCLUDE_ALL, UNLIMITED_INSTANCES_VALUE } from './constants'
+import { ALL_TYPES_REGEX, DATA_FILE_TYPES, DEFAULT_MAX_INSTANCES_PER_TYPE, DEFAULT_MAX_INSTANCES_VALUE, FILE_CABINET, GROUPS_TO_DATA_FILE_TYPES, INCLUDE_ALL, UNLIMITED_INSTANCES_VALUE } from './constants'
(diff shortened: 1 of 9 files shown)


packages/netsuite-adapter/src/config/config_creator.ts
```diff
@@ -77,11 +78,84 @@ const updatedFetchTarget = (
...
+  }))
+}
+
+const excludeDataFileTypes = (config: NetsuiteConfig): string[] => {
+  if (config.includeDataFileTypes === undefined) {
+    return []
+  }
+  const dataFileTypesToInclude = config.includeDataFileTypes.join('|')
+  const dataFileTypesToExclude = Object.values(DATA_FILE_TYPES)
+    .filter(fileType => !(new RegExp(`\\b${fileType}\\b`, 'i')).test(dataFileTypesToInclude))
```
ori-moisis, line 124: This implementation is really confusing to me
I think if this is supposed to set the example for future configuration refactors we should rethink this approach.
I had a really hard time figuring out where is the code handles `DATA_FILE_TYPES_GROUPS`, the relation here is very implicit, only after I read the code and understood the details I realized that this is actually the thing that makes it work.
if this was not a PR that had a small number of changes, I don't know if I would have been able to find this...

I also think that this is too much complicated logic for a configuration field.
it feels strange that a configuration field basically accepts anything as a valid value and silently ignores most of the words there except for a few specific words which are not really documented anywhere

I am not saying this won't technically work, but I really really think this is not something we would want to maintain and replicate to other use cases

packages/netsuite-adapter/src/config/types.ts
```diff
@@ -618,6 +636,60 @@ const deployConfigType = configUtils.createUserDeployConfigType(
...
+        [CORE_ANNOTATIONS.DESCRIPTION]: 'Salto only includes referenced searches by default.'
+         + ' Turning this option on will make Salto fetch all public records.'
+         + ' [Learn more](https://help.salto.io/en/articles/customize-netsuite-config)',
+      },
+    },
+    includeCustomRecords: {
+      refType: new ListType(BuiltinTypes.STRING),
+      annotations: {
+        [CORE_ANNOTATIONS.ALIAS]: 'Include custom records',
+        [CORE_ANNOTATIONS.DESCRIPTION]: 'Salto will only fetch the custom records that are included in this list.'
```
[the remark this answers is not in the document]
dantal4: @rozenmanofer @mikichernyak  can you help rephrasing it? this description is taken from the Figma.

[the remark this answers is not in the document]
dantal4: talked with @mikichernyak . the texts are not final, but it's fine to have it now because it's not represented in the UI yet.

packages/netsuite-adapter/src/config/config_creator.ts
dantal4: What are the alternatives for this field? support only `DATA_FILE_TYPES_GROUPS`? support `DATA_FILE_TYPES_GROUPS` and individual data types from the list? what should be under the restriction annotation?

ori-moisis: personally I would not implement groups for this, I think users can configure individual suffixes, I really think that would be more convenient and clear.

if you strongly think the groups are important here, then there are a couple of options:
1. only groups are allowed and when you use the configuration you behave according to the group names, the configuration itself only knows about group names and has no knowledge of suffixes at all. in this case the restriction should only allow group names.
2. both groups and suffixes are allowed, this requires a bit of code similar to the code you wrote - if you see a group, you need to replace it with the suffixes it represents (without parsing the group name, the groups and their meaning are hard coded). in this case the restriction must contain all the groups and all the suffixes

I think option 2 is still on the border of being overly complicated for this use case...

dantal4: I think we can start with option 1 and maybe change it in the future.

omrilit: @dantal4 why do we stick with the groups here? due to the product requirement? I think we can reconsider it and remove that requirement. Of course after verifying with Ofer & Miki.

PR #5283, opened 2024-01-08
SALTO-5215 - Salesforce: Improve dependency error text by updating the group names

tomsellek:
Change group names appear in user-visible error messages, so let's make them easier to understand by the end user.

---

TODO:
 - [ ] Final text from @tomermevorach 

---
_Release Notes_: 
Salesforce: Errors and warnings that contain the names of change groups are now clearer.

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/test/group_changes.test.ts
+++ b/packages/salesforce-adapter/test/group_changes.test.ts
@@ -151,12 +151,12 @@ describe('Group changes function', () => {
         expect(changeGroupIds.get(addInstance.elemID.getFullName()))
           .toEqual(changeGroupIds.get(anotherAddInstance.elemID.getFullName()))
         expect(changeGroupIds.get(addInstance.elemID.getFullName()))
-          .toEqual(`Add data instances of type '${customObjectName}'`)
+          .toEqual(`Addition of data instances of type '${customObjectName}'`)
       })
 
       it('should have a separate group for diff type', () => {
         expect(changeGroupIds.get(differentAddInstance.elemID.getFullName()))
-          .toEqual(`Add data instances of type '${differentCustomObjectName}'`)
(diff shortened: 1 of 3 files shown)


packages/salesforce-adapter/test/group_changes.test.ts
```diff
@@ -151,12 +151,12 @@ describe('Group changes function', () => {
...
           .toEqual(changeGroupIds.get(anotherAddInstance.elemID.getFullName()))
         expect(changeGroupIds.get(addInstance.elemID.getFullName()))
-          .toEqual(`add_${customObjectName}_instances`)
+          .toEqual(`Add data instances of type '${customObjectName}'`)
       })
 
       it('should have a separate group for diff type', () => {
         expect(changeGroupIds.get(differentAddInstance.elemID.getFullName()))
-          .toEqual(`add_${differentCustomObjectName}_instances`)
+          .toEqual(`Add data instances of type '${differentCustomObjectName}'`)
```
amitgoldarb, line 159: If we add this group name to the dependency error in core/deploy/deploy_actions.ts:174, we receive a sentence that isn't complete.
I think this should be:
`addition/modifcation/removal of data instance of type`
cc @tomermevorach WDYT? as well

amitgoldarb (approved):
Approving to save cycles, I think we should edit the text a bit though cc @tomermevorach

packages/salesforce-adapter/test/group_changes.test.ts
amitgoldarb: The text from the error in core is:
Element was not deployed, as it depends on ${groupName} which failed to deploy

tomermevorach: I am fine with @amitgoldarb's suggestion

PR #5285, opened 2024-01-09
SALTO-5155: Deploy First JSM Project 

idoamit8:
In this PR:
1. modified CV for locked field, so JSM related field will pass it. (They are expected to be changes in those fields only in the first time you deploy JSM project).
2. added dependency between JSM project and JSM Fields so Projects will be deployed before the fields in order to create them in the service. 
3. Added support for deploy of locked field.

---

1. Will add tests after first review of the code and code logic. 
2. Added a [video](https://www.loom.com/share/e4ab33f164aa4d08bff442dfd808eb1f) of successful deployment

---
_Release Notes_: 
_Jira Adpater:_
Now support deploy of first JSM project. 

---
_User Notifications_: 
None


packages/jira-adapter/src/filters/fields/field_deployment_filter.ts
```diff
@@ -64,14 +82,33 @@ const filter: FilterCreator = ({ client, config }) => ({
...
 
+    const allFieldsFromService = _.isEmpty(relevantChanges) ? undefined : await getAllFields(paginator)
+    const errors: SaltoElementError[] = []
+
     const deployResult = await deployChanges(
       relevantChanges.filter(isInstanceChange),
-      change => deployField(change, client, config),
+      async change => {
+        const inst = getChangeData(change)
+        if (isAdditionChange(change) && inst.value.isLocked === true && allFieldsFromService !== undefined) {
```
omrilit, line 92: Can you please extract a const for `isLocked` and reuse it?

packages/jira-adapter/src/filters/fields/field_deployment_filter.ts
```diff
@@ -64,14 +82,33 @@ const filter: FilterCreator = ({ client, config }) => ({
...
+          if (allFieldsFromService[inst.value.name] !== undefined) {
+            log.debug(`Field ${getChangeData(change).value.name} was auto-generated in Jira, skipping deployment`)
+            return
+          }
+          log.debug(`Field ${inst.value.name} is locked, but was not auto-generated in Jira.`)
+          errors.push(createSaltoElementError({
+            message: `Field ${inst.value.name} is locked, but was not auto-generates in Jira.`,
```
omrilit, line 99: I think that this error is not clear enough for users.
Maybe Lior should be involved here?

packages/jira-adapter/src/filters/fields/field_deployment_filter.ts
```diff
@@ -64,14 +82,33 @@ const filter: FilterCreator = ({ client, config }) => ({
         && getChangeData(change).elemID.typeName === FIELD_TYPE_NAME
     )
 
+    const allFieldsFromService = _.isEmpty(relevantChanges) ? undefined : await getAllFields(paginator)
```
omrilit, line 85: I am afraid that this change may have performance implications.
I would start with calling `getAllFields` only when having fields that are about to be deployed and have the `isLocked === true`.

In addition, assuming that there would be limited amount of possible locked fields, is it possible to query only them?
i.e. identify the fields with `isLocked === true` and query Jira to verify whether they exist there.

omrilit (commented):
Looks good!
I assume UTs are WIP?

packages/jira-adapter/src/dependency_changers/jsm_project_to_jsm_field.ts
```diff
@@ -0,0 +1,68 @@
...
+export const jsmProjectToJsmFieldDependencyChanger: DependencyChanger = async changes => {
+  const instanceChanges = Array.from(changes.entries())
+    .map(([key, change]) => ({ key, change }))
+    .filter(({ change }) => isAdditionChange(change))
+    .filter(
+      (change): change is deployment.dependency.ChangeWithKey<AdditionChange<InstanceElement>> =>
+        isInstanceChange(change.change)
+    )
+  const relevantChanges = instanceChanges
+    .filter(change => isAdditionChange(change.change))
```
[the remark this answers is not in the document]
idoamit8: Indeed

packages/jira-adapter/src/filters/fields/field_deployment_filter.ts
idoamit8: @liorn I will be glad to hear your input.

omrilit:
Could you please respond to my last question?
"In addition, assuming that there would be limited amount of possible locked fields, is it possible to query only them?
i.e. identify the fields with isLocked === true and query Jira to verify whether they exist there."

This is the only concern I have here.
Anyway, I'd wrap it with `log.time`.

packages/jira-adapter/src/filters/fields/field_deployment_filter.ts
idoamit8: I changed it to be called only when trying to deploy locked field, it doesn't suppose to happen every time. 
There is a problem to call only the locked field because they have different field Ids in different envs. 
For example, the field 'time to resolution'  can be called 'custom_10044' and 'custom_10448' in two different envs. 
I will wrap it with log.time

omrilit: Can we add some info to the request so it'd return only "standard" fields and not custom ones? or only JSM fields?
I am afraid that this query could take a lot of time...

idoamit8: I'll check to see how can I improve this request.

packages/jira-adapter/src/filters/fields/field_deployment_filter.ts
liorn: Message: "Locked field ${inst.value.name} was supposed to be auto-generated by Jira when deploying a project, but was unexpectedly not created."

packages/jira-adapter/src/filters/fields/field_deployment_filter.ts
idoamit8: Found this endpont https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-fields/#api-rest-api-3-field-get which seems to return what we need. I did couple of manual tests and it seems working good. Added `log. time` as well.

packages/jira-adapter/src/filters/fields/field_deployment_filter.ts
```diff
@@ -16,32 +16,46 @@
...
+  }).required().unknown(true)
+).required()
+
+const isFieldValueResponse = createSchemeGuard<fieldValuesResponse>(FIELD_RESPONSE_SCHEME)
 
-  const fieldNameToId = Object.fromEntries(fieldValues.map(field => [field.name, field.id]))
+const getAllFields = async (client: JiraClient)
+: Promise<Record<string, string>> => log.time(async () => {
+  const response = await client.getSinglePage({
+    url: '/rest/api/3/field',
```
omrilit, line 50: Why not using `/rest/api/3/field/search` with `type=system` query param?
https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-fields/#api-rest-api-3-field-search-get

idoamit8: I think that the new endpoint is great. Reduced the number of fields from 274 to 16, and returned the fields I expected it

packages/jira-adapter/src/filters/fields/field_deployment_filter.ts
```diff
@@ -13,17 +13,35 @@
...
 import { JiraConfig } from '../../config/config'
 
+const { toArrayAsync } = collections.asynciterable
+const { makeArray } = collections.array
 const { awu } = collections.asynciterable
+const log = logger(module)
+
+const getAllFields = async (paginator: clientUtils.Paginator): Promise<Record<string, string>> => {
+  const paginationArgs = {
+    url: '/rest/api/3/field/search?query=service',
```
omrilit, line 35: Isn't it too JSM specific while this filter is generic?
Maybe the `query` can hold a list of relevant fieldNames that we'd like to test whether they exist in the target account?

idoamit8: Currently only JSM fields will pass the CV. 
We generally don't want to allow to customers to deploy locked field. 
We can add more system fields if we see a use case.

omrilit: My main concern is that this is very implicit. nothing in the code indicates it, and in case tomorrow we will want some sort of similar behavior for other types of fields we wouldn't know to search for the `query=service` query param.
I think that some method renames would help to achieve that. WDYT?

idoamit8: Agreed. Will change

idoamit8: Changed methods names and made them a bit more general for future purposes.

packages/jira-adapter/src/change_validators/locked_fields.ts
```diff
@@ -37,7 +37,7 @@ export const lockedFieldsValidator: ChangeValidator = async changes => (
     .map(getChangeData)
     .filter(instance => instance.elemID.typeName === FIELD_TYPE_NAME)
     .filter(instance => instance.value.isLocked)
```
omrilit, line 39: Please use IS_LOCKED

omrilit (approved):
Great work!

PR #5289, opened 2024-01-09
SALTO-5261 Fix merge error on field type change in apply patch/changes

dantal4:
There is a merge error that happens when there are state-only changes in a type, and a field type changed too.

---

TODO:
- [x] add tests

---
_Release Notes_: 
Workspace:
- Fix merge error on field type change in apply patch/changes

---
_User Notifications_: 
None

### DIFF
--- a/packages/workspace/src/workspace/workspace.ts
+++ b/packages/workspace/src/workspace/workspace.ts
@@ -20,6 +20,7 @@ import {
   Value, toChange, isRemovalChange, getChangeData, isField, AuthorInformation, ReferenceInfo, ReferenceType,
   ReadOnlyElementsSource, isAdditionOrModificationChange, StaticFile, isInstanceElement, isObjectType,
   isModificationChange,
+  TypeReference,
 } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
 import {
@@ -878,7 +879,7 @@ export const loadWorkspace = async (
         modifiedFields
           .filter(field => element.fields[field.name] !== undefined)
           .forEach(field => {
-            element.fields[field.name].refType = field.refType
+            element.fields[field.name].refType = new TypeReference(field.refType.elemID)
           })
       })
   }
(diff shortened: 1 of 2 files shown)


packages/workspace/src/workspace/workspace.ts
```diff
@@ -849,6 +851,32 @@ export const loadWorkspace = async (
       return [toChange({ before, after })]
     }).toArray()
   }
+
+  const fixStateOnlyChangesFieldTypes = (
+    stateOnlyChanges: Change[],
+    changes: DetailedChange[],
+  ): void => {
+    const afterFieldsByParent = _.groupBy(
```
dantal4, line 859: @ori-moisis can you help me phrasing the comment that explains why this is a valid thing to do?

packages/workspace/src/workspace/workspace.ts
```diff
@@ -849,6 +851,32 @@ export const loadWorkspace = async (
...
+      changes
+        .filter(isAdditionOrModificationChange)
```
ori-moisis, line 861: I don't think we need to handle field addition
if a field was added there cannot be a conflict in its type because it doesn't exist in the state before

packages/workspace/src/workspace/workspace.ts
ori-moisis: the basic point to get across here is that when we get things from the state here, we want to get only the hidden parts.
for fields it is not possible to hide just the type of the field, so in theory the type of the field shouldn't be a part of the information (since the type is not hidden unless the whole field is hidden), but since we can't have a field without a type, we have to put something there...

using the type from the original change to make sure there is no conflict is a way to "remove" the information


something like that...

packages/workspace/src/workspace/workspace.ts
dantal4: I changed it to `isModificationChange`, for `changes` and `stateOnlyChanges`.

packages/workspace/src/workspace/workspace.ts
```diff
@@ -849,6 +852,37 @@ export const loadWorkspace = async (
...
+    stateOnlyChanges
+      .filter(isModificationChange)
+      .map(getChangeData)
+      .filter(isObjectType)
+      .forEach(element => {
+        const modifiedFields = modifiedFieldsByParent[element.elemID.getFullName()] ?? []
+        modifiedFields
+          .filter(field => element.fields[field.name] !== undefined)
+          .forEach(field => {
+            element.fields[field.name].refType = field.refType
```
ori-moisis, line 881: I think we might want to put a new type reference here with just the ID and no "resolved type"
otherwise we'd create a potentially dangerous mix where an element from one source (the state) points to an element from a different source (whatever the original change's source was)

ori-moisis (approved):
approving as the code looks ok

please add tests before you merge though

PR #5296, opened 2024-01-10
SALTO-2602: update bundle id table

Shulik95:
_Filling the bundleId to its elements record_

---

_The final part of SALTO-2602, The table is filled with almost all the bundles we currently have access to._

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


omrilit (approved):
LGTM!
This change has no effect yet on customers, right?
What are the next steps here?

Shulik95:
> LGTM! This change has no effect yet on customers, right? What are the next steps here?
@omrilit 
This does not affect users, addBundleId is behind a config flag. 
We can start opening the flag to users who have upgraded to the newest SuiteApp version.
I'll write an explanation on the ticket on how to open it and which bundles are supported so far.
To upgrade all of our users we are still waiting on the NS support.

PR #5298, opened 2024-01-10
SALTO 5270: Add Secuiry to workspaceId

idoamit8:
Currently, when customers have JSM Free but not JSM premium, they might get an errors for trying to access a source they don't have permission to. I wanted to add security that prevents the fetch or deployment from failing for not good reason. 

---

---
_Release Notes_: 
None.

---
_User Notifications_: 
None.

### DIFF
--- a/packages/jira-adapter/src/workspace_id.ts
+++ b/packages/jira-adapter/src/workspace_id.ts
@@ -18,6 +18,7 @@ import Joi from 'joi'
 import { logger } from '@salto-io/logging'
 import { createSchemeGuard } from '@salto-io/adapter-utils'
 import JiraClient from './client/client'
+import { JiraConfig } from './config/config'
 
 const log = logger(module)
 type WorkspaceResponse = {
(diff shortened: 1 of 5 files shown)


packages/jira-adapter/src/workspace_id.ts
```diff
@@ -34,13 +35,21 @@ const WORKSPACE_RESPONSE_SCHEME = Joi.object({
...
+      url: '/rest/servicedeskapi/assets/workspace',
+    })
+    if (!isWorkspaceResponse(response.data)) {
+      log.debug('Received invalid workspace response %o', response.data)
+      return undefined
+    }
+    return response.data.values[0].workspaceId
+  } catch (e) {
+    log.debug(`Failed to get workspace id: ${e}`)
     return undefined
```
omrilit, line 53: Could you please explain the try..catch reasoning here?
In case we fail here, wouldn't we like the operation to fail? can we proceed as if nothing happened?
In case we "swallow" the error, we should at least log an error/warn

idoamit8: It doesn't suppose to happen - 
But customers that have JSM Free and don't have JSM premium will call this endpoint add get `404 not found` of `401 Unauthorized`, but I am afraid that it will throw an error for some unknown reason and I don't want their fetch\deploy to fail since they don't use the workspaceId anyway.

omrilit (approved):

PR #5299, opened 2024-01-10
SALTO-5257: Adjust supported types for classic engine orgs (Okta)

shir-reifenberg:
Some types are not supported with the Classic Engine, we would like to avoid fetching them (currently we return fetch warnings)

---

---
_Release Notes_: 
_Okta_adapter_:
- Accounts with Okta Classic Engine will no longer get warnings when fetching Okta Identity Engine types fails.
---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/config.ts
+++ b/packages/okta-adapter/src/config.ts
@@ -1789,7 +1789,7 @@ export const getSupportedTypes = ({
   isClassicOrg: boolean
   supportedTypes: Record<string, string[]>
 }): Record<string, string[]> =>
-  (isClassicOrg ? _.omit(supportedTypes, CLASSIC_ENGINE_UNSUPPORTED_TYPES) : SUPPORTED_TYPES)
+  (isClassicOrg ? _.omit(supportedTypes, CLASSIC_ENGINE_UNSUPPORTED_TYPES) : supportedTypes)
 
 export type ChangeValidatorName = (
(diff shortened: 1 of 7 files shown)


packages/okta-adapter/src/config.ts
```diff
@@ -1778,6 +1779,18 @@ export const DEFAULT_CONFIG: OktaConfig = {
...
+  PROFILE_ENROLLMENT_POLICY_TYPE_NAME]
+
+export const getSupportedTypes = ({
+  isClassicOrg,
+  supportedTypes,
+}: {
+  isClassicOrg: boolean
+  supportedTypes: Record<string, string[]>
+}): Record<string, string[]> =>
+  (isClassicOrg ? _.omit(supportedTypes, CLASSIC_ENGINE_UNSUPPORTED_TYPES) : SUPPORTED_TYPES)
```
omrilit, line 1792: Shouldn't it be the following?
```suggestion
  (isClassicOrg ? _.omit(supportedTypes, CLASSIC_ENGINE_UNSUPPORTED_TYPES) : supportedTypes)
```

omrilit (approved):
Looks great! had only a single question

packages/okta-adapter/src/config.ts
shir-reifenberg: correct

PR #5300, opened 2024-01-11
SALTO-5274: Salesforce CustomField extra information feature

tamtamirr:
Implemented a feature that adds extra information on the fetched `CustomFields`.

---

Disabled by default. The current extraInformation I've added is for `defaultedOnCreate`
You can see diff on the opportunity object here as an example: https://github.com/salto-io/tamir-sf/commit/64e6bbeb40f41201ff2f4f75d6cccd1b9d52c491

---
_Release Notes_: 
Salesforce Adapter:
- Added the optional feature `extraCustomFieldInformation` that extends the annotations of the `CustomFields` when enabled.

---
_User Notifications_: 
Salesforce:
- When enabling the feature `extraCustomFieldInformation`, expect to see new values on each `CustomField`.


tomsellek (approved):
This is definitely too late to make this kind of change, but something to think about - would it make more sense to just let the user put a set of annotations in the config instead of one yes/no switch.

tamtamirr:
@tomsellek This feels like a premature optimization for me, we generally don't want to allow them to intercept the values, and possibly cause runtime errors.

PR #5301, opened 2024-01-11
SALTO-5271: Support label attribute

idoamit8:
In this PR:
1. Added CV that prevents the deletion of label attribute.
2. Added a filter that fetches and deploys the label attribute.
3. Modified the attribute deploy filter. 

---

None

---
_Release Notes_: 
_Jira Adapter_:
1. New instance element that holds for each objectType who its label attribute. 
2. New CV that prevents deletion of assets label attribute. 

---
_User Notifications_: 
_Jira Adapter_:
1. New instance element that holds for each objectType who its label attribute. 
2. New CV that prevents deletion of assets label attribute.

### DIFF
--- a/packages/jira-adapter/src/filters/assets/label_object_type_attribute.ts
+++ b/packages/jira-adapter/src/filters/assets/label_object_type_attribute.ts
@@ -102,7 +102,7 @@ const filterCreator: FilterCreator = ({ config, fetchQuery, client }) => ({
       }
       elements.push(labelAttributeInstance)
     })
-    // Remove label field field from attributes
+    // Remove label value from attributes
     objectTypeAttributeInstances.forEach(attribute => {
       delete attribute.value.label
     })
@@ -136,6 +136,8 @@ const filterCreator: FilterCreator = ({ config, fetchQuery, client }) => ({
     const deployResult = await deployChanges(
       relevantChanges.filter(isInstanceChange),
       async change => {
+        // We cover this case in a CV. Meaning that if we got here than the label attribute will be deleted
+        // With it's parent object type.
         if (isRemovalChange(change)) {
           return
         }
--- a/packages/jira-adapter/src/filters/assets/label_object_type_attribute.ts
+++ b/packages/jira-adapter/src/filters/assets/label_object_type_attribute.ts
@@ -1,0 +1,161 @@
+/*
+*                      Copyright 2024 Salto Labs Ltd.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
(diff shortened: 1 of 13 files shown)


packages/jira-adapter/src/filters/assets/label_object_type_attribute.ts
```diff
@@ -0,0 +1,161 @@
...
+        attributes,
+        labelAttributeType,
+        instanceNameToInstacne[parentName]
+      )
+      if (labelAttributeInstance === undefined) {
+        return
+      }
+      elements.push(labelAttributeInstance)
+    })
+    // Remove label field field from attributes
```
omrilit, line 105:
```suggestion
    // Remove label value from attributes
```

omrilit, line 105: Should we also remove the label field from the Attribute Type?

packages/jira-adapter/src/filters/assets/label_object_type_attribute.ts
```diff
@@ -0,0 +1,161 @@
...
+
+    const [relevantChanges, leftoverChanges] = _.partition(
+      changes,
+      change => getChangeData(change).elemID.typeName === OBJECT_TYPE_LABEL_ATTRIBUTE_TYPE
+    )
+    const workspaceId = await getWorkspaceId(client)
+    if (workspaceId === undefined) {
+      log.error(`Skip deployment of ${OBJECT_TYPE_LABEL_ATTRIBUTE_TYPE} types because workspaceId is undefined`)
+      const errors = relevantChanges.map(change => createSaltoElementError({
+        message: `The following changes were not deployed, due to error with the workspaceId: ${relevantChanges.map(c => getChangeData(c).elemID.getFullName()).join(', ')}`,
```
omrilit, line 127: Probably should be reviewed by Lior

packages/jira-adapter/src/filters/assets/label_object_type_attribute.ts
```diff
@@ -0,0 +1,161 @@
...
+      }))
+      return {
+        deployResult: { appliedChanges: [], errors },
+        leftoverChanges,
+      }
+    }
+    const deployResult = await deployChanges(
+      relevantChanges.filter(isInstanceChange),
+      async change => {
+        if (isRemovalChange(change)) {
```
omrilit, line 139: Could you please add a comment explaining that this is covered in the CV?
I wonder whether the CV relevant code should be written here as well (make sure that if the removal is allowed, there is also a removal of its parent Object Type), WDYT?

packages/jira-adapter/src/filters/assets/label_object_type_attribute.ts
```diff
@@ -0,0 +1,161 @@
...
+          return
+        }
+        const instance = getChangeData(change)
+        const attribute = isReferenceExpression(instance.value.labelAttribute)
+          ? instance.value.labelAttribute.value : undefined
+        if (attribute === undefined) {
+          return
+        }
+        await client.put({
```
omrilit, line 148: Is there a need to update also the other attribute that was previously marked as label?

omrilit (commented):
Looks great!
Could you please reason why you chose to model the label attribute as a dedicated instance and not just as a value on the attribute instance?

packages/jira-adapter/src/filters/assets/label_object_type_attribute.ts
idoamit8: No, it takes the label from it. (Automatically updated)

PR #5306, opened 2024-01-11
Catch error response on NS SOAP

dantal4:
Catch error response on NS SOAP and throw a more readable message.
I also refactored the `soap_client/schemas.ts`, to have much more DRY definitions.

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
```diff
@@ -751,6 +751,12 @@ export default class SoapClient {
       throw new Error(`VALIDATION_ERROR - Got invalid response from get all request. Errors: ${this.ajv.errorsText()}. Response: ${JSON.stringify(response, undefined, 2)}`)
     }
 
+    if (isGetAllErrorResponse(response)) {
+      const { code, message } = response.getAllResult.status.statusDetail[0]
+      log.error('Failed to run getAll request: %o', response)
+      throw new Error(`Failed to run getAll request: error code: ${code}, error message: ${message}`)
```
omrilit, line 757: This error will be user visible to customer.
I wonder whether it should be a bit more user friendly

omrilit (approved):
Looks great and thanks for the refactor!!

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
dantal4: We throw a similar message in all of the errors there. I agree that it can be improved, let's do it for all of them in another PR.

PR #5310, opened 2024-01-15
SALTO-5084 replace awu with promise-all

kami-delaroz:

---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/jira-adapter/src/filters/account_id/user_id_filter.ts
+++ b/packages/jira-adapter/src/filters/account_id/user_id_filter.ts
@@ -16,14 +16,11 @@
 */
 import { getChangeData, isAdditionOrModificationChange, isInstanceChange, isInstanceElement } from '@salto-io/adapter-api'
 import { walkOnElement } from '@salto-io/adapter-utils'
-import { collections } from '@salto-io/lowerdash'
 import _ from 'lodash'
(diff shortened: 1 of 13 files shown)


packages/jira-adapter/src/filters/account_id/user_id_filter.ts
```diff
@@ -63,15 +60,15 @@ const filter: FilterCreator = ({ client, config, getUserMapFunc, elementsSource
     if (userMap === undefined) {
       return
     }
-    await awu(elements)
+    await Promise.all(elements
```
omrilit, line 63: Isn't it risky to open promise per element? I think that in that magnitude of open promises we may reach some JS limits.

ori-moisis: in this specific example though there is really no point in using promises at all
there is nothing async happening...

kami-delaroz: thanks, I'll go over all changes with a deeper view :)
there are already a few examples of promise.all on elements, but we need to understand the risk here 
(this PR was supposed to be opened as draft anyway, I have a few more experiments I want to add)

PR #5311, opened 2024-01-15
SALTO-5283: convert requestForm properties to list - Jira

shayc331:
_convert requestForm properties to list_

---

requestForm properties can have an invalid field name as you can see in the attached image, the solution is to convert the field to a list in which each entry contains key and value (same as we do in [other places](https://github.com/salto-io/salto/blob/main/packages/jira-adapter/src/filters/workflow/workflow_properties_filter.ts#L92) in Jira).

before: 
Screen Shot 2024-01-15 at 16 35 01
after: 
Screen Shot 2024-01-15 at 16 33 50

---
_Release Notes_: 
None

---
_User Notifications_: 
Jira Adapter:
the properties of requestForm instances will be a list in which each entry will contain key and value

### DIFF
--- a/packages/jira-adapter/src/filters/layouts/request_types_layouts_to_values.ts
+++ b/packages/jira-adapter/src/filters/layouts/request_types_layouts_to_values.ts
@@ -14,7 +14,7 @@
 * limitations under the License.
 */
 
-import { CORE_ANNOTATIONS, Field, isInstanceElement, isObjectType } from '@salto-io/adapter-api'
+import { CORE_ANNOTATIONS, Field, InstanceElement, Values, isInstanceElement, isObjectType } from '@salto-io/adapter-api'
 import _ from 'lodash'
 import { logger } from '@salto-io/logging'
 import { FilterCreator } from '../../filter'
@@ -22,6 +22,15 @@ import { ISSUE_VIEW_TYPE, REQUEST_FORM_TYPE, REQUEST_TYPE_NAME } from '../../con
 
 const log = logger(module)
 const LAYOUT_TYPES_TO_ADJUST = [REQUEST_FORM_TYPE, ISSUE_VIEW_TYPE]
+
+const convertPropertiesToList = (instance: InstanceElement): void => {
+  instance.value.issueLayoutConfig.items.forEach((item: Values) => {
+    if (item.data?.properties !== undefined) {
+      item.data.properties = Object.entries(item.data.properties)
+        .map(([key, value]) => ({ key, value }))
+    }
+  })
+}
 
 /*
 * This filter is responsible for adding the requestForm and issueView fields to the requestType
@@ -62,6 +71,10 @@ const filter: FilterCreator = ({ config }) => ({
       delete layout.value.extraDefinerId
       delete layout.value.projectId
       if (layout.elemID.typeName === REQUEST_FORM_TYPE) {
+        if (layout.value.issueLayoutConfig?.items !== undefined
+          && _.isArray(layout.value.issueLayoutConfig.items)) {
+          convertPropertiesToList(layout)
+        }
         requestType.value.requestForm = layout.value
       } else {
         requestType.value.issueView = layout.value
--- a/packages/jira-adapter/src/filters/layouts/request_types_layouts_to_values.ts
+++ b/packages/jira-adapter/src/filters/layouts/request_types_layouts_to_values.ts
@@ -14,45 +14,47 @@
 * limitations under the License.
 */
 
-import { AdditionChange, Change, CORE_ANNOTATIONS, Field, InstanceElement, ModificationChange, Values, getChangeData, isAdditionOrModificationChange, isInstanceChange, isInstanceElement, isObjectType } from '@salto-io/adapter-api'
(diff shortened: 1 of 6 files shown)


omrilit:
@liorn @idoamit8 please consider whether we should notify customers with JSM enabled that there is such change to their configuration.
The user can notice that it happened as part of:
1. monitoring (@shayc331 did you check that there is no Notifications set to those envs?)
2. fetch summary (probably not interesting)
3. when comparing between envs that one of them was fetched with the new code and the other wasn't - they may see differences they will not understand
4. when deploying between envs that one of them was fetched with the new code and the other wasn't - the deploy may fail if deploying requestForm
5. audit (probably not interesting)

shayc331:
@omrilit 
re 1: 
I searched for envs with `enableJsm = true` that exists in `envNotification` as well. I found that there are 3 envs with monitoring and only one that has monitoring on this field (they have a monitor on `jira.*`). So there is one relevant customer, maybe we should just notify them.
re 4: 
IIUC `requestForm` is not deployable at the moment, it is inside `fieldToIgnore` list as you can see [here](https://github.com/salto-io/salto/blob/main/packages/jira-adapter/src/filters/request_type.ts#L47), @idoamit8 please correct me if I am wrong (thats why I didnt add preDeploy and onDeploy)

idoamit8:
@shayc331 It is in the list fields to ignore because it doesn't deploy through the requestType endpoint.
[here](https://github.com/salto-io/salto/blob/07b7e13e15249501c2f12476de885b90579736c6/packages/jira-adapter/src/filters/request_type.ts#L85) is the deployment function.

shayc331:
@idoamit8 thanks! I will add the relevant code

omrilit:
> @omrilit re 1: I searched for envs with `enableJsm = true` that exists in `envNotification` as well. I found that there are 3 envs with monitoring and only one that has monitoring on this field (they have a monitor on `jira.*`). So there is one relevant customer, maybe we should just notify them.

I think that `jira.*` will not catch **InstanceElements** that have changed. while `jira.*.instance.*` would.
We should either verify it or to ignore. worth making a decision with Lior probably

shayc331:
@omrilit they monitor `jira.*.instance.*` as well

omrilit:
> @omrilit they monitor `jira.*.instance.*` as well

Please consult Lior or the relevant CE whether to let them know or not. I don't think that it worth code changes in the noise reduction.

packages/jira-adapter/src/filters/layouts/request_types_layouts_to_values.ts
```diff
@@ -23,6 +23,37 @@ import { ISSUE_VIEW_TYPE, REQUEST_FORM_TYPE, REQUEST_TYPE_NAME } from '../../con
...
+}
+
+const isAdditionOrModificationRequestFormChange = (
+  change: Change
+): change is AdditionChange<InstanceElement> | ModificationChange<InstanceElement> =>
+  isInstanceChange(change)
+  && isAdditionOrModificationChange(change)
+  && change.data.after.elemID.typeName === REQUEST_TYPE_NAME
+  && change.data.after.value.requestForm?.issueLayoutConfig?.items !== undefined
+  && _.isArray(change.data.after.value.requestForm.issueLayoutConfig.items)
```
idoamit8, line 55: nit: There is some inconsistency between the function's name and its return values. The function does more than check for addition or modification changes. 
And it also returns true if the requestForm remains unchanged. 
Consider splitting it into two functions or relocating some of the checks to the changes themselves to improve readability.

packages/jira-adapter/src/filters/layouts/request_types_layouts_to_values.ts
```diff
@@ -23,6 +23,37 @@ import { ISSUE_VIEW_TYPE, REQUEST_FORM_TYPE, REQUEST_TYPE_NAME } from '../../con
...
+}
+
+const convertPropertiesToMap = (item: Values): void => {
+  item.data.properties = Object.fromEntries(
+    item.data.properties.map(({ key, value }: Values) => [key, value])
+  )
+}
+
+const convertProperties = (
+  issueLayoutConfig: Values,
```
idoamit8, line 38: nit: Why using type `Values` and not create a dedicated type if you just checked it before calling  `convertProperties` ?

idoamit8 (approved):
Great work ! 💯 
Loved the use of `Quack`
Added two suggestions, but approving to save cycle

packages/jira-adapter/src/filters/layouts/request_types_layouts_to_values.ts
shayc331: I split it, thanks

packages/jira-adapter/src/filters/layouts/request_types_layouts_to_values.ts
```diff
@@ -14,15 +14,48 @@
...
+    )
+  }
+}
+
+const convertProperties = (
+  issueLayoutConfig: IssueLayoutConfig,
+  convertFunc: (item: LayoutConfigItem) => void
+): void => {
+  issueLayoutConfig.items.forEach((item: LayoutConfigItem) => {
+    convertFunc(item)
```
idoamit8, line 47: nit: since you only have one line of shared code, IMO you can just have two separate functions.

PR #5314, opened 2024-01-16
Fix NS item types limiter exclusion

dantal4:
when `item` types reach the limit of max allowed instances, we should return the correct types names for the exclusion.

---

---
_Release Notes_: 
Netsuite Adapter:
- Fix NS item types limiter exclusion

---
_User Notifications_: 
None

### DIFF
--- a/packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
+++ b/packages/netsuite-adapter/src/client/suiteapp_client/soap_client/soap_client.ts
@@ -418,18 +418,25 @@ export default class SoapClient {
 
     const [itemTypes, otherTypes] = _.partition(types, isItemType)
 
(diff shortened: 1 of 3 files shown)


omrilit (commented):
Just to make sure I understand.
We will call a single search for all types together (as we do today), and will count their occurrences together and not per subtype.
If the sum of all subtypes will pass the limit, we will add an exclude section for each of the subtypes?

dantal4:
> Just to make sure I understand. We will call a single search for all types together (as we do today), and will count their occurrences together and not per subtype. If the sum of all subtypes will pass the limit, we will add an exclude section for each of the subtypes?

correct. we're already doing it, the change is to return the right type names.
we can also run a single query per item type and then the exclusion will be per type.

omrilit (approved):

PR #5317, opened 2024-01-16
SALTO-5206 - Salesforce: Warn when inaccessible fields are dropped from deploy

tomsellek:
Take 2 of https://github.com/salto-io/salto/pull/5239.

If the user doesn't have access to a field (i.e. it's not creatable for add or updateable for modify), remove it from the instance during deploy.
Only apply the logic for fields whose values changed, and that are neither hidden nor system fields.

---
Note that this logic is not identical to the code that removes fields in `toRecord`, because it is more selective in the fields it removes.

---
_Release Notes_: 
Salesforce: When deploying data instances with fields that are not creatable (for addition) or updateable (for modification) by the fetching user, the fields will now be deleted from the workspace and a warning will be issued.

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/custom_object_instances_deploy.ts
+++ b/packages/salesforce-adapter/src/custom_object_instances_deploy.ts
@@ -24,9 +24,9 @@ import {
   SaltoElementError,
   SaltoError,
   SeverityLevel,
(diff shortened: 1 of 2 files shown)


tamtamirr (approved):
LGTM

PR #5318, opened 2024-01-16
SALTO-5206 - Salesforce: Log when inaccessible fields are dropped from deploy

tomsellek:
This is a de-fanged version of https://github.com/salto-io/salto/pull/5317 where we only log the fields we would have removed.

---

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/custom_object_instances_deploy.ts
+++ b/packages/salesforce-adapter/src/custom_object_instances_deploy.ts
@@ -294,11 +294,22 @@ const removeFieldsWithNoPermission = async (
   instance: InstanceElement,
   permissionAnnotation: string
 ): Promise<SaltoElementError[]> => {
-  const shouldRemoveField = (type: ObjectType, fieldName: string, fieldValue: Value): boolean => (
-    fieldName !== CUSTOM_OBJECT_ID_FIELD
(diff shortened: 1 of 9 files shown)


packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -290,6 +292,53 @@ export const retryFlow = async (
...
+    }
+    return false
+  }
+  let namesOfFieldsThatChanged = Object.keys(getChangeData(instanceChange).value)
+  if (isModificationChange(instanceChange)) {
+    const [instanceBefore, instanceAfter] = getAllChangeData(instanceChange)
+    namesOfFieldsThatChanged = namesOfFieldsThatChanged
+      .filter(fieldName => instanceBefore.value[fieldName] !== instanceAfter.value[fieldName])
+  }
+  const instanceAfter = getChangeData(instanceChange)
```
tamtamirr, line 326: Can be simplified using `detailedCompare`. Here as an [example](https://github.com/salto-io/salto/blob/d085ee14ae3bc9d22129dd6bfa271bdc87b2df2d/packages/salesforce-adapter/src/filters/profile_deploy.ts#L50)

tamtamirr (approved):
Approving to save cycles. One small comment 🙏

PR #5319, opened 2024-01-16
SALTO-5287: Netsuite too many instances type exclusion description is not containing the type name

omrilit:

---

---
_Release Notes_: 
SALTO-5287: Netsuite too many instances type exclusion description is not containing the type name

---
_User Notifications_: 

### DIFF
--- a/packages/netsuite-adapter/src/config/suggestions.ts
+++ b/packages/netsuite-adapter/src/config/suggestions.ts
@@ -193,7 +193,7 @@ const updateConfigFromLargeTypes = (
       ...config.fetch,
       exclude: combineQueryParams(config.fetch.exclude, typesExcludeQuery),
     }
-    return excludedTypes
+    return excludedTypes.concat(failedCustomRecords)
   }
   return []
 }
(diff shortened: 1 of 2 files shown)


packages/netsuite-adapter/src/config/suggestions.ts
```diff
@@ -190,9 +193,9 @@ const updateConfigFromLargeTypes = (
       ...config.fetch,
       exclude: combineQueryParams(config.fetch.exclude, typesExcludeQuery),
     }
-    return true
+    return excludedTypes
```
dantal4, line 196: should we return `failedCustomRecords` too?

dantal4 (approved):
💯

packages/netsuite-adapter/src/config/suggestions.ts
omrilit: Nice catch, added this one too.
Please LMK if you think that we should rephrase the message to be more explicit for which of the elements is a "standard type" and which is CustomRecordType.
IMO it's good enough as is 🤷

PR #5320, opened 2024-01-17
SALTO-5295 reading properity of undefined fix (when deploying RequestTypeLayout)

omrifarkash:
When system is not defined we should use the name in lower case as the type

---

---
_Release Notes_: 
_Jira adapter_:
Fix reading system from undefined when deploying RequestTypeLayout

---
_User Notifications_: 
None


idoamit8 (approved):
Great! 💯 
Thanks for the fix 🥇

PR #5323, opened 2024-01-17
SALTO-5289 Jira adapter -  remove field reference from nacl

kami-delaroz:
SALTO-5289 remove field reference from nacl

---

[remove missing field configuration items in fix element](https://github.com/salto-io/salto/pull/5274)

---
_Release Notes_: 
_Jira adapter_:
* Fix oom issue on big fieldConfiguration by removing field references from nacl

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/field_configuration/replace_field_configuration_references.ts
+++ b/packages/jira-adapter/src/filters/field_configuration/replace_field_configuration_references.ts
@@ -47,10 +47,10 @@ const enrichFieldItem = async (
 ): Promise<EnrichedFieldItem | undefined> => {
   const elemId = new ElemID(JIRA, FIELD_TYPE_NAME, 'instance', fieldName)
   const fieldInstance = await elementSource.get(elemId)
-  // if (fieldInstance === undefined) {
-  //   log.debug(`Omitting element id ${elemId.getFullName()} since it does not exist in the account`)
-  //   return undefined
-  // } // TODO: is this needed ?
+  if (fieldInstance === undefined) {
+    log.debug(`Omitting element id ${elemId.getFullName()} since it does not exist in the account`)
+    return undefined
+  }
   return {
     id: new ReferenceExpression(elemId, fieldInstance),
     ...fieldItem,
--- a/packages/jira-adapter/src/filters/field_configuration/replace_field_configuration_references.ts
+++ b/packages/jira-adapter/src/filters/field_configuration/replace_field_configuration_references.ts
@@ -13,37 +13,86 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
-
 import _ from 'lodash'
-import { CORE_ANNOTATIONS, Field, getChangeData, InstanceElement, isInstanceElement, MapType, Values } from '@salto-io/adapter-api'
+import {
+  CORE_ANNOTATIONS,
+  ElemID,
+  Field,
+  getChangeData,
+  InstanceElement,
+  isInstanceElement,
+  MapType,
+  Values,
+  ReadOnlyElementsSource, ReferenceExpression,
+} from '@salto-io/adapter-api'
 import { isResolvedReferenceExpression } from '@salto-io/adapter-utils'
+import { collections, values } from '@salto-io/lowerdash'
+import { logger } from '@salto-io/logging'
 import { findObject } from '../../utils'
 import { FilterCreator } from '../../filter'
-import { FIELD_CONFIGURATION_TYPE_NAME } from '../../constants'
+import { FIELD_CONFIGURATION_TYPE_NAME, JIRA } from '../../constants'
+import { FIELD_TYPE_NAME } from '../fields/constants'
+import { FieldItem, isFieldConfigurationItems } from '../../weak_references/field_configuration_items'
+
+const log = logger(module)
+
+const { awu } = collections.asynciterable
+
+type EnrichedFieldItem = FieldItem & {id: ReferenceExpression}
+
+const enrichFieldItem = async (
+  fieldName: string,
+  fieldItem: FieldItem,
+  elementSource: ReadOnlyElementsSource,
+): Promise<EnrichedFieldItem | undefined> => {
+  const elemId = new ElemID(JIRA, FIELD_TYPE_NAME, 'instance', fieldName)
+  const fieldInstance = await elementSource.get(elemId)
+  // if (fieldInstance === undefined) {
+  //   log.debug(`Omitting element id ${elemId.getFullName()} since it does not exist in the account`)
+  //   return undefined
+  // } // TODO: is this needed ?
+  return {
+    id: new ReferenceExpression(elemId, fieldInstance),
+    ...fieldItem,
+  }
+}
 
 const replaceToMap = (instance: InstanceElement): void => {
-  instance.value.fields = _.keyBy(
-    instance.value.fields.filter((field: Values) => isResolvedReferenceExpression(field.id)),
-    field => field.id.elemID.name,
-  )
+  instance.value.fields = Object.fromEntries(instance.value.fields
+    .filter((field: Values) => isResolvedReferenceExpression(field.id))
+    .map((field: Values) => [
+      field.id.elemID.name,
+      _.omit(field, 'id'),
+    ]))
 }
 
-const replaceFromMap = (
+const replaceFromMap = async (
   instance: InstanceElement,
-): void => {
-  instance.value.fields = Object.values(instance.value.fields)
+  elementSource: ReadOnlyElementsSource,
+): Promise<void> => {
+  const fieldConfigurationItems = instance.value.fields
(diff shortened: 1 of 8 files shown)


packages/jira-adapter/src/filters/field_configuration/replace_field_configuration_references.ts
kami-delaroz, line 1: This is actually almost a revert of the changes introduced as part of the weak-references change.
a diff with the previous version:
Screen Shot 2024-01-17 at 17 57 39

packages/jira-adapter/src/weak_references/field_configuration_items.ts
```diff
@@ -13,40 +13,40 @@
...
+  [key: string]: FieldItem
 }
 
 const FIELD_CONFIGURATION_ITEM_SCHEME = Joi.object({
-  id: Joi.required(),
   description: Joi.optional(),
-  isHidden: Joi.boolean().required(),
-  isRequired: Joi.boolean().required(),
+  isHidden: Joi.boolean().optional(),
+  isRequired: Joi.boolean().optional(),
```
ori-moisis, line 43: is this supposed to be changed in this PR?

packages/jira-adapter/src/weak_references/field_configuration_items.ts
```diff
@@ -13,40 +13,40 @@
...
   renderer: Joi.optional(),
```
ori-moisis, line 44: unrelated to this change, but there is a mismatch between the typescript type (where there is a required field named `render` and this schema where there is an optional field named `renderer`)

packages/jira-adapter/src/weak_references/field_configuration_items.ts
```diff
@@ -57,12 +57,14 @@ const getFieldReferences = async (
     return []
   }
   return awu(Object.entries(fieldConfigurationItems))
-    .map(async ([id, config]) => (
-      isReferenceExpression(config.id)
-        ? { source: instance.elemID.createNestedID('fields', id, 'id'),
-          target: config.id.elemID,
-          type: 'weak' as const }
-        : undefined))
+    .map(([fieldName, _field]) => {
```
ori-moisis, line 60: if this is not async we should not use `awu` in the line above
also, looks like we are only using the key here, so we can iterate on `Object.keys` instead of `Object.entries`

packages/jira-adapter/src/weak_references/field_configuration_items.ts
```diff
@@ -57,12 +57,14 @@ const getFieldReferences = async (
...
+      const elemId = new ElemID(JIRA, FIELD_TYPE_NAME, 'instance', fieldName)
+      return {
+        source: instance.elemID.createNestedID('fields', fieldName, 'id'),
```
ori-moisis, line 63: this is setting the source on an element ID that does not exist (IIUC there is no `id` in that path)
the source should be something that exists...

packages/jira-adapter/src/weak_references/field_configuration_items.ts
```diff
@@ -13,40 +13,40 @@
...
-  isRequired: Joi.boolean().required(),
+  isHidden: Joi.boolean().optional(),
+  isRequired: Joi.boolean().optional(),
   renderer: Joi.optional(),
 })
 
 const FIELD_CONFIGURATION_ITEMS_SCHEME = Joi.object().pattern(/.*/, FIELD_CONFIGURATION_ITEM_SCHEME)
 
-const isFieldConfigurationItems = createSchemeGuard<FieldConfigurationItems>(FIELD_CONFIGURATION_ITEMS_SCHEME, 'Received an invalid field configuration items value')
+export const isFieldConfigurationItems = createSchemeGuard<FieldConfigurationItems>(FIELD_CONFIGURATION_ITEMS_SCHEME, 'Received an invalid field configuration items value')
```
ori-moisis, line 49: seems like every flow that uses the schema guard logs a warning on its own if it fails, in that case there is no point in passing an error message here as it would just cause double logging and in a more severe log level (internally it will be logged as `error`)

packages/jira-adapter/src/weak_references/field_configuration_items.ts
```diff
@@ -92,11 +102,8 @@ const removeMissingFields: WeakReferencesHandler['removeWeakReferences'] = ({ el
 
       const fixedInstance = instance.clone()
       fixedInstance.value.fields = Object.fromEntries(await awu(Object.entries(fieldConfigurationItems))
-        .filter(async ([_id, field]) =>
-          (isReferenceExpression(field.id)
-            // eslint-disable-next-line no-return-await
-            && await elementsSource.has(field.id.elemID)
-          )).toArray())
+        .filter(async ([fieldName, _field]) =>
+          fieldExists(fieldName, elementsSource)).toArray())
```
ori-moisis, line 106: no need to have `async` when there is no `await`
no need to specify additional arguments if they are no used...
```suggestion
        .filter(([fieldName]) => fieldExists(fieldName, elementsSource))
        .toArray())
```

also, this whole thing can be replaced with a much shorter and (IMO) more readable call to `pickAsync` from lowerdash

```typescript
fixedInstance.value.fields = await pickAsync(
  fieldConfigurationItems,
  (_field, fieldName) => fieldExists(fieldName),
)
```

packages/jira-adapter/src/weak_references/field_configuration_items.ts
ori-moisis: so update the type as well to say they are potentially `undefined`?

packages/jira-adapter/src/filters/field_configuration/replace_field_configuration_references.ts
```diff
@@ -53,29 +102,32 @@ const filter: FilterCreator = ({ config }) => ({
...
     }
 
-    changes
+    await awu(changes)
       .map(getChangeData)
       .filter(isInstanceElement)
       .filter(instance => instance.elemID.typeName === FIELD_CONFIGURATION_TYPE_NAME)
       .filter(instance => instance.value.fields !== undefined)
-      .forEach(replaceFromMap)
+      .forEach(async instance => replaceFromMap(instance, elementsSource))
```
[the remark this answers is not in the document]
ori-moisis: you don't have to write `async` before every function that returns a Promise.

I always suspected this, but this time I actually took the time to check and it seems that adding the redundant `async` actually has an effect on performance.
I tested this with a simple example:
```typescript
  const doSomethingAsync = (): Promise<void> => new Promise(resolve => setTimeout(resolve, 0))
  const callSomethingAsyncDirect = (): Promise<void> => doSomethingAsync()
  const callSomethingAsyncAsync = async (): Promise<void> => doSomethingAsync()
  
  await log.time(() => Promise.all(_.times(1000000).map(callSomethingAsyncDirect)), 'direct')
  await log.time(() => Promise.all(_.times(1000000).map(callSomethingAsyncAsync)), 'async')
```
(I ran this a few times and swapped which version ran first to make sure the difference isn't caused by the order of execution)
the results were pretty consistent with the redundant `async` version taking almost double the time:
```
2024-01-18T10:28:24.327Z debug salesforce-adapter async starting
2024-01-18T10:28:25.216Z debug salesforce-adapter async took 889 ms
2024-01-18T10:28:25.216Z debug salesforce-adapter direct starting
2024-01-18T10:28:25.671Z debug salesforce-adapter direct took 455 ms
```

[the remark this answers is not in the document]
kami-delaroz: interesting. thanks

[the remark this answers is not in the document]
kami-delaroz: but `await awu` is needed, right? to force waiting for the inner async function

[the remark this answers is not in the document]
ori-moisis: yes, there is still a promise there

packages/jira-adapter/src/weak_references/field_configuration_items.ts
```diff
@@ -13,68 +13,77 @@
...
  */
-const getFieldConfigurationItemsReferences: GetCustomReferencesFunc = async elements => log.time(() => awu(elements)
+const getFieldConfigurationItemsReferences: GetCustomReferencesFunc = async elements => log.time(() => elements
   .filter(isInstanceElement)
   .filter(instance => instance.elemID.typeName === FIELD_CONFIGURATION_TYPE_NAME)
-  .flatMap(getFieldReferences)
-  .toArray(), 'getFieldConfigurationItemsReferences')
+  .flatMap(getFieldReferences), 'getFieldConfigurationItemsReferences')
+
+const fieldExists = async (
```
ori-moisis, line 80: another `async` with no `await`
```suggestion
const fieldExists = (
```

ori-moisis (approved):
code LGTM
probably worth testing the behavior of the weak references before merging this to make sure it still works

PR #5324, opened 2024-01-17
SALTO-2392 fetch internal ids from SuiteQL tables

dantal4:
Fetch the internal ids & names of lots of data objects from SuiteQL tables and save them in hidden instances.
This will be used in the future for finding the real value behind an ACCOUNT_SPECIFIC_VALUE, in workflows & data elements.
In case of partial/quick fetch no new data will be fetched.
This feature is behind an adapter config value - `fetch.addSuiteQLTables`

---

TODO:
- [x] add tests

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/netsuite-adapter/src/adapter.ts
+++ b/packages/netsuite-adapter/src/adapter.ts
@@ -16,7 +16,7 @@
 import {
   FetchResult, AdapterOperations, DeployOptions, ElemIdGetter, ReadOnlyElementsSource, ProgressReporter,
   FetchOptions, DeployModifiers, getChangeData, isObjectType, isInstanceElement, ElemID, isSaltoElementError,
-  setPartialFetchData, InstanceElement, ObjectType, TypeElement,
+  setPartialFetchData, InstanceElement, ObjectType, TypeElement, ChangeDataType,
 } from '@salto-io/adapter-api'
 import _ from 'lodash'
 import { collections, values } from '@salto-io/lowerdash'
@@ -80,6 +80,7 @@ import { cloneChange } from './change_validators/utils'
 import { getChangeGroupIdsFunc } from './group_changes'
 import { getCustomRecords } from './custom_records/custom_records'
 import { getDataElements } from './data_elements/data_elements'
+import { getSuiteQLTableElements } from './data_elements/suiteql_table_elements'
 import { getStandardTypesNames } from './autogen/types'
 import { getConfigTypes, toConfigElements } from './suiteapp_config_elements'
 import { ImportFileCabinetResult } from './client/types'
@@ -373,9 +374,12 @@ export default class NetsuiteAdapter implements AdapterOperations {
         failures,
       },
       { elements: dataElements, requestedTypes: requestedDataTypes, largeTypesError: dataTypeError },
+      suiteQLTableElements,
     ] = await Promise.all([
       getStandardAndCustomElements(),
       getDataElements(this.client, fetchQuery, this.getElemIdFunc),
+      this.userConfig.fetch.addSuiteQLTables
+        ? getSuiteQLTableElements(this.client, this.elementsSource, isPartial) : [],
     ])
 
     progressReporter.reportProgress({ message: 'Running filters for additional information' })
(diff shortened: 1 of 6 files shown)


packages/netsuite-adapter/src/adapter.ts
```diff
@@ -373,9 +374,12 @@ export default class NetsuiteAdapter implements AdapterOperations {
         failures,
       },
       { elements: dataElements, requestedTypes: requestedDataTypes, largeTypesError: dataTypeError },
+      suiteQLTableElements,
     ] = await Promise.all([
       getStandardAndCustomElements(),
       getDataElements(this.client, fetchQuery, this.getElemIdFunc),
+      this.userConfig.fetch.addSuiteQLTables
```
omrilit, line 381: I think that `addSuiteQLTables` naming is super technical and will be hard to explain to customers what it actually does when we'll ask them to open it, and I think that we will have to gradually enable it to customer and not for all at once.

packages/netsuite-adapter/src/data_elements/types.ts
```diff
@@ -66,19 +65,17 @@ type TypeWithSingleFieldIdentifier = 'subsidiary' |
 export type SupportedDataType = ItemType | TypeWithMultiFieldsIdentifier | TypeWithSingleFieldIdentifier
 
 // Was taken from https://<account_id>.app.netsuite.com/app/help/helpcenter.nl?fid=section_n3432681.html&whence=
```
omrilit, line 67: is this comment still relevant?

packages/netsuite-adapter/src/data_elements/suiteql_table_elements.ts
```diff
@@ -0,0 +1,392 @@
...
+    existingInstance.refType = createRefToElmWithValue(suiteQLTableType)
+  }
+  if (isPartial) {
+    return existingInstance
+  }
+  const instance = existingInstance ?? new InstanceElement(tableName, suiteQLTableType)
+  instance.annotate({ [CORE_ANNOTATIONS.HIDDEN]: true })
+  if (instance.value[INTERNAL_IDS_MAP] === undefined) {
+    instance.value[INTERNAL_IDS_MAP] = {}
+  }
```
omrilit, line 349: nit: maybe all of the code after the `??` should be extracted to a method that creates the default instance?

packages/netsuite-adapter/src/data_elements/suiteql_table_elements.ts
```diff
@@ -0,0 +1,392 @@
...
+  customrecordtype: undefined,
+}
+
+const getInternalIdsMap = async (
+  client: NetsuiteClient,
+  tableName: string,
+  { internalIdField, nameField, lastModifiedDateField }: QueryParams,
+  lastFetchTime: Date | undefined,
+): Promise<Record<string, { name: string }>> => {
+  const whereLastModified = lastModifiedDateField !== undefined && lastFetchTime !== undefined
```
omrilit, line 308: I am a bit concerned from the potential duration this whole PR could add to fetch.
The fact that we have tables with no `lastmodifieddate` and means that on every fetch we will query all of them may be a bit problematic.
Not sure what the solution for it should be, and whether we should behave as if we didn't know about their table at all.

omrilit (approved):
👑

packages/netsuite-adapter/src/adapter.ts
dantal4: does `resolveAccountSpecificValues` a better name?

packages/netsuite-adapter/src/data_elements/types.ts
dantal4: yes

packages/netsuite-adapter/src/data_elements/suiteql_table_elements.ts
dantal4: setting `value[INTERNAL_IDS_MAP]` in case that it's undefined is needed also for existing instance - because that the field can be removed if it's value is an empty object (once we run `transformElement` on it, by purpose or not).

packages/netsuite-adapter/src/data_elements/suiteql_table_elements.ts
dantal4: I don't know if these tables can have lots of records "by their nature". I think that we can check the performance of it as we go, and modify the logic (limit/skip).

PR #5325, opened 2024-01-18
Fix custom record types with segments partial fetch

dantal4:
When running partial fetch for custom record types, the custom record types with custom segments were deleted.
It is because those custom record types are fetched by their custom segments, and since the custom segments weren't fetched we thought that the custom record types were deleted in the service and we marked them as "deleted elements".
The solution is to fetch the matching custom segments of those custom record types.

---

TODO:
- [x] add tests

---
_Release Notes_: 
Netsuite Adapter:
- Fix custom record types with segments partial fetch

---
_User Notifications_: 
None


omrilit (dismissed):
Nice catch!
I think that it'd be better to cover it with UTs

dantal4:
@omrilit I moved the login to `query.ts`

omrilit (approved):
Looks great!

PR #5326, opened 2024-01-18
SALTO-3080: deploy themes v2

edenhassid:
deploy guide themes

---

---
_Release Notes_: 
zendesk:
* support deploy of guide themes
---
_User Notifications_: 


packages/zendesk-adapter/e2e_test/adapter.test.ts
edenhassid, line 1: I think i'll add the e2e in a different PR

packages/zendesk-adapter/src/change_validator.ts
```diff
@@ -78,7 +80,6 @@ import {
   notEnabledMissingReferencesValidator,
   conditionalTicketFieldsValidator,
   dynamicContentDeletionValidator,
-  guideThemeReadonlyValidator,
```
edenhassid, line 81: I think we should keep this cv for now, @netama WDYT?

packages/zendesk-adapter/src/group_change.ts
```diff
@@ -71,9 +71,16 @@ const sectionChangeGroupId: deployment.ChangeIdFunction = async change =>
     ? getChangeData(change).elemID.getFullName()
     : undefined)
 
+// // themes and theme_settings need to be in the same group
+// const themeChangeGroupId: deployment.ChangeIdFunction = async change =>
+//   (([GUIDE_THEME_TYPE_NAME, THEME_SETTINGS_TYPE_NAME].includes(getChangeData(change).elemID.typeName))
+//     ? GUIDE_THEME_TYPE_NAME
+//     : undefined)
+
```
edenhassid, line 79: need to remove

packages/zendesk-adapter/src/group_change.ts
```diff
@@ -71,9 +71,16 @@ const sectionChangeGroupId: deployment.ChangeIdFunction = async change =>
...
 
 export const getChangeGroupIds = deployment.getChangeGroupIdsFunc([
   recurseIntoInstanceChangeToGroupId,
   sectionChangeGroupId,
+  // themeChangeGroupId,
```
edenhassid, line 84: need to remove

packages/zendesk-adapter/src/change_validators/guide_theme_delete_live.ts
```diff
@@ -0,0 +1,60 @@
...
+  if (elementSource === undefined) {
+    log.error('Failed to run guideThemeDeleteLiveValidator because no element source was provided')
+    return []
+  }
+  const deletedThemes = changes
+    .filter(isInstanceChange)
+    .filter(isRemovalChange)
+    .map(getChangeData)
+    .filter(instance => instance.elemID.typeName === GUIDE_THEME_TYPE_NAME)
+    .filter(instance => isReferenceExpression(instance.value.brand_id))
```
netama, line 41: this condition doesn't seem to be in the right place - even if the brand is not a reference we should block the deletion, no? maybe with a different error (or even with this one)? what are we trying to catch here exactly?

packages/zendesk-adapter/src/change_validators/guide_theme_update_metadata.ts
```diff
@@ -0,0 +1,63 @@
...
+  isInstanceChange,
+  isModificationChange, isReferenceExpression,
+  SeverityLevel,
+} from '@salto-io/adapter-api'
+import _ from 'lodash'
+import { GUIDE_THEME_TYPE_NAME } from '../constants'
+
+const MANIFEST_FIELDS = ['author', 'name', 'version']
+
+export const guideThemeUpdateMetadataValidator: ChangeValidator = async changes => {
```
netama, line 28: maybe add a docstring here as well?

packages/zendesk-adapter/src/change_validators/guide_theme_update_metadata.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  const unsupportedChanges = updatedThemes.filter(theme => {
+    const { before, after } = theme.data
+    if (isReferenceExpression(before.value.brand_id) && isReferenceExpression(after.value.brand_id)) {
+      return !before.value.brand_id.elemID.isEqual(after.value.brand_id.elemID)
+    }
+    if (_.isNumber(before.value.brand_id) && _.isNumber(after.value.brand_id)) {
+      return before.value.brand_id !== after.value.brand_id
+    }
+    return true
```
netama, line 54: if we don't expect to reach here maybe add a log?

packages/zendesk-adapter/src/change_validators/guide_theme_update_metadata.ts
```diff
@@ -0,0 +1,63 @@
...
+  }).map(theme => ({
+    elemID: getChangeData(theme).elemID,
+    message: 'Changing the brand on a theme is not supported',
```
netama, line 57: phrasing nit - maybe something like "Moving a theme to a different brand"? and technically we can write in the detailed message what the source and target brands are, right? (nice-to-have so can leave as-is for now)

packages/zendesk-adapter/src/config.ts
```diff
@@ -2669,15 +2669,26 @@ export const DEFAULT_TYPES: ZendeskApiConfig['types'] = {
   },
   theme: {
     transformation: {
-      idFields: ['&brand_id', ...DEFAULT_ID_FIELDS],
+      idFields: ['&brand_id', ...DEFAULT_ID_FIELDS, 'live'],
```
netama, line 2672: 1. do we want to cover excluding non-live themes as part of `omitInactive`?
2. are we sure we want this to be part of the id field? it can lead to weird behaviors between envs IMO

packages/zendesk-adapter/src/filters/guide_theme_settings.ts
```diff
@@ -0,0 +1,138 @@
...
+import {
+  GUIDE_THEME_TYPE_NAME, THEME_SETTINGS_TYPE_NAME, ZENDESK,
+} from '../constants'
+import { FilterCreator } from '../filter'
+import { publish } from './guide_themes/publish'
+
+const log = logger(module)
+
+
+const createThemeSettingsInstances = (guideThemes: InstanceElement[], config: ZendeskConfig): Element[] => {
```
netama, line 46: why not create these in the infra?

packages/zendesk-adapter/src/filters/guide_theme_settings.ts
```diff
@@ -0,0 +1,138 @@
...
+}
+
+/**
+ * Fetches and deploys guide theme settings
+ */
+const filterCreator: FilterCreator = ({ config, client }) => ({
+  name: 'guideThemesSettingsFilter',
+  onFetch: async elements => {
+    if (!isGuideEnabled(config[FETCH_CONFIG]) || !isGuideThemesEnabled(config[FETCH_CONFIG])) {
+      return undefined
```
netama, line 100: nit: I _think_ if all the `return`s are undefined you can just `return` (if you change both of them)?

netama (commented):
part 1 - didn't review the main flow yet, will do it later

packages/zendesk-adapter/src/change_validator.ts
edenhassid: I was thinking about deleting it only when we finish with developing everything

packages/zendesk-adapter/src/filters/guide_theme.ts
```diff
@@ -91,36 +144,94 @@ const filterCreator: FilterCreator = ({ config, client }) => ({
...
+      (await getInstancesFromElementSource(elementsSource, [THEME_SETTINGS_TYPE_NAME]))
+        .filter(instance => isReferenceExpression(instance.value.liveTheme)
+          && isReferenceExpression(instance.value.brand))
+        .map(instance => [instance.value.brand.elemID.getFullName(), instance.value.liveTheme.elemID.getFullName()])
+    )
+    const liveThemesNames = new Set(Object.values(liveThemesByBrand))
+
+    const processedChanges = await Promise.all(themeChanges
+      .map(async (change): Promise<{ appliedChange?: Change<InstanceElement>; errors: SaltoElementError[] }> => {
+        if (isRemovalChange(change)) {
```
netama, line 213: nit: maybe filter (again) for additions+modifications if this is an issue, instead of adding the `if`?

packages/zendesk-adapter/src/filters/guide_theme.ts
```diff
@@ -91,36 +144,94 @@ const filterCreator: FilterCreator = ({ config, client }) => ({
...
+          // Shouldn't happen, cleans up Typescript
+          return { errors: [] }
+        }
+        const isLiveTheme = liveThemesNames.has(getChangeData(change).elemID.getFullName())
+        const elementErrors = isModificationChange(change)
+          ? await updateTheme(change, client, isLiveTheme) : await createTheme(change, client, isLiveTheme)
```
netama, line 219: nit:
```suggestion
          ? await updateTheme(change, client, isLiveTheme)
          : await createTheme(change, client, isLiveTheme)
```

packages/zendesk-adapter/src/filters/guide_theme.ts
```diff
@@ -91,36 +144,94 @@ const filterCreator: FilterCreator = ({ config, client }) => ({
...
+        // Removal changes are handled in the default config.
+        isAdditionOrModificationChange(change)
+        && GUIDE_THEME_TYPE_NAME === getChangeData(change).elemID.typeName,
+    )
+
+    if (_.isEmpty(themeChanges)) {
+      return { deployResult: { appliedChanges: [], errors: [] }, leftoverChanges }
+    }
+
+    // to make sure that if there are multiple settings there is only one live per brand
```
netama, line 202: where do we verify that?

packages/zendesk-adapter/src/filters/guide_theme.ts
```diff
@@ -63,10 +83,43 @@ const addDownloadErrors = (
...
+  }
+  return elementErrors
+}
+
+const updateTheme = async (
+  change: ModificationChange<InstanceElement>, client: ZendeskClient, isLiveTheme: boolean
+): Promise<string[]> => {
+  const elementErrors = await createTheme(change, client, isLiveTheme)
+  if (elementErrors.length > 0) {
+    return elementErrors
```
netama, line 114: discussed IRL: we should make sure it hasn't been created in this case - if it has, we should probably delete the new (broken) theme and keep the old one

packages/zendesk-adapter/src/filters/guide_themes/api/createThemeExportJob.ts
```diff
@@ -25,20 +26,31 @@ export const createThemeExportJob = async (
...
+    })
+    if (![200, 202].includes(res.status)) {
+      log.warn(`Could not export a theme for themeId ${themeId}, received ${safeJsonStringify(res.data)}`)
+      return { job: undefined, errors: [safeJsonStringify(res.data)] }
+    }
+    return { job: isPendingJobResponse<DownloadJobData>(res.data) ? res.data.job : undefined, errors: [] }
+  } catch (e) {
+    if (e instanceof clientUtils.HTTPError) {
+      log.warn(`Could not update a theme for themeId ${themeId}. Received ${e.response.data}`)
+      if (e.response.data?.errors && Array.isArray(e.response.data.errors)) {
```
netama, line 49: why not just
```suggestion
      if (Array.isArray(e.response.data?.errors)) {
```

packages/zendesk-adapter/src/filters/guide_themes/api/createThemeImportJob.ts
```diff
@@ -0,0 +1,56 @@
+/*
```
netama, line 1: this file is almost identical to the export one except for a few minor differences. can we consolidate?
(we can also consolidate later, just seems like a simple one, especially since almost all the differences are in internal logs)

netama (commented):
did a relatively shallow review since you reviewed earlier, but LGTM overall except for a few questions. I think we can do a lot more to conoslidate the code, but if this is working and tested we can consolidate in a few weeks...

packages/zendesk-adapter/src/change_validators/guide_theme_delete_live.ts
edenhassid: I added this filter because we later run 
```
const deletedLiveThemes = deletedThemes
    .filter(instance => liveThemes[instance.value.brand_id.elemID.getFullName()] === instance.elemID.getFullName())
```
and I didn't want it to fail in case the brand_id is not a reference expression

packages/zendesk-adapter/src/change_validators/guide_theme_update_metadata.ts
edenhassid: we can reach there if once side has a number and the other a reference expression

packages/zendesk-adapter/src/change_validators/guide_theme_update_metadata.ts
```diff
@@ -0,0 +1,63 @@
...
+    .filter(change => getChangeData(change).elemID.typeName === GUIDE_THEME_TYPE_NAME)
+
+  const manifestChanges = updatedThemes
+    .filter(theme => {
+      const { before, after } = theme.data
+      return MANIFEST_FIELDS.some(field => before.value[field] !== after.value[field])
+    })
+    .map(theme => ({
+      elemID: getChangeData(theme).elemID,
+      message: 'Updating theme fields has no effect',
```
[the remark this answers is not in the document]
edenhassid: yes these are all the fields that cannot be updated

packages/zendesk-adapter/src/config.ts
edenhassid: discussed IRL, will remove the live from the id fields. if there will be a request to omit non live themes from the fetch we will add support for that in the future.

packages/zendesk-adapter/src/filters/guide_theme_settings.ts
```diff
@@ -0,0 +1,138 @@
...
+  return [themeSettingsType, ...guideThemeSettingsInstances]
+}
+
+/**
+ * Fetches and deploys guide theme settings
+ */
+const filterCreator: FilterCreator = ({ config, client }) => ({
+  name: 'guideThemesSettingsFilter',
+  onFetch: async elements => {
+    if (!isGuideEnabled(config[FETCH_CONFIG]) || !isGuideThemesEnabled(config[FETCH_CONFIG])) {
```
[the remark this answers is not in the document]
edenhassid: deleted the `isGuideEnabled` part

packages/zendesk-adapter/src/filters/guide_theme.ts
edenhassid: cause we create a record of brand and liveTheme, so there is only one per brand

packages/zendesk-adapter/src/change_validator.ts
edenhassid: will not delete and make a warning

packages/zendesk-adapter/src/change_validators/guide_theme_delete_live.ts
edenhassid: discussed IRL will leave like this

packages/zendesk-adapter/src/filters/guide_theme_settings.ts
```diff
@@ -0,0 +1,138 @@
...
+      annotations: {
+        [CORE_ANNOTATIONS.HIDDEN]: config[FETCH_CONFIG].hideTypes ?? true,
+      },
+    },
+  )
+
+  const guideThemeSettingsInstances = Object.values(
+    _.groupBy(
+      guideThemes.filter(theme => isReferenceExpression(theme.value.brand_id)),
+      theme => theme.value.brand_id.elemID.getFullName()
```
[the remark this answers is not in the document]
edenhassid: discussed IRL will leave like this

packages/zendesk-adapter/src/filters/guide_theme_settings.ts
```diff
@@ -0,0 +1,138 @@
...
+    )
+  ).map(themes => {
+    const brand = themes[0].value.brand_id
+    const brandName = brand.value.value.name
```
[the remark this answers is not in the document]
edenhassid: discussed IRL will leave like this

packages/zendesk-adapter/src/filters/guide_theme_settings.ts
```diff
@@ -0,0 +1,138 @@
...
+            message: e,
+            severity: 'Error' as SeverityLevel,
+          })),
+        }
+      }
+      return { appliedChange: change, errors: [] }
+    }))
+    const errorsOtherChanges = otherChanges.map(change => ({
+      elemID: getChangeData(change).elemID,
+      message: 'Non modification changes are not supported',
```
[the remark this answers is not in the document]
edenhassid: will change the message to be <..> instances cannot be added or removed

packages/zendesk-adapter/src/filters/guide_theme.ts
edenhassid: need to change the comment

packages/zendesk-adapter/src/filters/guide_theme.ts
edenhassid: will add an extra error in case it was created but not published

packages/zendesk-adapter/src/filters/guide_themes/publish.ts
```diff
@@ -0,0 +1,28 @@
...
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+import { logger } from '@salto-io/logging'
+import ZendeskClient from '../../client/client'
+
+const log = logger(module)
+
+export const publish = async (themeId: string, client: ZendeskClient): Promise<string[]> => {
+  const response = await client.post({ url: `/api/v2/guide/theming/themes/${themeId}/publish` })
```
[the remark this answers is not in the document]
edenhassid: discussed IRL, will be possible with the new infra

packages/zendesk-adapter/src/change_validators/guide_theme_readonly.ts
```diff
@@ -30,8 +30,8 @@ export const guideThemeReadonlyValidator: ChangeValidator = async changes => {
 
   return themeInstances.map(instance => ({
     elemID: instance.elemID,
-    severity: 'Error',
-    message: 'Cannot deploy this element because Guide Themes is currently under development.',
-    detailedMessage: 'Guide themes is in read only mode and not deployable at the moment.',
+    severity: 'Warning',
+    message: 'Guide Themes is currently under development.',
+    detailedMessage: 'The Guide Themes feature is currently under development. Please note that references to other elements will not function correctly upon deployment',
```
netama, line 35: nit: maybe
```suggestion
    detailedMessage: 'Guide Themes deploy support is currently under development. Please note that Theme code referencing other elements might not transfer correctly between environments',
```
(can also discuss IRL, either way not a blocker for merging)

packages/zendesk-adapter/src/change_validators/guide_theme_update_metadata.ts
```diff
@@ -0,0 +1,66 @@
...
+  isModificationChange, isReferenceExpression,
+  SeverityLevel,
+} from '@salto-io/adapter-api'
+import _ from 'lodash'
+import { GUIDE_THEME_TYPE_NAME } from '../constants'
+
+const MANIFEST_FIELDS = ['author', 'name', 'version']
+/**
+ * this filters check that there has been no change to the theme metadata. to change the theme metadata the user needs
+ * to update the manifest file and not the nacl
```
netama, line 29: 1. nits:
```suggestion

/**
 * this change validator checks that there has been no change to the theme metadata. to change the theme metadata the user needs
 * to update the manifest file and not the nacl
```
2. if these fields are also written in the manifest.json, is there a reason to have them in the nacl at all?
3. (related) when a user copies over both the manifest and the nacl changes from another env, won't they get these errors even though the change they made is valid?
4. seems we also have here "Moving a theme to a different brand is not supported" which is not listed?

packages/zendesk-adapter/src/filters/guide_theme.ts
```diff
@@ -13,41 +13,105 @@
...
+    fullPath: string,
+    pathParts: string[],
+    file: JSZip.JSZipObject,
+    currentDir: ThemeDirectory
+  ): Promise<void> => {
+    const [firstPart, ...rest] = pathParts
+
+    if (pathParts.length === 1) {
+      // It's a file
+      const filepath = `${ZENDESK}/themes/brands/${brandName}/${name}/${fullPath}`
```
netama, line 77: not related to this specifically, but I have a feeling this might not behave correctly with multiple-accounts (would the root be `zendesk` or the account name?)

not a blocker for merging, just something to be aware of

packages/zendesk-adapter/src/filters/guide_theme.ts
```diff
@@ -13,41 +13,105 @@
...
+  }
   await Promise.all(Object.entries(unzippedContents.files).map(async ([relativePath, file]): Promise<void> => {
     if (!file.dir) {
-      const content = await file.async('nodebuffer')
-      _.set(elements, relativePath.split('/').map(naclCase), {
-        filename: relativePath,
-        content: new StaticFile({ filepath: `${ZENDESK}/themes/brands/${brandName}/${name}/${relativePath}`, content }),
-      })
+      const pathParts = relativePath.split('/')
+      await addFile(relativePath, pathParts, file, elements)
```
netama, line 95: nits:
1.if we expect `fullPath` (which here is `relativePath`, which is it?) and `pathParts` to match, I think it'd be better to calculate `pathParts` inside the function to avoid inconsistencies
2. maybe use named parameters in `addFile`?

packages/zendesk-adapter/src/filters/guide_theme.ts
```diff
@@ -13,41 +13,105 @@
...
+        content: new StaticFile({ filepath, content }),
+      }
+    } else {
+      // It's a folder
+      if (!currentDir.folders[naclCase(firstPart)]) {
+        currentDir.folders[naclCase(firstPart)] = { files: {}, folders: {} }
+      }
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      await addFile(fullPath, rest, file, currentDir.folders[naclCase(firstPart)])
+    }
```
netama, line 90: 1. nit: can `return` early if there is nothing after the `else` and reduce nesting
```suggestion
      }
      return
    }
    // It's a folder
    if (!currentDir.folders[naclCase(firstPart)]) {
      currentDir.folders[naclCase(firstPart)] = { files: {}, folders: {} }
    }
    // eslint-disable-next-line @typescript-eslint/no-unused-vars
    await addFile(fullPath, rest, file, currentDir.folders[naclCase(firstPart)])
```
2. do we still need the `eslint-disable`?

packages/zendesk-adapter/src/filters/guide_theme.ts
```diff
@@ -13,41 +13,105 @@
...
     }
   }))
   return elements
 }
 
+const extractFilesFromThemeDirectory = (themeDirectory: ThemeDirectory): DeployThemeFile[] => {
+  let files: DeployThemeFile[] = []
+  // Add all files in the current directory
+  Object.values(themeDirectory.files).forEach(fileRecord => {
+    files.push(fileRecord as DeployThemeFile)
```
netama, line 105: can we use a type guard instead of the cast? there is no guarantee that the structure is as expected (didn't do another pass but maybe there are other places we should validate?)

packages/zendesk-adapter/src/filters/guide_theme.ts
```diff
@@ -63,16 +127,71 @@ const addDownloadErrors = (
...
+  isLiveTheme: boolean
+): Promise<string[]> => {
+  const live = isLiveTheme
+  const { brand_id: brandId, root } = change.data.after.value
+  const staticFiles = extractFilesFromThemeDirectory(root)
+  const { themeId, errors: elementErrors } = await create({ brandId, staticFiles }, client)
+  if (themeId === undefined) {
+    return [
+      ...elementErrors,
+      `Missing theme id from create theme response for theme ${change.data.after.elemID.getFullName()}`,
```
netama, line 142: I'm not in the details here so might be off, but is this error meaningful for caller? I'd assume it's a side-effect and not the root cause of the error, no? maybe we should only use it when there are no other errors (since we won't attempt to publish anyway in that case)

packages/zendesk-adapter/src/filters/guide_theme.ts
```diff
@@ -63,16 +127,71 @@ const addDownloadErrors = (
...
+
+const updateTheme = async (
+  change: ModificationChange<InstanceElement>, client: ZendeskClient, isLiveTheme: boolean
+): Promise<string[]> => {
+  const elementErrors = await createTheme(change, client, isLiveTheme)
+  if (elementErrors.length > 0) {
+    const lastError = elementErrors[elementErrors.length - 1]
+    if (lastError.includes('Failed to publish')) {
+      elementErrors[elementErrors.length - 1] = `${lastError}. The theme has been created but not published; you can manually publish it in the Zendesk UI.`
+    }
```
netama, line 161: nits:
1. can we use a different error type and search for it instead of parsing the string? (also do we need to assume it's the last one, or can we just find one?)
2. maybe pop the error and push a new one, instead of editing in-place?

packages/zendesk-adapter/src/filters/guide_theme.ts
```diff
@@ -63,16 +127,71 @@ const addDownloadErrors = (
...
+    return elementErrors
+  }
+  return deleteTheme(change.data.before.value.id, client)
+}
+
+const fixThemeTypes = (elements: Element[]): void => {
```
netama, line 167: can you please add a comment here that we should replace this once the fix is available?

packages/zendesk-adapter/src/change_validators/guide_theme_update_metadata.ts
netama: it's still an "error" flow though, no?

netama (approved):
cool :)
added a few more comments and questions, pre-approving since I probably won't do another review (didn't do a completely full now but ran into some things I missed in the previous iteration, sorry) 

(there's some more potential work here but as this is based on previous decisions and there will be some more capabilities soon, it's probably not worth iterating on more now)

(might be worth going over the resulting env before merging)

packages/zendesk-adapter/src/filters/guide_theme.ts
edenhassid: isn't this correct to all the other paths as well? in guide we also configure `zendesk` as the first folder

packages/zendesk-adapter/src/filters/guide_theme.ts
edenhassid: 1. path parts is reduced each call while the full path stays the same. I changed `relativePath` to be called `fullPath`
2. what do you mean by `named parameters`?

packages/zendesk-adapter/src/change_validators/guide_theme_update_metadata.ts
netama: re 1 - we can still generate even if the name is hidden, no?

packages/zendesk-adapter/src/filters/guide_theme.ts
netama: yep not specific for this PR (and it might work, will need to check)

packages/zendesk-adapter/src/filters/guide_theme.ts
edenhassid: 1. we can assume its the last one
2. 👍

packages/zendesk-adapter/src/change_validators/guide_theme_update_metadata.ts
edenhassid: right, so it will stay in the `unsupportedChanges` and later will be mapped to an error. what am i missing?

packages/zendesk-adapter/src/change_validators/guide_theme_update_metadata.ts
edenhassid: opened a ticket to improve this CV https://salto-io.atlassian.net/browse/SALTO-5315

packages/zendesk-adapter/src/filters/guide_theme.ts
edenhassid: I think you are right and this might not be relevant to the user. will remove this.

packages/zendesk-adapter/src/change_validators/guide_theme_update_metadata.ts
edenhassid: added a log warn in case we get there

packages/zendesk-adapter/src/filters/guide_theme.ts
adfineberg: You are probably right, it shouldn't happen according to the API. I added it as safeguard.

PR #5327, opened 2024-01-18
SALTO-5187: Add validate creds func (Jira)

shir-reifenberg:
add validate creds func as we have in other adapters

---

we will call validateCredentials for each function with `@requiresLogin()` decorator.
This way user will get unauthorized error when credentials are invalid

---
_Release Notes_: 

_Jira adapter_:
- When trying to deploy with invalid credentials, user will get unauthorized error.

---
_User Notifications_: 

None


omrilit (approved):
Thanks for that!

netama (approved):
🎈

shir-reifenberg:
@omrilit @netama This wasn't working for Jira DC so I had to adjust code a little (thanks e2e)
Can any of you please take another look? 🙏

PR #5329, opened 2024-01-18
SALTO-5129 fix workspace warning after fetch linkTypes on automation

omrifarkash:
When component inside automation instance is from type "jira.issue.delete.link" fetch is return an object with form: 
{
  id,
  direction,
  name
}

As linkTypes is list of strings, this returned object cause a workspace error.

Solution, map this case to a new field inside component.value `deleteLinkTypes` which is a list of matched objects

---

---
_Release Notes_: 
_Jira adapter_:
Fix workspace error after fetch on automation instance that has component with type "jira.issue.delete.link" 

---
_User Notifications_: 
_Jira adapter_:
linkTypes field inside a component from type "jira.issue.delete.link", will be rename to "deleteLinkTypes"


packages/jira-adapter/src/filters/automation/automation_structure.ts
```diff
@@ -341,6 +341,31 @@ const removeProjectsForGlobalDCAutomation = (instance: InstanceElement): void =>
   }
 }
 
+const transformDeleteLinkTypes = async (instance: InstanceElement, reverse?: boolean): Promise<void> => {
+  instance.value = (await transformElement({
```
shir-reifenberg, line 345: If we're only transforming instance' value, can we use `transformValues` instead? (or if possible `transformValuesSync`)

packages/jira-adapter/src/filters/automation/automation_structure.ts
```diff
@@ -341,6 +341,31 @@ const removeProjectsForGlobalDCAutomation = (instance: InstanceElement): void =>
...
+    element: instance,
+    allowEmpty: true,
+    strict: false,
+    transformFunc: ({ value, field }) => {
+      if (field?.name === 'components' && Array.isArray(value)) {
```
shir-reifenberg, line 350: Does it has to be under `components`? can't component be in top level (like in `trigger` field)? IMO I think it's safer to check `type === 'jira.issue.delete.link'` and that `linkTypes` is defined.

This way you can also avoid iterating `deleteIssuesComponent` (cause transformValues already does this for you)

packages/jira-adapter/src/filters/automation/automation_structure.ts
```diff
@@ -341,6 +341,31 @@ const removeProjectsForGlobalDCAutomation = (instance: InstanceElement): void =>
   }
 }
 
+const transformDeleteLinkTypes = async (instance: InstanceElement, reverse?: boolean): Promise<void> => {
```
shir-reifenberg, line 344: Can you please add a comment explaining why we need it ?

shir-reifenberg (commented):
Looks great!! 👑
Added a few comment mainly around simplifying the transform func

Also, as this changes nacl structure, we should probably add noise reduction in saas before merging this

packages/jira-adapter/src/filters/automation/automation_structure.ts
```diff
@@ -341,6 +341,30 @@ const removeProjectsForGlobalDCAutomation = (instance: InstanceElement): void =>
...
+// linkTypes field is returned from the service as a list of objects,
+// Since linkTypes is a string array in component objectType, we change the field name to deleteLinkTypes
+const transformDeleteLinkTypes = (instance: InstanceElement, reverse?: boolean): void => {
+  instance.value = (transformValuesSync({
+    type: instance.getTypeSync(),
+    values: instance.value,
+    allowEmpty: true,
+    strict: false,
+    transformFunc: ({ value }) => {
+      if (value?.type === 'jira.issue.delete.link' && value?.component === 'ACTION') {
```
shir-reifenberg, line 354: does `ACTION` is the only type of component supporting this ? (I think you can check this in Jira's UI)

packages/jira-adapter/src/filters/automation/automation_structure.ts
```diff
@@ -341,6 +341,30 @@ const removeProjectsForGlobalDCAutomation = (instance: InstanceElement): void =>
...
+        if (reverse) {
+          value.value.linkTypes = value.value?.deleteLinkTypes
+          delete value.value.deleteLinkTypes
+        } else {
+          value.value.deleteLinkTypes = value.value?.linkTypes
```
shir-reifenberg, line 359: Do we always have `linkTypes` when type is `jira.issue.delete.link` ? do we want to validate it's defined?

shir-reifenberg (approved):
Amazing! 🎉 
added 2 questions but code LGTM so approving

packages/jira-adapter/src/filters/automation/automation_structure.ts
omrifarkash: As dar as I understand, yes

packages/jira-adapter/src/filters/automation/automation_structure.ts
omrifarkash: We can have an empty list. but it is still works

PR #5330, opened 2024-01-19
SALTO-5302 - Salesforce: Don't create custom refs from PermissionSets.

tomsellek:
Permission sets are already handled by the normal reference mapping code. No need to create custom refs for them.
Additionally, if `generateRefsInProfiles` config is enabled then the normal reference mapping code will also generate refs for profiles, and we shouldn't try to create custom references from these generated `ReferenceExpression`s.

---

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/custom_references.ts
+++ b/packages/salesforce-adapter/src/custom_references.ts
@@ -139,7 +139,7 @@ const referencesToFields: RefTargetsGetter = (sectionEntry, sectionEntryKey) => 
 }
 
 const layoutReferences: RefTargetsGetter = sectionEntry => {
-  if (!_.isString(sectionEntry[0].layout)) {
+  if (!_.isString(sectionEntry[0]?.layout)) {
     return []
   }
   const layoutElemIdName = getMetadataElementName(sectionEntry[0].layout)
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/src/custom_references.ts
```diff
@@ -132,6 +139,9 @@ const referencesToFields: RefTargetsGetter = (sectionEntry, sectionEntryKey) =>
 }
 
 const layoutReferences: RefTargetsGetter = sectionEntry => {
+  if (!_.isString(sectionEntry[0].layout)) {
```
amitgoldarb, line 142: Is this required?
```suggestion
  if (!_.isString(sectionEntry[0]?.layout)) {
```

packages/salesforce-adapter/src/custom_references.ts
```diff
@@ -249,7 +262,13 @@ const customReferencesHandlers: Record<string, CustomRefsHandlersConfig> = {
 const getCustomReferencesConfig = (adapterConfig: InstanceElement): Record<string, boolean> => {
   const actualConfig = adapterConfig.value[FETCH_CONFIG]?.[DATA_CONFIGURATION]?.[CUSTOM_REFS_CONFIG] ?? {}
   const defaultConfig = _.mapValues(customReferencesHandlers, ({ isEnabledByDefault }) => isEnabledByDefault)
-  return _.defaults(actualConfig, defaultConfig)
+  const configWithAppliedDefaults = _.defaults(actualConfig, defaultConfig)
+
+  if (adapterConfig.value[FETCH_CONFIG]?.optionalFeatures?.generateRefsInProfiles
```
amitgoldarb, line 267: Should we exit earlier in this phase altogether? This logic is really only relevant when there are no references.

amitgoldarb (approved):
Approving to save cycles. Added 2 small comments

packages/salesforce-adapter/src/custom_references.ts
tomsellek: I am concerned about edge cases - e.g. what if the user set the config value but didn't fetch yet.

packages/salesforce-adapter/src/custom_references.ts
tomsellek: Seems unlikely, but we might as well...

PR #5331, opened 2024-01-21
SALTO-5304: sort password policy rule methods (Okta)

shir-reifenberg:
Sort unordered list password policy rule to avoid noise in subsequent fetches

---

there are no notifications enabled for this type, so no need of noise reduction

---

_Release Notes_: 

---
_User Notifications_: 

_Okta_adapter_:
- In password policy rules with more then one method used for password reset, the list of methods will be sorted.


omrilit (approved):
👑

PR #5332, opened 2024-01-21
SALTO-5303: Add validator to remind users to change tenant specific values when deploying application (Okta)

shir-reifenberg:
Add a validator to remind users to change tenant specific values

---

For more context see ticket details.
The validator is disabled by default and can be enabled from the adapter's config file

---
_Release Notes_: 

_Okta_adapter_:
- Add the ability to setup a reminder to change tenant specific values (such as URLs or subdomains) when deploying applications. To enable it, add the following config to the deploy section in adapter config: 

```
okta = {
  deploy = {
    changeValidators = {
      appUrls = true
    }
  }
}

```

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/change_validators/app_urls.ts
+++ b/packages/okta-adapter/src/change_validators/app_urls.ts
@@ -1,0 +1,36 @@
+/*
+*                      Copyright 2024 Salto Labs Ltd.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+import { ChangeValidator, getChangeData, isInstanceChange, isAdditionOrModificationChange } from '@salto-io/adapter-api'
+import { APPLICATION_TYPE_NAME } from '../constants'
+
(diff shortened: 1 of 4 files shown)


packages/okta-adapter/src/change_validators/app_urls.ts
```diff
@@ -0,0 +1,36 @@
...
+  changes
+    .filter(isInstanceChange)
+    .filter(isAdditionOrModificationChange)
+    .map(getChangeData)
+    .filter(instance => instance.elemID.typeName === APPLICATION_TYPE_NAME)
+    .map(instance => ({
+      elemID: instance.elemID,
+      severity: 'Warning',
+      message: 'Remember to update environment-specific fields when moving applications between Okta tenants',
+      detailedMessage: 'Ensure a smooth deployment by updating environment-specific fields, such as URLs and subdomains, when moving applications between Okta tenants. Easily adjust these fields by editing the relevant elements in Salto.',
```
omrilit, line 34: Do we have someone that usually goes over the CVs messages in Okta?
I think that we should use `values` instead of `fields`, but in general better to have some PM to approve it as well.
cc @iperelman

omrilit (approved):
Great work, thanks for taking this one!

packages/okta-adapter/src/change_validators/app_urls.ts
iperelman: Yes, we typically use 'values'.  I would consider the following changes (beyond fields / values):

- message: 'Update environment-specific values when deploying application elements between Okta tenants'.
- detailedMessage: 'Update environment-specific values, such as URLs and subdomains, when deploying application elements between Okta tenants.  Adjust these values by editing the relevant element in Salto.'.

shir-reifenberg: @iperelman thanks!

PR #5333, opened 2024-01-21
SALTO-5306: Removed recursed into fields

idoamit8:
Remove recursedInto fields in objectSchema. 

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


shir-reifenberg (approved):
👑  

Can you add some context in the related ticket so it would be easier to understand in the future why this change is needed?

PR #5334, opened 2024-01-21
SALTO-5109: Salesforce Metadata Important Values

tamtamirr:
Add initial implementation for Salesforce Important Values for Metadata Types.

---

- This feature is disabled by default. In order to enable it set the `importantValues` optional feature to true in the adapter config.
- Add the ability to configure additional important values under the adapter `fetch` config. If `important value` definition is already set for the `value`, we prefer the one from the `additional important values`.

Workspace diff: https://github.com/salto-io/tamir-sf/commit/6533cfebf854bd1a663bd825941e8a14346c49d9

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/types.ts
```diff
@@ -798,6 +802,10 @@ const fetchConfigType = createMatchingObjectType<FetchParameters>({
     preferActiveFlowVersions: { refType: BuiltinTypes.BOOLEAN },
     addNamespacePrefixToFullName: { refType: BuiltinTypes.BOOLEAN },
     warningSettings: { refType: warningSettingsType },
+    additionalImportantValues: {
+      // Exported type is downcasted to TypeElement
+      refType: CoreAnnotationTypes[CORE_ANNOTATIONS.IMPORTANT_VALUES] as ListType<ObjectType>,
+    },
```
tomsellek, line 808: Do we actually want a list of `ObjectType` in `salesforce.nacl`? Should this be a list of names or references or something?

packages/salesforce-adapter/src/filters/important_values.ts
```diff
@@ -0,0 +1,90 @@
...
+      const additionalImportantValues = makeArray(config.fetchProfile.additionalImportantValues)
+      if (additionalImportantValues.length === 0) {
+        return DEFAULT_IMPORTANT_VALUES
+      }
+      // The additional important values should override the default ones
+      const additionalImportantValuesValues = new Set(additionalImportantValues.map(value => value.value))
+      return DEFAULT_IMPORTANT_VALUES
+        .filter(importantValue => !additionalImportantValuesValues.has(importantValue.value))
+        .concat(additionalImportantValues)
+    }
```
tomsellek, line 71: Why not move this into the fetch profile? Perhaps create a method that takes a type+field and returns whether it's an important value?
Or maybe this is a fit for `MetadataQuery` instead?

tamtamirr: I think having all the logic that is relevant to `improtant values` with this feature is less complex for code readers.
In fact my initial implementation was this code within the `Fetch Profile` but it felt forgein there.
I also think that `FetchProfile` should only wrap the `fetch / deploy` settings (Yes, it's name is misleading) and not have more complex logic. WDYT?

packages/salesforce-adapter/src/types.ts
tamtamirr: Was completely tired when I wrote this :monkey: Thanks!

packages/adapter-api/src/builtins.ts
```diff
@@ -103,7 +103,7 @@ const restrictionType = new ObjectType({
   },
 })
 
-const importantValueType = new ObjectType({
+export const importantValueType = new ObjectType({
```
tamtamirr, line 106: @edenhassid are we okay with export? 🙏

edenhassid: yes

packages/salesforce-adapter/src/filters/important_values.ts
tomsellek: Personally, I like the abstraction of e.g. `MetadataQuery` as means of encapsulating the specifics of the config file structure, and ensuring the rest of the code doesn't need to know about them.

tamtamirr: I've done a refactor 🙏

tomsellek (approved):
:)

PR #5337, opened 2024-01-22
SALTO-5129 convert boolean value in component type to hasAttachmentsValue

omrifarkash:
For hasAttachment type of component in Jira, the api returns a boolean value.
In Salto matched objectType, value is an object.
To solve this, we're mapping the boolean value from the api to a new field "hasAttachmentsValue"

---

---
_Release Notes_: 
_Jira Adapter_:
For automation with component from type "hasAttachment", its boolean value will be store in "hasAttachmentsValue" field

---
_User Notifications_: 
Automation with component from type "hasAttachment", will have a new field named "hasAttachmentsValue" that will contain the boolean value

### DIFF
--- a/packages/jira-adapter/src/filters/automation/automation_structure.ts
+++ b/packages/jira-adapter/src/filters/automation/automation_structure.ts
@@ -277,7 +277,9 @@ const transformHasAttachmentValue = (instance: InstanceElement, reverse?: boolea
     values: instance.value,
     type: instance.getTypeSync(),
     transformFunc: ({ value }) => {
-      if (value?.type === 'jira.issue.hasAttachments' && value?.component === 'CONDITION') {
+      if (value?.type === 'jira.issue.hasAttachments'
+        && value?.component
+        && _.isBoolean(reverse ? value.hasAttachmentsValue : value.value)) {
         if (reverse) {
           value.value = value.hasAttachmentsValue
           delete value.hasAttachmentsValue
@@ -376,7 +378,7 @@ const transformDeleteLinkTypes = (instance: InstanceElement, reverse?: boolean):
     allowEmpty: true,
     strict: false,
     transformFunc: ({ value }) => {
-      if (value?.type === 'jira.issue.delete.link' && value?.component === 'ACTION') {
+      if (value?.type === 'jira.issue.delete.link' && value?.component) {
         if (reverse) {
           value.value.linkTypes = value.value?.deleteLinkTypes
           delete value.value.deleteLinkTypes
--- a/packages/jira-adapter/src/filters/automation/automation_structure.ts
+++ b/packages/jira-adapter/src/filters/automation/automation_structure.ts
@@ -266,6 +266,31 @@ const changeRawValueFieldsToValue = async (instance: InstanceElement): Promise<v
   })).value
 }
 
+
+// For components with type = "jira.issue.hasAttachments"
+// Value is a boolean, but component.value is an object in the objectType
+// So we transform the value to hasAttachmentsValue and vice versa
+const transformHasAttachmentValue = (instance: InstanceElement, reverse?: boolean): void => {
+  instance.value = transformValuesSync({
+    strict: false,
+    allowEmpty: true,
+    values: instance.value,
+    type: instance.getTypeSync(),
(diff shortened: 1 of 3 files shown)


packages/jira-adapter/src/filters/automation/automation_structure.ts
```diff
@@ -266,6 +266,31 @@ const changeRawValueFieldsToValue = async (instance: InstanceElement): Promise<v
...
+    allowEmpty: true,
+    values: instance.value,
+    type: instance.getTypeSync(),
+    transformFunc: ({ value }) => {
+      if (value?.type === 'jira.issue.hasAttachments' && value?.component === 'CONDITION') {
+        if (reverse) {
+          value.value = value.hasAttachmentsValue
+          delete value.hasAttachmentsValue
+        } else {
+          value.hasAttachmentsValue = value.value
```
shir-reifenberg, line 285: I think we should validate `value.value` exist and its what we think it is (same in deployment)

packages/jira-adapter/src/filters/automation/automation_structure.ts
```diff
@@ -266,6 +266,31 @@ const changeRawValueFieldsToValue = async (instance: InstanceElement): Promise<v
   })).value
 }
 
+
+// For components with type = "jira.issue.hasAttachments"
+// Value is a boolean, but component.value is an object in the objectType
+// So we transform the value to hasAttachmentsValue and vice versa
+const transformHasAttachmentValue = (instance: InstanceElement, reverse?: boolean): void => {
```
shir-reifenberg, line 273: Any reason not to consolidate with `transformDeleteLinkTypes` ? (and potentially with the other structure changes but it doesn't have to be part of this PR)

If we're already iterating the entire `Automation` instance, we call different functions inside the `transformFunc`
WDYT ?

packages/jira-adapter/src/filters/automation/automation_structure.ts
```diff
@@ -266,6 +266,31 @@ const changeRawValueFieldsToValue = async (instance: InstanceElement): Promise<v
...
+  instance.value = transformValuesSync({
+    strict: false,
+    allowEmpty: true,
+    values: instance.value,
+    type: instance.getTypeSync(),
+    transformFunc: ({ value }) => {
+      if (value?.type === 'jira.issue.hasAttachments' && value?.component === 'CONDITION') {
```
shir-reifenberg, line 280: Any chance `jira.issue.hasAttachments` can be in `TRIGGER` or `BRANCH` component ?

shir-reifenberg (commented):
Looks great!
Added few comments

Also, we should have a noise reduction here (can consolidate with the noise reduction for the previous PR)

packages/jira-adapter/src/filters/automation/automation_structure.ts
omrifarkash: Cannot create this in the UI but I can remove this condition

packages/jira-adapter/src/filters/automation/automation_structure.ts
omrifarkash: The only reason is that this is what we are already doing.
I agree that we should refactor this

omrifarkash: I think it supposed to be on a different task though

packages/jira-adapter/src/filters/automation/automation_structure.ts
omrifarkash: We haven't done it on the previous incident which is almost the same but on a different field.
Anyway, I added a validation

packages/jira-adapter/src/filters/automation/automation_structure.ts
shir-reifenberg: since you worked now on both I think it's a minimal effort to consolidate with `transformDeleteLinkTypes` in this PR and the rest can be done whenever we get to this ticket (thanks for opening this!) - but your call

shir-reifenberg (approved):
LGTM 😎 
please don't forger the noise reduction before you merge 🙏

PR #5339, opened 2024-01-22
SALTO-5212: Add support for different automation struct

idoamit8:
In this PR I support automation for assets components. 

---

Assets components can have only reference to `schemaId`.
Example can be found [here](https://ori-salto-test.atlassian.net/jira/settings/automation#/rule/17407491/386789326).
We want to support components with and without reference to `objectType`.  

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/jira-adapter/test/filters/automation/automation_fetch.test.ts
```diff
@@ -783,7 +784,6 @@ describe('automationFetchFilter', () => {
                       value: {
                         objectTypeId: '35',
                         schemaLabel: 'idoA Schema',
-                        schemaId: '5',
```
kami-delaroz, line 786: I'm not sure I understand why you decided to remove the _schemaId_ field (while the orig test had the _objectTypeId_ field)

packages/jira-adapter/test/filters/automation/automation_fetch.test.ts
```diff
@@ -731,6 +731,7 @@ describe('automationFetchFilter', () => {
             type: 'cmdb.object.create',
             value: {
               objectTypeId: '35',
+              schemaId: '5',
```
kami-delaroz, line 734: Maybe it's better to include a test of elements without _objectTypeId_ (both in fetch and in deploy flows)

kami-delaroz (approved):
🔥
Two small test-related comments. LGTM besides that

packages/jira-adapter/test/filters/automation/automation_fetch.test.ts
idoamit8: Because schemaId is mandatory and all the rest are not

packages/jira-adapter/test/filters/automation/automation_fetch.test.ts
idoamit8: Added :)

PR #5340, opened 2024-01-22
SALTO-5310 refactor automation structure to walk on instance once

omrifarkash:
Small refactor on automation_structor filter in Jira so we walk on each instance only once

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/automation/automation_structure.ts
+++ b/packages/jira-adapter/src/filters/automation/automation_structure.ts
@@ -318,9 +318,6 @@ const createTransformDeleteLinkTypesFunc = (reverse?: boolean): TransformFuncSyn
   return value
 }
 
-const isAutomationInstance = (elem: unknown): elem is InstanceElement =>
-  isInstanceElement(elem) && elem.elemID.typeName === AUTOMATION_TYPE
-
 const createAutomationTransformFunc = (transformFuncs: TransformFuncSync[]):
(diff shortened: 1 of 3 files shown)


packages/jira-adapter/src/filters/automation/automation_structure.ts
```diff
@@ -401,23 +336,27 @@ const filter: FilterCreator = ({ client }) => {
...
+            elementUtils.removeNullValuesTransformFunc,
+            removeRedundantKeys,
+            removeInnerIds,
+            separateLinkTypeField,
+            replaceStringValuesFieldName,
+            convertToCompareFieldValue,
+            createTransformDeleteLinkTypesFunc(),
+            createTransformHasAttachmentValueFunc(),
+          ])
+          instance.value = (await transformElement({
```
shir-reifenberg, line 354: if all function are sync now, can we use `transformValuesSync` ?

shir-reifenberg (approved):
Looks great 🌷 thank you!

packages/jira-adapter/src/filters/automation/automation_structure.ts
omrifarkash: Something isn't working with transformValuesSync or transformValues
I guess some of the applied logics need to go on the whole instance and not just the values.

Anyway, It is better than the current code. Merging :)

PR #5345, opened 2024-01-23
SALTO-5324: Remove serviceDeskId

idoamit8:
In this PR I removed the use of serviceDeskID and change to projectId identifier. 

---

In jira, a project has many unique Identifiers - ProjectKey, projectId, serviceDeskId.
Currently I see that we don't need to use serviceDeskId and we can use different Identifiers as shown [here](https://developer.atlassian.com/cloud/jira/service-desk/rest/intro/#project-identifiers).
In order to reduce api calles and re use common code - I changed the identifier from serviceDeskId to projectId

---
_Release Notes_: 
None

---
_User Notifications_: 
None


omrilit (approved):

packages/jira-adapter/src/adapter.ts
```diff
@@ -572,23 +574,28 @@ export default class JiraAdapter implements AdapterOperations {
...
       .filter(isInstanceElement)
       .filter(project => project.value.projectTypeKey === SERVICE_DESK)
-      .filter(project => {
-        if (project.value.serviceDeskId?.id === undefined) {
-          log.debug(`Service desk project ${project.elemID.name} does not have a service desk id`)
-          return false
-        }
-        return true
-      })
+      .filter(project => serviceDeskProjectIds.includes(project.value.id)))
```
omrilit, line 594: Worth adding a log indicating why the project was filtered so we could pinpoint the problem if needed, WDYT?

packages/jira-adapter/src/change_validators/jsm/jsm_permissions.ts
```diff
@@ -0,0 +1,73 @@
...
+      paginator(paginationArgs, page => makeArray(page.values) as clientUtils.ResponseValue[])
+    )).flat().map(project => project.projectId)
+
+    return awu(changes)
+      .filter(isInstanceChange)
+      .map(getChangeData)
+      .filter(instance => SUPPORTED_TYPES.has(instance.elemID.typeName))
+      .filter(instance => {
+        try {
+          const projectParent = getParent(instance)
```
omrilit, line 60: `getParent` could throw an exception that will be caught here and the error message will not be relevant to it

packages/jira-adapter/src/change_validators/jsm/jsm_permissions.ts
```diff
@@ -0,0 +1,73 @@
...
+          return !serviceDeskProjectIds.includes(projectParent.value.id)
+        } catch (e) {
+          return false
+        }
+      })
+      .map(instance => ({
+        elemID: instance.elemID,
+        severity: 'Error' as SeverityLevel,
+        message: 'Cannot deploy a JSM type without JSM permissions',
+        detailedMessage: `Cannot deploy ${instance.elemID.name} without JSM permissions`,
```
omrilit, line 70: Should Lior go over this one?

idoamit8: @liorn I would appreciate your opinion :)

PR #5346, opened 2024-01-24
SALTO-5316 fix IssueLayout id

kami-delaroz:
In some cases, IssueLayout ids are being changed over and over, causing fetches to be very big.
We need to replace the id of IssueLayouts with a combination of screenId and projectId

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
+++ b/packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
@@ -119,7 +119,7 @@ const fromLayoutConfigRespToLayoutConfig = (
   return { items }
 }
 
-const generateLayoutId = (
+export const generateLayoutId = (
   projectId: string, extraDefinerId: string | number
 ): string => projectId.concat('_').concat(extraDefinerId.toString())
 
--- a/packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
+++ b/packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
@@ -45,7 +45,7 @@ type QueryVariables = {
     projectId: string | number
     extraDefinerId: string | number
     layoutType?: string
-  }
+}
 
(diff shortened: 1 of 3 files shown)


packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
```diff
@@ -120,6 +119,10 @@ const fromLayoutConfigRespToLayoutConfig = (
   return { items }
 }
 
+export const generateLayoutId = (
+  projectId: string, extraDefinerId: string | number
```
shir-reifenberg, line 123: nit: personally I prefer using named params for functions with 2 args of the same type (especially if exported) to avoid confusions

packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
```diff
@@ -138,7 +141,7 @@ export const getLayout = async ({
   if (!Array.isArray(response.data) && isIssueLayoutResponse(response.data) && instance.path !== undefined) {
     const { issueLayoutResult } = response.data.issueLayoutConfiguration
     const value = {
-      id: issueLayoutResult.id,
+      id: generateLayoutId(instance.value.id, extraDefinerId),
```
shir-reifenberg, line 144: maybe add a comment explaining why are we creating this id ?

shir-reifenberg (approved):
Code LGTM so approving from my side 😎 
But I see `getLayout` is shared with JSM types (`RequestForm` and `IssueView`), are we OK with this changing those ids as well ? cc @idoamit8

packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
kami-delaroz: don't you think that the description in the PR / jira ticket is enough?

PR #5348, opened 2024-01-24
SALTO-5328: Omit condition's and trigger's ids in jiraWorkflow

shayc331:
_Omit condition's and trigger's ids in jiraWorkflow_

---

* it does not affect any user - the jiraWorkflow is behind a config flag

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


dantal4 (approved):
💯

PR #5349, opened 2024-01-25
SALTO-5285: Add allowedStatusCodesOnRemoval option to deploy function

shir-reifenberg:
Added option for adapter to pass `allowedStatusCodesOnRemoval` for adapter component's deploy 

---

For now I didn't include this in config, so each adapter can have global allowedStatusCodesOnRemoval, we might want to consider to define it per type (by adding this to deploy config)

I also removed some code in Okta we no longer need due to this change

---
_Release Notes_: 
None 

---
_User Notifications_: 
None


netama (approved):
🔥

PR #5351, opened 2024-01-25
SALTO-5331: Fixed jql in queue

idoamit8:
When fetching a queue, the API adds 'project = <ProjectKey>' at the beginning of the JQL. This doesn't appear to have any functional impact, but it is causing a lot of unnecessary noise

---

None

---
_Release Notes_: 
_Jira Adapter:_
The JQL for queues will no longer include an unnecessary 'project = <projectKey>' at the beginning.

---
_User Notifications_: 
_Jira Adapter:_
The JQL for queues will no longer include an unnecessary 'project = <projectKey>' at the beginning.


omrilit (approved):
🏅

PR #5353, opened 2024-01-27
Salto 5316 Fix issue layout id for request type

kami-delaroz:
In addition to the changes of IssueLayout id, should support also issueLayouts in requestType

---

[IssueLayout id change](https://github.com/salto-io/salto/pull/5346)

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/request_type.ts
+++ b/packages/jira-adapter/src/filters/request_type.ts
@@ -25,13 +25,7 @@ import { ISSUE_VIEW_TYPE, REQUEST_FORM_TYPE, REQUEST_TYPE_NAME } from '../consta
 import { deployChanges, defaultDeployChange } from '../deployment/standard_deployment'
 import JiraClient from '../client/client'
 import { LayoutConfigItem } from './layouts/layout_types'
-import {
-  LayoutTypeName,
-  getLayoutResponse,
-  isIssueLayoutResponse,
-  LAYOUT_TYPE_NAME_TO_DETAILS,
-  generateLayoutId,
-} from './layouts/layout_service_operations'
+import { LayoutTypeName, getLayoutResponse, isIssueLayoutResponse, LAYOUT_TYPE_NAME_TO_DETAILS } from './layouts/layout_service_operations'
 
 const { isDefined } = lowerDashValues
 const log = logger(module)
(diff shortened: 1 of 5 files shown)


packages/jira-adapter/src/filters/request_type.ts
```diff
@@ -140,20 +146,21 @@ const deployRequestTypeLayout = async (
...
-      extraDefinerId: requestType.value.id,
-      layoutType: LAYOUT_TYPE_NAME_TO_DETAILS[typeName].layoutType,
-    }
-    const response = await getLayoutResponse({ variables, client, typeName })
-    if (!isIssueLayoutResponse(response.data)) {
-      log.error(`Failed to deploy requestType's ${LAYOUT_TYPE_NAME_TO_DETAILS[typeName].fieldName} due to bad response from jira service`)
-      throw new Error(`Failed to deploy requestType's ${LAYOUT_TYPE_NAME_TO_DETAILS[typeName].fieldName} due to bad response from jira service`)
-    }
-    layout.id = response.data.issueLayoutConfiguration.issueLayoutResult.id
+    layout.id = generateLayoutId(variables.projectId, variables.extraDefinerId)
```
idoamit8, line 155: Now, that we are not saving the Id used for deploying the requestForm and issueView, I believe we no longer need to save an id.
This is because the issueView and requestForm are fields of requestType, and not instances of their own.

Additionally, I've noticed that the Id being saved is the Id of the requestType (extraDefinerId) duplicated, and not the id of the project and the extraDefinerID.

Does the Id of requestForm and issueView also changes during fetches? maybe we can save it for them and only change the logic for issue Layout to save requests.

idoamit8 (commented):
Great work! Thanks for the fix 🥇 
Left one comment, with a question :)

packages/jira-adapter/src/filters/request_type.ts
idoamit8, line 155: Now, that we are not saving the Id used for deploying the requestForm and issueView, I believe we no longer need to save an id.
This is because the issueView and requestForm are fields of requestType, and not instances of their own.

Additionally, I've noticed that the Id being saved is the Id of the requestType (extraDefinerId) duplicated, and not the id of the project and the extraDefinerID.

Does the Id of requestForm and issueView also changes during fetches? maybe we can save it for them and only change the logic for issue Layout to save requests.

idoamit8 (approved):
LGTM! 💯

PR #5356, opened 2024-01-28
salto-1872 - Netsuite: handle order problems in workbooks

SeggevHaimovich:
_Handle meaningless changes in the order of fields inside the xml_

---

_There are changes in the staging env: https://salto-io.slack.com/archives/C02BYQ46T34/p1705107160812559_

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


packages/netsuite-adapter/src/client/deploy_xml_utils.ts
```diff
@@ -27,6 +27,39 @@ const log = logger(module)
...
+const isDataset = (name:string): boolean =>
+  (name.startsWith('~/Objects/custdataset'))
+
+const isWorkbook = (name:string): boolean =>
+  (name.startsWith('~/Objects/custworkbook'))
+
+const isAnalyticsOrCustCollection = (name: string): boolean =>
+  (isCustCollection(name) || isDataset(name) || isWorkbook(name))
+
+const sortingAnalytics = (a:string, b:string): number => {
```
dantal4, line 42: 1. why is this needed? we already order `customTypeInfos` by their dependencies
2. if we need to sort the objects, it's nicer to do it on `customTypeInfos`, that is an array of objects with `typeName` attribute in them, and it's always better than string manipulations.

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -91,9 +91,91 @@ const fetchTransformFunc = async (
...
+    if (Array.isArray(newInstanceValues[arrName])) {
+      adjustWorkbookToolsOrder(newInstanceValues[arrName], oldTools)
+    }
+  })
+}
+
 const createAnalyticsInstance = async (
   instance: InstanceElement,
   analyticsType: ObjectType,
+  elementsSource: ReadOnlyElementsSource | undefined,
```
dantal4, line 178: nit - `elementsSource`is always defined in filters
```suggestion
  elementsSource: ReadOnlyElementsSource,
```

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -91,9 +91,91 @@ const fetchTransformFunc = async (
...
+        })
+      }
+    })
+}
+
+const discardUnrelevantChanges = (
+  newInstanceValues: Values,
+  oldTools: Record<string, Values>,
+): void => {
+  const innerArrayNames = [PIVOTS, CHARTS, DATASET_LINKS]
```
dantal4, line 167: should be defined as a const, probably in `analytics_constants.ts`

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -91,9 +91,91 @@ const fetchTransformFunc = async (
...
+  return _.isEqual(values1, values2)
+}
+
+const adjustWorkbookToolsOrder = (workbookTools: Value[], oldWorkbookToolsRecord: Record<string, Values>): void => {
+  workbookTools
+    .filter(values.isPlainObject)
+    .flatMap(obj => Object.values(obj))
+    .forEach((workbookTool: Values) => {
+      if (_.isArray(workbookTool.datasets)) {
+        workbookTool.datasets.sort()
```
dantal4, line 148: this is a list of strings, right?

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -91,9 +91,91 @@ const fetchTransformFunc = async (
...
+  })
+  return _.isEqual(values1, values2)
+}
+
+const adjustWorkbookToolsOrder = (workbookTools: Value[], oldWorkbookToolsRecord: Record<string, Values>): void => {
+  workbookTools
+    .filter(values.isPlainObject)
+    .flatMap(obj => Object.values(obj))
+    .forEach((workbookTool: Values) => {
+      if (_.isArray(workbookTool.datasets)) {
```
dantal4, line 147: can you use a const for `datasets`?

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -91,9 +91,91 @@ const fetchTransformFunc = async (
...
+        workbookTool.datasets.sort()
+      }
+      if (_.isString(workbookTool[SOAP_SCRIPT_ID]) && workbookTool[SOAP_SCRIPT_ID] in oldWorkbookToolsRecord) {
+        const oldPivot = oldWorkbookToolsRecord[workbookTool[SOAP_SCRIPT_ID]]
+        const innerXmlTitles = [DEFINITION, MAPPING]
+        innerXmlTitles.forEach((xmlTitle: string) => {
```
dantal4, line 153: this logic is very complicated and it's hard to follow... can you implement it using walkOnElement?

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
SeggevHaimovich: Yes the datasets is a list of strings.
For example:
datasets = [
  "custdataset_dilly_1",
  "custdataset_dilly_2",
]

This is part of what made the difference in the staging

packages/netsuite-adapter/src/filters/fix_order_problems_in_workbooks.ts
```diff
@@ -0,0 +1,199 @@
...
+const log = logger(module)
+const { awu } = collections.asynciterable
+
+const addWorkbookToolsToRecord = (
+  instance: InstanceElement,
+  workbookToolsRecord: Record<string, Values>,
+  name: string
+): void => {
+  if (Array.isArray(instance.value[name])) {
+    instance.value[name].forEach((tool:unknown) => {
```
dantal4, line 37:
```suggestion
    instance.value[name].forEach((tool: unknown) => {
```

packages/netsuite-adapter/src/filters/fix_order_problems_in_workbooks.ts
```diff
@@ -0,0 +1,199 @@
...
+const createRecordOfPivotsChartsAndDsLinks = async (
+  elementsSource: ReadOnlyElementsSource,
+  elemID: ElemID,
+): Promise<Record<string, Values>> => {
+  const oldWorkbookToolsRecord: Record<string, Values> = {}
+  const oldInstance = await elementsSource.get(elemID)
+  if (isInstanceElement(oldInstance)) {
+    addWorkbookToolsToRecord(oldInstance, oldWorkbookToolsRecord, PIVOTS)
+    addWorkbookToolsToRecord(oldInstance, oldWorkbookToolsRecord, CHARTS)
+    addWorkbookToolsToRecord(oldInstance, oldWorkbookToolsRecord, DATASET_LINKS)
```
dantal4, line 60: nit
```suggestion
    INNER_ARRAY_NAMES.forEach(arrayName => {
      addWorkbookToolsToRecord(oldInstance, oldWorkbookToolsRecord, arrayName)
    })
```

packages/netsuite-adapter/src/filters/fix_order_problems_in_workbooks.ts
```diff
@@ -0,0 +1,199 @@
...
+
+const adjustWorkbookToolsOrder = (workbookTools: Value[], oldWorkbookToolsRecord: Record<string, Values>): void => {
+  workbookTools
+    .filter(values.isPlainObject)
+    .flatMap(obj => Object.values(obj))
+    .forEach((workbookTool: Values) => {
+      if (_.isArray(workbookTool[DATASETS])) {
+        workbookTool[DATASETS].sort()
+      }
+      if (workbookTool[SOAP_SCRIPT_ID] in oldWorkbookToolsRecord) {
```
dantal4, line 105: nit - for safety
```suggestion
      if (_.isString(workbookTool[SOAP_SCRIPT_ID]) && workbookTool[SOAP_SCRIPT_ID] in oldWorkbookToolsRecord) {
```

packages/netsuite-adapter/src/filters/fix_order_problems_in_workbooks.ts
```diff
@@ -0,0 +1,199 @@
...
+        compareAndAssignInnerXmls(workbookTool, oldWorkbookTool)
+      }
+    })
+}
+
+const discardUnrelevantChanges = (
+  newInstanceValues: Values,
+  oldTools: Record<string, Values>,
+): void => {
+  INNER_ARRAY_NAMES.forEach((arrName: string) => {
```
dantal4, line 116: nit - type announcing is redundant when then type of `INNER_ARRAY_NAMES` is well defined:
```suggestion
  INNER_ARRAY_NAMES.forEach(arrName => {
```

packages/netsuite-adapter/src/filters/fix_order_problems_in_workbooks.ts
```diff
@@ -0,0 +1,199 @@
...
+      adjustWorkbookToolsOrder(newInstanceValues[arrName], oldTools)
+    }
+  })
+}
+
+const addPermissionToDataset = (
+  dataset1Name: string,
+  dataset2Name: string,
+  datasetMap: Map<string, InstanceElement>,
+): void => {
```
dantal4, line 127: naming:
1. why is it called "permission"?
2. please use "first" & "second" instead of dataset1/2

packages/netsuite-adapter/src/filters/fix_order_problems_in_workbooks.ts
```diff
@@ -0,0 +1,199 @@
...
+  if (isSameXmlValues(innerXmlNew, innerXmlOld)) {
+    newTool[innerXmlTitle] = oldTool[innerXmlTitle]
+  }
+}
+
+const adjustWorkbookToolsOrder = (workbookTools: Value[], oldWorkbookToolsRecord: Record<string, Values>): void => {
+  workbookTools
+    .filter(values.isPlainObject)
+    .flatMap(obj => Object.values(obj))
+    .forEach((workbookTool: Values) => {
```
dantal4, line 101: nit - for safety
```suggestion
    .flatMap(obj => Object.values(obj))
    .filter(values.isPlainObject)
    .forEach((workbookTool: Values) => {
```

packages/netsuite-adapter/src/filters/fix_order_problems_in_workbooks.ts
```diff
@@ -0,0 +1,199 @@
...
+  oldTools: Record<string, Values>,
+): void => {
+  INNER_ARRAY_NAMES.forEach((arrName: string) => {
+    if (Array.isArray(newInstanceValues[arrName])) {
+      adjustWorkbookToolsOrder(newInstanceValues[arrName], oldTools)
+    }
+  })
+}
+
+const addPermissionToDataset = (
```
dantal4, line 123: I think that this function requires a comment that explain why are we doing it

packages/netsuite-adapter/src/filters/fix_order_problems_in_workbooks.ts
```diff
@@ -0,0 +1,199 @@
...
+  datasetMap: Map<string, InstanceElement>,
+): void => {
+  const firstDataset = datasetMap.get(dataset1Name)
+  const secondDataset = datasetMap.get(dataset2Name)
+  if (
+    isInstanceElement(firstDataset)
+    && isInstanceElement(secondDataset)
+    && Array.isArray(secondDataset.value[DEPENDENCIES]?.[DEPENDENCY])
+  ) {
+    secondDataset.value[DEPENDENCIES]?.[DEPENDENCY]
```
dantal4, line 135: we shouldn't modify the dependencies list, as it is a part of the element that will be in the XML and will be deployed. we should use a new "made up" field, only so the dependencies mechanism will set a dependency between the datasets, and then delete it before creating the XML.
I think that we should have two functions in `client/utils.ts` for that:
```typescript
const addAdditionalDependency = (instance: InstanceElement, scriptId: string): void => {
  if (instance.value[ADDITIONAL_DEPS] === undefined) {
     instance.value[ADDITIONAL_DEPS] = []
  }
  instance.value[ADDITIONAL_DEPS].push(`[${SCRIPT_ID}=${scriptId}]`)
}

const removeAdditionalDependencies = (instance: InstanceElement): void => {
  delete instance.value[ADDITIONAL_DEPS]
}
```

packages/netsuite-adapter/src/filters/fix_order_problems_in_workbooks.ts
```diff
@@ -0,0 +1,199 @@
...
+      ?.push(new ReferenceExpression(firstDataset.elemID, firstDataset.value[SCRIPT_ID], firstDataset))
+  }
+}
+
+const getDatasetLinkList = (workbook: InstanceElement): Values[] =>
+  (Array.isArray(workbook.value[DATASET_LINKS])
+    ? workbook.value[DATASET_LINKS]
+      .filter(values.isPlainObject)
+      .map((obj:Values) => obj[DATASET_LINK])
```
dantal4, line 144: nit
```suggestion
      .map(obj => obj[DATASET_LINK])
```

packages/netsuite-adapter/src/filters/fix_order_problems_in_workbooks.ts
```diff
@@ -0,0 +1,199 @@
...
+      .filter(values.isPlainObject)
+    : [])
+
+const filterCreator: LocalFilterCreator = ({ elementsSource }) => ({
+  name: 'parseAnalytics',
```
dantal4, line 149: rename?

packages/netsuite-adapter/src/filters/fix_order_problems_in_workbooks.ts
```diff
@@ -0,0 +1,199 @@
...
+      .filter(change => change.data.after.elemID.typeName === DATASET)
+      .map(getChangeData)
+      .filter(isInstanceElement)
+
+    const datasetMap = new Map<string, InstanceElement>()
+    changedDatasets.forEach(dataset =>
+      datasetMap.set(
+        dataset.value[SCRIPT_ID],
+        dataset
+      ))
```
dantal4, line 173: nit - 
```suggestion
    const datasetMap = new Map<string, InstanceElement>(
      changedDatasets.map(dataset => [dataset.value[SCRIPT_ID], dataset]
    )
```

packages/netsuite-adapter/src/filters/fix_order_problems_in_workbooks.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    const workbooks = changes
+      .filter(isAdditionOrModificationChange)
+      .filter(change => change.data.after.elemID.typeName === WORKBOOK)
+      .map(getChangeData)
+      .filter(isInstanceElement)
+
+    workbooks.forEach(workbook => {
+      getDatasetLinkList(workbook)
+        .forEach((dslink: Values) => {
```
dantal4, line 183: dsLink is already `Values`
```suggestion
        .forEach(dslink => {
```

packages/netsuite-adapter/src/filters/fix_order_problems_in_workbooks.ts
```diff
@@ -0,0 +1,199 @@
...
+          if (Array.isArray(dslink[DATASETS])) {
+            const sortedDatasetArr = [...dslink[DATASETS]].sort()
```
dantal4, line 185: why do we need to sort it? isn't it sorted in fetch?

packages/netsuite-adapter/src/filters/fix_order_problems_in_workbooks.ts
```diff
@@ -0,0 +1,199 @@
...
+            if (dslink[DATASETS].length > 2) {
+              log.debug('There is a datasetLink with more than 2 datasets')
+              for (let i = 0; i < dslink[DATASETS].length - 1; i += 1) {
+                addPermissionToDataset(sortedDatasetArr[i], sortedDatasetArr[i + 1], datasetMap)
+              }
+            }
+            addPermissionToDataset(sortedDatasetArr[0], sortedDatasetArr[1], datasetMap)
```
dantal4, line 192: I think that this can be DRY, and please use `.reduce(...)` / `.forEach(...)` instead of a for loop.

packages/netsuite-adapter/src/filters/fix_order_problems_in_workbooks.ts
```diff
@@ -0,0 +1,199 @@
...
+      .filter(change => change.data.after.elemID.typeName === WORKBOOK)
+      .map(getChangeData)
+      .filter(isInstanceElement)
+
+    workbooks.forEach(workbook => {
+      getDatasetLinkList(workbook)
+        .forEach((dslink: Values) => {
+          if (Array.isArray(dslink[DATASETS])) {
+            const sortedDatasetArr = [...dslink[DATASETS]].sort()
+            if (dslink[DATASETS].length > 2) {
```
dantal4, line 186: maybe return if `dslink[DATASETS].length === 1` ?

dantal4 (commented):
Great work!

packages/netsuite-adapter/src/filters/fix_order_problems_in_workbooks.ts
SeggevHaimovich: not always
we saw that ti keeps happening in staging and if the order was the same in every fetch there was not a repeating problem

packages/netsuite-adapter/src/filters/fix_order_problems_in_workbooks.ts
SeggevHaimovich: The array represents a dataset link, which means there is a link between several datasets...
There shouldn't be any case with 1 dataset in this array.

dantal4: but how will the code behave if there's only 1 dataset?

SeggevHaimovich: after my change with the reduce there shouldn't be a problem anyway

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -306,6 +308,7 @@ const returnToOriginalShape = async (
   }).parse({ [ROOT]: definitionValues })
 
   return {
+    ..._.omit(instance.value, Object.keys(definitionValues)),
```
dantal4, line 311: in this case, setting explicitly the name, scriptid, dependencies & application_id is redundant, right?

packages/netsuite-adapter/src/filters/fix_order_problems_in_workbooks.ts
```diff
@@ -0,0 +1,200 @@
...
+const addDependencyToDataset = (
+  datasetFirstName: string,
+  datasetSecondName: string,
+  datasetMap: Map<string, InstanceElement>,
+): void => {
+  const firstDataset = datasetMap.get(datasetFirstName)
+  const secondDataset = datasetMap.get(datasetSecondName)
+  if (isInstanceElement(firstDataset) && isInstanceElement(secondDataset)) {
+    addAdditionalDependency(secondDataset, firstDataset.value[SCRIPT_ID])
+    // addAdditionalDependencyReference(secondDataset, firstDataset)
```
dantal4, line 136: leftover?

packages/netsuite-adapter/src/filters/fix_order_problems_in_workbooks.ts
```diff
@@ -0,0 +1,200 @@
...
+          if (isStringArray(datasetNamesArray)) {
+            if (datasetNamesArray.length !== 2) {
+              log.debug('There is a datasetLink with different number of datasets than 2: %o', dslink)
+            }
+            datasetNamesArray.reduce((prev, current, index) => {
+              if (index > 0) {
+                addDependencyToDataset(prev, current, datasetMap)
+              }
+              return current
+            }, '')
```
dantal4, line 193: nit - if you don't pass the initial value (''), then the reduce start running from index 1 (so you won't need the condition):
```typescript
> [1,2,3,4].reduce((x, y, i) => {console.log({x,y,i}); return y})
{ x: 1, y: 2, i: 1 }
{ x: 2, y: 3, i: 2 }
{ x: 3, y: 4, i: 3 }
```

dantal4 (approved):
💯

PR #5358, opened 2024-01-28
SALTO-5342: Salesforce Fetch With Changes Detection regression fixes

tamtamirr:
Fixed a couple regressions I've noticed in main.

---

Regressions fixed:
- CustomObject sub-instances are being deleted.
- Fetch fails due to issue with unresolved reference in referenceTo fields.
- TopicsForObjects instances are included in the workspace.
---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/src/filters/custom_objects_instances.ts
+++ b/packages/salesforce-adapter/src/filters/custom_objects_instances.ts
@@ -25,6 +25,7 @@ import {
   SaltoError,
   isInstanceElement,
   isPrimitiveType,
+  isReferenceExpression,
 } from '@salto-io/adapter-api'
 import { pathNaclCase, safeJsonStringify } from '@salto-io/adapter-utils'
 import {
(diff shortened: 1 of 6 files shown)


packages/salesforce-adapter/src/filters/custom_objects_instances.ts
```diff
@@ -634,6 +635,14 @@ const filterCreator: RemoteFilterCreator = ({ client, config }) => ({
...
       .filter(isCustomObjectSync)
       .toArray()
+
+    // Resolve referenceTo fields from Elements Source for partial fetch
+    await awu(customObjects)
+      .flatMap(customObject => Object.values(customObject.fields))
+      .filter(isReferenceField)
+      .flatMap(field => makeArray(field.annotations[FIELD_ANNOTATIONS.REFERENCE_TO]))
+      .filter(isReferenceExpression)
+      .forEach(async reference => { reference.value = await reference.getResolvedValue(elementsSource) })
```
tomsellek, line 645: This logic is incomplete - it doesn't cover the case of hierarchy fields. See `referenceFieldTargetType`.
Ideally, you could split `referenceFieldTargetType` into two functions - one that returns the types and one that returns the names of the types. You could use the former here, and the latter everywhere else `referenceFieldTargetType` is used.

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
```diff
@@ -634,6 +635,14 @@ const filterCreator: RemoteFilterCreator = ({ client, config }) => ({
     const customObjects = await awu(await elementsSource.getAll())
       .filter(isCustomObjectSync)
       .toArray()
+
+    // Resolve referenceTo fields from Elements Source for partial fetch
```
tomsellek, line 639: Why does the `customObjectInstances` filter need to resolve references? Why not handle this in `custom_object_instances_references`?

tomsellek (commented):
Thanks for finding these regressions!

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
tamtamirr: The only issue here is unresolved references. For hierarchy fields we simply return the `apiName` of the parent which is always resolved.

tomsellek (approved):
Approving due to urgency

PR #5359, opened 2024-01-29
SALTO-5343: convert statuses and transitions properties to a list

shayc331:
_convert statuses and transitions properties to a list_

---

* We use the same logic in other places in the adapter so I took it out as a more generic function

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/jira-adapter/src/filters/workflowV2/workflow_filter.ts
+++ b/packages/jira-adapter/src/filters/workflowV2/workflow_filter.ts
@@ -21,7 +21,7 @@ import { elements as adapterElements, config as configUtils, client as clientUti
 import { CORE_ANNOTATIONS, Element, InstanceElement, ObjectType, SaltoError, AdditionChange, Change, ChangeDataType, getChangeData, isAdditionChange, isInstanceElement, isModificationChange, ModificationChange, ReadOnlyElementsSource, ReferenceExpression, Values, ElemID } from '@salto-io/adapter-api'
 import { v4 as uuidv4 } from 'uuid'
 import { FilterCreator } from '../../filter'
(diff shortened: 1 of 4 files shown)


packages/jira-adapter/src/filters/workflowV2/workflow_filter.ts
```diff
@@ -439,6 +443,10 @@ const filter: FilterCreator = ({ config, client, paginator, fetchQuery, elements
           const statusIdToUuid = getUuidMap(statusInstances)
           const resolvedWorkflowInstance = await resolveValues(workflowInstance, getLookUpName)
           convertTransitionParametersFields(resolvedWorkflowInstance.value.transitions, convertParametersFieldsToString)
+          convertPropertiesToMap([
+            ...resolvedWorkflowInstance.value.statuses ?? [],
+            ...resolvedWorkflowInstance.value.transitions ?? [],
+          ])
```
idoamit8, line 449: I see that you are changing the struct of the filed in preDeploy and don't do anything in onDeploy.
Is it on purpose?

idoamit8 (commented):
Great work! 💯 
added one question and a small nit :)

packages/jira-adapter/src/filters/workflowV2/workflow_filter.ts
shayc331: the preDeploy changes the element structure completely, I save the original element [here](https://github.com/salto-io/salto/blob/main/packages/jira-adapter/src/filters/workflowV2/workflow_filter.ts#L437) and change it back in the onDeploy [here](https://github.com/salto-io/salto/blob/main/packages/jira-adapter/src/filters/workflowV2/workflow_filter.ts#L476)

idoamit8 (approved):
LGTM 💯

PR #5360, opened 2024-01-29
SALTO-5231: Added safeties to getParent

idoamit8:
In this PR I reused the function `isThereValidParent` to prevent errors when getParent throws. 

---

`getParent` throws an error when the parent is not a reference expression to an instance element. This issue can occur during fetch and deploy processes. When this error arises within the deploy filter in `deployChanges`, it is encapsulated in a try-catch block. Consequently, I modified the behavior in change validators and in the `onFetch` function.

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/adapter-utils/src/utils.ts
```diff
@@ -1152,6 +1152,14 @@ export const getParent = (instance: Element): InstanceElement => {
   return parents[0].value
 }
 
+export const isThereValidParent = (element: Element): boolean => {
```
ori-moisis, line 1155: naming nit
```suggestion
export const hasValidParent = (element: Element): boolean => {
```

packages/adapter-utils/src/utils.ts
```diff
@@ -1152,6 +1152,14 @@ export const getParent = (instance: Element): InstanceElement => {
...
+  try {
```
ori-moisis, line 1156: using exceptions for flow control is usually not good practice
I would question the original interface. do we really want something that throws exceptions there rather than returning `undefined`?
it feels like the case being solved in this PR won't be the only place that assumed `getParent` won't throw an exception, and for the code that does except the exception is probably less readable than the usual "or undefined" flow...

packages/jira-adapter/src/change_validators/last_queue.ts
```diff
@@ -40,15 +40,15 @@ export const deleteLastQueueValidator: (
     const projectToQueues = await awu(await elementsSource.list())
       .filter(id => id.typeName === QUEUE_TYPE && id.idType === 'instance')
       .map(id => elementsSource.get(id))
-      .filter(queue => queue.annotations[CORE_ANNOTATIONS.PARENT]?.[0] !== undefined)
+      .filter(queue => isReferenceExpression(queue.annotations[CORE_ANNOTATIONS.PARENT]?.[0]))
```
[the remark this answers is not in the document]
idoamit8: explained F2F

packages/adapter-utils/src/utils.ts
idoamit8: @ori-moisis Thanks! I'd like this to be included in tomorrow's deployment.
I'll update the interface to return undefined instead of throwing an error in a future PR.

PR #5361, opened 2024-01-29
SALTO-4564: Netsuite - warning about deleting permissions from role

SeggevHaimovich:
_Added a CV to warn the user of deleting permissions from role (can be done only in the UI).
Changed the CV that check the inner item list to warning instead of error.
Added filters to add the deleted items onDeploy to make the user's enviroment updated._

---

Jira ticket: https://salto-io.atlassian.net/browse/SALTO-4564

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/netsuite-adapter/src/adapter.ts
+++ b/packages/netsuite-adapter/src/adapter.ts
@@ -135,8 +135,8 @@ export const allFilters: (LocalFilterCreatorDefinition | RemoteFilterCreatorDefi
   // excludeCustomRecordTypes should run before customRecordTypesType,
   // because otherwise there will be broken references to excluded types.
   { creator: excludeCustomRecordTypes },
-  { creator: restoreDeletedListItems },
   { creator: restoreDeletedListItemsWithScriptId },
+  { creator: restoreDeletedListItems },
   // excludeCustomRecordTypes should run before customRecordTypesType,
   // because otherwise there will be broken references to excluded types.
   { creator: excludeCustomRecordTypes },
--- a/packages/netsuite-adapter/src/adapter.ts
+++ b/packages/netsuite-adapter/src/adapter.ts
@@ -86,14 +86,9 @@ import excludeCustomRecordTypes from './filters/exclude_by_criteria/exclude_cust
 import excludeInstances from './filters/exclude_by_criteria/exclude_instances'
 import workflowAccountSpecificValues from './filters/workflow_account_specific_values'
 import alignFieldNamesFilter from './filters/align_field_names'
-import {
-  Filter,
-  LocalFilterCreator,
-  LocalFilterCreatorDefinition,
-  RemoteFilterCreator,
-  RemoteFilterCreatorDefinition,
-  RemoteFilterOpts,
-} from './filter'
+import { Filter, LocalFilterCreator, LocalFilterCreatorDefinition, RemoteFilterCreator, RemoteFilterCreatorDefinition, RemoteFilterOpts } from './filter'
+import restoreDeletedListItems from './filters/restore_deleted_list_items_without_scriptid'
+import restoreDeletedListItemsWithScriptId from './filters/restore_deleted_list_items'
 import { getLastServerTime, getOrCreateServerTimeElements, getLastServiceIdToFetchTime } from './server_time'
 import { getChangedObjects } from './changes_detector/changes_detector'
 import { FetchDeletionResult, getDeletedElements } from './deletion_calculator'
@@ -190,6 +185,8 @@ export const allFilters: (LocalFilterCreatorDefinition | RemoteFilterCreatorDefi
   { creator: addAliasFilter },
   { creator: addImportantValuesFilter },
   { creator: fixOrderProblemsInWorkbooks },
+  { creator: restoreDeletedListItems }, // does it matter where it is?
+  { creator: restoreDeletedListItemsWithScriptId }, // does it matter where it is?
   // serviceUrls must run after suiteAppInternalIds and SDFInternalIds filter
   { creator: serviceUrls, addsNewInformation: true },
   { creator: addBundleReferences },
--- a/packages/netsuite-adapter/src/adapter.ts
+++ b/packages/netsuite-adapter/src/adapter.ts
@@ -86,7 +86,14 @@ import excludeCustomRecordTypes from './filters/exclude_by_criteria/exclude_cust
 import excludeInstances from './filters/exclude_by_criteria/exclude_instances'
 import workflowAccountSpecificValues from './filters/workflow_account_specific_values'
 import alignFieldNamesFilter from './filters/align_field_names'
-import { Filter, LocalFilterCreator, LocalFilterCreatorDefinition, RemoteFilterCreator, RemoteFilterCreatorDefinition, RemoteFilterOpts } from './filter'
+import {
+  Filter,
+  LocalFilterCreator,
+  LocalFilterCreatorDefinition,
+  RemoteFilterCreator,
+  RemoteFilterCreatorDefinition,
+  RemoteFilterOpts,
+} from './filter'
 import restoreDeletedListItemsWithoutScriptId from './filters/restore_deleted_list_items_without_scriptid'
 import restoreDeletedListItems from './filters/restore_deleted_list_items'
 import { getLastServerTime, getOrCreateServerTimeElements, getLastServiceIdToFetchTime } from './server_time'
--- a/packages/netsuite-adapter/src/adapter.ts
+++ b/packages/netsuite-adapter/src/adapter.ts
@@ -132,6 +132,14 @@ const { awu } = collections.asynciterable
 const log = logger(module)
 
 export const allFilters: (LocalFilterCreatorDefinition | RemoteFilterCreatorDefinition)[] = [
+  // excludeCustomRecordTypes should run before customRecordTypesType,
+  // because otherwise there will be broken references to excluded types.
+  { creator: excludeCustomRecordTypes },
+  { creator: restoreDeletedListItems },
+  { creator: restoreDeletedListItemsWithScriptId },
+  // excludeCustomRecordTypes should run before customRecordTypesType,
+  // because otherwise there will be broken references to excluded types.
+  { creator: excludeCustomRecordTypes },
   { creator: customRecordTypesType },
   { creator: omitSdfUntypedValues },
   { creator: dataInstancesIdentifiers },
@@ -185,6 +193,8 @@ export const allFilters: (LocalFilterCreatorDefinition | RemoteFilterCreatorDefi
   { creator: addAliasFilter },
   { creator: addImportantValuesFilter },
   { creator: fixOrderProblemsInWorkbooks },
+  { creator: restoreDeletedListItems }, // does it matter where it is?
+  { creator: restoreDeletedListItemsWithScriptId }, // does it matter where it is?
   { creator: restoreDeletedListItems }, // does it matter where it is?
   { creator: restoreDeletedListItemsWithScriptId }, // does it matter where it is?
   // serviceUrls must run after suiteAppInternalIds and SDFInternalIds filter
--- a/packages/netsuite-adapter/src/adapter.ts
+++ b/packages/netsuite-adapter/src/adapter.ts
@@ -87,8 +87,8 @@ import excludeInstances from './filters/exclude_by_criteria/exclude_instances'
 import workflowAccountSpecificValues from './filters/workflow_account_specific_values'
 import alignFieldNamesFilter from './filters/align_field_names'
 import { Filter, LocalFilterCreator, LocalFilterCreatorDefinition, RemoteFilterCreator, RemoteFilterCreatorDefinition, RemoteFilterOpts } from './filter'
-import restoreDeletedListItems from './filters/restore_deleted_list_items_without_scriptid'
-import restoreDeletedListItemsWithScriptId from './filters/restore_deleted_list_items'
+import restoreDeletedListItemsWithoutScriptId from './filters/restore_deleted_list_items_without_scriptid'
+import restoreDeletedListItems from './filters/restore_deleted_list_items'
 import { getLastServerTime, getOrCreateServerTimeElements, getLastServiceIdToFetchTime } from './server_time'
 import { getChangedObjects } from './changes_detector/changes_detector'
 import { FetchDeletionResult, getDeletedElements } from './deletion_calculator'
@@ -135,8 +135,8 @@ export const allFilters: (LocalFilterCreatorDefinition | RemoteFilterCreatorDefi
   // excludeCustomRecordTypes should run before customRecordTypesType,
   // because otherwise there will be broken references to excluded types.
   { creator: excludeCustomRecordTypes },
-  { creator: restoreDeletedListItemsWithScriptId },
   { creator: restoreDeletedListItems },
+  { creator: restoreDeletedListItemsWithoutScriptId },
   // excludeCustomRecordTypes should run before customRecordTypesType,
   // because otherwise there will be broken references to excluded types.
   { creator: excludeCustomRecordTypes },
@@ -193,10 +193,6 @@ export const allFilters: (LocalFilterCreatorDefinition | RemoteFilterCreatorDefi
   { creator: addAliasFilter },
   { creator: addImportantValuesFilter },
   { creator: fixOrderProblemsInWorkbooks },
-  { creator: restoreDeletedListItems }, // does it matter where it is?
-  { creator: restoreDeletedListItemsWithScriptId }, // does it matter where it is?
-  { creator: restoreDeletedListItems }, // does it matter where it is?
-  { creator: restoreDeletedListItemsWithScriptId }, // does it matter where it is?
   // serviceUrls must run after suiteAppInternalIds and SDFInternalIds filter
   { creator: serviceUrls, addsNewInformation: true },
   { creator: addBundleReferences },
--- a/packages/netsuite-adapter/src/change_validators/remove_list_item.ts
+++ b/packages/netsuite-adapter/src/change_validators/remove_list_item.ts
@@ -78,7 +78,7 @@ const changeValidator: NetsuiteChangeValidator = async changes => {
       elemID,
       severity: 'Warning',
       message: 'Can\'t remove inner elements',
-      detailedMessage: `Can't remove the inner element${removedListItems.length > 1 ? 's' : ''} ${removedListItems.join(', ')}. NetSuite supports the removal of inner elements only from their UI.`,
+      detailedMessage: `Can't remove the inner element${removedListItems.length > 1 ? 's' : ''} ${removedListItems.join(', ')}. NetSuite supports the removal of inner elements only from its UI.`,
     }))
     .toArray() as Promise<ChangeError[]>
 }
--- a/packages/netsuite-adapter/src/change_validators/remove_list_item.ts
+++ b/packages/netsuite-adapter/src/change_validators/remove_list_item.ts
@@ -76,8 +76,8 @@ const changeValidator: NetsuiteChangeValidator = async changes => {
     .filter(({ removedListItems }: { removedListItems: string[] }) => !_.isEmpty(removedListItems))
     .map(({ elemID, removedListItems }: { removedListItems: string[]; elemID: ElemID }) => ({
       elemID,
-      severity: 'Error',
-      message: "Can't remove inner elements",
+      severity: 'Warning',
+      message: 'Can\'t remove inner elements',
       detailedMessage: `Can't remove the inner element${removedListItems.length > 1 ? 's' : ''} ${removedListItems.join(', ')}. NetSuite supports the removal of inner elements only from their UI.`,
     }))
     .toArray() as Promise<ChangeError[]>
--- a/packages/netsuite-adapter/src/change_validators/remove_list_item.ts
+++ b/packages/netsuite-adapter/src/change_validators/remove_list_item.ts
@@ -77,12 +77,10 @@ const changeValidator: NetsuiteChangeValidator = async changes => {
     .map(({ elemID, removedListItems }: { removedListItems: string[]; elemID: ElemID }) => ({
       elemID,
       severity: 'Warning',
-      message: 'Can\'t remove inner elements',
-      detailedMessage: ((removedListItems.length > 1)
-        ? (`Can't remove the inner elements ${removedListItems.join(', ')}. NetSuite supports the removal of inner elements only from its UI.`
-          + ' Salto is going to ignore these removals.')
-        : (`Can't remove the inner element ${removedListItems.join(', ')}. NetSuite supports the removal of inner elements only from its UI.`
-          + ' Salto is going to ignore this removal.')),
+      message: 'Inner Element Removal Not Supported',
+      detailedMessage: `Netsuite doesn't support the removal of inner element${(removedListItems.length > 1) ? 's' : ''} via API; `
+        + `Salto will ignore ${(removedListItems.length > 1) ? 'these changes' : 'this change'} for this deployment. `
+        + `Please use Netuiste's UI to remove ${(removedListItems.length > 1) ? 'it' : 'them'}`,
     }))
     .toArray() as Promise<ChangeError[]>
 }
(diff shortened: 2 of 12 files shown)


packages/netsuite-adapter/src/change_validators/remove_list_item_without_scriptid.ts
```diff
@@ -0,0 +1,170 @@
...
+  elemID: ElemID
+  detailedMessage: string
+} => {
+  const getters = getGettersByType(instanceChange.data.before.elemID.typeName)
+  if (getters === undefined) {
+    return { removedListItems: [], elemID: instanceChange.data.before.elemID, detailedMessage: '' }
+  }
+  const { before, after } = instanceChange.data
+  const beforeItemList = getIdentifierList(before, getters)
+  const afterItemSet = new Set<string>(getIdentifierList(after, getters))
```
dantal4, line 144: nit - redundant
```suggestion
  const afterItemSet = new Set(getIdentifierList(after, getters))
```

packages/netsuite-adapter/src/change_validators/remove_list_item_without_scriptid.ts
```diff
@@ -0,0 +1,170 @@
...
+): {
+  removedListItems: string[]
+  elemID: ElemID
+  detailedMessage: string
+} => {
+  const getters = getGettersByType(instanceChange.data.before.elemID.typeName)
+  if (getters === undefined) {
+    return { removedListItems: [], elemID: instanceChange.data.before.elemID, detailedMessage: '' }
+  }
+  const { before, after } = instanceChange.data
```
dantal4, line 142: nit - put at the top of the function and use `before` directly instead of `instanceChange.data.before`

packages/netsuite-adapter/src/change_validators/remove_list_item_without_scriptid.ts
```diff
@@ -0,0 +1,170 @@
...
+    elemID: before.elemID,
+    detailedMessage: getters.getDetailedMessage(removedListItems),
+  }
+}
+
+const changeValidator: NetsuiteChangeValidator = async changes => {
+  const instanceChanges = await awu(changes)
+    .filter(isModificationChange)
+    .filter(isInstanceChange)
+    .toArray() as ModificationChange<InstanceElement>[]
```
dantal4, line 157: those are all synchronized functions, so  `awu` is redundant. In addition, the inferred type should be `ModificationChange<InstanceElement>[]` without specifying it

packages/netsuite-adapter/src/change_validators/remove_list_item_without_scriptid.ts
```diff
@@ -0,0 +1,170 @@
...
+export const getRemovedListItemDetails = (
+  instanceChange: ModificationChange<InstanceElement>,
+): {
+  removedListItems: string[]
+  elemID: ElemID
+  detailedMessage: string
+} => {
+  const getters = getGettersByType(instanceChange.data.before.elemID.typeName)
+  if (getters === undefined) {
+    return { removedListItems: [], elemID: instanceChange.data.before.elemID, detailedMessage: '' }
```
dantal4, line 140: nit - return `undefined`

packages/netsuite-adapter/src/change_validators/remove_list_item_without_scriptid.ts
```diff
@@ -0,0 +1,170 @@
...
+  }
+  const { before, after } = instanceChange.data
+  const beforeItemList = getIdentifierList(before, getters)
+  const afterItemSet = new Set<string>(getIdentifierList(after, getters))
+  const removedListItems = beforeItemList.filter(id => !afterItemSet.has(id))
+  return {
+    removedListItems,
+    elemID: before.elemID,
+    detailedMessage: getters.getDetailedMessage(removedListItems),
```
dantal4, line 149: does it makes sense to run `getDetailedMessage` on an empty list? I think it's worth returning undefined in that case.

packages/netsuite-adapter/src/change_validators/remove_list_item.ts
```diff
@@ -72,9 +72,9 @@ const changeValidator: NetsuiteChangeValidator = async changes => {
     .filter(({ removedListItems }: {removedListItems: string[]}) => !_.isEmpty(removedListItems))
     .map(({ elemID, removedListItems }: {removedListItems: string[]; elemID: ElemID}) => ({
       elemID,
-      severity: 'Error',
+      severity: 'Warning',
       message: 'Can\'t remove inner elements',
-      detailedMessage: `Can't remove the inner element${removedListItems.length > 1 ? 's' : ''} ${removedListItems.join(', ')}. NetSuite supports the removal of inner elements only from their UI.`,
+      detailedMessage: `Can't remove the inner element${removedListItems.length > 1 ? 's' : ''} ${removedListItems.join(', ')}. NetSuite supports the removal of inner elements only from its UI.`,
```
dantal4, line 77: maybe the message should be modified? something like "Salto is going to ignore those changes..." ?
@omrilit @liorn  WDYT?

packages/netsuite-adapter/src/change_validators/remove_list_item_without_scriptid.ts
```diff
@@ -0,0 +1,170 @@
...
+const changeValidator: NetsuiteChangeValidator = async changes => {
+  const instanceChanges = await awu(changes)
+    .filter(isModificationChange)
+    .filter(isInstanceChange)
+    .toArray() as ModificationChange<InstanceElement>[]
+
+  return instanceChanges
+    .map(getRemovedListItemDetails)
+    .filter(({ removedListItems }) => !_.isEmpty(removedListItems))
+    .map(({ detailedMessage, elemID }) => ({
```
dantal4, line 162: we should try not using `as` (line 167) - this should replace it
```suggestion
    .map(({ detailedMessage, elemID }): ChangeError => ({
```

packages/netsuite-adapter/src/filters/restore_deleted_list_items.ts
```diff
@@ -0,0 +1,82 @@
...
+
+const { awu } = collections.asynciterable
+
+const log = logger(module)
+
+const getRelativePathAndScriptId = (path: string): {
+  relativePath: string
+  scriptId: string
+} | undefined => {
+  const fullNameRegex = /\.instance\.(?<instance_name>\w+)\.(?<relative_path>[\w.]+(?=\.\w+$))\.(?<scriptid>\w+)$/
```
dantal4, line 32: elem id has a defined structure and lots of methods to help get what you want, without using regex on it.

packages/netsuite-adapter/src/filters/restore_deleted_list_items.ts
```diff
@@ -0,0 +1,82 @@
...
+    await awu(changes)
+      .filter(isModificationChange)
+      .filter(isInstanceChange)
+      .forEach(async instanceChange => {
+        const before = await getScriptIdsUnderLists(instanceChange.data.before)
+        const after = await getScriptIdsUnderLists(instanceChange.data.after)
+        before.forEach((value, key) => {
+          if (!after.has(key)) {
+            _.get(instanceChange.data.after.value, value.relativePath)[value.id] = value.val
+            log.debug('')
```
dantal4, line 75: ?

packages/netsuite-adapter/src/filters/restore_deleted_list_items.ts
```diff
@@ -0,0 +1,82 @@
...
+    log.debug('')
+    await awu(changes)
+      .filter(isModificationChange)
+      .filter(isInstanceChange)
+      .forEach(async instanceChange => {
+        const before = await getScriptIdsUnderLists(instanceChange.data.before)
+        const after = await getScriptIdsUnderLists(instanceChange.data.after)
+        before.forEach((value, key) => {
+          if (!after.has(key)) {
+            _.get(instanceChange.data.after.value, value.relativePath)[value.id] = value.val
```
dantal4, line 74: this can be done using more structured objects:
```typescript
setPath(after, elemId, value)
```

packages/netsuite-adapter/src/filters/restore_deleted_list_items.ts
```diff
@@ -0,0 +1,82 @@
...
+  const fullNameRegex = /\.instance\.(?<instance_name>\w+)\.(?<relative_path>[\w.]+(?=\.\w+$))\.(?<scriptid>\w+)$/
+  const match = path.match(fullNameRegex)
+  if (match && match.groups) {
+    return { relativePath: match.groups.relative_path, scriptId: match.groups.scriptid }
+  }
+  return undefined
+}
+
+const getScriptIdsUnderLists = async (instance: InstanceElement):
+  Promise<Map<string, { relativePath: string; id:string; val: Value}>> => {
```
dantal4, line 41: styling nit
```suggestion
const getScriptIdsUnderLists = async (
  instance: InstanceElement
): Promise<Map<string, SomeDefinedType>> => {
```

packages/netsuite-adapter/src/filters/restore_deleted_list_items_without_scriptid.ts
```diff
@@ -0,0 +1,61 @@
...
+      }
+      return { [id]: item }
+    })
+    .filter((val: { [id: string]: ItemInList } | undefined): val is { [id: string]: ItemInList } => val !== undefined)
+}
+
+const filterCreator: LocalFilterCreator = () => ({
+  name: 'restorDeletedListItems',
+  onDeploy: async changes => {
+    log.debug('')
```
dantal4, line 43: ?

packages/netsuite-adapter/src/filters/restore_deleted_list_items_without_scriptid.ts
```diff
@@ -0,0 +1,61 @@
...
+import _ from 'lodash'
+import { LocalFilterCreator } from '../filter'
+import { ItemInList, ItemListGetters, getGettersByType, getRemovedListItemDetails } from '../change_validators/remove_list_item_without_scriptid'
+
+const log = logger(module)
+
+const getRemovedListItems = (
+  instanceChange: ModificationChange<InstanceElement>,
+  getters: ItemListGetters,
+): { [id: string]: ItemInList }[] => {
```
dantal4, line 27: nit - `Record<string, ItemInList>[]`

or even better - return one `Record` object with all the `ItemInList` in it

packages/netsuite-adapter/src/filters/restore_deleted_list_items_without_scriptid.ts
```diff
@@ -0,0 +1,61 @@
...
+  const idsList = getRemovedListItemDetails(instanceChange)
+  return idsList.removedListItems
+    .map(id => {
+      const item = getters.getItemByID(instanceChange.data.before, id)
+      if (item === undefined) {
+        return undefined
+      }
+      return { [id]: item }
+    })
+    .filter((val: { [id: string]: ItemInList } | undefined): val is { [id: string]: ItemInList } => val !== undefined)
```
dantal4, line 37: `.filter(values.isDefined)` does the same

packages/netsuite-adapter/src/filters/restore_deleted_list_items_without_scriptid.ts
```diff
@@ -0,0 +1,61 @@
...
+      .filter(isModificationChange)
+      .filter(isInstanceChange)
+      .forEach(instanceChange => {
+        const getters = getGettersByType(instanceChange.data.before.elemID.typeName)
+        if (getters === undefined) {
+          return
+        }
+        const removedItems = getRemovedListItems(instanceChange, getters)
+        Object.assign(
+          _.get(instanceChange.data.after.value, getters.getListPath()),
```
dantal4, line 54: what will happen if `after.value.permissions.permission` is not defined?

dantal4 (commented):
Great work!

packages/netsuite-adapter/src/change_validators/remove_list_item.ts
omrilit: I agree with the general suggestion, as for phrasing, better to consult with a PM.
@rozenmanofer was the author of the messages in NS back then, so maybe better to loop him in

packages/netsuite-adapter/src/adapter.ts
```diff
@@ -143,6 +145,8 @@ export const allFilters: (LocalFilterCreatorDefinition | RemoteFilterCreatorDefi
   { creator: addInstancesFetchTime },
   { creator: addAliasFilter },
   { creator: addImportantValuesFilter },
+  { creator: restoreDeletedListItems }, // does it matter where it is?
+  { creator: restoreDeletedListItemsWithScriptId }, // does it matter where it is?
```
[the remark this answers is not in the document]
SeggevHaimovich: That's what I thought :)

packages/netsuite-adapter/src/change_validators/remove_list_item_without_scriptid.ts
```diff
@@ -0,0 +1,170 @@
...
+const PERMLEVEL = 'permlevel'
+const RESTRICTION = 'restriction'
+
+type RolePermissionObject = {
+  [PERMKEY]: string | ReferenceExpression
+  [PERMLEVEL]: string
+  [RESTRICTION]?: string
+}
+
+export type ItemInList = RolePermissionObject
```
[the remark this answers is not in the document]
SeggevHaimovich: Yes, that is what I thought about, to make it general as possible so we won't need to change all the places we use this type when we add more lists..

packages/netsuite-adapter/src/filters/restore_deleted_list_items.ts
SeggevHaimovich: used setPath with the elemID itself as you suggested so I deleted this func..

packages/netsuite-adapter/src/filters/restore_deleted_list_items_without_scriptid.ts
SeggevHaimovich: You are right.
I am wondering what we should do, do we let the user deploy after deleting the whole list (I know that in roles there is no problem to deploy without the permission list but I don't know about the future lists this CV and filter will deal with)?

The options are:
1. add this case in the CV to tell the user he can't delete the list and add this in the onDeploy filter.
2. in case of deletion of the entire list, both the CV and the filter does nothing.

SeggevHaimovich: nothing happens if the whole list is deleted, but I don't know if it will be the same case in the future lists we will put in this CV and filter.
For now I will implement option 1

packages/netsuite-adapter/src/change_validators/remove_list_item_without_scriptid.ts
```diff
@@ -0,0 +1,187 @@
...
+
+export const getGettersByType = (
+  typeName: string,
+): ItemListGetters | undefined => {
+  if (typeName === ROLE) {
+    return roleGetters
+  }
+  return undefined
+}
+const getIdentifierList = (
```
dantal4, line 127: styling nit - empty line between functions
```suggestion
}

const getIdentifierList = (
```

packages/netsuite-adapter/src/change_validators/remove_list_item_without_scriptid.ts
```diff
@@ -0,0 +1,187 @@
...
+const getDetailedMessage = (
+  instanceChange: ModificationChange<InstanceElement>,
+): {
+  elemID: ElemID
+  detailedMessage: string
+} | undefined => {
+  const { before, after } = instanceChange.data
+  const { elemID } = before
+  const getters = getGettersByType(elemID.typeName)
+  if (getters === undefined || _.isUndefined(_.get(before.value, getters.getListPath()))) {
```
dantal4, line 152: nit
```suggestion
  if (getters === undefined || _.get(before.value, getters.getListPath()) === undefined) {
```

packages/netsuite-adapter/src/change_validators/remove_list_item_without_scriptid.ts
```diff
@@ -0,0 +1,187 @@
...
+    return undefined
+  }
+
+  if (_.isUndefined(_.get(after.value, getters.getListPath()))) {
```
dantal4, line 156: I think that we should have a similar message as if all the permissions were deleted. if you remove this condition `getRemovedListItemIds` will take care of that case, right?

packages/netsuite-adapter/src/change_validators/remove_list_item_without_scriptid.ts
```diff
@@ -0,0 +1,187 @@
...
+  const afterItemSet = new Set(getIdentifierList(after, getters))
+  return beforeItemList.filter(id => !afterItemSet.has(id))
+}
+
+const getDetailedMessage = (
+  instanceChange: ModificationChange<InstanceElement>,
+): {
+  elemID: ElemID
+  detailedMessage: string
+} | undefined => {
```
dantal4, line 148: nit - return `ChangeError | undefined`

packages/netsuite-adapter/src/filters/restore_deleted_list_items.ts
```diff
@@ -0,0 +1,62 @@
...
+import { collections } from '@salto-io/lowerdash'
+import { ElemID, InstanceElement, Value, isInstanceChange, isModificationChange } from '@salto-io/adapter-api'
+import { WALK_NEXT_STEP, setPath, walkOnElement } from '@salto-io/adapter-utils'
+import _ from 'lodash'
+import { LocalFilterCreator } from '../filter'
+import { SCRIPT_ID } from '../constants'
+
+const { awu } = collections.asynciterable
+
+const getScriptIdsUnderLists = async (
```
dantal4, line 25: nit - can remove the `async`

packages/netsuite-adapter/src/adapter.ts
```diff
@@ -68,6 +68,8 @@ import addBundleReferences from './filters/bundle_ids'
 import excludeCustomRecordTypes from './filters/exclude_by_criteria/exclude_custom_record_types'
 import excludeInstances from './filters/exclude_by_criteria/exclude_instances'
 import workflowAccountSpecificValues from './filters/workflow_account_specific_values'
+import restoreDeletedListItems from './filters/restore_deleted_list_items_without_scriptid'
+import restoreDeletedListItemsWithScriptId from './filters/restore_deleted_list_items'
```
dantal4, line 72: the names are confusing - use "with" or "without" but not both

packages/netsuite-adapter/src/filters/restore_deleted_list_items.ts
```diff
@@ -0,0 +1,62 @@
...
+  instance: InstanceElement
+): Promise<Map<string, { elemID: ElemID; val: Value}>> => {
+  const pathToScriptIds = new Map<string, { elemID: ElemID; val: Value}>()
+  walkOnElement({
+    element: instance,
+    func: ({ value, path }) => {
+      if (path.isAttrID()) {
+        return WALK_NEXT_STEP.SKIP
+      }
+      if (_.isPlainObject(value) && SCRIPT_ID in value) {
```
dantal4, line 35: ` && !path.isTopLevel()` ?

packages/netsuite-adapter/src/filters/restore_deleted_list_items.ts
```diff
@@ -0,0 +1,62 @@
...
+        pathToScriptIds.set(path.getFullName(), { elemID: path, val: value })
+      }
+      return WALK_NEXT_STEP.RECURSE
+    },
+  })
+  return pathToScriptIds
+}
+
+const filterCreator: LocalFilterCreator = () => ({
+  name: 'restorDeletedListItems',
```
dantal4, line 45:
```suggestion
  name: 'restoreDeletedListItems',
```

packages/netsuite-adapter/src/filters/restore_deleted_list_items_without_scriptid.ts
```diff
@@ -0,0 +1,66 @@
...
+    const item = getters.getItemByID(instanceChange.data.before, id)
+    if (item !== undefined) {
+      itemsRecord[id] = item
+    }
+  })
+  return itemsRecord
+}
+
+const filterCreator: LocalFilterCreator = () => ({
+  name: 'restorDeletedListItems',
```
dantal4, line 38: should be a different name

packages/netsuite-adapter/src/filters/restore_deleted_list_items_without_scriptid.ts
```diff
@@ -0,0 +1,66 @@
...
+import { ItemInList, ItemListGetters, getGettersByType, getRemovedListItemIds } from '../change_validators/remove_list_item_without_scriptid'
+
+const getRemovedListItems = (
+  instanceChange: ModificationChange<InstanceElement>,
+  getters: ItemListGetters,
+): Record<string, ItemInList> => {
+  const idsList = getRemovedListItemIds(instanceChange, getters)
+  const itemsRecord: Record<string, ItemInList> = {}
+  idsList.forEach(id => {
+    const item = getters.getItemByID(instanceChange.data.before, id)
```
dantal4, line 29: can `idsList` be `ItemInList[]` instead of `string[]`, so you won't have to call `getters.getItemByID` ?

packages/netsuite-adapter/src/filters/restore_deleted_list_items_without_scriptid.ts
```diff
@@ -0,0 +1,66 @@
...
+const filterCreator: LocalFilterCreator = () => ({
+  name: 'restorDeletedListItems',
+  onDeploy: async changes => {
+    changes
+      .filter(isModificationChange)
+      .filter(isInstanceChange)
+      .forEach(instanceChange => {
+        const getters = getGettersByType(instanceChange.data.before.elemID.typeName)
+        const { before, after } = instanceChange.data
+        if (getters === undefined || _.isUndefined(_.get(before.value, getters.getListPath()))) {
```
dantal4, line 46: nit
```suggestion
        if (getters === undefined || _.get(before.value, getters.getListPath()) === undefined) {
```

packages/netsuite-adapter/src/filters/restore_deleted_list_items_without_scriptid.ts
```diff
@@ -0,0 +1,66 @@
...
+          return
+        }
+        if (_.isUndefined(_.get(after.value, getters.getListPath()))) {
```
dantal4, line 49: nit
```suggestion
        if (_.get(after.value, getters.getListPath())) === undefined) {
```

packages/netsuite-adapter/src/filters/restore_deleted_list_items_without_scriptid.ts
```diff
@@ -0,0 +1,66 @@
...
+          setPath(
+            after,
+            after.elemID.createNestedID(...getters.getListPath()),
+            _.get(before.value, getters.getListPath())
```
dantal4, line 53: can you call `getters.getListPath()` only once?

packages/netsuite-adapter/src/change_validators/remove_list_item_without_scriptid.ts
SeggevHaimovich: you mean the situation when the user deletes all of the permissions inside the list?
I think we should have different messages for deleting the list itself and deleting permissions in the list, and the situation when the user deletes permissions in the list fits to the second option better.

packages/netsuite-adapter/src/filters/restore_deleted_list_items.ts
SeggevHaimovich: I will add this to the CV as well

packages/netsuite-adapter/src/change_validators/remove_list_item_without_scriptid.ts
dantal4: we might not know the difference between empty list and removed list, because there are places that remove empty lists/objects, and I don't think that there is difference from the user's perspective, that will probably get to that situation by copying a role from one environment to the other, where in the source the role exist and have no permissions in it and in the target the role exists and have permissions in it.

SeggevHaimovich: So you think the message should be the same? write implicitly the whole permissions that have been deleted?

dantal4: yes 👍

packages/netsuite-adapter/src/change_validators/remove_list_item.ts
dantal4: image

dantal4 (approved):
Great Work!
Please update the CV message+detailedMessage according to @rozenmanofer 's comment in both CVs.

PR #5364, opened 2024-01-30
SALTO-5353: Fix CV for Automation to assets

idoamit8:
In this PR I fix the CV that warn when customer has automation to assets without assets support

---

None

---
_Release Notes_: 
_Jira Adapter:_
* Fixed automationToAssetsValidator validator in jira to warn when customer doesn't have assets support.

---
_User Notifications_: 
None


omrilit (approved):
As spoke f2f, please consult with Lior about the messaging here.
I think that we should consider telling the user that Salto doesn't support it yet

PR #5367, opened 2024-01-30
SALTO-5243 - Salesforce: Create deploy warning when data instance may have been partially deployed due to non existing fields

tomsellek:
Create Deploy Warning when attempting to deploy data instance with missing fields.

---
~Based on https://github.com/salto-io/salto/pull/5317~

Take 2 of https://github.com/salto-io/salto/pull/5277
When users attempt to deploy data instances with values to fields that do not exist in the CustomObject, we warn the users that the instance was most likely partially deployed in Salesforce.

The implementation does not intercept the applied changes, as we don't really know what was actually deployed to the service.
image
image

---
_Release Notes_: 
Salesforce Adapter:

Add Deploy Warning when attempting to deploy Data Instance with fields that do not exist on the CustomObject type.

---
_User Notifications_: 


tamtamirr (approved):
LGTM

PR #5369, opened 2024-01-30
SALTO-5356: Return unresolved elemIds from unresolved references validator

shir-reifenberg:
We should be able to return the list of missing elemIds from a change with unresolved references
---

See ticket for more details

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/adapter-api/src/adapter.ts
```diff
@@ -116,11 +116,17 @@ export type DependencyError = ChangeError & {
...
 }
 
+export type UnresolvedReferenceError = ChangeError & {
+  unresolvedElemIds: ElemID[]
+}
+
 export const isDependencyError = (err: ChangeError): err is DependencyError => 'causeID' in err
 
-export type ChangeValidator = (
+export const isUnresolvedReferenceError = (err: ChangeError): err is UnresolvedReferenceError => 'unresolvedElemIds' in err
```
ori-moisis, line 125: when adding a new error type, I think we need to be consistent with the SaltoErrorType definition:
https://github.com/salto-io/salto/blob/main/packages/adapter-api/src/error.ts#L21

adamro (approved):
LGTM what you did

I wonder if the Warnings returned from `incomingUnresolvedReferencesValidator` should also be of this type.

shir-reifenberg:
> LGTM what you did
> 
> I wonder if the Warnings returned from `incomingUnresolvedReferencesValidator` should also be of this type.

I think for now it's redundant... we can add it when we need it

PR #5370, opened 2024-01-30
SALTO-5357 - Salesforce: Log deployed data records

tomsellek:
~Introduce a new undocumented config `client.additionalDebugging`. When this config is set we~ log (at `trace` level) the parameters passed to `SalesforceClient.bulkLoadOperation`.

---
Test:
 - Deploy data records with the config set to `true`, observe trace-level logs
 - Deploy data records without the config, confirm no logs of data records

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/client/client.ts
+++ b/packages/salesforce-adapter/src/client/client.ts
@@ -968,8 +968,10 @@ export default class SalesforceClient {
     type: string,
     operation: BulkLoadOperation,
     records: SalesforceRecord[]
-  ):
-  Promise<BatchResultInfo[]> {
+  ): Promise<BatchResultInfo[]> {
+    if (this.config?.additionalDebugging) {
+      log.trace('client.bulkLoadOperation: %s %d records of type %s: %o', operation, records.length, type, records)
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/src/client/client.ts
```diff
@@ -968,8 +968,10 @@ export default class SalesforceClient {
     type: string,
     operation: BulkLoadOperation,
     records: SalesforceRecord[]
-  ):
-  Promise<BatchResultInfo[]> {
+  ): Promise<BatchResultInfo[]> {
+    if (this.config?.additionalDebugging) {
```
amitgoldarb, line 972: Considering this is trace, I think it's probably okay to not have this behind another debug flag. I think we can safely remove this for now.

amitgoldarb (approved):
Approved with 1 small comment

packages/salesforce-adapter/src/client/client.ts
tomsellek: Ack

PR #5371, opened 2024-01-31
SALTO-5319: support templated static files - move parser to a different package

edenhassid:
support templated static files - move parser to a different package

---

---
_Release Notes_: 
move parser to a different package

---
_User Notifications_: 
None


netama (approved):

netama:
let's make sure there are no dependencies in the SAAS before merging

PR #5374, opened 2024-01-31
SALTO-2392 more account specific values resolving in workflows

dantal4:
- map more fields with ASV to the data types they are referring
- fetch more SuiteQL tables
- fetch internal id maps from saved search too
- add option to limit saveSearch query results

---

---
_Release Notes_: 
Netsuite Adapter:
- More account specific values resolving in workflows

---
_User Notifications_: 
Netsuite Adapter:
- When `fetch.resolveAccountSpecificValues` is `true`, more `ACCOUNT_SPECIFIC_VALUE` values in workflows will be resolved.


omrilit (approved):
💯

PR #5375, opened 2024-01-31
Increase global concurrency limit in NS APIs

dantal4:
The global concurrency limit should be the maximum allowed limit from SDF+SuitApp APIs plus one, because every integration is allowed to have the maximum concurrency of the account minus one.

---

---
_Release Notes_: 
Netsuite Adapter:
- Increase global concurrency limit in NS APIs

---
_User Notifications_: 
None


omrilit (approved):
Thanks :)

PR #5376, opened 2024-01-31
SALTO-5365 - Salesforce: Create a deploy-time dependency between data records and their type

tomsellek:
We want to make sure changes to the data type are deployed before changes to data records of that type (and that removal of data records takes place before removal of the data type). Additionally, we suspect that deploying metadata along with data leads to contention on the Salesforce DB and to UNABLE_TO_LOCK_DB_ROW errors.

---

Right now we create a dependency on any kind of instance change. Is this overly strict?

---
_Release Notes_: 
Salesforce: Ensure types are deployed before their data records where needed

---
_User Notifications_: 
N/A


packages/salesforce-adapter/src/dependency_changer.ts
```diff
@@ -0,0 +1,77 @@
...
+      source: changeId,
+      target: typeChangeId,
+    },
+  }
+}
+
+const dataRecordToAssociatedType: DependencyChanger = async changes => {
+  // Note that we don't handle removal yet. We should probably create a reverse dependency to ensure we delete all
+  // records before we delete their type.
+  const instanceChanges: [ChangeId, Change<InstanceElement>][] = Array.from(changes.entries())
```
tamtamirr, line 49:
```suggestion
  const dataInstanceChanges: [ChangeId, Change<InstanceElement>][] = Array.from(changes.entries())
```
or `instanceOfCustomObjectChanges`

packages/salesforce-adapter/src/dependency_changer.ts
```diff
@@ -0,0 +1,77 @@
...
+    .filter(([, change]) => isInstanceChange(change))
+    .filter(([, change]) => isAdditionOrModificationChange((change)))
+    .filter(([, change]) => (
+      isInstanceOfCustomObjectSync(getChangeData(change))
+    )) as [ChangeId, Change<InstanceElement>][]
+  const typeChanges = Array.from(changes.entries())
+    .filter(([, change]) => isCustomObject(getChangeData(change)))
```
tamtamirr, line 56:
```suggestion
    .filter(([, change]) => isCustomObjectSync(getChangeData(change)))
```
This statement is always true...

Please make the UTs more robust so they would catch this edge-case. Considering they don't contain any other change but a single change on a type, they didn't catch this.

packages/salesforce-adapter/src/dependency_changer.ts
```diff
@@ -0,0 +1,77 @@
...
+  const typeToChangeIdMap = new Map<ElemID, ChangeId>(typeChanges
```
tamtamirr, line 57:
```suggestion
  const typeElemIdToChangeId = new Map<ElemID, ChangeId>(typeChanges
```

packages/salesforce-adapter/test/dependency_changer.test.ts
```diff
@@ -0,0 +1,107 @@
...
+  )
+)
+
+describe('Dependency Changer', () => {
+  let instanceChange: Change<InstanceElement>
+  let typeChange: Change<ObjectType>
+  let changes: Map<ChangeId, Change>
+  let deps: DependencyChange[]
+  beforeEach(() => {
+    typeChange = toChange({ after: mockTypes.Account })
```
tamtamirr, line 38: Nit: instead of invoking `getChangeData` across the code, I think it'd be better to assign it to a variable `changedType`

tamtamirr (commented):
Added some comments.
Let's also test this in various scenarios with various scopes.

packages/salesforce-adapter/src/dependency_changer.ts
tomsellek: So, funny story - this is actually (accidentally) not a bug, because this logic maps instances to types, and we filter the instances correctly. So there will be extra entries in `typeToChangeIdMap`, but no instance would map to them (because we only map instances of custom objects), and everything will come out OK. 
I will still fix this, of course, but I thought it's interesting.

packages/salesforce-adapter/src/dependency_changer.ts
tomsellek: ack

packages/salesforce-adapter/src/dependency_changer.ts
tomsellek: Ack

packages/salesforce-adapter/test/dependency_changer.test.ts
tomsellek: Ack

tamtamirr (approved):
LGTM

PR #5377, opened 2024-01-31
SALTO-5369 - Salesforce: Handle circular dependency of CPQ PriceRule/PriceCondition

tomsellek:
Similar to the special handling we have for SBAA `ApprovalRule`/`ApprovalCondition`, we need to do the same thing for CPQ `PriceRule`/`PriceCondition`.

---

I made the existing code more generic and implemented two special cases for SBAA and CPQ

---
_Release Notes_: 
Salesforce: Can now deploy new CPQ PriceRule records with 'Custom' ConditionMet value that reference new PriceCondition records.

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/custom_object_instances_deploy.ts
+++ b/packages/salesforce-adapter/src/custom_object_instances_deploy.ts
@@ -44,7 +44,7 @@ import {
 } from './constants'
 import { getIdFields, transformRecordToValues } from './filters/custom_objects_instances'
 import {
-  apiNameSync, buildSelectQueries, getFieldNamesForQuery, isHiddenField, isInstanceOfTypeChangeSync,
+  apiNameSync, buildSelectQueries, getFieldNamesForQuery, isHiddenField, isInstanceOfTypeChangeSync, SoqlQuery,
 } from './filters/utils'
 import { isListCustomSettingsObject } from './filters/custom_settings_filter'
 import { SalesforceRecord } from './client/types'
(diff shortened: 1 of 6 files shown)


tamtamirr (approved):
LGTM.
Before merging, let's run sanity on the flow of both SBQQ and SBAA use-cases (The old and the new)

packages/salesforce-adapter/src/custom_object_instances_deploy.ts
```diff
@@ -771,46 +819,109 @@ const deployAddCustomApprovalRulesAndConditions = async (
...
+  switch (groupId) {
+    case ADD_SBAA_CUSTOM_APPROVAL_RULE_AND_CONDITION_GROUP: {
+      return deployAddCustomApprovalRulesAndConditions(changes, client, dataManagement)
+    }
+    case ADD_CPQ_CUSTOM_PRICE_RULE_AND_CONDITION_GROUP: {
+      return deployAddCustomPriceRulesAndConditions(changes, client, dataManagement)
+    }
+    default: {
+      return deploySingleTypeAndActionCustomObjectInstancesGroup(changes, client, groupId, dataManagement)
+    }
```
tamtamirr, line 925: nit: Not sure why the switch is needed here? 🤔 
Couldn't we had `||` to the original if?
In general there's not a lot of use of switch cases inside the OSS.

packages/salesforce-adapter/test/group_changes.test.ts
```diff
@@ -232,8 +232,8 @@ describe('Group changes function', () => {
       result = await getChangeGroupIds(changeMap)
     })
     it('should create correct groups', () => {
```
tamtamirr, line 234: I think we should make this UT more robust and have all the different groupings:

1. SBAA Custom Approval Rules & Conditions.
2. SBQQ Custom Approval Rules & Conditions.
3. Salesforce Metadata

tamtamirr (commented):
LGTM! Added some comments

packages/salesforce-adapter/test/group_changes.test.ts
tomsellek: Ack

PR #5382, opened 2024-02-01
SALTO-5368: Log all errors when validating credentials in Okta

shir-reifenberg:
as title

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/client/connection.ts
+++ b/packages/okta-adapter/src/client/connection.ts
@@ -15,6 +15,7 @@
 */
 import { AccountInfo } from '@salto-io/adapter-api'
 import { client as clientUtils, client, auth as authUtils } from '@salto-io/adapter-components'
+import { safeJsonStringify } from '@salto-io/adapter-utils'
 import { logger } from '@salto-io/logging'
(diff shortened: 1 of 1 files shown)


omrilit (approved):
💯

packages/okta-adapter/src/client/connection.ts
```diff
@@ -52,11 +53,8 @@ export const validateCredentials = async (_creds: {
       isProduction: accountType === PRODUCTION_ACCOUNT_TYPE,
     }
   } catch (error) {
-    if (error.response?.status === 401) {
-      log.error('Failed to validate credentials: %s', error)
-      throw new client.UnauthorizedError('Invalid Credentials')
-    }
-    throw error
+    log.error('Failed to validate credentials, error: %s, stack: %s', safeJsonStringify({ data: error?.response?.data, status: error?.response?.status }), error.stack)
+    throw new client.UnauthorizedError('Invalid Credentials')
```
netama, line 57: do we want all errors treated as unauthorized, even if there was for example service downtime?

omrilit: This is the same behavior we currently have in NS, SF & Jira (maybe also others)

PR #5389, opened 2024-02-04
SALTO-5333 transformValues takes into count additionalProperties annotation 

omrifarkash:
Changes on this PR:

- In transformValues function, if a value doesn't have a corresponding field in the provided type, we will try to get a type from _additional_properties annotation.
if _additional_properties = { refType, annotations? } -> defined a Field with that type and annotations
if _additional_properties = false -> No field for the given value (same as today)
if _additional_properties is not defined -> defined a Field with that type unknown

- Fix side effects of this change (UTs and real behavior changes)

---

---
_Release Notes_: 
In transformValues function, if a value doesn't have a corresponding field in the provided type, we will try to get a type from _additional_properties annotation.

---
_User Notifications_: 
None


packages/netsuite-adapter/src/transformer.ts
```diff
@@ -116,7 +116,7 @@ export const createInstanceElement = async (
       // We sometimes get empty strings that we want to filter out
       return undefined
     }
-    if (!isPrimitiveType(fieldType) || !isPrimitiveValue(value)) {
+    if (!isPrimitiveType(fieldType) || !isPrimitiveValue(value) || fieldType.primitive === PrimitiveTypes.UNKNOWN) {
```
omrilit, line 119: NS changes LGTM, @dantal4 FYI

dantal4: how is this different from the previous code @omrifarkash ?

omrifarkash: @dantal4 
Since UKNOWN is a primitive type, on instances with values = { someField: 'T' } where the type of the field "someField" is UKNOWN we won't get inside that if.

This happens after the change I've made in transformValues. If some value doesn't have a matched field in the instance's objectType, and ADDITIONAL_PROPERTIES annotations isn't defined, we are handling the value as its field is UKNOWN

packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -355,10 +355,11 @@ export const extractFullNamesFromValueList = (instanceValues: { [INSTANCE_FULL_N
 
 export const buildAnnotationsObjectType = (annotationTypes: TypeMap): ObjectType => {
   const annotationTypesElemID = new ElemID(SALESFORCE, 'AnnotationType')
-  return new ObjectType({ elemID: annotationTypesElemID,
+  const annotationObjType = new ObjectType({ elemID: annotationTypesElemID,
     fields: Object.assign({}, ...Object.entries(annotationTypes)
       .concat(Object.entries(CoreAnnotationTypes))
       .map(([name, type]) => ({ [name]: { refType: createRefToElmWithValue(type) } }))) })
+  return setAdditionalPropertiesAnnotation(annotationObjType, false)
```
omrifarkash, line 362: @tamtamirr can you verify this change please?

packages/zendesk-adapter/src/config.ts
```diff
@@ -696,6 +696,10 @@ export const DEFAULT_TYPES: ZendeskApiConfig['types'] = {
           fieldType: 'string',
           restrictions: { enforce_value: true, values: ['unknown', 'verified', 'failed'] },
         },
+        {
+          fieldName: 'username',
+          fieldType: 'string',
+        },
```
omrifarkash, line 702: @edenhassid can you verify this change please?

packages/zendesk-adapter/src/filters/custom_field_options/custom_object_field_options.ts
```diff
@@ -49,6 +49,9 @@ export const customObjectFieldOptionType = new ObjectType({
     raw_name: { refType: BuiltinTypes.STRING },
     value: { refType: BuiltinTypes.STRING },
   },
+  annotations: {
+    [CORE_ANNOTATIONS.ADDITIONAL_PROPERTIES]: false,
+  },
```
omrifarkash, line 54: @edenhassid  can you verify this change please?

packages/adapter-utils/src/additional_properties.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export const extractAdditionalPropertiesField = (objType: ObjectType, name: string): Field | undefined => {
+  const additionalProperties = objType.annotations[CORE_ANNOTATIONS.ADDITIONAL_PROPERTIES]
+  if (additionalProperties === undefined) {
+    return new Field(objType, name, BuiltinTypes.UNKNOWN)
+  }
+  if (isAdditionalPropertiesAnnotation(additionalProperties)) {
+    return new Field(objType, name, additionalProperties.refType, additionalProperties.annotations)
+  }
+  // case additionalProperties = false
```
omrifarkash, line 44: We should reach this exit point only with additionalProperties = false, This will be enforced in the second part of this task (SALTO 5332).

packages/netsuite-adapter/test/change_validators/immutable_changes.test.ts
```diff
@@ -260,7 +260,7 @@ describe('customization type change validator', () => {
 
       new Array<ChangeDataType>(entityCustomFieldInstance, type, field).forEach(async element => {
         const change = await immutableChangesValidator([toChange({ after: element })])
-        expect(change).toBeFalsy()
+        expect(change).toEqual([])
```
omrifarkash, line 263: No idea why it passed until now but it got broken on my PR.
The return type of `immutableChangesValidator` is `ChangError[]` so I have no idea how `change` on line 263 was equal to `false`

cc: @dantal4

packages/adapter-utils/src/additional_properties.ts
```diff
@@ -0,0 +1,47 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+import { BuiltinTypes, CORE_ANNOTATIONS, Field, isType, Element, ObjectType, TypeElement, Value } from '@salto-io/adapter-api'
+import _ from 'lodash'
+
+type AdditionalPropertiesAnnotation = {
+  refType: TypeElement
```
ori-moisis, line 21: will this actually work?
I don't think we support putting elements directly as values in annotations / instances, that is why we have `ReferenceExpression`

PR #5390, opened 2024-02-04
[WIP] GetReferenceTargetDependencies

shir-reifenberg:

---

---
_Release Notes_: 

---
_User Notifications_: 


packages/workspace/src/workspace/workspace.ts
```diff
@@ -344,6 +343,53 @@ export const deserializeReferenceTree = async (data: string): Promise<ReferenceT
...
+  const wsSearchableNames = new Set(await ws.getSearchableNamesOfEnv(envName))
+  const elemIdsToSkipSet = new Set(elemIDsToSkip.map(id => id.getFullName()))
+  const baseLevelIds = elemIDs.map(id => id.createBaseID().parent)
+  const [resolvedIds, missingIds] = _.partition(baseLevelIds, elemID => wsSearchableNames.has(elemID.getFullName()))
+
+  const result: Record<string, ElemID[]> = {}
+  const stack = [...resolvedIds]
+  const visited: Set<string> = new Set()
+  while (stack.length > 0) {
+    const currentId = stack.pop() as ElemID
```
adamro, line 370: Is there really a reason to do this one by one and not each "level" in parallel with Promise.all?

packages/workspace/src/workspace/workspace.ts
```diff
@@ -344,6 +343,53 @@ export const deserializeReferenceTree = async (data: string): Promise<ReferenceT
...
+}): Promise<{ completed: Record<string, ElemID[]>; missing: ElemID[] }> => {
+  const wsSearchableNames = new Set(await ws.getSearchableNamesOfEnv(envName))
+  const elemIdsToSkipSet = new Set(elemIDsToSkip.map(id => id.getFullName()))
+  const baseLevelIds = elemIDs.map(id => id.createBaseID().parent)
+  const [resolvedIds, missingIds] = _.partition(baseLevelIds, elemID => wsSearchableNames.has(elemID.getFullName()))
+
+  const result: Record<string, ElemID[]> = {}
+  const stack = [...resolvedIds]
+  const visited: Set<string> = new Set()
+  while (stack.length > 0) {
```
adamro, line 369: No strong opinion, but what was the reasoning to choose while over recursion here?

packages/workspace/src/workspace/workspace.ts
```diff
@@ -344,6 +343,53 @@ export const deserializeReferenceTree = async (data: string): Promise<ReferenceT
...
+    const currentId = stack.pop() as ElemID
+    if (!visited.has(currentId.getFullName())) {
+      visited.add(currentId.getFullName())
+      // eslint-disable-next-line no-await-in-loop
+      const references = await ws.getElementOutgoingReferences(currentId, envName ?? ws.currentEnv())
+      const relevantElemIds = references
+        .filter(ref => ref.type !== 'weak')
```
adamro, line 376: Maybe add the decision to include weak or not to `getElementOutgoingReferences`? As an optional param that defaults to true so it's not a breaking change to the api

packages/workspace/src/workspace/workspace.ts
```diff
@@ -344,6 +343,53 @@ export const deserializeReferenceTree = async (data: string): Promise<ReferenceT
   return new collections.treeMap.TreeMap(elemIdsEntries)
 }
 
+/*
+* Get reference dependecies for a given list of elemIDs within a workspace
+* filters out weak references and any elemIDs from elemIDsToSkip
+*/
+export const getReferenceTargetDependencies = async ({
```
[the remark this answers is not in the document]
shir-reifenberg: The reason I put it here is that extracting it to a separate file will cause dependency cycle as this function is using workspace methods but also being used by `listUnresolved`.
I will add it to `index.ts` though

packages/workspace/src/workspace/workspace.ts
```diff
@@ -344,6 +343,53 @@ export const deserializeReferenceTree = async (data: string): Promise<ReferenceT
...
+  ws,
+  elemIDs,
+  elemIDsToSkip = [],
+  envName,
+}:{
+  ws: Workspace
+  elemIDs: ElemID[]
```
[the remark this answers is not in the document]
shir-reifenberg: mmm isn't those names are quite specific to our usage? what about something like `initialElemIDs`?

packages/workspace/src/workspace/workspace.ts
shir-reifenberg: I think it's more readable and also slightly better performance wise

packages/cli/src/commands/element.ts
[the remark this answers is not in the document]
shir-reifenberg: Didn't thought it deserves a separate PR 😄 Can split if you you prefer

packages/workspace/src/workspace/workspace.ts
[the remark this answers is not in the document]
adamro: You mean it would import the Workspace type cause it's one of the params, not the methods, but gotcha :D
Honestly there isn't a *real* reason this function isn't added to the Workspace api considering how it's written and what we did with other, similar functions. I know this is Ori's guidance so I won't push against it.

packages/workspace/src/workspace/workspace.ts
adamro: You didn't optimize for performance with doing this one by one and the readable part is debatable (I think the opposite honestly), but if you do the performance optimization from the other comment I am OK with keeping this

packages/workspace/src/workspace/workspace.ts
```diff
@@ -1250,22 +1315,7 @@ export const loadWorkspace = async (
...
-
-      const idPath = id.createBaseID().path.join(ElemID.NAMESPACE_SEPARATOR)
-      const references = idPath === ''
-      // Empty idPath means we are requesting the base level element, which includes all references
-        ? referencesTree.values()
-        : referencesTree.valuesWithPrefix(idPath)
-
-      return _.uniqBy(Array.from(references).flat(), ref => ref.id.getFullName())
-    },
+    getElementOutgoingReferences,
```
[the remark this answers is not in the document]
shir-reifenberg: otherwise I couldn't use it inside `listUnresolved`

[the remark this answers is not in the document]
shir-reifenberg: oops you're right, removed

packages/workspace/src/workspace/workspace.ts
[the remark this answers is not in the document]
shir-reifenberg: I agree, what about - 

```suggestion
export const listElementsDependenciesInWorkspace = async ({
  workspace,
  elemIDsToFind,
  elemIDsToSkip = [],
  envToListFrom,
}:{
  workspace: Workspace
  elemIDsToFind: ElemID[]
  elemIDsToSkip?: ElemID[]
  envToListFrom?: string
}): Promise<{ dependencies: Record<string, ElemID[]>; missing: ElemID[] }> => {
```

packages/workspace/src/workspace/workspace.ts
```diff
@@ -366,6 +370,57 @@ export const deserializeReferenceTree = async (data: string): Promise<ReferenceT
...
+    const [foundIds, missingIds] = _.partition(baseLevelIds, elemID => workspaceBaseLevelIds.has(elemID.getFullName()))
+
+    const result: Record<string, ElemID[]> = {}
+    const stack = [...foundIds]
+    const visited: Set<string> = new Set()
+    while (stack.length > 0) {
+      const currentLevelIds = stack.splice(0, stack.length)
+      // eslint-disable-next-line no-await-in-loop
+      await Promise.all(
+        currentLevelIds.map(async currentId => {
```
adamro, line 401: I would write it a bit differently - Run the Promise.all only on the async part and then do the sync part in a for loop
ie.
```
const nonVisitedCurrentLevelIds = currentLevelIds.filter(id => !visited.has(id.getFullName())
const idsToReferences = await Promise.all(
   nonVisitedCurrentLevelIds.map(async currentId => [currentId, await workspace.getElementOutgoingReferences(currentId, envToListFrom, false)]
)
.... (The for loop with the rest of the logic goes here :) )
```

(The visited check before the async code is also more efficient)

packages/workspace/src/workspace/workspace.ts
```diff
@@ -366,6 +370,57 @@ export const deserializeReferenceTree = async (data: string): Promise<ReferenceT
...
+  envToListFrom?: string
+}): Promise<{ dependencies: Record<string, ElemID[]>; missing: ElemID[] }> =>
+  log.time(async () => {
+    const workspaceBaseLevelIds = new Set(await workspace.getSearchableNamesOfEnv(envToListFrom))
+    const elemIdsToSkipSet = new Set(elemIDsToSkip.map(id => id.getFullName()))
+    const baseLevelIds = elemIDsToFind.map(id => id.createBaseID().parent)
+    const [foundIds, missingIds] = _.partition(baseLevelIds, elemID => workspaceBaseLevelIds.has(elemID.getFullName()))
+
+    const result: Record<string, ElemID[]> = {}
+    const stack = [...foundIds]
```
adamro, line 395: We should probably use a different name since we are no longer using it as a real stack, maybe `elementsToProccess`? 🤔

packages/workspace/src/workspace/workspace.ts
```diff
@@ -366,6 +370,57 @@ export const deserializeReferenceTree = async (data: string): Promise<ReferenceT
...
+  elemIDsToSkip = [],
+  envToListFrom,
+}: {
+  workspace: Workspace
+  elemIDsToFind: ElemID[]
+  elemIDsToSkip?: ElemID[]
+  envToListFrom?: string
+}): Promise<{ dependencies: Record<string, ElemID[]>; missing: ElemID[] }> =>
+  log.time(async () => {
+    const workspaceBaseLevelIds = new Set(await workspace.getSearchableNamesOfEnv(envToListFrom))
```
adamro, line 389: Can we make sure we have a way to understand how much time this specific line took? (ie. log.time or logs before/after etc.) cause I suspect we will end up wanting to see if this call is the problematic one or all the others

packages/workspace/src/workspace/workspace.ts
```diff
@@ -1469,56 +1536,29 @@ export const loadWorkspace = async (
...
+        Object.values(dependencies)
+          .flat()
+          .map(elemId => elemId.getFullName())
+          .concat(Object.keys(dependencies)),
+      )
       return {
-        found: compact(completed.sort().map(ElemID.fromFullName)),
-        missing: compact(missing.sort().map(ElemID.fromFullName)),
+        found: compact(completedElemIds.sort().map(ElemID.fromFullName)),
+        missing: compact(_.sortBy(missing, id => id.getFullName())),
```
adamro, line 1561: why this change?

packages/workspace/src/workspace/workspace.ts
```diff
@@ -1119,7 +1184,7 @@ export const loadWorkspace = async (
     return currentWorkspaceState.states[env].changedBy.isEmpty()
   }
 
-  return {
+  const ws: Workspace = {
```
adamro, line 1187: nit: We can't use `workspace` here instead of `ws` cause we already have it in the scope? Cause otherwise change please

packages/workspace/src/workspace/workspace.ts
```diff
@@ -366,6 +370,67 @@ export const deserializeReferenceTree = async (data: string): Promise<ReferenceT
...
+
+          const [foundElemIDs, missingElemIDs] = _.partition(relevantElemIds, elemID =>
+            workspaceBaseLevelIds.has(elemID.getFullName()),
+          )
+          result[currentId.getFullName()] = foundElemIDs
+          return { foundElemIDs, missingElemIDs }
+        }),
+      )
+
+      currentLevelResult.flatMap(entry => entry.missingElemIDs).forEach(id => missingIds.push(id))
```
adamro, line 423: I still think we are over-looping here but I don't think it will be critical. You can keep it like this :)

(Let's make sure we can know how much time each part takes)

adamro (approved):
🔥

packages/workspace/src/workspace/workspace.ts
shir-reifenberg: cause now `missing` is `ElemID[]` and not `string[]`, so it has to be sorted by full name (this is also aligned with this line https://github.com/salto-io/salto/blob/main/packages/workspace/src/workspace/workspace.ts#L1478)

PR #5392, opened 2024-02-04
SALTO-5378 rename getSinglePage to get

netama:
renamed the function, since there is now generic handling and we will soon generalize the requests to support all HTTP methods.

---
_Release Notes_: 
None

---
_User Notifications_: 
None


shir-reifenberg (approved):
🤯 🚀

PR #5393, opened 2024-02-04
SALTO-5264: Add check for unresolved reference in translation cv

DanielBarros1:
Return Change Validator error instead of crashing when a locale is unresolved while checking for translations.
Also added a test case to make sure we return an error elegantly.

---

_Test plan_:
Before applying the fix:
Screenshot 2024-02-04 at 16 49 51

After applying the fix:
Screenshot 2024-02-04 at 16 50 05

---
_Release Notes_: 

Zendesk: 
Return Change Validator error instead of crashing when a locale is unresolved while checking for translations

### DIFF
--- a/packages/zendesk-adapter/src/change_validators/translation_for_default_locale.ts
+++ b/packages/zendesk-adapter/src/change_validators/translation_for_default_locale.ts
@@ -16,10 +16,10 @@
 import {
   ChangeValidator,
   getChangeData, InstanceElement,
-  isAdditionOrModificationChange, isInstanceElement, isReferenceExpression,
+  isAdditionOrModificationChange, isInstanceElement,
 } from '@salto-io/adapter-api'
 import Joi from 'joi'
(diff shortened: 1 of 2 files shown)


packages/zendesk-adapter/src/change_validators/translation_for_default_locale.ts
```diff
@@ -53,7 +53,7 @@ const noTranslationForDefaultLocale = (instance: InstanceElement): boolean => {
   const sourceLocale = instance.value.source_locale
   const translation = instance.value.translations
     .filter(isTranslation)
-    .find(tran => (isReferenceExpression(tran.locale)
+    .find(tran => (isResolvedReferenceExpression(tran.locale)
       ? tran.locale.value.value.locale === sourceLocale
```
edenhassid, line 57: can you also check that `isInstanceElement(tran.locale.value)`? and if it is not add a `log.warn`

edenhassid (commented):
nice! 🤩 added one small comment

edenhassid (approved):
LGTM! 🔥

PR #5394, opened 2024-02-04
SALTO-4056: Allow configuring idFields that only exist in some elements

shir-reifenberg:
We now allow configuring idFields that might be null in some elements. 

---

This continues - https://github.com/salto-io/salto/pull/4056

---
_Release Notes_: 
None 

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/elements/instance_elements.ts
+++ b/packages/adapter-components/src/elements/instance_elements.ts
@@ -49,9 +49,8 @@ export const joinInstanceNameParts = (
   // if nameParts is empty, we assume it is intentional
   (nameParts.length === 0 || nameParts.some(part => part !== undefined && part !== ''))
     ? nameParts
-      .map(part => (part === undefined ? '' : part))
+      .filter(part => part !== undefined)
       .map(String)
-      .filter((part, idx) => part !== '' || idx !== nameParts.length - 1) // if the last part is empty, avoid adding an extra '_'
       .join('_')
     : undefined
 )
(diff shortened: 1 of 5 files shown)


packages/adapter-components/src/elements/instance_elements.ts
```diff
@@ -53,20 +53,29 @@ const getNameMapping = (
...
 
 export const joinInstanceNameParts = (
-  nameParts: string[],
-): string | undefined => (nameParts.every(part => part !== undefined && part !== '') ? nameParts.map(String).join('_') : undefined)
+  nameParts: unknown[],
+): string | undefined => (
+  // if nameParts is empty, we assume it is intentional
+  (nameParts.length === 0 || nameParts.some(part => part !== undefined && part !== ''))
+    ? nameParts
+      .filter(part => part !== undefined)
```
netama, line 61: comparing to the original PR - we don't want to avoid the extra `_` in case the last part is empty?

packages/adapter-components/src/filters/referenced_instance_names.ts
```diff
@@ -52,7 +52,7 @@ const getFirstParentElemId = (instance: InstanceElement): ElemID | undefined =>
 const createInstanceReferencedNameParts = (
   instance: InstanceElement,
   idFields: string[],
-): string[] => idFields.map(
+): unknown[] => idFields.map(
```
netama, line 55: nit: why not `(string | undefined)[]`?

netama (approved):
thanks

packages/adapter-components/src/elements/instance_elements.ts
shir-reifenberg: I'll replace it with .filter(part => !_.isEmpty(part))
then we shouldn't have an extra _

PR #5395, opened 2024-02-05
SALTO-5326: log errored GQL responses

omrilit:

---

---
_Release Notes_: 

---
_User Notifications_: 


shir-reifenberg (approved):
LGTM so approving
but isn't `IssueLayout` will fail silently as we're catching the error and returning undefined? (in `getLayoutResponse`)

omrilit:
> LGTM so approving but isn't `IssueLayout` will fail silently as we're catching the error and returning undefined? (in `getLayoutResponse`)

It returned as undefined but then checked whether it contains the requested data as part of isIssueLayoutResponse

PR #5400, opened 2024-02-05
SALTO-3432 resolve account specific values in data elements

dantal4:
Resolve account specific values in references in data elements.

In fetch:
- match the `internalId` of the reference in fetch to the suiteQL instances' internalId-to-name map, or use the reference `name`
- set the reference value to be:
```
ref = {
  id = "[ACCOUNT_SPECIFIC_VALUE] (<type>) (<name>)"
}
```

In deploy:
- find the matching `type` & `name` in the suiteQL instances' internalId-to-name map, and replace `id` with the real `interalId` of the referenced object
- return change error if `name` was not found

---

TODO:
- [x] add tests

---
_Release Notes_: 
Netsuite Adapter:
- Resolve account specific values in data elements in fetch & deploy (behind `fetch.resolveAccountSpecificValue`)

---
_User Notifications_: 
Netsuite Adapter:
- In case that `fetch.resolveAccountSpecificValue` is `true` in the adapter config, `ACCOUNT_SPECIFIC_VALUE` values will be resolved to the form of `[ACCOUNT_SPECIFIC_VALUE] (<type>) (<name>)`


packages/netsuite-adapter/src/filters/data_account_specific_values.ts
```diff
@@ -0,0 +1,393 @@
...
+  }
+  return unknownTypeReferencesInstance.value[path][internalId]
+}
+
+const getNameFromSuiteQLTableInstance = (
+  suiteQLTableInstance: InstanceElement,
+  internalId: string,
+  fallbackName: string | undefined
+): string | undefined => {
+  const internalIdsMap = getSuiteQLTableInternalIdsMap(suiteQLTableInstance)
```
omrilit, line 104: IIUC this method may be called a lot of in the fetch, once for WFs and once per data instance.
I think that we should make sure that we call it 1-2 times in the whole fetch as those maps could be quite large and it could have some impact on the fetch performance .

packages/netsuite-adapter/src/filters/data_account_specific_values.ts
```diff
@@ -0,0 +1,393 @@
...
+
+    addReferenceTypes(elements)
+  },
+  preDeploy: async changes => {
+    if (!config.fetch.resolveAccountSpecificValues) {
+      return
+    }
+    const relevantChangedInstances = changes
+      .filter(isAdditionOrModificationChange)
+      .flatMap(change => Object.values(change.data))
```
omrilit, line 378: why do we need to manipulate also the `before` data instance?

packages/netsuite-adapter/src/filters/data_account_specific_values.ts
```diff
@@ -0,0 +1,393 @@
...
+  if (!isNestedObject(path, value)) {
+    return { value }
+  }
+  if (typeof value[ID_FIELD] !== 'string' || !_.isEmpty(_.omit(value, ID_FIELD))) {
+    return { value }
+  }
+
+  const regexRes = value[ID_FIELD].match(resolvedAccountSpecificValueRegex)
+  if (regexRes?.groups === undefined) {
+    log.debug('replacing field id with internalId in path %s', path.getFullName())
```
omrilit, line 288: Could you please add comment explaining why?

packages/netsuite-adapter/src/filters/data_account_specific_values.ts
```diff
@@ -0,0 +1,393 @@
...
+const resolvedAccountSpecificValueRegex = new RegExp(`^${_.escapeRegExp(ACCOUNT_SPECIFIC_VALUE)} \\((?<${REGEX_TYPE}>\\w+)\\) \\((?<${REGEX_NAME}>.*)\\)$`)
+
+const toResolvedAccountSpecificValue = (
+  { type, name, internalId }: {
+    type: string
+    name: string | undefined
+    internalId: string
+  }
+): ResolvedAccountSpecificValue => ({
+  id: `${ACCOUNT_SPECIFIC_VALUE} (${type}) (${name ?? `unknown ${internalId}`})`,
```
omrilit, line 67: Why do we set the internalId in the id field? IIUC it's not used upon deploy but only exists there in the nacl.

packages/netsuite-adapter/src/filters/data_account_specific_values.ts
```diff
@@ -0,0 +1,393 @@
...
+})
+
+const getNameFromUnknownTypeReference = (
```
omrilit, line 70: Could you please explain the logic behind this method in a comment for future reader? I think it would be hard to understand it without context

omrilit (commented):
Great work!!

packages/netsuite-adapter/src/filters/data_instances_internal_id.ts
```diff
@@ -48,9 +48,12 @@ const isNestedPath = (path: ElemID | undefined): path is ElemID =>
  * (since the internal id is hidden, and we don't support hidden values in lists,
  * the objects in the list need to be extracted to new instances).
  */
-const filterCreator: LocalFilterCreator = () => ({
+const filterCreator: LocalFilterCreator = ({ config }) => ({
   name: 'dataInstancesInternalId',
   onFetch: async elements => {
+    if (config.fetch.resolveAccountSpecificValues) {
```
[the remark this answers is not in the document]
dantal4: The new filter & CV handles it all.
Additional noise will be the removal of `name` field (because it will be inside `'[ACCOUNT_SPECIFIC_VALUE] (type) (name)`, and removal of all subInstances in the environment.

packages/netsuite-adapter/src/filters/data_account_specific_values.ts
```diff
@@ -0,0 +1,393 @@
...
+  const existingInstance = await getUnknownTypeReferencesExistingInstance(elementsSource)
+  if (!isInstanceElement(existingInstance)) {
+    return {}
+  }
+  return _.mapValues(
+    existingInstance.value as Record<string, Record<string, string>>,
+    internalIdToName => _(internalIdToName)
+      .entries()
+      .groupBy(([_internalId, name]) => name)
+      .mapValues(row => row.map(([internalId]) => internalId))
```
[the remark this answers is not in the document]
dantal4: are you sure? there can be several internal ids for a single name, and in this case we choose the first one and return a warning

packages/netsuite-adapter/src/filters/data_account_specific_values.ts
dantal4: The `dataInstancesDiff` filter removes values that weren't change. if we don't run on both before & after then the resolved references fields will always be included in the deploy request, even if they weren't modified.

packages/netsuite-adapter/src/filters/data_account_specific_values.ts
dantal4: we can't use `internalId` because this field is hidden, and in deploy we replace `id` with `internalId`

packages/netsuite-adapter/src/filters/data_account_specific_values.ts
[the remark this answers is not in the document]
omrilit: Where do we return a warning?

packages/netsuite-adapter/src/filters/data_account_specific_values.ts
omrilit: I wonder whether in that case we should maybe set it only to ACCOUNT_SPECIFIC_VALUE and not support deploying it.
Otherwise we may corrupt the instance with deploying an incorrect value here.
WDYT?

packages/netsuite-adapter/src/filters/data_account_specific_values.ts
omrilit: You are right 😓

omrilit (approved):
💯

packages/netsuite-adapter/src/filters/data_account_specific_values.ts
dantal4: I now understand you first question.... it won't be used in deploy - we'll show an error of "could not identify object".
I thought that putting there "unknown" can have an edge case where there's a real object in the internal-id map named "unknown", but it's probably not a real case. Anyway I'll change it to "unknown object" and remove the internal id.

packages/netsuite-adapter/src/filters/data_account_specific_values.ts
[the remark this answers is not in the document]
dantal4: https://github.com/salto-io/salto/pull/5400/files#diff-f4346e0833717db74889cb3c22d5653befcc897ed3878f91784e7ccc3a95400bR318-R324

PR #5401, opened 2024-02-05
SALTO-5376: Fix bug in recreation of elemIds for trusted users group

shir-reifenberg:
Fix bug in recreation of elemIds for trusted users group

---

The bug was that in every fetch after the first fetch ran with `--regenerate-salto-ids`, we didn't rename the `trusted-users` group in Jira (because `elemIdGetter` wasn't provided so we returned original name).

This pr fixes this issue. In addition, instead of recreating all group instances, we'll only recreate the group we'd want to rename

---
_Release Notes_: 

_Jira_adapter_:
- Fix issue with the elemId of trusted-users group in Jira. If the elemId for the trusted-users group contains uuid, fetch with `regenerate salto ids` to resolve the issue.

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/group_name.ts
+++ b/packages/jira-adapter/src/filters/group_name.ts
@@ -16,10 +16,12 @@
 import { Element, ElemIdGetter, getChangeData, InstanceElement, isAdditionChange, isInstanceChange, isInstanceElement } from '@salto-io/adapter-api'
 import _ from 'lodash'
 import { elements as elementUtils, config as configUtils } from '@salto-io/adapter-components'
-import { naclCase, pathNaclCase } from '@salto-io/adapter-utils'
+import { logger } from '@salto-io/logging'
 import { FilterCreator } from '../filter'
-import { GROUP_TYPE_NAME, JIRA } from '../constants'
+import { GROUP_TYPE_NAME } from '../constants'
 import { JiraConfig } from '../config/config'
+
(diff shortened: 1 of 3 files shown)


packages/jira-adapter/src/filters/group_name.ts
```diff
@@ -31,54 +33,28 @@ const isGroupElement = (element: Element): boolean =>
...
-const getRenamedInstance = (
+const createRenamedTrustedGroupInstance = async (
   instance: InstanceElement,
   config: JiraConfig,
   getElemIdFunc?: ElemIdGetter,
-): InstanceElement => {
-  const elementName = getInstanceName(instance, config, getElemIdFunc)
-  const originalName = instance.value.name
+): Promise<InstanceElement> => {
   const newName = getGroupName(instance)
```
omrilit, line 47: I think that the function here is confusing as we always call it for a TrustedGroup instance, so can just use the const, WDYT?

omrilit (approved):
Thanks!!

packages/jira-adapter/src/filters/group_name.ts
shir-reifenberg: IMO because `createRenamedTrustedGroupInstance` can get any instance I think it's better to validate instead of using a const...

omrilit (approved):
LGTM!

PR #5402, opened 2024-02-05
SALTO-5164: [Zendesk] Add check for content in article attachment validator

DanielBarros1:
doing `getContent()` on anything that's not a static file throws, so add a check to elegantly leave if that is the case
Also added a test case

---
_Test Plan_:

Before:
Screenshot 2024-02-05 at 16 01 30
Screenshot 2024-02-05 at 16 01 53

After:
Screenshot 2024-02-05 at 16 02 07
Screenshot 2024-02-05 at 16 03 32

---
_Release Notes_: 
Zendesk: 
Add check in Article Attachment CV to prevent throws when content is not a static file

### DIFF
--- a/packages/zendesk-adapter/src/change_validators/article_attachment_size.ts
+++ b/packages/zendesk-adapter/src/change_validators/article_attachment_size.ts
@@ -13,7 +13,7 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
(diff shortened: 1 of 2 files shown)


edenhassid (approved):
looks good, added a small comment. pre-approving. notice that you should keep the `user notifications` part and just write none.

PR #5404, opened 2024-02-05
SALTO-2343 zendesk: move the omit inactive logic to the user facing part of the config

Uri-Bar:
SALTO-2343 zendesk: move the omit inactive logic to the user facing part of the config
---

I didnt remove any omitInactive logic as I assumed it would be part of the second PR

---
_Release Notes_: 
zendesk: add the ability to omit inactive from the user facing part of the config
---
_User Notifications_: 
none


packages/adapter-components/src/config/shared.ts
```diff
@@ -57,6 +57,7 @@ export type AdapterApiConfig<
 
 export type DefaultFetchCriteria = {
   name?: string
+  active?: boolean
```
netama, line 60: I'm not sure this has meaning in all adapters and it will likely require more customizations like we have here, probably best to start with a zendesk-only implementation? or do you have a plan to generalize?

packages/zendesk-adapter/src/fetch_criteria.ts
```diff
@@ -14,11 +14,29 @@
...
+    if (value === true) {
+      return instance.value.status !== 'inactive'
+    }
+    return instance.value.status === 'inactive'
+  }
+  // We can't omit inactive ticket_form instances because we need all the instance in order to reorder them
+  if (typeName === TICKET_FORM_TYPE_NAME || instance.value.active === undefined) {
+    return true
+  }
+  return instance.value.active === value
```
netama, line 32: I think this is slightly changing the behavior - in an arbitrary type, checking `active !== false` will also be truthy when the `active` field is not there at all - this will probably cause us now to omit instances that we fetch today. probably best to play with this on a local workspace - fetch before your code change (where we have `omitInactive` in the api definitions), then update the code and fetch again. the goal is for the workspace to still have exactly the same elements

netama (commented):
cool :) this is in the right direction, added some comments (can also discuss IRL). 
re keeping the old implementation - let's see if we want to do it in two PRs or in one (at least when testing locally, please remove it to make sure the behavior is as expected :)

packages/adapter-components/src/config/shared.ts
Uri-Bar: I do not have any plans here, where is the right place to add it to make it zendesk only implementation as it uses the shared ttype of `DefaultFetchCriteria` inside of `UserFetchConfig`

Uri-Bar:
> cool :) this is in the right direction, added some comments (can also discuss IRL). re keeping the old implementation - let's see if we want to do it in two PRs or in one (at least when testing locally, please remove it to make sure the behavior is as expected :)

when I tested it I obviously removed the filter. just for the PR I added it so we could discuss the release process

packages/zendesk-adapter/src/fetch_criteria.ts
Uri-Bar: I think I preserved the old behavior as I checked `instance.value.active === undefined`.  so if a type does not have active field it would return true and wont omit it no matter what

Uri-Bar:
we pivoted into different solution

PR #5405, opened 2024-02-05
SALTO-5351: minor bundle updates

Shulik95:
_A few bundles related updates_:
* bundles change validator updated to prevent all modifications on bundle elements and not only additions
* bundle elements renamed to have "bundle_" prefix and not only id

---

_None_

---
_Release Notes_: 
_Netsuite Adapter_:
* bundle elements will be renamed from `<bundleId>` to `bundle_<bundleId>`

---
_User Notifications_: 
_bundle elements will be renamed from `<bundleId>` to `bundle_<bundleId>`_

### DIFF
--- a/packages/netsuite-adapter/src/filters/bundle_ids.ts
+++ b/packages/netsuite-adapter/src/filters/bundle_ids.ts
@@ -81,7 +81,8 @@ const filterCreator: LocalFilterCreator = ({ config }) => ({
     const [bundleInstances, nonBundleElements] = _.partition(elements, isBundleInstance)
     const [existingBundles, missingBundles] = _.partition(bundleInstances, bundle =>
       bundle.value.id in BUNDLE_ID_TO_COMPONENTS)
-    log.debug('The following bundle ids are missing in the bundle record: %o', missingBundles.map(bundle => bundle.value.id))
+    log.debug('The following bundle ids are missing in the bundle record: %o',
+      missingBundles.map(bundle => [bundle.value.id, bundle.value.installedFrom.toUpperCase(), bundle.value.publisher]))
     existingBundles.forEach(bundle => {
       bundle.value.isPrivate = Object.keys(BUNDLE_ID_TO_COMPONENTS[bundle.value.id]).length === 0
     })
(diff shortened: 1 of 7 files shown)


packages/netsuite-adapter/src/filters/bundle_ids.ts
```diff
@@ -81,7 +81,8 @@ const filterCreator: LocalFilterCreator = ({ config }) => ({
     const [bundleInstances, nonBundleElements] = _.partition(elements, isBundleInstance)
     const [existingBundles, missingBundles] = _.partition(bundleInstances, bundle =>
       bundle.value.id in BUNDLE_ID_TO_COMPONENTS)
-    log.debug('The following bundle ids are missing in the bundle record: %o', missingBundles.map(bundle => bundle.value.id))
+    log.debug('The following bundle ids are missing in the bundle record: %o',
```
dantal4, line 84: nit - log only when the list is empty

packages/netsuite-adapter/src/filters/bundle_ids.ts
```diff
@@ -81,7 +81,8 @@ const filterCreator: LocalFilterCreator = ({ config }) => ({
...
+      missingBundles.map(bundle => [bundle.value.id, bundle.value.installedFrom.toUpperCase(), bundle.value.publisher]))
```
dantal4, line 85: nit- this will be easier to understand in the logs:
```suggestion
      missingBundles.map(bundle => { bundleId: bundle.value.id, installedFrom: bundle.value.installedFrom.toUpperCase(), publisher: bundle.value.publisher }))
```

packages/netsuite-adapter/src/transformer.ts
```diff
@@ -68,6 +68,8 @@ const getServiceIdFieldName = (type: ObjectType): string => {
   return isStandardType(type) ? SCRIPT_ID : PATH
 }
 
+const addBundlePrefix = (desiredName: string, type: ObjectType): string => (isBundleType(type) ? BUNDLE.concat(`_${desiredName}`) : desiredName)
```
dantal4, line 71: nit - use string formatting
```suggestion
const addBundlePrefix = (desiredName: string, type: ObjectType): string => (isBundleType(type) ? `${BUNDLE}_${desiredName}` : desiredName)
```

dantal4, line 71: Can you cover it with a test?

dantal4 (commented):
Great work!

packages/netsuite-adapter/src/filters/bundle_ids.ts
Shulik95: Just to verify. you mean when `missingBundles` isn't empty, right?

dantal4: yes, my mistake 👍

omrilit (approved):

PR #5407, opened 2024-02-05
SALTO-5091 Add multi nesting path support

adfineberg:
Enable nestStandalone for Zendesk and clean up nested instances created in Zendesk filters

---

---
_Release Notes_: 

---
_User Notifications_: 


packages/adapter-components/src/filters/referenced_instance_names.ts
```diff
@@ -101,6 +101,15 @@ const getInstanceNameDependencies = (
   return referencedInstances
 }
 
+const isStandalone = (instance: InstanceElement, configByType: Record<string, TransformationConfig>): boolean => {
```
adfineberg, line 104: Not sure if this should be in a different file

edenhassid (commented):
zendesk part looks good! @shir-reifenberg can you review the filter part?

packages/adapter-components/src/filters/referenced_instance_names.ts
```diff
@@ -113,7 +122,7 @@ const createInstanceNameAndFilePath = (
   const newName = joinInstanceNameParts(newNameParts) ?? instance.elemID.name
   const parentName = idConfig.extendsParentId ? getFirstParentElemId(instance)?.name : undefined
   const { typeName, adapter } = instance.elemID
-  const { fileNameFields, serviceIdField } = configByType[typeName]
+  const { fileNameFields, serviceIdField, nestStandaloneInstances } = configByType[typeName]
```
shir-reifenberg, line 125: maybe I'm missing something, but shouldn't the `nestStandaloneInstances` should be on the parent type ?

packages/adapter-components/src/filters/referenced_instance_names.ts
```diff
@@ -133,6 +142,8 @@ const createInstanceNameAndFilePath = (
     isSettingType: configByType[typeName].isSingleton ?? false,
     nameMapping: configByType[typeName].nameMapping,
     adapterName: adapter,
+    nestedPaths: nestStandaloneInstances && isStandalone(instance, configByType)
```
shir-reifenberg, line 145: Can we replace the condition with the same one we have in `toBasicInstance` ? ([here](https://github.com/salto-io/salto/blob/main/packages/adapter-components/src/elements/instance_elements.ts#L256))

and maybe create a shared util func

shir-reifenberg (commented):
Thanks! 🔥  added 2 questions
I also found the ticket I told you about (SALTO-5091), if you can please include in the title 🙏 
 And can you please validate fetching Jira with --regenerate-salto-ids doesn't cause any changes in paths / elemIds ? 🙏

packages/adapter-components/src/filters/referenced_instance_names.ts
adfineberg: NVM, changed it

packages/adapter-components/src/filters/referenced_instance_names.ts
```diff
@@ -132,6 +142,9 @@ const createInstanceNameAndFilePath = (
     isSettingType: configByType[typeName].isSingleton ?? false,
     nameMapping: configByType[typeName].nameMapping,
     adapterName: adapter,
+    nestedPaths: isStandalone(instance, configByType)
+      ? [...instance.path?.slice(2, instance.path?.length - 1) ?? []] : undefined,
+    hasNestStandAloneFields: nestStandaloneInstances && standaloneFields !== undefined,
```
idoamit8, line 147: @adfineberg @shir-reifenberg 
Does it support nested paths for more than one level of nested nacls? If we have:
Type 1
   |__ Type 2
             |__ Type 3
and Type 3 depends on Type 2, will the path change as expected? I believe that since we slice at index 2, it will become the child of Type 1.

adfineberg: Fixed the multi-nesting, take a look :)
Regarding the long folder names - I'm not sure it's a problem, as I set the nested folder names as the subtype. LMK if that makes sense

packages/adapter-components/src/filters/referenced_instance_names.ts
```diff
@@ -100,6 +101,30 @@ const getInstanceNameDependencies = (
...
+      || field.fieldName === instance.elemID.typeName)) ?? false
+}
+
+const nestedPath = (
+  instance: InstanceElement, configByType: Record<string, TransformationConfig>
+): string[] | undefined => {
+  if (!isStandalone(instance, configByType)) {
+    return undefined
+  }
+  const parent = getParent(instance)
```
idoamit8, line 122: getParent can throw if there is more than one parent or no parent at all. 
I think its worth rapping this line with try-catch or adding check for valid parent with `hasValidParent`.

idoamit8 (approved):
Looks great! 💯 
Thanks for modifying this utility function.
Added one comment that needs to resolve, and approving  to save cycle.

PR #5408, opened 2024-02-05
SALTO-5375 - Salesforce: Fix mapping of deploy errors of custom object and custom field

tomsellek:
Deploy errors where the `componentType` is `CustomObject` or `CustomField` were not mapped to the correct ElemID (e.g. they were mapped to `salesforce.CustomObject.instance.Account`.

---
At the point where we collect the names/elemIDs, custom objects are already converted back to instances. So our mapping is technically correct (which is the best kind of correct, of course), but not helpful.

---
_Release Notes_: 
Salesforce: Improved mapping of deploy errors to elements

---
_User Notifications_: 
N/A


tamtamirr (approved):
LGTM!

PR #5411, opened 2024-02-06
SALTO-5388 move existing logic to new locations

netama:
see https://github.com/salto-io/salto/pull/5409/files for context

starting by moving a few parts of the old implementation that will be reused in the new implementation to their new location + updating imports, as well as extending some existing types and functions

---

will add a few more missing tests tomorrow

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/adapter-components/src/fetch/resource/request_parameters.ts
```diff
@@ -41,6 +42,7 @@ export const simpleGetArgs: ComputeGetArgsFunc = (
   {
     url,
     queryParams,
+    // TODO deprecate once old config is gone, and replace with recurseInto
     recursiveQueryByResponseField,
```
shir-reifenberg, line 46: should we mark the field on `FetchRequestConfig` as deprecated ?

packages/adapter-components/src/fetch/request/utils.ts
```diff
@@ -0,0 +1,59 @@
...
+* limitations under the License.
+*/
+
+import _ from 'lodash'
+import { Values, isPrimitiveValue } from '@salto-io/adapter-api'
+import { ContextParams } from '../../definitions'
+
+export const ARG_PLACEHOLDER_MATCHER = /\{([\w_]+)\}/g
+
+export const findUnresolvedArgs = (value: string, definedParams: Set<string> = new Set()): string[] => {
```
shir-reifenberg, line 23: why all `url` replaced with `value`? can it be something else?

packages/adapter-components/src/fetch/request/utils.ts
```diff
@@ -0,0 +1,59 @@
...
+  const urlParams = value.match(ARG_PLACEHOLDER_MATCHER)?.map(m => m.slice(1, -1)) ?? []
+  return urlParams.filter(p => !definedParams.has(p))
+}
+
+export const findAllUnresolvedArgs = (value: unknown, definedParams: Set<string> = new Set()): string[] => {
```
shir-reifenberg, line 28: not sure why `value` is unknown, it looks like only works on string (in the inner call or when calling it on `requestDef.endpoint.path`
Can we talk IRL?

shir-reifenberg (approved):
Looks great 👑 🔥 🎉  added mainly questions

shir-reifenberg:
Forgot to add here as well cause we talked about it in the second PR - the client config is missing from user config

packages/adapter-components/src/fetch/resource/request_parameters.ts
netama: not yet because there's no alternative in the old infra and I haven't fully tested it yet 😅 but if this works as expected then I'll remove it once I upgrade workato

packages/adapter-components/src/fetch/request/utils.ts
netama: yes it can also be in query args / body etc

packages/adapter-components/src/fetch/resource/request_parameters.ts
```diff
@@ -123,7 +138,7 @@ const computeDependsOnURLs = (
     combo => potentialParams.map(param => ({ ...combo, ...param }))
   ))
 
-  return allArgCombinations.map(p => replaceUrlParams(url, p))
+  return allArgCombinations.map(p => replaceArgs(url, p))
 }
 
 export const createUrl = ({
```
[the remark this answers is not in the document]
netama: discussed IRL - will probably move this when adding the deploy

PR #5413, opened 2024-02-07
SALTO-5389 utils and DefaultWithCustomization, ArgsWithCustomizer, OptionsWithDefault

netama:
Add a few utility types and helper functions that will be used in the new infra.
See https://github.com/salto-io/salto/pull/5409/files for more context.
Merging this now because we want to start using `DefaultWithCustomization` and its utilities soon.

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/definitions/system/shared/types.ts
+++ b/packages/adapter-components/src/definitions/system/shared/types.ts
@@ -29,8 +29,8 @@ export type DefaultWithCustomizations<T, K extends string = string> = {
   customizations: string extends K ? Record<K, T> : Partial<Record<K, T>>
 }
 
-export type ArgsWithCustomizer<ResponseType, Args, Input = unknown, AdditionalArgs = {}> = Args & {
-  custom?: ((args: Partial<Args> & AdditionalArgs) => (input: Input) => ResponseType)
+export type ArgsWithCustomizer<ResultType, Args, Input = unknown, AdditionalArgs = {}> = Args & {
+  custom?: ((args: Partial<Args> & AdditionalArgs) => (input: Input) => ResultType)
 }
 
 export type OptionsWithDefault<T, K extends string> = {
--- a/packages/adapter-components/src/definitions/system/shared/types.ts
+++ b/packages/adapter-components/src/definitions/system/shared/types.ts
@@ -1,0 +1,39 @@
+/*
+*                      Copyright 2024 Salto Labs Ltd.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


packages/adapter-components/src/definitions/system/utils.ts
```diff
@@ -0,0 +1,115 @@
...
+    // TODO improve casting
+    return def.map(item => mergeSingleDefWithDefault(defaultDef, item)) as unknown as T
+  }
+  // we add a fake nesting level called "value" in order for ignoreDefaultFieldCustomizations to be caught even
+  // if it directly under the root.
+  return _.mergeWith(
+    { value: _.cloneDeep(defaultDef) },
+    { value: def },
+    (first, second) => {
+      if (lowerdashValues.isPlainObject(second) && _.get(second, 'ignoreDefaultFieldCustomizations') !== undefined) {
```
shir-reifenberg, line 48: is `ignoreDefaultFieldCustomizations` only relevant to `ElementFieldCustomization` ?

netama: yes, I think I will later generalize this to `ignoreX` so that it's more generic, it's a bit hacky this way but I don't think it's risky. I can change it now and make it a pattern, wdyt?

packages/adapter-components/src/definitions/system/shared/types.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+export type DefaultWithCustomizations<T, K extends string = string> = {
+  // if the customization is an array, the default will be applied to each array item
+  // (if the customization is empty, it will not be used)
+  default?: types.RecursivePartial<T extends (infer U)[] ? U : T>
+  // hack to avoid requiring all keys of an enum
+  customizations: string extends K ? Record<K, T> : Partial<Record<K, T>>
+}
+
+export type ArgsWithCustomizer<ResponseType, Args, Input = unknown, AdditionalArgs = {}> = Args & {
```
shir-reifenberg, line 32: nit: maybe `ResultType` instead of `ResponseType` ? 
for me `ResponseType` sounds more specific (something like client repsonse)

edenhassid (approved):
cool! 😎 added some small comments

packages/adapter-components/src/definitions/system/utils.ts
shir-reifenberg: mm I think it's ok not generalize now if we don't need it... but I wonder if this logic should be somehow separated cause it's kind of specific to `fieldCustomizations` 
or maybe just a comment explaining it' only relevant to `fieldCustomizations`
anyway whatever you think here :)

packages/adapter-components/src/definitions/system/shared/types.ts
```diff
@@ -0,0 +1,39 @@
...
+  custom?: ((args: Partial<Args> & AdditionalArgs) => (input: Input) => ResponseType)
```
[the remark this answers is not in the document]
netama: it was originally called `customizer` but I ended up thinking `custom` is cleaner, will take another look though :)

packages/adapter-components/src/definitions/system/utils.ts
netama: adding a comment for now, I think we might generalize this soon

PR #5415, opened 2024-02-07
SALTO-5255: [Zendesk] Fix bug when deploying an article with a change to User Segment to "Everyone"

DanielBarros1:
TSIA

---

Since the "Everyone" instance is a Salto concept and Zendesk doesn't know it, changing the User Segment to "Everyone" would fail on deploy.

---
_Release Notes_: 
Zendesk:

Changing `User Segment` in `Article` to `Everyone` now deploys correctly

---
_User Notifications_: 

None


edenhassid (approved):
LGTM! 🔥 💯

PR #5418, opened 2024-02-07
SALTO-5362: support scriptRunner references with the new workflow

shayc331:
_support scriptRunner references with the new workflow_

---

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


packages/jira-adapter/src/adapter.ts
```diff
@@ -206,7 +206,6 @@ export const DEFAULT_FILTERS = [
   fieldNameFilter,
   workflowStructureFilter,
   workflowFilter,
-  // must run before references are transformed and after workflowFilter
```
shayc331, line 209: note: 
I wrote this comment, it is a leftover

packages/jira-adapter/test/filters/script_runner/workflow/workflow_references.test.ts
```diff
@@ -64,22 +88,87 @@ const restoredInstance = new InstanceElement(
...
+    fieldName: string
+  }): Value => {
+    if (workflowVersion === WORKFLOW_V1) {
+      return instance.value.transitions[transitionKey].rules.postFunctions[postFunctionIndex].configuration[fieldName]
+    }
+    if (workflowVersion === WORKFLOW_V2) {
+      return workflowV2Instance.value.transitions[transitionKey].actions[postFunctionIndex].parameters[fieldName]
+    }
+    throw new Error('Unknown workflow version')
+  }
```
IdoZakk, line 161: This function both chooses an object (that is not in the parameters) and the path. Maybe split it into two functions?

IdoZakk (commented):
Nice work! 🐜 
Please increase back the coverage of workflow_references

packages/jira-adapter/test/filters/script_runner/workflow/workflow_references.test.ts
shayc331: Im not sure I understand, the object and the path are coupled together here

IdoZakk: Yes, but you have a function that both holds internal objects and finds the path. I think it might be confusing for a future person going into the code that will not understand what is being checked or if s/he will be using a different instance. 
I guess you can keep it, just feels a bit hacky this way to me

IdoZakk (approved):
🐺 Amazing coverage!!!

PR #5419, opened 2024-02-07
SALTO-5388 allow omitting the user fetch config elem id field

netama:
the new elem id field was added in https://github.com/salto-io/salto/pull/5411 as optional - I think it's safer to allow the adapter to not support it until we allow users to control the flow via that (forgot to add this in the previous PR).
also using this opportunity to switch the function to named parameters.

---
_Release Notes_: 
None

---
_User Notifications_: 
None


shir-reifenberg (approved):
🚀

PR #5420, opened 2024-02-07
SALTO-5399: Remove listUnresolved from core

shir-reifenberg:
implementation moved to workspace 
also `compact` is duplicated there

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


adamro (approved):
😱

PR #5421, opened 2024-02-07
SALTO-4692 - Salesforce: Move last-resort deploy validations to a change validator

tomsellek:
When unexpected things happen, like elements having no API name or changes not being instance changes at deploy time, we shouldn't create deploy errors - we should create validation errors that are more helpful.

---

I still removed the invalid changes from the changes that are deployed, does this make sense?

---
_Release Notes_: 
Salesforce: some deploy errors are now change validator errors

---
_User Notifications_: 
N/A


packages/salesforce-adapter/src/change_validators/metadata_deploy_validation.ts
```diff
@@ -0,0 +1,71 @@
...
+      log.info('API name changed for %s', changeElem.elemID.getFullName())
+      return {
+        elemID: changeElem.elemID,
+        message: 'Failed to update element because the API name before the change is different from the API name after it',
+        severity: 'Error',
+        detailedMessage: `Failed to update element because api names prev=${beforeName} and new=${afterName} are different`,
+      }
+    }
+  }
+  if (!isInstanceChange(change) || !await isMetadataInstanceElement(changeElem)) {
```
tomsellek, line 52: @tamtamirr this hits a bunch of field changes in the E2E tests. Do these field changes not make it to `deployMetadata`, or are we cool with dropping these changes?

PR #5424, opened 2024-02-08
SALTO-3080 Log handlebar references

adfineberg:
Extract references from handlebar templates. First stage, only log the results.

---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/zendesk-adapter/src/filters/guide_theme.ts
+++ b/packages/zendesk-adapter/src/filters/guide_theme.ts
@@ -30,7 +30,7 @@ import {
   isObjectType, Field, MapType, CORE_ANNOTATIONS,
 } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
-import { values, values as lowerdashValues } from '@salto-io/lowerdash'
+import { values, values as lowerdashValues, collections } from '@salto-io/lowerdash'
 import JSZip from 'jszip'
 import _, { remove } from 'lodash'
 import { getInstancesFromElementSource, naclCase, inspectValue } from '@salto-io/adapter-utils'
@@ -45,10 +45,12 @@ import { deleteTheme } from './guide_themes/delete'
 import { download } from './guide_themes/download'
 import { publish } from './guide_themes/publish'
 import { getBrandsForGuideThemes } from './utils'
+import { parseHandlebarReferences } from './template_engines/handlebar_parser'
 
 
 const log = logger(module)
 const { isPlainRecord } = lowerdashValues
+const { awu } = collections.asynciterable
 
 type ThemeFile = { filename: string; content: StaticFile }
 type DeployThemeFile = { filename: string; content: Buffer }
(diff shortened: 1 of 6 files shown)


packages/zendesk-adapter/src/filters/guide_theme.ts
```diff
@@ -246,6 +267,12 @@ const filterCreator: FilterCreator = ({ config, client, elementsSource }) => ({
       }
       return brandName
     }
+    const idsToElements = await awu(await elementsSource.getAll())
+      .filter(isInstanceElement)
+      .reduce<Record<string, Element>>((acc, elem) => {
+        acc[elem.value.id] = elem
```
edenhassid, line 273: what if id is undefined? i think you should fist filter all elements that have an id

packages/zendesk-adapter/src/filters/template_engines/handlebar_parser.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+  return undefined // When node is a Literal
+}
+
+/**
+ * Extracts all the number literals from a handlebar template that are used as arguments in relevant helper functions
+ * @example parseHandlebarReferences('Hello {{#is id 12345}}good name{{else}}bad name{{/is}}')
+ * // Returns [{ value: 12345, loc: { start: { line: 1, column: 18 }, end: { line: 1, column: 23 } } }]
+ */
+export const parseHandlebarReferences = (template: string): NumberLiteral[] => {
```
edenhassid, line 42: 1. why do you call the string `template`? i think it may be confusing with the functions that use template for template expression
2. maybe `parseHandlebarPotentialReferences` cause we don't turn it to reference yet

packages/zendesk-adapter/src/filters/template_engines/handlebar_parser.ts
```diff
@@ -0,0 +1,53 @@
...
+  const ast = parse(template)
+  return ast.body.map(node => {
+    if (isMustacheStatement(node) || isBlockStatement(node)) {
+      const helper = extractHelper(node.path)
+      if (helper && RELEVANT_HELPERS.includes(helper)) {
+        return node.params.filter(isNumberLiteral)
```
edenhassid, line 48: we return all the numbers? they can't be just a number and not an id?

packages/zendesk-adapter/test/filters/template_engines/handlebar_parser.test.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  it('should parse a template with a relevant helper function and a sub expression', () => {
+    const template = 'Hello {{#is (lookup id) 12345}}good name{{else}}bad name{{/is}}'
+    const parsed = parseHandlebarReferences(template)
+    expect(parsed).toHaveLength(1)
+    expect(parsed[0].value).toEqual(12345)
+  })
+
+  it('should not parse a template with an irrelevant helper function', () => {
+    const template = 'Hello {{#if id 123456}}good name{{else}}bad name{{/if}}{{notBlock 1231}}'
```
edenhassid, line 34: why we don't get this id from the if?

edenhassid (approved):
looks good! 🔥 😎 added a few questions and one comment. pre-approving :)

packages/zendesk-adapter/test/filters/template_engines/handlebar_parser.test.ts
adfineberg: I set only the relevant Zendesk handlebar helpers that have ids. I'll add a comment

PR #5425, opened 2024-02-08
SALTO-2392 add skipResolvingAccountSpecificValuesToTypes to NS config

dantal4:
Add `fetch.skipResolvingAccountSpecificValuesToTypes` config to skip some SuiteQL table queries, in case that they take too much time. In case that a SuiteQL table query is skipped, it's instance will be deleted.

Example:
```
netsuite {
  fetch = {
    skipResolvingAccountSpecificValuesToTypes = [
      "item",
      // regex supported
      ".*Category"
    ]
  }
}
```

---

---
_Release Notes_: 
Netsuite Adapter:
- Add `fetch.skipResolvingAccountSpecificValuesToTypes` config to skip some SuiteQL table queries, in case that they take too much time.

---
_User Notifications_: 
None


omrilit (approved):
Great work!
We should probably document this option in the README right?

dantal4:
> Great work! We should probably document this option in the README right?

I'm not sure. I think that we'd like it to be undocumented, but we'll be able to ask CE to guide customers to add it.

omrilit:
> > Great work! We should probably document this option in the README right?
> 
> I'm not sure. I think that we'd like it to be undocumented, but we'll be able to ask CE to guide customers to add it.

We can start without documenting it and consider doing it in the future

PR #5427, opened 2024-02-08
new Dependency Flow - changing removal depencdency 

oneYAD:
bugfix - add new dependency flow.
From now on - if A has reference to B and A modified and B removed, A would be dependency of B. so A would change before B removed.

---

see https://salto-io.atlassian.net/browse/SALTO-5402

---
_Release Notes_: 
Zendesk adapter - fix bug of getting Service error while removing trigger_category and changing trigger to be depend on other category simultaneously 

---
_User Notifications_: 

None

### DIFF
--- a/packages/core/src/core/plan/dependency/reference_dependency.ts
+++ b/packages/core/src/core/plan/dependency/reference_dependency.ts
@@ -18,9 +18,10 @@ import { collections, values } from '@salto-io/lowerdash'
 import {
   getChangeData, isReferenceExpression, ChangeDataType, Change, ChangeEntry, DependencyChange,
   addReferenceDependency, addParentDependency, isDependentAction, DependencyChanger, isObjectType,
-  ElemID, isModificationChange, isField, isEqualValues, dependencyChange,
+  ElemID, isModificationChange, isField, isEqualValues, dependencyChange, isRemovalChange,
 } from '@salto-io/adapter-api'
 import { getAllReferencedIds, getParents, resolvePath } from '@salto-io/adapter-utils'
+import { SetId } from '@salto-io/lowerdash/src/collections/set'
 
 const { awu } = collections.asynciterable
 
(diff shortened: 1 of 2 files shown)


packages/core/src/core/plan/dependency/reference_dependency.ts
```diff
@@ -46,37 +47,67 @@ const isReferenceValueChanged = (change: Change<ChangeDataType>, refElemId: Elem
...
-          return undefined
-        }
-        const [targetChangeEntry] = changesById.get(targetElemId) ?? []
+        const targetChangeEntry = getTarget(targetRefId, elemId)
         if (targetChangeEntry === undefined) {
           return undefined
         }
         const [targetId, targetChange] = targetChangeEntry
         if (isDependentAction(change.action, targetChange.action)) {
+          const parents = getParentIds(elem)
```
dantal4, line 110: we should calculate this only once, like it was before

packages/core/src/core/plan/dependency/reference_dependency.ts
```diff
@@ -86,7 +117,7 @@ export const addReferencesDependency: DependencyChanger = async changes => {
         }
         return undefined
       })
-      .filter(values.isDefined)
+      .filter(values.isDefined), ...addBeforeModificationDependency([id, change])]
```
dantal4, line 120: can you add the new call outside of this function, as it's not dependent.
you can instead do this at the outside `return`:
```typescript
return awu(changes)
    .flatMap(entry => [
      ...addChangeDependency(entry),
      ...addBeforeModificationDependency(entry),
    ])
    .toArray()
```

packages/core/src/core/plan/dependency/reference_dependency.ts
```diff
@@ -46,37 +47,67 @@ const isReferenceValueChanged = (change: Change<ChangeDataType>, refElemId: Elem
...
     const elemId = elem.elemID.getFullName()
+    // Not using ElementsSource here is legit because it's ran after resolve
+
     // Because fields are separate nodes in the graph, for object types we should only consider
     // references from the annotations
-    const onlyAnnotations = isObjectType(elem)
-    // Not using ElementsSource here is legit because it's ran
-    // after resolve
-    return wu(getAllReferencedIds(elem, onlyAnnotations))
+    return [...wu(getAllReferencedIds(elem, isObjectType(elem)))
```
dantal4, line 101: I think that a function of `onlyAnnotations = isObjectType` will increase the readability here and in line 78 (maybe with the related comment above it)

dantal4 (commented):
Looks great!
I added some comments, LMK what you think. please add a test too 🙏

packages/core/src/core/plan/dependency/reference_dependency.ts
```diff
@@ -46,37 +47,67 @@ const isReferenceValueChanged = (change: Change<ChangeDataType>, refElemId: Elem
...
   const changesById = collections.iterable.groupBy(
     changes,
     ([_id, change]) => getChangeElemId(change),
   )
 
+
+  const getTarget = (targetRefId: ElemID, elemId: string) : undefined | [SetId, Change] => {
+    const targetElemId = targetRefId.createBaseID().parent.getFullName()
+    // Ignore self references
+    if (targetElemId === elemId) {
```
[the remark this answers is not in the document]
oneYAD: it's not really an ElemID, is just its name, so the equation is between strings. I will change the names of the consts for better readability

packages/core/src/core/plan/dependency/reference_dependency.ts
oneYAD: my bad. done

packages/core/src/core/plan/dependency/reference_dependency.ts
oneYAD: I wasn't sure about it, I  wonder if add all this to function is the right thing. or only the code in the map (so the function would be per reference. but then, I think it would be function with a lot of arguments, so I'm not sure..

I can do all this more complicated (but maybe more accurate) if I merge the addBeforeModificationDependency into this loop, and transfer to a new "reference_handler" function which will get the element to get references from and the function to handle the refernece.

what do you think?

packages/core/src/core/plan/dependency/reference_dependency.ts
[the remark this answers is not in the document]
dantal4: Yes but why should it be a string and not ElemID? it was an ElemID before: `const elemId = before.elemID.getFullName()`, and I think that it's nicer to use the structured object, but that's a nit.

packages/core/src/core/plan/dependency/reference_dependency.ts
dantal4: I'm not following. maybe my comment wasn't clear - I talked about passing `isObjectType(elem)` to `getAllReferencedIds`, and that it's hard to understand why it's done (there's a comment for that but now it's called in two places).

oneYAD: mmmm, yes, I defiantly thought you mean something else. 
but I still not sure what is your suggestion.
do you mean to change getAllReferencedIds? or adding new function wrapped getAllReferencedIds only for  isObjectType would be a default argument?

dantal4: define
```typescript
// Because fields are separate nodes in the graph, for object types we should only consider references from the annotations
const onlyAnnotations = isObjectType
```

then use it with `getAllReferencedIds(elem, onlyAnnotation(elem))`

oneYAD: hhhhh
OK, tnx

packages/core/src/core/plan/dependency/reference_dependency.ts
```diff
@@ -52,32 +52,62 @@ export const addReferencesDependency: DependencyChanger = async changes => {
     ([_id, change]) => getChangeElemId(change),
   )
 
+  const getTarget = (targetRefId: ElemID, changeElemIdName: string) : undefined | [collections.set.SetId, Change] => {
```
dantal4, line 55: Is the return type here `ChangeEntry`?

dantal4 (approved):
:100:

packages/core/src/core/plan/dependency/reference_dependency.ts
oneYAD: sorry, dont see it.
fix it here - https://github.com/salto-io/salto/pull/5445

PR #5429, opened 2024-02-09
SALTO-5382 - Don't skip deploy groups if a dependency had a recoverable error

tomsellek:
No reason to skip the dependencies of a change group with e.g. 'Info' level warnings.

---

---
_Release Notes_: 
Dependencies of change groups that have low-severity errors will now be deployed.

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/core/src/core/deploy/deploy_actions.ts
+++ b/packages/core/src/core/deploy/deploy_actions.ts
@@ -170,17 +170,10 @@ export const deployActions = async (
             item.groupKey,
             nonFatalErrors.map(err => err.message).join('\n\n'),
           )
-          accumulatedNonFatalErrors.push(
-            ...nonFatalErrors
-              .map(err => ({ ...err, groupId: item.groupKey }))
-          )
+          accumulatedNonFatalErrors.push(...nonFatalErrors.map(err => ({ ...err, groupId: item.groupKey })))
         }
         if (fatalErrors.length > 0) {
-          log.warn(
-            'Failed to deploy %s, errors: %s',
-            item.groupKey,
-            fatalErrors.map(err => err.message).join('\n\n'),
-          )
+          log.warn('Failed to deploy %s, errors: %s', item.groupKey, fatalErrors.map(err => err.message).join('\n\n'))
           throw new WalkDeployError(fatalErrors)
         }
         reportProgress(item, 'finished')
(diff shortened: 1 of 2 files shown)


packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -154,13 +155,26 @@ export const deployActions = async (
...
+              .map(err => ({ ...err, groupId: item.groupKey }))
+          )
+        }
+        if (fatalErrors.length > 0) {
+          log.warn(
+            'Failed to deploy %s, fatal errors: %s, non-fatal errors: %s',
+            item.groupKey,
+            fatalErrors.map(err => err.message).join('\n\n'),
+            nonFatalErrors.map(err => err.message).join('\n\n'),
+          )
```
tamtamirr, line 176: The non-fatal errors were logged in L160. No reason to log them again.
I'd use the original error `Failed to deploy GROUP_ID, errors:` here.

tamtamirr (commented):
1. We should test this out if we haven't (Either reproduce the issue in Salesforce of field removals or patch the adapter deploy to contain warning/info error)
2. We should cover this behavior in UTs

packages/core/src/core/deploy/deploy_actions.ts
```diff
@@ -199,6 +213,6 @@ export const deployActions = async (
         })
       }
     }
-    return { errors: deployErrors, appliedChanges, extraProperties: { groups } }
+    return { errors: deployErrors.concat(accumulatedNonFatalErrors), appliedChanges, extraProperties: { groups } }
```
adamro, line 216: Aren't we not returning warnings now when we return outside of the catch (if there are only warnings)?

(I didn't do the leg work of following the code carefully cause I think unit-testing this will be the easiest way to show the behaviour, that was just my intuition)

adamro (commented):
Can you please add unit-tests that check the cases this change did? ie. a group that won't be skipped cause it depends on a group that had only warnings and before this change it would be skipped, also tests that make sure the returned errors work as we expect in all cases (only error severity, error and warnings severities, only warnings etc.)

tomsellek:
> 1. We should test this out if we haven't (Either reproduce the issue in Salesforce of field removals or patch the adapter deploy to contain warning/info error)

I tested by generting an explicit `Info` level error and confirming the dependent groups deployed

> 2. We should cover this behavior in UTs

Will do.

packages/core/src/core/deploy/deploy_actions.ts
tomsellek: Ack

adamro (approved):
🔥

PR #5432, opened 2024-02-11
SALTO-5411: make get content of article_attachment use relative_path field

edenhassid:
make get content of article_attachment use relative_path field

---

---
_Release Notes_: 
zendesk:
* fix fetch of attachments with non english characters

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/filters/article/utils.ts
+++ b/packages/zendesk-adapter/src/filters/article/utils.ts
@@ -130,11 +130,16 @@ const getAttachmentContent = async ({
     log.error(error)
     return contentWarning(error)
   }
+  if (attachment.value.relative_path === undefined) {
+    const error = `could not add attachment ${attachment.elemID.getFullName()}, as the relative_path is undefined`
+    log.error(error)
+    return contentWarning(error)
+  }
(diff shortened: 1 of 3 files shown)


packages/zendesk-adapter/src/filters/article/utils.ts
```diff
@@ -130,11 +130,16 @@ const getAttachmentContent = async ({
     log.error(error)
     return contentWarning(error)
   }
+  if (attachment.value.relative_path === undefined) {
+    const error = `could not add attachment ${attachment.elemID.getFullName()}, as the relative_path is undefined`
+    log.error(error)
```
adfineberg, line 135: Do we want to: (a) have this as an `error` level log, and (b) not try to use the previous heuristic as a backup?

edenhassid: I think we shouldn't use the previous heuristic as it may be incorrect and lead to errors. this is the right way to do it according to zendesk api.

edenhassid: turned to a `warn` level log

adfineberg (approved):

PR #5434, opened 2024-02-11
SALTO-5401 Improve handling publish workflowScheme draft

kami-delaroz:
Increase number of retries in checking workflowScheme draft change, and instead of sending a failure error, state that we run out of retries but don't know wether this was successful

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
None


shayc331 (approved):
🏆

PR #5435, opened 2024-02-11
SALTO-5360 align inactive fields

dantal4:
Align isinactive/inactive/isInactive fields to be all "isInactive", in top level & nested values.
In addition, changed the order of some filters and added removal of refTypes to excluded custom record types.

---

TODO:
- [x] add tests

---
_Release Notes_: 
Netsuite Adapter:
- align inactive fields

---
_User Notifications_: 
Netsuite Adapter:
- all isinactive/inactive fields will be replaced with "isInactive" field - in types, instances, and important values

### DIFF
--- a/packages/netsuite-adapter/src/config/suggestions.ts
+++ b/packages/netsuite-adapter/src/config/suggestions.ts
@@ -18,6 +18,7 @@ import { ElemID, InstanceElement } from '@salto-io/adapter-api'
 import { formatConfigSuggestionsReasons } from '@salto-io/adapter-utils'
 import { values } from '@salto-io/lowerdash'
 import { FailedFiles } from '../client/types'
+import { INACTIVE_FIELDS } from '../constants'
 import { FetchTypeQueryParams, LockedElementsConfig, NetsuiteConfig, NetsuiteFilePathsQueryParams, NetsuiteTypesQueryParams, QueryParams, configType } from './types'
 import { FetchByQueryFailures, convertToQueryParams, isCriteriaQuery } from './query'
 import { emptyQueryParams, fullFetchConfig } from './config_creator'
@@ -30,6 +31,8 @@ export const toLargeFoldersExcludedMessage = (updatedLargeFolders: NetsuiteFileP
 
 export const toLargeTypesExcludedMessage = (updatedLargeTypes: string[]): string => `The following types were excluded from the fetch as the elements of that type were too numerous: ${updatedLargeTypes.join(', ')}.`
  + ' To include them, increase the types elements\' size limitations and remove their exclusion rules.'
+
+export const ALIGNED_INACTIVE_CRITERIAS = 'The exclusion criteria of inactive elements was modified.'
 
 const createFolderExclude = (folderPaths: NetsuiteFilePathsQueryParams): string[] =>
   folderPaths.map(folder => `^${_.escapeRegExp(folder)}.*`)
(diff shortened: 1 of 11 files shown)


omrilit (commented):
The code looks great!
I wonder whether this PR should contain more content:
1. modify default config to use isInactive only
2. migration of the existing configs to use isInactive only
3. do we need to modify important attributes?

Anyway, this shouldn't be merged till the end of the week, so not approving yet

dantal4:
> The code looks great! I wonder whether this PR should contain more content:
> 
> 1. modify default config to use isInactive only
> 2. migration of the existing configs to use isInactive only
> 3. do we need to modify important attributes?
> 
> Anyway, this shouldn't be merged till the end of the week, so not approving yet

Good catch on the config! I'll change the default config and add a config suggestion to existing envs. With the important values there's no problem, because we set the for exisitng fields, that are now all `isInactive`.

omrilit:
> Good catch on the config! I'll change the default config and add a config suggestion to existing envs. With the important values there's no problem, because we set the for exisitng fields, that are now all `isInactive`.

But we should probably remove the isinactive & inactive important attributes, no?

dantal4:
> > Good catch on the config! I'll change the default config and add a config suggestion to existing envs. With the important values there's no problem, because we set the for exisitng fields, that are now all `isInactive`.
> 
> But we should probably remove the isinactive & inactive important attributes, no?

they are removed automatically

packages/netsuite-adapter/src/config/suggestions.ts
```diff
@@ -198,6 +201,22 @@ const updateConfigFromLargeTypes = (
...
 
+// TODO: remove at May 24.
+const alignInactiveExclusionCriterias = (config: NetsuiteConfig): boolean => {
+  let isUpdated = false
+  config.fetch.exclude.types.filter(isCriteriaQuery).forEach(item => {
+    Object.values(INACTIVE_FIELDS).forEach(fieldName => {
+      if (item.criteria[fieldName] !== undefined && fieldName !== INACTIVE_FIELDS.isInactive) {
+        item.criteria[INACTIVE_FIELDS.isInactive] = item.criteria[fieldName]
+        delete item.criteria[fieldName]
+        isUpdated = true
```
omrilit, line 212: I am afraid that this migration is inaccurate.
Think of the following existing config:
```
isinactive = ".*SuffixA"
inactive = ".*SuffixB"
isInactive = ".*SuffixC"
```

After the above migration it will be translated into:
```
isInactive = ".*SuffixA"
isInactive = ".*SuffixB"
isInactive = ".*SuffixC"
```

This doesn't have the same meaning as before.
I assume that it's not a real concerning issue as probably users didn't modify it that way, but I think that we should gather some info on the existing configs to verify that we are not causing issues.

omrilit (commented):
Looks great, had only 1 concern that should be checked.

packages/netsuite-adapter/src/config/suggestions.ts
dantal4: The criteria is not logged well ([logs](https://app.datadoghq.com/logs?query=updated%20netsuite%20config%20status%3Adebug%20criteria%20&cols=host%2Cservice&event=AgAAAY2epBRtOoKaVgAAAAAAAAAYAAAAAEFZMmVwQm01QUFDVi1XV0preHlTamdEbAAAACQAAAAAMDE4ZDllYTQtMWVlMi00Zjc1LWJmZTctYWNlNjU2MjY1ZWMw&index=%2A&messageDisplay=inline&refresh_mode=sliding&stream_sort=desc&viz=stream&from_ts=1706469451080&to_ts=1707765451080&live=true)) - but all of them are in one of the following formats:
1. `{ name: '.*', criteria: [Object] },`
2. `{ name: 'savedsearch', criteria: [Object] },`

so I think that we're ok with the migration 👍

omrilit: Perfect

omrilit (approved):

PR #5438, opened 2024-02-12
SALTO-5249: [Zendesk] Throw during `Workspace` deployment when the Zendesk client returns 200 with errors

DanielBarros1:
when the client fails to apply changes, it sometimes returns a status code 200 with errors. 
This actually means the deployment failed, and so we should throw on our side of the deployment.

---

Screenshot 2024-02-12 at 13 36 25

---
_Release Notes_: 
Zendesk:
Fail deployment when Zendesk returns errors when changing workspaces

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/filters/workspace.ts
+++ b/packages/zendesk-adapter/src/filters/workspace.ts
@@ -67,13 +67,11 @@ const filterCreator: FilterCreator = ({ config, client }) => ({
           && response.errors !== undefined
           && Array.isArray(response.errors)
           && response.errors.length > 0) {
-          if (response.errors !== undefined) {
-            throw createSaltoElementError({ // caught by deployChanges
-              message: response.errors[0],
-              severity: 'Error',
-              elemID: getChangeData(change).elemID,
-            })
-          }
+          throw createSaltoElementError({ // caught by deployChanges
+            message: response.errors[0],
+            severity: 'Error',
+            elemID: getChangeData(change).elemID,
+          })
         }
       },
     )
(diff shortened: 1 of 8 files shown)


packages/zendesk-adapter/src/filters/workspace.ts
```diff
@@ -65,9 +67,24 @@ const filterCreator: FilterCreator = ({ config, client }) => ({
...
+          if (Array.isArray(response) && response.length > 0) {
+            errors = response[0]?.errors
+          } else {
+            errors = (response as ResponseValue)?.errors
+          }
+          if (Array.isArray(errors) && errors.length > 0) {
+            throw createSaltoElementError({ // caught by deployChanges
+              message: errors[0],
+              severity: 'Error',
+              elemID: getChangeData(change).elemID,
```
edenhassid, line 84: i think you can add a schema guard and only if there is an error field throw an error

edenhassid (commented):
cool :) can you add tests?

packages/zendesk-adapter/src/filters/workspace.ts
```diff
@@ -65,9 +66,23 @@ const filterCreator: FilterCreator = ({ config, client }) => ({
...
+        const response = await deployChange(
           change, client, config.apiDefinitions, ['selected_macros'],
         )
+        // It's possible for the deployment to return with status 200 and still have errors.
+        if (response !== undefined
+          && !_.isArray(response)
+          && response.errors !== undefined
+          && Array.isArray(response.errors)
+          && response.errors.length > 0) {
+          if (response.errors !== undefined) {
```
edenhassid, line 78: why do you test this twice?

edenhassid (approved):
looks good! added one small comment :)

PR #5441, opened 2024-02-12
PROD-232 enable features

dantal4:
Enable the "fetch bundles" and "resolve account specific values" features by default.

---

---
_Release Notes_: 
Netsuite Adapter:
- Enable the "fetch bundles" and "resolve account specific values" features by default

---
_User Notifications_: 
Netsuite Adapter:
- `bundle` instances will be fetched
- `bundle` field will be added to instances that are part of a bundle
- `ACCOUNT_SPECIFIC_VALUE` will be resolved in workflows & data elements, on fetch & deploy


omrilit (approved):
LGTM!

PR #5443, opened 2024-02-13
SALTO-5332 keep additional properties from swagger as annotation and not field

omrifarkash:
Keep additional properties annotation from swagger and stop creating a field `additionalProperties` 

---

---
_Release Notes_: 
_Jira_Adapter_:
Instances with `additionalProperties` map field will be flatten upon fetch (types to be affected: Notification Events, Workflows and maybe more)

_Okta_Adapter_:
Instances with `additionalProperties` map field will be flatten upon fetch (types to be affected: Users, GroupRule, PasswordPolicyRule and maybe more)

---
_User Notifications_: 

_Jira_Adapter_:
Instances with `additionalProperties` map field will be flatten upon fetch (types to be affected: Notification Events, Workflows and maybe more)

_Okta_Adapter_:
Instances with `additionalProperties` map field will be flatten upon fetch (types to be affected: Users, GroupRule, PasswordPolicyRule and maybe more)

### DIFF
--- a/packages/adapter-components/src/elements/swagger/instance_elements.ts
+++ b/packages/adapter-components/src/elements/swagger/instance_elements.ts
@@ -15,25 +15,14 @@
  */
 import _ from 'lodash'
 import {
-  InstanceElement,
-  Values,
-  ObjectType,
-  isObjectType,
-  ReferenceExpression,
-  isReferenceExpression,
-  isListType,
-  isMapType,
-  TypeElement,
-  PrimitiveType,
-  MapType,
-  ElemIdGetter,
-  SaltoError,
+  InstanceElement, Values, ObjectType, isObjectType, ReferenceExpression, isReferenceExpression,
+  isListType, ElemIdGetter, SaltoError, isMapType,
 } from '@salto-io/adapter-api'
-import { transformElement, TransformFunc, safeJsonStringify, transformValues } from '@salto-io/adapter-utils'
+import { TransformFunc, safeJsonStringify, transformValues } from '@salto-io/adapter-utils'
 import { logger } from '@salto-io/logging'
 import { collections, values as lowerdashValues } from '@salto-io/lowerdash'
 import { ADDITIONAL_PROPERTIES_FIELD, ARRAY_ITEMS_FIELD } from './type_elements/swagger_parser'
-import { InstanceCreationParams, shouldRecurseIntoEntry, toBasicInstance } from '../instance_elements'
+import { shouldRecurseIntoEntry, toBasicInstance } from '../instance_elements'
 import { UnauthorizedError, Paginator, PageEntriesExtractor, HTTPError } from '../../client'
 import {
   TypeSwaggerDefaultConfig,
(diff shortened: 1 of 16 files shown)


omrifarkash:
Waiting for @netama to review this one, Zuora e2e is failing and I wonder if this is an expected side effect

packages/adapter-components/src/elements/swagger/instance_elements.ts
```diff
@@ -267,11 +210,11 @@ const generateInstancesForType = ({
     .toArray()
 }
 
+const isItemsOnlyObjectType = (type: ObjectType): boolean => _.isEqual(Object.keys(type.fields), [ARRAY_ITEMS_FIELD])
+
```
omrifarkash, line 214: Just order change by mistake, irrelevant

packages/adapter-utils/src/utils.ts
```diff
@@ -1184,25 +1184,26 @@ export const getSubtypes = async (types: ObjectType[], validateUniqueness = fals
...
-
-      if (!isObjectType(fieldType) || types.includes(fieldType)) {
-        return
-      }
-      if (fieldType.elemID.getFullName() in subtypes) {
-        if (validateUniqueness && !subtypes[fieldType.elemID.getFullName()].isEqual(fieldType)) {
-          log.warn(`duplicate ElemIDs of subtypes found. The duplicate is ${fieldType.elemID.getFullName()}`)
+    const fieldsTypes = await Promise.all(Object.values(type.fields).map(field => field.getType()))
+    const additionalPropertiesRefType = type.annotations[CORE_ANNOTATIONS.ADDITIONAL_PROPERTIES]?.refType
+    await awu(additionalPropertiesRefType ? [...fieldsTypes, additionalPropertiesRefType] : fieldsTypes).forEach(
```
netama, line 1189: nit: there's a preference for explicit checks
```suggestion
    await awu(additionalPropertiesRefType !== undefined ? [...fieldsTypes, additionalPropertiesRefType] : fieldsTypes).forEach(
```

packages/jira-adapter/src/filters/notification_scheme/notification_events.ts
```diff
@@ -92,7 +92,9 @@ export const transformNotificationEvent = (notificationEvent: NotificationEvent)
   notificationEvent.notifications?.forEach((notification: Values) => {
     notification.type = notification.notificationType
     delete notification.notificationType
-    delete notification.additionalProperties
+    delete notification.recipient
```
netama, line 95: probably best to ask @IdoZakk to review the jira part

packages/jira-adapter/src/filters/workflow/workflow_structure_filter.ts
```diff
@@ -40,7 +40,7 @@ const NOT_FETCHED_POST_FUNCTION_TYPES = ['GenerateChangeHistoryFunction', 'Issue
 const FETCHED_ONLY_INITIAL_POST_FUNCTION = ['UpdateIssueStatusFunction', 'CreateCommentFunction', 'IssueStoreFunction']
 
 const transformProperties = (item: Status | Transition): void => {
-  item.properties = item.properties?.additionalProperties
+  // item.properties = item.properties?.additionalProperties
```
netama, line 43: leftover?

packages/okta-adapter/src/filters/unordered_lists.ts
```diff
@@ -48,7 +48,6 @@ const orderPasswordPolicyRuleMethods = (instance: InstanceElement): void => {
   const methodsPath = instance.elemID.createNestedID(
     'actions',
     'selfServicePasswordReset',
-    'additionalProperties',
```
netama, line 51: @shir-reifenberg can you please review the okta changes?

netama (commented):
cool :) the fetch-related shared changes LGTM.
there is going to be noise here (we are changing element structure), so we should add release notes for each of the impacted adapters + a user notifications section and a noise-reduction PR (unless we decide it's not needed). this applies to jira+okta+zuora at least. (might be easier to discuss IRL)
also - can you please add TODO and mention the follow-up ticket for removing the deploy-side flattening?

shir-reifenberg (commented):
Okta changes LGTM 😃

packages/jira-adapter/src/filters/notification_scheme/notification_events.ts
```diff
@@ -92,7 +92,9 @@ export const transformNotificationEvent = (notificationEvent: NotificationEvent)
...
+    delete notification.field
+    delete notification.projectRole
```
IdoZakk, line 97: When testing this, I also found only those additional properties. 
Additional properties are not part of the possible answer in their documentation. Looking at the swagger, we can see that the type there is, in fact, PremissionHolder, which carries additional properties.
To increase confidence, I suggest you open a ticket to Atlassian (let me know if you need help) and ask if those are indeed the possible values here.

packages/jira-adapter/src/filters/workflow/workflow_structure_filter.ts
```diff
@@ -40,7 +40,6 @@ const NOT_FETCHED_POST_FUNCTION_TYPES = ['GenerateChangeHistoryFunction', 'Issue
 const FETCHED_ONLY_INITIAL_POST_FUNCTION = ['UpdateIssueStatusFunction', 'CreateCommentFunction', 'IssueStoreFunction']
 
 const transformProperties = (item: Status | Transition): void => {
-  item.properties = item.properties?.additionalProperties
```
IdoZakk, line 43: removing this might introduce fields that were there in `properties` but not in `additionalProperties`, but as we are changing the workflows already and this is undocumented API I think it is reasonable

IdoZakk (commented):
A bit risky but looks ok on the Jira side. I added something we can check to reduce mitigation.
Note that @shayc331 is working on the new workflow API. We need to verify it does not break it as they have some additional properties there

netama (approved):
LGTM (up to coordinating with the noise reduction)

please add release notes + notifications for zuora as well

PR #5450, opened 2024-02-13
SALTO-4741 - Salesforce: don't allow deploying instances without their types

tomsellek:
In some "Compare and Deploy" cases there may be cases where instances are deployed to the new environment but their types are not.
Let's create an explicit warning in such cases.
---

Tests:
 - Delete the `Account` folder in a workspace, add an `Account` instance and deploy. Observe the CV warning.

---
_Release Notes_: 
Salesforce: Warn when attempting to add instances without their types

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/change_validators/instance_with_unknown_type.ts
+++ b/packages/salesforce-adapter/src/change_validators/instance_with_unknown_type.ts
@@ -15,26 +15,29 @@
  */
 import {
   ChangeError,
-  ChangeValidator, getChangeData, InstanceElement, isAdditionOrModificationChange, isInstanceChange,
+  ChangeValidator,
+  getChangeData,
+  InstanceElement,
+  isAdditionOrModificationChange,
+  isInstanceChange,
   isPlaceholderObjectType,
 } from '@salto-io/adapter-api'
 
-const createInstanceWithoutTypeError = (instance: InstanceElement): ChangeError => (
-  {
-    elemID: instance.elemID,
-    message: 'Instance added without its type',
-    detailedMessage: `The record ${instance.elemID.getFullName()} is being added, but its type (${instance.getTypeSync().elemID.getFullName()}) is not.`,
-    severity: 'Error',
-  }
-)
+const createInstanceWithoutTypeError = (
+  instance: InstanceElement,
+): ChangeError => ({
+  elemID: instance.elemID,
+  message: 'Instance added without its type',
+  detailedMessage: `The record ${instance.elemID.getFullName()} is being added, but its type (${instance.getTypeSync().elemID.getFullName()}) is not.`,
+  severity: 'Error',
+})
 
-const changeValidator: ChangeValidator = async changes => (
+const changeValidator: ChangeValidator = async (changes) =>
   changes
     .filter(isAdditionOrModificationChange)
     .filter(isInstanceChange)
     .map(getChangeData)
-    .filter(instance => isPlaceholderObjectType(instance.getTypeSync()))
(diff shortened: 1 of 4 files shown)


packages/salesforce-adapter/src/change_validators/instance_with_unknown_type.ts
```diff
@@ -0,0 +1,40 @@
...
+  ChangeError,
+  ChangeValidator, getChangeData, InstanceElement, isAdditionOrModificationChange, isInstanceChange,
+  isPlaceholderObjectType,
+} from '@salto-io/adapter-api'
+
+const createInstanceWithoutTypeError = (instance: InstanceElement): ChangeError => (
+  {
+    elemID: instance.elemID,
+    message: 'Instance added without its type',
+    detailedMessage: `The record ${instance.elemID.getFullName()} is being added, but its type (${instance.getTypeSync().elemID.getFullName()}) is not.`,
```
tamtamirr, line 26: @tomermevorach WDYT?

I personally think the message is a bit misleading.
Maybe:

message: Instance without type
detailedMessage: The instance {instance elem ID} is of unknown type {type elem ID}

packages/salesforce-adapter/test/change_validators/instance_with_unknown_type.test.ts
```diff
@@ -0,0 +1,93 @@
...
+  describe('When an instance is added without its type', () => {
+    const instance = new InstanceElement('SomeAccount', new TypeReference(mockTypes.Account.elemID))
+    beforeEach(async () => {
+      validationResult = await changeValidator([
+        toChange({ after: instance }),
+      ])
+    })
+    it('Should raise an error', () => {
+      expect(validationResult).toHaveLength(1)
+      expect(validationResult).toSatisfyAll(error => error.elemID.isEqual(instance.elemID))
```
tamtamirr, line 64:
```suggestion
      expect(validationResult[0]).toSatisfy(error => error.elemID.isEqual(instance.elemID))
```

Nit: I wouldn't use `toSatisfyAll` on the array, and instead check the actual expected error (Considering L63 validates this is array with a single error)

tamtamirr (approved):
Approving to save cycles. Please run the message and detailed message through @tomermevorach please 🙏

packages/salesforce-adapter/test/change_validators/instance_with_unknown_type.test.ts
```diff
@@ -0,0 +1,93 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+import { ChangeError, InstanceElement, toChange, TypeReference } from '@salto-io/adapter-api'
+import changeValidator from '../../src/change_validators/instance_with_unknown_type'
+import { mockTypes } from '../mock_elements'
+
+describe('instanceWithUnknownType', () => {
+  let validationResult: ReadonlyArray<ChangeError>
```
tamtamirr, line 21:
```suggestion
  let errors: ReadonlyArray<ChangeError>
```
nit

packages/salesforce-adapter/src/change_validators/instance_with_unknown_type.ts
tomermevorach: How about this:
Message: Instance of unknown type
detailedMessage: Cannot add/modify instance {instance elem ID} because its type is unknown

tomsellek: Ack

PR #5451, opened 2024-02-14
SALTO-4536: Nest app schema and app logo under application type

shir-reifenberg:
This pr nests app-related instances under app

---

The new file tree structure will be - 

```md
── Application
    └── appName
       └── AppUserSchema
            └── appUserSchemaName.nacl
       └──AppLogo
            └── appLogoName.nacl
       └──appName.nacl
```

We should have group push instances nested as well, but there's currently a bug with `referencedInstanceNames` blocking it

---
_Release Notes_: 

_Okta_adapter_:
- On the next fetch, AppLogo and AppUserSchema types will now be nested under Application type in the element tree

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/config.ts
+++ b/packages/okta-adapter/src/config.ts
@@ -367,7 +367,7 @@ const DEFAULT_TYPE_CUSTOMIZATIONS: OktaSwaggerApiConfig['types'] = {
         },
         {
           type: 'AppUserSchema',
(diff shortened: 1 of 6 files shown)


omrilit (approved):
💯 
Maybe worth documenting in the description that it's done without migrating existing WSs file struct.
We should consider letting existing customers know about it, not sure how critical it is tho

PR #5452, opened 2024-02-14
SALTO-3483 return detailed change with base change

dantal4:
Create `DetailedChangeWithBaseChange` that is a `DetailedChange` with `baseChange: Change<Element>` in it, and return it in `PlanItem` & `FetchChange`

---

---
_Release Notes_: 
Core:
- Create `DetailedChangeWithBaseChange` and return it in `PlanItem` & `FetchChange`

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-api/src/change.ts
+++ b/packages/adapter-api/src/change.ts
@@ -94,12 +94,10 @@ export type DetailedChange<T = ChangeDataType | Values | Value> =
       after?: ElemID
     }
     path?: ReadonlyArray<string>
+    baseChange?: Change<Element>
   }
 
-export type DetailedChangeWithBaseChange<T = ChangeDataType | Values | Value> =
-  DetailedChange<T> & {
-    baseChange: Change<Element>
-  }
+export type DetailedChangeWithBaseChange = DetailedChange & Required<Pick<DetailedChange, 'baseChange'>>
 
 export type ChangeParams<T> = { before?: T; after?: T }
 
--- a/packages/adapter-api/src/change.ts
+++ b/packages/adapter-api/src/change.ts
@@ -18,7 +18,7 @@ import {
 } from '@salto-io/dag'
 import { values as lowerDashValues } from '@salto-io/lowerdash'
 import {
-  ObjectType, InstanceElement, Field, isInstanceElement, isObjectType, isField, TopLevelElement,
+  ObjectType, InstanceElement, Field, isInstanceElement, isObjectType, isField, TopLevelElement, Element,
 } from './elements'
 import { ElemID } from './element_id'
 import { Values, Value } from './values'
@@ -94,6 +94,11 @@ export type DetailedChange<T = ChangeDataType | Values | Value> =
       after?: ElemID
     }
     path?: ReadonlyArray<string>
+  }
+
+export type DetailedChangeWithBaseChange<T = ChangeDataType | Values | Value> =
+  DetailedChange<T> & {
+    baseChange: Change<Element>
   }
 
 export type ChangeParams<T> = { before?: T; after?: T }
--- a/packages/adapter-utils/src/compare.ts
+++ b/packages/adapter-utils/src/compare.ts
@@ -225,7 +225,7 @@ const getAnnotationTypeChanges = ({
   return []
 }
 
-export const toDetailedChangeWithBaseChange = (
+export const toDetailedChangeFromBaseChange = (
   baseChange: Change<Element>,
   elemIDs?: DetailedChangeWithBaseChange['elemIDs']
 ): DetailedChangeWithBaseChange => ({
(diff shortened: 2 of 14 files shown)


packages/core/src/core/restore.ts
```diff
@@ -113,20 +113,25 @@ export const createRestorePathChanges = async (
...
-    ...toChange({ before: element }),
-    id: element.elemID,
-  }))
+  const removalChanges = relevantElements.map(element => {
+    const baseChange = toChange({ before: element })
+    return {
+      ...baseChange,
+      id: element.elemID,
+      baseChange,
+    }
```
ori-moisis, line 125: maybe add a `toDetailedChange` helper to avoid repeating this logic?

packages/adapter-api/src/change.ts
```diff
@@ -96,6 +96,11 @@ export type DetailedChange<T = ChangeDataType | Values | Value> =
     path?: ReadonlyArray<string>
   }
 
+export type DetailedChangeWithBaseChange<T = ChangeDataType | Values | Value> =
+  DetailedChange<T> & {
+    baseChange: Change<Element>
+  }
```
ori-moisis, line 102: I worry that this is not a great way to maintain the interface for the future (the name is too long and specific)
I think we should aim that `DetailedChange` will contain this...

if this is really hard to accomplish right now (is it?), we can have `baseChange` as optional in `DetailedChange` and have this type as a temporary hack with
```
type DetailedChangeWithBaseChange = DetailedChange & Required<Pick<DetailedChange, 'baseChange'>>
```
but if we want to ensure a detailed change always contains a reference to its base change we should eventually make it non-optional in `DetailedChange` and remove this type...

packages/core/src/core/diff.ts
```diff
@@ -132,15 +132,15 @@ export function createDiffChanges(
...
+): Promise<DetailedChangeWithBaseChange[]>
 export async function createDiffChanges(
   toElementsSrc: elementSource.ElementsSource,
   fromElementsSrc: elementSource.ElementsSource,
   referenceSourcesIndex: remoteMap.ReadOnlyRemoteMap<ElemID[]> = new remoteMap.InMemoryRemoteMap<ElemID[]>(),
   elementSelectors: ElementSelector[] = [],
   topLevelFilters: IDFilter[] = [],
   resultType: 'changes' | 'detailedChanges' = 'detailedChanges'
-): Promise<DetailedChange[] | ChangeWithDetails[]> {
+): Promise<DetailedChangeWithBaseChange[] | ChangeWithDetails[]> {
```
ori-moisis, line 143: given that detailed changes now have a base change, do we still need the `ChangeWithDetails` type?

dantal4: I actually think that when we can return `ChangeWithDetails` it's preferred. In this function we eventually should return only `ChangeWithDetails[]`.

ori-moisis: ok

packages/adapter-api/src/change.ts
dantal4: There are too many places using/creating `DetailedChange`, so it breaks in many places when I add a required `baseChange` to `DetailedChange`.

packages/adapter-utils/src/compare.ts
```diff
@@ -223,12 +225,23 @@ const getAnnotationTypeChanges = ({
   return []
 }
 
+export const toDetailedChangeWithBaseChange = (
```
ori-moisis, line 228: since we can't really rely on typescript to validate the input here (IIRC we work with `Change<Value>` which is basically `Change<any>` in a bunch of places in the code), I think this can be confusing.
I suggest we call this `toDetailedChangeFromBaseChange` or something like that, so it would be more clear that the input here should be a base level change

packages/cli/test/mocks.ts
```diff
@@ -468,22 +469,34 @@ export const mockConfigType = (adapterName: string): ObjectType => {
...
 export const detailedChange = (
   action: 'add' | 'modify' | 'remove', path: ReadonlyArray<string> | ElemID,
   /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
   before: any, after: any,
-): DetailedChange => {
+): DetailedChangeWithBaseChange => {
   const id = path instanceof ElemID ? path : new ElemID('salesforce', ...path)
   if (action === 'add') {
-    return { action, id, data: { after } }
+    return { action, id, data: { after }, baseChange: baseChange(action) }
```
ori-moisis, line 491: note that the changes created here don't make sense (the path and value of the detailed change do not match anything in the base change)
this might be annoying for future tests

ori-moisis (commented):
missing tests for the compare functions to make sure they really return the base changes correctly?

ori-moisis (approved):
approving before the last round of fixes since I am not going to be available for a while

PR #5453, opened 2024-02-14
SALTO-5421 adjust adapter-components filters that only need user config so they will work with both versions of the system config

netama:
some of the other filters will need a "migration" from the relevant args of the old config to the new format, but these only rely on the user part so adjusting the types to have a reduced scope in order to avoid needing to support both formats. 

note this also contains some helper types + utils for filters using the new definitions.

---

this is technically ready, but I want to adjust some of the type names (used for now names that are very similar to the old format). if this is needed feel free to merge, or I'll adjust these before merging

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/definitions/system/references.ts
+++ b/packages/adapter-components/src/definitions/system/references.ts
@@ -16,6 +16,7 @@
 import { FieldReferenceDefinition } from '../../references'
 
 export type ReferenceDefinitions = {
-  // TODO use in filter
+  // rules for finding references - coverting values to references in fetch, and references to values in deploy.
+  // this is an array of arrays because rules can be run in multiple iterations during fetch,
   rules: FieldReferenceDefinition<never>[]
 }
--- a/packages/adapter-components/src/definitions/system/references.ts
+++ b/packages/adapter-components/src/definitions/system/references.ts
@@ -16,7 +16,7 @@
 import { FieldReferenceDefinition } from '../../references'
 
 export type ReferenceDefinitions = {
-  // rules for finding references - coverting values to references in fetch, and references to values in deploy.
-  // this is an array of arrays because rules can be run in multiple iterations during fetch,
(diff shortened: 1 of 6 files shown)


packages/adapter-components/src/definitions/system/references.ts
```diff
@@ -16,6 +16,7 @@
 import { FieldReferenceDefinition } from '../../references'
 
 export type ReferenceDefinitions = {
-  // TODO use in filter
+  // rules for finding references - coverting values to references in fetch, and references to values in deploy.
+  // this is an array of arrays because rules can be run in multiple iterations during fetch,
```
shir-reifenberg, line 20:
```suggestion
  // rules for finding references - converting values to references in fetch, and references to values in deploy.
  // this is an array of arrays because rules can be run in multiple iterations during fetch
```

packages/adapter-components/src/filter_utils.ts
```diff
@@ -25,6 +26,61 @@ export type FilterWith<M extends keyof Filter, TResult extends void | filter.Fil
...
+  Action extends string = ActionName,
+>(
+  opts: FilterOptions<TContext, TAdditional, ClientOptions, PaginationOptions, Action>,
+  filterCreators: ReadonlyArray<
+    AdapterFilterCreator<TContext, TResult, TAdditional, ClientOptions, PaginationOptions, Action>
+  >,
+  onFetchAggregator: (results: TResult[]) => TResult | void = () => undefined,
+): Required<Filter<TResult>> => filter.filtersRunner(opts, filterCreators, onFetchAggregator)
+
+// TODO deprecate when upgrading to new definitions
```
shir-reifenberg, line 82: Should we add a ticket here ?

shir-reifenberg (approved):
LGTM 💯

PR #5454, opened 2024-02-14
SALTO-5319: support template static file part2

edenhassid:
support template static files

---

1. added `staticFileToTemplateExpression` and `templateExpressionToStaticFile` in `parser` package
2. moved `resolveChangeElement` and `resolveValues` from `adapter-utils` to `adapter-components` because of dependency issues as they need to use `staticFileToTemplateExpression` from the `parser`.
3. moved `getAllReferencedIds` from `adapter-utils` to `adapter-components` because of dependency issues as it needs to use `staticFileToTemplateExpression` from the `parser`. also this function was turned to be async
4. added `isTemplate` flag to the `staticFile` class.

in the SAAS I need:
* to look at all the places that create a static file and make sure they take `isTemplate` into account
* to make sure that all places that use parser functions import them from the new package

---
_Release Notes_: 

---
_User Notifications_: 


packages/adapter-api/src/values.ts
```diff
@@ -48,18 +51,26 @@ type HashOrContent = {
 export type StaticFileParameters = {
   filepath: string
   encoding?: BufferEncoding
+  isTemplate?: boolean // if true then encoding has to be utf8
```
edenhassid, line 54: wasn't able to enforce this within the type

packages/workspace/src/workspace/static_files/functions.ts
```diff
@@ -22,7 +22,9 @@ export const getStaticFilesFunctions = (staticFilesSource: StaticFilesSource): p
   file: {
     parse: (parameters): Promise<StaticFile | InvalidStaticFile> => {
```
adfineberg, line 23: add tests for parse and dump here

adfineberg (commented):
Nice stuff! Added a few comments

packages/adapter-api/src/values.ts
edenhassid: tried that but then in every place a static file is created we need to add a check for the type before we create the static file.

packages/adapter-api/src/values.ts
```diff
@@ -48,18 +51,26 @@ type HashOrContent = {
...
 } & HashOrContent
 
 export const DEFAULT_STATIC_FILE_ENCODING: BufferEncoding = 'binary'
 
 export class StaticFile {
   public readonly filepath: string
   public readonly hash: string
+  public readonly isTemplate?: boolean
```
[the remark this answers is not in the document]
edenhassid: to have less code

packages/parser/src/utils/template_static_file.ts
```diff
@@ -0,0 +1,101 @@
...
+  const parts = createReferencesFromStringParts(splitByReferencesMarker(finalString))
+  const unescapedTemplateMarkerParts = parts.map(part =>
+    isReferenceExpression(part) ? part : unescapeTemplateMarker(part),
+  )
+
+  return createTemplateExpression({ parts: unescapedTemplateMarkerParts })
+}
+
+export const templateExpressionToStaticFile = (expression: TemplateExpression, filepath: string): StaticFile => {
+  const string = dumpValue(expression).toString()
```
adamro, line 85: The `toString()` here is actually legit? 🤔

packages/parser/src/utils/template_static_file.ts
```diff
@@ -0,0 +1,101 @@
...
+
+const isMultilineBuffer = (stringBuffer: string): boolean =>
+  stringBuffer.startsWith(MULTILINE_STRING_PREFIX) && stringBuffer.endsWith(MULTILINE_STRING_SUFFIX)
+
+const removeMultilineStringPrefixAndSuffix = (prim: string): string =>
+  _.trimStart(_.trimEnd(prim, MULTILINE_STRING_SUFFIX), MULTILINE_STRING_PREFIX)
+
+const returnDoubleTemplateMarkerEscaping = (prim: string): string => prim.replace(/\\\$\{/g, '\\\\${')
+
+const splitByReferencesMarker = (stringBuffer: string): string[] => {
```
adamro, line 35: nitagain: the word buffer is irrelevant at this point

packages/parser/src/utils/template_static_file.ts
```diff
@@ -0,0 +1,101 @@
...
+  const regex = /(?<!\\)\$\{[^}]+\}/g
```
adamro, line 36: What does this regex represent? How come it's only used here in the code base? (and why is it scoped to this function?)

(It's supposed to be what's inside a ${...}, right?)

packages/parser/src/utils/template_static_file.ts
adamro, line 35: suggestion I don't feel super strong about - instead of returning a simple string[] return an array of something that has both the string and an indication if it's just string or "templated" (inside a ${}). This way when you need to extract the references you only need to run on the "templated" areas and not run logic that is not tied to this to differentiate between the parts again (ie. line 57)

adamro (commented):
Mid-way reviewing but I have some comments and questions so putting it out there

packages/parser/src/utils/template_static_file.ts
edenhassid: dumpValue(expression) returns a list of strings

packages/parser/src/utils/template_static_file.ts
```diff
@@ -0,0 +1,101 @@
...
+      const ref = createReferenceExpresion(refParts[1].replace(/\s+/g, ''))
+      return ref instanceof IllegalReference ? part : ref
+    }
+    return part
+  })
+
+const parseBufferToTemplateExpression = (buffer: Buffer): TemplateExpression => {
+  const stringBuffer = buffer.toString()
+  const removed = removeMultilineStringPrefixAndSuffix(stringBuffer)
+  const finalString = isMultilineBuffer(stringBuffer)
```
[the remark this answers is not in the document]
edenhassid: discussed f2f

packages/parser/src/utils/template_static_file.ts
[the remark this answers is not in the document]
edenhassid: I am using `createTemplateExpression` all the other functions there prepare the template expression before deploy

packages/adapter-api/src/values.ts
```diff
@@ -49,18 +52,26 @@ type HashOrContent =
...
   public readonly encoding: BufferEncoding
   private internalContent?: Buffer
   constructor(params: StaticFileParameters) {
+    this.isTemplate = params.isTemplate
     this.filepath = path.normalize(params.filepath)
     this.encoding = params.encoding ?? DEFAULT_STATIC_FILE_ENCODING
+    if (this.isTemplate === true && params.encoding !== 'utf8') {
```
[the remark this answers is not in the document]
edenhassid: removing the log.warn

packages/parser/src/utils/template_static_file.ts
```diff
@@ -0,0 +1,101 @@
...
+import { unescapeMultilineMarker } from '../parser/internal/native/consumers/values'
+import { createReferenceExpresion, unescapeTemplateMarker } from '../parser/internal/native/helpers'
+import { IllegalReference } from '../parser'
+
+const log = logger(module)
+
+const isMultilineBuffer = (stringBuffer: string): boolean =>
+  stringBuffer.startsWith(MULTILINE_STRING_PREFIX) && stringBuffer.endsWith(MULTILINE_STRING_SUFFIX)
+
+const removeMultilineStringPrefixAndSuffix = (prim: string): string =>
```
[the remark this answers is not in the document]
edenhassid: just copied from `/parser/src/parser/internal/native/consumers/values.ts` where prim is used

packages/parser/src/utils/template_static_file.ts
```diff
@@ -0,0 +1,101 @@
...
+  _.trimStart(_.trimEnd(prim, MULTILINE_STRING_SUFFIX), MULTILINE_STRING_PREFIX)
+
+const returnDoubleTemplateMarkerEscaping = (prim: string): string => prim.replace(/\\\$\{/g, '\\\\${')
+
+const splitByReferencesMarker = (stringBuffer: string): string[] => {
+  const regex = /(?<!\\)\$\{[^}]+\}/g
+  const parts: string[] = []
```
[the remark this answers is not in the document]
edenhassid: it doesn't work

packages/core/test/common/workspace.ts
```diff
@@ -143,9 +143,11 @@ export const mockWorkspace = ({
     getParsedNaclFile: async (filename: string) => ({
       elements: async () => parsedNaclFiles[filename],
     }),
-    getStaticFile: ({ filepath, encoding }: { filepath: string; encoding: BufferEncoding }) =>
+    getStaticFile: (
+      { filepath, encoding, isTemplate }: { filepath: string; encoding: BufferEncoding; isTemplate?: boolean }
+    ) =>
       (staticFilesSource
-        ? staticFilesSource.getStaticFile(filepath, encoding)
+        ? staticFilesSource.getStaticFile(filepath, encoding, undefined, isTemplate)
```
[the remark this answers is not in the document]
edenhassid: changed in https://github.com/salto-io/salto/pull/5517

packages/workspace/src/workspace/static_files/common.ts
```diff
@@ -26,7 +26,12 @@ export abstract class InvalidStaticFile {
 export type StaticFilesSource = {
   // Load is optional for backwards compatibility
   load?(): Promise<string[]>
-  getStaticFile: (filepath: string, encoding: BufferEncoding, hash?: string) => Promise<StaticFile | InvalidStaticFile>
+  getStaticFile: (
```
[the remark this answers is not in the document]
edenhassid: changed in https://github.com/salto-io/salto/pull/5517

packages/parser/src/utils/template_static_file.ts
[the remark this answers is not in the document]
edenhassid: @tomermevorach said that we shouldn't change the static file besides adding references and escaping any other `${`. I'll fix that

packages/parser/src/utils/template_static_file.ts
```diff
@@ -0,0 +1,91 @@
...
+import { escapeTemplateMarker } from '../parser/internal/dump'
+import { createReferenceExpresion, unescapeTemplateMarker } from '../parser/internal/native/helpers'
+import { IllegalReference } from '../parser'
+import { REFERENCE_PART } from '../parser/internal/native/lexer'
+
+const log = logger(module)
+
+type PartWithReferenceIndicator = { isReference: boolean; part: string }
+
+const splitByReferencesMarker = (stringToParse: string): PartWithReferenceIndicator[] => {
```
adfineberg, line 28: Maybe generalize this (add the pattern to split in the params) and extract to a util function? A bit like `continuousSplit`. Although, not sure what our policy is about these things

packages/parser/src/utils/template_static_file.ts
```diff
@@ -0,0 +1,91 @@
...
+const createReferencesFromStringParts = (parts: PartWithReferenceIndicator[]): TemplatePart[] =>
+  parts.map(part => {
+    if (part.isReference) {
+      // remove the ${ and }
+      const refParts = part.part.split(/\${|}/)
+      if (refParts.length !== 3) {
+        log.warn(`refParts is invalid, received ${refParts.toString()}`)
+        return part.part
+      }
+      const ref = createReferenceExpresion(refParts[1].replace(/\s+/g, ''))
```
adfineberg, line 57: Are you trimming here, or did you mean to remove any spaces within the expression as well? If it's just trimming, just use `.trim()`

packages/workspace/src/workspace/reference_indexes.ts
```diff
@@ -76,21 +88,29 @@ const getReferences = (element: Element, customReferences: ReferenceInfo[]): Ref
   customReferences.forEach(refInfo => {
     references[getReferenceDetailsIdentifier(refInfo)] = refInfo
   })
+  await awu(templateStaticFiles)
```
adfineberg, line 91: Why have a separate loop here instead of running this logic in line 82?

packages/workspace/src/workspace/static_files/functions.ts
```diff
@@ -22,7 +22,9 @@ export const getStaticFilesFunctions = (staticFilesSource: StaticFilesSource): p
...
       const [filepath, encoding] = parameters
-      return staticFilesSource.getStaticFile({ filepath, encoding })
+      const finalEncoding = encoding === 'template' ? 'utf8' : encoding
```
adfineberg, line 25: WDYT about having `'template'` a const? Overkill?

adfineberg (commented):
Looks great! A few small comments :)

packages/workspace/src/workspace/reference_indexes.ts
edenhassid: cause we cant make `walkOnElement` to be async, so we have to collect all template static files and then run the logic in a separate loop

packages/core/src/core/plan/dependency/reference_dependency.ts
```diff
@@ -125,6 +128,10 @@ export const addReferencesDependency: DependencyChanger = async changes => {
   }
 
   return awu(changes)
-    .flatMap(entry => [...addChangeDependency(entry), ...addBeforeModificationDependency(entry)])
+    .flatMap(async entry => {
+      const changeDependency = await addChangeDependency(entry)
```
adamro, line 132: Maybe use Promise.all here as they are not dependent?

packages/parser/src/parser/internal/dump.ts
```diff
@@ -50,7 +50,7 @@ const separateByCommas = (items: string[][]): string[][] => {
 
 const isMultilineString = (prim: string): boolean => _.isString(prim) && prim.includes('\n')
 
-const escapeTemplateMarker = (prim: string): string => prim.replace(/\$\{/gi, '\\${')
+export const escapeTemplateMarker = (prim: string): string => prim.replace(/\$\{/gi, '\\${')
```
adamro, line 53: nit: I wonder if this is the best placing for this as an exported function, I am not sure. Feel a bit random :)

packages/parser/src/parser/internal/native/consumers/values.ts
```diff
@@ -256,7 +256,7 @@ const consumeArrayItems = (context: ParseContext, closingTokenType: string, idPr
   return items
 }
 
-const unescapeMultilineMarker = (prim: string): string => prim.replace(/\\'''/g, "'''")
+export const unescapeMultilineMarker = (prim: string): string => prim.replace(/\\'''/g, "'''")
```
adamro, line 259: nit: I wonder if this is the best placing for this as an exported function, I am not sure. Feel a bit random :)

packages/parser/src/utils/template_static_file.ts
```diff
@@ -0,0 +1,91 @@
...
+const parseBufferToTemplateExpression = (buffer: Buffer): TemplateExpression => {
+  const parts = createReferencesFromStringParts(splitByReferencesMarker(buffer.toString()))
+  const unescapedTemplateMarkerParts = parts.map(part =>
+    isReferenceExpression(part) ? part : unescapeTemplateMarker(part),
+  )
+
+  return createTemplateExpression({ parts: unescapedTemplateMarkerParts })
+}
+
+export const templateExpressionToStaticFile = (expression: TemplateExpression, filepath: string): StaticFile => {
```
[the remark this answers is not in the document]
edenhassid: no, it will only change the encoding to be 'utf8' and whenever `staticFileToTemplateExpression` will be used it will return a template expression with a single part of a string.

packages/core/test/common/workspace.ts
```diff
@@ -144,7 +144,14 @@ export const mockWorkspace = ({
...
-      staticFilesSource ? staticFilesSource.getStaticFile({ filepath, encoding }) : undefined,
+    getStaticFile: ({
+      filepath,
+      encoding,
+      isTemplate,
+    }: {
+      filepath: string
+      encoding: BufferEncoding
+      isTemplate?: boolean
+    }) => (staticFilesSource ? staticFilesSource.getStaticFile({ filepath, encoding, isTemplate }) : undefined),
```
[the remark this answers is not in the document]
edenhassid: the `getStaticFile` should create a staticfile for which isTemplate if needed. This is later used in other functions to know how to parse the file.

packages/parser/src/parser/internal/native/consumers/values.ts
edenhassid: moved to a utils file

packages/workspace/src/workspace/static_files/functions.ts
edenhassid: its only used in this file 3 times,  so I don't think its necessary :)

packages/parser/src/utils/template_static_file.ts
edenhassid: discussed f2f will leave it like this for now

packages/core/src/core/plan/dependency/reference_dependency.ts
```diff
@@ -58,73 +60,81 @@ const isReferenceValueChanged = (change: Change<ChangeDataType>, refElemId: Elem
...
+        .filter(values.isDefined)
+    }
 
-  return awu(changes)
-    .flatMap(entry => [...addChangeDependency(entry), ...addBeforeModificationDependency(entry)])
-    .toArray()
-}
+    const result = await Promise.all(
+      Array.from(changes.entries()).flatMap(async entry => {
+        const changeDependency = await addChangeDependency(entry)
```
adamro, line 134: nit: We can still Promise.all these 2 as well if we want more parallelism

adamro (approved):
🔥 🔥 🔥

PR #5456, opened 2024-02-14
SALTO-5440: adjust workflow properties cv to the new Workflow structure

shayc331:
_adjust workflow properties cv to the new Workflow structure_

---

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


packages/jira-adapter/src/constants.ts
```diff
@@ -24,6 +24,8 @@ export const PRIORITY_TYPE_NAME = 'Priority'
 export const RESOLUTION_TYPE_NAME = 'Resolution'
 export const WORKFLOW_TYPE_NAME = 'Workflow'
 export const JIRA_WORKFLOW_TYPE = 'JiraWorkflow'
+export const WORKFLOW_V1 = 'workflowV1'
```
shayc331, line 27: note: 
WDYT about locating those constants in this file?

kami-delaroz: Because it's in used only by the test, I think it should be in a test itself

packages/jira-adapter/test/change_validators/workflows/workflow_properties.test.ts
```diff
@@ -13,140 +13,139 @@
...
-      'instance',
-      type,
-      {
-        transitions:
-        { tran1: {
-          name: 'tran1',
-          properties: [
+      )
+    }
+    if (workflowVersion === WORKFLOW_V2) {
```
kami-delaroz, line 55: `else if`, or just `else`, is clearer IMO

kami-delaroz (approved):

PR #5460, opened 2024-02-15
SALTO-5445: Add the ability to configure FLS Profiles in Salesforce

tamtamirr:
Add the ability to configure FLS Profiles in Salesforce.

---

Prior to this PR, when deploying new CustomObjects and CustomFields, the adapter added visibility to the newly created fields only to the `System Administrator` Profile.

This PR allows configuring which Profiles Salto should add visibility for.

---
_Release Notes_: 
Salesforce Adapter:
- Add the ability to configure which Profiles should have visibility to newly created CustomObjects and CustomFields.

---
_User Notifications_: 
_None_


tomsellek (approved):
One comment. Approving to save cycles.

PR #5461, opened 2024-02-15
Salto 5244/exclude dependencies of zendesk config

oneYAD:
BugFix - now, when fetching guide with category\section excluded instead of getting Error, the dependent instances will excluded too.
the dependent hierarchy is article -> section -> category.

The user warned about exactly which item is dependent on another items. 
---

solved - https://salto-io.atlassian.net/browse/SALTO-5244 

---
_Release Notes_: 
Now you can exclude section, category and articles without getting Exception.

---
_User Notifications_: 
None


packages/zendesk-adapter/src/adapter.ts
```diff
@@ -281,6 +286,44 @@ const SKIP_RESOLVE_TYPE_NAMES = [
...
+    target: CATEGORY_TYPE_NAME,
+  },
+  {
+    source: ARTICLE_TYPE_NAME,
+    target: SECTION_TYPE_NAME,
+  },
+]
+
+
+const excludeGuideDependentTypes = (excludeElements: FetchEntry<DefaultFetchCriteria>[]):
```
edenhassid, line 307: did you try to use `fetchQuery.isTypeMatch`?

oneYAD: I didn't know it, thanks.
I'm not sure it can help me. I can change the if statement in line 314, but it will be less efficient and I'm not sure it will be more readable. 
what do you think?

packages/zendesk-adapter/src/adapter.ts
```diff
@@ -582,7 +625,7 @@ export default class ZendeskAdapter implements AdapterOperations {
     const combinedRes = {
       configChanges: (defaultSubdomainResult.configChanges ?? []),
       elements: defaultSubdomainResult.elements,
-      errors: (defaultSubdomainResult.errors ?? []),
+      errors: ([...errors, ...(defaultSubdomainResult.errors ?? [])]),
```
edenhassid, line 628:
```suggestion
      errors: errors.concat.(defaultSubdomainResult.errors ?? [])]),
```

packages/zendesk-adapter/src/adapter.ts
```diff
@@ -556,6 +597,8 @@ export default class ZendeskAdapter implements AdapterOperations {
 
   @logDuration('generating instances and types from service')
   private async getElements(): Promise<ReturnType<typeof getAllElements>> {
+    const { excludeEntries, errors } = excludeGuideDependentTypes(this.userConfig[FETCH_CONFIG].exclude)
+    this.userConfig[FETCH_CONFIG].exclude = excludeEntries
```
edenhassid, line 601: does this change their config after? if so I think this should be under config suggestion

packages/zendesk-adapter/src/adapter.ts
```diff
@@ -281,6 +284,44 @@ const SKIP_RESOLVE_TYPE_NAMES = [
...
+    target: CATEGORY_TYPE_NAME,
+  },
+  {
+    source: ARTICLE_TYPE_NAME,
+    target: SECTION_TYPE_NAME,
+  },
+]
+
+
+const excludeGuideDependentTypes = (excludeElements: definitions.FetchEntry<definitions.DefaultFetchCriteria>[]):
```
edenhassid, line 305: what happens if we exclude only category that has a certain name? we will exclude all sections right?

edenhassid (commented):
cool! 😎 added a few questions and comments

packages/zendesk-adapter/src/adapter.ts
```diff
@@ -281,6 +284,44 @@ const SKIP_RESOLVE_TYPE_NAMES = [
   ...GUIDE_ORDER_TYPES,
 ]
 
+// Source is dependent on target, when target is excluded source should be also excluded
+type ExcludeDependency = {
+  source: string
+  target: string
```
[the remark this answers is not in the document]
oneYAD: I used the core terminology for dependencies. I can change it, but maybe parent and child is confusing with parent and child of the instance element itself?

packages/zendesk-adapter/src/adapter.ts
oneYAD: sounds correct, cool feature

PR #5465, opened 2024-02-16
SALTO-4690 - Salesforce: CV for modification/deletion of CurrencyIsoCode instance

tomsellek:
`CurrencyIsoCodes` is a synthetic type we create. The type (and its instances) can't be deployed because they don't exist on the service.

---

image

---
_Release Notes_: 
Salesforce: Warn when attempting to modify the list of supported currencies.

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/change_validators/artifical_types.ts
+++ b/packages/salesforce-adapter/src/change_validators/artifical_types.ts
@@ -23,9 +23,9 @@ import { CURRENCY_CODE_TYPE_NAME } from '../constants'
 
 const createCurrencyIsoCodeError = (element: Element): ChangeError => ({
   elemID: element.elemID,
-  message: "The list of supported currency codes can't be changed via Salto.",
+  message: "The list of currency codes can't be changed via Salto.",
   detailedMessage:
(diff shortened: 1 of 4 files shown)


packages/salesforce-adapter/src/change_validators/artifical_types.ts
```diff
@@ -0,0 +1,38 @@
...
+  ChangeError,
+  ChangeValidator,
+  Element,
+  getChangeData,
+} from '@salto-io/adapter-api'
+import { CURRENCY_CODE_TYPE_NAME } from '../constants'
+
+const createCurrencyIsoCodeError = (element: Element): ChangeError => ({
+  elemID: element.elemID,
+  message: "The list of supported currency codes can't be changed via Salto.",
```
tomermevorach, line 26: I think we should remove the word 'supported' from the message as well as the detailed message.
Other than that it looks good to me

tamtamirr (approved):
LGTM.
We should add UT that covers a change on the `CurrencyIsoCodes` type and the `CurrencyIsoCodes` instance.

Approving to save cycles

PR #5468, opened 2024-02-18
SALTO-5453: Fix Jsm Fetch Failing

idoamit8:
In this PR I fix the bug when user tries to fetch JSM while he doesn't have JSM in service. 

---

---
_Release Notes_: 
_Jira Adapter:_
Fixed bug when JSMEnable = true but no JSM instance in the service.

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/adapter.ts
+++ b/packages/jira-adapter/src/adapter.ts
@@ -14,7 +14,7 @@
 * limitations under the License.
 */
 import _ from 'lodash'
-import { Element, FetchResult, AdapterOperations, DeployResult, InstanceElement, TypeMap, isObjectType, FetchOptions, DeployOptions, Change, isInstanceChange, ElemIdGetter, ReadOnlyElementsSource, ProgressReporter, FixElementsFunc, isInstanceElement } from '@salto-io/adapter-api'
+import { Element, FetchResult, AdapterOperations, DeployResult, InstanceElement, TypeMap, isObjectType, FetchOptions, DeployOptions, Change, isInstanceChange, ElemIdGetter, ReadOnlyElementsSource, ProgressReporter, FixElementsFunc, isInstanceElement, Value } from '@salto-io/adapter-api'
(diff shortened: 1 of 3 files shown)


packages/jira-adapter/src/adapter.ts
```diff
@@ -577,6 +577,20 @@ export default class JiraAdapter implements AdapterOperations {
...
     }
+    const accountLicense = await getCloudLicense(this.client)
+    const isJsmEnabledInService = accountLicense.applications?.some((app: Value) => app.id === 'jira-servicedesk')
+    if (!isJsmEnabledInService) {
+      log.debug('enableJSM set to true, but JSM is not enabled in the service, skipping fetching JSM elements')
+      return {
+        elements: [],
+        errors: [
+          {
+            message: 'Jira Service Management is not enabled in the service, skipping fetching JSM elements',
```
idoamit8, line 588: @liorn Would appreciate your input for the fetch warning.
This is how it looks at the moment: 
image

liorn: "Jira Service Management is not enabled in this Jira instance. Skipping fetch of JSM elements."

That's enough

packages/jira-adapter/src/adapter.ts
```diff
@@ -577,6 +577,20 @@ export default class JiraAdapter implements AdapterOperations {
       || !this.userConfig.fetch.enableJSM) {
       return { elements: [] }
     }
+    const accountLicense = await getCloudLicense(this.client)
```
omrilit, line 580: There is an assumption here that JSM is supported in Salto only in Cloud and not in DC, which is indeed the case TODAY.
Could you please think on a way to make sure in the code that it will work also once JSM DC is supported? maybe defining something similar to `isJsmEnabledInService` func in a relevant location would help to achieve it?

omrilit (approved):
Looks great!
had 1 comment but it doesn't block merging.

omrilit (approved):

PR #5470, opened 2024-02-18
SALTO-5010: [Zendesk] Fix race condition when creating new variants for a dynamic_content_item

DanielBarros1:
When creating new variants, we send 2 requests:
1. to modify the parent item
2. to create the variants.
This can create a race condition in Zendesk that returns an error.
By ordering the deployments correctly + ignoring the variants field on the item, we make sure the 2 requests don't contradict each other

---

_Test Plan_:

Cases i tested manually:
1. Add 2 variants and modify item
2. Create item + variant
3. Delete variant
4. Delete item + variant
5. Modify variant
6. Modify item

---

_Release Notes_: 
Zendesk: 
Fix race condition when creating new variants for a dynamic_content_item

---

_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/filters/dynamic_content.ts
+++ b/packages/zendesk-adapter/src/filters/dynamic_content.ts
@@ -76,6 +76,7 @@ const filterCreator: FilterCreator = ({ config, client }) => ({
       relevantChanges,
       change => getChangeData(change).elemID.typeName === DYNAMIC_CONTENT_ITEM_TYPE_NAME,
     )
+    relevantChanges.map(change => console.log(change.data))
     if (itemChanges.length === 0 || itemChanges.every(isModificationChange)) {
       // The service does not allow us to have an item with no variant - therefore, we need to do
       //  the removal changes last
--- a/packages/zendesk-adapter/src/filters/dynamic_content.ts
+++ b/packages/zendesk-adapter/src/filters/dynamic_content.ts
@@ -1,18 +1,21 @@
 /*
- *                      Copyright 2024 Salto Labs Ltd.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with
- * the License.  You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
+*                      Copyright 2024 Salto Labs Ltd.
+*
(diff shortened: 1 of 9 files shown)


packages/zendesk-adapter/src/filters/dynamic_content.ts
```diff
@@ -71,12 +71,25 @@ const filterCreator: FilterCreator = ({ config, client }) => ({
...
+      )
+      const [itemRemovalChanges, itemNonRemovalChanges] = _.partition(itemChanges, isRemovalChange)
+
       const deployResult = await deployChangesByGroups(
-        [nonRemovalChanges, removalChanges] as Change<InstanceElement>[][],
+        [
+          variantAdditionChanges,
+          itemNonRemovalChanges,
+          variantModificationChanges,
+          itemRemovalChanges,
```
edenhassid, line 90: you don't get an error if you remove an item and its variant after?

packages/zendesk-adapter/src/filters/dynamic_content.ts
```diff
@@ -71,12 +71,25 @@ const filterCreator: FilterCreator = ({ config, client }) => ({
       relevantChanges,
       change => getChangeData(change).elemID.typeName === DYNAMIC_CONTENT_ITEM_TYPE_NAME,
     )
+
     if (itemChanges.length === 0 || itemChanges.every(isModificationChange)) {
```
edenhassid, line 75: we are keeping the every? what happens if we have and addition of an item and a modification in the same deploy? do they get to this filter in different groups?

edenhassid (commented):
looks good!! 💯 🤩 added some small questions

edenhassid (commented):
did you also test modifying an item by adding a variant and removing another variant?

packages/zendesk-adapter/src/filters/dynamic_content.ts
DanielBarros1: I checked a deployment with both addition and modification and it seems like it's working like you suggested - two different groups, so it works well i guess :)

packages/zendesk-adapter/src/filters/dynamic_content.ts
DanielBarros1: Nope! deleted a variant and an item and the deployment didn't throw

packages/zendesk-adapter/src/filters/dynamic_content.ts
edenhassid, line 75: i think that if we get here then there is only one item in the change group, and then we don't need the `itemRemovalChanges,` in the code. WDYT?

DanielBarros1: I think you're probably right, but i'd probably opt to leave it as is - this way the ordering of the deployment is as explicit and verbose as possible, saves us from having to add a bunch of comments explaining the precise circumstances.

edenhassid (approved):
LGTM!! 😄 💯

PR #5473, opened 2024-02-18
SALTO-5438: Removed values from Application settings are not getting deployed (Okta)

shir-reifenberg:
Fix an issue with application deployment when removing fields from `settings` object didn't actually removed those from the service

---

The issue was when deploying an app with removed properties from `settings` - the request returned successfully, but the properties doesn't get removed from the service, so we basically report that the deploy was successful though it's not.
The solution is to set `null` explicitly for removed fields. As this might be a pattern we'll encounter in the future, I added function to adapter components.
 
In Okta, I made some research and it seems like this pattern only relevant for App instances under `settings` object (and for schema which we already have a dedicated filter)

---
_Release Notes_: 

_Okta_adapter_:
- Fix a bug in application deployment when removing application settings
 
---
_User Notifications_: 
None


packages/adapter-components/src/deployment/deployment.ts
```diff
@@ -76,6 +76,36 @@ export const filterIgnoredValues = async (
...
+    elemId,
+    value: resolvePath(before, elemId),
+    func: ({ value, path }) => {
+      const valueInAfter = resolvePath(after, path)
+      if (valueInAfter === undefined) {
+        setPath(after, path, null)
+        return WALK_NEXT_STEP.SKIP
+      }
+      // Arrays are being skipped to avoid setting null to removed array elements
+      if (Array.isArray(value)) {
```
omrilit, line 100: Do we have other data types that should get teh same treatment? Map for example?

omrilit (approved):
Great work!!
1 question for the general exposed function, I assume that it's not a blocker for Okta specific use case so approving to unblock.

packages/adapter-components/src/deployment/deployment.ts
shir-reifenberg: I just updated the PR cause I realized I should handle objects better - 
1. for removed primitives and arrays - replace entire value with null (and don't recurse into arrays values)
2. for objects - recurse into and replace internal keys with null

packages/adapter-components/test/deployment/deployment.test.ts
```diff
@@ -268,3 +268,84 @@ describe('filterUndeployableValues', () => {
...
+    const change = toChange({ before, after }) as ModificationChange<InstanceElement>
+    const result = transformRemovedValuesToNull(change)
+    expect(result.data.before.value).toEqual(before.value)
+    expect(result.data.after.value).toEqual({ name: 'inst',
+      status: 'active',
+      nested1: {
+        field: [1, 2],
+        nested2: {
+          some: 'value',
+          another: { type: null },
```
omrilit, line 322: I am not sure if this behavior is the correct one.
We could decide also to have here:
`another: {}, ` or even remove the `another` key as a whole.
WDYT?

shir-reifenberg: At first I went with the `another: null` approach but it seems not to work with certain cases for Okta. Once we'll encounter another usecase I guess we can add a flag to control this behavior, WDYT ?

omrilit (approved):

PR #5483, opened 2024-02-20
SALTO-4022: Exclude by bundleId

Shulik95:
_Allow users to exclude elements according the their bundle_

---

_None_

---
_Release Notes_: 
_Netsuite Adapter_:
* excludeBundles added to config, All elements from listed bundles Ids will be excluded.

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/netsuite-adapter/src/adapter.ts
+++ b/packages/netsuite-adapter/src/adapter.ts
@@ -1,18 +1,18 @@
 /*
-*                      Copyright 2024 Salto Labs Ltd.
-*
-* Licensed under the Apache License, Version 2.0 (the "License");
-* you may not use this file except in compliance with
-* the License.  You may obtain a copy of the License at
-*
-*     http://www.apache.org/licenses/LICENSE-2.0
-*
-* Unless required by applicable law or agreed to in writing, software
-* distributed under the License is distributed on an "AS IS" BASIS,
-* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-* See the License for the specific language governing permissions and
-* limitations under the License.
-*/
+ *                      Copyright 2024 Salto Labs Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 import {
   FetchResult,
   AdapterOperations,
--- a/packages/netsuite-adapter/src/adapter.ts
+++ b/packages/netsuite-adapter/src/adapter.ts
@@ -344,7 +344,7 @@ export default class NetsuiteAdapter implements AdapterOperations {
     const fetchQueryWithBundles = andQuery(fetchQuery, netsuiteBundlesQuery)
     const timeZoneAndFormat = getTimeDateFormat(configRecords)
     const { changedObjectsQuery, serverTime } = await this.runSuiteAppOperations(
-      fetchQuery,
+      fetchQueryWithBundles,
       useChangesDetection,
       timeZoneAndFormat,
     )
@@ -376,7 +376,7 @@ export default class NetsuiteAdapter implements AdapterOperations {
         importFileCabinetContent(),
         this.client.getCustomObjects(getStandardTypesNames(), {
           updatedFetchQuery,
-          originFetchQuery: fetchQuery,
+          originFetchQuery: fetchQueryWithBundles,
         }),
       ])
       const bundlesCustomInfo = bundlesToInclude.map(bundle => ({
@@ -395,7 +395,7 @@ export default class NetsuiteAdapter implements AdapterOperations {
       const { elements: customRecords, largeTypesError: failedCustomRecords } = await getCustomRecords(
         this.client,
         customRecordTypes,
-        fetchQuery,
+        fetchQueryWithBundles,
         this.getElemIdFunc,
       )
       return {
@@ -414,7 +414,7 @@ export default class NetsuiteAdapter implements AdapterOperations {
       { elements: suiteQLTableElements, largeSuiteQLTables },
     ] = await Promise.all([
       getStandardAndCustomElements(),
-      getDataElements(this.client, fetchQuery, this.getElemIdFunc),
+      getDataElements(this.client, fetchQueryWithBundles, this.getElemIdFunc),
       getSuiteQLTableElements(this.userConfig, this.client, this.elementsSource, isPartial),
     ])
 
@@ -422,7 +422,7 @@ export default class NetsuiteAdapter implements AdapterOperations {
 
     failures.failedTypes.excludedTypes = failures.failedTypes.excludedTypes.concat(dataTypeError)
     const suiteAppConfigElements = this.client.isSuiteAppConfigured()
-      ? toConfigElements(configRecords, fetchQuery).concat(getConfigTypes())
+      ? toConfigElements(configRecords, fetchQueryWithBundles).concat(getConfigTypes())
       : []
 
     // we calculate deleted elements only in partial-fetch mode
@@ -431,7 +431,7 @@ export default class NetsuiteAdapter implements AdapterOperations {
         ? await getDeletedElements({
             client: this.client,
             elementsSource: this.elementsSource,
-            fetchQuery,
+            fetchQuery: fetchQueryWithBundles,
             serviceInstanceIds: instancesIds,
             requestedCustomRecordTypes: customRecordTypes,
             serviceCustomRecords: customRecords,
--- a/packages/netsuite-adapter/src/adapter.ts
+++ b/packages/netsuite-adapter/src/adapter.ts
@@ -130,7 +130,7 @@ import {
 } from './config/query'
 import { getConfigFromConfigChanges } from './config/suggestions'
 import { NetsuiteConfig, AdditionalDependencies, QueryParams, NetsuiteQueryParameters, ObjectID } from './config/types'
-import { BundleInfo, buildNetsuiteBundlesQuery } from './config/bundle_query'
+import { buildNetsuiteBundlesQuery } from './config/bundle_query'
 import { SuiteAppBundleType } from './types/bundle_type'
 import { BUNDLE_ID_TO_COMPONENTS } from './autogen/bundle_components/bundle_components'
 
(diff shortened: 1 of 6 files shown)


packages/netsuite-adapter/src/adapter.ts
```diff
@@ -326,20 +330,40 @@ export default class NetsuiteAdapter implements AdapterOperations {
...
     fetchQuery: NetsuiteQuery,
     progressReporter: ProgressReporter,
     useChangesDetection: boolean,
     isPartial: boolean,
   ): Promise<FetchByQueryReturnType> => {
     const configRecords = await this.client.getConfigRecords()
+    const installedBundles = await this.client.getInstalledBundles()
+    const [bundlesToExclude, bundlesToInclude] = this.getBundlesToExclude(installedBundles)
+    const bundlesToExcludeInfo = bundlesToExclude.map(bundle => ({ id: bundle.id.toString(), version: bundle.version }))
+    const supportedBundles = bundlesToInclude.filter(bundle => bundle.id in BUNDLE_ID_TO_COMPONENTS)
```
Shulik95, line 355: This is added to avoid creating bundle elements for unsupported bundles, currently we create elements for bundles even if we don't have their components which is weird IMO

packages/netsuite-adapter/src/config/bundle_query.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+  // version is unsupported, check all supported versions of this bundle
+  const versionComponentsUnion = new Set(
+    Object.values(BUNDLE_ID_TO_COMPONENTS[bundleId]).flatMap(versionComponents => Array.from(versionComponents))
+  )
+  return versionComponentsUnion.has(instanceIdOrType)
+}
+
+const isContainedInSomeBundle = (instanceIdOrType: string, bundlesInfo: BundleInfo[]): boolean =>
+  bundlesInfo.some(({ id, version }) =>
```
dantal4, line 39: For every instanceId, we're going to iterate over all of the bundles, looking for the instanceId in its components. That is not much effiecient, especially if versionId is undefined. It will be much more efficient to first get all the components of the excluded bundles into one unified `Set`, then look for instanceId in that set.

packages/netsuite-adapter/src/config/bundle_query.ts
```diff
@@ -0,0 +1,100 @@
...
+    isBundleComponent(instanceIdOrType, id, version))
+
+const buildTypesQuery = (bundlesInfo: BundleInfo[]): TypesQuery => ({
+  isTypeMatch: () => true,
+  areAllObjectsMatch: () => false,
+  isObjectMatch: ({ instanceId }) => !isContainedInSomeBundle(instanceId, bundlesInfo),
+})
+
+const buildFileCabinetQuery = (bundlesInfo: BundleInfo[]): FileCabinetQuery => {
+  const regexPatterns = bundlesInfo.map(bundleInfo => new RegExp(`Bundle ${bundleInfo.id}`))
```
dantal4, line 49: In bundle_ids filter we're also using this regex, right? Can we share them with a const?
Also, this will catch "Bundle 12345" for "Bundle 123" too, better adding `/` at the end (and start too?)

packages/netsuite-adapter/src/config/bundle_query.ts
```diff
@@ -0,0 +1,100 @@
...
+    isCustomRecordTypeMatch,
+    areAllCustomRecordsMatch,
+    isCustomRecordMatch,
+  } = buildCustomRecordQuery(bundlesInfo)
+
+  return {
+    isTypeMatch,
+    areAllObjectsMatch,
+    isObjectMatch: object => isObjectMatch(object)
+    || (object.type === CUSTOM_SEGMENT && isObjectMatch({
```
dantal4, line 89: I think that this OR is redundent- If a custom segment belongs to a bundle and it should be excluded, I think that we should exclude its related custom record type. I also believe that the custom record type will be in the bundle's components too.

dantal4 (commented):
Great work!

packages/netsuite-adapter/src/config/bundle_query.ts
```diff
@@ -0,0 +1,100 @@
...
+import { BUNDLE_ID_TO_COMPONENTS } from '../autogen/bundle_components/bundle_components'
+import { CUSTOM_RECORD_TYPE, CUSTOM_SEGMENT } from '../constants'
+import { addCustomRecordTypePrefix } from '../types'
+
+export type BundleInfo = {
+  id: string
+  version: string | undefined
+}
+
+const isBundleComponent = (instanceIdOrType: string, bundleId: string, bundleVersion: string | undefined): boolean => {
```
dantal4, line 27: The "OrType" is confusing IMO. I guess that it refers to a custom record type, but "instanceId" match for that too, because it's a scriptid and not an SDF type.

packages/netsuite-adapter/src/config/bundle_query.ts
```diff
@@ -0,0 +1,100 @@
...
+    isParentFolderMatch: () => true,
+    areSomeFilesMatch: () => true,
+  }
+}
+
+const buildCustomRecordQuery = (bundlesInfo: BundleInfo[]): CustomRecordsQuery => ({
+  isCustomRecordTypeMatch: typeName => !isContainedInSomeBundle(typeName, bundlesInfo),
+  areAllCustomRecordsMatch: () => false,
+  isCustomRecordMatch: ({ type, instanceId }) => !bundlesInfo.some(({ id, version }) =>
+    isBundleComponent(type, id, version) || isBundleComponent(instanceId, id, version)),
```
dantal4, line 62: Similar comment here- we can just check the instanceId, becausw we already checked the custom record type.

packages/netsuite-adapter/src/adapter.ts
```diff
@@ -418,22 +440,20 @@ export default class NetsuiteAdapter implements AdapterOperations {
...
+        fetchQuery,
+        serviceInstanceIds: instancesIds,
+        requestedCustomRecordTypes: customRecordTypes,
+        serviceCustomRecords: customRecords,
+        requestedDataTypes,
+        serviceDataElements: dataElements.filter(isInstanceElement),
+      })
+      : {}
+
+    const serverTimeElements = serverTime !== undefined ? await getOrCreateServerTimeElements(serverTime, this.elementsSource, isPartial) : []
```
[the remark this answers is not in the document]
Shulik95: Nope, has some weird formatting issue.

packages/netsuite-adapter/src/adapter.ts
Shulik95: Currently we log missing bundles in `filters/bundle_ids.ts`, If we're ok with creating elements for unsupported bundles we can skip this here and leave the log in the filter. If we prefer not to have these elements we should filter and move the log here.

Shulik95: Spoke f2f with @omrilit and we should create bundle elements for bundles even if we dont have their components. We log the missing bundles in the filter so I think were good.

packages/netsuite-adapter/src/adapter.ts
```diff
@@ -332,14 +333,23 @@ export default class NetsuiteAdapter implements AdapterOperations {
...
-    const configRecords = await this.client.getConfigRecords()
+    const [configRecords, installedBundles] = await Promise.all([
+      this.client.getConfigRecords(),
+      this.client.getInstalledBundles(),
+    ])
+    const { query: netsuiteBundlesQuery, bundlesToInclude } = buildNetsuiteBundlesQuery(
+      installedBundles,
+      this.userConfig.excludeBundles ?? [],
+    )
+    const fetchQueryWithBundles = andQuery(fetchQuery, netsuiteBundlesQuery)
```
dantal4, line 344: there are more places here that use `fetchQuery`, and they should be replaced with `fetchQueryWithBundles` - meaning that `fetchQuery` shouldn't be used at all after this line

packages/netsuite-adapter/src/config/bundle_query.ts
```diff
@@ -0,0 +1,109 @@
...
+import { BUNDLE } from '../constants'
+
+export type BundlesQueryAndSupportedBundles = {
+  query: NetsuiteQuery
+  bundlesToInclude: SuiteAppBundleType[]
+}
+
+const getBundleComponents = (bundleId: string, bundleVersion: string | undefined): Set<string> => {
+  if (bundleVersion) {
+    return BUNDLE_ID_TO_COMPONENTS[bundleId][bundleVersion]
```
dantal4, line 33: don't you need to check if `bundleVersion` in `BUNDLE_ID_TO_COMPONENTS[bundleId]` ?

packages/netsuite-adapter/src/config/bundle_query.ts
```diff
@@ -0,0 +1,109 @@
...
+import { getGroupItemFromRegex } from '../client/utils'
+import { BUNDLE } from '../constants'
+
+export type BundlesQueryAndSupportedBundles = {
+  query: NetsuiteQuery
+  bundlesToInclude: SuiteAppBundleType[]
+}
+
+const getBundleComponents = (bundleId: string, bundleVersion: string | undefined): Set<string> => {
+  if (bundleVersion) {
```
dantal4, line 32: can `bundleVersion` be undefined? in bundle_ids we're not checking it... https://github.com/salto-io/salto/blob/1c5929604b6021e8098b82c30c460f6d67bee4e5/packages/netsuite-adapter/src/filters/bundle_ids.ts#L40

packages/netsuite-adapter/src/config/bundle_query.ts
```diff
@@ -0,0 +1,109 @@
...
+  bundlesToExclude: string[],
+): [SuiteAppBundleType[], SuiteAppBundleType[]] => {
+  if (bundlesToExclude.length === 0) {
+    return [[], installedBundles]
+  }
+  if (bundlesToExclude.includes(INCLUDE_ALL)) {
+    return [installedBundles, []]
+  }
+  const bundleMatchers = bundlesToExclude.map(
+    matcher => new RegExp(matcher === INCLUDE_ALL ? ALL_TYPES_REGEX : matcher),
```
dantal4, line 77: this is redundant because of the condition above:
```suggestion
    matcher => new RegExp(matcher),
```

dantal4 (commented):
Looks great! I added some more comments 👍

packages/netsuite-adapter/src/config/bundle_query.ts
Shulik95: It can be undefined. I've added a check for this in `bundle_ids`, will be handled the same way as a missing version. we'll log it and return the union of all the other versions we have.

packages/netsuite-adapter/src/filters/bundle_ids.ts
```diff
@@ -34,18 +34,21 @@ const log = logger(module)
...
-  if (serviceIdsInBundle) {
-    return serviceIdsInBundle
+export const getServiceIdsOfVersion = (
+  bundleVersions: Record<string, Set<string>>,
+  bundleId: string,
+  bundleVersion: string | undefined,
+): Set<string> => {
+  if (bundleVersion === undefined || !(bundleVersion in bundleVersions)) {
+    log.debug(
+      `Version ${`${bundleVersion} ` ?? ''} of bundle %s is missing or not supported in the record, use a union of all existing versions`,
```
dantal4, line 46: \`${bundleVersion} \` will never be undefined.....

packages/netsuite-adapter/src/filters/bundle_ids.ts
```diff
@@ -34,18 +34,21 @@ const log = logger(module)
...
 const BUNDLE = 'bundle'
-const bundleIdRegex = RegExp(`Bundle (?<${BUNDLE}>\\d+)`, 'g')
+export const bundleIdRegex = RegExp(`Bundle (?<${BUNDLE}>\\d+)`, 'g')
 
-const getServiceIdsOfVersion = (bundleVersions: Record<string, Set<string>>, bundle: InstanceElement): Set<string> => {
-  const serviceIdsInBundle = bundleVersions[bundle.value.version]
-  if (serviceIdsInBundle) {
-    return serviceIdsInBundle
+export const getServiceIdsOfVersion = (
+  bundleVersions: Record<string, Set<string>>,
```
dantal4, line 40: nit - since this function gets the `bundleId`, it can create `bundleVersions` in it instead of passing from outside.

dantal4 (approved):
💯

PR #5485, opened 2024-02-20
Limit SuiteQL tables query to 100K records

dantal4:
Avoid querying SuiteQL tables that have more that 100K records in them, and exclude those tables' instances.

---

---
_Release Notes_: 
Netsuite Adapter:
- Limit SuiteQL tables query to 100K records

---
_User Notifications_: 
None


packages/adapter-components/src/config/shared.ts
dantal4, line 1: change because of prettier

omrilit (approved):
💯

PR #5489, opened 2024-02-20
Add config suggestion on skipping SuiteQL tables

dantal4:
In case that some SuiteQL tables were excluded, we want to return a config suggestion that excludes the table from the next fetches.

---

---
_Release Notes_: 
Netsuite Adapter:
- Add config suggestion on skipping SuiteQL tables

---
_User Notifications_: 
None


packages/netsuite-adapter/src/config/suggestions.ts
```diff
@@ -42,6 +42,10 @@ export const toLargeTypesExcludedMessage = (updatedLargeTypes: string[]): string
   `The following types were excluded from the fetch as the elements of that type were too numerous: ${updatedLargeTypes.join(', ')}.` +
   " To include them, increase the types elements' size limitations and remove their exclusion rules."
 
+export const toLargeSuiteQLTablesExcludedMessage = (largeSuiteQLTables: string[]): string =>
```
omrilit, line 45: I think that the user will not understand what is their motivation to include them back. Should we consider to explain that it's used for ASVs?

omrilit (approved):
Great work!

PR #5490, opened 2024-02-20
SALTO-5450_Add_Aliases_to_Jira_Statuses

omrifarkash:
Add Aliases to Jira Statuses

---

---
_Release Notes_: 
_Jira_adapter_:
Add aliases to Jira Statuses based on the instance name

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/add_alias.ts
+++ b/packages/jira-adapter/src/filters/add_alias.ts
@@ -216,6 +216,14 @@ const aliasMap: Record<string, AliasData> = {
     ],
     separator: ':',
   },
+  Status: {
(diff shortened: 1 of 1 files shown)


packages/jira-adapter/src/filters/add_alias.ts
```diff
@@ -216,6 +216,14 @@ const aliasMap: Record<string, AliasData> = {
     ],
     separator: ':',
   },
+  Status: {
```
IdoZakk, line 219: Please use the constants

IdoZakk (approved):
👍 
See the comment

PR #5492, opened 2024-02-21
Run SuiteQL queries concurrently

dantal4:
Use the first SuiteQL result's "last page" link to determine how many pages are there for the SuiteQL query, and then query all pages using `Promise.all`.

NOTE - there's still the bottleneck limiter that run max of 4/5 requests at the same time.

---

---
_Release Notes_: 
Netsuite Adapter:
- Run SuiteQL queries concurrently

---
_User Notifications_: 
None


omrilit (approved):
Looks great!

PR #5493, opened 2024-02-21
SALTO-5461 - Salesforce: Propagate warnings from successful deploy results

tomsellek:
See ticket for an example of such a message on a successful deploy of a component

---

I made a couple of fixes to the E2E utils while I was at it.

---
_Release Notes_: 
Salesforce: info/warning messages from successfully deployed components will now be shown

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/metadata_deploy.ts
+++ b/packages/salesforce-adapter/src/metadata_deploy.ts
@@ -295,6 +295,14 @@ const processDeployResponse = (
     return rawElemId
   }
 
+  const problemTypeToSeverity = (messageType: string): SeverityLevel => {
+    if (['Info', 'Warning', 'Error'].includes(messageType)) {
+      return messageType as SeverityLevel
+    }
+    log.warn('unknown messageType %s', messageType)
+    return 'Warning'
+  }
+
   const allFailureMessages = makeArray(result.details).flatMap((detail) =>
     makeArray(detail.componentFailures),
   )
@@ -324,7 +332,7 @@ const processDeployResponse = (
     .map((message) => ({
       elemID: getElemIdForDeployError(message),
       message: message.problem,
-      severity: message.problemType as SeverityLevel, // Salesforce problem types happen to match our severity levels
+      severity: problemTypeToSeverity(message.problemType),
     }))
 
   const componentErrors = failedComponentErrors.concat(
--- a/packages/salesforce-adapter/src/metadata_deploy.ts
+++ b/packages/salesforce-adapter/src/metadata_deploy.ts
@@ -299,7 +299,11 @@ const processDeployResponse = (
     makeArray(detail.componentFailures),
   )
 
-  const componentErrors = allFailureMessages
+  const allSuccessMessages = makeArray(result.details).flatMap((detail) =>
+    makeArray(detail.componentSuccesses),
+  )
+
+  const failedComponentErrors = allFailureMessages
     .filter((failure) => !isUnFoundDelete(failure, deletionsPackageName))
     .map(getUserFriendlyDeployMessage)
     .map((failure) => ({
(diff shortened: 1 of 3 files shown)


packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -308,6 +314,23 @@ const processDeployResponse = (
...
+    .filter((message) => message.problem)
+    .filter((message) => !isUnFoundDelete(message, deletionsPackageName))
+    .filter((message) => {
+      log.info('fullName: %s', message.fullName)
+      return true
+    })
+    .map((message) => ({
+      elemID: getElemIdForDeployError(message),
+      message: message.problem,
+      severity: message.problemType as SeverityLevel, // Salesforce problem types happen to match our severity levels
```
amitgoldarb, line 327: This should probably be explicit. It's safer if we find an unexpected problemType or if this changes over time.

tomsellek: Ack

packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -295,11 +295,25 @@ const processDeployResponse = (
     return rawElemId
   }
 
+  const problemTypeToSeverity = (messageType: string): SeverityLevel => {
```
tamtamirr, line 298: Why do we need this method?
Isn't the code in L325 is only relevant for warnings `Warnings`?

packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -308,6 +322,23 @@ const processDeployResponse = (
...
+    .filter((message) => message.problem)
+    .filter((message) => !isUnFoundDelete(message, deletionsPackageName))
+    .filter((message) => {
+      log.info('fullName: %s', message.fullName)
+      return true
+    })
+    .map((message) => ({
+      elemID: getElemIdForDeployError(message),
+      message: message.problem,
+      severity: problemTypeToSeverity(message.problemType),
```
tamtamirr, line 335: Shouldn't this be a Warning?

packages/salesforce-adapter/test/adapter.crud.test.ts
```diff
@@ -1164,6 +1162,47 @@ describe('SalesforceAdapter CRUD', () => {
           expect(result.appliedChanges).toBeEmpty()
         })
       })
+      describe('when the successful element has warnings', () => {
```
tamtamirr, line 1165: Warnings should have severity `Warning` and not Info.
Do we handle Info as-well? Does the code handle it? If so, we need to change the terminology in this PR. (As it explicitly states Warning and no Info)

tamtamirr (commented):
Good work!
Added some comments. Mainly confused on the scope we tackle here. Do we also handle Info?

Were you able to create a deployment that returns with warnings?
If so, let's test how the code behaves when we use `ignoreWarnings = false` in the deploy config. Will we have both warnings and errors? Should we change the severity of the Warning to Error in that use-case?

packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -308,6 +322,23 @@ const processDeployResponse = (
       severity: 'Error' as SeverityLevel,
     }))
 
+  const successfulComponentWarnings = allSuccessMessages
+    .filter((message) => message.problem)
+    .filter((message) => !isUnFoundDelete(message, deletionsPackageName))
+    .filter((message) => {
+      log.info('fullName: %s', message.fullName)
```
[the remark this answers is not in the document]
tomsellek: Sorry, leftover debugging stuff

packages/salesforce-adapter/src/metadata_deploy.ts
tomsellek: We actually saw `Info` level too (which the Salesforce docs don't mention)

packages/salesforce-adapter/src/metadata_deploy.ts
tomsellek: Not if it's `Info`, I think

packages/salesforce-adapter/test/adapter.crud.test.ts
tomsellek: Ack

packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -308,6 +320,19 @@ const processDeployResponse = (
       severity: 'Error' as SeverityLevel,
     }))
 
+  const successfulComponentProblems = allSuccessMessages
```
tamtamirr, line 323: I think we should log these problems so we can monitor this easily

tamtamirr (approved):
LGTM! Added a small comment to add visibility.
Approving to save cycles

PR #5497, opened 2024-02-22
SALTO-5457: Add Remaining Types In Fetch

idoamit8:
In this PR I added support to add all the remaining types while fetch. 

---

More context in [SALTO-5467](https://salto-io.atlassian.net/browse/SALTO-5467)
https://salto-io.atlassian.net/browse/SALTO-5467

---
_Release Notes_: 
_Jira Adpater:_
always have at least an empty type for any deployable instance

---
_User Notifications_: 
_Jira Adapter:_
new empty objectTypes will be added

### DIFF
--- a/packages/jira-adapter/src/adapter.ts
+++ b/packages/jira-adapter/src/adapter.ts
@@ -197,7 +197,7 @@ import { hasSoftwareProject } from './utils'
 import { getWorkspaceId } from './workspace_id'
 import { JSM_ASSETS_DUCKTYPE_SUPPORTED_TYPES } from './config/api_config'
 
-const { getAllElements } = elementUtils.ducktype
+const { getAllElements, addRemainingTypes } = elementUtils.ducktype
 const { findDataField } = elementUtils
 const { computeGetArgs } = fetchUtils.resource
 const { generateTypes, getAllInstances, loadSwagger, addDeploymentAnnotations } = elementUtils.swagger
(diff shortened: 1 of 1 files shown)


packages/jira-adapter/src/adapter.ts
```diff
@@ -720,6 +720,22 @@ export default class JiraAdapter implements AdapterOperations {
       ...jsmElements.elements,
       ...jsmAssetsElements.elements,
     ]
+    const typesConfig = {
+      ...this.userConfig.apiDefinitions.types,
+      ...this.userConfig.jsmApiDefinitions?.types,
+      ...this.userConfig.scriptRunnerApiDefinitions?.types,
```
shir-reifenberg, line 726: can `jsmApiDefinitions` or `scriptRunnerApiDefinitions` be undefined (asking cause your'e using `?`) ? if so, we might not get the remaining types we wanted :)

packages/jira-adapter/src/adapter.ts
```diff
@@ -720,6 +720,22 @@ export default class JiraAdapter implements AdapterOperations {
...
+    }
+    
+    // Remaining types should be added once to avoid overlaps between the generated elements,
+    // so we add them once after all elements are generated
+    addRemainingTypes({
+      adapterName: JIRA,
+      elements,
+      typesConfig,
+      supportedTypes: _.merge(supportedTypes, this.userConfig.jsmApiDefinitions?.supportedTypes, this.userConfig.scriptRunnerApiDefinitions?.supportedTypes),
+      typeDefaultConfig: this.userConfig.apiDefinitions.typeDefaults,
```
shir-reifenberg, line 736: should we merge all types default here as well ?

shir-reifenberg (approved):
Loos great 👑
Please add a UT, and there's no need in the user notifications cause types are hidden :) Also, IMO worth updating the release notes with exactly what bug is solves

PR #5504, opened 2024-02-23
SALTO-2392: turn resolveAccountSpecificValues off by default

omrilit:
**This PR is not rebased on the newest main but on the bumped version**

---

_Additional context for reviewer_

---
_Release Notes_: 
SALTO-2392: disable resolveAccountSpecificValues by default

---
_User Notifications_: 
ACCOUNT_SPECIFIC_VALUE will not be resolved in workflows & data elements, on fetch & deploy

PR #5506, opened 2024-02-24
SALTO-2392: change the default behavior of resolveAccountSpecificValues so if undefined do not fetch SuiteQL tables

omrilit:
This changes the default behavior of fetching the SuiteQL tables for ASV resolving - we will fetch the tables only when `fetch.resolveAccountSpecificValues` is true.
We still transform data elements to be in the new format in data_account_sepcific_values filter.

---

This PR supposes to replace the intended behavior in https://github.com/salto-io/salto/pull/5504 as it will still allow us to resolve ASVs in a smarter way but without persisting the SuiteQL tables by default, unless `fetch.resolveAccountSpecificValues` is set to true.

---
_Release Notes_: 
Netsuite Adapter:
- change the default behavior of fetching the SuiteQL tables for ASV resolving

---
_User Notifications_: 


omrilit:
Closing as it was deployed to production as a hotfix.
The same content should get merged in https://github.com/salto-io/salto/pull/5505

PR #5507, opened 2024-02-25
SALTO-5161: [Zendesk] Move caching for users request in Zendesk to be in the Adapter

DanielBarros1:
The getUsers query currently runs on a global scope and caches the user data.
This means having more than one workspace on a single process can cause data leakage and unexpected behaviour, which happens when running SAAS locally.

In this PR I move the scope of the query from global to the zendesk adapter, which prevents the leak.
This is essentially identical to the pattern added [here in Okta](https://github.com/salto-io/salto/pull/5261/files#diff-fdd8ade86a6a37213a71a3465db95e5e94e20ec7bc026beab906303cc6630a1aR127).

---

I ran Saas locally with my changes, added logs and made sure that the query executes correctly and that fetch and deployment works properly.

---

_Release Notes_: 
Zendesk:
Fix leakage of users when running Saas locally with multiple environments

---

_User Notifications_: 
None


edenhassid (commented):
looks good! 🏆  added one small question

packages/zendesk-adapter/src/adapter.ts
```diff
@@ -880,6 +899,7 @@ export default class ZendeskAdapter implements AdapterOperations {
       changeValidator: createChangeValidator({
         client: this.client,
         config: this.userConfig,
+        usersPromise: this.getUsersPromise(),
```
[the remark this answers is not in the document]
DanielBarros1: We talked about awaiting the promise and passing in the results instead of the promise itself, which would mean making this function async which i didn't love :)

packages/zendesk-adapter/test/filters/custom_objects/custom_object_fields.test.ts
```diff
@@ -35,16 +35,22 @@ import filterCreator from '../../../src/filters/custom_objects/custom_object_fie
 import { createFilterCreatorParams } from '../../utils'
 import { DEFAULT_CONFIG, DEPLOY_CONFIG } from '../../../src/config'
 
-const USER = { id: 111, email: 'User' }
-const DEFAULT_USER = { id: 222, email: 'DefaultUser' }
-jest.mock('../../../src/user_utils', () => ({
-  ...jest.requireActual<{}>('../../../src/user_utils'),
+const USER = { id: 111, email: 'User', role: 'admin', custom_role_id: 123, name: 'c', locale: 'en-US' }
+const DEFAULT_USER = { id: 222, email: 'DefaultUser', role: 'admin', custom_role_id: 123, name: 'c', locale: 'en-US' }
```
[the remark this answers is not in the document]
DanielBarros1: Pretty sure it's a copy/paste thing, let me check if i can revert this change lol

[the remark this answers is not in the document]
DanielBarros1: It's not a copy paste thing, it's because we apply a type to the users returned and that type has all these additional fields. :)

packages/zendesk-adapter/src/users/user_utils.ts
```diff
@@ -300,20 +281,14 @@ export const getUserFallbackValue = async (
   return defaultMissingUserFallback
 }
 
-const getIdByEmailFunc = (): ((
-  paginator: clientUtils.Paginator,
-  runQuery: boolean | undefined,
-) => Promise<Record<string, string>>) => {
+const getIdByEmailFunc = (): ((getUsersPromise: Promise<GetUsersResponse>) => Promise<Record<string, string>>) => {
   let idToEmail: Record<string, string>
```
edenhassid, line 285: i think you don't need this, its the same issue from get users with the cache. I think its ok to just calculate the map each time.

packages/zendesk-adapter/src/users/user_utils.ts
```diff
@@ -326,20 +301,14 @@ const getIdByEmailFunc = (): ((
 
 export const getIdByEmail = getIdByEmailFunc()
 
-const getIdByNameFunc = (): ((
-  paginator: clientUtils.Paginator,
-  runQuery: boolean | undefined,
-) => Promise<Record<string, string>>) => {
+const getIdByNameFunc = (): ((getUsersPromise: Promise<GetUsersResponse>) => Promise<Record<string, string>>) => {
   let idToName: Record<string, string>
```
edenhassid, line 305: same here

packages/zendesk-adapter/test/user_utils.test.ts
```diff
@@ -174,7 +148,8 @@ describe('userUtils', () => {
           },
         ]
       })
-      const idByEmail = await userUtils.getIdByEmail(mockPaginator, true)
+      const getUsersPromise = userUtils.getUsers(mockPaginator, true)
```
edenhassid, line 151: nit: maybe you can run one of these at the beginning, WDYT?

edenhassid (commented):
looks good! 🤯 🔥 💯 added some comments :)

packages/zendesk-adapter/test/user_utils.test.ts
DanielBarros1: The paginator mock changes for every test so i think it has to be here :P

PR #5508, opened 2024-02-25
SALTO-5436: [Okta] Nest GroupPush and GroupPushRule under the parent app path

shir-reifenberg:
Nest group push types under the parent app

---

This is in a dedicated filter and not using "nestStandaloneFields" as group push types are not standalone fields of apps cause the types are in different configs.
This is a temp filter that will be removed once we migrate Okta to the new infra

This change is done without any migration to existing files.

---
_Release Notes_: 

_Okta_adapter_:
- On the next fetch, GroupPush and GroupPushRule types will now be nested under the parent Application type in the element tree

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/test/filters/group_push_path.test.ts
+++ b/packages/okta-adapter/test/filters/group_push_path.test.ts
@@ -1,0 +1,104 @@
+/*
+ *                      Copyright 2024 Salto Labs Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


packages/okta-adapter/test/filters/group_push_path.test.ts
```diff
@@ -0,0 +1,104 @@
...
+      name: 'test rule',
+      status: 'ACTIVE',
+      searchExpression: 'sales',
+      searchExpressionType: 'ENDS_WITH',
+    },
+    undefined,
+    { [CORE_ANNOTATIONS.PARENT]: [new ReferenceExpression(appWithGroupPush.elemID, appWithGroupPush)] },
+  )
+
+  describe('fetch', () => {
```
idoamit8, line 77: Can you please add a test when the parent is not as expected (Undefined or not resolved reference expression)

idoamit8 (commented):
Great work! 💯 
Very clean and organized code. I really loved the documentation of the filter :) 
Add one small comment about test coverage. 

One small question
I see that we have `group_push` filter that fetches those types.  Why don't we include the path handling inside of it?

PR #5514, opened 2024-02-26
SALTO-5460: [Zendesk] Fix dynamic_content placeholder regex catching non-DC strings

DanielBarros1:
The current regex for catching dynamic content placeholders allows for some pretty whacky stuff;
For example, having `{{ blip bloop blep dc.content_item }}` will catch `dc.content_item` as a dynamic content placeholder, even though Zendesk wouldn't allow it. 
Another example is that this issue was caught when a customer tried to note his email, which had the string `dc.stuttgart` in it, and it got replaced by a dynamic content ref - without any brackets at all!

In this PR, i've changed the regex so that it forces the content to be inside brackets, and furthermore forces the brackets to be "close" to the placeholder (so if you have content between the brackets and the placeholder it won't catch it anymroe).

---

Ironically, we already had a test for this in place, but it was testing the wrong thing (i.e the test was *expecting* a dynamic content ref where there shouldn't be one

---
_Release Notes_: 
Zendesk:
Fix bug where non-"Dynamic content item" strings were being parsed as dynamic content and creating broken references

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/filters/handle_template_expressions.ts
+++ b/packages/zendesk-adapter/src/filters/handle_template_expressions.ts
@@ -327,11 +327,15 @@ const formulaToTemplate = ({
   // we continuously split the expression to find all kinds of potential references
   return extractTemplate(seekAndMarkPotentialReferences(formula), potentialRegexes, expression => {
     const zendeskReference = expression.match(potentialReferenceTypeRegex)
+    if (expression.includes('dc')) {
+      console.log(formula, expression, zendeskReference)
+    }
     if (zendeskReference) {
       return handleZendeskReference(expression, zendeskReference)
     }
     const dynamicContentReference = expression.match(DYNAMIC_CONTENT_REGEX)
     if (dynamicContentReference) {
+      console.log('do i get here? ', dynamicContentReference)
       return handleDynamicContentReference(expression, dynamicContentReference)
     }
     if (extractReferencesFromFreeText) {
--- a/packages/zendesk-adapter/src/filters/handle_template_expressions.ts
+++ b/packages/zendesk-adapter/src/filters/handle_template_expressions.ts
@@ -61,10 +61,8 @@ const BRACKETS = [
   ['{%', '%}'],
 ]
 const REFERENCE_MARKER_REGEX = /\$\{({{.+?}})\}/
-// const DYNAMIC_CONTENT_NAME_REGEX = /(dc\.[\w-]+)/g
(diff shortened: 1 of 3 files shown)


packages/zendesk-adapter/test/filters/handle_template_expressions.test.ts
```diff
@@ -568,9 +568,7 @@ describe('handle templates filter', () => {
           parts: [
             `{{some other irrelevancies-${TICKET_TICKET_FIELD}_`,
             new ReferenceExpression(placeholder1.elemID, placeholder1),
-            ' | something irrelevant | dynamic content now: ',
-            new ReferenceExpression(dynamicContentRecord.elemID, dynamicContentRecord),
-            ' | and done}}',
+            ' | something irrelevant | dynamic content now: dc.dynamic_content_test | and done}}',
```
adfineberg, line 571: can you call the `dc.dynamic_content_test` something else? as this is on purpose not supposed to catch, right?
Also, can you add a test for the email case - to show that we don't catch it. I think it's a reasonable test to lean on in the future, no?

packages/zendesk-adapter/src/filters/handle_template_expressions.ts
```diff
@@ -60,8 +60,9 @@ const BRACKETS = [
   ['{{', '}}'],
   ['{%', '%}'],
 ]
-const REFERENCE_MARKER_REGEX = /\$\{(.+?)}/
+const REFERENCE_MARKER_REGEX = /\$\{({{.+?}})\}/
 const DYNAMIC_CONTENT_REGEX = /(dc\.[\w-]+)/g
+const DYNAMIC_CONTENT_REGEX_WITH_BRACKETS = /({{dc\.[\w-]+}})/g
```
adfineberg, line 65: some duplication here, I guess it's not critical to merge them. Up to you :)

adfineberg (commented):
Are we 100% sure that dynamic content can only be wrapped by `{{`? Just double (or triple?) checking because someone specifically wrote tests for it

DanielBarros1:
> Are we 100% sure that dynamic content can only be wrapped by `{{`? Just double (or triple?) checking because someone specifically wrote tests for it

100% fair question @adfineberg , i had the same thought when i saw the code - no idea why it was coded this way, the zendesk docs (as well as actual trial and error inside the zendesk ui) say that it can only ever appear {{dc.like_this}}.

adfineberg (approved):
Looks good! 🔥

PR #5521, opened 2024-02-29
SALTO-5404: Add format on save for VS Code

shoshana-houri:
Header is self explanatory

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


adfineberg (commented):
Not sure we need the react cases in this repo. Also, WDYT about reverting the addition of prettier in the `yarn build` flow? It created some bugs @IdoZakk

shoshana-houri:
@adfineberg You're right, I'll remove it. And sure :)

adfineberg (approved):

PR #5522, opened 2024-02-29
SALTO-5429: [MA] Switch elem id name mapping to functions that can be overwritten

shoshana-houri:
Currently, the only name mappings defined are `lowercase` and `uppercase`, without a simple structured way to add custom mappings to meet adapter requirements. For example, in Jira, there's a **filter** designed to omit the internal id from the name, as mentioned in [this comment](https://github.com/salto-io/salto/pull/5409#discussion_r1481069530).

In this PR, we suggest a structurally typed approach to adding custom mapping options, where the default ones (`lowercase` and `uppercase`) remain common and do not require a special definition.

It works by defining the possible options as an enum, since this mapping is a user-facing config, and providing a corresponding map from these enums to their mapping functions.

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/config/ducktype.ts
+++ b/packages/adapter-components/src/config/ducktype.ts
@@ -33,7 +33,7 @@ import {
 } from './transformation'
 
 import { validateRequestConfig } from './request'
-import { UserFetchConfig } from '../definitions/user'
+import { UserFetchConfig, UserFetchConfigOptions } from '../definitions/user'
 import { DefaultWithCustomizations } from '../definitions'
 
 const { isDefined } = lowerDashValues
@@ -107,9 +107,9 @@ export const validateApiDefinitionConfig = (
  * Verify that all fetch types exist in the endpoint definitions.
  * Note: This validation is only relevant for ducktype adapters.
  */
-export const validateFetchConfig = (
+export const validateFetchConfig = <Options extends UserFetchConfigOptions>(
   fetchConfigPath: string,
-  userFetchConfig: UserFetchConfig,
+  userFetchConfig: UserFetchConfig<Options>,
   adapterApiConfig: AdapterApiConfig,
 ): void => {
   validateSupportedTypes(fetchConfigPath, userFetchConfig, Object.keys(adapterApiConfig.supportedTypes))
--- a/packages/adapter-components/src/definitions/user/fetch_config.ts
+++ b/packages/adapter-components/src/definitions/user/fetch_config.ts
@@ -26,6 +26,8 @@ import {
 } from '@salto-io/adapter-api'
 import { createMatchingObjectType } from '@salto-io/adapter-utils'
 import { ElemIDDefinition, FieldIDPart } from '../system/fetch/element'
+import { NameMappingOptions, ResolveCustomNameMappingOptionsType } from '../system'
+import { FetchApiDefinitionsOptions } from '../system/fetch'
 
 export type DefaultFetchCriteria = {
   name?: string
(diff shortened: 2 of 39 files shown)


packages/adapter-components/src/definitions/user/fetch_config.ts
```diff
@@ -95,7 +105,8 @@ export const createUserFetchConfigType = ({
     fetchEntryType.fields.criteria = new Field(fetchEntryType, 'criteria', fetchCriteriaType)
   }
 
-  const elemIDPartType = createMatchingObjectType<Omit<FieldIDPart, 'condition' | 'custom'>>({
+  // TODO: find if there's a way to specify the generic type here of TCustomNameMappingOptions
+  const elemIDPartType = createMatchingObjectType<Omit<FieldIDPart<NameMappingOptions>, 'condition' | 'custom'>>({
```
shoshana-houri, line 109: Left this comment so we can discuss it during the review -
For some reason, when I specify the generic type as:

```suggestion
  // TODO: find if there's a way to specify the generic type here of TCustomNameMappingOptions
  const elemIDPartType = createMatchingObjectType<Omit<FieldIDPart<TCUstomNameMappingOptions>, 'condition' | 'custom'>>({
```

The resulting type causes an unexpected issue with `fieldName` under `fields`: TypeScript expects its `_annotations: { required: ?}` field to simultaneously satisfy contradictory conditions (both falsy and truthy) 😐. I've attempted to resolve this without success.
It seems like TypeScript is not accurately recognizing which elements should be considered optional, leading to confusion in determining the appropriate `SaltoAnnotationsForField` type for this field.

netama: weird, gonna need to look more later - either way since we currently don't enforce the enum values it's not critical :)

netama (commented):
very cool :) 
added some comments, lmk if you want to discuss IRL
the extra `Resolve` types are interesting, I'm wondering if there's a way to reduce their number somehow as well (but not critical for the first iteration)

packages/adapter-components/src/fetch/element/element.ts
```diff
@@ -39,15 +40,16 @@ export type ElementGenerator = {
...
   defQuery,
   predefinedTypes,
+  customNameMapping,
   getElemIdFunc,
 }: {
   adapterName: string
-  fetchQuery: ElementQuery
-  defQuery: ElementAndResourceDefFinder
+  defQuery: ElementAndResourceDefFinder<Options>
+  customNameMapping: NameMappingFunctionMap<ResolveCustomNameMappingOptionsType<Options>>
```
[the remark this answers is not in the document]
shoshana-houri: I initially made it optional but then realized that type enforcement is more effective when it's not optional. If it's not optional, we have to pass an empty record, which is slightly inconvenient but manageable. The advantage is significant: if I define custom options, TypeScript will require me to specify them in the custom mapping as well. Conversely, if it's optional, not specifying it at all won't enforce the specification of the custom options, which is prone to errors.

I haven't found a straightforward way to enforce this parameter in TS such that if the generic type is defined, this parameter must also be defined. I attempted using an interface with call signatures to simulate overloads, as in the following example for the base function:

```
interface GetNameMapping {
  (options: {
    name: string;
    mappingOption?: NameMappingOptions;
    nameMapping?: never;
  }): string;

  <TCustomNameMappingOptions extends string>(options: {
    name: string;
    mappingOption: TCustomNameMappingOptions;
    nameMapping: Required<NameMapping<TCustomNameMappingOptions>>;
  }): string;
}
```

However, this didn't work well and would require similar definitions throughout the hierarchy. I'd be happy to hear suggestions if you have any!

packages/adapter-components/src/config/ducktype.ts
```diff
@@ -107,9 +107,9 @@ export const validateApiDefinitionConfig = (
  * Verify that all fetch types exist in the endpoint definitions.
  * Note: This validation is only relevant for ducktype adapters.
  */
-export const validateFetchConfig = (
+export const validateFetchConfig = <Options extends UserFetchConfigOptions>(
```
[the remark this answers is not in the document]
shoshana-houri: Not sure I understand it completely, all the values under `Options` are optional with default. Do you mean:
```suggestion
export const validateFetchConfig = <Options extends UserFetchConfigOptions = {}>(
```
It seems like it's working ok atm without explicitly providing this default (for example in the UTs of this file), so I think I'm missing something here. I'd be happy to understand it better f2f :)

packages/adapter-components/src/fetch/element/id_utils.ts
```diff
@@ -69,27 +74,52 @@ export const serviceIDKeyCreator =
...
+}): string => {
+  if (!nameMapping) {
+    return String(name)
   }
+
+  const nameMappingFunctions = {
+    ...DEFAULT_NAME_MAPPING_FUNCTIONS,
+    ...customNameMappingFunctions,
+  } as Record<NameMappingOptions | TCustomNameMappingOptions, NameMappingFunction>
+  // Unfortunately, TypeScript doesn't recognize that a union of NameMappingOptions and
```
netama, line 100: please add comments above the line they refer to and not below :)

packages/adapter-components/src/fetch/element/id_utils.ts
```diff
@@ -69,27 +74,52 @@ export const serviceIDKeyCreator =
...
-      return name.toUpperCase()
-    default:
-      return name
+/**
+ * Returns the name part of an elemId based on the provided name mapping options, which can be
+ * either a built-in name mapping option or a custom name mapping option.
+ * If the name mapping option or the relevant custom mapping is not provided,
+ * the function returns the name as is, after converting it to a string.
+ */
+export const getNameMapping = <TCustomNameMappingOptions extends string>({
```
netama, line 83: maybe default to `never`?

packages/adapter-components/src/definitions/user/user_config.ts
```diff
@@ -19,24 +19,28 @@ import { createClientConfigType, ClientBaseConfig, ClientRateLimitConfig } from
 import { UserFetchConfig, createUserFetchConfigType } from './fetch_config'
 import { UserDeployConfig, createChangeValidatorConfigType, createUserDeployConfigType } from './deploy_config'
 
-export type UserConfig = {
+export type UserConfig<TCustomNameMappingOptions extends string> = {
```
netama, line 22: maybe add
```suggestion
export type UserConfig<TCustomNameMappingOptions extends string = never> = {
```
so that we don't have to override it when there are no customizations? (might apply to other types as well)

netama (approved):
great ⭐  added a few comments/questions, mostly around not having to define extra args when not customizing (which will be the common use case), other than that LGTM so pre-approving (if there are issues with setting the defaults let's discuss :) )

PR #5523, opened 2024-02-29
SALTO-5483 Exclude Assets Types in free jsm

idoamit8:
Exclude Assets Types in free jsm.

---

Users who set enableJSM to true gain support for assets. However, they don't necessarily have assets in their Jira account, as only JSM Premium includes them. I modified the code to verify the presence of JSM Premium.
---
_Release Notes_: 

---
_User Notifications_: 


packages/jira-adapter/src/filters/account_info.ts
```diff
@@ -112,6 +112,12 @@ export const isJsmEnabledInService = async (client: JiraClient): Promise<boolean
   return accountLicense.applications?.some((app: Value) => app.id === 'jira-servicedesk')
 }
 
+export const hasJsmPremium = async (client: JiraClient): Promise<boolean> => {
```
omrilit, line 115: The name here could be ambiguous for whether the check is in the service or in Salto.
Maybe have a naming convention similar to isJsmEnabledInService and call it isJsmPremiumEnabledInService?

omrilit (approved):
LGTM!

PR #5525, opened 2024-02-29
SALTO-5415 support customizing the http response codes

Uri-Bar:
SALTO-5415 support customizing the http response codes
---

tested in zendesk by removing this: https://github.com/salto-io/salto/blob/main/packages/zendesk-adapter/src/client/client.ts#L97-L100

---
_Release Notes_: 
when creating a connection, we are now able to set an array of validStatuses that the client will not throw an error for them

---
_User Notifications_: 
none

### DIFF
--- a/packages/adapter-components/src/client/http_connection.ts
+++ b/packages/adapter-components/src/client/http_connection.ts
@@ -61,7 +61,11 @@ export interface Connection<TCredentials> {
   login: LoginFunc<TCredentials>
 }
 
-export type ConnectionCreator<TCredentials> = (retryOptions: RetryOptions, timeout?: number) => Connection<TCredentials>
+export type ConnectionCreator<TCredentials> = (
+  retryOptions: RetryOptions,
+  timeout?: number,
+  validStatuses?: number[],
+) => Connection<TCredentials>
 
 const getRetryDelayFromHeaders = (headers: Record<string, string>): number | undefined => {
   // Although the standard is 'Retry-After' is seems that some servers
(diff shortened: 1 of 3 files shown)


packages/adapter-components/src/client/http_connection.ts
```diff
@@ -194,13 +205,17 @@ export const axiosConnection = <TCredentials>({
   baseURLFunc,
   credValidateFunc,
   timeout = 0,
+  validStatuses = [],
```
netama, line 208: nit: maybe `additionalValidStatuses` (since 2xx is always considered valid)?

packages/adapter-components/test/client/http_connection.test.ts
```diff
@@ -30,30 +31,46 @@ describe('client_http_connection', () => {
...
-        accountId: 'ACCOUNT_ID:user123',
-        accountType: 'Sandbox',
-        isProduction: false,
+        expect(mockAxiosAdapter.history.get.length).toBe(1)
+        const req = mockAxiosAdapter.history.get[0]
+        expect(req.url).toEqual('/users/me')
+        expect(req.auth).toEqual({ username: 'user123', password: 'pass' })
+        // already verified the customheader1 header in the onGet header matcher
+      })
+      it('should use the provided validStatuses', async () => {
```
netama, line 65: maybe extend the tests a bit? make sure it fails when the arg is not provided and succeeds when it is?

netama (commented):
cool :) added some minor comments.
let's have a follow-up that "upgrades" the existing adapters to use this arg instead of the custom checks today, to make sure all cases are covered?

PR #5528, opened 2024-03-03
SALTO-4246: Fetch and Deploy issue type icons

idoamit8:
In this PR:
1. I added filter to fetch and deploy issue type icons (static files).
2. I added CV to check if it is possible to delete issue type icon. 

---

1. Jira doesn't allow to delete issueType icon, only add or update one. 

---
_Release Notes_: 
_Jira Adapter:_
* Now support fetching and deploying issue type icons. 

---
_User Notifications_: 
_Jira Adapter:_
* Users will see new nacls of type `issueTypeIcon` upon their next fetch.


packages/jira-adapter/src/change_validators/delete_issue_type_icon.ts
```diff
@@ -0,0 +1,55 @@
...
+
+/*
+ * This validator checks whether an issue type icon can be deleted.
+ * It is not allowed to delete an issue type icon without deleting the issue type itself.
+ */
+export const deleteIssueTypeIconValidator: ChangeValidator = async (changes, elementSource) => {
+  if (elementSource === undefined) {
+    return []
+  }
+  const issueTypesNames = (await getInstancesFromElementSource(elementSource, [ISSUE_TYPE_NAME])).map(
```
IdoZakk, line 31: If you only need the names it is better to go over the list (which returns elemIDs)

packages/jira-adapter/src/filters/issue_type_icon.ts
```diff
@@ -0,0 +1,248 @@
...
+      throw new Error('Failed to deploy logo to Jira: Invalid response from Jira API')
+    }
+    await client.put({
+      url: `/rest/api/3/issuetype/${parent.id}`,
+      data: {
+        avatarId: resp.data.id,
+        description: parent.description,
+        name: parent.name,
+      },
+    })
```
IdoZakk, line 180: This means we might update the issue type twice.
More importantly, it breaks the usage of the swagger. If Jira will add a new field to the issue type API we will need to manually add it here (and we might forget....)
I think the best path here is to create a change of the issue type.
There is here also the problem of multiple parents, we can think together if you want how to do it

packages/jira-adapter/src/change_validators/delete_issue_type_icon.ts
IdoZakk, line 31: Typo: you should use `elementsSource`, not `elementSource`

IdoZakk (commented):
Nice work 🦸 
There are a few things to consider that might require non-trivial changes, most notably using the swagger infrastructure, choosing the correct API, and how to address possible multiple parents.
I did not review the tests yet.

packages/jira-adapter/src/filters/issue_type_icon.ts
```diff
@@ -0,0 +1,248 @@
...
+  isAdditionChange,
+  isAdditionOrModificationChange,
+  isInstanceElement,
+  isRemovalChange,
+  isStaticFile,
+} from '@salto-io/adapter-api'
+import _ from 'lodash'
+import { createSchemeGuard, getParent, pathNaclCase } from '@salto-io/adapter-utils'
+import { elements as adapterElements } from '@salto-io/adapter-components'
+import { ResponseValue } from '@salto-io/adapter-components/src/client'
```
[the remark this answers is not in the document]
idoamit8: Fixed.

packages/jira-adapter/src/filters/issue_type_icon.ts
```diff
@@ -0,0 +1,248 @@
...
+    id: Joi.string().required(),
+  })
+    .required()
+    .unknown(true),
+})
+  .required()
+  .unknown(true)
+
+type QueryParamsType = {
+  foramt: string
```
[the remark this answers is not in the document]
idoamit8: Indeed.

packages/jira-adapter/src/filters/issue_type_icon.ts
```diff
@@ -0,0 +1,248 @@
...
+    fields: {
+      id: {
+        refType: BuiltinTypes.SERVICE_ID_NUMBER,
+        annotations: { [CORE_ANNOTATIONS.HIDDEN_VALUE]: true },
+      },
+      content: { refType: BuiltinTypes.STRING },
+      contentType: { refType: BuiltinTypes.STRING },
+      fileName: { refType: BuiltinTypes.STRING },
+    },
+    path: [JIRA, adapterElements.TYPES_PATH, adapterElements.TYPES_PATH, ISSUE_TYPE_ICON_NAME],
```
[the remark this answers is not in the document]
idoamit8: By mistake. Removed

packages/jira-adapter/src/change_validators/delete_issue_type_icon.ts
idoamit8: I delete this FIle since you gave me an endpoint to delete the icon.

packages/jira-adapter/src/change_validators/delete_issue_type_icon.ts
```diff
@@ -0,0 +1,55 @@
...
+    .filter(change => {
+      try {
+        const instance = getChangeData(change)
+        const parentName = getParent(instance).elemID.name
+        return issueTypesNames.includes(parentName)
+      } catch (e) {
+        return false
+      }
+    })
+    .map(getChangeData)
```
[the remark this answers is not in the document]
idoamit8: I deleted this FIle since you gave me an endpoint to delete the icon.

packages/jira-adapter/src/filters/issue_type_icon.ts
```diff
@@ -0,0 +1,248 @@
...
+    const logoType = createIconType()
+    setTypeDeploymentAnnotations(logoType)
+    await addAnnotationRecursively(logoType, CORE_ANNOTATIONS.CREATABLE)
+    await addAnnotationRecursively(logoType, CORE_ANNOTATIONS.UPDATABLE)
+    await addAnnotationRecursively(logoType, CORE_ANNOTATIONS.DELETABLE)
+    elements.push(logoType)
+    const errors: SaltoError[] = []
+    try {
+      await Promise.all(
+        issueTypes.map(async issueType => {
```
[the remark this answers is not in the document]
idoamit8: Its not recommended to use async in forEach, as we can see here - https://codeburst.io/javascript-async-await-with-foreach-b6ba62bbf404.

packages/jira-adapter/src/filters/issue_type_icon.ts
```diff
@@ -0,0 +1,248 @@
...
+    headers: { ...PRIVATE_API_HEADERS, 'Content-Type': 'image/png' },
+  })
+  return resp
+}
+
+const deployLogo = async (change: Change<InstanceElement>, client: JiraClient): Promise<void> => {
+  try {
+    const logoInstance = getChangeData(change)
+    const parent = getParent(logoInstance).value
+    const logoUrl = `/rest/api/3/issuetype/${parent.id}/avatar2`
```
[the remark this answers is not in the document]
idoamit8: Great ! changed to the documented endpoint :)

packages/jira-adapter/src/change_validators/delete_issue_type_icon.ts
idoamit8: file deleted

packages/jira-adapter/src/filters/issue_type_icon.ts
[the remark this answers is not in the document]
IdoZakk: The two of them are documented....
https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-types/#api-rest-api-3-issuetype-id-avatar2-post

packages/jira-adapter/src/filters/issue_type_icon.ts
idoamit8: I removed it. Not sure why I put in the first place

packages/jira-adapter/src/filters/issue_type_icon.ts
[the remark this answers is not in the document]
IdoZakk: We use `awu` in case we need to do an async foreach. No reason to map and return undefined in the end.
Also I just noticed that the try is outside the loop, which means you will stop with the first error, in which case you can return at that point. 
I think it is cleaner to convert to a list of errors, if you feel strongly about this way and Omri thinks it is better go ahead

packages/jira-adapter/src/filters/issue_type_icon.ts
```diff
@@ -0,0 +1,241 @@
...
+    throw new Error(`Failed to deploy icon to Jira: ${e.message}`)
+  }
+}
+
+const filter: FilterCreator = ({ client }) => ({
+  name: 'issueTypeIconFilter',
+  onFetch: async elements => {
+    const issueTypes = elements
+      .filter(isInstanceElement)
+      .filter(instance => instance.elemID.typeName === ISSUE_TYPE_NAME)
```
IdoZakk, line 191: We should add a check here to see this is not running on DC

packages/jira-adapter/test/filters/issue_type_icon.test.ts
```diff
@@ -0,0 +1,163 @@
...
+        if (params.url === '/rest/api/3/universal_avatar/view/type/issuetype/avatar/101') {
+          return {
+            status: 200,
+            data: content,
+          }
+        }
+        throw new Error('Err')
+      })
+      await filter.onFetch(elements)
+      const iconInstanse = elements.filter(isInstanceElement).find(e => e.elemID.typeName === ISSUE_TYPE_ICON_NAME)
```
IdoZakk, line 72: Typo: should be iconInstance

packages/jira-adapter/test/filters/issue_type_icon.test.ts
```diff
@@ -0,0 +1,163 @@
...
+        fileName: 'issueType1.png',
+        id: 101,
+        content: new StaticFile({
+          filepath: 'jira/IssueTypeIcon/issueType1.png',
+          encoding: 'binary',
+          content,
+        }),
+      })
+    })
+    it('should not add issue layout if it is a bad response', async () => {
```
IdoZakk, line 86: Test title is about issue layouts, also in other tests

packages/jira-adapter/test/filters/issue_type_icon.test.ts
```diff
@@ -0,0 +1,163 @@
...
+        contentType: 'png',
+        fileName: 'issueType1.png',
+        id: 101,
+        content: new StaticFile({
+          filepath: 'jira/IssueTypeIcon/issueType1.png',
+          encoding: 'binary',
+          content,
+        }),
+      })
+    })
```
IdoZakk, line 85: Please also verify the type element you added

packages/jira-adapter/test/filters/issue_type_icon.test.ts
```diff
@@ -0,0 +1,163 @@
...
+        id: 101,
+        content: new StaticFile({
+          filepath: 'jira/IssueTypeIcon/issueType1.png',
+          encoding: 'binary',
+          content,
+        }),
+      })
+      issueTypeIconInstance.annotate({
+        [CORE_ANNOTATIONS.PARENT]: [new ReferenceExpression(issueTypeInstance.elemID, issueTypeInstance)],
+      })
```
IdoZakk, line 122: nit: you can add it as part of the constructor of InstanceElement

packages/jira-adapter/src/filters/issue_type_icon.ts
```diff
@@ -0,0 +1,241 @@
...
+  return resp
+}
+
+const deployIcon = async (change: Change<InstanceElement>, client: JiraClient): Promise<void> => {
+  try {
+    const iconInstance = getChangeData(change)
+    const parent = getParent(iconInstance)
+    if (isRemovalChange(change)) {
+      await client.delete({
+        url: `/rest/api/3/universal_avatar/type/issuetype/owner/${parent.value.id}/avatar/${iconInstance.value.id}`,
```
IdoZakk, line 169: This call fails with 401 for all avatars that are built-in. It only works for uploaded avatars

IdoZakk (commented):
LGTM
I tested it a bit, and I saw a few problems.
* The main one is storing the same icon more than once. If two types have the same built-in icon, they return the same icon ID, but currently, we will store this icon (from the same URL) twice. For customers with many issue types this can be a big problem. ( Also, if you have several issue types with the same icon, all their references will point to a single icon as they all have the same ID)
* avatarId is currently not creatable, so copying a new issueType between envs will not update the icon.
* We need to find a way to avoid the deployment of native icons. You can use the following [API](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-avatars#api-rest-api-3-avatar-type-system-get); no reason to re-deploy those. We should consider if we want to bring them (I guess we do so users will be able to know which icon they are using), and we can't give them automatic names. Either use the ID or name them all hardcoded; I guess you can talk to Lior about it; naming them will be nice)
* I wrote in one of the comments that we can delete only uploaded icons.

packages/jira-adapter/src/filters/issue_type_icon.ts
idoamit8: I removed all system avatars 🙅🏻

PR #5531, opened 2024-03-03
fetch support for basic workato deploy

oneYAD:
not omit extended_input/output_schema when using new flag - FETCH_V2_FIELDS_TO_OMIT
From now on, users should use the FETCH_V2_FIELDS_TO_OMIT on theirs config and start getting the extended_schemas

---

preperation for deploy support - https://github.com/salto-io/salto/pull/4096 
When deploying Workato we will use the Recipe Lifecycle Managment feature. When using it without extended_schema recipe could stop, recipe parameters could change without announcing and hazards could pop up. So, we change the fetch structure, such that extended_schema enter to the Nacl

---
_Release Notes_: 

---
_User Notifications_: 
None

### DIFF
--- a/packages/workato-adapter/src/adapter_creator.ts
+++ b/packages/workato-adapter/src/adapter_creator.ts
@@ -24,7 +24,7 @@ import {
   CLIENT_CONFIG,
   validateFetchConfig,
   FETCH_CONFIG,
-  DEFAULT_CONFIG,
+  getDefaultConfig,
   WorkatoFetchConfig,
 } from './config'
 import WorkatoClient from './client/client'
@@ -41,13 +41,13 @@ const credentialsFromConfig = (config: Readonly<InstanceElement>): Credentials =
 
 const adapterConfigFromConfig = (config: Readonly<InstanceElement> | undefined): WorkatoConfig => {
   const configValue = config?.value ?? {}
-
+  const defaultConfig = getDefaultConfig(configValue.fetch?.enableFetchSturctureV2)
   const apiDefinitions = configUtils.mergeWithDefaultConfig(
-    DEFAULT_CONFIG.apiDefinitions,
+    defaultConfig.apiDefinitions,
     config?.value.apiDefinitions,
   ) as configUtils.AdapterDuckTypeApiConfig
 
-  const fetch = configUtils.mergeWithDefaultConfig(DEFAULT_CONFIG.fetch, config?.value.fetch) as WorkatoFetchConfig
+  const fetch = configUtils.mergeWithDefaultConfig(defaultConfig.fetch, config?.value.fetch) as WorkatoFetchConfig
 
   const adapterConfig: { [K in keyof Required<WorkatoConfig>]: WorkatoConfig[K] } = {
     client: configValue.client,
@@ -71,7 +71,7 @@ export const adapter: Adapter = {
     const updatedConfig = configUtils.configMigrations.migrateDeprecatedIncludeList(
       // Creating new instance is required because the type is not resolved in context.config
       new InstanceElement(ElemID.CONFIG_NAME, configType, context.config?.value),
-      DEFAULT_CONFIG,
+      getDefaultConfig(context.config?.value.fetch?.enableFetchSturctureV2),
     )
 
     const config = adapterConfigFromConfig(updatedConfig?.config[0] ?? context.config)
--- a/packages/workato-adapter/src/adapter_creator.ts
+++ b/packages/workato-adapter/src/adapter_creator.ts
@@ -26,6 +26,7 @@ import {
   FETCH_CONFIG,
   getDefaultConfig,
   WorkatoFetchConfig,
+  ENABLE_DEPLOY_SUPPORT_FLAG,
 } from './config'
 import WorkatoClient from './client/client'
 import { createConnection } from './client/connection'
@@ -41,7 +42,7 @@ const credentialsFromConfig = (config: Readonly<InstanceElement>): Credentials =
 
 const adapterConfigFromConfig = (config: Readonly<InstanceElement> | undefined): WorkatoConfig => {
   const configValue = config?.value ?? {}
-  const defaultConfig = getDefaultConfig(configValue.fetch?.enableFetchSturctureV2)
+  const defaultConfig = getDefaultConfig(configValue.enableDeploySupport)
   const apiDefinitions = configUtils.mergeWithDefaultConfig(
     defaultConfig.apiDefinitions,
     config?.value.apiDefinitions,
@@ -54,6 +55,7 @@ const adapterConfigFromConfig = (config: Readonly<InstanceElement> | undefined):
     fetch,
     apiDefinitions,
     deploy: configValue.deploy,
+    enableDeploySupport: configValue.enableDeploySupport,
   }
 
   validateClientConfig(CLIENT_CONFIG, adapterConfig.client)
@@ -71,7 +73,7 @@ export const adapter: Adapter = {
     const updatedConfig = configUtils.configMigrations.migrateDeprecatedIncludeList(
       // Creating new instance is required because the type is not resolved in context.config
       new InstanceElement(ElemID.CONFIG_NAME, configType, context.config?.value),
-      getDefaultConfig(context.config?.value.fetch?.enableFetchSturctureV2),
+      getDefaultConfig(context.config?.value.fetch?.[ENABLE_DEPLOY_SUPPORT_FLAG]),
     )
 
     const config = adapterConfigFromConfig(updatedConfig?.config[0] ?? context.config)
--- a/packages/workato-adapter/src/config.ts
+++ b/packages/workato-adapter/src/config.ts
@@ -38,11 +38,28 @@ const { createDucktypeAdapterApiConfigType, validateDuckTypeFetchConfig } = conf
 
 export const DEFAULT_SERVICE_ID_FIELD = 'id'
 export const DEFAULT_ID_FIELDS = ['name']
-export const FIELDS_TO_OMIT: configUtils.FieldToOmitType[] = [
+export const DEFAULT_FIELDS_TO_OMIT: configUtils.FieldToOmitType[] = [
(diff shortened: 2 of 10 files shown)


packages/workato-adapter/src/config.ts
```diff
@@ -38,12 +38,29 @@ const { createDucktypeAdapterApiConfigType, validateDuckTypeFetchConfig } = conf
...
   { fieldName: 'updated_at', fieldType: 'string' },
   { fieldName: 'extended_input_schema' },
   { fieldName: 'extended_output_schema' },
 ]
+export const FETCH_V2_FIELDS_TO_OMIT: configUtils.FieldToOmitType[] = [
+  { fieldName: 'created_at', fieldType: 'string' },
+  { fieldName: 'updated_at', fieldType: 'string' },
+  { fieldName: 'dynamicPickListSelection' },
+  { fieldName: 'visible_config_fields' },
+]
```
shir-reifenberg, line 52: Didn't we say we keep those for now ?

shir-reifenberg (commented):
Great job!
I might be missing something, but I think `getDefaultConfig` can be simplified to only change the `fieldsToOmit` of recipe code instead of changing the entire config
`DEFAULT_FIELDS_TO_OMIT` can not include external and internal output schema, and then we can change only recipe code config in  `getDefaultConfig` which should only modify the types config
WDYT?

oneYAD:
> Great job! 🎉
> 
> I might be missing something, but I think `getDefaultConfig` can be simplified to only change the `fieldsToOmit` of recipe code instead of changing the entire config `DEFAULT_FIELDS_TO_OMIT` can not include external and internal output schema, and then we can change only recipe code config in `getDefaultConfig` which should only modify the types config WDYT?
> 
> few more things -
> 
>     * can you please add the related ticket to the PR title and remove the release notes ? (as this is not user facing for now)
> 
>     * can you elaborate in the description why this change is needed?
> 
>     * please add a unit test to `getDefaultConfig` as this will be exported function

packages/workato-adapter/src/config.ts
```diff
@@ -55,6 +72,7 @@ export type WorkatoClientConfig = definitions.ClientBaseConfig<definitions.Clien
 
 export type WorkatoFetchConfig = definitions.UserFetchConfig & {
   serviceConnectionNames?: Record<string, string[]>
+  enableFetchSturctureV2?: boolean
```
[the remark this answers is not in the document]
oneYAD: It was like that in the beginning, but after discussion with @yuribro we decided to do it like this.
The only thing we really change in this PR is the Fetch structure, and we don't want to put fetch change flag outside the fetch config. And if we do so, the relevant name would be the change of the structure and not the deploy support. WDYT?

packages/workato-adapter/src/config.ts
oneYAD: yes, changed

packages/workato-adapter/src/config.ts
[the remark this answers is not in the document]
shir-reifenberg: I don't see a reason to decouple the two... The fetch structure changes are only relevant for deployment support. IMO we should have one flag to control the entire feature, having 2 separate flags might result in wired config variations

packages/workato-adapter/src/config.ts
```diff
@@ -38,12 +38,29 @@ const { createDucktypeAdapterApiConfigType, validateDuckTypeFetchConfig } = conf
 
 export const DEFAULT_SERVICE_ID_FIELD = 'id'
 export const DEFAULT_ID_FIELDS = ['name']
-export const FIELDS_TO_OMIT: configUtils.FieldToOmitType[] = [
+export const DEFAULT_FIELDS_TO_OMIT: configUtils.FieldToOmitType[] = [
   { fieldName: 'created_at', fieldType: 'string' },
   { fieldName: 'updated_at', fieldType: 'string' },
   { fieldName: 'extended_input_schema' },
   { fieldName: 'extended_output_schema' },
```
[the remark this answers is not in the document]
shir-reifenberg: not sure if it's before or after we talked, but it worked for me 🙃

packages/workato-adapter/src/config.ts
```diff
@@ -38,18 +38,35 @@ const { createDucktypeAdapterApiConfigType, validateDuckTypeFetchConfig } = conf
...
+  { fieldName: 'authorized_at', fieldType: 'string' },
+  { fieldName: 'authorization_status', fieldType: 'string' },
+]
 export const FIELDS_TO_HIDE: configUtils.FieldToHideType[] = []
 
 export const CLIENT_CONFIG = 'client'
 export const FETCH_CONFIG = 'fetch'
 export const DEPLOY_CONFIG = 'deploy'
 export const API_DEFINITIONS_CONFIG = 'apiDefinitions'
+export const SUPPORT_DEPLOY_FLAG = 'enableDeploySupport'
```
shir-reifenberg, line 69: nit
```suggestion
export const ENABLE_DEPLOY_SUPPORT_FLAG = 'enableDeploySupport'
```

packages/workato-adapter/src/config.ts
```diff
@@ -189,7 +196,7 @@ export const DEFAULT_TYPES: Record<string, configUtils.TypeDuckTypeConfig> = {
   },
 }
 
-export const DEFAULT_CONFIG: WorkatoConfig = {
+const DEFAULT_CONFIG: WorkatoConfig = {
```
shir-reifenberg, line 199: Pease add `enableDeploySupport = false` to the DEFAULT_CONFIG const (and just overridde it in `getDefaultConfig` when needed)

packages/workato-adapter/src/config.ts
```diff
@@ -209,6 +216,16 @@ export const DEFAULT_CONFIG: WorkatoConfig = {
   },
 }
 
+export const getDefaultConfig = (deploySupported = false): WorkatoConfig => {
+  const defaultConfig = _.cloneDeep(DEFAULT_CONFIG)
+  defaultConfig[SUPPORT_DEPLOY_FLAG] = deploySupported
```
shir-reifenberg, line 221: let's just update it when necessary

packages/workato-adapter/src/config.ts
```diff
@@ -209,6 +216,16 @@ export const DEFAULT_CONFIG: WorkatoConfig = {
...
+  const typeDefaultTransformation = defaultConfig[API_DEFINITIONS_CONFIG].typeDefaults.transformation
+  if (typeDefaultTransformation !== undefined && deploySupported) {
+    typeDefaultTransformation.fieldsToOmit = typeDefaultTransformation.fieldsToOmit?.filter(field => !EXTENDED_SCHEMA_FIELDS.map(ex => ex.fieldName).includes(field.fieldName))
```
shir-reifenberg, line 224: can you please add a log.debug here with the updated default fieldsToOmit ? 🙏

shir-reifenberg, line 224: super nit
```suggestion
    typeDefaultTransformation.fieldsToOmit = typeDefaultTransformation.fieldsToOmit?.filter(field => !EXTENDED_SCHEMA_FIELDS.map(field => field.fieldName).includes(field.fieldName))
```

packages/workato-adapter/src/adapter_creator.ts
```diff
@@ -41,19 +42,20 @@ const credentialsFromConfig = (config: Readonly<InstanceElement>): Credentials =
 
 const adapterConfigFromConfig = (config: Readonly<InstanceElement> | undefined): WorkatoConfig => {
   const configValue = config?.value ?? {}
-
+  const defaultConfig = getDefaultConfig(configValue.fetch?.deploySupportEnable)
```
shir-reifenberg, line 45: looks like the wrong path (fetch.deploySupportEnable), no?

packages/workato-adapter/src/adapter_creator.ts
```diff
@@ -41,19 +42,20 @@ const credentialsFromConfig = (config: Readonly<InstanceElement>): Credentials =
...
-  const fetch = configUtils.mergeWithDefaultConfig(DEFAULT_CONFIG.fetch, config?.value.fetch) as WorkatoFetchConfig
+  const fetch = configUtils.mergeWithDefaultConfig(defaultConfig.fetch, config?.value.fetch) as WorkatoFetchConfig
 
   const adapterConfig: { [K in keyof Required<WorkatoConfig>]: WorkatoConfig[K] } = {
     client: configValue.client,
     fetch,
     apiDefinitions,
     deploy: configValue.deploy,
+    enableDeploySupport: configValue.deploySupportEnable,
   }
```
shir-reifenberg, line 59: should we change all values here to use `defaultConfig` ?

shir-reifenberg (commented):
Looks good, added a few comments

packages/workato-adapter/src/config.ts
oneYAD: I change it to ex, because the filter is also with field expression, and I thought it will be more readable, WDYT?

packages/workato-adapter/src/config.ts
oneYAD: It makes problem with the tests. when you just update two tests will use the same DEAFULT_CONFIG

packages/workato-adapter/src/config.ts
```diff
@@ -207,6 +216,23 @@ export const DEFAULT_CONFIG: WorkatoConfig = {
     types: DEFAULT_TYPES,
     supportedTypes: SUPPORTED_TYPES,
   },
+  [ENABLE_DEPLOY_SUPPORT_FLAG]: false,
+}
+
+export const getDefaultConfig = (deploySupported = false): WorkatoConfig => {
+  const defaultConfig = _.cloneDeep(DEFAULT_CONFIG)
```
shir-reifenberg, line 223: if `deploySupported` is false we can early return here with `DEFAULT_CONFIG`, right ?

shir-reifenberg (approved):
Looks great! 🎉  had one comment but approving to save cycle :)
Please update the PR title to include ticket number before merging

PR #5532, opened 2024-03-03
SALTO-5474: fix idFields mechanism for newly created fields

edenhassid:
fix idFields mechanism for newly created fields

---

---
_Release Notes_: 
zendesk:
*  fix idFields mechanism for newly created fields

---
_User Notifications_: 

### DIFF
--- a/packages/zendesk-adapter/src/client/client.ts
+++ b/packages/zendesk-adapter/src/client/client.ts
@@ -54,7 +54,8 @@ export type AttachmentRes = {
   }
 }
 
-const getSchema = (field: string): Joi.AnySchema => Joi.object({
+const getSchema = (field: string): Joi.AnySchema =>
+  Joi.object({
     data: Joi.object({
       [field]: Joi.array().items(Joi.object().unknown(true)).required(),
     })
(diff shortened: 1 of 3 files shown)


packages/zendesk-adapter/src/client/client.ts
```diff
@@ -89,7 +163,9 @@ export default class ZendeskClient extends clientUtils.AdapterHTTPClient<
     args: clientUtils.ClientBaseParams,
   ): Promise<clientUtils.Response<clientUtils.ResponseValue | clientUtils.ResponseValue[]>> {
     try {
-      return await super.get(args)
+      // will be deleted once we move to new infra
```
adfineberg, line 166: Do you have a ticket number for this?

adfineberg (approved):

packages/zendesk-adapter/src/client/client.ts
edenhassid: I'll check with @netama

packages/zendesk-adapter/src/client/client.ts
```diff
@@ -35,6 +36,74 @@ type LogsFilterConfig = {
...
+  let usernameCounter = 0
+  let attachmentCounter = 0
+  return (
+    url: string,
+    response: clientUtils.Response<clientUtils.ResponseValue | clientUtils.ResponseValue[]>,
+  ): clientUtils.Response<clientUtils.ResponseValue | clientUtils.ResponseValue[]> => {
+    // for endpoint /api/v2/business_hours/schedules/{scheduleId}/holidays
+    if (url.includes('holidays') && isHolidayResponse(response)) {
+      response.data.holidays.forEach(day => {
+        day.start_year = holidayCounter
```
netama, line 86: this one could just have the "final" values and we can remove the filter, right?

packages/zendesk-adapter/src/client/client.ts
```diff
@@ -35,6 +36,74 @@ type LogsFilterConfig = {
...
+        day.end_year = holidayCounter
+        holidayCounter += 1
+      })
+      // for endpoint /api/v2/recipient_addresses
+    } else if (url.includes('recipient_addresses') && isSupportAddressResponse(response)) {
+      response.data.recipient_addresses.forEach(email => {
+        email.username = usernameCounter
```
netama, line 93: same

packages/zendesk-adapter/src/client/client.ts
edenhassid: right, thought we could have this for now and change it later when the infra changes

PR #5535, opened 2024-03-03
SALTO-5424 support defining aliases and importantValues in definitions

omrifarkash:
- support defining aliases and importantValues in definitions
- build dependencies graph for addAlias function to calculate the order we should create aliases

---

---
_Release Notes_: 
- addAlias function is no longer supports `secondInterationsTypes` parameter, instead it builds a dependencies graph

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/filters/add_alias.ts
+++ b/packages/adapter-components/src/filters/add_alias.ts
@@ -1,0 +1,55 @@
+/*
+ *                      Copyright 2024 Salto Labs Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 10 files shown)


packages/adapter-components/src/filters/add_alias.ts
```diff
@@ -0,0 +1,56 @@
+/*
```
omrifarkash, line 1: To be used in adapter-wrapper once it merged :)

edenhassid (approved):
LGTM! 🔥 💯 I think @netama also wanted to take a look at the definitions before you merge :)

PR #5543, opened 2024-03-04
SALTO-4077 - Salesforce: Ignore duplicate retrieve config changes

tomsellek:
When creating a config change suggestion, make sure it isn't already excluded

---

For now we'll only log the cases where this change would take effect. If we see things are OK we can fully enable it.

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A


tamtamirr (approved):
LGTM

PR #5544, opened 2024-03-04
SALTO-5501: fix account id filter before bug

IdoZakk:
See details in the Jira Ticket.
The `account_id_filter` is using a dictionary cache that is based on path. When we run it on both before and after, and when they are not the same structure, we create two entries for the same change and override data that is not account_id.

I had to make a small change in notification scheme that was based on the before having the `account_id_filter` changes.
This fix also fixed another small bug that existed only in the OSS

---

None

---
_Release Notes_: 
Jira Adapter:
* Fixed a bug that could cause partial success for deployments 

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/account_id/account_id_filter.ts
+++ b/packages/jira-adapter/src/filters/account_id/account_id_filter.ts
@@ -15,7 +15,6 @@
  */
 import {
   BuiltinTypes,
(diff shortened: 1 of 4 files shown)


shir-reifenberg (approved):
💯 
Thanks!

PR #5547, opened 2024-03-05
SALTO-5334, SALTO-5346 - Salesforce: move PermissionSet refs to custom references code

tomsellek:
Move the extraction of references from the reference mapper to custom references. This will speed fetches in cases where `PermissionSet`s are included and some of them are very large.

---

I merged two tickets into this PR - the one for treating the `PermissionSet` references the same as `Profile` references in the reference mapper, and the one for treating `PermissionSet` references the same as `Profile` references in the custom references code.

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/transformers/reference_mapping.ts
+++ b/packages/salesforce-adapter/src/transformers/reference_mapping.ts
@@ -972,7 +972,9 @@ export const getReferenceMappingDefs = (args: {
     refDefs = refDefs.concat(fieldPermissionEnumDisabledExtraMappingDefs)
   }
   if (args.otherProfileRefs) {
-    refDefs = refDefs.concat(referencesFromProfile)
+    refDefs = refDefs
+      .concat(referencesFromProfile)
+      .concat(referencesFromPermissionSets)
   }
   if (args.permissionsSetRefs) {
     refDefs = refDefs.concat(referencesFromPermissionSets)
--- a/packages/salesforce-adapter/src/transformers/reference_mapping.ts
+++ b/packages/salesforce-adapter/src/transformers/reference_mapping.ts
@@ -894,7 +894,7 @@ export const fieldPermissionEnumDisabledExtraMappingDefs: FieldReferenceDefiniti
     },
   ]
 
-export const referencesFromProfile: FieldReferenceDefinition[] = [
+const referencesFromProfile: FieldReferenceDefinition[] = [
   {
     src: { field: 'object', parentTypes: ['ProfileObjectPermissions'] },
     target: { type: CUSTOM_OBJECT },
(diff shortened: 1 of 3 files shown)


packages/salesforce-adapter/src/transformers/reference_mapping.ts
```diff
@@ -960,7 +977,9 @@ export const getReferenceMappingDefs = (args: {
     refDefs = refDefs.concat(fieldPermissionEnumDisabledExtraMappingDefs)
   }
   if (args.otherProfileRefs) {
-    refDefs = refDefs.concat(referencesFromProfile)
+    refDefs = refDefs
+      .concat(referencesFromProfile)
+      .concat(referencesFromPermissionSets)
```
tamtamirr, line 982: Do we want to couple the `otherProfileRefs` with `PermissionSets` as-well?

packages/salesforce-adapter/src/transformers/reference_mapping.ts
```diff
@@ -934,6 +933,23 @@ export const referencesFromProfile: FieldReferenceDefinition[] = [
   },
 ]
 
+export const referencesFromPermissionSets = [
```
tamtamirr, line 936: Nit: shouldn't be exported

packages/salesforce-adapter/test/custom_references.test.ts
```diff
@@ -43,12 +43,17 @@ import {
...
+    fields: Values,
+  ): [InstanceElement, InstanceElement] => [
+    new InstanceElement('test', mockTypes.Profile, fields),
+    new InstanceElement('test', mockTypes.PermissionSet, fields),
+  ]
 
   describe('when profile refs are disabled', () => {
     beforeEach(async () => {
-      profileInstance = createTestInstance({
+      ;[permissionSetInstance, profileInstance] = createTestInstances({
```
tamtamirr, line 56: Is the `;` needed here?

packages/salesforce-adapter/src/custom_references.ts
tamtamirr, line 1: The code in this file is referring in many places only to `Profiles` even though we handle both `PermissionSets & Profiles`

tamtamirr (commented):
Looking good! Added some small comments

packages/salesforce-adapter/test/custom_references.test.ts
tomsellek: It's not, but prettier auto-inserts it for some reason

tomsellek: https://github.com/prettier/prettier/issues/4193

packages/salesforce-adapter/src/transformers/reference_mapping.ts
tomsellek: Yes, that's the master switch for references in profiles and should be extended to permission sets.

tomsellek: Yeah, that's a good point.

tamtamirr (approved):
LGTM! Let's not merge until it's communicated properly through PROD ticket

PR #5552, opened 2024-03-06
SALTO-5448: New CustomFields and CustomObjects FLS Change Validator

tamtamirr:
Add a Change Validator that notifies users that the adapter will enable the FLS to certain Profiles on addition of new CustomField or CustomObject 

---

_Release Notes_: 
Salesforce Adapter:
- Implemented a Change Validator that notifies users that the adapter will enable the FLS to certain Profiles on addition of new CustomField or CustomObject 

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/change_validators/new_fields_and_objects_fls.ts
```diff
@@ -0,0 +1,54 @@
...
+import { SalesforceConfig } from '../types'
+
+const createFLSInfo = (
+  element: Element,
+  flsProfiles: string[],
+): ChangeError => {
+  const typeOrField = isField(element) ? 'CustomField' : 'CustomObject'
+  return {
+    message: `${typeOrField} visibility in Profiles.`,
+    detailedMessage: `Deploying this new ${typeOrField} will add access to it in the following Profiles: [${flsProfiles.join(', ')}].`,
```
tomsellek, line 36: Nit: perhaps "Will make it accessible by" instead of "will add access to it in"?

packages/salesforce-adapter/src/change_validators/new_fields_and_objects_fls.ts
```diff
@@ -0,0 +1,54 @@
...
+const changeValidator =
+  (config: SalesforceConfig): ChangeValidator =>
+  async (changes) => {
+    const flsProfiles = getFLSProfiles(config)
+    return changes
+      .filter(isAdditionChange)
+      .map((change) => getChangeData(change))
+      .filter((element) => isObjectType(element) || isField(element))
+      .filter((element) => isCustom(apiNameSync(element)))
+      .map((element) => createFLSInfo(element, flsProfiles))
```
tomsellek, line 51: In the case where Salto is used to manage profiles, we expect the appropriate change to have been made by the user instead of by the filter. Is there a way to avoid this warning if Profiles are managed by Salto?

tomsellek (approved):
Approving to save cycles, but please see my comment about avoiding spurious messages.

packages/salesforce-adapter/src/change_validators/new_fields_and_objects_fls.ts
tamtamirr: IMO it's still better to notify regardless if users defined `flsProfiles` or not as the deploying user may be unaware to this configuration.

PR #5557, opened 2024-03-06
SALTO-5522: Add data to trace logs of http calls

IdoZakk:
Data logs were missing in POST requests that were part of the deploy

---

I removed the existing data from the deploy log to avoid duplicate write

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/client/http_client.ts
+++ b/packages/adapter-components/src/client/http_client.ts
@@ -234,6 +234,7 @@ export abstract class AdapterHTTPClient<TCredentials, TRateLimitConfig extends C
     }
 
     const { url, queryParams, headers, responseType } = params
+    const data = isMethodWithData(params) ? params.data : undefined
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const logResponse = (res: Response<any>): void => {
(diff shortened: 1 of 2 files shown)


packages/adapter-components/src/client/http_client.ts
```diff
@@ -61,6 +61,8 @@ export type ClientDataParams = ClientBaseParams & {
 
 export type ClientParams = ClientBaseParams | ClientDataParams
 
+const isClientDataParams = (params: ClientParams): params is ClientDataParams => 'data' in params
```
shir-reifenberg, line 64: is there any difference from `isMethodWithData` ?

shir-reifenberg (approved):
LGTM 😎  with one question

packages/adapter-components/src/client/http_client.ts
IdoZakk: No, missed that 🤭

packages/adapter-components/src/client/http_client.ts
```diff
@@ -248,6 +249,7 @@ export abstract class AdapterHTTPClient<TCredentials, TRateLimitConfig extends C
           ? `<omitted buffer of length ${res.data.length}>`
           : this.clearValuesFromResponseData(res.data, url),
         headers: this.extractHeaders(res.headers),
+        data,
```
netama, line 252: please add a check that this is an object / array, if it's a buffer we should handle it like above

IdoZakk: I added a similar check to the above. Is this good, or do you prefer a more specific check?

IdoZakk: @netama I see you changed the file to be named deployment_old. Could not find the new place for deployment, should I make the change there as well?

netama: the logs are different in the new infra so no need, will try to pair them more before we upgrade the existing adapters but I don't think you have something specific to remove right now, thanks

PR #5560, opened 2024-03-07
 SALTO-5249: [Zendesk] Throw during Workspace deployment when the Zendesk client returns 200 with an error

DanielBarros1:
Same thing as https://github.com/salto-io/salto/pull/5438 - we saw that an error can also be a string, so add a case to catch that
---

---
_Release Notes_: 

Zendesk:
Fail deployment when Zendesk returns errors when changing workspaces

---
_User Notifications_: 
None


edenhassid (approved):
looks good! 🔥 thanks!

PR #5564, opened 2024-03-08
SALTO-5533 - Salesforce: Fix Quickfetch for allowReferenceTo records

tomsellek:
When fetch-with-changes-detection is enabled, we only attempt to fetch instances that are referred to from instances that changed.
Instead, we should fetch instances that are referenced by instances in the workspace too, in case any of the referred-to instances changed.

---

~We cheat a little by treating `InstanceElement.value` as a `SalesforceRecord` for the purpose of finding references. This is because the data in the workspace/`elementsSource` is stored as `InstanceElement`s, but the fetched records are `SalesforceRecord`s.~

There is an issue where if the user deleted the last reference to a data record, the referred record will not be removed from the workspace. This is because we don't identify the deletion of the referring record (because we don't support deletions in quickfetch)

---
_Release Notes_: 
Salesforce: Quickfetch now works for data records of types listed in `allowReferenceTo`

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/filters/custom_objects_instances.ts
+++ b/packages/salesforce-adapter/src/filters/custom_objects_instances.ts
@@ -27,6 +27,7 @@ import {
   isPrimitiveType,
   isReferenceExpression,
   ReadOnlyElementsSource,
+  Values,
 } from '@salto-io/adapter-api'
 import { pathNaclCase, safeJsonStringify } from '@salto-io/adapter-utils'
 import {
@@ -48,7 +49,7 @@ import {
   LAST_MODIFIED_DATE,
   DATA_INSTANCES_CHANGED_AT_MAGIC,
 } from '../constants'
-import { FilterResult, RemoteFilterCreator } from '../filter'
+import { FilterContext, FilterResult, RemoteFilterCreator } from '../filter'
 import {
   apiName,
   Types,
@@ -71,6 +72,7 @@ import {
   SoqlQuery,
   buildElementsSourceForFetch,
   getChangedAtSingletonInstance,
+  isInstanceOfCustomObjectSync,
 } from './utils'
 import { ConfigChangeSuggestion, DataManagement } from '../types'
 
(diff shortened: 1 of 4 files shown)


packages/salesforce-adapter/src/filters/custom_objects_instances.ts
```diff
@@ -536,19 +541,72 @@ export const getAllInstances = async (
...
+
+  const fetchedBaseRecordByTypeAndId = _.mapValues(
+    fetchedBaseRecordsByType,
+    (records) => _.keyBy(records, (record) => record[CUSTOM_OBJECT_ID_FIELD]),
   )
+
+  const elementSourceRecordsByTypeAndId = _.mapValues(
+    elementSourceRecordsByType,
+    (records) => _.keyBy(records, (record) => record[CUSTOM_OBJECT_ID_FIELD]),
+  )
```
tamtamirr, line 589: I would export all the logic of creating **elementSourceRecordsByTypeAndId** to a dedicated method

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
```diff
@@ -778,6 +836,80 @@ const filterCreator: RemoteFilterCreator = ({ client, config }) => ({
   name: 'customObjectsInstancesFilter',
   remote: true,
   onFetch: async (elements: Element[]): Promise<FilterResult> => {
+    const getAllCustomObjects = async (
```
tamtamirr, line 839: Is this existing code that moved?

tamtamirr (commented):
LGTM! Let's test older version with this new one, and see there's no regression in full fetch. Added some comments. There's quite a diff in this PR, but overall looks good!

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
tomsellek: Yes, from line 793 of the old file

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
tomsellek: Ack

tamtamirr (approved):
LGTM!

PR #5565, opened 2024-03-10
SALTO-4864: Okta: Add a Change Validator to make sure ProfileMappings can only be removed alongside either its source or target

Nurdok:
`ProfileMapping`s can't actually be removed - they are automatically removed by Okta when either side of the mapping (source or target) is removed. This CV will block `ProfileMapping` removal if this isn't the case.

---

Note that currently `ProfileMapping` removals aren't enabled at all - this PR is in preparations for allowing removals.

---

_Release Notes_: 
None

---
_User Notifications_: 
None


packages/okta-adapter/src/change_validators/profile_mapping_removal.ts
```diff
@@ -0,0 +1,60 @@
...
+ * When removing a Profile Mapping, validate that either its source or target are also removed.
+ */
+export const profileMappingRemovalValidator: ChangeValidator = async changes => {
+  const removeInstanceChanges = changes.filter(isInstanceChange).filter(isRemovalChange).map(getChangeData)
+
+  const removedProfileMappingInstances = removeInstanceChanges.filter(
+    instance => instance.elemID.typeName === PROFILE_MAPPING_TYPE_NAME,
+  )
+
+  const removedNames = new Set(removeInstanceChanges.map(instance => instance.elemID.getFullName()))
```
shir-reifenberg, line 33: naming nit 
```suggestion
  const removedElemIDs = new Set(removeInstanceChanges.map(instance => instance.elemID.getFullName()))
```

packages/okta-adapter/test/change_validators/profile_mapping_removal.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    source: new ReferenceExpression(app.elemID, app),
+    target: new ReferenceExpression(userType.elemID, userType),
+  })
+
+  const profileMappingB = new InstanceElement('mappingB', profileMappingType, {
+    source: new ReferenceExpression(userType.elemID, userType),
+    target: new ReferenceExpression(identityProvider.elemID, identityProvider),
+  })
+
+  it('should return an error when ProfileMapping is deleted without its parent Application', async () => {
```
shir-reifenberg, line 46: nit (cause parent has a different meaning in our code)
```suggestion
  it('should return an error when ProfileMapping is deleted without its source or target Application', async () => {
```

packages/okta-adapter/src/change_validators/profile_mapping_removal.ts
```diff
@@ -0,0 +1,60 @@
...
+    })
+    .map(profileMapping => ({
+      elemID: profileMapping.elemID,
+      severity: 'Error',
+      message: 'Cannot remove profile mapping if neither its source nor target are also removed',
+      detailedMessage:
+        `In order to remove ${profileMapping.elemID.name}, either its source (instance ` +
+        `${profileMapping.value.source?.elemID.name} of type ${profileMapping.value.source?.elemID.typeName})` +
+        ` or target (instance ${profileMapping.value.target?.elemID.name} of type` +
+        ` ${profileMapping.value.target?.elemID.typeName}) must be removed as well.`,
```
shir-reifenberg, line 58: detailedMessage is user facing, so we usually avoid using phrases like instances and types in message (can use element instead).
Also, we don't need to repeat the information in message itself.
maybe something like (but fee free to re-phrase...) - 

`To deploy this change, source ${typename} element: ${name} or target ${typename} element: ${name} must be removed as well.`

packages/okta-adapter/src/change_validators/profile_mapping_removal.ts
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+import { ChangeValidator, getChangeData, isInstanceChange, isRemovalChange } from '@salto-io/adapter-api'
+import { logger } from '@salto-io/logging'
+import { PROFILE_MAPPING_TYPE_NAME } from '../constants'
+
+const log = logger(module)
+
+/**
+ * When removing a Profile Mapping, validate that either its source or target are also removed.
```
shir-reifenberg, line 24: can you please add a short explanation why this needed?

shir-reifenberg (commented):
Looks great!!
added mainly nits 😎

packages/okta-adapter/src/change_validators/profile_mapping_removal.ts
```diff
@@ -0,0 +1,60 @@
...
+  )
+
+  const removedNames = new Set(removeInstanceChanges.map(instance => instance.elemID.getFullName()))
+
+  return removedProfileMappingInstances
+    .filter(profileMapping => {
+      try {
+        return !(
+          removedNames.has(profileMapping.value.source?.elemID.getFullName()) ||
+          removedNames.has(profileMapping.value.target?.elemID.getFullName())
```
omrilit, line 40: Threre is an implied assumption here that `source` & `target` are ReferenceExpressions, can we be more explicit and assert that before accessing that data?

packages/okta-adapter/src/change_validators/profile_mapping_removal.ts
```diff
@@ -0,0 +1,60 @@
...
+        )
+      } catch (e) {
+        log.error(
+          'Could not run profileMappingAndAppValidator validator for instance ' +
+            `${profileMapping.elemID.getFullName}: ${e}`,
+        )
+        return false
+      }
```
omrilit, line 48: I think that we can omit the try...catch block here, unless there is something unexpected that may fail here.
IIUC the only thing that may fail is the access to the `source` & `target` which can be handled without catching an exception. WDYT?

packages/okta-adapter/src/change_validators/profile_mapping_removal.ts
omrilit: @shir-reifenberg do we have a PM that reviews the messages in Okta? I think that it's better to ask for a review of validator messages in general.

omrilit (commented):
Great work Amir :)

packages/okta-adapter/src/change_validators/profile_mapping_removal.ts
Nurdok: But this is storing the result of `getFullName` here, not the top-level `ElemID` type, why call it `ElemIDs`?

packages/okta-adapter/test/change_validators/profile_mapping_removal.test.ts
Nurdok: Thanks, this was copy-pasta from another test :)

packages/okta-adapter/src/change_validators/profile_mapping_removal.ts
Nurdok: I added a long explanation :)

packages/okta-adapter/src/change_validators/profile_mapping_removal.ts
Nurdok: Done, PTAL.

packages/okta-adapter/src/change_validators/profile_mapping_removal.ts
shir-reifenberg: Nice!

packages/okta-adapter/src/change_validators/profile_mapping_removal.ts
shir-reifenberg: you're right

shir-reifenberg (approved):
Looks great!
Please review the validator message with someone from product before merging as @omrilit suggested, but approving from my end :)

packages/okta-adapter/src/change_validators/profile_mapping_removal.ts
liorn: I think they will understand it anyway, but I agree with Shir's general statement.

Let's gently change to `In order to remove this Profile Mapping, remove its source (<typename> <name>) or target (<typename> <name>) as well. `

ALSO, if it's easy and possible and profile mappings have them, let's use aliases instead of element names.

PR #5567, opened 2024-03-10
SALTO-3080 Add greedy id extraction from JS files

adfineberg:
This is the trivial strategy of extracting all over 6 digit long matches and trying to find a reference. 
In next PRs (based on discussions with clients), we will add some config to allow digit amount changes, and to choose different strategies.

---

---
_Release Notes_: 

---
_User Notifications_: 


edenhassid (approved):
looks good! 🔥 🔥 🔥

PR #5569, opened 2024-03-10
SALTO-4404: Task or Event Fields Modifications Change Validator

tamtamirr:
Task or Event Fields Modifications Change Validator.

---

Modifying fields on the Event / Task Objects directly is forbidden and instead needs to be done through the Activity Object.

Here's the CV output for both cases:

```
Please review the following messages; changes with errors will not be deployed.
    Error: Modifying a field of Task or Event is not allowed in the following elements:
      salesforce.Event.field.Test_Field__c
        Modifing the field Test_Field__c of the Event object directly is forbidden. Instead, modify the corresponding field in the Activity Object
      salesforce.Task.field.Test_Field__c
        Modifing the field Test_Field__c of the Task object directly is forbidden. Instead, modify the corresponding field in the Activity Object
```

---
_Release Notes_: 
Salesforce Adapter:
- Add Change Validator that forbids changes on fields of Task and Event.

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/change_validators/task_or_event_fields_modifications.ts
```diff
@@ -0,0 +1,44 @@
...
+import { apiNameSync, isCustomObjectSync } from '../filters/utils'
+
+const isFieldOfTaskOrEvent = ({ parent }: Field): boolean =>
+  isCustomObjectSync(parent) &&
+  [TASK_CUSTOM_OBJECT, EVENT_CUSTOM_OBJECT].includes(apiNameSync(parent) ?? '')
+
+const createFieldOfTaskOrEventChangeError = (field: Field): ChangeError => ({
+  elemID: field.elemID,
+  severity: 'Error',
+  message: 'Modifying a field of Task or Event is not allowed',
```
tomsellek, line 33: Nit: since this error will also be shown for addition and removal, how about passing the change as a parameter and doing `Unable to ${change.action} a field of ...`?

tamtamirr: I don't think it's worth the complexity overhead 🙏

PR #5575, opened 2024-03-11
SALTO-5195 - Salesforce: propagate messages from retrieve

tomsellek:
We currently discard any messages we receive in calls to `retrieve`. Let's pass them up so they can be returned to the user.

---

This required changing the return type of `retrieveMetadataInstances.retrieveInstances` and propagating the return type change up through all callers.

Note that we don't associate the messages with an `ElemID` - that would require mapping a file name to an `ElemID`, which would, in turn, require us to adapt and reuse the relevant code from the deploy flow.

Test:
 - Run a fetch

---
_Release Notes_: 
Salesforce: Messages received from Salesforce API during successful fetch operations will now be shown

---
_User Notifications_: 
N/A


amitgoldarb (commented):
Pending discussion via slack, we understood this information we are propagating could be non-intuitive for our users, hence we decided to "Allow-list" the errors we propogate. We also want to group some of the propogated messages in order for them to be less "spammy".

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -596,13 +598,20 @@ export default class SalesforceAdapter implements AdapterOperations {
...
     } = await metadataInstancesPromise
     const elements = [
       ...fieldTypes,
       ...hardCodedTypes,
       ...metadataTypes,
       ...metadataInstancesElements,
     ]
+    const metadataInstancesErrors: SaltoError[] = metadataInstancesMessages.map(
+      (message) => ({
+        severity: 'Warning',
```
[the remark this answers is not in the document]
tomsellek: Sadly, retrieve messages don't come with a severity:

```
retrieve result for types <...> : {
  done: 'true',
  id: '09SWR0000000vr72AA',
  messages: [
    {
      fileName: 'unpackaged/package.xml',
      problem: "Entity of type 'InstalledPackage' named 'usf' cannot be found"
    },
    ...
}
```

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -345,6 +351,98 @@ const getTypesWithMetaFile = async (
...
+    messageLine: (matchResult) =>
+      (matchResult &&
+        `Failed to fetch record ${matchResult.groups?.recordId} [${matchResult.groups?.type}] - duplicate developerName as ${matchResult.groups?.duplicate}`) ??
+      '',
+  },
+  unsupportedExperienceBundleTemplate: {
+    regex:
+      /ExperienceBundle Metadata API doesn't support the template of (?<name>\w+)/,
+    header:
+      "ExperienceBundle Metadata API doesn't support the template of the following records:",
```
amitgoldarb, line 407:
```suggestion
      'ExperienceBundle Metadata API doesn't support the template of the following records:',
```

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -345,6 +351,98 @@ const getTypesWithMetaFile = async (
...
+    header:
+      'The following records could not be retrieved because you do not have access to view or run the required flow types:',
+    messageLine: (matchResult) =>
+      (matchResult &&
+        `Failed to fetch record ${matchResult.groups?.recordId} [${matchResult.groups?.type}] - insufficient permissions for flows of type ${matchResult.groups?.flowType}`) ??
+      '',
+  },
+}
+
+const groupAndFilterRetrieveMessages = (
```
amitgoldarb, line 423: Can we possibly add some UT for this function? I find it a bit difficult to understand the final outcome string that we have here.

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -345,6 +351,98 @@ const getTypesWithMetaFile = async (
...
+      "ExperienceBundle Metadata API doesn't support the template of the following records:",
+    messageLine: (matchResult) =>
+      (matchResult && matchResult.groups?.name) ?? '',
+  },
+  noFlowAccess: {
+    regex:
+      /Unable to retrieve file for id (?<recordId>\w+) of type (?<type>\w+). You don't have access to view or run flows of type (?<flowType>\w+)/,
+    header:
+      'The following records could not be retrieved because you do not have access to view or run the required flow types:',
+    messageLine: (matchResult) =>
```
amitgoldarb, line 416: As these errors are customer facing, can we please ping @tomermevorach with them?

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -345,6 +351,98 @@ const getTypesWithMetaFile = async (
...
+      (matchResult && matchResult.groups?.name) ?? '',
+  },
+  duplicateNamesInPackage: {
+    regex:
+      /Duplicate object names in the same package, '(?<packageName>[^']+)'; please rename one/,
+    header: 'The following packages contain duplicate names:',
+    messageLine: (matchResult) =>
+      (matchResult && matchResult.groups?.packageName) ?? '',
+  },
+  duplicateDeveloperName: {
```
amitgoldarb, line 393: Is this a client error or something that is because we are malforming the package?

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -345,6 +351,98 @@ const getTypesWithMetaFile = async (
...
+  experienceBundleNotEnabled: {
+    regex:
+      // note the use of the Unicode apostrophe!
+      /(?<name>\w+) wasn’t retrieved because ExperienceBundle isn’t enabled for Aura sites\. To enable ExperienceBundle, in Setup, select Enable ExperienceBundle Metadata API in Digital Experiences \| Settings/,
+    header:
+      'The following records were not retrieved because ExperienceBundle is not enabled for Aura sites. To enable ExperienceBundle, in Setup, select Enable ExperienceBundle Metadata API in Digital Experiences | Settings:',
+    messageLine: (matchResult) =>
+      (matchResult && matchResult.groups?.name) ?? '',
+  },
+  duplicateNamesInPackage: {
```
amitgoldarb, line 386: Is this a client error or something that is because we are malforming the package?

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -345,6 +351,98 @@ const getTypesWithMetaFile = async (
...
   )
 
+type retrieveMessageHandler = {
+  regex: RegExp
+  header: string
+  messageLine: (matchResult: ReturnType<typeof RegExp.prototype.exec>) => string
+}
+
+const KNOWN_RETRIEVE_MESSAGES: Record<string, retrieveMessageHandler> = {
+  entityNotFound: {
```
amitgoldarb, line 361: Is this a client error or something that is because we are malforming the package?

packages/salesforce-adapter/src/fetch.ts
amitgoldarb: Nit: I think we can also extract it possibly to it's own file

amitgoldarb (commented):
Looks good! Added a few comments.

packages/salesforce-adapter/src/config_change.ts
```diff
@@ -240,9 +240,14 @@ export const createListMetadataObjectsConfigChange = (
 ): ConfigChangeSuggestion =>
   createSkippedListConfigChange({ type: res.type, instance: res.folder })
 
+type ConfigChangeCreationResult = {
+  configChanges: ConfigChangeSuggestion[]
+  errorsWithoutConfigChanges: string[]
```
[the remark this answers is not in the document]
tomsellek: Ack

packages/salesforce-adapter/src/fetch.ts
tomsellek: I don't know. Perhaps @tamtamirr does?

packages/salesforce-adapter/src/fetch.ts
tomsellek: There's an apostrophe in the string

PR #5577, opened 2024-03-12
SALTO-5263 - Salesforce: Exclude DiscoveryAIModel from Salesforce fetch

tomsellek:
Exclude this metadata type, because apparently we should?

---
Added a couple of exclusions that were mistakenly added only to the CPQ default config.

---
_Release Notes_: 
Salesforce: DiscoveryAIModel metadata type is excluded in new workspaces

---
_User Notifications_: 
Salesforce: DiscoveryAIModel metadata type is now excluded by default from new environments


tamtamirr (approved):
LGTM

PR #5581, opened 2024-03-13
SALTO-5421 support definitions in serviceUrl generic filter

omrifarkash:
Make service url filter support the new definition type

---

---
_Release Notes_: 
- Service url filter support the new definition type, should have no effect on users

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/filters/service_url.ts
+++ b/packages/adapter-components/src/filters/service_url.ts
@@ -24,6 +24,7 @@ import {
   isInstanceChange,
   isInstanceElement,
   isReferenceExpression,
+  ReadOnlyElementsSource,
   Values,
 } from '@salto-io/adapter-api'
 import { filter } from '@salto-io/adapter-utils'
@@ -38,8 +39,8 @@ const log = logger(module)
 
 export const transformConfigToFetchDefinitionsForServiceUrl = (
   configDef: TypeConfig<TransformationConfig, ActionName>,
-  additionalConfigDef?: TypeConfig<TransformationConfig, ActionName>
-  ): InstanceFetchApiDefinitions => {
+  additionalConfigDef?: TypeConfig<TransformationConfig, ActionName>,
+): InstanceFetchApiDefinitions => {
   const serviceUrl = additionalConfigDef?.transformation?.serviceUrl ?? configDef.transformation?.serviceUrl
   return {
     element: {
@@ -51,10 +52,11 @@ export const transformConfigToFetchDefinitionsForServiceUrl = (
   }
 }
 
-export const addUrlToInstance: <
-ClientOptions extends string = 'main'
->(instance: InstanceElement, baseUrl: string, apiDef: InstanceFetchApiDefinitions<ClientOptions> | undefined) => void =
-(instance, baseUrl, apiDef) => {
+export const addUrlToInstance: <ClientOptions extends string = 'main'>(
+  instance: InstanceElement,
+  baseUrl: string,
+  apiDef: InstanceFetchApiDefinitions<ClientOptions> | undefined,
+) => void = (instance, baseUrl, apiDef) => {
   const serviceUrl = apiDef?.element?.topLevel?.serviceUrl
   if (serviceUrl === undefined) {
     return
@@ -75,16 +77,16 @@ ClientOptions extends string = 'main'
 }
 
 export const serviceUrlFilterCreator: <
-TContext,
-TResult extends void | filter.FilterResult = void,
-TAdditional = {},
-ClientOptions extends string = 'main',
-PaginationOptions extends string | 'none' = 'none',
-Action extends string = ActionName,
+  TContext,
+  TResult extends void | filter.FilterResult = void,
+  TAdditional = {},
+  ClientOptions extends string = 'main',
+  PaginationOptions extends string | 'none' = 'none',
+  Action extends string = ActionName,
 >(
   baseUrl: string,
 ) => AdapterFilterCreator<TContext, TResult, TAdditional, ClientOptions, PaginationOptions, Action> =
-  (baseUrl) =>
+  baseUrl =>
   ({ definitions }) => {
     if (definitions.fetch === undefined) {
       log.warn('No fetch definitions were found, skipping service url filter')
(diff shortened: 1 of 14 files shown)


omrifarkash:
Have been tested, and no side effect detected

packages/adapter-components/src/filters/service_url.ts
```diff
@@ -23,18 +24,44 @@ import {
...
+export const addUrlToInstance: <ClientOptions extends string = 'main'>(
+  instance: InstanceElement,
+  baseUrl: string,
+  apiDef: InstanceFetchApiDefinitions<ClientOptions> | undefined,
+) => void = (instance, baseUrl, apiDef) => {
+  const serviceUrl = apiDef?.element?.topLevel?.serviceUrl
   if (serviceUrl === undefined) {
     return
   }
+  // TODO_F use custom function from serviceUrl SALTO-5580
```
netama, line 64: if you want to merge this let's make it a "regular" TODO (or just remove)?

packages/adapter-components/src/filters/service_url.ts
```diff
@@ -45,38 +72,73 @@ export const addUrlToInstance = (instance: InstanceElement, baseUrl: string, api
...
+        config.apiDefinitions.types[typeName],
+        additionalApiDefinitions?.types[typeName],
+      ),
+    ]),
+  )
+
+  const definitions = {
+    fetch: {
+      instances: {
+        customizations,
```
netama, line 139: there's also a `typeDefaults` in the old config which is equivalent to `default`, let's add it as well (maybe also add a test)? (e.g. maybe there's a shared `/view/{id}` that is shared across all types)

packages/adapter-components/src/filters/service_url.ts
```diff
@@ -45,38 +72,73 @@ export const addUrlToInstance = (instance: InstanceElement, baseUrl: string, api
...
+  ])
+  const customizations = Object.fromEntries(
+    Array.from(typeNamesSet).map(typeName => [
+      typeName,
+      configDefToInstanceFetchApiDefinitionsForServiceUrl(
+        config.apiDefinitions.types[typeName],
+        additionalApiDefinitions?.types[typeName],
+      ),
+    ]),
+  )
```
netama, line 134: I think this can be simpler - e.g.
```suggestion
  const customizations = _({ ...config.apiDefinitions.types, ...additionalApiDefinitions?.types,})
    .mapValues(configDefToInstanceFetchApiDefinitionsForServiceUrl)
    .pickBy(lowerdashValues.isDefined)
    .value()
```
and we can have `configDefToInstanceFetchApiDefinitionsForServiceUrl` accept a single arg. wdyt?
(but maybe I'm missing something)

packages/adapter-components/src/filters/service_url.ts
```diff
@@ -23,18 +24,44 @@ import {
...
 import { createUrl } from '../fetch/resource'
+import { ApiDefinitions, queryWithDefault } from '../definitions'
+import { InstanceFetchApiDefinitions } from '../definitions/system/fetch'
+import { AdapterApiConfig, TransformationConfig, TypeConfig } from '../config'
 
-export const addUrlToInstance = (instance: InstanceElement, baseUrl: string, apiDefs: AdapterApiConfig): void => {
-  const serviceUrl = apiDefs.types[instance.elemID.typeName]?.transformation?.serviceUrl
+const log = logger(module)
+
+export const configDefToInstanceFetchApiDefinitionsForServiceUrl = (
```
netama, line 40: continuing from the comment below, I think we can make this function accept only a single param

packages/adapter-components/src/filters/service_url.ts
```diff
@@ -23,18 +24,44 @@ import {
...
+  configDef?: TypeConfig<TransformationConfig, ActionName>,
+  additionalConfigDef?: TypeConfig<TransformationConfig, ActionName>,
+): InstanceFetchApiDefinitions => {
+  const serviceUrl = additionalConfigDef?.transformation?.serviceUrl ?? configDef?.transformation?.serviceUrl
+  return {
+    element: {
+      topLevel: {
+        isTopLevel: true as const,
+        serviceUrl: serviceUrl ? { path: serviceUrl } : undefined,
```
netama, line 49: if there's no service url maybe just return `undefined` from the entire function?

packages/jira-adapter/src/change_validators/workflow_scheme_migration.ts
```diff
@@ -259,7 +259,11 @@ export const workflowSchemeMigrationValidator =
       .map(async change => {
         await updateSchemeId(change, client, paginator, config)
         const instance = getChangeData(change)
-        addUrlToInstance(instance, client.baseUrl, config.apiDefinitions)
+        addUrlToInstance(
+          instance,
+          client.baseUrl,
+          configDefToInstanceFetchApiDefinitionsForServiceUrl(config.apiDefinitions.types[instance.elemID.typeName]),
```
netama, line 265: you're losing the `typeDefaults` def here as well, please use it as a fallback

packages/adapter-components/src/filters/service_url.ts
```diff
@@ -45,38 +72,73 @@ export const addUrlToInstance = (instance: InstanceElement, baseUrl: string, api
...
+
+  const definitions = {
+    fetch: {
+      instances: {
+        customizations,
+      },
     },
-  })
+  } as ApiDefinitions
+  return serviceUrlFilterCreator(baseUrl)({ ...args, definitions, elementSource: {} as ReadOnlyElementsSource })
```
netama, line 143: maybe 
```suggestion
  return serviceUrlFilterCreator(baseUrl)({ ...args, definitions, elementSource: buildElementsSourceFromElements([])
```

netama (commented):
looks good, added some comments - mainly around simplifying, but also making sure we don't lose the default in the upgrades

packages/adapter-components/src/filters/service_url.ts
omrifarkash: You're not missing anything, just using a lot of magic that I wasn't familiar with :)

packages/jira-adapter/src/change_validators/workflow_scheme_migration.ts
omrifarkash: I think we haven't used it before either.
I can start using it as a fallback but it might change the current behavior

image

packages/adapter-components/src/filters/service_url.ts
omrifarkash: I think the old code hasn't used typeDefaults

packages/adapter-components/src/filters/service_url.ts
```diff
@@ -25,13 +26,40 @@ import {
...
 
-export const addUrlToInstance = (instance: InstanceElement, baseUrl: string, apiDefs: AdapterApiConfig): void => {
-  const serviceUrl = apiDefs.types[instance.elemID.typeName]?.transformation?.serviceUrl
+const log = logger(module)
+
+export const configDefToInstanceFetchApiDefinitionsForServiceUrl = (
+  configDef?: TypeConfig<TransformationConfig, ActionName>,
+): InstanceFetchApiDefinitions | undefined => {
+  const serviceUrl = configDef?.transformation?.serviceUrl
+  return serviceUrl
```
netama, line 45: nit: 
```suggestion
  return serviceUrl !== undefined
```

packages/adapter-components/src/filters/service_url.ts
netama: sorry my bad. so let's add a comment explaining it wasn't supported?

netama (approved):

PR #5582, opened 2024-03-13
SALTO-5498: [Zendesk] Move reference creation for macro attachment to macro filter

DanielBarros1:
This is meant to avoid an id clash that happens when attachments from different macros have the same ID

---

---
_Release Notes_: 

Zendesk:
Fix bug where macro attachments had clashing IDs on fetch

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/filters/macro_attachments.ts
+++ b/packages/zendesk-adapter/src/filters/macro_attachments.ts
@@ -244,13 +244,16 @@ const getMacroAttachments = async ({
   if (!isAttachments(attachments)) {
     return []
   }
-  return (
+  const attachmentsContent = (
     await Promise.all(
       attachments.map(async attachment => getAttachmentContent({ client, attachment, macro, attachmentType })),
     )
   )
     .flat()
     .filter(values.isDefined)
+
+  macro.value.attachments = attachmentsContent.map(attachment => new ReferenceExpression(attachment.elemID, attachment))
+  return attachmentsContent
 }
 
 /**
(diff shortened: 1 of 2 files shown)


packages/zendesk-adapter/test/filters/macro_attachments.test.ts
edenhassid, line 1: can you add a test that would fail with the previous code but will succeed with this one?

packages/zendesk-adapter/src/filters/macro_attachments.ts
edenhassid, line 1: I think we should change the name `getMacroAttachments` cause it also updates the macros now, or we can move the logic of updating the macro itself to be elsewhere.

edenhassid (commented):
great! 👏 💯 🔥  added some comments :) did you also check that the deploy works and that the references are resolved to ids correctly?

packages/zendesk-adapter/src/filters/macro_attachments.ts
```diff
@@ -244,13 +244,16 @@ const getMacroAttachments = async ({
...
-  return (
+  const attachmentsContent = (
     await Promise.all(
       attachments.map(async attachment => getAttachmentContent({ client, attachment, macro, attachmentType })),
     )
   )
     .flat()
     .filter(values.isDefined)
+
+  macro.value.attachments = attachmentsContent.map(attachment => new ReferenceExpression(attachment.elemID, attachment))
```
adfineberg, line 255: Maybe verify that all the original ids get attachments? Wouldn't want to mistakenly delete an ID if the API call failed or something

DanielBarros1:
> great! 👏 💯 🔥 added some comments :) did you also check that the deploy works and that the references are resolved to ids correctly?

yup on both

packages/zendesk-adapter/src/filters/macro_attachments.ts
DanielBarros1: I'll change the name :)

PR #5583, opened 2024-03-13
SALTO-5585 - Salesforce: Log bulk load results

tomsellek:
We're already doing this for metadata, let's do it for data as well

---
```
2024-03-13T13:39:49.877Z trace salesforce-adapter/client/client client.bulkLoadOperation result: [
  { id: '001Dn0000063rYsIAI', success: true, errors: [ [length]: 0 ] },
  [length]: 1
]
```

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A


amitgoldarb (approved):
Thanks for this!

PR #5585, opened 2024-03-13
SALTO-5576: Modify default page size in Okta

shir-reifenberg:
Until now we used the maximum page size for specific endpoint, I switched it to be used by default

---

- Not all endpoints support page size but it has no affect for unsupported endpoints
- I prefer to put this as default as every now and then we find out we can increase page size for another type (usually default is 20 or 50)
- groups maximum page size is different from all other endpoints and it defined to be 10000 (which is the default one, but now that we override the default it needs to be mentioned explicitly)

---
_Release Notes_: 
None

---
_User Notifications_: 
None


idoamit8 (approved):
Great ! 
LGTM 🥇

PR #5590, opened 2024-03-14
SALTO-5523: add template staticFile to dummy

edenhassid:
1. add template staticFile to dummy
2. add template expression to dummy

---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/dummy-adapter/test/adapter.test.ts
+++ b/packages/dummy-adapter/test/adapter.test.ts
@@ -142,4 +142,21 @@ describe('dummy adapter', () => {
       expect(await adapterWithDeployModifiers.deployModifiers.changeValidator?.([myInst1Change])).toHaveLength(0)
     })
   })
+  describe('fixElements', () => {
+    it('should return correct value', async() => {
+      const mockReporter = { reportProgress: jest.fn() }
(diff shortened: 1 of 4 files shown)


packages/dummy-adapter/test/adapter.test.ts
```diff
@@ -142,4 +142,21 @@ describe('dummy adapter', () => {
       expect(await adapterWithDeployModifiers.deployModifiers.changeValidator?.([myInst1Change])).toHaveLength(0)
     })
   })
+  describe('fixElements', () => {
```
edenhassid, line 145: added this test to increase coverage

adamro (approved):
Beyond what you did, we might wanna consider adding this to the data in `data/fixtures` for the `generateExtraElements` flow that we use a lot for testing (making sure it works properly + possibly adding an example now)

edenhassid:
> Beyond what you did, we might wanna consider adding this to the data in `data/fixtures` for the `generateExtraElements` flow that we use a lot for testing (making sure it works properly + possibly adding an example now)
> 
> and, make this pretty :)

I think it will be better to add a template staticFile to fixtures when we have a ui for it.

PR #5592, opened 2024-03-15
SALTO-5602 - Dummy: Support loadElementsFromFolder

tomsellek:
Support `loadElementsFromFolder` in dummy adapter as a first step towards running "Deploy from PR" tests using the dummy adapter instead of a real adapter

---

N/A

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/dummy-adapter/src/adapter_creator.ts
+++ b/packages/dummy-adapter/src/adapter_creator.ts
@@ -13,6 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import fs from 'fs'
+import _ from 'lodash'
 import {
   Adapter,
   ElemID,
@@ -21,8 +23,10 @@ import {
   ObjectType,
   ListType,
   GetCustomReferencesFunc,
+  FetchResult,
+  LoadElementsFromFolderArgs,
+  ProgressReporter,
 } from '@salto-io/adapter-api'
-import _ from 'lodash'
 import DummyAdapter from './adapter'
 import { GeneratorParams, DUMMY_ADAPTER, defaultParams, changeErrorType } from './generator'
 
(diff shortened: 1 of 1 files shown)


packages/dummy-adapter/src/adapter_creator.ts
```diff
@@ -56,6 +60,25 @@ const getCustomReferences: GetCustomReferencesFunc = async elements =>
...
+  const otherConfigPath = `${baseDir}/dummy.nacl`
+  let otherConfig: Record<string, unknown>
+
+  try {
+    otherConfig = JSON.parse(fs.readFileSync(otherConfigPath, { encoding: 'utf8' }))
+  } catch (err) {
+    throw new Error(`Failed to read config file for loadElementsFromFolder in ${otherConfigPath}: ${err.message}`)
+  }
+
+  const otherAdapter = new DummyAdapter(otherConfig as GeneratorParams)
```
omrifarkash, line 78: I think we should think of a way to verify that `otherConfig` is a valid Dummy configuration

omrifarkash (commented):
The code looks fine to me.

I have another suggestion, instead of having a dummy configuration nacl saved in some folder (probably remotely) we can have the entire workspace (nacl files) save in the remote and just parse the nacl files when calling `loadElementsFromWorkspace`

This way we can make any change we want in the after branch and not relying only on dummy configuration features for creating changes (note that we can easily using dummy configuration features this way, by fetching with some configuration and push the result nacls to a remote repo). 

I vote to whatever you think is simpler to implement and to use
WDYT?
cc: @amitgoldarb @adamro @tomsellek

tomsellek:
@omrifarkash I believe I reimplemented this per your suggestion. Could you please re-review?

PR #5594, opened 2024-03-17
SALTO-4864: Enable the "remove" deploy action for `ProfileMapping`

Nurdok:
Enable the "remove" deploy action for `ProfileMapping`.

`ProfileMapping`s are removed automatically by Okta when either side of the mapping is removed, and there's no endpoint in Okta's API to manually remove mappings.
Prior to this PR, `ProfileMapping` instances did not support the "remove" action at all and any deploy attempt would result in an error. This PR allows `ProfileMapping`s to be manually removed as long as either side of the mapping is removed in the same deploy action. There's no API call happening, just marking the mapping as removed. This PR also makes the mapping "removal" dependent on the mapping side(s) being removed, to make sure we'll only report success on the mapping removal if the side removal succeeded.

---

Note that PR #5565 added a CV that blocks standalone `ProfileMapping` removal as preparation for this PR.

---
_Release Notes_: 
_Okta Adapter_:
- `ProfileMapping` now supports removal action in deployment.
---
_User Notifications_: 
None


packages/okta-adapter/src/config.ts
```diff
@@ -1311,7 +1311,17 @@ const DEFAULT_TYPE_CUSTOMIZATIONS: OktaSwaggerApiConfig['types'] = {
...
       },
-      // TODO SALTO-4769 add support in remove
+      remove: {
+        // ProfileMappings are removed automatically by Okta when either side of the mapping is removed.
+        // We use an empty URL here to mark this action as supported in case a user removed the mapping
+        // alongside either side.
+        // A separate Change Validator ensures that mappings aren't removed by themselves.
+        url: '',
+        method: 'delete',
+        urlParamsToFields: {
```
shir-reifenberg, line 1321: I think `urlParamsToFields` is redundant

packages/okta-adapter/src/filters/profile_mapping_removal.ts
```diff
@@ -0,0 +1,40 @@
...
+      changes,
+      change =>
+        isInstanceChange(change) &&
+        isRemovalChange(change) &&
+        getChangeData(change).elemID.typeName === PROFILE_MAPPING_TYPE_NAME,
+    )
+    return {
+      leftoverChanges,
+      // Mark the removal as success
+      deployResult: { errors: [], appliedChanges: relevantChanges },
```
shir-reifenberg, line 35: I know I initially said it's ok to just mark it as a success, but I think it's safer to make a `get` request to validate it indeed doesn't exist anymore (returns 404 status code). At start I thought we're gonna make the deploy plan to be visible to the adapter but seems like it's not going to happen soon..

you can do something similar to `verifyProjectDeleted` (in Jira)

packages/okta-adapter/src/dependency_changers/remove_profile_mapping_after_deps.ts
```diff
@@ -0,0 +1,64 @@
...
+} from '@salto-io/adapter-api'
+import { deployment } from '@salto-io/adapter-components'
+import { PROFILE_MAPPING_TYPE_NAME } from '../constants'
+
+/**
+ * Remove ProfileMapping only *after* one of its dependencies is removed.
+ *
+ * ProfileMappings are removed automatically by Okta when either side of the mapping is removed.
+ * The actual remove deploy for ProfileMapping does nothing - this just makes sure that we report success on its
+ * removal only after removing the dependencies (otherwise it will still exist in Okta).
```
shir-reifenberg, line 34: I think this is more relevant to the filter, and you can move the explanation from line 56 to here :)

packages/okta-adapter/src/dependency_changers/remove_profile_mapping_after_deps.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  const profileMappingRemovals = removals.filter(
+    change => getChangeData(change.change).elemID.typeName === PROFILE_MAPPING_TYPE_NAME,
+  )
+
+  return profileMappingRemovals.flatMap(profileMappingRemoval => {
+    const { source, target } = getChangeData(profileMappingRemoval.change).value
+    return removals
+      .filter(removal =>
+        [source.id.elemID.getFullName(), target.id.elemID.getFullName()].includes(
```
shir-reifenberg, line 52: can't `source` / `target` be undefined? should we verify `source.id` is a reference ?

shir-reifenberg (commented):
Looks great! 🚀  added some comments

packages/okta-adapter/src/filters/profile_mapping_removal.ts
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Override the default ProfileMapping removal with a verification of removal.
+ *
+ * ProfileMappings are removed automatically by Okta when either side of the mapping is removed, so instead only
+ * verify that they are removed.
+ * Separate change validator and dependency changer ensure that this is only executed if one of the mapping side was
+ * removed in the same deploy action.
+ */
+const filterCreator: FilterCreator = ({ client }) => ({
+  // If a ProfileMapping is removed, mark the result as success without actually using the endpoint.
```
shir-reifenberg, line 50: can remove now, no?

packages/okta-adapter/src/filters/profile_mapping_removal.ts
```diff
@@ -0,0 +1,74 @@
...
+ */
+import _ from 'lodash'
+import { isInstanceChange, getChangeData, isRemovalChange } from '@salto-io/adapter-api'
+import { client as clientUtils } from '@salto-io/adapter-components'
+import { FilterCreator } from '../filter'
+import { PROFILE_MAPPING_TYPE_NAME } from '../constants'
+import { deployChanges } from '../deployment'
+import OktaClient from '../client/client'
+
+const verifyProfileMappingIsDeleted = async (profileMappingId: number, client: OktaClient): Promise<boolean> => {
```
shir-reifenberg, line 24: isn't `profileMappingId` a string ?

packages/okta-adapter/src/filters/profile_mapping_removal.ts
```diff
@@ -0,0 +1,74 @@
...
+  try {
+    return (
+      (
+        await client.get({
+          url: `/api/v1/mappings/${profileMappingId}`,
+        })
+      ).status === 404
```
shir-reifenberg, line 31: can it happen ? should we just return `false` if we get a valid response ?

shir-reifenberg (approved):
Added a few questions but nothing major 🔥

PR #5598, opened 2024-03-17
SALTO-5613 make the customizations optional in defaultWithCustomizations type

Uri-Bar:
SALTO-5613 make the customizations optional in defaultWithCustomizations type
---

---
_Release Notes_: 
none

---
_User Notifications_: 
none

### DIFF
--- a/packages/adapter-components/src/definitions/system/utils.ts
+++ b/packages/adapter-components/src/definitions/system/utils.ts
@@ -71,10 +71,7 @@ export const queryWithDefault = <T, K extends string = string>(
   defsWithDefault: DefaultWithCustomizations<T, K>,
 ): DefQuery<T, K> => {
   const query: DefQuery<T, K>['query'] = key =>
-    mergeSingleDefWithDefault(
-      defsWithDefault.default,
-      defsWithDefault.customizations ? defsWithDefault.customizations[key] : undefined,
-    )
+    mergeSingleDefWithDefault(defsWithDefault.default, defsWithDefault.customizations?.[key] ?? undefined)
   return {
     query,
     allKeys: () => Object.keys(defsWithDefault.customizations ?? {}) as K[],
@@ -131,8 +128,7 @@ export const mergeWithUserElemIDDefinitions = <Options extends FetchApiDefinitio
   return _.merge({}, fetchConfig, {
     instances: {
       customizations: _.mapValues(userElemID, ({ extendSystemPartsDefinition, ...userDef }, type) => {
-        const customizations = fetchConfig?.instances.customizations ?? {}
-        const { elemID: systemDef } = customizations[type]?.element?.topLevel ?? {}
+        const { elemID: systemDef } = fetchConfig?.instances.customizations?.[type]?.element?.topLevel ?? {}
         const elemIDDef = {
           ..._.defaults({}, userDef, systemDef),
           parts: extendSystemPartsDefinition ? (systemDef?.parts ?? []).concat(userDef.parts ?? []) : userDef.parts,
--- a/packages/adapter-components/src/definitions/system/utils.ts
+++ b/packages/adapter-components/src/definitions/system/utils.ts
@@ -71,7 +71,10 @@ export const queryWithDefault = <T, K extends string = string>(
   defsWithDefault: DefaultWithCustomizations<T, K>,
 ): DefQuery<T, K> => {
   const query: DefQuery<T, K>['query'] = key =>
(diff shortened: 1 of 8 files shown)


packages/adapter-components/src/definitions/system/utils.ts
```diff
@@ -71,7 +71,10 @@ export const queryWithDefault = <T, K extends string = string>(
   defsWithDefault: DefaultWithCustomizations<T, K>,
 ): DefQuery<T, K> => {
   const query: DefQuery<T, K>['query'] = key =>
-    mergeSingleDefWithDefault(defsWithDefault.default, defsWithDefault.customizations[key])
+    mergeSingleDefWithDefault(
+      defsWithDefault.default,
+      defsWithDefault.customizations ? defsWithDefault.customizations[key] : undefined,
```
netama, line 76: nit: I think you can just
```suggestion
      defsWithDefault.customizations?.[key],
```

packages/adapter-components/src/definitions/system/utils.ts
```diff
@@ -118,7 +121,8 @@ export const mergeWithUserElemIDDefinitions = <Options extends FetchApiDefinitio
   userElemID: UserFetchConfig<Options>['elemID']
   fetchConfig?: FetchApiDefinitions<Options>
 }): FetchApiDefinitions<Options> => {
-  if (userElemID === undefined) {
+  const customizations = fetchConfig?.instances.customizations
+  if (userElemID === undefined || costumizations === undefined) {
     return (
       fetchConfig ?? {
         instances: { customizations: {} },
```
netama, line 128: nit: there's no point defining `customizations` here if it's not required and empty, right?

EDIT: I don't think we should get into this case, we should get the elem id from the user if the user defined it. what am I missing?

packages/adapter-components/test/deployment/deploy/graph.test.ts
```diff
@@ -105,6 +99,9 @@ describe('createDependencyGraph', () => {
   describe('with custom actions and dependencies', () => {
     let graph: DAG<NodeType<'activate' | 'deactivate'>>
     beforeEach(() => {
+      if (!deployDef.instances.customizations) {
+        deployDef.instances.customizations = {}
```
netama, line 103: not sure I understand why this is done in a test? if it's valid as a definition then the code should be able to handle it, no?

packages/adapter-components/src/definitions/system/utils.ts
Uri-Bar: typescript didnt like it

packages/adapter-components/src/definitions/system/utils.ts
Uri-Bar: this is just type stuff for the test, ts wont allow me touch customizations.typeB otherwise.
because it in the test I dont think it is critical

packages/adapter-components/test/deployment/deploy/graph.test.ts
Uri-Bar: this is just type stuff for the test, ts wont allow me touch customizations.typeB otherwise.
because it in the test I dont think it is critical

packages/adapter-components/src/definitions/system/utils.ts
Uri-Bar: I think you are righ

packages/adapter-components/src/definitions/system/utils.ts
netama: weird, it's working for me (notice you need `?.` and not `?`)

packages/adapter-components/src/definitions/system/utils.ts
netama: same as above (unless I'm missing something :) ) - you need `?.[...]` and not `?[...]` and then I _think_ it will work?

packages/adapter-components/src/definitions/system/utils.ts
Uri-Bar: my bad, fixed it

packages/adapter-components/test/deployment/deploy/graph.test.ts
netama: I'm worried this will mean we don't have tests that only have defaults without customizations. let's pick some use case and add an equivalent test without customizations and only with defaults?

netama (approved):
LGTM 🎈 , one remaining comment on the tests

PR #5599, opened 2024-03-18
SALTO-5596: Classify NS testing accounts as non-production

omrilit:

---

---
_Release Notes_: 
NetSuite Adapter:
SALTO-5596: do not classify testing accounts as production

---
_User Notifications_: 


shoshana-houri (approved):
Looks good to me 🤓

PR #5602, opened 2024-03-18
SALTO-5609 - Salesforce: Exclude SF translations by default

tomsellek:
We've seen errors associated with this type, and it is not used by customers.

---

---
_Release Notes_: 
Salesforce: The 'Translations' type is now excluded from fetches in new workspaces.

---
_User Notifications_: 
N/A


amitgoldarb (approved):
Thanks for this one!

PR #5609, opened 2024-03-19
SALTO-4832 - Salesforce: handle the case where SF returns an HTML page as an error

tomsellek:
Sometimes Salesforce returns a whole HTML page when it encounters an internal connection error. Let's handle this.

---

---
_Release Notes_: 
Salesforce: better handling of internal Salesforce errors.

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/client/user_facing_errors.ts
+++ b/packages/salesforce-adapter/src/client/user_facing_errors.ts
@@ -24,6 +24,7 @@ import {
   ERROR_PROPERTIES,
   INVALID_GRANT,
   isSalesforceError,
+  SALESFORCE_CONNECTION_ERROR,
   SALESFORCE_ERRORS,
   SalesforceError,
 } from '../constants'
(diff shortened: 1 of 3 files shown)


packages/salesforce-adapter/src/client/user_facing_errors.ts
```diff
@@ -73,6 +78,11 @@ const withSalesforceError = (
 ): string => `${saltoErrorMessage}\n\nUnderlying Error: ${salesforceError}`
 
 export const ERROR_MAPPERS: ErrorMappers = {
+  [SALESFORCE_CONNECTION_ERROR]: {
```
adamro, line 81: Why did we do something new for this and not include it as an OR in `ERROR_HTTP_502`?

adamro (approved):

packages/salesforce-adapter/src/client/user_facing_errors.ts
tomsellek: Just because it's a different "symptom". I could definitely merge the two.

PR #5584, opened 2024-03-13
SALTO-5414 support polling as a capability in the new infra

Uri-Bar:
SALTO-5414 support polling as a capability in the new infra
---

_Additional context for reviewer_
added the polling option for both deploy + fetch
completed the support in additional status codes for deploy

---
_Release Notes_: 
none

---
_User Notifications_: 
none

PR #5611, opened 2024-03-19
fix build

netama:
looks like https://github.com/salto-io/salto/pull/5584 was merged without a rebase

---
_Release Notes_: 
None

---
_User Notifications_: 
None


Uri-Bar (approved):
thanks

PR #5614, opened 2024-03-20
SALTO-5641: add data and headers to log when http request fails

edenhassid:
add data and headers to log when http request fails

---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/adapter-components/src/client/http_client.ts
+++ b/packages/adapter-components/src/client/http_client.ts
@@ -15,7 +15,7 @@
  */
 import _ from 'lodash'
 import { ResponseType } from 'axios'
-import {inspectValue, safeJsonStringify} from '@salto-io/adapter-utils'
+import { safeJsonStringify } from '@salto-io/adapter-utils'
 import { values } from '@salto-io/lowerdash'
 import { Values } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
--- a/packages/adapter-components/src/client/http_client.ts
+++ b/packages/adapter-components/src/client/http_client.ts
@@ -15,7 +15,7 @@
  */
 import _ from 'lodash'
 import { ResponseType } from 'axios'
-import { safeJsonStringify } from '@salto-io/adapter-utils'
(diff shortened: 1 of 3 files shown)


packages/adapter-components/src/client/http_client.ts
```diff
@@ -288,8 +288,9 @@ export abstract class AdapterHTTPClient<TCredentials, TRateLimitConfig extends C
         headers: this.extractHeaders(res.headers),
       }
     } catch (e) {
+      const dataToLog = Buffer.isBuffer(data) ? `<omitted buffer of length ${data.length}>` : data
       log.warn(
-        `failed to ${method} ${url} ${safeJsonStringify(queryParams)}: ${e}, data: ${safeJsonStringify(e?.response?.data)}, stack: ${e.stack}`,
+        `failed to ${method} ${url} ${safeJsonStringify(queryParams)}: ${e}, data: ${safeJsonStringify(e?.response?.data ?? dataToLog)}, headers: ${safeJsonStringify(this.extractHeaders(headers))} stack: ${e.stack}`,
```
netama, line 293: we log twice now if there is an `e.response`, no? any chance we can use `logResponse` and pass in the relevant args below instead? e.g.
```
logResponse({ data: params.data, headers: this.extractHeaders(headers), status: '<unknown>' })
```

netama (approved):
approving in case this is blocking, added a small suggestion (we will need to clean up all of these logs a bit more soon so not critical)

packages/adapter-components/src/client/http_client.ts
edenhassid: I fist thought about it but in logResponse we have logs that specifically write `Full HTTP response` and  `Received response for` and its to really true in this case. And even if I use  logResponse it will still be logged twice. I can move the log to be right before we throw the error. WDYT?

netama: the other option is to pass in an optional error and adjust the logs accordingly, might be easier IRL :)

netama (approved):

PR #5616, opened 2024-03-20
SALTO-5388 fix omitting values of overridden types

netama:
* make sure predefined types are used for field type overrides
* first override fields types and then remove omitted field values, to fix a bug where we didn't omit nested values of overwritten fields

(this is in the new infra, not impacting existing adapters)

---
_Release Notes_: 
None

---
_User Notifications_: 
None


shir-reifenberg (approved):
👑 🚀

PR #5617, opened 2024-03-20
SALTO-5642: Import OpenAPI / Swagger types

shir-reifenberg:
Import types from OpenAPI / swagger spec

---

This PR add an initial support for fetching types from OpenAPI / swagger.
We parse all `schemas` and convert them to types, then filter only for types and subtypes used by instances
I still need to add UT for `getReachableTypes`

Still not supported - 
- support `prefix`
- support converting response schemas which doesn't have their own schema defined

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/adapter-components/src/elements/swagger/type_elements/element_generator.ts
```diff
@@ -317,3 +319,61 @@ export const generateTypes = async (
...
+export const generateOpenApiTypes = async ({
+  adapterName,
+  openApiDefs,
+}: {
+  adapterName: string
+  openApiDefs: Omit<OpenAPIDefinition<never>, 'toClient'>
+}): Promise<TypeMap> => {
+  const { url: swaggerPath, endpointsOnly } = openApiDefs
+  if (endpointsOnly) {
+    log.debug('Skipping types generation for adapter %s, as only endpoints are required', adapterName)
```
netama, line 335: nits: 
1. maybe also log the url, in case there are multiple imported swaggers?
2. let's mention `openAPI` in the log?

packages/adapter-components/src/elements/swagger/type_elements/element_generator.ts
```diff
@@ -317,3 +319,61 @@ export const generateTypes = async (
     parsedConfigs,
   }
 }
+
+/**
```
netama, line 323: doesn't have to be in this PR, but the `elements` folder will soon be removed and shouldn't contain "new infra" logic, eventually this will need to move to the `fetch/element` folder or to a dedicated `openAPI` folder (the latter is better IMO).

packages/adapter-components/src/fetch/element/type_utils.ts
```diff
@@ -300,3 +305,41 @@ export const hideAndOmitFields = <Options extends FetchApiDefinitionsOptions>({
...
+export const getReachableTypes = <Options extends FetchApiDefinitionsOptions>({
+  instances,
+  types,
+  defQuery,
+}: {
+  instances: InstanceElement[]
+  types: ObjectType[]
+  defQuery: ElementAndResourceDefFinder<Options>
+}): ObjectType[] => {
+  const topLevelTypeNames = new Set(defQuery.allKeys().concat(instances.map(inst => inst.elemID.typeName)))
```
netama, line 321: nit: `topLevel` is a little misleading here since we're not looking at the `topLevel` attribute in the definitions, maybe use something like `root<>` if the goal is to say these are the types used for the traversal?

packages/adapter-components/src/fetch/element/type_utils.ts
```diff
@@ -300,3 +305,41 @@ export const hideAndOmitFields = <Options extends FetchApiDefinitionsOptions>({
...
+
+  const topLevelTypes = types.filter(type => topLevelTypeNames.has(type.elemID.name))
+  const nestedTypesToReturn: Set<string> = new Set()
+
+  topLevelTypes.forEach(type => {
+    walkOnElement({
+      element: type,
+      func: ({ value }) => {
+        if (isField(value)) {
+          const fieldType = getDeepInnerTypeSync(value.getTypeSync())
```
netama, line 331: I think there are potential endless recursions here, e.g. if a type refers to itself (even directly as a `next` field or within a list/map, but also with more elaborate cycles). let's add a test for it?
(and then a follow-up - if we check in every step, we can probably look in a single set instead of 2 if we add the type itself when starting the walk)

packages/adapter-components/src/elements/swagger/type_elements/element_generator.ts
```diff
@@ -147,10 +149,10 @@ const typeAdder = ({
     definedTypes[naclObjName] = type
 
     const { allProperties, additionalProperties } = extractProperties(schemaDef, refs)
-
+    const naclCasedProperties = _.mapKeys(allProperties, (_value, key) => naclCase(key))
```
netama, line 152: thanks - this can have side-effects in existing adapters, let's map them out and see if we should add a flag for the new behavior to simplify upgrades?

packages/adapter-components/src/fetch/element/type_utils.ts
```diff
@@ -300,3 +305,41 @@ export const hideAndOmitFields = <Options extends FetchApiDefinitionsOptions>({
     })
   })
 }
+
+/**
+ * Filter for types that are either used by instance or defined in fetch definitions
+ */
+export const getReachableTypes = <Options extends FetchApiDefinitionsOptions>({
```
netama, line 312: added some comments below, but looking more - is there a reason not to use the existing `getSubtypes` for the traversal (and pass in the `topLevelTypes`)? we can probably make it run sync but other than that is there new logic here?
(also please notice that the `additionalProperties` logic for extracting fields from the annotation was merged and isn't covered here, was probably added after you implemented this)

netama (commented):
thanks, cool :) added some comments

packages/adapter-components/src/fetch/element/type_utils.ts
shir-reifenberg: missed that, will reuse

packages/core/src/core/adapters/adapters.ts
```diff
@@ -89,19 +88,17 @@ const getAdapterConfigFromType = async (adapterName: string): Promise<InstanceEl
...
-        adapterCreator.configType
-          ? [adapterCreator.configType, ...(await getSubtypes([adapterCreator.configType], true))]
-          : [],
+      _.mapValues(adapterCreators, adapterCreator =>
+        adapterCreator.configType ? [adapterCreator.configType, ...getSubtypes([adapterCreator.configType], true)] : [],
       ),
     ).filter(entry => entry[1].length > 0),
   )
 
 export const getAdaptersConfigTypes = async (): Promise<ObjectType[]> =>
```
shir-reifenberg, line 100: left async cause it's exported from core and being currently used

packages/adapter-components/src/elements/swagger/type_elements/element_generator.ts
```diff
@@ -71,20 +71,22 @@ type TypeAdderType = (
...
  */
-const typeAdder = ({
+export const typeAdder = ({
   adapterName,
   schemas,
   toUpdatedResourceName,
   definedTypes,
   parsedConfigs,
   refs,
+  naclCaseFields = false,
```
netama, line 81: nit: maybe default to true and set to false for existing adapters, so that the default usage is correct? technically false could result in broken elements

packages/adapter-components/src/fetch/element/openAPI/types.ts
```diff
@@ -0,0 +1,106 @@
...
+import { defineAdditionalTypes } from '../../../elements/swagger/type_elements/type_config_override'
+import { typeAdder } from '../../../elements/swagger/type_elements/element_generator'
+import { ElementAndResourceDefFinder } from '../../../definitions/system/fetch/types'
+import { FetchApiDefinitionsOptions } from '../../../definitions/system/fetch'
+import { TYPES_PATH, SUBTYPES_PATH } from '../../../elements/constants'
+
+const log = logger(module)
+
+/**
+ * openAPI types are created by under the subtypes folder, modify path of top level instances to be under the types folder
```
netama, line 31:
```suggestion
 * openAPI types are created under the subtypes folder, modify path of top level instances to be under the types folder
```
(or did you mean "by default"?)

packages/adapter-components/test/elements/swagger/type_elements.test.ts
```diff
@@ -33,7 +33,7 @@ jest.mock('@salto-io/lowerdash', () => {
 })
 
 const ADAPTER_NAME = 'myAdapter'
-const BASE_DIR = __dirname.replace('/dist', '')
+const BASE_DIR = __dirname.replace('elements/swagger', 'fetch/element/openAPI')
```
netama, line 36: can you please explain this change?

netama (approved):
🎈 looks great. just making sure the Types folder now looks identical for existing adapters?

packages/adapter-components/test/elements/swagger/type_elements.test.ts
shir-reifenberg: when I console logged `__dirname` it was not in `/dist` so I thought it's redundant

PR #5619, opened 2024-03-20
Remove a couple of fields from zendesk e2etest

DanielBarros1:
we need this to be able to run the test on an additional instance of zd

---

---

_Release Notes_: 
Zendesk:
Modifications to the Zendesk e2e test

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/e2e_test/adapter.test.ts
+++ b/packages/zendesk-adapter/e2e_test/adapter.test.ts
@@ -945,7 +945,6 @@ describe('Zendesk adapter E2E', () => {
       const article3Instance = createInstanceElement({
         type: ARTICLE_TYPE_NAME,
         valuesOverride: {
-          author_id: 'neta.marcus+zendesk@salto.io',
           draft: true,
           promoted: false,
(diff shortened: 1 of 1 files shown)


packages/zendesk-adapter/e2e_test/adapter.test.ts
```diff
@@ -945,7 +945,6 @@ describe('Zendesk adapter E2E', () => {
       const article3Instance = createInstanceElement({
         type: ARTICLE_TYPE_NAME,
         valuesOverride: {
-          author_id: 'neta.marcus+zendesk@salto.io',
```
edenhassid, line 948: maybe put another author instead?

edenhassid (approved):
LGTM! added one small comment, pre-approving :)

packages/zendesk-adapter/e2e_test/adapter.test.ts
DanielBarros1: It won't work, since we want to run this test in 2 separate instances, we would need a user that is a member of both, and i don't know if that's a thing :P

PR #5620, opened 2024-03-20
SALTO-5534 - Salesforce: support data records deletions in quickfetch

tomsellek:
In the quickfetch case we need to determine which records were deleted on the Salesforce side so we can delete them on our end. The simple way to do this would be to use a SOQL query with all the IDs of the records that exist in the workspace and see which of them are not returned. 

---

There's a small optimization of not querying for IDs of records that we just fetched, because we know those records were not deleted.

Tests:
- Do a full fetch, delete an account, do a quickfetch, observe the account being deleted.

---
_Release Notes_: 
Salesforce: quickfetch will now correctly identify deleted records

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/adapter.ts
+++ b/packages/salesforce-adapter/src/adapter.ts
@@ -137,11 +137,14 @@ import {
   apiNameSync,
   getFLSProfiles,
   getFullName,
+  instanceInternalId,
   isCustomObjectSync,
   isCustomType,
+  isInstanceOfCustomObjectSync,
   isMetadataInstanceElementSync,
   listMetadataObjects,
   metadataTypeSync,
+  queryClient,
 } from './filters/utils'
 import {
   retrieveMetadataInstances,
@@ -944,13 +947,10 @@ export default class SalesforceAdapter implements AdapterOperations {
       .value()
   }
 
-  private async getDeletedElemIdsForPartialFetch({
-    fetchProfile,
-    fetchElements,
-  }: {
-    fetchElements: ReadonlyArray<Element>
-    fetchProfile: FetchProfile
-  }): Promise<Required<PartialFetchData>['deletedElements']> {
+  private async getDeletedMetadataForPartialFetch(
+    fetchElements: ReadonlyArray<Element>,
+    fetchProfile: FetchProfile,
+  ): Promise<Required<PartialFetchData>['deletedElements']> {
     const createElemId = (type: ObjectType, fullName: string): ElemID => {
       const typeName = apiNameSync(type)
       return typeName === CUSTOM_OBJECT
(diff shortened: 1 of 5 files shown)


packages/salesforce-adapter/src/filters/custom_objects_instances.ts
```diff
@@ -863,6 +900,24 @@ const filterCreator: RemoteFilterCreator = ({ client, config }) => ({
...
+          instances.map(instanceInternalId),
+        ),
+      )
+      _.remove(
+        elements,
+        (element) =>
+          isInstanceOfCustomObjectSync(element) &&
+          !idsInSalesforce.has(instanceInternalId(element)),
+      )
+    }
```
tamtamirr, line 919: Will this code actually delete Records that were not changed?
The `elements` list will only contain fetched Elements.. No?

I think this logic needs to sit next to where we detected deletions for Metadata (in `adapter.ts` fetch)

tamtamirr (commented):
Added a comment. I might need a sync on this PR 🙏

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
tomsellek: Oh. Yeah. Right. Crap.

tomsellek: Why not add removal changes in the filter's `onDeploy`?

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -973,4 +975,82 @@ export default class SalesforceAdapter implements AdapterOperations {
...
+      await querySalesforceForRecordIdsOfInstances(workspaceInstances),
+    )
+
+    return _(workspaceInstances)
+      .filter(
+        (instance) =>
+          !instanceIdsInSalesforce.has(instanceInternalId(instance)),
+      )
+      .map((instance) => instance.elemID)
+      .value()
```
tamtamirr, line 1038: nit: Lodash is not required here..?

packages/salesforce-adapter/test/adapter.discover.test.ts
```diff
@@ -3073,3 +3085,152 @@ describe('Fetch via retrieve API', () => {
...
+      it('should return the modified instance', () => {
+        expect(fetchResult.elements).toContainEqual(
+          expect.objectContaining({
+            elemID: testInstance.elemID,
+          }),
+        )
+      })
+    })
+  })
+})
```
tamtamirr, line 3236: What is this addition?
I would also expect the test of deleted data records to be within `partial fetch deletion support` describe.
Please also note: This support you've added is also relevant for targeted fetch (partial)

tamtamirr (commented):
LGTM! Added some comments

packages/salesforce-adapter/src/adapter.ts
tomsellek: Ack. Must be leftovers.

packages/salesforce-adapter/test/adapter.discover.test.ts
tomsellek: > What is this addition?

Unit tests for the new code?

> I would also expect the test of deleted data records to be within partial fetch deletion support describe.

Ack

> Please also note: This support you've added is also relevant for targeted fetch (partial)

Ack.

tamtamirr (approved):
LGTM!

PR #5623, opened 2024-03-20
allow setting a standalone instance's type using predefined types

netama:
we should set standalone instance types based on the defined type, even if it's different from the current field type

---
_Release Notes_: 
None

---
_User Notifications_: 
None


shir-reifenberg (approved):
🔥

PR #5626, opened 2024-03-21
SALTO-5493: add solar bundle to record and json

Shulik95:
_Update the bundle record and json_

---

_None_

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


omrilit (approved):
Thanks!

PR #5628, opened 2024-03-21
SALTO-5647 - Saving test state in tmp dir

yelly:
updateConfig creates a real content provider instead of the mock, so every time the test was run it saved the state file to the local directory. Instead, saving it to the tmp dir.
Also handled some spelling warnings in the file.

---

---
_Release Notes_: 
None.

---
_User Notifications_: 
None.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,5 +6,10 @@
   "[json]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode",
     "editor.formatOnSave": true
-  }
+  },
+  "cSpell.words": [
+    "asynciterable",
+    "lowerdash",
+    "netsuite"
+  ]
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,10 +6,5 @@
   "[json]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode",
     "editor.formatOnSave": true
-  },
-  "cSpell.words": [
-    "asynciterable",
-    "lowerdash",
-    "netsuite"
-  ]
+  }
 }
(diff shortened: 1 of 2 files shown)


.vscode/settings.json
```diff
@@ -6,5 +6,10 @@
   "[json]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode",
     "editor.formatOnSave": true
-  }
+  },
+  "cSpell.words": [
```
yuribro, line 10: I think the user settings are a better place than here

packages/core/test/workspace/local/state/state.test.ts
```diff
@@ -471,7 +473,13 @@ describe('localState', () => {
       })
       overridingStateFilesSource = mockStaticFilesSource([])
       contentProvider = mockContentProvider({})
-      state = localState('empty', '', inMemRemoteMapCreator(), contentProvider, overridingStateFilesSource)
+      state = localState(
+        path.join(os.tmpdir(), 'empty'),
```
yuribro, line 477: It is still leaking the file (just in the tmp dir), maybe clean up this folder at the end?

yelly: Here's a (very partial) view of the tmp folder. It's already full of stuff that the tests put there that we have no control over.
From what I can tell, getting the exact name of the file to delete will require breaking an abstraction layer because the `State` interface doesn't expose a method to do this. Do you think it's worth it?
Screenshot 2024-03-21 at 12 21 13

.vscode/settings.json
yelly: Why? That way every person working on the code will need to manually add all of the "misspelled" words in the repo and there is no single source of truth for what we consider OK.

yelly: Removed.

packages/core/test/workspace/local/state/state.test.ts
yelly: Using `setupTmpDir`.

yuribro (approved):

PR #5633, opened 2024-03-24
Increase log level to error on deploy failures

shir-reifenberg:
and remove some dup log

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/client/http_client.ts
+++ b/packages/adapter-components/src/client/http_client.ts
@@ -240,7 +240,13 @@ export abstract class AdapterHTTPClient<TCredentials, TRateLimitConfig extends C
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const logResponse = (res: Response<any>, error?: any): void => {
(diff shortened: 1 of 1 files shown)


packages/adapter-components/src/client/http_client.ts
```diff
@@ -263,7 +269,9 @@ export abstract class AdapterHTTPClient<TCredentials, TRateLimitConfig extends C
           responseText,
         )
       } else {
-        log.warn(`failed to ${method} ${url} with error: ${error}, stack: ${error.stack}, ${responseText}`)
+        log.error(
```
netama, line 272: I don't think we should increase this to `error` since we might treat these as valid if they're listed as http success codes. what's the purpose of this change?

omrilit (approved):
🏅

netama:
please don't miss my question, also is there a ticket?

shir-reifenberg:
closing after discussion with @netama

PR #5642, opened 2024-03-24
SALTO-5637: Don't fetch ProfileMapping properties for new envs

shir-reifenberg:
modify the default value of `includeProfileMappingProperties` flag to false

---

The change will only apply for new envs

---
_Release Notes_: 

_Okta_adapter_:
- In new environments, ProfileMapping properties object will not be fetch by default, and can be included by turning `includeProfileMappingProperties` flag to true. Existing envs will not be affected.

---
_User Notifications_: 
None


omrilit (approved):
🙏

PR #5646, opened 2024-03-26
SALTO-5658: remove empty strings from transition parameters

shayc331:
_remove empty strings from transition parameters_

---

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/jira-adapter/src/filters/workflowV2/workflow_filter.ts
+++ b/packages/jira-adapter/src/filters/workflowV2/workflow_filter.ts
@@ -67,6 +67,7 @@ import {
   isAdditionOrModificationWorkflowChange,
   CONDITION_GROUPS_PATH_NAME_TO_RECURSE,
   WorkflowStatus,
+  EMPTY_STRINGS_PATH_NAME_TO_RECURSE,
 } from './types'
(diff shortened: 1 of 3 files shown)


packages/jira-adapter/src/filters/workflowV2/workflow_filter.ts
```diff
@@ -191,13 +210,15 @@ const createWorkflowInstances = async ({
...
           }
-          return toBasicInstance({
+          const instance = await toBasicInstance({
             entry: workflow,
             type: workflowConfigurationType,
             transformationConfigByType: getTransformationConfigByType(DEFAULT_API_DEFINITIONS.types),
             transformationDefaultConfig: DEFAULT_API_DEFINITIONS.typeDefaults.transformation,
             defaultName: workflow.name,
           })
+          walkOnElement({ element: instance, func: removeParametersEmptyStrings })
```
IdoZakk, line 220: You can call first WalkOnValue, then return the `toBasicInstance` and keep the await outside of this function. It will also simplify a bit your internal function

IdoZakk (approved):
🥇

PR #5648, opened 2024-03-26
SALTO-5659 avoid adding additional delimiter when element only extendsParent

omrifarkash:
Fixed a bug in the new create elemId function.  Where for an elemId with extendsParent = true but without any parts. We added a redundant delimiter at the end of the elemId

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


shir-reifenberg (approved):
💯

PR #5651, opened 2024-03-26
SALTO-5242: [Zendesk] Fix bug for deployment with additions of app installations

DanielBarros1:
When creating an app installation, you *have* to send all required field. This is not really a problem normally, unless you duplicate an existing app installation from another env, in which case fields that are secret will not be present in the nacl, which results in a deployment error.
In this PR, i add a call to ZD API during deployment to see if we are missing any required fields, and if we are - fill them with placeholders.
I also added a CV with instructions to the user to edit those fields after the deployment.

---
Screenshot 2024-03-26 at 13 09 10
Screenshot 2024-03-26 at 13 09 16

---
_Release Notes_: 
Zendesk:
Fix issue when creating app installations with missing fields

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/change_validators/app_installation.ts
+++ b/packages/zendesk-adapter/src/change_validators/app_installation.ts
@@ -1,0 +1,106 @@
+/*
+ *                      Copyright 2024 Salto Labs Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
+  ChangeError,
+  ChangeValidator,
+  ElemID,
+  getChangeData,
+  isAdditionChange,
+  isInstanceChange,
+} from '@salto-io/adapter-api'
+import { collections, values as lowerDashValues } from '@salto-io/lowerdash'
+import { logger } from '@salto-io/logging'
+import ZendeskClient from '../client/client'
+import { APP_INSTALLATION_TYPE_NAME } from '../constants'
+import { isValidApp } from '../filters/app_installations'
+
+const log = logger(module)
+const { awu } = collections.asynciterable
+
+const { isDefined } = lowerDashValues
+
+const APPS_SECTION_URL = 'admin/apps-integrations/apps/support-apps'
+
+export const createChangeError = (
+  instanceElemId: ElemID,
+  prefilledFields: Array<string>,
+  baseUrl: string,
+): ChangeError => ({
+  elemID: instanceElemId,
+  severity: 'Info',
+  message: 'App installation creation detected',
+  detailedMessage: 'App installation creation detected',
+  deployActions: {
+    preAction: {
+      title: 'Some App installation required fields will not be set',
+      description: `App installation ${instanceElemId.name} will be deployed with the placeholder values for the following fields: ${prefilledFields.length === 1 ? prefilledFields[0] : prefilledFields.join(', ')}. Salto will guide you on modifying them post deploy`,
+      subActions: [],
+    },
+    postAction: {
+      title: 'Set app installation fields',
+      description: `Please manually set the following fields for app installation ${instanceElemId.name} via the Zendesk UI: ${prefilledFields.length === 1 ? prefilledFields[0] : prefilledFields.join(', ')}`,
+      showOnFailure: false,
+      subActions: [
+        `Go to Zendesk Apps panel at ${baseUrl}${APPS_SECTION_URL}`,
+        'Hover over the created app',
+        'Click on the Gear icon > Change Settings',
+        'Edit the fields in the "App Installation" section',
+        'Click "Update"',
+      ],
+    },
+  },
+})
+
+export const appInstallationValidator: (client: ZendeskClient) => ChangeValidator = client => async changes => {
+  const potentialChanges = changes
+    .filter(isAdditionChange)
+    .filter(isInstanceChange)
+    .map(getChangeData)
+    .filter(changeData => changeData.elemID.typeName === APP_INSTALLATION_TYPE_NAME)
+
+  const errors = await awu(potentialChanges)
+    .flatMap(async changeData => {
+      try {
+        const res = (
+          await client.get({
+            url: `/api/support/apps/${changeData.value.app_id}/installations/new`,
+            responseType: 'json',
+          })
+        ).data
+        if (isValidApp(res) && res.installation !== undefined) {
+          const prefilledFields = res.installation.settings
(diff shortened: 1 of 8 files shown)


packages/zendesk-adapter/src/change_validators/app_installation.ts
```diff
@@ -0,0 +1,106 @@
...
+        'Hover over the created app',
+        'Click on the Gear icon > Change Settings',
+        'Edit the fields in the "App Installation" section',
+        'Click "Update"',
+      ],
+    },
+  },
+})
+
+export const appInstallationValidator: (client: ZendeskClient) => ChangeValidator = client => async changes => {
```
edenhassid, line 68: can you add docs?

packages/zendesk-adapter/src/change_validators/app_installation.ts
```diff
@@ -0,0 +1,106 @@
...
+  const potentialChanges = changes
+    .filter(isAdditionChange)
+    .filter(isInstanceChange)
+    .map(getChangeData)
+    .filter(changeData => changeData.elemID.typeName === APP_INSTALLATION_TYPE_NAME)
```
edenhassid, line 73: can you also make sure that there is an app_id?

packages/zendesk-adapter/src/change_validators/app_installation.ts
```diff
@@ -0,0 +1,106 @@
...
+  ElemID,
+  getChangeData,
+  isAdditionChange,
+  isInstanceChange,
+} from '@salto-io/adapter-api'
+import { collections, values as lowerDashValues } from '@salto-io/lowerdash'
+import { logger } from '@salto-io/logging'
+import ZendeskClient from '../client/client'
+import { APP_INSTALLATION_TYPE_NAME } from '../constants'
+import { isValidApp } from '../filters/app_installations'
```
edenhassid, line 29: maybe move this function to some util file? so that the cv will not have to import from a filter?

packages/zendesk-adapter/src/change_validators/app_installation.ts
```diff
@@ -0,0 +1,106 @@
...
+const log = logger(module)
+const { awu } = collections.asynciterable
+
+const { isDefined } = lowerDashValues
+
+const APPS_SECTION_URL = 'admin/apps-integrations/apps/support-apps'
+
+export const createChangeError = (
+  instanceElemId: ElemID,
+  prefilledFields: Array<string>,
```
edenhassid, line 40:
```suggestion
  prefilledFields: string[],
```

packages/zendesk-adapter/src/filters/app_installations.ts
```diff
@@ -41,20 +41,58 @@ const log = logger(module)
...
+  number: 123,
+}
+
 type JobStatus = {
   status: string
   message?: string
 }
 
+type ValidApp = {
+  installation: { settings: Array<{ secure: boolean; required: boolean; key: string; type: string }> }
```
edenhassid, line 63:
```suggestion
  installation: { settings: { secure: boolean; required: boolean; key: string; type: string }[] }
```

packages/zendesk-adapter/src/filters/app_installations.ts
```diff
@@ -130,6 +168,37 @@ const filterCreator: FilterCreator = ({ config, client }) => ({
...
+          if (isValidApp(res) && res.installation.settings !== undefined) {
+            const fieldsToAdd = res.installation.settings.filter(
+              field => field.secure === true && field.required === true,
+            )
+            fieldsToAdd.forEach(field => {
+              if (!(field.key in changeData.value.settings)) {
+                const placeholder =
+                  field.key in APP_FIELD_TYPE_TO_PLACEHOLDER_VALUE
+                    ? APP_FIELD_TYPE_TO_PLACEHOLDER_VALUE[field.key]
+                    : STR_PLACEHOLDER
```
edenhassid, line 193: maybe add a trace log if we get here? so that we will know to add this key in the `APP_FIELD_TYPE_TO_PLACEHOLDER_VALUE`, WDYT?

packages/zendesk-adapter/src/filters/app_installations.ts
```diff
@@ -130,6 +168,37 @@ const filterCreator: FilterCreator = ({ config, client }) => ({
...
+                changeData.value.settings[field.key] = placeholder
+              }
+            })
+          }
+        } catch (e) {
+          log.warn(`Failed to fetch app installation details. error: ${e}`)
```
edenhassid, line 199:
```suggestion
          log.warn(`Failed to fetch app installation details, missing required fields are not added to the app ${changeData.elemId.name}. error: ${e}`)
```

packages/zendesk-adapter/src/filters/app_installations.ts
```diff
@@ -130,6 +168,37 @@ const filterCreator: FilterCreator = ({ config, client }) => ({
...
+              url: `/api/support/apps/${changeData.value.app_id}/installations/new`,
+              responseType: 'json',
+            })
+          ).data
+          if (isValidApp(res) && res.installation.settings !== undefined) {
+            const fieldsToAdd = res.installation.settings.filter(
+              field => field.secure === true && field.required === true,
+            )
+            fieldsToAdd.forEach(field => {
+              if (!(field.key in changeData.value.settings)) {
```
edenhassid, line 189: you cannot edit the change itself, any additional fields you have will show on the nacl after the deploy. you need to either do this in pre-deploy and on deploy mechanism or clone the change and make sure the deploy runs on the cloned change.

packages/zendesk-adapter/test/filters/app_installations.test.ts
edenhassid, line 1: can you add a test for this too? make sure that it adds the placeholders, etc..

packages/zendesk-adapter/src/filters/app_installations.ts
```diff
@@ -130,7 +144,46 @@ const filterCreator: FilterCreator = ({ config, client }) => ({
...
+        clonedChange = await applyFunctionToChangeData(change, inst => inst.clone())
+        const changeData = getChangeData(clonedChange)
+        try {
+          const res = (
+            await client.get({
+              url: `/api/support/apps/${changeData.value.app_id}/installations/new`,
+              responseType: 'json',
+            })
+          ).data
+          if (isValidApp(res) && res.installation.settings !== undefined) {
```
edenhassid, line 163: why here you don't check that it doesn't exist already?

edenhassid (approved):
looks good! 💯 😎  added one comment, pre-approving :)

packages/zendesk-adapter/src/filters/app_installations.ts
```diff
@@ -130,7 +144,46 @@ const filterCreator: FilterCreator = ({ config, client }) => ({
         isAdditionOrModificationChange(change) && getChangeData(change).elemID.typeName === APP_INSTALLATION_TYPE_NAME,
     )
     const deployResult = await deployChanges(relevantChanges, async change => {
-      const response = await deployChange(change, client, config.apiDefinitions, [
+      let clonedChange
```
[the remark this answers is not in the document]
DanielBarros1: Wanted to avoid the need to clone on every deployment. My thought was that most of the deployments are modifications and then we don't run unnecessary code for those. I guess the code to clone is not really a problem from a performance perspective, but didn't want to have operations running for no reason :)

PR #5654, opened 2024-03-26
SALTO-5332 stop creating additionalProperties map field and use the annotation instead

omrifarkash:
- Stop creating additionalProperties field upon fetch
- Save reference to a type in _additional_properties annotation

---

---

_Release Notes_: 

_Jira_Adapter_:
Instances with additionalProperties map field will be flatten upon fetch (types to be affected: Notification Events, Workflows and maybe more)

_Okta_Adapter_:
Instances with additionalProperties map field will be flatten upon fetch (types to be affected: Users, GroupRule, PasswordPolicyRule and maybe more)

_Zuora_billing_Adapter_:
Instances with additionalProperties map field will be flatten upon fetch

_Stripe_Adapter_:
Instances with additionalProperties map field will be flatten upon fetch

---

_User Notifications_: 

_Jira_Adapter_:
Instances with additionalProperties map field will be flatten upon fetch (types to be affected: Notification Events, Workflows and maybe more)

_Okta_Adapter_:
Instances with additionalProperties map field will be flatten upon fetch (types to be affected: Users, GroupRule, PasswordPolicyRule and maybe more)

_Zuora_billing_Adapter_:
Instances with additionalProperties map field will be flatten upon fetch

_Stripe_Adapter_:
Instances with additionalProperties map field will be flatten upon fetch

### DIFF
--- a/packages/adapter-utils/src/additional_properties.ts
+++ b/packages/adapter-utils/src/additional_properties.ts
@@ -43,7 +43,9 @@ export const setAdditionalPropertiesAnnotation = <T extends Element>(
 }
 
 const isAdditionalPropertiesAnnotation = (value: Value): value is AdditionalPropertiesAnnotation =>
-  (value?.annotations === undefined || _.isPlainObject(value?.annotations)) && isReferenceExpression(value?.refType)
+  (value?.annotations === undefined || _.isPlainObject(value?.annotations)) &&
+  isReferenceExpression(value?.refType) &&
+  value.refType.elemID.idType === 'type'
 
 export const extractAdditionalPropertiesField = (objType: ObjectType, name: string): Field | undefined => {
   const additionalProperties = objType.annotations[CORE_ANNOTATIONS.ADDITIONAL_PROPERTIES]
--- a/packages/adapter-utils/src/additional_properties.ts
+++ b/packages/adapter-utils/src/additional_properties.ts
@@ -22,15 +22,15 @@ import {
   Element,
   ObjectType,
   Value,
-  TypeReference,
-  isTypeReference,
+  ReferenceExpression,
+  isReferenceExpression,
 } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
 import _ from 'lodash'
 
 const log = logger(module)
 type AdditionalPropertiesAnnotation = {
-  refType: TypeReference
+  refType: ReferenceExpression
   annotations?: Record<string, Value>
 }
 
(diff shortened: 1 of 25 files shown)


packages/zuora-billing-adapter/src/filters/object_defs.ts
```diff
@@ -120,60 +114,58 @@ const addRelationships = (
       }
       const refObjectNames = new Set<string>()
       // sort in order to keep relationship fields list stable
-      _.sortBy(relationships, 'object', ({ fields }) => Object.values(fields.additionalProperties ?? {})[0]).forEach(
-        rel => {
-          const { cardinality, namespace, object, fields, recordConstraints: constraints, ...additionalDetails } = rel
-          // the only cardinalities currently in use are oneToMany / manyToOne
-          if (cardinality === 'oneToMany') {
-            // each relationship appears in both directions, so it's enough to look at manyToOne
+      _.sortBy(relationships, 'object', ({ fields }) => Object.values(fields ?? {})[0]).forEach(rel => {
```
omrifarkash, line 117: The flatten change is reorder the list of relationships, this cause changes relatedTo and cardinality

packages/adapter-utils/src/additional_properties.ts
```diff
@@ -40,15 +43,20 @@ export const setAdditionalPropertiesAnnotation = <T extends Element>(
 }
 
 const isAdditionalPropertiesAnnotation = (value: Value): value is AdditionalPropertiesAnnotation =>
-  (value?.annotations === undefined || _.isPlainObject(value?.annotations)) && isType(value?.refType)
+  (value?.annotations === undefined || _.isPlainObject(value?.annotations)) && isReferenceExpression(value?.refType)
```
netama, line 46: please also check that the elem id if an elem id of a type (can use `.idType`)

packages/okta-adapter/src/filters/user_schema.ts
```diff
@@ -42,7 +42,7 @@ const log = logger(module)
 const { getTransformationConfigByType } = configUtils
 const { toBasicInstance } = elementUtils
 const { isDefined } = values
-const LINK_PATH = [LINKS_FIELD, 'additionalProperties', 'schema', 'href']
```
netama, line 45: I think @shir-reifenberg has a new case, please coordinate with her

packages/zuora-billing-adapter/src/filters/object_defs.ts
netama: not sure I understand - it sounds like we should ignore these explicitly-defined fields (either by name or based on the type's fields), no?

netama (approved):
reviewing compared to https://github.com/salto-io/salto/pull/5443

looks good, added one (blocking) comment on zuora and a nit (+ we should make sure Shir's PR takes this into account). a couple more nits -
* please add a TODO to remove also the usages of `flattenAdditionalProperties` in the deploy functions (e.g. in jira and okta `defaultDeployChange`)
* please add release notes for zuora and not only user notifications :)

I'm approving, but let's discuss if the needed zuora fix isn't clear

thanks!

packages/zuora-billing-adapter/src/filters/object_defs.ts
omrifarkash: The sortBy is yielding new results now so there are some side effects. We can discuss about it f2f and decide what to do

netama: I understand, I think we should avoid them

PR #5656, opened 2024-03-27
fix recurseInto with single result

netama:
the `single: true` case wasn't working properly

---
_Release Notes_: 
None

---
_User Notifications_: 
None


Uri-Bar (approved):
🔩

PR #5658, opened 2024-03-27
SALTO-5667: Omit required field from AccessPolicyRule constraints when it's optional

shir-reifenberg:
Fix issue with deployment of AccessPolicyRule

---

the `required` field is optional and should be set to true when no authentication method is excluded.
The filter omits this field when it optional in order to handle misalignments in envs (when source env supports this but target env doesn't)

---
_Release Notes_: 

_Okta_Adapter_:
- Fix bug in deployment of Access Policy Rules with `required` field set

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/filters/access_policy_rule_constraints.ts
+++ b/packages/okta-adapter/src/filters/access_policy_rule_constraints.ts
@@ -22,7 +22,7 @@ import {
   isInstanceElement,
   Values,
 } from '@salto-io/adapter-api'
-import { createSchemeGuard, resolvePath } from '@salto-io/adapter-utils'
+import { createSchemeGuard, resolvePath, safeJsonStringify } from '@salto-io/adapter-utils'
 import { logger } from '@salto-io/logging'
 import { FilterCreator } from '../filter'
 import { ACCESS_POLICY_RULE_TYPE_NAME } from '../constants'
@@ -31,12 +31,12 @@ const log = logger(module)
 const AUTHENTICATOR_CONSTRAINTS_PATH = ['actions', 'appSignOn', 'verificationMethod', 'constraints']
 
 type ConstraintsEntry = {
-  knowledge: Values
-  possession: Values
+  knowledge?: Values
+  possession?: Values
 }
 const CONSTRAINTS_ENTRY_SCHEMA = Joi.object({
-  knowledge: Joi.object(),
-  possession: Joi.object(),
+  knowledge: Joi.object().optional(),
+  possession: Joi.object().optional(),
 }).unknown(true)
 
 const CONSTRAINTS_ARRAY_SCHEMA = Joi.array().items(CONSTRAINTS_ENTRY_SCHEMA).required()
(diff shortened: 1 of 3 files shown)


packages/okta-adapter/src/filters/access_policy_rule_constraints.ts
```diff
@@ -0,0 +1,115 @@
...
+import { logger } from '@salto-io/logging'
+import { FilterCreator } from '../filter'
+import { ACCESS_POLICY_RULE_TYPE_NAME } from '../constants'
+
+const log = logger(module)
+const AUTHENTICATOR_CONSTRAINTS_PATH = ['actions', 'appSignOn', 'verificationMethod', 'constraints']
+
+type ConstraintsEntry = {
+  knowledge: Values
+  possession: Values
```
netama, line 35: aren't these both optional?

packages/okta-adapter/src/filters/access_policy_rule_constraints.ts
```diff
@@ -0,0 +1,115 @@
...
+
+const CONSTRAINTS_ARRAY_SCHEMA = Joi.array().items(CONSTRAINTS_ENTRY_SCHEMA).required()
+
+const isConstraintsArray = createSchemeGuard<ConstraintsEntry[]>(
+  CONSTRAINTS_ARRAY_SCHEMA,
+  'Received an invalid constraints',
+)
+
+/**
+ * Removes the `required` field from `possession` or `knowledge` object when it is optionl.
```
netama, line 50:
```suggestion
 * Removes the `required` field from `possession` or `knowledge` object when it is optional.
```

packages/okta-adapter/src/filters/access_policy_rule_constraints.ts
```diff
@@ -0,0 +1,115 @@
...
+ * By default, this field is true. If the `knowledge` or `possession` constraint has values for excludedAuthenticationMethods then the required value is false.
+ * (source: https://developer.okta.com/docs/reference/api/policy/#constraints)
```
netama, line 52: just making sure - we'll be able to remove this filter and replace it with an `adjust` function in the new infra, right?

packages/okta-adapter/src/filters/access_policy_rule_constraints.ts
```diff
@@ -0,0 +1,115 @@
...
+          originalChangeByElemID[instance.elemID.getFullName()] = instance.clone()
+          const constraintsArray = resolvePath(
+            instance,
+            instance.elemID.createNestedID(...AUTHENTICATOR_CONSTRAINTS_PATH),
+          )
+          if (isConstraintsArray(constraintsArray)) {
+            constraintsArray.forEach(constraint => {
+              const { possession, knowledge } = constraint
+              if (
+                possession?.additionalProperties?.required === true &&
```
netama, line 74: please align this with @omrifarkash 's PR which is supposed to be merged this week (he mapped out all the places using `additionalProperties`)

packages/okta-adapter/src/filters/access_policy_rule_constraints.ts
```diff
@@ -0,0 +1,115 @@
...
+                possession?.excludedAuthenticationMethods === undefined
+              ) {
+                log.debug(
+                  'omitting "required" field from possession object %o in instance %s',
+                  possession,
```
netama, line 79: `safeJsonStringify` instead? otherwise you might get `[object Object]` (or some endless recursion of there's a reference expression)

netama (commented):
cool, added some minor comments

netama (approved):
pre-approving, for the `additionalProperties` flattening maybe make the filter handle both locations?

PR #5655, opened 2024-03-27
SALTO-5650 SALTO-5651: Salesforce Fetch With Changes Detection Deleted Custom Fields Support

tamtamirr:
Fixed various issues when Custom Fields are deleted in Fetch With Changes Detection mode.

---

1. Returning the `elemID` of deleted CustomFields from `Partial Fetch Data`.
2. Fixed a bug where the adapter attempted to query deleted fields.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

PR #5661, opened 2024-03-28
SALTO-5650 SALTO-5651: Salesforce Fetch With Changes Detection CustomFields Deletions

tamtamirr:
Salesforce Fetch With Changes Detection CustomFields Deletions

---

Changed the implementation to treat Object with Deleted Field as updated.
We detect deleted fields based on the `list` result of `CustomField`.

For reference, here's an alternate solution for the linked tickets that we decided makes less sense product wise:  https://github.com/salto-io/salto/pull/5655

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/adapter.ts
```diff
@@ -558,6 +577,8 @@ export default class SalesforceAdapter implements AdapterOperations {
           fetchParams,
           elementsSource: this.elementsSource,
           lastChangeDateOfTypesWithNestedInstances,
+          customObjectsWithDeletedFields:
+            await this.getCustomObjectsWithDeletedFields(),
```
tomsellek, line 581: Doesn't this happen before we called `listMetadata` (and thus before the proxy had a chance to discover what fields exist in the env)?

tomsellek (commented):
I'm not sure I understand how this logic works, flow-wise. Please see my comment.

packages/salesforce-adapter/src/adapter.ts
tamtamirr: The list already happens in `getLastChangeDateOfTypesWithNestedInstances`, but in order to be more explicit and future proof I've added another list call on CustomFields as part of the implementation, since we cache the list calls either way :)

PR #5662, opened 2024-03-28
SALTO-5161: [Zendesk] Move caching for users request in Zendesk to be…

DanielBarros1:
Re-creating the PR that i previously reverted. 

The getUsers query currently runs on a global scope and caches the user data.
This means having more than one workspace on a single process can cause data leakage and unexpected behaviour, which happens when running SAAS locally.

In this PR I move the scope of the query from global to the zendesk adapter, which prevents the leak.
This is essentially identical to the pattern added [here in Okta](https://github.com/salto-io/salto/pull/5261/files#diff-fdd8ade86a6a37213a71a3465db95e5e94e20ec7bc026beab906303cc6630a1aR127).

---

I ran Saas locally with my changes, added logs and made sure that the query executes correctly and that fetch and deployment works properly.

---

_Release Notes_: 
Zendesk:
Fix leakage of users when running Saas locally with multiple environments

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/adapter.ts
+++ b/packages/zendesk-adapter/src/adapter.ts
@@ -149,7 +149,7 @@ import organizationsFilter from './filters/organizations'
 import hideAccountFeatures from './filters/hide_account_features'
 import auditTimeFilter from './filters/audit_logs'
 import sideConversationsFilter from './filters/side_conversation'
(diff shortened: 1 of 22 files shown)


DanielBarros1:
Still need to fix the unit tests, but wanted to have the PR in place either way

PR #5663, opened 2024-03-28
SALTO-5408 - Salesforce: Validate Salesforce CustomField businessOwnerUser user existence

tomsellek:
There's an annotation on custom fields that refers to a user. Let's validate that user too

---
This is different from the current validations because it runs on annotations of all custom fields as opposed to the values of specific fields. Thus a separate function.

image

---
_Release Notes_: 
Salesforce: Will now validate custom fields' Data Owner property

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/change_validators/unknown_users.ts
+++ b/packages/salesforce-adapter/src/change_validators/unknown_users.ts
@@ -31,12 +31,13 @@ import {
 import { transformElement, TransformFuncArgs } from '@salto-io/adapter-utils'
 import _ from 'lodash'
 import {
+  apiNameSync,
   buildSelectQueries,
-  isInstanceOfType,
+  isCustomObjectSync,
+  isInstanceOfTypeSync,
   queryClient,
 } from '../filters/utils'
 import SalesforceClient from '../client/client'
-import { isFieldOfCustomObject } from '../transformers/transformer'
 
 const { awu } = collections.asynciterable
 const { makeArray } = collections.array
@@ -69,7 +70,7 @@ type TypeWithUserFields = (typeof TYPES_WITH_USER_FIELDS)[number]
 type TypesWithUserFields = Record<TypeWithUserFields, UserFieldGetter[]>
 
 type UserRef = {
-  elemId: ElemID
+  elemID: ElemID
   field: string
   user: string
 }
@@ -201,7 +202,7 @@ const getUsersFromInstance = async (
   const users: UserRef[] = topLevelGetters.flatMap((getter) =>
     getter
       .getter(instance.value)
-      .map((user) => ({ user, elemId: instance.elemID, field: getter.field })),
+      .map((user) => ({ user, elemID: instance.elemID, field: getter.field })),
   )
 
   const gettersBySubType = new Map(
@@ -217,7 +218,7 @@ const getUsersFromInstance = async (
     if (subTypeGetter && path) {
       const userRefs = subTypeGetter
         .getter(value)
-        .map((user) => ({ user, elemId: path, field: subTypeGetter.field }))
+        .map((user) => ({ user, elemID: path, field: subTypeGetter.field }))
       users.push(...userRefs)
     }
     return value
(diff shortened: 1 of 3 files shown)


tomermevorach (approved):

packages/salesforce-adapter/src/change_validators/unknown_users.ts
```diff
@@ -263,34 +267,87 @@ const unknownUserError = ({ elemId, field, user }: UserRef): ChangeError => ({
...
+    .filter(isFieldChange)
+    .filter((change) => isFieldOfCustomObject(getChangeData(change)))
+    .map(getChangeData)
+
+  const userRefs = changedCustomField
+    .filter(
+      (customField) => customField.annotations?.businessOwnerUser !== undefined,
+    )
+    .flatMap((customField) => ({
+      user: customField.annotations.businessOwnerUser as string,
```
tamtamirr, line 325: I think we should avoid the casting here.
I would add a typeguard `isFieldWithBusinessOwnerUser` that validates `customField.annotations.businnessOwnerUser` is of type string

packages/salesforce-adapter/src/change_validators/unknown_users.ts
```diff
@@ -263,34 +267,87 @@ const unknownUserError = ({ elemId, field, user }: UserRef): ChangeError => ({
...
+      field: customField,
+    }))
+
+  const existingUsers = new Set(
+    await getSalesforceUsers(
+      client,
+      userRefs.map(({ user }) => user),
+    ),
+  )
```
tamtamirr, line 334: We should try to avoid executing this query twice.
Instead of receiving `client` to your methods, I'd recommend passing the `existingUsers` as arguments to both methods.

packages/salesforce-adapter/src/change_validators/unknown_users.ts
```diff
@@ -263,34 +267,87 @@ const unknownUserError = ({ elemId, field, user }: UserRef): ChangeError => ({
...
-        client,
-        userRefs.map(({ user }) => user),
-      ),
-    )
-
-    return userRefs
-      .filter(({ user }) => !existingUsers.has(user))
-      .map(unknownUserError)
+    const instancesErrors = await validateInstances(changes, client)
+    const customFieldsErrors = await validateCustomFields(changes, client)
```
tamtamirr, line 349: Not sure this separation to a different methods is necessary. I think it adds complexity
If you'd like to extract the validation to another code, I would recommend the following method (a type parameter would be better):
`validateUsersExist(instances: InstanceElement[], businnessOwnerFields: BusinessOwnerField[], knownUsers: Set<string>)`

packages/salesforce-adapter/test/change_validators/unknown_users.test.ts
```diff
@@ -559,4 +576,29 @@ describe('unknown user change validator', () => {
       )
     })
   })
+  describe('when the user information is in a CustomField annotation', () => {
```
tamtamirr, line 579:
```suggestion
  describe('CustomFields businessOwnerUser', () => {
```
Nit

packages/salesforce-adapter/test/change_validators/unknown_users.test.ts
```diff
@@ -559,4 +576,29 @@ describe('unknown user change validator', () => {
...
+        'SomeField',
+        {
+          businessOwnerUser: TEST_USERNAME,
+        },
+      )
+      changes = [toChange({ after: fieldAfter })]
+
+      setupClientMock([IRRELEVANT_USERNAME])
+    })
+    it('should fail validation', async () => {
```
tamtamirr, line 595:
```suggestion
    it('should have change error', async () => {
```

tamtamirr (commented):
Looking good! Added some comments

packages/salesforce-adapter/src/change_validators/unknown_users.ts
tomsellek: Ack

tomsellek: They're not the same users, though - in each of the functions we query for which of the users that we found in [fields/annotations] exist in Salesforce. Thus `existingUsers` will be a subset of the `userRefs` in each function.
We could have each function return an array of user references, combine them and call the query once, but given that we want to have a different error message for the different cases (so we'll need to keep track of which function a missing user came from) and that we're not querying for all the users on the system but only for those that are referenced, I think it's better to keep things as they are.

packages/salesforce-adapter/src/change_validators/unknown_users.ts
tomsellek: The logic of extracting the users is completely different though?

packages/salesforce-adapter/test/change_validators/unknown_users.test.ts
tomsellek: Ack

tamtamirr (approved):
LGTM :)

PR #5665, opened 2024-03-28
SALTO-5569 - The state file should be un-changed after a Fetch with no semantic changes

tomsellek:
We want the state file to remain unchanged if there were no changes to the workspace.

This requires removing the feature of recommending 'align' mode if the accounts are too old.

---

These changes were *not* exhaustively tested.

---
_Release Notes_: 
Will no longer keep track of the last fetch date for accounts. 

---
_User Notifications_: 
N/A


packages/core/src/local-workspace/state/state.ts
```diff
@@ -77,7 +77,7 @@ const parseStateContent = async (contentStreams: AsyncIterable<NamedStream>): Pr
           } else if (key === 1) {
             // line 2 - update dates, e.g.
             //   {"dummy":"2023-01-09T15:57:59.322Z"}
-            res.updateDates.push(value)
+            res.accounts.push(value)
```
amitgoldarb, line 80: Is this code backwards compatible to state files with existing timestamps?  Also the comment above refers to the the format of the original code. We should either edit it or remove it I think

tomsellek: Ack

packages/cli/test/commands/fetch.test.ts
```diff
@@ -612,189 +611,6 @@ describe('fetch command', () => {
       })
     })
   })
-  describe('multienv - new account in env, with existing common elements', () => {
```
amitgoldarb, line 615: Do we really need to remove this entire test?

packages/core/src/local-workspace/state/state.ts
```diff
@@ -75,9 +75,14 @@ const parseStateContent = async (contentStreams: AsyncIterable<NamedStream>): Pr
             //   [{"elemID":{...},"annotations":{...}},{"elemID":{...},"annotations":{...}},...]
             res.elements = res.elements.concat(await deserializeParsed(value))
           } else if (key === 1) {
-            // line 2 - update dates, e.g.
-            //   {"dummy":"2023-01-09T15:57:59.322Z"}
-            res.accounts.push(value)
+            // line 2 - configured accounts, e.g.
+            //   [dummy, ...]
+            if (_.isPlainObject(value)) {
+              // old-format
```
amitgoldarb, line 77: I think we can elaborate a bit on these comments.

amitgoldarb (commented):
This LGTM.
Let's wait for @ori-moisis's approval as well 🙏 
Added 2 small comments

packages/cli/test/commands/fetch.test.ts
tomsellek: I think so. AFAICT it tests the "suggest different fetch mode if the last fetch was too long ago" feature?

packages/cli/test/mocks.ts
```diff
@@ -326,11 +325,13 @@ export const mockWorkspace = ({
...
-      key: wsState.StateMetadataKey
-      value: string
-    }[]),
+    accounts: [],
+    saltoMetadata: new InMemoryRemoteMap(
+      [] as {
+        key: wsState.StateMetadataKey
+        value: string
+      }[],
+    ),
```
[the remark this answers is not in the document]
tomsellek: Ack

packages/core/src/local-workspace/state/state.ts
```diff
@@ -75,19 +71,22 @@ const parseStateContent = async (contentStreams: AsyncIterable<NamedStream>): Pr
             //   [{"elemID":{...},"annotations":{...}},{"elemID":{...},"annotations":{...}},...]
             res.elements = res.elements.concat(await deserializeParsed(value))
           } else if (key === 1) {
-            // line 2 - update dates, e.g.
-            //   {"dummy":"2023-01-09T15:57:59.322Z"}
-            res.updateDates.push(value)
+            // line 2 - configured accounts, e.g.
+            //   [dummy, ...]
```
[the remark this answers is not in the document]
tomsellek: Ack. AFAICT we only keep the state of one account per file, so I'll make it a scalar value rather than an array.

packages/core/src/local-workspace/state/state.ts
tomsellek: Ack.

packages/core/src/local-workspace/state/state.ts
```diff
@@ -75,19 +71,22 @@ const parseStateContent = async (contentStreams: AsyncIterable<NamedStream>): Pr
...
             // line 3 - path index, e.g.
             //   [["dummy.aaa",[["dummy","Types","aaa"]]],["dummy.aaa.instance.bbb",[["dummy","Records","aaa","bbb"]]]]
             res.pathIndices = res.pathIndices.concat(value)
           } else if (key === 3) {
             // line 4 - version, e.g.
             //   "0.1.2"
-            if (!_.isEmpty(value)) {
-              res.versions.push(value)
-            }
+            log.trace('Old format state file contains the OSS version information')
```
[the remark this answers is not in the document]
tomsellek: Ack.

packages/core/src/local-workspace/state/state.ts
```diff
@@ -155,21 +154,10 @@ export const localState = (
...
-    if (stateUpdateDate !== undefined) {
-      await stateUpdateDate.clear()
-      await stateUpdateDate.setAll(awu(Object.entries(updateDatesByAccount).map(([key, value]) => ({ key, value }))))
-    }
-    const currentVersion = semver.minSatisfying(res.versions, '*') ?? undefined
-    if (currentVersion) {
-      await stateData.saltoMetadata.set('version', currentVersion)
+    const accounts = res.accounts ?? []
+    const stateAccounts = stateData.accounts
+    if (stateAccounts !== undefined) {
```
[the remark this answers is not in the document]
tomsellek: Ack

packages/core/src/local-workspace/state/state.ts
```diff
@@ -224,9 +211,8 @@ export const localState = (
       }
       yield* yieldWithEOL([
         accountToElementStreams[account],
-        awu([safeJsonStringify({ [account]: accountToDates[account] } || {})]),
+        awu([safeJsonStringify(account)]),
```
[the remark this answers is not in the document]
tomsellek: `yieldWithEOL` requires all parameters to be `AsyncIterable`s.

packages/workspace/src/workspace/state/in_mem.ts
```diff
@@ -36,9 +36,7 @@ const { awu } = collections.asynciterable
 
 const log = logger(module)
 
-type InMemoryState = State & {
-  setVersion(version: string): Promise<void>
-}
+type InMemoryState = State
```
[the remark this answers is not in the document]
tomsellek: That sort of depends on your feelings about types-as-documentation. For my part, I think yes.

packages/workspace/src/workspace/state/in_mem.ts
```diff
@@ -61,8 +59,7 @@ export const buildInMemState = (loadData: () => Promise<StateData>, persistent =
 
   const updateAccounts = async (accounts?: string[]): Promise<void> => {
     const data = await stateData()
-    const newAccounts = accounts ?? (await awu(data.accountsUpdateDate.keys()).toArray())
-    return data.accountsUpdateDate.setAll(newAccounts.map(s => ({ key: s, value: new Date(Date.now()) })))
+    data.accounts = accounts ?? data.accounts
```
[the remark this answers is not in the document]
tomsellek: Ack.

packages/workspace/src/workspace/state/in_mem.ts
```diff
@@ -148,19 +145,15 @@ export const buildInMemState = (loadData: () => Promise<StateData>, persistent =
...
+    existingAccounts: async (): Promise<string[]> => (await stateData()).accounts,
     getPathIndex: async (): Promise<PathIndex> => (await stateData()).pathIndex,
     getTopLevelPathIndex: async (): Promise<PathIndex> => (await stateData()).topLevelPathIndex,
     clear: async () => {
       const currentStateData = await stateData()
       await currentStateData.elements.clear()
       await currentStateData.pathIndex.clear()
       await currentStateData.topLevelPathIndex.clear()
-      await currentStateData.accountsUpdateDate.clear()
+      currentStateData.accounts.length = 0
```
ori-moisis, line 156: seems like we were ok with replacing the list instead of updating it in place on `updateAccounts`, why can't we do the same here?

just feels like in terms of readability
```typescript
currentStateData.accounts = []
```
gets the point across more clearly

packages/workspace/src/workspace/state/state.ts
```diff
@@ -95,12 +92,7 @@ export const buildStateData = async (
     deserialize: async data => JSON.parse(data),
     persistent,
   }),
-  accountsUpdateDate: await remoteMapCreator<Date>({
-    namespace: createStateNamespace(envName, 'service_update_date'),
-    serialize: async date => date.toISOString(),
-    deserialize: async data => new Date(data),
-    persistent,
-  }),
+  accounts: [],
```
ori-moisis, line 95: isn't this a problem to always initialize this empty?
now that this is no longer persisted in the cache, won't it cause the state to always think it has an empty account list when the cache is valid?

packages/core/src/local-workspace/state/state.ts
```diff
@@ -75,19 +71,24 @@ const parseStateContent = async (contentStreams: AsyncIterable<NamedStream>): Pr
...
           } else if (key === 3) {
-            // line 4 - version, e.g.
+            // In the old format, there was a line 4 - version, e.g.
             //   "0.1.2"
-            if (!_.isEmpty(value)) {
-              res.versions.push(value)
-            }
+            // We do not expect this in new-format files, but keep the code here so older state files are parsed as valid.
+            // once we are certain all state files are updated, we can remove the `key === 3` section.
+            log.trace('Old format state file contains the Salto version information')
```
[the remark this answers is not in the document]
tomsellek: Ack

packages/core/test/workspace/local/state/state.test.ts
```diff
@@ -47,24 +46,17 @@ const { toMD5 } = hash
...
 const mockStateContent = async ({
   elements,
-  date,
   pathIndexLine = '[]',
-  version = '0.0.1',
   extraLine,
 }: MockStateContentArgs): Promise<Buffer> => {
   const elementsLine = await serialization.serialize(elements)
-  const dateLine = safeJsonStringify({ [elements[0].elemID.adapter]: date ?? new Date() })
+  const dateLine = safeJsonStringify(elements[0].elemID.adapter)
```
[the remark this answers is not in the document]
tomsellek: Ack

packages/workspace/src/workspace/state/in_mem.ts
tomsellek: Ack

packages/workspace/src/workspace/state/state.ts
tomsellek: Indeed. My bad for misunderstanding what this data structure was used for.

packages/workspace/src/workspace/state/state.ts
```diff
@@ -95,10 +92,10 @@ export const buildStateData = async (
     deserialize: async data => JSON.parse(data),
     persistent,
   }),
-  accountsUpdateDate: await remoteMapCreator<Date>({
-    namespace: createStateNamespace(envName, 'service_update_date'),
-    serialize: async date => date.toISOString(),
-    deserialize: async data => new Date(data),
+  accounts: await remoteMapCreator<string>({
```
ori-moisis, line 95: 1. why is this a map structured like this? this feels like an unintuitive way to store a list of strings...
2. how does this work in an upgrade scenario? seems like the new code can't read the old cache, so until the state is updated it will be as if there are no accounts, no?

packages/workspace/src/workspace/state/state.ts
```diff
@@ -95,10 +92,10 @@ export const buildStateData = async (
...
+    namespace: createStateNamespace(envName, 'service_names'),
+    serialize: async data => data,
+    deserialize: async data => data,
     persistent,
   }),
   saltoMetadata: await remoteMapCreator<string, 'version'>({
```
ori-moisis, line 101: is this still used anywhere?
seems like this remote map only stores the version, but IIUC we don't read the version or store it anymore

ori-moisis (commented):
seems like there are still issues around the migration from the old format to the new one...

I suggest to focus a bit more on manually testing these scenarios (loading existing state/cache with the new code) and making sure the results are correct because it is very hard to catch these kinds of issues in a review.

this can be done with unit tests as well, but only if you feel like it justifies the effort, my intuition is it doesn't, but I may be wrong...

packages/workspace/src/workspace/state/state.ts
tomsellek: > how does this work in an upgrade scenario? seems like the new code can't read the old cache, so until the state is updated it will be as if there are no accounts, no?

Yes. I modified `loadStateData` to account for this.

> also, much less important - why is this a map structured like this? this feels like an unintuitive way to store a list of strings...

I agree, but I don't know of another way to store a list of strings in our cache?

packages/workspace/src/workspace/state/state.ts
tomsellek: So this is kind of weird, but from what I understand this is where we keep the cache's hash in addition to the version.

tomsellek:
> seems like there are still issues around the migration from the old format to the new one...
> 
> I suggest to focus a bit more on manually testing these scenarios (loading existing state/cache with the new code) and making sure the results are correct because it is very hard to catch these kinds of issues in a review.
> 
> this can be done with unit tests as well, but only if you feel like it justifies the effort, my intuition is it doesn't, but I may be wrong...

I loaded an existing workspace and it seems to have worked as expected (following the latest fix to `loadStateData`)

packages/core/test/workspace/local/state/state.test.ts
```diff
@@ -336,15 +303,6 @@ describe('localState', () => {
       state = localState('empty', '', inMemRemoteMapCreator(), contentProvider)
     })
 
-    it('should return an undefined hash', async () => {
-      const stateHash = await state.getHash()
-      expect(stateHash).toBeUndefined()
-    })
```
tomsellek, line 342: We now fall back to loading the state from the file if the cached state is empty, so there will always be a hash value.

packages/workspace/src/workspace/state/state.ts
tomsellek: nvm. Figured it out.

packages/core/test/workspace/local/state/state.test.ts
ori-moisis: are you sure we don't rely on this behavior anywhere?
I think this might break things in the workspace code

packages/core/src/local-workspace/state/state.ts
```diff
@@ -184,11 +171,13 @@ export const localState = (
     const filePaths = await currentContentProvider.findStateFiles(currentFilePrefix)
     const stateFilesHash = await currentContentProvider.getHash(filePaths)
     const quickAccessHash = (await quickAccessStateData.saltoMetadata.get('hash')) ?? toMD5(safeJsonStringify([]))
-    if (quickAccessHash !== stateFilesHash) {
+    const cachedAccounts = await awu(quickAccessStateData.accounts.entries()).toArray()
+    if (quickAccessHash !== stateFilesHash || cachedAccounts.length === 0) {
```
ori-moisis, line 175: this change would effectively invalidate all of the state files caches, this is not something that we should do since reloading the state file can take a lot of time and for sufficiently large state file it could actually fail and cause the workspace to basically be broken with no way to recover (this is a fairly rare case).

please change it back

packages/workspace/src/workspace/state/in_mem.ts
```diff
@@ -148,19 +149,15 @@ export const buildInMemState = (loadData: () => Promise<StateData>, persistent =
...
+    existingAccounts: async (): Promise<string[]> => (await (await stateData()).accounts.get('account_names')) ?? [],
     getPathIndex: async (): Promise<PathIndex> => (await stateData()).pathIndex,
     getTopLevelPathIndex: async (): Promise<PathIndex> => (await stateData()).topLevelPathIndex,
     clear: async () => {
       const currentStateData = await stateData()
       await currentStateData.elements.clear()
       await currentStateData.pathIndex.clear()
       await currentStateData.topLevelPathIndex.clear()
-      await currentStateData.accountsUpdateDate.clear()
+      await currentStateData.accounts.clear()
```
ori-moisis, line 160: we need to keep the existing remote map in the code until we are done with migrating state files (we need to define some time period where we support the migration).
we can't just ignore the fact that the old information exists before we migrate to the new format
```suggestion
      await currentStateData.accountsUpdateDate.clear()
      await currentStateData.accounts.clear()
```

packages/workspace/src/workspace/state/in_mem.ts
```diff
@@ -148,19 +149,15 @@ export const buildInMemState = (loadData: () => Promise<StateData>, persistent =
     setAll: async (elements: ThenableIterable<Element>): Promise<void> => awu(elements).forEach(setElement),
     remove: removeId,
     isEmpty: async (): Promise<boolean> => (await stateData()).elements.isEmpty(),
-    getAccountsUpdateDates: async () => {
-      const stateDataVal = await awu((await stateData()).accountsUpdateDate.entries()).toArray()
-      return Object.fromEntries(stateDataVal.map(e => [e.key, e.value]))
-    },
-    existingAccounts: async (): Promise<string[]> => awu((await stateData()).accountsUpdateDate.keys()).toArray(),
+    existingAccounts: async (): Promise<string[]> => (await (await stateData()).accounts.get('account_names')) ?? [],
```
ori-moisis, line 152: this should use the same implementation as `updateAccounts`
it should also _account_ for the possibility that this is an old format cache

```suggestion
    const existingAccounts = async (): Promise<string[]> => {
      const data = await stateData()
      const accounts = await data.accounts.get('account_names')
      if (accounts !== undefined) {
        return accounts
      }
      // Fallback to old format state cache where this information was written in a different map
      return await awu(data.accountsUpdateDate.keys()).toArray() ?? []
    }
```

packages/workspace/src/workspace/state/state.ts
ori-moisis: so let's fix the type here?
should be
```typescript
saltoMetadata: await remoteMapCreator<string, 'hash'>({
```
right?

packages/core/src/local-workspace/state/state.ts
```diff
@@ -75,19 +71,24 @@ const parseStateContent = async (contentStreams: AsyncIterable<NamedStream>): Pr
...
           } else if (key === 3) {
-            // line 4 - version, e.g.
+            // In the old format, there was a line 4 - version, e.g.
             //   "0.1.2"
-            if (!_.isEmpty(value)) {
-              res.versions.push(value)
-            }
+            // We do not expect this in new-format files, but keep the code here so older state files are parsed as valid.
+            // once we are certain all state files are updated, we can remove the `key === 3` section.
+            log.trace('Old format state file %s contains the Salto version information', name)
```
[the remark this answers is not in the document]
tomsellek: 🤷

packages/workspace/src/workspace/state/in_mem.ts
tomsellek: Ack

packages/workspace/src/workspace/state/in_mem.ts
tomsellek: Ack. Also, I see what you did there.

packages/workspace/src/workspace/state/in_mem.ts
```diff
@@ -60,9 +58,12 @@ export const buildInMemState = (loadData: () => Promise<StateData>, persistent =
   }
 
   const updateAccounts = async (accounts?: string[]): Promise<void> => {
+    if (!accounts) {
+      return
+    }
     const data = await stateData()
-    const newAccounts = accounts ?? (await awu(data.accountsUpdateDate.keys()).toArray())
-    return data.accountsUpdateDate.setAll(newAccounts.map(s => ({ key: s, value: new Date(Date.now()) })))
+    const existingAccounts = (await data.accounts.get('account_names')) ?? []
```
ori-moisis, line 65: this needs to consider the possibility that the cache has not been updated yet (and therefore the accounts are still in `accountsUpdateDate`).
it would probably be best to reuse the logic in `existingAccounts`

packages/workspace/src/workspace/state/state.ts
ori-moisis: also, in the type this is defined using the type `StateMetadataKey`, we should probably fix that type and use it here as well

ori-moisis (approved):
one last thing to fix, approving to save the last cycle

packages/core/src/local-workspace/state/state.ts
```diff
@@ -184,11 +178,13 @@ export const localState = (
     const filePaths = await currentContentProvider.findStateFiles(currentFilePrefix)
     const stateFilesHash = await currentContentProvider.getHash(filePaths)
     const quickAccessHash = (await quickAccessStateData.saltoMetadata.get('hash')) ?? toMD5(safeJsonStringify([]))
+    const cachedAccounts = await loadAccountNames(quickAccessStateData)
     if (quickAccessHash !== stateFilesHash) {
       log.debug(
-        'found different hash - loading state data (quickAccessHash=%s stateFilesHash=%s)',
+        'found different hash - loading state data (quickAccessHash=%s stateFilesHash=%s cachedAccounts = %o)',
```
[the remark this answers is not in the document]
tomsellek: Ack

packages/workspace/src/workspace/state/in_mem.ts
tomsellek: Ack

PR #5667, opened 2024-03-31
SALTO-5618: fix a bug in fetching transition with the same key

shayc331:
_fix a bug in fetching transition with the same key_

---

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


IdoZakk (approved):
💯

PR #5668, opened 2024-03-31
SALTO: 5690 upgrade log.time args

edenhassid:
upgrade log.time args

---

will delete the backward compatibility once the saas is also upgraded

---
_Release Notes_: 

---
_User Notifications_: 


edenhassid:
decided to go with another approach https://github.com/salto-io/salto/pull/5674

PR #5673, opened 2024-03-31
SALTO-5281: [Zendesk] Add CV to prevent race condition in source_locale changes

DanielBarros1:
When trying to delete the source_locale translation of a guide item (category/section/article) in the same time as creating a new one, we hit a race condition that breaks the deployment.
Fixing the race condition requires some pretty elaborate re-jigging of the groups and dependencies between items and translations, which would be drastically easier with the new infra. 
In the meantime, in this PR i'm adding a CV which makes sure users don't attempt to deploy these changes simultaneously.

---

Screenshot 2024-03-31 at 14 23 22
Screenshot 2024-03-31 at 14 23 46

---

_Release Notes_: 
Zendesk: 
Add CV to prevent deployments with race-condition failures in the guide

---
_User Notifications_: 
None


packages/zendesk-adapter/src/change_validators/guide_default_translation_change.ts
```diff
@@ -0,0 +1,157 @@
...
+ * 2. Change a guide item's source locale to the new created translation
+ * 3. Remove the translation that used to be the item's source local
+ * This CV is meant to block these deployments, and explain that they should be split into separate deployments
+ */
+export const guideDefaultTranslationChangeValidator: ChangeValidator = async changes => {
+  const guideItemErrors = changes
+    .filter(isModificationChange)
+    .filter(isInstanceChange)
+    .filter(change => change.data.before.value.source_locale !== change.data.after.value.source_locale)
+    .filter(change => GUIDE_ITEM_TYPE_NAMES.includes(getChangeData(change).elemID.typeName))
```
edenhassid, line 84: this filter needs to be before the source_locale filter

packages/zendesk-adapter/src/change_validators/guide_default_translation_change.ts
```diff
@@ -0,0 +1,157 @@
...
+  isAdditionChange,
+  isInstanceChange,
+  isModificationChange,
+  isReferenceExpression,
+  isRemovalChange,
+} from '@salto-io/adapter-api'
+import { getParent } from '@salto-io/adapter-utils'
+import { GUIDE_ITEM_TYPE_NAMES, TRANSLATION_TYPE_NAMES } from '../constants'
+
+const getLocaleChange = (
```
edenhassid, line 36: why not make this function get all the translation changes and then search on them? instead of filtering again for each modification?

packages/zendesk-adapter/src/change_validators/guide_default_translation_change.ts
```diff
@@ -0,0 +1,157 @@
...
+            instance,
+            getChangeData(addedSourceLocale),
+            getChangeData(removedSourceLocale),
+          ),
+        ]
+      }
+      return []
+    })
+
+  const translationErrors = changes
```
edenhassid, line 111: I don't understand why you do it twice, why not look for the modification changes, find the translation changes from them and then log the error?

edenhassid (commented):
looks good 😄 added some comments

packages/zendesk-adapter/src/change_validators/guide_default_translation_change.ts
DanielBarros1: oh right that would be simpler lol

packages/zendesk-adapter/src/change_validators/guide_default_translation_change.ts
DanielBarros1: Not sure i understand what you mean here

PR #5675, opened 2024-03-31
SALTO-5514: SFDC client retry on 406

yelly:
SFDC client retry on 406.

---

None.

---
_Release Notes_:

_Salesforce_:
- Automatically retrying requests which fail with a 406 status. We are seeing these errors in the wild and they are always resolved by retrying.

---
_User Notifications_: 
None.

### DIFF
--- a/packages/salesforce-adapter/src/client/client.ts
+++ b/packages/salesforce-adapter/src/client/client.ts
@@ -389,8 +389,8 @@ const retryErrorsByCodeWrapper =
       return true
     }
     if (errorCodesToRetry.some((code) => response.statusCode === code)) {
-      log.trace(
-        `Retrying on ${response.statusCode} due to known salesforce issues. Err: ${err}`,
+      log.warn(
+        `Retrying on ${response.statusCode} due to known salesforce issues. Err: ${err}, headers: ${response.headers}, status message: ${response.statusMessage}`,
       )
       return true
     }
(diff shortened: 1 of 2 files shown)


ori-moisis:
why no release notes?
is this not solving an issue that users can encounter?

packages/salesforce-adapter/src/client/client.ts
```diff
@@ -380,14 +382,16 @@ const sendChunked = async <TIn, TOut>({
...
+const retryErrorsByCodeWrapper =
   (strategy: RetryStrategy): RetryStrategy =>
   (err, response, body) => {
     if (strategy(err, response, body)) {
       return true
     }
-    if (response.statusCode === 400) {
-      log.trace(`Retrying on 400 due to known salesforce issues. Err: ${err}`)
+    if (errorCodesToRetry.some((code) => response.statusCode === code)) {
+      log.trace(
```
ori-moisis, line 392: IIUC we don't know what causes these 406 errors, so it might be worth logging the response (at least the headers and body).
also, since I think we don't expect many of those, to happen, maybe we should log them not as `trace`?

yelly: I've bumped this up to `warn` and added the response headers and status message to the log. The body isn't available in the interface so can't print that (maybe there's some way to get to it but not sure it's worth the effort).

ori-moisis: so, the 3rd argument in line 386 named `body` is not actually the body?

ori-moisis (approved):

packages/salesforce-adapter/src/client/client.ts
yelly: 🤦
My bad, I was looking in the response object expecting it to be there.

PR #5676, opened 2024-03-31
SALTO-5684: Rename the default access policy

shir-reifenberg:
This fixes an issue with deployment of default access policies when it was renamed in one env but not the other

---

every Okta tenant has one default access policy, the default policy has a certain rule in the configuration and therefore cannot be added (only modified in certain properties). The issue occurred during env compare when the default policies have different names across envs. As the elemIDs are different, salto identifies the changes as addition and removal instead of modification. To resolve the issue, we create a constant elemID for the default policy, alias and path remains the same for better UX.

I also had to rename all children rules as the point to the parent policy.

---
_Release Notes_: 
_Okta_adapter_:
- Fix an issue with deployment of default authentication policies in certain cases. To resolve, run next fetch with "regenerate elemIDs"

---
_User Notifications_: 
_Okta_adapter_:
- When running fetch with "regenerate elemIDs", elemID for the default access policy will change

### DIFF
--- a/packages/okta-adapter/src/filters/rename_default_access_policy.ts
+++ b/packages/okta-adapter/src/filters/rename_default_access_policy.ts
@@ -1,0 +1,103 @@
+/*
+ *                      Copyright 2024 Salto Labs Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import _ from 'lodash'
+import {
+  CORE_ANNOTATIONS,
+  ElemIdGetter,
+  InstanceElement,
+  ReferenceExpression,
+  isInstanceElement,
+} from '@salto-io/adapter-api'
+import { getParent, hasValidParent, naclCase } from '@salto-io/adapter-utils'
+import { elements as elementUtils } from '@salto-io/adapter-components'
+import { logger } from '@salto-io/logging'
+import { FilterCreator } from '../filter'
+import { ACCESS_POLICY_RULE_TYPE_NAME, ACCESS_POLICY_TYPE_NAME, OKTA } from '../constants'
+
+const log = logger(module)
+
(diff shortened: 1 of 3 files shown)


packages/okta-adapter/src/filters/rename_default_access_policy.ts
```diff
@@ -0,0 +1,103 @@
...
+ * */
+const filter: FilterCreator = ({ getElemIdFunc }) => ({
+  name: 'renameDefaultAccessPolicy',
+  onFetch: async elements => {
+    const instances = elements.filter(isInstanceElement)
+    const defaultAccessPolicy = instances
+      .filter(instance => instance.elemID.typeName === ACCESS_POLICY_TYPE_NAME)
+      .find(instance => instance.value.system === true)
+    if (defaultAccessPolicy === undefined) {
+      log.debug('The default access policy was not found, skipping renaming')
```
edenhassid, line 61: is this a valid scenario? that there is no default?

packages/okta-adapter/src/filters/rename_default_access_policy.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    const defaultPolicyRules = instances
+      .filter(instance => instance.elemID.typeName === ACCESS_POLICY_RULE_TYPE_NAME)
+      .filter(instance =>
+        hasValidParent(instance)
+          ? getParent(instance).elemID.getFullName() === defaultAccessPolicy.elemID.getFullName()
+          : false,
+      )
+
+    const defaultName = 'Default Policy'
```
edenhassid, line 73: can a non default policy have this name?

packages/okta-adapter/src/filters/rename_default_access_policy.ts
```diff
@@ -0,0 +1,103 @@
...
+    const updatedName = getDefaultAccessPolicyName(defaultName, defaultAccessPolicy, getElemIdFunc)
+    const renamedPolicy = new InstanceElement(
+      updatedName,
+      defaultAccessPolicy.getTypeSync(),
+      defaultAccessPolicy.value,
+      defaultAccessPolicy.path,
+      defaultAccessPolicy.annotations,
```
edenhassid, line 80: wouldn't you want them to also have the same alias?

edenhassid (approved):
LGTM! 🔥 added some small questions, pre-approving :)

packages/okta-adapter/src/filters/rename_default_access_policy.ts
shir-reifenberg: Not really, but it can happen when policies are excluded for example

packages/okta-adapter/src/filters/rename_default_access_policy.ts
shir-reifenberg: technically yes but i think it's very unlikely, do you hav a suggestion ?

PR #5677, opened 2024-04-01
SATLO-5705 fix copyFromResponse on additions

netama:
1. Fix a bug in the new `overrideInstanceTypeForDeploy` that caused service id fields to not be marked correctly on additions if they did not exist on the instance
2. improve the `copyFromResponse` logic - separate the service id copy from custom transformations in order to avoid confusing / misaligned cases, and change the default behavior when copying service ids to use the request's `nestUnderField` as its `root`

---
_Release Notes_: 
None (not used yet)

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/deployment/definition_helpers.ts
+++ b/packages/adapter-components/src/deployment/definition_helpers.ts
@@ -83,9 +83,6 @@ export const createStandardItemDeployDefinition = <AdditionalAction extends stri
             nestUnderField,
           },
         },
-        copyFromResponse: {
-          root: nestUnderField,
-        },
       },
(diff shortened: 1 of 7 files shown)


packages/adapter-components/src/deployment/definition_helpers.ts
```diff
@@ -83,9 +83,6 @@ export const createStandardItemDeployDefinition = <AdditionalAction extends stri
             nestUnderField,
           },
         },
-        copyFromResponse: {
-          root: nestUnderField,
-        },
```
edenhassid, line 88: why do you delete this and not put it under additional?

edenhassid (approved):
LGTM! 🔥 😎

PR #5679, opened 2024-04-01
SALTO-4246: Fetch issue type icon

idoamit8:
In this PR:
1. I added filter to fetch and deploy issue type icons (static files).
2. **Please look only on commit 2 and any sub commit.** 

---

1. I address system issueType icons as custom icons since either way each issueType point to its own icon
3. New issueType nacl:
image

4. I checked and if you change the static file without the field `avatar` it still considered modification change. 
5. I will add tests after review.  

---
_Release Notes_: 
_Jira Adapter:_
* Now support fetching and deploying issue type icons. 

---
_User Notifications_: 
_Jira Adapter:_
* Users will see new fields in their issueType nacls named `avatar`.


packages/jira-adapter/src/filters/issue_type.ts
```diff
@@ -33,10 +42,29 @@ const SUBTASK_TYPE = 'subtask'
...
+): Promise<void> => {
+  const instance = getChangeData(change)
+  try {
+    const headers = { ...PRIVATE_API_HEADERS, 'Content-Type': 'image/png' }
+    const url = `/rest/api/3/universal_avatar/type/issuetype/owner/${instance.value.id}`
+    const resp = await sendIconRequest({ client, change, url, fieldName: 'avatar', headers })
+    if (!isIconResponse(resp)) {
+      throw new Error('Failed to deploy icon to Jira issue type: Invalid response from Jira API')
+    }
+    instance.value.avatarId = Number(resp.data.id)
```
IdoZakk, line 57: Continuing the messages from the previous commit:
In the deployIcon function, you change the instance of the issue type (not the icon), and I think it might be confusing. I think it will be easier to read if the deployIcon function returns the icon ID, and the caller will use it to update the issue type icon

packages/jira-adapter/src/filters/issue_type.ts
```diff
@@ -48,10 +76,25 @@ const filter: FilterCreator = ({ client }) => ({
...
     }
-
     issueTypes.forEach(issueType => {
       delete issueType.value.subtask
     })
+    if (client.isDataCenter) {
+      return { errors: [] }
+    }
+    const errors: SaltoError[] = []
+    await awu(issueTypes).forEach(async issueType => {
```
IdoZakk, line 86: To make sure we do not do one by one please change this to Promise.all

packages/jira-adapter/src/filters/issue_type.ts
```diff
@@ -48,10 +76,25 @@ const filter: FilterCreator = ({ client }) => ({
...
+      try {
+        if (issueType.value.avatarId === undefined) {
+          return
```
IdoZakk, line 89: log a warning here so it will be easier to understand in case it happens

packages/jira-adapter/src/filters/issue_type.ts
```diff
@@ -71,6 +114,59 @@ const filter: FilterCreator = ({ client }) => ({
...
+          })
+        }
+      } else {
+        // Load the avatar to get avatarId
+        const avatarBefore = change.data.before.value.avatar
+        const avatarAfter = change.data.after.value.avatar
+        if (avatarBefore !== undefined && avatarAfter !== undefined && !avatarBefore.isEqual(avatarAfter)) {
+          await deployIcon(change, client)
+        }
+        // update the issueTpype with the avatarId
```
IdoZakk, line 155: Nit typo in the comment

IdoZakk (approved):
🎈 〽️ 🍾 
Minor comments, make sure to merge only after the noise reduction is merged

PR #5682, opened 2024-04-01
SALTO-4730: Do not suggest to exclude types with many instances in allowReferenceTo

tamtamirr:
Do not suggest to exclude types with many instances in allowReferenceTo.

---

_Release Notes_: 
Salesforce Adapter:
- Do not suggest to exclude types with many instances that are under `allowReferenceTo`.

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/src/filters/custom_objects_instances.ts
+++ b/packages/salesforce-adapter/src/filters/custom_objects_instances.ts
@@ -794,8 +794,10 @@ const filterTypesWithManyInstances = async ({
   const heavyTypesSuggestions: ConfigChangeSuggestion[] = []
 
   // Creates a lists of typeNames and changeSuggestions for types with too many instances
-  await awu(Object.keys(validChangesFetchSettings)).forEach(
-    async (typeName) => {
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/src/filters/custom_objects_instances.ts
```diff
@@ -794,8 +794,10 @@ const filterTypesWithManyInstances = async ({
   const heavyTypesSuggestions: ConfigChangeSuggestion[] = []
 
   // Creates a lists of typeNames and changeSuggestions for types with too many instances
-  await awu(Object.keys(validChangesFetchSettings)).forEach(
-    async (typeName) => {
+  await awu(Object.entries(validChangesFetchSettings))
+    // Do not handle types that are in `allowReferenceTo`
```
adamro, line 798: supernit: 
Be positive :D 
// Only handle base types

(tbh, I think the comment is redundant :) )

adamro (approved):
🔥 

Can you please add release notes?

PR #5684, opened 2024-04-01
SALTO-5704: Allow using transformation in recurseInto

shir-reifenberg:
Allow passing transformation for recurseInto in order to adjust values before using in recurse into

---

---
_Release Notes_: 
None 

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/definitions/system/fetch/dependencies.ts
+++ b/packages/adapter-components/src/definitions/system/fetch/dependencies.ts
@@ -13,7 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { TransformDefinition } from '../shared'
+import { types } from '@salto-io/lowerdash'
+import { ContextParams, TransformDefinition } from '../shared'
 
 export type DependsOnDefinition = {
   parentTypeName: string
@@ -32,9 +33,11 @@ export type Condition = ConditionByField | ConditionByContext
 
 export const isConditionByField = (condition: Condition): condition is ConditionByField => 'fromField' in condition
 
-type RecurseIntoContextParamDefinition = {
-  fromField: string // TODO replace with transformation config to align
-}
+type RecurseIntoContextParamDefinition<TContext = ContextParams> = types.OneOf<{
+  fromField: string,
+  transformation: TransformDefinition<TContext, string> & { single?: true } // set single to true as transform func should return a single string
+}>
+
 type RecurseIntoContext = {
   args: Record<string, RecurseIntoContextParamDefinition>
 }
--- a/packages/adapter-components/src/definitions/system/fetch/dependencies.ts
+++ b/packages/adapter-components/src/definitions/system/fetch/dependencies.ts
@@ -34,7 +34,7 @@ export type Condition = ConditionByField | ConditionByContext
 export const isConditionByField = (condition: Condition): condition is ConditionByField => 'fromField' in condition
 
 type RecurseIntoContextParamDefinition<TContext = ContextParams> = types.OneOf<{
-  fromField: string,
+  fromField: string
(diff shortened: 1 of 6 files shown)


packages/adapter-components/src/definitions/system/fetch/dependencies.ts
```diff
@@ -32,9 +33,11 @@ export type Condition = ConditionByField | ConditionByContext
 
 export const isConditionByField = (condition: Condition): condition is ConditionByField => 'fromField' in condition
 
-type RecurseIntoContextParamDefinition = {
-  fromField: string // TODO replace with transformation config to align
-}
+type RecurseIntoContextParamDefinition<TContext = ContextParams> = types.OneOf<{
+  fromField: string
```
netama, line 37: hmm maybe we should just use a transformation here after all, since `fromField` is equivalent to `root`, right? I think we don't need to nest the transformation field, just use it directly (`RecurseIntoContextParamDefinition<...> = TransformationDefinition<...>`)
(but we can also do it later)

packages/adapter-components/src/definitions/system/fetch/dependencies.ts
```diff
@@ -32,9 +33,11 @@ export type Condition = ConditionByField | ConditionByContext
...
+  transformation: TransformDefinition<TContext, string | string[]>
```
netama, line 38: what guarantees that a string is returned? in general there is an option to provide a context "object" and have the args used by something nested under it, I don't think there's a reason to restrict it here

packages/adapter-components/src/fetch/resource/subresources.ts
```diff
@@ -15,16 +15,42 @@
...
 ) => Promise<Record<string, ValueGeneratedItem[] | ValueGeneratedItem>>
 
+const extractRecurseIntoContext = (
+  item: ValueGeneratedItem,
+  recurseIntoDef: RecurseIntoDefinition,
+): Record<string, string | string[]> => {
+  const { args: contextArgs } = recurseIntoDef.context
+  const context = _.mapValues(contextArgs, contextDef => {
+    if (contextDef.fromField !== undefined) {
+      return _.get(item.value, contextDef.fromField)
```
netama, line 38: continuing from above - there is no guarantee that this is a string, please either use a guard before returning or change the signature to `unknown`
(the old code just assumes `unknown`)

packages/adapter-components/src/fetch/resource/subresources.ts
```diff
@@ -15,16 +15,42 @@
...
+    }
+    const transformer = createValueTransformer(contextDef.transformation)
+    return collections.array
```
netama, line 41: can you please explain why this part is needed?

netama (commented):
thanks 🎈 added some comments and questions

packages/adapter-components/src/fetch/resource/subresources.ts
shir-reifenberg: cause if definition is `single` it would return { value: {}, typeName: '', context: {} }
no ?

packages/adapter-components/src/definitions/system/fetch/dependencies.ts
shir-reifenberg: yeah it's the same
will change

packages/adapter-components/src/fetch/resource/subresources.ts
netama: the lack of validation comes from the fact that `nestedRequestContext` is passed to the type fetcher which has ```
context?: Record<string, unknown>
```
so anything accessing the context needs to check what it contains. here we're casting it to a string without checking

netama (approved):
🎉

PR #5687, opened 2024-04-02
SALTO-5672: Retrying SFDC requests on disconnect/reset before headers

yelly:
Retrying SFDC requests on disconnect/reset before headers.

---

None.

---
_Release Notes_: 

_Salesforce_:
- Retrying SFDC requests on disconnect/reset before headers.

---
_User Notifications_: 
None.


amitgoldarb (approved):
Thanks for this!

PR #5690, opened 2024-04-02
SALTO-5044: move logs to trace level

edenhassid:
move logs to trace level

---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/adapter-components/src/client/http_client.ts
+++ b/packages/adapter-components/src/client/http_client.ts
@@ -240,7 +240,7 @@ export abstract class AdapterHTTPClient<TCredentials, TRateLimitConfig extends C
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const logResponse = (res: Response<any>, error?: any): void => {
-      log.debug('Received response for %s on %s', method.toUpperCase(), url)
+      log.trace('Received response for %s on %s', method.toUpperCase(), url)
 
       const responseText = safeJsonStringify({
         url,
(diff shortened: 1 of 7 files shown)


packages/adapter-components/src/client/pagination/pagination_async.ts
```diff
@@ -213,7 +213,7 @@ export const traverseRequestsAsync: (
       }
     }
     // the number of results may be lower than actual if the instances are under a nested field
-    log.info('Received %d results for endpoint %s', numResults, getParams.url)
```
shir-reifenberg, line 216: Why are we changing this specific log to trace? I think it's pretty useful and there are certain things that can't be done with rehydrate (like looking on a few ops together)

edenhassid: discussed IRL, i'll leave as info

packages/adapter-components/src/client/pagination/pagination_async.ts
```diff
@@ -106,7 +106,7 @@ class PromisesQueue {
     }
     const promise = singlePagePagination(pageArgs, additionalArgs)
     this.promises.push(promise)
-    log.debug(`Added promise to pagination queue. Queue size: ${this.promises.length}`)
+    log.trace(`Added promise to pagination queue. Queue size: ${this.promises.length}`)
```
netama, line 109: this is getting deprecated, if you want this to take effect in the new infra as well it should be in https://github.com/salto-io/salto/blob/main/packages/adapter-components/src/fetch/request/pagination/queue.ts (I'll mark the old code as deprecated next week)

netama (commented):
please be aware that except for the client itself, the rest of the shared code is getting deprecated soon so most of these changes will not apply for long

packages/adapter-components/src/client/http_client.ts
```diff
@@ -240,7 +240,7 @@ export abstract class AdapterHTTPClient<TCredentials, TRateLimitConfig extends C
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const logResponse = (res: Response<any>, error?: any): void => {
-      log.debug('Received response for %s on %s', method.toUpperCase(), url)
+      log.trace('Received response for %s on %s', method.toUpperCase(), url)
```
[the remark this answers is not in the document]
netama: this is the important log IMO, I think if there's one line we keep as debug from the client it should be this one as it's the most useful. there's also no need to keep the line below that was converted to `log.log`, we just need to add here ` with status %d` at the end with the right arg and avoid the duplication

packages/adapter-components/src/client/http_client.ts
```diff
@@ -282,13 +282,15 @@ export abstract class AdapterHTTPClient<TCredentials, TRateLimitConfig extends C
             url,
             isMethodWithData(params) ? { ...requestConfig, data: params.data } : requestConfig,
           )
-      log.debug(
+      log.log(
+        method.toUpperCase() === 'GET' ? 'trace' : 'debug',
```
netama, line 286: I don't think this can be a differentiator, fetches can also have non-GET requests. I think this isn't the place for this logic as I said yesterday, sorry...

PR #5691, opened 2024-04-02
SALTO-5711: Sort group membership members

shir-reifenberg:
To avoid comparison noise, we should sort group members
this must be done after user filter to avoid sorting based on internal ids 

---

No noise reduction is needed as this is behind an FF 

---

_Release Notes_: 
None

---
_User Notifications_: 

_Okta_adapter_:
- members list in GroupMembership instances will be sorted


idoamit8 (approved):
💯

PR #5695, opened 2024-04-02
SALTO-5718: Add managedPackages to SalesforceConfigOptionsType

tamtamirr:
Add managedPackages to SalesforceConfigOptionsType

---

_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/config_creator.ts
```diff
@@ -181,19 +184,38 @@ export const configWithCPQ = new InstanceElement(
...
+      cpq: { refType: BuiltinTypes.BOOLEAN },
+      managedPackages: {
+        refType: new ListType(BuiltinTypes.STRING),
+        annotations: {
+          [CORE_ANNOTATIONS.RESTRICTION]: createRestriction({
+            values: MANAGED_PACKAGES,
+            enforce_value: true,
+          }),
+          [CORE_ANNOTATIONS.DESCRIPTION]:
+            'You can choose to add managed packages e.g. sbaa, SBQQ (CPQ)',
```
ori-moisis, line 211: I think we may want a little bit more info in the description here eventually.
also, I think usually documentation shouldn't start with "you", it should usually describe the behavior of the code, not of the user.
so here it can be something like
```
Names of managed packages to fetch into the environment [Learn more](...)
```
I wonder about including the example in the description, since we have the values in the `values` restriction...

ori-moisis (approved):

packages/salesforce-adapter/src/config_creator.ts
tamtamirr: Do we have a learn more here?

ori-moisis: @rozenmanofer sent me this link - https://help.salto.io/en/articles/9164974-extending-your-salesforce-configuration-with-managed-packages

PR #5697, opened 2024-04-03
SALTO-5721: Consolidate missing ref strategies between okta and jira

shir-reifenberg:
Consolidate missing ref strategies between okta and jira and allow reusing missing ref strategies for other adapters

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


netama (approved):
cool :) I was just starting to do this as part of a refactor of the shared base class, so great timing 😅

PR #5699, opened 2024-04-03
SALTO-5599 Fix multiline regex to support $ before a reference

kami-delaroz:
Fix multiline regex to support $ before a reference
For example: "Cost:  ${{issue.Cost}} for year" - we should be able to escape the given $, and match the reference to cost

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/parser/src/parser/internal/native/lexer.ts
+++ b/packages/parser/src/parser/internal/native/lexer.ts
@@ -50,7 +50,14 @@ export const TOKEN_TYPES = {
 
 const WORD_PART = '[a-zA-Z_][\\w.@]*'
 const NEWLINE_CHARS = '\r\n\u2028\u2029'
-const MULTILINE_CONTENT = new RegExp(`.*\\\\\\$\\{.*[${NEWLINE_CHARS}]|.*?(?=\\$\\{)|.*[${NEWLINE_CHARS}]`)
+// In multiline strings, we want to match content till a reference expression, so next token matches the reference.
+// to do this we use a lookahead that finds the '${', but matches the string without it. the next token should catch the full ${<reference>} expression.
+// But, if the origin string contains something like ${<some text> {<reference>}}, the $ will be escaped by us and we get \${<some text> {<reference>}}
+// So, we need to distinguish between '${' and '\${', and the first alternative in the regex actually checks for a '${' preceded by at least one char that isn't \
+// The second alternative checks for strings containing '\${', and stops right after, to avoid including real references that comes after
+// The third alternative matches any other string, till a new line.
+const MULTILINE_CONTENT = new RegExp(`.*?[^\\\\](?=\\$\\{)|.*\\\\\\$\\{|.*[${NEWLINE_CHARS}]`)
+const ALL_TILL_NEW_LINE = new RegExp(`.*[${NEWLINE_CHARS}]`)
 // Template markers are added to prevent incorrect parsing of user created strings that look like Salto references.
 // We accept unicode newline characters because the string dump code does not treat them as newlines so they can appear
 // in a single line string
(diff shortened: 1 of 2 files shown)


packages/parser/src/parser/internal/native/helpers.ts
```diff
@@ -55,7 +55,7 @@ export const parseTopLevelID = (context: ParseContext, fullname: string, range:
   return new ElemID(adapter, name)
 }
 
-export const createReferenceExpresion = (ref: string): ReferenceExpression | IllegalReference => {
+export const createReferenceExpression = (ref: string): ReferenceExpression | IllegalReference => {
```
ori-moisis, line 58: not that I am against fixing typos
but parser PRs are at a high risk of getting reverted, so might be best to split this to a separate PR

packages/parser/test/parser/parse.test.ts
```diff
@@ -733,6 +736,29 @@ value
           '}}\nvalue',
         ])
       })
+
+      it('should parse references to complex multiline template as TemplateExpression', () => {
```
ori-moisis, line 740: I think the name of the test should be move specific about the use case and not generally say "complex".
complex does not give any information, it is like calling something "quick", it is meaningless...

specifically we are testing the case where a template exists on the same line as an escaped template marker

packages/parser/test/parser/parse.test.ts
```diff
@@ -733,6 +736,29 @@ value
           '}}\nvalue',
         ])
       })
+
+      it('should parse references in multiline that exists on the same line as an escaped template marker as TemplateExpression', () => {
```
adfineberg, line 740: Not sure, but maybe also add a test in `dump.test.ts` that dumps and parses the dump?

packages/parser/src/parser/internal/native/lexer.ts
```diff
@@ -50,7 +50,14 @@ export const TOKEN_TYPES = {
...
 const NEWLINE_CHARS = '\r\n\u2028\u2029'
-const MULTILINE_CONTENT = new RegExp(`.*\\\\\\$\\{.*[${NEWLINE_CHARS}]|.*?(?=\\$\\{)|.*[${NEWLINE_CHARS}]`)
+// In multiline strings, we want to match content till a reference expression, so next token matches the reference.
+// to do this we use a lookahead that finds the '${', but matches the string without it. the next token should catch the full ${<reference>} expression.
+// But, if the origin string contains something like ${<some text> {<reference>}}, the $ will be escaped by us, and we get \${<some text> {<reference>}}
+// So, we need to distinguish between '${' and '\${'.
+// The first alternative in the regex actually checks for a '${' preceded by at least one char that isn't \
+// The second alternative checks for strings containing '\${', and stops right after, to avoid including real references that comes after
+// The third alternative matches any other string, until a new line.
+const MULTILINE_CONTENT = new RegExp(`.*?[^\\\\](?=\\$\\{)|.*\\\\\\$\\{|.*[${NEWLINE_CHARS}]`)
```
adfineberg, line 60: Didn't we discuss splitting this into 3 different tokenizers?

kami-delaroz: I'm not sure it's clearer, given the explanation comment

adfineberg: Not sure, because if you split it into 3 tokens, only the last one needs to be with the `new Regexp` format, so it's **MUCH** more readable. (also, I prefer to limit comments if there's an option to have it encapsulated in the variable name)

packages/parser/test/parser/parse.test.ts
kami-delaroz: I'll have a look. but prefer not to risk this PR from getting into next deployment as this is an important fix

packages/parser/src/parser/internal/native/lexer.ts
kami-delaroz: after wasting so much time understanding what was the intention in these regexes I highly disagree and will keep this explanation anyway :)
I tried to slit it back then and it got totally broken. will give it another try if I get to it today

kami-delaroz: OK. Just verifying we're aligned that the conflicts matcher is using the same multi-line-content regex, that it has an open bug ticket and will need a refactor soon, and that the split of the regex will be in both matchers

ori-moisis: what? no
the conflict content was supposed to use a different regex, I am pretty sure the version I approved had that fix
when and why was this changed back?

kami-delaroz: look at commit number 3 :) it was before the approval
I'm still investigating this area

PR #5708, opened 2024-04-04
SALTO-5404: Set prettier to ignore json files under test folder

shoshana-houri:
We want to be able to define mock replies JSONs (like [this one](https://github.com/salto-io/salto/blob/main/packages/serviceplaceholder-adapter/test/fetch_mock_replies.json)) in a more concise format. Therefore, we should disable Prettier for JSON files located directly within test directories.

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,1 @@
+packages/*/test/*.json


netama (approved):
thanks!

PR #5711, opened 2024-04-04
SALTO-5565: Allow sorting an array field as an infra capability

Nurdok:
This PR introduces a new common filter that allows to sort list fields.
The sorting configuration is part of the fetch definition under `ElementFieldCustomization`.

---

Note for reviewers: I still don't know what I'm doing :)
PTAL specifically at the filter type - I used `FilterCreator` directly and not `AdapterFilterCreator` because I didn't understand how to provide the correct types / signature to make it work.

I'm not entirely sure if I used the new infra here as the types are in the same files and there's no clear boundary between what configuration is "old" and what is "new" and I'm not sure I figured it out correctly.

Should this filter work for the old infra too? If so, I can try to convert the existing sorting use cases.

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/adapter-components/src/filters/sort_lists.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
+  if (rest.length === 0) {
+    return next
+  }
+  return get(next, rest)
+}
+
+const sortLists = async (instance: InstanceElement, defQuery: DefQuery<ElementFetchDefinition>): Promise<void> => {
+  instance.value =
+    (await transformValues({
```
netama, line 52: in fetch we can use `transformValuesSync` which is much faster, because we have all the types loaded. there's also `.getTypeSync()` on the instances

packages/adapter-components/src/filters/sort_lists.ts
```diff
@@ -0,0 +1,104 @@
...
+
+const { awu } = collections.asynciterable
+
+/*
+ * Resolve a string path within an instance value.
+ *
+ * This function recursively resolves the path within the instance value, including hops into
+ * values from resolved reference expressions.
+ */
+const get = (current: Value, tail: string[]): Value => {
```
netama, line 32: any chance we can instead use `resolvePath` from adapter-utils, to keep consistent with other parts of the code (and some nuances in how elem ids are resolved)?

packages/adapter-components/src/filters/sort_lists.ts
```diff
@@ -0,0 +1,104 @@
...
+        }
+        const fieldCustomizations = defQuery.query(field.parent.elemID.typeName)?.fieldCustomizations
+        if (fieldCustomizations === undefined) {
+          return value
+        }
+        const sortFields = fieldCustomizations[field.name]?.sort?.sortByProperties
+
+        if (sortFields !== undefined) {
+          _.assign(
+            value,
```
netama, line 69: can we avoid modifying the original value here? (we're not consistent enough in enforcing this, but it is preferred where possible)

packages/adapter-components/src/definitions/system/fetch/element.ts
```diff
@@ -77,13 +77,23 @@ type StandaloneFieldDefinition = {
   nestPathUnderParent?: boolean
 }
 
+// Settings for sorting a list field.
+type SortFieldDefinition = {
+  // The fields to sort by, in order of precedence.
+  // Each string is a dot-separated path to a property of the objects in the list field.
+  // Paths can include reference expressions, e.g. `myList: ['path.to.property.refToAnotherElement.id']`
+  // Selected properties should be env-independent to avoid sorting differences between envs.
+  sortByProperties: string[]
```
netama, line 86: 1. I know you wrote you didn't see an immediate requirement to ascending vs descending from the current cases, but I think we will have some (for example we might want to move here the logic from the [zendesk guide order instances](https://github.com/salto-io/salto/blob/main/packages/zendesk-adapter/src/filters/guide_order/guide_order_utils.ts#L103-L108), sorry I think I just mentioned something but forgot to send you the example). how would that look? might be worth making this a list of objects so that it's more straightforward to extend?
2. no objections to starting with just paths if it's covering everything we've seen so far, just from the examples we saw in other places over the last few months, there's almost always a need for something custom for some case (and then the transformation usually gives some basic out-of-the-box support)

packages/adapter-components/test/filters/sort_lists.test.ts
```diff
@@ -0,0 +1,240 @@
...
+        field1: { sort: { sortByProperties: ['ref.prop1'] } },
+      }),
+    ) as FilterWith<'onFetch'>
+    const mainObjType = new ObjectType({
+      elemID: new ElemID('adapter', 't1'),
+      fields: { field1: { refType: new ListType(BuiltinTypes.UNKNOWN) } },
+    })
+    const instance1 = new InstanceElement('inst1', mainObjType, {
+      field1: [
+        { ref: new ReferenceExpression(new ElemID('adapter', 't2'), { prop1: 2 }) },
```
netama, line 125: almost all references we have are to full elements (usually instances) and not to inner values. how would the path look in such cases? (also can we cover them in at least some of the tests?)

netama (commented):
very cool :) added some comments and questions, lmk if more context will help as I assume some of these are related to things you haven't seen yet

packages/adapter-components/src/filters/sort_lists.ts
Nurdok: Done

packages/adapter-components/test/filters/sort_lists.test.ts
Nurdok: [Your comment on the proposal](https://www.notion.so/SALTO-5565-Allow-sorting-an-array-field-as-an-infra-capability-2f09e083a60248ddbcd162c796990011?d=c22f7fdeb8db4231a50e8e3f8d436e96&pvs=4#70852ef65b8e4093806424ee362cc53d) said:

> elem ids are not usually good for sorting because they’re not that predictable and might differ between fetches and add noise in some surprising cases, so we shouldn’t use them unless we have to

I made the filter forbit using full references without an inner value, to avoid using unstable element IDs which might be a source for subtle bugs. I assumed all instances should have _some_ property that is a "primary key" which can distinguish them from other instances of the same type, like an ID or equivalent. Is this a bad assumption?

packages/adapter-components/src/definitions/system/fetch/element.ts
Nurdok: 1. Added ascending / descending support.
2. Let's start with just paths then :) It'd be easier for me to add customization in a followup PR if needed.

packages/adapter-components/src/filters/sort_lists.ts
Nurdok: Done

packages/adapter-components/src/filters/sort_lists.ts
Nurdok: I could get this to work, but there are a few issues:

1. It requires that the property path includes `.values` if there a reference. In the tests, for example, instead of `'ref.ref2.prop'` we'll need to specify `'ref.values.ref2.values.prop'` which isn't as nice.
2. The custom function parameter for `_.orderBy` doesn't accept an index of the value to sort, while `resolvePath` requires a full element ID from the element to the leaf property, which includes the index in the field being sorted. So I had to lookup the value in the list (`indexOf` below) which is also kind of awkward:  
```
    _.assign(
            value,
            _.orderBy(
              value,
              sortFields.map(({ path }: PropertySortDefinition) => (item: Value) => {
                const nestedID = instance.elemID.createNestedID(
                  field?.name,
                  String(value.indexOf(item)),
                  ...path.split('.'),
                )
                const resolved = resolvePath(instance, nestedID)
                return resolved
              }),
              sortFields.map(({ order }: PropertySortDefinition) => order ?? 'asc'),
            ),
          )
  ```
3. `resolvePath` runs `_.get` on the entire path, and it doesn't have protection from, e.g., users specifying reference expression fields like `ref.elemID.fullName`. In some sense it "leaks" the implementation, while the current `get` function in this PR examines each part in the path and automatically drills down to `value` inside any reference expression to prevent this.

packages/adapter-components/src/filters/sort_lists.ts
Nurdok: Done

packages/adapter-components/test/filters/sort_lists.test.ts
Nurdok: PTAL, I hope I understood correctly this time :)

netama: this one still doesn't look right, let's discuss IRL :)

netama (commented):
partial review, I think some of the threads will be easier IRL

packages/adapter-components/src/filters/sort_lists.ts
netama: discussed IRL, keeping this one

netama (approved):
LGTM 🎉  there are still some minor comments to address, pre-approving for once they are resolved. and no need for release notes :)

PR #5713, opened 2024-04-07
SALTO-4142: add permissions to roles and custom record types

SeggevHaimovich:
Roles and Custom Record Types references each other in the permissions field.
When deploying both a role and a custom record type that one reference the other but the opposite direction is missing the deployment fails.
This PR adds missing permissions in roles/custom record types on preDeploy

Points worth mentioning:
* the problem we fix happens only when deploying both the role and the custom record type, if only one of them is deployed netsuite will handle the corresponding permission on their own.
* We don't handle a broken env: If the user already have for some reason roles and custom record types with unmatched permissions there would be a problem to deploy them. A fetch should fix the problem.
---

_jira ticket: https://salto-io.atlassian.net/browse/SALTO-4142_

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


packages/netsuite-adapter/src/client/client.ts
```diff
@@ -115,13 +115,18 @@ type DependencyInfo = {
...
   startNode: GraphNode<SDFObjectNode>,
   endNode: GraphNode<SDFObjectNode>,
   serviceId: string,
 ): boolean =>
-  (startNode.value.changeType === 'addition' || startNode.value.addedObjects.has(serviceId)) &&
-  startNode.id !== endNode.id
+  startNode.id !== endNode.id &&
+  (startNode.value.changeType === 'addition' ||
+    startNode.value.addedObjects.has(serviceId) ||
+    isRoleToCustomRecordType(startNode, endNode))
```
kami-delaroz, line 129: could you please elaborate more on this change?

packages/netsuite-adapter/src/filters/add_permissions_to_cutomRecord_and_roles.ts
```diff
@@ -0,0 +1,199 @@
...
+
+const filterCreator: LocalFilterCreator = () => ({
+  name: 'addPermissions',
+  preDeploy: async changes => {
+    const customRecordTypeChangedMap = new Map<string, AdditionOrModificationRelevantChange>(
+      changes
+        .filter(isAdditionOrModificationChange)
+        .filter(isObjectTypeChange)
+        .filter(change => isCustomRecordType(change.data.after))
+        .map(custRecordChange => [custRecordChange.data.after.elemID.typeName, custRecordChange]),
```
kami-delaroz, line 179: all "customRecord" should be "customRecordType"

kami-delaroz (approved):
🥇🥇🥇
left a few nit comments

PR #5718, opened 2024-04-07
SALTO-5691: Upgrade Okta to use new infra on fetch

shir-reifenberg:
Upgrade Okta to use new infra on fetch

---

Upgrade from old api definitions to new definitions 🎉 

Issues left to resolve -

- [X] AppGroupAssignment are added and removed
- [X] default policies are renamed on regenerateSaltoIds

Still TODO before merging - 
- [X]  make all tests work
- [X] more manual testing
- [X] elemIDs migration - rebase on PR
- [X]  test assign service ID code after deployment
- [X]  separate GroupAssignment path change to different PR, role assignment config
- [X] add noise reduction for nacl case on fields (this is the only visible change on nacls)

---
_Release Notes_: 

---
_User Notifications_: 


packages/okta-adapter/src/filters/app_logo.ts
```diff
@@ -90,29 +88,22 @@ const getAppLogo = async ({
...
 
-  const idFields = configUtils.getTypeTransformationConfig(
-    APPLICATION_TYPE_NAME,
-    config.apiDefinitions.types,
-    config.apiDefinitions.typeDefaults,
-  ).idFields ?? [app.value.label]
-
-  const name = naclCase(getInstanceName(app.value, idFields, APP_LOGO_TYPE_NAME))
+  // TODO - use new definitions with IDPartsDefinition
+  const name = naclCase(getInstanceName(app.value, ['label'], APP_LOGO_TYPE_NAME))
```
shir-reifenberg, line 100: Now we must use the new definitions cause it's customized for some envs

packages/okta-adapter/src/adapter.ts
```diff
@@ -192,72 +179,75 @@ export default class OktaAdapter implements AdapterOperations {
...
+        shouldAccessPrivateAPIs(this.isOAuthLogin, this.userConfig),
+        getAdminUrl(this.client.baseUrl),
+      ),
+      sources: { openAPI: [OPEN_API_DEFINITIONS] },
+    }
+
+    this.definitions = {
+      ...definitions,
+      fetch: definitionsUtils.mergeWithUserElemIDDefinitions({
+        userElemID: userConfig.fetch.elemID as OktaUserFetchConfig['elemID'],
```
netama, line 197: I think the cast is not needed?

packages/okta-adapter/src/adapter.ts
```diff
@@ -192,72 +179,75 @@ export default class OktaAdapter implements AdapterOperations {
...
-        })),
-      },
+  private async handleClassicEngineOrg(): Promise<definitionsUtils.ConfigChangeSuggestion | undefined> {
+    const { isClassicOrg: isClassicOrgByConfig } = this.userConfig[FETCH_CONFIG]
+    const isClassicOrg = isClassicOrgByConfig ?? (await isClassicEngineOrg(this.client))
+    if (isClassicOrg) {
+      const updatedCustomizations = _.pickBy(
+        this.definitions.fetch.instances.customizations,
+        (_v, key) => !CLASSIC_ENGINE_UNSUPPORTED_TYPES.includes(key),
+      )
```
netama, line 233: nit: I think this can also be
```suggestion
      const updatedCustomizations = _.omit(
        this.definitions.fetch.instances.customizations,
        CLASSIC_ENGINE_UNSUPPORTED_TYPES,
      )
```

packages/okta-adapter/src/deployment.ts
```diff
@@ -137,17 +137,23 @@ export const deployStatusChange = async (
   }
 }
 
-export const assignServiceIdToAdditionChange = (
+export const assignServiceIdToAdditionChange = async (
```
netama, line 140: I'll review this part after the rebases are done, there were some fixes in the infra since this was added
once deploy is upgraded this code will go away, so we can also just "port" it back to the old api definitions format if we want to be on the safe side

(might be easier to discuss IRL :) )

packages/okta-adapter/src/client/client.ts
```diff
@@ -265,15 +264,7 @@ export default class OktaClient extends clientUtils.AdapterHTTPClient<Credential
...
-      log.debug('Received response for resource request %s with status %d', url, status)
-      log.trace(
-        'Full HTTP response for resource %s: %s',
-        url,
-        safeJsonStringify({
-          url,
-          response: data,
-        }),
-      )
+      log.trace('Received response for resource request %s with status %d', url, status)
```
netama, line 267: is this related?

packages/okta-adapter/src/definitions/fetch.ts
```diff
@@ -0,0 +1,1094 @@
...
+
+const NAME_ID_FIELD: definitions.fetch.FieldIDPart = { fieldName: 'name' }
+const DEFAULT_ID_PARTS = [NAME_ID_FIELD]
+
+const DEFAULT_FIELD_CUSTOMIZATIONS: Record<string, definitions.fetch.ElementFieldCustomization> = _.merge(
+  {},
+  DEFAULT_FIELDS_TO_OMIT,
+)
+
+const getPrivateAPICustomizations = ({
```
netama, line 51: nice

packages/okta-adapter/src/config.ts
```diff
@@ -304,26 +200,7 @@ const getPolicyConfig = (): OktaSwaggerApiConfig['types'] => {
...
-  api__v1__groups: {
-    request: {
-      url: '/api/v1/groups',
-      queryParams: { limit: '10000' }, // maximum page size allowed
-    },
-  },
   Group: {
-    transformation: {
-      fieldTypeOverrides: [
-        { fieldName: 'roles', fieldType: 'list<RoleAssignment>' },
```
netama, line 316: I'm probably missing something but what's replacing this (and the standalone def below)?

packages/okta-adapter/src/filters/group_push.ts
```diff
@@ -82,230 +51,24 @@ const PUSH_RULE_SCHEMA = Joi.object({
...
-      }),
-    ),
-  )
-}
-
 /**
- * Fetch group push instances and group push rule instances using private API
+ * Prepare payload for group push and group push rule removals
  */
-const groupPushFilter: FilterCreator = ({ config, adminClient, getElemIdFunc }) => ({
```
netama, line 249: cool

netama (commented):
did a partial pass, sorry about the delay - looks good :) I sampled the defs and added some questions, will sample some more (and lmk if you want more thorough reviews on specific cases), but I think if the fetches are identical then it's probably a better check for the config-to-defs migrations anyway :)
let's review the general upgrade plan as well to see if any additional support is still needed

packages/okta-adapter/src/client/client.ts
shir-reifenberg: no 🙄

packages/okta-adapter/src/config.ts
shir-reifenberg: good catch :)
we're not fetching role assignments for a while, it was just a left over in the config I forgot to remove...

netama (approved):
diff looks good :)

PR #5723, opened 2024-04-07
SALTO-5754: Merge user config with default recursively in adapterConfigFromConfig

shir-reifenberg:
Changed `adapterConfigFromConfig` for merge config the same way existing adapters do
also merged any additional props config might contain

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/adapter-components/src/definitions/user/user_config.ts
```diff
@@ -98,7 +98,7 @@ export const adapterConfigFromConfig = <
   defaultConfig: Co,
 ): Co & UserConfig<TCustomNameMappingOptions> => {
   // TODO extend validations SALTO-5584
-  const adapterConfig = _.defaults({}, config?.value, defaultConfig)
+  const adapterConfig = _.defaultsDeep({}, config?.value, defaultConfig)
```
netama, line 101: there are some potential surprises when defaulting everything recursively since there's no way to "turn off" something that's been set in the default config. maybe use defaults one level deep (so that it conforms with https://github.com/salto-io/salto/blob/main/packages/okta-adapter/src/adapter_creator.ts#L83-L93 , though I'm not sure we need the handling for arrays)?

packages/adapter-components/src/definitions/user/user_config.ts
```diff
@@ -98,7 +103,24 @@ export const adapterConfigFromConfig = <
...
+  )
+
+  const adapterConfig = {
+    fetch,
+    deploy,
+    client,
+    ...additionalConfigProps,
+  }
+
+  return adapterConfig as Co & UserConfig<TCustomNameMappingOptions>
```
shir-reifenberg, line 125: not sure how to avoid casting here

PR #5724, opened 2024-04-07
SALTO-5755: Expose a 'scrollingPagination' function within the infra pagination utils

shoshana-houri:
Intercom public API exposes a [scrolling API](https://developers.intercom.com/docs/references/rest-api/api.intercom.io/Companies/scrollOverAllCompanies/) request, which offers better performance than using the standard [‘list’ endpoint](https://developers.intercom.com/docs/references/rest-api/api.intercom.io/Companies/listAllCompanies/). In this PR we add this scrolling mechanism as part of our pagination functions.

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/fetch/request/pagination/pagination_functions.ts
+++ b/packages/adapter-components/src/fetch/request/pagination/pagination_functions.ts
@@ -208,6 +208,32 @@ export const cursorPagination = ({
   return nextPageCursorPages
 }
 
+export const scrollingPagination = ({
+  scrollingParam,
+  stopCondition,
+}: {
+  scrollingParam: string
+  stopCondition?: (data: unknown) => boolean
+}): PaginationFunction => {
+  const nextPageScrollingPages: PaginationFunction = ({ currentParams, responseData }) => {
+    const scrollParam = _.get(responseData, scrollingParam)
+    if (typeof scrollParam !== 'string' || stopCondition?.(responseData)) {
+      return []
+    }
+    return [
+      _.merge({}, currentParams, {
+        queryParams: {
+          [scrollingParam]: scrollParam,
+          // This is a workaround to to handle our pagination mechanism which expects each pagination request to have a unique query param.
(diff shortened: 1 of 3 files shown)


packages/adapter-components/src/fetch/request/pagination/pagination_functions.ts
```diff
@@ -208,6 +208,32 @@ export const cursorPagination = ({
...
+    if (typeof scrollParam !== 'string' || stopCondition?.(responseData)) {
+      return []
+    }
+    return [
+      _.merge({}, currentParams, {
+        queryParams: {
+          [scrollingParam]: scrollParam,
+          // This is a workaround to to handle our pagination mechanism which expects each pagination request to have a unique query param.
+          // It is not the case for the scrolling param - which is the same for all requests (until the stop condition is met).
+          timestamp: new Date().toISOString(),
```
netama, line 229: are these not supposed to be sent to the service?

shoshana-houri: The timestamp is not required as a query parameter, it's here only as workaround for our infra. Including it doesn't have any side effects (at least from what I have seen).

netama: hmm I think it might be better to fix this the "right" way and have an option in the infra to allow this, since we shouldn't be sending out unneeded args this way. but let's discuss IRL, unless we have another use case it might not be worth solving just for this case

shoshana-houri: Talked f2f - we decided to maintain the current setup for now. However, we will relocate it to the Intercom package instead of under the shared adapter-components, since it's not how we ideally want things to work.

shoshana-houri:
Closing. Context in https://github.com/salto-io/salto/pull/5724#discussion_r1557644677

PR #5726, opened 2024-04-08
SALTO-4094: Fetch priority instance for policy rules

idoamit8:
In this PR I support fetching policy rules priories into an instance. 

---

1. **I will add tests after first review when the tech design will be approve.** 
2. View of the nacl:
image

3. View of the element tree
image

4. Priority without rules will not have a priority rule instance
image

---
_Release Notes_: 
_Okta Adapter:_
* New instances responsible for policy rules priorities were added. 

---
_User Notifications_: 
_Okta Adapter:_
* New instances responsible for policy rules priorities were added.

### DIFF
--- a/packages/okta-adapter/src/adapter.ts
+++ b/packages/okta-adapter/src/adapter.ts
@@ -87,6 +87,7 @@ import profileMappingPropertiesFilter from './filters/profile_mapping_properties
 import profileMappingAdditionFilter from './filters/profile_mapping_addition'
 import profileMappingRemovalFilter from './filters/profile_mapping_removal'
 import omitAuthenticatorMappingFilter from './filters/omit_authenticator_mapping'
+import policyRulePrioritiesFilter from './filters/policy_rule_priorities'
 import groupPushFilter from './filters/group_push'
 import addImportantValues from './filters/add_important_values'
 import groupPushPathFilter from './filters/group_push_path'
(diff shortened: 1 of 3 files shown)


shir-reifenberg (commented):
Overall design looks good 😎  only looked briefly on code
A few thoughts - 
- Should we make a single "Prioritiy" type for all types of policies or create one for each policy type ? don't have a really strong opinion here, WDYT?  also not sure about the policy type in the policy name (sorry if you asked me about it earlier and I miss understand u)
- Do we have a usecase when there are no rules? cause I think all policies have at least the default rule so IMO we should create it with just one rule for consistency, WDYT?

I know it initial design, just keep in mind we'd want to avoid making multiple changes to existing workspaces so only add things when the entire feature is ready (including the deploy support)

idoamit8:
> Overall design looks good 😎 only looked briefly on code A few thoughts -
> 
> * Should we make a single "Prioritiy" type for all types of policies or create one for each policy type ? don't have a really strong opinion here, WDYT?  also not sure about the policy type in the policy name (sorry if you asked me about it earlier and I miss understand u)
> * Do we have a usecase when there are no rules? cause I think all policies have at least the default rule so IMO we should create it with just one rule for consistency, WDYT?
> 
> I know it initial design, just keep in mind we'd want to avoid making multiple changes to existing workspaces so only add things when the entire feature is ready (including the deploy support)

* I don't see any reason to have an order type to each policy type, its a code duplicaiton without gaining anything. 
The only reason to do it is in order prettdy the instance name and I don't think its a good enough reason.
* There is a usecase when there are no rules (I even added a photo of this usecase)

PR #5733, opened 2024-04-09
SALTO-3080 Add template static file support to resolver

adfineberg:
1. move RestoreValuesFunc, restoreChangeElement and restoreValues to adapter-component
2. add template static file logic to restore
3. fix resolve logic for template static file

---

---
_Release Notes_: 

---
_User Notifications_: 


packages/adapter-components/src/restore_utils.ts
```diff
@@ -0,0 +1,161 @@
...
+      if (isEqualResolvedValues(refValue, value)) {
+        return ref
+      }
+    }
+
+    const file = allStaticFilesPaths.get(path.getFullName())
+    if (file !== undefined) {
+      if (file.isTemplate) {
+        return parserUtils.templateExpressionToStaticFile(value, file.filepath)
+      }
```
edenhassid, line 106: everything is moved from the adapter-utils/utils except for this part which is added

packages/adapter-components/test/restore_utils.test.ts
edenhassid, line 1: copied from adapter-utils

packages/adapter-utils/src/utils.ts
```diff
@@ -539,121 +534,6 @@ export type ResolveValuesFunc = <T extends Element>(
   allowEmpty?: boolean,
 ) => Promise<T>
 
-export type RestoreValuesFunc = <T extends Element>(
```
ori-moisis, line 542: why was this moved?

packages/adapter-components/src/resolve_utils.ts
```diff
@@ -32,21 +32,33 @@ import { parserUtils } from '@salto-io/parser'
...
     if (isStaticFile(value)) {
       if (value.isTemplate) {
-        return parserUtils.staticFileToTemplateExpression(value)
+        const templateExpression = await parserUtils.staticFileToTemplateExpression(value)
+        if (templateExpression)
+          templateExpression.parts = await Promise.all(
+            templateExpression?.parts.map(async part =>
+              isReferenceExpression(part) ? resolveReferenceExpression(part) : part,
+            ),
+          )
```
ori-moisis, line 60: if we are starting to resolve template expressions here, we should be consistent about it.
IIUC this code will only resolve the references in template expressions that come from static files, but it won't resolve references in template expressions that came from NaCl

packages/adapter-components/src/restore_utils.ts
```diff
@@ -0,0 +1,161 @@
...
+      const refValue = await getLookUpName({ ref, field, path, element: targetElement })
+      if (isEqualResolvedValues(refValue, value)) {
+        return ref
+      }
+    }
+
+    const file = allStaticFilesPaths.get(path.getFullName())
+    if (file !== undefined) {
+      if (file.isTemplate) {
+        return parserUtils.templateExpressionToStaticFile(value, file.filepath)
```
ori-moisis, line 105: I might be missing something, but it seems like there is no check that the current value is actually supposed to be a template static file (i.e - what if it is just a string and not a template expression?)
won't that be an issue? if the `value` is just a string, won't it cause the code to crash?

packages/adapter-components/src/restore_utils.ts
```diff
@@ -0,0 +1,161 @@
...
+      }
+      const content = file.encoding === 'binary' ? value : Buffer.from(value, file.encoding)
```
ori-moisis, line 107: similar issue (though probably less prone to crashing) - should we check the value is a buffer / string before we make it a buffer and put it in a static file?

packages/adapter-components/src/resolve_utils.ts
edenhassid: references from template expressions are also resolved here, the `valuesReplacer` is called from `transformElement` which if I understand correctly calls `valuesReplacer` for all the values in the template expression at some point.

packages/adapter-components/src/restore_utils.ts
edenhassid: I'll add a test that value is indeed a template expression

packages/adapter-components/src/resolve_utils.ts
ori-moisis: I also thought so, but if that is true, then there is no need to have this code here

edenhassid: why if it is true there is no need to have this code here? we resolve a template static file to template expression and then in this case we don't go inside the template expression like we do if it just appears in the nacl

edenhassid: discussed IRL

packages/adapter-components/src/restore_utils.ts
```diff
@@ -0,0 +1,163 @@
...
+        return ref
+      }
+    }
+
+    const file = allStaticFilesPaths.get(path.getFullName())
+    if (file !== undefined) {
+      if (file.isTemplate && isTemplateExpression(value)) {
+        return parserUtils.templateExpressionToStaticFile(value, file.filepath)
+      }
+      if (isBuffer(value) || isString(value)) {
```
ori-moisis, line 108: what if `file.isTemplate` and `value` is a string?
we would not enter the first condition, and we would enter this one, so IIUC we would end up creating a template static file (because we take the encoding from the original file) that isn't actually a template

edenhassid: I think its valid

PR #5734, opened 2024-04-09
SALTO-3080 Fetch guide theme with references

adfineberg:
Add support to fetch template static files in Zendesk guide themes

---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/zendesk-adapter/src/config.ts
+++ b/packages/zendesk-adapter/src/config.ts
@@ -73,15 +73,21 @@ export type IdLocator = {
   type: string[]
 }
 
-export type Themes =
-  | {
-      javascriptStrategy: 'greedy'
-      javascriptDigitAmount: number
-    }
-  | {
-      javascriptStrategy: 'prefix'
-      javascriptPrefix: string
-    }
+export type Themes = {
+  brands?: string[]
+  referenceOptions: {
+    enableReferenceLookup: boolean
+    javascriptReferenceLookupStrategy?:
+      | {
+          strategy: 'numericValues'
+          minimumDigitAmount: number
+        }
+      | {
+          strategy: 'varNamePrefix'
+          prefix: string
+        }
+  }
+}
 
 export type Guide = {
   brands: string[]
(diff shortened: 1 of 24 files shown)


edenhassid (commented):
wow 🤯 looks great!! 🔥 😎 💯 added some comments and questions :)

packages/zendesk-adapter/src/config.ts
```diff
@@ -73,8 +73,19 @@ export type IdLocator = {
...
+      javascriptDigitAmount: number
+    }
+  | {
+      javascriptStrategy: 'prefix'
+      javascriptPrefix: string
+    }
+
 export type Guide = {
   brands: string[]
+  themes?: Themes
```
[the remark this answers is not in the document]
edenhassid: we will have to add config suggestion since we have users that already fetch themes

packages/zendesk-adapter/src/config.ts
```diff
@@ -2838,6 +2855,68 @@ const IdLocatorType = createMatchingObjectType<IdLocator>({
...
+    minimumDigitAmount: {
+      refType: BuiltinTypes.NUMBER,
+      annotations: {
+        _required: true,
+      },
+    },
+    prefix: {
+      refType: BuiltinTypes.STRING,
+      annotations: {
+        _required: true,
```
edenhassid, line 2878: isn't this optional?

packages/zendesk-adapter/src/config.ts
```diff
@@ -2847,6 +2926,9 @@ const GuideType = createMatchingObjectType<Guide>({
         _required: true,
       },
     },
+    themes: {
+      refType: ThemesType,
+    },
     themesForBrands: {
```
edenhassid, line 2932: maybe also add a comment here that this is deprecated

packages/zendesk-adapter/src/filters/utils.ts
```diff
@@ -143,9 +143,9 @@ export const isCorrectConditions = (value: unknown, typeName: string): value is
 const getBrandsForFilter = (
   elements: InstanceElement[],
   fetchConfig: ZendeskFetchConfig,
-  filter: 'brands' | 'themesForBrands',
+  filter: 'brands' | 'themesForBrands' | 'themes.brands',
```
edenhassid, line 146: why do we have 'brands' here?

edenhassid (commented):
LGTM! 🔥 🤯 💯  added some comments, pre-approving 😄

edenhassid (approved):

packages/zendesk-adapter/src/config.ts
adfineberg: Yeah, but the type system in the config doesn't recognize it (and the `minimumDigitAmount`) as optional from the conditional type... Do you think it's reasonable to leave it like this?

packages/zendesk-adapter/src/filters/utils.ts
adfineberg: This method is both for the brand filter and the guide theme brand filter

packages/zendesk-adapter/src/config.ts
edenhassid: I am not sure how the `_require` work and if it will cause errors if this field doesn't exist

PR #5738, opened 2024-04-09
SALTO-4094: Deploy priority instance for policy

idoamit8:
In this PR:
1. I added a filter that fetches and deploys policy rules priories into an instance.
2. Added a weak reference mechanism to allow deploying order without all the instances in it.  

---

1. **I will add tests after first review when the tech design will be approve.** 
3. View of the nacl:
image

4. View of the element tree
image
@shir-reifenberg I don't think that the priority should actually be inside its own folder. WDYT?

---
_Release Notes_: 
_Okta Adapter:_
* New instances responsible for policy rules priorities were added. 

---
_User Notifications_: 
_Okta Adapter:_
* New instances responsible for policy rules priorities were added.

### DIFF
--- a/packages/okta-adapter/src/filters/policy_rule_priority.ts
+++ b/packages/okta-adapter/src/filters/policy_rule_priority.ts
@@ -27,12 +27,12 @@ import {
   isInstanceElement,
   isModificationChange,
   isReferenceExpression,
-  ElemIdGetter,
 } from '@salto-io/adapter-api'
 import _ from 'lodash'
 import { collections } from '@salto-io/lowerdash'
 import { getParents, invertNaclCase, naclCase, pathNaclCase } from '@salto-io/adapter-utils'
-import { fetch, elements as adapterElements, config as configUtils } from '@salto-io/adapter-components'
+import { fetch, elements as adapterElements } from '@salto-io/adapter-components'
+import { logger } from '@salto-io/logging'
 import { FilterCreator } from '../filter'
 import {
   ACCESS_POLICY_RULE_PRIORITY_TYPE_NAME,
@@ -57,15 +57,14 @@ import {
 } from '../constants'
 import { deployChanges } from '../deployment'
 import OktaClient from '../client/client'
-import { API_DEFINITIONS_CONFIG, OktaConfig } from '../config'
+import { OktaConfig } from '../config'
 
+const log = logger(module)
 const { awu } = collections.asynciterable
 const { createUrl } = fetch.resource
-const { toBasicInstance } = adapterElements
-const { getTransformationConfigByType } = configUtils
-
 const ALL_SUPPORTED_POLICY_NAMES = POLICY_TYPE_NAMES.concat([AUTHORIZATION_POLICY])
 const ALL_SUPPORTED_POLICY_RULE_NAMES = POLICY_RULE_TYPE_NAMES.concat([AUTHORIZATION_POLICY_RULE])
+// Automation is not included in the list of supported policy rules because it is not supported
 const POLICY_NAME_TO_PRIORITY_NAME: Record<string, string> = {
   [ACCESS_POLICY_TYPE_NAME]: ACCESS_POLICY_RULE_PRIORITY_TYPE_NAME,
   [IDP_POLICY_TYPE_NAME]: IDP_RULE_PRIORITY_TYPE_NAME,
@@ -89,27 +88,18 @@ const createPriorityType = (typeName: string): ObjectType =>
         annotations: { [CORE_ANNOTATIONS.CREATABLE]: true, [CORE_ANNOTATIONS.UPDATABLE]: true },
       },
     },
-    path: undefined,
-    annotations: {
-      [CORE_ANNOTATIONS.CREATABLE]: true,
-      [CORE_ANNOTATIONS.UPDATABLE]: true,
-      [CORE_ANNOTATIONS.DELETABLE]: true,
-    },
+    path: [OKTA, adapterElements.TYPES_PATH, typeName],
   })
 
 const createPolicyRulePriorityInstance = ({
   rules,
   type,
   policy,
-  getElemIdFunc,
-  config,
 }: {
   rules: InstanceElement[]
   type: ObjectType
   policy: InstanceElement
-  getElemIdFunc?: ElemIdGetter
-  config: OktaConfig
-}): Promise<InstanceElement> => {
+}): InstanceElement => {
   const name = naclCase(`${invertNaclCase(policy.elemID.name)}_priority`)
   const defaultRule = rules.find(rule => rule.value.system === true)
   const value = {
@@ -121,16 +111,8 @@ const createPolicyRulePriorityInstance = ({
   const fullValue = defaultRule
     ? { ...value, defaultRule: new ReferenceExpression(defaultRule.elemID, defaultRule) }
     : value
-
-  return toBasicInstance({
-    entry: fullValue,
-    type,
-    transformationConfigByType: getTransformationConfigByType(config[API_DEFINITIONS_CONFIG].types),
-    transformationDefaultConfig: config[API_DEFINITIONS_CONFIG].typeDefaults.transformation,
-    parent: policy,
-    defaultName: name,
-    nestedPath: [...(policy.path ?? []).slice(2, (policy.path ?? []).length - 1), pathNaclCase(name)],
-    getElemIdFunc,
+  return new InstanceElement(name, type, fullValue, [...(policy.path ?? []).slice(0, -1), pathNaclCase(name)], {
+    [CORE_ANNOTATIONS.PARENT]: new ReferenceExpression(policy.elemID, policy),
   })
 }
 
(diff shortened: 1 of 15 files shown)


packages/okta-adapter/src/filters/policy_rule_priority.ts
```diff
@@ -0,0 +1,249 @@
...
+        refType: new ListType(BuiltinTypes.NUMBER),
+        annotations: { [CORE_ANNOTATIONS.CREATABLE]: true, [CORE_ANNOTATIONS.UPDATABLE]: true },
+      },
+      defaultValue: {
+        refType: BuiltinTypes.STRING,
+        annotations: { [CORE_ANNOTATIONS.CREATABLE]: true, [CORE_ANNOTATIONS.UPDATABLE]: true },
+      },
+    },
+    path: undefined,
+    annotations: {
```
shir-reifenberg, line 93: isn't this defaults to true ? and we should only put `false` on the things that create be created/updated/deleted ?
Also, deletion is not implemented, no?

packages/okta-adapter/src/filters/policy_rule_priority.ts
```diff
@@ -0,0 +1,249 @@
...
+import { API_DEFINITIONS_CONFIG, OktaConfig } from '../config'
+
+const { awu } = collections.asynciterable
+const { createUrl } = fetch.resource
+const { toBasicInstance } = adapterElements
+const { getTransformationConfigByType } = configUtils
+
+const ALL_SUPPORTED_POLICY_NAMES = POLICY_TYPE_NAMES.concat([AUTHORIZATION_POLICY])
+const ALL_SUPPORTED_POLICY_RULE_NAMES = POLICY_RULE_TYPE_NAMES.concat([AUTHORIZATION_POLICY_RULE])
+const POLICY_NAME_TO_PRIORITY_NAME: Record<string, string> = {
```
shir-reifenberg, line 69: maybe add a note about the policies that are not included ?

packages/okta-adapter/src/filters/policy_rule_priority.ts
```diff
@@ -0,0 +1,249 @@
...
+      ALL_SUPPORTED_POLICY_RULE_NAMES.includes(instance.elemID.typeName),
+    )
+    const policiesToPoliciesRules = _.groupBy(policiesRules.filter(rule => isReferenceExpression(getParents(rule)[0])), rule =>
+      // AuthorizationServerPolicyRule has two parents. 
+      getParents(rule)[0].elemID.getFullName(),
+    )
+    const priorityTypes = POLICY_RULE_PRIORITY_TYPE_NAMES.map(name => createPriorityType(name))
+    priorityTypes.forEach(type => elements.push(type))
+    const priorityNameToPriorityType = _.keyBy(priorityTypes, type => type.elemID.typeName)
+    await awu(Object.entries(policiesToPoliciesRules)).forEach(async ([policyName, rules]) => {
```
shir-reifenberg, line 194: as you don't use `policiesToPoliciesRules` as a map, you can just instead create an object 
`{ policy: InstanceElement;  rule: InstanceElement[] }`, and iterate it instead of trying to get the parent policy which you already had in line 187

packages/okta-adapter/src/filters/policy_rule_priority.ts
```diff
@@ -0,0 +1,249 @@
...
+  const deployUrl = deployRequest?.url
+  if (deployUrl === undefined) {
+    throw new Error('Failed to deploy priority change due to missing url')
+  }
+  const url = createUrl({ instance: rule.value, url: deployUrl, additionalUrlVars: { ruleId: id, policyId } })
+
+  await client.put({ url, data })
+}
+
+/* Manages the priorities of policyRules within each Policy
```
shir-reifenberg, line 173: Can you please extend this a bit - 
- why we have a field for default policy
- what is happening in deploy
- etc

packages/okta-adapter/src/filters/policy_rule_priority.ts
```diff
@@ -0,0 +1,249 @@
...
+      elements.push(priorityInstance)
+    })
+    // Remove priority field from the policy rules
+    policiesRules.forEach(rule => {
+      delete rule.value.priority
+    })
+  },
+  deploy: async changes => {
+    const [relevantChanges, leftoverChanges] = _.partition(changes, change =>
+      POLICY_RULE_PRIORITY_TYPE_NAMES.includes(getChangeData(change).elemID.typeName),
```
shir-reifenberg, line 213: nit: create a `Set` for it (to be used also in fetch)

packages/okta-adapter/src/filters/policy_rule_priority.ts
```diff
@@ -0,0 +1,249 @@
...
+            await deployPriorityChange({
+              client,
+              priority,
+              rule,
+              config,
+            })
+          })
+      }
+      if (isModificationChange(change)) {
+        const positionsBefore = change.data.before.value.priorities.filter(isReferenceExpression)
```
shir-reifenberg, line 230: `priorities` might be undefined

packages/okta-adapter/src/filters/policy_rule_priority.ts
```diff
@@ -0,0 +1,249 @@
...
+
+const deployPriorityChange = async ({
+  client,
+  priority,
+  rule,
+  config,
+}: {
+  client: OktaClient
+  priority: number
+  rule: ReferenceExpression
```
shir-reifenberg, line 153: any reason not to pass the InstanceElement instead of the ref ?

packages/okta-adapter/src/filters/policy_rule_priority.ts
```diff
@@ -0,0 +1,249 @@
...
+  config: OktaConfig
+}): Promise<void> => {
+  const { id } = rule.value.value
+  const { type } = rule.value.value
+  const policy = getParents(rule.value)[0].value
```
shir-reifenberg, line 158: policy might no exist?

packages/okta-adapter/src/weak_references/policy_rule_priorities.ts
```diff
@@ -0,0 +1,110 @@
...
+        if (priorities === undefined) {
+          return undefined
+        }
+        const fixedInstance = instance.clone()
+        fixedInstance.value.priorities = await awu(priorities)
+          .filter(
+            async rule =>
+              rule !== undefined &&
+              isReferenceExpression(rule) &&
+              // eslint-disable-next-line no-return-await
```
shir-reifenberg, line 83: needed ?

packages/okta-adapter/src/weak_references/policy_rule_priorities.ts
```diff
@@ -0,0 +1,110 @@
...
+
+        return fixedInstance
+      })
+      .filter(values.isDefined)
+      .toArray()
+
+    const errors = fixedElements.map(instance => ({
+      elemID: instance.elemID.createNestedID('priorities'),
+      severity: 'Info' as const,
+      message: `${instance.elemID.typeName} will be deployed without priorities defined on non-existing fields`,
```
shir-reifenberg, line 100: TBH I don't understand what does it mean 😅

packages/okta-adapter/src/weak_references/policy_rule_priorities.ts
```diff
@@ -0,0 +1,110 @@
...
+      detailedMessage:
```
shir-reifenberg, line 101: Maybe align with the message we have in zendesk with adjustments? (`order_elements.ts`)

packages/okta-adapter/src/weak_references/policy_rule_priorities.ts
```diff
@@ -0,0 +1,110 @@
...
+ */
+const getPolicyPriorityReferences: GetCustomReferencesFunc = async elements =>
+  awu(elements)
+    .filter(isInstanceElement)
+    .filter(instance => POLICY_RULE_PRIORITY_TYPE_NAMES.includes(instance.elemID.typeName))
+    .flatMap(getFieldReferences)
+    .toArray()
+
+/**
+ * Remove invalid rules (not references or missing references) from policyPriority.
```
shir-reifenberg, line 64: Please be more specific

shir-reifenberg (commented):
Looks great! 👯  added some comments
some other general thoughts -
- do we want to add alias to the created instances? (in alias filter)
- what would be the behavior if the user selected to deploy a policy without its order element? rules will be created in random order? should we add a validator for this?
- Please verify if a noise reduction is required and add one if needed
- We should also handle policy priorities when relevant, it can defently be in a different PR but they should be included in the same deployment

packages/okta-adapter/src/filters/policy_rule_priority.ts
```diff
@@ -0,0 +1,249 @@
...
+}): Promise<InstanceElement> => {
+  const name = naclCase(`${invertNaclCase(policy.elemID.name)}_priority`)
+  const defaultRule = rules.find(rule => rule.value.system === true)
+  const value = {
+    priorities: rules
+      .filter(rule => rule.value.system !== true)
+      .sort((a, b) => a.value.priority - b.value.priority)
+      .map(inst => new ReferenceExpression(inst.elemID, inst)),
+  }
+  const fullValue = defaultRule
```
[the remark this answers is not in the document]
idoamit8: I don't think we should log an error. Its to specific. 
two policies of the same type can have different behavior. One has default rule and the other don't.

packages/okta-adapter/src/filters/policy_rule_priority.ts
idoamit8: Deletion is fake. We aren't really do anything with the instance.

packages/okta-adapter/src/filters/policy_rule_priority.ts
idoamit8: So why we don't save it as set in constants?

packages/okta-adapter/src/weak_references/policy_rule_priorities.ts
idoamit8: Yea

packages/okta-adapter/src/weak_references/policy_rule_priorities.ts
idoamit8: Changed 🦸

packages/okta-adapter/src/weak_references/policy_rule_priorities.ts
idoamit8: WDYM?

idoamit8:
> Looks great! 👯 added some comments some other general thoughts -
> 
> * do we want to add alias to the created instances? (in alias filter)
> * how do we handle partial failures ? if deploying priority for one rule succeeded and other fails we'll get some odd state in the target env and we should probably show a more specific error for this case and make it look like a partial failure, WDYT?
> * what would be the behavior if the user selected to deploy a policy without its order element? rules will be created in random order? should we add a validator for this?
> * Please verify if a noise reduction is required and add one if needed
> * We should also handle policy priorities when relevant, it can definitely be in a different PR but they should be included in the same deployment
> * the release notes and user notifications in this case shouldn't be exactly the same.. in release notes we should elaborate what this feature do and mention if it fix any known gap or bug
> 
> and I saw I didn't reply earlier, but I agree priority instance shouldn't be in its own folder :)

* we can add an alias with the name of the policy and suffix priority?
* Currently we don't handle partial failures. We can think on a way to support this or start without it, but it is not intuitive. 
* Yes, it will be random. But I don't think we need to add CV. 
* What should be the release notes then? what is exectly the bug we want to remove? 

I will also writing noise reduction for deleting the priority.

packages/okta-adapter/src/filters/policy_rule_priority.ts
```diff
@@ -0,0 +1,252 @@
...
+          return acc
+        },
+        {} as {
+          [key: string]: {
+            policy: InstanceElement
+            rules: InstanceElement[]
+          }
+        },
+      ),
+    )
```
shir-reifenberg, line 196: nit - personally i think reduce is sometimes difficult to read, and also I think we can easily avoid the casting and add using push here

I think this should work, but you can leave with reduce if its gets complicated (didn't include the entire logic)
```suggestion
        const policyAndRules = _(policiesRules).map(rule => {
      const parent = getParents(rule).find(isParentPolicy) // your code here...
      return parent ? ({ rule, parent }) : undefined
    })
    .filter(values.isDefined)
    .groupBy(({ parent }) => parent?.elemID.getFullName())
    .mapValues(v => v.map(({ rule }) => rule))
```

packages/okta-adapter/src/filters/policy_rule_priority.ts
```diff
@@ -0,0 +1,252 @@
...
+    const priorityTypes = POLICY_RULE_PRIORITY_TYPE_NAMES.map(name => createPriorityType(name))
+    priorityTypes.forEach(type => elements.push(type))
+    const priorityNameToPriorityType = _.keyBy(priorityTypes, type => type.elemID.typeName)
+    await awu(policyAndrules).forEach(async ({ policy, rules }) => {
```
shir-reifenberg, line 200: is there anything async?

packages/okta-adapter/src/filters/policy_rule_priority.ts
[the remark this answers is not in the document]
shir-reifenberg: so if it's a legit scenario please add a comment about when we expect to have default rule and when not ?

packages/okta-adapter/src/filters/policy_rule_priority.ts
```diff
@@ -0,0 +1,252 @@
...
+  config,
+}: {
+  client: OktaClient
+  priority: number
+  rule: InstanceElement
+  config: OktaConfig
+}): Promise<void> => {
+  const { id } = rule.value
+  const { type } = rule.value
+  const policy = getParents(rule).find(parent => ALL_SUPPORTED_POLICY_NAMES.includes(parent.elemID.typeName))?.value
```
shir-reifenberg, line 140: can probably extract to getParentPolicy that would just return getParents[0] for all policies expect for authorization policy and would check for the right parent for Authorization policy rule

packages/okta-adapter/src/weak_references/policy_rule_priorities.ts
```diff
@@ -0,0 +1,110 @@
...
+
+const { awu } = collections.asynciterable
+
+const log = logger(module)
+
+const getFieldReferences = async (instance: InstanceElement): Promise<ReferenceInfo[]> => {
+  const { priorities } = instance.value
+  if (priorities === undefined || !Array.isArray(priorities)) {
+    // priorities can be undefined if the policy has no custom rules
+    log.debug(
```
shir-reifenberg, line 37: let's move to `trace` since it's a valid scenario ?

packages/okta-adapter/src/weak_references/policy_rule_priorities.ts
```diff
@@ -0,0 +1,110 @@
...
+import { collections, values } from '@salto-io/lowerdash'
+import { logger } from '@salto-io/logging'
+import { POLICY_RULE_PRIORITY_TYPE_NAMES } from '../constants'
+import { WeakReferencesHandler } from './weak_references_handler'
+
+const { awu } = collections.asynciterable
+
+const log = logger(module)
+
+const getFieldReferences = async (instance: InstanceElement): Promise<ReferenceInfo[]> => {
```
shir-reifenberg, line 33: just a suggestion, but maybe use a more descriptive name ?
```suggestion
const markRuleAsWeakReference = async (instance: InstanceElement): Promise<ReferenceInfo[]> => {
```

shir-reifenberg (approved):
Looks great!!!
added some comment, nothing too major so approving
re you comments in the previous thread -
- Alias - your suggestion sounds great, please add it 🙏 
- partial failures - agree, maybe worth opening a ticket and link to this one?
- validator for policy without priority - not sure I agree.. I think most cases will be addition of policy and all its rules so I think that if the priority is not deployed it should be visible somehow.. let's discuss IRL
- Something about now able to support changes in priorities correctly + if the bug assigned to you now resolved it's worth mentioning
-

packages/okta-adapter/src/filters/policy_rule_priority.ts
idoamit8: Nope :)
It's a leftover from `toBasicInstance`

PR #5741, opened 2024-04-10
SALTO-5406 pass in custom serialization mapping to delayed deploy resolution as well

netama:
Changes resolved right before deploy did not have access to the resolver that had the custom definitions, causing them to incorrectly resolve as `fullValue`.

(will add more unittests next week as part of extending the coverage in this area in general)

---
_Release Notes_: 
None (not used in existing adapters)

---
_User Notifications_: 
None


Uri-Bar (approved):
thanks, also tested the fix locally

PR #5745, opened 2024-04-10
SALTO-5742 confluence template content to be template expression

omrifarkash:
---

---
_Release Notes_: 
confluence: templates body will be template expression to support references inside the body

---
_User Notifications_: 
confluence: templates body will be template expression to support references inside the body


packages/confluence-adapter/src/adapter_creator.ts
```diff
@@ -44,6 +45,16 @@ export const adapter = createAdapter<Credentials, Options, UserConfig>({
...
+    customizeFilterCreators: (config, definitions, resolverCreator) => [
+      ...Object.values(
+        filters.createCommonFilters<Options, UserConfig>({
+          config,
+          definitions,
+          fieldReferenceResolverCreator: resolverCreator,
+        }),
+      ),
+      transformTemplateBodyToTemplateExpressionFilterCreator,
+    ],
```
netama, line 57: after this can be 
```suggestion
    customizeFilterCreators: args => ({
      ...filters.createCommonFilters<Options, UserConfig>(args),
      transformTemplateBodyToTemplateExpressionFilterCreator,
    }),

```

packages/confluence-adapter/src/filters/transform_template_body_to_template_expression.ts
```diff
@@ -0,0 +1,203 @@
...
+}
+
+const prepRef = (ref: ReferenceExpression): TemplatePart => {
+  if (ref.value instanceof UnresolvedReference) {
+    log.trace(
+      'prepRef received a part as unresolved reference, returning an empty string, instance fullName: %s ',
+      ref.elemID.getFullName(),
+    )
+    return ''
+  }
```
omrifarkash, line 119: uncovered lines

packages/confluence-adapter/src/filters/transform_template_body_to_template_expression.ts
```diff
@@ -0,0 +1,203 @@
...
+  if (ref.value.elemID.typeName === SPACE_TYPE_NAME && _.isString(ref.value.value?.key)) {
+    return ref.value.value.key
+  }
+  if (ref.value.elemID.typeName === PAGE_TYPE_NAME && _.isString(ref.value.value?.title)) {
+    return ref.value.value.title
+  }
+  log.warn('prepRef received a part that is not a space or page reference %o', ref)
+  // fallback to the original reference
+  return ref
```
omrifarkash, line 128: uncovered lines

packages/confluence-adapter/src/filters/transform_template_body_to_template_expression.ts
```diff
@@ -0,0 +1,203 @@
...
+          pageBySpaceFullNameAndTitle: {},
+        },
+      )
+
+      const templateInstances = instances.filter(inst => TEMPLATE_TYPE_NAMES.includes(inst.elemID.typeName))
+      templateInstances.forEach(templateInst => {
+        const bodyValue = _.get(templateInst.value, 'body.storage.value')
+        if (!_.isString(bodyValue)) {
+          log.warn('Body value is not a string for template instance %s', templateInst.elemID.getFullName())
+          return
```
omrifarkash, line 164: uncovered lines

omrifarkash:
There are some uncovered lines and branches. All of them are cases that shouldn't happen and hard to be tested with valid instances

packages/confluence-adapter/src/adapter_creator.ts
```diff
@@ -45,6 +46,10 @@ export const adapter = createAdapter<Credentials, Options, UserConfig>({
   operationsCustomizations: {
     connectionCreatorFromConfig: () => createConnection,
     credentialsFromConfig: defaultCredentialsFromConfig,
+    customizeFilterCreators: args => ({
+      ...filters.createCommonFilters<Options, UserConfig>(args),
+      transformTemplateBodyToTemplateExpressionFilterCreator,
```
netama, line 51: please don't forget to add the comment if the order is important (IIUC it is)

packages/confluence-adapter/src/error_utils.ts
```diff
@@ -18,26 +18,39 @@ import _ from 'lodash'
 import { isSaltoElementError, isSaltoError } from '@salto-io/adapter-api'
 import { deployment } from '@salto-io/adapter-components'
 
+const isJavaNullPointerErrorOrUndefined = (err: Error): boolean => {
+  const status = _.get(err, 'response.status')
+  const message = _.get(err, 'response.data.message')
+  return status === 500 && message.startsWith('java.lang.NullPointerException: Cannot invoke')
+}
+
+const suppressedErrorsCheckerList: ((err: Error) => boolean)[] = [isJavaNullPointerErrorOrUndefined]
```
netama, line 27: nits:
1. maybe define a type for the error checker signature?
2. this is a constant so should be 
```suggestion
const SUPPRESSED_ERRORS_CHECKER_LIST: ((err: Error) => boolean)[] = [isJavaNullPointerErrorOrUndefined]
```

packages/confluence-adapter/src/error_utils.ts
```diff
@@ -18,26 +18,39 @@ import _ from 'lodash'
...
-    const status = _.get(err, 'response.status')
-    return (
-      status === 409 &&
-      Array.isArray(errorsArray) &&
-      errorsArray.length > 0 &&
-      _.isString(errorsArray[0].title) &&
-      errorsArray[0].title.startsWith('Version')
-    )
+  if (shouldSuppressError(error)) {
+    return undefined
```
netama, line 52: maybe add a log saying what error is suppressed?

packages/confluence-adapter/src/filters/transform_template_body_to_template_expression.ts
```diff
@@ -0,0 +1,203 @@
...
+  if (spaceMatches !== null) {
+    return handleSpaceRefMatch(spaceMatches, indices.spaceByKey, expression)
+  }
+  return expression
+}
+
+const prepRef = (ref: ReferenceExpression): TemplatePart => {
+  if (ref.value instanceof UnresolvedReference) {
+    log.trace(
+      'prepRef received a part as unresolved reference, returning an empty string, instance fullName: %s ',
```
netama, line 115: nits: 
1. I think this should be a debug log
2. we should also log the unresolved reference's elem id (there can be many)
3. 
```suggestion
      'prepRef received a part as unresolved reference, returning an empty string, instance fullName: %s',
```

packages/confluence-adapter/src/filters/transform_template_body_to_template_expression.ts
netama: why is this one hard to cover?

packages/confluence-adapter/src/filters/transform_template_body_to_template_expression.ts
netama: why is this hard to cover?

packages/confluence-adapter/src/filters/transform_template_body_to_template_expression.ts
```diff
@@ -0,0 +1,203 @@
...
+const SPACE_REF_REGEX = /(<ri:space\sri:space-key="[^"]*"\s*\/>)/
+const SPLIT_SPACE_REF_REGEX = /(<ri:space\sri:space-key=")([^"]*)("\s*\/>)/
+
+const handlePageRefMatch = (
+  matches: RegExpMatchArray,
+  indices: PossibleRefsInTemplateIndices,
+  fallback: string,
+): TemplatePart | TemplatePart[] => {
+  const { spaceByKey, pageBySpaceFullNameAndTitle } = indices
+  const [, spaceKey, spaceKeyValue, contentTitle, contentTitleValue, versionAtSave] = matches
```
netama, line 65: can you please explain why it's ok to drop the first part (here and below)?

packages/confluence-adapter/src/filters/transform_template_body_to_template_expression.ts
```diff
@@ -0,0 +1,203 @@
...
+            replaceTemplatesWithValues(
+              { values: [instance.value.body?.storage], fieldName: 'value' },
+              deployTemplateMapping,
+              prepRef,
+            )
+            return instance
+          })
+        })
+    },
+    onDeploy: async changes => {
```
netama, line 188: let's add a TODO to move these to an `adjust` function? (no need to do it now, but should be clear that this is just reverting the `preDeploy` and we have better ways to do it now)

packages/confluence-adapter/test/adapter.test.ts
```diff
@@ -201,7 +202,7 @@ describe('adapter', () => {
       results.push(
         await operations.deploy({
           changeGroup: {
-            groupID: 'space',
+            groupID: SPACE_TYPE_NAME,
```
netama, line 205: nit: I think it might be better to keep these tests without referencing the constants

packages/confluence-adapter/test/filters/transform_template_body_to_template_expression.test.ts
```diff
@@ -0,0 +1,210 @@
...
+    })
+  })
+
+  describe('on fetch', () => {
+    const checkInstanceAfterFilter = (instanceName: string, elements: InstanceElement[]): void => {
+      const instance = elements.find(e => e.elemID.name === instanceName)
+      const value = instance?.value.body.storage.value
+      expect(value).toEqual(getTemplateBody(instanceName, false))
+    }
+    it('should create template expression when there are references on template body', async () => {
```
netama, line 162: maybe split the "failure" cases into another test case?

packages/confluence-adapter/test/filters/transform_template_body_to_template_expression.test.ts
```diff
@@ -0,0 +1,210 @@
...
+      checkChangeAfterFilter('templateWithoutReferences', templateAdditionChanges, false)
+      checkChangeAfterFilter('templateWithPageReference', templateAdditionChanges, false)
+      checkChangeAfterFilter('templateWithRefToPageWhichCannotBeFound', templateAdditionChanges, false)
+      checkChangeAfterFilter('templateWithRefToSpace', templateAdditionChanges, false)
+    })
+    it('should work correctly on modification', async () => {
+      const templateAdditionChanges = elements
+        .filter(e => TEMPLATE_TYPE_NAMES.includes(e.elemID.typeName))
+        .map(template => toChange({ before: template, after: template }))
+      await filter.preDeploy?.(templateAdditionChanges)
```
netama, line 203: let's also add tests for the `onDeploy`?

netama (commented):
looks good 💪 , added some comments

packages/confluence-adapter/src/filters/transform_template_body_to_template_expression.ts
omrifarkash: This code is from Zendesk, should I change it there as well?

packages/confluence-adapter/test/adapter.test.ts
omrifarkash: Why? don't we want to control the typeName from one place?

packages/confluence-adapter/test/filters/transform_template_body_to_template_expression.test.ts
omrifarkash: There are both tested on the same block. I saw we are doing it on other filter test files. Should I split them?

packages/confluence-adapter/src/filters/transform_template_body_to_template_expression.ts
omrifarkash: Actually I think this part will never be reached. (Copied it from Zendesk)
prepRef won't get something that isn't instance of ReferenceExpression
image

packages/confluence-adapter/src/filters/transform_template_body_to_template_expression.ts
omrifarkash: I played with it a bit and it wasn't trivial. I can keep trying though

packages/confluence-adapter/src/filters/transform_template_body_to_template_expression.ts
netama: let's change only this one for now, I assume it was noisy in zendesk but let's see what happens here because it might indicate a bug (in zenedesk it's probably due to missing references which are unresolved in valid scenarios)

packages/confluence-adapter/test/adapter.test.ts
netama: more likely to catch issues if they come up in the future. but up to you

packages/confluence-adapter/test/filters/transform_template_body_to_template_expression.test.ts
netama: I don't see it at all in this test, what am I missing?

packages/confluence-adapter/src/filters/transform_template_body_to_template_expression.ts
netama: doesn't it just need to be a reference to an instance from some other type?

packages/confluence-adapter/src/filters/transform_template_body_to_template_expression.ts
omrifarkash: Maybe we can remove it? (I originally copied it from Zendesk)

netama (approved):
approving to unblock

PR #5747, opened 2024-04-10
SALTO-5392 Simplify filter customization

netama:
Simplify interfaces for defining custom filters in the wrapper - for use in https://github.com/salto-io/salto/pull/5745 

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/adapter-components/src/adapter-wrapper/adapter_creator.ts
```diff
@@ -110,16 +111,13 @@ export const createAdapter = <
...
-              : Object.values(
-                  createCommonFilters<Options, Co>({
-                    config,
-                    definitions,
-                    fieldReferenceResolverCreator: resolverCreator,
-                  }),
-                ),
+          filterCreators: Object.values(
+            (customizeFilterCreators ?? createCommonFilters)({
+              config,
```
omrifarkash, line 116: Maybe we can merge `createCommonFilters` and `customizeFilterCreators` results?
In this way adapter can define custom filters and still get the common filter. If adapter wants to control the order of some common filter, it can override it in `customizeFilterCreators`

WDYT?

omrifarkash (approved):

packages/adapter-components/src/adapter-wrapper/adapter_creator.ts
netama: we talked about this one IRL briefly, we have cases where we don't want all the common filters or we want a different order so we'd need some more control around that. I think we can start with this for now and once we haev a few more cases we can generalize, it's a single line in the adapters so won't take much to change later

netama: will continue the discussion offline to unblock the next PR

PR #5753, opened 2024-04-11
SALTO-5786 Change duplicate id message

kami-delaroz:
Change duplicate id message for better visualization

---

None

---
_Release Notes_: 
duplicate id message in Jira was changed for better visualization

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/duplicate_ids.ts
+++ b/packages/jira-adapter/src/filters/duplicate_ids.ts
@@ -59,11 +59,13 @@ const filter: FilterCreator = ({ config }) => ({
       )
     })
 
+    const prettifiesName = (instance: InstanceElement): string =>
+      instance.annotations[CORE_ANNOTATIONS.ALIAS] !== undefined
+      ? instance.annotations[CORE_ANNOTATIONS.ALIAS]
+      : instance.elemID.name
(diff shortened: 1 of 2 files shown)


packages/jira-adapter/src/filters/duplicate_ids.ts
```diff
@@ -59,12 +59,20 @@ const filter: FilterCreator = ({ config }) => ({
       )
     })
 
+    const duplicateInstanceNames = _.uniq(
+      duplicateInstances.flatMap(instance =>
+        instance.annotations[CORE_ANNOTATIONS.ALIAS] !== undefined
+          ? instance.annotations[CORE_ANNOTATIONS.ALIAS]
+          : instance.elemID.getFullName(),
```
shayc331, line 66: maybe we can use `elemID.name`, then the name wont contain `jira.objectType.instance`

shayc331 (approved):
🥇

packages/jira-adapter/src/filters/duplicate_ids.ts
kami-delaroz: this was the origin behavior, the request was to use alias name if exists.
honestly I'm not sure what's better, but prefer to leave it the way it was

PR #5756, opened 2024-04-11
SALTO-5751 Remove elements and numbers from change validator messages

benisraeladi:
Remove numbers and names from change validator messages so that they will be groupable

---

The information still exists in the detailed message field of the change validator.

---
_Release Notes_: _None_ 

---
_User Notifications_: _None_


edenhassid (approved):
amazing!! 🔥 💯 🏆

PR #5757, opened 2024-04-11
SALTO-5630: Refactoring SFDC custom references to be more similar to other adapters

yelly:
No new logic yet, but this gives us a natural place to implement the profiles fix elements.

---

---
_Release Notes_: 
None.

---
_User Notifications_: 
None.


packages/salesforce-adapter/src/weak_references/index.ts
tamtamirr, line 1: Not a fan of having code in `index.ts` files.
I suggest we move the content of this file into `weak_references_handler.ts`

tamtamirr (commented):
A small comment. 🙏

packages/salesforce-adapter/src/weak_references/index.ts
yelly: Can't put the content of both files together because it introduces a dependency cycle, but renamed the files.
I originally did this to align with the Jira implementation, but I agree that `index.ts` files with logic don't smell great.

tamtamirr (approved):
LGTM!

PR #5758, opened 2024-04-11
SALTO-5293 hide types by default, avoid requiring client config

netama:
minor improvements to the user defaults to clean up the resulting env and config

---
_Release Notes_: 
None

---
_User Notifications_: 
None


shoshana-houri (approved):
🔥

PR #5760, opened 2024-04-11
SALTO-5788 do not rely on types being loaded

netama:
Fix unintentional fallback to `additionalProperties = false` on loadWorkspace

---

will also check if there are other similar assumptions before merging, and try to add a UT

---
_Release Notes_: 
_Okta Adapter_:
* Fix issue that would cause false pending changes to remain after deploy

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/e2e_test/adapter.test.ts
+++ b/packages/jira-adapter/e2e_test/adapter.test.ts
@@ -26,7 +26,7 @@ import {
   ReadOnlyElementsSource,
   toChange,
 } from '@salto-io/adapter-api'
-import { elements as elementUtils, resolveValues } from '@salto-io/adapter-components'
+import { elements as elementUtils, resolveValues, client as clientUtils } from '@salto-io/adapter-components'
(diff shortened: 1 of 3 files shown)


ori-moisis (approved):

packages/jira-adapter/e2e_test/adapter.test.ts
```diff
@@ -253,15 +253,25 @@ each([
...
-        removalChanges.map(change =>
-          adapter.deploy({
-            changeGroup: {
-              groupID: getChangeData(change).elemID.getFullName(),
-              changes: [change],
-            },
-            progressReporter: nullProgressReporter,
-          }),
-        ),
+        removalChanges.map(change => {
```
netama, line 256: temporary workaround, there should be a follow up with the real fix as regardless this error should be caught (and we shouldn't fail on 404 in deletions)

ori-moisis:
bypassing tests as the failure is unrelated

PR #5762, opened 2024-04-14
SALTO-5787 google ws fix elemId for orgUnit

Uri-Bar:
SALTO-5787 google ws fix elemId for orgUnit

---

---
_Release Notes_: 
none

---
_User Notifications_: 
none


packages/google-workspace-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -251,6 +251,9 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
                   path: '/admin/directory/v1/customer/my_customer/orgunits{orgUnitPath}',
                   method: 'delete',
                 },
+                transformation: {
+                  omit: ['orgUnitPath'],
```
netama, line 255: maybe add an explanation about why we need to omit (and why it's ok)?

netama (approved):

PR #5763, opened 2024-04-14
SALTO-5728 Google WS omit domainAliases deploy when creating domain

Uri-Bar:
SALTO-5728 Google WS omit domainAliases deploy when creating domain
---

---
_Release Notes_: 
none

---
_User Notifications_: 
none

### DIFF
--- a/packages/google-workspace-adapter/src/definitions/deploy/deploy.ts
+++ b/packages/google-workspace-adapter/src/definitions/deploy/deploy.ts
@@ -90,11 +90,12 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
                 },
                 transformation: {
                   // add CV to inform the user that this fields are read-only
-                  omit: ['verified', 'isPrimary'],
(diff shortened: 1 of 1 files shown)


packages/google-workspace-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -90,11 +90,12 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
                 },
                 transformation: {
                   // add CV to inform the user that this fields are read-only
```
netama, line 92: nit: this should also be a TODO, right?

netama (approved):

packages/google-workspace-adapter/src/definitions/deploy/deploy.ts
Uri-Bar: I didnt open a ticket op added todo to any of the CVs I will do it in the comming days

PR #5766, opened 2024-04-14
SALTO-5789: remove workflowV2 warnings when workflow version is workflowV2

shayc331:
_remove workflowV2 warnings when workflow version is workflowV2_

---

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/jira-adapter/src/filters/workflow/transition_structure.ts
+++ b/packages/jira-adapter/src/filters/workflow/transition_structure.ts
@@ -17,12 +17,26 @@
 import { invertNaclCase, naclCase } from '@salto-io/adapter-utils'
 import { SaltoError, Value } from '@salto-io/adapter-api'
 import { collections } from '@salto-io/lowerdash'
+import Joi from 'joi'
 import _ from 'lodash'
 import { Status, Transition as WorkflowTransitionV1, WorkflowV1Instance } from './types'
 import { SCRIPT_RUNNER_POST_FUNCTION_TYPE } from '../script_runner/workflow/workflow_cloud'
-import { isWorkflowV2Transition, WorkflowTransitionV2, WORKFLOW_V1 } from '../workflowV2/types'
+import { WorkflowStatusAndPort, WorkflowTransitionV2 } from '../workflowV2/types'
 
 const { makeArray } = collections.array
+
+const TRANSITION_FROM_V2_SCHEME = Joi.array().items(
+  Joi.object({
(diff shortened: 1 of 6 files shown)


IdoZakk (commented):
🦞 
I think it is a patchy solution. I know you have spent way too much time on those workflows, but let's try to make this right.

Currently you use `isWorkflowV2Transition` for two things- decide if it is new or old workflow and verify the structure of the transition, but only for V2....
 I can think of several other solutions. 
 -If we want to verify the structure we can do it when we verify the general workflow structure
 - We can check V2 structure, if it fails check V1, and if both fail issue a warning
 - We can split to two functions. One to check the version (which you take from outside), the second one to check structure in case it is V2

packages/jira-adapter/src/filters/workflow/transition_structure.ts
```diff
@@ -17,13 +17,27 @@
...
+import { WorkflowStatusAndPort, WorkflowTransitionV2 } from '../workflowV2/types'
 
 const { makeArray } = collections.array
 
+const TRANSITION_FROM_V2_SCHEME = Joi.array().items(
+  Joi.object({
+    statusReference: Joi.alternatives(Joi.object(), Joi.string()).required(),
+    port: Joi.number(),
+  }),
+)
```
IdoZakk, line 33: Isn't the array required? If so better to add it here. Also making sure the object cannot contain more fields

packages/jira-adapter/src/filters/workflow/transition_structure.ts
```diff
@@ -17,13 +17,27 @@
...
+
+// we already validate the workflow structure in workflow_filter, so we just want to differ between the two versions
+const isTransitionFromV2 = (value: unknown): value is WorkflowStatusAndPort[] => {
+  const { error } = TRANSITION_FROM_V2_SCHEME.validate(value)
+  return error === undefined
+}
+
```
IdoZakk, line 40: Check to use schemeGuard without the error message

IdoZakk (approved):
Great, look at the comments, and maybe add a test to make sure we cover both cases (V1, V2) properly

packages/jira-adapter/src/filters/workflow/transition_structure.ts
shayc331: 1. the array is not required, for example global transition does not have `from`
2. yes

PR #5767, opened 2024-04-14
SALTO-5793: Add TypeValueMulti References in ObjectTypeAttribute

idoamit8:
Added support for reference expression in TypeValueMulti field 

---

| Status | ObjectSchemaStatus | Group |
|--------|--------------------|-------|
| Screenshot 2024-04-14 at 12 27 08 | Screenshot 2024-04-14 at 12 27 59 | Screenshot 2024-04-14 at 12 28 26 |

---
_Release Notes_: 
_Jira Adapter:_
* Fixed bug SALTO-5793 by added references from objectTypeAttribute to statuses, objectSchemaStatuses and groups. 

---
_User Notifications_: 
_Jira Adapter:_
* In the nacl `objectTypeAttribute` the field `TypeValueMulti` changed from string to referenceExpression.


IdoZakk (approved):
Great!
Can you please
1. Add a missing strategy in the last one (unless there are cases with a value that we do not want to reference
2. Add those cases to our E2E (I would also say to our monitor but we do not have premium there. I think Lior mentioned that he might help with it, not sure....)

PR #5770, opened 2024-04-14
SALTO-5801 invert Nacl case in the new infra deploy

Uri-Bar:
SALTO-5801 invert Nacl case in the new infra deploy
---

---
_Release Notes_: 
none

---
_User Notifications_: 
none


netama (approved):
thanks

PR #5776, opened 2024-04-15
SALTO-5807: Add the option to add an object to s3 with a tag

idoamit8:
Added the option to add 'tagging` when adding an object to S3. 

---

* We want to save changes in s3 for operation-changes. We decided that we will insert each operation change with a lifecycle rule of 30 days. In order to do so, I added this rule. 
* tag can be undefined.

---
_Release Notes_: 
_Core:_
* Added the support to add tagging to s3 objects when using s3_dir_store. 

---
_User Notifications_: 


omrilit (commented):
This PR can be closed right @idoamit8 ?

PR #5780, opened 2024-04-15
SALTO-5639 allow removing restriction field in views

benisraeladi:
When removing the restriction field in views, it must be set to null for the service to understand that it was on purpose.

---

When changing restrictions on views to "all agents", Zendesk sets the field to null. We should exhibit the same behavior as the service.

---
_Release Notes_: _None_

---
_User Notifications_: _None_


packages/zendesk-adapter/src/filters/view.ts
edenhassid, line 1: can you add a test for this?

edenhassid (commented):
great! 🏆 💯

edenhassid (approved):
nice! 😄

PR #5783, opened 2024-04-16
SALTO-5800 handle homepage upon space addition (WIP)

omrifarkash:
works but probably bad idea, @netama I'd like to discuss about a different solution.

---

---
_Release Notes_: 

---
_User Notifications_: 


packages/confluence-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -217,5 +267,11 @@ export const createDeployDefinitions = (): definitions.deploy.DeployApiDefinitio
     },
     customizations: createCustomizations(),
   },
-  dependencies: [],
+  dependencies: [
+    {
+      // TODO check space add + homepage move from another space
+      second: { type: SPACE_TYPE_NAME, action: 'add' },
+      first: { type: PAGE_TYPE_NAME },
```
netama, line 274: nit:
```suggestion
      first: { type: PAGE_TYPE_NAME },
      second: { type: SPACE_TYPE_NAME, action: 'add' },
```

packages/confluence-adapter/src/definitions/references.ts
```diff
@@ -38,6 +38,17 @@ const REFERENCE_RULES: referenceUtils.FieldReferenceDefinition<never>[] = [
     serializationStrategy: 'id',
     target: { type: 'label' },
   },
+  // This rule should be defined before the same rule with serializationStrategy = 'id'
+  // {
```
netama, line 42: is this needed?

packages/confluence-adapter/src/definitions/transformation_utils/space.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import { definitions } from '@salto-io/adapter-components'
+import _ from 'lodash'
+import { values as lowerdashValues } from '@salto-io/lowerdash'
+import { assertValue } from './generic'
+
+export const adjustHomepageToIdFunc: definitions.AdjustFunction = item => {
+  const value = assertValue(item.value)
+  const homepage = _.get(value, 'homepage')
+  if (!lowerdashValues.isPlainRecord(homepage) || !('id' in homepage)) {
```
netama, line 25: nit: please use 
```suggestion
  if (!lowerdashValues.isPlainRecord(homepage) || !Object.prototype.hasOwnProperty.call(homepage, 'id')) {
```
or something else that is guaranteed to look at the homepage and not some prototype (should probably be ok in this case, but we try not to have these in the code because they sometimes have unexpected results e.g. on `toString`)

packages/confluence-adapter/src/definitions/transformation_utils/generic.ts
```diff
@@ -14,9 +14,10 @@
  * limitations under the License.
  */
 
+import { Value } from '@salto-io/adapter-api'
 import { values as lowerdashValues } from '@salto-io/lowerdash'
 
-export const assertValue = (value: unknown): Record<string, unknown> => {
+export const assertValue = (value: unknown): Record<string, Value> => {
```
netama, line 20: 1. nit: maybe rename to `validateValue` to avoid using `assert` which is usually a test-related utility?
2. how do you know it's a `Value`? `isPlainRecord` only guarantees `unknown`

packages/confluence-adapter/src/definitions/transformation_utils/page.ts
```diff
@@ -15,27 +15,44 @@
...
 
 /**
  * AdjustFunction that increases the version number of a page for deploy modification change.
  */
-export const increasePagesVersion: definitions.AdjustFunction<definitions.deploy.ChangeAndContext> = item => {
-  const value = assertValue(item.value)
-  const version = assertValue(value.version)
+export const increasePagesVersion = (value: Record<string, Value>): void => {
+  const { version } = value
+  if (!lowerdashValues.isPlainRecord(version)) {
```
netama, line 29: do you want to override even if it's not undefined? we should probably at least add a log if we're overriding something in a different format, no?

packages/confluence-adapter/src/definitions/transformation_utils/page.ts
```diff
@@ -15,27 +15,44 @@
...
+    value.version = { number: 2 }
+    return
+  }
   const { number } = version
   if (!isNumber(number)) {
-    return { value }
+    return
```
netama, line 35: if this is unexpected can we log?

packages/confluence-adapter/src/definitions/transformation_utils/page.ts
```diff
@@ -15,27 +15,44 @@
...
+  if (changeFromChangeGroup !== undefined && isInstanceChange(changeFromChangeGroup)) {
+    const newId = getChangeData(changeFromChangeGroup).value.id
+    if (newId !== undefined) {
+      changeData.value.id = newId
+      value.id = newId
+    }
   }
 }
+
+export const adjustPageModificationFunction: definitions.AdjustFunction<definitions.deploy.ChangeAndContext> = item => {
```
netama, line 53: please add a docstring explaining what this does (and why)

netama (commented):
looks good, added some comments (and I assume there will be some more tests) but let's discuss IRL if you want to try something else

packages/confluence-adapter/src/definitions/utils/page.ts
```diff
@@ -0,0 +1,103 @@
...
+} from '@salto-io/adapter-api'
+import { definitions } from '@salto-io/adapter-components'
+import { values } from '@salto-io/lowerdash'
+import _ from 'lodash'
+import { SPACE_TYPE_NAME } from '../../constants'
+import { AdditionalAction } from '../types'
+import { validateValue } from './generic'
+
+/**
+ * If page is a homepage of a space and both are being deployed as addition to some environment,
```
netama, line 32: can there be a case where an existing page becomes a homepage (either of a new space or an existing one)?

packages/confluence-adapter/src/definitions/utils/page.ts
```diff
@@ -0,0 +1,103 @@
...
+ * We first deploy the space and then we modify its homepage (default homepage is created in the service when creating new space).
+ * This function will switch the action from 'add' to 'modify' in case of homepage addition.
+ */
+export const homepageAdditionToModification: ({
+  change,
+  changeGroup,
+  elementSource,
+}: definitions.deploy.ChangeAndContext) => (ActionName | AdditionalAction)[] = ({ change, changeGroup }) => {
+  const spaceChange = changeGroup.changes.find(c => getChangeData(c).elemID.typeName === SPACE_TYPE_NAME)
```
netama, line 41: just making sure - do we know a space cannot be the homepage of two spaces?

packages/confluence-adapter/src/definitions/utils/page.ts
```diff
@@ -0,0 +1,103 @@
...
+  if (isAdditionChange(change) && spaceChange !== undefined && isInstanceChange(spaceChange)) {
+    const changeData = getChangeData(change)
+    const spaceChangeData = getChangeData(spaceChange)
+    const spaceRef = _.get(changeData.value, 'spaceId')
+    if (isReferenceExpression(spaceRef) && spaceRef.elemID.isEqual(spaceChangeData.elemID)) {
+      return ['modify']
```
netama, line 47: maybe add a debug log when this happens?

packages/confluence-adapter/src/definitions/utils/page.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+  return [change.action]
+}
+
+const isNumber = (value: unknown): value is number => typeof value === 'number'
+
+/**
+ * AdjustFunction that increases the version number of a page for deploy modification change.
+ */
+const increasePagesVersion: definitions.AdjustFunction<definitions.deploy.ChangeAndContext> = args => {
```
netama, line 58: tiny nit (though existing code) - why `Pages` and not `Page`?

packages/confluence-adapter/src/definitions/utils/page.ts
```diff
@@ -0,0 +1,103 @@
...
+  return { ...args, value }
+}
+
+/**
+ * AdjustFunction that runs all page modification adjust functions.
+ */
+export const adjustPageOnModification: definitions.AdjustFunction<definitions.deploy.ChangeAndContext> = args => {
+  const value = validateValue(args.value)
+  const argsWithValidatedValue = { ...args, value }
+  return [increasePagesVersion, updateHomepageId].reduce(
```
netama, line 99: style/readability nit - usually if `reduce` is not required it's better to avoid it, in this case maybe we can simplify a bit by making the inner functions actually be inner functions (so - avoid the triple run of `validateValue`, and just run them in sequence)? unless you think there will be something else here that would benefit from the generalization?

packages/confluence-adapter/src/definitions/utils/page.ts
```diff
@@ -0,0 +1,103 @@
...
+ * AdjustFunction that increases the version number of a page for deploy modification change.
+ */
+const increasePagesVersion: definitions.AdjustFunction<definitions.deploy.ChangeAndContext> = args => {
+  const value = validateValue(args.value)
+  const version = _.get(value, 'version')
+  if (!values.isPlainRecord(version) || !isNumber(version.number)) {
+    return { ...args, value }
+  }
+  return {
+    ...args,
```
netama, line 65: nit: pretty sure this is not needed (and same below) - if you only return the `value` the rest will be added by the infra

packages/confluence-adapter/src/definitions/utils/space.ts
```diff
@@ -85,12 +85,37 @@ export const restructurePermissionsAndCreateInternalIdMap = (value: Record<strin
...
+ * Group space with its homepage upon addition.
+ * We want to first deploy the space, a default homepage will be created in the service. We want to modify it
+ */
+export const spaceChangeGroupWithItsHomepage: deployment.grouping.ChangeIdFunction = async change => {
+  const changeData = getChangeData(change)
+  if (isInstanceElement(changeData)) {
+    const homepageFullName = changeData.value.homepage?.elemID?.getFullName()
+    // in case of addition, we want the space to be in the same group as its homepage
+    if (isAdditionChange(change) && _.isString(homepageFullName)) {
+      return homepageFullName
```
netama, line 117: note that the homepage could also remain a number if for some reason we didn't convert the homepage to a reference (e.g. if we include space but exclude page in the fetch config). it sounds like in this case it's "ok" since there is no point grouping with something that doesn't exist - but please be aware of that when checking for references and adding fallbacks. in this case, might be worth IMO to split this into two checks - one checks if the homepage is `!== undefined`, and then check if it's not a reference log a warning. (and then I think you won't need to check for `isString`)

netama (approved):
looks good for the main flow so approving and I think we should get this in (with a few minor comments). 
I think we should follow up with checks for a bunch of edge cases around non-standard combinations (e.g. making an existing page the homepage of a new/existing space, making an existing homepage the homepage of another space, checking if a page can somehow be the homepage of two spaces, changing just the homepage of an existing space, etc)

packages/confluence-adapter/src/definitions/utils/page.ts
omrifarkash: 1. Page can only exists in a single space.
2. To make existing page a homepage of a space, I think we need to make two modifications, one for the page and second for the homepage, but the internalIds will remain the same.

packages/confluence-adapter/src/definitions/utils/page.ts
omrifarkash: You meant "a `page`* cannot be the homepage of two spaces?" ?
If so, then yes. Page has a single space pointer

packages/confluence-adapter/src/definitions/utils/page.ts
omrifarkash: typo

packages/confluence-adapter/src/definitions/utils/page.ts
omrifarkash: I think there might be more adjust function in the future

packages/confluence-adapter/src/definitions/utils/page.ts
netama: :+1: thanks. that's a valod scenario though, right? a deploy that has (maybe among other things) a change of the page's parent to another space, and a change of the other space's homepage to the moved page?

PR #5784, opened 2024-04-16
SALTO-5807: Added S3 dir store Deletion

idoamit8:
In this Pr I added support for deletion with S3 dir store.

---

---
_Release Notes_: 
_Core Adapter:_
* Added support for deletion in S3 dir store.

---
_User Notifications_: 

### DIFF
--- a/packages/core/src/local-workspace/s3_dir_store.ts
+++ b/packages/core/src/local-workspace/s3_dir_store.ts
@@ -23,10 +23,12 @@ import Bottleneck from 'bottleneck'
 import getStream from 'get-stream'
 import { Readable } from 'stream'
 import { values } from '@salto-io/lowerdash'
+import _ from 'lodash'
 
 const log = logger(module)
 
 const DEFAULT_CONCURRENCY_LIMIT = 100
+const DELETION_BATCH_SIZE = 1000 // AWS limit
 
 export const buildS3DirectoryStore = ({
   bucketName,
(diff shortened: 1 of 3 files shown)


packages/core/src/local-workspace/s3_dir_store.ts
```diff
@@ -125,10 +126,44 @@ export const buildS3DirectoryStore = ({
...
+  }
   const flush = async (): Promise<void> => {
     const files = Object.values(updated)
     updated = {}
     await Promise.all(files.map(f => writeFile(f)))
+
+    if (deleted.size > 0) {
+      const toDelete = Array.from(deleted)
+      await deleteMany(toDelete)
+      deleted = new Set()
```
omrilit, line 165: Same as for `updated`, you wouldn't like to report that you treated "delete" requests that happened during the `flush` although they weren't.
```suggestion
      deleted = new Set()
      await deleteMany(toDelete)

```

packages/core/src/local-workspace/s3_dir_store.ts
```diff
@@ -125,10 +126,44 @@ export const buildS3DirectoryStore = ({
...
     }, `listing s3 objects for ${baseDir}`)
 
+  const deleteMany = async (objectPaths: string[]): Promise<void> => {
+    const objectIdentifiers: AWS.ObjectIdentifier[] = objectPaths.map(objectPath => ({ Key: getFullPath(objectPath) }))
+
+    // Split the deletion into batches of 1000 items each (AWS limit)
+    const batchSize = 1000
+    const batchPromises = []
+
+    for (let i = 0; i < objectIdentifiers.length; i += batchSize) {
```
omrilit, line 136: you can use `_.chunk()`

omrilit (commented):
Looks good!

omrilit (approved):
LGTM!

packages/core/src/local-workspace/s3_dir_store.ts
```diff
@@ -125,10 +128,39 @@ export const buildS3DirectoryStore = ({
       return Array.from(paths)
     }, `listing s3 objects for ${baseDir}`)
 
+  const deleteMany = async (objectPaths: string[]): Promise<void> => {
+    const objectIdentifiers: AWS.ObjectIdentifier[] = objectPaths.map(objectPath => ({ Key: getFullPath(objectPath) }))
+
+    const batches = _.chunk(objectIdentifiers, DELETION_BATCH_SIZE)
+    const batchPromises = batches.map(batch =>
+      bottleneck.schedule(async () => {
+        log.trace('Deleting batch of objects from S3 bucket %s', bucketName)
```
ori-moisis, line 137: instead of logging before and after, use `log.timeTrace`

packages/core/src/local-workspace/s3_dir_store.ts
```diff
@@ -139,5 +171,8 @@ export const buildS3DirectoryStore = ({
     list,
     getFullPath: filePath => `s3://${bucketName}/${getFullPath(filePath)}`,
     flush,
+    delete: async fileName => {
+      deleted.add(fileName)
```
ori-moisis, line 175: this is not sufficient to just add something to `deleted`
you can look at the implementation of the "regular" dir store, but, just an example of a scenario that won't work:
```typescript
await dirStore.set(someFile)
await dirStore.delete(someFile.filename)
const f = await dirStore.get(someFile.filename) // Expected - undefined, actual - it behaves as though the file exists
```
also the other way around would be wrong, e.g
```typescript
await dirStore.delete(someFile.filename)
await dirStore.set(someFile)
await dirStore.flush()
// Expected - "someFile" will exist, Actual - it will be written and deleted, depending on the order and timing, it may or may not be there after the flush
```

PR #5787, opened 2024-04-16
SALTO-5749 custom request context

netama:
Allow using a custom function for defining context args for request url/body, both in fetch and in deploy.

---

The immediate use case is for google workspace, where we need the "before" name in a request url.

---
_Release Notes_: 
None

---
_User Notifications_: 
None


Uri-Bar (approved):
looking good, also tested it locally

PR #5788, opened 2024-04-17
SALTO-5630: SFDC removing missing references from profiles in fix elements

yelly:
Adds a new elements fixer to SFDC adapter which removes broken references from profiles.
Also refactored `combineElementFixers` to support disabling them with config.

---

---
_Release Notes_: 
_Salesforce_:
* Adds a new elements fixer that removes broken references from profiles. Disabled by default.

---
_User Notifications_: 
None.


packages/salesforce-adapter/src/weak_references/profiles.ts
```diff
@@ -260,21 +257,48 @@ const sectionsReferenceParams: Record<section, ReferenceFromSectionParams> = {
...
+    sourceField?: string,
+  ) => T,
+): T[] =>
   Object.entries(sectionsReferenceParams).flatMap(([sectionName, params]) =>
-    referencesFromSection(profile, sectionName as section, params),
+    mapSectionEntries(
+      profile,
+      sectionName as section,
+      params,
+      _.curry(f)(sectionName),
```
tomsellek, line 274: why `_.curry` and not `bind`?

packages/salesforce-adapter/src/weak_references/profiles.ts
```diff
@@ -284,7 +308,58 @@ const getProfilesCustomReferences: WeakReferencesHandler['findWeakReferences'] =
     return refs
   }
 
+const profileEntriesTargets = (profile: InstanceElement): Dictionary<ElemID> =>
+  _(
+    mapProfileSections(
+      profile,
+      (sectionName, sectionEntryKey, target, sourceField): [string, ElemID] => [
+        [sectionName, sectionEntryKey, ...makeArray(sourceField)].join('.'),
```
tomsellek, line 316: Am I missing something or are you now no longer calling `createNestedID` on the source field?

tomsellek (commented):
Left a couple of comments. Overall I must say this change feels like it's trying too hard to avoid code duplication, and ends up with code that is more complex than necessary. It may be because I have an aversion to this functional style of super generic functions that act on functions that act on functions.

packages/salesforce-adapter/src/weak_references/handlers.ts
```diff
@@ -42,7 +43,7 @@ const defaultHandlersConfiguration: Record<Handlers, boolean> = {
   profiles: false,
 }
 
-const handlersConfiguration = (
+const customReferencesConfiguration = (
```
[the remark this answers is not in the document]
yelly: Done.

packages/salesforce-adapter/src/weak_references/profiles.ts
yelly: Why would I want to set `this`?

packages/salesforce-adapter/src/weak_references/profiles.ts
yelly: We are, on line 282.

packages/salesforce-adapter/src/types.ts
```diff
@@ -656,6 +672,9 @@ const dataManagementType = new ObjectType({
     [CUSTOM_REFS_CONFIG]: {
       refType: customReferencesSettingsType,
     },
+    [FIX_ELEMENTS_CONFIG]: {
+      refType: fixElementsSettingsType,
+    },
```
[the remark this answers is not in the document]
yelly: I can imagine users wanting to see the references but not to fix them automatically for example.

yelly:
Re your main comment - I think having duplication for the logic which creates references from the profile would be pretty painful and hard to maintain and I couldn't really find a more readable way to avoid it. I think the code being somewhat complicated is a result of profiles being pretty complex inherently.

packages/adapter-components/src/references/element_fixers.ts
```diff
@@ -29,12 +30,17 @@ const getFixedElements = (elements: Element[], fixedElements: Element[]): Elemen
...
- * Combine several fixElements functions into one that will run all of them.
+ * Combine several fixElements functions into one that will run the enabled ones.
  */
 export const combineElementFixers =
-  (fixers: FixElementsFunc[]): FixElementsFunc =>
-  async elements =>
-    awu(fixers).reduce(
+  (
+    namedFixers: Record<string, FixElementsFunc>,
+    getFixersConfig: (adapterConfig: InstanceElement) => Record<string, boolean> = () => ({}),
```
tamtamirr, line 38: Nit: I'm not a fan of this approach.
I would make the `getFixersConfig` possibly undefined, and use empty record when it's undefined.
This is semantically different than having empty config value.

packages/adapter-components/src/references/element_fixers.ts
```diff
@@ -29,12 +30,17 @@ const getFixedElements = (elements: Element[], fixedElements: Element[]): Elemen
...
+  ): FixElementsFunc =>
+  async (elements, adapterConfig) => {
+    const adapterConfigValues = adapterConfig ? getFixersConfig(adapterConfig) : {}
+    const fixers = Object.values(getEnabledEntries(namedFixers, adapterConfigValues))
```
tamtamirr, line 42: Can we add doc/comment on the `getEnabledEntries` that the default when entry is undefined is `Enabled`?

packages/salesforce-adapter/src/custom_references/profiles.ts
```diff
@@ -284,7 +308,58 @@ const getProfilesCustomReferences: WeakReferencesHandler['findWeakReferences'] =
...
+      isInstanceOfTypeSync(PROFILE_METADATA_TYPE),
+    )
+    const entriesTargets: Dictionary<ElemID> = _.merge(
+      {},
+      ...profiles.map(profileEntriesTargets),
+    )
+    const brokenReferenceFields = Object.keys(
+      await pickAsync(
+        entriesTargets,
+        async (target) => !(await elementsSource.has(target)),
```
tamtamirr, line 337: To avoid having many calls to the `Rocks DB`, I suggest listing all the **ElemIDs** once by invoking `elementsSource.list()` and hold a set of all the Element IDs full names.

packages/salesforce-adapter/src/custom_references/profiles.ts
```diff
@@ -284,7 +308,58 @@ const getProfilesCustomReferences: WeakReferencesHandler['findWeakReferences'] =
...
+      ),
+    )
+    const brokenElements = profiles.filter((profile) =>
```
tamtamirr, line 340:
```suggestion
    const profilesWithBrokenReferences = profiles.filter((profile) =>
```

packages/salesforce-adapter/src/custom_references/profiles.ts
```diff
@@ -284,7 +308,58 @@ const getProfilesCustomReferences: WeakReferencesHandler['findWeakReferences'] =
...
+      brokenReferenceFields.some((field) => _(profile.value).has(field)),
+    )
+    const fixedElements = brokenElements.map((profile) => {
+      const fixed = profile.clone()
+      fixed.value = _.omit(fixed.value, ...brokenReferenceFields)
```
tamtamirr, line 345:
```suggestion
      fixed.value = _.omit(fixed.value, brokenReferenceFields)
```
It's better to avoid spread operator on possibly large arrays, as spread happens in the stack and can lead to overflow. (OOM exception)

packages/salesforce-adapter/src/custom_references/profiles.ts
```diff
@@ -284,7 +308,58 @@ const getProfilesCustomReferences: WeakReferencesHandler['findWeakReferences'] =
...
+    const errors = brokenElements.flatMap((profile) =>
+      brokenReferenceFields
+        .filter((field) => _(profile.value).has(field))
+        .map((field) => ({
+          elemID: profile.elemID.createNestedID(...field.split('.')),
+          severity: 'Info' as const,
+          message: 'Dropping profile fields which reference missing types',
+          detailedMessage: `The field references the type ${entriesTargets[field].getFullName()} which is not available in the workspace.`,
+        })),
+    )
```
tamtamirr, line 357: Creating ChangeError on nested ElemIDs is not well supported.
Also, I think this code can possibly yield too many Errors.
We should create Errors only on top level Elements, in this case it's the Profile instances.

packages/salesforce-adapter/test/custom_references/handlers.test.ts
```diff
@@ -0,0 +1,68 @@
...
+
+import { InstanceElement, ElemID, ObjectType } from '@salto-io/adapter-api'
+import { buildElementsSourceFromElements } from '@salto-io/adapter-utils'
+import { adapter } from '../../src/adapter_creator'
+import {
+  fixElementsFunc,
+  getCustomReferences,
+} from '../../src/custom_references/handlers'
+
+describe('custom references handlers', () => {
```
tamtamirr, line 25: Not sure what these tests are good for..? What are we covering here?

packages/salesforce-adapter/src/custom_references/profiles.ts
```diff
@@ -284,7 +308,58 @@ const getProfilesCustomReferences: WeakReferencesHandler['findWeakReferences'] =
...
+      (sectionName, sectionEntryKey, target, sourceField): [string, ElemID] => [
+        [sectionName, sectionEntryKey, ...makeArray(sourceField)].join('.'),
+        target,
+      ],
+    ),
+  )
+    .fromPairs()
+    .value()
+
+const removeMissingReferences: WeakReferencesHandler['removeWeakReferences'] =
```
tamtamirr, line 324: Aren't we lacking UTs here?

tamtamirr (commented):
Good work! Added some comments

packages/adapter-components/src/references/element_fixers.ts
```diff
@@ -29,12 +30,17 @@ const getFixedElements = (elements: Element[], fixedElements: Element[]): Elemen
...
 /**
- * Combine several fixElements functions into one that will run all of them.
+ * Combine several fixElements functions into one that will run the enabled ones.
  */
 export const combineElementFixers =
-  (fixers: FixElementsFunc[]): FixElementsFunc =>
-  async elements =>
-    awu(fixers).reduce(
+  (
+    namedFixers: Record<string, FixElementsFunc>,
```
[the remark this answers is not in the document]
yelly: Done.

packages/adapter-components/src/references/element_fixers.ts
yelly: Done.

packages/salesforce-adapter/src/custom_references/profiles.ts
yelly: The file is nearly 100% covered by `profiles.test.ts` (the only uncovered lines are a couple of edge error cases).
Do you have specific flows in mind you think we should add?

packages/salesforce-adapter/src/custom_references/profiles.ts
yelly: Done.

packages/adapter-api/src/adapter.ts
```diff
@@ -135,7 +135,10 @@ export type DeployModifiers = {
 
 export type ValidationModifiers = Pick<DeployModifiers, 'changeValidator'>
 
-export type FixElementsFunc = (elements: Element[]) => Promise<{
+export type FixElementsFunc = (
+  elements: Element[],
+  adapterConfig?: InstanceElement,
```
ori-moisis, line 140: why are you changing the interface?
the adapter config is already accessible to element fixers because they are created in the context of "adapter operations", meaning, the function that creates the element fixer gets the config as a parameter ([this](https://github.com/salto-io/salto/blob/main/packages/adapter-api/src/adapter.ts/#L213)).
we should not have multiple options for passing the configuration - that could lead to conflicting definitions and bugs

ori-moisis (commented):
Did not read most of the PR, just commenting on the API change

packages/salesforce-adapter/src/custom_references/profiles.ts
yelly: Done.

yelly: Done.

packages/salesforce-adapter/src/custom_references/profiles.ts
yelly: Done. Didn't realise I can just pass the array as-is.

packages/salesforce-adapter/test/custom_references/handlers.test.ts
yelly: This exactly covers `handlers.ts` and even caught some sneaky mistakes I made.
I could add some profiles that will get references/fixed, but that will just duplicate tests from `profiles.test.ts`.

packages/adapter-api/src/adapter.ts
yelly: Good point.
I was trying to stick to the custom references implementation as much as possible, but I guess in this case it's better to deviate.
Moved this logic into the adapter specific logic.

packages/salesforce-adapter/src/custom_references/profiles.ts
yelly: Done.

packages/salesforce-adapter/src/custom_references/profiles.ts
```diff
@@ -284,7 +311,61 @@ const getProfilesCustomReferences: WeakReferencesHandler['findWeakReferences'] =
...
+        .join(', ')}`,
+    }))
+
+    return { fixedElements, errors }
+  }
+
 export const profilesHandler: WeakReferencesHandler = {
   findWeakReferences: getProfilesCustomReferences,
-  removeWeakReferences: () => async () => ({ fixedElements: [], errors: [] }),
+  removeWeakReferences: removeMissingReferences,
```
tamtamirr, line 370: Nit: not sure why we have different names here that basically indicate the same thing.
Consider renaming your `removeWeakReferences` method to `removeWeakReferences`

tamtamirr (approved):
LGTM! Added a small nit.
Did we test this thoroughly? Is it blocked by some other work / optional features?

packages/salesforce-adapter/src/custom_references/profiles.ts
yelly: Done.

PR #5790, opened 2024-04-17
SALTO-2688: fix zendesk cleanup e2e

edenhassid:
fix zendesk cleanup e2e

---

---
_Release Notes_: 

---
_User Notifications_: 


packages/zendesk-adapter/e2e_test/adapter.test.ts
```diff
@@ -102,6 +112,25 @@ const ALL_SUPPORTED_TYPES = {
 
 const HELP_CENTER_BRAND_NAME = 'e2eHelpCenter'
 
+const TYPES_TO_REMOVE = new Set<string>([
```
adfineberg, line 115: Another thought - not sure if it makes sense. Maybe instead of a list of what to remove, we keep a list of types that are automatically removed? This feels a bit more future-proof... WDYT?

adfineberg (approved):
Yay!

PR #5791, opened 2024-04-17
SALTO-5703 convert create-check-deployment-based-on-config-validator to work with definitions

Uri-Bar:
SALTO-5703 convert create-check-deployment-based-on-config-validator to work with definitions
---

Also added the option to pass additional CV to the new createAdapter func and pass the new definitions CV to google WS adapter

---
_Release Notes_: 
none

---
_User Notifications_: 
none

### DIFF
--- a/packages/adapter-components/src/adapter-wrapper/adapter/adapter.ts
+++ b/packages/adapter-components/src/adapter-wrapper/adapter/adapter.ts
@@ -56,6 +56,7 @@ import { ElementQuery, createElementQuery } from '../../fetch/query'
 import {
   createChangeValidator,
   deployNotSupportedValidator,
+  createCheckDeploymentBasedOnDefinitionsValidator,
   getDefaultChangeValidators,
 } from '../../deployment/change_validators'
 import { generateOpenApiTypes } from '../../openapi/type_elements/type_elements'
@@ -145,6 +146,13 @@ export class AdapterImpl<
     this.changeValidators = {
       ...getDefaultChangeValidators(),
       ...(this.definitions.deploy?.instances === undefined ? { deployNotSupported: deployNotSupportedValidator } : {}),
+      ...(this.definitions.deploy !== undefined
+        ? {
+            createCheckDeploymentBasedOnDefinitions: createCheckDeploymentBasedOnDefinitionsValidator({
+              deployDefinitions: this.definitions.deploy,
+            }),
+          }
+        : {}),
       ...additionalChangeValidators,
     }
     // TODO combine with infra changers after SALTO-5571
--- a/packages/adapter-components/src/adapter-wrapper/adapter/adapter.ts
+++ b/packages/adapter-components/src/adapter-wrapper/adapter/adapter.ts
@@ -145,14 +145,13 @@ export class AdapterImpl<
     this.configInstance = configInstance
     this.changeValidators = {
       ...getDefaultChangeValidators(),
-      ...(this.definitions.deploy?.instances === undefined ? { deployNotSupported: deployNotSupportedValidator } : {}),
-      ...(this.definitions.deploy !== undefined
-        ? {
+      ...(this.definitions.deploy?.instances === undefined
+        ? { deployNotSupported: deployNotSupportedValidator }
+        : {
             createCheckDeploymentBasedOnDefinitions: createCheckDeploymentBasedOnDefinitionsValidator({
               deployDefinitions: this.definitions.deploy,
             }),
-          }
-        : {}),
+          }),
       ...additionalChangeValidators,
     }
     // TODO combine with infra changers after SALTO-5571
--- a/packages/adapter-components/src/deployment/change_validators/check_deployment_based_on_definitions.ts
+++ b/packages/adapter-components/src/deployment/change_validators/check_deployment_based_on_definitions.ts
@@ -30,16 +30,22 @@ import { collections } from '@salto-io/lowerdash'
 import { DeployApiDefinitions, DeployableRequestDefinition } from '../../definitions/system/deploy'
 import { DefaultWithCustomizations, queryWithDefault } from '../../definitions'
 import { ERROR_MESSAGE, detailedErrorMessage } from './check_deployment_based_on_config'
+import {
+  APIDefinitionsOptions,
+  ResolveAdditionalActionType,
+  ResolveClientOptionsType,
+} from '../../definitions/system/api'
 
 const { awu } = collections.asynciterable
+const { makeArray } = collections.array
 
-const createChangeErrors = (
-  typeConfig: DefaultWithCustomizations<DeployableRequestDefinition<never>[], ActionName>,
+const createChangeErrors = <Options extends APIDefinitionsOptions = {}>(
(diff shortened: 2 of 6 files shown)


packages/google-workspace-adapter/src/adapter_creator.ts
```diff
@@ -77,6 +79,13 @@ export const adapter = createAdapter<Credentials, Options, UserConfig>({
   operationsCustomizations: {
     connectionCreatorFromConfig: () => createConnectionForApp(DIRECTORY_APP_NAME),
     credentialsFromConfig: defaultCredentialsFromConfig,
+    additionalChangeValidators: {
```
Uri-Bar, line 82: later I can create a `change_validator` file like in the other adapters and move it there

packages/adapter-components/src/deployment/change_validators/check_deployment_based_on_definitions.ts
```diff
@@ -0,0 +1,100 @@
...
+
+const { awu } = collections.asynciterable
+const { makeArray } = collections.array
+
+const createChangeErrors = <Options extends APIDefinitionsOptions = {}>(
+  typeConfig: DefaultWithCustomizations<DeployableRequestDefinition<ResolveClientOptionsType<Options>>[], ActionName>,
+  instanceElemID: ElemID,
+  action: ActionName,
+): ChangeError[] => {
+  const requestsByAction = makeArray(queryWithDefault(typeConfig).query(action))
```
Uri-Bar, line 47: this sometimes returns an array and sometimes a single object

packages/adapter-components/src/deployment/change_validators/check_deployment_based_on_definitions.ts
```diff
@@ -0,0 +1,100 @@
...
+  APIDefinitionsOptions,
+  ResolveAdditionalActionType,
+  ResolveClientOptionsType,
+} from '../../definitions/system/api'
+
+const { awu } = collections.asynciterable
+const { makeArray } = collections.array
+
+const createChangeErrors = <Options extends APIDefinitionsOptions = {}>(
+  typeConfig: DefaultWithCustomizations<DeployableRequestDefinition<ResolveClientOptionsType<Options>>[], ActionName>,
```
netama, line 43: nits: 
1. if we can use def / definition instead of config it will be clearer IMO
2. this is not the entire type definition - maybe `typeRequestDefinition` or `requestsByActionDefQuery`?

packages/adapter-components/src/deployment/change_validators/check_deployment_based_on_definitions.ts
```diff
@@ -0,0 +1,100 @@
...
+  instanceElemID: ElemID,
+  action: ActionName,
+): ChangeError[] => {
+  const requestsByAction = makeArray(queryWithDefault(typeConfig).query(action))
+  if (requestsByAction.length === 0 || requestsByAction.every(req => req.request.endpoint === undefined)) {
```
netama, line 48: probably not critical for now, but can we distinguish between undefined an empty array? ideally empty would still mean supported if possible (if we don't `makeArray` I think we _should_ know?)

packages/adapter-components/src/deployment/change_validators/check_deployment_based_on_definitions.ts
netama, line 47: another naming nit - this also seems a bit misleading, maybe `requestsForAction` / `actionRequests`? `requestsByAction` is the map before choosing an action

packages/adapter-components/src/deployment/change_validators/check_deployment_based_on_definitions.ts
```diff
@@ -0,0 +1,100 @@
...
+        message: ERROR_MESSAGE,
+        detailedMessage: detailedErrorMessage(action, instanceElemID),
+      },
+    ]
+  }
+  return []
+}
+
+export const createCheckDeploymentBasedOnDefinitionsValidator = <Options extends APIDefinitionsOptions = {}>({
+  typesConfig,
```
netama, line 62: same re `typeDefinitions` and not `config`

packages/adapter-components/src/deployment/change_validators/check_deployment_based_on_definitions.ts
```diff
@@ -0,0 +1,100 @@
...
+        severity: 'Error',
+        message: ERROR_MESSAGE,
+        detailedMessage: detailedErrorMessage(action, instanceElemID),
+      },
+    ]
+  }
+  return []
+}
+
+export const createCheckDeploymentBasedOnDefinitionsValidator = <Options extends APIDefinitionsOptions = {}>({
```
netama, line 61: let's add a comment (here or in the original) that these are the old and new versions of the same validator?

packages/google-workspace-adapter/src/adapter_creator.ts
netama: we should have this in all adapters that are using the infra (and for now - also in confluence) -
1. we can add it this way in this PR if you prefer, but if so can you please do the same here for confluence (since it also needs it now)?
2. I think you can instead add this one specifically in https://github.com/salto-io/salto/blob/main/packages/adapter-components/src/adapter-wrapper/adapter/adapter.ts#L148 next to `deployNotSupported`

netama (approved):
cool :) added some comments but pre-approving since the logic looks good 🎈

packages/adapter-components/src/deployment/change_validators/check_deployment_based_on_definitions.ts
Uri-Bar: how does empty array is supported?

packages/google-workspace-adapter/src/adapter_creator.ts
Uri-Bar: added it to the adapter file, and removed it from google

packages/adapter-components/src/adapter-wrapper/adapter/adapter.ts
```diff
@@ -145,6 +146,13 @@ export class AdapterImpl<
     this.changeValidators = {
       ...getDefaultChangeValidators(),
       ...(this.definitions.deploy?.instances === undefined ? { deployNotSupported: deployNotSupportedValidator } : {}),
+      ...(this.definitions.deploy !== undefined
+        ? {
+            createCheckDeploymentBasedOnDefinitions: createCheckDeploymentBasedOnDefinitionsValidator({
+              deployDefinitions: this.definitions.deploy,
+            }),
+          }
+        : {}),
```
netama, line 155: can't this be
```suggestion
      ...(this.definitions.deploy?.instances === undefined
      ? {
        deployNotSupported: deployNotSupportedValidator.
      } : {
        createCheckDeploymentBasedOnDefinitions: createCheckDeploymentBasedOnDefinitionsValidator({
              deployDefinitions: this.definitions.deploy,
            }),
        }),
```
(up to indentation etc)?

Uri-Bar: of course

PR #5792, opened 2024-04-17
SALTO-5814 default service id field type if missing when replacing types during deploy

netama:
we create synthetic types during deploy in order to support creating the first instance of its type / other cases in ducktype adapters that might not have all fields defined accurately during fetch.
added a fallback in case a service id field does not exist at all (which happens when the first instance of its type is created)

---
_Release Notes_: 
None

---
_User Notifications_: 
None


shir-reifenberg (approved):
💯

PR #5793, opened 2024-04-17
SALTO-5815: fix a bug in account ids for automations

IdoZakk:
Good thing Alon is not here to see that, he warned about this but I waved him off

---

I do not think there will be noise as we would have crashed like we already saw in the bug

---
_Release Notes_: 
Jira Adapter:
* Fixed a bug in automation that could cause fetch to fail

---
_User Notifications_: 
None


shayc331 (approved):
🏆

PR #5795, opened 2024-04-17
SALTO-5737 space deployment filter to handle permission changes

omrifarkash:
Note, on this PR I also make template to nest under space instance and add deploy definitions to global templates. These two changes are not the main changes on this PR, just leftovers from other tasks

---

For future reader:
explanation of the two main changes on this PR

1. upon fetch, we restructure permissions object and save internalIds in a hidden map on the space instance. We need these ids the handle deployment of these permissions. For each permission "instance" we calculate a unique key (given its 4 fields), and save its internal id as the value of this key

2. deploy_space_and_permission is a complex filter that handle deployment of space instance with its permissions.

- If user modify space instance, we first modify the space instance without its permissions (first api call), then we calculate the diff in the permissions between the before instance and the after one. When the diff is calculated, we first delete the unwanted permissions by creating a fake permission instance removal change (we use the internalIds we save upon fetch, see first bullet). Then we do similar thing with the new added permissions, and fake addition changes. After we deploy the new added permissions, we update the hidden mapIds to include the new permissions ids.

- If user add a new space instance we first create the instance (first api call). When creating a space in the service, default permissions are created. To handle it, in this point we fetch the new space default permissions, creating a fake space instance with these permissions, then we create a fake modification change between the fake space instance (as the "before") and the real added space instance (as the "after"). Now with this fake modification change we can operate exactly as we did in the above case (space modification case).

---

_Release Notes_: 

_Confluence Adapter_:
- Change structure of permissions objects array (on space instance)
- Support deployment of space permissions

---

_User Notifications_: 
_Confluence Adapter_:
- Change structure of permissions objects array (on space instance)
- Support deployment of space permissions


packages/confluence-adapter/src/filters/deploy_space_permissions.ts
```diff
@@ -0,0 +1,295 @@
...
+      { beforePermissions, afterPermissions },
+      getChangeData(change).elemID.getFullName(),
+    )
+    return { permissionsToDeleted: [], permissionsToAdd: [] }
+  }
+
+  const beforePermissionsMap = createPermissionMap(beforePermissions)
+  const afterPermissionsMap = createPermissionMap(afterPermissions)
+  beforePermissionsMap.forEach(permission => {
+    const key = createPermissionUniqueKey(permission)
```
netama, line 96: isn't it already the item's key in the map?

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
```diff
@@ -0,0 +1,295 @@
...
+    if (beforePermissionsMap.has(key)) {
+      beforePermissionsMap.delete(key)
+      afterPermissionsMap.delete(key)
+    }
+  })
+  return {
+    permissionsToDeleted: Array.from(beforePermissionsMap.values()),
+    permissionsToAdd: Array.from(afterPermissionsMap.values()),
+  }
+}
```
netama, line 113: this can work - but I think you can simplify it into something like
```suggestion
  const overlappingPermissionKeys = new Set(_.intersection(beforePermissions.map(createPermissionUniqueKey), afterPermissions.map(createPermissionUniqueKey)))
  const removeOverlaps = (permissionList: PermissionObject[]): PermissionObject[] => permissionList.filter(permission => !overlappingPermissionKeys.has(createPermissionUniqueKey(permission))
  return {
    permissionsToDeleted: removeOverlaps(beforePermissions),
    permissionsToAdd: removeOverlaps(afterPermissions),
  }
}
```

(can also take it in different directions, there's `intersectionBy` and `differenceBy`)

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
```diff
@@ -0,0 +1,295 @@
...
+  changeResolver: ChangeElementResolver<Change<InstanceElement>>
+}
+
+const deployPermissions = async ({
+  spaceModificationChange,
+  ...args
+}: DeployPermissionsInput): Promise<DeployResult['errors']> => {
+  const changeData = getChangeData(spaceModificationChange)
+  const { permissionsToDeleted, permissionsToAdd } = calculatePermissionsDiff(spaceModificationChange)
+  log.debug('Found %o permissions to deleted and %o permissions to add', permissionsToDeleted, permissionsToAdd)
```
netama, line 130: nits:
1. `deleted` -> `delete`
2. maybe use counts here as well, and move the details to the end (or even to a trace log)?

(and the previous comment about using `%o` applies here as well)

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
```diff
@@ -0,0 +1,295 @@
...
+  const permissionsToDeletedIds = permissionsToDeleted
```
netama, line 131: typos
```suggestion
  const permissionsToDeleteIds = permissionsToDelete
```

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
```diff
@@ -0,0 +1,295 @@
...
+    .map(
+      permissionToDelete =>
+        changeData.value.permissionInternalIdMap?.[naclCase(createPermissionUniqueKey(permissionToDelete))],
```
netama, line 134: re the nacl-case - is any of the strings "custom" / are there always characters that will need to be escaped? (it's not critical because this field is hidden, but still if it's not needed it might be less error-prone to avoid)

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
```diff
@@ -0,0 +1,295 @@
...
+    )
+    .filter(values.isDefined)
+    .filter(_.isString)
+  const createPermissionRemoveChange = (id: string): RemovalChange<InstanceElement> => ({
+    action: 'remove',
+    data: {
+      before: new InstanceElement(id, permissionObjectType, { id }, [], {
```
netama, line 141: nits: 
1. please use `undefined` if there's no path and not an empty array (which, if used, could create a file in an unexpected location)
2. the elem id will be logged and I assume this permission might exist in other changes - maybe have the elem id also contain the parent's elem id (the name part), to simplify debugging?

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
```diff
@@ -0,0 +1,295 @@
...
+  const removalRes = await deployment.deployChanges({
+    changes: removalChanges,
+    ...args,
+  })
+  const additionChanges = permissionsToAdd.map(createPermissionAdditionChange)
+
+  const additionRes = await deployment.deployChanges({
+    changes: additionChanges,
+    ...args,
+  })
```
netama, line 178: these can fail in various scenarios (if some API calls succeed and some fail), let's add a ticket to follow up with more safeties

specifically, what happens if we try to create a permission that already exists in the service? should we silently catch that specific error without failing?

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
```diff
@@ -0,0 +1,295 @@
...
+  }
+  return [...removalRes.errors, ...additionRes.errors]
+}
+
+const filter =
+  ({
+    convertError = deployment.defaultConvertError,
+  }: filters.FilterCreationArgs<Options, UserConfig>): AdapterFilterCreator<{}, FilterResult, {}, Options> =>
+  ({ definitions, elementSource }) => ({
+    name: 'deploySpacePermissionsFilter',
```
netama, line 198: nit: we're also deploying the space, right? should it be `...SpaceAnd...`?

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
```diff
@@ -0,0 +1,295 @@
...
+    }
+  })
+  changeData.value.permissionInternalIdMap = {
+    ...changeData.value.permissionInternalIdMap,
+    ...newIdsToAdd,
+  }
+  return [...removalRes.errors, ...additionRes.errors]
+}
+
+const filter =
```
netama, line 193: please add a docstring

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
```diff
@@ -0,0 +1,295 @@
...
+  ({
+    convertError = deployment.defaultConvertError,
+  }: filters.FilterCreationArgs<Options, UserConfig>): AdapterFilterCreator<{}, FilterResult, {}, Options> =>
+  ({ definitions, elementSource }) => ({
+    name: 'deploySpacePermissionsFilter',
+    deploy: async (changes, changeGroup) => {
+      const { deploy, fetch, ...otherDefs } = definitions
+      if (deploy === undefined || fetch === undefined) {
+        throw new Error('could not find deploy or fetch definitions')
```
netama, line 202: nit:
```suggestion
        throw new Error('could not find space and space_permission definitions')
```

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
```diff
@@ -0,0 +1,295 @@
...
+        const { errors: spaceDeploymentErrors } = await deployment.deployChanges({
+          changes: [spaceChange],
+          ...deployArgsWithoutChanges,
+        })
+        if (!_.isEmpty(spaceDeploymentErrors)) {
+          errors.concat(spaceDeploymentErrors)
+          return
+        }
+        log.debug('successfully deployed space %s, starting deploy permissions', spaceChangeData.elemID.getFullName())
+        appliedChanges.concat(spaceChange)
```
netama, line 240: `.concat()` returns a new array, it doesn't add anything in-place. did you mean to use `.push()`?

(same for `errors.concat()` above)

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
```diff
@@ -0,0 +1,295 @@
...
+        const spaceChangeData = getChangeData(spaceChange)
+        const { errors: spaceDeploymentErrors } = await deployment.deployChanges({
+          changes: [spaceChange],
+          ...deployArgsWithoutChanges,
+        })
+        if (!_.isEmpty(spaceDeploymentErrors)) {
+          errors.concat(spaceDeploymentErrors)
+          return
+        }
+        log.debug('successfully deployed space %s, starting deploy permissions', spaceChangeData.elemID.getFullName())
```
netama, line 239:
```suggestion
        log.debug('successfully deployed space %s, starting to deploy permissions', spaceChangeData.elemID.getFullName())
```

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
```diff
@@ -0,0 +1,295 @@
...
+      const lookupFunc = references.generateLookupFunc(definitions.references?.rules ?? [])
+      const changeResolver = createChangeElementResolver<Change<InstanceElement>>({ getLookUpName: lookupFunc })
+      const deployArgsWithoutChanges = {
+        definitions: definitionsWithDeployAndFetch,
+        changeGroup,
+        elementSource,
+        convertError,
+        changeResolver,
+      }
+      await awu(spaceNonRemovalChanges).forEach(async spaceChange => {
```
netama, line 226: we can keep the `awu` call but just making sure, with the current grouping function there will only be one space per change group, right?

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
```diff
@@ -0,0 +1,295 @@
...
+              ),
+            })
+            // fetch default permissions created in the service
+            log.debug('fetching permissions created in the service for space %s', spaceChangeData.elemID.getFullName())
+            const itemsWithContext = await requester.requestAllForResource({
+              callerIdentifier: { typeName: PERMISSION_TYPE_NAME },
+              contextPossibleArgs: { id: [spaceChangeData.value.id] },
+            })
+            // spaceChangeData must be an InstanceElement. filtered changes at the beginnings of this filter
+            const fakeBeforeSpaceInstance = spaceChangeData.clone() as InstanceElement
```
netama, line 265: maybe just add a `.filter(isInstanceChange).filter(isAdditionOrModificationChange)` in the beginning after the `_.partition()`, and then we'll know all `spaceNonRemovalChanges` changes will already have the correct type and we can avoid this cast and the extra checks?

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
```diff
@@ -0,0 +1,295 @@
...
+            fakeBeforeSpaceInstance.value.permissions = itemsWithContext.map(item => item.value)
+            restructurePermissionsAndCreateInternalIdMap(fakeBeforeSpaceInstance.value)
+            // copy the created `permissionInternalIdMap` on the fake instance
```
netama, line 268: we're copying to the real instance here, no?

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
```diff
@@ -0,0 +1,295 @@
...
+            spaceChangeData.value.permissionInternalIdMap = fakeBeforeSpaceInstance.value.permissionInternalIdMap
+            const fakeModificationChange = toChange({
+              before: fakeBeforeSpaceInstance,
+              after: spaceChangeData,
+            }) as ModificationChange<InstanceElement>
```
netama, line 273: do we need the cast?

packages/confluence-adapter/src/definitions/transformation_utils/space.ts
```diff
@@ -0,0 +1,87 @@
...
+const log = logger(module)
+
+export type PermissionObject = {
+  type: string
+  principalId: string
+  key: string
+  targetType: string
+}
+
+export const isSpaceChange = ({ change }: definitions.deploy.ChangeAndContext): boolean => {
```
netama, line 33: I know this is existing code, but can't it just be a transformation with
```
condition: {
  transformForCheck: {
    omit: ['permissions'],
  }
}
```
?

netama (commented):
very cool 💪 added some comments and suggestions. I think we'll have some longtail of edge cases to cover, let's see if we cover all of them now or add some follow-ups

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
```diff
@@ -0,0 +1,295 @@
...
+const permissionObjectType = new ObjectType({ elemID: new ElemID(ADAPTER_NAME, PERMISSION_TYPE_NAME) })
+
+const isPermissionObjectArray = (value: unknown[]): value is PermissionObject[] => value.every(isPermissionObject)
+
+const createPermissionMap = (permissions: PermissionObject[]): Map<string, PermissionObject> => {
+  const map = new Map<string, PermissionObject>()
+  permissions.forEach(permission => {
+    map.set(createPermissionUniqueKey(permission), permission)
+  })
+  return map
```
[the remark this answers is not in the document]
omrifarkash: I started with it, this returns a Dictionary and not a Map and I think it is easier to work with map here to `.has` function

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
```diff
@@ -0,0 +1,295 @@
...
+              after: spaceChangeData,
+            }) as ModificationChange<InstanceElement>
+            const permissionErrors = await deployPermissions({
+              spaceModificationChange: fakeModificationChange,
+              ...deployArgsWithoutChanges,
+            })
+            errors.concat(permissionErrors)
+          } catch (e) {
+            log.error(
+              'failed to fetch permissions for space %s with error: %o',
```
netama, line 281: nit: failed to deploy?

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
omrifarkash: Yes

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
omrifarkash: I need a `ModificationChange` but the function returns `Change`. I can add a check but this is clearly a `ModificationChange`

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
omrifarkash: Actually I think the code that might throws an error here is the attempt to fetch the permissions (As `deployChanges` returns an error array if some request thrown an error)

packages/confluence-adapter/src/definitions/transformation_utils/space.ts
```diff
@@ -0,0 +1,87 @@
...
+  if (!isModificationChange(change)) {
+    return false
+  }
+  return !_.isEqual(_.omit(change.data.before.value, 'permissions'), _.omit(change.data.after.value, 'permissions'))
+}
+
+export const createPermissionUniqueKey = ({ type, principalId, key, targetType }: PermissionObject): string =>
+  `${type}_${principalId}_${key}_${targetType}`
+
+export const isPermissionObject = (value: unknown): value is PermissionObject =>
```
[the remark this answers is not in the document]
omrifarkash: I am familiar with `Joi` I think this is more readable because the type isn't complex

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
```diff
@@ -0,0 +1,295 @@
...
+  const beforePermissions = _.get(change.data.before.value, 'permissions')
+  const afterPermissions = _.get(change.data.after.value, 'permissions')
+  if (
+    !Array.isArray(beforePermissions) ||
+    !Array.isArray(afterPermissions) ||
+    !isPermissionObjectArray(beforePermissions) ||
+    !isPermissionObjectArray(afterPermissions)
+  ) {
+    log.warn(
+      'permissions are not in expected format: %o, skipping permissions deploy filter on space %s',
```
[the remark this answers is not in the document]
omrifarkash: I throw a SaltoError. Do you think this will do the trick? I probably should check it

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
omrifarkash: 100%

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
omrifarkash: I like

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
omrifarkash: https://salto-io.atlassian.net/browse/SALTO-5823

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
omrifarkash: Not working :/

packages/confluence-adapter/src/definitions/transformation_utils/space.ts
omrifarkash: Is it needed at all? (given that the filter will run before the default deploy)

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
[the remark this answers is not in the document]
omrifarkash: NVM, I changed this logic according to your suggestion below

packages/confluence-adapter/src/definitions/fetch/fetch.ts
```diff
@@ -150,7 +157,7 @@ const createCustomizations = (): Record<string, definitions.fetch.InstanceFetchA
             typeName: TEMPLATE_TYPE_NAME,
             addParentAnnotation: true,
             referenceFromParent: false,
-            nestPathUnderParent: false,
+            nestPathUnderParent: true,
```
omrifarkash, line 160: not relevant to the main change on this PR.
Just nesting templates under their space

packages/confluence-adapter/src/filters/deploy_space_permissions.ts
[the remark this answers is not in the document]
omrifarkash: Didn't work, I am leaving this to later
https://salto-io.atlassian.net/browse/SALTO-5823

packages/confluence-adapter/src/filters/deploy_space_and_permissions.ts
```diff
@@ -0,0 +1,301 @@
...
+        changeResolver,
+        sharedContext: {},
+      }
+      await awu(spaceNonRemovalChanges)
+        // This filter is for TypeScript, we already filtered the changes at the beginning of this filter
+        .filter(isInstanceChange)
+        .forEach(async spaceChange => {
+          if (!isInstanceChange(spaceChange)) {
+            return
+          }
```
netama, line 229: I think we can remove this check now?

netama (approved):
looks great 💪

packages/confluence-adapter/src/filters/deploy_space_and_permissions.ts
netama: not sure it's worth the extra build though :)

PR #5796, opened 2024-04-17
SALTO-5820 SALTO-5757 SALTO-5668 improve folder strucutre in google-workspace + confluence

netama:
* Use the filter from https://github.com/salto-io/salto/pull/5797 and adjust the definitions accordingly for Google Workspace `orgUnit` and for confluence `page`s
* Also fix the Confluence adapter unittests, which did not mock spaces correctly

---
_Release Notes_: 

_Google-Workspace Adapter_:
* Improve folder structure for org units

_Confluence Adapter_:
* Improve folder structure for pages

---
_User Notifications_: 
None


packages/google-workspace-adapter/src/adapter_creator.ts
```diff
@@ -77,6 +78,11 @@ export const adapter = createAdapter<Credentials, Options, UserConfig>({
   operationsCustomizations: {
     connectionCreatorFromConfig: () => createConnectionForApp(DIRECTORY_APP_NAME),
     credentialsFromConfig: defaultCredentialsFromConfig,
+    customizeFilterCreators: args => ({
+      ...filters.createCommonFilters<Options, UserConfig>(args),
```
Uri-Bar, line 82: maybe we can change some logic in here: https://github.com/salto-io/salto/blob/586e3b878823e6ff6c9a22f8c6ba71edce2d01cb/packages/adapter-components/src/adapter-wrapper/adapter_creator.ts#L119 so all the adapters wont need to pass the common filter, as all of the adapter will probably have some filters.
```suggestion
    operationsCustomizations: {
        runCommonFilters: true
        customizeFilterCreators: args => ({
              // customPathsFilterCreator must run after fieldReferencesFilter
              customPathsFilterCreator,
```

packages/google-workspace-adapter/src/definitions/fetch/fetch.ts
```diff
@@ -379,10 +379,7 @@ const createCustomizations = (): Record<string, definitions.fetch.InstanceFetchA
           aliasComponents: [{ fieldName: 'name' }],
         },
         path: {
-          pathParts: [
-            { parts: [{ fieldName: 'parentOrgUnitId', isReference: true }] },
-            { parts: [{ fieldName: 'name' }] },
-          ],
+          pathParts: [{ parts: [{ fieldName: 'name' }] }, { parts: [{ fieldName: 'name' }] }],
```
Uri-Bar, line 382: 🔥

packages/adapter-components/src/filter_utils.ts
```diff
@@ -49,6 +49,11 @@ export type AdapterFilterCreator<
   TOptions extends APIDefinitionsOptions = {},
 > = filter.FilterCreator<TResult, FilterOptions<TOptions, TContext, TAdditional>>
 
+export type NoOptionsFilterCreator<TResult extends void | filter.FilterResult = void> = filter.FilterCreator<
```
Uri-Bar, line 52: didnt fully understand, why do we need new type of filter?

packages/google-workspace-adapter/src/filters/custom_paths.ts
```diff
@@ -0,0 +1,47 @@
...
+    return undefined
+  }
+  return {
+    nestUnder: parentOrgUnitId.elemID,
+    pathSuffix: [currentFileName, currentFileName],
+  }
+}
+
+const filter: filterUtils.NoOptionsFilterCreator<filterUtils.FilterResult> =
+  filters.customPathsFilterCreator(pathMapper)
```
Uri-Bar, line 45: to connected with what amir wrote in the second PR. not for now but I vote for path mapper per type

packages/google-workspace-adapter/src/filters/custom_paths.ts
```diff
@@ -0,0 +1,47 @@
...
+    return undefined
+  }
+  const currentFileName = _.last(inst.path)
+  if (currentFileName === undefined) {
+    log.warn('org unit %s has no path, not updating', inst.elemID.getFullName())
+    return undefined
+  }
+  return {
+    nestUnder: parentOrgUnitId.elemID,
+    pathSuffix: [currentFileName, currentFileName],
```
Uri-Bar, line 40: this is a duplicate of the fetch definition ornUnit path. IIUC we have:
`pathParts: [{ parts: [{ fieldName: 'name' }] }, { parts: [{ fieldName: 'name' }] }],`
for the root orgUnit only. right?

Uri-Bar (approved):
looks great, left few questions and comments

packages/adapter-components/src/filter_utils.ts
netama: just to avoid having to pass unnecessary args that are not used, not strictly needed

packages/google-workspace-adapter/src/filters/custom_paths.ts
netama: yes definitely

packages/google-workspace-adapter/src/adapter_creator.ts
netama: the order is important so we'd need to add before and after filters and it complicates the interface a bit. we can do something similar to what we're doing in the change validators (Farkash had a suggestion that would allow full control of the order with some games), but I thought for now it's clearer to just require the extra line

packages/google-workspace-adapter/src/filters/custom_paths.ts
netama: it has the same result, but right now this is a custom function so it won't look there. I can change it to take the last two parts of the existing path instead

netama: changed, not sure this is what you were aiming for though

PR #5797, opened 2024-04-18
SALTO-5820 add filter supporting custom path nesting logic

netama:
Add shared filter that accepts a function defined by the adapter, and supports nesting elements under other elements. Once we gather more cases, we can formalize the common ones into the element definitions

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/adapter-components/src/filters/custom_paths.ts
```diff
@@ -0,0 +1,86 @@
...
+import { NoOptionsFilterCreator } from '../filter_utils'
+
+const log = logger(module)
+
+export type PathMapperFunc = (instance: InstanceElement) =>
+  | {
+      // the elem id of the instance to nest this instance's path under
+      nestUnder: ElemID
+      // the suffix to add when nesting
+      pathSuffix: string[]
```
Nurdok, line 31: Suggestion: make this optional, and the filter can use a default suffix that's the same as if the nesting was done via `nestStandaloneInstances` (`<type name>/<instance name>` I believe?)

I imagine that would be the usual case?

packages/adapter-components/src/filters/custom_paths.ts
```diff
@@ -0,0 +1,86 @@
...
+    })
+    graph.walkSync(id => {
+      const mapper = mappersByElemID[id]
+      if (mapper === undefined) {
+        // nothing to update
+        return
+      }
+      const { nestUnder, pathSuffix } = mapper
+      const instance = instancesByElemID[id]
+      const parent = instancesByElemID[nestUnder.getFullName()]
```
Nurdok, line 63: This is called `parent` here, but IIUC it doesn't have to be an actual parent (as in the parent annotation).

In Okta we have a use case where we need to manually make one type a child of a different type and also nest it under its new parent (see #5777). Maybe this filter could also accept a `makeParent: boolean` option to do that? 
Otherwise, maybe not call this `parent` here to avoid confusion?

packages/adapter-components/src/filters/custom_paths.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+  | undefined
+
+/**
+ * Shared filter for customizing element paths based on other elements, with a custom function.
+ * Dependencies are used for correct traversal (cycles should be avoided).
+ * Paths are nested as follows: <the path of the nestUnder instance, excluding the filename>/<path suffix>
+ */
+export const customPathsFilterCreator: <TResult extends void | filter.FilterResult = void>(
+  pathMapper: PathMapperFunc,
```
Nurdok, line 41: This filter accepts a single function, but an adapter might have several unrelated use cases for nesting. I guess it will either need a single `pathMapper` function to handle all the cases or instantiate the filter multiple times with different functions. The first option violates separation of concerns and will result in a monolithic function, while the second is problematic because you lose the correct traversal handling (as each filter instantiation will only consider a subset of the actual dependency graph).

Maybe take an array of functions and run them in order, using the first value that's not `undefined`?

Nurdok (commented):
Left some comments, but I'm not familiar enough with the codebase and use cases so I might be way off. Feel free to ignore :)

packages/adapter-components/src/filters/custom_paths.ts
netama: thanks, this is overriding anything that was done earlier (similarly to a filter we have in zendesk) - this is not intended to be generic, this is handling some specific cases that we need today :)

packages/adapter-components/src/filters/custom_paths.ts
netama: this actually is a logical parent for the path (and in these cases it's not written in the annotation because of some "product" decision but it is really a parent), it is following a similar pattern to `additionalParentFields` in the query filter. the goal here is just to update the paths for some cases that are not yet supported via the infra.
(btw @Uri-Bar @omrifarkash we might want to mark these fields as `additionalParentFields` for the query eventually, but let's see if it makes sense)

packages/adapter-components/src/filters/custom_paths.ts
netama: we can generalize it later (and move to the infra) - there are other cases not covered e.g. functions might need access to the entire elements list :)

Uri-Bar (approved):

packages/adapter-components/src/filters/custom_paths.ts
netama: will add a link to this suggestion from the follow-up ticket

packages/adapter-components/src/filters/custom_paths.ts
netama: re the filter from the linked PR - I _think_ you'll be able to just do this in the new infra using an `adjust` function, but we can also formalize as part of the transformation the option to explicitly define a parent (and other annotations). do you want to open a ticket for it? :)

PR #5802, opened 2024-04-18
SALTO-5738 support deploy for page restrictions

netama:
Clean up redundant details on restrictions on fetch + support deploy.

---

Tested on addition + modification

---
_Release Notes_: 
_Confluence Adapter_:
* Add support for deploying page restrictions

---
_User Notifications_: 
None

### DIFF
--- a/packages/confluence-adapter/src/definitions/deploy/deploy.ts
+++ b/packages/confluence-adapter/src/definitions/deploy/deploy.ts
@@ -42,6 +42,11 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
         customizations: {
           add: [
             {
+              condition: {
+                transformForCheck: {
+                  omit: ['restriction', 'version'],
+                },
+              },
               request: {
                 endpoint: {
                   path: '/wiki/api/v2/pages',
@@ -52,9 +57,26 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
                 },
               },
             },
+            {
+              request: {
+                endpoint: {
+                  path: '/wiki/rest/api/content/{id}/restriction',
+                  method: 'put',
+                },
+                transformation: {
+                  pick: ['restriction'],
+                  adjust: ({ value }) => ({ value: { results: _.get(value, 'restriction') } }),
+                },
+              },
+            },
           ],
           modify: [
             {
+              condition: {
+                transformForCheck: {
+                  omit: ['restriction', 'version'],
+                },
+              },
               request: {
                 endpoint: {
                   path: '/wiki/api/v2/pages/{id}',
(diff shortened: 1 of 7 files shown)


packages/confluence-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -50,6 +50,11 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
         customizations: {
           add: [
             {
+              condition: {
+                transformForCheck: {
+                  omit: ['restriction', 'version'],
```
omrifarkash, line 55: Does it matter on addition change? 
I think in addition this condition will always be true with or without omitting those fields?

packages/confluence-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -60,9 +65,26 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
...
             },
+            {
+              request: {
+                endpoint: {
+                  path: '/wiki/rest/api/content/{id}/restriction',
+                  method: 'put',
+                },
+                transformation: {
+                  pick: ['restriction'],
+                  adjust: ({ value }) => ({ value: { results: _.get(value, 'restriction') } }),
```
omrifarkash, line 76: `nestUnderField` won't work?

packages/confluence-adapter/src/definitions/fetch/fetch.ts
```diff
@@ -305,15 +305,20 @@ const createCustomizations = (): Record<string, definitions.fetch.InstanceFetchA
...
           root: 'results',
-          adjust: item => ({
-            value: {
-              operation: _.get(item.value, 'operation'),
-              restrictions: {
-                user: _.get(item.value, 'restrictions.user.results'),
-                group: _.get(item.value, 'restrictions.group.results'),
+          adjust: ({ value }) => {
+            const userRestrictions = _.get(value, 'restrictions.user.results')
+            return {
```
omrifarkash, line 310: Maybe extract the function to transform_utils folder? just to be aligned with the rest of the code in the adapter.
You can open a ticket about it and put it on me 🙏

omrifarkash (approved):
Thanks for the help!

packages/confluence-adapter/src/definitions/deploy/deploy.ts
netama: `transformForCheck` only runs on modifications (nothing to compare to on addition)

netama: I can remove it

packages/confluence-adapter/src/definitions/deploy/deploy.ts
netama: no, it converts it to an object instead of an array

PR #5807, opened 2024-04-21
SALTO-5807: Fix S3 dir store deletion

idoamit8:
In this PR, I've addressed the issue with the deletion of the S3 directory store.

---

1. The main bug I fixed is that `deleteObjects` expects relative paths, but I provided full paths.
2. I handled the interaction between `set` and `delete` operations to prevent conflicts.
3. I will add the `ignoreDeletionsCache` and revise the logic to support this in a separate PR, as I would like this to be insert today.

---
_Release Notes_: 
_Core_:
* Modified the deletion for s3 dir store. 

---
_User Notifications_: 
None


packages/core/src/local-workspace/s3_dir_store.ts
```diff
@@ -128,8 +130,10 @@ export const buildS3DirectoryStore = ({
       return Array.from(paths)
     }, `listing s3 objects for ${baseDir}`)
 
-  const deleteMany = async (objectPaths: string[]): Promise<void> => {
-    const objectIdentifiers: AWS.ObjectIdentifier[] = objectPaths.map(objectPath => ({ Key: getFullPath(objectPath) }))
+  const deleteMany = async (fileNames: string[]): Promise<void> => {
+    const objectIdentifiers: AWS.ObjectIdentifier[] = fileNames.map(fileName => ({
+      Key: getRelativeFileName(fileName),
```
ori-moisis, line 135: this is the actual path being sent to AWS, right?
if it doesn't include the baseDir, this is wrong, isn't it?

idoamit8: In the tests I added (Also in this PR), I think that the bestDir isn't included in the relativePath.

packages/core/src/local-workspace/s3_dir_store.ts
```diff
@@ -128,8 +128,10 @@ export const buildS3DirectoryStore = ({
       return Array.from(paths)
     }, `listing s3 objects for ${baseDir}`)
 
-  const deleteMany = async (objectPaths: string[]): Promise<void> => {
-    const objectIdentifiers: AWS.ObjectIdentifier[] = objectPaths.map(objectPath => ({ Key: getFullPath(objectPath) }))
+  const deleteMany = async (fileNames: string[]): Promise<void> => {
```
eshedg, line 131: this change seems to be meaningless...

packages/core/test/workspace/local/s3_dir_store.test.ts
```diff
@@ -278,23 +278,43 @@ describe('buildS3DirectoryStore', () => {
         Body: Buffer.from('aaa'),
       })
     })
-
-    it('should write the file only once', async () => {
```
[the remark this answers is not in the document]
idoamit8: Its here by mistake and its a leftover. 
The test is under the 'set' description and not delete

packages/core/src/local-workspace/s3_dir_store.ts
ori-moisis: it was actually a good change to be more consistent with the terminology of the dir store
can you please change it back?

ori-moisis (approved):

PR #5810, opened 2024-04-21
SALTO-2242 separate e2es into their own workflow

benisraeladi:
Separate each package's e2e test into its own job in CircleCI so that they run in parallel

---

This still runs all e2e tests regardless of which package changed, but will make it easier to see which package's e2e tests failed, and rerun only that package instead of all e2es

---
_Release Notes_: _None_

---
_User Notifications_: _None_


eshedg (approved):
🚀 🔥 👍

PR #5813, opened 2024-04-21
SALTO-1707: Removing previous newline and indentation when removing elements

yelly:
This solves the empty lines bug.

---

---
_Release Notes_: 
_Core_:
* Solves the bug that inserts empty lines into NaCl files when elements are removed.

---
_User Notifications_: 
None.

### DIFF
--- a/packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
+++ b/packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
@@ -114,8 +114,8 @@ const fixEdgeIndentation = (data: string, action: ActionName, initialIndentation
   if (action === 'add') {
     /* When adding the placement we are given is right before the closing bracket.
      * The string that dump gave us has an empty last line, meaning we have to recreate the
-     * indentation that was there previously. We also have to slice from the beggining of the first
-     * line the initial indentation that was there in the begginging.
+     * indentation that was there previously. We also have to slice from the beginning of the first
+     * line the initial indentation that was there in the beginning.
      */
     return [
       firstLine.slice(initialIndentationLevel),
@@ -247,10 +247,20 @@ export const updateNaclFileData = async (
         const nextChangeStart = change.start
         if (lastPartEnd !== nextChangeStart) {
           // Add slice from the current data to fill the gap between the last part and the next
+          let data = currentData.slice(lastPartEnd, nextChangeStart)
+
+          if (change.newData === '' && nextChangeStart !== Infinity) {
+            // For removals, removing newline and indent at the end of the block to avoid empty lines
+            const lastNewlineIndex = data.search(/\n[ \t]*$/)
+            if (lastNewlineIndex !== -1) {
+              data = data.slice(0, lastNewlineIndex)
+            }
+          }
+
           parts.push({
             start: lastPartEnd,
             end: nextChangeStart,
-            newData: currentData.slice(lastPartEnd, nextChangeStart),
+            newData: data,
           })
         }
         parts.push(change)
@@ -261,8 +271,7 @@ export const updateNaclFileData = async (
       [{ start: 0, end: 0, newData: '' }],
     )
 
-  const ret = allBufferParts.map(part => part.newData).join('')
-  return ret
+  return allBufferParts.map(part => part.newData).join('')
 }
 
 const wrapAdditions = (nestedAdditions: DetailedAddition[]): DetailedAddition => {
--- a/packages/workspace/test/workspace/nacl_files/nacl_file_update.test.ts
+++ b/packages/workspace/test/workspace/nacl_files/nacl_file_update.test.ts
@@ -27,6 +27,7 @@ import {
   getNestedStaticFiles,
   getChangeLocations,
   DetailedChangeWithSource,
+  updateNaclFileData,
 } from '../../../src/workspace/nacl_files/nacl_file_update'
 
 const mockType = new ObjectType({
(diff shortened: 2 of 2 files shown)


packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -247,10 +247,20 @@ export const updateNaclFileData = async (
         const nextChangeStart = change.start
         if (lastPartEnd !== nextChangeStart) {
           // Add slice from the current data to fill the gap between the last part and the next
+          let data = currentData.slice(lastPartEnd, nextChangeStart)
+
+          if (change.newData === '' && nextChangeStart !== Infinity) {
+            // For removals, removing newline and indent at the end of the block to avoid empty lines
+            const lastNewlineIndex = data.search(/\n[ \t]*$/)
```
ori-moisis, line 254: do we need to support all newline characters (e.g - https://github.com/salto-io/salto/blob/main/packages/parser/src/parser/internal/native/lexer.ts/#L52)?

I wonder if maybe it would be better to say that on removal, we remove all whitespace leading up to the thing we removed and change the pattern here to be just `/\s*$/`

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -247,10 +247,20 @@ export const updateNaclFileData = async (
         const nextChangeStart = change.start
         if (lastPartEnd !== nextChangeStart) {
           // Add slice from the current data to fill the gap between the last part and the next
+          let data = currentData.slice(lastPartEnd, nextChangeStart)
+
+          if (change.newData === '' && nextChangeStart !== Infinity) {
```
ori-moisis, line 252: why is the second condition needed?
how can there be a removal change at infinity?

packages/workspace/test/workspace/nacl_files/nacl_file_update.test.ts
```diff
@@ -110,3 +111,110 @@ describe('getChangeLocations', () => {
...
+          end: { col: 1, line: 10, byte: mockTypeNacl.length },
+        },
+      },
+    ]
+
+    const result = await updateNaclFileData(mockTypeNacl, changes, {})
+    expect(result).toEqual('')
+  })
+
+  it('should add fields to elements', async () => {
```
ori-moisis, line 169: usually the pattern you want to follow with tests is

```typescript
describe('myFunction', () => {
  describe('when <scenario / input>', () => {
    it('should <expected behavior / output>', () => {
    })
  })
})
```

the problem with how it is currently written is that it isn't clear what these test are actually expecting the result to be.
for example "should add fields to elements", when is that supposed to happen? what does it mean to "add a field to an element" in a function that only returns a string and (hopefully) has no side effects?


also, as a rule of thumb, in every PR that solves a specific bug, there should be at least one unit test that explicitly states the scenario of the bug and the expected behavior after the fix.
in this example, there should be a test which explicitly says when something (type, field, value, annotation, etc.) is removed, there are no "leftover" newlines / indentation.
e.g
```typescript
describe('updateNaclFileData', () => {
  ...
  describe('when removing a value from an instance', () => {
    beforeEach(() => {
      // implementation of the scenario
    })
    it('should remove the indentation of the value and preceding newline', () => {
    })
  })
})
```

the "beforeEach" part isn't mandatory, it is usually helpful when there are multiple things to check on the same scenario...

packages/workspace/test/workspace/nacl_files/nacl_file_update.test.ts
```diff
@@ -110,3 +111,110 @@ describe('getChangeLocations', () => {
...
+    const changes: DetailedChangeWithSource[] = [
+      {
+        ...toChange({ after: mockType.fields.numArray }),
+        id: mockType.fields.numArray.elemID,
+        location: {
+          filename: 'file',
+          start: { col: 1, line: 4, byte: 38 },
+          end: { col: 1, line: 4, byte: 38 },
```
ori-moisis, line 177: this is not a realistic scenario in the current code, right?
we never actually add things in the middle of a block

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
yelly: Using `/\s*#/` will cause extra spacing inserted by the user to be removed as well - do we want this?
I'm making the change but let's consider this before merging.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
yelly: Without the second condition we will remove the newline at the end of the file. Can you think of a cleaner way to write this?

packages/workspace/test/workspace/nacl_files/nacl_file_update.test.ts
yelly: Done.
The "should remove fields from elements" case covers the bug being fixed (this case fails without the change in the code). It seemed silly to explicitly state "and leaves no newlines", do you think I need to do that anyway?

packages/workspace/test/workspace/nacl_files/nacl_file_update.test.ts
yelly: `updateNaclFileData` makes no such assumption, but it is true that it is not used in this way (yet).
IMO verifying that this case works is actually something we want though, especially since this is something we want to do very soon.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
ori-moisis: your code does that as well (we never ident with tabs but you included tabs in the pattern), that is why I assumed that was your intent to remove all spacing

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
ori-moisis: oh. now I see.
I assumed `newData === ''` is always a removal but that is not actually the case.
maybe it would be clearer if this was more explicit - only do this for removal changes, not all changes where `newData === ''` (yes, I know that would require adding a bit more information to the change)

packages/workspace/test/workspace/nacl_files/nacl_file_update.test.ts
ori-moisis: the thing is, if you don't explicitly say what the expected result is, you are running the risk that future developers may "fix" your test in a way you did not intend.

consider someone changes the code in a way that breaks your fix, they would see a failing test that says "should remove field" and see that the field is indeed removed (but not the newline).
the easy way to "solve" this problem would be to change the test since the thing it seems to be testing is happening.

technically, it would be possible for that future dev to also notice the difference in whitespaces, and hopefully someone who remembers that we used to have this issue would catch it, so I won't say you have to have it explicit.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
yelly: Not quite, the original code only removes from the last newline, so if there are multiple newlines they won't be removed.
For example, if we look at this block:
```
type salto.mock {
  string file {
  }

  "List<number>" numArray {
  }

  "List<string>" strArray {
  }

  "List<salto.obj>" obj {
  }
}
```
And we remove `numArray`, the first iteration of the code will give:
```
type salto.mock {
  string file {
  }


  "List<string>" strArray {
  }

  "List<salto.obj>" obj {
  }
}
```
And the current iteration will give:
```
type salto.mock {
  string file {
  }

  "List<string>" strArray {
  }

  "List<salto.obj>" obj {
  }
}
```
Which I guess is better (at least in this case), but maybe we can find example where it isn't.

ori-moisis: this is not how nacl is formatted by default.
there aren't supposed to be empty lines between blocks

packages/workspace/test/workspace/nacl_files/nacl_file_update.test.ts
yelly: Makes sense, made the description more verbose.

packages/workspace/test/workspace/nacl_files/nacl_file_update.test.ts
yelly: Broke up the test cases to make everything more explicit.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
yelly: Done.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
yelly: That's what I'm saying - there will be no difference under default formatting, the only difference will be when there are changes made by the user.
Since this is not a main use case and since even in that case removing multiple newlines is probably better anyway let's stick with the current solution.

ori-moisis (approved):
overall looks good

one thing to consider - the unit tests added here are fairly targeted and isolated.
they don't cover a lot of different cases (only adding and removing fields in a type, which is not a very common case in reality), but for the abstraction level they are at, it doesn't really matter what kind of edit is tested so that is ok for these unit tests.

it might be good to also include a test for this behavior in the higher level test, like [here](https://github.com/salto-io/salto/blob/main/packages/workspace/test/workspace/workspace.test.ts/#L1623).
I think the relevant scenarios should already exist there (removing values from an instance / annotations from a field / type, etc...), it is just about adding a few `it(...)` blocks or editing a few existing ones to also verify the whitespace removal.

low level unit tests as the ones that were added are good, but they are not always good enough on their own...

yelly:
Sounds good. In order to avoid waiting another week to get this in, I'll merge now and add the extra tests in another PR.

PR #5816, opened 2024-04-21
SALTO-2688: fix e2e

edenhassid:
fix e2e

---

---
_Release Notes_: 

---
_User Notifications_: 


netama (approved):
LGTM, we don't have hard limits on these in the service?

PR #5823, opened 2024-04-24
SALTO-5845: fix a bug of projectIds in issue layouts

IdoZakk:
Not sure what the problem is. Added some guards to prevent, and logs to understand the root cause

---

None

---
_Release Notes_: 
Jira adapter:
* fixed an issue that caused some users to fail fetch in the issue layouts filter

---
_User Notifications_: 
None


shayc331 (approved):
🎖️

PR #5825, opened 2024-04-24
SALTO-5846: Okta: Add support for customized pages per brand [DEPRECATED]

Nurdok:
Add support for customized pages (Sign-in, Error) per brand - fetch, add, modify and remove.

---

I would like for this to have e2e coverage but it looks like the e2e environments do not support multiple brands.

---
_Release Notes_: 
_Okta adapter_:
- Added support for customized pages (Sign-in, Error) per brand - fetch, add, modify and remove.

---
_User Notifications_: 
None


packages/okta-adapter/src/config.ts
```diff
@@ -839,6 +849,74 @@ const DEFAULT_TYPE_CUSTOMIZATIONS: OktaSwaggerApiConfig['types'] = {
       dataField: '.',
     },
   },
+  SignInPage: {
+    request: {
+      url: '/api/v1/brands/{brandId}/pages/sign-in/customized',
+    },
+    transformation: {
```
shir-reifenberg, line 856: Do we have a `serviceIdField` we can use ?

shir-reifenberg (commented):
Yay 🎈🎉 looks good
Can you please add aliases for the new instances?
As for the E2E, we can test the modification flow, no? 
Also, please update the user notifications section as those will be added in the next fetch (and do we need noise reduction here? let's discuss IRL)

PR #5827, opened 2024-04-24
SALTO-5703 allow earlySuccess

netama:
Treat `earlySuccess: true` as a valid request and do not block it

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/adapter-components/src/deployment/change_validators/check_deployment_based_on_definitions.ts
```diff
@@ -48,7 +48,10 @@ const createChangeErrors = <Options extends APIDefinitionsOptions = {}>(
   action: ActionName,
 ): ChangeError[] => {
   const actionRequests = makeArray(queryWithDefault(typeRequestDefinition).query(action))
-  if (actionRequests.length === 0 || actionRequests.every(req => req.request.endpoint === undefined)) {
+  if (
+    actionRequests.length === 0 ||
+    actionRequests.every(req => req.request.endpoint === undefined && req.request.earlySuccess !== true)
```
Uri-Bar, line 53: I also prefer to us the flags as we usr the makeArray

Uri-Bar (approved):
looks good

PR #5830, opened 2024-04-24
SALTO-5427: Better error handling during fetch when some requests fail

shoshana-houri:
This PR adds a new field to the `FetchResourceDefinition` type, `onError`, which allows for custom error handling for a specific resource.
The error handler receives the error that was thrown during the fetch and can return exactly one of the following: `FailEntireFetch`, `CustomSaltoError`, `ConfigSuggestion`.
By default, which is also the behavior until now, any error thrown during the fetch is treated as if `FailEntireFetch` was set to false, meaning the fetch will log the error and continue, with only this resource missing.

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/definitions/system/fetch/resource.ts
+++ b/packages/adapter-components/src/definitions/system/fetch/resource.ts
@@ -13,10 +13,29 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import { SaltoError } from '@salto-io/adapter-api'
+import { types } from '@salto-io/lowerdash'
 import { ContextCombinationDefinition, RecurseIntoDefinition } from './dependencies'
 import { AdjustFunction, GeneratedItem, TransformDefinition } from '../shared'
+import { ConfigChangeSuggestion } from '../../user'
 
 export type ResourceTransformFunc = AdjustFunction<{ fragments: GeneratedItem[] }>
+
+type FailEntireFetch = {
+  // If set to true, the entire fetch will fail if an error is encountered while fetching this resource.
+  // By default - false, meaning the fetch will log the error and continue, with only this resource missing.
+  failEntireFetch: boolean
+}
+
+type CustomSaltoError = {
+  customSaltoError: SaltoError
+}
+
+type ConfigSuggestion = {
+  configSuggestion: ConfigChangeSuggestion
+}
+
+type OnErrorHandler = (e: Error) => types.XOR<types.XOR<FailEntireFetch, CustomSaltoError>, ConfigSuggestion>
 
 export type FetchResourceDefinition = {
   // set to true if the resource should be fetched on its own. set to false for types only fetched via recurseInto
@@ -39,4 +58,10 @@ export type FetchResourceDefinition = {
   // note: on overlaps the latest fragment wins ??
   // note: concatenation order between fragments is not defined.
   mergeAndTransform?: TransformDefinition<{ fragments: GeneratedItem[] }>
+
+  // Error handler for a specific resource.
+  // The error handler receives the error that was thrown during the fetch and can return exactly
+  // one of the following: FailEntireFetch, CustomSaltoError, ConfigSuggestion.
+  // By default, any error thrown during the fetch is treated as if FailEntireFetch was set to false.
+  onError?: OnErrorHandler
 }
--- a/packages/adapter-components/src/definitions/system/fetch/resource.ts
+++ b/packages/adapter-components/src/definitions/system/fetch/resource.ts
@@ -14,9 +14,8 @@
  * limitations under the License.
  */
 import { SaltoError } from '@salto-io/adapter-api'
-import { types } from '@salto-io/lowerdash'
 import { ContextCombinationDefinition, RecurseIntoDefinition } from './dependencies'
-import { AdjustFunction, GeneratedItem, TransformDefinition } from '../shared'
+import { AdjustFunction, GeneratedItem, ResultWithCustomizer, TransformDefinition } from '../shared'
 import { ConfigChangeSuggestion } from '../../user'
 
 export type ResourceTransformFunc = AdjustFunction<{ fragments: GeneratedItem[] }>
@@ -24,18 +23,21 @@ export type ResourceTransformFunc = AdjustFunction<{ fragments: GeneratedItem[] 
 type FailEntireFetch = {
   // If set to true, the entire fetch will fail if an error is encountered while fetching this resource.
   // By default - false, meaning the fetch will log the error and continue, with only this resource missing.
-  failEntireFetch: boolean
+  action: 'failEntireFetch'
+  value: boolean
 }
 
 type CustomSaltoError = {
-  customSaltoError: SaltoError
+  action: 'customSaltoError'
+  value: SaltoError
 }
 
 type ConfigSuggestion = {
-  configSuggestion: ConfigChangeSuggestion
+  action: 'configSuggestion'
+  value: ConfigChangeSuggestion
 }
 
-type OnErrorHandler = (e: Error) => types.XOR<types.XOR<FailEntireFetch, CustomSaltoError>, ConfigSuggestion>
+type OnErrorHandler = ResultWithCustomizer<FailEntireFetch | CustomSaltoError | ConfigSuggestion, Error>
 
 export type FetchResourceDefinition = {
   // set to true if the resource should be fetched on its own. set to false for types only fetched via recurseInto
@@ -60,8 +62,10 @@ export type FetchResourceDefinition = {
   mergeAndTransform?: TransformDefinition<{ fragments: GeneratedItem[] }>
 
   // Error handler for a specific resource.
-  // The error handler receives the error that was thrown during the fetch and can return exactly
-  // one of the following: FailEntireFetch, CustomSaltoError, ConfigSuggestion.
+  // The error handler is used to customize the behavior when an error occurs while fetching the resource.
+  // It can be one of the following: FailEntireFetch, CustomSaltoError, ConfigSuggestion.
+  // This can also be customized using a custom function that receives the error and returns the desired action.
+  // If defined - the custom handling will be prioritized.
   // By default, any error thrown during the fetch is treated as if FailEntireFetch was set to false.
   onError?: OnErrorHandler
 }
--- a/packages/adapter-components/src/definitions/system/fetch/resource.ts
+++ b/packages/adapter-components/src/definitions/system/fetch/resource.ts
@@ -15,7 +15,7 @@
  */
 import { SaltoError } from '@salto-io/adapter-api'
 import { ContextCombinationDefinition, RecurseIntoDefinition } from './dependencies'
-import { AdjustFunction, GeneratedItem, ResultWithCustomizer, TransformDefinition } from '../shared'
+import { AdjustFunction, ArgsWithCustomizer, GeneratedItem, TransformDefinition } from '../shared'
 import { ConfigChangeSuggestion } from '../../user'
 
 export type ResourceTransformFunc = AdjustFunction<{ fragments: GeneratedItem[] }>
@@ -37,7 +37,9 @@ type ConfigSuggestion = {
   value: ConfigChangeSuggestion
 }
 
-type OnErrorHandler = ResultWithCustomizer<FailEntireFetch | CustomSaltoError | ConfigSuggestion, Error>
+type OnErrorHandlerAction = FailEntireFetch | CustomSaltoError | ConfigSuggestion
+
+type OnErrorHandler = ArgsWithCustomizer<OnErrorHandlerAction, OnErrorHandlerAction, Error>
 
 export type FetchResourceDefinition = {
   // set to true if the resource should be fetched on its own. set to false for types only fetched via recurseInto
--- a/packages/adapter-components/src/definitions/system/shared/types.ts
+++ b/packages/adapter-components/src/definitions/system/shared/types.ts
@@ -33,6 +33,13 @@ export type ArgsWithCustomizer<ResultType, Args, Input = unknown, AdditionalArgs
   custom?: (args: Partial<Args> & AdditionalArgs) => (input: Input) => ResultType
 }
 
+export type ResultWithCustomizer<ResultType, Input = unknown> = types.XOR<
+  { result: ResultType },
+  {
+    custom: (input: Input) => ResultType
+  }
+>
+
 export type OptionsWithDefault<T, K extends string> = {
   options: Record<K, T>
   default: K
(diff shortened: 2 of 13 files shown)


packages/adapter-components/src/definitions/system/fetch/resource.ts
```diff
@@ -13,11 +13,30 @@
...
+
+type CustomSaltoError = {
+  customSaltoError: SaltoError
+}
+
+type ConfigSuggestion = {
+  configSuggestion: ConfigChangeSuggestion
+}
+
+type OnErrorHandler = (e: Error) => types.XOR<types.XOR<FailEntireFetch, CustomSaltoError>, ConfigSuggestion>
```
netama, line 38: 1. styling nit: maybe 
```suggestion
type OnErrorHandler = (e: Error) => types.OneOf<{
  // If set to true, the entire fetch will fail if an error is encountered while fetching this resource.
  // By default - false, meaning the fetch will log the error and continue, with only this resource missing.
  failEntireFetch: boolean
  customSaltoError: SaltoError
  configSuggestion: ConfigChangeSuggestion
}>
```
2. might be worth maintaining the special handling for permission errors that we had in the [old infra](https://github.com/salto-io/salto/blob/main/packages/adapter-components/src/elements_deprecated/ducktype/transformer.ts#L455-L458)? and similarly the basic handling for [config suggestions](https://github.com/salto-io/salto/blob/main/packages/adapter-components/src/elements_deprecated/ducktype/transformer.ts#L441-L447)
3. I'm not sure how we combine a default `failEntireFetch = true` with a type override that has, say, a config suggestion. wouldn't we need to override both args?

packages/adapter-components/src/fetch/element/element.ts
```diff
@@ -76,6 +82,29 @@ export const getElementGenerator = <Options extends FetchApiDefinitionsOptions>(
...
+      throw new AbortFetchOnFailure({ adapterName, typeName, message: error.message })
+    }
+
+    if (onError?.customSaltoError !== undefined) {
+      log.warn('failed to fetch type %s:%s, generating custom Salto error', adapterName, typeName)
+      customSaltoErrors.push(onError.customSaltoError)
+    } else if (onError?.configSuggestion !== undefined) {
+      log.warn('failed to fetch type %s:%s, generating config suggestions', adapterName, typeName)
+      configSuggestions.push(onError.configSuggestion)
+    } else log.warn('failed to fetch type %s:%s: %s', adapterName, typeName, error.message)
```
netama, line 105: readability nit: 
```suggestion
    } else {
      log.warn('failed to fetch type %s:%s: %s', adapterName, typeName, error.message)
    }
```

packages/adapter-components/src/fetch/resource/resource_manager.ts
```diff
@@ -91,21 +97,29 @@ export const createResourceManager = <ClientOptions extends string>({
...
           typeFetcherCreator: createTypeFetcher, // used for recurseInto calls
         })
+
+        const typeNameAsStr = String(typeName)
         if (!res.success) {
-          log.warn('failed to fetch type %s:%s:', adapterName, typeName)
+          try {
+            elementGenerator.handleError({ typeName: typeNameAsStr, error: res.error })
+          } catch (err) {
+            e = err
```
netama, line 110: would be nice if we could avoid this hack (which also spreads to a few places and doesn't let us aggregate the errors) - it looks like the `WalkError` still keeps the inner errors under `handlerErrors`, can we use these instead?

packages/adapter-components/src/fetch/resource/resource_manager.ts
```diff
@@ -80,7 +81,12 @@ export const createResourceManager = <ClientOptions extends string>({
         lowerdashValues.isDefined,
       )
       const graph = createDependencyGraph(resourceDefQuery.getAll())
+      let e: unknown
       await graph.walkAsync(async typeName => {
+        // If we already have an error, we don't want to continue actually fetching resources.
+        // We don't throw the error here, since we don't want it to be wrapped by the graph error.
+        if (e) return
```
netama, line 88: I think this code will disappear, but in general please avoid if's with inline logic :)

packages/adapter-components/src/fetch/resource/type_fetcher.ts
```diff
@@ -184,15 +192,9 @@ export const createTypeResourceFetcher = <ClientOptions extends string>({
...
-      if (_.isError(e)) {
-        return {
-          success: false,
-          errors: [e],
-        }
-      }
       return {
         success: false,
-        errors: [new Error(String(e))],
+        error: _.isError(e) ? e : new Error(String(e)),
```
netama, line 197: I think we should still assume there can be more than one error (this applies to the entire solution). is there a reason we can't support it?

packages/adapter-components/src/fetch/types.ts
```diff
@@ -30,10 +30,14 @@ export type ResourceIdentifier = {
...
-  errors?: Error[]
-}
+type ResourceFetchResult =
+  | {
+      success: true
+    }
+  | {
+      success: false
+      error: Error
+    }
```
netama, line 40: why did this need to change? (besides the switch to a single error, which I think we shouldn't assume)

netama (commented):
cool, added some comments and questions

packages/adapter-components/src/definitions/system/fetch/resource.ts
shoshana-houri: 1+2 🚀 🚀  Thanks!!
As for 3 I'm not sure I understand, the field `failEntireFetch` is set to false by default 🤔

packages/adapter-components/src/fetch/resource/type_fetcher.ts
shoshana-houri: Perhaps I’m overlooking something, but since this error originates solely from the try & catch clause, there couldn’t be multiple errors from a single resource. However, we do support handling multiple errors for **multiple** resources, including sub-resources.

packages/adapter-components/src/fetch/resource/resource_manager.ts
shoshana-houri: We can wrap the entire graph walk in a try & catch clause and then extract the `handlerErrors`. You're right that this will solve the original intent of this hack, which was to eliminate the error wrapping performed by the graph. However, when I reviewed the code, I realized that even if an error is thrown during the async walk, we continue fetching other resources not dependent on the failed one. Since we only throw the error when we want the entire fetch to fail, this continuation is unnecessary. This is why I added [this check](https://github.com/salto-io/salto/pull/5830/files/3efcf11718f973313c604991fea7995d1b2bee2c#diff-b2b3e8caff7e00cfb9b1193cdcde0e66b077a3e9295ff925d71a89184628af90R88) to avoid redundant calls.
What do you think?

netama: hmm I think the check doesn't do exactly this though, because you don't know inside the next step whether it depends on the errored one or on something else that completed in parallel / later, no?

packages/adapter-components/src/definitions/system/fetch/resource.ts
shoshana-houri: Got it! That would be true if we had set these fields directly on the resource. 
However, the definition is not: `default.resource.failEntireFetch = true`,
instead, it's: `default.resource.onError = (e) => {...someLogic; return something;}` (in order to be able to define specific actions for specific errors). 
So, when we merge the definitions, the custom one will entirely override the default one :)

packages/adapter-components/src/fetch/resource/resource_manager.ts
shoshana-houri: Yes! The code section you sent is exactly what I meant by wrapping it with a try and catch, and extracting the `handlerErrors` :)
However, I think this use case is a bit different. If we actively set `failEntireFetch` to true and aim to maintain the behavior of failing fetches for ducktypes, it implies that we don't want a partial success, as happens when some deploy requests fail. Instead, we prefer not to apply any changes at all. If that's indeed the case, then fetching other resources that are not dependent could be a waste of time when we know the fetch will fail anyway.
We can also discuss this f2f :)

packages/adapter-components/src/definitions/system/fetch/resource.ts
shoshana-houri: Discussed f2f - decided to redefine the type in a way that is similar to `ArgsWithCustomizer`

packages/adapter-components/src/fetch/resource/resource_manager.ts
shoshana-houri: Discussed f2f - decided to update our DAG code to allow throwing an error during the walk

packages/adapter-components/src/fetch/resource/type_fetcher.ts
shoshana-houri: Discussed f2f - it's ok (at least for now)

packages/dag/src/nodemap.ts
```diff
@@ -40,6 +40,12 @@ export class NodeSkippedError extends Error {
   }
 }
 
+export class FatalError extends Error {
```
shoshana-houri, line 43: can make these changes in a separate PR, but I think it's somewhat better to have the usage context as well. Let me know if you think otherwise :)

packages/adapter-components/src/definitions/system/fetch/resource.ts
```diff
@@ -13,11 +13,32 @@
...
-import { AdjustFunction, GeneratedItem, TransformDefinition } from '../shared'
+import { AdjustFunction, GeneratedItem, ResultWithCustomizer, TransformDefinition } from '../shared'
+import { ConfigChangeSuggestion } from '../../user'
 
 export type ResourceTransformFunc = AdjustFunction<{ fragments: GeneratedItem[] }>
 
+type FailEntireFetch = {
+  // If set to true, the entire fetch will fail if an error is encountered while fetching this resource.
+  // By default - false, meaning the fetch will log the error and continue, with only this resource missing.
+  action: 'failEntireFetch'
```
shoshana-houri, line 26: This format (`action` and `value`) ensures that we won't have multiple conflicting values if we also have a default definition. This way the customized one will just override the default.

packages/dag/src/nodemap.ts
netama: @ori-moisis probably best that you review this part as well (and say what you prefer re splitting the PRs)?

packages/adapter-components/src/definitions/system/shared/types.ts
```diff
@@ -33,6 +33,13 @@ export type ArgsWithCustomizer<ResultType, Args, Input = unknown, AdditionalArgs
   custom?: (args: Partial<Args> & AdditionalArgs) => (input: Input) => ResultType
 }
 
+export type ResultWithCustomizer<ResultType, Input = unknown> = types.XOR<
```
netama, line 36: let's discuss IRL - I'm not sure the difference from `ArgsWithCustomizer` "justifies" introducing another concept if it's not strictly needed (trying to keep the generic ones to a minimum) and I _think_ you can do the same with the existing one, but I might be missing something about what we lose

packages/adapter-components/src/fetch/resource/resource_manager.ts
```diff
@@ -80,32 +83,48 @@ export const createResourceManager = <ClientOptions extends string>({
...
-        if (!res.success) {
-          log.warn('failed to fetch type %s:%s:', adapterName, typeName)
-          return
+      } catch (e) {
+        if (e instanceof WalkError) {
+          // In case we decided to fail the entire fetch we don't want the error to be wrapped in a WalkError
+          const failFetchError = wu(e.handlerErrors.values()).find(err => err instanceof AbortFetchOnFailure)
+          if (failFetchError !== undefined) {
+            // Since there may be other errors in the handlerErrors, we log them all
+            log.error('Received at least one AbortFetchOnFailure error, failing the entire fetch. Full error:', e)
```
netama, line 120: just making sure there are enough details in the log? I think we might want to e.g. log the stack trace as well?

packages/adapter-components/src/fetch/resource/resource_manager.ts
```diff
@@ -80,32 +83,48 @@ export const createResourceManager = <ClientOptions extends string>({
...
+            throw failFetchError
+          }
         }
-        elementGenerator.pushEntries({
-          typeName: String(typeName),
-          entries: resourceFetcher.getItems()?.map(item => item.value) ?? [],
-        })
-      })
+        // If we got here, it means that the error is not an AbortFetchOnFailure error, so we revert to
+        // the original error behavior, where this section wasn't wrapped in a try-catch.
```
netama, line 125: nit: this comment may explain the diff, but I think people reading the code will be more interested in the behavior and the "justification" than the historical context :)

netama (commented):
looks great :) the remaining point to discuss from my POV is whether we need `ResultWithCustomizer` or can make this work with `ArgsWithCustomizer`, and will be good if @ori-moisis can take a look at the graph part as well

packages/dag/test/nodemap.test.ts
```diff
@@ -885,6 +886,34 @@ describe('NodeMap', () => {
...
+                return new Promise(r => {
+                  resolve = r
+                })
+              }
+              return undefined
+            })
+            .catch(e => {
+              error = e
+            })
+          expect(error).toBeDefined()
```
ori-moisis, line 908: we should separate the setup from the test.
either have the setup like this but don't `await` for the `walkAsync` in the setup (only `await` in the test).

also, what about `walkSync`? do we expect the same behavior? (if not, why not?), there should be a test for it too IMO

packages/adapter-components/src/fetch/element/element.ts
```diff
@@ -76,6 +82,38 @@ export const getElementGenerator = <Options extends FetchApiDefinitionsOptions>(
...
+    // This can happen if the error was thrown inside a sub-type that has failEntireFetch set to true.
+    // In this case we should not call the parent's onError function.
+    if (error instanceof AbortFetchOnFailure) {
+      throw error
+    }
+
+    const { resource: resourceDef } = defQuery.query(typeName) ?? {}
+    const onError = resourceDef?.onError
+
+    const onErrorResult = onError?.custom?.({})(error) ?? onError
```
netama, line 95: nit: even if today we don't pass other args we might in the future - maybe pass the args to `custom(...)` instead of the empty object?

netama (approved):
LGTM (reviewed only the changes and approved based on the previous review, lmk if you want another pass)

PR #5832, opened 2024-04-25
SALTO-4976: remove missing reference when transitionId is empty

shayc331:
_remove missing reference when transitionId is empty_

---

* the users can use the transition name as a reference. Currently when they do that the id field is empty and they get a missingReference  that creates a fetch warning and blocks their deployment

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/jira-adapter/src/filters/workflow/transition_structure.ts
+++ b/packages/jira-adapter/src/filters/workflow/transition_structure.ts
@@ -151,10 +151,12 @@ export const walkOverTransitionIds = (transition: WorkflowTransitionV1, func: (v
   transition.rules?.postFunctions
     ?.filter(postFunction => postFunction.type === SCRIPT_RUNNER_POST_FUNCTION_TYPE)
     .forEach(postFunction => {
-      if (postFunction.configuration?.scriptRunner?.transitionId === undefined) {
-        return
+      if (
+        postFunction.configuration?.scriptRunner?.transitionId !== undefined &&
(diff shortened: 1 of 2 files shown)


packages/jira-adapter/src/filters/workflow/transition_structure.ts
```diff
@@ -151,10 +151,12 @@ export const walkOverTransitionIds = (transition: WorkflowTransitionV1, func: (v
   transition.rules?.postFunctions
     ?.filter(postFunction => postFunction.type === SCRIPT_RUNNER_POST_FUNCTION_TYPE)
     .forEach(postFunction => {
-      if (postFunction.configuration?.scriptRunner?.transitionId === undefined) {
-        return
+      if (
+        postFunction.configuration?.scriptRunner?.transitionId !== undefined &&
+        !_.isEmpty(postFunction.configuration.scriptRunner.transitionId)
+      ) {
+        func(postFunction.configuration.scriptRunner)
```
IdoZakk, line 158: I am a bit worried that the change is in a generic place. Please verify there is no existing situation in which the value is empty, but we do want to run the func

IdoZakk (approved):
👌 Please review my comment

packages/jira-adapter/src/filters/workflow/transition_structure.ts
shayc331: for now when its empty we are getting a missing reference. It will be empty on purpose when the user wants to use the name as a reference (we do not support it yet)

PR #5834, opened 2024-04-25
SALTO-5834: Omit valueSet from Standard Picklist Fields (Salesforce)

tamtamirr:
Omit valueSet from Standard Picklist Fields.

---

Standard Picklist Fields that are not attached to a StandardValueSet, have a constant valueSet that cannot be modified.

The existing **picklistStandardField Change Validator** is redundant and blocks valid changes on standard picklist fields when the feature is enabled. This also indicate that omitting the valueSet added support in deploying valid changes that were not supported before.

---
_Release Notes_: 
Salesforce Adapter:
- Added the feature `omitStandardFieldsNonDeployableValues` that omits non deployable valueSet from Standard Picklist Fields that are not attached to a **StandardValueSet**.

---
_User Notifications_: 
Salesforce:
- Enabling the feature **omitStandardFieldsNonDeployableValues** will remove the `valueSet` annotation from Standard Picklist Fields.


packages/salesforce-adapter/test/change_validators/picklist_standard_field.test.ts
tamtamirr, line 1: Most of the test cases didn't make sense at all, and are not aligned with the expected behavior.
I've created a single test that covers all the possible cases. 💯 % coverage

packages/salesforce-adapter/src/filters/omit_standard_fields_non_deployable_values.ts
```diff
@@ -0,0 +1,61 @@
...
+      'Error occurred when attempting to omit standard fields non deployable values',
+    fetchFilterFunc: async (elements) => {
+      const standardFields = elements
+        .filter(isCustomObjectSync)
+        .flatMap((customObject) => Object.values(customObject.fields))
+        .filter(isStandardField)
+      const standardFieldsWithValueSet = standardFields.filter(
+        (field) => field.annotations[FIELD_ANNOTATIONS.VALUE_SET] !== undefined,
+      )
+      log.debug(
```
tomsellek, line 45: Maybe consider making this a `trace` if there might be many fields like this?

packages/salesforce-adapter/src/filters/omit_standard_fields_non_deployable_values.ts
```diff
@@ -0,0 +1,61 @@
...
+        'omitting valueSet from the following standard fields: %s',
+        safeJsonStringify(
+          standardFieldsWithValueSet.map((field) => apiNameSync(field)),
+        ),
+      )
+
+      standardFields.forEach((field) => {
```
tomsellek, line 52: Why not use `standardFieldsWithValueSet` here?

tomsellek (commented):
Overall looks good, but I had a couple of questions.

packages/salesforce-adapter/src/change_validators/picklist_standard_field.ts
```diff
@@ -14,41 +14,19 @@
...
-const shouldCreateChangeError = async (
-  changeData: ChangeDataType,
-): Promise<boolean> =>
-  isPicklistField(changeData) &&
-  !isCustom(await apiName(changeData)) &&
-  !(await isStandardValueSet(changeData))
+const isStandardPicklistFieldWithValueSet = (field: Field): boolean =>
+  isStandardField(field) &&
+  isPicklistField(field) &&
+  field.annotations.valueSet !== undefined
```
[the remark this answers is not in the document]
tamtamirr: This is on purpose. The code is incorrect and yielded errors on valid changes.
I don't think it's necessary to validate references as part of a change validator has it's already covered in many other flows.

packages/salesforce-adapter/src/filters/omit_standard_fields_non_deployable_values.ts
tamtamirr: There aren't too many, but I'll slice for 100

packages/salesforce-adapter/jest.config.js
```diff
@@ -25,7 +25,7 @@ module.exports = deepMerge(require('../../jest.base.config.js'), {
     : undefined,
   coverageThreshold: {
     global: {
-      branches: 88.69,
+      branches: 88.67,
```
tamtamirr, line 28: The ternary operator when using trace/debug isn't covered.

packages/salesforce-adapter/src/change_validators/picklist_standard_field.ts
```diff
@@ -62,12 +35,11 @@ const createChangeError = (field: Field): ChangeError => ({
...
+  changes
+    .filter(isFieldChange)
     .filter(isModificationChange)
     .map(getChangeData)
-    .filter(shouldCreateChangeError)
-    // We can cast since shouldCreateChangeError only return true to fields
-    .map((field) => createChangeError(field as Field))
-    .toArray()
+    .filter(isStandardPicklistFieldWithValueSet)
+    .map((field) => createChangeError(field))
```
yelly, line 43: I think we should update the error message here because it isn't really accurate any more.

packages/salesforce-adapter/src/filters/omit_standard_fields_non_deployable_values.ts
```diff
@@ -0,0 +1,65 @@
...
+      'Error occurred when attempting to omit standard fields non deployable values',
+    fetchFilterFunc: async (elements) => {
+      const standardFields = elements
+        .filter(isCustomObjectSync)
+        .flatMap((customObject) => Object.values(customObject.fields))
+        .filter(isStandardField)
+      const standardFieldsWithValueSet = standardFields.filter(
+        isStandardPicklistFieldWithValueSet,
+      )
+      ;(standardFieldsWithValueSet.length > 100 ? log.trace : log.debug)(
```
yelly, line 46: This is a bit weird IMO. Makes more sense to me to only print the first N field names instead of demoting the log in cases when there are lots (which is actually probably the more interesting cases).
Actually, why do we even need this log at all? And why are we interested in `valueSet` but not `referenceTo`?

packages/salesforce-adapter/src/filters/omit_standard_fields_non_deployable_values.ts
```diff
@@ -0,0 +1,65 @@
...
+        'omitting valueSet from the following standard fields: %s',
+        safeJsonStringify(
+          standardFieldsWithValueSet.map((field) => apiNameSync(field)),
+        ),
+      )
+
+      standardFields.forEach((field) => {
+        if (field.annotations[FIELD_ANNOTATIONS.VALUE_SET] !== undefined) {
```
yelly, line 54: No need for this check (and the next one as well), `delete` works just fine for undefined keys.

packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -1165,3 +1165,8 @@ export const toCustomProperties = async (
     ..._.pickBy(element.annotations, (_val, name) => isAllowed(name)),
   }
 }
+
+export const isStandardPicklistFieldWithValueSet = (field: Field): boolean =>
```
yelly, line 1169: Why put this in utils? We only use it once.

packages/salesforce-adapter/src/filters/omit_standard_fields_non_deployable_values.ts
yelly: Because then you won't catch the fields with just `referenceTo`.

packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -1165,3 +1165,8 @@ export const toCustomProperties = async (
...
+  isStandardField(field) &&
+  isPicklistField(field) &&
+  field.annotations.valueSet !== undefined
```
yelly, line 1172: Maybe use the `FIELD_ANNOTATIONS.REFERENCE_TO` const here.

packages/salesforce-adapter/test/filters/omit_standard_fields_non_deployable_values.test.ts
```diff
@@ -0,0 +1,128 @@
...
+                ),
+              ),
+              referenceTo: ['Case', 'User'],
+            },
+          },
+        },
+      })
+    })
+
+    it('should omit valueSet from standard fields', async () => {
```
yelly, line 88: Missing a test case for `referenceTo`.

packages/salesforce-adapter/src/filters/utils.ts
tamtamirr: This is used both in the Change Validator and the Filter

packages/salesforce-adapter/src/change_validators/picklist_standard_field.ts
tamtamirr: This is the same error returnd from Salesforce actually. I'd rather stick to this. Otherwise, what did you have in mind?

packages/salesforce-adapter/src/filters/utils.ts
tamtamirr: Since I've removed the log, there was only one usage in CV. I've moved it to a private method within the CV

packages/salesforce-adapter/test/filters/omit_standard_fields_non_deployable_values.test.ts
tamtamirr: This test covers both actually :) I've updated it's name

packages/salesforce-adapter/src/filters/omit_standard_fields_non_deployable_values.ts
tamtamirr: removed log

packages/salesforce-adapter/src/change_validators/picklist_standard_field.ts
yelly: We can make this more specific, something like "Standard picklist fields cannot have a value set defined." (and also update the detailed message accordingly). The current wording is confusing IMO (for example, IIUC, standard fields can be defined a picklist, we just can't deploy a value set).

packages/salesforce-adapter/src/filters/utils.ts
yelly: It's still here. Did you forget to push?

packages/salesforce-adapter/src/filters/omit_standard_fields_non_deployable_values.ts
```diff
@@ -0,0 +1,46 @@
...
+        .filter(isCustomObjectSync)
+        .flatMap((customObject) => Object.values(customObject.fields))
+        .filter(isStandardField)
+
+      standardFields.forEach((field) => {
+        delete field.annotations[FIELD_ANNOTATIONS.VALUE_SET]
+        delete field.annotations[FIELD_ANNOTATIONS.REFERENCE_TO]
+      })
+    },
+  }),
```
yelly, line 43: Just a stylistic preference, but this can just be a long chain.
```suggestion
    fetchFilterFunc: async (elements) => elements
      .filter(isCustomObjectSync)
      .flatMap((customObject) => Object.values(customObject.fields))
      .filter(isStandardField)
      .forEach((field) => {
        delete field.annotations[FIELD_ANNOTATIONS.VALUE_SET]
        delete field.annotations[FIELD_ANNOTATIONS.REFERENCE_TO]
      }),
```

yelly (approved):

PR #5835, opened 2024-04-25
SALTO-5738 handle default restriction on page deployment

omrifarkash:
- When deploying a page addition with default restriction we should not modify restriction after addition as default restrictions are created anyway
- When deploying a page modification with default restriction we should not modify restriction but delete the existing restriction. Deletion, resets the restriction to the default

---

---

_Release Notes_: 
_Confluence Adapter_:
- Fully support restrictions on page deployment

---

_User Notifications_: 
_Confluence Adapter_:
- Fully support restrictions on page deployment

### DIFF
--- a/packages/confluence-adapter/src/definitions/transformation_utils/restriction.ts
+++ b/packages/confluence-adapter/src/definitions/transformation_utils/restriction.ts
@@ -15,7 +15,7 @@
  */
 import _ from 'lodash'
 import { definitions } from '@salto-io/adapter-components'
-import { getChangeData, isModificationChange } from '@salto-io/adapter-api'
+import { getChangeData, isEqualValues, isModificationChange } from '@salto-io/adapter-api'
 
 export const DEFAULT_RESTRICTION = [
   {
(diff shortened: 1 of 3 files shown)


packages/confluence-adapter/src/definitions/transformation_utils/restriction.ts
```diff
@@ -15,6 +15,38 @@
...
+/**
+ * Check if need to modify restriction on page modification
+ * If user change page restriction to the default, we need to delete the existing restriction
+ */
+export const shouldDeleteRestrictionOnPageModification = (args: definitions.deploy.ChangeAndContext): boolean => {
+  if (!isModificationChange(args.change)) {
+    return false
+  }
+  const afterRestriction = _.get(args.change.data.after.value, 'restriction')
+  if (_.isEqual(args.change.data.before.value.restriction, afterRestriction)) {
```
netama, line 45: please use `isEqualValues` here, since there might be references in the restrictions

netama (approved):
LGTM with one comment, thanks!

PR #5836, opened 2024-04-26
SALTO-5674 - Salesforce:Quick Fetch omits data instances due to missing refs

tomsellek:
When we fetch instances in fetch-with-changes-detection, we may fetch referring instances without the instances they refer to. In order to resolve these references we need to look for the reference targets in the workspace.

---

Tests:
* Run the test case described in the ticket. Confirm it fails before this change and passes after it.

---
_Release Notes_: 
Salesforce: References from custom object instances will now be resolved correctly in quickfetch.

---
_User Notifications_: 
N/A


tamtamirr (approved):
LGTM!I However the diff appears bigger than it should be. This is due to reformat?

PR #5839, opened 2024-04-30
SALTO-5603: Do not fetch Partial requestType

idoamit8:
In this PR I excluded requestTypes that being partial fetched. 

---

There is a known issue where lacking the "Browse Projects" and "Service Project Agent" permissions for a specific project prevents access to the RequestType. In the UI, this discrepancy is evident, whereas the API returns a partial RequestType, contradicting the user's experience and causing issues during deployment attempts. The error message received in such cases is: "Entity associated with issue layout does not exist or user does not have required permissions." Consequently, when this message appears, we should avoid fetching the RequestType altogether.

---
_Release Notes_: 
_Jira Adapter:_
* Stop fetching requestsType that has no requestForm or issueView.

---
_User Notifications_: 
_Jira Adapter:_
* Stop fetching requestsType that has no requestForm or issueView.

### DIFF
--- a/packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
+++ b/packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
@@ -41,6 +41,8 @@ import {
   createLayoutType,
   IssueLayoutConfiguration,
   IssueLayoutConfig,
+  ERROR_RESPONSE_SCHEME,
+  errorResponse,
 } from './layout_types'
 import { ISSUE_LAYOUT_TYPE, ISSUE_VIEW_TYPE, JIRA, REQUEST_FORM_TYPE, REQUEST_TYPE_NAME } from '../../constants'
 import { DEFAULT_API_DEFINITIONS } from '../../config/api_config'
@@ -86,6 +88,7 @@ export const LAYOUT_TYPE_NAME_TO_DETAILS: Record<LayoutTypeName, LayoutTypeDetai
 
 export const isIssueLayoutResponse = createSchemeGuard<IssueLayoutResponse>(ISSUE_LAYOUT_RESPONSE_SCHEME)
 const isLayoutConfigItem = createSchemeGuard<LayoutConfigItem>(ISSUE_LAYOUT_CONFIG_ITEM_SCHEME)
+const isErrorResponse = createSchemeGuard<errorResponse>(ERROR_RESPONSE_SCHEME)
 
 export const getLayoutResponse = async ({
   variables,
@@ -234,7 +237,7 @@ export const fetchRequestTypeDetails = async ({
 
   const { layoutType: issueLayoutType } = createLayoutType(typeName)
   elements.push(issueLayoutType)
-
+  const requestsTypeToRemove: string[] = []
   const layouts = (
     await Promise.all(
       Object.entries(requestTypeIdToRequestType).flatMap(async ([requestTypeId, requestTypeInstance]) => {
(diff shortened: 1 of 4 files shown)


packages/jira-adapter/src/filters/layouts/request_types_layouts_to_values.ts
```diff
@@ -57,7 +57,7 @@ const filter: FilterCreator = ({ config }) => ({
     const requestFormType = objectTypes.find(e => e.elemID.typeName === REQUEST_FORM_TYPE)
     const issueViewType = objectTypes.find(e => e.elemID.typeName === ISSUE_VIEW_TYPE)
     if (requestTypeType === undefined || requestFormType === undefined || issueViewType === undefined) {
-      log.warn('Feild to find requestTypeType or requestFormType or issueViewType')
+      log.warn('Faild to find requestTypeType or requestFormType or issueViewType')
```
IdoZakk, line 60: Nice catch, this is still wrong :-)

IdoZakk (commented):
As discussed please add the error message to the condition for removal
I will do a quick approval afterwards

IdoZakk (approved):
Please add noise reduction

packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
```diff
@@ -249,6 +252,19 @@ export const fetchRequestTypeDetails = async ({
           client,
           typeName,
         })
+        if (
+          Array.isArray(response.errors) &&
+          isErrorResponse(response.errors[0]) &&
+          response.errors[0].extensions.statusCode === 404 &&
+          response.errors[0].message ===
```
shayc331, line 259: do we have to rely on the error message? I think it is a bit risky because they may change it in the future, is there a scenario when we get 404 and dont want to remove it?

shayc331, line 259: do we have to rely on the error message? I think it is a bit risky because they may change it in the future, is there a scenario when we get 404 and dont want to remove it?

packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
```diff
@@ -249,6 +252,19 @@ export const fetchRequestTypeDetails = async ({
           client,
           typeName,
         })
+        if (
+          Array.isArray(response.errors) &&
+          isErrorResponse(response.errors[0]) &&
```
shayc331, line 257: does it have to be in the first element in the errors list? why not go over the list?

packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
```diff
@@ -267,4 +283,6 @@ export const fetchRequestTypeDetails = async ({
   await addAnnotationRecursively(issueLayoutType, CORE_ANNOTATIONS.CREATABLE)
   await addAnnotationRecursively(issueLayoutType, CORE_ANNOTATIONS.UPDATABLE)
   await addAnnotationRecursively(issueLayoutType, CORE_ANNOTATIONS.DELETABLE)
+  // Remove request types that doesn't exist or user does not have required permissions to their layouts
+  _.remove(elements, e => requestsTypeToRemove.includes(e.elemID.getFullName()) && isInstanceElement(e))
```
shayc331, line 287: why do we need to remove them if we returned undefined?
nit: the second condition is less complicated so maybe do it first

shayc331 (commented):
looking great 🏀 
had a few comments

PR #5840, opened 2024-04-30
SALTO-5865: Fix jsm permissions change validator 

idoamit8:
In this PR I fixed JSM permission change validator

---

When a user has enableJSM set to true, but no JSM enabled in the service. he is failing in jsm_permissions CV since he is trying to access an endpoint that is JSM related. We should check if JSM is enabled in the CV before we check separately for each project. 

---
_Release Notes_: 
_Jira Adapter:_
* The bug where deploy fails when a user has `enableJSM = true` but no JSM in the service is solved. 

---
_User Notifications_: 
None


IdoZakk (approved):
Approving assuming the change in the last test we discussed

PR #5842, opened 2024-04-30
SALTO-5769 - Salesforce: Log when unable to map deploy messages to ElemIDs

tomsellek:
When we receive a deploy error we can't map to an ElemID, let's log some context to help us repro and debug.

---

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/metadata_deploy.ts
+++ b/packages/salesforce-adapter/src/metadata_deploy.ts
@@ -279,6 +279,11 @@ const processDeployResponse = (
   }: DeployMessage): ElemID | undefined => {
     const rawElemId = typeAndNameToElemId[componentType]?.[fullName]
     if (rawElemId === undefined) {
+      log.debug(
+        'Unable to match deploy message for %s[%s] with an ElemID.',
+        fullName,
+        componentType,
+      )
       return undefined
     }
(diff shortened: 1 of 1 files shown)


packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -320,6 +325,13 @@ const processDeployResponse = (
       severity: 'Error' as SeverityLevel,
     }))
 
+  if (failedComponentErrors.some((error) => error.elemID === undefined)) {
+    log.debug(
+      'Some deploy messages could not be mapped to an ElemID. typeAndNameToElemId=%s',
```
amitgoldarb, line 330: Is this potentially a large map?
Also should this be %o
```suggestion
      'Some deploy messages could not be mapped to an ElemID. typeAndNameToElemId=%o',
```

amitgoldarb (approved):
Approving to save cycles. If this is a huge map, maybe we should slice it somehow? Or place this in trace?

tomsellek:
> Approving to save cycles. If this is a huge map, maybe we should slice it somehow? Or place this in trace?

I don't think the additional code for slicing this object is worth it just for logging purposes, so I'll just make it a `trace`.

PR #5844, opened 2024-04-30
SALTO-5843 Google ws - support default domain for group deploy

Uri-Bar:
SALTO-5843 Google ws - support default domain for group deploy
---

---
_Release Notes_: 
none

---
_User Notifications_: 
none


packages/google-workspace-adapter/src/fix_elements/replace_groups_domain.ts
```diff
@@ -0,0 +1,87 @@
...
+    const groups = elements.filter(isInstanceElement).filter(e => e.elemID.typeName === GROUP_TYPE_NAME)
+    const domains = await awu(await elementsSource.getAll())
+      .filter(isInstanceElement)
+      .filter(e => e.elemID.typeName === DOMAIN_TYPE_NAME)
+      .toArray()
+    if (groups.length === 0) {
+      return { errors: [], fixedElements: [] }
+    }
+    const groupsWithNoDomain = groups.filter(group => !isDomainExist(group, domains))
+    if (defaultDomain === '') {
```
netama, line 72: continuing from above - should this include the undefined case? is there a reason to allow empty vs requiring either undefined or a non-empty value (can add restrictions on the field so that users will get a warning)?

packages/google-workspace-adapter/src/fix_elements/replace_groups_domain.ts
```diff
@@ -0,0 +1,87 @@
...
+const { awu } = collections.asynciterable
+
+export const domainNotExistError = (group: InstanceElement): ChangeError => ({
+  elemID: group.elemID,
+  severity: 'Error',
+  message: 'Group Domain does not exist in the target environment',
+  detailedMessage: `The domain of the group ${group.value.name} does not exist in the target environment. You can manually update the domain of the group to an existing domain or set the default domain in the deploy defaultDomain configuration to ${DEFAULT_PRIMARY_DOMAIN} to use the primary domain.`,
+})
+
+export const noPrimaryError = (group: InstanceElement): ChangeError => ({
```
netama, line 33: nit: in general there's a preference not to export just for the sake of unittest

packages/adapter-components/src/adapter-wrapper/adapter/adapter.ts
```diff
@@ -95,6 +96,7 @@ export class AdapterImpl<
   protected changeValidators: Record<string, ChangeValidator>
   protected dependencyChangers: DependencyChanger[]
   protected definitions: RequiredDefinitions<Options>
+  protected fixElementsFunc: FixElementsFunc
```
netama, line 99: I think we should also support here the full `WeakReferencesHandler` (which has a `findWeakReferences` part as well). if we don't need it in this flow we can start with just the fixer - but in general when both are needed they should come in pairs like we have in the existing adapters, so this API will change

packages/google-workspace-adapter/src/fix_elements/replace_groups_domain.ts
```diff
@@ -0,0 +1,87 @@
...
+import { DEFAULT_PRIMARY_DOMAIN, UserConfig } from '../config'
+import { Options } from '../definitions/types'
+import { DOMAIN_TYPE_NAME, GROUP_TYPE_NAME } from '../constants'
+
+const { awu } = collections.asynciterable
+
+export const domainNotExistError = (group: InstanceElement): ChangeError => ({
+  elemID: group.elemID,
+  severity: 'Error',
+  message: 'Group Domain does not exist in the target environment',
```
netama, line 29: tiny nit: 
```suggestion
  message: 'Group domain does not exist in the target environment',
```

packages/google-workspace-adapter/src/fix_elements/replace_groups_domain.ts
```diff
@@ -0,0 +1,87 @@
...
+  clone.value.email = `${emailPrefix}@${domain}`
+  return clone
+}
+
+export const replaceGroupsDomainHandler: FixElementsHandler<Options, UserConfig> =
+  ({ config, elementsSource }) =>
+  async elements => {
+    const defaultDomain = config.deploy?.defaultDomain
+    const groups = elements.filter(isInstanceElement).filter(e => e.elemID.typeName === GROUP_TYPE_NAME)
+    const domains = await awu(await elementsSource.getAll())
```
netama, line 64: somewhat of an edge case, but I think we should also return early if domains were excluded in the fetch config (maybe with a warning, but can also just log)

packages/google-workspace-adapter/src/fix_elements/replace_groups_domain.ts
```diff
@@ -0,0 +1,87 @@
...
+    const groupsWithNoDomain = groups.filter(group => !isDomainExist(group, domains))
+    if (defaultDomain === '') {
+      const nonExistErrors = groupsWithNoDomain.map(group => domainNotExistError(group))
+      return { errors: nonExistErrors, fixedElements: [] }
+    }
+    const primaryDomain = domains.find(domain => domain.value.isPrimary === true)
+    if (defaultDomain === DEFAULT_PRIMARY_DOMAIN && !primaryDomain) {
+      return { errors: groupsWithNoDomain.map(noPrimaryError), fixedElements: [] }
+    }
+    const domainReplacement = defaultDomain === DEFAULT_PRIMARY_DOMAIN ? primaryDomain?.value.domainName : defaultDomain
```
netama, line 80: not critical at all, but if the config uses a non-default domain we don't validate it - is this intentional? (I think it's an acceptable behavior, just making sure it's intentional)

netama (approved):
looks good 💪 added a bunch of comments but most are nits / related to fallbacks, so pre-approving. if there are big changes lmk and I'll review again

netama:
a couple more things:
1. please add release notes for GW
2. (not new but related to this) I think we should consider splitting the username and the domain in groups into different fields, and converting the domain to a reference. otherwise we will always see differences in env compare (might be easier to discuss IRL)

packages/adapter-components/src/adapter-wrapper/adapter/adapter.ts
```diff
@@ -351,4 +356,6 @@ export class AdapterImpl<
       changeValidator,
     }
   }
+
+  fixElements: FixElementsFunc = elements => this.fixElementsFunc(elements)
```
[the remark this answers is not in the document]
Uri-Bar: I changed it, but I it doesnt seems clean to me, I would like you to take a look

packages/google-workspace-adapter/src/fix_elements/replace_groups_domain.ts
```diff
@@ -0,0 +1,87 @@
...
+import { Options } from '../definitions/types'
+import { DOMAIN_TYPE_NAME, GROUP_TYPE_NAME } from '../constants'
+
+const { awu } = collections.asynciterable
+
+export const domainNotExistError = (group: InstanceElement): ChangeError => ({
+  elemID: group.elemID,
+  severity: 'Error',
+  message: 'Group Domain does not exist in the target environment',
+  detailedMessage: `The domain of the group ${group.value.name} does not exist in the target environment. You can manually update the domain of the group to an existing domain or set the default domain in the deploy defaultDomain configuration to ${DEFAULT_PRIMARY_DOMAIN} to use the primary domain.`,
```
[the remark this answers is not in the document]
Uri-Bar: will move it to the CV

packages/google-workspace-adapter/src/fix_elements/replace_groups_domain.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export const replaceDomainInfo = (group: InstanceElement, domain: string): ChangeError => ({
+  elemID: group.elemID,
+  severity: 'Info',
+  message: `replaced to domain of the ${group.value.name} group to ${domain}`,
+  detailedMessage: `replaced to domain of the ${group.value.name} group to ${domain}`,
+})
+
+const isDomainExist = (group: InstanceElement, domains: InstanceElement[]): boolean => {
+  const domainName = group.value.email.split('@')[1]
```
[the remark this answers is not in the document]
Uri-Bar: did yo mean I need to check if group.value.email is not empty or domainName? 
In most CVs, we assume some instance structure and if domainName is empty/undefined the `isDomainExist` will return false as expected

packages/google-workspace-adapter/src/fix_elements/replace_groups_domain.ts
```diff
@@ -0,0 +1,87 @@
...
+  const clone = group.clone()
+  const emailPrefix = clone.value.email.split('@')[0]
+  clone.value.email = `${emailPrefix}@${domain}`
+  return clone
+}
+
+export const replaceGroupsDomainHandler: FixElementsHandler<Options, UserConfig> =
+  ({ config, elementsSource }) =>
+  async elements => {
+    const defaultDomain = config.deploy?.defaultDomain
```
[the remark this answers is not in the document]
Uri-Bar: because I added the defaultDomain into the default CONFIG it was never undefined. I removed it from there and added handle for this.

packages/google-workspace-adapter/src/fix_elements/replace_groups_domain.ts
```diff
@@ -0,0 +1,87 @@
...
+    const groups = elements.filter(isInstanceElement).filter(e => e.elemID.typeName === GROUP_TYPE_NAME)
+    const domains = await awu(await elementsSource.getAll())
+      .filter(isInstanceElement)
+      .filter(e => e.elemID.typeName === DOMAIN_TYPE_NAME)
+      .toArray()
+    if (groups.length === 0) {
+      return { errors: [], fixedElements: [] }
+    }
+    const groupsWithNoDomain = groups.filter(group => !isDomainExist(group, domains))
```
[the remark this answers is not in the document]
Uri-Bar: I moved this part into the CV

packages/google-workspace-adapter/src/fix_elements/replace_groups_domain.ts
Uri-Bar: answered above

packages/google-workspace-adapter/src/fix_elements/replace_groups_domain.ts
Uri-Bar: Im not sure I agree, if someone didnt want to fetch domain but is managing groups and set defaultDomain='legit.com' I would still want the fixer to run and change the group nacl

packages/google-workspace-adapter/src/fix_elements/replace_groups_domain.ts
Uri-Bar: this is intentional and now we will have the CV to take care of that

packages/google-workspace-adapter/src/change_validators/group_domain.ts
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * This CV warns if a group email is in a domain that does not exist in the deployed environment.
+ */
+export const groupDomainValidator =
+  (config: UserConfig): ChangeValidator =>
+  async (changes, elementSource) => {
+    if (elementSource === undefined) {
+      return []
+    }
+    const isDomainFetched = config.fetch.exclude.some(exclude => exclude.type === DOMAIN_TYPE_NAME)
```
netama, line 55: 1. this is not accurate, both include and exclude use regular expressions so this would only catch the specific case where all types are included and domain is excluded specifically. 
you can use `isTypeMatch` like we use during the fetch to determine the types to fetch (see `createTypeResourceFetcher`).
(lmk if this is not clear, the include/exclude config is pretty central so it's worth understanding)
2. isn't the name reversed? why is it true if domains are _not_ fetched?

packages/google-workspace-adapter/src/change_validators/group_domain.ts
```diff
@@ -0,0 +1,68 @@
...
+    const domains = await awu(await elementSource.getAll())
+      .filter(isInstanceElement)
+      .filter(e => e.elemID.typeName === DOMAIN_TYPE_NAME)
+      .toArray()
```
netama, line 59: nit (I think continuing from last round but maybe different): no reason to make this call if domains were not fetched - maybe turn this into a function (`getDomains()`) and call it where relevant? this is loading _all_ elements in the element source so can be heavy

packages/google-workspace-adapter/src/fix_elements/replace_groups_domain.ts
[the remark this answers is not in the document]
netama: why is false the correct result? if there's nothing missing why should we replace it? (also we should probably log some error).
in most CVs we actually use type guards that check the structure, or have safeties :) here the expected structure is not just string but a string in a specific format

packages/google-workspace-adapter/src/fix_elements/replace_groups_domain.ts
```diff
@@ -0,0 +1,77 @@
...
+import { DEFAULT_PRIMARY_DOMAIN, UserConfig } from '../config'
+import { Options } from '../definitions/types'
+import { DOMAIN_TYPE_NAME, GROUP_TYPE_NAME } from '../constants'
+
+const { awu } = collections.asynciterable
+
+const replaceDomainInfo = (group: InstanceElement, domain: string): ChangeError => ({
+  elemID: group.elemID,
+  severity: 'Info',
+  message: `Replaced to domain of the ${group.value.name} group to ${domain}.`,
```
netama, line 29: this doesn't look right :) maybe 
```suggestion
  message: `Replaced the domain of group ${group.value.name} with ${domain}.`,
```
or
```suggestion
  message: `Replaced domain ${<domain>} of group ${group.value.name} with ${domain}.`,
```

packages/google-workspace-adapter/src/fix_elements/replace_groups_domain.ts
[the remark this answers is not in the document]
netama: even if it's there by default, we shouldn't assume users won't edit it

packages/google-workspace-adapter/src/fix_elements/replace_groups_domain.ts
```diff
@@ -0,0 +1,77 @@
...
+export const replaceGroupsDomainHandler: FixElementsHandler<Options, UserConfig> =
+  ({ config, elementsSource }) =>
+  async elements => {
+    const defaultDomain = config.deploy?.defaultDomain
+    if (defaultDomain === undefined) {
+      return { errors: [], fixedElements: [] }
+    }
+    const groups = elements.filter(isInstanceElement).filter(e => e.elemID.typeName === GROUP_TYPE_NAME)
+    if (groups.length === 0) {
+      return { errors: [], fixedElements: [] }
```
netama, line 58: I think some debug logs about early returns will come in handy when you debug user issues. up to you

packages/google-workspace-adapter/src/fix_elements/replace_groups_domain.ts
netama: what would happen though? you would replace _all_ domains even when they do exist

netama (commented):
added some comments about things that I think are still logically not "correct", did a relatively quick pass so hope I didn't miss anything (but we can also fix later). will look at the typescript issue separately but not a blocker

packages/google-workspace-adapter/src/fix_elements/replace_groups_domain.ts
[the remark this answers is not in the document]
Uri-Bar: if there is nothing missing it wont return false, I do you want it to throw if the sturct is not as expected?

packages/google-workspace-adapter/src/fix_elements/replace_groups_domain.ts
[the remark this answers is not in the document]
Uri-Bar: agree, I removed it from the default. If I havent removed it from the default, users would try to remove it and from the nacl and still get the default when we merge it to get the merged config

packages/google-workspace-adapter/src/fix_elements/replace_groups_domain.ts
Uri-Bar: if a user didnt fetch domains and he have defultDomain set to something, and all his groups are "legit", I would still change them al the CV will **warn** the user that we dont know if the fixed domain exists is the service

PR #5849, opened 2024-05-01
SALTO-5878 when resolving values, give precedence for the already-resolved value if available

netama:
this can fix a regression introduced in zendesk by https://github.com/salto-io/salto/pull/5733/files#diff-89893d1af4da92a0633bf2148169be4c5bfbc87b840ef9a7eb4e82ebdc75682cR825 , that caused the ducktype call to [replaceInstanceTypeForDeploy](https://github.com/salto-io/salto/blob/main/packages/zendesk-adapter/src/adapter.ts#L808) to have no effect during the [resolveChangeElement](https://github.com/salto-io/salto/blob/main/packages/zendesk-adapter/src/adapter.ts#L820) call.

~I'm not 100% about the reasons for giving the old precedence, so this is still WIP. we can also solve this by adding a variant of `resolveValues` that will use the element source for the values but not for the types, but we prefer not to use it if we can avoid it since it should be ok to pass an element source in these cases as well.~
cc @edenhassid 

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/core/src/core/fetch.ts
+++ b/packages/core/src/core/fetch.ts
@@ -36,7 +36,6 @@ import {
   isAdditionOrModificationChange,
   isElement,
   isEqualElements,
-  isField,
   isInstanceElement,
   isModificationChange,
   isObjectType,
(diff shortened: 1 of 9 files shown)


packages/jira-adapter/src/filters/fields/field_structure_filter.ts
```diff
@@ -243,7 +243,7 @@ const filter: FilterCreator = ({ config, getElemIdFunc }) => ({
       new ListType(BuiltinTypes.STRING),
     )
     fieldContextType.fields.defaultValue = new Field(fieldContextType, 'defaultValue', fieldContextDefaultValueType)
-    fieldContextType.fields.options = new Field(fieldType, 'options', new MapType(fieldContextOptionType))
+    fieldContextType.fields.options = new Field(fieldContextType, 'options', new MapType(fieldContextOptionType))
```
netama, line 246: this was a bug discovered by the safety log

packages/core/src/core/fetch.ts
```diff
@@ -470,6 +473,34 @@ const runPostFetch = async ({
...
+const updateInconsistentTypes = (validAccountElements: Element[]): void => {
+  const objectTypesByElemID = _.keyBy(validAccountElements.filter(isObjectType), e => e.elemID.getFullName())
+  const isInstanceOrField = (e: Element): e is InstanceElement | Field => isInstanceElement(e) || isField(e)
+  const elementsWithInconsistentTypes = validAccountElements
+    .filter(isInstanceOrField)
+    .filter(
+      e =>
+        e.refType.type !== undefined &&
+        objectTypesByElemID[e.refType.elemID.getFullName()] !== undefined &&
+        !isEqualElements(e.refType.type, objectTypesByElemID[e.refType.elemID.getFullName()]),
```
netama, line 486: this comparison is not completely flat, but it stops at the field level (see the `.isEqual()` implementations) so I hope it won't be too bad. added the `else` log so we can see if it slows us down too much and remove this more quickly

packages/core/src/core/fetch.ts
```diff
@@ -470,6 +473,34 @@ const runPostFetch = async ({
...
+    )
+
+  if (elementsWithInconsistentTypes.length > 0) {
+    log.error(
```
ori-moisis, line 495: IIUC changing the ref type below to be unresolved "solves" the issue and allows the code to continue properly.
if that is correct, then this shouldn't be an error log, it should be a warning

packages/core/src/core/fetch.ts
ori-moisis: I think it would be easier to monitor if you add a log.timeDebug on the entire function (updateInconsistentTypes) instead.

ori-moisis (approved):

PR #5850, opened 2024-05-01
SALTO-5872: add new useElementSourceForTypes flag for resolveValues func

edenhassid:
add new useElementSourceForTypes flag for resolveValues func

---

---
_Release Notes_: 
Zendesk adapter:
* Fix issues on deploy of some values containing reference expressions

---
_User Notifications_: 


netama (approved):
LGTM
not a blocker for the fix but may be worth adding a UT later, and potentially also converting to named args if we keep this

PR #5852, opened 2024-05-01
Additions for tickets SALTO-4142, SALTO-5537

SeggevHaimovich:
_Found a few issues related to things handled in the tickets:
SALTO-4142 - We created a CV that prevents the users from deploying a custom record type without permissions while its access type is defined as `USEPERMISSIONLIST`, we discovered we only need to prevent the deployment on addition change
SALTO-5537 - On the fix element we deleted all the unresolved references but we need to delete all the places with service id info as well._

---

Jira tickets:
[SALTO-4142](https://salto-io.atlassian.net/browse/SALTO-4142)
[SALTO-5537](https://salto-io.atlassian.net/browse/SALTO-5537)

example of service id info:
image

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


kami-delaroz (approved):
🥇

PR #5854, opened 2024-05-02
SALTO-954 SALTO-5558 allow multiple pointers to static file

benisraeladi:
Do not delete static file unless all pointers to it are removed

---

Please think about how I can create a test for this, because this is very internal

To check this, I ran the following scenario in Zendesk with and without the change:

1. add an article named A with an attachment
2. fetch
3. add an article with the same attachment called B
4. fetch
5. change the name of the article from A to B
6. fetch
7. Make sure that A and B point to the same static file
8. delete the original B

Without the new code, the result is a workspace error with a missing static file.
With the code, this goes smoothly and the static file is preserved

---
_Release Notes_: 
Workspace:
* Allow multiple pointers to the same static file

---
_User Notifications_: _None_

### DIFF
--- a/packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
+++ b/packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
@@ -851,7 +851,9 @@ const buildNaclFilesSource = (
     const removeDanglingStaticFiles = async (allChanges: DetailedChange[]): Promise<void> => {
       const { staticFilesIndex } = await getState()
 
-      await Promise.all(getDanglingStaticFiles(allChanges, staticFilesIndex).map(file => staticFilesSource.delete(file)))
+      await Promise.all(
+        getDanglingStaticFiles(allChanges, staticFilesIndex).map(file => staticFilesSource.delete(file)),
+      )
     }
     const changesByFileName = await groupChangesByFilename(changes)
     log.debug(
--- a/packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
+++ b/packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
@@ -846,10 +846,16 @@ const buildNaclFilesSource = (
 
     const removeDanglingStaticFiles = async (allChanges: DetailedChange[]): Promise<void> => {
       const {staticFilesIndex} = await getState()
+
       await Promise.all(
         getDanglingStaticFiles(allChanges).map(
-          file => staticFilesIndex.get(file.filepath) ?? staticFilesSource.delete(file),
-        ),
+          async file => {
+            const filePath = await staticFilesIndex.get(file.filepath)
+            if (!filePath) {
+              await staticFilesSource.delete(file)
+            }
+          }
+        )
       )
     }
     const changesByFileName = await groupChangesByFilename(changes)
--- a/packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
+++ b/packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
@@ -583,10 +583,13 @@ const logNaclFileUpdateErrorContext = (
   log.debug('data after:\n%s', naclDataAfter)
 }
 
-// Returns a list of all static files that existed in the changes 'before' and doesn't exist in the 'after'
-export const getDanglingStaticFiles = (fileChanges: DetailedChange[]): StaticFile[] => {
-  // Using filepath is currently enough because all implementations of static files have unique file paths
-  // The only exception is 'buildHistoryStateStaticFilesSource' but it doesn't support deletion at the moment
+export const getDanglingStaticFiles = (
+  fileChanges: DetailedChange[],
+  staticFileIndex?: RemoteMap<string[]>,
+): StaticFile[] => {
+  // Returns a list of all static files that existed in the changes 'before' and doesn't exist in the 'after'
+  // If staticFileIndex is defined, it also checks whether there are any other elements that still point to
+  //  the specific file
   const afterFilePaths = new Set<string>(
     fileChanges
       .filter(isAdditionOrModificationChange)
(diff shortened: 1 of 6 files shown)


packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
```diff
@@ -844,10 +844,17 @@ const buildNaclFilesSource = (
...
     }
 
-    // This method was written with the assumption that each static file is pointed by no more
-    // then one value in the nacls. A ticket was open to fix that (SALTO-954)
     const removeDanglingStaticFiles = async (allChanges: DetailedChange[]): Promise<void> => {
-      await Promise.all(getDanglingStaticFiles(allChanges).map(file => staticFilesSource.delete(file)))
+      const { staticFilesIndex } = await getState()
+
+      await Promise.all(
+        getDanglingStaticFiles(allChanges).map(async file => {
```
ori-moisis, line 851: isn't it a bug to let `getDanglingStaticFiles` return things that are not actually dangling static files?
I assume it is used in another place...

benisraeladi: The one other place this is used in in `workspace/state/in_mem.ts`. The problem is that we can't reach `staticFilesIndex` from there, so I can't do the same logic there, although I would say that this would be the correct thing to do

packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
```diff
@@ -599,6 +602,7 @@ export const getDanglingStaticFiles = (fileChanges: DetailedChange[]): StaticFil
     .map(change => change.data.before)
     .flatMap(getNestedStaticFiles)
     .filter(file => !afterFilePaths.has(file.filepath))
+    .filter(async file => !staticFileIndex?.has(file.filepath))
```
ori-moisis, line 605: async function on a sync filter is like returning `true` (because a Promise is considered `true`)
so, in practice this line doesn't do anything

packages/workspace/src/workspace/workspace.ts
```diff
@@ -315,7 +315,7 @@ export type Workspace = {
     isTemplate?: boolean
   }): Promise<StaticFile | undefined>
   getStaticFilePathsByElemIds(elementIds: ElemID[], envName?: string): Promise<string[]>
-  getElemIdsByStaticFilePaths(filePaths?: Set<string>, envName?: string): Promise<Record<string, string>>
+  getElemIdsByStaticFilePaths(filePaths?: Set<string>, envName?: string): Promise<Record<string, string[]>>
```
ori-moisis, line 318: this is a breaking change to the API
we should not do it in a single commit, we should add a new function with the new signature, and keep the existing one (as a deprecated option) for a little while (we can probably remove it in a few days, we really just need to have one version with both interfaces)

packages/workspace/src/workspace/workspace.ts
```diff
@@ -1185,18 +1185,23 @@ export const loadWorkspace = async (
...
+          .filter(filename => filePaths === undefined || filePaths.has(filename))
+          .map(filename => [filename, id] as [string, string]),
+      )
+      .forEach(([filename, id]) => {
+        if (!result[filename]) {
+          result[filename] = []
+        }
+        result[filename].push(id)
+      })
+    return result
```
ori-moisis, line 1204: convention nit: prefer code without side effects
```suggestion
    const idsAndFileByFileName = await awu(currentWorkspaceState.states[env].referencedStaticFiles.entries())
      .flatMap(({ key: id, value: fileNames }) =>
        fileNames
          .filter(filename => filePaths === undefined || filePaths.has(filename))
          .map(filename => ({ filename, id })),
      )
      .groupBy(({ filename }) => filename)

    return _.mapValues(idsAndFileByFileName, idAndFiles => idAndFiles.map(({ id }) => id))
```

packages/workspace/test/workspace/workspace.test.ts
```diff
@@ -2588,6 +2594,20 @@ describe('workspace', () => {
         },
       })
     })
+    describe('multiple pointers to single static file', () => {
+      it('should not delete static file if there is an element that still points to it', async () => {
+        const changes: DetailedChange[] = [
+          {
+            id: new ElemID('salesforce', 'lead', 'instance', 'someName1', 'salesforce', 'text'),
+            action: 'remove',
+            data: { before: 'file("static2.nacl")' },
```
ori-moisis, line 2603: this is not how a static file remove change looks like.
I suggest you run a fetch in a debugger and see what the change actually looks like to create a correct test

packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
benisraeladi: The `await` is implied because it's a function and you can't write `await` on `return`

Also, this is what fixed the UTs (adding the `async`), so I'm surprised that you say that this line doesn't do anything

benisraeladi: @ori-moisis We have to add the `async` because `staticFileIndex.has` returns `Promise<boolean>`. Since (as you said) a promise is always true, if we don't await then the response is wrong

packages/workspace/src/workspace/workspace.ts
benisraeladi: This changed a unit test in the workspace. You can see the changes at the bottom of the diff view

packages/workspace/test/workspace/workspace.test.ts
benisraeladi: We have such tests in `nacl_file_source.test.ts`

ori-moisis: I know, hence the
> there are quicker ways to figure it out specifically for building the change, but I think it would be worth the extra time to actually recreate the full scenario

edenhassid:
closing as its no longer relevant

PR #5856, opened 2024-05-03
SALTO-5441: update error message for failing to get issueLayouts

izibr:
updating the error message the client gets when he tries to deploy change (add, remove, or modify) regarding issue layout and he gets a failed response from jira service.
it can happen because the issue layout was in the environment and there were changes in the environment and now the issue layout no longer exists in this environment and the problem was that is environment was not sync 
I changed the error message so it will explain this possibility

---

---
_Release Notes_: 
jira adapter:
- improved the error message for some failed issue layouts deployments
---
_User Notifications_: 
N/A


IdoZakk (approved):
🏆 
The release notes can be improved:
* we write 'jira adapter', it makes the deployer life easier
* The message is very long. It should not get into the exact details, but be more of a header. Something like 'improved the error message for some failed issue layouts deployments'

PR #5858, opened 2024-05-05
SALTO-5886: Salesforce Profiles with loginFlows deploy bugfix

tamtamirr:
Salesforce Profiles with loginFlows deploy bugfix

---

This field is required when deploying modifications to Profile instances, and should not be removed as part of the `minify_deploy.ts` filter

---
_Release Notes_: 
_Salesforce Adapter_:
- Fixed a bug where attempting to deploy modifications to a Profile with loginFlows failed.

---
_User Notifications_: 
_None_


amitgoldarb (approved):
Thanks for this!

PR #5866, opened 2024-05-06
SALTO-5799: Move fixElements and customReferences from Salesforce data config to the top level

tamtamirr:
These config values are currently misplaced under the data configuration.

---

_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/types.ts
```diff
@@ -131,6 +131,7 @@ export type OptionalFeatures = {
   extendedCustomFieldInformation?: boolean
   importantValues?: boolean
   hideTypesFolder?: boolean
+  omitStandardFieldsNonDeployableValues?: boolean
```
yelly, line 134: Why is this here?

packages/salesforce-adapter/src/custom_references/handlers.ts
yelly, line 1: You also need to change `customReferencesConfiguration` (on line 51).

yelly (changes requested):

yelly (approved):

PR #5869, opened 2024-05-07
SALTO-5553 update field context counting

izibr:
Change the counting of the options in field-context so that it will start from 1 instead of 0 to make it compatible with the jira UI
---

---
_Release Notes_: 

_Jira Adapter_:
- update the counting of the options in field-context 
_User Notifications_: 

N/A

### DIFF
--- a/packages/jira-adapter/e2e_test/instances/field.ts
+++ b/packages/jira-adapter/e2e_test/instances/field.ts
@@ -32,7 +32,7 @@ const p1Option = {
       position: 1,
     },
   },
-  position: 0,
+  position: 1,
 }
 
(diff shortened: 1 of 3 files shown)


IdoZakk:
Regarding the User notification comment: you broke the structure, it will cause problems to the person doing the bump.
Also, as you change the NaCls you should write about it in the user notifications

IdoZakk (commented):
Awesome! 🎸 
Please fix E2E. I will approve after it is fixed + the noise reduction is deployed

IdoZakk (commented):
Please see my previous comment about the PR comment

IdoZakk (approved):

PR #5872, opened 2024-05-08
SALTO-5980: Update Google Workspace README with reauthentication policy instructions

shoshana-houri:
We recently found out that the default reauthentication policy defined on GW accounts requires requthentication after 16 hours, which causes us to receive an error when trying to get an access token using only our stored refresh token.
This can be configured in the admin console in 2 ways. We added guidance for both in the GW README.

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


Uri-Bar (approved):
thanks!

PR #5873, opened 2024-05-08
SALTO-5555: errors in fetch JSM forms

idoamit8:
In this PR I fixed the multiple errors in JSM Forms. 

---

Currently, we are logging many errors related to failures in fetching JSM forms.
|Was|Now|
|----|-----| 
|Previously, when we failed to fetch forms for an entire project, we logged an error for each project in the account.| Now, we log a single error for all projects.|
|When forms had missing titles, we logged an error for each form. |Now, we log one error covering all forms and projects.|
  
Example in oriSaltoTest:
image

In addition, I decreased the logs from Error to debug and added fetch warnings.  
#### Important Note:
**I will add tests after the initial review to ensure that my solution is acceptable.**

---
_Release Notes_: 
_Jira adapter_:
* Optimized error logging for JSM form fetching to reduce noise: Now logging a single error per account for project-wide fetch failures and a unified error for all forms with missing titles.

---
_User Notifications_: 
_Jira adapter_:
* Optimized error logging for JSM form fetching to reduce noise: Now logging a single error per account for project-wide fetch failures and a unified error for all forms with missing titles.

### DIFF
--- a/packages/jira-adapter/src/filters/forms/forms.ts
+++ b/packages/jira-adapter/src/filters/forms/forms.ts
@@ -136,10 +136,6 @@ const filter: FilterCreator = ({ config, client, fetchQuery }) => ({
               log.error(
                 `Failed to fetch forms for project ${project.value.name} with the following response: ${inspectValue(res)}`,
               )
-              errors.push({
-                message: `Failed to fetch forms for project ${project.value.name}`,
-                severity: 'Error' as SeverityLevel,
-              })
             }
             return undefined
           }
--- a/packages/jira-adapter/src/filters/forms/forms.ts
+++ b/packages/jira-adapter/src/filters/forms/forms.ts
@@ -30,6 +30,7 @@ import {
 import { values as lowerDashValues } from '@salto-io/lowerdash'
 import { getParent, invertNaclCase, mapKeysRecursive, naclCase, pathNaclCase } from '@salto-io/adapter-utils'
 import _ from 'lodash'
+import { logger } from '@salto-io/logging'
 import { resolveValues } from '@salto-io/adapter-components'
 import { FilterCreator } from '../../filter'
 import { FORM_TYPE, JSM_DUCKTYPE_API_DEFINITIONS, PROJECT_TYPE, SERVICE_DESK } from '../../constants'
@@ -41,7 +42,7 @@ import { setTypeDeploymentAnnotations, addAnnotationRecursively } from '../../ut
 import { getLookUpName } from '../../reference_mapping'
 
 const { isDefined } = lowerDashValues
-
+const log = logger(module)
 const deployForms = async (change: Change<InstanceElement>, client: JiraClient): Promise<void> => {
   const form = getChangeData(change)
   const project = getParent(form)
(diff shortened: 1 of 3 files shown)


packages/jira-adapter/src/filters/forms/forms.ts
```diff
@@ -156,6 +156,23 @@ const filter: FilterCreator = ({ config, client, fetchQuery }) => ({
       form.value = mapKeysRecursive(form.value, ({ key }) => naclCase(key))
       elements.push(form)
     })
+    if (projectsWithOutForms.length > 0) {
+      const message = `Unable to fetch forms for the following projects: ${projectsWithOutForms.join(', ')}. This issue is likely due to insufficient permissions.`
```
idoamit8, line 160: @liorn , how does this message look? It will be shown to the user at the end of the fetch process if there are any projects for which forms could not be fetched.

packages/jira-adapter/src/filters/forms/forms.ts
```diff
@@ -156,6 +156,23 @@ const filter: FilterCreator = ({ config, client, fetchQuery }) => ({
...
+      log.debug(message)
+      errors.push({
+        message,
+        severity: 'Warning' as SeverityLevel,
+      })
+    }
+    const projectsWithUntitledFormsArr = Array.from(projectsWithUntitledForms)
+    if (projectsWithUntitledFormsArr.length > 0) {
+      const message = `Unable to fetch untitled forms for the following projects: ${projectsWithUntitledFormsArr.join(', ')}. Please provide a title for them or delete them.`
```
idoamit8, line 169: @liorn , how does this message look? It will be shown to the user at the end of the fetch process if there are any projects with untitled forms.

packages/jira-adapter/src/filters/forms/forms.ts
```diff
@@ -109,24 +110,23 @@ const filter: FilterCreator = ({ config, client, fetchQuery }) => ({
...
     const errors: SaltoError[] = []
+    const projectsWithOutForms: string[] = []
+    const projectsWithUntitledForms: Set<string> = new Set()
     const forms = (
       await Promise.all(
         jsmProjects.flatMap(async project => {
           const url = `/gateway/api/proforma/cloudid/${cloudId}/api/1/projects/${project.value.id}/forms`
           const res = await client.get({ url })
           if (!isFormsResponse(res)) {
+            projectsWithOutForms.push(project.elemID.name)
```
IdoZakk, line 121: As the schema guard can fail due to structure we do not understand I think we might want to log in trace level the response

packages/jira-adapter/src/filters/forms/forms.ts
IdoZakk, line 160: From what I remember we mentioned the permissions for the detailed forms calls. Is there a different permission issue also here? If so, can we be more specific when we issue this message?

IdoZakk (commented):
⭐ ⭐ ⭐ ⭐ ⭐ 
Thank you for the very detailed commit comment. It made the review much easier.
I think the general concept is great, see my comments

packages/jira-adapter/src/filters/forms/forms.ts
liorn: If it gives the projet names (and not weird salto element IDs), LGTM

packages/jira-adapter/src/filters/forms/forms.ts
liorn: I would say something simpler; `Salto does not support fetching untitled forms, found in the following projects: ${projectsWithUntitledFormsArr.join(', ')}`

packages/jira-adapter/src/filters/forms/forms.ts
idoamit8: When a user lacking the necessary permissions attempts to access the forms via this endpoint, they receive the following response:
`{
data: [],
status 200
}`
This response is not very informative. Based on my experimentation with OriSaltoTest, it appears that modifying the permissions enables access to the forms. However, access to detailed forms is unsuccessful due to the absence of a title.

IdoZakk: So maybe limit this error to when we get this exact message, and issue a regular error message (including to the logs) when we get other bad formats. 
The problem I think about is that in several cases we saw very different behaviors for customers envs

IdoZakk (approved):
🚀 
I am approving, please look at the comments and most importantly the one about different types of errors

PR #5877, opened 2024-05-08
SALTO-5643: Adding members in the order they are defined

yelly:
Instead of adding new fields, annotations or values at the end of their parent in NaCl files, add them in the order they are defined in their parent object. This will ensure that members always show up in the same place in the NaCl if they are removed and added again.

---

Still need to add higher level tests, IMO it's cleaner to split into a separate PR.

---
_Release Notes_: 
_Core_
* Element members (field, annotations or values) are added in the same location consistently instead of being appended.

---
_User Notifications_: 
None.

### DIFF
--- a/packages/adapter-api/src/element_id.ts
+++ b/packages/adapter-api/src/element_id.ts
@@ -258,14 +258,6 @@ export class ElemID {
     return new ElemID(this.adapter, this.typeName)
   }
 
-  createSiblingID(name: string): ElemID {
-    if (this.isTopLevel()) {
-      throw new Error(`Cannot create sibling for top level element ID ${this.getFullName()}.`)
-    }
-
-    return new ElemID(this.adapter, this.typeName, this.idType, ...this.nameParts.slice(0, -1), name)
-  }
-
   createAllElemIdParents(): ElemID[] {
     return this.isTopLevel() ? [this] : [this, ...this.createParentID().createAllElemIdParents()]
   }
@@ -296,6 +288,14 @@ export class ElemID {
       }
     }
     return { parent, path }
+  }
+
+  createSiblingID(name: string): ElemID {
+    if (this.isTopLevel()) {
+      throw new Error(`Cannot create sibling for top level element ID ${this.getFullName()}.`)
+    }
+
+    return new ElemID(this.adapter, this.typeName, this.idType, ...this.nameParts.slice(0, -1), name)
   }
 
   getRelativePath(other: ElemID): ReadonlyArray<string> {
--- a/packages/adapter-api/src/element_id.ts
+++ b/packages/adapter-api/src/element_id.ts
@@ -258,6 +258,14 @@ export class ElemID {
     return new ElemID(this.adapter, this.typeName)
   }
 
+  createSiblingID(name: string): ElemID {
+    if (this.isTopLevel()) {
+      throw new Error(`Cannot create sibling for top level element ID ${this.getFullName()}.`)
+    }
+
+    return new ElemID(this.adapter, this.typeName, this.idType, ...this.nameParts.slice(0, -1), name)
+  }
+
   createAllElemIdParents(): ElemID[] {
     return this.isTopLevel() ? [this] : [this, ...this.createParentID().createAllElemIdParents()]
   }
--- a/packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
+++ b/packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
@@ -55,9 +55,52 @@ export type DetailedChangeWithSource = DetailedChange & {
 const createFileNameFromPath = (pathParts?: ReadonlyArray<string>): string =>
   `${path.join(...(pathParts ?? ['unsorted']))}${FILE_EXTENSION}`
 
-type positionInParent = {
+type PositionInParent = {
   followingElementIDs: ElemID[]
   indexInParent?: number
+}
+
+const getPositionInParent = (change: DetailedChange): PositionInParent => {
+  if (change.baseChange === undefined) {
+    log.warn('No base change: %s', inspectValue(change))
+    return { followingElementIDs: [] }
+  }
+
+  const changeData = getChangeData(change)
+  const parent = isField(changeData) ? changeData.parent : getChangeData(change.baseChange)
+  const pathInParent = getPath(parent, change.id)
+  if (pathInParent === undefined) {
+    log.warn('Could not get path for %s in parent: %s', change.id.getFullName(), inspectValue(parent))
+    return { followingElementIDs: [] }
+  }
+  if (pathInParent.length === 0) {
+    return { followingElementIDs: [] }
+  }
+
+  const container = _.get(parent, pathInParent.slice(0, -1))
+  if (!_.isObjectLike(container)) {
+    log.warn(
+      'Got non object container at path %s: %s',
+      change.id.createParentID().getFullName(),
+      inspectValue(container),
+    )
+    return { followingElementIDs: [] }
+  }
+
+  const idNameParts = change.id.getFullNameParts().slice(0, -1)
+  const elementName = change.id.name
+  const index = Object.keys(container).indexOf(elementName)
+  if (index === -1) {
+    log.warn('Element %s not found in container: %s', elementName, inspectValue(container))
+    return { followingElementIDs: [] }
+  }
+
+  return {
+    followingElementIDs: Object.keys(container)
+      .slice(index + 1)
+      .map(k => ElemID.fromFullNameParts([...idNameParts, k])),
+    indexInParent: index,
+  }
 }
 
 export const getChangeLocations = (
@@ -79,47 +122,6 @@ export const getChangeLocations = (
     }
   }
 
-  const getPositionInParent = (): positionInParent => {
-    if (change.baseChange === undefined) {
-      log.warn('No base change: %s', inspectValue(change))
-      return { followingElementIDs: [] }
-    }
-
-    const changeData = getChangeData(change)
-    const parent = isField(changeData) ? changeData.parent : getChangeData(change.baseChange)
-    const pathInParent = getPath(parent, change.id)
-    if (pathInParent === undefined) {
-      log.warn('Could not get path for %s in parent: %s', inspectValue(change.id), inspectValue(parent))
-      return { followingElementIDs: [] }
-    }
-
-    const container = _.get(parent, pathInParent.slice(0, -1))
-    if (!_.isObjectLike(container)) {
-      log.warn(
-        'Got non object container %s for path %s in parent: %s',
-        inspectValue(container),
-        inspectValue(change.id.createParentID()),
-        inspectValue(parent),
-      )
-      return { followingElementIDs: [] }
-    }
-
-    const idNameParts = change.id.getFullNameParts().slice(0, -1)
-    const elementName = change.id.name
-    const index = Object.keys(container).indexOf(elementName)
-    if (index === -1) {
-      log.warn('Element %s not found in container: %s', elementName, inspectValue(container))
-      return { followingElementIDs: [] }
-    }
-
-    return {
-      followingElementIDs: Object.keys(container)
-        .slice(index + 1)
-        .map(k => ElemID.fromFullNameParts([...idNameParts, k])),
-      indexInParent: index,
-    }
-  }
-
   const findLocations = (): { location: Location; requiresIndent?: boolean }[] => {
     if (change.action !== 'add') {
       // We want to get the location of the existing element
@@ -135,19 +137,18 @@ export const getChangeLocations = (
       }
     } else if (!change.id.isTopLevel()) {
       const fileName = createFileNameFromPath(change.path)
-      const { followingElementIDs, indexInParent } = getPositionInParent()
-      const possibleFollowingElementsRanges = followingElementIDs
-        .map(elemID => sourceMap.get(elemID.getFullName()))
+      const { followingElementIDs, indexInParent } = getPositionInParent(change)
+      const possibleFollowingElementsRange = followingElementIDs
+        .flatMap(elemID => sourceMap.get(elemID.getFullName()))
         .filter(isDefined)
-        .flat()
-        .filter(sr => sr.filename === fileName)
-      if (possibleFollowingElementsRanges.length > 0) {
+        .find(sr => sr.filename === fileName)
+      if (possibleFollowingElementsRange !== undefined) {
         return [
           {
             location: {
               filename: fileName,
-              start: possibleFollowingElementsRanges[0].start,
-              end: possibleFollowingElementsRanges[0].start,
+              start: possibleFollowingElementsRange.start,
+              end: possibleFollowingElementsRange.start,
               indexInParent,
             },
           },
@@ -166,6 +167,7 @@ export const getChangeLocations = (
         // TODO: figure out how to choose the correct location if there is more than one option
         return [{ location: lastNestedLocation(foundInPath ?? possibleLocations[0]), requiresIndent: true }]
       }
+      log.error('No possible locations found for %s.', parentID.getFullName())
     }
     // Fallback to using the path from the element itself
     const naclFilePath = change.path ?? getChangeData(change).path
@@ -341,10 +343,7 @@ export const updateNaclFileData = async (
 
           if (change.action === 'remove') {
             // For removals, dropping newline and indent at the end of the block to avoid empty lines
-            const lastNewlineIndex = data.search(/\s+$/)
-            if (lastNewlineIndex !== -1) {
-              data = data.slice(0, lastNewlineIndex)
-            }
+            data = data.trimEnd()
           }
 
           parts.push({
--- a/packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
+++ b/packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
@@ -33,7 +33,7 @@ import {
   isTypeReference,
 } from '@salto-io/adapter-api'
 import { AdditionDiff, ActionName } from '@salto-io/dag'
-import { inspectValue, resolvePath, walkOnElement, WalkOnFunc, WALK_NEXT_STEP } from '@salto-io/adapter-utils'
+import { inspectValue, getPath, walkOnElement, WalkOnFunc, WALK_NEXT_STEP } from '@salto-io/adapter-utils'
 import { collections, values } from '@salto-io/lowerdash'
 import { parser } from '@salto-io/parser'
 
@@ -87,8 +87,13 @@ export const getChangeLocations = (
 
     const changeData = getChangeData(change)
     const parent = isField(changeData) ? changeData.parent : getChangeData(change.baseChange)
+    const pathInParent = getPath(parent, change.id)
+    if (pathInParent === undefined) {
+      log.warn('Could not get path for %s in parent: %s', inspectValue(change.id), inspectValue(parent))
+      return { followingElementIDs: [] }
+    }
 
-    const container = resolvePath(parent, change.id.createParentID())
+    const container = _.get(parent, pathInParent.slice(0, -1))
     if (!_.isObjectLike(container)) {
       log.warn(
         'Got non object container %s for path %s in parent: %s',
@@ -99,7 +104,7 @@ export const getChangeLocations = (
       return { followingElementIDs: [] }
     }
 
-    const idNameParts = change.id.getFullNameParts()
+    const idNameParts = change.id.getFullNameParts().slice(0, -1)
     const elementName = change.id.name
     const index = Object.keys(container).indexOf(elementName)
     if (index === -1) {
--- a/packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
+++ b/packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
@@ -15,7 +15,9 @@
  */
 import _ from 'lodash'
 import path from 'path'
+import { logger } from '@salto-io/logging'
 import {
+  Field,
   getChangeData,
   isElement,
   ObjectType,
@@ -31,16 +33,23 @@ import {
   isTypeReference,
 } from '@salto-io/adapter-api'
 import { AdditionDiff, ActionName } from '@salto-io/dag'
-import { walkOnElement, WalkOnFunc, WALK_NEXT_STEP } from '@salto-io/adapter-utils'
-import { collections } from '@salto-io/lowerdash'
+import { getPath, walkOnElement, WalkOnFunc, WALK_NEXT_STEP } from '@salto-io/adapter-utils'
+import { collections, values } from '@salto-io/lowerdash'
 import { parser } from '@salto-io/parser'
 
 const { awu } = collections.asynciterable
+const { isDefined } = values
+
+const log = logger(module)
 
 // Declared again to prevent cyclic dependency
 const FILE_EXTENSION = '.nacl'
 
-export type DetailedChangeWithSource = DetailedChange & { location: parser.SourceRange }
+export type DetailedChangeWithSource = DetailedChange & {
+  location: parser.SourceRange
+  indexInParent?: number
+  requiresIndent?: boolean
+}
 
 const createFileNameFromPath = (pathParts?: ReadonlyArray<string>): string =>
   `${path.join(...(pathParts ?? ['unsorted']))}${FILE_EXTENSION}`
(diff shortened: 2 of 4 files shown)


packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -64,41 +73,107 @@ export const getChangeLocations = (
...
+    if (!(container instanceof Object)) {
+      log.warn(`Got non object container ${container} for path ${pathInParent.slice(0, -1)} in parent: ${parent}`)
+      return [[], undefined]
+    }
+
+    const idNameParts = change.id.getFullNameParts()
+    const elementName = idNameParts.pop() as string // Element can't have an empty ID
+    const index = Object.keys(container).indexOf(elementName)
+    if (index === -1) {
+      log.warn(`Element ${elementName} not found in container: ${container}`)
```
ori-moisis, line 100: nit - I tend to prefer string formatting over template strings (I do it because it allows the code to skip actually creating the log string if it is not going to be logged)

also, you shouldn't log element values directly.
elements values can contain reference expressions that could lead to a LOT of information that you do not want to log (makes to log potentially very large and hard to understand), we have a few overrides to make `inspect` output a more log friendly representation of values ([example](https://github.com/salto-io/salto/blob/main/packages/adapter-api/src/values.ts#L160))

```suggestion
      log.warn('Element %s not found in container: %s', elementName, inspectValue(container))
```

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -252,10 +330,7 @@ export const updateNaclFileData = async (
 
           if (change.action === 'remove') {
             // For removals, dropping newline and indent at the end of the block to avoid empty lines
-            const lastNewlineIndex = data.search(/\s+$/)
-            if (lastNewlineIndex !== -1) {
-              data = data.slice(0, lastNewlineIndex)
-            }
+            data = data.trimEnd()
```
ori-moisis, line 333: unrelated change?

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -191,18 +266,15 @@ export const updateNaclFileData = async (
...
   const toBufferChange = async (change: DetailedChangeWithSource): Promise<BufferChange> => {
     const elem = change.action === 'remove' ? undefined : change.data.after
     let newData: string
     let indentationLevel = (change.location.start.col - 1) / 2
-    /* When adding a nested change we need to increase one level of indentation because
-     * we get the placement of the closing brace of the next line. The closing brace will
-     * be indented one line less then wanted change.
-     */
-    if (change.action === 'add' && !change.id.isTopLevel()) {
+    if (change.requiresIndent) {
```
ori-moisis, line 277: you changed the logic that calculates this and now this happens in different conditions
why did you do that?

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -64,41 +73,107 @@ export const getChangeLocations = (
...
+
+    return [
+      Object.keys(container)
+        .slice(index + 1)
+        .map(k => ElemID.fromFullNameParts([...idNameParts, k])),
+      index,
+    ]
+  }
+
+  const findLocations = (): { location: parser.SourceRange; indexInParent?: number; requiresIndent?: boolean }[] => {
```
ori-moisis, line 112: nit - I would have put the `indexInParent` as part of the location (I'd create a local `type Location = parser.SourceRange & { indexInParent: number }`)
I think it would have made the the changes here less severe and kept the code a bit cleaner

ori-moisis (commented):
didn't finish reviewing the logic, but I think there are enough comments to do a round...

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -64,41 +73,107 @@ export const getChangeLocations = (
     }
   }
 
-  const findLocations = (): parser.SourceRange[] => {
+  const getFollowingElementsInParent = (): [ElemID[], number | undefined] => {
```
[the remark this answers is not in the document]
yelly: Done.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -64,41 +73,107 @@ export const getChangeLocations = (
...
+    if (change.baseChange === undefined) {
+      log.warn(`No base change: ${change}`)
+      return [[], undefined]
+    }
+
+    const changeData = getChangeData(change)
+    const parent = changeData instanceof Field ? changeData.parent : getChangeData(change.baseChange)
```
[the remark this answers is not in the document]
yelly: Done.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -64,41 +73,107 @@ export const getChangeLocations = (
...
+    const pathInParent = getPath(parent, change.id)
+    if (pathInParent === undefined) {
+      log.warn(`Could not get path for ${change.id} in parent: ${parent}`)
+      return [[], undefined]
+    }
+
+    const container = _.get(parent, pathInParent.slice(0, -1))
+    if (!(container instanceof Object)) {
```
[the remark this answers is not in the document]
yelly: How do you suggest we support arrays? We don't have an element ID per array element (or do we)?

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -64,41 +73,107 @@ export const getChangeLocations = (
...
+
+    const changeData = getChangeData(change)
+    const parent = changeData instanceof Field ? changeData.parent : getChangeData(change.baseChange)
+    const pathInParent = getPath(parent, change.id)
+    if (pathInParent === undefined) {
+      log.warn(`Could not get path for ${change.id} in parent: ${parent}`)
+      return [[], undefined]
+    }
+
+    const container = _.get(parent, pathInParent.slice(0, -1))
```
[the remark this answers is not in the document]
yelly: Done.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -64,41 +73,107 @@ export const getChangeLocations = (
...
+    if (!(container instanceof Object)) {
+      log.warn(`Got non object container ${container} for path ${pathInParent.slice(0, -1)} in parent: ${parent}`)
+      return [[], undefined]
+    }
+
+    const idNameParts = change.id.getFullNameParts()
+    const elementName = idNameParts.pop() as string // Element can't have an empty ID
```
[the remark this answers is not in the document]
yelly: Done.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
yelly: Done.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
yelly: Done.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
yelly: Yeah, should I do a separate PR for this?

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
yelly: Because I need to differentiate between additions at the end of the parent and in the middle. Until now this logic didn't support adding in the middle (which is also why I had to change the UT - it wasn't a good test because it compensated for this problem).

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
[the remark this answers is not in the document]
yelly: Actually I take it back - using `change.id.createParentID()` doesn't work because it omits the container.
For example, in the case of a field, the change ID is of the form `adapter.type.field.fieldName`, so for the container we need `adapter.type.field`, but `createParentID()` returns `adapter.type`.
We could add a bunch of logic to figure out the container name but that feels much worse than the current approach.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
ori-moisis: probably.
not because it is big or something, just because of the potential need to revert one thing without the other

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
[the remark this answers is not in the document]
ori-moisis: we do.
generally, you can test this out using the vscode extension - it adds a "copy element ID" option in the context menu - so you can right click on any part of the nacl and use that to know what the element ID is...

I am pretty sure the rest of the code would just work if you let arrays through (`Object.keys` returns the indices of an array IIRC)

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
yelly: Done.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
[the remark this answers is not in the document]
yelly: So it seems the current code already supports arrays without needing any changes. Added a unit test to validate this.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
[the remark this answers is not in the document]
ori-moisis: ok, I understand
also, looking again at the code I see the manipulations here (`.slice(0, -1)`) are done on the path, not the element ID so that should be ok

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -64,41 +79,112 @@ export const getChangeLocations = (
...
       if (possibleLocations.length > 0) {
-        const foundInPath = possibleLocations.find(sr => sr.filename === createFileNameFromPath(change.path))
+        const foundInPath = possibleLocations.find(sr => sr.filename === fileName)
+        /* When adding a nested change we need to increase one level of indentation because
+         * we get the placement of the closing brace of the next line. The closing brace will
+         * be indented one line less then wanted change.
+         */
         // TODO: figure out how to choose the correct location if there is more than one option
-        return [lastNestedLocation(foundInPath || possibleLocations[0])]
+        return [{ location: lastNestedLocation(foundInPath ?? possibleLocations[0]), requiresIndent: true }]
```
ori-moisis, line 167: maybe instead of doing all this with the `requiresIndent`, if this is really the only case where this is needed, it would be cleaner to just change the implementation of `lastNestedLocation` so it would return a more nested column?
that way there won't be a need to handle special cases and increase the complexity in multiple functions, it would just hide one edge case in the place that creates that edge case...

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -252,7 +341,10 @@ export const updateNaclFileData = async (
 
           if (change.action === 'remove') {
             // For removals, dropping newline and indent at the end of the block to avoid empty lines
-            data = data.trimEnd()
+            const lastNewlineIndex = data.search(/\s+$/)
+            if (lastNewlineIndex !== -1) {
+              data = data.slice(0, lastNewlineIndex)
+            }
```
ori-moisis, line 347: rebase mistake?

packages/workspace/test/workspace/nacl_files/nacl_file_update.test.ts
```diff
@@ -90,24 +101,283 @@ describe('getNestedStaticFiles', () => {
...
+      ])
+    })
+  })
+
+  describe('with addition of field', () => {
+    it('should add the field before a following field', () => {
+      const change: DetailedChange = {
+        ...toChange({ after: mockType.fields.numArray }),
+        id: mockType.fields.numArray.elemID,
+        baseChange: toChange({ after: mockType }),
```
ori-moisis, line 144: in this scenario the base change cannot be an addition of the whole type.
it should be a modification where this field was added...

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -64,41 +79,112 @@ export const getChangeLocations = (
...
-        const foundInPath = possibleLocations.find(sr => sr.filename === createFileNameFromPath(change.path))
+        const foundInPath = possibleLocations.find(sr => sr.filename === fileName)
+        /* When adding a nested change we need to increase one level of indentation because
+         * we get the placement of the closing brace of the next line. The closing brace will
+         * be indented one line less then wanted change.
+         */
         // TODO: figure out how to choose the correct location if there is more than one option
-        return [lastNestedLocation(foundInPath || possibleLocations[0])]
+        return [{ location: lastNestedLocation(foundInPath ?? possibleLocations[0]), requiresIndent: true }]
       }
```
ori-moisis, line 168: unrelated to your change, but if we get here and don't find possible locations, it is always a bug (because we are going to write something that should be nested at the end of the file, and that is not a valid thing to do).

I am not sure if throwing an error here is correct (just in terms of the logging, it might be more helpful to let this finish so we can log the result as we do today), but feels like a least an error log would be nice to add...

packages/workspace/test/workspace/nacl_files/nacl_file_update.test.ts
```diff
@@ -90,24 +101,283 @@ describe('getNestedStaticFiles', () => {
...
+        path: ['file'],
+      }
+      const sourceMap = new Map([
+        [
+          mockType.elemID.getFullName(),
+          [
+            {
+              filename: 'file.nacl',
+              start: { line: 1, col: 1, byte: 0 },
+              end: { line: 10, col: 2, byte: 126 },
```
ori-moisis, line 187: something with the column numbers doesn't feel right
where did you get these numbers?

I wonder if it would make more sense to have the test work with actual nacl data and parse it to create the elements and source maps?
I think it might be easier to both read and write new tests that way...

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -31,20 +33,33 @@ import {
...
+
+export type DetailedChangeWithSource = DetailedChange & {
+  location: Location
+  requiresIndent?: boolean
+}
 
 const createFileNameFromPath = (pathParts?: ReadonlyArray<string>): string =>
   `${path.join(...(pathParts ?? ['unsorted']))}${FILE_EXTENSION}`
 
+type positionInParent = {
```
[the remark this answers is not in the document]
yelly: Done.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -64,41 +79,112 @@ export const getChangeLocations = (
...
+    if (change.baseChange === undefined) {
+      log.warn('No base change: %s', inspectValue(change))
+      return { followingElementIDs: [] }
+    }
+
+    const changeData = getChangeData(change)
+    const parent = isField(changeData) ? changeData.parent : getChangeData(change.baseChange)
+    const pathInParent = getPath(parent, change.id)
+    if (pathInParent === undefined) {
+      log.warn('Could not get path for %s in parent: %s', inspectValue(change.id), inspectValue(parent))
```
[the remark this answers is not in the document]
yelly: Done.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -64,41 +79,112 @@ export const getChangeLocations = (
...
+      return { followingElementIDs: [] }
+    }
+
+    const container = _.get(parent, pathInParent.slice(0, -1))
+    if (!_.isObjectLike(container)) {
+      log.warn(
+        'Got non object container %s for path %s in parent: %s',
+        inspectValue(container),
+        inspectValue(change.id.createParentID()),
```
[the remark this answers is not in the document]
yelly: Done.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -64,41 +79,112 @@ export const getChangeLocations = (
...
+        inspectValue(parent),
+      )
+      return { followingElementIDs: [] }
+    }
+
+    const idNameParts = change.id.getFullNameParts().slice(0, -1)
```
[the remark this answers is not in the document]
yelly: `createSiblingID` won't do everything we need because we also need the name for finding the index.
I've added an explicit check for top kevek elements when getting the path (even though everything works fine without is as far as I can tell) - does this address all the edge cases you have in mind?

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -64,41 +79,112 @@ export const getChangeLocations = (
     }
   }
 
-  const findLocations = (): parser.SourceRange[] => {
+  const getPositionInParent = (): positionInParent => {
```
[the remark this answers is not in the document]
yelly: Agreed, done.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -64,41 +79,112 @@ export const getChangeLocations = (
...
     } else if (!change.id.isTopLevel()) {
-      // We add new values / elements as the last part of a parent scope unless the parent scope
-      // is a config element
+      const fileName = createFileNameFromPath(change.path)
+      const { followingElementIDs, indexInParent } = getPositionInParent()
+      const possibleFollowingElementsRanges = followingElementIDs
+        .map(elemID => sourceMap.get(elemID.getFullName()))
+        .filter(isDefined)
+        .flat()
+        .filter(sr => sr.filename === fileName)
```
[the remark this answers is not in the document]
yelly: Nice, done.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
yelly: Yes, nice catch.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
yelly: I tried doing this, but it doesn't work. It seems to be throwing off the alignment between the column and the bytes which actually ends up causing the following part to get indented. Happy to go over this with you if you think I'm missing something here.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
yelly: Done.

packages/workspace/test/workspace/nacl_files/nacl_file_update.test.ts
ori-moisis: I agree that an integration test could be valuable, and I also agree a test at this level is valuable, my comment was about the test at this level.
my suggestion was to have the same test at the same level, just generate the inputs for this test using `parse` instead of manually.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
ori-moisis: seems like there is [some code](https://github.com/salto-io/salto/blob/main/packages/workspace/src/workspace/nacl_files/nacl_file_update.ts/#L115-L119) which still assumes that additions are always before the closing bracket.
maybe that code needs to change as well?
at least the comment isn't true anymore...

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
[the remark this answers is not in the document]
yelly: Discussed F2F and fixed.

packages/workspace/test/workspace/nacl_files/nacl_file_update.test.ts
yelly: Done.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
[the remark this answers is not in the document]
yelly: Turns out this isn't true because the source map doesn't include members of arrays.
@ori-moisis do you think we should put in the work to support this?

packages/adapter-api/src/element_id.ts
```diff
@@ -258,6 +258,14 @@ export class ElemID {
     return new ElemID(this.adapter, this.typeName)
   }
 
+  createSiblingID(name: string): ElemID {
```
ori-moisis, line 261: please add unit tests so the behavior of this is clear
I think this should be tested with the different id types (annotation, attr, instance, field), even though in the code these are not different cases

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -31,20 +33,75 @@ import {
...
+
+  const changeData = getChangeData(change)
+  const parent = isField(changeData) ? changeData.parent : getChangeData(change.baseChange)
+  const pathInParent = getPath(parent, change.id)
+  if (pathInParent === undefined) {
+    log.warn('Could not get path for %s in parent: %s', change.id.getFullName(), inspectValue(parent))
+    return { followingElementIDs: [] }
+  }
+  if (pathInParent.length === 0) {
+    return { followingElementIDs: [] }
```
ori-moisis, line 77: can you please add a comment saying when this is possible?

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -31,20 +33,75 @@ import {
...
+    log.warn(
+      'Got non object container at path %s: %s',
+      change.id.createParentID().getFullName(),
+      inspectValue(container),
+    )
+    return { followingElementIDs: [] }
+  }
+
+  const elementName = change.id.name
+  const index = Object.keys(container).indexOf(elementName)
```
ori-moisis, line 91: nit: get the `Object.keys` once instead of twice

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -31,20 +33,75 @@ import {
...
 
 const createFileNameFromPath = (pathParts?: ReadonlyArray<string>): string =>
   `${path.join(...(pathParts ?? ['unsorted']))}${FILE_EXTENSION}`
 
+type PositionInParent = {
+  followingElementIDs: ElemID[]
+  indexInParent?: number
+}
+
+const getPositionInParent = (change: DetailedChange): PositionInParent => {
```
ori-moisis, line 63: nit: can we have a more specific type here?
IIUC this is only ever called with addition changes, so maybe we can define this as
```suggestion
const getPositionInParent = (change: DetailedChange & AdditionChange<unknown>): PositionInParent => {
```
or something like that?

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -64,41 +121,71 @@ export const getChangeLocations = (
...
+      const fileName = createFileNameFromPath(change.path)
+      const { followingElementIDs, indexInParent } = getPositionInParent(change)
+      const possibleFollowingElementsRange = followingElementIDs
+        .flatMap(elemID => sourceMap.get(elemID.getFullName()))
+        .filter(isDefined)
+        .find(sr => sr.filename === fileName)
+      if (possibleFollowingElementsRange !== undefined) {
+        return [
+          {
+            location: {
```
ori-moisis, line 147: nit: worth adding a comment about the behavior we expect with this return

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -64,41 +121,71 @@ export const getChangeLocations = (
...
+      }
+
+      // If we can't find an element after this one in the parent we put it at the end
       const parentID = change.id.createParentID()
-      const possibleLocations = sourceMap.get(parentID.getFullName()) || []
+      const possibleLocations = sourceMap.get(parentID.getFullName()) ?? []
       if (possibleLocations.length > 0) {
-        const foundInPath = possibleLocations.find(sr => sr.filename === createFileNameFromPath(change.path))
+        const foundInPath = possibleLocations.find(sr => sr.filename === fileName)
+        /* When adding a nested change we need to increase one level of indentation because
```
ori-moisis, line 162: nit: comment style - we mostly use `//` even for multiline comments (maybe worth aligning the rest of the comments in the file as well)

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -112,7 +199,7 @@ const fixEdgeIndentation = (data: string, action: ActionName, initialIndentation
     lines.push(lastLine)
   }
   if (action === 'add') {
-    /* When adding the placement we are given is right before the closing bracket.
+    /* When adding the placement we are given is right before following member or the closing bracket of the parent.
```
ori-moisis, line 202: how does this actually work when adding a value before a following member?
the code here de-indents the next line under the assumption that the next line is supposed to be one level less than the addition, but when there is a following member we should not do it, no?

ori-moisis (approved):

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
yelly: Done.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
yelly: Done.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
yelly: Done.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
yelly: "...recreate the indentation that was there previously".
If we are in the middle, we have one more indentation than if we are at the end, which is the desired behaviour.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
yelly: Done.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
yelly: Done.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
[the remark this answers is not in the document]
yelly: Discussed OOB.

packages/adapter-api/src/element_id.ts
yelly: Done.

PR #5878, opened 2024-05-09
SALTO-5516 order apps field in workspace by ID as well as position

benisraeladi:
Fix issue where two apps with the same position can flip places by adding the ID to the sorting

---

---
_Release Notes_: 
Zendesk Adapter:
* Two app installation with the same position in workspace will keep their set position

---
_User Notifications_: _None_


edenhassid (approved):
LGTM! 🔥

PR #5879, opened 2024-05-09
SALTO-5916: [Zendesk] Add a second sorting field for conditions in ticket forms

DanielBarros1:
Ticket field conditions are sorted by "value". if there are dup values (like `true` or `false`), the sorting is not robust enough. In this PR we add an additional field, parent_field_id, so that sorting stays consistent
---

---
_Release Notes_: 
Zendesk: 
Make sorting of conditions in Ticket Forms more consistent

---
_User Notifications_: 
Zendesk: 
Make sorting of conditions in Ticket Forms more consistent


edenhassid (approved):
LGTM! 🔥

PR #5880, opened 2024-05-09
SALTO-5477: Improving performance object comparison in plan

yelly:
When comparing objects we first check if they have the same keys by doing a set comparison. Turns out `_.isEqual` has terrible performance for sets, so replaced it with a more performant implementation.

---

None

---
_Release Notes_: 
None.

---
_User Notifications_: 
None.

### DIFF
--- a/packages/core/src/core/plan/plan.ts
+++ b/packages/core/src/core/plan/plan.ts
@@ -139,7 +139,11 @@ const compareValuesAndLazyResolveRefs = async (
   }
 
   if (_.isPlainObject(first) && _.isPlainObject(second)) {
-    if (!values.setsEqual(new Set(Object.keys(first)), new Set(Object.keys(second)))) {
+    if (Object.keys(first).length !== Object.keys(second).length) {
+      return false
+    }
+    const secondKeys = new Set(Object.keys(second))
+    if (Object.keys(first).some(k => !secondKeys.has(k))) {
       return false
     }
     return !(await awu(Object.keys(first)).some(
--- a/packages/core/src/core/plan/plan.ts
+++ b/packages/core/src/core/plan/plan.ts
@@ -139,7 +139,7 @@ const compareValuesAndLazyResolveRefs = async (
   }
 
   if (_.isPlainObject(first) && _.isPlainObject(second)) {
-    if (!_.isEqual(new Set(Object.keys(first)), new Set(Object.keys(second)))) {
+    if (!values.setsEqual(new Set(Object.keys(first)), new Set(Object.keys(second)))) {
       return false
     }
     return !(await awu(Object.keys(first)).some(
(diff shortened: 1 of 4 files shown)


packages/lowerdash/src/values.ts
```diff
@@ -32,3 +32,6 @@ export const lookupValue = (blob: unknown, lookupFunc: (val: unknown) => boolean
   }
   return false
 }
+
+// _.isEqual has terrible performance for sets, prefer this when possible
+export const setsEqual = <T>(a: Set<T>, b: Set<T>): boolean => a.size === b.size && [...a].every(x => b.has(x))
```
ori-moisis, line 37: couple of problems here
1. using `...` is problematic as it unpacks the input iterator on the stack, with a long enough iterator (usually on the order of 400k) this overflows. this happened to us in several cases so we try to avoid doing this unless we know the input is small
2. This does not do the same thing as `_.isEqual`, AFAIK the check for `.has()` will do a strict comparison (`===`), whereas `_.isEqual` checks for value equality (similar to `==` but actually more accurate than that as well)

I think for the specific use case of comparing strings this would work, but for the general case this is not equivalent.

for the first problem, you could use `wu(a.values()).every(...)` to operate directly on the iterator without cloning the whole list, or compare the performance to the built in `isSubset`/`isSuperset` (which, after comparing sizes, should be equivalent I think)

for the second problem, maybe change the name, restrict the type, or just have comment warning about the difference in behavior or something?

yelly: I've just moved to implementing this straight inline where it used. We might end up never needing this anywhere else. It also saves creating the sets when we don't need them.
WDYT?

packages/core/test/core/plan/plan.test.ts
```diff
@@ -170,6 +175,23 @@ describe('getPlan', () => {
     expect(planItems).toHaveLength(6)
   })
 
+  it('should create plan with a huge change in reasonable time', async () => {
```
ori-moisis, line 178: I don't think we should have this kind of test in our code.
it is not really a "behavior" to do something quickly, and these kinds of tests tend to become flaky and dependent on where they run

packages/core/src/core/plan/plan.ts
```diff
@@ -139,7 +139,11 @@ const compareValuesAndLazyResolveRefs = async (
   }
 
   if (_.isPlainObject(first) && _.isPlainObject(second)) {
-    if (!_.isEqual(new Set(Object.keys(first)), new Set(Object.keys(second)))) {
+    if (Object.keys(first).length !== Object.keys(second).length) {
```
ori-moisis, line 142: do we have anything to gain from not calling `Object.keys` twice (here and in lines 145-146)?

ori-moisis (approved):

packages/core/test/core/plan/plan.test.ts
yelly: OK, removing.

packages/core/src/core/plan/plan.ts
yelly: Well turns out `Object.keys()` can be [`O(nlogn)`](https://stackoverflow.com/a/64912755)(!?), so I guess we're better off only running it once.

ori-moisis: cool

PR #5881, opened 2024-05-09
SALTO-5915 Google ws - fix - roleAssignment deployment

Uri-Bar:
SALTO-5915 Google ws - fix - roleAssignment deployment
---

---
_Release Notes_: 
none 

---
_User Notifications_: 
none

### DIFF
--- a/packages/adapter-components/src/client/polling.ts
+++ b/packages/adapter-components/src/client/polling.ts
@@ -16,6 +16,7 @@
 import { retry } from '@salto-io/lowerdash'
 import { ResponseValue, Response } from './http_connection'
 import { PollingArgs } from '../definitions'
+import { HTTPError } from './http_client'
 
 const {
   withRetry,
@@ -32,7 +33,7 @@ export const executeWithPolling = async <T>(
       const response = await singleClientCall(args)
       return polling.checkStatus(response) ? response : undefined
     } catch (e) {
-      if (polling.retryOnStatus?.includes(e.response?.status)) {
+      if (e instanceof HTTPError && polling.retryOnStatus?.includes(e.response?.status)) {
         return undefined
       }
       throw e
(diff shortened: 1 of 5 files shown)


packages/adapter-components/src/client/polling.ts
```diff
@@ -28,10 +28,16 @@ export const executeWithPolling = async <T>(
   singleClientCall: (args: T) => Promise<Response<ResponseValue | ResponseValue[]>>,
 ): Promise<Response<ResponseValue | ResponseValue[]>> => {
   const pollingFunc = async (): Promise<Response<ResponseValue | ResponseValue[]> | undefined> => {
-    const response = await singleClientCall(args)
-    return polling.checkStatus(response) ? response : undefined
+    try {
+      const response = await singleClientCall(args)
+      return polling.checkStatus(response) ? response : undefined
+    } catch (e) {
+      if (polling.retryOnStatus?.includes(e.response?.status)) {
```
shir-reifenberg, line 35: let's verify as well that ?
`e instanceof clientUtils.HTTPError`

packages/adapter-components/src/definitions/system/requests/types.ts
```diff
@@ -38,6 +38,7 @@ export type PollingArgs = {
   interval: number
   retries: number
   checkStatus: (response: Response<ResponseValue | ResponseValue[]>) => boolean
+  retryOnStatus?: number[]
```
shir-reifenberg, line 41: nit - maybe worth adding a comment about why this is needed and how it's different from `checkStatus` ?

shir-reifenberg (approved):
looks great! 👯 
for next prs, I think it's better to split adapter changes from infra changes, and link each one to it's own ticket. It makes it easier to understand later why we done each change
Also, please update PR title to include info about the infra change 🙏

PR #5883, opened 2024-05-09
SALTO-5940: Fix Forms fetching keep failing due to 403 error

idoamit8:
In this PR I fixed the Forms fetching errors that were cause by 403 forbidden access. 

---

Previously, attempts to access forms without valid permissions would yield a 200 OK status with an empty data array. Our handling of this scenario was based on the assumption of successful access.

With recent updates to the API, which is private, unauthorized access now correctly returns a 403 Forbidden status. This change prompted the addition of a try-catch block to properly manage fetch errors.

Example from oriSaltoTest:
`Failed to fetch forms for projects: Dan_Avigdor_JSM@s, AlonITSM3, ITSM_project@s, Sagi_s_JSM@ts, testProject, testFacilities3, ITSM_sample_space@s, testFacilities4, test, testFacilities2, testFacilities due to insufficient permissions`

---
_Release Notes_: 
_Jira Adpater:_
* The issue where the fetch failed due to receiving a 403 error when attempting to access forms without proper permissions has been fixed. Now, the fetch will succeed and the user will see a clearer error message, making the process smoother and more straightforward.

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/forms/forms.ts
+++ b/packages/jira-adapter/src/filters/forms/forms.ts
@@ -114,41 +114,42 @@ const filter: FilterCreator = ({ config, client, fetchQuery }) => ({
       await Promise.all(
         jsmProjects.flatMap(async project => {
           try {
-          const url = `/gateway/api/proforma/cloudid/${cloudId}/api/1/projects/${project.value.id}/forms`
-          const res = await client.get({ url })
-          if (!isFormsResponse(res)) {
-            return undefined
-          }
-          return await Promise.all(
-            res.data.map(async formResponse => {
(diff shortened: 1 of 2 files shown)


packages/jira-adapter/src/filters/forms/forms.ts
```diff
@@ -156,7 +163,12 @@ const filter: FilterCreator = ({ config, client, fetchQuery }) => ({
       form.value = mapKeysRecursive(form.value, ({ key }) => naclCase(key))
       elements.push(form)
     })
-
+    if (projectsNamesWithPermissionErrors.length > 0) {
```
omrilit, line 166: do we want to have logs somewhere? or is the response already logged and it's enough?

omrilit (approved):
👑

PR #5886, opened 2024-05-12
SALTO-5334, SALTO-5346 - Salesforce: move PermissionSet refs to custom references code

tamtamirr:
This implementation is based on old PR by @tomsellek  https://github.com/salto-io/salto/pull/5547

Move the extraction of references from the reference mapper to custom references. This will speed fetches in cases where `PermissionSet`s are included and some of them are very large.

---

_Release Notes_: 
Salesforce Adapter:
- Add PermissionSet custom references handler.

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/src/custom_references/permission_sets.ts
+++ b/packages/salesforce-adapter/src/custom_references/permission_sets.ts
@@ -1,0 +1,65 @@
+/*
+ *                      Copyright 2024 Salto Labs Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ReferenceInfo, Element, InstanceElement } from '@salto-io/adapter-api'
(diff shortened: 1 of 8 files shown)


packages/salesforce-adapter/src/custom_references/permission_sets.ts
```diff
@@ -0,0 +1,65 @@
...
+ * limitations under the License.
+ */
+
+import { ReferenceInfo, Element, InstanceElement } from '@salto-io/adapter-api'
+import { logger } from '@salto-io/logging'
+import { collections } from '@salto-io/lowerdash'
+import { PERMISSION_SET_METADATA_TYPE } from '../constants'
+import { isInstanceOfTypeSync } from '../filters/utils'
+import { WeakReferencesHandler } from '../types'
+import { mapProfileOrPermissionSetSections } from './profiles'
```
yelly, line 23: Maybe move all the common code to a separate file?

packages/salesforce-adapter/src/custom_references/permission_sets.ts
```diff
@@ -0,0 +1,65 @@
...
+    'Generated %d references for %d elements.',
+    refs.length,
+    elements.length,
+  )
+  return refs
+}
+
+export const permissionSetsHandler: WeakReferencesHandler = {
+  findWeakReferences,
+  removeWeakReferences: () => async () => ({ fixedElements: [], errors: [] }),
```
yelly, line 64: We will need to also implement fix elements here for the same reason. OK to do in a separate PR.

amitgoldarb (approved):
Thanks for this!

packages/salesforce-adapter/src/custom_references/permission_sets.ts
tamtamirr: Will do in a separate PR 🙏

PR #5893, opened 2024-05-13
SALTO-5537 (Netsuite): remove unresolved netsuite references in forms

SeggevHaimovich:
Forms in netsuite tend to have unresolved Netsuite references, this PR enables the users to deploy forms with unresolved Netsuite references between environments by removing references to elements that are not exist in the target env.

We created a new hidden element that saves all the `scripid`s in the client during the fetch to check the existence of a `scriptid` in the env later.

---

Jira ticket about removing unresolved references from forms: https://salto-io.atlassian.net/browse/SALTO-5537

---
_Release Notes_: 
_Salto users will be able to deploy forms with unresolved netsuite references_

---
_User Notifications_: 
_None_


packages/netsuite-adapter/src/filters/remove_missing_references_from_forms.ts
```diff
@@ -0,0 +1,120 @@
...
+      return value
+    }
+    const { id } = value
+    if (id === undefined) {
+      return value
+    }
+    if (
+      _.isString(id) &&
+      isUnresolvedStringReference(id, generatedDependencies) &&
+      isAdditionField(path, formChange, id)
```
dantal4, line 76: should we check that the path contains `.field.`? because in the form type there are more things that have `.id` in them (e.g button, tab, etc.)

packages/netsuite-adapter/src/filters/remove_missing_references_from_forms.ts
```diff
@@ -0,0 +1,120 @@
...
+  )
+
+  const pathsToRemove = await getPathsToRemove(formChange, generatedDependencies)
+
+  if (pathsToRemove.length === 0) {
+    return
+  }
+
+  pathsToRemove.forEach(path =>
+    _.unset(formChange.data.after.value, formChange.data.after.elemID.getRelativePath(path).join('.')),
```
dantal4, line 107: you can use `setPath` here

dantal4 (commented):
Great work!
We also talked about a CV that tells the user that we're going to remove unresolved references, right?

packages/netsuite-adapter/src/scriptid_list.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export const getScriptIdList = async (elementsSource: ReadOnlyElementsSource): Promise<string[]> => {
+  const scriptIdListElement = await elementsSource.get(SCRIPT_ID_LIST_INSTANCE_ID)
+
+  if (!isInstanceElement(scriptIdListElement)) {
+    log.warn('scriptid list instance not found in elements source')
+    return []
+  }
+
+  return collections.array.makeArray(scriptIdListElement.value.scriptid_list)
```
dantal4, line 95: please use a field name constant - in type and instance creation too.

dantal4 (commented):
Looks good!

packages/netsuite-adapter/src/scriptid_list.ts
```diff
@@ -0,0 +1,96 @@
...
+    [CORE_ANNOTATIONS.HIDDEN]: true,
+  },
+})
+
+export const createScriptIdListElements = (instancesIds: ObjectID[]): [ObjectType, InstanceElement] => {
+  log.debug('Creating script id list elements')
+  const instance = new InstanceElement(
+    ElemID.CONFIG_NAME,
+    scriptIdListObjectType,
+    { scriptid_list: instancesIds.map(id => id.instanceId) },
```
dantal4, line 55: 1. I think that it would be good to save the typeName of each object too- meaning the whole `ObjectID`
2. We should make sure that custom record types of custom segments (e.g `customrecord_cseg*`) are written too- because I think that only the segment is returned from SDF list command (and we get the custom record type on the SDF import object command)

packages/netsuite-adapter/src/custom_references/weak_references/fields_references.ts
```diff
@@ -101,65 +102,48 @@ const getFormReferences = (formInstance: InstanceElement): ReferenceInfo[] => {
...
-    isUnresolvedReference(dep.reference, elementsSource),
-  )
-
-  return {
-    fixedGeneratedDependencies: resolved,
-    unresolvedGeneratedDependencies: new Set<string>(
-      unresolved.map(dep => dep.reference.elemID.createTopLevelParentID().parent.name),
-    ),
-  }
+): Promise<boolean> => (await elementsSource.get(ref.elemID.createTopLevelParentID().parent)) !== undefined
```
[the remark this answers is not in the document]
SeggevHaimovich: I think that in the fix element stage the references are still unresolved but I am not sure, I'll double check it.

packages/netsuite-adapter/src/custom_references/weak_references/fields_references.ts
```diff
@@ -186,7 +185,7 @@ const fixIndexInMappedList = (mappedList: MappedList): void =>
       mappedList[key][INDEX] = index
     })
 
-const fixIndexes = (form: InstanceElement, pathsToRemove: ElemID[]): void => {
+export const fixIndexes = (form: InstanceElement, pathsToRemove: ElemID[]): void => {
```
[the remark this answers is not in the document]
SeggevHaimovich: leftover from the use in the filter

packages/netsuite-adapter/src/scriptid_list.ts
dantal4: @SeggevHaimovich  see [this fix](https://github.com/salto-io/salto/pull/6014) for bullet 2

packages/netsuite-adapter/src/custom_references/weak_references/fields_references.ts
```diff
@@ -202,51 +201,50 @@ const fixIndexes = (form: InstanceElement, pathsToRemove: ElemID[]): void => {
...
+  envScriptIds: Set<string>,
+): Promise<{ instance: InstanceElement; paths: string[] } | undefined> => {
   const fixedForm = form.clone()
 
-  const { fixedGeneratedDependencies, unresolvedGeneratedDependencies } = await handleGeneratedDependencies(
-    form,
-    elementsSource,
+  const fixedGeneratedDependencies = await getResolvedGeneratedDependencies(fixedForm, elementsSource)
+  const generatedDependenciesScriptId = new Set<string>(
+    fixedGeneratedDependencies.map(dep => dep.reference.elemID.createTopLevelParentID().parent.name),
```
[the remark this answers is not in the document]
SeggevHaimovich: Here too I can't use the `reference.value` because the references are not resolved yet, the only way to get the scriptid field is to use the elementsSource to get the element.
I combined the creation of the set of `scriptid`s and the creation of filtered generated dependency to get from elementsSource only once per reference.

packages/netsuite-adapter/src/custom_references/weak_references/fields_references.ts
```diff
@@ -187,12 +209,10 @@ const fixIndexInMappedList = (mappedList: MappedList): void =>
     })
 
 const fixIndexes = (form: InstanceElement, pathsToRemove: ElemID[]): void => {
-  const pathsToFixIndex = new Set<string>(
-    pathsToRemove.map(path => form.elemID.getRelativePath(path.createParentID()).join('.')),
-  )
+  const pathsToFixIndex = new Set<ElemID>(pathsToRemove.map(path => path.createParentID()))
```
dantal4, line 212: set doesn't work well with non primitive values...
```suggestion
  const pathsToFixIndex = _.uniqBy(pathsToRemove.map(path => path.createParentID()), id => id.getFullName())
```

packages/netsuite-adapter/src/custom_references/weak_references/fields_references.ts
```diff
@@ -202,51 +222,45 @@ const fixIndexes = (form: InstanceElement, pathsToRemove: ElemID[]): void => {
...
-  pathsToRemove.forEach(path => _.unset(fixedForm.value, fixedForm.elemID.getRelativePath(path).join('.')))
+  pathsToRemove.forEach(path => setPath(fixedForm, path, undefined))
 
   fixIndexes(fixedForm, pathsToRemove)
 
-  fixedForm.annotations[CORE_ANNOTATIONS.GENERATED_DEPENDENCIES] = fixedGeneratedDependencies
+  fixedForm.annotations[CORE_ANNOTATIONS.GENERATED_DEPENDENCIES] = generatedDependencies
 
-  return { instance: fixedForm, paths: pathsToRemove }
+  return { instance: fixedForm, relatedPaths: pathsToRemove.map(path => form.elemID.getRelativePath(path).join('.')) }
```
dantal4, line 243: nit - keep `relatedPaths` as `elemID[]` here, and make it a string only when you return the error message. IMO we should keep dealing with structured objects as much as possible, and avoid "losing info" along the way.

packages/netsuite-adapter/src/scriptid_list.ts
```diff
@@ -0,0 +1,106 @@
...
+const OBJECT_ID_LIST_TYPE_ID = new ElemID(NETSUITE, OBJECT_ID_LIST_TYPE_NAME)
+const OBJECT_ID_LIST_INSTANCE_ID = new ElemID(NETSUITE, OBJECT_ID_LIST_TYPE_NAME, 'instance', ElemID.CONFIG_NAME)
+
+const objectIdType = new ObjectType({
+  elemID: new ElemID(NETSUITE, 'objectid'),
+  isSettings: true,
+  fields: {
+    instanceId: { refType: BuiltinTypes.STRING },
+    type: { refType: BuiltinTypes.STRING },
+  },
```
dantal4, line 46: isn't it missing the `_hidden` annotation?

packages/netsuite-adapter/src/scriptid_list.ts
```diff
@@ -0,0 +1,106 @@
...
+    ElemID.CONFIG_NAME,
+    objectIdListObjectType,
+    { [OBJECT_ID_LIST_FIELD_NAME]: instancesIds },
+    undefined,
+    {
+      [CORE_ANNOTATIONS.HIDDEN]: true,
+    },
+  )
+
+  return [objectIdListObjectType, instance]
```
dantal4, line 72: we should return `objectIdType` too

packages/netsuite-adapter/src/scriptid_list.ts
```diff
@@ -0,0 +1,106 @@
...
+  },
+  annotations: {
+    [CORE_ANNOTATIONS.HIDDEN]: true,
+  },
+})
+
+export const createObjectIdListElements = (instancesIds: ObjectID[]): [ObjectType, InstanceElement] => {
+  log.debug('Creating object id list elements')
+  const instance = new InstanceElement(
+    ElemID.CONFIG_NAME,
```
dantal4, line 63: nit 
```suggestion
    OBJECT_ID_LIST_INSTANCE_ID.name,
```

dantal4 (commented):
Great work!

dantal4 (approved):
Really great work! I'm approving the PR but please fix the comments before merging 👍

PR #5894, opened 2024-05-15
SALTO-5879: Okta: Add `Brand` and `BrandTheme` to e2e test.

Nurdok:
This PR adds `Brand` and `BrandTheme` instances to the Okta e2e test.
Since the e2e environment doesn't support multiple `Brand`s, there's some complication as we need to make a separate `GET` request to find the id for the singleton `Brand`.

---

---
_Release Notes_: 
None

---

_User Notifications_: 
None


packages/okta-adapter/e2e_test/adapter.test.ts
```diff
@@ -103,12 +114,9 @@ const createInstance = ({
...
   }
-  const { idFields } = configUtils.getConfigWithDefault(
-    DEFAULT_CONFIG[API_DEFINITIONS_CONFIG].types[typeName].transformation ?? {},
-    DEFAULT_CONFIG[API_DEFINITIONS_CONFIG].typeDefaults.transformation,
-  )
-  const instName = name ?? getInstanceName(instValues, idFields, typeName)
-  const naclName = naclCase(parent ? `${parent.elemID.name}__${instName}` : String(instName))
+  const { idFields } = getTransformationConfig(typeName)
+  const instName = name ?? getInstanceName(instValues, idFields ?? [], typeName)
+  const naclName = naclCase(parent && extendsParentId ? `${parent.elemID.name}__${instName}` : String(instName))
```
shir-reifenberg, line 119: This part is going to change in https://github.com/salto-io/salto/pull/5718 so I think we shouldn't rely on `extendsParentId` from the old config if that's possible..

packages/okta-adapter/e2e_test/adapter.test.ts
```diff
@@ -256,6 +286,66 @@ const createChangesForDeploy = (types: ObjectType[], testSuffix: string): Change
...
+      primaryColorHex: '#1662dd',
+      primaryColorContrastHex: '#ffffff',
+      secondaryColorHex: '#ebebed',
+      secondaryColorContrastHex: '#000000',
+      signInPageTouchPointVariant: 'OKTA_DEFAULT',
+      endUserDashboardTouchPointVariant: 'OKTA_DEFAULT',
+      errorPageTouchPointVariant: 'OKTA_DEFAULT',
+      emailTemplateTouchPointVariant: 'OKTA_DEFAULT',
+      loadingPageTouchPointVariant: 'OKTA_DEFAULT',
+    },
```
shir-reifenberg, line 325: any reason not to put those in `mockDefaultValues` ?  (same for `theme`)

packages/okta-adapter/e2e_test/adapter.test.ts
```diff
@@ -390,7 +482,7 @@ describe('Okta adapter E2E', () => {
       const types = fetchBeforeCleanupResult.elements.filter(isObjectType)
       await deployCleanup(adapterAttr, fetchBeforeCleanupResult.elements.filter(isInstanceElement))
 
-      const changesToDeploy = createChangesForDeploy(types, testSuffix)
+      const changesToDeploy: Change[] = await createChangesForDeploy(types, testSuffix, adapterAttr.client)
```
shir-reifenberg, line 485: nit - 
```suggestion
      const changesToDeploy = await createChangesForDeploy(types, testSuffix, adapterAttr.client)
```

packages/okta-adapter/e2e_test/adapter.test.ts
```diff
@@ -529,11 +621,18 @@ describe('Okta adapter E2E', () => {
         .map(change => getChangeData(change)) as InstanceElement[]
 
       deployInstances.forEach(deployedInstance => {
+        elements.filter(isInstanceElement).forEach(e => log.info('Instance %s', e.elemID.getFullName()))
```
shir-reifenberg, line 624: please move to `log.trace`

packages/okta-adapter/e2e_test/adapter.test.ts
```diff
@@ -529,11 +621,18 @@ describe('Okta adapter E2E', () => {
...
         // Omit hidden fields
-        const originalValue = _.omit(instance?.value, ['_links', 'customName'])
-        const isEqualResult = isEqualValues(originalValue, deployedInstance.value)
+        const typeTransformation = getTransformationConfig(deployedInstance.elemID.typeName)
+        const fieldsToIgnoreInComparison: string[] = [
+          ...(typeTransformation?.fieldsToOmit ?? []),
+          ...(typeTransformation.fieldsToHide ?? []),
+        ].map(f => f.fieldName)
+        const originalValue = _.omit(instance?.value, fieldsToIgnoreInComparison)
+        const deployedValue = _.omit(deployedInstance.value, fieldsToIgnoreInComparison)
```
shir-reifenberg, line 634: why would the deployedValue contain `fieldsToOmit` or `fieldsToHide` ?  
Also, if there are not many of them, I prefer we use fields name explicitly cause the `transformation` will be deprecated soon 🤞

packages/okta-adapter/e2e_test/adapter.test.ts
Nurdok: Done

packages/okta-adapter/e2e_test/adapter.test.ts
Nurdok: Done

packages/okta-adapter/e2e_test/adapter.test.ts
Nurdok: Done

packages/okta-adapter/e2e_test/adapter.test.ts
Nurdok: This isn't `extendsParentId` from the old config, it's just a parameter for this local function. The issue is that it automatically concatenated the parent name in `naclName` when the `BrandTheme` behavior in single-brand Okta environment is that it's just named `unnamed_0` without a trailing underscore. Leaving as it is for now as I'm not relying on the old config.

PR #5895, opened 2024-05-15
SALTO-5927: fix parser to not use trim in createStringValue

edenhassid:
fix parser to not use trim in createStringValue

---

---
_Release Notes_: 
core:
* fix issue where pending changes appear in multi-line string when the unicode U+200D is the last character

---
_User Notifications_: 

### DIFF
--- a/packages/parser/src/parser/internal/native/consumers/values.ts
+++ b/packages/parser/src/parser/internal/native/consumers/values.ts
@@ -258,10 +258,11 @@ const consumeMultilineString: Consumer<string | TemplateExpression> = context =>
     tokens.push(context.lexer.next())
   }
   if (tokens.length > 0) {
+    // We get rid of the trailing newline
     tokens[tokens.length - 1].value = tokens[tokens.length - 1].value.slice(0, -1)
     tokens[tokens.length - 1].text = tokens[tokens.length - 1].text.slice(0, -1)
   }
-  // We get rid of the trailing newline
+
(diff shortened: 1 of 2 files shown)


packages/parser/src/parser/internal/native/consumers/values.ts
```diff
@@ -269,11 +259,12 @@ const consumeMultilineString: Consumer<string | TemplateExpression> = context =>
   }
   if (tokens.length > 0) {
     tokens[tokens.length - 1].value = tokens[tokens.length - 1].value.slice(0, -1)
+    tokens[tokens.length - 1].text = tokens[tokens.length - 1].text.slice(0, -1)
```
ori-moisis, line 262: maybe add a comment saying what this part of the code is doing?
it is not trivial...

packages/parser/test/parser/parse.test.ts
```diff
@@ -802,6 +802,22 @@ multiline
     expect(elements[0].annotations.str.length).toEqual(stringLength)
   })
 
+  it('test', async () => {
```
ori-moisis, line 805: maybe worth adding a more descriptive name

also, since the unicode character that makes this test worthwhile is not visible, maybe add a comment about it and what exactly the scenario under test here (e.g - why this specific character, why is specifically has to be the last character in the content, etc...)

ori-moisis (approved):

PR #5897, opened 2024-05-15
SALTO-5800 fix homepage deployment

omrifarkash:
In context to resolveAllArgs, we didn't passed the data (after transformation), but only the value from the original change

---

---
_Release Notes_: 
_Confluence Adapter_:
- Fix homepage deployment

---
_User Notifications_: 
_Confluence Adapter_:
- Fix homepage deployment

### DIFF
--- a/packages/confluence-adapter/src/definitions/utils/page.ts
+++ b/packages/confluence-adapter/src/definitions/utils/page.ts
@@ -71,7 +71,9 @@ const increasePageVersion: definitions.AdjustFunction<definitions.deploy.ChangeA
       value: {
         ...value,
         version: {
-          number: 2, // case where homepage is created in the service with version = 1
+          // In case of homepage addition, we don't have a version number yet but it is "1" in the service
+          // It has been set to one when we created the space and the default homepage was created
+          number: 2,
         },
       },
     }
--- a/packages/confluence-adapter/src/definitions/utils/page.ts
+++ b/packages/confluence-adapter/src/definitions/utils/page.ts
@@ -67,7 +67,14 @@ const increasePageVersion: definitions.AdjustFunction<definitions.deploy.ChangeA
   const value = validateValue(args.value)
   const version = _.get(value, 'version')
   if (!values.isPlainRecord(version) || !isNumber(version.number)) {
-    return { ...args, value }
+    return {
+      value: {
+        ...value,
+        version: {
+          number: 2, // case where homepage is created in the service with version = 1
+        },
+      },
+    }
   }
   return {
     value: {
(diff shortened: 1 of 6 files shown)


packages/adapter-components/src/deployment/deploy/requester.ts
```diff
@@ -297,7 +297,7 @@ export const getRequester = <TOptions extends APIDefinitionsOptions>({
 
     const finalEndpointIdentifier = replaceAllArgs({
       value: mergedEndpointDef,
-      context: _.merge({}, value, additionalContext),
+      context: _.merge({}, value, additionalContext, data),
```
netama, line 300: let's discuss this one IRL? (also will be good to have a UT demonstrating the behavior change - but I'm not sure this is the correct order of precedence)

packages/confluence-adapter/src/definitions/utils/page.ts
```diff
@@ -67,7 +67,14 @@ const increasePageVersion: definitions.AdjustFunction<definitions.deploy.ChangeA
   const value = validateValue(args.value)
   const version = _.get(value, 'version')
   if (!values.isPlainRecord(version) || !isNumber(version.number)) {
-    return { ...args, value }
+    return {
+      value: {
+        ...value,
+        version: {
+          number: 2, // case where homepage is created in the service with version = 1
```
netama, line 74: nit: not sure the comment will be that clear in a few months

packages/confluence-adapter/test/definitions/utils/page.test.ts
```diff
@@ -52,7 +52,7 @@ describe('page definitions utils', () => {
         expect(adjustPageOnModification(args).value.version.number).toEqual(2)
       })
 
-      it('should return the same value if the version number is not a number (should not happen)', () => {
+      it('should return version = 2 if the version number is not a number (should not happen)', () => {
```
netama, line 55: is it still true that this should never happen?

packages/adapter-components/src/deployment/deploy/requester.ts
shoshana-houri: Encountered a similar problem and @netama referenced me here, and it did help! However, I then tried adding the relevant info to the context using `context.custom` on the deploy definitions, and it also worked without modifying the line here, because the result [is stored](https://github.com/salto-io/salto/blob/2c54785fe5ac9bfe92d5f4eb5892ddf463b1676e/packages/adapter-components/src/deployment/deploy/requester.ts#L265) as the additionalContext which is already used here.

I'm not sure it's the best solution, and it will require using `updateHomepageId` under `context.custom` instead of under `transformation.adjust`, but thought it's best sharing in case you'll prefer using this approach.

omrifarkash: You're right, this is working. Not pushing this change

packages/confluence-adapter/src/definitions/utils/page.ts
omrifarkash: will improve it

PR #5899, opened 2024-05-15
SALTO-5545 fix confluence build

omrifarkash:
Fix broken build, caused by merging [this](https://github.com/salto-io/salto/pull/5876) PR

---

---
_Release Notes_: 
None
---
_User Notifications_: 
None

PR #5900, opened 2024-05-15
SALTO-5817 support deployment of settings instances

omrifarkash:
- Resolve args on request.queryArgs (adapter-component)
- Initial support in deployment of space settings. There still some things todo but I think this adds some value and can be merge until we will support more complex cases.
- Convert action to modification from addition when deploying new space with its settings. 

TODO:
On this PR we support modification of custom settings. Global is not configurable.
- Allow to set which settings definition to use on a space ('custom', 'global', 'theme'). using [this](https://developer.atlassian.com/cloud/confluence/rest/v1/api-group-settings/#api-wiki-rest-api-settings-lookandfeel-put) API
- Block somehow modification on global field on settings instance, as this is not editable in the sevice
- Explore more about 'theme' option. As I understand this option is from some plugins 

---

---
_Release Notes_: 
_Confluence Adapter_:
- support deployment of custom space settings

---
_User Notifications_: 
_Confluence Adapter_:
- support deployment of custom space settings

### DIFF
--- a/packages/adapter-components/src/deployment/deploy/requester.ts
+++ b/packages/adapter-components/src/deployment/deploy/requester.ts
@@ -282,7 +282,13 @@ export const getRequester = <TOptions extends APIDefinitionsOptions>({
     throwOnUnresolvedReferences(data)
 
     const callArgs = {
-      queryParams: mergedEndpointDef.queryArgs,
+      queryParams:
+        mergedEndpointDef.queryArgs !== undefined
+          ? replaceAllArgs({
+              value: mergedEndpointDef.queryArgs,
+              context: _.merge({}, value, additionalContext),
+            })
+          : undefined,
       headers: mergedEndpointDef.headers,
       data,
     }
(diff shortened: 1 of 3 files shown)


packages/confluence-adapter/src/adapter_creator.ts
```diff
@@ -58,7 +58,7 @@ export const adapter = createAdapter<Credentials, Options, UserConfig>({
       // customPathsFilterCreator must run after fieldReferencesFilter
       customPathsFilterCreator,
     }),
-    additionalChangeValidators: createChangeValidator(),
+    additionalChangeValidators: createChangeValidator,
```
netama, line 61: doesn't this break https://github.com/salto-io/salto/pull/5899/files again?

packages/adapter-components/src/deployment/deploy/requester.ts
```diff
@@ -282,7 +282,13 @@ export const getRequester = <TOptions extends APIDefinitionsOptions>({
     throwOnUnresolvedReferences(data)
 
     const callArgs = {
-      queryParams: mergedEndpointDef.queryArgs,
+      queryParams:
+        mergedEndpointDef.queryArgs !== undefined
```
netama, line 286: can you add a UT please?

netama (approved):
LGTM, added a couple of comments

packages/confluence-adapter/src/adapter_creator.ts
omrifarkash: Just need to rebase. this is not a real change on this PR

PR #5902, opened 2024-05-16
SALTO-5861: [Zendesk] Fix fetching of article attachments

DanielBarros1:
When fetching article attachment, we use a url that has the filename in it - which breaks when the file name is weird. In this PR i switched it so that we use the same url but remove the filename (which was suggested by zendesk support)
---

---
_Release Notes_: 
Zendesk:
Fix fetching of article attachments with weird names

---
_User Notifications_: 
None

### DIFF
--- a/packages/zendesk-adapter/src/filters/article/utils.ts
+++ b/packages/zendesk-adapter/src/filters/article/utils.ts
@@ -164,7 +164,7 @@ const getAttachmentContent = async ({
   try {
     const path = attachment.value.relative_path
     res = await client.get({
-      url: `${path.substring(0, path.lastIndexOf('/') + 1)}`,
+      url: `${path.substring(0, path.lastIndexOf('/'))}`,
       responseType: 'arraybuffer',
     })
   } catch (e) {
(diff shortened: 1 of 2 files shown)


packages/zendesk-adapter/src/filters/article/utils.ts
```diff
@@ -162,8 +162,9 @@ const getAttachmentContent = async ({
   const client = brandIdToClient[attachment.value.brand]
   let res
   try {
+    const path = attachment.value.relative_path
     res = await client.get({
-      url: `${attachment.value.relative_path}`,
+      url: `${path.substring(0, path.lastIndexOf('/') + 1)}`,
```
edenhassid, line 167: do we need the last `/`?

edenhassid (commented):
looks good :) added one small question

packages/zendesk-adapter/src/filters/article/utils.ts
DanielBarros1: not really lol, i'll remove it

edenhassid (approved):

PR #5904, opened 2024-05-16
SALTO-5914: retry on Automation's 504

IdoZakk:
Added a retry to prevent failing on 504

---

None

---
_Release Notes_: 
Jira Adapter:
* Added a retry when automation rules fetch fail with 504

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/automation/automation_fetch.ts
+++ b/packages/jira-adapter/src/filters/automation/automation_fetch.ts
@@ -27,7 +27,14 @@ import { values as lowerdashValues } from '@salto-io/lowerdash'
 import { logger } from '@salto-io/logging'
 import Joi from 'joi'
 import _ from 'lodash'
-import { AUTOMATION_TYPE, fetchFailedWarnings, JIRA, PROJECT_TYPE, PROJECTS_FIELD } from '../../constants'
+import {
+  AUTOMATION_RETRY_PERIODS,
+  AUTOMATION_TYPE,
+  fetchFailedWarnings,
+  JIRA,
+  PROJECT_TYPE,
+  PROJECTS_FIELD,
+} from '../../constants'
 import JiraClient from '../../client/client'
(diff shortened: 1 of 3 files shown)


packages/jira-adapter/src/filters/automation/automation_fetch.ts
```diff
@@ -78,33 +85,54 @@ const PAGE_RESPONSE_SCHEME = Joi.object({
...
+  try {
+    const response = await client.post({
+      url,
+      data: {
+        offset,
+        limit: pageSize,
+      },
+    })
+    if (!isPageResponse(response.data)) {
+      throw new Error('Failed to get response page, received invalid response')
```
kami-delaroz, line 117: instead of try-catch and double exception throwing, why not something like:

```
if (!isPageResponse(response.data)) {
    if (!(e instanceof clientUtils.HTTPError && e.response?.status === 504)) {
      throw new Error('Failed to get response page, received invalid response')
    }
    // we get an occasional 504 from the Automation's APIs, Atlassian's solution is to retry
    return requestPageRecurse({ url, client, offset, pageSize, retriesUsed: retriesUsed + 1 })
}
return response.data
```

kami-delaroz (approved):

packages/jira-adapter/src/filters/automation/automation_fetch.ts
IdoZakk: If we get a 504 it throws an exception, we need to catch it

kami-delaroz: Oh, I thought it was our error that was thrown. so we catch both "our" error and 504 error in this scenario?

IdoZakk: We catch our errors and all the the other http errors

PR #5905, opened 2024-05-16
SALTO-5812: Okta: Add deploy support for Domain instances

Nurdok:
Add deploy support for Domain instances (add, modify, remove).

---

---
_Release Notes_: 
_Okta adapter_:
- Added deploy support for Domain instances

---
_User Notifications_: 
None


packages/okta-adapter/src/change_validators/domain_addition.ts
```diff
@@ -0,0 +1,32 @@
...
+    .filter(isInstanceChange)
+    .filter(isAdditionChange)
+    .map(getChangeData)
+    .filter(instance => instance.elemID.typeName === DOMAIN_TYPE_NAME)
+    .filter(instance => instance.value.brandId === undefined)
+    .map(instance => ({
+      elemID: instance.elemID,
+      severity: 'Error',
+      message: 'Cannot add domain without a brand',
+      detailedMessage: `Domain ${instance.value.domain} must have a brand.`,
```
shir-reifenberg, line 31: the assumption under using `value.domain` in detailed message is that it's part of the elemID, but it can be configured by the user so I think we should avoid using it.
elemID contains the domain details so I'm ok with not including it in the detailed message

packages/okta-adapter/src/change_validators/domain_modification.ts
```diff
@@ -0,0 +1,39 @@
...
+  changes
+    .filter(isInstanceChange)
+    .filter(isModificationChange)
+    .filter(change => getChangeData(change).elemID.typeName === DOMAIN_TYPE_NAME)
+    .map(getAllChangeData)
+    .filter(([before, after]) => !_.isEqual(_.omit(before.value, 'brandId'), _.omit(after.value, 'brandId')))
+    .map(([instance]) => ({
+      elemID: instance.elemID,
+      severity: 'Error',
+      message: 'Cannot modify any domain fields except its brand',
```
shir-reifenberg, line 37: sorry if we already discussed this and I forgot, but why is that? can't we change the domain name for example ?

packages/okta-adapter/e2e_test/adapter.test.ts
```diff
@@ -144,6 +146,77 @@ const getIdForDefaultBrandTheme = async (client: OktaClient, brandId: string): P
...
+  // And even more conveniently, this e2e test doesn't run any change validator prior to deploying.
+  // We use this to create a domain here without a brand ID, even though it's not supported in production code.
+  // This allows us to both test the domain deploy actions (even if they're not exactly like production), but moreover
+  // it allows us to test custom page creation, which requires a custom domain to exist.
+  const domain = createInstance({
+    typeName: DOMAIN_TYPE_NAME,
+    types,
+    valuesOverride: {
+      domain: 'subdomain.salto.io',
+      // If Okta's free tier ever supports multiple brands in non-paid accounts, we can add a brandId here.
```
shir-reifenberg, line 210: nit: I think this comment is redundant 😅 we can have a ticket if we want...

shir-reifenberg (commented):
Looks good! 😎 
added one question and nits :)

packages/okta-adapter/e2e_test/adapter.test.ts
Nurdok: Removed.

shir-reifenberg (approved):
🚀

PR #5907, opened 2024-05-16
SALTO-5918: Using element source in field references filter

yelly:
Resolving a very old TODO to use the element source directly in the field references filter.
This doesn't make a huge difference in runtime, but leads to a cleaner implementation. We will chase performance improvements in later PRs.

---

This filter is super slow in some SFDC cases. We might also filter out profiles in that specific case in the future, but this PR is already big enough.

---
_Release Notes_: 
None.

---
_User Notifications_: 
None.

### DIFF
--- a/packages/adapter-components/src/references/context.ts
+++ b/packages/adapter-components/src/references/context.ts
@@ -31,12 +31,12 @@ const log = logger(module)
 export type ContextValueMapperFunc = (val: string) => string | undefined
 export type ContextFunc = ({
   instance,
-  elementsSource,
+  elementSource,
   field,
   fieldPath,
 }: {
   instance: InstanceElement
-  elementsSource: ReadOnlyElementsSource
+  elementSource: ReadOnlyElementsSource
   field: Field
   fieldPath?: ElemID
   levelsUp?: number
@@ -78,7 +78,7 @@ export const neighborContextGetter =
     contextValueMapper?: ContextValueMapperFunc
     getLookUpName: GetLookupNameFunc
   }): ContextFunc =>
-  async ({ instance, elementsSource, fieldPath }) => {
+  async ({ instance, elementSource, fieldPath }) => {
     if (fieldPath === undefined || contextFieldName === undefined) {
       return undefined
     }
@@ -87,7 +87,7 @@ export const neighborContextGetter =
       const contextField = await getField(await instance.getType(), fieldPath.createTopLevelParentID().path)
       const refWithValue = new ReferenceExpression(
         context.elemID,
-        context.value ?? (await elementsSource.get(context.elemID)),
+        context.value ?? (await elementSource.get(context.elemID)),
       )
       return getLookUpName({ ref: refWithValue, field: contextField, path, element: instance })
     }
--- a/packages/adapter-components/src/references/context.ts
+++ b/packages/adapter-components/src/references/context.ts
@@ -15,29 +15,28 @@
  */
 import _ from 'lodash'
 import {
-  Element,
   Field,
   ReferenceExpression,
   InstanceElement,
   ElemID,
   getField,
   isReferenceExpression,
+  ReadOnlyElementsSource,
 } from '@salto-io/adapter-api'
 import { resolvePath, GetLookupNameFunc } from '@salto-io/adapter-utils'
 import { logger } from '@salto-io/logging'
-import { multiIndex } from '@salto-io/lowerdash'
 
 const log = logger(module)
 
 export type ContextValueMapperFunc = (val: string) => string | undefined
 export type ContextFunc = ({
   instance,
-  elemByElemID,
+  elementsSource,
   field,
   fieldPath,
 }: {
   instance: InstanceElement
-  elemByElemID: multiIndex.Index<[string], Element>
+  elementsSource: ReadOnlyElementsSource
   field: Field
   fieldPath?: ElemID
   levelsUp?: number
@@ -79,7 +78,7 @@ export const neighborContextGetter =
     contextValueMapper?: ContextValueMapperFunc
     getLookUpName: GetLookupNameFunc
   }): ContextFunc =>
-  async ({ instance, elemByElemID, fieldPath }) => {
+  async ({ instance, elementsSource, fieldPath }) => {
     if (fieldPath === undefined || contextFieldName === undefined) {
       return undefined
     }
@@ -88,7 +87,7 @@ export const neighborContextGetter =
       const contextField = await getField(await instance.getType(), fieldPath.createTopLevelParentID().path)
       const refWithValue = new ReferenceExpression(
         context.elemID,
-        context.value ?? elemByElemID.get(context.elemID.getFullName()),
+        context.value ?? (await elementsSource.get(context.elemID)),
       )
       return getLookUpName({ ref: refWithValue, field: contextField, path, element: instance })
     }
--- a/packages/adapter-components/src/references/field_references.ts
+++ b/packages/adapter-components/src/references/field_references.ts
@@ -25,8 +25,10 @@ import {
   ElemID,
   cloneDeepWithoutRefs,
   isElement,
+  ReadOnlyElementsSource,
 } from '@salto-io/adapter-api'
 import {
+  buildElementsSourceFromElements,
   GetLookupNameFunc,
   GetLookupNameFuncArgs,
   TransformFunc,
@@ -69,16 +71,16 @@ const isRelativeSerializer = <CustomIndexField extends string>(
 export const replaceReferenceValues = async <TContext extends string, CustomIndexField extends string>({
   instance,
   resolverFinder,
-  elemLookupMaps,
+  elemIDLookupMaps,
   fieldsWithResolvedReferences,
-  elemByElemID,
+  elementsSource,
   contextStrategyLookup = emptyContextStrategyLookup,
 }: {
   instance: InstanceElement
   resolverFinder: ReferenceResolverFinder<TContext, CustomIndexField>
-  elemLookupMaps: Record<string, multiIndex.Index<[string, string], Element>>
+  elemIDLookupMaps: Record<string, multiIndex.Index<[string, string], ElemID>>
   fieldsWithResolvedReferences: Set<string>
-  elemByElemID: multiIndex.Index<[string], Element>
+  elementsSource: ReadOnlyElementsSource
   contextStrategyLookup?: Record<TContext, ContextFunc>
 }): Promise<Values> => {
   const getRefElem = async ({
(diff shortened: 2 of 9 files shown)


packages/adapter-utils/src/element_source.ts
```diff
@@ -76,11 +76,22 @@ export const buildElementsSourceFromElements = (
...
+    let element: Readonly<Element> | undefined = elementsMap[id.getFullName()]
     if (element !== undefined) {
       return element
     }
 
+    if (!id.isTopLevel()) {
+      const { parent } = id.createTopLevelParentID()
+      const topLevel = elementsMap[parent.getFullName()]
+      if (topLevel !== undefined) {
+        element = resolvePath(topLevel, id)
```
ori-moisis, line 88: at this point it is not an element, it is a `Value` (which is basically `any`)
it is confusing to use the same variable name here, also because the type explicitly says this is an element.

I think `element` should remain a `const` and this should use a different `const`

packages/salesforce-adapter/src/filters/field_references.ts
```diff
@@ -176,39 +175,29 @@ export const addReferences = async (
...
+  await awu(instances).forEach(async (instance) => {
+    instance.value = await replaceReferenceValues({
+      instance,
+      resolverFinder,
+      elemIDLookupMaps: { elemIDLookup },
+      fieldsWithResolvedReferences,
+      elementsSource: referenceElements,
+      contextStrategyLookup,
     })
+  })
```
ori-moisis, line 200: would this be faster if we let it run on more instances concurrently?

```suggestion
  await withLimitedConcurrency(
    elements
      .filter(isInstanceElement)
      .map(instance => async () => {
        instance.value = await replaceReferenceValues({
          instance,
          resolverFinder,
          elemIDLookupMaps: { elemIDLookup },
          fieldsWithResolvedReferences,
          elementsSource: referenceElements,
          contextStrategyLookup,
        })
      }),
    20 // any reasonable number that is larger than 1 but not too much to avoid memory issues, and of course it should be a const and not literally 20 here
  )
```

packages/adapter-utils/src/element_source.ts
yelly: How is this possible? Doesn't that mean that the type of `elementsMap` (`_.Dictionary<Readonly<Element>>`) is wrong?

packages/salesforce-adapter/src/filters/field_references.ts
yelly: Only one way to find out.

packages/adapter-utils/src/element_source.ts
yelly: Done.

packages/salesforce-adapter/src/filters/field_references.ts
yelly: From a single (large) test this actually seems to be slower so I'll go back to the first implementation.

packages/adapter-components/src/references/field_references.ts
```diff
@@ -98,15 +100,20 @@ export const replaceReferenceValues = async <TContext extends string, CustomInde
...
         return undefined
       }
 
-      return lookup.get(targetType, value)
+      const elemID = lookup.get(targetType, value)
+      if (elemID === undefined) {
+        return undefined
+      }
+
+      return elementsSource.get(elemID)
```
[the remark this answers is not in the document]
ori-moisis: even if we stay just in SF for now, does this change specifically (removing the index and changing the mapping to be async here) make it better or worse? (noticeably)

packages/adapter-utils/test/element_source.test.ts
```diff
@@ -58,13 +65,23 @@ describe('elementSource', () => {
...
           expect(await elementsSource.get(new ElemID('adapter', 'type3'))).toBeUndefined()
         })
+
+        it('should return a nested field if it exists', async () => {
+          expect(await elementsSource.get(new ElemID('adapter', 'nestedType', 'field', 'strField'))).toBe(
+            elements[2].fields.strField,
+          )
+        })
+
+        it('should return a nested field if it does not exist', async () => {
```
ori-moisis, line 82:
```suggestion
        it('should return undefined if a nested field does not exist', async () => {
```

packages/salesforce-adapter/src/filters/field_references.ts
```diff
@@ -176,39 +174,29 @@ export const addReferences = async (
...
-      instance.value = await replaceReferenceValues({
-        instance,
-        resolverFinder,
-        elemLookupMaps: { elemLookup },
-        fieldsWithResolvedReferences,
-        elemByElemID,
-        contextStrategyLookup,
-      })
+  const instances = elements.filter(isInstanceElement)
+  await awu(instances).forEach(async (instance) => {
```
ori-moisis, line 190: nit
```suggestion
  await awu(instances).forEach(async instance => {
```

ori-moisis (commented):
overall LGTM
will let @netama review as well since this is touching a fairly deep part of adapter components

packages/adapter-utils/test/element_source.test.ts
```diff
@@ -58,13 +65,23 @@ describe('elementSource', () => {
       })
 
       describe('get', () => {
-        it('should return element if exists', async () => {
+        it('should return element if it exists', async () => {
           expect(await elementsSource.get(new ElemID('adapter', 'type1'))).toBe(elements[0])
         })
 
-        it('should return undefined if not exists', async () => {
+        it('should return undefined if it does not exists', async () => {
```
netama, line 72:
```suggestion
        it('should return undefined if it does not exist', async () => {
```

packages/adapter-utils/src/element_source.ts
```diff
@@ -76,9 +76,19 @@ export const buildElementsSourceFromElements = (
   }
 
   const get = async (id: ElemID): Promise<Value> => {
-    const element = elementsMap[id.getFullName()]
-    if (element !== undefined) {
-      return element
+    if (id.getFullName() in elementsMap) {
```
netama, line 79: probably not a huge difference, but why not keep the old (single-lookup) logic and check if undefined? do we expect something to be in the mapping _and_ be undefined?

packages/adapter-utils/src/element_source.ts
```diff
@@ -76,9 +76,19 @@ export const buildElementsSourceFromElements = (
...
+      return elementsMap[id.getFullName()]
+    }
+
+    if (!id.isTopLevel()) {
+      const { parent } = id.createTopLevelParentID()
+      const topLevel = elementsMap[parent.getFullName()]
+      if (topLevel !== undefined) {
+        const value = resolvePath(topLevel, id)
+        if (value !== undefined) {
```
netama, line 88: not sure how much of this is new, but it looks to me like there's some risk in allowing to get nested paths inside the same element from different sources (e.g. one source has `a.b` and the other `a.c` and they're never supposed to be defined together)? @ori-moisis @adamro wdyt?

packages/adapter-components/src/references/context.ts
```diff
@@ -15,29 +15,28 @@
...
 import { logger } from '@salto-io/logging'
-import { multiIndex } from '@salto-io/lowerdash'
 
 const log = logger(module)
 
 export type ContextValueMapperFunc = (val: string) => string | undefined
 export type ContextFunc = ({
   instance,
-  elemByElemID,
+  elementsSource,
```
netama, line 34: nit: please use `elementSource` and not `elementsSource`, trying to get rid of this in new places at least so that eventually it will disappear...

packages/adapter-components/src/references/context.ts
```diff
@@ -88,7 +87,7 @@ export const neighborContextGetter =
       const contextField = await getField(await instance.getType(), fieldPath.createTopLevelParentID().path)
       const refWithValue = new ReferenceExpression(
         context.elemID,
-        context.value ?? elemByElemID.get(context.elemID.getFullName()),
+        context.value ?? (await elementsSource.get(context.elemID)),
```
netama, line 90: won't this slow us down rather than speed up? what am I missing?

packages/adapter-components/src/references/field_references.ts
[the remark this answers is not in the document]
netama: also confused here, I thought the goal was to reduce the async

netama (commented):
started reviewing and added some questions, but this is going in a different direction that I understood so might be easier to discuss IRL first :)

packages/adapter-components/src/references/field_references.ts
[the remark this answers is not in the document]
yelly: From my tests it doesn't really move the needle either way.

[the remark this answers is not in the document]
ori-moisis: can you share the numbers please?

packages/adapter-utils/src/element_source.ts
ori-moisis: +1 on this comment, I missed this

the `in` operator is generally not something that we should use.
when we do need to check for `has` in an object, we generally use `hasOwnProperty` to avoid the strange edge cases (e.g - `toString` is always defined)

ori-moisis: also, this isn't correct, if you allow nested IDs as input, this needs to be
```typescript
const element = elementsMap[id.createTopLevelParentID().parent.getFullName()]
```

packages/adapter-utils/src/element_source.ts
ori-moisis: so you suggest to remove this `if (value !== undefined)` and always return the `resolvePath` if the top level ID was found?

packages/adapter-components/src/references/field_references.ts
```diff
@@ -266,31 +273,29 @@ export const addReferences = async <
...
+    fieldsToGroupBy.forEach(fieldName =>
+      indexer.addIndex({
+        name: fieldName,
+        filter: e => isInstanceElement(e) && e.value[fieldName] !== undefined,
+        key: (inst: InstanceElement) => [inst.refType.elemID.name, inst.value[fieldName]],
+        map: inst => inst.elemID,
+      }),
+    )
+    fieldLookups = await indexer.process(awu(contextElements))
+  }
```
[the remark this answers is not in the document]
yelly: Done.

packages/adapter-utils/test/element_source.test.ts
yelly: Done.

packages/adapter-utils/test/element_source.test.ts
yelly: Done.

packages/salesforce-adapter/src/filters/field_references.ts
yelly: The formatter forces this style.

packages/adapter-utils/src/element_source.ts
yelly: Done.

packages/adapter-components/src/references/context.ts
yelly: Done.

packages/salesforce-adapter/src/filters/field_references.ts
ori-moisis: huh, didn't know that was changed (seems like the prettier commit changed this)

yelly:
Answering all the remaining questions together here.
The original goal of this PR was to reduce the runtime, but it evolved into solving the lingering TODO to of moving from the ad-hoc indices to using the element source.
It turns out that this doesn't really change the runtime at all (at least in my tests). My best guess is that on one hand we are paying a price for moving to async but making gains from not needing to load all the elements into the index.
For the sake of example, I profiled the Salesforce filter with Nitzan's huge env and got that the old implementation ran in 466.8 seconds and the new implementation ran in 467.5 seconds. This difference is small enough to ignore. I've attached the profiles here.

I'll amend the PR description to reflect all of this. We can chase performance improvements in later CLs over the (IMO) simpler implementation.

[old.cpuprofile](https://github.com/user-attachments/files/15752623/old.cpuprofile)
[new.cpuprofile](https://github.com/user-attachments/files/15752622/new.cpuprofile)

ori-moisis:
in that case I think we shouldn't merge this PR

if we were to continue with this PR we'd have to prove this doesn't significantly hurt the other scenarios - (i.e - compare a large fetch in ZD and Jira as well where we know there are very large envs), I am not sure it is worth the effort given that we do not actually expect to gain performance here.

btw - the "not loading all elements into the index" assumption seems unlikely - the elements are all in memory anyway in this scenario, especially in a full fetch.

yelly:
So do you think the TODO was out of place? Why did you put it there in the first place?

ori-moisis:
IIRC I added it under the assumption that it would improve performance
this was before I knew how bad the effect of async code was on performance

PR #5909, opened 2024-05-16
SALTO-5933: fix type for section

edenhassid:
fix type for section

---

adding direct_parent_type to filedTypeOverrides so that it will appear in the type as this field is added in a filter

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/zendesk-adapter/src/config.ts
+++ b/packages/zendesk-adapter/src/config.ts
@@ -2172,6 +2172,7 @@ export const DEFAULT_TYPES: ZendeskApiConfig['types'] = {
       fieldTypeOverrides: [
         { fieldName: 'id', fieldType: 'number' },
         { fieldName: 'parent_section_id', fieldType: 'number' },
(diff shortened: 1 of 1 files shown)


packages/zendesk-adapter/src/config.ts
```diff
@@ -2172,6 +2172,7 @@ export const DEFAULT_TYPES: ZendeskApiConfig['types'] = {
       fieldTypeOverrides: [
         { fieldName: 'id', fieldType: 'number' },
         { fieldName: 'parent_section_id', fieldType: 'number' },
+        { fieldName: 'direct_parent_type', fieldType: 'string' },
```
ori-moisis, line 2175: what about `'position'`, it also appears above in the `fieldsToHide`

edenhassid: we get the position from the response of the service. its already in the type

DanielBarros1 (approved):
lgtm

PR #5910, opened 2024-05-16
SALTO 4242 (Okta): support additions of app user schema

SeggevHaimovich:
_This PR will make Salto support additions of appUserSchema._
When adding Application, the client creates automatically the related appUserSchema.
We behave as if the user added the appUserSchema but behind the scenes we fetch the auto-created appUserSchema and modifying it.

In addition, changes in the base field of appUserSchema are unsupported with Okta API, therefore we warn about these changes and ignore them.

Notes:
* **We are currently ignore the option to change the `required` field within base, we may change it in the future**
* When modifying the base field, the user will get a warning that he cannot do it and that other changes will be deployed. In his env after the deployment he will see the element as if the base have changed. After fetch, he will see the real base field.

---

Jira ticket: https://salto-io.atlassian.net/browse/SALTO-4242

---
_Release Notes_:
Okta Adapter:
* support additions of app user schema

---
_User Notifications_: 
_None_


idoamit8 (commented):
I haven't looked at the code yet, but I have two quick high-level questions:

Why are there separate filters for removal and for modifying/adding? Having two different filters makes maintenance harder.

Regarding this point:

> When modifying the base field, the user will get a warning that they cannot do it and that other changes will be deployed. In their environment after the deployment, they will see the element as if the base has changed. After fetching, they will see the real base field.


I don't understand the reason for this inconsistency between deployment and fetching. Why does this happen? We can modify the instance in onDeploy if necessary.

packages/okta-adapter/src/change_validators/app_user_schema_base_changes.ts
```diff
@@ -0,0 +1,84 @@
...
+  ChangeError,
+} from '@salto-io/adapter-api'
+import { values } from '@salto-io/lowerdash'
+import _ from 'lodash'
+import { APP_USER_SCHEMA_TYPE_NAME } from '../constants'
+
+const getErrorFromChange = (
+  change: ModificationChange<InstanceElement> | AdditionChange<InstanceElement>,
+): ChangeError | undefined => {
+  if (isAdditionChange(change)) {
```
idoamit8, line 34: Since this message will be visible to users with every addition, we may want to reduce its severity to an informational level. Additionally, it would be beneficial to have a Product Manager review the message for clarity and appropriateness.

packages/okta-adapter/src/change_validators/app_user_schema_base_changes.ts
```diff
@@ -0,0 +1,84 @@
...
+  const afterWithSameBase = after.clone()
+
+  if (_.isUndefined(beforeBase)) {
+    _.unset(afterWithSameBase.value, ['definitions', 'base'])
+  } else {
+    _.set(afterWithSameBase.value, ['definitions', 'base'], beforeBase)
+  }
+
+  if (!_.isEqual(beforeBase, afterBase)) {
+    if (_.isEqual(afterWithSameBase.value, change.data.before.value)) {
```
idoamit8, line 56:
```suggestion
if (_.isEqual(afterWithSameBase.value, before.value)) {
```

packages/okta-adapter/src/change_validators/app_user_schema_base_changes.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  const afterWithSameBase = after.clone()
+
+  if (_.isUndefined(beforeBase)) {
+    _.unset(afterWithSameBase.value, ['definitions', 'base'])
+  } else {
+    _.set(afterWithSameBase.value, ['definitions', 'base'], beforeBase)
+  }
+
+  if (!_.isEqual(beforeBase, afterBase)) {
```
idoamit8, line 55: nit: Could you check this condition first?

```
if (_.isEqual(beforeBase, afterBase)) {
  return undefined;
}
```
Then proceed with the logic if they are not equal. This will make the code cleaner and reduce indentation.

packages/okta-adapter/src/config.ts
```diff
@@ -505,11 +505,15 @@ const DEFAULT_TYPE_CUSTOMIZATIONS: OktaSwaggerApiConfig['types'] = {
       serviceUrl: '/admin/universaldirectory#app/{_parent.0.id}',
     },
     deployRequests: {
-      // Hack to pass through createCheckDeploymentBasedOnConfigValidator validator for removals
+      // Hack to pass through createCheckDeploymentBasedOnConfigValidator validator for removals and additions
```
idoamit8, line 508: Lets keep the order of add, modify, remove

idoamit8 (commented):
Great work! 💯 I have added a few comments.

packages/okta-adapter/src/change_validators/app_user_schema_base_changes.ts
```diff
@@ -0,0 +1,84 @@
...
+): ChangeError | undefined => {
+  if (isAdditionChange(change)) {
+    return {
+      elemID: change.data.after.elemID,
+      severity: 'Warning',
+      message: 'Cannot deploy the base field in App User Schema',
+      detailedMessage:
+        'Okta API does not support deploying the base field in App User Schema. Salto will deploy the element without the base field. After fetch the field will be updated as the client',
+    }
+  }
```
shir-reifenberg, line 42: I'm not sure if we should have 3 different messages here.. cause the bottom line is that we don't support changes to the base attributes (also change terminology to "base attributes" and not "base field" and this is the terms in the UI). 

therefore, I would phrase something in the spirit of  -
message: `Base attributes will not be deployed`
detailedMessage: `Salto cannot make changes to base attributes, as they are determined by the associated application.`

packages/okta-adapter/src/change_validators/app_user_schema_base_changes.ts
```diff
@@ -0,0 +1,84 @@
...
+  const beforeBase = before.value.definitions?.base
+  const afterBase = after.value.definitions?.base
+
+  const afterWithSameBase = after.clone()
+
+  if (_.isUndefined(beforeBase)) {
+    _.unset(afterWithSameBase.value, ['definitions', 'base'])
+  } else {
+    _.set(afterWithSameBase.value, ['definitions', 'base'], beforeBase)
+  }
```
shir-reifenberg, line 53: why is this needed?

shir-reifenberg (commented):
Looks good (only did partial review, will leave @idoamit8 to continue :) )
I think we should changes messages a bit, we can also discuss IRL

packages/okta-adapter/src/filters/app_user_schema_deployment.ts
```diff
@@ -0,0 +1,201 @@
...
+ * Deploy changes of appUserSchema.
+ * additions - changing them to modification changes,
+ * because appUserSchema automatically created by the service when deploying new Application
+ * modifications - changing the base field to the original value because Okta's API doesn't support changing it
+ * removals - verifying the parent application is deleted. appUserSchema is deleted if and only if the parent application is deleted
+ */
+const filterCreator: FilterCreator = ({ definitions, oldApiDefinitions }) => ({
+  name: 'appUserSchemaDeployment',
+  deploy: async changes => {
+    const client = definitions.clients.options.main.httpClient as OktaClient
```
idoamit8, line 187:
```suggestion
const client = definitions.clients.options.main.httpClient
```

packages/okta-adapter/src/filters/app_user_schema_deployment.ts
```diff
@@ -0,0 +1,201 @@
...
+  const appUserSchemaInstance = getChangeData(change)
+  const parentApplicationId = getParents(appUserSchemaInstance)[0]?.id
+  if (!_.isString(parentApplicationId) || !(await verifyApplicationIsDeleted(parentApplicationId, client))) {
+    throw new Error('Expected the parent Application to be deleted')
+  }
+}
+
+const deployChange = async (
+  change: Change<InstanceElement>,
+  client: OktaClient,
```
idoamit8, line 165:
```suggestion
client: clientUtils.HTTPWriteClientInterface & clientUtils.HTTPReadClientInterface,
```

Also in all other functions in this file.

packages/okta-adapter/test/change_validators/app_user_schema_base_changes.test.ts
```diff
@@ -0,0 +1,137 @@
...
+  beforeEach(() => {
+    jest.clearAllMocks()
+    appUserSchemaInstance = appUserSchema.clone()
+  })
+  describe('addition changes', () => {
+    it('should return info when adding an appUserSchema', async () => {
+      const changes = [toChange({ after: appUserSchemaInstance })]
+      const changeErrors = await appUserSchemaBaseChangesValidator(changes)
+      expect(changeErrors).toHaveLength(1)
+      expect(changeErrors[0]).toEqual(getErrorWithSeverity(appUserSchemaInstance.elemID, 'Info'))
```
idoamit8, line 74: When we write tests, we usually write the text as-is, without calling functions or using ${}. This acts as a safeguard to verify there are no spelling mistakes, and that the text appears as expected.

idoamit8 (commented):
Great work! 💯
I've added some small comments.

@shir-reifenberg  I see it interfaces with the new infrastructure. Can you look only at these parts?

packages/okta-adapter/src/filters/app_user_schema_deployment.ts
```diff
@@ -0,0 +1,201 @@
...
+
+const makeModificationFromAddition = async (
+  change: AdditionChange<InstanceElement>,
+  client: OktaClient,
+  apiDefinitions: OktaSwaggerApiConfig,
+): Promise<ModificationChange<InstanceElement>> => {
+  const appUserSchemaInstance = getChangeData(change)
+  const parentApplicationId = getParents(appUserSchemaInstance)[0]?.id
+  if (parentApplicationId === undefined) {
+    log.error(`Error while trying to get parent id for user schema ${appUserSchemaInstance.elemID.getFullName()}`)
```
shir-reifenberg, line 112: nit: there's another type called `user schema` so maybe change to `AppUserSchema` ? (same with the error below)

packages/okta-adapter/src/change_validators/app_user_schema_base_properties.ts
```diff
@@ -23,53 +23,50 @@ import {
...
-  if (isAdditionChange(change)) {
-    return {
-      elemID: change.data.after.elemID,
-      severity: 'Warning',
-      message: 'Cannot deploy the base field in App User Schema',
-      detailedMessage:
-        'Okta API does not support deploying the base field in App User Schema. Salto will deploy the element without the base field. After fetch the field will be updated as the client',
-    }
-  }
+const getBasePath = (): string[] => [DEFINITIONS_FIELD, BASE_FIELD]
```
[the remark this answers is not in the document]
shir-reifenberg: let's also export this and re-use in `app_user_schema_deployment` ?

packages/okta-adapter/src/filters/app_user_schema_deployment.ts
```diff
@@ -0,0 +1,201 @@
...
+  autoCreatedappUserSchema: AppUserSchema,
+  appUserSchemaInstance: InstanceElement,
+  apiDefinitions: OktaSwaggerApiConfig,
+): InstanceElement => {
+  const createdAppUserSchemaInstance = appUserSchemaInstance.clone()
+  const { fieldsToOmit } = configUtils.getTypeTransformationConfig(
+    createdAppUserSchemaInstance.elemID.typeName,
+    apiDefinitions.types,
+    apiDefinitions.typeDefaults,
+  )
```
shir-reifenberg, line 96: this part is deprecated, please use `fieldCustomizations` from `fetchDefinitions` instead.
there's an example for omitting hidden values [here](https://github.com/salto-io/salto/blob/main/packages/okta-adapter/e2e_test/adapter.test.ts#L466)
so it should be something like this but with `omit` instead of `hide`

packages/okta-adapter/src/filters/app_user_schema_deployment.ts
```diff
@@ -0,0 +1,201 @@
...
+}
+
+const AppUserSchemaSchema = Joi.object({
+  id: Joi.string().required(),
+}).unknown(true)
+
+const isAppUserSchema = (values: unknown): values is AppUserSchema => {
+  const { error } = AppUserSchemaSchema.validate(values)
+  return _.isUndefined(error)
+}
```
shir-reifenberg, line 54: please use `createSchemeGuard`

shir-reifenberg (commented):
Looks good 😎 
Didn't review the entire PR, added mainly nits and one comment about something that is probably not working after the fetch upgrade

shir-reifenberg (commented):
adapter-components changes LGTM 😎 leaving the rest to @idoamit8

idoamit8 (approved):
LGTM! 💯 
NICLEY DONE 🍏

PR #5911, opened 2024-05-16
SALTO-5935: Return empty results for fix elements when provided element list is empty

shir-reifenberg:
Fixed a bug which caused `fixElements` to throw when element list was empty

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/core/src/api.ts
+++ b/packages/core/src/api.ts
@@ -792,6 +792,7 @@ export const fixElements = async (
   workspace: Workspace,
   selectors: ElementSelector[],
 ): Promise<{ errors: ChangeError[]; changes: DetailedChange[] }> => {
+  log.trace('Fixing elements by the following selectors: %o', selectors.map(selector => selector.origin))
   const accounts = workspace.accounts()
   const adapters = await getAdapters(
     accounts,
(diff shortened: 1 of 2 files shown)


packages/core/src/api.ts
```diff
@@ -792,6 +792,7 @@ export const fixElements = async (
   workspace: Workspace,
   selectors: ElementSelector[],
 ): Promise<{ errors: ChangeError[]; changes: DetailedChange[] }> => {
+  log.trace('Fixing elements by the following selectors: %o', selectors.map(selector => selector.origin))
```
ori-moisis, line 795: I think this can be a debug level log.
I do think though that it would be more useful if we put it after applying the selectors, so we can log both the selectors and the relevant IDs

packages/core/src/api.ts
```diff
@@ -821,6 +822,10 @@ export const fixElements = async (
     .filter(values.isDefined)
     .toArray()
 
+  if (_.isEmpty(elements)) {
```
ori-moisis, line 825: maybe add a debug log here as well?

ori-moisis (approved):

PR #5912, opened 2024-05-17
SALTO-5346 - Salesforce: Implement fixElements for permission sets

tomsellek:
Implement `fixElements` for permission sets by reusing the code from profiles. Move common code to a shared utils file. Add unit tests and refactor the profiles tests a bit.

---

---
_Release Notes_: 
N/A

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/custom_references/profile_permission_set_utils.ts
+++ b/packages/salesforce-adapter/src/custom_references/profile_permission_set_utils.ts
@@ -1,0 +1,350 @@
+/*
+ *                      Copyright 2024 Salto Labs Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import _ from 'lodash'
+import { collections, promises } from '@salto-io/lowerdash'
+import {
+  ChangeError,
+  ElemID,
+  FixElementsFunc,
+  InstanceElement,
+  ReadOnlyElementsSource,
+  Values,
+} from '@salto-io/adapter-api'
+import { logger } from '@salto-io/logging'
+import { Types } from '../transformers/transformer'
+import {
+  APEX_CLASS_METADATA_TYPE,
+  APEX_PAGE_METADATA_TYPE,
+  API_NAME_SEPARATOR,
+  CUSTOM_APPLICATION_METADATA_TYPE,
+  FLOW_METADATA_TYPE,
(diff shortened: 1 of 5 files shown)


packages/salesforce-adapter/src/custom_references/profile_permission_set_utils.ts
```diff
@@ -0,0 +1,350 @@
...
+  sectionEntryKey: string,
+) => ReferenceInSection[]
+
+type ReferenceFromSectionParams = {
+  filter?: (sectionEntry: Values) => boolean
+  targetsGetter: RefTargetsGetter
+}
+
+const mapSectionEntries = <T>(
+  profile: InstanceElement,
```
yelly, line 77: This isn't necessarily a profile.

packages/salesforce-adapter/src/custom_references/profile_permission_set_utils.ts
```diff
@@ -0,0 +1,350 @@
...
+  objectAccessSectionEntry: Values,
+): boolean =>
+  [
+    objectAccessSectionEntry.allowCreate,
+    objectAccessSectionEntry.allowDelete,
+    objectAccessSectionEntry.allowEdit,
+    objectAccessSectionEntry.allowRead,
+    objectAccessSectionEntry.modifyAllRecords,
+    objectAccessSectionEntry.viewAllRecords,
+  ].some((permission) => permission === true)
```
yelly, line 116: Not sure what is considered better style, but this could just use the identity here:
```suggestion
  ].some((x) => x)
```
Actually it's a bit weird to me that you can't call `some` with no argument and get this behaviour, but that's JS for you. Maybe I'll write a proposal.

packages/salesforce-adapter/src/custom_references/profile_permission_set_utils.ts
```diff
@@ -0,0 +1,350 @@
...
+      sectionName as section,
+      params,
+      _.curry(f)(sectionName),
+    ),
+  )
+
+const profileEntriesTargets = (
+  profile: InstanceElement,
+): _.Dictionary<ElemID> =>
+  _(
```
yelly, line 279: Maybe use `Object.fromEntries` instead of going through lodash here?

yelly (approved):
Just some small comments, some of them may well be on code that I originally wrote.

packages/salesforce-adapter/src/custom_references/profile_permission_set_utils.ts
tomsellek: Ack

packages/salesforce-adapter/src/custom_references/profile_permission_set_utils.ts
tomsellek: For my part I'd rather be explicit about this sort of thing. I wonder about the semantic difference, though - what do we want to happen if there's a bug and any of these fields contain something that's not a boolean?

packages/salesforce-adapter/src/custom_references/profile_permission_set_utils.ts
tomsellek: Sure, but note the weird thing we do here - `mapProfileOrPermissionSetSections` returns the result of `Object.entries.flatMap` only for the calling function to use `Object.fromEntries` on the result. I wonder if the whole thing could be replaced with a call to `_.mapValues`

PR #5914, opened 2024-05-19
SALTO-5945: add references between workflowV2 action to Resolution and Priority

shayc331:
_add references between workflowV2 action to Resolution and Priority_

---

before:
Screen Shot 2024-05-19 at 11 16 17

after:
Screen Shot 2024-05-19 at 11 12 16

---
_Release Notes_: 
Jira Adapter:
* add a reference from WorkflowConfiguration (new Workflow API) action to Resolution and Priority

---
_User Notifications_: 
_None_


IdoZakk (approved):
💣 💣 💣 
I think we should add a release note so we have visibility that it entered

PR #5916, opened 2024-05-19
remove duplicate filter calls

dantal4:
Remove duplicate filters, that were there probably because of a rebase mistake

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


SeggevHaimovich (approved):
You are right, my mistake

PR #5919, opened 2024-05-20
SALTO-5948: Allow throwing an error during the DAG walk

shoshana-houri:
Currently, if we throw an error during the DAG walk, the error isn’t thrown immediately but is instead caught and only thrown as a combined error (containing all the errors that occurred) once the graph walk is complete. 
We want to optimize this behavior and allow certain errors to be thrown before the graph walk is complete. 

To achieve this, we have exposed a `FatalError` error class. If the handler throws such an error, then we throw and exit during the walk. It's important to note that for `walkAsync`, jobs that have already started will not be interrupted; however, the function will return with an error.

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/dag/src/nodemap.ts
+++ b/packages/dag/src/nodemap.ts
@@ -40,6 +40,14 @@ export class NodeSkippedError extends Error {
   }
 }
 
+// Errors that should be thrown *during* the walk. (The default behavior is to finish the walk and only then throw the errors that were encountered).
+// Please note that async jobs that have already started will not be interrupted; however, the function will return with an error.
+export class FatalError extends Error {
+  constructor(message: string) {
+    super(message)
+  }
+}
+
 export class WalkError extends Error {
   readonly handlerErrors: ReadonlyMap<NodeId, Error>
   readonly circularDependencyError?: CircularDependencyError
@@ -86,6 +94,10 @@ class WalkErrors<T> extends Map<NodeId, Error> {
     }
     visited.add(idAsString)
     super.set(nodeId, value)
+    if (value instanceof FatalError) {
+      this.nodeMap.forEach((_, id) => this.set(id, new NodeSkippedError(nodeId), visited))
+      throw new WalkError(this, undefined)
+    }
     this.nodeMap
       .getReverse(nodeId)
       .forEach(dependentNode => this.set(dependentNode, new NodeSkippedError(nodeId), visited))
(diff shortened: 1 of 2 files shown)


packages/dag/src/nodemap.ts
```diff
@@ -401,10 +413,11 @@ export class DAG<T> extends DataNodeMap<T> {
       wu(this.freeNodes(affectedNodes)).forEach(node => {
         try {
           handler(node)
-          next(this.deleteNode(node))
         } catch (e) {
           errors.set(node, e)
+          return
```
shoshana-houri, line 418: Please note that this change now aligns the behavior of the synchronous walk with that of the asynchronous walk
For the async walk we have:
```
          handler(node).then(
            () => next(this.deleteNode(node)),
            e => {
              errors.set(node, e)
            },
          ),
```

This means that any error occurring in this line `next(this.deleteNode(node))` will not be caught by the parent.

packages/dag/test/nodemap.test.ts
```diff
@@ -885,6 +913,42 @@ describe('NodeMap', () => {
...
+              error = e
+            })
+        })
+
+        it('should throw before all promises are resolved and add NodeSkippedError for remaining nodes', () => {
+          expect(error).toBeDefined()
+          expect(error).toBeInstanceOf(WalkError)
+
+          const errors = (error as WalkError).handlerErrors
+          expect(errors.size).toBe(3)
```
ori-moisis, line 943: I don't know if it is correct to force the number of errors to be exactly 3.
I think the behavior on all the "unrelated" nodes (1, 2 and 5 in this case) is really undefined, they could potentially get a `NodeSkippedError` and it would be a valid result, right?

shoshana-houri: Agreed :) thanks for noticing

packages/dag/test/nodemap.test.ts
```diff
@@ -643,6 +644,50 @@ describe('NodeMap', () => {
         })
       })
 
+      describe('when the handler throws a FatalError', () => {
+        const handlerMock = jest.fn()
```
ori-moisis, line 648: you should not reuse a mock function between tests, it retains the history from previous tests (we even have a [blog post](https://www.salto.io/blog-posts/typescript-unit-testing-pitfalls-with-jest-and-how-to-work-around-them) about it, see the `Sharing context through mock functions` section)

the `jest.fn()` should move into the `beforeEach`

ori-moisis (approved):

PR #5925, opened 2024-05-20
SALTO-5956: Change getInstancesFromElementSource to use .list() instead of .getAll()

shir-reifenberg:
change getInstancesFromElementSource utility not to use elementSource.getAll()

---

after discussion with @ori-moisis, and as the common usecase in code is getting elements of specific type, we believe using elementSource.list() and multiple elementSource.get() will have better performance.

tested locally on an account with 1000~ elements and a few validators and seems slightly better results (but I only tested with a small amount of changes)

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/adapter-utils/test/utils.test.ts
```diff
@@ -2734,4 +2735,25 @@ describe('Test utils.ts', () => {
...
+    it('should return elements of type from element source', async () => {
+      const type = new ObjectType({ elemID: new ElemID('salto', 'type') })
+      const instA = new InstanceElement('instA', type)
+      const instB = new InstanceElement('instB', type)
+      const source = buildElementsSourceFromElements([instA, instB, type])
+      const res = await getInstancesFromElementSource(source, ['type'])
+      expect(res).toHaveLength(2)
+      expect(res).toEqual([instA, instB])
+    })
+    it('should use elementSource.list() and not elementSource.getAll()', async () => {
```
ori-moisis, line 2748: tests should not enforce a specific implementation, they should test functionality.
if for some reason we decide in the future that a different implementation is more efficient, we should not have to handle broken tests...
generally when an implementation changes to be more efficient but does not change anything in functionality, I would not expect to see any change in the tests.

it seems like you added the first tests for this function here, which is good (well, bad for the original author...)
while you are at it, maybe add a "negative" test as well? (where we show that the function does not return instances from types we did not request)

PR #5928, opened 2024-05-21
SALTO-5761: Upgrade non guide types to be fetchable using the new infra

DanielBarros1:
In this PR, i add the ability to getElements using the new infra. This fetch only applies to non-guide types (103 types), as we decided to leave the guide up for a followup PR;

Important stuff to know:
1. The flag for using the new infra is `useNewInfra`. It is a config flag, and is false by default. It's not documented anywhere so users wouldn't know it exists.
2. The purpose of having both fetch methods is for further testing - ideally, we would like to try testing on a staging or even prod environment to make doubly sure that we don't break anything by switching to the new infra.
3. There are 2 live issues that still need to be dealt with, which create a divergence between the old infra and the new one. Not super sure if we would consider these blocking - my gut feeling is that SALTO-5934 is blocking and the other issue isn't. 
4. There were a couple of decision made along the way w.r to some types:
     a. Some types were not implemented, since they are either unused or just plain broken in the current infra
     b. Some types have inconsistencies w.r to how they are named. For example, `api_token`, which is an optional type, is inconsistent in naming when the field used as elemID is not defined. This will only be an issue if the user cleans their workspace between fetches.
     c. If there are multiple elements within the same `.nacl` file, and you clean your workspace between fetches, the order of the elements within the file might be inconsistent.
     d. you can read more about these issues in the notion link in the Ticket, and in the list of types linked above.
5. In this PR there are also a couple of changes to `adapter-components` code; This was necessary in order to enable filtering out inactive elements, and make some fixes to the order of the elements list.

In terms of next steps, here is the plan:
1. Once we're confident with the changes, we can kill the old infra fetch, and keep only the new one.
2. Enable using the new infra to fetch guide elements
3. Start Removing config code so that we only use definitions code going forward

Once these 3 steps are done, we can move on to the deploy, addition and removal part of the adapter, and then follow up with taking care of the filters code.

---

Here are some steps i took to test my changes:
1. For each type, fetch using the old infra, commit the workspace, fetch using the new infra, and observe that there are no changes in before/after. This includes optional types ( => types that are not in `SupportedTypes`) 
2. Once all non-guide types were implemented, fetch them all at once and follow the same steps
3. Fetch vs main to make sure i didn't break anything in the old infra
4. Fetch with Guide elements to make sure it still works
5. Deploy, add and remove some elements to make sure other core usecases are not broken

All tests were performed on our test env, `d3v-salto2.zendesk.com`.

---
_Release Notes_: 
Zendesk:
Add option to fetch non-guide types using the new infra

---
_User Notifications_: 
None

[SALTO-5860]: https://salto-io.atlassian.net/browse/SALTO-5860?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SALTO-5934]: https://salto-io.atlassian.net/browse/SALTO-5934?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/zendesk-adapter/src/adapter.ts
```diff
@@ -553,6 +571,11 @@ export default class ZendeskAdapter implements AdapterOperations {
     if (!isGuideThemesEnabledInConfig) {
       keysToOmit.push(GUIDE_THEME_TYPE_NAME)
     }
+    return keysToOmit
+  }
+
+  private filterSupportedTypes(): Record<string, string[]> {
```
edenhassid, line 577: why did you split `filterSupportedTypes` to 2 functions?

packages/zendesk-adapter/src/adapter.ts
```diff
@@ -562,21 +585,45 @@ export default class ZendeskAdapter implements AdapterOperations {
...
+    } else {
+      defaultSubdomainResult = await fetchUtils.getElements({
+        adapterName: ZENDESK,
+        fetchQuery: this.fetchQuery,
+        getElemIdFunc: this.getElemIdFunc,
+        definitions: this.adapterDefinitions,
+        customItemFilter: filterOutInactiveItemForType(this.userConfig),
+      })
+      if (!isGuideInFetch) {
+        addRemainingTypes({
```
edenhassid, line 618: not sure i understand why you add this here, we didn't have this in the old infra. what am I missing?

packages/zendesk-adapter/src/definitions/deploy/deploy.ts
edenhassid, line 1: should we have deploy in this pr?

packages/zendesk-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -0,0 +1,63 @@
...
+  >({
+    group: { bulkPath: '/api/v2/groups', nestUnderField: 'group' },
+  })
+  const customDefinitions: Record<string, Partial<InstanceDeployApiDefinitions>> = {
+    business_hours_schedule: getBusinessHoursScheduleDefinition(),
+  }
+  return _.merge(standardRequestDefinitions, customDefinitions)
+}
+
+export const createDeployDefinitions = (): definitions.deploy.DeployApiDefinitions<never, ClientOptions> => ({
```
edenhassid, line 38: I think we should make this empty for starters no?

packages/zendesk-adapter/src/definitions/deploy/business_hours_schedule.ts
edenhassid, line 1: do we need this file in this pr?

packages/zendesk-adapter/src/definitions/fetch/fetch.ts
```diff
@@ -0,0 +1,1359 @@
...
+  view__restriction: {
+    element: {
+      fieldCustomizations: {
+        id: { fieldType: 'unknown' },
+        type: { fieldType: 'string', restrictions: { enforce_value: true, values: ['Group', 'User'] } },
+      },
+    },
+  },
+
+  view_order: {},
```
edenhassid, line 147: can you add a comment why its empty?

packages/zendesk-adapter/src/definitions/fetch/fetch.ts
```diff
@@ -0,0 +1,1359 @@
...
+  workspace__selected_macros: {
+    element: { fieldCustomizations: { usage_7d: { omit: true } } },
+  },
+
+  workspace__selected_macros__restriction: {
+    element: { fieldCustomizations: { id: { fieldType: 'unknown' } } },
+  },
+
+  workspace__apps: {
+    element: {},
```
edenhassid, line 1238: why add element if its empty?

packages/zendesk-adapter/src/definitions/fetch/transforms/business_hours_schedule.ts
```diff
@@ -0,0 +1,33 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import _ from 'lodash'
+import { definitions } from '@salto-io/adapter-components'
+import { values as lowerdashValues } from '@salto-io/lowerdash'
+
+export const transform: definitions.AdjustFunction = ({ value }) => {
```
edenhassid, line 20: I think we should add this code only when we remove the current code from the filter

packages/zendesk-adapter/src/inactive.ts
```diff
@@ -15,6 +15,7 @@
  */
 import { InstanceElement } from '@salto-io/adapter-api'
 import { definitions } from '@salto-io/adapter-components'
+import { ValueGeneratedItem } from '@salto-io/adapter-components/src/fetch/types'
```
edenhassid, line 18: the import should not look like this, you should add this to the `adapter-components` index

packages/zendesk-adapter/src/inactive.ts
```diff
@@ -47,3 +48,27 @@ export const filterOutInactiveInstancesForType = (
     return instances.filter(instance => instance.value.active !== false)
   }
 }
+
+/**
+ * Same as above, but for the new infra. The above function is removable after the migration to the new
+ * infra is complete (SALTO-5760)
+ */
+export const filterOutInactiveItemForType = (config: ZendeskConfig): ((item: ValueGeneratedItem) => boolean) => {
```
edenhassid, line 56: can you add a test for this?

edenhassid (commented):
nice!! this looks great! 💯 🏆 🤩  added some small comments :)

packages/zendesk-adapter/src/adapter.ts
DanielBarros1: I needed the types that are unsupported and need to be omited, so i split it into its own function instead of duplicating code :)

packages/zendesk-adapter/src/adapter.ts
DanielBarros1: It happens inside the `getAllElements`, and in the new infra it doesn't. Check it out [here](https://github.com/salto-io/salto/blob/main/packages/adapter-components/src/elements_deprecated/ducktype/transformer.ts#L473)

packages/zendesk-adapter/src/definitions/deploy/deploy.ts
DanielBarros1: i'll remove!

packages/zendesk-adapter/src/definitions/deploy/deploy.ts
DanielBarros1: removing this file completely

packages/zendesk-adapter/src/definitions/deploy/business_hours_schedule.ts
DanielBarros1: removing

packages/zendesk-adapter/src/definitions/fetch/transforms/business_hours_schedule.ts
DanielBarros1: will remove

edenhassid (commented):
LGTM! 🔥 waiting for @netama to give the final approve

packages/adapter-components/src/fetch/resource/resource_manager.ts
```diff
@@ -95,6 +98,8 @@ export const createResourceManager = <ClientOptions extends string>({
             fetcher => fetcher.getItems(),
           )
 
+          // TODO wrap in try-catch and support turning into config suggestions or fetch warnings (SALTO-5427)
```
netama, line 101: SALTO-5427 is now done :) is there something left?

packages/adapter-components/src/fetch/fetch.ts
```diff
@@ -53,13 +53,15 @@ export const getElements = async <Options extends APIDefinitionsOptions>({
   predefinedTypes,
   getElemIdFunc,
   additionalRequestContext,
+  customItemFilter,
```
netama, line 56: discussed offline - reviewing here but this will be better as a separate PR :)

packages/adapter-components/src/fetch/resource/type_fetcher.ts
```diff
@@ -144,17 +146,20 @@ export const createTypeResourceFetcher = <ClientOptions extends string>({
...
-            [fieldName]: Array.isArray(fieldItems) ? fieldItems.map(({ value }) => value) : fieldItems.value,
-          }))
-          return {
-            ...item,
-            value: _.defaults({}, item.value, ...fieldValues),
-            context: item.context,
-          }
-        }),
+        itemsWithContext
+          .filter(item => (customItemFilter === undefined ? true : customItemFilter(item)))
```
netama, line 150: tiny nit: maybe just create a filter before this and use it here (so that we don't run the `if` so many times)? can use `() => true` as the "fallback" function

packages/adapter-components/src/fetch/resource/type_fetcher.ts
```diff
@@ -144,17 +146,20 @@ export const createTypeResourceFetcher = <ClientOptions extends string>({
...
+          .map(async item => {
+            const nestedResources = await recurseIntoFetcher(item)
+            const fieldValues = Object.entries(nestedResources).map(([fieldName, fieldItems]) => ({
+              [fieldName]: Array.isArray(fieldItems) ? fieldItems.map(({ value }) => value) : fieldItems.value,
+            }))
+            const a = _.defaults({}, item.value, ...fieldValues)
```
netama, line 156: please use meaningful names for consts/variables. did anything change in this line? or was this just for debugging and can just move back to `value`?

packages/zendesk-adapter/src/adapter.ts
```diff
@@ -502,6 +507,19 @@ export default class ZendeskAdapter implements AdapterOperations {
...
 
+    const otherDefinitions = {
+      clients: createClientDefinitions({ main: this.client }),
+      pagination: PAGINATION,
+      fetch: createFetchDefinitions(this.userConfig, this.nonSupportedTypesToOmit()),
+    }
+
+    this.adapterDefinitions = {
+      ...otherDefinitions,
+      fetch: definitions.mergeWithUserElemIDDefinitions({
```
netama, line 518: doesn't this override the `fetch` from `otherDefinitions`? what am I missing?

packages/zendesk-adapter/src/adapter.ts
```diff
@@ -544,7 +562,7 @@ export default class ZendeskAdapter implements AdapterOperations {
     )
   }
 
-  private filterSupportedTypes(): Record<string, string[]> {
+  private nonSupportedTypesToOmit(): string[] {
```
netama, line 565: naming nit: missing a verb here :)

packages/zendesk-adapter/src/definitions/fetch/fetch.ts
netama: not sure I understand :) e.g. shouldn't it have `isTopLevel = true`? and maybe some field customizations? (or - I think you can create it here and have it depend on the `view` type - but doesn't have to be in this PR)

netama (commented):
looks good :) didn't review all the definition changes, I assume @edenhassid did that and either way it should be covered in the tests.
let's split the adapter-components changes into a separate PR (but reviewed here)
and let's sync re the test and upgrade plan IRL?

netama (approved):
pre-approving as most of the remaining comments are nits + discussions

packages/adapter-components/src/fetch/resource/type_fetcher.ts
DanielBarros1: Totally just debugging :P will remove

packages/adapter-components/src/fetch/fetch.ts
DanielBarros1: [new pr](https://github.com/salto-io/salto/pull/5979)

packages/adapter-components/src/fetch/resource/type_fetcher.ts
DanielBarros1: solved in [new pr](https://github.com/salto-io/salto/pull/5979)

packages/adapter-components/src/fetch/resource/type_fetcher.ts
DanielBarros1: removed in [new pr](https://github.com/salto-io/salto/pull/5979)

PR #5929, opened 2024-05-21
SALTO-5965: Pass spread request context to extractor context

shir-reifenberg:
we need to spread context otherwise real context will be nested twice under context

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


netama (approved):
thanks

PR #5930, opened 2024-05-21
SALTO-5958: Adding an SFDC filter that drops all generated dependencies pre deploy

yelly:
They aren't used in the deploy and just seem to create bugs when resolving references.

---

---
_Release Notes_: 
None.

---
_User Notifications_: 
None.

### DIFF
--- a/packages/adapter-api/src/change.ts
+++ b/packages/adapter-api/src/change.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 import { AdditionDiff, ModificationDiff, RemovalDiff, ActionName } from '@salto-io/dag'
-import { values as lowerDashValues } from '@salto-io/lowerdash'
+import { values as lowerDashValues, types as lowerDashTypes } from '@salto-io/lowerdash'
 import {
   ObjectType,
   InstanceElement,
@@ -37,6 +37,7 @@ export type AdditionChange<T> = AdditionDiff<T>
 export type ModificationChange<T> = ModificationDiff<T>
 export type RemovalChange<T> = RemovalDiff<T>
 export type Change<T = ChangeDataType> = AdditionChange<T> | ModificationChange<T> | RemovalChange<T>
+export type ChangeDataKeys = lowerDashTypes.ExtractKeys<Change['data']>
 
 export type ChangeData<T extends Change<unknown>> = T extends Change<infer U> ? U : never
 export const isModificationChange = <T extends Change<unknown>>(
--- a/packages/adapter-api/src/change.ts
+++ b/packages/adapter-api/src/change.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 import { AdditionDiff, ModificationDiff, RemovalDiff, ActionName } from '@salto-io/dag'
-import { values as lowerDashValues, types as lowerDashTypes } from '@salto-io/lowerdash'
+import { values as lowerDashValues } from '@salto-io/lowerdash'
 import {
   ObjectType,
   InstanceElement,
@@ -37,7 +37,6 @@ export type AdditionChange<T> = AdditionDiff<T>
 export type ModificationChange<T> = ModificationDiff<T>
 export type RemovalChange<T> = RemovalDiff<T>
 export type Change<T = ChangeDataType> = AdditionChange<T> | ModificationChange<T> | RemovalChange<T>
-export type ChangeDataKeys = lowerDashTypes.ExtractKeys<Change['data']>
 
 export type ChangeData<T extends Change<unknown>> = T extends Change<infer U> ? U : never
 export const isModificationChange = <T extends Change<unknown>>(
--- a/packages/salesforce-adapter/src/adapter.ts
+++ b/packages/salesforce-adapter/src/adapter.ts
@@ -115,6 +115,7 @@ import centralizeTrackingInfoFilter from './filters/centralize_tracking_info'
 import changedAtSingletonFilter from './filters/changed_at_singleton'
 import importantValuesFilter from './filters/important_values_filter'
 import omitStandardFieldsNonDeployableValuesFilter from './filters/omit_standard_fields_non_deployable_values'
+import generatedDependenciesFilter from './filters/generated_dependencies'
 import {
   CUSTOM_REFS_CONFIG,
   FetchElements,
@@ -280,6 +281,7 @@ export const allFilters: Array<
   { creator: hideTypesFolder },
   // createChangedAtSingletonInstanceFilter should run last
   { creator: changedAtSingletonFilter },
+  { creator: generatedDependenciesFilter },
 ]
 
 // By default we run all filters and provide a client
--- a/packages/salesforce-adapter/src/filters/generated_dependencies.ts
+++ b/packages/salesforce-adapter/src/filters/generated_dependencies.ts
@@ -15,55 +15,60 @@
  */
 import {
   ChangeDataType,
-  ChangeDataKeys,
-  changeId,
   CORE_ANNOTATIONS,
+  getChangeData,
+  isAdditionOrModificationChange,
 } from '@salto-io/adapter-api'
 import { DetailedDependency } from '@salto-io/adapter-utils'
 import { LocalFilterCreator } from '../filter'
+
+type ElementWithGeneratedDependencies = ChangeDataType & {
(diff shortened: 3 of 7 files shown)


packages/adapter-api/src/change.ts
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 import { AdditionDiff, ModificationDiff, RemovalDiff, ActionName } from '@salto-io/dag'
-import { values as lowerDashValues } from '@salto-io/lowerdash'
+import { values as lowerDashValues, types as lowerDashTypes } from '@salto-io/lowerdash'
```
tamtamirr, line 17: Nit: Is the as required here?

packages/salesforce-adapter/src/filters/generated_dependencies.ts
```diff
@@ -0,0 +1,71 @@
...
+    Record<ChangeDataKeys, DetailedDependency[]>
+  >
+
+  const popGeneratedDependencies = (
+    element?: ChangeDataType,
+  ): DetailedDependency[] | undefined => {
+    const deps = element?.annotations[CORE_ANNOTATIONS.GENERATED_DEPENDENCIES]
+    delete element?.annotations[CORE_ANNOTATIONS.GENERATED_DEPENDENCIES]
+    return deps
+  }
```
tamtamirr, line 40: Please note that we are dealing with `any` here. The return type is actually lying and might cause issues (as the code assumes we are dealing with `DetailedDependency[]` but it's never forced).

I suggest the following approach:

- Add type `ElementWithGeneratedDependencies`
```
type ElementWithGeneratredDependencies = Element & {
  annotations: Element['annotations'] & {
    [CORE_ANNOTATIONS.GENERATED_DEPENDENCIES]: DetailedDependency[]
 }
}
```

- Add a typeguard and filter the Elements by it. 
- I think `popGeneratedDependencies` is redundant and can be replaced with forEach
```
.forEach(element => {
  generatedDependencies[element.id.getFullName()] = element.annotations[CORE_ANNOTATIONS.GENERATED_DEPENDENCIES]
  delete element.annotations[CORE_ANNOTATIONS.GENERATED_DEPENDENCIES]
})
```

packages/salesforce-adapter/src/filters/generated_dependencies.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  return {
+    name: 'generatedDependencies',
+    preDeploy: async (changes) => {
+      generatedDependencies = Object.fromEntries(
+        changes.map((change) => [
```
tamtamirr, line 46: We should only handle addition / modification changes here

packages/salesforce-adapter/test/filters/generated_dependencies.test.ts
```diff
@@ -0,0 +1,152 @@
...
+      expect(preDeployChanges).toEqual(generateChangesNoGeneratedDependencies())
+    })
+  })
+
+  describe('onDeploy', () => {
+    it('should restore generated dependencies', async () => {
+      const onDeployChanges = await runOnDeploy(...preDeployChanges)
+      expect(onDeployChanges).toEqual(generateChanges())
+    })
+  })
```
tamtamirr, line 151: Nit: I prefer testing `preDeploy and onDeploy` together as they are heavy reliant (Your beforeAll scope implies that).
Consider using beforeEach and have a single test case `preDeploy & onDeploy flow`

packages/salesforce-adapter/test/filters/generated_dependencies.test.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  const runPreDeploy = async (...changes: Change[]): Promise<Change[]> => {
+    await filter.preDeploy(changes)
+    return changes
+  }
+
+  const runOnDeploy = async (...changes: Change[]): Promise<Change[]> => {
+    await filter.onDeploy(changes)
+    return changes
+  }
```
tamtamirr, line 133: These are not needed IMO. The test can invoke `preDeploy` and `onDeploy` directly, which will modify the passed changes in place.

packages/salesforce-adapter/test/filters/generated_dependencies.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    )
+
+    return [
+      toChange({ before: typeBefore, after: typeAfter }),
+      toChange({ after: instanceToAdd }),
+      toChange({ before: instanceToRemove }),
+    ]
+  }
+
+  const generateChangesNoGeneratedDependencies = (): Change[] => {
```
tamtamirr, line 113:
```suggestion
  const createChangesWithNoGeneratedDependencies = (): Change[] => {
```
nit

packages/salesforce-adapter/src/filters/generated_dependencies.ts
```diff
@@ -0,0 +1,71 @@
...
+import { DetailedDependency } from '@salto-io/adapter-utils'
+import { LocalFilterCreator } from '../filter'
+
+/**
+ * Remove generated dependencies before deploy.
+ */
+const filterCreator: LocalFilterCreator = () => {
+  let generatedDependencies: Record<
+    string,
+    Record<ChangeDataKeys, DetailedDependency[]>
```
tamtamirr, line 31: This is overly complicated and unnecessary to save both the `before / after` for each change.
We should only care about the `after` here, and not modify the `before` as this is meaningless and we will receive the same result. We can discuss that F2F as-well as this might come off as not very intuitive :)

tamtamirr (commented):
Looking good! Added some comments :)

packages/salesforce-adapter/src/filters/generated_dependencies.ts
yelly: Done following F2F.

packages/salesforce-adapter/src/filters/generated_dependencies.ts
yelly: Did most of what you said. I still think the full assignment is better because there might be some weird edge cases if the filter is called twice so I prefer to just replace all of `generatedDependencies` at once and not worry about it.

packages/salesforce-adapter/src/filters/generated_dependencies.ts
yelly: Done.

packages/salesforce-adapter/src/filters/generated_dependencies.ts
```diff
@@ -0,0 +1,76 @@
...
+  isAdditionOrModificationChange,
+} from '@salto-io/adapter-api'
+import { DetailedDependency } from '@salto-io/adapter-utils'
+import { LocalFilterCreator } from '../filter'
+
+type ElementWithGeneratedDependencies = ChangeDataType & {
+  annotations: ChangeDataType['annotations'] & {
+    [CORE_ANNOTATIONS.GENERATED_DEPENDENCIES]: DetailedDependency[]
+  }
+}
```
tamtamirr, line 29: Nit: It feels very off to use `ChangeDataType` and `Element` in the same terminology.
Can we use `Element` in the type definition instead to avoid confusion?

tamtamirr (approved):
LGTM 👑 
One small nit. Approving to save cycles

packages/salesforce-adapter/src/filters/generated_dependencies.ts
yelly: `ChangeDataType` is more specific than `Element`, so this won't work.

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -280,6 +281,7 @@ export const allFilters: Array<
   { creator: hideTypesFolder },
   // createChangedAtSingletonInstanceFilter should run last
   { creator: changedAtSingletonFilter },
+  { creator: generatedDependenciesFilter },
```
adamro, line 284: How come we are ignoring the comment that `changedAtSingletonFilter` should run last? 🤔

yelly: Fair, I'll move it to be one before. It doesn't matter in this case but you're right that we should respect the comment.

yelly: https://github.com/salto-io/salto/pull/5941

PR #5931, opened 2024-05-21
Salto 2242 Only run e2es if they or the packages they depend on have changed

benisraeladi:
When no changes in packages are found: https://app.circleci.com/pipelines/github/salto-io/salto/26792/workflows/253e2fd0-10ad-43fe-941d-e069eb78ce3b

Changes in Okta: https://app.circleci.com/pipelines/github/salto-io/salto/26696/workflows/3faab7ad-9b79-4670-8e3e-e86afbad6026

Changes in salesforce-adapter, Okta, and adapter-components: https://app.circleci.com/pipelines/github/salto-io/salto/26699/workflows/e52fdeb9-56f1-4156-870d-fca75fd9fa74

---

---
_Release Notes_: 

---
_User Notifications_: 


eshedg:
the circleci pipeline lgtm.

eshedg (approved):

.circleci/config_template.yml
ori-moisis, line 1: there seem to be a lot of differences between this and the current config.
I did not expect there to be so many differences, I expected that only the last part (which tasks need to run) would change.
are all the other changes on purpose?

.circleci/scripts/update_config_template
ori-moisis, line 1: is there a reason why this isn't one script with the `find_changed_packages`?
the implementation seems to be very tightly coupled

.circleci/scripts/update_config_template
```diff
@@ -0,0 +1,35 @@
+#! /usr/bin/env node
+
+const { execSync } = require('child_process')
+const { readFileSync, writeFileSync, existsSync, readdirSync } = require('fs')
+const path = require('path')
+
+
+function main() {
+  const continueConfigFilePath = path.join(__dirname, '..', 'continue_config.yml')
+  const configTemplate = readFileSync(path.join(__dirname, '..', 'config_template.yml'), 'utf8')
```
ori-moisis, line 10: why create the path twice?

would be irrelevant if this is unified with the other script

.circleci/scripts/find_changed_packages
```diff
@@ -0,0 +1,133 @@
...
+
+  return Array.from(changedPackages)
+}
+
+function main() {
+  const args = process.argv.slice(2)
+  console.log(`User provided arguments: ${args}`)
+
+  const branchLocation = args.length > 0 ? args[0] : 'origin'
+  console.log(`attempting to use ${branchLocation} as main branch location`)
```
ori-moisis, line 90: doesn't seem like this is actually called with arguments from the config.yml, do it make sense?

.circleci/scripts/find_changed_packages
```diff
@@ -0,0 +1,133 @@
+#! /usr/bin/env node
+
+const { execSync } = require('child_process')
+const { readFileSync, writeFileSync, existsSync, readdirSync } = require('fs')
+const path = require('path')
+
+function getChangedFiles(branchLocation) {
+  const output = execSync(`git log --name-status ${branchLocation}/main..HEAD`).toString()
```
ori-moisis, line 8: IIUC this compares to the current main.
this would lead to false changes if a PR is not fully rebased.

I think there is a way to get the PR base from the environment so this can be done properly

.circleci/scripts/find_changed_packages
```diff
@@ -0,0 +1,133 @@
...
+  return output
+    .split('\n')
+    .filter(line => line.startsWith('M') || line.startsWith('A') || line.startsWith('D'))
+    .map(line => line.split('\t')[1])
```
ori-moisis, line 12: you can use `git diff <base>..<head> --name-only` to get just the names of the changed files
no need to do all this complicated parsing

.circleci/scripts/find_changed_packages
```diff
@@ -0,0 +1,133 @@
...
+    .filter(file => file && file.startsWith('packages/'))
```
ori-moisis, line 13: why filter only things under `packages`?
won't this miss the case where, for example, someone changes a build script or the main `package.json` which could potentially affect everything?

.circleci/scripts/find_changed_packages
```diff
@@ -0,0 +1,133 @@
...
+}
+
+function fetchWorkspacesInfo() {
+  const output = execSync('yarn workspaces info', { encoding: 'utf8' })
+  const jsonOutput = output.substring(output.indexOf('{'), output.lastIndexOf('}') + 1)
+  return JSON.parse(jsonOutput)
```
ori-moisis, line 19: should not have to do string manipulations, you can ask `yarn` to output a json

```suggestion
  const output = execSync('yarn --json workspaces info', { encoding: 'utf8' })
  return JSON.parse(JSON.parse(output).data)
```

.circleci/scripts/find_changed_packages
```diff
@@ -0,0 +1,133 @@
...
+
+function findChangedPackages(branchLocation) {
+  const changedFiles = getChangedFiles(branchLocation)
+  console.log('Changed files:', changedFiles)
+  const changedPackages = new Set()
+
+  for (const file of changedFiles) {
+    const parts = file.split('/')
+    if (parts.length > 1) {
+      changedPackages.add('@salto-io/' + parts[1])
```
ori-moisis, line 78: the mapping between package name and the path of the package exists in the output of `yarn workspaces info`, we should use that and not rely on heuristics.

specifically, this doesn't work correctly for one of the packages we have today:
```
  "salto-vscode": {
    "location": "packages/vscode",
    "workspaceDependencies": [
      "@salto-io/core",
      "@salto-io/file",
      "@salto-io/lang-server",
      "@salto-io/lowerdash",
      "@salto-io/parser",
      "@salto-io/workspace"
    ],
    "mismatchedWorkspaceDependencies": []
  },
```

.circleci/scripts/find_changed_packages
```diff
@@ -0,0 +1,133 @@
...
+function getChangedFiles(branchLocation) {
+  const output = execSync(`git log --name-status ${branchLocation}/main..HEAD`).toString()
+  return output
+    .split('\n')
+    .filter(line => line.startsWith('M') || line.startsWith('A') || line.startsWith('D'))
+    .map(line => line.split('\t')[1])
+    .filter(file => file && file.startsWith('packages/'))
+}
+
+function fetchWorkspacesInfo() {
```
ori-moisis, line 16: nit: why `fetch`? why not `get` like in `getChangedFiles`?

.circleci/scripts/find_changed_packages
```diff
@@ -0,0 +1,133 @@
...
+      resolveTransitiveDependencies(dep, workspaceInfo, resolved, seen)
+    }
+  }
+}
+
+function generateDependencyGraph(workspaceInfo) {
+  const graph = {}
+  const invertedGraph = {}
+
+  for (const packageName of Object.keys(workspaceInfo)) {
```
ori-moisis, line 45: it seems like both scripts are written completely different from our coding conventions (using named functions, using explicit loops, using side effects instead of return values, etc)

is there a reason for this?
if not, please align this with our coding conventions

.circleci/scripts/find_changed_packages
```diff
@@ -0,0 +1,133 @@
...
+    if (!resolved.has(dep)) {
+      resolved.add(dep)
+      resolveTransitiveDependencies(dep, workspaceInfo, resolved, seen)
+    }
+  }
+}
+
+function generateDependencyGraph(workspaceInfo) {
+  const graph = {}
+  const invertedGraph = {}
```
ori-moisis, line 43: nit: technically this is a graph in the same sense that any mapping is a graph, but it is never actually used as a graph, this is just a mapping.

I think it would be clearer to give these mappings names according to their purpose and not just a generic "graph", so, `graph` is actually something like `dependenciesByPackage`.

.circleci/scripts/update_config_template
```diff
@@ -0,0 +1,35 @@
...
+      - test_e2e:
+          requires:
+            - build
+          matrix:
+            parameters:
+              package_name:
+                - ${e2ePackagesToTest.join('\n                - ')}`
+
+
+  let alteredConfigTemplate = e2ePackagesToTest.length > 0 ?
```
ori-moisis, line 25: why is this `let` and not `const`?
(another thing I'd expect the linter to catch)

.circleci/scripts/find_changed_packages
```diff
@@ -0,0 +1,133 @@
...
+      packagesToTest.add(pkg)
+      for (const dep of dependencyGraph[pkg]) {
+        packagesToTest.add(dep)
+      }
+    }
+  }
+
+  const packagesArray = Array.from(packagesToTest, pkg => pkg.replace('@salto-io/', '')).sort()
+  const e2ePackagesToTest = []
+  const utPackagesToTest = []
```
ori-moisis, line 119: unused?
I assume this would have been caught if this file was named properly (the linter should have caught that)

.circleci/scripts/find_changed_packages
```diff
@@ -0,0 +1,133 @@
...
+    // Otherwise, run tests for all changed packages and their dependencies
+    for (const pkg of changedPackages) {
+      packagesToTest.add(pkg)
+      for (const dep of dependencyGraph[pkg]) {
+        packagesToTest.add(dep)
+      }
+    }
+  }
+
+  const packagesArray = Array.from(packagesToTest, pkg => pkg.replace('@salto-io/', '')).sort()
```
ori-moisis, line 117: why replace again? didn't you already replace in line 99?

.circleci/scripts/find_changed_packages
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+
+  return invertedGraph
+}
+
+function hasE2eTests(packageName) {
+  const e2eDir = path.join(__dirname, '..', '..', 'packages', packageName, 'e2e_test')
+  console.log('e2eDir:', e2eDir, 'exists:', existsSync(e2eDir))
+  return existsSync(e2eDir) && readdirSync(e2eDir).some(file => file.endsWith('.test.ts'))
```
ori-moisis, line 67: the log might be misleading given that the existence of an e2e dir is not the only condition you are checking here.

not sure we need the extra test, but as long as they are there...
also, super nit - shouldn't all `existsSync` twice

.circleci/config.yml
```diff
@@ -1,102 +1,16 @@
...
+  find_changed_packages:
     docker:
       - image: cimg/node:14.15
-
-    resource_class: xlarge
-    working_directory: /mnt/ramdisk/project
+    resource_class: large
 
     steps:
       - restore_cache:
```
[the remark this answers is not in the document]
benisraeladi: It should, but to create the configuraiton I need the git repository. Having the cache means that we don't have to do everything over again in the build step, and we know there won't be changes between this step and the next

.circleci/config_template.yml
benisraeladi: These other changes are due to the fact that `config.yml` is now the file that creates the dynamic configuration, and the entire configuration (all of what is red) is now mostly in `config_template.yml`

.circleci/scripts/find_changed_packages
benisraeladi: This allows the script to be run locally without change, since locally we need to use `upstream/main` while in CircleCI we need `origin/main`. If we don't have this, the script fails in cryptic ways

.circleci/scripts/update_config_template
benisraeladi: This was a request by @eshedg so that the script can be taken and used in other places without being altered

.circleci/config_template.yml
ori-moisis: maybe it wasn't clear
I compared the current `config.yml` to `config_template.yml` in this PR

benisraeladi: The change is due to the fact that now I don't have a good way of knowing if we need to install Java or not for an e2e without having a list. Given that everything is dynamic, I would rather not have to maintain this list, and the additional 30 seconds in any single test is not as important imho

.circleci/scripts/find_changed_packages
benisraeladi: We only know the dependency between files under `packages`. It sounds weird to me to run all the e2es when I think that building and unit tests should probably be enough in the case of a build script change or a change in `package.json`

.circleci/scripts/find_changed_packages
benisraeladi: I still need to do complicated parsing because I need to differentiate between commit+message and file name, and the only way I found to do that was by using regex. I don't think that's better, but I don't have a strong opinion either way

.circleci/scripts/find_changed_packages
benisraeladi: @ori-moisis There is a way to get the PR base, but it requires the github auth token. Currently, we only use it in things that run on main (`publish_canary` and `publish_on_version_change`). I don't know the reasoning behind it, but maybe you might.
If this is okay, then I'll go ahead and add it (and then it should work). If not, I don't think there's a real problem with comparing to main. The worst case is that we run more tests than necessary, which might actually make people rebase more.

ori-moisis:
```bash
git diff --name-only $(git merge-base main HEAD) HEAD
```
I was a bit worried this won't work when there are merge commits from "main" into the branch, but I tested it a bit and it seems to work.
feel free to test it more, maybe you can find a case I missed where this doesn't work...

.circleci/scripts/find_changed_packages
ori-moisis: you are basically claiming that for the most fundamental changes possible (changing a dependency version, changing the build process itself) we don't need to run the tests that are specifically there for the purpose of catching integration issues.

I don't have a better response than to just say that is wrong.
so, unfortunately, even if it sounds weird to you, we must run all the e2e tests if we make such a change.

if any file changes outside of the packages folder, all tests must run.
not running a test when we should have is far more dangerous than running a test when we didn't have to

.circleci/config.yml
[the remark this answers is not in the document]
ori-moisis: I don't understand what you are trying to say here

are you sure that you need to install everything and restore the whole workspace cache and everything in order to run the two scripts that AFAICT don't have any requirements beyond having `node` in the container?

.circleci/scripts/update_config_template
ori-moisis: if they need to be reusable outside the context of this task (which, tbh, I am not sure why), I think there are a few changes that would need to be done - they make some assumptions about their runtime environment and directory structure that I am not sure are ok for a script that is meant to be reusable in different contexts.

@eshedg - what other places did you have in mind for running these scripts? which assumptions are valid and which are not in those other places?

.circleci/scripts/find_changed_packages
ori-moisis: what parsing?
that command just outputs file names, one file name per line

benisraeladi: The command outputs not only file names, but also the commit+message:
```
db7dd2bb7 Fixed condition
.circleci/scripts/find_changed_packages.js
9254a6ba0 Removed old files
.circleci/config.yml
.circleci/config_template.yml
.circleci/scripts/find_changed_packages
.circleci/scripts/update_config_template
```

So I have to differentiate between lines with commit messages and lines without. It's not a big deal, it just didn't make the code any simpler

ori-moisis: what command are you talking about?
I don't see any reason why git diff would output a commit message and it did not happen in any of my attempts to test this

.circleci/config.yml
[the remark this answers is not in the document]
benisraeladi: You're right, I didn't need to install. Deleted all this

.circleci/scripts/find_changed_packages
benisraeladi: I found the reason for our misunderstanding. Changed it and now it works

.circleci/scripts/find_changed_packages
benisraeladi: Fixed this, now it works correctly for all packages

.circleci/config.yml
```diff
@@ -1,476 +1,38 @@
...
 
 jobs:
-  build:
+  find_changed_packages:
     docker:
       - image: cimg/node:14.15
-
-    resource_class: xlarge
-    working_directory: /mnt/ramdisk/project
+    resource_class: large
```
ori-moisis, line 13: do we really need a large instance for this?

.circleci/config.yml
```diff
@@ -1,476 +1,38 @@
...
-                - main
+      - find_changed_packages
+      - continuation/continue:
           requires:
-            - build
-            - unit_test
-            - e2e_tests_without_java
-            - e2e_tests_with_java
+            - find_changed_packages
+          checkout: false
```
ori-moisis, line 36: why the `checkout: false` here and then a `checkout` step in the job?

.circleci/scripts/find_changed_packages.js
```diff
@@ -0,0 +1,120 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const { execSync, exec } = require('child_process')
+const { readFileSync, writeFileSync, existsSync, readdirSync } = require('fs')
+const path = require('path')
+
+const getChangedFiles = (userInputBaseCommit) => {
+  const commitHashRegex = /\b[0-9a-f]{9}/g
```
ori-moisis, line 22: unused

.circleci/scripts/find_changed_packages.js
```diff
@@ -0,0 +1,120 @@
...
+  const baseCommit = userInputBaseCommit ?? execSync('git merge-base main HEAD').toString().trim()
+  console.log('base commit:', baseCommit)
+  const output = execSync(`git diff --oneline --name-only ${baseCommit}..HEAD`).toString().split('\n')
+  console.log('git diff output:', output)
+  return output.filter(file => file && file.startsWith('packages/'))
```
ori-moisis, line 27: we should not filter out files at the root of the project
this will miss changes in files that should trigger all the tests

.circleci/scripts/find_changed_packages.js
```diff
@@ -0,0 +1,120 @@
...
+}
+
+const resolveTransitiveDependencies = (packageName, workspaceInfo) => {
+  const packageInfo = workspaceInfo[packageName]
+  if (!packageInfo) {
+    return
+  }
+
+  return packageInfo['workspaceDependencies'].map(dep => resolveTransitiveDependencies(dep, workspaceInfo))
+}
```
ori-moisis, line 42: unused

.circleci/scripts/find_changed_packages.js
```diff
@@ -0,0 +1,120 @@
...
+  )
+  .flat()
+  .reduce((acc, [dep, packageName]) => {
+    if (!acc[dep]) {
+      acc[dep] = []
+    }
+    acc[dep].push(packageName)
+    return acc
+  }, {})
+}
```
ori-moisis, line 58: I might be missing something, but does that actually get all the dependencies beyond the direct ones?

this seems like a bit of a strange implementation of `_.groupBy`, doesn't seem like it recurses into the dependencies as it should

.circleci/scripts/find_changed_packages.js
```diff
@@ -0,0 +1,120 @@
...
+  const e2eDirExists = existsSync(e2eDir)
+  const e2eDirHasTestFiles = e2eDirExists && readdirSync(e2eDir).some(file => file.endsWith('.test.ts'))
+  console.log('e2eDir:', e2eDir, 'exists:', e2eDirExists, 'hasTestFiles:', e2eDirHasTestFiles)
+  return e2eDirHasTestFiles
+}
+
+const findChangedPackages = (userInputBaseCommit, workspaceInfo) => {
+  const changedFiles = getChangedFiles(userInputBaseCommit)
+  console.log('Changed files:', changedFiles)
+  const changedPackageLocation = changedFiles.map(path => path.split('/').slice(0, 2).join('/')).filter((value, index, self) => self.indexOf(value) === index)
```
ori-moisis, line 71: what is the filter at the end here supposed to do?

.circleci/scripts/find_changed_packages.js
```diff
@@ -0,0 +1,120 @@
...
+
+  const changedPackages = changedPackageLocation.map(packageDir =>Object.keys(workspaceInfo).find(key => workspaceInfo[key].location === packageDir))
```
ori-moisis, line 73: any chance this file is not covered by any linter?
```suggestion
  const changedPackages = changedPackageLocation.map(packageDir => Object.keys(workspaceInfo).find(key => workspaceInfo[key].location === packageDir))
```

.circleci/scripts/find_changed_packages.js
```diff
@@ -0,0 +1,120 @@
...
+  return changedPackages ? getDependenciesFromChangedPackages(changedPackages, workspaceInfo) : allPackages
+
+}
+
+const main = () => {
+  const args = process.argv.slice(2)
+  console.log(`User provided arguments: ${args}`)
+
+  const userInputBaseCommit = args.length > 0 ? args[0] : null
+  console.log(`attempting to use ${userInputBaseCommit} as base commit`)
```
ori-moisis, line 107: this is a confusing log, when I looked at the run on circleci I thought there was a bug

.circleci/scripts/find_changed_packages.js
```diff
@@ -0,0 +1,120 @@
...
+}
+
+const findChangedPackages = (userInputBaseCommit, workspaceInfo) => {
+  const changedFiles = getChangedFiles(userInputBaseCommit)
+  console.log('Changed files:', changedFiles)
+  const changedPackageLocation = changedFiles.map(path => path.split('/').slice(0, 2).join('/')).filter((value, index, self) => self.indexOf(value) === index)
+
+  const changedPackages = changedPackageLocation.map(packageDir =>Object.keys(workspaceInfo).find(key => workspaceInfo[key].location === packageDir))
+  // if there are undefined values, it means that there are files that changed outside of the packages directory. Return null to indicate that we should test all packages
+  const hasChangedFilesOutsidePackage = changedPackages.length === 0 || changedPackages.some(pkg => !pkg)
```
ori-moisis, line 75: this condition doesn't really make sense to me
why is this the condition?

.circleci/scripts/find_changed_packages.js
benisraeladi: I'll change the function name. When it returns, it checks whether this is empty. If so, we run all tests because some file outside of `packages/` has changed

.circleci/scripts/find_changed_packages.js
benisraeladi: It returns only unique packages. It was either this or to `Array.from(Set(changedFiles...)`
I felt like the filter was cleaner

ori-moisis: I can honestly say it would have taken me minutes to figure out that this is what it does
so, not great on readability IMO

if you are really intent on using this trick, maybe put it in a clearly named function or something?

.circleci/config.yml
benisraeladi: In the continuation step, we don't want to checkout code again, because it will overwrite `continue_config.yml`

.circleci/scripts/find_changed_packages.js
benisraeladi: I found out that I had this due to an edge case (at first `pagerduty-adapter` was `pager-duty-adapter` and my remote branch still had that)
I changed it so that any packages that return `undefined` are removed, since this shouldn't happen if we're running on CircleCI

ori-moisis: that part I can understand (I read it as "any file that we don't know to explicitly link to a known package should trigger a full test")
but that doesn't explain the `changedPackages.length === 0` part, right?

benisraeladi: That's because the plan was for `getChangedPackages` to only return the packages, and if there were no packages changed the idea was to return null. I now changed it so that instead it returns both the changed packages and whether there were changes outside the packages, which makes `getPackagesToTest` clearer

.circleci/config_template.yml
```diff
@@ -0,0 +1,447 @@
...
+      - aws-cli/install:
+          version: 2.2.2
+      - aws-cli/setup:
+          aws-region: AWS_DEFAULT_REGION
+
+      - run:
+          name: Release version
+          command: ./.circleci/scripts/release_version.sh << pipeline.git.base_revision >>
+
+  e2e_tests:
```
ori-moisis, line 208: is it ok to have this defined even when we don't need to run any e2e tests?
wondering if circleci won't complain about having a job defined but not part of any workflow...

ori-moisis (commented):
mostly looks ok
how updated are the test cases?
maybe worth checking again with the latest version of the code

.circleci/config_template.yml
benisraeladi: It doesn't fail anything.
The linter does show that this functionality is unused, but if someone decides to remove it, the workflow won't be able to run (since we run all the tests if this file changes)

ori-moisis (approved):

PR #5933, opened 2024-05-22
SALTO-5906: Move board columns API calls to be async and called in th…

izibr:
…e start of the filters

Move request board columns API calls to be async and called at the start of the filters

---

---
_Release Notes_: 
_Jira Adapter_:
 - hopefully, make the fetch quicker  by updating the code so the API calls of the board columns will occur at the start of the filters

---
_User Notifications_: 

N/A


IdoZakk (commented):
Nice Job :-)
Reviewing the logic of this filter (it is calling the get only for kanban boards that have two backlogs columns) I do not think it is worth the code complication, it happens rarely and will save very little time.

PR #5935, opened 2024-05-22
SALTO-5699: Upgrade elemID config

shir-reifenberg:
Added utility to upgrade deprecated elemID config to the new infra type

---

The upgrade function - 
1. Convert any elemID related config from `apiDefinitions` to `elemID` config
2. remove any empty parts of `apiDefinitions`

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/adapter-components/test/definitions/user/user_config.test.ts
```diff
@@ -142,4 +142,181 @@ describe('config_shared', () => {
...
+              typeDefaults: {
+                transformation: {},
+              },
+              types: {
+                foo: {
+                  transformation: { idFields: ['name', 'status'] },
+                  request: {},
+                },
+                bar: {
+                  request: { recurseInto: [] } // not sure this case is legit
```
shir-reifenberg, line 240: In this case `recurseInto` config will be removed, but in some cases putting `recurseInto: []` is to override existing `recurseInto` config... as we don't handle it anyway and I think there isn't a real case for now, I guess it's ok
@netama WDYT?

packages/adapter-components/src/definitions/user/user_config.ts
```diff
@@ -124,3 +129,97 @@ export const adapterConfigFromConfig = <
...
+  if (_.isEmpty(types)) {
+    return {}
+  }
+
+  const convertedIds = _.mapValues(types, (defs, typeName) => {
+    const transformationConfig = defs.transformation
+    if (!transformationConfig) {
+      return undefined
+    }
+    const updatedElemIDConfig: ElemIDCustomization<TCustomNameMappingOptions> = {}
```
netama, line 152: if you prefer to update this this way and not functional-style then ok, but I _think_ the underlying issue here is the decisions of sometimes not deleting - I think we should either delete-and-log or throw, we shouldn't have leftovers from the config after the migration

packages/adapter-components/src/definitions/user/user_config.ts
```diff
@@ -124,3 +129,97 @@ export const adapterConfigFromConfig = <
...
+      )
+      updatedElemIDConfig.parts = updatedIdFields
+      delete transformationConfig.idFields
+    }
+    if (nameMapping !== undefined) {
+      if (updatedElemIDConfig.parts) {
+        updatedElemIDConfig.parts.forEach(part => _.set(part, 'mapping', nameMapping))
+        delete transformationConfig.nameMapping
+      } else {
+        // handaling the case of when mappings exists but idFields are missing is not trivial, because we don't know which parts should be updated
```
netama, line 172: typo:
```suggestion
        // handling the case of when mappings exists but idFields are missing is not trivial, because we don't know which parts should be updated
```

packages/adapter-components/src/definitions/user/user_config.ts
```diff
@@ -124,3 +129,97 @@ export const adapterConfigFromConfig = <
...
+    )
+    return updatedElemIDConfig
+  })
+
+  const typesWithIds = _.omitBy(convertedIds, _.isEmpty)
+
+  return _.isEmpty(typesWithIds) ? {} : { elemID: typesWithIds }
+}
+
+export const updateDeprecatedConfig = (
```
netama, line 191: nit: document? (or if you want to document only one of the functions, let's make it the exported one :) )

packages/adapter-components/src/definitions/user/user_config.ts
```diff
@@ -124,3 +129,97 @@ export const adapterConfigFromConfig = <
...
+  config: InstanceElement,
+): { config: InstanceElement; message: string } | undefined => {
+  const apiDefs = config.value?.fetch?.apiDefinitions
+  if (!apiDefs) {
+    return undefined
+  }
+
+  log.warn('adapter config contains deprecated api definitions: %s', safeJsonStringify(apiDefs))
+  const updatedConfig = config.clone()
+  const updatedElemIDs = updateElemIDDefinitions(updatedConfig.value.fetch.apiDefinitions)
```
netama, line 201: what's `value.fetch.apiDefinitions`? should it have been `value.apiDefinitions`?

packages/adapter-components/src/definitions/user/user_config.ts
```diff
@@ -124,3 +129,97 @@ export const adapterConfigFromConfig = <
...
+    return undefined
+  }
+  
+  const cleanedApiDefs = objects.cleanEmptyObjects(updatedConfig.value?.fetch?.apiDefinitions)
+  updatedConfig.value = {
+    ...updatedConfig.value,
+    fetch: {
+      ...updatedConfig.value.fetch,
+      ...updatedElemIDs,
+      apiDefinitions: cleanedApiDefs ? { ...cleanedApiDefs } : undefined,
```
netama, line 213: 1. continuing from above - isn't this _next_ to the `fetch` defs?
2. why do we need the spread? and not just `apiDefinitions: cleanedApiDefs`?

packages/adapter-components/src/definitions/user/user_config.ts
```diff
@@ -124,3 +129,97 @@ export const adapterConfigFromConfig = <
...
+    },
+  }
+
+  if (updatedConfig.value.fetch.apiDefinitions) {
+    log.error('some deprecated apiDefinitions remained in config: %s', safeJsonStringify(updatedConfig.value.fetch.apiDefinitions))
+  }
+  return {
+    config: updatedConfig,
+    message:
+      'The configuration options "apiDefinitions" is deprecated. The following changes will update the deprecated options to the "fetch" configuration option.',
```
netama, line 223: nit: maybe something like 
```suggestion
      'Elem ID customizations are now under `fetch.elemID`. The following changes will upgrade the deprecated definitions from `apiDefinitions` to the new location.',
```
(not sure if we can use the backticks here or not, if not we can quote)

packages/lowerdash/src/objects.ts
```diff
@@ -27,3 +27,21 @@ export const concatObjects = <T extends Record<string, ReadonlyArray<unknown> |
         .filter(isDefined)
         .flat(),
   ) as T
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const cleanEmptyObjects = (object: any): any | undefined => {
```
netama, line 32: can we adjust the types to make this actually run on and return objects / records?

packages/lowerdash/src/objects.ts
```diff
@@ -27,3 +27,21 @@ export const concatObjects = <T extends Record<string, ReadonlyArray<unknown> |
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const cleanObject = (obj: any): any => {
+    if (Array.isArray(obj)) {
```
netama, line 35: empty arrays should not be treated as meaningless, we should keep them (so should probably explain in the function documentation as well)

netama (commented):
thanks! added some comments, lmk if you want to discuss or if I misread something

packages/adapter-components/test/definitions/user/user_config.test.ts
netama: sorry somehow missed this, commented on the util function that I think we shouldn't remove empty arrays so this should be kept as well, right?

packages/adapter-components/src/definitions/user/user_config.ts
```diff
@@ -124,3 +129,97 @@ export const adapterConfigFromConfig = <
...
+      updatedElemIDConfig.parts = updatedIdFields
+      delete transformationConfig.idFields
+    }
+    if (nameMapping !== undefined) {
+      if (updatedElemIDConfig.parts) {
+        updatedElemIDConfig.parts.forEach(part => _.set(part, 'mapping', nameMapping))
+        delete transformationConfig.nameMapping
+      } else {
+        // handaling the case of when mappings exists but idFields are missing is not trivial, because we don't know which parts should be updated
+        log.warn('found nameMapping without idFields for type %s', typeName)
```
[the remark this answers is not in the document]
shir-reifenberg: you're right

packages/adapter-components/src/definitions/user/user_config.ts
shir-reifenberg: oops :( your'e right, I tested it in the wrong hierarchy 😅

packages/adapter-components/src/definitions/user/user_config.ts
```diff
@@ -124,3 +129,97 @@ export const adapterConfigFromConfig = <
...
+/**
+ * Convert elemID definitions from deprecated apiDefinitions to the new format
+ * Note: 
+ * 1. config is changed in place
+ * 2. only elemID related definitions are converted
+ */
+const updateElemIDDefinitions = <TCustomNameMappingOptions extends string>(
+  apiDefinitions: AdapterApiConfig,
+): Record<string, ElemIDCustomization<TCustomNameMappingOptions>> => {
+  const { types } = apiDefinitions
```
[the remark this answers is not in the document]
shir-reifenberg: discussed IRL, any non empty `typeDefaults` would be logged in the later log

packages/adapter-components/src/definitions/user/config_upgrade_utils.ts
```diff
@@ -0,0 +1,126 @@
...
+    log.debug('found no elemID definitions to update in config')
+    return undefined
+  }
+
+  const cleanedApiDefs = objects.cleanEmptyObjects(updatedConfig.value?.apiDefinitions)
+  updatedConfig.value = {
+    ...updatedConfig.value,
+    fetch: {
+      ...updatedConfig.value.fetch,
+      ...updatedElemIDs,
```
netama, line 110: can we add a check in case there already are `elemIDs` in the old config?

packages/lowerdash/src/objects.ts
```diff
@@ -27,3 +27,24 @@ export const concatObjects = <T extends Record<string, ReadonlyArray<unknown> |
         .filter(isDefined)
         .flat(),
   ) as T
+
+/*
+ * Cleans empty objects recursively
+ * Notice: arrays are ignored and treated like primitives
+ */
+export const cleanEmptyObjects = (object: Record<string, unknown>): Record<string, unknown> | undefined => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
netama, line 36: 🙈

packages/lowerdash/src/objects.ts
```diff
@@ -27,3 +27,24 @@ export const concatObjects = <T extends Record<string, ReadonlyArray<unknown> |
...
+  const cleanObject = (obj: any): any => {
+    if (_.isPlainObject(obj)) {
+      const mapped = _.mapValues(obj, val => {
+        const cleanedVal = cleanObject(val)
+        return _.isPlainObject(val) && _.isEmpty(cleanedVal) ? undefined : cleanedVal
+      })
+      return _.omitBy(mapped, _.isUndefined)
```
netama, line 43: nit: I think there's a preference to check explicitly `val => val !== undefined` (or you can `pickBy` `isDefined`)

netama (approved):
LGTM 🎈 added one comment and a nit

PR #5936, opened 2024-05-22
SALTO-5741 support users and groups in confluence adapter

omrifarkash:
Support displayName on group and user references

---

---
_Release Notes_: 
_Confluence Adapter_:
- Support displayName on group and user references

---
_User Notifications_: 
_Confluence Adapter_:
- Support displayName on group and user references


packages/confluence-adapter/src/get_users_and_groups.ts
```diff
@@ -0,0 +1,140 @@
...
+const USERS_FETCH_DEF: definitionsUtils.fetch.InstanceFetchApiDefinitions<Options> = {
+  requests: [
+    {
+      endpoint: {
+        method: 'get',
+        path: '/rest/api/3/users/search',
+        queryArgs: {
+          maxResults: '10000',
+        },
+      },
```
omrifarkash, line 81: For this API I need to use different pagination, not sure how to do it yet.

packages/confluence-adapter/src/get_users_and_groups.ts
```diff
@@ -0,0 +1,140 @@
...
+    userResults = await requester.requestAllForResource({
+      callerIdentifier: { typeName: USER_TYPE_NAME },
+      contextPossibleArgs: {},
+    })
+    usersIndex = _.keyBy(userResults.map(item => item.value).filter(isUser), user => user.accountId)
+  } catch (e) {
+    log.error('Failed to fetch users with error %o', e)
+    usersIndex = {}
+  }
+  return { groupsIndex, usersIndex }
```
shoshana-houri, line 139: Nit (possible way to share code here):
```suggestion
const callSingleResource = async <T>(
    typeName: string,
    filterFn: (item: any) => item is T,
    idKey: string
  ): Promise<Record<string, T>> => {
    try {
      const result = await requester.requestAllForResource({
        callerIdentifier: { typeName },
        contextPossibleArgs: {},
      })
      return _.keyBy(result.map(item => item.value).filter(filterFn), idKey)
    } catch (e) {
      log.error(`Failed to fetch ${typeName.toLowerCase()}s with error %o`, e)
      return {}
    }
  }

  const [groupsIndex, usersIndex] = await Promise.all([
    callSingleResource<Group>(GROUP_TYPE_NAME, isGroup, 'id'),
    callSingleResource<User>(USER_TYPE_NAME, isUser, 'accountId'),
  ])

  return { groupsIndex, usersIndex }
```

packages/confluence-adapter/src/get_users_and_groups.ts
```diff
@@ -0,0 +1,140 @@
...
+        {
+          ...definitionsWithFetch.fetch.instances,
+          customizations: createCustomizations(),
+        },
+        'requests',
+      ),
+    ),
+  })
+  let groupResult: definitionsUtils.GeneratedItem<Record<string, unknown>, Values>[] = []
+  let userResults: definitionsUtils.GeneratedItem<Record<string, unknown>, Values>[] = []
```
shoshana-houri, line 116: Nit: In any case I would remove these, they are used in the same scope so IIUC they can be constants

packages/confluence-adapter/test/definitions/utils/page.test.ts
```diff
@@ -121,6 +126,32 @@ describe('page definitions utils', () => {
...
+              changes: [],
+              groupID: 'group-id',
+            },
+            sharedContext: {},
+            change: pageChange,
+          },
+          value: {
+            authorId: { accountId: 'authorId', displayName: 'authorId' },
+            ownerId: { accountId: 'ownerId', displayName: 'ownerId' },
+            notUser: 'not',
```
shoshana-houri, line 145: 😄

packages/confluence-adapter/test/get_users_and_groups.test.ts
```diff
@@ -0,0 +1,103 @@
...
+  it('should return empty users index and correct groups index when failed to fetch users', async () => {
+    mockRequestAllResources
+      .mockResolvedValueOnce(mockGroupsResponse)
+      .mockRejectedValueOnce(new Error('Failed to fetch users'))
+    expect(await getUsersAndGroups(mockDefinitions)).toEqual({
+      groupsIndex: expectedGroupsIndex,
+      usersIndex: {},
+    })
+  })
+  it('should return user and group indices when succeed to fetch groups and users', async () => {
```
shoshana-houri, line 96: I think it's worth adding another one that checks the object validation :)

shoshana-houri (commented):
Looks great!! 🚀 🚀 
Added nits and questions.

As a more general question - I see that their API supports fetching [specific users](https://developer.atlassian.com/cloud/confluence/rest/v1/api-group-users/#api-wiki-rest-api-user-bulk-get) and [specific groups](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-groups/#api-rest-api-3-group-bulk-get). Perhaps it's worth first retrieving the list of accountIds and groupIds from the pages and permissions result, and only then fetching the relevant users/groups?

packages/confluence-adapter/src/get_users_and_groups.ts
```diff
@@ -0,0 +1,140 @@
...
+}
+
+type Group = {
+  name: string
+  id: string
+}
+
+const USER_SCHEMA = Joi.object({
+  accountId: Joi.string().required(),
+  emailAddress: Joi.string(),
```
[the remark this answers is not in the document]
omrifarkash: Currently we are not using it, I can remove it but we might want to use it in the future (can be added then)

packages/confluence-adapter/src/get_users_and_groups.ts
```diff
@@ -0,0 +1,140 @@
...
+}
+
+export const getUsersAndGroups = async (
+  definitions: definitionsUtils.ApiDefinitions<Options>,
+): Promise<{ usersIndex: Record<string, User>; groupsIndex: Record<string, Group> }> => {
+  const { fetch } = definitions
+  if (fetch === undefined) {
+    throw new Error('could not find fetch definitions')
+  }
+  const createCustomizations = (): Record<string, definitionsUtils.fetch.InstanceFetchApiDefinitions<Options>> => ({
```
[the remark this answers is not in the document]
omrifarkash: Just to be consistent with how we create customization on real fetch definitions https://github.com/salto-io/salto/blob/main/packages/confluence-adapter/src/definitions/fetch/fetch.ts#L443

I don't mind changing it to const

packages/confluence-adapter/src/get_users_and_groups.ts
omrifarkash: Nice

packages/confluence-adapter/test/get_users_and_groups.test.ts
omrifarkash: I do check it as the mock response contains one user and one group in invalid format, that expected to be filtered out

shoshana-houri: Great! Sorry I missed it

omrifarkash:
@shoshana-houri I didn't add comment for adding pagination as I am planning to do it before I merge this PR

packages/confluence-adapter/src/get_users_and_groups.ts
```diff
@@ -0,0 +1,134 @@
...
+    },
+  ],
+}
+
+const USERS_FETCH_DEF: definitionsUtils.fetch.InstanceFetchApiDefinitions<Options> = {
+  requests: [
+    {
+      endpoint: {
+        method: 'get',
+        path: '/wiki/rest/api/user/bulk?accountId={idsToQuery}',
```
omrifarkash, line 74: @netama 
@shoshana-houri and I have found out we can query specific users by ids in a single call.
But, I am not sure if this can work for a big mass of users as I'm adding each id to the path. 

WDYT about it? we can talk IRL if you prefer :)

shoshana-houri (approved):
Looks great to me 🚀
Approving on my side. Re the user's scaling concern, it's probably best to get @netama's opinion.

PR #5937, opened 2024-05-22
SALTO-5969: Clean RoleAssignment Support

shir-reifenberg:
Clean all RoleAssigment related code

---

The type is already excluded for a while and not being used (or asked for)
gonna remove this for now cause it's not gonna be part of the upgrade to the new infra

no changes are expected as this is not in use

---
_Release Notes_: 
None

---
_User Notifications_: 
None


omrilit (approved):
Nice move :)

PR #5938, opened 2024-05-22
SALTO-5970: Fix ApplicationGroupAssignment paths

shir-reifenberg:
There was a bug in `ApplicationGroupAssignment` path

---

This is gonna change in the upgrade PR anyway, extracting this to a separate PR incase we'll need to revert the upgrade PR to minimize user visible changes

---
_Release Notes_: 

_Okta_adapter_:
- Fix a bug causing ApplicationGroupAssignment instances not to be nested under the parent application in the files tree.

---
_User Notifications_:

_Okta_adapter_:
- ApplicationGroupAssignment element files will be nested under the parent application


omrilit (approved):
To reduce noise to existing WSs, maybe it would be better to merge this one along with the new infra, just in a different commit?

Nurdok (approved):

PR #5939, opened 2024-05-22
SALTO-5867 Google ws add e2e logs

Uri-Bar:
SALTO-5867 Google ws add e2e logs
---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/google-workspace-adapter/e2e_test/adapter.test.ts
+++ b/packages/google-workspace-adapter/e2e_test/adapter.test.ts
@@ -179,18 +179,13 @@ describe('Google Workspace adapter E2E', () => {
 
     const deployAndFetch = async (changes: Change[]): Promise<void> => {
       deployResults = await deployChanges(adapterAttr, changes)
+      // Google workspace API takes time to reflect the changes from the deploy.
+      // So, we need to wait for some time before fetching in order to get the new elements.
       await sleep(5000)
       const fetchResult = await adapterAttr.adapter.fetch({
         progressReporter: { reportProgress: () => null },
       })
       elements = fetchResult.elements
-      log.info(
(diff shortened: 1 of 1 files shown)


packages/google-workspace-adapter/e2e_test/adapter.test.ts
```diff
@@ -177,10 +179,18 @@ describe('Google Workspace adapter E2E', () => {
 
     const deployAndFetch = async (changes: Change[]): Promise<void> => {
       deployResults = await deployChanges(adapterAttr, changes)
+      await sleep(5000)
```
shir-reifenberg, line 182: nit: `jest.setTimeout(5000)`
also please add a comment to explain this ?

packages/google-workspace-adapter/e2e_test/adapter.test.ts
```diff
@@ -177,10 +179,18 @@ describe('Google Workspace adapter E2E', () => {
...
         progressReporter: { reportProgress: () => null },
       })
       elements = fetchResult.elements
+      log.info(
+        'Fetched these Test elements: %o',
+        elements
+          .filter(isInstanceElement)
+          .filter(checkNameField)
+          .map(e => e.elemID.getFullName()),
+      )
```
shir-reifenberg, line 193: I think we should only log on failure scenario
IIUC the e2e fails on `expect(instance).toBeDefined()` 
so please add an if to check and only log if instance is undefined

packages/google-workspace-adapter/e2e_test/adapter.test.ts
Uri-Bar: set timeout is something else.
I will add a comment

PR #5940, opened 2024-05-22
SALTO-5928: Using strong references for managed elements

yelly:
They should never have been weak.

---

---
_Release Notes_: 
_Salesforce_:
* Using strong custom references for managed elements.

---
_User Notifications_: 
None.


tamtamirr (changes requested):
Please bump the reference index 🙏

tamtamirr (approved):
LGTM.

PR #5943, opened 2024-05-22
SALTO-4858 Add missing extensions transition rules change validator

IlayDavid95:
Jira:
- Added change validator for workflows that will prevent deployment of transition rules that depend on Jira apps not installed in the target Jira Cloud env. This will only work for the latest iteration of Workflows by Jira (AKA WorkflowV2).

---

_Additional context for reviewer_
This PR is a part of a series of changes, followed by a FixElement and a Filter, for allowing deployment without the faulty transition rules and fixing special Forge apps transition rules, respectively.

---
_Release Notes_: 
Jira:
- Prevent deployment of workflows with transition rules that are dependent on missing Jira apps. This is only for the new version of Workflows by Jira (enabled by the enableNewWorkflowAPI flag)

PR #5944, opened 2024-05-22
SALTO-5975 refactor out old style jira get cloud id and use the one in the jira client

IlayDavid95:
Refactor to Jira adapter, specifically using the getCloudId in JiraClient which utilizes the Jira API according to their [documentation](https://confluence.atlassian.com/jirakb/how-to-find-cloud-site-id-1272283178.html).

Mainly test fixes.
---

This PR is dependent on the changes in branch IlayDavid95:SALTO-4858-prevent-deployment-of-unknown-types-of-validators-and-post-functions and should be reviewed only after PR #5943 is reviewed and merged.

---
_Release Notes_: 
No effect on users.
---
_User Notifications_: 
No effect on users.


packages/jira-adapter/src/client/client.ts
```diff
@@ -247,4 +287,60 @@ export default class JiraClient extends clientUtils.AdapterHTTPClient<Credential
...
+      const response = await this.get({
+        url: GET_CLOUD_ID_URL,
+      })
+      if (!isCloudIdResponse(response.data)) {
+        throw new Error(`'Failed to get cloud id, received invalid response' ${response.data}`)
+      }
+      this.cloudId = response.data.cloudId
+    }
+
+    return this.cloudId
```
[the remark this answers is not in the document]
IlayDavid95: I'm not sure I understand. If the request fails then this.cloudId will remain undefined and so will not be cached.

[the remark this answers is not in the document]
IdoZakk: Sorry, you are correct

packages/jira-adapter/test/filters/automation/automation_fetch.test.ts
```diff
@@ -325,23 +314,36 @@ describe('automationFetchFilter', () => {
...
     })
 
     it('should throw if cloud resource is not an object', async () => {
-      connection.post.mockImplementation(async url => {
-        if (url === '/rest/webResources/1.0/resources') {
+      const { client: cli, paginator, connection: conn } = mockClient()
+      client = cli
+      connection = conn
+      connection.get.mockImplementation((url: string): Value => {
+        if (url === GET_CLOUD_ID_URL) {
```
IdoZakk, line 336: You do this init several times, you should unify the tests in a describe. But I think it won't work - once the call fails we will not call it again. If I am correct let's add a way for the client to recover

packages/jira-adapter/test/filters/automation/automation_label/label_fetch.test.ts
```diff
@@ -207,63 +179,6 @@ describe('automationLabelFetchFilter', () => {
...
-          return {
-            status: 200,
-            data: {
-              unparsedData: {},
-            },
-          }
-        }
-
-        throw new Error(`Unexpected url ${url}`)
-      })
```
IdoZakk, line 222: Keep one test to make sure the failure is not swallowed for labels

packages/jira-adapter/test/filters/automation/automation_label/label_fetch.test.ts
```diff
@@ -293,7 +208,7 @@ describe('automationLabelFetchFilter', () => {
     })
   })
   it('should warn if response is 403', async () => {
-    const { client: cli, connection: conn } = mockClient(true)
+    const { client: cli, connection: conn } = mockClient(true, DEFAULT_CLOUD_ID)
```
IdoZakk, line 211: If mockClient is called with true it is Jira DC, no need for cloud ID

packages/jira-adapter/test/filters/automation/automation_fetch.test.ts
```diff
@@ -445,7 +458,7 @@ describe('automationFetchFilter', () => {
     })
   })
   it('should warn if response is 403', async () => {
-    const { client: cli, connection: conn } = mockClient(true)
+    const { client: cli, connection: conn } = mockClient(true, DEFAULT_CLOUD_ID)
```
IdoZakk, line 461: I did not review the new mock client version last time. If most of the time you call it with `DEFAULT_CLOUD_ID` make it the default behavior

IdoZakk (commented):
Very nice! I was not aware we are switching cloudId API, good catch
I gave some tests comments

packages/jira-adapter/test/filters/automation/automation_fetch.test.ts
IlayDavid95: Giving a CloudId as an argument means it is not undefined and therefore the getCloudId call is mocked. I'd rather someone explicitly mock something than explicitly unmock it. What do you think?

packages/jira-adapter/test/filters/automation/automation_label/label_fetch.test.ts
IlayDavid95: Done

packages/jira-adapter/test/filters/automation/automation_fetch.test.ts
IlayDavid95: These inits are actually overriding the most common behaviour initialized in the beforeEach.
Added cache invalidating in the client when getCloudId fails.

packages/jira-adapter/test/filters/automation/automation_label/label_fetch.test.ts
IlayDavid95: There is already a test for client that checks whether an invalid response to getCloudId throws.

packages/jira-adapter/test/filters/automation/automation_fetch.test.ts
IdoZakk: I meant the part about creating a new client and connection every time. Why not create them once in the before each?

packages/jira-adapter/test/filters/automation/automation_label/label_fetch.test.ts
IdoZakk: But what if label_fetch ignores it or does not handle it properly?

packages/jira-adapter/test/filters/automation/automation_fetch.test.ts
IdoZakk: The mock client helps programmers write tests with clients without the need to understand the logic of the client. They want to mock the client, usually the are indifferent to the cloud id. In most cases they do not care or know what and how the cloud id is brought, especially now that it is a feature of the client. Keeping it will make test writing more difficult and less clean

packages/jira-adapter/test/filters/forms/forms.test.ts
```diff
@@ -591,7 +570,7 @@ describe('forms filter', () => {
     beforeEach(async () => {
       const config = _.cloneDeep(getDefaultConfig({ isDataCenter: false }))
       config.fetch.enableJSM = true
-      const { client: cli, connection: conn } = mockClient(false)
+      const { client: cli, connection: conn } = mockClient(false, DEFAULT_CLOUD_ID)
```
IdoZakk, line 573: Here you left DEFAULT_CLOUD_ID
I think I am missing something- you removed the default value but did not add it as default to the mockClient code

packages/jira-adapter/test/adapter.test.ts
```diff
@@ -512,12 +511,8 @@ describe('adapter', () => {
...
           })
         // mock as we call getCloudId in the forms filter.
-        mockAxiosAdapter.onPost().reply(200, {
-          unparsedData: {
-            [CLOUD_RESOURCE_FIELD]: safeJsonStringify({
-              tenantId: 'cloudId',
-            }),
-          },
+        mockAxiosAdapter.onGet(GET_CLOUD_ID_URL).replyOnce(200, {
+          cloudId: DEFAULT_CLOUD_ID,
```
IdoZakk, line 515: We should remove those calls as well

packages/jira-adapter/test/filters/automation/automation_label/label_fetch.test.ts
IdoZakk: I think you missed this comment

packages/jira-adapter/test/filters/automation/automation_fetch.test.ts
```diff
@@ -69,33 +69,31 @@ describe('automationFetchFilter', () => {
...
-
   const mockPostResponse = (url: string): Value => {
-    if (url === '/rest/webResources/1.0/resources') {
+    if (url === DEFAULT_URL) {
+      return automationResponse
+    }
+
+    throw new Error(`Unexpected url ${url}`)
+  }
+
```
IdoZakk, line 79: Same here, as we moved it to the client no test need to do it here

IdoZakk (commented):
I added some comments. Also please review the previous comments, you did not fix/comment on at least two of them

packages/jira-adapter/test/adapter.test.ts
IlayDavid95: We need them because the tests don't use the mockClient.

packages/jira-adapter/test/filters/automation/automation_fetch.test.ts
IlayDavid95: Done

packages/jira-adapter/test/filters/automation/automation_fetch.test.ts
IlayDavid95: They are not consistent between each test case. Some initialize as DC, some initialize without default cloud id, some initialize with default cloud id.

packages/jira-adapter/test/filters/automation/automation_label/label_fetch.test.ts
IlayDavid95: Re-added one test.

packages/jira-adapter/test/filters/forms/forms.test.ts
IlayDavid95: I don't think I understand what you mean 😅

IdoZakk (commented):
Nice :-)
There was a comment from the first review you did not answer/resolve. I really think you should add the default cloud ID as the default of the mock client and not make the implementor add it himself every time

packages/jira-adapter/test/filters/automation/automation_fetch.test.ts
IlayDavid95: Extracted the mocked response to a constant in utils

packages/jira-adapter/test/filters/automation/automation_fetch.test.ts
IlayDavid95: Renamed the parameter name to something more indicative.

IdoZakk (commented):
Discussed IRL

packages/jira-adapter/test/utils.ts
```diff
@@ -58,7 +58,11 @@ type ClientWithMockConnection = {
   getUserMapFunc: GetUserMapFunc
   scriptRunnerClient: ScriptRunnerClient
 }
-export const mockClient = (isDataCenter = false, mockedCloudID?: string): ClientWithMockConnection => {
+export const mockClient = (
+  isDataCenter = false,
+  mockGetCloudId: boolean = true,
```
IdoZakk, line 63: You can remove the boolean and keep the old condition.
There will be a default value of DEFAULT_CLOUD_ID for mockGetCloudId, and if someone wants to set it as undefined they can

IdoZakk (approved):
💯 
I am approving, but please see my comment.
Also, if the linter does not run automatically for you on save you should fix the vscode settings, it is very annoying otherwise

packages/jira-adapter/test/utils.ts
IlayDavid95: Done. Changed the type to string | null so that you can pass a null if you don't want to mock.

PR #5946, opened 2024-05-22
SALTO-5847 resolve ASV references of the taxSchedule field

dantal4:
Recognize that the ASV references of the `taxSchedule` field belongs to the taxSchedule SuiteQL table, that we're already fetching (when `resolveAccountSpecificValues=true`).

---

---
_Release Notes_: 
Netsuite Adapter:
- Resolve ASV references of the `taxSchedule` field

---
_User Notifications_: 
Netsuite Adapter:
- ASV references of the `taxSchedule` field will change the type in them: from `object` to `taxSchedule`:
```
taxSchedule = {
  id = "[ACCOUNT_SPECIFIC_VALUE] (taxSchedule) (Tax Schedule 123)"
}
```

### DIFF
--- a/packages/netsuite-adapter/src/filters/data_account_specific_values.ts
+++ b/packages/netsuite-adapter/src/filters/data_account_specific_values.ts
@@ -42,6 +42,7 @@ import {
   NAME_FIELD,
   NETSUITE,
   SUBTYPES_PATH,
+  TAX_SCHEDULE,
   TYPES_PATH,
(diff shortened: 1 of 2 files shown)


omrilit (commented):
Great work!
Since this change would introduce noise to existing WSs (that have the ASV flag enabled), we should communicate it properly to those customers before merging.

omrilit (approved):

PR #5947, opened 2024-05-23
SALTO-5866 confluence e2e

omrifarkash:
---

- add e2e_utils to adapter-components

---
_Release Notes_: 
Confluence Adapter_:
- e2e test

---
_User Notifications_: 
None


packages/confluence-adapter/e2e_test/mock_elements.ts
```diff
@@ -0,0 +1,54 @@
...
+      labels: {
+        start: 0,
+        limit: 200,
+        size: 0,
+      },
+    },
+    type: 'global',
+    status: 'current',
+  },
+  [PAGE_TYPE_NAME]: {
```
shir-reifenberg, line 41: is this type usually that simple ?

packages/confluence-adapter/e2e_test/mock_elements.ts
```diff
@@ -0,0 +1,54 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { Values } from '@salto-io/adapter-api'
+import { SPACE_TYPE_NAME, PAGE_TYPE_NAME, TEMPLATE_TYPE_NAME } from '../src/constants'
+
+export const mockDefaultValues: Record<string, Values> = {
```
shir-reifenberg, line 19: nit: I think it's worth adding the instances we test something with reference / parent annotation, so that the E2E would cover the reference resolving as well.

packages/confluence-adapter/e2e_test/adapter.test.ts
```diff
@@ -0,0 +1,284 @@
...
+        .map(res => res.appliedChanges)
+        .flat()
+        .map(change => getChangeData(change)) as InstanceElement[]
+
+      deployInstances.forEach(deployedInstance => {
+        const { typeName } = deployedInstance.elemID
+        const instance = elements
+          .filter(isInstanceElement)
+          .filter(inst => inst.elemID.typeName === typeName)
+          .find(inst => inst.value[nameFieldPerType[typeName]] === deployedInstance.value[nameFieldPerType[typeName]])
```
shir-reifenberg, line 267: why are you comparing `nameFieldPerType` and not elemID?

Also, probably worth creating `_.keyBy(elements, e => e.elemID.getFullName())` once instead of filtering the list for each `deployedInstance`

packages/confluence-adapter/e2e_test/adapter.test.ts
```diff
@@ -0,0 +1,284 @@
...
+
+const TEST_PREFIX = 'Test'
+
+const nameFieldPerType: Record<string, string> = {
+  [SPACE_TYPE_NAME]: 'key',
+  [PAGE_TYPE_NAME]: 'title',
+  [TEMPLATE_TYPE_NAME]: 'name',
+}
+
+const fieldsToComparePerType: Record<string, string[]> = {
```
shir-reifenberg, line 75: Why are we limiting the compare to specific fields ?

packages/confluence-adapter/e2e_test/adapter.test.ts
```diff
@@ -0,0 +1,284 @@
...
+          if (planElement !== undefined) {
+            applyDetailedChanges(planElement, detailedCompare(planElement, updatedElement))
+          }
+        })
+      return deployResult
+    })
+    .toArray()
+}
+
+const checkNameField = (instance: InstanceElement): boolean => {
```
shir-reifenberg, line 166: why is this needed ?
Also I think we shouldn't have `nameFieldPerType` cause it's a duplication for `elemID` definitions, and changing the fields used for elemID would break this test

packages/confluence-adapter/e2e_test/adapter.test.ts
```diff
@@ -0,0 +1,284 @@
...
+
+  return [toChange({ after: spaceInstance }), toChange({ after: pageInstance }), toChange({ after: templateInstance })]
+}
+
+const nullProgressReporter: ProgressReporter = {
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  reportProgress: () => {},
+}
+
+const deployChanges = async (adapterAttr: Reals, changes: Change[]): Promise<DeployResult[]> => {
```
shir-reifenberg, line 142: is this just a copy paste from google-workspace `deployChanges`? if so lets create e2e utils and make both e2e's use this

packages/confluence-adapter/e2e_test/adapter.test.ts
```diff
@@ -0,0 +1,284 @@
...
+const checkNameField = (instance: InstanceElement): boolean => {
+  const { typeName } = instance.elemID
+  const nameField = instance.value[nameFieldPerType[typeName]]
+  return typeof nameField === 'string' && nameField.startsWith(TEST_PREFIX)
+}
+
+const getChangesForInitialCleanup = (elements: InstanceElement[]): Change<InstanceElement>[] =>
+  elements.filter(checkNameField).map(instance => toChange({ before: instance }))
+
+const deployCleanup = async (adapterAttr: Reals, elements: InstanceElement[]): Promise<void> => {
```
shir-reifenberg, line 175: relevant for here too, let's not duplicate?

packages/confluence-adapter/e2e_test/adapter.test.ts
```diff
@@ -0,0 +1,284 @@
...
+    types,
+    uniqueValue: createUniqueValue(PAGE_TYPE_NAME),
+  })
+
+  const templateInstance = createInstance({
+    typeName: TEMPLATE_TYPE_NAME,
+    types,
+    uniqueValue: createUniqueValue(TEMPLATE_TYPE_NAME),
+  })
+  spaceInstance.value.name = createUniqueValue(SPACE_TYPE_NAME)
```
shir-reifenberg, line 129: nit: you can add `createInstance` with `valuesOverride` param and then override default values provided by `mockDefaultValues`

packages/confluence-adapter/e2e_test/adapter.test.ts
```diff
@@ -0,0 +1,284 @@
...
+  const spaceRef = new ReferenceExpression(spaceInstance.elemID, spaceInstance)
+  pageInstance.value.spaceId = spaceRef
+  templateInstance.annotate({ [CORE_ANNOTATIONS.PARENT]: [spaceRef] })
```
shir-reifenberg, line 132: I now see you did included a reference :) please generalize `createInstance` to have a parent param for easier extension of this test

shir-reifenberg (commented):
Looks good 😎  Added some questions and comments
As is seems like there's a pretty similar structure to this test and google workspace test, I think we should consider consolidate what ever is possible for faster creation of next tests

packages/confluence-adapter/e2e_test/mock_elements.ts
omrifarkash: Yes

packages/confluence-adapter/e2e_test/mock_elements.ts
omrifarkash: I am doing it on the test

packages/confluence-adapter/e2e_test/adapter.test.ts
omrifarkash: - Because I created the instances without a real computeElemId func so the elemIds won't be the same.
- 👍

packages/confluence-adapter/e2e_test/adapter.test.ts
omrifarkash: Because there are fields we expect to be different. I can work with black list if you prefer

packages/confluence-adapter/e2e_test/adapter.test.ts
omrifarkash: These are not necessarily equal to elemId parts. but fields that must be unique

For example, key on space instance

packages/confluence-adapter/e2e_test/adapter.test.ts
omrifarkash: It is as most of the code here. I think Neta wants to make something generic after we will see more adapters. There is a different task for it

packages/confluence-adapter/e2e_test/adapter.test.ts
shir-reifenberg: I think that if something changes in the APIs it would be difficult to detect it this way.
in what cases we expect fields to be different?

packages/confluence-adapter/e2e_test/adapter.test.ts
shir-reifenberg: not sure I understand, let's discuss IRL ?

packages/confluence-adapter/e2e_test/adapter.test.ts
shir-reifenberg: whatever you and @netama decide, but IMO as there are at least 3 more adapters in progress now, I think consolidating the code some of this code is not a lot of effort and will save us times in the next E2Es

packages/adapter-components/src/e2e_utils.ts
shir-reifenberg, line 1: maybe just for consistency with other packages put this in `packages/adapter-components/e2e_test/utils.ts` ?

shir-reifenberg (approved):
Looks great! 👑 
added mainly nits but nothing major

PR #5948, opened 2024-05-23
SALTO-5978: Adding separate Organization API Version settings in SFDC

yelly:
We will have the duplication for a bit until we enbale the optional feature.

---

---
_Release Notes_: 
_Salesforce_:
* Adding a hidden Organization API Version settings type and instance which stores the latest supported API version. This is only temporary and will be removed soon.

---
_User Notifications_: 
None.

### DIFF
--- a/packages/salesforce-adapter/src/filters/organization_settings.ts
+++ b/packages/salesforce-adapter/src/filters/organization_settings.ts
@@ -32,6 +32,7 @@ import {
 } from '../transformers/transformer'
 import {
   API_NAME,
+  ORGANIZATION_API_VERSION,
   ORGANIZATION_SETTINGS,
   RECORDS_PATH,
   SALESFORCE,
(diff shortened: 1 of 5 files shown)


packages/salesforce-adapter/src/filters/organization_settings.ts
```diff
@@ -165,9 +166,36 @@ const createOrganizationInstance = (
...
+
+  const instance = new InstanceElement(ElemID.CONFIG_NAME, objectType)
+
+  return [objectType, instance]
+}
+
 const addLatestSupportedAPIVersion = async (
   client: SalesforceClient,
-  instance: InstanceElement,
+  instances: InstanceElement[],
```
tamtamirr, line 198: This method is pretty odd and confusing. It requires following the code to understand the input here.
To improve it, I suggest the following method:
```
type AddLatestSupportedAPIVersionParams = {
  client: SalesforceClient
  apiVersionInstance: InstanceElement
  organizationInstance?: InstanceElement
}
const addLatestSupportedAPIVersion = async ({
  client,
  apiVersionInstance,
  organizationInstance,
}: AddLatestSupportedAPIVersionParams => ...
```

tamtamirr (commented):
LGTM! Added one small comment.
Before merging, can we run a sanity fetch that make sure this implementation works as expected in `fetch with changes detection mode`? (use the `-cd` flag)

packages/salesforce-adapter/src/filters/organization_settings.ts
yelly: Done.

yelly:
Ran with changes detection, everything worked fine.

tamtamirr (approved):
LGTM!

PR #5949, opened 2024-05-23
SALTO-5668: Google Workspace OAuth - enforce consent + login screen

shoshana-houri:
We want to be able to choose the desired google workspace account even if we're already logged in to this account/another one.

---
_Release Notes_: 
None

---
_User Notifications_: 
None


Uri-Bar (approved):
שאי ברכה

PR #5950, opened 2024-05-23
SALTO-5979: FixErrorDeployingDefaultAccessPolicyRule

idoamit8:
This PR fixes a bug where the deployment of a default access policy rule would fail.

---

When deploying a system rule of type AccessPolicyRule, you need to set its priority to 99, similar to how we handle PROFILE_ENROLLMENT_RULE_TYPE_NAME.

---
_Release Notes_: 
_Okta Adapter:_
* Fixed the deployment issue for default access policy rules by ensuring the priority is set correctly to 99.

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/filters/default_rule_deployment.ts
+++ b/packages/okta-adapter/src/filters/default_rule_deployment.ts
@@ -39,6 +39,8 @@ type PolicyRule = {
   system: boolean
 }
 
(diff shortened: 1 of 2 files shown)


shir-reifenberg (approved):
Thanks for the quick fix 💪 👑 
please add a unit test 🙏

PR #5951, opened 2024-05-23
SALTO-4584 keep elemID of data elements when identifier is renamed

dantal4:
Set `internalId` field as `serviceid` in data elements, and use its value to get existing elements' elemIDs, instead of using the instance's identifier (name/scriptid etc) that can be renamed. This prevents the removal & addition of an element in case that it was renamed.

A side effect of this change will be that the `immutable_changes` CV won't return error when a data element is deployed with a modified field that was considered a serviceId until now (e.g `name` in `netsuite.location`). I think that it's OK since NS allows modifying those fields.

---

---
_Release Notes_: 
Netsuite Adapter:
- keep elemID of data elements when identifier is renamed

---
_User Notifications_: 
Netsuite Adapter:
- In data types, the `internalId` field will be the `serviceid` field, instead of another field that was the `serviceid`


dantal4:
> 💯 Just making sure that the only diff existing WSs will have is around the Types folder due to field types changes, right?

@omrilit also in custom record types- we set the internalId fieldType there too. I'll wait with merging this PR to be with the other 2 PRs that need customer communication.

PR #5952, opened 2024-05-23
SALTO-5983: Okta: Add deploy support for `EmailDomain`

Nurdok:
Add deploy support for `EmailDomain`: add, modify, remove.

---

---
_Release Notes_: 
_Okta adapter_:
- Added deploy support for `EmailDomain`
---
_User Notifications_: 
None


packages/okta-adapter/src/change_validators/email_domain_addition.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  return awu(changes)
+    .filter(isInstanceChange)
+    .filter(isAdditionChange)
+    .map(getChangeData)
+    .filter(instance => instance.elemID.typeName === EMAIL_DOMAIN_TYPE_NAME)
+    .filter(async emailDomainInstance =>
+      // Check if there is at least one brand that uses the email domain, return True if there isn't.
+      _.isEmpty(await findReferencingBrands(emailDomainInstance.elemID, elementsSource)),
```
shir-reifenberg, line 31: I think we should first get all brands from the element source, and not doing this per change

packages/okta-adapter/src/change_validators/email_domain_addition.ts
```diff
@@ -0,0 +1,40 @@
...
+    )
+    .map(instance => ({
+      elemID: instance.elemID,
+      severity: 'Error' as const,
+      message: 'Cannot add email domain without at least one brand that uses it',
+      detailedMessage: 'Cannot add email domain without at least one brand that uses it',
```
shir-reifenberg, line 37: the `detailedMessage` should ideally give a user some instructions about how he can resolve this issue.
maybe change to something in the spirit of -
`To deploy this change, ....`

packages/okta-adapter/src/definitions/deploy/types/email_domain.ts
```diff
@@ -0,0 +1,62 @@
...
+export const addBrandIdToRequest: definitions.AdjustFunction<definitions.deploy.ChangeAndContext> = async ({
+  value,
+  context,
+}) => {
+  validatePlainObject(value, EMAIL_DOMAIN_TYPE_NAME)
+  const emailDomainElemId = getChangeData(context.change).elemID
+  const [brand] = (await findReferencingBrands(emailDomainElemId, context.elementSource)) ?? []
+
+  if (brand === undefined) {
+    throw new Error(`Brand not found for email domain ${emailDomainElemId.getFullName()}`)
```
shir-reifenberg, line 53: please add a `log.error` here as well

packages/okta-adapter/src/definitions/deploy/types/email_domain.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+  // Use the ID directly instead of a reference value to avoid circular references.
```
shir-reifenberg, line 55: I think this comment is redundant

shir-reifenberg (commented):
Looks great, added some comments (mainly around the usage of the `elementSource`)

packages/okta-adapter/src/change_validators/email_domain_addition.ts
Nurdok: We do this per-change in the transform function, so if it's critical we perform this filtering only once, then maybe the transformation should happen in a filter instead so we can optimize it too?

Not sure if this is really costly in absolute time.

shir-reifenberg: there is actually cache in the `elementSource`, so I _think_ we should be fine here  (we can start like this and adjust if needed)
another alternative for the deploy is store the results of `findReferencingBrands` in the `sharedContext` and then call `findReferencingBrands` only if it's empty.

packages/okta-adapter/src/definitions/deploy/types/email_domain.ts
Nurdok: I actually tried a reference before using the ID and found this out the hard way, so I'd rather keep it :)

PR #5954, opened 2024-05-23
SALTO-5910:  Add sourceScope to reference indexes

tamtamirr:
Add sourceScope to ReferenceInfo.

---

This is a required step for Salesforce Profiles Custom References.

PR changes:
1. Add **sourceScope** to **referenceTargets** entries. (optional, defaults to `baseId` when undefined)
2. Return `sourceScope: value` for Salesforce **Profiles** custom references.
3. Add **getElementIncomingReferenceInfos** method to workspace, that returns additional info on the incoming references (reference `type` and `sourceScope`).
4. **Bumping** the reference index to apply the changes.

---
_Release Notes_: 
Workspace:
- `getElementOutgoingReferences` will now additionally return the `sourceScope` for each reference.
- Implemented `getElementIncomingReferenceInfos` that returns additional information on the incoming reference (The reference type and the sourceType)

---
_User Notifications_: 
_None_


packages/workspace/src/workspace/reference_indexes.ts
```diff
@@ -42,15 +41,33 @@ import { RemoteMap, RemoteMapEntry } from './remote_map'
...
-export type ReferenceTargetIndexValue = collections.treeMap.TreeMap<{ id: ElemID; type: ReferenceType }>
+export type ReferenceIndexEntry = {
+  id: ElemID
+} & Pick<ReferenceInfo, 'type' | 'sourceScope'>
+
+export type SerializedReferenceIndexEntry = {
+  id: string
+} & Pick<ReferenceInfo, 'type' | 'sourceScope'>
+
+const isValidReferenceSourceScope = (value: unknown): value is ReferenceInfo['sourceScope'] =>
```
ori-moisis, line 60: since that type is exported anyway, might as well use it directly.
It would also make it easier to find this place if anyone ever changes to possible values in `ReferenceSourceScope`
```suggestion
const isValidReferenceSourceScope = (value: unknown): value is ReferenceSourceScope =>
```

This is still not ideal because there is nothing enforcing that all the possible values are caught here AFAICT

packages/workspace/src/workspace/reference_indexes.ts
```diff
@@ -42,15 +41,33 @@ import { RemoteMap, RemoteMapEntry } from './remote_map'
...
+  value === undefined || value === 'baseId' || value === 'value'
+
+const isValidReferenceType = (value: unknown): value is ReferenceInfo['type'] => value === 'strong' || value === 'weak'
```
ori-moisis, line 63:
```suggestion
const isValidReferenceType = (value: unknown): value is ReferenceType => value === 'strong' || value === 'weak'
```

packages/workspace/src/workspace/reference_indexes.ts
```diff
@@ -131,7 +148,14 @@ const createReferenceTree = (references: ReferenceInfo[], rootFields = false): R
         rootFields && ref.source.idType === 'field'
           ? ''
           : ref.source.createBaseID().path.join(ElemID.NAMESPACE_SEPARATOR)
-      return [key, [{ id: ref.target, type: ref.type }]]
+      return [
```
ori-moisis, line 151: would it be a bit cleaner to implement this with `_.pickBy(..., values.isDefined)` maybe?

packages/workspace/src/workspace/reference_indexes.ts
```diff
@@ -189,21 +213,27 @@ const updateReferenceTargetsIndex = async (
...
-  const referenceSourcesChanges = _(references)
+const getReferenceSourcesMap = (references: ReferenceInfo[]): Record<string, ReferenceIndexEntry[]> => {
+  const referenceSourcesChanges: Record<string, ReferenceIndexEntry[]> = _(references)
     .groupBy(({ target }) => target.createBaseID().parent.getFullName())
-    .mapValues(refs => refs.map(ref => ref.source))
+    .mapValues(refs =>
+      refs.map(ref =>
+        ref.sourceScope
+          ? { id: ref.source, type: ref.type, sourceScope: ref.sourceScope }
+          : { id: ref.source, type: ref.type },
```
ori-moisis, line 234: similar comment, would this look nicer with a `_.pickBy`?

packages/workspace/src/workspace/reference_indexes.ts
```diff
@@ -256,7 +286,7 @@ const updateReferenceSourcesIndex = async (
   const [deletions, modifications] = _.partition(updates, update => update.value.length === 0)
   const uniqueModification = modifications.map(modification => ({
     ...modification,
-    value: _.uniqBy(modification.value, id => id.getFullName()),
+    value: _.uniqBy(modification.value, entry => entry.id.getFullName()),
```
ori-moisis, line 289: not actually related to the `uniqBy` part, but there is filtering in `updateIdOfReferenceSourcesIndex` that could cause the attributes of a reference not to be updated

need to test the case where an existing reference changes its type or source type

packages/workspace/src/workspace/reference_indexes.ts
tamtamirr: Using _.pickBy will lose the type of the object and return `Dictonary<value type>` instead. This yields the following complilation error:
`Property 'id' is missing in type 'Dictionary<ElemID | "baseId" | "value" | "strong" | "weak">' but required in type '{ id: ElemID; }'.`

packages/workspace/src/workspace/reference_indexes.ts
tamtamirr: Changed the code so adding new values won't break :)

packages/workspace/src/workspace/reference_indexes.ts
tamtamirr: This is covered in UT now :) This is actually working as expected. See L287 in reference_indexes.test.ts

packages/workspace/src/workspace/workspace.ts
```diff
@@ -1302,13 +1332,25 @@ export const loadWorkspace = async (
...
+      }))
     },
     getElementIncomingReferences: async (id, envName = currentEnv()) => {
       if (!id.isBaseID()) {
         throw new Error(`getElementIncomingReferences only support base ids, received ${id.getFullName()}`)
       }
-      return (await (await getWorkspaceState()).states[envName].referenceSources.get(id.getFullName())) ?? []
+      return makeArray(await (await getWorkspaceState()).states[envName].referenceSources.get(id.getFullName())).map(
+        entry => entry.id,
+      )
```
ori-moisis, line 1346: readability nit - I think it would be nicer to separate getting the entires from running the `.map` (like you did in the next function)

packages/workspace/src/workspace/workspace.ts
```diff
@@ -1302,13 +1332,25 @@ export const loadWorkspace = async (
...
+    },
+    getElementIncomingReferenceInfos: async (id, envName = currentEnv()) => {
+      if (!id.isBaseID()) {
+        throw new Error(`getElementIncomingReferenceInfos only support base ids, received ${id.getFullName()}`)
+      }
+      const entries = (await (await getWorkspaceState()).states[envName].referenceSources.get(id.getFullName())) ?? []
```
ori-moisis, line 1352: consistency nit: `makeArray` or `?? []` in adjacent functions, pick one and go with it

packages/workspace/src/workspace/workspace.ts
```diff
@@ -180,6 +186,10 @@ export type FromSourceWithEnv = {
   envName?: string
 }
 
+type IncomingReferenceInfo = {
+  id: ReferenceInfo['source']
+} & Required<Pick<ReferenceInfo, 'type' | 'sourceScope'>>
```
ori-moisis, line 191: is this just `Required<ReferenceIndexEntry>`?
maybe we don't need this type anymore?

packages/workspace/src/workspace/workspace.ts
```diff
@@ -354,6 +365,27 @@ const compact = (sortedIds: ElemID[]): ElemID[] => {
...
+
+export const deserializeReferenceSourcesEntries = async (data: string): Promise<ReferenceIndexEntry[]> => {
+  const parsedEntries: unknown = JSON.parse(data)
+  if (!Array.isArray(parsedEntries)) {
+    log.warn('failed to deserizlize reference sources entries. Parsed Entries: %s', inspectValue(parsedEntries))
+    throw new Error('Failed to deserialize reference sources entries')
+  }
+  // Handle old format
+  if (parsedEntries.some(_.isString)) {
+    log.debug('Deserializing old reference sources entries format')
```
ori-moisis, line 379: isn't this log extremely spammy?
meaning, if it happens, it will be logged a lot of times, no?

not sure if there is a way to avoid it easily and IIUC because of the bump to the index version it shouldn't really happen. but still, if you can think of an easy way to restrict this log from happening too much, that would be nice

packages/workspace/src/workspace/workspace.ts
```diff
@@ -354,6 +365,27 @@ const compact = (sortedIds: ElemID[]): ElemID[] => {
   return ret
 }
 
+export const serializeReferenceSourcesEntries = async (entries: ReferenceIndexEntry[]): Promise<string> =>
```
ori-moisis, line 368: should we maybe move these to the reference indexes file?

packages/workspace/src/workspace/workspace.ts
```diff
@@ -1302,13 +1341,25 @@ export const loadWorkspace = async (
...
+      return entries.map(entry => entry.id)
+    },
+    getElementIncomingReferenceInfos: async (id, envName = currentEnv()) => {
+      if (!id.isBaseID()) {
+        throw new Error(`getElementIncomingReferenceInfos only support base ids, received ${id.getFullName()}`)
+      }
+      const entries = makeArray(
+        await (await getWorkspaceState()).states[envName].referenceSources.get(id.getFullName()),
+      )
+      return entries.map(toReferenceIndexEntryWithDefaults)
```
ori-moisis, line 1362: instead of repeating the implementation of `getElementIncomingReferences`, maybe make `getElementIncomingReferences` use the new function?
IIUC the implementation of `getElementIncomingReferences` could just be
```typescript
(await getElementIncomingReferenceInfos(id, envName)).map(info => info.id)
```

ori-moisis (approved):
mostly style comments at this point
logic lgtm

packages/workspace/src/workspace/workspace.ts
tamtamirr: Nope, as `ReferenceInfo` contains `source` and `target` fields. We need only `id` at this point. (either of source or target)

packages/workspace/src/workspace/workspace.ts
tamtamirr: Considering the serializer & deserializer of the `targetsIndex` sits here I've decided to implement them here as-well for alignment. Would you prefer moving them all?

packages/workspace/src/workspace/workspace.ts
ori-moisis: I wrote `ReferenceIndexEntry`, not `ReferenceInfo`

packages/workspace/src/workspace/workspace.ts
```diff
@@ -368,6 +368,7 @@ const compact = (sortedIds: ElemID[]): ElemID[] => {
 export const serializeReferenceSourcesEntries = async (entries: ReferenceIndexEntry[]): Promise<string> =>
   safeJsonStringify(entries.map(entry => ({ ...entry, id: entry.id.getFullName() })))
 
+let hasLoggedOldFormatWarning = false
```
ori-moisis, line 371: we should not use global state
this is setting a very very dangerous precedent

I'd rather have the spammy log than this

tamtamirr:
This is total size of the caches for

```
➜  caches du -sh ./branch
 22M	./branch
➜  caches du -sh ./main
 22M	./main
➜  caches du -sh ./main-np  // main no profiles
 16M	./main-np
➜  caches du -sh ./branch-np  // branch no profiles
 16M	./branch-np
```

PR #5955, opened 2024-05-23
SALTO-5884: Support Microsoft Entra adapter

shoshana-houri:
Support basic fetch + partial deploy for Microsoft Entra adapter, using [Microsoft Graph API](https://learn.microsoft.com/en-us/graph/api/overview?view=graph-rest-1.0).

---
_Release Notes_: 
Microsoft Entra adapter:
* First version (basic fetch + partial deploy) is ready to use.

---
_User Notifications_: 
Microsoft Entra adapter:
* First version (basic fetch + partial deploy) is ready to use.


packages/adapter-components/src/definitions/system/fetch/element.ts
```diff
@@ -105,7 +105,7 @@ export type ElementFieldCustomization = types.XOR<
   },
   types.OneOf<{
     // omit the field
-    omit: true
+    omit: boolean
```
shoshana-houri, line 108: I wanted to be able to omit some fields by default but still be able to override them for certain types. Please LMK if you think I should achieve this in a different way.

packages/microsoft-entra-adapter/README.md
```diff
@@ -0,0 +1,5 @@
+# Microsoft Entra adapter
+
+Microsoft Entra adapter for salto.io
+
+TODO: Add more information about the adapter
```
netama, line 5: let's replace this (even with 1-2 sentences)?

packages/microsoft-entra-adapter/specific-cli-options.md
```diff
@@ -0,0 +1,17 @@
+# Microsoft Entra CLI options
+
+## Non interactive Login Parameters
+
+Supported parameters are:
+
+TODO adjust according to auth
```
netama, line 7: same?

packages/microsoft-entra-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -0,0 +1,366 @@
...
+      ...customization.requestsByAction,
+      customizations: {
+        ..._.mapValues(customization.requestsByAction?.customizations, actionCustomizations =>
+          (actionCustomizations ?? []).map(action => ({
+            ...action,
+            request: {
+              ...action.request,
+              ...((action.request.endpoint
+                ? { endpoint: { ...action.request.endpoint, path: `${basePath}${action.request.endpoint.path}` } }
+                : {}) as DeployRequestDefinition['endpoint']),
```
netama, line 65: the cast is not valid technically (it doesn't match any of the XOR options), why do we need it to be this way?

packages/microsoft-entra-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -0,0 +1,366 @@
...
+              transformation: {
+                omit: ['id'],
+                adjust: item => {
+                  if (!isPlainObject(item.value) || !isAdditionChange(item.context.change)) {
+                    throw new Error('Unexpected value, expected a plain object')
+                  }
+                  return {
+                    value: {
+                      ...item.value,
+                      principalId: getParent(item.context.change.data.after).value.id,
```
netama, line 235: `getParent` can throw, is this the intended flow? can we maybe add a check + confirm the id is defined? (and same for other relevant places)

not critical for the first PR, just something to plan for

netama (approved):
very cool ⭐

packages/microsoft-entra-adapter/src/definitions/deploy/deploy.ts
shoshana-houri: @netama Maybe I'm missing something:
Since we first check `action.request.endpoint ?` doesn't it mean we're in the case where the endpoint is provided, and not `earlySuccess`? Therefore matching exactly one of the XOR options?

packages/microsoft-entra-adapter/src/definitions/deploy/deploy.ts
shoshana-houri: @netama Since `principalId` is [required](https://learn.microsoft.com/en-us/graph/api/group-post-approleassignments?view=graph-rest-1.0&tabs=http) when handling with `appRoleAssignment` I think it's ok that it throws - otherwise we have nothing to do.
We can possibly add a CV? I'll add a TODO here and let's discuss this f2f :)

packages/microsoft-entra-adapter/src/definitions/deploy/deploy.ts
netama: then why do you need the cast?

packages/microsoft-entra-adapter/src/definitions/deploy/deploy.ts
netama: the `.id` can be undefined anyway. I think we should have an explicit check here

PR #5956, opened 2024-05-26
SALTO-5984: Okta: Omit `EmailDomain` instances marked as `DELETED`

Nurdok:
This PR fixes a bug when fetching EmailDomain instances:

When deleting an email domain through the UI, the API still returns it, but with validationStatus = "DELETED"

When recreating the same email domain through the UI (same email address and sender name), the API will now return two entries with the same details (one DELETED and one not)

When Salto performs a fetch, we get

```
Error merging okta.EmailDomain.instance.Team_Rachum@s: duplicate key <somekey>
[..]
dropped elements: okta.EmailDomain.instance.Team_Rachum@s
```

We solve this by omitting entries with `DELETED` status on fetch.

---

---
_Release Notes_: 
- Fixed a bug where email domains are dropped after being deleted and recreated, by omitting all currently-deleted email domains in fetch
---
_User Notifications_: 
- Email domains that have `validationStatus = "DELETED"` will be omitted in future fetches.


shir-reifenberg (commented):
If we're waiting for the merge of fetch changes anyway, I think this can be done without a filter using the new infra `valueGuard` option under `element.topLevel`

PR #5959, opened 2024-05-26
SALTO-3621: Handling type definition cases

yelly:
We currently don't support 2 type definition cases we claim to support:
1. `type <name> is object`
2. Dropping the `is` (this should add an error but still parse correctly)

This PR adds support for both.

---

This is on the way to supporting meta types.

---
_Release Notes_: 
_Core_:
* Added support for `type <name> is object` type definition syntax.

---
_User Notifications_: 
None.


packages/parser/src/parser/internal/native/helpers.ts
```diff
@@ -121,18 +121,10 @@ export const createFieldRefType = (context: ParseContext, blockType: string, ran
...
-  }
-  if (typeName === Keywords.TYPE_UNKNOWN) {
-    return PrimitiveTypes.UNKNOWN
-  }
-  if (typeName === Keywords.TYPE_BOOL) {
-    return PrimitiveTypes.BOOLEAN
-  }
-  return undefined
-}
+export const primitiveType = (typeName: string): PrimitiveTypes | undefined =>
```
ori-moisis, line 124: is this the same mapping just reversed?
how hard would it be to share it between the parse and the dump code?

packages/parser/src/parser/internal/native/consumers/top_level.ts
```diff
@@ -222,10 +231,10 @@ export const consumeVariableBlock = (context: ParseContext): ConsumerReturnType<
   }
 }
 
-// We consider a block def with 2 labels to be a primitive type def with the 'is'
+// We consider a block def with 2 labels to be a specified type def with the 'is'
 // keyword missing, since in all other block types there is only 1 legal label.
-// the primitive type consumer handles the missing 'is'.
-const isPrimitiveTypeDef = (elementType: string, elementLabels: string[]): boolean =>
+// the specified type consumer handles the missing 'is'.
```
ori-moisis, line 236: what is a `specified type`?
this is not a term I am familiar with...

packages/parser/src/parser/internal/native/consumers/top_level.ts
```diff
@@ -263,10 +272,10 @@ export const consumeElement = (context: ParseContext): ConsumerReturnType<Elemen
...
   let consumedElement: ConsumerReturnType<Element | undefined>
   const isSettings = elementType === Keywords.SETTINGS_DEFINITION
-  // Primitive type def actually needs 3 labels, but we assume that 2 labels
-  // is a primitive type def with no inheritance operator
-  if (isPrimitiveTypeDef(elementType, elementLabels)) {
-    consumedElement = consumePrimitive(context, consumedLabels)
+  // Specified type def actually needs 3 labels, but we assume that 2 labels
+  // is a specified type def with no inheritance operator
+  if (isSpecifiedTypeDef(elementType, elementLabels)) {
+    consumedElement = consumeType(context, consumedLabels)
```
ori-moisis, line 278: the flow here seems a bit strange.
if `consumeType` is also supposed to handle creating `ObjectType` it is unclear why there is a separate `consumeObjectType` at this level of the code.
I would expect that if you have a function that handles all types of type definitions and handles the different cases internally anyway, there would not be an extra check at this level

packages/parser/src/parser/dump.ts
```diff
@@ -40,21 +40,13 @@ import { addValuePromiseWatcher, replaceValuePromises } from './internal/native/
...
-  }
-  return Keywords.TYPE_OBJECT
-}
+const getPrimitiveTypeName = (primitiveType: PrimitiveTypes): string =>
+  ({
+    [PrimitiveTypes.STRING]: Keywords.TYPE_STRING,
+    [PrimitiveTypes.NUMBER]: Keywords.TYPE_NUMBER,
+    [PrimitiveTypes.BOOLEAN]: Keywords.TYPE_BOOL,
+    [PrimitiveTypes.UNKNOWN]: Keywords.TYPE_UNKNOWN,
+  })[primitiveType] || Keywords.TYPE_OBJECT
```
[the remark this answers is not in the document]
yelly: Just style, it took me a while to parse the old code. I'll give it a name.

packages/parser/src/parser/internal/native/helpers.ts
yelly: Apparently it's pretty much impossible (unless we're missing something). Reversing object keys and values while keeping the types just doesn't seem to be an option in javascript because it casts all keys to strings, and in this case the keys are numbers.
We can cast back to numbers manually but at this point I think it's better to just leave as is.

yelly: I am moving these out to a different file to keep them together though.

packages/parser/src/parser/dump.ts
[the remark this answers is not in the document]
yelly: Done.

packages/parser/src/parser/internal/native/helpers.ts
yelly: Done.

packages/parser/src/parser/internal/native/consumers/top_level.ts
yelly: A term a just invented to mean a type defined with "is". Do you have a better suggestion?

packages/parser/src/parser/internal/native/consumers/top_level.ts
yelly: Refactored. Now all object consumption is done in one function. This also allows us to support the error case where the type name is missing.

packages/parser/src/parser/internal/native/consumers/top_level.ts
yelly: Went away in the refactor.

packages/parser/src/parser/internal/native/consumers/top_level.ts
```diff
@@ -55,50 +57,69 @@ const INSTANCE_ANNOTATIONS_ATTRS: string[] = Object.values(INSTANCE_ANNOTATIONS)
...
-  const elemID = parseTopLevelID(context, typeName, { ...labels.range, filename: context.filename })
+  isSettings: boolean,
+): ConsumerReturnType<PrimitiveType | ObjectType | undefined> => {
+  // Note - this method is called when the first label is 'type' or 'settings'.
+  let [typeName, kw, baseType] = labels.value.slice(1)
+  const range = { ...labels.range, filename: context.filename }
+
+  // If there is no type name we just generate one.
+  if (typeName === undefined) {
+    typeName = _.uniqueId('UnnamedType')
```
ori-moisis, line 71: AFAICT, this is an actual behavior change, and to me it seems like this is a change for the worse.

IIUC before this change, if the user had a type definition with no name, there would be a parse error and no element, so there would be no effect on the workspace in terms of which elements exist.
this changes it so an actual element with an arbitrary name is created, which means the effect on this parsing issue becomes much more widespread and this could affect many more things down the line (e.g - it will show up in a deploy plan, there could be validation errors on it, etc...)

packages/parser/src/parser/internal/native/consumers/top_level.ts
```diff
@@ -55,50 +57,69 @@ const INSTANCE_ANNOTATIONS_ATTRS: string[] = Object.values(INSTANCE_ANNOTATIONS)
...
+  if (kw === undefined || isSettings) {
+    kw = Keywords.TYPE_INHERITANCE_SEPARATOR
+    baseType = Keywords.TYPE_OBJECT
+  }
 
   // We create an error if some other token is used instead of the 'is' keyword.
-  // We don't need to recover. We'll just pretend the wrong word is 'is' (hihi)
+  // We don't need to recover. We'll just pretend the wrong word is 'is'
   // and parse as usual.
+  // If this leaves us with no baseName, we assume the 'is' was dropped.
```
ori-moisis, line 90: baseName should be typeName or baseType?

packages/parser/src/parser/internal/native/consumers/top_level.ts
```diff
@@ -55,50 +57,69 @@ const INSTANCE_ANNOTATIONS_ATTRS: string[] = Object.values(INSTANCE_ANNOTATIONS)
...
   if (kw !== Keywords.TYPE_INHERITANCE_SEPARATOR) {
-    context.errors.push(
-      invalidPrimitiveTypeDef(
-        {
-          ...labels.range,
-          filename: context.filename,
-        },
-        kw,
+    context.errors.push(invalidTypeDefOperator(range, kw))
+    baseType = baseType ?? kw
```
ori-moisis, line 93: another case where I think you added a recovery that I think is worse than the previous behavior (or at least worse than what I think it was)
IIUC, this will recover no only if the `is` is missing (which, btw, I am not sure why we even need to allow that...), this will recover regardless of what was there, right?
so something like
```
type dummy.MyType dummy.Base1 dummy.Base2 {
```
will now be considered a type definition that is recoverable (meaning, the type will exist).
I don't think we should allow that

packages/parser/src/parser/internal/native/consumers/top_level.ts
```diff
@@ -222,15 +221,6 @@ export const consumeVariableBlock = (context: ParseContext): ConsumerReturnType<
   }
 }
 
-// We consider a block def with 2 labels to be a primitive type def with the 'is'
-// keyword missing, since in all other block types there is only 1 legal label.
-// the primitive type consumer handles the missing 'is'.
-const isPrimitiveTypeDef = (elementType: string, elementLabels: string[]): boolean =>
-  elementType === Keywords.TYPE_DEFINITION && elementLabels.length >= 2 && elementLabels.length < 4
```
ori-moisis, line 229: IIUC removing all checks on the number of labels means cases that used to skip parsing the type and emit an error will no longer do so.
so, things like
```
type dummy.Bla is string and also more text that should be an error {
```
that used to be a parse error will now just be valid, I think that is not what we want...

packages/parser/test/parser/parse.test.ts
```diff
@@ -28,93 +28,826 @@ import {
...
+    describe('string', () => {
+      const body = `
+        type salesforce.string is string {
+        }
+      `
 
-      type salesforce.number is number {
-      }
+      beforeEach(async () => {
+        await parseBody(body)
```
ori-moisis, line 82: not loving the side-effect based programming style
makes it harder to follow where things are initialized

yelly: How would you suggest doing this then? Replacing all of these with:
```
{ elements, sourceMap, errors } = await parseBody(body)
```
?

packages/parser/src/parser/internal/native/consumers/top_level.ts
yelly: Yeah, `baseType`.

packages/parser/src/parser/internal/native/consumers/top_level.ts
yelly: Fair enough. I'll add a check to make sure we don't get here with no type name.

packages/parser/src/parser/internal/native/errors.ts
```diff
@@ -28,24 +28,31 @@ const createError = (range: SourceRange, summary: string, message?: string): Par
...
     range,
-    'unknown primitive type',
-    token ? `Unknown primitive type ${token}.` : 'Expected a primitive type definition.',
+    'Unknown primitive type',
+    token ? `Unknown primitive type ${token}.` : 'Expected a primitive type definition, using unknown.',
   )
 
-export const invalidPrimitiveTypeDef = (range: SourceRange, token: string): ParseError =>
-  createError(range, 'invalid type definition', `Expected inheritance operator 'is' found ${token} instead`)
+export const invalidTypeDefOperator = (range: SourceRange, token: string): ParseError =>
```
ori-moisis, line 35: nit: type is not an operator

packages/parser/test/parser/parse.test.ts
ori-moisis: now sure if you can do it in a single line, but essentially, yes

packages/parser/src/parser/internal/native/consumers/top_level.ts
yelly: Per our conversation F2F, moved to stricter validations. Now all of these cases will give an "ambiguous block" error.

packages/parser/src/parser/internal/native/errors.ts
yelly: This error no longer exists.

packages/parser/test/parser/parse.test.ts
yelly: Done.

packages/parser/test/parser/parse.test.ts
```diff
@@ -28,93 +28,780 @@ import {
...
+      })
+
+      it('should parse type', () => {
+        expect(elements).toHaveLength(1)
+        const numberType = elements[0] as PrimitiveType
+        expect(numberType.primitive).toBe(PrimitiveTypes.NUMBER)
+        expect(numberType.elemID).toEqual(new ElemID('salesforce', 'number'))
+      })
+
+      it('should contain all elements in source map', validateSourceMap)
```
ori-moisis, line 118: this feels like a very superficial test of the source map, IIUC it only checks that the top level element exists somewhere in the source map.
there is no validation of the actual source range, there is no check of the internal parts.

not sure if this was tested previously, but I do think that with the new structure, there is no place that explicitly tests the source map, so it is harder to keep track of that and you'd have to split all the use cases to be tested throughout the different scenarios, so edge cases for the source map will need to appear somewhere (e.g - comments, extra spaces, extra line breaks, etc...)

packages/parser/test/parser/parse.test.ts
```diff
@@ -28,93 +28,780 @@ import {
...
+      const body = `
+        type salesforce.unknown is invalid {
+        }
+      `
+
+      beforeEach(async () => {
+        ;({ elements, sourceMap, errors } = await parseBody(body))
+      })
+
+      it('should parse type', () => {
```
ori-moisis, line 177: this is not really a good description of the expected behavior

packages/parser/test/parser/parse.test.ts
```diff
@@ -28,93 +28,780 @@ import {
...
-          fieldLevelSecurity = {
-            all_profiles = {
-              visible = false
-              read_only = false
-              //comment inside an object
+      beforeEach(async () => {
+        ;({ elements, sourceMap, errors } = await parseBody(body))
+      })
+
+      it('should parse type', () => {
```
ori-moisis, line 155: this is not a good description of the expected behavior IMO, the previous tests were more explicit
e.g -
```typescript
describe('number type', () => {
      it('should have the correct type', () => {
        expect(numberType.primitive).toBe(PrimitiveTypes.NUMBER)
      })
```

also, since you opened the door to writing a full suite of new tests, shouldn't there be a test for having annotations / annotation types inside a primitive type somewhere?
do we have a test for a primitive type trying to define fields?

packages/parser/test/parser/parse.test.ts
```diff
@@ -28,93 +28,780 @@ import {
...
+        const error = errors[0]
+        expect(error.summary).toEqual('Ambiguous block definition')
+        expect(elements).toHaveLength(0)
+      })
+    })
+  })
+
+  describe('object type', () => {
+    describe('labels', () => {
+      describe('full definition', () => {
```
ori-moisis, line 215: I think you are under-describing the scenarios
the target is that it should be clear what the scenario under test is for people who have not read the code

rule of thumb, similar to how every every `it(` is followed by `'should ...`, every `describe(` should be followed by `'when ...` or `'with ...`
```
describe('when/with <description of scenario / input>', () => {
  it('should <expected behavior / output>', () => {
```

packages/parser/test/parser/parse.test.ts
```diff
@@ -28,93 +28,780 @@ import {
...
-            {
-              input = "bla"
-              output = "foo"
+        beforeEach(async () => {
+          ;({ elements, sourceMap, errors } = await parseBody(body))
+        })
+
+        it('should have errors', () => {
+          expect(errors).toHaveLength(2)
+          expect(errors[0].summary).toEqual('Invalid type name')
```
ori-moisis, line 327: what is the second error supposed to be?

ori-moisis (commented):
didn't finish going over the tests

packages/parser/src/parser/internal/native/consumers/top_level.ts
```diff
@@ -222,14 +195,14 @@ export const consumeVariableBlock = (context: ParseContext): ConsumerReturnType<
...
-
-const isObjectTypeDef = (elementType: string, elementLabels: string[], isSettings: boolean): boolean =>
-  (elementType === Keywords.TYPE_DEFINITION || isSettings) && elementLabels.length === 1
+// Type or settings.
+// Settings can only have a name, types can also be of the form "type <name> is <meta type>".
+const isTypeDef = (elementType: string, elementLabels: string[]): boolean =>
```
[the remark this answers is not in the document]
yelly: Done.

packages/parser/src/parser/internal/native/consumers/top_level.ts
```diff
@@ -55,50 +54,46 @@ const INSTANCE_ANNOTATIONS_ATTRS: string[] = Object.values(INSTANCE_ANNOTATIONS)
...
-): ConsumerReturnType<PrimitiveType | undefined> => {
-  // Note - this method is called *only* if labels has 4 tokens (the first of which
-  // is 'type' which we can ignore
-  const [typeName, kw, baseType] = labels.value.slice(1)
-  const elemID = parseTopLevelID(context, typeName, { ...labels.range, filename: context.filename })
+): ConsumerReturnType<PrimitiveType | ObjectType | undefined> => {
+  // We know the labels are in one of the following formats:
+  // * type <name>
+  // * settings <name>
+  // * type <name> is <meta type>
```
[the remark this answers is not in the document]
yelly: Well, more technically
```
// * type <name> is <primitive type / object>
```
and in the future this will be
```
// * type <name> is <primitive type / object / meta type>
```
But that's pretty verbose, so changed to using the `type category` nomenclature used in the docs.

packages/parser/test/parser/parse.test.ts
```diff
@@ -28,93 +28,780 @@ import {
...
+      })
+
+      it('should contain all elements in source map', validateSourceMap)
+
+      it('should have no errors', () => {
+        expect(errors).toHaveLength(1)
+        const error = errors[0]
+        expect(error.summary).toEqual('Unknown primitive type')
+        expect(error.message).toEqual("Unknown primitive type 'invalid'.")
+      })
```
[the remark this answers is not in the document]
yelly: Yes.

packages/parser/test/parser/parse.test.ts
yelly: Done.

packages/parser/test/parser/parse.test.ts
yelly: This gives the same coverage as we had before. No tests here actually inspected ranges.
We could improve this, but that's more work. Do you think it's in scope?

packages/parser/test/parser/parse.test.ts
yelly: It seems that according to this approach, we can't really use `describe` blocks to gather tests that cover the same area.
These describe blocks (at least apart from the last one) are similar to the top level describe block in that they talk about the logic being tested and not the case.
I see how the smallest describe block can try to always use "when/with", so I'll do that. If you think I should also change the rest of the hierarchy let's talk about it becuase I don't really how we can do that without making the tests more difficult to parse.

packages/parser/test/parser/parse.test.ts
yelly: Nice catch.

packages/parser/test/parser/parse.test.ts
yelly: Done.
As for more coverage - maybe in a separate PR? We're already doing a lot here.
We can also merge these tests with the tests in errors.test.ts which are doing some very similar things, and figure out a better strategy for testing the source map.

packages/parser/test/parser/errors.test.ts
```diff
@@ -171,19 +113,15 @@ describe('parsing errors', () => {
...
+            end: { byte: 60, col: 10, line: 3 },
             filename: 'file.nacl',
           })
-          expect(res.errors[0].message).toBe("Expected inheritance operator 'is' found tanananana instead")
-          expect(res.errors[0].summary).toBe('invalid type definition')
-        })
-        it('should still create the element', async () => {
-          const element = (await awu(res.elements).toArray())[0] as PrimitiveType
-          expect(element.primitive).toBe(PrimitiveTypes.STRING)
+          expect(res.errors[0].summary).toBe('Ambiguous block definition')
```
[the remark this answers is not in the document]
yelly: Improved the errors.

packages/parser/test/parser/parse.test.ts
ori-moisis: I don't really know where to draw the line on this PR

personally, I would not have re-written all of the tests in this PR, but given that you already did, I have to review the tests, and if I review them, I comment about what I am seeing, including missing tests.

this is part of why I don't like mixing feature PRs with refactor PRs - in a refactor PR it would be clearly out of scope to add new tests, but in a PR where you are adding new features, it isn't out of scope to add tests...

if this really was not covered before, then adding the coverage can be a follow up PR.

packages/parser/test/parser/errors.test.ts
```diff
@@ -140,26 +94,14 @@ describe('parsing errors', () => {
...
-            filename: 'file.nacl',
-          })
-          expect(res.errors[1].message).toBe('Expected a primitive type definition.')
-          expect(res.errors[1].summary).toBe('unknown primitive type')
-          expect(await awu(res.elements).toArray()).toHaveLength(1)
-        })
-        it('should use unknown as the primitive type', async () => {
-          const element = (await awu(res.elements).toArray())[0] as PrimitiveType
-          expect(element.primitive).toBe(PrimitiveTypes.UNKNOWN)
+          expect(res.errors[0].summary).toBe('Invalid type definition')
```
ori-moisis, line 103: are all the new cases of invalid definition covered?

generally, seems like coverage has decreased in the PR (maybe it just needs a rebase, not sure...)

ori-moisis (approved):

packages/parser/test/parser/parse.test.ts
yelly: Discussed F2F. Maybe we'll add more coverage in a different PR.

packages/parser/test/parser/parse.test.ts
yelly: Done.

packages/parser/test/parser/errors.test.ts
yelly: All the new cases are covered.

yelly: I was mistaken, added the missing test case.

PR #5961, opened 2024-05-27
SALTO-5994 Add updatedConfig option to dummy adapter

adfineberg:
Adding the field `updatedConfig` to the dummy adapter configuration will have the adapter return the given value.

---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/dummy-adapter/src/adapter_creator.ts
+++ b/packages/dummy-adapter/src/adapter_creator.ts
@@ -21,6 +21,7 @@ import {
   ObjectType,
   ListType,
   GetCustomReferencesFunc,
+  InstanceElement,
 } from '@salto-io/adapter-api'
 import _ from 'lodash'
 import DummyAdapter from './adapter'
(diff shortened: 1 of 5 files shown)


packages/dummy-adapter/src/adapter_creator.ts
```diff
@@ -59,7 +61,13 @@ const getCustomReferences: GetCustomReferencesFunc = async elements =>
     : []
 
 export const adapter: Adapter = {
-  operations: context => new DummyAdapter(context.config?.value as GeneratorParams),
+  operations: context => {
+    const genParams = context.config?.value as GeneratorParams
+    const updatedConfig = genParams.updatedConfig
+      ? new InstanceElement(ElemID.CONFIG_NAME, configType, genParams.updatedConfig)
+      : undefined
+    return new DummyAdapter(genParams, updatedConfig)
```
tamtamirr, line 69: Considering we create the `updatedConfig` instance from the `GeneratorParams`, I think it makes more sense to have this logic within the constructor of `DummyAdapter`. This is safer as-well IMO. (Considering we receive an InstanceElement and we don't validate it's **ElemID** / values)

packages/dummy-adapter/test/adapter.test.ts
```diff
@@ -130,10 +130,23 @@ describe('dummy adapter', () => {
         }
       })
     })
+    it('should replace the config with the updatedConfig', async () => {
```
tamtamirr, line 133:
```suggestion
    it('should replace the config with the updatedConfig on fetch', async () => {
```

tamtamirr (commented):
LGTM! Added small comments

PR #5964, opened 2024-05-27
SALTO-6000: Add typeName to error handler

shir-reifenberg:
To include the typeName in the fetch warning message / suggestion, we should be able to access it from handler 

---

---
_Release Notes_: 
None 

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/definitions/system/fetch/index.ts
+++ b/packages/adapter-components/src/definitions/system/fetch/index.ts
@@ -18,7 +18,7 @@
 export { ContextCombinationDefinition } from './dependencies'
 export { FetchApiDefinitions, InstanceFetchApiDefinitions, FetchApiDefinitionsOptions } from './fetch'
 export { FetchRequestDefinition } from './request'
-export { ResourceTransformFunc } from './resource'
+export { ResourceTransformFunc, OnErrorHandlerAction } from './resource'
 export {
   ElementFieldCustomization,
   FieldIDPart,
--- a/packages/adapter-components/src/definitions/system/fetch/resource.ts
+++ b/packages/adapter-components/src/definitions/system/fetch/resource.ts
@@ -39,7 +39,7 @@ type ConfigSuggestion = {
 
 export type OnErrorHandlerAction = FailEntireFetch | CustomSaltoError | ConfigSuggestion
 
-type OnErrorHandler = ArgsWithCustomizer<OnErrorHandlerAction, OnErrorHandlerAction, { error: Error, typeName: string }>
+type OnErrorHandler = ArgsWithCustomizer<OnErrorHandlerAction, OnErrorHandlerAction, { error: Error; typeName: string }>
 
 export type FetchResourceDefinition = {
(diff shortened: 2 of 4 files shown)


packages/adapter-components/src/definitions/system/fetch/resource.ts
```diff
@@ -37,9 +37,9 @@ type ConfigSuggestion = {
   value: ConfigChangeSuggestion
 }
 
-type OnErrorHandlerAction = FailEntireFetch | CustomSaltoError | ConfigSuggestion
+export type OnErrorHandlerAction = FailEntireFetch | CustomSaltoError | ConfigSuggestion
 
-type OnErrorHandler = ArgsWithCustomizer<OnErrorHandlerAction, OnErrorHandlerAction, Error>
+type OnErrorHandler = ArgsWithCustomizer<OnErrorHandlerAction, OnErrorHandlerAction, { error: Error, typeName: string }>
```
shoshana-houri, line 42: What do you think about adding it as the `AdditionalArgs` of the `ArgsWithCustomizer` instead?
Which means:
```suggestion
type OnErrorHandler = ArgsWithCustomizer<OnErrorHandlerAction, OnErrorHandlerAction, Error, { typeName: string }>
```

I'm not sure if it really matters, the naming just makes sense to me. Your call :)

shoshana-houri (approved):
Great addition 🌟 🌟 
Left tiny nits and a suggestion :)

packages/adapter-components/src/definitions/system/fetch/resource.ts
```diff
@@ -37,9 +37,9 @@ type ConfigSuggestion = {
   value: ConfigChangeSuggestion
 }
 
-type OnErrorHandlerAction = FailEntireFetch | CustomSaltoError | ConfigSuggestion
+export type OnErrorHandlerAction = FailEntireFetch | CustomSaltoError | ConfigSuggestion
```
shir-reifenberg, line 40: exported to later use in the adapter

packages/adapter-components/src/definitions/system/fetch/index.ts
```diff
@@ -18,7 +18,7 @@
 export { ContextCombinationDefinition } from './dependencies'
 export { FetchApiDefinitions, InstanceFetchApiDefinitions, FetchApiDefinitionsOptions } from './fetch'
 export { FetchRequestDefinition } from './request'
-export { ResourceTransformFunc } from './resource'
+export { ResourceTransformFunc, OnErrorHandlerAction } from './resource'
```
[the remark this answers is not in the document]
shir-reifenberg: no, I export it to use in the adapter

packages/adapter-components/src/definitions/system/fetch/resource.ts
shir-reifenberg: I think the difference is `AdditionalArgs` are used to create the `custom` func, so i'd say in this case `typeName` is more like the `custom` function input and not it's creator

PR #5965, opened 2024-05-28
SALTO-5967: JSM pagination

IdoZakk:
Added pagination to JSM

---

There are two different pagination types. Object Schema uses the regular [Jira API pagination](https://developer.atlassian.com/cloud/jira/platform/rest/v3/intro/#pagination) and the [JSM API pagination](https://developer.atlassian.com/cloud/jira/service-desk/rest/intro/#pagination)

I went over all of the JSM APIs, both in api_config and all the get calls in the filters.

I did not add noise reduction, as the noise would be too big

---
_Release Notes_: 
Jira Adapter:
* Fixed a bug where Object Schemas, Queues, and Request Types would bring partial results for big environments

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/client/pagination/pagination.ts
+++ b/packages/adapter-components/src/client/pagination/pagination.ts
@@ -178,17 +178,16 @@ export const getWithPageOffsetAndLastPagination = (firstPage: number): Paginatio
   return nextPageFullPages
 }
 
-export const getWithOffsetAndLimit = (): PaginationFunc => {
-  // Hard coded "isLast" and "values" in order to fit the configuration scheme which only allows
+export const getWithOffsetAndLimit = (isLastPageField: string): PaginationFunc => {
+  // Hard coded "values" in order to fit the configuration scheme which only allows
   // "paginationField" to be configured
   type PageResponse = {
-    isLast: boolean
     values: unknown[]
     [k: string]: unknown
   }
   const isPageResponse = (responseData: unknown, paginationField: string): responseData is PageResponse =>
     _.isObject(responseData) &&
-    _.isBoolean(_.get(responseData, 'isLast')) &&
+    _.isBoolean(_.get(responseData, isLastPageField)) &&
     Array.isArray(_.get(responseData, 'values')) &&
     _.isNumber(_.get(responseData, paginationField))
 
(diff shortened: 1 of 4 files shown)


packages/adapter-components/src/client/pagination/pagination.ts
```diff
@@ -178,17 +178,16 @@ export const getWithPageOffsetAndLastPagination = (firstPage: number): Paginatio
   return nextPageFullPages
 }
 
-export const getWithOffsetAndLimit = (): PaginationFunc => {
-  // Hard coded "isLast" and "values" in order to fit the configuration scheme which only allows
+export const getWithOffsetAndLimit = (isLastPageField: string): PaginationFunc => {
```
IdoZakk, line 181: This function was not called anywhere before

shir-reifenberg: I think this function has a duplicate version named `offsetAndLimitPagination` that is used in the old infra
maybe we should keep those aligned?

packages/jira-adapter/src/client/pagination.ts
```diff
@@ -19,6 +19,8 @@ import { client as clientUtils } from '@salto-io/adapter-components'
 
 const { makeArray } = collections.array
 
+const JSM_IS_LAST_PAGE = 'isLastPage'
+const JSM_PAGINATION_URL_PATTERN = /^\/rest\/servicedeskapi\/servicedesk/
```
kami-delaroz, line 23: [This line](https://github.com/salto-io/salto/pull/5965/files?diff=unified&w=0#diff-30fe015aa09d08972c73cad827947ad3e4d1c41302efc129301dcdc3d57176a7L2167) doesn't fall in the pattern. what am I missing?

kami-delaroz (approved):

packages/jira-adapter/src/client/pagination.ts
IdoZakk: It is on purpose. I mentioned it in the context- the object schema uses the regular Jira pagination, while the other two use the JSM pagination.

packages/adapter-components/src/client/pagination/pagination.ts
IdoZakk: Thank you! I will ask Neta if she wants me to change the new pagination as well

PR #5966, opened 2024-05-28
SALTO-5691: Add fetch warning for 403

shir-reifenberg:
Add the fetch warning support using custom handler

---

No release notes as this continues the behavior from old infra

---
_Release Notes_: 
None

---
_User Notifications_: 
None


shoshana-houri (approved):
💯 🚀 yay!

PR #5968, opened 2024-05-28
SALTO-5997: Improve JSM object type icon performance

IdoZakk:
Improve the performance of object type icon fetch. Currently, it is done one-by-one.

---

None

---
_Release Notes_: 
Jira Adapter:
* Improved the fetch performance for JSM

---
_User Notifications_: 
None


kami-delaroz (approved):
🔥

PR #5970, opened 2024-05-28
SALTO-6008: Fix Priority Attribute Handling for Okta Policies

idoamit8:
Fixed a bug related to modifying priority attributes in various default Okta policies to ensure consistent deployment.

---

This PR addresses the issue of modifying the priority attribute in various default Okta policies.
The following changes have been made to handle priority attributes more effectively:

AccessPolicyRule Default Policy Rule - Added `priority=99` for modification changes.
ProfileEnrollmentPolicyRule Default Policy Rule - 
Removed these rules from the priority filter as they don't require prioritization.
Added priority=99 for modification changes, not just for addition changes.
MultifactorEnrollmentPolicy Default Policy:

Ensured the priority field is set to its current value during deployment.
Two options were considered:
Deploy it last if there are other priority additions and calculate its priority based on the existing list.
Query its current priority directly before deployment.
AccessPolicy Default Policy:

Added priority: 1 in the preDeploy phase to align with the handling of other default policies/policyRules, as setting this priority resolved the deployment issue.

---
_Release Notes_: 
_Jira Adpater:_
* Fixed a bug in Okta's deployment of several policies and policyRules so we can deploy them. 

---
_User Notifications_: 
_Jira Adpater:_
* Removed the MultifactorEnrollmentPolicy priority instance.


packages/okta-adapter/src/filters/default_rule_deployment.ts
```diff
@@ -22,24 +22,47 @@ import {
...
 type PolicyRule = {
   id: string
   system: boolean
 }
+type PolicyResponse = {
+  data: {
+    priority: number
+  }
+}
+const POLICY_RESPISE_SCHEMA = Joi.object({
```
shir-reifenberg, line 54: typo

packages/okta-adapter/src/filters/policy_priority.ts
```diff
@@ -63,7 +62,9 @@ import { API_DEFINITIONS_CONFIG, OktaSwaggerApiConfig } from '../config'
 const log = logger(module)
 const { awu } = collections.asynciterable
 const { createUrl } = fetch.resource
-export const ALL_SUPPORTED_POLICY_RULE_NAMES = POLICY_RULE_TYPE_NAMES.concat([AUTHORIZATION_POLICY_RULE])
+export const ALL_SUPPORTED_POLICY_RULE_NAMES = POLICY_RULE_TYPE_NAMES.filter(
+  typeName => typeName !== PROFILE_ENROLLMENT_RULE_TYPE_NAME,
+).concat([AUTHORIZATION_POLICY_RULE])
```
shir-reifenberg, line 67: I think it would be easier to follow if we'll just have an explicit list of the supported type names
and we can call it something like `POLICY_RULE_TYPES_WITH_PRIORITY_INSTANCE` (or anything in the spirit)

packages/okta-adapter/src/filters/policy_priority.ts
```diff
@@ -63,7 +62,9 @@ import { API_DEFINITIONS_CONFIG, OktaSwaggerApiConfig } from '../config'
...
 export const ALL_SUPPORTED_POLICY_NAMES = [SIGN_ON_POLICY_TYPE_NAME, MFA_POLICY_TYPE_NAME, PASSWORD_POLICY_TYPE_NAME]
 // Automation is not included in the list of supported policy rules because it is not supported
```
shir-reifenberg, line 69: `// Automation and PofileEnrollmentPolicyRule is not included in the list of supported policy rules because it is not supported`

packages/okta-adapter/src/filters/policy_priority.ts
```diff
@@ -261,9 +261,12 @@ const filter: FilterCreator = ({ definitions, oldApiDefinitions }) => ({
       elements.push(priorityInstance)
     })
     // Remove priority field from the instances
-    policiesRules.concat(Object.values(policyTypeNameToPolicies).flat()).forEach(rule => {
-      delete rule.value.priority
-    })
+    policiesRules
+      .concat(Object.values(policyTypeNameToPolicies).flat())
+      .concat(instances.filter(instance => instance.elemID.typeName === PROFILE_ENROLLMENT_RULE_TYPE_NAME))
```
shir-reifenberg, line 266: complexity wise I think we should extract 
`instances.filter(instance => instance.elemID.typeName === PROFILE_ENROLLMENT_RULE_TYPE_NAME)` 
to be outside of the `concat`

Also, for policies not handled in this filter, I think we can just add the `priority` field into `fieldsToOmit` 
(or in the new version - to `element.fieldCustomizations` (can discuss IRL 😄)

packages/okta-adapter/src/filters/default_rule_deployment.ts
```diff
@@ -119,26 +142,38 @@ const deployDefaultPolicy = async (
   )
 }
 
-/**
- * Deploy addition changes of default rules for AccessPolicy and ProfileEnrollmentPolicy
- * by changing them to modification changes, because default rules automatically created by the service
- */
```
shir-reifenberg, line 125: Plese don't remove it from here
If we add additional purposes for the filter, we should update this comment

packages/okta-adapter/src/filters/default_rule_deployment.ts
```diff
@@ -119,26 +142,38 @@ const deployDefaultPolicy = async (
...
+      .forEach(async instance => {
+        if (instance.elemID.typeName === ACCESS_POLICY_TYPE_NAME) {
+          instance.value.priority = 1
+        } else if (instance.elemID.typeName === MFA_POLICY_TYPE_NAME) {
+          // get the priority from the service
+          const response = await definitions.clients.options.main.httpClient.get({
+            url: `/api/v1/policies/${instance.value.id}`,
+          })
+          if (isPolicyResponse(response)) {
+            instance.value.priority = response.data?.priority
```
shir-reifenberg, line 166: maybe extract the `MFA_POLICY_TYPE_NAME` part to a different section of this filter for readability ? cause the usecase for it is a little different

Also, we should make sure we're making the request for modification changes only, as we're trying to access `id` field. (I'm pretty sure we have a validator blocking addition changes of default policies, but this should be still reflected in this code).

packages/okta-adapter/test/filters/default_rule_deployment.test.ts
shir-reifenberg, line 1: can you please add a test covering the case for MFA policy ? (test the default rule case and the regular rule case)

packages/okta-adapter/src/dependency_changers/mfa_policies_to_default_mfa.ts
```diff
@@ -0,0 +1,52 @@
...
+ * This dependency changer is used to add a dependency from custom MFA policies to the default MFA policy
+ * This dependency is necessary to obtain the default MFA priority before deployment,
+ * preventing race conditions.
+ */
+export const changeDependenciesFromMfaPoliciesToDefaultMfa: DependencyChanger = async changes => {
+  const instanceChanges = Array.from(changes.entries())
+    .map(([key, change]) => ({ key, change }))
+    .filter((change): change is deployment.dependency.ChangeWithKey<ModificationChange<InstanceElement>> =>
+      isInstanceChange(change.change),
+    )
```
shir-reifenberg, line 39: we're casting to  `ModificationChange` without filtering here
do we even need the casting ? and should it only be on modification changes ?

packages/okta-adapter/test/dependency_changers/mfa_policies_to_default_mfa.test.ts
```diff
@@ -0,0 +1,58 @@
...
+    id: '2',
+    name: 'mfaInstanceTwo',
+    system: false,
+  })
+  const defaultMfaInstance = new InstanceElement('defaultMfaInstance', MfaPolicyType, {
+    id: '3',
+    name: 'defaultMfaInstance',
+    system: true,
+  })
+  it('should add dependencies from MfaPolicy to DefaultMfa', async () => {
```
shir-reifenberg, line 40: nit: maybe use the type's name (and in all other places too ?) - `MultifactorEnrollmentPolicy`

packages/okta-adapter/test/dependency_changers/mfa_policies_to_default_mfa.test.ts
```diff
@@ -0,0 +1,58 @@
...
+    const defaultMfaInstanceAfter = defaultMfaInstance.clone()
+    defaultMfaInstanceAfter.value.name = 'afterMfaInstance'
+    const inputChanges = new Map([
+      [0, toChange({ before: defaultMfaInstance, after: defaultMfaInstanceAfter })],
+      [1, toChange({ after: mfaInstnace })],
+      [2, toChange({ after: mfaInstanceTwo })],
```
shir-reifenberg, line 46: let's have one as removal change ?

packages/okta-adapter/test/dependency_changers/mfa_policies_to_default_mfa.test.ts
```diff
@@ -0,0 +1,58 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { InstanceElement, toChange, DependencyChange, ElemID, ObjectType } from '@salto-io/adapter-api'
+import { collections } from '@salto-io/lowerdash'
+import { MFA_POLICY_TYPE_NAME, OKTA } from '../../src/constants'
+import { changeDependenciesFromMfaPoliciesToDefaultMfa } from '../../src/dependency_changers/mfa_policies_to_default_mfa'
+
+describe('changeDependenciesFromMfaPoliciesToDefaultMfa', () => {
```
shir-reifenberg, line 22: can you please add a test to check we're not adding dependencies when the default  MultifactorEnrollmentPolicy is not deployed ?

packages/okta-adapter/src/dependency_changers/mfa_policies_to_default_mfa.ts
```diff
@@ -0,0 +1,52 @@
...
+import _ from 'lodash'
+import { deployment } from '@salto-io/adapter-components'
+import { MFA_POLICY_TYPE_NAME } from '../constants'
+
+/*
+ * This dependency changer is used to add a dependency from custom MFA policies to the default MFA policy
+ * This dependency is necessary to obtain the default MFA priority before deployment,
+ * preventing race conditions.
+ */
+export const changeDependenciesFromMfaPoliciesToDefaultMfa: DependencyChanger = async changes => {
```
shir-reifenberg, line 34: nit
```suggestion
export const defaultMultifactorEnrollmentPolicyDependency: DependencyChanger = async changes => {
```

packages/okta-adapter/src/dependency_changers/mfa_policies_to_default_mfa.ts
```diff
@@ -0,0 +1,52 @@
...
+  dependencyChange,
+  getChangeData,
+  isInstanceChange,
+} from '@salto-io/adapter-api'
+import _ from 'lodash'
+import { deployment } from '@salto-io/adapter-components'
+import { MFA_POLICY_TYPE_NAME } from '../constants'
+
+/*
+ * This dependency changer is used to add a dependency from custom MFA policies to the default MFA policy
```
shir-reifenberg, line 30: nit: maybe it's just me but I think it would clearer to understand if you write that we're deploying the default MFA policy last

shir-reifenberg (commented):
Great job! and thanks for the quick fix 💪 
few more general questions / comments - 
1. do we need a special treatment of setting fake priority to `IdentityProviderPolicy` too? (see SALTO-5325)
2. The change of removing the priority instance is noise, we should either have a prod ticket / noise reduction for this one (
3. can you please update release notes and user notifcations ?

idoamit8:
> Great job! and thanks for the quick fix 💪 few more general questions / comments -
> 
> 1. do we need a special treatment of setting fake priority to `IdentityProviderPolicy` too? (see SALTO-5325)
> 2. The change of removing the priority instance is noise, we should either have a prod ticket / noise reduction for this one (
> 3. can you please update release notes and user notifcations ?

1. I think we can't modify the default IdentityProviderPolicy. Am I correct?
2.  I will add noise reduction. 
3. Updated.

packages/okta-adapter/src/filters/default_rule_deployment.ts
idoamit8: I just moved it to be on top of the deployment section, but I put it back on top of the filter..

packages/okta-adapter/src/filters/policy_priority.ts
idoamit8: Lets discuss the other part IRL.

packages/okta-adapter/src/dependency_changers/mfa_policies_to_default_mfa.ts
idoamit8: The casting is by mistake to ModificationChange. I changed it to change.

packages/okta-adapter/test/dependency_changers/mfa_policies_to_default_mfa.test.ts
idoamit8: not sure I understand

PR #5974, opened 2024-05-29
SALTO-6008: Fix Priority Attribute Handling for Okta Policies

idoamit8:
Fixed a bug related to modifying priority attributes in various default Okta policies to ensure consistent deployment.

---

This PR addresses the issue of modifying the priority attribute in various default Okta policies.
The following changes have been made to handle priority attributes more effectively:

AccessPolicyRule Default Policy Rule - Added `priority=99` for modification changes.
ProfileEnrollmentPolicyRule Default Policy Rule - 
Removed these rules from the priority filter as they don't require prioritization.
Added priority=99 for modification changes, not just for addition changes.
MultifactorEnrollmentPolicy Default Policy:

Ensured the priority field is set to its current value during deployment.
Two options were considered:
Deploy it last if there are other priority additions and calculate its priority based on the existing list.
Query its current priority directly before deployment.
AccessPolicy Default Policy:

Added priority: 1 in the preDeploy phase to align with the handling of other default policies/policyRules, as setting this priority resolved the deployment issue.

---
_Release Notes_: 
_Jira Adpater:_
* Fixed a bug in Okta's deployment of several policies and policyRules so we can deploy them. 

---
_User Notifications_: 
_Jira Adpater:_
* Removed the MultifactorEnrollmentPolicy priority instance.


packages/okta-adapter/src/dependency_changers/mfa_policies_to_default_mfa.ts
shir-reifenberg, line 1: please change file name to be aligned with the dependency changer name

packages/okta-adapter/src/filters/default_rule_deployment.ts
```diff
@@ -126,17 +150,39 @@ const deployDefaultPolicy = async (
...
   name: 'defaultPolicyRuleDeployment',
   preDeploy: async changes => {
+    // Get the priority of the default MultifactorEnrollmentPolicy
+    await awu(changes)
+      .filter(isInstanceChange)
+      .filter(isModificationChange)
+      .filter(
+        change =>
+          getChangeData(change).elemID.typeName === MFA_POLICY_TYPE_NAME && getChangeData(change).value.system === true,
+      )
```
shir-reifenberg, line 160: there should be only one with `system = true`, no?
In this case please use `.find` and add logs on failures scenarios

packages/okta-adapter/src/filters/default_rule_deployment.ts
```diff
@@ -126,17 +150,39 @@ const deployDefaultPolicy = async (
 const filterCreator: FilterCreator = ({ definitions, oldApiDefinitions }) => ({
   name: 'defaultPolicyRuleDeployment',
   preDeploy: async changes => {
+    // Get the priority of the default MultifactorEnrollmentPolicy
```
shir-reifenberg, line 153: Let's update the main comment of the filter with this information, cause the filter now responsible for more than just turning additions to modifications

packages/okta-adapter/src/filters/default_rule_deployment.ts
```diff
@@ -126,17 +150,39 @@ const deployDefaultPolicy = async (
...
+        const response = await definitions.clients.options.main.httpClient.get({
+          url: `/api/v1/policies/${instance.value.id}`,
+        })
+        if (isPolicyResponse(response)) {
+          instance.value.priority = response.data?.priority
+        }
+      })
+
     changes
       .filter(isInstanceChange)
```
shir-reifenberg, line 172: nit: add this in the top of the filter instead of filtering twice

const instances = changes.filter(isInstanceChange)

packages/okta-adapter/src/filters/default_rule_deployment.ts
```diff
@@ -126,17 +150,39 @@ const deployDefaultPolicy = async (
...
           getChangeData(change).value.system === true,
       )
       .map(change => getChangeData(change))
       .forEach(instance => {
-        instance.value.priority = 99
+        if (instance.elemID.typeName === ACCESS_POLICY_TYPE_NAME) {
+          instance.value.priority = 1
+        } else {
+          instance.value.priority = 99
+        }
```
shir-reifenberg, line 185: maybe add comment explaining those seems to be random numbers ? 😄 
(e.g `default AccessPolicyRule and ProfileEnrollmentRule must last and therefore deployed with priority 99`, and same for the `AccessPolicy`)

packages/okta-adapter/src/filters/default_rule_deployment.ts
```diff
@@ -126,17 +150,39 @@ const deployDefaultPolicy = async (
...
+        change =>
+          getChangeData(change).elemID.typeName === MFA_POLICY_TYPE_NAME && getChangeData(change).value.system === true,
+      )
+      .map(change => getChangeData(change))
+      .forEach(async instance => {
+        const response = await definitions.clients.options.main.httpClient.get({
+          url: `/api/v1/policies/${instance.value.id}`,
+        })
+        if (isPolicyResponse(response)) {
+          instance.value.priority = response.data?.priority
```
shir-reifenberg, line 167: nit: I think `?` is redundant as you already validated the schema
```suggestion
          instance.value.priority = response.data.priority
```

shir-reifenberg (approved):
Looks great, thank you! 👑 
Added some more comments and nits but LGTM otherwise so approving
review was mainly in - https://github.com/salto-io/salto/pull/5970

PR #5975, opened 2024-05-29
SALTO-6015: (Bugfix) CustomMetadata types are being corrupt on targeted fetch of CustomObjects

tamtamirr:
CustomMetadata types are being corrupt on targeted fetch of CustomObjects.

---

This caused having invalid types in the Workspace until the next full fetch, which yields warnings on all of the CustomMetadata Records and cause partial fetch to fail constantly afterwards until the next full fetch.

---
_Release Notes_: 
Salesforce Adapter:
- Bugfix: CustomMetadata types are being corrupt on targeted fetch of CustomObjects.
---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/test/filters/custom_metadata_to_object_type.test.ts
+++ b/packages/salesforce-adapter/test/filters/custom_metadata_to_object_type.test.ts
@@ -27,6 +27,7 @@ import {
   getChangeData,
   CORE_ANNOTATIONS,
 } from '@salto-io/adapter-api'
+import { buildElementsSourceFromElements } from '@salto-io/adapter-utils'
 import { collections } from '@salto-io/lowerdash'
 import filterCreator from '../../src/filters/custom_metadata_to_object_type'
 import { defaultFilterContext } from '../utils'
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/test/filters/custom_metadata_to_object_type.test.ts
```diff
@@ -150,6 +156,58 @@ describe('customMetadataToObjectTypeFilter', () => {
...
+            }),
+          },
+        }) as FilterWith<'onFetch' | 'preDeploy' | 'onDeploy'>
+        await partialFetchFilter.onFetch(elements)
+        customMetadataRecordType = elements
+          .filter(isObjectType)
+          .find(
+            (element) =>
+              apiNameSync(element) === CUSTOM_METADATA_RECORD_TYPE_NAME,
+          ) as ObjectType
```
yelly, line 183: Why do you need the explicit typing if you pass through the `isObjectType` type guard?

tamtamirr: As find will return possibly undefined. While this is hacky, thats the same reason the next line validates the `customMetadataRecordType` is defined.

yelly (approved):

PR #5981, opened 2024-05-30
SALTO-6022 allow isTopLevel to be false

omrifarkash:
- In case directFetch is true we want to allow topLevelDef to be undefined without throwing an error. In this case Instances will not be created but the fetched data can be used for recurseInto or dependsOn types

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/test/fetch/element/element.test.ts
+++ b/packages/adapter-components/test/fetch/element/element.test.ts
@@ -52,7 +52,7 @@ describe('element', () => {
       const generator = getElementGenerator({
         adapterName: 'myAdapter',
         defQuery: queryWithDefault<InstanceFetchApiDefinitions, string>({
-          customizations: { myType: { element: { topLevel: { isTopLevel: true } } } },
+          customizations: { myType: { element: {} } },
         }),
         customNameMappingFunctions: {},
       })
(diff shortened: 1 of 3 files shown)


packages/adapter-components/test/fetch/element/element.test.ts
```diff
@@ -52,7 +52,7 @@ describe('element', () => {
       const generator = getElementGenerator({
         adapterName: 'myAdapter',
         defQuery: queryWithDefault<InstanceFetchApiDefinitions, string>({
-          customizations: { myType: { element: { topLevel: { isTopLevel: true } } } },
+          customizations: { myType: { element: { topLevel: { isTopLevel: false } } } },
```
omrifarkash, line 55: Wasn't this UTwrong?

packages/adapter-components/src/fetch/element/instance_element.ts
```diff
@@ -38,8 +38,8 @@ export const generateInstancesWithInitialTypes = <Options extends FetchApiDefini
     log.error('could not find any element definitions for type %s:%s', adapterName, typeName)
   }
   if (!elementDef?.topLevel?.isTopLevel) {
-    const error = `type ${adapterName}:${typeName} is not defined as top-level, cannot create instances`
-    throw new Error(error)
+    log.debug(`type ${adapterName}:${typeName} is not defined as top-level, cannot create instances`)
```
netama, line 41: can we instead add an early return in the one place calling this function (`generate` in `getElementGenerator`)?

netama (approved):
approving since this works - I think it might be better to move it one step up if possible, and keep this function with the assumption it should be called only for top-level. not critical if you have a strong preference though

PR #5982, opened 2024-05-30
SALTO-6016: Removing tests from build

yelly:

---

---
_Release Notes_: 

---
_User Notifications_: 


dantal4:
@yelly are the e2es going to be built? because I __think__ that they're not built automatically by jest.

yelly:
> @yelly are the e2es going to be built? because I **think** that they're not built automatically by jest.

The E2E tests are not build because they actually are build by Jest. The only thing we need to build in the main build are the E2E test environments - see `packages/cli/tsconfig.json` for an example.
This is definitely the case because all the E2Es are passing on this PR.

packages/confluence-adapter/tsconfig.json
```diff
@@ -2,18 +2,9 @@
...
   },
-  "include": [
-    "src/**/*",
-    "src/**/*.json",
-    "test/**/*",
-    "test/**/*.json",
-    "index.ts",
-    "../adapter-components/src/definitions/system/helpers/utils.ts"
-  ],
+  "include": ["src/**/*", "src/**/*.json", "index.ts", "../adapter-components/src/definitions/system/helpers/utils.ts"],
```
ori-moisis, line 7: not your change, but why is there a specific file from adapter-components here?
also, I think that file doesn't actually exist, so.... what?

yelly: Good catch, I wonder why we had this.
Anyway, deleted everywhere.

yelly:
> > This doesn't actually seem to be saving time in the build but is probably a good thing to do anyway.
> 
> there is some cleanup you did as a side effect which is good, but splitting the `tsconfig` for all packages does add some complexity, so, not sure I see why this is a good thing to do with it doesn't save time, what else do we gain here?

It's a good question. I think the main upside is that we will stop shipping our tests (which it seems we currently do) which probably isn't actually a problem but IMO is a smell.
It should also be reducing the size of the output of the build which gets passed around a lot in the CI. The difference is likely not large enough to notice for a specific CI run but might add up.
Finally, I just think this is generally a best practice. I relied a bit on [this](https://kettanaito.com/blog/one-thing-nobody-explained-to-you-about-typescript) blog post in this PR which makes the case better than I can.

ori-moisis:
we don't ship test code
this is configured in the "files" key in each package.json
see https://www.npmjs.com/package/@salto-io/core?activeTab=code for example

yelly:
> we don't ship test code this is configured in the "files" key in each package.json see https://www.npmjs.com/package/@salto-io/core?activeTab=code for example

Fair. So we're left with the output size in the CI and the best practice argument.
Honestly I'm not sure if it's enough to justify the complexity. I lean on the side of "yes" because I think the price is pretty low but if anyone has a strong opinion on this we'll probably just go with that.

ori-moisis (approved):
I don't have a strong objection to the added complexity

yelly:
I ran a full diff compare between the build output and main and it all looks good.

PR #5988, opened 2024-05-30
SALTO-6024: Enforce forward slash on endpoint path

shir-reifenberg:
Enforce forward slash on endpoints in request definitions for both fetch and deploy

---

see https://github.com/salto-io/salto/pull/5984

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/definitions/system/requests/types.ts
+++ b/packages/adapter-components/src/definitions/system/requests/types.ts
@@ -18,10 +18,12 @@ import { Response, ResponseValue } from '../../../client'
 
 export type HTTPMethod = 'get' | 'post' | 'put' | 'patch' | 'delete' | 'head' | 'options'
 
+export type EndpointPath = `/${string}`
+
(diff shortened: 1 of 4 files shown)


packages/adapter-components/src/definitions/system/requests/types.ts
```diff
@@ -21,7 +21,7 @@ export type HTTPMethod = 'get' | 'post' | 'put' | 'patch' | 'delete' | 'head' |
 export type HTTPEndpointIdentifier<ClientOptions extends string> = {
   // specify the client to use to call the endpoint - defaults to the default client as specified in client.default
   client?: ClientOptions
-  path: string
+  path: `/${string}`
```
netama, line 24: nit: wdyt about defining this explicitly as e.g. `export type EndpointPath = `/${string}` (and using this in the other places you're changing)?

netama (approved):
cool :) added a nit but LGTM

PR #5990, opened 2024-06-02
SAAS-6023 Config bug

adfineberg:

---

---
_Release Notes_: 

---
_User Notifications_: 


DanielBarros1 (approved):

ori-moisis:
@adfineberg @DanielBarros1 - this is quite under documented
we should not have PRs where `Replace me with a description of the changes in this PR` is the description
also looks like the ticket has nothing to do with the change, I assume the mistake is that is should be SALTO and not SAAS?

PR #5995, opened 2024-06-02
SALTO-3621: Specifying type references using generics

yelly:
This allows us to enforce that instance types are object types using the type system and will allow us to do the same for meta types in the next PR.

---

---
_Release Notes_: 
None.

---
_User Notifications_: 
None.

### DIFF
--- a/packages/adapter-api/src/elements.ts
+++ b/packages/adapter-api/src/elements.ts
@@ -32,18 +32,12 @@ const log = logger(module)
 
 export const BuiltinTypesRefByFullName: Record<string, TypeReference<PrimitiveType>> = {}
 
-export const createRefToElmWithValue = <T extends TypeElement>(element: T): TypeReference<T> => {
-  // For BuiltinTypes we use a hardcoded list of refs with values to avoid duplicate instances
-  if (isPrimitiveType(element) && element.elemID.getFullName() in BuiltinTypesRefByFullName) {
-    // Technically, all TS knows here is that the type of `element` is `T & PrimitiveType`,
-    // but that doesn't necessarily mean that `T === PrimitiveType`.
-    // We know this is the case though because there aren't (and probably never will be)
-    // any types which extend `PrimitiveType`.
-    return BuiltinTypesRefByFullName[element.elemID.getFullName()] as TypeReference<T>
-  }
-
-  return new TypeReference(element.elemID, element)
-}
+export const createRefToElmWithValue = <T extends TypeElement>(element: T): TypeReference<T> =>
+  // For BuiltinTypes we use hardcoded refs with values to avoid duplicate instances.
+  // If the element ID is in the present we know the element is of type PrimitiveType but TS does not,
+  // so we need to tell it.
+  (BuiltinTypesRefByFullName[element.elemID.getFullName()] as TypeReference<T>) ??
+  new TypeReference(element.elemID, element)
 
 // This is used to allow constructors Elements with Placeholder types
 // to receive TypeElement and save the appropriate Reference
--- a/packages/adapter-api/src/elements.ts
+++ b/packages/adapter-api/src/elements.ts
@@ -32,28 +32,23 @@ const log = logger(module)
 
 export const BuiltinTypesRefByFullName: Record<string, TypeReference<PrimitiveType>> = {}
 
-type CreateRefToElmWithValueFunc = {
-  (element: PrimitiveType): TypeReference<PrimitiveType>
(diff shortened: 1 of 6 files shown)


packages/adapter-api/src/elements.ts
```diff
@@ -30,16 +30,30 @@ const { mapValuesAsync } = promises.object
...
 // This is used to allow constructors Elements with Placeholder types
 // to receive TypeElement and save the appropriate Reference
-const getRefType = (typeOrRef: TypeOrRef): TypeReference =>
-  isTypeReference(typeOrRef) ? typeOrRef : createRefToElmWithValue(typeOrRef)
+const getRefType: GetRefTypeFunc = <T extends TypeElement>(
+  typeOrRef: TypeOrRef<T>,
+): TypeReference<T | PrimitiveType> =>
+  ((value: TypeOrRef<T>): value is TypeReference<T> => isTypeReference(value))(typeOrRef)
+    ? typeOrRef
+    : createRefToElmWithValue(typeOrRef as T)
```
yelly, line 56: Not particularly happy with this but couldn't find a better solution. Open to suggestions.

packages/workspace/src/serializer/elements.ts
```diff
@@ -325,7 +325,7 @@ const generalDeserializeParsed = async <T>(parsed: unknown, staticFileReviver?:
       InstanceElement: v =>
         new InstanceElement(
           v.elemID.nameParts[0],
-          reviveRefTypeOfElement(v),
+          reviveRefTypeOfElement(v) as TypeReference<ObjectType>,
```
yelly, line 328: Feel like maybe this is surfacing somewhere we should be doing some validations but I'm not sure - this whole file seems to be pretty liberal with the `any`s.

packages/adapter-api/src/elements.ts
```diff
@@ -30,16 +30,30 @@ const { mapValuesAsync } = promises.object
...
 // This is used to allow constructors Elements with Placeholder types
 // to receive TypeElement and save the appropriate Reference
-const getRefType = (typeOrRef: TypeOrRef): TypeReference =>
-  isTypeReference(typeOrRef) ? typeOrRef : createRefToElmWithValue(typeOrRef)
+const getRefType: GetRefTypeFunc = <T extends TypeElement>(
+  typeOrRef: TypeOrRef<T>,
+): TypeReference<T | PrimitiveType> =>
+  ((value: TypeOrRef<T>): value is TypeReference<T> => isTypeReference(value))(typeOrRef)
+    ? typeOrRef
+    : createRefToElmWithValue(typeOrRef)
```
yelly, line 56: Not particularly happy with this but couldn't find a better solution. Open to suggestions.

packages/adapter-api/src/elements.ts
```diff
@@ -30,16 +30,30 @@ const { mapValuesAsync } = promises.object
...
+  <T extends TypeElement>(typeOrRef: TypeOrRef<T>): TypeReference<T>
+}
 // This is used to allow constructors Elements with Placeholder types
 // to receive TypeElement and save the appropriate Reference
-const getRefType = (typeOrRef: TypeOrRef): TypeReference =>
-  isTypeReference(typeOrRef) ? typeOrRef : createRefToElmWithValue(typeOrRef)
+const getRefType: GetRefTypeFunc = <T extends TypeElement>(
+  typeOrRef: TypeOrRef<T>,
+): TypeReference<T | PrimitiveType> =>
+  ((value: TypeOrRef<T>): value is TypeReference<T> => isTypeReference(value))(typeOrRef)
```
yelly, line 54: Not particularly happy with this but couldn't find a better solution. Open to suggestions.

ori-moisis (approved):

yelly:
Fixed the tests. There seems to be a race with when `BuiltinTypesRefByFullName` is populated.
At this point, we could return to the original implementation of the function with an explicit case:
```
export const createRefToElmWithValue = <T extends TypeElement>(element: TypeElement): TypeReference<T> =>
  (BuiltinTypesRefByFullName[element.elemID.getFullName()] as TypeReference<T>) ?? new TypeReference(element.elemID, element)
```
With a relevant comment, but I don't think that's as readable as the current implementation (though it is more efficient).
@ori-moisis WDYT?

ori-moisis:
you had me at "more efficient" (I know it is the last thing you wrote, but still...)
I think this might be called a lot...

ori-moisis (approved):

PR #5998, opened 2024-06-03
SALTO-5782: return only cycle from getCycle

ori-moisis:
Before this fix, getCycle would return edges that lead up to the cycle along with the actual cycle This causes issues in depending code which assumed the return of "getCycle" is just the cycle

---

---
_Release Notes_: 
Core:
- Fix issue where plan would go into infinite loop on specific cases of circular dependencies

---
_User Notifications_: 
_None_


shir-reifenberg (approved):
🚀

PR #6002, opened 2024-06-03
SALTO-6036: Modify selectElementsBySelectorsWithoutReferences to also support returning emptyList 

idoamit8:
In this PR I added functionality to  selectElementsBySelectorsWithoutReferences so we will return empty list when no selectors given.

---

The desired behavior for monitoring is that if we set a monitor on our elements without specifying a selector, the monitor should apply to all elements. 

Conversely, for the "Exclude from Compare" feature, if no selector is defined, nothing should be selected. These two behaviors are opposite, and I want to support both.

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/workspace/src/workspace/element_selector.ts
+++ b/packages/workspace/src/workspace/element_selector.ts
@@ -127,13 +127,15 @@ export const selectElementsBySelectorsWithoutReferences = ({
   elementIds,
   selectors,
   includeNested = false,
+  returnEmptyOnEmptySelectors = false,
 }: {
   elementIds: ElemID[]
   selectors: ElementSelector[]
   includeNested?: boolean
+  returnEmptyOnEmptySelectors?: boolean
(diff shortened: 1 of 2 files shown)


packages/workspace/src/workspace/element_selector.ts
```diff
@@ -127,13 +127,15 @@ export const selectElementsBySelectorsWithoutReferences = ({
   elementIds,
   selectors,
   includeNested = false,
+  returnEmptyOnEmptySelectors = false,
```
dantal4, line 130: nit- add that option to `selectElementsBySelectors` too, as those functions intend to be aligned.

dantal4 (approved):
LGTM

ori-moisis:
I don't think we should merge this
it doesn't feel like the right place to put this kind of behavior

PR #6009, opened 2024-06-04
SALTO-6042 remove source.list override patch from ci pipeline

eshedg:
removing a patch that replaced the Ubuntu sources with an alternative mirror, due to some network issue in the past, and is no longer required.

---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,18 +17,6 @@ commands:
       package_name:
         type: string
     steps:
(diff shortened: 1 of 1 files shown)


ori-moisis:
cc: @benisraeladi  - may want to remove this in #5931 as well

ori-moisis (approved):

PR #6015, opened 2024-06-05
SALTO-6034: Okta: e2e: Actually perform the cleanup :/

Nurdok:
TIL `concat` doesn't modify arrays in place, whaddya know...

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/e2e_test/adapter.test.ts
+++ b/packages/okta-adapter/e2e_test/adapter.test.ts
@@ -457,7 +457,7 @@ const getChangesForInitialCleanup = async (
         }),
       ),
   )
-  log.debug(
+  log.info(
     'Cleaning up the environment before starting e2e test: %s',
     cleanupChanges.map(change => getChangeData(change).elemID.getFullName()),
   )
--- a/packages/okta-adapter/e2e_test/adapter.test.ts
+++ b/packages/okta-adapter/e2e_test/adapter.test.ts
@@ -438,7 +438,9 @@ const getChangesForInitialCleanup = async (
   types: ObjectType[],
   client: OktaClient,
 ): Promise<Change<InstanceElement>[]> => {
-  let cleanupChanges: Change<InstanceElement>[] = []
+const removalChanges = element.filter(....)
+const brandChanges = (await createBrandChangesForDeploy(types, client)).filter(...)
+const cleanupChanges = removalChanges.concat(brandChanges)
   cleanupChanges = cleanupChanges.concat(
     elements
       .filter(isInstanceElement)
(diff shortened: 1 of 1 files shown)


packages/okta-adapter/e2e_test/adapter.test.ts
```diff
@@ -379,6 +380,7 @@ const deployChanges = async (adapterAttr: Reals, changes: Change[]): Promise<Dep
   const planElementById = _.keyBy(changes.map(getChangeData), inst => inst.elemID.getFullName())
   return awu(changes)
     .map(async change => {
+      log.debug('Deploying change for %s (%s)', getChangeData(change).elemID.getFullName(), change.action)
```
shir-reifenberg, line 382: is there actually a way to see this log? also, IMO it's not worth having it on `info`..

packages/okta-adapter/e2e_test/adapter.test.ts
```diff
@@ -437,19 +439,28 @@ const getChangesForInitialCleanup = async (
...
-      .map(([before, after]) => toChange({ before: after, after: before })),
+      .filter(isModificationChange) // We can't reverse "add" actions because we'd need the service ID.
+      .map(change =>
+        toChange({
+          before: change.data.after,
+          after: change.data.before,
+        }),
+      ),
+  )
+  log.debug(
```
shir-reifenberg, line 460:
```suggestion
  log.info(
```

packages/okta-adapter/e2e_test/adapter.test.ts
```diff
@@ -437,19 +438,28 @@ const getChangesForInitialCleanup = async (
   types: ObjectType[],
   client: OktaClient,
 ): Promise<Change<InstanceElement>[]> => {
-  const cleanupChanges: Change<InstanceElement>[] = []
-  cleanupChanges.concat(
+  let cleanupChanges: Change<InstanceElement>[] = []
```
shir-reifenberg, line 441: There's a preferences to avoid `let`
so maybe something like - 
```suggestion
const removalChanges = element.filter(....)
const brandChanges = (await createBrandChangesForDeploy(types, client)).filter(...)
const cleanupChanges = removalChanges.concat(brandChanges)
```

shir-reifenberg (approved):
Thanks! 💯 
added a few comments but pre approving

packages/okta-adapter/e2e_test/adapter.test.ts
Nurdok: Yes, if you set `SALTO_LOG_LEVEL` when running locally (it's helpful IMO). I have it configured in my IDE so it's easier to debug these e2e tests.

PR #6018, opened 2024-06-05
SALTO-6049: Fix missing Reference bug in ObjectTypeAttribute

idoamit8:
In this PR I fixed missing Reference bug in ObjectTypeAttribute

---

In the ObjectTypeAttribute, there is a field named additionalValue.
This field can hold two types: ObjectSchemaDefaultReferenceType or ObjectSchemaReferenceType.
Initially, I assumed they could have the same IDs and attempted a hack to determine the reference type.
I have now updated the logic to match the approach used for the typeValueMulti field in ObjectTypeAttribute.

---
_Release Notes_: 
_Jira Adapter:_
* Fixed missing reference bug in ObjectTypeAttribute in the `additionalValue` field.

---
_User Notifications_: 
_Jira Adapter:_
* Users might encounter changes in the additionalValue field within the ObjectTypeAttribute instances, changing from a missing reference to an actual referenc

### DIFF
--- a/packages/jira-adapter/src/reference_mapping.ts
+++ b/packages/jira-adapter/src/reference_mapping.ts
@@ -1306,24 +1306,24 @@ export const referencesRules: JiraFieldReferenceDefinition[] = [
     missingRefStrategy: 'typeAndValue',
     target: { type: OBJECT_TYPE_TYPE },
   },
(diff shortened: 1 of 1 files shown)


IdoZakk (approved):
💯 just fix the user notification part in the PR message

PR #6022, opened 2024-06-05
SALTO-5860-allowEmpty-for-recurseInto

Uri-Bar:
SALTO-5860-allowEmpty-for-recurseInto
---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/adapter-components/src/fetch/resource/subresources.ts
+++ b/packages/adapter-components/src/fetch/resource/subresources.ts
@@ -84,36 +84,34 @@ export const recurseIntoSubresources =
   }): NestedResourceFetcher =>
   async item =>
     Object.fromEntries(
-      (
-        await Promise.all(
-          Object.entries(def.recurseInto ?? {})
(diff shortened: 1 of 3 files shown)


packages/adapter-components/src/fetch/resource/subresources.ts
```diff
@@ -84,36 +84,34 @@ export const recurseIntoSubresources =
...
-            .map(async ([fieldName, recurseDef]) => {
-              const nestedRequestContext = extractRecurseIntoContext(item, recurseDef)
-              const typeFetcher = typeFetcherCreator({
-                typeName: recurseDef.typeName,
-                context: { ...item.context, ...nestedRequestContext },
-              })
-              if (typeFetcher === undefined) {
-                log.debug('no resource fetcher defined for type %s, cannot recurse into resource', recurseDef.typeName)
-                return []
+      await Promise.all(
```
Uri-Bar, line 87: all the diff in this file is the removal of the isEmpty filter in the last line

PR #6023, opened 2024-06-05
SALTO-6057 Fix hanging e2e tests

adfineberg:
Catch final event in jest circus to make sure the printer doesn't hang.

---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/e2e-credentials-store/src/jest-environment/index.ts
+++ b/packages/e2e-credentials-store/src/jest-environment/index.ts
@@ -44,6 +44,10 @@ export class SaltoE2EJestEnvironment extends NodeEnvironment {
   }
 
   handleTestEvent(event: Event): void {
+    if (event.name === 'teardown') {
+      this.runningTasksPrinter.clear()
+      return
+    }
     const { id, type, status } = extractStatus(event) ?? {}
 
     if (!id) return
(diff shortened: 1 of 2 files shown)


packages/e2e-credentials-store/src/jest-environment/index.ts
```diff
@@ -44,6 +44,10 @@ export class SaltoE2EJestEnvironment extends NodeEnvironment {
   }
 
   handleTestEvent(event: Event): void {
+    if (event.name === 'teardown') {
+      this.runningTasksPrinter.clear()
```
ori-moisis, line 48: just in case we don't follow through with fixing the root cause, at least add a log here if the running tasks are not empty?

ori-moisis (approved):

packages/e2e-credentials-store/src/jest-environment/index.ts
```diff
@@ -45,6 +45,10 @@ export class SaltoE2EJestEnvironment extends NodeEnvironment {
 
   handleTestEvent(event: Event): void {
     if (event.name === 'teardown') {
+      this.log.warn(
```
ori-moisis, line 48: if the size is 0 it is not an issue
a warning that always appears even if everything is fine is bad practice

adfineberg: Oops! But it looks like it won't be long for the next step: https://github.com/salto-io/salto/pull/6024

PR #6028, opened 2024-06-06
SALTO_5950: (Jira) fix warnings on requestType field in Automations

SeggevHaimovich:
There were warnings on `requestType` field in `Automation` with an “edit request type” action after fetch.
The error happened because we defined this field as string.
Fixed this by changing the type to unknown.

* As we know this field can be either reference expression or `{ type: string; value: string }`

---

jira ticket: https://salto-io.atlassian.net/browse/SALTO-5950

---
_Release Notes_:
Jira Adapter:
* Fixed warnings on `requestType` field in `Automation`

---
_User Notifications_: 
_None_


IdoZakk (approved):
🍏

PR #6029, opened 2024-06-06
SALTO-5692: Okta: Upgrade `Group` deploy definitions to new infra

Nurdok:
This PR:
- Modifies the current "default deploy" Okta filter to ignore types that have new-infra definitions
- Adds a new "default deploy" filter to catch these upgraded types (runs after the old "default deploy")
- Upgrades "Group" to the new deploy definitions

---

This is the first PR for upgrading Okta deploy definitions to the new infra.

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/okta-adapter/src/adapter.ts
```diff
@@ -138,9 +143,17 @@ const DEFAULT_FILTERS = [
   profileMappingRemovalFilter,
   // should run after fieldReferences
   ...Object.values(commonFilters),
+
```
shir-reifenberg, line 146: please remove

packages/okta-adapter/src/adapter.ts
```diff
@@ -138,9 +143,17 @@ const DEFAULT_FILTERS = [
...
   // should run last
   privateApiDeployFilter,
   defaultDeployFilter,
+
+  // This catches types we moved to the new infra definitions
+  filterUtils.defaultDeployFilterCreator<FilterResult, OktaOptions>({
```
shir-reifenberg, line 152: nit: maybe extract to a separate file ?

packages/okta-adapter/src/adapter.ts
```diff
@@ -357,38 +371,77 @@ export default class OktaAdapter implements AdapterOperations {
...
+    try {
+      await runner.preDeploy(resolvedChanges)
+    } catch (e) {
+      if (!isSaltoError(e)) {
+        throw e
+      }
+      return {
+        appliedChanges: [],
+        errors: [e],
+      }
```
shir-reifenberg, line 422: mmm I'm not sure about this
I think we currently have filters in preDeploy that produces errors and then return them during `deploy`
for example see `oktaExpressionLanguageFilter`
It's probably not the best either but this is kind of confusing to me

packages/okta-adapter/src/adapter.ts
```diff
@@ -357,38 +371,77 @@ export default class OktaAdapter implements AdapterOperations {
...
 
-    const restoredAppliedChanges = await awu(appliedChangesBeforeRestore)
+    const appliedChanges = await awu(appliedChangesBeforeRestore)
       .map(change => restoreChangeElement(change, sourceChanges, getLookUpName))
       .toArray()
-
+    const restoredAppliedChanges = deploymentUtils.restoreInstanceTypeFromChange({
+      appliedChanges,
+      originalInstanceChanges: instanceChanges,
+    })
```
shir-reifenberg, line 441: do we need this ? cause we deleted the "fix types" code, no?

packages/okta-adapter/src/filters/default_deploy.ts
```diff
@@ -13,21 +13,26 @@
...
 /**
  * Deploys all the changes that were not deployed by the previous filters
  */
 const filterCreator: FilterCreator = ({ definitions, oldApiDefinitions }) => ({
-  name: 'defaultDeployFilter',
+  name: 'oldDefaultDeployFilter',
   deploy: async (changes: Change<InstanceElement>[]) => {
+    const [newInfraChanges, oldInfraChanges] = _.partition(
+      changes,
+      change => definitions.deploy?.instances.customizations?.[getChangeData(change).elemID.typeName] !== undefined,
```
shir-reifenberg, line 29: nit: maybe use `queryWithDefault` and then `allKeys()` ?

shir-reifenberg (commented):
Looks great!! 😎 
Added mainly questions inside

shir-reifenberg:
Also forgot to say earlier, but maybe add unit tests to adapter.test ? we can probably use a similar pattern to `adapter.test` in serviceplaceholder

packages/okta-adapter/src/change_validators/index.ts
```diff
@@ -49,26 +49,43 @@ import {
...
   client: OktaClient
   userConfig: OktaUserConfig
+  definitions: definitionUtils.ApiDefinitions<OktaOptions>
   oldApiDefsConfig: OldOktaDefinitionsConfig
 }): ChangeValidator => {
+  const typesDeployedWithOldInfra = [
+    ...Object.keys(oldApiDefsConfig[API_DEFINITIONS_CONFIG].types),
+    ...Object.keys(oldApiDefsConfig[PRIVATE_API_DEFINITIONS_CONFIG].types),
+  ]
+  const typesDeployedWithNewInfra = Object.keys(definitions.deploy?.instances.customizations ?? {})
```
shir-reifenberg, line 76: nit: cafn we use `defQuery` with `.keys()` ?

packages/okta-adapter/src/change_validators/index.ts
```diff
@@ -49,26 +49,43 @@ import {
...
   const validators: Record<ChangeValidatorName, ChangeValidator> = {
     ...getDefaultChangeValidators(),
     createCheckDeploymentBasedOnConfig: createCheckDeploymentBasedOnConfigValidator({
       typesConfig: _.merge(
         oldApiDefsConfig[API_DEFINITIONS_CONFIG].types,
         oldApiDefsConfig[PRIVATE_API_DEFINITIONS_CONFIG].types,
       ),
+      typesWithNoDeploy: typesDeployedWithNewInfra,
```
shir-reifenberg, line 84: just making sure I understand - this will skip creating error for any type in `typesWithNoDeploy` ?

packages/okta-adapter/src/definitions/deploy.ts
```diff
@@ -0,0 +1,90 @@
...
+          },
+          copyFromResponse: {
+            updateServiceIDs: true,
+          },
+        },
+      },
+      referenceResolution: {
+        when: 'early',
+      },
+      changeGroupId: deployment.grouping.selfGroup,
```
shir-reifenberg, line 85: don't we need to use the adapter's changeGroups ? cause not all changes are in `selfGroup`

packages/okta-adapter/src/adapter.ts
```diff
@@ -357,38 +363,64 @@ export default class OktaAdapter implements AdapterOperations {
...
+    const { deployResult } = await runner.deploy(resolvedChanges, changeGroup)
+    const appliedChangesBeforeRestore = [...deployResult.appliedChanges]
+    try {
+      await runner.onDeploy(appliedChangesBeforeRestore)
+    } catch (e) {
+      if (!isSaltoError(e)) {
+        throw e
+      }
+      saltoErrors.push(e)
+    }
```
shir-reifenberg, line 416: should we skip this too ?

shir-reifenberg (approved):
Nice ! added mainly questions but other than that LGTM :)

packages/okta-adapter/src/definitions/deploy.ts
shir-reifenberg: looking on this again - I think it has no meaning cause we're not using adapter-wrapper ?

packages/okta-adapter/src/change_validators/index.ts
Nurdok: Yes. I agree the parameter name is confusing...

packages/okta-adapter/src/change_validators/index.ts
Nurdok: Done.

packages/okta-adapter/src/definitions/deploy.ts
Nurdok: Removed

packages/okta-adapter/src/adapter.ts
Nurdok: Done

packages/okta-adapter/src/adapter.ts
```diff
@@ -357,38 +362,57 @@ export default class OktaAdapter implements AdapterOperations {
...
+    const resolvedChanges = await awu(instanceChanges)
       .map(async change =>
-        SKIP_RESOLVE_TYPE_NAMES.includes(getChangeData(change).elemID.typeName)
-          ? change
-          : resolveChangeElement(change, getLookUpName),
+        deployDefQuery.query(getChangeData(change).elemID.typeName)?.referenceResolution?.when === 'early' &&
+        !SKIP_RESOLVE_TYPE_NAMES.includes(getChangeData(change).elemID.typeName)
+          ? changeResolver(change)
+          : change,
       )
```
Nurdok, line 401: Reviewing my own code :)

I copied this from the new infra deploy function, but now that I look at it, I think that this will prevent resolving types for old-infra types (the condition should probably check for `early` or `undefined`). BUT I didn't actually see anything fail in the e2e tests which have some reference expressions. I'd like to find the scenario where this fails before I revert or change this, so I understand what this affects.

@shir-reifenberg Any suggestion on what deployment scenario would trigger this potential bug? E.g., some resolutions that needs to happen in old-infra types (anything other than `Group` at this point).

shir-reifenberg: Good catch! I missed it myself. 
wouldn't this will do the trick ?
```
       deployDefQuery.query(getChangeData(change).elemID.typeName)?.referenceResolution?.when === 'early' || 
        !SKIP_RESOLVE_TYPE_NAMES.includes(getChangeData(change).elemID.typeName)
          ? changeResolver(change)
          : change,
      )
```

I guess it didn't trigger the E2E cause we don't have much references and the ones we have (I think policies on apps are not changed in the e2e flow and hence we don't send the request).
basically anything with a reference should fail, right ?
you can try modifying a `GroupRule` or change the `Application`'s accessPolicy field

Nurdok: It took a while to debug, but this is actually fine 😅 
Using `deployDefQuery.query` will return the default value for types that we didn't move to the new infra types, which is `early`, so they'll get resolved like before. No code change required here.

Thanks!

packages/okta-adapter/src/definitions/deploy.ts
shir-reifenberg: Still relevant ?

packages/okta-adapter/src/adapter.ts
```diff
@@ -377,43 +382,62 @@ export default class OktaAdapter implements AdapterOperations {
...
       .toArray()
-    const runner = this.createFiltersRunner()
+    const saltoErrors: SaltoError[] = []
```
shir-reifenberg, line 428: leftover?

shir-reifenberg (approved):
LGTM 💯 with 2 small comments
also, can you please add a unit test to `default_deploy.ts` to verify it's using a custom lookup func when provided?

packages/okta-adapter/src/adapter.ts
Nurdok: Removed

PR #6035, opened 2024-06-06
SALTO-6048 - Salesforce: Remove Salesforce CPQ Billing unique ID field

tomsellek:
This field can't be deployed, so no reason to have it around.
While we're at it, change a bunch of async stuff to sync.
---
The path of least resistance seems to be to delete the field. Is there a reason to hide it instead?

Tests:
 - fetch an env with the CPQ Billing package installed with/without this PR. Observe the field `blng__UniqueId__c` removed from `blng__RevenueRecognitionTreatment__c`, and that the fields `blng__Family__c`, `blng__NextOpenPeriod__c` and `blng__UniqueId__c` are removed from `blng__FinancePeriod__c`.

---
_Release Notes_: 

Salesforce: CPQ Billing - automatically-generated fields in the 'RevenueRecognitionTreatment' and 'FinancePeriod' will no longer be fetched.

---
_User Notifications_: 
Salesforce: CPQ Billing - The `blng__UniqueId__c` field is removed from the `blng__RevenueRecognitionTreatment__c` type and all its instances. The `blng__Family__c`, `blng__NextOpenPeriod__c` and `blng__UniqueId__c` fields are removed from the `blng__FinancePeriod__c` type abd all its instances.

### DIFF
--- a/packages/salesforce-adapter/src/filters/remove_fields_and_values.ts
+++ b/packages/salesforce-adapter/src/filters/remove_fields_and_values.ts
@@ -26,7 +26,10 @@ import { apiNameSync } from './utils'
 const TYPE_NAME_TO_FIELD_REMOVALS: Map<string, string[]> = new Map([
   ['Profile', ['tabVisibilities']],
   ['blng__RevenueRecognitionTreatment__c', ['blng__UniqueId__c']],
-  ['blng__FinancePeriod__c', ['blng__Family__c', 'blng__NextOpenPeriod__c']],
+  [
+    'blng__FinancePeriod__c',
+    ['blng__Family__c', 'blng__NextOpenPeriod__c', 'blng__UniqueId__c'],
+  ],
 ])
 
 const fieldRemovalsForType = (
--- a/packages/salesforce-adapter/src/filters/remove_fields_and_values.ts
+++ b/packages/salesforce-adapter/src/filters/remove_fields_and_values.ts
@@ -25,7 +25,10 @@ import { apiNameSync, isCustomObjectSync, metadataTypeSync } from './utils'
 
 const TYPE_NAME_TO_FIELD_REMOVALS: Map<string, string[]> = new Map([
   ['Profile', ['tabVisibilities']],
-  ['blng__RevenueRecognitionTreatment__c', ['blng__UniqueId__c']],
+  [
+    'blng__RevenueRecognitionTreatment__c',
+    ['blng__UniqueId__c', 'blng__Family__c', 'blng__NextOpenPeriod__c'],
+  ],
 ])
 
 const fieldRemovalsForType = (
--- a/packages/salesforce-adapter/src/filters/remove_fields_and_values.ts
+++ b/packages/salesforce-adapter/src/filters/remove_fields_and_values.ts
@@ -25,10 +25,8 @@ import { apiNameSync } from './utils'
 
 const TYPE_NAME_TO_FIELD_REMOVALS: Map<string, string[]> = new Map([
   ['Profile', ['tabVisibilities']],
-  [
-    'blng__RevenueRecognitionTreatment__c',
-    ['blng__UniqueId__c', 'blng__Family__c', 'blng__NextOpenPeriod__c'],
-  ],
+  ['blng__RevenueRecognitionTreatment__c', ['blng__UniqueId__c']],
+  ['blng__FinancePeriod__c', ['blng__Family__c', 'blng__NextOpenPeriod__c']],
 ])
 
 const fieldRemovalsForType = (
--- a/packages/salesforce-adapter/src/filters/remove_fields_and_values.ts
+++ b/packages/salesforce-adapter/src/filters/remove_fields_and_values.ts
@@ -13,66 +13,78 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { isObjectType, Element, isInstanceElement } from '@salto-io/adapter-api'
-import { transformValues, TransformFunc } from '@salto-io/adapter-utils'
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/src/filters/remove_fields_and_values.ts
```diff
@@ -13,65 +13,78 @@
...
-  await awu(elements)
+  elements
     .filter(isInstanceElement)
     // The below filter is temporary optimization to save calling transformValues for all instances
     // since TYPE_NAME_TO_FIELD_REMOVALS contains currently only top level types
-    .filter(async (inst) =>
-      typeNameToFieldRemovals.has(await metadataType(inst)),
+    .filter(
+      (inst) =>
+        fieldRemovalsForType(inst.getTypeSync(), typeNameToFieldRemovals)
```
yelly, line 75: Nit: why this more verbose implementation instead of the current `typeNameToFieldRemovals.has`?

yelly (approved):
Looks good to me. Just a few small comments which you can take or leave.

packages/salesforce-adapter/src/filters/remove_fields_and_values.ts
```diff
@@ -13,65 +13,78 @@
...
+import { TransformFunc, transformValuesSync } from '@salto-io/adapter-utils'
 import { LocalFilterCreator } from '../filter'
-import { metadataType } from '../transformers/transformer'
-
-const { awu } = collections.asynciterable
+import { apiNameSync, isCustomObjectSync, metadataTypeSync } from './utils'
 
 const TYPE_NAME_TO_FIELD_REMOVALS: Map<string, string[]> = new Map([
   ['Profile', ['tabVisibilities']],
+  ['blng__RevenueRecognitionTreatment__c', ['blng__UniqueId__c']],
```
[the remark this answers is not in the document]
tomsellek: My B.

packages/salesforce-adapter/src/filters/remove_fields_and_values.ts
```diff
@@ -13,65 +13,81 @@
...
 import { LocalFilterCreator } from '../filter'
-import { metadataType } from '../transformers/transformer'
-
-const { awu } = collections.asynciterable
+import { apiNameSync, isCustomObjectSync, metadataTypeSync } from './utils'
 
 const TYPE_NAME_TO_FIELD_REMOVALS: Map<string, string[]> = new Map([
   ['Profile', ['tabVisibilities']],
+  [
+    'blng__RevenueRecognitionTreatment__c',
```
tamtamirr, line 29: Are we sure this is the only type with these fields in `blng`?
If not, we should implement a more generic solution that removes these fields from all the types.

packages/salesforce-adapter/src/filters/remove_fields_and_values.ts
```diff
@@ -13,65 +13,81 @@
...
 ])
 
-const removeFieldsFromTypes = async (
+const fieldRemovalsForType = (
+  type: ObjectType,
+  typeNameToFieldRemovals: Map<string, string[]>,
+): string[] => {
+  const typeName = isCustomObjectSync(type)
+    ? apiNameSync(type) ?? ''
+    : metadataTypeSync(type)
```
tamtamirr, line 40:
```suggestion
  const typeName = apiNameSync(type) ?? ''
```
apiName of a type (metadataType / CustomObject) should be correct.

packages/salesforce-adapter/src/filters/remove_fields_and_values.ts
```diff
@@ -13,65 +13,81 @@
...
-  await awu(elements)
+  elements
     .filter(isInstanceElement)
     // The below filter is temporary optimization to save calling transformValues for all instances
     // since TYPE_NAME_TO_FIELD_REMOVALS contains currently only top level types
-    .filter(async (inst) =>
-      typeNameToFieldRemovals.has(await metadataType(inst)),
+    .filter(
+      (inst) =>
+        fieldRemovalsForType(inst.getTypeSync(), typeNameToFieldRemovals)
```
tamtamirr, line 78:
```suggestion
    .filter(inst =>
      typeNameToFieldRemovals.has(apiNameSync(inst.getTypeSync()) ?? ''),
```

tamtamirr (commented):
Added a few comments 🙏

packages/salesforce-adapter/src/filters/remove_fields_and_values.ts
```diff
@@ -13,65 +13,81 @@
...
-import { metadataType } from '../transformers/transformer'
-
-const { awu } = collections.asynciterable
+import { apiNameSync, isCustomObjectSync, metadataTypeSync } from './utils'
 
 const TYPE_NAME_TO_FIELD_REMOVALS: Map<string, string[]> = new Map([
   ['Profile', ['tabVisibilities']],
+  [
+    'blng__RevenueRecognitionTreatment__c',
+    ['blng__UniqueId__c', 'blng__Family__c', 'blng__NextOpenPeriod__c'],
```
amitgoldarb, line 30: There is also some issue with:
`blng__StartDateTime__c` right @tamtamirr ? We should probably look into that field as well.

tamtamirr: @amitgoldarb Yes, needs investigation though 🙏

packages/salesforce-adapter/src/filters/remove_fields_and_values.ts
tomsellek: There will certainly be follow-up PRs as we discover more issues.

PR #6037, opened 2024-06-09
SALTO-3621: Adding meta type to Object type

yelly:
This PR should add everything needed for meta types including validations, parsing and dumping.

---

Hope I didn't miss anything...

---
_Release Notes_: 
None.

---
_User Notifications_: 
None.

### DIFF
--- a/packages/adapter-api/src/elements.ts
+++ b/packages/adapter-api/src/elements.ts
@@ -359,11 +359,30 @@ export type FieldDefinition = {
   annotations?: Values
 }
 
+const validateMetaType = (metaType?: ObjectType): ObjectType | undefined => {
+  if (metaType === undefined) {
+    return undefined
+  }
+
+  if (!isObjectType(metaType)) {
+    log.error(`Got an invalid meta type which is not an object type with ID ${(metaType as Element).elemID}.`)
+    return undefined
+  }
+
+  if (isPlaceholderObjectType(metaType)) {
+    log.warn(`Meta type with ID ${metaType.elemID.getFullName()} not found in elements source.`)
+    return undefined
+  }
+
+  return metaType
+}
+
 /**
  * Defines a type that represents an object (Also NOT auto generated)
  */
 export class ObjectType extends Element {
   fields: FieldMap
+  metaType: TypeReference<ObjectType> | undefined
   isSettings: boolean
 
   constructor({
@@ -371,6 +390,7 @@ export class ObjectType extends Element {
     fields = {},
     annotationRefsOrTypes = {},
     annotations = {},
+    metaType = undefined,
     isSettings = false,
     path = undefined,
   }: {
@@ -378,6 +398,7 @@ export class ObjectType extends Element {
     fields?: Record<string, FieldDefinition>
     annotationRefsOrTypes?: TypeRefMap
     annotations?: Values
+    metaType?: TypeOrRef<ObjectType>
     isSettings?: boolean
     path?: ReadonlyArray<string>
   }) {
@@ -386,6 +407,9 @@ export class ObjectType extends Element {
       fields,
       (fieldDef, name) => new Field(this, name, getRefType(fieldDef.refType), fieldDef.annotations),
     )
+    if (metaType !== undefined) {
+      this.metaType = getRefType(metaType)
+    }
     this.isSettings = isSettings
   }
 
@@ -409,6 +433,14 @@ export class ObjectType extends Element {
     )
   }
 
+  async getMetaType(elementsSource?: ReadOnlyElementsSource): Promise<ObjectType | undefined> {
+    return validateMetaType(await this.metaType?.getResolvedValue(elementsSource))
+  }
+
+  getMetaTypeSync(): ObjectType | undefined {
+    return validateMetaType(this.metaType?.getResolvedValueSync())
+  }
+
   /**
    * Return an independent copy of this instance.
    * @return {ObjectType} the cloned instance
@@ -421,6 +453,7 @@ export class ObjectType extends Element {
       fields: this.cloneFields(),
       annotationRefsOrTypes: this.cloneAnnotationTypes(),
       annotations: this.cloneAnnotations(),
+      metaType: this.metaType?.clone(),
       isSettings,
       path: this.path !== undefined ? [...this.path] : undefined,
     })
@@ -435,7 +468,10 @@ export class ObjectType extends Element {
   }
 }
 
-export class PlaceholderObjectType extends ObjectType {}
+export class PlaceholderObjectType extends ObjectType {
+  // This is required to allow typescript to reason about ObjectTypes which are not PlaceholderObjectTypes.
+  isPlaceholder = true
+}
 
 const objectTypeOrPlaceholder = (type: TypeElement | undefined, elemID: ElemID): ObjectType => {
   // This can happen when the user has an instance like
@@ -457,6 +493,7 @@ const validateType = (type: TypeElement | undefined, elemID: ElemID): TypeElemen
   }
   return type
 }
+
 export class InstanceElement extends Element {
   public refType: TypeReference<ObjectType>
   constructor(
--- a/packages/adapter-api/src/elements.ts
+++ b/packages/adapter-api/src/elements.ts
@@ -369,7 +369,7 @@ const validateMetaType = (metaType?: ObjectType): ObjectType | undefined => {
     return undefined
   }
 
-  if (isPlaceholderObjectType(metaType)) {
+  if (metaType instanceof PlaceholderObjectType) {
     log.warn(`Meta type with ID ${metaType.elemID.getFullName()} not found in elements source.`)
   }
 
@@ -480,10 +480,7 @@ export class ObjectType extends Element {
   }
 }
 
-export class PlaceholderObjectType extends ObjectType {
-  // This is required to allow typescript to reason about ObjectTypes which are not PlaceholderObjectTypes.
-  isPlaceholder = true
-}
+export class PlaceholderObjectType extends ObjectType {}
 
 const objectTypeOrPlaceholder = (type: TypeElement | undefined, elemID: ElemID): ObjectType => {
   // This can happen when the user has an instance like
--- a/packages/adapter-api/src/elements.ts
+++ b/packages/adapter-api/src/elements.ts
@@ -428,9 +428,22 @@ export class ObjectType extends Element {
         _.mapValues(this.fields, f => f.elemID.getFullName()),
         _.mapValues(other.fields, f => f.elemID.getFullName()),
       ) &&
+      this.isMetaTypeEqual(other) &&
       _.isEqual(this.isSettings, other.isSettings) &&
       _.every(Object.keys(this.fields).map(n => this.fields[n].isEqual(other.fields[n], options)))
     )
+  }
+
+  isMetaTypeEqual(other: ObjectType): boolean {
+    if (this.metaType === undefined && other.metaType === undefined) {
+      return true
+    }
+
+    if (this.metaType === undefined || other.metaType === undefined) {
+      return false
+    }
+
+    return this.metaType.elemID.isEqual(other.metaType.elemID)
   }
 
   async getMetaType(elementsSource?: ReadOnlyElementsSource): Promise<ObjectType | undefined> {
--- a/packages/adapter-api/src/elements.ts
+++ b/packages/adapter-api/src/elements.ts
@@ -365,13 +365,12 @@ const validateMetaType = (metaType?: ObjectType): ObjectType | undefined => {
   }
 
   if (!isObjectType(metaType)) {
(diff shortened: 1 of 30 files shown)


packages/adapter-api/src/elements.ts
```diff
@@ -359,25 +359,46 @@ export type FieldDefinition = {
   annotations?: Values
 }
 
+const validateMetaType = (metaType?: ObjectType): ObjectType | undefined => {
+  if (metaType === undefined) {
+    return undefined
+  }
+
+  if (!isObjectType(metaType)) {
+    log.error(`Got an invalid meta type which is not an object type with ID ${(metaType as Element).elemID}.`)
```
ori-moisis, line 368: IIUC this log will use the `inspect` customizer for the element ID, so it will be
```
... with ID ElemID(...)
```
feels redundant to have ID twice...

packages/adapter-api/src/elements.ts
```diff
@@ -359,25 +359,46 @@ export type FieldDefinition = {
...
+    return undefined
+  }
+
+  if (isPlaceholderObjectType(metaType)) {
+    log.warn(`Meta type with ID ${metaType.elemID.getFullName()} not found in elements source.`)
+    return undefined
```
ori-moisis, line 374: why return undefined in this case?
would it not be better to return the placeholder so it will at least have the ID?

packages/adapter-api/test/elements.test.ts
```diff
@@ -35,6 +35,7 @@ import {
   isContainerType,
   createRefToElmWithValue,
   PlaceholderObjectType,
+  placeholderReadonlyElementsSource,
```
ori-moisis, line 38: should we really be relying on this thing?
this will make it harder to remove as the comment above it says we should...

packages/adapter-api/src/elements.ts
```diff
@@ -435,7 +481,10 @@ export class ObjectType extends Element {
   }
 }
 
-export class PlaceholderObjectType extends ObjectType {}
+export class PlaceholderObjectType extends ObjectType {
+  // This is required to allow typescript to reason about ObjectTypes which are not PlaceholderObjectTypes.
+  isPlaceholder = true
```
ori-moisis, line 486: is this really needed?
the build doesn't seem to break if this is removed, so, wondering where this is actually needed...

packages/adapter-api/test/elements.test.ts
```diff
@@ -110,7 +181,7 @@ describe('Test elements.ts', () => {
       annotations: {},
     })
 
-    const strField = new Field(objT, 'str_field', primStr)
+    const strField = new Field(objT, 'strField', primStr)
```
ori-moisis, line 184: is it really needed to change the names of the fields?
if it is really that important, do we need this to be consistent (e.g - the next two fields...)

packages/adapter-api/test/elements.test.ts
```diff
@@ -132,14 +203,26 @@ describe('Test elements.ts', () => {
...
+    it('should identify object types with different meta types', () => {
+      const otDiff1 = ot.clone()
+      const otDiff2 = ot.clone()
+
+      otDiff1.metaType = new TypeReference(new ElemID('test', 'meta1'))
+      otDiff2.metaType = undefined
+
+      expect(isEqualElements(ot, otDiff1)).toBeFalsy()
+      expect(isEqualElements(otDiff1, otDiff2)).toBeFalsy()
+      expect(isEqualElements(otDiff2, otDiff2)).toBeTruthy()
```
ori-moisis, line 215: supposed to compare to `ot`?
```suggestion
      expect(isEqualElements(ot, otDiff2)).toBeTruthy()
```

packages/parser/src/parser/internal/native/consumers/top_level.ts
```diff
@@ -87,11 +88,32 @@ const consumeType = (
...
+            elemID,
+            fields: consumedBlock.value.fields,
+            annotationRefsOrTypes: consumedBlock.value.annotationRefTypes,
+            annotations: consumedBlock.value.attrs,
+            metaType: new TypeReference(metaElemID),
+          }),
+        ),
+        range: consumedBlock.range,
+      }
+    }
```
ori-moisis, line 116: this is kind of hard to follow the paths here, unclear which cases are going to `return` and which cases continue to the rest of the function

also, it is a bit confusing that there are now two different places that create object types in this function and the only difference is the meta type, I think we should try to simplify this a bit

ori-moisis (commented):
Reviewed half, posting comments so far

packages/adapter-api/src/elements.ts
yelly: Done.

packages/adapter-api/src/elements.ts
yelly: Done.

packages/adapter-api/src/elements.ts
yelly: I've added a test which won't transpile if this is removed. A quirk of TS.
Right now the tests pass even if they have TS errors, I'm working on it, see https://salto-io.atlassian.net/browse/SALTO-6129.

packages/adapter-api/test/elements.test.ts
yelly: Done.

packages/adapter-api/test/elements.test.ts
yelly: Not that important, just having all of these eslint exceptions seems unnecessary.
Happy to revert and happy to align, whatever you think is better.

packages/adapter-api/test/elements.test.ts
yelly: The alternative is writing another degenerate element source here for the tests - do you think that is better?

packages/parser/src/parser/internal/native/consumers/top_level.ts
yelly: Good point, simplified.

packages/adapter-api/test/elements.test.ts
ori-moisis: generally, I will always vote for not mixing in unrelated changes into a feature PR.
specifically with tests and things that might seem on the surface as strange and unimportant, I would never touch them without really understanding what the scenario was or unless there is a different reason to change (e.g - the code changed / the expectations changed)

packages/adapter-api/src/elements.ts
yelly: OK, I've played with this a bit more. Turns out this is specifically a shortcoming of the way type guards work (which is super weird). You can see an example of this [here](https://www.typescriptlang.org/play/?#code/MYGwhgzhAECC0G8BQ1XQNYFMCe0C80A5AG5ggCumhSAvkkqJDAELSYAeALpgHYAmMeAjpIAZuR7BOASwD2PaNIjMAFOwBc0Ceh6yA7jwCUm9opaIUaAE6ZO5KwtPSeETmEmZZo6M1r1xkjLy0GAAcrKcqhpwxtCuVs4A5hZoit4qSlGGhimp1rb2CoS61KkiqTZ2DtDsAHRY2LRAA). You will notice that if you add anything to the definition of `B` it fixes the error.
I've moved to explicitly using `instanceof` where I need it in `validateMetaType` which solves the typing problem. Gotta love typescript.

As for building the tests, we are explicitly telling Jest to ignore transpilation errors which IMO is bad. Also we had some TS errors in the tests which the old build was not catching for some reason, so I still think this was a good move.

packages/adapter-api/test/elements.test.ts
yelly: OK cool, going back to the old style.

packages/adapter-api/src/elements.ts
ori-moisis: cool

packages/adapter-api/test/elements.test.ts
ori-moisis: probably not

packages/adapter-api/test/elements.test.ts
yelly: `ot` absolutely has a meta type.

packages/workspace/src/serializer/elements.ts
```diff
@@ -336,6 +339,7 @@ const generalDeserializeParsed = async <T>(parsed: unknown, staticFileReviver?:
           fields: reviveFieldDefinitions(v.fields),
           annotationRefsOrTypes: reviveAnnotationRefTypes(v),
           annotations: restoreClasses(v.annotations),
+          metaType: v.metaType ? (restoreClasses(v.metaType) as TypeReference<ObjectType>) : undefined,
```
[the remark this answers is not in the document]
yelly: Done.

packages/workspace/src/serializer/elements.ts
```diff
@@ -215,7 +215,10 @@ export const serializeStream = async <T = Element>(
   const resolveCircles = (v: any): any =>
     isPrimitiveType(v)
       ? new PrimitiveType({ elemID: v.elemID, primitive: v.primitive })
-      : new ObjectType({ elemID: v.elemID })
+      : new ObjectType({
+          elemID: v.elemID,
+          metaType: v.metaType !== undefined ? new TypeReference<ObjectType>(v.metaType.elemID) : undefined,
```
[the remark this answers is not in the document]
yelly: Without this everything breaks.
My understanding is that if you hit a circular dependency when restoring you need to create a node in the graph that you can then merge the content of the element into later. Because we use both the ID and the meta type to decide that two object types are the same element then we need to include it in the shell, otherwise we get a meta type conflict.
We can discuss further if this explanation isn't enough.

packages/workspace/src/validator.ts
```diff
@@ -695,14 +743,41 @@ const validateField = (field: Field): ValidationError[] => {
...
 const validateType = (element: TypeElement): ValidationError[] => {
   const annotationTypes = syncGetElementAnnotationTypes(element)
   const errors = Object.keys(element.annotations)
     .filter(k => annotationTypes[k] !== undefined)
     .flatMap(k => validateValue(element.elemID.createNestedID('attr', k), element.annotations[k], annotationTypes[k]))
   if (isObjectType(element)) {
+    let metaTypeErrors: ValidationError[] = []
+    if (element.metaType !== undefined) {
+      metaTypeErrors = validateMetaType(element.elemID, element.metaType.type)
+    }
```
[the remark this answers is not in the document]
yelly: Done.

packages/parser/src/parser/internal/native/consumers/top_level.ts
```diff
@@ -85,10 +97,6 @@ const consumeType = (
   }
 
   let primitive = primitiveType(baseType)
-
-  // If the base type token can't be resolved to a specific primitive type, we will
-  // just treat the type as unknown and add an error.
-  // No need to recover since structure is unharmed.
```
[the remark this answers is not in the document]
yelly: Done.

packages/adapter-api/test/elements.test.ts
```diff
@@ -94,10 +108,67 @@ describe('Test elements.ts', () => {
...
+          elemID: otID,
+          metaType: primNum as unknown as ObjectType,
+        })
+        expect(objectTypeInvalidMeta.getMetaTypeSync()).toBeUndefined()
+      })
+
+      it('should return placeholder when meta type is placeholder', () => {
+        const objectTypeInvalidMeta = new ObjectType({
+          elemID: otID,
+          metaType: new PlaceholderObjectType({ elemID: metaTypeID }),
```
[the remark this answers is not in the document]
yelly: Unless I'm missing some super cool TS trick, the answer seems to be no.
Since PlaceholderObjectType extends ObjectType this makes sense, because you can't no if an ObjectType is actually a PlaceholderObjectType.

packages/workspace/src/serializer/elements.ts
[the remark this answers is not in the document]
ori-moisis: 1. this code is in serialize, not in restore
2. the element ID is the only thing that should decide if two object types are the same element
3. if we get a conflict on the meta type, the exact same thing should have happened on `isSettings` but that doesn't seem to be here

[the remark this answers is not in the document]
yelly: Removed.

ori-moisis (approved):

packages/parser/src/parser/internal/native/consumers/top_level.ts
```diff
@@ -69,26 +70,36 @@ const consumeType = (
...
       range: consumedBlock.range,
     }
   }
 
-  let primitive = primitiveType(baseType)
-
   // If the base type token can't be resolved to a specific primitive type, we will
   // just treat the type as unknown and add an error.
   // No need to recover since structure is unharmed.
+  let primitive = primitiveType(baseType)
```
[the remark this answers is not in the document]
yelly: Done.

PR #6041, opened 2024-06-09
SALTO-6078: Filter out null-values from instances upon creation and filter out empty instances

DanielBarros1:
see SALTO-6078 for details - basically, in the new infra in ZD, we create instance whether they have fields or not. So in the example of the salto3 environment, we have a `macro_categories` singleton instance that has a single field called 'categories'. The list in the field is empty, and so we end up with an instance who's value is `undefined`, which breaks all sorts of things in our filters.

AFAIU, it's not possible to address this in a transformer on the type because we would like to just not have this instance in the first place (which is what happens in the old infra).

So to deal with that, in this PR i basically retraced what the old infra does (see code pointers in the jira ticket)

to test this change, i pulled before/after on 2 zendesk subdomains and made sure that the results are identical (w.r to the old infra)

### DIFF
--- a/packages/adapter-components/src/elements_deprecated/ducktype/transformer.ts
+++ b/packages/adapter-components/src/elements_deprecated/ducktype/transformer.ts
@@ -22,7 +22,6 @@ import {
   ObjectType,
   ElemIdGetter,
   SaltoError,
-  isInstanceElement,
 } from '@salto-io/adapter-api'
 import { naclCase } from '@salto-io/adapter-utils'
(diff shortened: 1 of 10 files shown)


netama (approved):
LGTM, thanks. is there some UT that can demonstrate the behavior that was being fixed? (something that would return a bad result before the fix)

also nit: I think this shouldn't be in the release notes, at least not under core - might be easier to just remove the notes :) as there is probably no impact on users

PR #6043, opened 2024-06-09
SALTO-6080 pushing entities one by one upon fetch to prevent RangeError

omrifarkash:
We originally used `...` to push all validEntries, when there are a lot of entries, this will cause `Maximum call stack size exceeded` Error

---

- I also added a log of the original error stack. It was very hard to debug since we lose the original error stack while walking on a DAG.

---
_Release Notes_: 
_Confluence Adapter_:
- fix fetch for large amount of pages

---
_User Notifications_: 
_Confluence Adapter_:
- fix fetch for large amount of pages

### DIFF
--- a/packages/dag/src/nodemap.ts
+++ b/packages/dag/src/nodemap.ts
@@ -90,7 +90,7 @@ class WalkErrors<T> extends Map<NodeId, Error> {
   }
 
   set(nodeId: NodeId, value: Error, visited: Set<string> = new Set()): this {
-    log.error('Error encountered while walking on node %s: original stack: %s', nodeId, value.stack)
+    log.error('Error encountered while walking on node %s: %s\n stack: %s', nodeId, value, value.stack)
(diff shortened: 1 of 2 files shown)


netama (approved):
thanks!

packages/dag/src/nodemap.ts
```diff
@@ -88,6 +90,7 @@ class WalkErrors<T> extends Map<NodeId, Error> {
   }
 
   set(nodeId: NodeId, value: Error, visited: Set<string> = new Set()): this {
+    log.error('Error encountered while walking on node %s: original stack: %s', nodeId, value.stack)
```
ori-moisis, line 93: please add the error itself as well, this will only log the node ID and the stack, but not the error

omrifarkash: I think the error it self is being thrown later. However I will add it here also 👍

PR #6044, opened 2024-06-09
SALTO-6047 pagerDuty enforce subdomain

Uri-Bar:
SALTO-6047 pagerDuty enforce subdomain
---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/pagerduty-adapter/src/client/connection.ts
+++ b/packages/pagerduty-adapter/src/client/connection.ts
@@ -28,10 +28,10 @@ export const validateCredentials = async ({
   credentials: Credentials
 }): Promise<AccountInfo> => {
   try {
-    const users = await connection.get('/users?include[]=subdomains')
+    const users = await connection.get('/users?limit=1&include[]=subdomains')
     const subdomain = users.data.users[0].subdomains[0]
     if (subdomain !== credentials.subdomain) {
-      throw new Error('Subdomain does not match')
+      throw new Error('The subdomain given does not match with the subdomain that associated with the token')
     }
     return { accountId: credentials.subdomain }
   } catch (e) {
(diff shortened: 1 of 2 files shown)


packages/pagerduty-adapter/src/client/connection.ts
```diff
@@ -28,7 +28,11 @@ export const validateCredentials = async ({
   credentials: Credentials
 }): Promise<AccountInfo> => {
   try {
-    await connection.get('/services')
+    const users = await connection.get('/users?include[]=subdomains')
```
netama, line 31: can we get this with something other than the users endpoint? usually we keep the users under a config flag so that we don't have to get it. maybe `/users/me` or something like that if available?

packages/pagerduty-adapter/src/client/connection.ts
```diff
@@ -28,7 +28,11 @@ export const validateCredentials = async ({
...
+    const subdomain = users.data.users[0].subdomains[0]
+    if (subdomain !== credentials.subdomain) {
+      throw new Error('Subdomain does not match')
```
netama, line 34: the text is user-facing, right? can we make the error clearer?

packages/pagerduty-adapter/src/client/connection.ts
Uri-Bar: its just a call, I call limit it to 1 user. but we cant use users/me because it could be a group token and then it not support /me

packages/pagerduty-adapter/src/client/connection.ts
Uri-Bar: sure

packages/pagerduty-adapter/src/client/connection.ts
Uri-Bar: I changed it to query only one user

PR #6047, opened 2024-06-10
SALTO-6081 remove-airbnb-eslint-plugin

eshedg:
removing the airbnb eslint plugin
---

the airbnb eslint plugin is a curated set of eslint rules, which we've been using since this project started.
we are in the process of bumping to Node 18 (and then to later versions), which is being done in another PR.
the airbnb eslint plugin breaks a lot of code, when mixed with Node 18 types.
this PR was created in order to keep the Node 18 PR smaller, by splitting off the airbnb removal.
once we land on the proper Node version, we may come back and revisit this eslint plugin in a separate PR.

---
_Release Notes_: 

---
_User Notifications_: 


eshedg:
dropped

PR #6048, opened 2024-06-10
SALTO-5860: add a filter which re-adds omitted recurseInto fields 

DanielBarros1:
As per the discussion in SALTO-5860, since this is 'correct' behavior for this infra, we would need to deal with the issue on the ZD level. In this PR i add a filter which re-adds omitted fields

---
_Release Notes_: 
Zendesk: 
Add filter to re-add omitted fields in the new infra fetch

---
_User Notifications_: 
None


packages/zendesk-adapter/src/filters/add_recurse_into_field.ts
```diff
@@ -0,0 +1,79 @@
...
+import { FilterCreator } from '../filter'
+import { FETCH_CONFIG } from '../config'
+
+const log = logger(module)
+
+type RecurseIntoFieldComponent = {
+  recurseIntoFields: { fieldName: string }[]
+}
+
+const recurseIntoFieldMap: Record<string, RecurseIntoFieldComponent> = {
```
Uri-Bar, line 27: 🔥

packages/zendesk-adapter/src/filters/add_recurse_into_field.ts
```diff
@@ -0,0 +1,79 @@
...
+ * As part of the migration to the new infra (SALTO-5761), there are some types for whom the
+ * 'recurseInto' fields get omitted when no values are present (details: SALTO-5860).
+ * While this is 'correct' behavior, it diverges from the behavior of the old infra, which means it creates a diff
+ * while fetching for clients.  * We would like to not have diff, so in this filter we re-add the omitted fields.
+ */
+const filterCreator: FilterCreator = ({ config }) => ({
+  name: 'addRecurseIntoField',
+  onFetch: async (elements: Element[]): Promise<void> => {
+    if (config[FETCH_CONFIG].useNewInfra !== true) {
+      log.info('not running addRecurseIntoField filter as the fetch is not using the new infra')
```
Uri-Bar, line 61: does the filter will run in the old infra as well?
if so it does, it will create a lot of garbage logs.
and if not it necessary

so I think we can remove it

packages/zendesk-adapter/src/filters/add_recurse_into_field.ts
```diff
@@ -0,0 +1,79 @@
...
+      fields.recurseIntoFields.forEach(field => {
+        if (element.value[field.fieldName] === undefined) {
+          element.value[field.fieldName] = []
+        }
+      })
+    })
+  },
+})
+
+export default filterCreator
```
Uri-Bar, line 79: please add UT 📚

Uri-Bar (commented):
looks good

Uri-Bar (approved):

PR #6049, opened 2024-06-11
SALTO-6086: Bugfix: actionDependencies cause errors when they refer to actions that aren't part of the currently deployed change

Nurdok:
This PR adds an option to avoid creating missing nodes when adding edges in a DAG, and uses it where it causes errors in deployment. See SALTO-6086 for a detailed explanation.

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/adapter-components/src/deployment/deploy/graph.ts
```diff
@@ -92,7 +92,7 @@ export const createDependencyGraph = <ClientOptions extends string, AdditionalAc
       })
     })
     defQuery.query(typeName)?.actionDependencies?.forEach(({ first, second }) => {
-      graph.addEdge(toNodeID(typeName, second), toNodeID(typeName, first))
+      graph.addEdge(toNodeID(typeName, second), toNodeID(typeName, first), { createMissingNodes: false })
```
netama, line 95: I think this logic is problematic, because if we have a << b << c and b is missing, we will lose the implied dependency a << c.
wdyt about instead ensuring there are nodes, e.g.
```suggestion
      // add empty nodes if they don't already exist
      [first, second].forEach(n => {
        const id = toNodeID(typeName, n)
        if (!graph.has(id)) {
          graph.addNode(id, [], {
            action: n as ActionName | AdditionalAction,
            typeActionChanges: [],
            typeName,
          })
        }
      })
      graph.addEdge(toNodeID(typeName, second), toNodeID(typeName, first))
```
(and not changing the dag)?

netama: I'm honestly not sure we should fix this, as it likely hides a development error. can throw a clearer error instead...

netama (approved):

PR #6052, opened 2024-06-13
SALTO-6089: calling isPathIncluded on staticFileSource would yield Ma…

dorsev:
…ximum call stack size exceeded

---

---

_Release Notes_: 
None

---

_User Notifications_: 
None


ori-moisis (approved):

dorsev:
Thank you for the quick review @ori-moisis , I really appreciate it 🙏

PR #6055, opened 2024-06-13
SALTO-5918: Performance improvements in field references filter

yelly:
1. Made the instance filter sync.
2. Dropping profiles when not needed.

Also removed the comment to move to using the element source, we decided against that in #5907.

---

This filter is very slow.

---
_Release Notes_: 
None.

---
_User Notifications_: 
None.

### DIFF
--- a/packages/salesforce-adapter/src/filters/field_references.ts
+++ b/packages/salesforce-adapter/src/filters/field_references.ts
@@ -45,11 +45,13 @@ import {
   GROUP_METADATA_TYPE,
   ROLE_METADATA_TYPE,
   FLOW_METADATA_TYPE,
+  PROFILE_METADATA_TYPE,
 } from '../constants'
 import {
   buildElementsSourceForFetch,
   extractFlatCustomObjectFields,
   hasApiName,
+  isInstanceOfTypeSync,
 } from './utils'
 
 const { awu } = collections.asynciterable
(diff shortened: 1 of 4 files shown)


tamtamirr (approved):
LGTM

packages/salesforce-adapter/src/filters/field_references.ts
```diff
@@ -233,6 +237,9 @@ const filter: LocalFilterCreator = ({ config }) => ({
       elements,
       buildElementsSourceForFetch(elements, config),
       refDef,
+      config.fetchProfile.isFeatureEnabled('generateRefsInProfiles')
```
tamtamirr, line 240: Can we also do the same logic for Permission Sets

packages/salesforce-adapter/src/filters/field_references.ts
tamtamirr, line 1: Since you filter out **Profiles** and **PermissionSets** in advance, you can delete the older logic that manipulates the list of the reference mappers  definitions.

tamtamirr (commented):
Added some small comments we've discussed F2F 🙏

packages/salesforce-adapter/src/filters/field_references.ts
yelly: Done.

packages/salesforce-adapter/src/filters/field_references.ts
yelly: Done.

tamtamirr (approved):

PR #6057, opened 2024-06-13
SALTO-6093 - Salesforce: Remove some more UniqueID fields from CPQ Billing types

tomsellek:
These fields can't be deployed, and contain no useful information. We can just discard them.

---

---
_Release Notes_: 
Salesforce: All 'Unique ID' fields of CPQ Billing objects are now discarded.

---
_User Notifications_: 
Salesforce: All 'Unique ID' fields of CPQ Billing objects will be removed from the workspace.

### DIFF
--- a/packages/salesforce-adapter/src/filters/remove_fields_and_values.ts
+++ b/packages/salesforce-adapter/src/filters/remove_fields_and_values.ts
@@ -30,6 +30,14 @@ const TYPE_NAME_TO_FIELD_REMOVALS: Map<string, string[]> = new Map([
     'blng__FinancePeriod__c',
     ['blng__Family__c', 'blng__NextOpenPeriod__c', 'blng__UniqueId__c'],
   ],
+  ['blng__AccountBalanceSnapshot__c', ['blng__UniqueId__c']],
+  ['blng__ErrorLog__c', ['blng__UniqueId__c']],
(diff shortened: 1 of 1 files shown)


packages/salesforce-adapter/src/filters/remove_fields_and_values.ts
```diff
@@ -30,6 +30,14 @@ const TYPE_NAME_TO_FIELD_REMOVALS: Map<string, string[]> = new Map([
     'blng__FinancePeriod__c',
     ['blng__Family__c', 'blng__NextOpenPeriod__c', 'blng__UniqueId__c'],
   ],
+  ['blng__AccountBalanceSnapshot__c', ['blng__UniqueId__c']],
```
yelly, line 33: Just a general note - don't we want to just always remove every `blng__UniqueId__c` field instead of needing include every type as it comes up? Or are we just confident that this is all of them?

yelly (approved):

packages/salesforce-adapter/src/filters/remove_fields_and_values.ts
tomsellek: We probably do, but for now the quickest solution was to piggyback on this filter. If this turns into a whack-a-mole situation we should certainly add this functionality of removing a field from all types that contain it.

PR #6058, opened 2024-06-14
SALTO-6096 - Salesforce: Handle circular dependency between ProductRule and ErrorCondition

tomsellek:
Similar to other types that refer to each other implicitly (rather than via explicit references), we need to create a special deploy group and handle it explicitly.

---

Test:
 - Change the NaCl files of a `ProductRule` instance and the related `ErrorCondition` instance. Observe the correct group ID.

---
_Release Notes_: 
Salesforce: Circular dependency between CPQ ProductRule instances with 'Custom' ConditionsMet value and ErrorCondition instances are handled correctly.

---
_User Notifications_: 
N/A


tamtamirr (approved):
Thanks for this! Is this PR well tested? Feel free to merge if so

PR #6061, opened 2024-06-16
SALTO-6090 Fix SuiteQL query of over 100K items

dantal4:
SuiteQL returns only the first 100K items. In this PR we fix it and retrieve whole items when there are more than 100K items, by making more queries with filtering by the last queried item id - `WHERE ... AND id > {lastQueriedItemId}`

---

---
_Release Notes_: 
Netsuite Adapter:
- Fix SuiteQL query of over 100K items

---
_User Notifications_: 
Netsuite Adapter:
- Missing files/folders will be fetched & added to the FileCabinet folder


omrilit (approved):
LGTM!

PR #6064, opened 2024-06-16
SALTO-6051: Microsoft Entra: Add filter to support array fields deployment

shoshana-houri:
We have two types, administrativeUnit and directoryRole, that have a members field, which is an array of references. Each change to this array requires a different API call—either addition or deletion. To handle this, we've created a custom filter (shared under adapter-components) that finds the array differences and makes the necessary unknown number of calls.

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/adapter-components/src/filters/array_fields.ts
```diff
@@ -0,0 +1,292 @@
...
+  }
+  log.debug('Found %d values to delete and %d values to add', valuesToDelete.length, valuesToAdd.length)
+  log.trace(
+    'values to delete: %s, values to add: %s',
+    safeJsonStringify(valuesToDelete, elementExpressionStringifyReplacer),
+    safeJsonStringify(valuesToAdd, elementExpressionStringifyReplacer),
+  )
+
+  const fieldObjectType = new ObjectType({
+    elemID: new ElemID(singleFieldDefinitions.adapterName, singleFieldDefinitions.fieldTypeName),
```
shoshana-houri, line 145: Please note that we create a new `elemId` using the field type name, which is not a top-level element. This way, we can set the deploy definition directly on the field instead of working around it in the parent's definition. This approach is also aligned with how we achieve this for [Confluence](https://github.com/salto-io/salto/blob/95a13ccb91ce8144a8b5668ea4b3c380e80cea35/packages/confluence-adapter/src/filters/deploy_space_and_permissions.ts#L62).
Please LMK if you have any objection :)

packages/adapter-components/src/filters/array_fields.ts
```diff
@@ -0,0 +1,292 @@
...
+  fieldName,
+  valueMapper,
+}: { change: AdditionOrModificationChange } & CommonDefinition & Omit<SingleFieldDefinition, 'fieldTypeName'>): {
+  valuesToDelete: ArrayElementChangeData[]
+  valuesToAdd: ArrayElementChangeData[]
+} => {
+  const beforeValues = isModificationChange(change) ? _.get(change.data.before.value, fieldName, []) : []
+  const afterValues = _.get(change.data.after.value, fieldName, [])
+  if (!_.isArray(beforeValues) || !_.isArray(afterValues)) {
+    throw new Error(
```
shir-reifenberg, line 92: instead of throwing I think you can return `SaltoElementError` (which should contain the top level change `elemID`. The error message itself is user facing so it should be indicative (something like `Expected ${fieldName} to be an array` 
In the `log.error` include as much details as you want :)

packages/adapter-components/src/filters/array_fields.ts
```diff
@@ -0,0 +1,292 @@
...
+      `Failed to calculate diff for ${parentTypeName}:${fieldName} ${getChangeData(change).elemID.getFullName()}, expected ${fieldName} to be an array, got ${inspectValue(beforeValues)} and ${inspectValue(afterValues)}`,
+    )
+  }
+
+  const [resolvedBeforeValues, resolvedAfterValues] = [beforeValues, afterValues].map(arrayValues =>
+    arrayValues.map(valueMapper),
+  )
+  const overlappingValues = _.intersectionWith(resolvedBeforeValues, resolvedAfterValues, _.isEqual)
+  const valuesToDelete = _.differenceWith(resolvedBeforeValues, overlappingValues, _.isEqual)
```
shir-reifenberg, line 101: in case the values identifier remains the same, but other properties changes, do we need to make a request?

packages/adapter-components/src/filters/array_fields.ts
```diff
@@ -0,0 +1,292 @@
...
+      appliedChanges: [],
+      errors: [
+        {
+          severity: 'Error',
+          message: e.message,
+        },
+      ],
+    }
+  }
+  log.debug('Found %d values to delete and %d values to add', valuesToDelete.length, valuesToAdd.length)
```
shir-reifenberg, line 137: redundant I think ? if we already have the `trace` below ?

packages/adapter-components/src/filters/array_fields.ts
shir-reifenberg: another question, can `values` contain references? or in this case it's already resolved ?

packages/adapter-components/src/filters/array_fields.ts
```diff
@@ -0,0 +1,292 @@
...
+    arrayChangeData: ArrayElementChangeData
+  }): RemovalChange<InstanceElement> | AdditionChange<InstanceElement> => {
+    const instance = new InstanceElement(
+      `${arrayChangeData.name}_${arrayChangeData.id}`,
+      fieldObjectType,
+      arrayChangeData,
+      undefined,
+      {
+        [CORE_ANNOTATIONS.PARENT]: [parentChangeData.value],
+      },
```
shir-reifenberg, line 162: needed ?

packages/adapter-components/src/filters/array_fields.ts
```diff
@@ -0,0 +1,292 @@
...
+    )
+
+    return action === 'remove'
```
shir-reifenberg, line 165: nit: if you first create the instances and then calculate the diff, you'll already know how to devide those to addition and removal change

packages/adapter-components/src/filters/array_fields.ts
```diff
@@ -0,0 +1,292 @@
...
+  const removalChanges = valuesToDelete.map(vals => createChange({ action: 'remove', arrayChangeData: vals }))
+  const removalRes = await deployChanges({
+    changes: removalChanges,
+    ...args,
+  })
+  const additionChanges = valuesToAdd.map(vals => createChange({ action: 'add', arrayChangeData: vals }))
+  const additionRes = await deployChanges({
+    changes: additionChanges,
+    ...args,
+  })
```
shir-reifenberg, line 189:
```suggestion
  const removalChanges = valuesToDelete.map(vals => createChange({ action: 'remove', arrayChangeData: vals }))
 const additionChanges = valuesToAdd.map(vals => createChange({ action: 'add', arrayChangeData: vals }))
  const result = await deployChanges({
    changes: removalChanges.concat(additionChanges),
    ...args,
  })
```

packages/adapter-components/src/filters/array_fields.ts
```diff
@@ -0,0 +1,292 @@
...
+    'values to delete: %s, values to add: %s',
+    safeJsonStringify(valuesToDelete, elementExpressionStringifyReplacer),
+    safeJsonStringify(valuesToAdd, elementExpressionStringifyReplacer),
+  )
+
+  const fieldObjectType = new ObjectType({
+    elemID: new ElemID(singleFieldDefinitions.adapterName, singleFieldDefinitions.fieldTypeName),
+  })
+
+  const createChange = ({
```
shir-reifenberg, line 148: styling nit: (feel free to ignore!!!) I prefer not having functions definitions inside other functions unless I have to

packages/adapter-components/src/filters/array_fields.ts
```diff
@@ -0,0 +1,292 @@
...
+              deployedIds.includes(valueMapper(val).id),
+            )
+            if (_.isEmpty(arrayWithDeployedChangesOnly)) {
+              delete change.data.after.value[fieldName]
+            } else {
+              change.data.after.value[fieldName] = arrayWithDeployedChangesOnly
+            }
+          }
+        })
+        appliedChanges.push(change)
```
shir-reifenberg, line 285: Sorry if i'm missing something here, but don't we need to consider the type of change here ?
i.e what happened if a removal change failed to be applied, wouldn't this code remove it?

shir-reifenberg (commented):
Nice, great job 😎  
I didn't review the entire PR yet, but few things to consider - 
1. I'm not sure this filter should be in adapter-components, as IMO it's a little too entra specific and there are some other things to consider when generalizing it, such as - handling arrays where the order matters, what counts as a change in inner values (only identifier changes or inner values changes), handling arrays not as top level
Also I think that by having the filter only in entra we can simplify it a little. WDYT?
2. error handling - ultimately, deploy result contain `SaltoElementError` that includes the top level element ID. In the case of arrays this is kind of tricky cause it's not obvious when something counts as a complete failure. 
if all changes were failed / we got bad format, we should create a `SaltoElementError`. If some changes were applied, then we shouldn't create any `SaltoElementError`, and let core calculate the differences between the applied change and the requested change.

We can also discuss IRL 🙃

packages/adapter-components/src/filters/array_fields.ts
```diff
@@ -0,0 +1,292 @@
...
+        convertError,
+        changeResolver: createChangeElementResolver<Change<InstanceElement>>({
+          getLookUpName: generateLookupFunc(definitions.references?.rules ?? []),
+        }),
+        sharedContext,
+      }
+
+      const appliedChanges: Change[] = []
+      const errors: (SaltoError | SaltoElementError)[] = []
+      const [relevantChangesInstances, otherChanges] = _.partition(
```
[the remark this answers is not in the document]
shoshana-houri: Oops, confused myself with the `getChangeData` that checks the typename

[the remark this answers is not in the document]
shoshana-houri: Oops, confused myself with the `getChangeData` that checks the `typename`

packages/adapter-components/src/filters/array_fields.ts
```diff
@@ -0,0 +1,292 @@
...
+        changes,
+        change =>
+          isInstanceChange(change) &&
+          isAdditionOrModificationChange(change) &&
+          getChangeData(change).elemID.typeName === parentTypeName,
+      ) as [AdditionOrModificationChange[], Change[]]
+
+      await awu(relevantChangesInstances).forEach(async change => {
```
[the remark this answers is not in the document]
shoshana-houri: We want it to be deployed before we deploy the array changes, and as mentioned [here](https://github.com/salto-io/salto/blob/ec427bf2dd709ce8779d6b35a82c03fc0a5a059c/packages/adapter-components/src/filters/common_filters.ts#L70), the `default Deploy` filter (which handles the general deployment) should run after other deploy filters.

packages/adapter-components/src/filters/array_fields.ts
shoshana-houri: Ohhhh, that's a good point! In this case, these objects only include one field, which is indeed a reference. I guess using isEqualValues would be appropriate in this case? :)

packages/adapter-components/src/filters/array_fields.ts
shoshana-houri: Mmm, trace is only available when rehydrating logs, so I think it makes sense to log a shorter version as debug and add trace with the full info. WDYT?

packages/adapter-components/src/filters/array_fields.ts
shoshana-houri: AFAIU yes, since we need the `parent_id` for the deploy request.

packages/adapter-components/src/filters/array_fields.ts
shoshana-houri: Talked face-to-face – will remove since we should try reducing logs, and rehydrating the trace should be okay.

packages/adapter-components/src/filters/array_fields.ts
shoshana-houri: Talked f2f - will skip that

packages/adapter-components/src/filters/array_fields.ts
shoshana-houri: 100% right, thank you!

packages/microsoft-entra-adapter/test/deploy/utils/group_deployment_life_cycle_policy.test.ts
```diff
@@ -74,3 +77,43 @@ describe(`${getGroupLifecyclePolicyGroupModificationRequest.name}`, () => {
     })
   })
 })
+
+describe(`${createDefinitionForGroupLifecyclePolicyGroupModification.name}`, () => {
```
shoshana-houri, line 81: Not related to this PR, but noticed this one is not being tested.

packages/adapter-components/src/deployment/deploy/deploy.ts
```diff
@@ -52,7 +52,7 @@ export const defaultConvertError: ConvertError = (elemID, error) => {
 /**
  * Deploy a change with standard (add / modify / remove) and custom-defined actions based on the provided deploy definitions
  */
-const createSingleChangeDeployer = <TOptions extends APIDefinitionsOptions>({
```
shir-reifenberg, line 55: we can revert this change cause it's not in use, right ?

shir-reifenberg (approved):
Great work! 🎉 
added mainly nit so approving :)

packages/microsoft-entra-adapter/src/filters/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+  validatePlainObject(value, 'member reference')
+
+  const memberRef = _.get(value, 'id')
+  const odataType = _.get(value, ODATA_TYPE_FIELD_NACL_CASE)
+  if (!isReferenceExpression(memberRef) || !_.isString(odataType)) {
+    throw new Error('Invalid member reference, missing id reference or OData type')
+  }
+
+  const refValue = _.get(memberRef, 'resValue.value')
+  validatePlainObject(refValue, 'member reference value')
```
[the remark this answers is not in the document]
shoshana-houri: Agree, but for some reason `memberRef.value` is of type `any`, so I can't use it directly. If it's better I guess I can validate that `memberRef.value` is InstanceElement and then proceed :)

packages/microsoft-entra-adapter/src/filters/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+export const mapMemberRefToChangeData: MapArrayValueToChangeData = value => {
+  validatePlainObject(value, 'member reference')
+
+  const memberRef = _.get(value, 'id')
+  const odataType = _.get(value, ODATA_TYPE_FIELD_NACL_CASE)
+  if (!isReferenceExpression(memberRef) || !_.isString(odataType)) {
+    throw new Error('Invalid member reference, missing id reference or OData type')
+  }
+
+  const refValue = _.get(memberRef, 'resValue.value')
```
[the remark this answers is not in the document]
shoshana-houri: But - I'll use the approach [suggested here](https://github.com/salto-io/salto/pull/6064#discussion_r1655527209) instead

packages/microsoft-entra-adapter/src/filters/array_fields_deployment.ts
```diff
@@ -0,0 +1,349 @@
...
+            errors: [
+              {
+                severity: 'Error',
+                message: 'change group not provided',
+              },
+            ],
+          },
+          leftoverChanges: changes,
+        }
+      }
```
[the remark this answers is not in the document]
shoshana-houri: Would love to, but it's required as one of the args to `deployChanges`, which is used later.

packages/adapter-components/src/filters/array_fields.ts
```diff
@@ -0,0 +1,292 @@
...
+        const changeData = getChangeData(change)
+        const { errors: deployErrors } = await deployChanges({
+          changes: [change],
```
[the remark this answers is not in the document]
shoshana-houri: Talked f2f - decided to use `deployChanges` to avoid repeating the logic

PR #6065, opened 2024-06-16
SALTO-5773: Bumping fast-xml-parser

yelly:
This is a major which required a bunch of changes. The reason for the bump is the fix for https://github.com/NaturalIntelligence/fast-xml-parser/issues/350 which solves the bug, but there are also a lot of additional bugfixes. Also had to copy some code from the SF E2E utils to the UT utils to get the linter to stop complaining about the TS scope.

---

WS diff: https://github.com/salto-io/ariel-multi/commit/2da3c534fe1a9745b7ddf192403842ae852edcca (nothing of note in NS or SF).

---
_Release Notes_: 
_Salesforce_:
* Fixes a bug where some fetches can result in broken NaCl files.

_General_:
* Upgrading `fast-xml-parser` (used in NS and SF) from 3.15.0 to 4.4.0.

---
_User Notifications_: 
None.


packages/netsuite-adapter/src/client/sdf_parser.ts
```diff
@@ -75,11 +74,10 @@ const DEFAULT_FOLDER_ATTRIBUTES =
   `  <isprivate>F</isprivate>${os.EOL}` +
   `</folder>${os.EOL}`
 
-const XML_PARSE_OPTIONS: xmlParser.J2xOptionsOptional = {
+const xmlParser = new XMLParser({
   attributeNamePrefix: ATTRIBUTE_PREFIX,
   ignoreAttributes: false,
-  tagValueProcessor: val => he.decode(val),
```
dantal4, line 81: what about this arg? we need it for special characters

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -129,15 +128,13 @@ const fetchTransformFunc = async ({ value, field, path }: TransformFuncArgs, typ
...
+  attributeNamePrefix: ATTRIBUTE_PREFIX,
+  ignoreAttributes: false,
+})
+
 const createAnalyticsInstance = async (instance: InstanceElement, analyticsType: ObjectType): Promise<void> => {
-  const definitionValues = _.omit(
-    parse(instance.value[DEFINITION], {
-      attributeNamePrefix: ATTRIBUTE_PREFIX,
-      ignoreAttributes: false,
-      tagValueProcessor: val => decode(val),
```
dantal4, line 137: here too

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
```diff
@@ -317,6 +314,14 @@ const createDefinitionName = (instance: InstanceElement, definitionValues: Value
       : instance.value[NAME]
 }
 
+const xmlBuilder = new XMLBuilder({
+  attributeNamePrefix: ATTRIBUTE_PREFIX,
+  format: true,
+  ignoreAttributes: false,
+  cdataPropName: CDATA_TAG_NAME,
+  tagValueProcessor: (_name, val) => String(val),
```
dantal4, line 322: here too

packages/netsuite-adapter/test/client/mocks.ts
```diff
@@ -108,7 +108,7 @@ export const MOCK_ORIGINAL_DEPLOY_XML = `<deploy>
 
 export const OBJECT_XML_WITH_HTML_CHARS =
   '<entitycustomfield scriptid="custentity_my_script_id">' +
-  '<label>Golf &#x26; Co&#x2019;Co element&#x200B;Name</label>' +
+  '<label>Golf &amp; Co’Co element​Name</label>' +
```
dantal4, line 111: SDF expects the format that was there, @omrilit  right?

packages/netsuite-adapter/test/filters/analytics_tests_constants.ts
```diff
@@ -679,7 +679,7 @@ export const definitionWithArrays = `<root>
...
...x3E;dataDimension&#x3C;/_T_&#x3E;&#x3C;id&#x3E;4&#x3C;/id&#x3E;&#x3C;items type=&#x22;array&#x22;&#x3E;&#x3C;_ITEM_&#x3E;&#x3C;id&#x3E;1&#x3C;/id&#x3E;&#x3C;expression&#x3E;&#x3C;function&#x3E;field&#x3C;/function&#x3E;&#x3C;args&#x3E;&#x3C;type&#x3E;&#x3C;_T_&#x3E;textType&#x3C;/_T_&#x3E;&#x3C;/type&#x3E;&#x3C;dataSourceId&#x3E;custdataset_dilly_1&#x3C;/dataSourceId&#x3E;&#x3C;fieldId&#x3E;description&#x3C;/fieldId&#x3E;&#x3C;/args&#x3E;&#x3C;/expression&#x3E;&#x3C;/_ITEM_&#x3E;&#x3C;/items&#x3E;&#x3C;/_ITEM_&#x3E;&#x3C;/children&#x3E;&#x3C;/dimensionTree&#x3E;&#x3C;/rowAxis&#x3E;&#x3C;measures type=&#x22;array&#x22;&#x3E;&#x3C;_ITEM_&#x3E;&#x3C;_T_&#x3E;dataMeasure&#x3C;/_T_&#x3E;&#x3C;id&#x3E;3&#x3C;/id&#x3E;&#x3C;aggregation&#x3E;count&#x3C;/aggregation&#x3E;&#x3C;expression&#x3E;&#x3C;function&#x3E;field&#x3C;/function&#x3E;&#x3C;args&#x3E;&#x3C;type&#x3E;&#x3C;_T_&#x3E;recordType&#x3C;/_T_&#x3E;&#x3C;id&#x3E;record_accttype&#x3C;/id&#x3E;&#x3C;/type&#x3E;&#x3C;dataSourceId&#x3E;custdataset_dilly_1&#x3C;/dataSourceId&#x3E;&#x3C;fieldId&#x3E;accttype&#x3C;/fieldId&#x3E;&#x3C;/args&#x3E;&#x3C;/expression&#x3E;&#x3C;/_ITEM_&#x3E;&#x3C;/measures&#x3E;&#x3C;/root&#x3E;</definition>
```
dantal4, line 682: same here

packages/netsuite-adapter/src/client/deploy_xml_utils.ts
```diff
@@ -27,6 +27,16 @@ const log = logger(module)
 // The '/' prefix is needed to make osPath.resolve treat '~' as the root
 const PROJECT_ROOT_TILDE_PREFIX = `${osPath.sep}~`
 
+const xmlParser = new XMLParser({
```
dantal4, line 30: shouldn't we create those objects inside the function that uses them?

dantal4 (commented):
NS looks good 👍  added some comments

packages/netsuite-adapter/src/client/sdf_parser.ts
yelly: `fast-xml-parser` now encodes (some) special characters itself. I have a feeling this may be enough so I'm seeing if everything works without this. If not, we have a bit of a headache because some characters (most notably &) will get encoded twice.
If I can get the E2E tests passing, let's discuss how we can get confidence that everything works the way we want. Do you know of any NS docs which describe the encoding they are expecting?

packages/netsuite-adapter/test/client/mocks.ts
yelly: See above, let's see if we can find any docs. `fast-xml-parser` now encodes whatever it needs to in order to ensure a valid XML output. It might be that NS is more strict than that in which case we need to do some legwork to make sure we don't get double encoding.

packages/netsuite-adapter/src/client/deploy_xml_utils.ts
yelly: I guess it's a matter of convention. Do we prefer to create them every time instead of sharing?

packages/netsuite-adapter/src/client/sdf_parser.ts
dantal4: @omrilit  do you know?

packages/netsuite-adapter/src/client/deploy_xml_utils.ts
dantal4: not specifically, just wondered.

packages/netsuite-adapter/src/client/sdf_parser.ts
omrilit: I am not familiar with such documentation.
I think that we should fetch before this change and after this change and make sure that there are no changes.
We should probably do it in the Development & Extended Dev & Dilly accounts to make sure we wrap as much use cases as we can. WDYT @dantal4 ?

dantal4: @omrilit I agree. we should check the deployment of special characters too.

dantal4: Here's an example of the "carriage return" character that isn't parsed correctly using this PR's code:
```diff
@@ -18,7 +18,7 @@ netsuite.entitycustomfield custentity11 {
   checkspelling = false
   defaultchecked = false
   description = '''
-desc from UI
+desc from UI&#xd;
 test
 '''
   displaytype = "NORMAL"
```

yelly: OK I've dug deeper and done modelling and turns out we do need this.

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
yelly: Yes.

packages/netsuite-adapter/src/filters/analytics_definition_handle.ts
yelly: This actually required finding another option but I did.

packages/netsuite-adapter/src/filters/fix_order_problems_in_workbooks.ts
```diff
@@ -41,21 +41,19 @@ const isStringArray = (val: unknown): val is string[] => Array.isArray(val) && _
...
+  attributeNamePrefix: ATTRIBUTE_PREFIX,
+  ignoreAttributes: false,
+  tagValueProcessor: (_name, val) => he.decode(val),
+})
+
 const isSameXmlValues = (xml1: string, xml2: string): boolean => {
-  const values1 = parse(xml1, {
-    attributeNamePrefix: ATTRIBUTE_PREFIX,
-    ignoreAttributes: false,
-    tagValueProcessor: val => decode(val),
```
yelly, line 48: Note that here we are just comparing so the decoding is unnecessary work.

packages/netsuite-adapter/test/filters/analytics_tests_constants.ts
yelly: Done.

dantal4 (commented):
NS looks good in general, but there are still some things that changes in the workspace because of that:

inner xmls in workbooks shouldn't be parsed:
```diff
@@ -132,7 +330,41 @@ netsuite.workbook custworkbook_seggev_table_dslink_pivot_and_chart {
           "custdataset_seggev_account_basic",
           "custdataset_seggev_research",
         ]
-        mapping = "<root><_T_>reportDataSourceExpressionMapping</_T_><expressions type=\"array\"><_ITEM_ type=\"array\"><_ITEM_><function>field</function><args><type><_T_>recordKeyType</_T_><valueType><_T_>integerType</_T_></valueType>
</type><dataSourceId>custdataset_seggev_account_basic</dataSourceId><fieldId>id</fieldId></args></_ITEM_><_ITEM_><function>field</function><args><type><_T_>recordKeyType</_T_><valueType><_T_>integerType</_T_></valueType></type><dataSourceId>custdataset_seggev_research</dataSourceId><fieldId>id</fieldId></args></_ITEM_></_ITEM_></expressions></root>"
+        mapping = {
+          root = {
+            fieldType = "reportDataSourceExpressionMapping"
+            expressions = [
+              [
+                {
+                  function = "field"
+                  args = {
+                    type = {
+                      fieldType = "recordKeyType"
+                      valueType = {
+                        fieldType = "integerType"
+                      }
+                    }
+                    dataSourceId = "custdataset_seggev_account_basic"
+                    fieldId = "id"
+                  }
+                },
+                {
+                  function = "field"
+                  args = {
+                    type = {
+                      fieldType = "recordKeyType"
+                      valueType = {
+                        fieldType = "integerType"
+                      }
+                    }
+                    dataSourceId = "custdataset_seggev_research"
+                    fieldId = "id"
+                  }
+                },
+              ],
+            ]
+          }
+        }
       }
     },
   ]
```

places where `&amp;` changed to `&`:
```diff
@@ -72,7 +72,7 @@ settings netsuite.companyFeatures {
     label = "Expense migration scheduled from start"
   }
   boolean DUPLICATES {
-    label = "Duplicate Detection &amp; Merge"
+    label = "Duplicate Detection & Merge"
   }
   boolean FCADVANCEDSECURITY {
     label = "File Cabinet Advanced Security"
@@ -288,7 +288,7 @@ settings netsuite.companyFeatures {
     label = "Payment Link"
   }
   boolean DROPSHIPMENTS {
-    label = "Drop Shipments &amp; Special Orders"
+    label = "Drop Shipments & Special Orders"
   }
   boolean MATRIXITEMS {
     label = "Matrix Items"
```

```diff
@@ -122,7 +122,7 @@ type netsuite.customrecord11 {
     encryptatrest = false
     fieldtype = "HELP"
     globalsearch = false
-    help = "&lt;p&gt;"
+    help = "<p>"
     isformula = false
     ismandatory = false
     isparent = false
@@ -144,7 +144,7 @@ type netsuite.customrecord11 {
     encryptatrest = false
     fieldtype = "HELP"
     globalsearch = false
-    help = "&lt;p&gt;"
+    help = "<p>"
     isformula = false
     ismandatory = false
     isparent = false
@@ -191,7 +191,7 @@ type netsuite.customrecord11 {
     isformula = false
     ismandatory = false
     isparent = false
-    label = "&gt;"
+    label = ">"
     onparentdelete = "NO_ACTION"
     rolerestrict = false
     searchlevel = "2"
```

packages/adapter-utils/src/utils.ts
```diff
@@ -822,16 +822,28 @@ type MapKeysRecursiveArgs = {
...
+    return obj.map((val, idx) => mapKeysRecursive(val, func, pathID?.createNestedID(String(idx)), keySortFunc))
   }
   if (_.isPlainObject(obj)) {
-    return _(obj)
-      .mapKeys((_val, key) => func({ key, pathID: pathID?.createNestedID(String(key)) }))
-      .mapValues((val, key) => mapKeysRecursive(val, func, pathID?.createNestedID(String(key))))
-      .value()
+    const entries = Object.entries(obj)
+    if (keySortFunc !== undefined) {
+      entries.sort(([aKey, _aVal], [bKey, _bVal]) => keySortFunc(aKey, bKey))
```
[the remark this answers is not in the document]
yelly: Thanks!

packages/netsuite-adapter/src/transformer.ts
```diff
@@ -184,7 +185,16 @@ export const createInstanceElement = async (
   const transformAttributeKey: MapKeyFunc = ({ key }) =>
     key.startsWith(ATTRIBUTE_PREFIX) ? key.slice(ATTRIBUTE_PREFIX.length) : key
 
-  const valuesWithTransformedAttrs = mapKeysRecursive(customizationInfo.values, transformAttributeKey)
+  // We later assume that the `scriptid` value (which comes from an attribute) is always first, so we put them first.
+  const nonAttributeKeyIndicator = (k: string): number => (k.startsWith(ATTRIBUTE_PREFIX) ? 0 : 1)
+  const sortKeysAttributesFirst: SortKeysFunc = (a, b) => nonAttributeKeyIndicator(a) - nonAttributeKeyIndicator(b)
```
ori-moisis, line 190: Is this not overkill to recursively sort everything if we only need one specific value to be first?

also, how hard would it be to change the assumption that scriptid is first? where is that assumption made?

packages/adapter-utils/src/utils.ts
```diff
@@ -822,16 +822,28 @@ type MapKeysRecursiveArgs = {
...
-
-export const mapKeysRecursive = (obj: Values, func: MapKeyFunc, pathID?: ElemID): Values => {
+export type SortKeysFunc = (a: string, b: string) => number
+
+export const mapKeysRecursive = (
+  obj: Values,
+  func: MapKeyFunc,
+  pathID?: ElemID,
+  keySortFunc?: SortKeysFunc,
+): Values => {
```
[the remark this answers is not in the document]
yelly: Done.

packages/netsuite-adapter/src/transformer.ts
yelly: The assumption is made at `packages/netsuite-adapter/src/filters/fix_order_problems_in_workbooks.ts` (I've added a comment there in this PR).
There are a few possible solutions here which I discussed with @dantal4 before we landed on this, but we can change the decision:
1. Change `walkOnElement` (and possibly other walk functions) to be BFS instead of DFS.
2. Change the code in `getServiceIdsToElemIds` to look for a `serviceID` value in the object being walked instead of waiting to get to it in the walk, removing the need for it to be first.
3. Specifically hoist `serviceID` before we get to `getServiceIdsToElemIds`. This will require another walk.
4. Hoist all attributes before we get to `getServiceIdsToElemIds`. This will require another walk.
5. Sort keys in the walk (what we ended up going with).
Another thing to keep in mind is if we don't move the attributes to the start this will cause the values that come from attributes in elements to move from the beginning to the end. This will cause a diff in the NaCl if users clean and fetch or if the value is removed and returned. Not necessarily a huge deal but it's a consideration.
WDYT?

yelly:
> NS looks good in general, but there are still some things that changes in the workspace because of that:
> 
> inner xmls in workbooks shouldn't be parsed:
> ...
> 
> places where `&amp;` changed to `&`:
> ...

IMO decoding named entities (`&amp;`, `&lt;`, `&gt;` etc.) is actually a feature. It makes the NaCls more readable and editable. We encode them back in deploy so not a problem.

As for the inner XMLs - I can't reproduce this. Are these from the `tstdrv2257860` app?

@dantal4

packages/adapter-utils/src/utils.ts
```diff
@@ -822,18 +822,47 @@ type MapKeysRecursiveArgs = {
...
+export const mapKeysRecursive = ({
+  values,
+  func,
+  pathID,
+  keySortFunc,
+}: {
+  values: Values
+  func: MapKeyFunc
+  pathID?: ElemID
+  keySortFunc?: SortKeysFunc
```
ori-moisis, line 836: missing tests for this new feature
I think it might be best to put this in a separate PR btw as this is really detached from bumping fast-xml-parser

packages/adapter-utils/src/utils.ts
```diff
@@ -822,18 +822,47 @@ type MapKeysRecursiveArgs = {
...
+    )
   }
-  if (_.isPlainObject(obj)) {
-    return _(obj)
-      .mapKeys((_val, key) => func({ key, pathID: pathID?.createNestedID(String(key)) }))
-      .mapValues((val, key) => mapKeysRecursive(val, func, pathID?.createNestedID(String(key))))
-      .value()
+  if (_.isPlainObject(values)) {
+    const entries = Object.entries(values)
+    if (keySortFunc !== undefined) {
```
ori-moisis, line 850: not trivial to understand that the sort function operations on the keys before the mapping function.

not sure how to solve it (one thing is to allow the called to choose when the sort runs, but that would require more implementation that isn't really needed), if it can't be solved by naming / api, maybe a comment?

and a test should explicitly state this, because it can be fairly confusing (e.g - if I sort alphabetically, and then map `a -> z`, I'd end up with `z` as the first key)

dantal4:
> > NS looks good in general, but there are still some things that changes in the workspace because of that:
> > inner xmls in workbooks shouldn't be parsed:
> > ...
> > places where `&amp;` changed to `&`:
> > ...
> 
> IMO decoding named entities (`&amp;`, `&lt;`, `&gt;` etc.) is actually a feature. It makes the NaCls more readable and editable. We encode them back in deploy so not a problem.
> 
> As for the inner XMLs - I can't reproduce this. Are these from the `tstdrv2257860` app? You can see the diff I got here: [salto-io/ariel-multi@d87c1bd#diff-b63bf86accd68d7549ae0ae87739147bf9e7d9712d65cdc6d83ba82777729160](https://github.com/salto-io/ariel-multi/commit/d87c1bda0f35a952dc2b151add42989749ebb87e#diff-b63bf86accd68d7549ae0ae87739147bf9e7d9712d65cdc6d83ba82777729160).
> 
> @dantal4

inner xmls - you need to remove the exclusion of workbooks & datasets in the [adapter config](https://github.com/salto-io/ariel-multi/blob/d87c1bda0f35a952dc2b151add42989749ebb87e/salto.config/adapters/netsuite/netsuite.nacl#L31-L36),

decoding named entities - any change in NS customers' data requires a PROD ticket and a communication email that explains the steps that they should do in order to keep their audit log organized (fetch>push>fetch after deployment>push). This can be done, but it means that this PR will have to wait ~1 week before merging it. @omrilit FYI

packages/netsuite-adapter/src/transformer.ts
dantal4: it's not only the top level - inner objects have the scriptid field too, which we'd like to be at the top as it's the object's id.

ori-moisis: I think there are two separate things here.
1. I assume you mean `packages/netsuite-adapter/src/service_id_info.ts` and not `packages/netsuite-adapter/src/filters/fix_order_problems_in_workbooks.ts`.
I would go for 2 (change the code the not assume order). there is no reason to have this assumption there, it shouldn't really make the code more complex and it will be more robust, I don't see a down side really...
2. the thing about the values changing order is a thing, so if the xml parser changed the order compared to how it used to be, I agree it is something we would want to fix.
but then I am not sure that fixing it just here is good enough - if this is a general change in the parser, don't we need to fix the order anywhere that we parse things which could contain attributes?

yelly: 1. Yes. I'll change to 2.
2. Just NS and SF. The order change is because values that are parsed from attributes go to the end instead of the beginning. I haven't yet found any cases in SF where attributes turn into visible values, so I don't think that's a problem there, leaving only NS. Personally I don't think this is such a big deal because it will only manifest if the value is removed and then returned and the field ordering is a new feature so it's not like everyone expects it.

packages/adapter-utils/src/utils.ts
yelly: Removing it anyway.

packages/adapter-utils/src/utils.ts
yelly: Removing this anyway.

packages/netsuite-adapter/src/transformer.ts
dantal4: I think that in NS it's important that we'll keep the scriptid at top, because we want the elements and inner objects to be aligned - if a new inner object will have the scriptid at the bottom and existing inner objects in that element will have the scriptid at top it won't look good. In addition, if I'm not wrong, *important values* aren't sorting inner objects, so users won't see the scriptid of inner objects at top.

IMO it's not a overkill to run the keys sort in `mapKeysRecursive`, but we can change it to use `_.sortBy`, that doesn't compare each 2 keys.

@omrilit  FYI

ori-moisis: +1 - the order of values is important

this is not just about changes, this also affects new workspaces - the script id being at the top is part of the UX of the nacl

as for this being the only place where attributes are values, there is a filter in SF that transforms attributes in values - https://github.com/salto-io/salto/blob/main/packages/salesforce-adapter/src/filters/xml_attributes.ts/#L43
I assume that, given it exists, there are values that come from attributes (seems like this happens at least in `LightningComponentBundle` instances, but maybe others as well...)

packages/netsuite-adapter/src/client/deploy_xml_utils.ts
```diff
@@ -56,8 +55,5 @@ export const reorderDeployXml = (deployContent: string, dependencyGraph: Graph<S
   }
 
   // eslint-disable-next-line new-cap
```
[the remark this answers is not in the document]
yelly: Yes.

packages/netsuite-adapter/src/service_id_info.ts
```diff
@@ -106,11 +107,25 @@ export const getServiceIdsToElemIds = (element: Element): ServiceIdRecords => {
...
+      let container = value
+      let containerPath = path
+      if (isInstanceElement(value)) {
+        container = value.value
+      } else if (isElement(value)) {
+        container = value.annotations
+        if (isType(value)) {
+          containerPath = path.createNestedID('attr')
+        }
+      }
```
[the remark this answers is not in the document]
yelly: Nice. Also changed the flow a bit to improve typing.

packages/netsuite-adapter/src/service_id_info.ts
```diff
@@ -106,11 +107,25 @@ export const getServiceIdsToElemIds = (element: Element): ServiceIdRecords => {
...
+      if (_.isPlainObject(container) && typeof container[SCRIPT_ID] === 'string') {
+        const scriptID = container[SCRIPT_ID]
```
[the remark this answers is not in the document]
yelly: Done.

packages/netsuite-adapter/src/client/sdf_parser.ts
```diff
@@ -95,8 +98,35 @@ export const getDeployFilePath = (projectPath: string): string => osPath.resolve
 export const getFeaturesXmlPath = (projectPath: string): string =>
   osPath.resolve(projectPath, SRC_DIR, ACCOUNT_CONFIGURATION_DIR, FEATURES_XML)
 
+type XMLNode = Dictionary<XMLNode> | XMLNode[] | string
```
dantal4, line 101: we commonly use `Record<string, abc>` instead of `Dictionary`.

packages/netsuite-adapter/src/client/sdf_parser.ts
```diff
@@ -95,8 +98,35 @@ export const getDeployFilePath = (projectPath: string): string => osPath.resolve
...
+    return _.mapValues(values, flattenCDATA)
+  }
+
+  return values
+}
+
 const convertToCustomizationInfo = (xmlContent: string): CustomizationInfo => {
-  const parsedXmlValues = xmlParser.parse(xmlContent, XML_PARSE_OPTIONS)
+  // We need to remove all CDATA tags we added to avoid decoding.
+  const parsedXmlValues = flattenCDATA(xmlParser.parse(xmlContent)) as Dictionary<Dictionary<XMLNode>>
```
dantal4, line 129: can we avoid casting? maybe by wrapping `flattenCDATA` with a function that receives `Record<..>` and returns `Record<..>`?

dantal4 (commented):
💯  on NS code - I also checked the fetch & deploy of inner XMLs and it works like a charm 🥇

packages/netsuite-adapter/src/client/sdf_parser.ts
yelly: Done.

packages/netsuite-adapter/src/client/sdf_parser.ts
yelly: We could maybe do this with some fancy overloading but I think it's overkill in this case.

packages/salesforce-adapter/src/filters/xml_attributes.ts
```diff
@@ -31,27 +31,35 @@ import { metadataTypesWithAttributes } from '../transformers/xml_transformer'
...
-      : key
-  })
+  // We put the attributes first for backwards compatibility.
+  const [attrEntries, entries] = _(value)
+    .entries()
+    .partition(
+      ([key, _val]) =>
+        type.fields[key.replace(XML_ATTRIBUTE_PREFIX, '')]?.annotations[
+          IS_ATTRIBUTE
+        ],
```
tamtamirr, line 46: Make this type safe by validating the type of the annotation and existence

packages/salesforce-adapter/test/utils.ts
```diff
@@ -51,6 +57,61 @@ import {
   OptionalFeatures,
 } from '../src/types'
 
+export const nullProgressReporter: ProgressReporter = {
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  reportProgress: () => {},
+}
+
+export const createElement = async <T extends InstanceElement | ObjectType>(
```
tamtamirr, line 65: Let's not duplicate this implementation

tamtamirr (approved):
LGTM!

packages/salesforce-adapter/test/utils.ts
yelly: Done.

PR #6066, opened 2024-06-16
SALTO-6105: split allow empty in transformValues and transformValuesSync

shir-reifenberg:
Split `allowEmpty` param `transformValues` and `transformValuesSync` to distinguish between arrays and objects

---

though `transformValues` and `transformValuesSync` are exported but currently not in use, I removed it entirely.
added `allowEmptyArrays` and `allowEmptyObjects` to `transformElement, left `allowEmpty` in `transformElement` for backward compatibility and will remove it in a separate PR

---
_Release Notes_: 

_Core_:
- Change `transformValues` and `transformValuesSync` params by splitting `allowEmpty` param to `allowEmptyArray` and `allowEmptyObjects`
- Add `allowEmptyArrays` and `allowEmptyObjects` to `transformElement`, `allowEmpty` is deprecated and will be deleted soon

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-utils/src/utils.ts
+++ b/packages/adapter-utils/src/utils.ts
@@ -240,7 +240,7 @@ const recurseIntoValue = ({
       // because we cannot get here with an object
       const isEmptyString = _.isString(newVal) && _.isEmpty(newVal)
       if (isEmptyString) {
-        log.warn('found empty string %s in field %s, string will be omitted: %s', newVal, field?.name, allowEmptyArrays)
+        log.warn('found empty string found in path %s, string will be omitted: %s', keyPathID, !allowEmptyArrays)
       }
       const valueIsEmpty = (Array.isArray(newVal) && _.isEmpty(newVal)) || isEmptyString
       return valueIsEmpty && !allowEmptyArrays ? undefined : newVal
--- a/packages/adapter-utils/src/utils.ts
+++ b/packages/adapter-utils/src/utils.ts
@@ -85,7 +85,8 @@ type TransformValuesBaseArgs = {
   strict?: boolean
   pathID?: ElemID
   isTopLevel?: boolean
-  allowEmpty?: boolean
+  allowEmptyArrays?: boolean
+  allowEmptyObjects?: boolean
 }
 
 type TransformValuesSyncArgs = TransformValuesBaseArgs & { transformFunc: TransformFuncSync }
@@ -149,14 +150,22 @@ const fieldMapperGenerator = (type: ObjectType | TypeMap | MapType | ListType, v
       : undefined
 }
 
-const removeEmptyParts = (value: Value, allowEmpty: boolean): Value => {
+const removeEmptyParts = ({
+  value,
+  allowEmptyArrays,
+  allowEmptyObjects,
+}: {
+  value: Value
+  allowEmptyArrays: boolean
+  allowEmptyObjects: boolean
+}): Value => {
   if (Array.isArray(value)) {
     const filtered = value.filter(isDefined)
-    return filtered.length === 0 && (value.length > 0 || !allowEmpty) ? undefined : filtered
+    return filtered.length === 0 && (value.length > 0 || !allowEmptyArrays) ? undefined : filtered
   }
   if (_.isPlainObject(value)) {
     const filtered = _.omitBy(value, _.isUndefined)
-    return _.isEmpty(filtered) && (!_.isEmpty(value) || !allowEmpty) ? undefined : filtered
+    return _.isEmpty(filtered) && (!_.isEmpty(value) || !allowEmptyObjects) ? undefined : filtered
   }
   return value
 }
@@ -165,7 +174,7 @@ type recurseIntoValueArgs = {
   newVal: Value
   transformFunc: (value: Value, keyPathID?: ElemID, field?: Field) => Value
   strict: boolean
-  allowEmpty: boolean
+  allowEmptyArrays: boolean
   isAsync: boolean
   keyPathID?: ElemID
   field?: Field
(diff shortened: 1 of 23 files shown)


packages/adapter-utils/test/utils.test.ts
```diff
@@ -864,24 +864,48 @@ describe('Test utils.ts', () => {
...
           type: new ListType(BuiltinTypes.NUMBER),
           transformFunc: ({ value }) => value,
-          allowEmpty: true,
+          allowEmptyArrays: true,
         })
 
         expect(result).toEqual([])
       })
 
+      it('should remove empty object', async () => {
```
ori-moisis, line 879: these tests are not sufficient IMO.
they only test the code of the "top level" transform values, they don't test the logic for nested values.

I think generally we should have tests for both options, but this is even more important given that (for some reason I don't fully understand) the implementation is actually different, and I think actually has different behavior for top level and for nested values

packages/adapter-utils/src/utils.ts
```diff
@@ -422,7 +434,8 @@ export const transformElement = async <T extends Element>({
         strict,
         elementsSource,
         pathID: element.elemID,
-        allowEmpty,
```
ori-moisis, line 425: don't we want to expose the new options in `transformElement` as well?
I thought the point was to replace the old argument eventually, no?

shir-reifenberg: I can do this for completeness, but to resolve the bug in Okta I only need `transformValues`. So I can either open a ticket / do it in a separate PR - your call

packages/adapter-utils/src/utils.ts
```diff
@@ -230,7 +239,7 @@ const recurseIntoValue = ({
       // objects, arrays and strings. we don't need to check for objects here
       // because we cannot get here with an object
       const valueIsEmpty = (Array.isArray(newVal) || _.isString(newVal)) && _.isEmpty(newVal)
-      return valueIsEmpty && !allowEmpty ? undefined : newVal
+      return valueIsEmpty && !allowEmptyArrays ? undefined : newVal
```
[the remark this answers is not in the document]
shir-reifenberg: this doesn't change existing behavior, but I agree it's inconsistent
let's discuss IRL ?

packages/adapter-utils/src/utils.ts
```diff
@@ -229,8 +238,12 @@ const recurseIntoValue = ({
       // but we do not want to omit those, we only want to omit empty
       // objects, arrays and strings. we don't need to check for objects here
       // because we cannot get here with an object
-      const valueIsEmpty = (Array.isArray(newVal) || _.isString(newVal)) && _.isEmpty(newVal)
-      return valueIsEmpty && !allowEmpty ? undefined : newVal
+      const isEmptyString = _.isString(newVal) && _.isEmpty(newVal)
+      if (isEmptyString) {
+        log.warn('found empty string %s in field %s, string will be omitted: %s', newVal, field?.name, allowEmptyArrays)
```
ori-moisis, line 243: do we only need to log this if `allowEmpty` is true?

packages/adapter-utils/test/utils.test.ts
```diff
@@ -864,27 +864,78 @@ describe('Test utils.ts', () => {
...
+              fields: {
+                nestedArr: { refType: new ListType(BuiltinTypes.STRING) },
+                val: { refType: BuiltinTypes.STRING },
+              },
+            }),
+          },
+          emptyObj: { refType: new ObjectType({ elemID: new ElemID('adapter', 'emptyType') }) },
+        },
       })
+      describe('with allowEmptyArrays', () => {
```
ori-moisis, line 884: nit:
1. there should be a test for allowing both to be empty
2. super nit - in the `it` description, would be better to describe what it should do before describing what it shouldn't do (e.g - `should remove lists and not remove objects` is better than `should not remove objects and remove lists`) because the "remove lists" is the main purpose

ori-moisis (approved):
LGTM
please open a follow up PR to remove all remaining usage of `allowEmpty`

packages/adapter-utils/src/utils.ts
ori-moisis, line 243: two things about this
1. is the field name enough? can we get the path of the value here? if not, maybe the full ID of the field?
2. is the boolean at the end here opposite to what the log is saying? (one is saying the string will be omitted, but the boolean is the one that would allow it to remain there IIUC)

shir-reifenberg: 1. ok
2. you're right
3. I thought of is as a property and it's values, like `omit: false / true` but I understand what you're saying

ori-moisis: I am very bad at boolean logic, but I think if the log wants to say omit, then the value needs to be `!allowEmpty`

shir-reifenberg: you're right 🫠

packages/adapter-utils/src/utils.ts
```diff
@@ -229,8 +238,12 @@ const recurseIntoValue = ({
       // but we do not want to omit those, we only want to omit empty
       // objects, arrays and strings. we don't need to check for objects here
       // because we cannot get here with an object
-      const valueIsEmpty = (Array.isArray(newVal) || _.isString(newVal)) && _.isEmpty(newVal)
-      return valueIsEmpty && !allowEmpty ? undefined : newVal
+      const isEmptyString = _.isString(newVal) && _.isEmpty(newVal)
+      if (isEmptyString) {
+        log.warn('found empty string found in path %s, string will be omitted: %s', keyPathID, !allowEmptyArrays)
```
ori-moisis, line 243: `keyPathID` is an elem ID
this should be `keyPathID.getFullName()`
also
```suggestion
log.warn('found empty string in path %s, string will be omitted: %s', keyPathID.getFullName(), !allowEmptyArrays)
```

shir-reifenberg: thanks 🙏 
opened - https://github.com/salto-io/salto/pull/6075

PR #6067, opened 2024-06-16
SALTO-6106: Add optional allowEmptyArrays option to top level element and enable it in Okta Application

shir-reifenberg:

---

** **please only review last commit** **

Still TODO - 

- [x] test in Zendesk to verify no side effects
- [x] uts
- [x] noise reduction in Okta

---
_Release Notes_: 

_Okta_adapter_:
- Fix a bug in deployment of OIDC application causing applications with empty scopes to fail

---
_User Notifications_: 

_Okta_adapter_:
- Empty arrays in Okta `Application` will be visible in `Application` elements
-


netama (approved):
🎉

PR #6069, opened 2024-06-18
SALTO-5383: (Jira) prevent deploying portal groups with an non-unique name field

SeggevHaimovich:
Some types have fields that should be unique in the cross-environment.
Created a CV in the adapter-components to verify uniqueness of a specific field in a specific type.
Used it in Jira with the types: issueType, portalGroups

---

---
_Release Notes_: 
Jira-adapter:
* prevent deploying portal groups with an non-unique name field

---
_User Notifications_: 
_None_


IdoZakk:
Minor typo in the release notes comment- use `a non-unique`

packages/adapter-components/src/deployment/change_validators/unique_fields.ts
```diff
@@ -0,0 +1,116 @@
...
+): (ModificationChange<InstanceElement> | AdditionChange<InstanceElement>)[] =>
+  changes
+    .filter(isAdditionOrModificationChange)
+    .filter(change => getChangeData(change).elemID.typeName in typeToFieldRecord)
+    .filter(isInstanceChange)
+
+const getFieldValue = (instance: InstanceElement, fieldName: string): string | undefined => {
+  const fieldValue = resolvePath(instance, instance.elemID.createNestedID(...fieldName.split('.')))
+  if (!_.isString(fieldValue)) {
+    log.error('Unique field value is not a string')
```
IdoZakk, line 68: I do not think we need an error here. The user can set the value as a number, which might work with the service. I think a warning can be enough

packages/adapter-components/src/deployment/change_validators/unique_fields.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+  return getSameUniqueFieldError(change, otherInstance, fieldName)
+}
+
+const getErrorsForType = (
+  typeName: string,
+  instances: InstanceElement[],
+  changes: (ModificationChange<InstanceElement> | AdditionChange<InstanceElement>)[],
+  typeToFieldRecord: Record<string, string>,
+): ChangeError[] => {
```
IdoZakk, line 97: With this amount of inputs variables, it is better to define it as an object

packages/adapter-components/src/deployment/change_validators/unique_fields.ts
```diff
@@ -0,0 +1,116 @@
...
+  )
+  return changesOfType
+    .map(change => getErrorForChange(change, typeToFieldRecord[typeName], fieldValueToInstancesRecord))
+    .filter(isDefined)
+}
+
+export const uniqueFieldsChangeValidatorCreator =
+  (typeToFieldRecord: Record<string, string>): ChangeValidator =>
+  async (changes, elementSource) => {
+    const relevantChanges = getRelevantChanges(changes, typeToFieldRecord)
```
IdoZakk, line 112: You can create a function that is a bit more complicated but will return a map of relevant changes keyed by their type. It will save you the need to get the types and filter per type later on

packages/adapter-components/src/deployment/change_validators/unique_fields.ts
```diff
@@ -0,0 +1,116 @@
...
+  fieldName: string,
+  fieldValueToInstancesRecord: Record<string, InstanceElement[]>,
+): ChangeError | undefined => {
+  const fieldValue = getFieldValue(getChangeData(change), fieldName)
+  if (fieldValue === undefined) {
+    return undefined
+  }
+  const otherInstance = fieldValueToInstancesRecord[fieldValue].find(
+    instance => !instance.elemID.isEqual(getChangeData(change).elemID),
+  )
```
IdoZakk, line 85: You can unite those lines, by adding ? before the `.find`. If you try to get a value of undefined you will get undefined back

IdoZakk (commented):
👍 
Very nice!
See my comments

packages/adapter-components/src/deployment/change_validators/unique_fields.ts
```diff
@@ -0,0 +1,116 @@
...
+    .map(change => getErrorForChange(change, typeToFieldRecord[typeName], fieldValueToInstancesRecord))
+    .filter(isDefined)
+}
+
+export const uniqueFieldsChangeValidatorCreator =
+  (typeToFieldRecord: Record<string, string>): ChangeValidator =>
+  async (changes, elementSource) => {
+    const relevantChanges = getRelevantChanges(changes, typeToFieldRecord)
+    const relevantTypes = _.uniq(relevantChanges.map(change => getChangeData(change).elemID.typeName))
+    const instances = await getInstances(elementSource, relevantTypes)
```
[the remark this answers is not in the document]
SeggevHaimovich: In the getInstances func I return an empty list if the typeNames is with length 0.
If there is no changes the typeNames list is empty

packages/adapter-components/test/deployment/change_validators/unique_fields.test.ts
```diff
@@ -0,0 +1,184 @@
...
+    expect(deletionErrors).toHaveLength(0)
+  })
+  it('should not return error for elements with different types and the same value in the unique field', async () => {
+    elementSource = buildElementsSourceFromElements([relevantInstance1, otherRelevantInstance1])
+    const deletionErrors = await changeValidator([toChange({ after: otherRelevantInstance1 })], elementSource)
+    expect(deletionErrors).toHaveLength(0)
+  })
+  it('should not return an error if the field is undefined or is not a string', async () => {
+    const instance1 = relevantInstance1.clone()
+    instance1.value.uniqueField = undefined
```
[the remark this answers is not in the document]
SeggevHaimovich: I use clone only in 2 tests, I thought it would be inefficient to clone in the beforeEach for all the tests because of that.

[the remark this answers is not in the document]
IdoZakk: No need to clone in the before all, just to set the relevant values. 
You declare the elements in the general scope, and create them in the beforeEach

packages/adapter-components/src/deployment/change_validators/unique_fields.ts
SeggevHaimovich: I think I shouldn't do this because record can have an `undefined` key and it could be risky.
image
WDYT?

packages/adapter-components/test/deployment/change_validators/unique_fields.test.ts
```diff
@@ -0,0 +1,212 @@
...
+    })
+  })
+
+  it('should not return error for removal changes', async () => {
+    const elementSource = buildElementsSourceFromElements([])
+    const changeErrors = await changeValidator(
+      [toChange({ before: relevantInstance1 }), toChange({ before: relevantInstance2 })],
+      elementSource,
+    )
+    expect(changeErrors).toHaveLength(0)
```
IdoZakk, line 88: I think it is clearer and safer to test here against the basic scenario of an existing instance (in the elements source)

packages/adapter-components/test/deployment/change_validators/unique_fields.test.ts
```diff
@@ -0,0 +1,212 @@
...
+        toChange({ after: relevantInstance1 }),
+        toChange({ after: irrelevantInstance1 }),
+        toChange({ after: otherRelevantInstance1 }),
+      ],
+      elementSource,
+    )
+    expect(changeErrors).toHaveLength(0)
+  })
+  it('should not return an error if the field is undefined or is not a string', async () => {
+    const instance1 = relevantInstance1.clone()
```
IdoZakk, line 117: Instead of clone you can override the value of relevantInstance1 and relevantInstance2

packages/adapter-components/test/deployment/change_validators/unique_fields.test.ts
```diff
@@ -0,0 +1,212 @@
...
+      ],
+      elementSource,
+    )
+    expect(changeErrors).toHaveLength(2)
+    expect(changeErrors[0]).toEqual({
+      elemID: instance1After.elemID,
+      severity: 'Error',
+      message: "The field 'uniqueField' in type relevantType must have a unique value",
+      detailedMessage:
+        "This relevantType have the same 'uniqueField' as the instance adapter.relevantType.instance.relevantInstance2, and can not be deployed.",
```
IdoZakk, line 163: While true it is a bit confusing to read, as the "other" instance is instance2After, but the name indicates relevantInstance2.

IdoZakk (commented):
Great 🍏 
Some small comments.
Also, something general about the change validator- 
The practice of taking code into a function in complex logic really helps readability. When done for single use straight forward code segments it makes the code harder to read and longer. With 9 internal functions it was a bit hard to follow

packages/adapter-components/test/deployment/change_validators/unique_fields.test.ts
SeggevHaimovich: I changed the names to: `relevantInstance1After` and `relevantInstance2After`.
Is this clearer?

packages/adapter-components/src/deployment/change_validators/unique_fields.ts
```diff
@@ -0,0 +1,125 @@
...
+    .filter(isInstanceChange)
+  return _.groupBy(relevantChanges, change => getChangeData(change).elemID.typeName)
+}
+
+const createTypeToInstancesRecord = async (
+  elementsSource: ReadOnlyElementsSource | undefined,
+  relevantTypes: string[],
+): Promise<Record<string, InstanceElement[]>> => {
+  const instances =
+    elementsSource === undefined || relevantTypes.length === 0
```
IdoZakk, line 61: If elementsSource is undefined there is no point in running the validator, and also it will be good to log that the validator did not run.
Instead of doing the test at each type and allowing undefined let's check it at the start of the filter and also log if undefined

IdoZakk (approved):
Very good, one small comment, make sure to fix before merge

PR #6071, opened 2024-06-18
SALTO-6111 Add logs in fetch resources for debugging

omrifarkash:
---
Adding log time to fetch resources flow and generate elements

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/fetch/resource/resource_manager.ts
+++ b/packages/adapter-components/src/fetch/resource/resource_manager.ts
@@ -67,8 +67,8 @@ export const createResourceManager = <ClientOptions extends string>({
   initialRequestContext?: Record<string, unknown>
   customItemFilter?: (item: ValueGeneratedItem) => boolean
 }): ResourceManager => ({
-  fetch: log.timeDebug(
-    () => async query => {
+  fetch: query => log.timeDebug(
+    async () => {
       const createTypeFetcher: TypeFetcherCreator = ({ typeName, context }) =>
         createTypeResourceFetcher({
           adapterName,
@@ -87,7 +87,7 @@ export const createResourceManager = <ClientOptions extends string>({
       )
       const graph = createDependencyGraph(resourceDefQuery.getAll())
       try {
-        await graph.walkAsync(async typeName => {
+        await graph.walkAsync(typeName => log.timeDebug(async () => {
           const resourceFetcher = resourceFetchers[typeName]
           if (resourceFetcher === undefined) {
             log.debug('no resource fetcher defined for type %s:%s', adapterName, typeName)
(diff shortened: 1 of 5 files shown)


packages/adapter-components/src/fetch/resource/resource_manager.ts
```diff
@@ -97,22 +97,24 @@ export const createResourceManager = <ClientOptions extends string>({
...
           }
-
+          const items = resourceFetcher.getItems()
+          log.info('pushing %d resources for %s:%s', items?.length, adapterName, typeNameAsStr)
           elementGenerator.pushEntries({
             typeName: typeNameAsStr,
-            entries: resourceFetcher.getItems()?.map(item => item.value) ?? [],
+            entries: items?.map(item => item.value) ?? [],
           })
+          log.info('done walk on graph for %s:%s', adapterName, typeNameAsStr)
```
omrifarkash, line 117:
```suggestion
          log.info('done pushing resources for type %s:%s', adapterName, typeNameAsStr)
```

packages/adapter-components/src/fetch/resource/resource_manager.ts
```diff
@@ -67,8 +67,8 @@ export const createResourceManager = <ClientOptions extends string>({
   initialRequestContext?: Record<string, unknown>
   customItemFilter?: (item: ValueGeneratedItem) => boolean
 }): ResourceManager => ({
-  fetch: log.timeDebug(
-    () => async query => {
+  fetch: query => log.timeDebug(
+    async () => {
```
ori-moisis, line 71: note - before this change the `timeDebug` would count the time it took to define the function instead of the time it took the function to run (we noticed in the logs the time it took was always 1-2 ms)

packages/adapter-components/src/fetch/resource/resource_manager.ts
```diff
@@ -67,8 +67,8 @@ export const createResourceManager = <ClientOptions extends string>({
   initialRequestContext?: Record<string, unknown>
   customItemFilter?: (item: ValueGeneratedItem) => boolean
 }): ResourceManager => ({
-  fetch: log.timeDebug(
-    () => async query => {
+  fetch: query => log.timeDebug(
```
omrifarkash, line 70: The log time here was used wrongly. It measured the time to define a function and not the time to execute it

packages/adapter-components/src/fetch/resource/resource_manager.ts
```diff
@@ -87,7 +87,7 @@ export const createResourceManager = <ClientOptions extends string>({
       )
       const graph = createDependencyGraph(resourceDefQuery.getAll())
       try {
-        await graph.walkAsync(async typeName => {
+        await graph.walkAsync(typeName => log.timeDebug(async () => {
```
shir-reifenberg: nit: maybe `log.timeTrace` if it's per type ?

shir-reifenberg (approved):
LGTM with 1 nit 🎉

packages/adapter-components/src/fetch/resource/resource_manager.ts
omrifarkash: as all the others log.time we added in this PR, its per type, not per instance. shouldn't happen more than 10-100 time for a fetch

PR #6079, opened 2024-06-18
SALTO-6133: Fix issue with client config type in Okta

shir-reifenberg:
We accidentally passed `usePrivateAPI` to the wrong place in the config object type 

---

will add test soon 

---
_Release Notes_: 

_Okta_adapter_:
- Fix a bug in fetch causing all fetch to include invalid config event

---
_User Notifications_: 


omrilit (commented):
Looks good!
@netama I assume that you'd like to give it a look as well?

netama (approved):
thanks!

PR #4451, opened 2023-06-01
SALTO-4278 Add parser support for unicode newline

adfineberg:
_Add support for unicode line terminator parsing_

---

_Additional context for reviewer_
* Javascript has 2 unicode line terminators (\u2028 and \u2029)

---
_Release Notes_: 
_Core_
* Unicode line terminators will no longer throw parse errors.

---
_User Notifications_:

PR #6080, opened 2024-06-18
SALTO-6116 Ignore conflicts caused by different line separators

dantal4:
We should be able to auto merge strings when there are different line separators between the base/current/incoming versions. The easiest solution would be to take the line separator of the `current` version (if there are mixed line separators in current - take the most common one*) and use it in the merged version**.

*we’ll lose the other used separators in current, but I guess that it’s fine.

**only in conflicts auto-merge - other fetched files will be kept as fetched even they have mixed line separators.

---

TODO:
- [x] add tests

---
_Release Notes_: 
Core:
- Ignore conflicts caused by different line separators

---
_User Notifications_: 
None

### DIFF
--- a/packages/core/src/core/merge_content.ts
+++ b/packages/core/src/core/merge_content.ts
@@ -41,8 +41,9 @@ const splitToLinesWithTerminators = (multilineString: string): LineWithTerminato
       linesWithTerminators.push(Object.assign(currentLine, { terminator: '\r\n' }))
       currentLine = ''
       i += 1 // Skip the \n part
-    } else if (char === '\r' || char === '\n' || char === '\u2028' || char === '\u2029') {
-      // Other terminators: \r, \n, \u2028, \u2029
+    } else if (char === '\r' || char === '\n') {
+      // Other terminators: \r, \n. In general, \u2028 & \u2029 are considered line terminators too,
+      // but Monaco Editor doesn't treat them as so (shows an unknown char), so we can ignore them too.
       linesWithTerminators.push(Object.assign(currentLine, { terminator: char }))
       currentLine = ''
     } else {
--- a/packages/core/src/core/merge_content.ts
+++ b/packages/core/src/core/merge_content.ts
@@ -25,13 +25,26 @@ const { humanFileSize } = strings
 const MAX_MERGE_CONTENT_SIZE = 10 * 1024 * 1024
 const MERGE_TIMEOUT = 10 * 1000
 
+const getLineSeparator = (str: string): string => {
+  const numOfCrlf = str.match(/\r\n/g)?.length ?? 0
+  const numOfLf = (str.match(/\n/g)?.length ?? 0) - numOfCrlf
+  const numOfCr = (str.match(/\r/g)?.length ?? 0) - numOfCrlf
+  if (numOfCrlf > numOfCr && numOfCrlf > numOfLf) {
+    return '\r\n'
+  }
+  if (numOfCr > numOfLf) {
+    return '\r'
+  }
+  return '\n'
+}
+
 const isConflictChunk = (chunk: diff3.ICommResult<string>): boolean =>
   !('common' in chunk) && chunk.buffer1.length > 0 && chunk.buffer2.length > 0
 
 const mergeTwoStrings = (
   first: string,
   second: string,
-  { stringSeparator }: { stringSeparator: string },
+  { stringSeparator }: { stringSeparator: RegExp },
   timeout: number,
 ): { conflict: boolean; result: string[] } => {
   const result = diff3.diffComm(first.split(stringSeparator), second.split(stringSeparator), timeout)
@@ -71,7 +84,7 @@ export const mergeStrings = (
         )
         return undefined
       }
-      const options = { excludeFalseConflicts: true, stringSeparator: '\n' }
+      const options = { excludeFalseConflicts: true, stringSeparator: /\r\n|\n|\r/g }
       try {
         const { conflict, result } =
           base !== undefined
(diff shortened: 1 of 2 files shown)


packages/core/src/core/merge_content.ts
```diff
@@ -71,7 +84,7 @@ export const mergeStrings = (
         )
         return undefined
       }
-      const options = { excludeFalseConflicts: true, stringSeparator: '\n' }
+      const options = { excludeFalseConflicts: true, stringSeparator: /\r\n|\n|\r/g }
```
ori-moisis, line 87: 1. should a single `\r` really be considered a line break?
2. don't we also need to consider unicode line breaks? (#4451)

packages/core/src/core/merge_content.ts
```diff
@@ -80,8 +93,11 @@ export const mergeStrings = (
         if (conflict) {
           log.debug('conflict found in %s', changeId)
         } else {
-          log.debug('merged %s successfully', changeId)
-          return result.join(options.stringSeparator)
+          // getting the line separator of the current version in the workspace
+          const lineSeparator = getLineSeparator(current)
+          // eslint-disable-next-line no-restricted-syntax
+          log.debug('merged %s successfully- using line separator %s', changeId, JSON.stringify(lineSeparator))
+          return result.join(lineSeparator)
```
ori-moisis, line 100: can't we preserve the original line breaks per chunk?

IIUC each chunk comes from one of the the sources, so in terms of pure information, we do know which line ending should be after each chunk, no?

dantal4: I think that we can do it, although it won't be easy because `result` is just `string[]` and there's no info about the origin of each line.

packages/core/src/core/merge_content.ts
dantal4: 1. we will probably never see it. I have no strong opinion about leaving it or not.
2. I'll add them too.

packages/core/src/core/merge_content.ts
ori-moisis: if we can provide a custom compare function to diff3 maybe we could keep the line endings in the strings and use our special string compare function? (so here we will do `result.join('')` because the line endings will already be here)

dantal4: I found a nice solution, by supplying diff3 a list of `string & { terminator: string }`. diff3 compares the lines as strings (without the terminator), and at the end we can restore the terminator of the lines, because diff3 merge the lines but doesn't change them.

packages/core/src/core/merge_content.ts
```diff
@@ -25,16 +25,50 @@ const { humanFileSize } = strings
...
 const MERGE_TIMEOUT = 10 * 1000
 
-const isConflictChunk = (chunk: diff3.ICommResult<string>): boolean =>
+type LineWithTerminator = string & { terminator: string }
+
+const splitToLinesWithTerminators = (multilineString: string): LineWithTerminator[] => {
+  const linesWithTerminators: LineWithTerminator[] = []
+  let currentLine = ''
+
+  for (let i = 0; i < multilineString.length; i += 1) {
```
ori-moisis, line 34: according to the comment [here](https://github.com/salto-io/salto/blob/main/packages/adapter-api/src/comparison.ts/#L149-L157) this might be noticeably slow.

I am not 100% sure what kind of benchmark we need to pass here in order to say it is ok, but I think it might be worth at least testing the performance of this with a few large strings and seeing if this makes a significant difference compared to the time it took before this change

dantal4: I tried it on several files - from 160KB to 5MB - the added time of `splitToLinesWithTerminators` & `joinLinesWithTerminators` is much less than the total time of `mergeStrings`.

ori-moisis: can you share the numbers please?

packages/core/src/core/merge_content.ts
```diff
@@ -25,16 +25,50 @@ const { humanFileSize } = strings
...
+  // Add the last line if there is no terminator at the end
+  if (currentLine !== '') {
+    linesWithTerminators.push(Object.assign(currentLine, { terminator: '' }))
+  }
+
+  return linesWithTerminators
+}
+
+const joinLinesWithTerminators = (lines: LineWithTerminator[]): string =>
+  lines.reduce((res, line) => res.concat(line, line.terminator ?? '\n'), '')
```
ori-moisis, line 61: just making sure, this fallback should never actually happen, right?

packages/core/src/core/merge_content.ts
dantal4: 160KB file (a JS file with ~6000 lines):
- splitToLinesWithTerminators took 7/9/28 ms (on the different versions)
- joinLinesWithTerminators took 2 ms
- mergeStrings took 8523 ms

5MB file (just repeating `abcd\n`):
- splitToLinesWithTerminators took 365/611/331 ms (on the different versions)
- mergeStrings reached timeout of 10 seconds (it makes sense on a repetitive file)

ori-moisis (commented):
code LGTM overall, just missing a bunch of tests

packages/core/src/core/merge_content.ts
dantal4: right

packages/core/src/core/merge_content.ts
ori-moisis: what are the different versions?

dantal4: base, current, incoming

ori-moisis: any intuition as to why one of them takes significantly longer than the others?

dantal4: no.... the main difference is that in `current` I used `\n` and in base&incoming (that both came from the service) I used `\r\n`

ori-moisis (approved):
code LGTM
I think the tests need to be improved and be more explicit in what is the scenario under test and what is the expected result.
Also not sure all the interesting scenarios are actually covered (2/3 way diff, mix of newline characters, merging chunks from 3 way diff where each chunk has a different newline, etc...)

PR #6081, opened 2024-06-18
Improve FileCabinet files query by excluding excluded file extensions

dantal4:
It happens quite often that we're fetching a large amount of files, then we're excluding most of them because of their extensions. The query can be improved by adding it a regex of the excluded file extensions from the user's adapter config.

---

---
_Release Notes_: 
Netsuite Adapter:
- Improve FileCabinet files query by excluding excluded file extensions

---
_User Notifications_: 
None

### DIFF
--- a/packages/netsuite-adapter/test/client/suiteapp_client/suiteapp_file_cabinet.test.ts
+++ b/packages/netsuite-adapter/test/client/suiteapp_client/suiteapp_file_cabinet.test.ts
@@ -286,7 +286,7 @@ describe('suiteapp_file_cabinet', () => {
 
   describe('importFileCabinet', () => {
     const maxFileCabinetSizeInGB = 1
-    const extensionsToExclude = ['.*\\csv']
+    const extensionsToExclude = ['.*\\.csv']
     it('should return all the files', async () => {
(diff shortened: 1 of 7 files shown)


packages/netsuite-adapter/test/client/suiteapp_client/suiteapp_file_cabinet.test.ts
```diff
@@ -286,9 +286,14 @@ describe('suiteapp_file_cabinet', () => {
 
   describe('importFileCabinet', () => {
     const maxFileCabinetSizeInGB = 1
+    const extensionsToExclude = ['.*\\csv']
```
omrilit, line 289: Shouldn't it be the below?
Maybe we are missing another UT checking it?
```suggestion
    const extensionsToExclude = ['.*\\.csv']
```

omrilit (approved):
Looks great!
My single comment is only about the UT

packages/netsuite-adapter/test/client/suiteapp_client/suiteapp_file_cabinet.test.ts
dantal4: You're right. I'll fix it and add a test to `adapter.test.ts` where we get the list of excluded extensions.

PR #6085, opened 2024-06-19
SALTO-6123: fix workflow condition type guard

shayc331:
_fix workflow condition type guard_

---

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


IdoZakk (approved):
🐰

PR #6086, opened 2024-06-19
SALTO-6124: Remove noisy error logs on undefined link property

omrilit:

---

---
_Release Notes_: 

---
_User Notifications_: 


shir-reifenberg (approved):
Thanks for that!
when links property is missing I think it's an actual error cause we use it to create references from `Application` type to `AccessPolicy` and `ProfileEnrollment` policy

I _think_ that what we saw in the logs is the for okta classic account, where I _think_ we don't get some of the policy types (didn't actually validated this)
we should probably only call `extractIdsFromUrls` or log this error when the account is not a classic engine

but this is a bigger change then feel free to merge it and we can address the issue some other time (we should at least log.trace in this scenario for debugging)

PR #6087, opened 2024-06-19
SALTO-5520: Fixed a bug in issue layouts

IdoZakk:
There was a bug in the code and in the tests that blocked Node 18 upgrade

---

For some legacy reasons, the test did not properly check a promise. The result was a transient test that failed depending on whether the promise was fulfilled or not.
The bug in the code caused us to send a wrong call in case there are no screens in a relevant screen scheme

---
_Release Notes_: 
Jira Adapter:
* A small fix in the issue layouts for Node 18

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/layouts/issue_layout.ts
+++ b/packages/jira-adapter/src/filters/layouts/issue_layout.ts
@@ -136,7 +136,8 @@ const getProjectToScreenMappingUnresolved = (elements: Element[]): Record<string
               .map(
                 (issueTypeMapping: issueTypeMappingStruct) =>
                   screensSchemesToDefaultOrViewScreens[issueTypeMapping.screenSchemeId],
-              ),
+              )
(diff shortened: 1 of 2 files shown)


packages/jira-adapter/src/filters/layouts/issue_layout.ts
```diff
@@ -136,7 +136,8 @@ const getProjectToScreenMappingUnresolved = (elements: Element[]): Record<string
               .map(
                 (issueTypeMapping: issueTypeMappingStruct) =>
                   screensSchemesToDefaultOrViewScreens[issueTypeMapping.screenSchemeId],
-              ),
+              )
+              .filter(isDefined),
```
izibr, line 140: This should not occur in a valid environment
There should always be a default screen or a view screen
In that case it might be worth adding a log because it is something that is not valid in the environment and that way we can monitor it

PR #6092, opened 2024-06-20
SALTO-5692: Okta: Upgrade `Brand` deploy definitions to new infra

Nurdok:
Other than adding `Brand` support, I also transitioned to using the standard definition helper.

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/okta-adapter/src/deployment.ts
```diff
@@ -194,6 +194,9 @@ export const defaultDeployChange = async (
     }
   }
 
+  if (apiDefinitions.types[getChangeData(change).elemID.typeName] === undefined) {
+    throw new Error(`No deployment configuration found for type ${getChangeData(change).elemID.typeName}`)
+  }
```
shir-reifenberg, line 199: Isn't `deployChange` in adapter components already verifies there's endpoint for deploy?  [here](https://github.com/salto-io/salto/blob/main/packages/adapter-components/src/deployment/deployment_deprecated.ts#L72)

shir-reifenberg (approved):
NICE! 🎉

packages/okta-adapter/src/deployment.ts
Nurdok: This is the old infra. When developing, the next line (destructuring `deployRequests`) would trigger an exception, but because it gets translated to `OktaError` it's very hard to find what's wrong / the exact line that's failing. This just makes migrating from old to new infra a bit easier.

PR #6094, opened 2024-06-20
SALTO-6100 - Salesforce: Fallback to element source if referred element was not fetched in QuickFetch

tomsellek:
In changed-based fetch, if we fetch a record that contains a reference, and the referred record was not fetched (because it was not modified), we need to look for the referred record in the workspace. 

---

Test:
 - Run the test described in the ticket. Observe no issues with missing references.

---
_Release Notes_: 
Salesforce: will no longer show an error when using changed-based fetch to fetch records that refer to unchanged records.

---
_User Notifications_: 
N/A

### DIFF
--- a/packages/salesforce-adapter/src/filters/custom_objects_instances.ts
+++ b/packages/salesforce-adapter/src/filters/custom_objects_instances.ts
@@ -421,10 +421,61 @@ const getTargetRecordIds = (
   )
 }
 
+type SalesforceRecordsByType = Record<string, SalesforceRecord>
+type SalesforceRecordsByTypeAndId = Record<string, SalesforceRecordsByType>
+
+const getElementSourceRecordsByTypeAndId = async (
+  elementsSource: ReadOnlyElementsSource,
+  types: Set<string>,
+  ignoredRecordIds: Set<string>,
+): Promise<SalesforceRecordsByTypeAndId> => {
+  const instancesToRecords = async (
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/src/filters/custom_objects_instances.ts
```diff
@@ -474,7 +525,24 @@ const getReferencedRecords = async (
           customObjectFetchSettings: fetchSettings,
           ids,
         })
-        return _.keyBy(records, (record) => record[CUSTOM_OBJECT_ID_FIELD])
+        const fetchedRecordsById = _.keyBy(
+          records,
+          (record) => record[CUSTOM_OBJECT_ID_FIELD],
+        )
+        const missingIds = ids.filter((id) => !_.has(fetchedRecordsById, id))
```
yelly, line 532: Why not `!fetchedRecordsById.hasOwnProperty(id)`?

yelly (approved):
LGTM, maybe @tamtamirr should also take a look though.

packages/salesforce-adapter/src/filters/custom_objects_instances.ts
tomsellek: Because `hasOwnProperty` has to be used in a somewhat awkward way, like `Object.prototype.hasOwnProperty.call(instance, property)`.

PR #6095, opened 2024-06-20
SALTO-5884: Microsoft Entra: Deploy improvements

shoshana-houri:
Fix several issues with deployment:
1. Group type name:
   1. Some fields on the group type are read-only on *addition* as well, and cannot be specified in the addition request. If they are specified the request fails with error. We've added them to the `readOnlyFields` validator, and make sure to omit them in the request.
   2. Some fields cannot be specified in the addition request, but only in a subsequent modification request.
   3. We needed to add polling for the assignment of `lifeCyclePolicy` when creating a group, since the returned group id takes some time to be recognized by the service in the context of the life cycle policy request.
2. Some references were missing - which caused deploy requests to fail since the referred object didn't exist in the environment, and we had no indication of unresolved reference in Salto...

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/microsoft-entra-adapter/src/change_validators/read_only_fields_validator.ts
```diff
@@ -59,15 +109,22 @@ export const TYPE_NAME_TO_READ_ONLY_FIELDS: Record<string, string[]> = {
...
+    .filter(isAdditionOrModificationChange)
     .filter(isInstanceChange)
     .filter(change => relevantTypes.includes(change.data.after.elemID.typeName))
 
-  return changesWithReadOnlyFields
-    .map((change): ChangeError | undefined => {
-      const readOnlyFields = TYPE_NAME_TO_READ_ONLY_FIELDS[change.data.after.elemID.typeName]
-      const modifiedFields = readOnlyFields.filter(
-        fieldName => !_.isEqual(_.get(change.data.before.value, fieldName), _.get(change.data.after.value, fieldName)),
+  const lookupFunc = referenceUtils.generateLookupFunc(REFERENCES?.rules ?? [])
```
shir-reifenberg, line 116: I'm not sure resolving references in validator is the correct thing here... for example, if we add a instance of a type that doesn't exist in the target env, the current type of the instance will not exist and the references will not be resolved correctly.. (in the adapter's deploy we handle this using `overrideInstanceTypeForDeploy`)

packages/microsoft-entra-adapter/src/change_validators/read_only_fields_validator.ts
```diff
@@ -59,15 +109,22 @@ export const TYPE_NAME_TO_READ_ONLY_FIELDS: Record<string, string[]> = {
...
+  return awu(changesWithReadOnlyFields)
+    .map(async (change): Promise<ChangeError | undefined> => {
+      const readOnlyFields =
+        change.action === 'add'
+          ? TYPE_NAME_TO_READ_ONLY_FIELDS_ADDITION[change.data.after.elemID.typeName]
+          : TYPE_NAME_TO_READ_ONLY_FIELDS_MODIFICATION[change.data.after.elemID.typeName]
+      const before = isModificationChange(change) ? await resolveValues(change.data.before, lookupFunc) : { value: {} }
+      const after = await resolveValues(change.data.after, lookupFunc)
+      const modifiedFields = readOnlyFields?.filter(
+        fieldName => !_.isEqual(_.get(before.value, fieldName), _.get(after.value, fieldName)),
```
shir-reifenberg, line 127: I think you can use `isEqualValues` which will also compare references and you can avoid resolving them

packages/microsoft-entra-adapter/src/definitions/deploy/utils/condition_empty_fields.ts
```diff
@@ -0,0 +1,31 @@
...
+/*
+ * Creates a custom condition that returns true if at least one of the specified fields is not empty
+ * This custom condition is used to filter out changes that are not relevant for *addition* deploy requests,
+ * since the transformForCheck field is irrelevant for addition changes
+ */
+export const createCustomConditionEmptyFields = (fieldNames: string[]): definitions.deploy.DeployRequestCondition => ({
+  custom:
+    () =>
+    ({ change }) =>
+      _.some(fieldNames, fieldName => !_.isEmpty(_.get(getChangeData(change).value, fieldName))),
```
shir-reifenberg, line 30: if it's only for addition change can we please add `isAdditionChange` and add an indication in the function name?

packages/microsoft-entra-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -215,15 +226,31 @@ const graphV1CustomDefinitions: DeployCustomDefinitions = {
...
+            request: {
+              endpoint: {
+                path: '/groups/{id}',
+                method: 'patch',
+              },
+              transformation: {
+                pick: GROUP_UNDEPLOYABLE_FIELDS_ON_ADDITION,
               },
             },
+            condition: createCustomConditionEmptyFields(GROUP_UNDEPLOYABLE_FIELDS_ON_ADDITION),
```
shir-reifenberg, line 249: very cool :)

packages/microsoft-entra-adapter/src/definitions/deploy/utils/read_only_fields.ts
```diff
@@ -16,13 +16,23 @@
...
+  TYPE_NAME_TO_READ_ONLY_FIELDS_ADDITION,
+  TYPE_NAME_TO_READ_ONLY_FIELDS_MODIFICATION,
+} from '../../../change_validators'
 import { AdjustFunction } from '../types'
 
 export const omitReadOnlyFields: AdjustFunction = ({ typeName, value, context }) => {
   validatePlainObject(value, typeName)
-  const readOnlyFieldsToOmit = TYPE_NAME_TO_READ_ONLY_FIELDS[typeName]
-  if (readOnlyFieldsToOmit === undefined || context.action !== 'modify') {
+  if (context.action === 'remove') {
```
shir-reifenberg, line 27: shouldn't this be `context.change.action` ?

shir-reifenberg (commented):
Looks good! 😎 
Added some comment (one that I believe might be buggy in some cases)

packages/microsoft-entra-adapter/src/definitions/requests/clients.ts
```diff
@@ -36,6 +36,20 @@ export const createClientDefinitions = (
             omitBody: true,
           },
         },
+        customizations: {
+          '/v1.0/groupLifecyclePolicies/{lifeCyclePolicyId}/addGroup': {
+            post: {
+              // After creating a group it takes a while for the group to be available for assigning it to a lifecycle policy
+              // See a similar issue in https://stackoverflow.com/questions/47303158/add-group-member-fails-with-404-error
+              polling: {
+                interval: 6000,
```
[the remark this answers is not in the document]
shoshana-houri: Yes, when considering intervals of 1000 jumps.

packages/microsoft-entra-adapter/src/definitions/deploy/utils/condition_empty_fields.ts
```diff
@@ -23,9 +23,15 @@ import _ from 'lodash'
...
+export const createCustomConditionEmptyFieldsOnAddition = (
+  fieldNames: string[],
+): definitions.deploy.DeployRequestCondition => ({
   custom:
     () =>
-    ({ change }) =>
-      _.some(fieldNames, fieldName => !_.isEmpty(_.get(getChangeData(change).value, fieldName))),
+    ({ change }) => {
+      if (!isAdditionChange(change)) {
+        return true
```
shir-reifenberg, line 33: I thought we said the opposite? (make the request only if it addition change && _.some(...) )

packages/microsoft-entra-adapter/src/definitions/deploy/utils/read_only_fields.ts
```diff
@@ -24,7 +24,7 @@ import { AdjustFunction } from '../types'
 
 export const omitReadOnlyFields: AdjustFunction = ({ typeName, value, context }) => {
   validatePlainObject(value, typeName)
-  if (context.action === 'remove') {
+  if (context.change.action === 'remove') {
```
shir-reifenberg, line 27: I think also in line 32 you check you `context.action`

shir-reifenberg (approved):
Nice 🎈

packages/microsoft-entra-adapter/src/definitions/deploy/utils/condition_empty_fields.ts
shoshana-houri: Oh, I thought that we'll behave as though there's no condition (which means to always perform this request 😅 ) I don't mind changing it though

PR #6100, opened 2024-06-23
SALTO-5472 query referenced ASV only

dantal4:
Instead of fetching all SuiteQL tables data, we want to fetch only the referenced internalIds in workflows and data instances.

---

~~This PR is based on #6102 . Please review it first.~~

TODO:
- [x] add tests

---
_Release Notes_: 
Netsuite Adapter:
- Query referenced ASV only

---
_User Notifications_: 
None


omrilit (approved):
Very very impressive! 👏 
It seems like there are only 2 customers with resolveAccountSpecificValues set to true explicitly ([logs](https://app.datadoghq.com/logs?query=%22resolveAccountSpecificValues%3A%20true%22%20user%20netsuite%20%40saltoStage%3Aproduction%20%40isExternalUser%3Atrue%20&agg_m=count&agg_m_source=base&agg_t=count&cols=%40workspaceId%2C%40orgName%2C%40envName&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1709930186988&to_ts=1710102986988&live=true)).
What is our rollout plan?
I assume that the first frop will affect only those 2 orgs, right?

PR #6102, opened 2024-06-23
Change NS changeValidators params

dantal4:
change NS changeValidators params to be an object instead of params list.

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


omrilit (approved):
LGTM!

PR #6104, opened 2024-06-23
SALTO-6137: Log users count in Okta

shir-reifenberg:
Add api call to get the number of users

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


omrilit (approved):
Thanks!

PR #6105, opened 2024-06-24
SALTO-6140: Recalculate custom references on reference indexes version mismatch

tamtamirr:
Recalculate custom references on reference indexes version mismatch

---

We should invoke `getCustomReferences` with all the changes from the Elements Source on outdated indexes.

---
_Release Notes_: 
Core:
- Fixed a bug where custom references did not recalculate on outdated reference indexes.

---
_User Notifications_: 
_None_


packages/workspace/test/workspace/reference_indexes.test.ts
```diff
@@ -525,9 +529,17 @@ describe('updateReferenceIndexes', () => {
           mapVersions,
           elementsSource,
           true,
-          async () => [],
+          async elements => {
+            elements.forEach(e => inputElements.push(e))
+            return []
+          },
```
ori-moisis, line 535: not sure I fully understand what you tried to do here, but if what you tried to do is check that the references function was called with a specific argument, it should be done with a jest mock function and `expect(...).toHaveBeenCalledWith(...)`

packages/workspace/test/workspace/reference_indexes.test.ts
```diff
@@ -515,8 +516,10 @@ describe('updateReferenceIndexes', () => {
     })
 
     describe('When indexes are out of date', () => {
+      let mockedGetCustomReferences: jest.MockedFunction<ReferenceIndexesGetCustomReferencesFunc>
       beforeEach(async () => {
         await elementsSource.set(instance)
+        mockedGetCustomReferences = jest.fn().mockResolvedValue([])
```
ori-moisis, line 522: should use typed mock functions
```suggestion
        mockedGetCustomReferences = jest.fn().mockResolvedValue([])
```
```suggestion
        mockedGetCustomReferences = mockFunction<typeof mockedGetCustomReferences>().mockResolvedValue([])
```

ori-moisis (approved):

packages/workspace/test/workspace/reference_indexes.test.ts
tamtamirr: Not sure this is required, considering the assignment will fail if the `mockResolvedValue` type is not matching

PR #6108, opened 2024-06-24
SALTO-3518 bump node 18 and yarn berry 

eshedg:
Update to Node-18 and Yarn Berry
---

## Background

OSS has been lagging behind SAAS with old versions of Node and Yarn.
Several attempts in the past to bump these failed for various reasons, most notably an issue that rose with Jest getting stuck.

In this PR, I have tried to keep the changes and upgrades to a minimum, although it still ended up being a large (but not huge) change).

I left out additional improvements that can be done later (especially around parallelizing the tests in a similar manner to what is done in SAAS pipeline, and also around bumping other aging packages, getting rid of Nexe, and others).

I tried to get the build time to be about the same as before the bump, despite some constraints that demanded some things to become slower (hence the caching improvements I added and the all-out UT parallelization).

## Overview

This PR is composed of multiple updates and necessary adjustments in the following areas:
1. Node package versions
2. Yarn and build scripts
3. Code fixes
4. CI configuration

## Node package versions

There are major and minor updates.  The minor updates are responsible for there being a lot more touched files than I would otherwise prefer (all the `package.json` were affected).

### Major updates

- Node bumped to 18.9.0 (aligned with SAAS)
- Yarn bumped to 3.1.0 (aligned with SAAS)
- Turbocache introduced and set to v2 (aligned with SAAS)

### Minor updates

- Typescript bumped to 4.8.4 (aligned with SAAS)
- Jest bumped slightly (still in v27, behind SAAS with v29)
- JsZip bumped slightly

## Yarn and build scripts

With the upgrade to Yarn Berry there had to be made several adaptations, mainly around the Workspaces plugin.
These changes include:
1. Using the Workspaces `forEach` instead of `Lerna run --parallel` 
2. Adding some JS code in a couple of scripts to organize the monorepo information that yarn provides in a format that's compatible with the existing and tested script logic
3. Remove the existing hoisting configuration - The only place where it mattered was the CLI package, which received its special treatment (see codefixes)

I also introduced Turbocache to accelerate build and lint (works both in local and CI).  That required a few more script adjustments.

## Code fixes

There were mainly 3 areas that required code fixing:
1. TS Linting (mainly due to node 18 types, and not due to TS 4.8.4)
2. Jest
3. CLI packaging

### TS Linting issues

There were a few necessary updates in `eslint.rc` to workaround rules that were imported by the airbnb plugin and conflicted with node18 types.  We may want to resolve these in a separate PR.

There was also a small update in `mock.ts` (props to @ori-moisis) that was necessary to get past node18 constraints.

### Jest

For unknown reasons, running the unit tests with Jest multi-worker is causing race conditions and random deadlocks between tests, which cause them to hang.

Jest has a feature `--detectedOpenHandles` which is known to _not_ work as expected, so is pretty useless in finding out why tests hang.

Jest also has a feature `--forceExit` that tells it to terminate if all the tests have already ran (the summary was already reported) but some worker threads haven't exited (maybe promises that have remained unfulfilled).  

I encountered both types of 'jest not terminating' behavior, and ended up having to use both flags (if you just do the forceExit, you get a warning telling you to use the detectOpenHandles.  If you use them both, it works and it doesn't complain).

For the other problem, where forceExit is not helpful, I had to move the tests to run in Jest "inband" mode using the `-i` flag - This results in a much longer runtime, as the tests need to run serially and do not take advantage of multi-cores.

In order to keep from causing a major regression on the total UT running time in CI, I parallelized the UTs (see the CI Config changes).

### CLI Packaging

Our CLI packaging uses some outdated methods and tools, and revamping how we do it is out of scope for this PR.
Moving to the new Yarn Berry introduced a challenge because the way that package hoisting is done, and Yarn Berry's lack of configuration granularity to adjust this behavior for specific packages, in particular `@salto-io/rocksdb`.
In the spirit of keeping this PR small as it can be, I added a workaround for that, which effectively "de-hoists" rocksdb's dependencies that get hoisted by Yarn.  
This is only done during the CLI packaging step in the CI.

## CI Configuration

The configuration changes in the CI are in these areas:

1. UT parallelization
2. Orb bumping
3. Docker image versions
4. Caching improvements
5. Yarn updates
6. Tidying-up

### UT parallelization

As mentioned above, I introduced UT parallelization to mitigate the performance regression due to running Jest in a single threaded mode, which was done to go around its "not exiting" predicament.
There could be improvements here, like parallelizing the UT suites themselves using CircleCI's test splitting (as we do in SAAS E2E), which can reduce the longer suites (jira, sf, okta and ns) by a factor of 8.

### Orb bumping

We were using ancient orb versions, and as I was already doing a lot of revamping in the CI configuration, I bumped them and adjusted for the minor changes they introduced

### Docker image versions

Nothing interesting here - Just moved to `cimg/node:18.9`

### Caching improvements

1. I introduced Turbocache, which creates its own files, so I added CI caching for those to be shared across builds.
2. Yarn Berry manages its cache differently, so I added CI caching for it to reduce the time it takes for Yarn to load packages from the internet
3. `node_modules` were being cached in the Workspace, which is both slower than the caching mechanism, as well as not shared between builds.  Since Node_Modules are tightly coupled with the `yarn.lock` (same as the Yarn cache), there was a large performance benefit in caching them, and saving time during the Workspace attach (which now only handles our build artifacts), as well as sharing it across builds to compound this time saving across all builds.

### Yarn updates

There were a few adjustments needed to run Yarn Berry and handle its proper execution (mainly around upgrading it to Berry and adding necessary plugins such as Workspaces)

### Tidying-up

I moved repeated configuration segments to use parameterized commands (such as the Test_Cli part), and introduced simple commands instead of inline bash scripts in various places, to keep the flow of the steps easier to follow.

---
_Release Notes_: 

---
_User Notifications_: 


Dockerfile
```diff
@@ -1,12 +1,22 @@
...
 WORKDIR /app
 
-RUN apt update && apt install -y openssl build-essential libgssapi-krb5-2 git ccache
+RUN \
+    apt update && \
+    apt install -y openssl build-essential libgssapi-krb5-2 git ccache
 
 COPY . .
-RUN echo "Running yarn install" && yarn || yarn || yarn # run again if failing due to rimraf
+RUN echo "Updating yarn" && YARN_IGNORE_NODE=1 yarn set version 3.1.0
```
yelly, line 9: Also here, why `YARN_IGNORE_NODE`?

README.md
```diff
@@ -39,9 +39,12 @@ docker run salto-cli
 
 ### Building from source
 
-  1. Install Node.js 14 (for M1 Macs use 14.15). You can download it directly from [here](https://nodejs.org/en/download/releases/), or use [Node Version Manager (NVM)](https://github.com/nvm-sh/nvm) (simply run `nvm use`) to install it.
+  1. Install Node.js 18.9.0. You can download it directly from [here](https://nodejs.org/en/download/releases/), or use [Node Version Manager (NVM)](https://github.com/nvm-sh/nvm) (simply run `nvm use`) to install it.
   2. Install [yarn 1](https://yarnpkg.com/en/docs/install).
-  3. Fetch dependencies and build:
+  3. Update Yarn to Berry 3.1.0 by running `YARN_IGNORE_NODE=1 yarn set version 3.1.0`
```
yelly, line 44: Also here, why `YARN_IGNORE_NODE`?

build_utils/find_conflicting_versions.js
```diff
@@ -54,10 +58,13 @@ const main = async () => {
     {},
   )
 
+  // jest-circus: create-react-app does is using jest 27 via but jest-circus v27 was not installed
```
yelly, line 61: Nit: something is off with the grammar here.

build_utils/import-plugins.js
```diff
@@ -0,0 +1,52 @@
...
+*/
+const { execSync } = require('child_process')
+
+// List of plugins to check and import
+const plugins = ['@yarnpkg/plugin-workspace-tools', '@yarnpkg/plugin-interactive-tools']
+
+// Function to check if a plugin is installed
+function isPluginInstalled(plugin) {
+  try {
+    const result = execSync('yarn plugin runtime --json').toString()
```
yelly, line 26: Listing all of the plugins for every plugin we want to check is a bit inefficient, we can just list them all once.

build_utils/turbo_run.sh
```diff
@@ -0,0 +1,24 @@
...
+
+  # running in CI with 100% CPU will OOM
+  TURBO_CONCURRENCY="50%"
+
+  container_cpu_quota="/sys/fs/cgroup/cpu/cpu.cfs_quota_us"
+  container_cpu_scale="/sys/fs/cgroup/cpu/cpu.cfs_period_us"
+
+  if [ -f "$container_cpu_quota" ] && [ -f "$container_cpu_scale" ]; then
+
+    # if CI runs in containers, so finding the total allotted CPU is trickier
```
yelly, line 17:
```suggestion
    # When CI runs in containers finding the total allotted CPU is trickier
```

packages/adapter-components/jest.config.js
```diff
@@ -22,7 +22,7 @@ module.exports = deepMerge(require('../../jest.base.config.js'), {
   collectCoverageFrom: ['!<rootDir>/**/index.ts'],
   coverageThreshold: {
     global: {
-      branches: 85,
+      branches: 84,
```
yelly, line 25: Why does this need to come down? We're not touching any code or tests...

packages/cli/package.sh
```diff
@@ -0,0 +1,15 @@
...
+    cp -R "../../node_modules/$1" ../../node_modules/@salto-io/rocksdb/node_modules/
+}
+
+echo "workaround for yarn berry hoisting @salto-io/rocksdb dependencies"
+missing_dependencies='["is-buffer","catering","queue-tick","node-gyp-build"]'
+for dep in $(echo "$missing_dependencies" | jq -r '.[]'); do
+    de_hoist_package "$dep"
+done
+
+node ./package_native.js
```
yelly, line 15: Nit: add newline at end of file.

packages/file/test/file.test.ts
```diff
@@ -269,16 +268,22 @@ describe('file', () => {
         let dir: string
         let dirTmp: tmp.DirectoryResult
 
-        beforeEach(async () => {
+        // beforeEach(async () => {
+        // })
```
yelly, line 272: Remove this.

packages/file/test/file.test.ts
```diff
@@ -269,16 +268,22 @@ describe('file', () => {
...
-
-        it('removes the dir', async () => {
           expect(await file.exists(dir)).toBeFalsy()
+          if (await file.exists(dir)) {
+            await dirTmp.cleanup()
+          }
         })
+
+        // afterEach(async () => {
+        // })
```
yelly, line 286: Remove this.

packages/e2e-credentials-store/jest.config.js
```diff
@@ -24,10 +24,10 @@ module.exports = deepMerge(require('../../jest.base.config.js'), {
...
     // Slowly start increasing here, never decrease!
     global: {
-      branches: 34,
-      functions: 59,
-      lines: 64,
-      statements: 63,
+      branches: 32,
+      functions: 55,
+      lines: 62,
+      statements: 60,
```
yelly, line 30: Why do all of these need to come down?

packages/intercom-adapter/jest.config.js
```diff
@@ -23,7 +23,7 @@ module.exports = deepMerge(require('../../jest.base.config.js'), {
   testEnvironment: undefined,
   coverageThreshold: {
     global: {
-      branches: 88,
+      branches: 84,
```
yelly, line 26: Why does this need to come down?

packages/logging/jest.config.js
```diff
@@ -22,7 +22,7 @@ module.exports = deepMerge(require('../../jest.base.config.js'), {
   coveragePathIgnorePatterns: ['src/index.ts'],
   coverageThreshold: {
     global: {
-      branches: 97,
+      branches: 96,
```
yelly, line 25: Why does this need to come down?

packages/lowerdash/jest.config.js
```diff
@@ -23,7 +23,7 @@ module.exports = deepMerge(require('../../jest.base.config.js'), {
   coverageThreshold: {
     global: {
       branches: 99,
-      functions: 100,
+      functions: 98,
```
yelly, line 26: Why does this need to come down?

packages/microsoft-entra-adapter/jest.config.js
```diff
@@ -23,10 +23,10 @@ module.exports = deepMerge(require('../../jest.base.config.js'), {
...
   coverageThreshold: {
     global: {
-      branches: 84.8,
-      functions: 89.2,
-      lines: 92.75,
-      statements: 93.21,
+      branches: 83,
+      functions: 88.7,
+      lines: 92.58,
+      statements: 93.02,
```
yelly, line 29: Why does these need to come down?

packages/netsuite-adapter/jest.config.js
```diff
@@ -23,7 +23,7 @@ module.exports = deepMerge(require('../../jest.base.config.js'), {
   testEnvironment: process.env.RUN_E2E_TESTS ? '@salto-io/netsuite-adapter/dist/e2e_test/jest_environment' : undefined,
   coverageThreshold: {
     global: {
-      branches: 90,
+      branches: 89,
```
yelly, line 26: Why does this need to come down?

packages/salesforce-adapter/jest.config.js
```diff
@@ -25,7 +25,7 @@ module.exports = deepMerge(require('../../jest.base.config.js'), {
     : undefined,
   coverageThreshold: {
     global: {
-      branches: 88.65,
+      branches: 88,
```
yelly, line 28: Why does this need to come down?

tsconfig.base.json
```diff
@@ -16,6 +16,7 @@
     "composite": true,
     "noEmitOnError": true,
     "experimentalDecorators": true,
+    "skipLibCheck": true,
```
yelly, line 19: Why? Is the build time a big problem?

.circleci/config_template.yml
```diff
@@ -2,13 +2,120 @@
...
+
+  restore_node_modules:
+    description: restore node modules
+    steps:
+      - run:
+          name: restore node modules
+          command: echo ''
+      - restore_cache:
+          keys:
+            - &node-modules-cache node-modules-v4-{{ arch }}}-{{ .Branch }}-{{ checksum "yarn.lock" }}
```
amitizle, line 44: I'm no expert in YAML but wouldn't it transform it to:
```yaml
keys:
  - node-modules-v4-{{ arch }}}-{{ .Branch }}-{{ checksum "yarn.lock" }}:
      paths:
        - /mnt/ramdisk/project/node_modules
  - node-modules-v2-{{ arch }}}-{{ .Branch }}
  - node-modules-v2-{{ arch }}}-main
 ```

Meaning, will the `path` value be macro'd back to the same level as `keys`?

File: .gitignore [hunk 19]
```diff
@@ -19,6 +19,8 @@ junit.xml
 
 # yarn
 yarn-error.log
+/.yarn/
+.yarnrc.yml
```
amitizle, line 23: Should be:
```
.yarn
!.yarn/releases
!.yarn/plugins
!.yarn/sdks
!.yarn/versions
```
.yarnrc.yml should definitely be also checked in

.turbo/config.json
```diff
@@ -0,0 +1,4 @@
+{
+    "teamId": "SaltoDev",
+    "apiUrl": "https://turbocache.salto-internal.io"
```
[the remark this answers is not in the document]
amitizle: +1, remove it, we're gonna remove this thing in any case.

Dockerfile
```diff
@@ -1,12 +1,22 @@
...
+RUN \
+    echo "Setting up yarnrc" && \
+    yarn config set cacheFolder .yarn/cache && \
+    yarn config set nodeLinker node-modules
```
amitizle, line 13: Same as above, make the modifications to `.gitignore` and it'll be fine, no need to explicitly set it up

Dockerfile
```diff
@@ -1,12 +1,22 @@
...
+
+RUN echo "Running yarn install" && yarn 
 
 RUN \
-    echo "Running yarn build" && yarn pre-build && yarn build-ts
+    echo "Running yarn build" && \
+    yarn pre-build && \
+    yarn build-ts
 
 ENTRYPOINT ["/usr/bin/env", "node", "./packages/cli/bin/salto"]
```
amitizle, line 22: In general, why are we rebuilding everything and not copying the Linux binary?

build_utils/clean-old-ts.sh
```diff
@@ -15,7 +15,7 @@ diff --strip-trailing-cr <(
   find ${PWD}/dist -type f |
   sort
 ) <(
-  yarn -s tsc -b --clean --dry |
+  yarn tsc -b --clean --dry |
   grep -e '^ ' |
   sed -e 's/^.*[*] //' |
   sort
```
amitizle, line 21: With turborepo, I am not sure anything from here is needed. I mean, this entire cryptic script

build_utils/verify_tsconfig_references.js
```diff
@@ -36,9 +36,33 @@ const filterValues = (o, f) => Object.fromEntries(
   Object.entries(o).filter(([k, v], i) => f(v, k, i))
 )
 
+const readWorkspacesFromYarnBerry = async () => {
```
amitizle, line 39: What are we doing in this file I'm so confused

packages/vscode/scripts/package.sh
```diff
@@ -1,12 +1,23 @@
...
+yarn workspaces focus salto-vscode
+
+sudo apt update && sudo apt install -y rsync
+
+rm -rf ./tmp_pkg
 mkdir -p ./tmp_pkg && cp ../../LICENSE . && vsce package --yarn -o ./tmp_pkg/salto.vsix && rm -f LICENSE
-cd ./tmp_pkg && unzip salto.vsix
-cp -r ../node_modules/!(@salto-io) extension/node_modules/
+
+pushd ./tmp_pkg
```
amitizle, line 13: nit: you can use `mktemp -d` like `cd $(mktemp -d)`

build_utils/import-plugins.js
eshedg: we have 2.  total. ;)

build_utils/verify_tsconfig_references.js
eshedg: not sure, i only fixed the input so it can work.  i think that deprecating stuff like this is a bit outside the scope of this PR.  wdyt?

packages/adapter-components/jest.config.js
eshedg: evidently, it failed with reduced branch coverage.

build_utils/clean-old-ts.sh
amitizle: @ori-moisis and I checked and by using `turbo`, you don't need this.
We can make sure that from within a package, the `turbo` command will be running.
For example, from withing `logging`, you can run `yarn turbo run -F '@salto-io/logging' build-ts`.

.circleci/config_template.yml
```diff
@@ -2,13 +2,120 @@
...
+  restore_intermediate:
+    description: restore intermediate artifacts between steps
+    steps:
+      - attach_workspace:
+          at: .workspace
+      - run:
+          name: restore intermediate artifacts between steps
+          command: |
+            tar -xzf .workspace/workspace.tar.gz
+            rm .workspace/workspace.tar.gz
```
amitizle, line 34: Instead of storing and restoring we can use `persist_to_workspace` with globs, like:
```yaml
persist_to_workspace:
  root: ~/project
  paths: packages/*/dist/**/*
```

build_utils/import-plugins.js
amitizle: Nothing is needed if you're commiting `.yarnrc.yml` and `.yarn/plugins` and `.yarn/releases`

build_utils/import-plugins.js
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env node
+
+/*
+*                      Copyright 2024 Salto Labs Ltd.
```
amitizle, line 4: This script can be removed

packages/cli/package_native.js
```diff
@@ -44,18 +44,19 @@ const fontFiles = require('./dist/src/fonts').fontFiles
  * and the binaries you uploaded should be used automatically.
  * */
 
-const BUILD_NODE_EXECUTABLE = false
+const BUILD_NODE_EXECUTABLE = process.env.BUILD_NODE_EXECUTABLE ? true : false
```
amitizle, line 47: Who's setting up this this environment variable?

packages/jira-adapter/src/filters/layouts/issue_layout.ts
```diff
@@ -136,7 +136,8 @@ const getProjectToScreenMappingUnresolved = (elements: Element[]): Record<string
               .map(
                 (issueTypeMapping: issueTypeMappingStruct) =>
                   screensSchemesToDefaultOrViewScreens[issueTypeMapping.screenSchemeId],
-              ),
+              )
+              .filter(isDefined),
```
amitizle, line 140: What is this change

packages/cli/package.sh
```diff
@@ -0,0 +1,15 @@
...
+
+function de_hoist_package() {
+    cp -R "../../node_modules/$1" ../../node_modules/@salto-io/rocksdb/node_modules/
+}
+
+echo "workaround for yarn berry hoisting @salto-io/rocksdb dependencies"
+missing_dependencies='["is-buffer","catering","queue-tick","node-gyp-build"]'
+for dep in $(echo "$missing_dependencies" | jq -r '.[]'); do
+    de_hoist_package "$dep"
+done
```
ori-moisis, line 13: 1. do we need to undo this at the end?
2. should this be part of `package_native.js`? why separate the logic between the two?

packages/jira-adapter/test/filters/layouts/issue_layout.test.ts
```diff
@@ -366,9 +366,7 @@ describe('issue layout filter', () => {
         projectInstance1.value.issueTypeScheme = { issueTypeScheme: { id: 10 } }
         const res = getLayoutRequestsAsync(client, config, fetchQuery, elements)
         expect(Object.entries(res)).toHaveLength(1)
-        expect(res['11111']).toEqual({
-          undefined: Promise.resolve({ data: {} }),
-        })
+        expect(res['11111']).toEqual({})
```
ori-moisis, line 369: This seems like an actual logic change.
why do we need to do it?

if we do, can we put it in a separate PR please?

packages/vscode/scripts/package.sh
```diff
@@ -1,12 +1,23 @@
 #!/bin/bash
+
 set -xeuo pipefail
 shopt -s extglob
 
+yarn workspaces focus salto-vscode
+
+sudo apt update && sudo apt install -y rsync
```
ori-moisis, line 8: we should not run this every time we package the extension

turbo.json
```diff
@@ -0,0 +1,26 @@
+{
+  "$schema": "https://turborepo.org/schema.json",
+  "tasks": {
+    "build-ts": {
+      "outputs": [
+        "dist/**"
+      ],
+      "dependsOn": [
+        "^build-ts"
```
ori-moisis, line 9: is it really ok to run `build-ts` without the `generate` step first?

packages/cli/package_native.js
```diff
@@ -44,18 +44,19 @@ const fontFiles = require('./dist/src/fonts').fontFiles
...
 
 const TARGET_FILE_BASENAME = 'salto'
 const TARGET_DIR = 'pkg'
 const TARGET_ARCH = BUILD_NODE_EXECUTABLE ? os.arch() : 'x64'
-const TARGET_NODE_VERSION = '14.15.3'
+const TARGET_NODE_VERSION = '18.12.1'
```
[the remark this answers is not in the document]
eshedg: this was the version that Tom used to pre-build the NEXE images.
it doesn't really matter, and I think we can align it when we bump to the next node ver (or get rid of Nexe altogether).
for now, i didn't see this as a good-enough reason to pre-build Nexe images for the 3 platforms from scratch.

packages/file/test/file.test.ts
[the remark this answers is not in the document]
eshedg: I suspected it was involved in the hanging Jest issue for not closing files properly.
I put it thru Chatgpt and it proposed some changes which made sense, and are not related to the actual functionality that the test is checking, so I kept them.

packages/cli/package.sh
eshedg: 1. this is done in a final step, it has zero affect, as the workspace is not saved.  
2. i think it's more an issue whether to impl in bash or js.  in any case, this would be a-hack-to-solve-a-hack and I expect this to go away when we move away from the Nexe approach

packages/jira-adapter/test/filters/layouts/issue_layout.test.ts
eshedg: this was a bug that was discovered by moving to Node18 (some race condition was hiding it), and @IdoZakk fixed it.  I just added it, or there would be a Jira test that never passes.

packages/vscode/scripts/package.sh
eshedg: why? what is the concern?

turbo.json
eshedg: the Generate step is only implemented in core, cli and dummy, and it only copies some stuff, not something that is related to turbo.  build-ts is the step that Turbo actually involved with.

.circleci/config.yml
```diff
@@ -9,17 +9,27 @@ orbs:
...
+          name: set up yarn
           command: |
-            pushd .circleci/scripts
-            node find_changed_packages.js
-            node update_config_template.js
+            yarn set version 3.1.0
+            yarn config set cacheFolder /mnt/ramdisk/project/.yarn/cache
+            yarn config set nodeLinker node-modules
+          environment:
+            YARN_IGNORE_NODE: 1
```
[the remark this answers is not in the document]
eshedg: 1. re set yarn version - per Yarn's documentation (https://yarnpkg.com/configuration/yarnrc#yarnPath), specifying yarnPath used to be the way, but not anymore.  the recommended way to do it is by enabling corepack (which is done by default in the Circle CI node docker image), and then use yarn set version, which adds a `packageManager` entry in the main package.json.

2. the `cacheFolder` setting is specific to the CI here, so it should remain in the config.yml.

3. the nodeLinker can move to the .yarnrc.yml - I'll do that.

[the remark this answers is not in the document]
eshedg: > Why? I thought you said you're not doing this any more.

i assume you mean the `YARN_IGNORE_NODE=1`?
that's actually a guard against yarn failing to set the version to 3.1.0 because of requirements on the node version.
it is not relevant for the build itself.

.gitignore
eshedg: .yarnrc.yml - added to commit
all the rest I'm leaving as is, since I don't want to add the plugins etc to our repo.

Dockerfile
eshedg: when you update yarn ver, it moves from yarn 1 to yarn latest, then evaluates its Node version minimal requirements, before fulfilling your request.
so, in this case, you get:
```
0.130 Updating yarn                                                                                                                                                                                                                                                                      
2.362 Usage Error: This tool requires a Node version compatible with >=18.12.0 (got 18.9.0). Upgrade Node, or set `YARN_IGNORE_NODE=1` in your environment.                                                                                                                              
```

Now, you are still right - we don't need the override... provided we use `corpack enable`
I'm fixing that.

build_utils/update_package.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+if [ -z "$1" ] || [ -z "$2" ]; then
+    echo "$0 package_name version search_path"
+    echo "example:"
```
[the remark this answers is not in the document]
eshedg: it's a utility i added that replaces a version of an import in all package.json.

build_utils/import-plugins.js
eshedg: yes, but not committing .yarn at all.

Dockerfile
eshedg: this is just a rudimentary Dockerfile that is customary in open source projects

build_utils/find_conflicting_versions.js
eshedg: not sure why i added this, i'll try to remove

packages/cli/package_native.js
eshedg: this is only for a scenario where you want to build a new Nexe binary version.  the caller will turn this flag on.  on all other times, we use the prebuilt image from the remote storage.

packages/intercom-adapter/jest.config.js
eshedg: i don't know

packages/jira-adapter/src/filters/layouts/issue_layout.ts
eshedg: fixed by @IdoZakk - it was a bug, apparently

packages/logging/jest.config.js
eshedg: idk

packages/lowerdash/jest.config.js
eshedg: idk

packages/microsoft-entra-adapter/jest.config.js
eshedg: idk

packages/netsuite-adapter/jest.config.js
eshedg: idk

packages/test-utils/src/mock.ts
```diff
@@ -14,7 +14,11 @@
  * limitations under the License.
  */
 export type MockInterface<T extends {}> = {
-  [k in keyof T]: T[k] extends (...args: never[]) => unknown ? jest.MockedFunction<T[k]> : MockInterface<T[k]>
+  [k in keyof T]: T[k] extends (...args: never[]) => unknown
+    ? jest.MockedFunction<T[k]>
+    : T[k] extends {}
+      ? MockInterface<T[k]>
+      : never
```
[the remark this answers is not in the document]
eshedg: @ori-moisis fixed it

packages/vscode/scripts/package.sh
eshedg: i kept the change to minimum

tsconfig.base.json
eshedg: the problem is not build time.  it's linter issues with ancient imports that need to be bumped.

packages/file/test/file.test.ts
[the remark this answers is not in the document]
ori-moisis: if you don't have to change something in this PR, please don't

packages/cli/package.sh
ori-moisis: 1. please remember that people also work locally. this is not just about the CI
2. unless you know that there is a nexe alternative and that you or someone else is going to work on it soon (i.e - start within at most 2 months), we have to assume nexe is here to stay for a while and we should not make things worse

packages/jira-adapter/test/filters/layouts/issue_layout.test.ts
ori-moisis: > if we do, can we put it in a separate PR please?

packages/vscode/scripts/package.sh
ori-moisis: time
side effects on every developer's computer
requiring `sudo` permissions to use our repo

packages/cli/package_native.js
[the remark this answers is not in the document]
ori-moisis: ok

packages/jira-adapter/test/filters/layouts/issue_layout.test.ts
eshedg: #6147

ori-moisis: thanks!

packages/cli/package.sh
eshedg: point taken.  
1. the action doesn't need to be undone, it doesn't change anything observable
2. i don't think it makes anything worse by enabling the packaging to work (vs. not work).  the alternative is to not upgrade to Yarn Berry, which has much more adverse effect.

packages/file/test/file.test.ts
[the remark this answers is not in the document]
eshedg: I'll try to revert this file and if there is no regression on the UTs, I'll keep it out of this PR

.circleci/config_template.yml
eshedg: the purpose of the `&node-modules-cache` is to store the key definition, so it can be used in the `store_cache` command without defining it twice (so you maintain it in one place).  i followed examples in implementing this, and also tested it to behave as I want.
there was a bug there in the other keys, they should be the same `v4` but had a typo and were `v2`, but it's fixed in a later commit.

.circleci/config_template.yml
eshedg: we are also persisting the checked out code.  so it's a bit more complicated to express in a glob, especially since CircleCI don't support excluding from the path.

packages/vscode/scripts/package.sh
eshedg: i didn't realize this was something people may run themselves (not in CI), and in any case using `apt` is really debian/ubuntu minded.
I'm adding code to check if rsync is installed, and only install it if running in CI (CI==1)
users running this script will get an error message.

packages/salesforce-adapter/jest.config.js
eshedg: we discussed, we do not know why this happens, resolving.

packages/netsuite-adapter/jest.config.js
eshedg: we discussed, we do not know why this happens, resolving.

packages/microsoft-entra-adapter/jest.config.js
eshedg: we discussed, we do not know why this happens, resolving.

packages/lowerdash/jest.config.js
eshedg: we discussed, we do not know why this happens, resolving.

.circleci/scripts/update_config_template.js
```diff
@@ -51,19 +55,54 @@ const main = () => {
...
+  }
+  
+const generateUtMatrix = (templateName, outputName) => {
+  const configTemplate = readFileSync(templateName, 'utf8')
+  const utPackagesToTest = readFileSync(path.join(__dirname, '..', 'ut_packages_to_test.txt'), 'utf8').split('\n')
+  const utPackagesWithPrettyName = utPackagesToTest.map((pkg) => { return `${pkg.split('/').pop()}:${pkg}` })
   
-  writeFileSync(path.join(__dirname, '..', 'continue_config.yml'), alteredConfigTemplate)
-  console.log(alteredConfigTemplate)
+  // TODO: optimize on the resource_class for each package, using the CI resource utilization analysis
```
yelly, line 79: Can we open a JIRA issue for this?

eslintrc.js
```diff
@@ -155,6 +155,9 @@ module.exports = {
         '@typescript-eslint/ban-ts-ignore': [0],
         'max-classes-per-file': [0],
         'import/no-unresolved': [0], // unable to read paths from tsconfig.json
+        'import/no-cycle': [0], // airbnb rules collide with node-18 types
```
yelly, line 158: Maybe we should add a todo to get rid of all of this Airbnb stuff?

yelly (commented):
Overall LGTM but I'll leave approval to someone else.

.circleci/config.yml
```diff
@@ -9,17 +9,24 @@ orbs:
...
       - checkout
       - run:
-          name: Run Find Changed Packages Script
+          name: set up yarn
           command: |
-            pushd .circleci/scripts
-            node find_changed_packages.js
-            node update_config_template.js
+            sudo corepack enable
+            yarn config set cacheFolder /mnt/ramdisk/project/.yarn/cache
```
amitizle, line 21: You're using `ramdisk` with `small` `resource_class`, I think this can be close to the ramdisk limit

Dockerfile
amitizle: Yes but we already have the binary no? Why build the Typescript project again?

build_utils/clean-old-ts.sh
amitizle: Ping

build_utils/find_conflicting_versions.js
```diff
@@ -54,10 +58,14 @@ const main = async () => {
     {},
   )
 
+  const ignoreDeps = []
```
amitizle, line 61: Why?

build_utils/turbo_run.sh
```diff
@@ -0,0 +1,24 @@
...
+  # running in CI with 100% CPU will OOM
+  TURBO_CONCURRENCY="50%"
+
+  container_cpu_quota="/sys/fs/cgroup/cpu/cpu.cfs_quota_us"
+  container_cpu_scale="/sys/fs/cgroup/cpu/cpu.cfs_period_us"
+
+  if [ -f "$container_cpu_quota" ] && [ -f "$container_cpu_scale" ]; then
+
+    # when CI runs in containers, finding the total allotted CPU is trickier
+    TURBO_CONCURRENCY="$(awk -v quota="$(cat /sys/fs/cgroup/cpu/cpu.cfs_quota_us)" -v period="$(cat /sys/fs/cgroup/cpu/cpu.cfs_period_us)" 'BEGIN { printf "%.0f\n", (quota / period) * 0.5 }')"
```
amitizle, line 18: I'm OK with leaving it in for now, but I would much rather having some more readable script for that.
See for example the comment by the CircleCI employee in [this thread](https://circleci.canny.io/cloud-feature-requests/p/have-nproc-accurately-reporter-number-of-cpus-available-to-container)

build_utils/find_conflicting_versions.js
```diff
@@ -32,11 +32,15 @@ const filterValues = (o, f) => Object.fromEntries(
 )
 
 const readManifests = async () => {
-  const { stdout } = await exec('yarn workspaces -s info')
-  return mapValuesAsync(JSON.parse(stdout), async info => {
-    const manifestFile = path.join(info.location, 'package.json')
-    return JSON.parse(await fs.promises.readFile(manifestFile))
-  })
+  const { stdout } = await exec('yarn workspaces list --json -v')
+  const entries = await Promise.all(
```
amitizle, line 36: For this entire file: wouldn't [constraints](https://yarnpkg.com/features/constraints) solve it ootb?

build_utils/update_package.sh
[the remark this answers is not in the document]
amitizle: Can you use [interactive upgrade](https://yarnpkg.com/cli/upgrade-interactive) instead?

build_utils/verify_tsconfig_references.js
amitizle: Not sure since there's a new logic in here.
I vote for removing this file and its usage altogether.
@ori-moisis @yelly do you agree?

Dockerfile
ori-moisis: It is actually nice that the docker doesn't rely on the packaged executable, both because it means less dependencies and less things that could go wrong, and also because it makes it is possible to "hack" the code in the docker if needed.

tsconfig.base.json
eshedg: i believe we have several open tasks to bump various libraries.
as for a TODO - it's a json file, you can't put comments inside...

Dockerfile
eshedg: maybe, but it's not in the scope of this PR (the change above was only line breaks).

build_utils/clean-old-ts.sh
eshedg: we talked, and you said to leave it alone

build_utils/update_package.sh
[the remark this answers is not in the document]
eshedg: it's not replacing interactive_upgrade - - it's a fast find&replace tool for package versions.  yarn's interactive upgrade only lets you go up to the latest minor, or the latest version.  if you want to specifically pin some arbitrary version (which was what I used this for - - to align with versions that SAAS uses), then this does the trick.  in any case.

tsconfig.base.json
yelly: So I think we need an open issue specifically to remove this config - I think we are opening ourselves up to new bugs as long as it's here.

build_utils/verify_tsconfig_references.js
eshedg: actually, it is not new logic - i added this function to re-arrange the output of `yarn workspaces list --json -v` to fit the existing logic.
Berry's `workspaces` plugin's output is slightly different than what it used to be in old Yarn (the package name used to be the key, and it's now one level down).  
without this function, the script fails to parse.

build_utils/verify_tsconfig_references.js
```diff
@@ -82,12 +106,11 @@ const main = async () => {
...
   )
-
-  const workspacePackageToReference = (
-    package, refPackage,
-  ) => ({
-    path: path.relative(workspaces[package].location, workspaces[refPackage].location)
-  })
+  
+  const workspacePackageToReference = (package, refPackage) => {
+    const relativePath = path.relative(workspaces[package].location, workspaces[refPackage])
```
ori-moisis, line 111: the old code was correct, this seems to be broken.
either revert this change, or fix it
```suggestion
    const relativePath = path.relative(workspaces[package].location, workspaces[refPackage].location)
```

build_utils/verify_tsconfig_references.js
ori-moisis: IIUC this script ensures that the dependencies in `package.json` match the references in `tsconfig.json`.
ACAICT, nothing else enforces this, so, why would it be ok to remove?

build_utils/find_conflicting_versions.js
eshedg: maybe, but it's out of scope for this PR

.circleci/config.yml
eshedg: nope.  this step only creates the continuation pipeline, and it takes 11% of the memory at max.

.turbo/config.json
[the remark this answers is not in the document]
eshedg: added by mistake.  it's autogenerated.

.circleci/scripts/update_config_template.js
eshedg: https://salto-io.atlassian.net/browse/SALTO-6217

eslintrc.js
eshedg: whoever bumps to node22 will need to tackle this anyhow

.turbo/config.json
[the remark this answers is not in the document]
yelly: It's still here. Should it be or does it belong in the `.gitignore`?

.circleci/config.yml
```diff
@@ -9,17 +9,24 @@ orbs:
...
     docker:
-      - image: cimg/node:14.15
+      - image: cimg/node:18.9
     resource_class: small
 
     steps:
       - checkout
       - run:
-          name: Run Find Changed Packages Script
+          name: set up yarn
```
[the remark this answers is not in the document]
yelly: Did you mean to fix this?

.circleci/scripts/update_config_template.js
yelly: Can you add the issue to the comment so we can easily get back to it?
```suggestion
  // TODO (SALTO-6217): optimize on the resource_class for each package, using the CI resource utilization analysis
```

build_utils/find_conflicting_versions.js
eshedg: must be a mistake.  reverting this change

turbo.json
```diff
@@ -0,0 +1,27 @@
...
+      ]
+    },
+    "test": {
+      "outputs": ["coverage/**"]
+    },
+    "clean": {
+      "cache": false
+    }
+  }
+}
```
amitizle, line 27: You might want to add:
```json
{
  "globalDependencies": ["tsconfig.json"],
}
```

.circleci/scripts/update_config_template.js
```diff
@@ -17,12 +17,16 @@
...
 const { readFileSync, writeFileSync, existsSync, readdirSync } = require('fs')
 const path = require('path')
+const { config } = require('process')
 
-
-const main = () => {
-  const configTemplate = readFileSync(path.join(__dirname, '..', 'config_template.yml'), 'utf8')
-  const adaptersWithJava = readFileSync(path.join(__dirname, '..', 'adapters_with_java.txt'), 'utf8').split('\n').filter(Boolean)
-  const e2ePackagesToTest = readFileSync(path.join(__dirname, '..', 'e2e_packages_to_test.txt'), 'utf8').split('\n').filter(Boolean)
+const generateE2eMatrix = (templateName, outputName) => {
```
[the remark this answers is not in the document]
eshedg: for the future @amitizle : we may be able to just use a dockerized java without installing it

amitizle (approved):
@eshedg ⭐ please plan with the team on how to rebase and what is changing etc before merging.

PR #6111, opened 2024-06-24
SALTO-6074 PagerDuty - resolve userId to email

Uri-Bar:
SALTO-6074 PagerDuty - resolve userId to email
---

---
_Release Notes_: 
none

---
_User Notifications_: 
none

### DIFF
--- a/packages/pagerduty-adapter/src/adapter_creator.ts
+++ b/packages/pagerduty-adapter/src/adapter_creator.ts
@@ -13,7 +13,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { client as clientUtils, createAdapter, credentials as defaultCredentials } from '@salto-io/adapter-components'
+import {
+  client as clientUtils,
+  createAdapter,
+  credentials as defaultCredentials,
+  filters,
+} from '@salto-io/adapter-components'
 import { Credentials, credentialsType } from './auth'
 import createChangeValidator from './change_validator'
 import { DEFAULT_CONFIG, UserConfig } from './config'
@@ -23,6 +28,7 @@ import { createClientDefinitions, createDeployDefinitions, createFetchDefinition
 import { PAGINATION } from './definitions/requests/pagination'
 import { Options } from './definitions/types'
 import { REFERENCES } from './definitions/references'
+import convertUsersIds from './filters/users'
 
 const { defaultCredentialsFromConfig } = defaultCredentials
 
@@ -57,6 +63,10 @@ export const adapter = createAdapter<Credentials, Options, UserConfig>({
     connectionCreatorFromConfig: () => createConnection,
     credentialsFromConfig: defaultCredentialsFromConfig,
     additionalChangeValidators: createChangeValidator,
+    customizeFilterCreators: args => ({
+      ...filters.createCommonFilters<Options, UserConfig>(args),
+      convertUsersIds,
+    }),
   },
   initialClients: {
     main: undefined,
--- a/packages/pagerduty-adapter/src/adapter_creator.ts
+++ b/packages/pagerduty-adapter/src/adapter_creator.ts
@@ -19,6 +19,7 @@ import {
   credentials as defaultCredentials,
   filters,
 } from '@salto-io/adapter-components'
+import { BuiltinTypes } from '@salto-io/adapter-api'
 import { Credentials, credentialsType } from './auth'
 import createChangeValidator from './change_validator'
 import { DEFAULT_CONFIG, UserConfig } from './config'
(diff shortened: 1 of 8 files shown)


packages/pagerduty-adapter/src/adapter_creator.ts
```diff
@@ -57,6 +69,10 @@ export const adapter = createAdapter<Credentials, Options, UserConfig>({
     connectionCreatorFromConfig: () => createConnection,
     credentialsFromConfig: defaultCredentialsFromConfig,
     additionalChangeValidators: createChangeValidator,
+    customizeFilterCreators: args => ({
+      ...filters.createCommonFilters<Options, UserConfig>(args),
+      convertUsersIds,
```
edenhassid, line 74: why is this separate and not part of all the filters?

packages/pagerduty-adapter/src/filters/users.ts
```diff
@@ -0,0 +1,172 @@
...
+  const targets = _.get(value, 'targets')
+  return Array.isArray(targets) && targets.every(target => _.isString(target.id) && _.isString(target.type))
+}
+
+const isScheduleLayerUser = (value: unknown): value is ScheduleLayerUser => {
+  const user = _.get(value, 'user')
+  return _.isPlainObject(user) && _.isString(user.id) && _.isString(user.type)
+}
+
+const replaceValues = (instance: InstanceElement, mapping: Record<string, string>): void => {
```
edenhassid, line 51: can you add a comment with an example for each type?

packages/pagerduty-adapter/src/filters/users.ts
```diff
@@ -0,0 +1,172 @@
...
+      break
+
+    case SCHEDULE_LAYERS_TYPE_NAME:
+      makeArray(instance.value.users)
+        .filter(isScheduleLayerUser)
+        .forEach(userObj => {
+          const userIdentifier = _.get(userObj, 'user.id')
+          if (mapping[userIdentifier]) {
+            _.set(userObj, 'user.id', mapping[userIdentifier])
+          }
```
edenhassid, line 75: maybe add a log if `mapping[userIdentifier]` is undefined? (same for both cases)

packages/pagerduty-adapter/src/filters/users.ts
```diff
@@ -0,0 +1,172 @@
...
+        })
+      break
+    default:
+      break
+  }
+}
+
+export const replaceValuesForChanges = async (
```
edenhassid, line 83: why do you export this?

packages/pagerduty-adapter/src/filters/users.ts
```diff
@@ -0,0 +1,172 @@
...
+  return {
+    name: 'usersFilter',
+    onFetch: async elements => {
+      if (!(config.fetch.convertUsersIds ?? DEFAULT_CONVERT_USERS_IDS_VALUE)) {
+        log.debug('Converting user ids was disabled (onFetch)')
+        return
+      }
+
+      // Using casting as the difference between definitions type and RequiredDefinitions is that the fetch is not mandatory in  the definitions, but I override it.
+      const users = await fetchUtils.getElements({
```
edenhassid, line 115: 1. do we really want to use getElements? isn't it enough just to have a map of user id to email?

2. also we would want this code to run only if we have relevant instances right?

packages/pagerduty-adapter/src/filters/users.ts
```diff
@@ -0,0 +1,172 @@
...
+      if (!users || _.isEmpty(users)) {
+        log.warn('Could not find any users (onFetch)')
+      }
+
+      userIdToLogin = Object.fromEntries(
+        users.elements.filter(isInstanceElement).map(user => [user.value.id, user.value.email]),
+      )
+      const loginToUserId = Object.fromEntries(
+        users.elements.filter(isInstanceElement).map(user => [user.value.email, user.value.id]),
+      )
```
edenhassid, line 155: I think you can put all this logic in the utils as well

packages/pagerduty-adapter/src/filters/users.ts
```diff
@@ -0,0 +1,172 @@
...
+      await replaceValuesForChanges(relevantChanges, loginToUserId)
+    },
+    onDeploy: async (changes: Change<InstanceElement>[]) => {
```
edenhassid, line 158: I am not sure we need the onDeploy actually, cause we clone the changes before we change them right? so there is no need to bring them back to the prev state

edenhassid (commented):
looks good! added some comments

packages/pagerduty-adapter/src/adapter_creator.ts
Uri-Bar: talkef f2f

packages/pagerduty-adapter/src/filters/users.ts
Uri-Bar: 1. talked f2f
2. fixed

packages/pagerduty-adapter/src/filters/users.ts
Uri-Bar: I think its just one liner, if it ok I will leave it her.

packages/pagerduty-adapter/src/filters/users.ts
Uri-Bar: removing the on deploy broke the flow, so its stays

edenhassid (approved):
great!

PR #6114, opened 2024-06-24
SALTO-6135: Add the ability to always regenerate salto Ids for Salesforce Data Instances

tamtamirr:
Add the ability to always regenerate salto Ids for Salesforce Data Instances

---

Added a config field `regenerateSaltoIds` under the `DataManagement` config.

---
_Release Notes_: 
Salesforce Adapter:
- Add the ability to always regenerate salto Ids for Salesforce Data Instances by setting `data.regenerateSaltoIds` to true. 

---
_User Notifications_: 
_None_


packages/salesforce-adapter/config_doc.md
```diff
@@ -212,6 +212,7 @@ salesforce {
 | [saltoManagementFieldSettings](#salto-management-field-configuration-options) | {}                                               | Configuration for managed-by-Salto field                                                                  |
 | [brokenOutgoingReferencesSettings](#broken-outgoing-references-settings)      | {}                                               | Configuration for handling broken references                                                              |
 | omittedFields                                                                 | []                                               | List of API names of fields to discard when fetching data records.                                        |
+| regenerateSaltoIds                                                            | false                                            | When set to true, the adapter will regenerate the Ids of Data Instances.                                  |
```
yelly, line 215: Maybe make it clear that it will regenerate on every fetch.

packages/salesforce-adapter/src/types.ts
```diff
@@ -677,6 +678,9 @@ const dataManagementType = new ObjectType({
     omittedFields: {
       refType: new ListType(BuiltinTypes.STRING),
     },
+    regenerateSaltoIds: {
+      refType: BuiltinTypes.BOOLEAN,
```
yelly, line 682: Did you consider putting the in the `saltoIDSettings`?

yelly (approved):
LGTM, just a small comment.

packages/salesforce-adapter/config_doc.md
tamtamirr: On every fetch as long as it's in the config though..?

packages/salesforce-adapter/src/types.ts
tamtamirr: Yes, either is fine by me. Considering we put `omittedFields` and other settings under the data config I didn't mins much.

packages/salesforce-adapter/config_doc.md
yelly: Not sure I understand.
Is the following accurate?
```
When set to true, the adapter will regenerate the Ids of Data Instances on every fetch.
```

packages/salesforce-adapter/src/types.ts
yelly: I think putting it with the other ID settings makes sense, but open to other opinions here.

PR #6115, opened 2024-06-25
SALTO-6141: support extendParent for pathParts

edenhassid:
support extendParent for pathParts

---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/adapter-components/test/fetch/element/id_utils.test.ts
+++ b/packages/adapter-components/test/fetch/element/id_utils.test.ts
@@ -248,6 +248,29 @@ describe('id utils', () => {
         })({ entry: { a: 'A', b: 'B', c: 'C' }, defaultName: 'unnamed' }),
       ).toEqual(['myAdapter', 'Records', 'myType', 'thisIsACuteCustomName'])
     })
+    it('should calculate correct path when extendParent is true', () => {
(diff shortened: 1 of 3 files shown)


packages/adapter-components/test/fetch/element/id_utils.test.ts
shir-reifenberg, line 1: maybe I'm missing something but shouldn't this usecase also be covered in `referenced_instance_names` test ?

shir-reifenberg (commented):
🎈 
I have a few questions -
1. what is the behavior when the `extendsParent` is not on the part that defines the file name ? I'm wondering if `extendsParent` should be something we can define once on path and not for each part... WDYT?
2. can we make sure it works as expected when regenerating ids using `referencedInstanceNamesFilterCreator` ?

shir-reifenberg (approved):
Discussed IRL and agreed to allow `extendsParent` not on filename

edenhassid:
> 🎈 I have a few questions -
> 
> 1. what is the behavior when the `extendsParent` is not on the part that defines the file name ? I'm wondering if `extendsParent` should be something we can define once on path and not for each part... WDYT?
> 2. can we make sure it works as expected when regenerating ids using `referencedInstanceNamesFilterCreator` ?

discussed f2f

PR #6117, opened 2024-06-25
SALTO-6150 Fix merge error on custom record type addition in the apply-patch flow

dantal4:
In the apply-patch flow we’re not creating the `internalId` annotation ref on added custom record types, that is a hidden string, and the result is that after the deploy we populate the `internalId` annotation of the custom record type and it is visible and not hidden. Then in the next fetch we fetch the created custom record type, this time with the `internalId` annotation ref, and it makes a merge error because `internalId` is already visible in the nacl.

---

---
_Release Notes_: 
Netsuite Adapter:
- Fix merge error on custom record type addition in the apply-patch flow

---
_User Notifications_: 
None


omrilit (approved):
👑

PR #6118, opened 2024-06-25
SAAS-9608: Returning SF instance URL in client extra info

yelly:
It's good to have even for non sandbox orgs.

---

---
_Release Notes_: 
None.

---
_User Notifications_: 
None.


tamtamirr (approved):
LGTM

PR #6119, opened 2024-06-26
SALTO-6153: Including fields when searching in profiles element fixer

yelly:
In the profiles element fixer we listed the element source which did not give us fields, which led to us evaluating all field permissions as broken references and removing them.

---

---
_Release Notes_: 
_Salesforce_:
* Fixed bug in Salesforce profiles element fixer which caused all field permissions to be dropped.

---
_User Notifications_: 
None.


tamtamirr (approved):
LGTM

PR #6121, opened 2024-06-26
SALTO-5783:  Support deploying Users in Okta

shir-reifenberg:
Add support in deployments of users in Okta, and changed some things in the fetch behavior.

---

Changes to Fetch (no user impact, it's behind a config `exclude`):
- added the `search` query param that fetches deactivated (deprovisioned) users as well.
- removed the `OMIT_CREDENTIALS` header in order to fetch the `provider` fields. credentials fields were added to `omit` + cleaned from any logged data.

Deploy:
- allow addition, modifications and removals of users, and some of the status transitions (we can probably do better here, but I think this is enough for starting).
- Block certain status transitions that are not allowed. 

For context re user statuses se [this doc](https://help.okta.com/en-us/content/topics/users-groups-profiles/usgp-end-user-states.htm) and this nice diagram - 

image

Still TODO - 
- add Users to E2E flows 
- add unit tests 

---
_Release Notes_: 

_Okta_adapter_:
- Support users deployments.

---
_User Notifications_: 
None


packages/okta-adapter/src/definitions/deploy.ts
```diff
@@ -30,7 +31,136 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
...
+              condition: {
+                custom:
+                  () =>
+                  ({ change }) =>
+                    isModificationChange(change) &&
+                    change.data.before.value.status === 'ACTIVE' &&
+                    getChangeData(change).value.status === 'SUSPENDED',
+              },
+            },
+            // unsuspend a user, and change its to status active
```
Nurdok, line 88: -> change its status to active?

(for the next request, too)

packages/okta-adapter/src/definitions/deploy.ts
```diff
@@ -30,7 +31,136 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
...
+                endpoint: { path: '/api/v1/users/{id}/lifecycle/deactivate', method: 'post' },
+              },
+              condition: {
+                custom:
+                  () =>
+                  ({ change }) =>
+                    // user must be in status DEPROVISIONED before it can be deleted
+                    getChangeData(change).value.status !== 'DEPROVISIONED',
+              },
+            },
```
Nurdok, line 153: Just a thought -
This endpoint is duplicated between modify and remove. I think it's fine, but I guess this can also be solved by adding `activate` and `deactivate` actions and using that mechanism. I'm not sure when's the best scenario to use either approach.

Nurdok (approved):

packages/okta-adapter/src/definitions/deploy.ts
```diff
@@ -30,7 +31,136 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
...
+      requestsByAction: {
+        customizations: {
+          add: [
+            {
+              request: { endpoint: { path: '/api/v1/users', method: 'post', queryArgs: { activate: 'false' } } },
+              condition: {
+                custom:
+                  () =>
+                  ({ change }) =>
+                    // To create user in STAGED status, we need to provide 'activate=false' query param
```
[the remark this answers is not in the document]
shir-reifenberg: it works! thanks for that 👑

packages/okta-adapter/src/definitions/deploy.ts
shir-reifenberg: I agree, but since I had more statuses than the usual activate / deactivate i decided to do it this way 😅

PR #6124, opened 2024-06-26
SALTO-5884: [WIP] Microsoft Entra: deploy fixes & improvements

shoshana-houri:

---

---
_Release Notes_: 

---
_User Notifications_: 


packages/microsoft-entra-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -41,34 +41,70 @@ import {
...
   createDefinitionForAppRoleAssignment,
   createDefinitionForGroupLifecyclePolicyGroupModification,
   getGroupLifecyclePolicyGroupModificationRequest,
   omitReadOnlyFields,
+  adjustParentWithAppRoles,
 } from './utils'
 
+// We use filters to handle the deployment of certain types. As a workaround, we define this request,
+// which will fail, to ensure that the CV checking for the deployment definition's existence will pass.
+const FAILURE_REQUEST: DeployableRequestDefinition = {
```
shir-reifenberg, line 64: Can providing `typesWithNoDeploy` to `createCheckDeploymentBasedOnDefinitionsValidator` work instead?

packages/microsoft-entra-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -215,6 +253,17 @@ const graphV1CustomDefinitions: DeployCustomDefinitions = {
       },
     },
   },
+  [APP_ROLE_TYPE_NAME]: {
+    changeGroupId: deployment.grouping.groupWithFirstParent,
```
shir-reifenberg, line 257: Just verifying there's no scenario when the first has a reference to the service principal and the second to the associated application

packages/microsoft-entra-adapter/src/definitions/deploy/utils/app_role.ts
```diff
@@ -0,0 +1,41 @@
...
+import { adjustWrapper } from './read_only_fields'
+
+/*
+ * Adjusts the appRoles field in the parent object to:
+ * 1. Add an id to each appRole that does not have one
+ * 2. Remove the parent_id field from each appRole (we manually add this field during fetch and it is not deployable)
+ */
+export const adjustParentWithAppRoles: AdjustFunction = adjustWrapper(async ({ value, typeName }) => {
+  validatePlainObject(value, typeName)
+  const adjustedAppRoles = _.get(value, APP_ROLES_FIELD_NAME, []).map((appRole: Record<string, unknown>) => ({
```
shir-reifenberg, line 31: please validate the type here cause type is `any`

packages/microsoft-entra-adapter/src/definitions/deploy/utils/app_role.ts
```diff
@@ -0,0 +1,41 @@
...
+    id: uuid4(),
+    ..._.omit(appRole, PARENT_ID_FIELD_NAME),
+  }))
+  return {
+    value: {
+      ...value,
+      ...(_.isEmpty(adjustedAppRoles) ? {} : { appRoles: adjustedAppRoles }),
```
shir-reifenberg, line 38: nit 
```suggestion
      ...(_.isEmpty(adjustedAppRoles) ? {} : { [APP_ROLES_FIELD_NAME]: adjustedAppRoles }),
```

packages/microsoft-entra-adapter/src/definitions/deploy/utils/app_role.ts
```diff
@@ -0,0 +1,41 @@
...
+
+/*
+ * Adjusts the appRoles field in the parent object to:
+ * 1. Add an id to each appRole that does not have one
+ * 2. Remove the parent_id field from each appRole (we manually add this field during fetch and it is not deployable)
+ */
+export const adjustParentWithAppRoles: AdjustFunction = adjustWrapper(async ({ value, typeName }) => {
+  validatePlainObject(value, typeName)
+  const adjustedAppRoles = _.get(value, APP_ROLES_FIELD_NAME, []).map((appRole: Record<string, unknown>) => ({
+    id: uuid4(),
```
shir-reifenberg, line 32: why do we need this one ?

packages/microsoft-entra-adapter/src/definitions/deploy/utils/condition_empty_fields.ts
```diff
@@ -29,7 +29,7 @@ export const createCustomConditionEmptyFieldsOnAddition = (
   custom:
     () =>
     ({ change }) => {
-      if (!isAdditionChange(change)) {
+      if (isRemovalChange(change)) {
```
shir-reifenberg, line 32: please update the name of `createCustomConditionEmptyFieldsOnAddition` cause it's now also applies to modifications? (also the documentation needs some adjustment)

packages/microsoft-entra-adapter/src/definitions/deploy/utils/conditional_access_policy.ts
```diff
@@ -0,0 +1,49 @@
...
+    value: {
+      ...value,
+      conditions: {
+        ..._.get(value, 'conditions', {}),
+        users: {
+          ..._.get(value, 'conditions.users', {}),
+          // We store the includeUsers field only on specific conditions.
+          // However, this field is required on addition, so we set it to a default of 'none' when it's not specified.
+          includeUsers: _.get(value, 'conditions.users.includeUsers') ?? ['none'],
+        },
```
shir-reifenberg, line 45: what happened if the user has something else there? do we just override it ?
should it maybe be in a fixer and not during deployment ?

Also, maybe worth using `getPath` and `setPath` (using the the elemID in `change`)

packages/microsoft-entra-adapter/src/definitions/deploy/utils/read_only_fields.ts
```diff
@@ -16,24 +16,39 @@
...
   if (readOnlyFieldsToOmit === undefined) {
     return { value }
   }
   return { value: _.omit(value, readOnlyFieldsToOmit) }
 }
+
+/*
+ * Adjust the value of the object using the provided adjust function, and then omit read-only fields from the result.
+ */
+export const adjustWrapper: (adjust: AdjustFunction) => AdjustFunction = adjust => async args => {
```
shir-reifenberg, line 47: I don't really see a reason that the `adjust` will override any other defined functions, what you added here seems like a pretty common usecase to me 🤔 
@netama is there a specific reason that `adjust` overrides any other functions definitions ?

packages/microsoft-entra-adapter/src/filters/app_role.ts
```diff
@@ -0,0 +1,235 @@
...
+ * We set the app roles field in the application/service principal as top level in order to be able to reference it.
+ * However, it creates a problem both in the fetch and deploy phases:
+ * - In the fetch phase, the app roles can be retrieved either from the application or from the service principal, with the same values.
+ *   We cannot filter them in advance (i.e. fetch only the application's/SP's app roles), since the application/SP might not be fetched at all.
+ *   Therefore, we remove the duplicates after fetching all the elements.
+ * - In the deploy phase, we need to deploy the app roles as part of the application/SP. They don't have a separate API call.
+ */
+export const appRolesFilter: FilterCreator = ({ definitions, elementSource, sharedContext }) => ({
+  name: 'appRolesFilter',
+  onFetch: async (elements: Element[]): Promise<void> => {
```
shir-reifenberg, line 52:
```suggestion
  onFetch: async elements => {
```

packages/microsoft-entra-adapter/src/filters/app_role.ts
```diff
@@ -0,0 +1,235 @@
...
+        return appRoles.filter((_appRole, index) => index !== appRoleWithApplicationParentIdx)
+      })
+
+    log.trace(
+      'Removing the following app roles: %s',
+      appRolesToRemove.map(({ elem }) => elem.elemID.getFullName()).join(', '),
+    )
+    // We remove the app roles in reverse order to avoid changing the indices of the elements we still need to remove
+    const indicesToRemove = appRolesToRemove.map(({ index }) => index).sort((a, b) => b - a)
+    indicesToRemove.forEach(index => elements.splice(index, 1))
```
shir-reifenberg, line 81: nit: I think you can just use `_.pullAll`

packages/microsoft-entra-adapter/src/filters/app_role.ts
```diff
@@ -0,0 +1,235 @@
...
+            {
+              severity: 'Error',
+              message: 'Deploy not supported',
+            },
+          ],
+        },
+        leftoverChanges: changes,
+      }
+    }
+    if (changeGroup === undefined) {
```
shir-reifenberg, line 100: is it really needed ?

packages/microsoft-entra-adapter/src/filters/app_role.ts
```diff
@@ -0,0 +1,235 @@
...
+      changes,
+      change => isInstanceChange(change) && getChangeData(change).elemID.typeName === APP_ROLE_TYPE_NAME,
+    ) as [Change<InstanceElement>[], Change[]]
+
+    // We use 'fromEntries' as a "trick" to save each parent only once, without actually comparing the objects
+    const parentsWithAppRoleChanges = Object.fromEntries(
+      appRoleChanges
+        .map((change): [string, InstanceElement] | undefined => {
+          const parent = getParents(getChangeData(change))[0]?.value
+          return parent ? [parent.elemID.getFullName(), parent] : undefined
```
shir-reifenberg, line 126: what will happen if there is no parent ?

packages/microsoft-entra-adapter/src/filters/app_role.ts
```diff
@@ -0,0 +1,235 @@
...
+        })
+        .filter(isDefined),
+    )
+    const allAppRoleInstancesWithParentName = awu(await elementSource.getAll())
```
shir-reifenberg, line 130: when possible, there's a preference to use elementSource.list() instead of `getAll()` (which loads all elements) and then use multiple `get` calls (in case you want a specific type)

but you can also use `getInstancesFromElementSource` instead

shir-reifenberg (commented):
Great work handling of the use cases 🎉  
Didn't review anything yet, but added some comments (will continue later)
Also, if it's not too much effort at this point, I think it would help splitting this PR to the different issues it addresses 😅

packages/microsoft-entra-adapter/src/definitions/fetch/utils/conditional_access_policy.ts
```diff
@@ -0,0 +1,48 @@
...
+  CONDITIONAL_ACCESS_POLICY_TYPE_NAME,
+  EXCLUDE_USERS_FIELD_NAME,
+  INCLUDE_USERS_FIELD_NAME,
+} from '../../../constants'
+
+/*
+ * Adjust the conditional access policy object to include the includeUsers field and excludeUsers fields
+ * only if they do not include user ids, which we don't handle.
+ */
+export const adjustConditionalAccessPolicy: AdjustFunction = async ({ value }) => {
```
shir-reifenberg, line 30: IMO it's ok to leave it as is and handle it in a different PR / ticket cause I'm not sure this is the best approach and ultimately we'd want to change it

packages/microsoft-entra-adapter/src/definitions/fetch/utils/application.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+export const adjustApplication: AdjustFunction = async ({ value }) => {
+  validatePlainObject(value, 'application')
+  const identifierUris = _.get(value, IDENTIFIER_URIS_FIELD_NAME, [])
+  validateArray(identifierUris, IDENTIFIER_URIS_FIELD_NAME)
+
+  return {
+    value: {
+      ...value,
+      identifierUris: identifierUris.filter(uri => uri !== `api://${_.get(value, 'appId')}`),
```
shir-reifenberg, line 36: The reason it's not deployable is that it includes the `appId`? if so, I think this can be converted into `TemplateExpression` (string combined with a reference)

packages/microsoft-entra-adapter/src/definitions/fetch/fetch.ts
```diff
@@ -326,7 +347,24 @@ const graphV1Customizations: FetchCustomizations = {
...
           },
         },
+        [APP_ROLES_FIELD_NAME]: APP_ROLES_FIELD_CUSTOMIZATIONS,
+      },
+    },
+  },
+  [APP_ROLE_TYPE_NAME]: {
+    resource: {
+      directFetch: false,
+      serviceIDFields: [PARENT_ID_FIELD_NAME, 'id'],
```
shir-reifenberg, line 357: please ignore my previous comment about `parent_id`, I get it now :)

packages/microsoft-entra-adapter/src/filters/app_role.ts
```diff
@@ -0,0 +1,235 @@
...
+          const applyFunction = async (instance: InstanceElement): Promise<InstanceElement> => {
+            const instanceDup = instance.clone()
+            instanceDup.value[APP_ROLES_FIELD_NAME] = await allAppRoleInstancesWithParentName
+              .filter(({ parentName }) => parentName === parentFullName)
+              .map(({ elem }) => elem.value)
+              .toArray()
+            return instanceDup
+          }
+
+          if (isAdditionChange(parentChange)) {
```
shir-reifenberg, line 176: I think you'll want to use `applyFunctionToChangeData` ?

packages/microsoft-entra-adapter/src/filters/app_role.ts
```diff
@@ -0,0 +1,235 @@
...
+        elem,
+        parentName: getParents(elem)[0]?.elemID.getFullName(),
+      }))
+    const parentsOriginalChange: Change[] = []
+    const parentsChangeWithAppRoles = await Promise.all(
+      Object.entries(parentsWithAppRoleChanges).map(
+        async ([parentFullName, parent]): Promise<
+          AdditionChange<InstanceElement> | ModificationChange<InstanceElement> | undefined
+        > => {
+          const parentExistingChangeIdx = otherChanges.findIndex(
```
shir-reifenberg, line 142: why do we cate about the parent changes ?

shir-reifenberg (commented):
Done reviewing now :) added some more comments
Can you please explain why app_role filter is using the parent change? shouldn't we just collect all app roles and deploy them as is? i'm probably missing bunch of stuff here so maybe worth discussing IRL :)

packages/microsoft-entra-adapter/src/definitions/deploy/deploy.ts
shoshana-houri: Oohh I checked and looks like `typesDeployedViaParent` arg is the one we want here, thanks!
However, it's currently impossible to provide this arg when creating the adapter with the new infra. Will open a ticket and implement it separately

packages/microsoft-entra-adapter/src/definitions/deploy/deploy.ts
shoshana-houri: It shouldn't be. If it then is it means we have a bug in the `onFetch` part in our filter.

packages/microsoft-entra-adapter/src/definitions/deploy/utils/app_role.ts
shoshana-houri: App roles must be assigned a unique id when first created. If they already have an id it will be overridden by the next line

packages/microsoft-entra-adapter/src/definitions/deploy/utils/conditional_access_policy.ts
shoshana-houri: It's only relevant for addition changes, so we don't exactly override it. I would be happy to discuss the fixer option.
Also, I'm not so sure how `getPath` and `setPath` come in handy here, I need the path for a specific field inside the change 🤔

packages/microsoft-entra-adapter/src/definitions/deploy/utils/read_only_fields.ts
shoshana-houri: Discussed f2f - it doesn't override other functions definitions, only the same adjust definition

packages/microsoft-entra-adapter/src/filters/app_role.ts
shoshana-houri: Sadly, yes. When calling `deployChanges` we need to call it with a valid `changeGroup`

packages/microsoft-entra-adapter/src/filters/app_role.ts
shoshana-houri: ~~We'll skip the deployment of this app role and it will later throw an error when trying to deploy it with the default definitions and not via the filter~~
It will actually throw an error in line 230 when trying to get the parent using getParent (and not getParents). I think we should actually return a more user friendly error here if we can't find the parent. Will update the code, thanks!

packages/microsoft-entra-adapter/src/definitions/fetch/constants.ts
```diff
@@ -47,12 +50,6 @@ export const DEFAULT_FIELD_CUSTOMIZATIONS: Record<string, ElementFieldCustomizat
   expirationDateTime: {
     omit: true,
   },
-  includeUsers: {
-    omit: true,
-  },
-  excludeUsers: {
-    omit: true,
-  },
```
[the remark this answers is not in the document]
shoshana-houri: Good point, created https://salto-io.atlassian.net/browse/SALTO-6331 :)

packages/microsoft-entra-adapter/src/definitions/deploy/utils/conditional_access_policy.ts
shoshana-houri: Will remove this logic from this PR. Created a dedicated ticket https://github.com/salto-io/salto/pull/6124#discussion_r1685949099

shoshana-houri: Will remove this logic from this PR. Created a dedicated ticket https://salto-io.atlassian.net/browse/SALTO-6332

packages/microsoft-entra-adapter/src/definitions/fetch/utils/conditional_access_policy.ts
shoshana-houri: Will remove it from this PR. Opened a dedicated ticket https://salto-io.atlassian.net/browse/SALTO-6332 :)

packages/microsoft-entra-adapter/src/definitions/fetch/utils/application.ts
shoshana-houri: Cool! I opened a dedicated ticket https://salto-io.atlassian.net/browse/SALTO-6333
Let's leave it as it's for now, to avoid failing the deployment, and handle it separately afterwards?

packages/microsoft-entra-adapter/src/filters/app_role.ts
shoshana-houri: Talked f2f - the problem is that this function modified the change both before and after. Instead, @shir-reifenberg suggested directly modifying the change `change.data.after = ....` which seems to work :)

packages/microsoft-entra-adapter/src/filters/app_role.ts
shoshana-houri: Talked f2f - the app role changes are deployed as a **field** on the parent, and they have the same group Id as their parent.

packages/microsoft-entra-adapter/src/filters/app_role.ts
shoshana-houri: Giving it another though - I'm afraid that modifying the change directly will cause the deployment to be marked as partial failure (applied changes differ from originals)

packages/microsoft-entra-adapter/src/filters/app_role.ts
```diff
@@ -0,0 +1,333 @@
...
+              severity: 'Error',
+              message: 'Deploy not supported',
+            },
+          ],
+        },
+        leftoverChanges: changes,
+      }
+    }
+
+    const deployChangesFunc: CustomDeployChangesFunc = ({
```
shir-reifenberg, line 312: why is this defined here and passed to `deployAppRoleChangesViaParent` instead of just calling 
```
deployment.deployChanges({
        changes: adjustedChanges,
        changeGroup: adjustedChangeGroup,
        definitions: { deploy, ...otherDefs },
        elementSource,
        convertError: customConvertError,
        changeResolver,
        sharedContext,
      })
```

inside?

shir-reifenberg (approved):
Great work 🚀 I added some suggestions that I think might simplify the code a bit (in app roles filter)
otherwise LGTM 💪

packages/microsoft-entra-adapter/test/mocks.ts
```diff
@@ -27,17 +39,46 @@ export const instanceElementMock = new InstanceElement('testInstance', objectTyp
...
-export const modificationChangeMock: Change<InstanceElement> = toChange({
-  before: instanceElementMock,
-  after: instanceElementMock,
+export const objectTypeElementMock = new ObjectType({
+  elemID: new ElemID(ADAPTER_NAME, 'testType'),
 })
 
-export const removalChangeMock: Change<InstanceElement> = toChange({
-  before: instanceElementMock,
+export const objectTypeChangeMock: Change<ObjectType> = toChange({
```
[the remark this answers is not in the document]
shoshana-houri: I wanted to test that we handle correctly changes that are not instance changes

packages/microsoft-entra-adapter/src/filters/app_role.ts
```diff
@@ -0,0 +1,333 @@
...
+  const parentToAppRolesMap = await groupAppRolesByParent(elementSource)
+  const getAppRolesByFullName = (parentFullName: string): Values[] =>
+    parentToAppRolesMap[parentFullName].map(appRole => appRole.value)
+
+  const parentChanges = (
+    await Promise.all(
+      uniqueParents.map(parent =>
+        setOrCreateParentChangeWithAppRoles({
+          parent,
+          getAppRolesByFullName,
```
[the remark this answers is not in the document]
shoshana-houri: I thought it's redundant passing the map itself since what we really need is this ability, but I don't mind changing back if it's confusing :)

packages/microsoft-entra-adapter/src/filters/app_role.ts
```diff
@@ -0,0 +1,333 @@
...
+  | undefined
+> => {
+  const parentFullName = parent.elemID.getFullName()
+  const existingParentChangeIdx = otherChanges.findIndex(
+    change => getChangeData(change).elemID.getFullName() === parentFullName,
+  )
+  const existingParentChange = existingParentChangeIdx !== -1 ? otherChanges[existingParentChangeIdx] : undefined
+  if (existingParentChange !== undefined) {
+    // We shouldn't really reach this point, as we validate the parent in 'extractUniqueParentsFromAppRoleChanges', but just for TS to be happy
+    if (!isInstanceChange(existingParentChange)) {
```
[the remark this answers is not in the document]
shoshana-houri: But they're not necessarily `Change<InstanceElement>` 🥲  It's all the changes originally received, which are not app role changes

packages/microsoft-entra-adapter/src/filters/app_role.ts
```diff
@@ -0,0 +1,333 @@
...
+      log.error('Parent %s is not an instance change, skipping its app roles deployment', parentFullName)
+      return undefined
+    }
+
+    otherChanges.splice(existingParentChangeIdx, 1)
```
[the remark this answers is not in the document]
shoshana-houri: A significant part of this function's logic is to determine if we want to use the parent's change or not, so I think we shouldn't filter them in advance. But, I agree that it can be a surprising side effect, so I'll change the function to return the index, and then we'll change the array from outside.

packages/microsoft-entra-adapter/src/filters/app_role.ts
```diff
@@ -0,0 +1,333 @@
...
+          otherChanges,
+        }),
+      ),
+    )
+  ).filter(isDefined)
+
+  const changesToDeploy = parentChanges.map(({ adjustedParentChange }) => adjustedParentChange)
+  const existingParentChanges = parentChanges.map(({ existingParentChange }) => existingParentChange).filter(isDefined)
+
+  const calculateResult = async ({
```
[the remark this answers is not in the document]
shoshana-houri: We do deploy all parent changes, but we need to further calculate the result for two reasons:
1. If the parent doesn't have other changes, it means we created a new one, which should not be returned as an applied change.
2. We have changes on app roles that may or may not succeed. The way to determine this is by checking if their matching parent change succeeded.

packages/microsoft-entra-adapter/src/filters/app_role.ts
```diff
@@ -0,0 +1,333 @@
...
+      deployResult: {
+        errors: errors.concat(appRoleWithNoParentErrors),
+        appliedChanges: appliedExistingParentChanges.concat(appliedAppRoleChanges),
+      },
+      leftoverChanges: otherChanges,
+    }
+  }
+
+  if (_.isEmpty(changesToDeploy)) {
+    return calculateResult({ errors: [], appliedChanges: [] })
```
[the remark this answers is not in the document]
shoshana-houri: Not exactly, because we want to return the errors we created for `appRoleWithNoParentErrors`. `calculateResult` unites all of that

packages/microsoft-entra-adapter/src/filters/app_role.ts
shoshana-houri: Just nit reasons - This will require us to pass the `definitions`, the `elementsSource`, the `sharedContext`... this way we only pass 1 arg :)

PR #6127, opened 2024-06-27
SALTO-6165_Improve_referenced_instance_name_filter_efficiency

omrifarkash:
Improving running time for referenced instance name filter

---

findIndex is O(n) and we are doing it in the worst case n times. this make the filter running time O(n^2)
adding actual data from fetch on 42K instances (most of them are from type `page` in confluence, which has a reference in its element name)

before:
"(referencedInstanceNames):onFetch took 32647 ms"
after:
"(referencedInstanceNames):onFetch took 7697 ms"

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/filters/referenced_instance_names.ts
+++ b/packages/adapter-components/src/filters/referenced_instance_names.ts
@@ -207,9 +207,9 @@ const updateAllReferences = ({
 
 /* Create a graph with instance names as nodes and instance name dependencies as edges */
 const createGraph = <TOptions extends APIDefinitionsOptions = {}>(
-  instances: InstanceElement[],
-  fetchDefinitionInstanceName: Record<string, InstanceFetchApiDefinitions<TOptions>>,
-): DAG<InstanceElement> => {
+  instances: Element[],
+  fetchDefinitionByType: Record<string, InstanceFetchApiDefinitions<TOptions>>,
+): DAG<{ instance: InstanceElement; index: number }> => {
   const duplicateElemIds = new Set(findDuplicates(instances.map(i => i.elemID.getFullName())))
   const duplicateIdsToLog = new Set<string>()
   const isDuplicateInstance = (instanceFullName: string): boolean => {
@@ -220,9 +220,12 @@ const createGraph = <TOptions extends APIDefinitionsOptions = {}>(
     return false
   }
 
-  const graph = new DAG<InstanceElement>()
-  instances.forEach(instance => {
-    const { elemID } = fetchDefinitionInstanceName[instance.elemID.getFullName()]?.element?.topLevel ?? {}
+  const graph = new DAG<{ instance: InstanceElement; index: number }>()
+  instances.forEach((instance, index) => {
+    if (!isInstanceElement(instance)) {
+      return
+    }
+    const { elemID } = fetchDefinitionByType[instance.elemID.typeName]?.element?.topLevel ?? {}
     const parts = elemID?.parts ?? []
     const extendsParent = elemID?.extendsParent
     if (extendsParent || parts.some(part => part.isReference)) {
@@ -232,7 +235,7 @@ const createGraph = <TOptions extends APIDefinitionsOptions = {}>(
         const nameDependencies = getInstanceNameDependencies(instance, parts, extendsParent).filter(
           instanceName => !isDuplicateInstance(instanceName),
         )
-        graph.addNode(instance.elemID.getFullName(), nameDependencies, instance)
+        graph.addNode(instance.elemID.getFullName(), nameDependencies, { instance, index })
       }
     }
   })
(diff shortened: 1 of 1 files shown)


packages/adapter-components/src/filters/referenced_instance_names.ts
```diff
@@ -339,9 +340,13 @@ export const addReferencesToInstanceNames = async <TOptions extends APIDefinitio
...
-
-      const originalInstanceIdx = elements.findIndex(e => e.elemID.getFullName() === originalFullName)
-      elements.splice(originalInstanceIdx, 1, newInstance)
+      elementsToReplaceIndex[originalFullName] = newInstance
+    }
+  })
+  elements.forEach((elem, index) => {
+    const elemFullName = elem.elemID.getFullName()
+    if (elementsToReplaceIndex[elemFullName] !== undefined) {
+      elements.splice(index, 1, elementsToReplaceIndex[elemFullName])
```
shir-reifenberg, line 349: I _think_ there might be an issue with updating `elements` at the end of the graph iteration instead of in-place.

consider the scenario when we have a reference from `A->B`, and both elements are renamed. 
then we create `A'`, but in `elements` we still have `A`. then we create `B`, but when we update all references to it, we actually update `A` to point `B'`, instead of `A'` to point `B'` 

please verify if this indeed a bug here 😅  I would expect it to be covered in tests 🤔

shir-reifenberg (commented):
I think there might be scenario in which this can be problematic, please see my comment 🙏

packages/adapter-components/src/filters/referenced_instance_names.ts
omrifarkash: So we need to use a data structure that will save the index per element, I don't think we should iterate over all elements to extract the index every time we want to replace one element.

Also it is probably not covered by UT as all tests passed, I'll add a test for it

packages/adapter-components/src/filters/referenced_instance_names.ts
```diff
@@ -207,9 +207,9 @@ const updateAllReferences = ({
 
 /* Create a graph with instance names as nodes and instance name dependencies as edges */
 const createGraph = <TOptions extends APIDefinitionsOptions = {}>(
-  instances: InstanceElement[],
-  fetchDefinitionInstanceName: Record<string, InstanceFetchApiDefinitions<TOptions>>,
-): DAG<InstanceElement> => {
+  instances: Element[],
+  fetchDefinitionByType: Record<string, InstanceFetchApiDefinitions<TOptions>>,
+): DAG<{ instance: InstanceElement; index: number }> => {
```
omrifarkash, line 212: Data of each node is the instance itself (as it was) plus its original index

packages/adapter-components/src/filters/referenced_instance_names.ts
```diff
@@ -207,10 +207,10 @@ const updateAllReferences = ({
 
 /* Create a graph with instance names as nodes and instance name dependencies as edges */
 const createGraph = <TOptions extends APIDefinitionsOptions = {}>(
-  instances: InstanceElement[],
-  fetchDefinitionInstanceName: Record<string, InstanceFetchApiDefinitions<TOptions>>,
-): DAG<InstanceElement> => {
-  const duplicateElemIds = new Set(findDuplicates(instances.map(i => i.elemID.getFullName())))
+  elements: Element[],
```
omrifarkash, line 210: Passing all elements and not just instances to keep the original index

packages/adapter-components/src/filters/referenced_instance_names.ts
```diff
@@ -220,9 +220,12 @@ const createGraph = <TOptions extends APIDefinitionsOptions = {}>(
...
 
-  const graph = new DAG<InstanceElement>()
-  instances.forEach(instance => {
-    const { elemID } = fetchDefinitionInstanceName[instance.elemID.getFullName()]?.element?.topLevel ?? {}
+  const graph = new DAG<{ instance: InstanceElement; index: number }>()
+  elements.forEach((instance, index) => {
+    if (!isInstanceElement(instance)) {
+      return
+    }
+    const { elemID } = fetchDefinitionByType[instance.elemID.typeName]?.element?.topLevel ?? {}
```
omrifarkash, line 228: No need definition per instance name. definition per type is the same and is much smaller

packages/adapter-components/src/filters/referenced_instance_names.ts
```diff
@@ -207,10 +207,10 @@ const updateAllReferences = ({
...
 /* Create a graph with instance names as nodes and instance name dependencies as edges */
 const createGraph = <TOptions extends APIDefinitionsOptions = {}>(
-  instances: InstanceElement[],
-  fetchDefinitionInstanceName: Record<string, InstanceFetchApiDefinitions<TOptions>>,
-): DAG<InstanceElement> => {
-  const duplicateElemIds = new Set(findDuplicates(instances.map(i => i.elemID.getFullName())))
+  elements: Element[],
+  fetchDefinitionByType: Record<string, InstanceFetchApiDefinitions<TOptions>>,
+): DAG<{ instance: InstanceElement; index: number }> => {
+  const duplicateElemIds = new Set(findDuplicates(elements.map(i => i.elemID.getFullName())))
```
shir-reifenberg, line 213: nit: no reason to look for dups in object types right ? should we filter to instance elements ?

shir-reifenberg (approved):
Looks great, thanks for that 👑 👑 👑 
can you please add a test to cover the scenario we discussed ? (elements name were updated but reference points to the old instance)

PR #6128, opened 2024-06-27
SALTO-5323 dump order changes correctly

dantal4:
Make `updateNaclFiles` be able to dump order changes in lists. This is a preparation for calculating order changes in lists, in order to auto merge semantic changes in lists.
Currently, we'll calculate order changes only when the env var `SALTO_COMPARE_LIST_ITEMS` is passed.

---

---
_Release Notes_: 
Workspace:
- Dump order changes correctly

Core:
- Calculate order changes when the SALTO_COMPARE_LIST_ITEMS env var is passed

---
_User Notifications_: 
None


packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -122,71 +192,76 @@ export const getChangeLocations = (
...
+    if (possibleLocations.length > 0) {
+      const foundInPath = possibleLocations.find(sr => sr.filename === fileName)
+      // When adding a nested change we need to increase one level of indentation because
+      // we get the placement of the closing brace of the next line. The closing brace will
+      // be indented one line less then wanted change.
+      // TODO: figure out how to choose the correct location if there is more than one option
+      return [
+        {
+          ...change,
+          location: { ...lastNestedLocation(foundInPath ?? possibleLocations[0]), indexInParent, newInParent: true },
```
yelly, line 245: This doesn't necessarily mean that the value is new in the parent, it could just be last or even just last in the file.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -281,40 +356,50 @@ export const updateNaclFileData = async (
...
-    let indentationLevel = (change.location.start.col - 1) / 2
-    if (change.requiresIndent) {
-      indentationLevel += 1
-    }
-    if (elem !== undefined) {
-      const changeKey = change.id.name
-      const isListElement = changeKey.match(/^\d+$/) !== null
+    const changeKey = change.id.name
+    const isListElement =
+      !change.id.isAnnotationTypeID() && !change.id.createParentID().isBaseID() && isIndexPathPart(changeKey)
```
yelly, line 361: Are the first two cases here even legal? Feels like they shouldn't be...

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -281,40 +356,50 @@ export const updateNaclFileData = async (
...
+    const initialIndentationLevel = change.location.start.col - 1
```
yelly, line 362: This isn't the indentation level, it's the indentation width (we divide by 2 to get the level because each indentation is 2 spaces).

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -281,40 +356,50 @@ export const updateNaclFileData = async (
...
+      const dumpedObj = await parser.dumpValues({ [changeKey]: elem }, functions, indentationLevel - 1)
+      // once we have the "new scope", we want to take just the serialized values because the
+      // brackets already exist in the original scope.
+      return removeBracketLines(dumpedObj)
     }
+
+    const getNewData = async (): Promise<string> => {
+      if (isListElement) {
+        if (elem === undefined) {
+          return `\n${parser.createIndentation(indentationLevel)}`
```
yelly, line 387: Why do we need to add an extra newline when removing a list item? Don't we have the opposite problem of needing to remove a newline?

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -98,11 +103,76 @@ const getPositionInParent = <T>(change: DetailedChange<T> & AdditionChange<T>):
...
+  if (afterContainer.length >= beforeContainer.length) {
+    return [change]
+  }
+
+  const oldLocationIndex = Number(oldLocationChangeId.name)
+  const newLocationIndex = Number(newLocationChangeId.name)
+  const oldLastItemIndex = beforeContainer.length - 1
+  const newLastItemIndex = afterContainer.length - 1
+
+  // we create a change for the last item with a modified location
```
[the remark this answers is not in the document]
dantal4: writing a better comment

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
dantal4: you're correct, but there's no validation that a field name, non-nested value in element, etc. are not just a number.
the current code "deals" with it because we first check `change.id.isAnnotationTypeID()` then `isElement(elem)` only then `isListElement`....

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
dantal4: I'm not exactly sure why, but otherwise the result is (if `3` was removed):
```
list = [
  1,
  2,4
]
```

packages/netsuite-adapter/src/adapter.ts
```diff
@@ -553,6 +553,14 @@ export default class NetsuiteAdapter implements AdapterOperations {
...
   @logDuration('fetching account configuration')
   public async fetch({ progressReporter, withChangesDetection = false }: FetchOptions): Promise<FetchResult> {
+    if (process.env.DUMMY_DATA !== undefined) {
+      const dummy = new ObjectType({
+        elemID: new ElemID('netsuite', 'dummy'),
+        annotations: JSON.parse(process.env.DUMMY_DATA),
+        path: ['netsuite', 'dummy'],
+      })
+      return { elements: [dummy], partialFetchData: { isPartial: true } }
+    }
```
dantal4, line 563: please ignore in the review

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
yelly: This is confusing me but if that's the behaviour you are seeing then it's ok.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -98,95 +102,240 @@ const getPositionInParent = <T>(change: DetailedChange<T> & AdditionChange<T>):
...
-      line: parentScope.end.line,
-      col: parentScope.start.col,
-      byte: parentScope.end.byte - 1,
+  const { baseChange } = change
+  if (baseChange === undefined) {
+    log.warn('No base change: %s', inspectValue(change))
+    return [change]
+  }
+
+  if (baseChange.action !== 'modify') {
```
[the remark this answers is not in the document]
dantal4: in addition/removal the baseChange is the element addition/removal itself

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -98,95 +102,240 @@ const getPositionInParent = <T>(change: DetailedChange<T> & AdditionChange<T>):
...
+    return [change]
+  }
+
+  const beforeLocationChangeId = change.elemIDs?.before ?? change.id
+  const afterLocationChangeId = change.elemIDs?.after ?? change.id
```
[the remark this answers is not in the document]
dantal4: 1. If `SALTO_COMPARE_LIST_ITEMS` is disabled, and then there won't be a change inside a list if the list length was modified
2. If it's not a change inside list

in both cases we can return if `elemIDs` is undefined 👍

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -98,95 +102,240 @@ const getPositionInParent = <T>(change: DetailedChange<T> & AdditionChange<T>):
...
+        log.warn('missing location for %s', afterContainerChangeId.getFullName())
+        return [change]
+      }
+      return [
+        {
+          id: afterContainerChangeId,
+          action: 'modify',
+          data: {
+            before: beforeContainer,
+            after: afterContainer,
```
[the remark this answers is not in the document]
dantal4: I think that I prefer using the value of `afterContainer`, that is the value from `resolvePath` on the parent id

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -98,95 +105,254 @@ const getPositionInParent = <T>(change: DetailedChange<T> & AdditionChange<T>):
...
-        // TODO: figure out how to choose the correct location if there is more than one option
-        return [{ location: lastNestedLocation(foundInPath ?? possibleLocations[0]), requiresIndent: true }]
-      }
-      log.error('No possible locations found for %s.', parentID.getFullName())
+  if (change.elemIDs === undefined) {
+    return [change]
+  }
+
+  const beforeChangeId = change.elemIDs.before
+  const afterChangeId = change.elemIDs.after
```
yelly, line 132:
```suggestion
  const { before: beforeChangeId, after: afterChangeId } = change.elemIDs
```

packages/netsuite-adapter/src/adapter.ts
yelly: Do you want to delete this?

yelly (commented):
Just a couple last comments, apart from that LGTM.
Maybe just get final approval from @ori-moisis, I might have missed some delicate edge case here.

dantal4:
Currently we went in the different direction that doesn't require the ability to dump order changes.

PR #6129, opened 2024-06-27
SALTO-5432 enable merging overriding system definitions from env var

Uri-Bar:
SALTO-5432 enable merging overriding system definitions from env var
---

---
_Release Notes_: 
Added the ability to override definitions with the `DEFINITION_OVERRIDES` env var

---
_User Notifications_: 
none

### DIFF
--- a/packages/adapter-components/src/definitions/system/overrides.ts
+++ b/packages/adapter-components/src/definitions/system/overrides.ts
@@ -20,29 +20,35 @@ import { RequiredDefinitions } from './types'
 import { APIDefinitionsOptions } from './api'
 
 const log = logger(module)
-export const DEFINITIONS_OVERRIDES = 'SALTO_DEFINITION_OVERRIDES'
+export const DEFINITIONS_OVERRIDES = 'SALTO_DEFINITIONS_OVERRIDES'
 
 export const getParsedDefinitionsOverrides = (): Values => {
-  const flagValue = process.env[DEFINITIONS_OVERRIDES]
-  let parsedFlagValue: unknown
+  const overrides = process.env[DEFINITIONS_OVERRIDES]
   try {
-    parsedFlagValue = flagValue === undefined ? undefined : JSON.parse(flagValue)
+    const parsedOverrides = overrides === undefined ? undefined : JSON.parse(overrides)
+    if (parsedOverrides !== undefined && typeof parsedOverrides === 'object') {
+      return parsedOverrides as Values
+    }
   } catch (e) {
     if (e instanceof SyntaxError) {
-      log.error('There was a syntax error in the JSON while parsing a flag:', e.message)
+      log.error('There was a syntax error in the JSON while parsing the overrides:', e.message)
     } else {
-      log.error('An unknown error occurred while parsing a flag:', e)
+      log.error('An unknown error occurred while parsing the overrides:', e)
     }
-  }
-  if (parsedFlagValue !== undefined && typeof parsedFlagValue === 'object') {
-    return parsedFlagValue as Values
   }
   return {}
 }
 
+/**
+ * merge definitions with overrides from the environment variable SALTO_DEFINITIONS_OVERRIDES
+ * the merge is done as follows:
+ * - overrides takes precedence over definitions
+ * - when merging an array the overrides array is used instead of the definitions array completely (no merge)
+ * - when merging an object the overrides object is merged with the definitions object recursively
+ * - we can use null to remove a field from the definitions as we override the original value with null and then remove it
+ */
 export const mergeDefinitionsWithOverrides = <Options extends APIDefinitionsOptions>(
   definitions: RequiredDefinitions<Options>,
-  overrides: Values,
 ): RequiredDefinitions<Options> => {
   const customMerge = (objValue: Value, srcValue: Value): Value => {
     if (_.isArray(objValue)) {
@@ -53,10 +59,12 @@ export const mergeDefinitionsWithOverrides = <Options extends APIDefinitionsOpti
     }
     return srcValue
   }
+  const overrides = getParsedDefinitionsOverrides()
   if (_.isEmpty(overrides)) {
     return definitions
   }
-  const merged = _.mergeWith(definitions, overrides, customMerge)
+  const cloneDefinitions = _.cloneDeep(definitions)
+  const merged = _.mergeWith(cloneDefinitions, overrides, customMerge)
   const removeNullObjects = (obj: Value): Value => {
     if (_.isArray(obj)) {
       return obj.map(removeNullObjects)
--- a/packages/adapter-components/src/definitions/system/overrides.ts
+++ b/packages/adapter-components/src/definitions/system/overrides.ts
@@ -16,22 +16,14 @@
 import _ from 'lodash'
 import { Value, Values } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
-import { types } from '@salto-io/lowerdash'
 import { RequiredDefinitions } from './types'
 import { APIDefinitionsOptions } from './api'
 
 const log = logger(module)
+export const DEFINITIONS_OVERRIDES = 'DEFINITION_OVERRIDES'
 
-export const DEFINITIONS_FLAGS = {
-  definitionOverride: 'DEFINITION_OVERRIDE',
-} as const
-
-type CoreFlagName = types.ValueOf<typeof DEFINITIONS_FLAGS>
-
-const getFlag = (flagName: CoreFlagName): string | undefined => process.env[flagName]
-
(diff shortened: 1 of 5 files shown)


packages/adapter-components/src/definitions/system/overrides.ts
```diff
@@ -0,0 +1,64 @@
...
+ * limitations under the License.
+ */
+import _ from 'lodash'
+import { Value, Values } from '@salto-io/adapter-api'
+import { logger } from '@salto-io/logging'
+import { RequiredDefinitions } from './types'
+import { APIDefinitionsOptions } from './api'
+
+const log = logger(module)
+export const DEFINITIONS_OVERRIDES = 'DEFINITION_OVERRIDES'
```
ori-moisis, line 23: didn't review the code itself, just one thing about this - every env var used by salto should have a `SALTO_` prefix

Uri-Bar: yep np

packages/adapter-components/src/definitions/system/overrides.ts
```diff
@@ -0,0 +1,71 @@
...
+      log.error('An unknown error occurred while parsing a flag:', e)
+    }
+  }
+  if (parsedFlagValue !== undefined && typeof parsedFlagValue === 'object') {
+    return parsedFlagValue as Values
+  }
+  return {}
+}
+
+export const mergeDefinitionsWithOverrides = <Options extends APIDefinitionsOptions>(
```
shir-reifenberg, line 43: 1. can you please add documentation about the behavior of this merge ? similar to what we have in `mergeSingleDefWithDefault` 
2. I think `mergeDefinitionsWithOverrides` should handle parsing the env var, so we would expose a single API for the adapter to use to get definitions with overrides
3. not completely sure about this one - should we align arrays behavior with `mergeSingleDefWithDefault` ? WDYT ?

packages/adapter-components/src/definitions/system/overrides.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+    if (_.isObject(objValue) && _.isObject(srcValue)) {
+      return _.mergeWith({}, objValue, srcValue, customMerge)
+    }
+    return srcValue
+  }
+  if (_.isEmpty(overrides)) {
+    return definitions
+  }
+  const merged = _.mergeWith(definitions, overrides, customMerge)
```
shir-reifenberg, line 59: this changes `definitions` inplace, right ? should we clone this or indicate somewhere in the function name that it changes definitions inplace ?

packages/adapter-components/src/definitions/system/overrides.ts
```diff
@@ -0,0 +1,71 @@
...
+    if (e instanceof SyntaxError) {
+      log.error('There was a syntax error in the JSON while parsing a flag:', e.message)
+    } else {
+      log.error('An unknown error occurred while parsing a flag:', e)
+    }
+  }
+  if (parsedFlagValue !== undefined && typeof parsedFlagValue === 'object') {
+    return parsedFlagValue as Values
+  }
+  return {}
```
shir-reifenberg, line 40: maybe return just `undefined` when there's no override and avoid merging ?

packages/adapter-components/test/definitions/system/overrides.test.ts
```diff
@@ -0,0 +1,257 @@
...
+          },
+        },
+      }
+      const merged = mergeDefinitionsWithOverrides(mockedDefinitions, overrides)
+      expect(merged.fetch.instances.default?.element?.topLevel?.elemID).toEqual(
+        overrides.fetch.instances.default.element.topLevel.elemID,
+      )
+      expect(merged.fetch.instances.default?.element?.topLevel?.singleton).toBeTruthy()
+    })
+    it('should remove a type from the fetch config', () => {
```
shir-reifenberg, line 243: maybe change test cases title to describe the specific case handle and not the specific value we override?
i.e instead of `should remove a type from the fetch config`  -> `should override existing keys when provided with null` 
(also in the array cases)

shir-reifenberg (commented):
Very cool!!! 👑  Added some comments, we can also discuss IRL

packages/adapter-components/src/definitions/system/overrides.ts
```diff
@@ -0,0 +1,71 @@
...
+import { Value, Values } from '@salto-io/adapter-api'
+import { logger } from '@salto-io/logging'
+import { RequiredDefinitions } from './types'
+import { APIDefinitionsOptions } from './api'
+
+const log = logger(module)
+export const DEFINITIONS_OVERRIDES = 'SALTO_DEFINITION_OVERRIDES'
+
+export const getParsedDefinitionsOverrides = (): Values => {
+  const flagValue = process.env[DEFINITIONS_OVERRIDES]
```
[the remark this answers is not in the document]
Uri-Bar: change to value

packages/adapter-components/test/definitions/system/overrides.test.ts
Uri-Bar: add describe with explanation

packages/adapter-components/src/definitions/system/overrides.ts
Uri-Bar: added clone

packages/adapter-components/src/definitions/system/overrides.ts
```diff
@@ -0,0 +1,79 @@
...
+}
+
+/**
+ * merge definitions with overrides from the environment variable SALTO_DEFINITIONS_OVERRIDES
+ * the merge is done as follows:
+ * - overrides takes precedence over definitions
+ * - when merging an array the overrides array is used instead of the definitions array completely (no merge)
+ * - when merging an object the overrides object is merged with the definitions object recursively
+ * - we can use null to remove a field from the definitions as we override the original value with null and then remove it
+ */
```
shir-reifenberg, line 49: nice 🥇

packages/adapter-components/src/definitions/system/overrides.ts
```diff
@@ -0,0 +1,79 @@
...
+  definitions: RequiredDefinitions<Options>,
+): RequiredDefinitions<Options> => {
+  const customMerge = (objValue: Value, srcValue: Value): Value => {
+    if (_.isArray(objValue)) {
+      return srcValue
+    }
+    if (_.isObject(objValue) && _.isObject(srcValue)) {
+      return _.mergeWith({}, objValue, srcValue, customMerge)
+    }
+    return srcValue
```
shir-reifenberg, line 60: Looking at this again - I think this can be simplified by just returning `undefined` and let `_.mergeWith` handle the merge (cause besides objects, the behavior we want is like any other merge, right ) ?
also, I think there's no need to the double recursive iterations using `_.mergeWith` inside `_.mergeWith`

packages/adapter-components/src/definitions/system/overrides.ts
```diff
@@ -0,0 +1,79 @@
...
+import _ from 'lodash'
+import { Value, Values } from '@salto-io/adapter-api'
+import { logger } from '@salto-io/logging'
+import { RequiredDefinitions } from './types'
+import { APIDefinitionsOptions } from './api'
+
+const log = logger(module)
+export const DEFINITIONS_OVERRIDES = 'SALTO_DEFINITIONS_OVERRIDES'
+
+export const getParsedDefinitionsOverrides = (): Values => {
```
shir-reifenberg, line 25: no need for the export now, right ?

packages/adapter-components/src/definitions/system/overrides.ts
```diff
@@ -0,0 +1,79 @@
...
+      return _.mergeWith({}, objValue, srcValue, customMerge)
+    }
+    return srcValue
+  }
+  const overrides = getParsedDefinitionsOverrides()
+  if (_.isEmpty(overrides)) {
+    return definitions
+  }
+  const cloneDefinitions = _.cloneDeep(definitions)
+  const merged = _.mergeWith(cloneDefinitions, overrides, customMerge)
```
shir-reifenberg, line 67: please don't forget to add some logs 🙏

shir-reifenberg (approved):
LGTM 🚀 , added some comments but approving to save cycle

packages/adapter-components/src/definitions/system/overrides.ts
Uri-Bar: for testing purposes, I already wrote the tests

Uri-Bar: removed

packages/adapter-components/src/definitions/system/overrides.ts
Uri-Bar: 🔥

PR #6131, opened 2024-06-28
SALTO-6172 - Salesforce: Missing reference from LightningPage to QuickAction

tomsellek:
`LightningPage.flexiPageRegions.itemInstances.componentInstances.componentInstanceProperties` may contain references to `QuickAction` when the name value is `"actionNames"`. 
e.g.

```
salesforce.LightningPage SomePage {
  flexiPageRegions = [
    {
      itemInstances = [
        {
          componentInstance = {
            componentInstanceProperties = [
              {
                name = "actionNames"
                valueList = {
                  valueListItems = [
                    {
                      value = "Global.NewTask"
                    },
                    {
                      value = "Opportunity.SomeQuickAction"
                    },
                ]
            }
    ...
```

---

Note the potential `Global.` prefix that needs to be handled if the `QuickAction` is not nested under any object type

Tests: 
 - Created a Lightning page with a global `QuickAction` and verified it becomes a reference after a fetch.

Workspace diff: https://github.com/salto-io/tomsellek-sf-6172/pull/1/files

---
_Release Notes_: 
Salesforce: Will now generate references from `LightningPage` to `QuickAction`.

---
_User Notifications_: 
Salesforce: Will now generate references from `LightningPage` to `QuickAction`.

### DIFF
--- a/packages/salesforce-adapter/src/transformers/reference_mapping.ts
+++ b/packages/salesforce-adapter/src/transformers/reference_mapping.ts
@@ -911,6 +911,11 @@ export const fieldNameToTypeMappingDefs: FieldReferenceDefinition[] = [
     src: { field: 'object', parentTypes: ['PermissionSetObjectPermissions'] },
     target: { type: CUSTOM_OBJECT },
   },
+  {
+    src: { field: 'value', parentTypes: ['ComponentInstancePropertyListItem'] },
+    sourceTransformation: 'globalPrefix',
+    target: { type: 'QuickAction' },
+  },
 ]
 
 const matchName = (name: string, matcher: string | RegExp): boolean =>
--- a/packages/adapter-components/src/references/reference_mapping.ts
+++ b/packages/adapter-components/src/references/reference_mapping.ts
@@ -84,7 +84,10 @@ export type ReferenceSourceTransformation = {
   validate: (referringFieldValue: string | number, serializedRefExpr: string) => boolean
 }
 
-export type ReferenceSourceTransformationName = 'exact' | 'asString' | 'asCaseInsensitiveString'
+const removeGlobalPrefix = (fieldValue: string): string =>
+  fieldValue.startsWith('Global.') ? fieldValue.substr(7) : fieldValue
+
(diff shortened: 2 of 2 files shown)


packages/salesforce-adapter/src/transformers/reference_mapping.ts
```diff
@@ -942,6 +942,11 @@ export const fieldNameToTypeMappingDefs: FieldReferenceDefinition[] = [
     src: { field: 'object', parentTypes: ['PermissionSetObjectPermissions'] },
     target: { type: CUSTOM_OBJECT },
   },
+  {
+    src: { field: 'value', parentTypes: ['ComponentInstancePropertyListItem'] },
```
amitgoldarb, line 946: I might be missing something, but can't this reference a great number of different types? I wonder if this is like formula's syntax. Please see this ticket - https://salto-io.atlassian.net/browse/SALTO-1878

amitgoldarb (commented):
Added a small question 🙏

packages/salesforce-adapter/src/transformers/reference_mapping.ts
tomsellek: No, you're totally right. I don't think this is like the formula syntax in this case, it's just that we don't have a mechanism in place to say "only look for references if the parent has `name = "actionNames"`. So it's going to look for `QuickAction` instances with the given name, and if the name is of something from a different type it's just not going to find a reference. So I think it should all work out in the end. Unless I'm missing something?

amitgoldarb: I think I might be missing something here, but when reading this code I think we are losing information in the serialization right? The resolved value needs to return to be `global.API_NAME` right? Doesn't that mean we need this to be a `ReferenceSerializationStrategyLookup` and not this way? Did you attempt to fetch and then deploy an instance like this? I think it will probably fail or have some unwanted side effect

tomsellek: I rewrote this to be based on the formula-based references code. The new PR is at https://github.com/salto-io/salto/pull/6131

tomsellek:
Moved to https://github.com/salto-io/salto/pull/6131

PR #6135, opened 2024-06-30
SALTO-6173: omit user segment id field from deployment of article

edenhassid:
omit user segment id field from deployment of article

---

---
_Release Notes_: 
zendesk:
* fix deployment of articles

---
_User Notifications_: 


adfineberg (approved):
💯

PR #6137, opened 2024-06-30
SALTO-6178 Allow running fetch with regenerate Salto IDs by selectors

dantal4:
Add the option to regenerate only selected element ids, that match the input element selectors.

---

TODO:
- [x] add tests

---
_Release Notes_: 
Core:
- add the `ignoreStateElemIdMappingForSelectors` param to fetch API, in order to regenerate Salto IDs by selectors

CLI:
- add the `regenerateSaltoIdsForSelectors` param to fetch command, in order to regenerate Salto IDs by selectors

---
_User Notifications_: 
None


packages/core/src/core/fetch.ts
```diff
@@ -1282,24 +1284,57 @@ export const createElemIdGetter = async (
...
+  workspace: Workspace,
+  accountToServiceNameMap: Record<string, string>,
+  elementsSource: ReadOnlyElementsSource,
+  ignoreStateElemIdMapping: boolean,
+  ignoreStateElemIdMappingForSelectors: ElementSelector[],
+): Promise<Record<string, ElemIdGetter>> => {
+  if (ignoreStateElemIdMapping && ignoreStateElemIdMappingForSelectors.length === 0) {
+    return {}
+  }
+  const ignoreStateElemIdMappingForElements = new Set(
```
ori-moisis, line 1297: there is already an iteration in `generateServiceIdToStateElemId` where it would be fairly easy to filter.

alternatively, if you already have exactly all the IDs that you want to build the index from, use that list instead of having another `getAll` on the elements source which forces us to read all the elements

dantal4: Do you mean to pass `ignoreStateElemIdMappingForElements` to `generateServiceIdToStateElemId` or to pass it the list of selectors and use it to filter there? There's no exported function in `element_selectors.ts` that receives an element id and checks if it matches the selectors.

ori-moisis: in the first option, `selectElementsBySelectorsWithoutReferences` essentially does that (instead of a boolean it would return an empty / non-empty array, but the point is the same...)

for the second option you don't have to change `generateServiceIdToStateElemId`

dantal4: for the first option - isn't there any reason to allow passing selectors with a `referenceBy`?
for the second option - I guess that eventually most elements won't need to be regenerated, so do you think that it's preferred to call `awu(elementsSource.list()).map(elemId => elementsSource.get(elemId))` instead of calling `elementsSource.getAll()` ?

ori-moisis: if you really want to get into the details, `getAll` is faster if and only if there is one account in the workspace (and even then, only if the workspace is very large, to the point where it does not fit in the remote map cache).

so, if you really want to optimize, you could do something like

```typescript
const getElementsToMaintain = (...): AsyncIterable<Element> => {
  if (ignoreStateElemIdMappingForSelectors.length === 0 && Object.keys(accountToServiceMap).length === 1) {
    return elementsSource.getAll()
  }
  const accountElementIDs = awu(elementsSource.list()).filter(id => id.adapter === account))
  const elementIDsToMaintain = ignoreStateElemIdMappingForSelectors.length === 0
    ? accountElementIDs
    : awu(selectElementsBySelectors({
        elementIds: accountElementIDs,
        ...
      })
  return elementIDsToMaintain.map(id => elementsSource.get(id))
}
```

in practice it would probably be ok to omit the first condition though...

ori-moisis (approved):

packages/core/test/core/fetch.test.ts
```diff
@@ -2216,14 +2223,79 @@ describe('fetch from workspace', () => {
...
+      const idGetters = await createElemIdGetters(
+        mockWorkspace({ stateElements: [type, instance] }),
+        { salesforce: 'salesforce' },
+        createElementSource([type, instance]),
+        true,
+        [],
+      )
+      expect(idGetters).toEqual({})
+    })
+    it('should not get elem id for elements that match ignoreStateElemIdMappingForSelectors', async () => {
```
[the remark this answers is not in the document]
dantal4: aren't instance & instanceToIgnore test that?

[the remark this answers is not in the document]
ori-moisis: kind of hard to figure that out (tbh I am still not sure what that `createAdapterReplacedID` is all about)

it would be better if tests say exactly what they are testing, if you are testing two things, it would be better if there are two tests, and each test would clearly say what it is testing.
here the test says it is testing that things which are ignored do not get an element ID (which, in itself is confusing, they do get an element ID, they just don't keep the old one), it does not say anything about things which are not ignored...

packages/core/test/core/fetch.test.ts
```diff
@@ -2216,14 +2223,79 @@ describe('fetch from workspace', () => {
...
+      const serviceIdsToIgnore = {
+        serviceIdField: '1',
+        [OBJECT_SERVICE_ID]: toServiceIdsString({
+          [OBJECT_NAME]: typeToIgnore.elemID.getFullName(),
+        }),
+      }
+      expect(idGetter('salesforce', serviceIds, 'newInst')).toEqual(
+        createAdapterReplacedID(instance.elemID, 'salesforce'),
+      )
+      expect(idGetter('salesforce', serviceIdsToIgnore, 'newInst')).toEqual(new ElemID('salesforce', 'newInst'))
```
ori-moisis, line 2298: why did this turn into a type ID? isn't it supposed to be an instance?

hadard:
@dantal4, thanks for running this so quickly! Does the plan include adding a text box to the UI near the regenerate IDs in order to pass the selectors to the core API?

packages/core/test/core/fetch.test.ts
dantal4: this is how elemIDGetter works - in case that the elemId doesn't exist it returns a new elem id with the given service name & name

dantal4:
> @dantal4, thanks for running this so quickly! Does the plan include adding a text box to the UI near the regenerate IDs in order to pass the selectors to the core API?

@hadard NP 👍  the quickest solution will be to use a FF that passes selectors to the fetch call.

PR #6138, opened 2024-06-30
SAAS-9608: Move account URL in to AccountInfo type

yelly:
This field can be relevant for all adapters and has semantic meaning (it should be a URL leading to the account in the service).

---

None.

---
_Release Notes_: 
None.

---
_User Notifications_: 
None.


tamtamirr (approved):
LGTM 💯

PR #6141, opened 2024-06-30
SALTO-6059: projects fields dependency

IdoZakk:
Removed the previous handling of the JSM-locked fields and added a new CV based on the field type
---

There is a map of all templates and the fields created by adding them [here](https://www.notion.so/saltoio/Fields-per-template-4d1e03e130084f18ba076a77083f0741)

I am still waiting for Lior's final decision about the CV message. It might complicate the logic, as we might want to give the minimal number of templates to add. Another option is to use an article.

---
_Release Notes_: 
Jira Adapter: 
* Fixed a potential circular dependency when deploying a JSM project

---
_User Notifications_: 
None


packages/jira-adapter/src/change_validators/jsm/jsm_locked_fields.ts
```diff
@@ -0,0 +1,64 @@
...
+ * limitations under the License.
+ */
+import { ChangeValidator, getChangeData, isInstanceElement, SeverityLevel } from '@salto-io/adapter-api'
+import { collections } from '@salto-io/lowerdash'
+import { FIELD_TYPE_NAME, IS_LOCKED, SERVICE, WORK_CATEGORY_FIELD } from '../../filters/fields/constants'
+import { isRelatedToSpecifiedTerms } from '../../common/fields'
+
+const { awu } = collections.asynciterable
+
+const JSM_FIELD_TYPE_MAP = (): Record<string, string[]> => ({
```
idoamit8, line 23: nit: Since it doesn't receive any parameters, I don't see any reason to use a function here instead of a const record, as we do in other places.
In addition I suggest to name it `getProjectTemplatesToFields` if its a function or `PROJECT_TEMPLATE_TO_FIELDS` if its a const

packages/jira-adapter/src/change_validators/locked_fields.ts
```diff
@@ -13,31 +13,20 @@
...
 export const lockedFieldsValidator: ChangeValidator = async changes =>
   awu(changes)
-    .filter(isInstanceChange)
     .map(getChangeData)
+    .filter(isInstanceElement)
     .filter(instance => instance.elemID.typeName === FIELD_TYPE_NAME)
-    .filter(instance => instance.value?.[IS_LOCKED] === true)
-    .filter(instance => !isRelatedToSpecifiedTerms(instance, [SERVICE]))
+    .filter(instance => instance.value[IS_LOCKED] === true)
+    .filter(instance => !isRelatedToSpecifiedTerms(instance, [SERVICE, WORK_CATEGORY_FIELD]))
```
idoamit8, line 29: Why don't we use one CV with different messages for JSM lock fields and non-JSM locked fields? I feel it would be more maintainable and less prone to errors.

packages/jira-adapter/src/common/fields.ts
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { InstanceElement } from '@salto-io/adapter-api'
+import { logger } from '@salto-io/logging'
+
+const log = logger(module)
+export const isRelatedToSpecifiedTerms = (instance: InstanceElement, terms: string[]): boolean => {
+  const includesTerm = (term: string): boolean => instance.value.type?.includes(term)
```
idoamit8, line 21: Question: There are a lot of field without a field named `type`. Are we sure each filed that is relevant here has type field? it can't be related to some other fields?

packages/jira-adapter/src/change_validators/locked_fields.ts
```diff
@@ -13,31 +13,20 @@
...
-    return true
-  }
-  return false
-}
 
 export const lockedFieldsValidator: ChangeValidator = async changes =>
   awu(changes)
-    .filter(isInstanceChange)
     .map(getChangeData)
+    .filter(isInstanceElement)
```
idoamit8, line 26:
```suggestion
.filter(isInstanceChange)
.map(getChangeData)
```

packages/jira-adapter/src/change_validators/jsm/jsm_locked_fields.ts
```diff
@@ -0,0 +1,64 @@
...
+
+const jsmFieldToTemplateName = (field: string): string => {
+  const map = JSM_FIELD_TYPE_MAP()
+  return Object.keys(map).find(key => map[key].includes(field)) ?? 'the relevant type'
+}
+
+export const jsmLockedFieldsValidator: ChangeValidator = async changes =>
+  awu(changes)
+    .map(getChangeData)
+    .filter(isInstanceElement)
```
idoamit8, line 54:
```suggestion
.filter(isInstanceChange)
.map(getChangeData)
```
If possible, it's better to filter before applying a function. This way, we avoid applying a function to an object that will be filtered out.

idoamit8 (commented):
Great work! 💯 
Really just few comments

packages/jira-adapter/src/change_validators/jsm/jsm_locked_fields.ts
IdoZakk: I started using this with inspiration from Netsuite. The idea behind it is that a const string indeed cannot be changed, but a const array/record can be modified. To overcome this they create const arrays/records using a function. This way if someone edits it (by mistake or on purpose) it will not affect the others using it.
 I will change the name

packages/jira-adapter/src/common/fields.ts
IdoZakk: I mapped all the fields of all the different templates, and all had types. Maybe I am missing something, but it seems safe to me ATM

packages/jira-adapter/src/change_validators/locked_fields.ts
IdoZakk: It was like that before, this way the compiler does not know it is an instanceElement, and you need to add `?` for value

packages/jira-adapter/src/change_validators/jsm/jsm_locked_fields.ts
IdoZakk: As I wrote in the previous function it is better for the compiler. And we do not get any changes that are not instance elements in Jira

packages/jira-adapter/src/change_validators/locked_fields.ts
```diff
@@ -13,36 +13,55 @@
...
     .map(async instance => ({
       elemID: instance.elemID,
       severity: 'Error' as SeverityLevel,
       message: 'Cannot deploy a locked field',
-      detailedMessage:
-        'The field is locked and cannot be deployed. Learn more here: https://help.salto.io/en/articles/6933969-the-field-is-locked-and-cannot-be-deployed',
+      detailedMessage: isRelatedToSpecifiedTerms(instance, [SERVICE, WORK_CATEGORY_FIELD])
+        ? 'The field is auto-generated by Atlassian. To create or modify it in the target environment, create a first project ' +
+          `from a template of ${jsmFieldToTemplateName(instance.elemID.name)}. For more information, see the documentation at: ` +
+          'https://help.salto.io/en/articles/9556414-jira-service-management-locked-fields'
```
kami-delaroz, line 64: nice mapping here!

kami-delaroz (approved):
💯

PR #6148, opened 2024-07-02
Remove cycle dependents from deploy.xml

dantal4:
We should remove from deploy.xml not only objects that are in a dependency cycle, but also objects that are dependent on them.

---

---
_Release Notes_: 
Netsuite Adapter:
- Fix "unresolved reference" error in SDF deploy

---
_User Notifications_: 
None


omrilit (approved):
Looks good!

PR #6155, opened 2024-07-03
SALTO-6196: New Fields and Objects FLS CV CustomMetadata additions bugfix

tamtamirr:
Fixed an issue when we created FLS CV message on added fields of CustomMetadata types, in case their object was added. 

---

_Release Notes_: 
Salesforce Adapter:
- Fixed an issue when we created FLS CV message on added fields of CustomMetadata types, in case their object was added. 

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/change_validators/new_fields_and_objects_fls.ts
```diff
@@ -61,7 +64,8 @@ const changeValidator =
       .filter(isAdditionChange)
       .map(getChangeData)
       .filter(isObjectType)
```
yelly, line 66: Nit: isn't it better to just filter custom objects here?

tamtamirr: filtering for only `isCustomObjectSync` is not good enough here, as this will not take into consideration `CustomMetadata types`. I've updated the code to be more precise though 🙏

yelly (approved):

yelly (approved):

PR #6157, opened 2024-07-03
SALTO-6198: Clean noisy logs from adapters infra

shir-reifenberg:
Clean noisy logs

---

in ticket

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/fetch/element/type_utils.ts
+++ b/packages/adapter-components/src/fetch/element/type_utils.ts
@@ -252,11 +252,10 @@ export const overrideFieldTypes = <Options extends FetchApiDefinitionsOptions>({
         }
         const field = type.fields[fieldName]
         if (field === undefined) {
-          log.debug('field %s.%s is undefined, not applying customizations', typeName, fieldName)
           return
         }
         if (restrictions) {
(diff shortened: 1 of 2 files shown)


packages/adapter-components/src/fetch/element/type_utils.ts
```diff
@@ -290,17 +289,16 @@ export const hideAndOmitFields = <Options extends FetchApiDefinitionsOptions>({
       Object.entries(elementDef?.fieldCustomizations ?? {}).forEach(([fieldName, customization]) => {
         const field = type.fields[fieldName]
         if (field === undefined) {
-          log.debug('field %s.%s is undefined, not applying customizations', typeName, fieldName)
```
netama, line 293: can we add this back as trace as well please?

netama (approved):
LGTM with a small change, thanks!!

packages/adapter-components/src/fetch/element/type_utils.ts
shir-reifenberg: discussed IRL and decided to remove for now

PR #6158, opened 2024-07-04
SALTO-6129: Improving profiles element fixer message

yelly:
Improving profiles element fixer message

---

---
_Release Notes_: 
None.

---
_User Notifications_: 
None.

### DIFF
--- a/packages/salesforce-adapter/src/custom_references/profiles.ts
+++ b/packages/salesforce-adapter/src/custom_references/profiles.ts
@@ -367,8 +367,9 @@ const removeWeakReferences: WeakReferencesHandler['removeWeakReferences'] =
       return {
         elemID: profile.elemID,
         severity: 'Info' as const,
-        message: 'Dropping profile fields which reference missing types',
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/src/custom_references/profiles.ts
```diff
@@ -367,8 +367,9 @@ const removeWeakReferences: WeakReferencesHandler['removeWeakReferences'] =
       return {
         elemID: profile.elemID,
         severity: 'Info' as const,
-        message: 'Dropping profile fields which reference missing types',
-        detailedMessage: `The profile has ${profileBrokenReferenceFields.length} fields which reference types which are not available in the workspace.`,
+        message: 'Omitting profile entries which reference unavailable types',
```
hadard, line 370: @yelly I'm not sure this message explains to the user what happened here. I think the way you phrased it in the slack thread is better, WDYT about: 
message: "This profile contains some permissions for types that are not available in the target environment"
detailed message: "The profile contains some permissions for types that are not available in the target environment and will not be deployed" 
@tomermevorach @amitgoldarb ?

PR #6160, opened 2024-07-04
SALTO-4153: Salesforce Custom References between CPQ Rules and their Conditions

tamtamirr:
Add references from CPQ Rules with Advanced Condition to the Condition Instances they require

---

This PR covers the different use-cases for rules and conditions in the CPQ area.
We can easily expand on that with new permutations of the same problem we introduce to in the future

---
_Release Notes_: 
Salesforce Adapter:
- Improved the dependencies between CPQ Rules and their conditions to avoid deploy issues due to missing reference.

---
_User Notifications_: 
_None_


tamtamirr (commented):
added a small comment

packages/salesforce-adapter/src/custom_references/rules_and_conditions.ts
```diff
@@ -0,0 +1,208 @@
...
+const createReferencesFromRuleInstance = (
+  ruleInstance: InstanceElement,
+  conditionsByIndex: Record<string, InstanceElement>,
+  def: RuleAndConditionDef,
+): ReferenceInfo[] => {
+  const condition = ruleInstance.value[def.rule.customConditionField]
+  if (!_.isString(condition)) {
+    return []
+  }
+  const regexMatch = condition.match(/?\d+/g)
```
yelly, line 119: Use `(?<!\S)\d+(?!\S)` to avoid matching numbers in operator names (who knows what those maniacs at CPQ will do).

packages/salesforce-adapter/src/custom_references/rules_and_conditions.ts
```diff
@@ -0,0 +1,208 @@
...
+  if (regexMatch == null) {
+    return []
+  }
+  // We should avoid creating mutliple references to the same condition
+  // e.g. the following custom condition is valid: "0 OR (0 AND 1)"
+  const indexes = _.uniq(Array.from(regexMatch.values()))
+  return indexes
+    .map<ReferenceInfo | undefined>((index) => {
```
yelly, line 127: I think it's more readable to put the type on the function:
```suggestion
    .map((index): ReferenceInfo | undefined => {
```

packages/salesforce-adapter/src/custom_references/rules_and_conditions.ts
```diff
@@ -0,0 +1,208 @@
...
+      apiName: SBAA_APPROVAL_CONDITION,
+      indexField: SBAA_INDEX_FIELD,
+      ruleField: SBAA_APPROVAL_RULE,
+    },
+  },
+]
+
+const createReferencesFromRuleInstance = (
+  ruleInstance: InstanceElement,
+  conditionsByIndex: Record<string, InstanceElement>,
```
yelly, line 112: I think it's more clear and accurate to use a `Record<number, InstanceElement>` here.

packages/salesforce-adapter/src/custom_references/rules_and_conditions.ts
```diff
@@ -0,0 +1,208 @@
...
+      return {
+        source: ruleInstance.elemID,
+        target: conditionInstance.elemID,
+        type: 'strong',
+      }
+    })
+    .filter(isDefined)
+}
+
+type CreateReferencesFromDefArgs = {
```
yelly, line 144: You only use this once, why not define inline?

packages/salesforce-adapter/src/custom_references/rules_and_conditions.ts
```diff
@@ -0,0 +1,208 @@
...
+  instancesByType: Record<string, InstanceElement[]>
+}
+
+const isConditionOfRuleFunc =
+  (rule: InstanceElement, ruleField: string) =>
+  (condition: InstanceElement): boolean => {
+    const ruleRef = condition.value[ruleField]
+    return (
+      isReferenceExpression(ruleRef) &&
+      ruleRef.elemID.getFullName() === rule.elemID.getFullName()
```
yelly, line 155:
```suggestion
      ruleRef.elemID.isEqual(rule.elemID)
```

packages/salesforce-adapter/src/custom_references/rules_and_conditions.ts
```diff
@@ -0,0 +1,208 @@
...
+    )
+  }
+
+const getConditionIndex = (
+  condition: InstanceElement,
+  indexField: string,
+): number => {
+  const index = condition.value[indexField]
+  return _.isNumber(index) ? index : -1
```
yelly, line 164: Not sure about returning `-1` in the error case, IMO `undefined` is preferred.

packages/salesforce-adapter/src/custom_references/rules_and_conditions.ts
```diff
@@ -0,0 +1,208 @@
...
+}
+
+const createReferencesFromDef = ({
+  def,
+  instancesByType,
+}: CreateReferencesFromDefArgs): ReferenceInfo[] => {
+  const rules = instancesByType[def.rule.apiName]
```
yelly, line 171: I think it's a bit confusing to call this `apiName` if we're comparing it to `elemID.typeName`.

packages/salesforce-adapter/src/custom_references/rules_and_conditions.ts
```diff
@@ -0,0 +1,208 @@
...
+      getConditionIndex(condition, def.condition.indexField),
+    )
+    return createReferencesFromRuleInstance(rule, conditionsByIndex, def)
+  })
+}
+
+const findWeakReferences: WeakReferencesHandler['findWeakReferences'] = async (
+  elements,
+) => {
+  log.debug('rules and conditions handler started')
```
yelly, line 193: This log is just spam, no?

tamtamirr: This log should help understanding when we started and finished executing the logic of this customReferences func. IDM the extra log.

packages/salesforce-adapter/src/custom_references/rules_and_conditions.ts
tamtamirr: This is actually the Salesforce terminology for the unique identifier of the type. In case the types were correct here, I'd use `apiNameSync(instance.getTypeSync())` instead. But this is a workaround

yelly: So maybe add a comment about this.

packages/salesforce-adapter/src/custom_references/rules_and_conditions.ts
```diff
@@ -0,0 +1,207 @@
...
+    const ruleConditions = makeArray(
+      instancesByType[def.condition.apiName],
+    ).filter(isConditionOfCurrentRule)
+    const conditionsByIndex: Record<number, InstanceElement> = {}
+    ruleConditions.forEach((condition) => {
+      const index = getConditionIndex(condition, def.condition.indexField)
+      if (index !== undefined) {
+        conditionsByIndex[index] = condition
+      }
+    })
```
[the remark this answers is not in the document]
tamtamirr: changed my implementation to `reduce` instead

packages/salesforce-adapter/src/custom_references/rules_and_conditions.ts
tamtamirr: Changed to `typeApiName` to make this more clear 🙏

tamtamirr:
Closing in favor of the Template Expressions solution in https://github.com/salto-io/salto/pull/6225

PR #6161, opened 2024-07-04
SALTO-5926 manage pages for predefined spaces only

omrifarkash:
Changes on this PR:
- Add to Confluence user fetch config a new field `managePagesForSpaces: string[]`
- Changed the way we fetch pages. Now we bring pages per space with `recurseInto` definitions
- Add condition to pages recurseInto, only if space name is matching one of the strings on `managePagesForSpaces` field
- Default for `managePagesForSpaces` is undefined which means bringing no pages at all

---
This commit was tested on an existing workspace in the following way:
1. fetch the workspace with build of main (without this commit). and commit the current workspace change the git
2. build this commit, change nothing in the adapter config
3. fetch the workspace again and verify that the workspace hasn't changed
4. add to the adapter config `managePagesForSpaces = ['1 real space name', '2 real space name']
5. fetch again and verify that pages from all other spaces than '1 real space name', '2 real space name' were deleted

---
_Release Notes_: 
_Confluence Adapter_:
- new ability to manage pages only for specific spaces (these space can be defined under fetch configuration in the adapter config)

---
_User Notifications_: 
_Confluence Adapter_:
- new ability to manage pages only for specific spaces (these space can be defined under fetch configuration in the adapter config)


packages/confluence-adapter/src/definitions/fetch/fetch.ts
```diff
@@ -240,7 +267,7 @@ const createCustomizations = (): Record<string, definitions.fetch.InstanceFetchA
       },
     ],
     resource: {
-      directFetch: true,
+      directFetch: false,
```
netama, line 270: will they still be able to include/exclude pages? also I assume this will prevent them from fetching pages without fetching spaces - we can start with it for now, but will probably want to change it later (which will also require changing the url to not depend on spaces for that case)

packages/confluence-adapter/src/definitions/fetch/fetch.ts
```diff
@@ -147,6 +150,22 @@ const createCustomizations = (): Record<string, definitions.fetch.InstanceFetchA
             },
           },
         },
+        pages: {
+          typeName: PAGE_TYPE_NAME,
+          conditions: [
+            {
+              fromField: 'name',
+              match: userConfig.fetch.managePagesForSpaces ?? [],
```
netama, line 158: 1. can't we just not have conditions when the flag is not defined?
2. I think the phrasing is a bit misleading, maybe `manageOnly...`? (I personally think it'd be better either as a fetch criterion or at least as some object that can also contain the template info, but since it's not persisted and will be relatively simple to migrate later we can keep it here for now)

netama (commented):
added a few comments/questions (mostly for discussion, if you and Lior agree on the solution someone else can review the rest :) )

packages/confluence-adapter/src/definitions/fetch/fetch.ts
omrifarkash: I don't see why we'd like to change it. Not specifying `managePagesForSpaces` will yield the same results as exclude page if it was directFetch

packages/confluence-adapter/src/definitions/fetch/fetch.ts
omrifarkash: This comment is irrelevant since we change the default to not bringing pages at all

packages/confluence-adapter/test/adapter.test.ts
```diff
@@ -100,7 +100,9 @@ describe('adapter', () => {
               token: 'pass',
               subdomain: 'subdomain',
             }),
-            config: new InstanceElement('config', adapter.configType as ObjectType, DEFAULT_CONFIG),
+            config: new InstanceElement('config', adapter.configType as ObjectType, {
+              fetch: { ...DEFAULT_CONFIG.fetch, managePagesForSpaces: ['My first space'] },
+            }),
```
omrifarkash, line 105: Testing bringing pages only from 1 space (instead of 2 spaces exists in this test).
In mockFetchResponse, there are responses for pages out of `My first space` but this is testing that we are not fetching them

packages/confluence-adapter/e2e_test/adapter.test.ts
```diff
@@ -57,7 +58,7 @@ const fieldsToOmitOnComparisonPerType: Record<string, string[]> = {
   [TEMPLATE_TYPE_NAME]: [],
 }
 
-const fetchDefinitions = createFetchDefinitions()
+const fetchDefinitions = createFetchDefinitions({ fetch: { ...DEFAULT_CONFIG.fetch, managePagesForSpaces: ['.*'] } })
```
shir-reifenberg, line 61: shouldn't the default be an empty array ? also, can you please have a const for the default value ?

shir-reifenberg (commented):
Config looks good to me, a few comments / suggestions - 
1. I don't have all the context so maybe I'm missing something, but why adding another config flag and not utilizing the `include` / `exclude` config? you can add additional `criteria` of `space` then have something like - 
```
include = [ 
   { typeName: 'page', space: "mySpace.*" }, 
]
```
If there's additional things we can get only per space I think this interface will be better

2. Should we add a config suggestion to update the config for existing envs?
3. Can you please add a config example to `config_doc` + update the release notes with an example ?

packages/confluence-adapter/e2e_test/adapter.test.ts
omrifarkash: The default is undefined, in this test I want to test fetching pages so I am overriding the default

omrifarkash:
> Config looks good to me, a few comments / suggestions -
> 
> 1. I don't have all the context so maybe I'm missing something, but why adding another config flag and not utilizing the `include` / `exclude` config? you can add additional `criteria` of `space` then have something like -
> 
> ```
> include = [ 
>    { typeName: 'page', space: "mySpace.*" }, 
> ]
> ```
> 
> If there's additional things we can get only per space I think this interface will be better
> 
> 2. Should we add a config suggestion to update the config for existing envs?
> 3. Can you please add a config example to `config_doc` + update the release notes with an example ?

1. Criteria is used in query filter which is running after we fetch the data. We want to save these api calls and we're doing it using conditions of recurseInto. I can somehow use your `include` suggestion, but then, I will need to provide a dummy criteria function that does nothing, O.W query filter will through an error. I prefer this way and it was approved by Lior

2. We don't want to update existing environment. The existing configs have `managePagesForSpaces ` = undefined, this will not fetch pages and that is the behavior we want to all the existing environment

3. 👍

shir-reifenberg (approved):
I won't be here tomorrow so I don't want to block, but please address my comments :)

PR #6162, opened 2024-07-04
SALTO-6199: Fix `Brand` addition API calls following a service change

Nurdok:
See SALTO-6199 for more details. 
TL;DR: To add a new `Brand`, we are now required to call `POST` with only the `name` field and then `PUT` with the other fields (add and then modify).

---

I also included a change for better Jest error messages in the Okta e2e test  🎉 

---
_Release Notes_: 
None

---
_User Notifications_: 
None


shir-reifenberg (approved):
🕺

PR #6163, opened 2024-07-04
SALTO-6184: Only adding errors to successful elements if the deploy failed

yelly:
We currently have a bug where we add errors to successful elements even if they only have info or warning messages and the deployment succeeded.

---

---
_Release Notes_: 
_Salesforce_:
* Fixed a bug where elements in a successful deployment got an error with the message: `Element was not deployed because other elements had errors and the 'rollbackOnError' option is enabled (or not set).`

---
_User Notifications_: 
None.


amitgoldarb (approved):
LGTM!
Thanks for this

PR #6165, opened 2024-07-07
SALTO-5692: Okta: Upgrade Application deploy definitions to new infra

Nurdok:
This PR migrates `Application` to the new infra for deploy.

Note: there are also some external behavior changes as part of this migration:
- There was a CV that prevented an active application from being removed. This PR removes that CV and allows to remove an active application by first deactivating it (this was a feature request).
- We now pass `activate=true` query params when creating an active application, instead of relying on Okta's default behavior.

Check out the diff in mock replies in the HTTP recording tests to clearly see these changes in action.

---

None

---
_Release Notes_: 
_Okta Adapter_:
- Active `Application` instances can now be removed and do not require manual deactivation first.

---
_User Notifications_: 
None


packages/okta-adapter/src/deployment.ts
shir-reifenberg, line 1: As most of the code in this file (`deployChanges`, `deployEdges`, `defaultDeployWithStatus`)  should be deleted once we migrate to the new infra maybe worth moving the utils to `deployment_utils` file and mark this file as deprecated?

packages/okta-adapter/src/deployment.ts
```diff
@@ -108,20 +118,52 @@ export const getOktaError = (elemID: ElemID, error: Error): Error => {
...
+  _.isEqual(
+    getAllChangeData(change).map(data => data.value.status),
+    [ACTIVE_STATUS, INACTIVE_STATUS],
+  )
 
 const DEACTIVATE_BEFORE_REMOVAL_TYPES = new Set([NETWORK_ZONE_TYPE_NAME])
 const shouldDeactivateBeforeRemoval = (change: Change<InstanceElement>): boolean =>
   isRemovalChange(change) && DEACTIVATE_BEFORE_REMOVAL_TYPES.has(getChangeData(change).elemID.typeName)
 
+export const isInactiveCustomAppChange = (change: Change<InstanceElement>): boolean =>
```
shir-reifenberg, line 145: I think we should probably have `application.ts` file under `deployment/types` (same as we have in `fetch`) and have all the app related code there

packages/okta-adapter/src/filters/app_deployment.ts
```diff
@@ -93,98 +59,22 @@ export const isAppResponse = createSchemeGuard<Application>(
...
-    }
-  } catch (err) {
-    throw getOktaError(getChangeData(change).elemID, err)
-  }
-}
-
 /**
  * Application type is deployed separately to update application's configuration, status and application's policies
  */
-const filterCreator: FilterCreator = ({ elementSource, definitions, oldApiDefinitions }) => ({
```
shir-reifenberg, line 187: can you please update the name of this filter (and file) and update the docstring ?

packages/okta-adapter/src/definitions/deploy.ts
```diff
@@ -43,6 +100,172 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
...
+              request: {
+                endpoint: {
+                  path: '/api/v1/apps/{applicationId}/lifecycle/activate',
+                  method: 'post',
+                },
+                context: {
+                  applicationId: '{id}',
+                },
+                transformation: {
+                  pick: [],
```
shir-reifenberg, line 213: does `omitBody` works ?

packages/okta-adapter/src/definitions/deploy.ts
```diff
@@ -43,6 +100,172 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
...
+              request: {
+                endpoint: {
+                  path: '/api/v1/apps/{id}',
+                  method: 'put',
+                },
+                transformation: {
+                  adjust: async ({ value, context }) => {
+                    validatePlainObject(value, APPLICATION_TYPE_NAME)
+                    const transformed = getChangeData(
+                      deployment.transformRemovedValuesToNull(context.change as ModificationChange<InstanceElement>, [
```
shir-reifenberg, line 165: can we avoid the cast ?

packages/okta-adapter/src/definitions/deploy.ts
```diff
@@ -43,6 +100,172 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
...
+            {
+              request: {
+                endpoint: {
+                  path: '/api/v1/apps/{applicationId}',
+                  method: 'delete',
+                },
+                context: {
+                  applicationId: '{id}',
+                },
+              },
```
shir-reifenberg, line 191: nit (also in the `activate` and `deactivate` requests)
```suggestion
              request: {
                endpoint: {
                  path: '/api/v1/apps/{id}',
                  method: 'delete',
                },
              },
```

packages/okta-adapter/src/definitions/deploy.ts
```diff
@@ -43,6 +100,172 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
...
+            },
+          ],
+          deactivate: [
+            {
+              condition: {
+                custom:
+                  () =>
+                  ({ change }) =>
+                    (isRemovalChange(change) && getChangeData(change).value.status !== INACTIVE_STATUS) ||
```
shir-reifenberg, line 224: can you please add a comment in the spirit of
`// Active app must be deactivated before removal`

packages/okta-adapter/src/definitions/deploy.ts
```diff
@@ -43,6 +100,172 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
...
+                },
+                transformation: {
+                  pick: [],
+                },
+              },
+            },
+          ],
+        },
+      },
+      toActionNames: ({ change }) => {
```
shir-reifenberg, line 245: NICE!

packages/okta-adapter/src/definitions/deploy.ts
```diff
@@ -15,22 +15,79 @@
...
+
+const { isDefined } = values
 
 type InstanceDeployApiDefinitions = definitions.deploy.InstanceDeployApiDefinitions<AdditionalAction, ClientOptions>
 export type DeployApiDefinitions = definitions.deploy.DeployApiDefinitions<AdditionalAction, ClientOptions>
 
+/**
+ * Create a deploy request for setting a policy associated with an `Application`.
+ */
+const createDeployAppPolicyRequest = (
```
shir-reifenberg, line 61: I think this as well can fit in `application.ts` file

packages/okta-adapter/src/definitions/deploy.ts
```diff
@@ -15,22 +15,79 @@
...
+  policyName: string,
+): definitions.deploy.DeployableRequestDefinition<ClientOptions> => ({
+  condition: {
+    custom:
+      () =>
+      ({ change }) =>
```
shir-reifenberg, line 67: in can of a modification change shouldn't we validate the policy changed ?

packages/okta-adapter/src/filters/app_deployment.ts
```diff
@@ -93,98 +59,22 @@ export const isAppResponse = createSchemeGuard<Application>(
...
-    new ElemID(OKTA, ORG_SETTING_TYPE_NAME, 'instance', ElemID.CONFIG_NAME),
-  )
-  if (!isInstanceElement(orgSettingInstance)) {
-    log.error(`Failed to get ${ORG_SETTING_TYPE_NAME} instance, can not find subdomain`)
-    return undefined
-  }
-  return orgSettingInstance.value.subdomain
-}
-
 export const isInactiveCustomAppChange = (change: ModificationChange<InstanceElement>): boolean =>
```
shir-reifenberg, line 68: is this needed ?

shir-reifenberg (commented):
Looks great!!! 👯 
added some comments, but nothing major
do you think we can get rid of the `preDeploy` code in the `app_deployment` filter ?

packages/okta-adapter/src/filters/app_deployment.ts
Nurdok: Moved elsewhere.

packages/okta-adapter/src/definitions/deploy.ts
Nurdok: Nice catch! This was a bug, fixed.

packages/okta-adapter/src/definitions/deploy.ts
Nurdok: Done, and check out SALTO-6313 😅

packages/okta-adapter/src/filters/app_fetch.ts
```diff
@@ -0,0 +1,81 @@
...
+  name: 'appFetchFilter',
+  onFetch: async (elements: Element[]) => {
+    const instances = elements.filter(isInstanceElement)
+    const appInstances = instances.filter(instance => instance.elemID.typeName === APPLICATION_TYPE_NAME)
+    // OrgSetting is settings type
+    const orgInstance = instances.find(instance => instance.elemID.typeName === ORG_SETTING_TYPE_NAME)
+    const subdomain = orgInstance?.value?.subdomain
+    if (!_.isString(subdomain)) {
+      log.error('Could not create customName field for custom apps because subdomain was missing')
+      return
```
shir-reifenberg, line 57: not related to your code but I think the early return here is a bug 🤔 
 we still want to set the deployment annotations and delete the feature array even if `subdomain` is not string, right ? 

I think we can just put the code related to the custom name change under
``` 
if (_.isString(subodmain)) {
  // custom name code
} else {
  log.error()
}
```

packages/okta-adapter/test/filters/app_fetch.test.ts
```diff
@@ -0,0 +1,101 @@
...
+import { filterUtils } from '@salto-io/adapter-components'
+import { buildElementsSourceFromElements } from '@salto-io/adapter-utils'
+import { createDefinitions, getFilterParams, mockClient } from '../utils'
+import OktaClient from '../../src/client/client'
+import appDeploymentFilter from '../../src/filters/app_fetch'
+import { APPLICATION_TYPE_NAME, OKTA, ORG_SETTING_TYPE_NAME } from '../../src/constants'
+
+describe('appFetchFilter', () => {
+  let client: OktaClient
+  type FilterType = filterUtils.FilterWith<'onFetch' | 'preDeploy' | 'deploy' | 'onDeploy'>
```
shir-reifenberg, line 36: nit
```suggestion
  type FilterType = filterUtils.FilterWith<'onFetch'>
```

packages/okta-adapter/src/definitions/deploy/types/application.ts
```diff
@@ -0,0 +1,96 @@
...
+      () =>
+      ({ change }) => {
+        const getPolicy = (value: Values): boolean => _.get(value, policyName)
+        if (isDefined(getPolicy(getChangeData(change).value))) {
+          if (isAdditionChange(change)) {
+            return true
+          }
+          if (isModificationChange(change)) {
+            const [before, after] = getAllChangeData(change).map(data => getPolicy(data.value))
+            return isDefined(before) && isDefined(after) && before !== after
```
shir-reifenberg, line 53: When this code runs do we compare references or values ?
maybe use `isEqualValues` in case we decide to change when we resolve references ? (I think it would be harder to find in the future)

shir-reifenberg (approved):
Looks great! 🚢 🚀 🛩️

PR #6170, opened 2024-07-08
SALTO-5766: Reduce log chunk size

IdoZakk:
Datadog does not index for search log messages larger than 25K. Our log chunk size was 200K, which made big parts of long messages unsearchable. 
This change reduces the chunk size. It might increase our costs but according to the analysis done by @ori-moisis it should be OK (available in the [Jira ticket](salto-io.atlassian.net/browse/SALTO-5776))
---

None
---
_Release Notes_: 
Core:
* Reduced the default log chunk size to 25K

---
_User Notifications_: 
None

### DIFF
--- a/packages/logging/src/internal/config.ts
+++ b/packages/logging/src/internal/config.ts
@@ -44,7 +44,7 @@ export const DEFAULT_CONFIG: Readonly<Config> = Object.freeze({
   namespaceFilter: '*',
   colorize: null,
   globalTags: {},
-  maxJsonLogChunkSize: 200 * 1024, // 200K
+  maxJsonLogChunkSize: 25 * 1024, // 25K, DataDog recommended limit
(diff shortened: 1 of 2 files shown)


packages/logging/src/internal/config.ts
```diff
@@ -44,7 +44,7 @@ export const DEFAULT_CONFIG: Readonly<Config> = Object.freeze({
   namespaceFilter: '*',
   colorize: null,
   globalTags: {},
-  maxJsonLogChunkSize: 200 * 1024, // 200K
+  maxJsonLogChunkSize: 25 * 1024, // 25K, DataDog recommended limit
```
ori-moisis, line 47: can we remove the reference to DataDog from the OSS please?

ori-moisis (approved):

PR #6171, opened 2024-07-08
SALTO-5692: Okta: Add HTTP recording tests for `Application` deploy

Nurdok:
See title.

---

This just adds a test, without migrating `Application` to the new infra, just to make it clear that the migration PR doesn't cause any regressions here.

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/okta-adapter/test/adapter.test.ts
```diff
@@ -1013,5 +1015,72 @@ describe('adapter', () => {
...
+      beforeEach(() => {
+        appType = new ObjectType({
+          elemID: new ElemID(OKTA, APPLICATION_TYPE_NAME),
+          fields: {
+            id: {
+              refType: BuiltinTypes.SERVICE_ID,
+            },
+          },
+        })
+        app = new InstanceElement('app', appType, {
```
shir-reifenberg, line 1030: should we test application with `profileEnrollment` and `accessPolicy` to verify those as well ? (there are separate requests to update those)

packages/okta-adapter/test/adapter.test.ts
```diff
@@ -1013,5 +1015,72 @@ describe('adapter', () => {
...
+          id: 'app-fakeid1',
+          label: 'app1',
+          status: INACTIVE_STATUS,
+        })
+      })
+
+      it('should successfully add an application', async () => {
```
shir-reifenberg, line 1037: can we make a separate test to deployment of added application in status `ACTIVE` and in status `INACTIVE` ? they differentiate in query params

packages/okta-adapter/test/adapter.test.ts
```diff
@@ -1013,5 +1015,72 @@ describe('adapter', () => {
...
+        const result = await operations.deploy({
+          changeGroup: {
+            groupID: 'app',
+            changes: [toChange({ after: appWithoutId })],
+          },
+          progressReporter: nullProgressReporter,
+        })
+        expect(result.errors).toHaveLength(0)
+        expect(result.appliedChanges).toHaveLength(1)
+        expect(getChangeData(result.appliedChanges[0] as Change<InstanceElement>).value.id).toEqual('app-fakeid1')
```
shir-reifenberg, line 1049: I think we should have a test for "regular" app and "custom" app, cause in custom app we assign the name field back on the application

shir-reifenberg (approved):
Thanks! 🎉 
I added some comments about scenarios that I believe should be tested as well, but approving to unblock.
In addition, is there anyway to verify the order of requests ? there are some nuances in modifications that we should cover IMO

packages/okta-adapter/test/adapter.test.ts
Nurdok: Done

packages/okta-adapter/test/adapter.test.ts
Nurdok: Done

packages/okta-adapter/test/adapter.test.ts
Nurdok: Done

PR #6172, opened 2024-07-08
SAAS-11472: Salesforce Fetch Targets with dependencies bug-fix

tamtamirr:
Fixed a bug where we didn't include the correct dependencies for the provided fetch targets.

---

This caused a weird side-effect where we deleted all the `WorkflowRule` instances in `fetch with changes detection mode` and a targeted fetch on `WorkflowRule`.

- If a type wasn't listed for targeted fetch this is actually bug. Added handling with a log which we will monitor.

---
_Release Notes_: 
Salesforce Adapter:
- Fixed a bug where we didn't include the correct dependencies for the provided fetch targets.

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/fetch_profile/metadata_types.ts
```diff
@@ -653,35 +653,31 @@ export const isMetadataTypeWithoutDependencies = (
...
     metadataType,
   )
 
-export const getFetchTargets = (
-  target: SupportedMetadataType[],
-): SalesforceMetadataType[] => {
-  const allTypes: SalesforceMetadataType[] = [
-    ...target.filter(isMetadataTypeWithoutDependencies),
-  ]
+export const getFetchTargetsWithDependencies = (target: string[]): string[] => {
```
yelly, line 662: Just a semantic issue - having a parameter `target` of type `string[]` just sounds wrong.

packages/salesforce-adapter/src/fetch_profile/metadata_types.ts
```diff
@@ -653,35 +653,31 @@ export const isMetadataTypeWithoutDependencies = (
...
+  const result = _.clone(target)
```
yelly, line 663:
```suggestion
  const result = [...target]
```

yelly (approved):
Just a couple style comments.

PR #6178, opened 2024-07-09
SALTO-6146 PagerDuty cache users between deploy change groups

Uri-Bar:
SALTO-6146 PagerDuty cache users between deploy change groups
---

---
_Release Notes_: 
none

---
_User Notifications_: 
none

### DIFF
--- a/packages/pagerduty-adapter/src/filters/users.ts
+++ b/packages/pagerduty-adapter/src/filters/users.ts
@@ -137,6 +137,7 @@ const filter: filterUtils.AdapterFilterCreator<UserConfig, filterUtils.FilterRes
   config,
   definitions,
   fetchQuery,
+  sharedContext,
 }) => {
   let userIdToLogin: Record<string, string> = {}
   const userDefinition = { ...definitions, fetch: { instances: USER_FETCH_DEFINITIONS } }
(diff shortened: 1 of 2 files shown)


packages/pagerduty-adapter/src/filters/users.ts
```diff
@@ -179,12 +180,14 @@ const filter: filterUtils.AdapterFilterCreator<UserConfig, filterUtils.FilterRes
...
-        definitions: userDefinition as definitionsUtils.RequiredDefinitions<Options>,
-      })
+      if (!sharedContext.users) {
+        sharedContext.users = fetchUtils.getElements({
+          adapterName: ADAPTER_NAME,
+          fetchQuery,
+          definitions: userDefinition as definitionsUtils.RequiredDefinitions<Options>,
+        })
+      }
+      const users = (await sharedContext.users) as fetchUtils.FetchElements<InstanceElement[]>
```
netama, line 190: nit - I know I suggested this as the quick fix, but may be worth adding some type guard instead of the cast (also since the rest of the code always looks at `users.elements`). 
can also start with this since there is probably low risk

netama (approved):
LGTM with two nits

packages/pagerduty-adapter/src/filters/users.ts
Uri-Bar: lets do it later, I gave it few min but it started getting messy so I stopped

PR #6185, opened 2024-07-10
SALTO-6235: Supporting all non-standard types in standard object CV

yelly:
Also improved a warning message that came up during testing.

---

Tested on my env with a custom event.

---
_Release Notes_: 
_Salesforce_:
* Fixed a bug where Custom Events were misidentified as Standard Objects and could not be deployed.

---
_User Notifications_: 
None.


tamtamirr (commented):
I don't think we should change the implementation of `isCustom`.
I'd change the implementation of the problematic CV to use `isStandardObjectSync` which covers all the possible cases :)

tamtamirr (approved):
LGTM

PR #6187, opened 2024-07-10
SALTO-6236: Salesforce CustomMetadata Records deploy fail when attempting to delete field values

tamtamirr:
Bugfix: Salesforce CustomMetadata Records deploy fail when attempting to delete field values

---

It seems like Salesforce started to enforce providing value to the xsi:nil attribute which indicates no value (null). This caused deletions in CustomMetadata Records to fail as we used to send them as <xsi:nil/> instead of <xsi:nil=”true”/>. 

---
_Release Notes_: 
Salesforce Adapter:
- Bugfix: Salesforce CustomMetadata Records deploy fail when attempting to delete field values

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/src/filters/custom_metadata.ts
+++ b/packages/salesforce-adapter/src/filters/custom_metadata.ts
@@ -115,6 +115,8 @@ export type ServiceMDTRecordValue = MetadataValues & {
   values: ServiceMDTRecordFieldValue | ServiceMDTRecordFieldValue[]
 }
 
+// Salesforce expects xsi:nil="true" for null values
+// We pass "true" here since fast-xml-parser  omits these values and no value is needed here
 export const XSI_NIL_TRUE = { [`${XML_ATTRIBUTE_PREFIX}xsi:nil="true"`]: true }
 
 const isServiceMDTRecordFieldValue = (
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/src/filters/custom_metadata.ts
```diff
@@ -219,11 +221,13 @@ const formatRecordValuesForService = async (
           value: { '#text': fieldValue, 'attr_xsi:type': xsdType },
         }
       }
+      // No value was provided for the field or field value was removed
       return {
         field: fieldName,
-        value: { 'attr_xsi:nil': 'true' },
```
yelly, line 224: I'm confused, these should yield the same output.

packages/salesforce-adapter/src/filters/custom_metadata.ts
```diff
@@ -115,6 +115,10 @@ export type ServiceMDTRecordValue = MetadataValues & {
   values: ServiceMDTRecordFieldValue | ServiceMDTRecordFieldValue[]
 }
 
+// Salesforce expects xsi:nil="true" for null values
+// We pass "true" here since fast-xml-parser  omits these values and no value is needed here
```
yelly, line 119:
```suggestion
// Salesforce expects xsi:nil="true" for null values explicitly.
// We put the "true" value in the key since fast-xml-parser omits boolean values.
```

yelly (approved):

PR #6191, opened 2024-07-11
SALTO-6241 turbocache unit tests

eshedg:
add turbocache to unit tests

---

this part was split off of SALTO-3518

---
_Release Notes_: 

with this PR, we now support turbocache for both local tests and CircleCI.
if you need to test a specific package, do (for example):
```
yarn test --filter @salto-io/zendesk-adapter
```

before the turbocache run, we will check if any source files have been modified in any of the dependencies of the package you are testing (or the entire project, if you run all tests).

---
_User Notifications_:

### DIFF
--- a/.circleci/config_template.yml
+++ b/.circleci/config_template.yml
@@ -43,9 +43,9 @@ commands:
       - restore_cache:
           name: Restore node modules
           keys:
-            - &node-modules-cache node-modules-v2-{{ arch }}-{{ .Branch }}-{{ checksum "yarn.lock" }}
-            - node-modules-v2-{{ arch }}-{{ .Branch }}
-            - node-modules-v2-{{ arch }}-main
+            - &node-modules-cache node-modules-v2-{{ .Branch }}-{{ checksum "yarn.lock" }}
+            - node-modules-v2-{{ .Branch }}
+            - node-modules-v2-main
   
   store_node_modules:
     description: Cache store node_modules
@@ -63,9 +63,9 @@ commands:
       - restore_cache:
           name: Restore yarn cache
           keys:
-            - &yarn-cache yarn-cache-v2-{{ arch }}-{{ .Branch }}-{{ checksum "yarn.lock" }}
-            - yarn-cache-v2-{{ arch }}-{{ .Branch }}
-            - yarn-cache-v2-{{ arch }}-main
+            - &yarn-cache yarn-cache-v2-{{ .Branch }}-{{ checksum "yarn.lock" }}
+            - yarn-cache-v2-{{ .Branch }}
+            - yarn-cache-v2-main
 
   store_yarn_cache:
     description: Cache store yarn cache
@@ -131,6 +131,34 @@ commands:
       - restore_intermediate
       - restore_node_modules
       - setup_yarn
+
+  cached_unit_test:
+    parameters:
+      package_name:
+        type: string
+    steps:
+      - step_setup
+      - restore_turbo_cache:
+          cache_name: jest_ut-<< parameters.package_name >>
+      - run:
+          name: Check dependencies for changes
+          command: |
+            export SALTO_DEPENDENCIES_HASH="$(node ./build_utils/hash_dependencies.js -p "$(echo << parameters.package_name >> | cut -d':' -f2)")"
+      - run:
+          name: << parameters.package_name >> unit tests
+          command: |
+            yarn test --filter="$(echo "<< parameters.package_name >>" | cut -d':' -f2)" --  --forceExit --detectOpenHandles --reporters=default --reporters=jest-junit --ci --colors
+          environment:
+            JEST_JUNIT_OUTPUT_DIR: 'reports/unittests'
+
+      - store_test_results:
+          path: reports
+      - store_turbo_cache:
+          cache_name: jest_ut-<< parameters.package_name >>
+
+      - coveralls/upload:
+          parallel: true
+          flag_name: << parameters.package_name >>  
 
   run_e2e_tests:
     parameters:
--- a/.circleci/config_template.yml
+++ b/.circleci/config_template.yml
@@ -132,36 +132,6 @@ commands:
       - restore_node_modules
       - setup_yarn
 
-  cached_unit_test:
-    parameters:
-      package_name:
-        type: string
-    steps:
-      - step_setup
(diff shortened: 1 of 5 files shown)


build_utils/turbo_run.sh
```diff
@@ -19,6 +20,39 @@ if [ -n "$CI" ]; then
...
+      shift
+      filter="$1"
+      break
+    else
+      shift
+    fi
+  done
+
+  if [ -n "$filter" ]; then
+    echo "node ./build_utils/hash_dependencies.js -p $filter"
```
yelly, line 50: Instead of these echos you can just pass `-x` to bash in the shbang:
```
#!/usr/bin/env bash -x
```

turbo.json
```diff
@@ -20,6 +20,16 @@
         ".check-format.cache"
       ]
     },
+    "test": {
+      "dependsOn": [ "^build-ts"],
```
yelly, line 24: The tests shouldn't depend on the build, jest builds on the fly.

build_utils/hash_dependencies.js
```diff
@@ -0,0 +1,76 @@
...
+  } else {
+    if (argv.dependencies) {
+      const allPackages = execSync("yarn workspaces list --json | jq '.name'").toString().trim()
+      console.log(allPackages)
+    }
+    fs.appendFileSync(tempFile.name, execSync(`find packages -type f -name "*.ts"`).toString().trim() + '\n')
+  }
+
+  if (argv.files) {
+    // console.log(dependenciesSourceFiles.split(' ').join('\n'))
```
yelly, line 69: Looks like leftover debug.

build_utils/hash_dependencies.js
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env node
```
yelly, line 1: You could use `ts-node` instead and be able to code like a human here, but up to you.

build_utils/hash_dependencies.js
```diff
@@ -0,0 +1,76 @@
...
+      description: 'Show the dependencies for package_name',
+      demandOption: false,
+    })
+    .option('files', {
+      alias: 'f',
+      type: 'boolean',
+      description: 'Show all the files that package_name depends on',
+      demandOption: false,
+    })
+    .help().argv
```
yelly, line 43: Why all these options if this script is only used once?

build_utils/turbo_run.sh
eshedg: not sure i understand....
what I did here was piggyback on the `--filter` arg that's passed anyhow for the yarn command.  i search for it (-n thingy) and if it's there, i use it.  if not, i do the whole project.  not sure how -x is helpful there...?

turbo.json
eshedg: well _actually_... this is something @ori-moisis and I had a discussion about.  turbo will invalidate tests if the build results have changed.  so what happens is I bring the turbo build cache into the unit test job, and then this will either cache-hit the tests or cache-miss them, if the code changed somewhere.  otherwise, turbo will not detect the change and keep cache-hitting test results that should've failed.

build_utils/turbo_run.sh
yelly: Aren't these `echo`s for debugging?
It looks like you are just printing out what command you are going to run before running it, `-x` gives you that without extra `echo`s.

build_utils/hash_dependencies.js
eshedg: using js for build utils is pretty ubiquitous, so I aligned.  the other option is bash, if you prefer..?

build_utils/hash_dependencies.js
eshedg: it is used in various ways - based on the `--filter` i explained earlier.  
i did a complete implementation for this, because we're in opensource, and because this might be useful in other ways i did not think about.

build_utils/hash_dependencies.js
yelly: No leave it. IMO we should eventually migrate everything to TS but not something critical for now obviously.

.circleci/config_template.yml
```diff
@@ -132,6 +132,36 @@ commands:
...
+    steps:
+      - step_setup
+      - restore_turbo_cache:
+          cache_name: jest_ut-<< parameters.package_name >>
+      - restore_turbo_cache:
+          cache_name: build      
+      - run:
+          name: Check dependencies for changes
+          command: |
+            export SALTO_DEPENDENCIES_HASH="$(node ./build_utils/hash_dependencies.js -p "$(echo << parameters.package_name >> | cut -d':' -f2)")"
```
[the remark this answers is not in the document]
eshedg: this is actually fixed in #6252 , but I'll carry the fix over.

.circleci/config_template.yml
```diff
@@ -132,6 +132,36 @@ commands:
       - restore_node_modules
       - setup_yarn
 
+  cached_unit_test:
```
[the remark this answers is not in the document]
eshedg: wanted to keep the unit_test command clean, but I guess it's not a big deal - I'll merge them.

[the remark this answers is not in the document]
yelly: Whatever works for you, just made the file a bit harder to parse for me.

build_utils/turbo_run.sh
eshedg: oh, gotcha.
yes, i'm echoing the commands to the ci console, so it's easier to understand what's going on.
if i use -x it will be too verbose and less helpful.

.circleci/config_template.yml
```diff
@@ -285,16 +285,30 @@ jobs:
...
       - step_setup
+      - restore_turbo_cache:
+          cache_name: jest_ut-<< parameters.package_name >>
+      - restore_turbo_cache:
+          cache_name: build      
+      - run:
+          name: Extract package name
+          command: |
+            PACKAGE_NAME="$(echo "<< parameters.package_name >>" | cut -d':' -f2)"
+            echo "export E2E_PACKAGE_NAME=$PACKAGE_NAME" >> $BASH_ENV
```
yelly, line 296: Nit: I don't think you mean `E2E` here.

yelly (approved):
LGTM, just fix the last comment.

PR #6199, opened 2024-07-14
SALTO-6250: fix missing user fallback logic when resolveUserIds is false

edenhassid:
fix missing user fallback logic when resolveUserIds is false

---

---
_Release Notes_: 
zendesk:
* fix missing user fallback logic when resolveUserIds is false. setting `defaultMissingUserFallback =  "##DEPLOYER##"` will work.

---
_User Notifications_: 
none


adfineberg (approved):
Nice! A couple of small comments, pre approving

PR #6200, opened 2024-07-14
SALTO-6251: Bump version of jest

ori-moisis:
---

Older versions of jest do not support importing modules according to “exports” definitions in “package.json”.
This is needed for SFDX support because it depends on packages which utilize this feature.

A few categories for changes here:
1. all eslint configuration files were moved and re-formatted to fit the new configuration format (so, .eslintrc and .eslintignore moved to eslint.config.mjs) - migrated automatically with the `@eslint/migrate-config` command
2. all jest config files were changed to fit the new configuration format
3. new eslint configuration warns about unused eslint disables, so a lot of those were removed
4. typescript version bumped to 4.8.4 which is a bit more strict, required a few type fixes

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


packages/adapter-components/test/client/rate_limit.test.ts
```diff
@@ -34,20 +34,20 @@ class A {
     })
   }
 
-  @throttle<MyRateLimitConfig>({})
   // eslint-disable-next-line class-methods-use-this
+  @throttle<MyRateLimitConfig>({})
```
ori-moisis, line 38: new eslint version finally knows how to parse decorators, so the lint disable had to move

packages/adapter-components/src/fetch/element/type_element.ts
```diff
@@ -170,12 +170,14 @@ export const generateType = <Options extends FetchApiDefinitionsOptions>(
   const { element: elementDef } = defQuery.query(typeName) ?? {}
   if (elementDef === undefined) {
     log.debug('could not find any element definitions for type %s', typeName)
-  }
-  if (elementDef?.topLevel?.custom !== undefined) {
-    log.info('found custom override for type %s:%s, using it to generate types', adapterName, typeName)
-    const { types } = elementDef?.topLevel?.custom(elementDef)(args)
```
ori-moisis, line 176: the issue here:
according to the type information at this point, `elementDef?.topLevel?.custom(elementDef)(args)` could potentially be undefined (we know it won't because we checked `elementDef?.topLevel?.custom !== undefined` above, but this is not good enough for typescript) and `const { types } = undefined` is invalid.

changed the conditions a bit to ensure the `custom` function is not undefined at this point.
this should not change the logic (please review and ensure)

packages/core/src/local-workspace/state/state.ts
```diff
@@ -224,7 +224,7 @@ export const localState = (
       }
       yield* yieldWithEOL([
         accountToElementStreams[account],
-        awu([safeJsonStringify({ [account]: accountToDates[account] } || {})]),
```
ori-moisis, line 227: this was an unreachable branch (the `|| {}` could never happen because `{ [account]: ... }` can never be falsy

packages/core/test/telemetry.test.ts
```diff
@@ -325,5 +325,6 @@ describe('telemetry', () => {
       expect(timesHTTPCalled).toEqual(MAX_CONSECUTIVE_RETRIES)
       expect(testDone).toBeTruthy()
     })
+    nock.abortPendingRequests()
```
ori-moisis, line 328: there is a long `delayConnection` above which creates timers.
adding this call in order to clear those timers, otherwise we get leftover timers from this test

packages/salesforce-adapter/test/tsconfig.json
```diff
@@ -4,5 +4,6 @@
     "outDir": "dist/test"
   },
   "include": ["**/*"],
+  "files": ["../../../node_modules/jest-extended/types/index.d.ts"],
```
ori-moisis, line 7: needed because of the newer `ts-jest` version (see https://jest-extended.jestcommunity.dev/docs/getting-started/typescript)

packages/stripe-adapter/test/adapter.test.ts
```diff
@@ -56,7 +56,7 @@ jest.mock('@salto-io/adapter-components', () => {
   }
 })
 
-const mockedAdapterComponents = jest.mocked(adapterComponents, true)
+const mockedAdapterComponents = jest.mocked(adapterComponents)
```
ori-moisis, line 59: new jest api requires a parameter object for the second param (should be `{ shallow: true }`) but `true` is the default (not sure if it was before)

amitizle (commented):
LGTM, the only thing that kinda "scares" me is that `eslint` is not passing on my laptop.
My worry is that something in the cache is broken.
According to [these docs](https://turbo.build/repo/docs/guides/tools/eslint) it's configured fine, maybe we can omit the `outputs` key altogether and make it just `lint: {}` like it's done in there?

jest.base.config.js
```diff
@@ -38,6 +40,8 @@ module.exports = {
     '!coverage/**',
     '!test/**',
     '!e2e_test/**',
```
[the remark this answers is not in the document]
amitizle: We were talking face to face, understanding that `collectCoverageFrom` is unrelated to `testMatch`, so that should stay there.

package.json
```diff
@@ -6,7 +6,7 @@
     "packages/*"
   ],
   "resolutions": {
-    "typescript": "4.1.3"
+    "typescript": "4.8.4"
   },
```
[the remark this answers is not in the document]
amitizle: We're not sure why it's there but for the sake of minimizing the scope of this PR we gonna let it be

amitizle:
We have ran (@ori-moisis and myself) `yarn lint` in a few machines and it looks like it works fine now, so we can move forward with this bump.

amitizle (approved):

PR #6201, opened 2024-07-14
SALTO-6243: Core - getElementHiddenParts attributes handling bug-fix

tamtamirr:
Core - getElementHiddenParts attributes handling bug-fix

---

We should treat attribute values the same way we treat instance values.

---
_Release Notes_: 

_Core_:
- Improved the handling of attribute values in `getElementHiddenParts`. 

---
_User Notifications_: 
_None_


packages/workspace/src/workspace/hidden_values.ts
```diff
@@ -92,6 +92,10 @@ export const getElementHiddenParts = async <T extends Element>(
   const storeHiddenPaths: TransformFunc = async ({ value, field, path }) => {
     if (await hiddenFunc(field, elementsSource)) {
       if (path !== undefined) {
+        // We should hide attribute only when its annotation ref type is _hidden_value
+        if (path.idType === 'attr' && !isHiddenValue(field)) {
+          return undefined
+        }
```
yelly, line 98: We can do this before we check `await hidddenFunc(...)`.

tamtamirr:
Converted to draft since additional handling is needed here for Field annotations (This only tackles ObjectTypes)

packages/workspace/src/workspace/hidden_values.ts
```diff
@@ -90,6 +92,15 @@ export const getElementHiddenParts = async <T extends Element>(
   // or a _hidden_value annotation on its type definition.
   const hiddenFunc = isInstanceElement(stateElement) ? isHiddenValue : isHidden
   const storeHiddenPaths: TransformFunc = async ({ value, field, path }) => {
+    // If the field type is hidden value, we should hide all of its attributes
+    if (isField(value) && isHiddenValue(await value.getType(elementsSource))) {
+      hiddenPaths.add(value.elemID.getFullName())
+      return undefined
+    }
+    // We should hide an attribute only when its annotation ref type is a hidden value
+    if (path && (path?.isAttrID() || isFieldAnnotationId(path)) && !isHiddenValue(field)) {
```
yelly, line 101: You shouldn't need to use optional field access on `path` after checking it's defined, maybe you need to explicitly check `path !== undefined`.

packages/workspace/src/workspace/hidden_values.ts
```diff
@@ -90,6 +92,15 @@ export const getElementHiddenParts = async <T extends Element>(
...
+      return undefined
```
yelly, line 102: Don't we need to `return value` here?

packages/workspace/src/workspace/hidden_values.ts
```diff
@@ -144,10 +156,11 @@ export const getElementHiddenParts = async <T extends Element>(
       // Keep traversing as long as we might reach nested hidden parts.
       // Note: it is ok to check isAncestorOfHiddenPath before isNestedHiddenPath, because there is
       // no overlap between ancestorsOfHiddenPaths and hiddenPaths
-      path !== undefined && (isAncestorOfHiddenPath(path) || isNestedHiddenPath(path)) ? value : undefined,
+      path !== undefined && (isNestedHiddenPath(path) || isAncestorOfHiddenPath(path)) ? value : undefined,
```
yelly, line 159: Why the change of order here?

packages/workspace/test/workspace/hidden_values.test.ts
```diff
@@ -728,4 +730,124 @@ describe('handleHiddenChanges', () => {
...
     })
   })
+  describe('getElemHiddenParts', () => {
+    describe('ObjectType attribute handling', () => {
+      let elementsSource: ReadOnlyElementsSource
+      let testElement: ObjectType
+      let annotationType: ObjectType
+
+      beforeEach(() => {
+        annotationType = new ObjectType({
```
yelly, line 740: Nit: this should be a primitive string type, no?

packages/workspace/test/workspace/hidden_values.test.ts
```diff
@@ -728,4 +730,124 @@ describe('handleHiddenChanges', () => {
...
+            value1: 'test',
+            value2: 'test',
+          },
+          annotationRefsOrTypes: {
+            value1: annotationType,
+            value2: BuiltinTypes.STRING,
+          },
+        })
+        elementsSource = buildElementsSourceFromElements([annotationType, testElement])
+      })
```
yelly, line 755: Nit: newline here.

packages/workspace/test/workspace/hidden_values.test.ts
```diff
@@ -728,4 +730,124 @@ describe('handleHiddenChanges', () => {
...
+        elementsSource = buildElementsSourceFromElements([
+          testElement,
+          hiddenFieldWithNoHiddenAnnotations,
+          hiddenFieldWithNoHiddenAnnotations,
+          hiddenValueFieldWithNoHiddenAnnotations,
+        ])
+      })
+
+      it('should return ObjectType with all of the hiddenValueFieldWithNoHiddenAnnotations and the specific annotations from fieldWithPartiallyHiddenAnnotations', async () => {
+        const result = (await getElementHiddenParts(testElement, elementsSource)) as ObjectType
```
yelly, line 838: Why do you need the case here? The generic should take care of it.
If it's just to deal with it being defined, I personally think it's better to explicitly check for that instead of `expect(result).toBeDefined()`, but up to you.

packages/workspace/test/workspace/hidden_values.test.ts
```diff
@@ -728,4 +730,124 @@ describe('handleHiddenChanges', () => {
...
+        const { fields } = result
+        expect(Object.keys(fields)).toIncludeSameMembers([
+          'fieldWithPartiallyHiddenAnnotations',
+          'hiddenValueFieldWithNoHiddenAnnotations',
+        ])
+        const { fieldWithPartiallyHiddenAnnotations, hiddenValueFieldWithNoHiddenAnnotations } = fields
+        expect(fieldWithPartiallyHiddenAnnotations).toBeDefined()
+        expect(hiddenValueFieldWithNoHiddenAnnotations).toBeDefined()
+        expect(fieldWithPartiallyHiddenAnnotations.annotations).toEqual({ hiddenStringValue: 'testHidden' })
+        expect(hiddenValueFieldWithNoHiddenAnnotations.annotations).toEqual({ stringValue: 'test', booleanValue: true })
```
yelly, line 849: Why not just match the whole object explicitly?

packages/workspace/src/workspace/hidden_values.ts
tamtamirr: leftover from previous iteration

packages/workspace/src/workspace/hidden_values.ts
tamtamirr: nope, the reason I return undefined here is in order to stop looping deeper. At this point we will hide all the values nested under the field.

tamtamirr (commented):
added some comments for clarification 🙏

packages/workspace/src/workspace/hidden_values.ts
tamtamirr: I forgot to post my clarifications comments they were pending 🤦 my bad

packages/workspace/test/workspace/hidden_values.test.ts
tamtamirr: I use it later in my tests and mark this as hidden_value

ori-moisis (approved):

PR #6202, opened 2024-07-14
SALTO-6251: remove import between salesforce test and e2e

ori-moisis:
Since the tests and e2e tests are not part of the same typescript project, there can't be imports from one to the other

---

This was the minimal code duplication I could make, trying to have this really shared (by letting one project reference the other or creating a third project) turned out to be more difficult...

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


yelly (approved):

ori-moisis:
not merging, will disable the check in my branch (as it is disabled today) to allow solving different issues (this issue vs. upgrading the jest version) in different PRs

PR #6203, opened 2024-07-14
SALTO-6188 Add support for API Token for Zendesk

adfineberg:
Zendesk is deprecating username/password and adding API token support

---

---
_Release Notes_: 
__Zendesk Adapter:__ 
* Users can now authenticate to Zendesk with username and API Token

---
_User Notifications_: 


edenhassid (approved):
LGTM :) I'm not sure I understand how do we know if the user is using a token or a password

PR #6205, opened 2024-07-14
SALTO-3845: cancel workflow post functions sort in DC

IdoZakk:
Cancled the sorting of post functions
---

See Jira ticket for details. We are skipping noise reduction is it is complicated and will only influence one customer which will be communicated personally
---
_Release Notes_: 
Jira adapter:
* keep post functions order in DC

---
_User Notifications_: 
Jira adapter:
* for DC users the order of post functions in workflow will change

### DIFF
--- a/packages/jira-adapter/src/filters/sort_lists.ts
+++ b/packages/jira-adapter/src/filters/sort_lists.ts
@@ -145,7 +145,10 @@ const VALUES_TO_SORT: Record<string, Record<string, string[]>> = {
 
 const getValue = (value: Value): Value => (isResolvedReferenceExpression(value) ? value.elemID.getFullName() : value)
 
-const sortLists = async (instance: InstanceElement): Promise<void> => {
+const sortLists = async (instance: InstanceElement, isDataCenter: boolean): Promise<void> => {
+  if (isDataCenter) {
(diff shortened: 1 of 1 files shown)


packages/jira-adapter/src/filters/sort_lists.ts
```diff
@@ -145,7 +145,10 @@ const VALUES_TO_SORT: Record<string, Record<string, string[]>> = {
 
 const getValue = (value: Value): Value => (isResolvedReferenceExpression(value) ? value.elemID.getFullName() : value)
 
-const sortLists = async (instance: InstanceElement): Promise<void> => {
+const sortLists = async (instance: InstanceElement, isDataCenter: boolean): Promise<void> => {
+  if (isDataCenter) {
```
kami-delaroz, line 149: Could you please add a comment about removing this once PROD-321 fully rolled-out?

kami-delaroz (approved):
🥇

packages/jira-adapter/src/filters/sort_lists.ts
kami-delaroz: (after merging and bumping and staff)

PR #6214, opened 2024-07-15
SALTO-6266 Upgrade Stripe to new infra

netama:
Upgrade to new infra and cleanup unneeded code.
Removed the e2e's since they were not running on a real account and only tested the types - we can add them back with better coverage later.
(not attempting to change the implementation - only marking service ids)

(to review the minor changes in the mock replies you can look at the first commit)

---
_Release Notes_: 
None

---
_User Notifications_: 
None


edenhassid (approved):
LGTM! 😄 🔥 why did you remove the change validator?

netama:
@edenhassid  the adapter-wrapper already provides the same validators (including the no-deploy one when there are no deploy definitions)

PR #6215, opened 2024-07-15
SALTO-5392 simplify default field csutomizations

netama:
no need for the two separate structures

---
_Release Notes_: 
None

---
_User Notifications_: 
None


shoshana-houri (approved):
🧼 🚀

PR #6217, opened 2024-07-16
SALTO-6221 and SALTO-6060 support criteria in adapter wrapper and use criteria definition to exclude unwanted spaces

omrifarkash:
- SALTO-6060 add support in giving criteria functions to `adapter_creator` of `adapter-wrapper`. If no criteria functions were provided, use default name criterion function
- SALTO-6221 change default config to exclude space with attribute `type = 'personal'` or  `status = 'archived'`.
Provide relevant criteria functions to these criterions so user can define those on other types than `space`.
Add special logic that calculates spaces fetch requests given user config. so if user exclude some space types or space statuses we won't fetch them in the first place

---

---

_Release Notes_: 
_Confluence Adapter_:
- Support exclude space specific types or statuses through user config

---

_User Notifications_: 
_Confluence Adapter_:
- Support exclude space specific types or statuses through user config


packages/adapter-components/src/adapter-wrapper/adapter/adapter.ts
```diff
@@ -120,6 +120,7 @@ export class AdapterImpl<
     dependencyChangers,
     referenceResolver,
     fixElements,
+    criterionFunctions,
```
netama, line 123: nit: maybe `fetchCriteria` so that this is aligned with the existing terms?

packages/confluence-adapter/src/definitions/utils/space.ts
```diff
@@ -109,3 +114,29 @@ export const spaceChangeGroupWithItsHomepage: deployment.grouping.ChangeIdFuncti
   }
   return changeData.elemID.getFullName()
 }
+
+const addParamsToSpaceUrl = (params: Record<string, string[]>): `/${string}` => {
+  const paramEntries = Object.entries(params)
+  return paramEntries.reduce(
```
netama, line 120: in general, `reduce` is usually hard to read and I don't think there's a need for the complexity here - I think this could have been written using a simple `join` and some conditions instead?
in this case though - the args shouldn't be in the path, they should be in the query args (which are a dictionary already and would have saved this extra work), no? what am I missing?

packages/confluence-adapter/src/definitions/utils/space.ts
```diff
@@ -109,3 +114,29 @@ export const spaceChangeGroupWithItsHomepage: deployment.grouping.ChangeIdFuncti
...
+    `${SPACE_URL_WITHOUT_PARAMS}?`,
+  ) as `/${string}`
+}
+
+/*
+ * Fetch spaces endpoint with the relevant params according to the user config.
+ * We fetch all space types and statuses by default, unless specified otherwise in the user config.
+ */
+export const getFetchSpacesEndpoint = (userConfig: UserConfig): `/${string}` => {
+  const excludeSpaceDefs = userConfig.fetch.exclude.filter(e => e.type === 'space')
```
netama, line 132: we are making too specific assumptions here around what the user would type in exactly - for example, I think we should be leveraging the query's `isTypeMatch` to decide if a specific rule applies to spaces or not, so that we also catch "types starting with `sp`" for example.
we should be careful of over-excluding though - let's start by mapping out the cases (e.g. a few "include" rules with different space types, vs a few "exclude" rules with different space types, vs additional conditions)?

that said - since this is an optimization and there can be some edge cases, it's ok if we only optimize some approaches for now as long as we can explain what is and isn't covered.

packages/confluence-adapter/src/definitions/utils/space.ts
omrifarkash: We cannot add the same param more than onces in the query args. For example, we are querying more than 1 type and we need to concat it like `?type=firstType&type=secondType`
We can do it I think by setting a single query arg with multiple values. something like 
`queryArgs = {
   type: 'firstType&type=secondType'
}`

I didn't like this way very much. WDYT?

packages/confluence-adapter/src/definitions/utils/space.ts
omrifarkash: I don't think we want to add some queryArgs to types that match `isTypeMatch`, I think we need to be very specific here that we are adding those to space fetch URL only. `type = "s"` or `type = ".*"` will catch many unwanted types, and we could not expect how the query will react to adding these params.
This is the reason I am working with a "black list" and excluding things from a predefined types and statuses that I know are valid.

Also, I think that is ok to work only with exclude here and let it overcome the include definitions. But we can talk about it f2f if you like

packages/confluence-adapter/src/definitions/utils/space.ts
netama: I think we had some places where we did it in the past, maybe an array? it depends on how the underlying library converts it. either way I think making it work "correctly" is better than hiding in the URL itself, unless there's no reasonable way to do it

omrifarkash: I haven't succeed to do it other way, we can try together if you like.
I am not sure that this is "hiding" the params in the URL, as `axios` eventually resolves these params into the URL in the same way, also the util function name is applying of what we are doing here

packages/adapter-components/src/adapter-wrapper/adapter/adapter.ts
omrifarkash: Named it `allCriteria` as this is the name of the parameter of `createElementQuery`

packages/confluence-adapter/test/definitions/utils/space.test.ts
```diff
@@ -261,4 +266,71 @@ describe('space definitions utils', () => {
...
+        queryArgs: {
+          anExistingQueryArg: 'someValue',
+        },
+      },
+    } as unknown as definitions.fetch.FetchRequestDefinition<definitions.ResolveClientOptionsType<Options>>
+    const createMockUserConfig = (
+      statusesToExclude: string[],
+      typesToExclude: string[],
+      statusesToInclude: string[] = [],
+      typesToInclude: string[] = [],
```
kami-delaroz, line 281: nit: this should be an object, for easier tests reading

packages/confluence-adapter/test/definitions/utils/space.test.ts
```diff
@@ -261,4 +266,71 @@ describe('space definitions utils', () => {
...
+      expect(requests).toHaveLength(1)
+      expect(requests).toEqual([mockRequest])
+    })
+    it('should return the given request with modified status param when user include all types and a single status', () => {
+      const requests = getSpaceRequests(
+        createMockUserConfig([], [], ['current'], ['knowledge_base', 'global', 'personal', 'collaboration']),
+        mockRequest,
+      )
+      expect(requests).toHaveLength(1)
+      expect(requests).toEqual([_.merge({}, mockRequest, { endpoint: { queryArgs: { status: 'current' } } })])
```
kami-delaroz, line 312: nit: in tests the best practice is to have the "flat" object, like you have in the next test, and not the same logic as within the code you're testing. 
In this particular case it's fine as it's regular array functions, but I think it could be cleared if you put the expected result like you have in the next test. No string opinion though

packages/confluence-adapter/src/definitions/utils/space.ts
```diff
@@ -109,3 +115,54 @@ export const spaceChangeGroupWithItsHomepage: deployment.grouping.ChangeIdFuncti
...
+export const getSpaceRequests = (
+  userConfig: UserConfig,
+  spaceRequest: definitions.fetch.FetchRequestDefinition<definitions.ResolveClientOptionsType<Options>>,
+): definitions.fetch.FetchRequestDefinition<definitions.ResolveClientOptionsType<Options>>[] => {
+  const excludeSpaceDefs = userConfig.fetch.exclude.filter(exclude => isSpaceTypeMatch(exclude.type))
+  const includeSpaceDefs = userConfig.fetch.include.filter(include => isSpaceTypeMatch(include.type))
+  const spaceTypesToFetch = getTypesOrStatusesToFetch({ excludeSpaceDefs, includeSpaceDefs, query: 'type' })
+  const spaceStatusesToFetch = getTypesOrStatusesToFetch({ excludeSpaceDefs, includeSpaceDefs, query: 'status' })
+  const requestQueryArgs = spaceRequest.endpoint?.queryArgs
+  if (spaceTypesToFetch.length === 0 && spaceStatusesToFetch.length === 0) {
```
kami-delaroz, line 158: should be ||

packages/confluence-adapter/test/definitions/utils/space.test.ts
```diff
@@ -261,4 +266,71 @@ describe('space definitions utils', () => {
...
+        ],
+        exclude: [
+          ...statusesToExclude.map(status => ({ type: 's.*', criteria: { status } })),
+          ...typesToExclude.map(type => ({ type: 'space', criteria: { type } })),
+        ],
+      },
+    })
+    it('should return no requests when user exclude all statuses and types', () => {
+      const requests = getSpaceRequests(
+        createMockUserConfig(['current'], ['personal'], ['current'], ['personal']),
```
kami-delaroz, line 296: nit: a test with only status in include and exclude (not both status and type) - could catch the bug above

kami-delaroz (approved):

PR #6221, opened 2024-07-16
SALTO-6276: Format salesforce deployed XMLs

ori-moisis:
This is a step towards supporting SFDX format.
Formatting the XMLs generated by the adapter to match the formatting options that SFDX uses

---

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/src/transformers/xml_transformer.ts
+++ b/packages/salesforce-adapter/src/transformers/xml_transformer.ts
@@ -516,17 +516,18 @@ const builder = new XMLBuilder({
   indentBy: '    ',
 })
 
-const XML_VERSION_ENCODING_HEADER = '<?xml version="1.0" encoding="UTF-8"?>\n'
 const SALESFORCE_XML_NAMESPACE_URL = 'http://soap.sforce.com/2006/04/metadata'
 
 const toMetadataXml = (name: string, values: Values): string =>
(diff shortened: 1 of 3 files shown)


packages/salesforce-adapter/src/transformers/xml_transformer.ts
```diff
@@ -512,10 +512,21 @@ export const fromRetrieveResult = async (
...
+  format: true,
+  indentBy: '    ',
 })
 
+const XML_VERSION_ENCODING_HEADER = '<?xml version="1.0" encoding="UTF-8"?>\n'
+const SALESFORCE_XML_NAMESPACE_URL = 'http://soap.sforce.com/2006/04/metadata'
+
 const toMetadataXml = (name: string, values: Values): string =>
-  builder.build({ [name]: _.omit(values, INSTANCE_FULL_NAME_FIELD) })
+  XML_VERSION_ENCODING_HEADER.concat(
```
yelly, line 523: This feels a bit hacky to me. We can achieve the same outcome by passing `builder.build` an array. You can look at the value returned from `parser.parse` to see what it needs to look like to recreate the same XML.

ori-moisis: good point, fixed

yelly (approved):

PR #6222, opened 2024-07-16
SALTO-6278 ensure that updateNaclFiles get detailedChanges with baseChange in them

dantal4:
we expect the input changes in `updateNaclFiles` to have `baseChange` in them (`DetailedChangeWithBaseChange`). this is trivial in most places because the core returns `DetailedChangeWithBaseChange[]` in most of its flows, but in some places we were building new `DetailedChange[]` and there we should make sure that they have `baseChange` too.

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/core/src/core/rename.ts
+++ b/packages/core/src/core/rename.ts
@@ -5,13 +5,21 @@
  *
  * CERTAIN THIRD PARTY SOFTWARE MAY BE CONTAINED IN PORTIONS OF THE SOFTWARE. See NOTICE FILE AT https://github.com/salto-io/salto/blob/main/NOTICES
  */
-import { ElemID, Element, isElement, InstanceElement, DetailedChange } from '@salto-io/adapter-api'
+import {
+  ElemID,
+  Element,
+  isElement,
+  InstanceElement,
+  DetailedChangeWithBaseChange,
+  toChange,
+} from '@salto-io/adapter-api'
 import { collections, values } from '@salto-io/lowerdash'
-import { transformElement, references as referencesUtils } from '@salto-io/adapter-utils'
+import { references as referencesUtils, detailedCompare, setPath, getDetailedChanges } from '@salto-io/adapter-utils'
 import { ElementsSource, getElementsPathHints, PathIndex, splitElementByPath, Workspace } from '@salto-io/workspace'
 
 const { awu } = collections.asynciterable
 const { isDefined } = values
+const { getReferences, getUpdatedReference } = referencesUtils
 
 export class RenameElementIdError extends Error {
   constructor(message: string) {
@@ -61,69 +69,47 @@ export const renameChecks = async (workspace: Workspace, sourceElemId: ElemID, t
   }
 }
 
-export const updateElementReferences = async (
-  element: InstanceElement,
-  sourceElemId: ElemID,
-  targetElemId: ElemID,
-  elementsSource?: ElementsSource,
-): Promise<InstanceElement> =>
-  transformElement({
-    element,
-    transformFunc: referencesUtils.createReferencesTransformFunc(sourceElemId, targetElemId),
-    elementsSource,
-    strict: false,
-  })
-
-const getRenameElementChanges = async (
-  elementsSource: ElementsSource,
+const getRenameElementChanges = (
   sourceElemId: ElemID,
   targetElemId: ElemID,
   sourceElements: InstanceElement[],
-): Promise<DetailedChange[]> => {
-  const removeChange = {
-    id: sourceElemId,
-    action: 'remove' as const,
-    data: {
-      before: sourceElements[0],
-    },
-  }
+): DetailedChangeWithBaseChange[] => {
+  const removeChanges = getDetailedChanges(toChange({ before: sourceElements[0] }))
 
-  // updating references inside the renamed element
-  const updatedElements = await Promise.all(
-    sourceElements.map(element => updateElementReferences(element, sourceElemId, targetElemId, elementsSource)),
+  const newElements = sourceElements.map(
+    element =>
+      new InstanceElement(targetElemId.name, element.refType, element.value, element.path, element.annotations),
   )
 
-  const addChanges = updatedElements.map(element => ({
-    id: targetElemId,
-    action: 'add' as const,
-    data: {
-      after: new InstanceElement(targetElemId.name, element.refType, element.value, element.path, element.annotations),
-    },
-  }))
+  newElements.forEach(element =>
+    getReferences(element, sourceElemId).forEach(({ path, value }) =>
+      setPath(element, path, getUpdatedReference(value, targetElemId)),
+    ),
+  )
 
-  return [removeChange, ...addChanges]
+  const addChanges = newElements.flatMap(element => getDetailedChanges(toChange({ after: element })))
+
+  return removeChanges.concat(addChanges)
 }
 
 const getRenameReferencesChanges = async (
   elementsSource: ElementsSource,
   sourceElemId: ElemID,
   targetElemId: ElemID,
-): Promise<DetailedChange[]> => {
-  const references = await awu(await elementsSource.getAll())
+): Promise<DetailedChangeWithBaseChange[]> =>
+  awu(await elementsSource.getAll())
     // filtering the renamed element - its references are taken care in getRenameElementChanges
     .filter(element => !sourceElemId.isEqual(element.elemID))
-    .flatMap(element => referencesUtils.getReferences(element, sourceElemId))
+    .flatMap(async element => {
+      const references = getReferences(element, sourceElemId)
+      if (references.length === 0) {
+        return []
+      }
+      const cloned = element.clone()
+      references.forEach(({ path, value }) => setPath(cloned, path, getUpdatedReference(value, targetElemId)))
+      return detailedCompare(element, cloned, { createFieldChanges: true })
+    })
     .toArray()
-
-  return references.map(reference => ({
-    id: reference.path,
-    action: 'modify',
-    data: {
-      before: reference.value,
-      after: referencesUtils.getUpdatedReference(reference.value, targetElemId),
-    },
-  }))
-}
 
 const renameElementPathIndex = async (
   index: PathIndex,
@@ -152,11 +138,11 @@ export const renameElement = async (
   sourceElemId: ElemID,
   targetElemId: ElemID,
   index?: PathIndex,
-): Promise<DetailedChange[]> => {
+): Promise<DetailedChangeWithBaseChange[]> => {
   const source = await elementsSource.get(sourceElemId)
   const elements = isDefined(index) ? await splitElementByPath(source, index) : [source]
 
-  const elementChanges = await getRenameElementChanges(elementsSource, sourceElemId, targetElemId, elements)
+  const elementChanges = getRenameElementChanges(sourceElemId, targetElemId, elements)
   const referencesChanges = await getRenameReferencesChanges(elementsSource, sourceElemId, targetElemId)
 
   if (isDefined(index)) {
--- a/packages/workspace/src/workspace/hidden_values.ts
+++ b/packages/workspace/src/workspace/hidden_values.ts
@@ -738,7 +738,7 @@ export const filterOutHiddenChanges = async (
       return { visible: change, hidden: change }
     }
 
-    const { parent } = change.id.createTopLevelParentID()
+    const { parent, path } = change.id.createTopLevelParentID()
     const baseElem = change.id.isTopLevel() ? change.data.after : await state.get(parent)
 
     if (baseElem === undefined) {
@@ -775,7 +775,6 @@ export const filterOutHiddenChanges = async (
         changeType: TypeElement | undefined
         changePath: ReadonlyArray<string>
       }> => {
-        const { path } = change.id.createTopLevelParentID()
         if (change.id.isAttrID()) {
           return {
             changeType: (await elementAnnotationTypes(baseElem, state))[path[0]],
--- a/packages/workspace/src/workspace/hidden_values.ts
+++ b/packages/workspace/src/workspace/hidden_values.ts
@@ -16,8 +16,9 @@ import {
   elementAnnotationTypes,
   safeJsonStringify,
   resolvePath,
-  applyDetailedChanges,
   getDetailedChanges,
+  detailedCompare,
+  setPath,
 } from '@salto-io/adapter-utils'
 import {
   CORE_ANNOTATIONS,
@@ -51,10 +52,11 @@ import {
   isTypeReference,
   DetailedChangeWithBaseChange,
   toChange,
+  isFieldChange,
+  getChangeData,
 } from '@salto-io/adapter-api'
 import { mergeElements, MergeResult } from '../merger'
 import { State } from './state'
-import { createAddChange, createRemoveChange } from './nacl_files/multi_env/projections'
 import { ElementsSource } from './elements_source'
 import { splitElementByPath } from './path_index'
 
@@ -449,7 +451,7 @@ const getHiddenFieldAndAnnotationValueChanges = async (
   )
 
   const createHiddenValueChangeIfNeeded =
-    (hiddenValueChanges: DetailedChange[]): TransformFunc =>
+    (visibleElement: Element): TransformFunc =>
     ({ value, field, path }) => {
       if (path === undefined) {
         return value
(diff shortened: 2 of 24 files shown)


packages/core/src/core/rename.ts
```diff
@@ -87,27 +94,29 @@ const getRenameElementChanges = async (
   sourceElemId: ElemID,
   targetElemId: ElemID,
   sourceElements: InstanceElement[],
-): Promise<DetailedChange[]> => {
+): Promise<DetailedChangeWithBaseChange[]> => {
+  const baseRemoveChange = toChange({ before: sourceElements[0] })
```
ori-moisis, line 98: instead of building the change explicitly
```suggestion
  const removeChanges = getDetailedChanges(toChange({ before: sourceElements[0] }))
```

applies to several places in this PR (in other files as well)

packages/core/src/core/rename.ts
```diff
@@ -87,27 +94,29 @@ const getRenameElementChanges = async (
...
   // updating references inside the renamed element
   const updatedElements = await Promise.all(
     sourceElements.map(element => updateElementReferences(element, sourceElemId, targetElemId, elementsSource)),
   )
 
-  const addChanges = updatedElements.map(element => ({
-    id: targetElemId,
-    action: 'add' as const,
-    data: {
+  const addChanges = updatedElements.map(element => {
```
ori-moisis, line 110: similar case
```suggestion
  const addChanges = updatedElements.flatMap(element => getDetailedChanges(toChange({ after: ... })))
```

packages/core/src/core/rename.ts
```diff
@@ -116,23 +125,16 @@ const getRenameReferencesChanges = async (
...
-  const references = await awu(await elementsSource.getAll())
+): Promise<DetailedChangeWithBaseChange[]> =>
+  awu(await elementsSource.getAll())
     // filtering the renamed element - its references are taken care in getRenameElementChanges
     .filter(element => !sourceElemId.isEqual(element.elemID))
-    .flatMap(element => referencesUtils.getReferences(element, sourceElemId))
+    .flatMap(async element => {
+      const updatedElement = await updateElementReferences(element, sourceElemId, targetElemId)
+      return detailedCompare(element, updatedElement)
+    })
```
ori-moisis, line 135: this will create a huge impact on runtime
running `transformElement` on all the elements of a workspace could take a lot of time and it will create a significant increase in memory consumption.

I wonder if this pushes us over the edge and we would need to optimize `updateElementReferences` now (avoid using transformElement there, instead, go over references and replace each one separately)

packages/workspace/src/workspace/nacl_files/multi_env/projections.ts
```diff
@@ -111,6 +113,14 @@ export const createAddChange = (value: Element | Value, id: ElemID, path?: Reado
   path,
 })
 
+export const createElementAddChange = (
```
ori-moisis, line 116: I think we should not have this here
this is a very generic utility function that kind of already exists in `adapter-utils`.
if we need an even shorter way to write `getDetailedChanges(toChange({ ...}))` we can add that to `adapter-utlis`, but I don't think it makes sense to put that in `workspace/nacl_files/multi_env/projections`...

packages/workspace/src/workspace/nacl_files/multi_env/routers.ts
```diff
@@ -575,17 +588,25 @@ export const routeChanges = async (
...
   return awu(Object.entries(groupedByTopLevel))
     .flatMap(async ([key, groupedIds]) => {
       const targetTopElement = await targetSource.get(ElemID.fromFullName(key))
       if (targetTopElement === undefined) {
         return []
       }
-      return groupedIds.map(id => createRemoveChange(resolvePath(targetTopElement, id), id))
+      const removeChanges = groupedIds.map(id => createRemoveChange(resolvePath(targetTopElement, id), id))
+      if (removeChanges.find(change => change.id.isEqual(targetTopElement.elemID)) !== undefined) {
+        return removeChanges.map(change => ({ ...change, baseChange: toChange({ before: targetTopElement }) }))
```
ori-moisis, line 604: can't we find this case before creating all the remove changes? (check if the top level ID is in `groupedIds`?)
and why do we need to edit the change here? can't we create it with the right base change in the first place in this case?


also, I think this is wrong, it will put the top level element as a `baseChange`, but "base" and "top" are not the same thing, this will create the wrong thing for field changes

packages/workspace/src/workspace/nacl_files/multi_env/routers.ts
```diff
@@ -430,16 +442,16 @@ export const routeIsolated = async (
   return {
     // No need to apply addToSource to primary env changes since it was handled by the original plan
     primarySource: [...currentEnvChanges, ...addCommonProjectionToCurrentChanges],
-    commonSource: [createRemoveChange(currentCommonElement, change.id, pathHint)],
+    commonSource: [{ ...createRemoveChange(currentCommonElement, change.id, pathHint), baseChange: change.baseChange }],
```
ori-moisis, line 445: I might be wrong, but IIUC the `baseChange` here might not fit a removal change, right?
meaning, this could create a detailed removal change where the base change is a modification that did not delete that value, right?

not sure if this is a very important issue, just trying to understand if this is correct so we'll know that it is a scenario that needs to be considered in every place that deals with removal changes (we can't rely on the base change there...)

packages/core/src/core/rename.ts
dantal4: we can do that

dantal4: `updateElementReferences` isn't used outside of `rename.ts` so it would be easy to do that.

packages/workspace/src/workspace/nacl_files/multi_env/routers.ts
dantal4: you're correct.
that raises a similar question in more places like the separation of visible & hidden parts - we don't separate `baseChange` into visible & hidden parts, so `visible.baseChange` includes also hidden parts in it.

packages/workspace/src/workspace/hidden_values.ts
```diff
@@ -505,27 +516,31 @@ const getHiddenFieldAndAnnotationValueChanges = async (
...
-        transformFunc: createHiddenValueChangeIfNeeded,
+        transformFunc: createHiddenValueChangeIfNeeded(hiddenValueChanges),
         strict: true,
         elementsSource: state,
         runOnFields: true,
         allowEmptyArrays: true,
         allowEmptyObjects: true,
       })
+      const after = element.clone()
+      applyDetailedChanges(after, hiddenValueChanges)
```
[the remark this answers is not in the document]
dantal4: fixed

packages/workspace/src/workspace/nacl_files/multi_env/multi_env_source.ts
```diff
@@ -355,6 +356,7 @@ const buildMultiEnvSource = (
             id: changes[0].id,
             elemIDs: { before: changes[0].id },
             path: changes[0].path,
+            baseChange: changes[0].baseChange,
```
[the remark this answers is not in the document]
dantal4: changing this to be a removal too.

packages/workspace/src/workspace/nacl_files/multi_env/routers.ts
```diff
@@ -443,16 +454,24 @@ export const routeIsolated = async (
...
     // No need to apply addToSource to primary env changes since it was handled by the original plan
     primarySource: [...currentEnvChanges, ...addCommonProjectionToCurrentChanges],
-    commonSource: [createRemoveChange(currentCommonElement, change.id, pathHint)],
+    commonSource: [
+      {
+        data: { before: currentCommonElement },
+        action: 'remove',
+        id: change.id,
+        path: pathHint,
+        baseChange: change.baseChange,
```
dantal4, line 463: Should fix

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -233,6 +239,7 @@ export const groupAnnotationTypeChanges = (
           .fromPairs()
           .value(),
       },
+      baseChange: change.baseChange,
```
dantal4, line 242: Should fix

packages/workspace/src/workspace/hidden_values.ts
```diff
@@ -505,27 +516,31 @@ const getHiddenFieldAndAnnotationValueChanges = async (
...
+    .flatMap(async element => {
+      const before = await visibleElementSource.get(element.elemID)
+      if (!isElement(before)) {
+        return []
+      }
+      const after = before.clone()
       await transformElement({
         element,
-        transformFunc: createHiddenValueChangeIfNeeded,
+        transformFunc: createHiddenValueChangeIfNeeded(after),
```
ori-moisis, line 528: this is a non trivial implementation and using a methodology that we really don't want to encourage (having a function modify its input and relying on side effects).
I think we should add a comment saying what we are doing here and why (the comment can probably go above the `const after = before.clone()`)

it is not easy to understand just from the code...

packages/workspace/src/workspace/nacl_files/multi_env/multi_env_source.ts
```diff
@@ -334,18 +335,24 @@ const buildMultiEnvSource = (
...
-  ): DetailedChange<T> & AdditionChange<T> => ({
+    change: DetailedChangeWithBaseChange & ModificationChange<T>,
+  ): DetailedChangeWithBaseChange => ({
     action: 'add',
     data: { after: change.data.after },
     id: change.id,
     elemIDs: change.elemIDs?.after ? { after: change.elemIDs.after } : undefined,
+    baseChange: change.id.isBaseID()
+      ? toChange({ after: change.data.after })
+      : // in this case `baseChange` will be incorrect because `change.baseChange.before` will contain
```
ori-moisis, line 346: if this is true, it means the condition `change.id.isBaseID()` is always true and this just adds dead code.
in that case, why not have the comment saying this can only be called with a base change and remove this branch of the code?

if you want to be extra safe you could add a check and a warning log, but tbh, I am not sure that is really needed since this is a fairly "local" function, so the risk of anyone trying to use it in a different scenario without encountering the comment saying it must be called with a base ID is fairly low

packages/workspace/src/workspace/nacl_files/multi_env/routers.ts
```diff
@@ -417,12 +428,23 @@ export const routeIsolated = async (
...
-  const commonChangeProjection = projectElementOrValueToEnv(getChangeData(change), currentCommonElement)
+  const currentCommonValue = resolvePath(currentCommonElement, change.id)
+  if (currentCommonValue === undefined) {
+    return { primarySource: [change] }
+  }
+
+  const updatedCommonElement = currentCommonElement.clone()
+  setPath(updatedCommonElement, change.id, undefined)
+  // Note that this base change includes only the removal of change.id from common.
+  // It doesn't include any other change of the same element.
```
ori-moisis, line 444: maybe be a bit more explicit saying this base change is wrong?

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -53,8 +57,9 @@ type PositionInParent = {
   indexInParent?: number
 }
 
-const getPositionInParent = <T>(change: DetailedChange<T> & AdditionChange<T>): PositionInParent => {
+const getPositionInParent = (change: DetailedAddition): PositionInParent => {
```
ori-moisis, line 60: did the old type not work here? why did we need a new type?
(also, seems like a "Detailed(Addition/Removal/Modification)" was needed at least twice in the PR
if for some reason to old way of doing this doesn't work anymore, maybe worth putting these types somewhere common...

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -203,50 +211,57 @@ const fixEdgeIndentation = (data: string, action: ActionName, initialIndentation
   return [firstLine.trimLeft(), ...lines.slice(1)].join('\n')
 }
 
-type DetailedAddition = AdditionDiff<Element> & {
-  id: ElemID
-  path: string[]
-}
-
 export const groupAnnotationTypeChanges = (
```
ori-moisis, line 214: nit - doesn't have to be exported?

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -402,34 +417,35 @@ const wrapAdditions = (nestedAdditions: DetailedAddition[]): DetailedAddition =>
     id: wrapperObject.elemID,
     path: nestedAdditions[0].path,
     data: {
-      after: wrapperObject as Element,
+      after: wrapperObject,
     },
-  } as DetailedAddition
+    baseChange: toChange({ after: wrapperObject }),
+  }
```
ori-moisis, line 423: can be `getDetailedChanges(toChange({ after: wrapperObject })).map(change => ({ ...change, path: nestedAdditions[0].path })`?
I wonder if this pattern of create detailed changes with a path is common enough to have a utility function for it maybe? it did repeat a few times in this PR...

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -203,50 +211,57 @@ const fixEdgeIndentation = (data: string, action: ActionName, initialIndentation
...
-  const createGroupedAddAnnotationTypesChange = (annotationTypesAddChanges: DetailedChange[]): DetailedChange => {
+  const createGroupedAddAnnotationTypesChange = (annotationTypesAddChanges: DetailedAddition[]): DetailedAddition => {
     const change = annotationTypesAddChanges[0]
+    const addedAnnotations = Object.fromEntries(annotationTypesAddChanges.map(c => [c.id.name, c.data.after]))
+
+    const baseBefore = isRemovalOrModificationChange(change.baseChange)
+      ? change.baseChange.data.before.clone()
+      : undefined
+    if (baseBefore !== undefined) {
+      baseBefore.annotations = {}
```
ori-moisis, line 233: why do we need to do this?

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -402,34 +417,35 @@ const wrapAdditions = (nestedAdditions: DetailedAddition[]): DetailedAddition =>
...
-    DetailedChange[],
-  ]
+  const [nestedAdditionsWithPath, otherChanges] = _.partition(changes, isNestedAddition)
   const wrappedNestedAdditions: DetailedAddition[] = _(nestedAdditionsWithPath)
     .groupBy(addition => [addition.path, addition.id.createTopLevelParentID().parent])
     .values()
     .map(wrapAdditions)
     .value()
-  return groupAnnotationTypeChanges(_.concat(otherChanges, wrappedNestedAdditions), sourceMap)
+  return groupAnnotationTypeChanges([...otherChanges, ...wrappedNestedAdditions], sourceMap)
```
ori-moisis, line 448: we should avoid using `...` on arrays

packages/workspace/test/common/nacl_file_source.ts
```diff
@@ -134,7 +133,7 @@ export const createMockNaclFileSource = (
     load: mockFunction<NaclFilesSource['load']>().mockImplementation(
       returnChanges({
         cacheValid: true,
-        changes: currentElements.map(e => createAddChange(e, e.elemID)),
+        changes: currentElements.map(e => ({ data: { after: e as ChangeDataType }, action: 'add', id: e.elemID })),
```
ori-moisis, line 136: why not `toChange`?

packages/workspace/test/workspace/hidden_values.test.ts
```diff
@@ -301,10 +301,14 @@ describe('handleHiddenChanges', () => {
...
       const { field } = objectType.fields
-      const change: DetailedChange = {
+      const change: DetailedChangeWithBaseChange = {
         id: field.elemID,
         action: 'add',
         data: { after: field },
+        baseChange: {
+          action: 'add',
+          data: { after: field },
+        },
```
ori-moisis, line 311:
```suggestion
      const change = toDetailedChangeFromBaseChange(toChange({ after: field }))
```

same for other places in this file

ori-moisis (approved):

packages/workspace/src/workspace/hidden_values.ts
dantal4: changed it to be more clear.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
dantal4: `DetailedAddition` was already defined in this file, so I decided to modify it and use it.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
dantal4: it should be `annotationRefTypes`

ori-moisis: what do we need to do that then?

dantal4: we create there an addition change with id `adapter.type.annotation`, and the change's data is an `annotationRefTypes` object, so I wanted the `baseChange` to have no `annotationRefTypes` in the before, and to have the change's data's `annotationRefTypes` in the after.

ori-moisis: but why do you have to do that?
how can there be something not empty there?

PR #6224, opened 2024-07-17
SALTO-6277: Support base elements in applyDetailedChanges

yelly:
---

---
_Release Notes_: 
None.

---
_User Notifications_: 
None.

### DIFF
--- a/packages/adapter-api/src/elements.ts
+++ b/packages/adapter-api/src/elements.ts
@@ -120,11 +120,11 @@ export abstract class Element {
   abstract clone(annotations?: Values): Element
 
   /**
-   * Replace the content of the element with the content of other.
+   * Assign all element fields from other.
    * Needs to be overridden by each subclass as this is structure dependent.
-   * Note that the element ID is not replaced.
+   * Note that the element ID is not changed.
    */
-  replace(other: Element): void {
+  assign(other: Element): void {
     this.annotationRefTypes = other.annotationRefTypes
     this.annotations = other.annotations
     this.path = other.path
@@ -181,8 +181,8 @@ export class ListType<T extends TypeElement = TypeElement> extends Element {
     return new ListType(this.refInnerType.clone())
   }
 
-  replace(other: ListType): void {
-    super.replace(other)
+  assign(other: ListType): void {
+    super.assign(other)
     this.refInnerType = other.refInnerType
   }
 
@@ -243,8 +243,8 @@ export class MapType<T extends TypeElement = TypeElement> extends Element {
     return new MapType(this.refInnerType.clone())
   }
 
-  replace(other: MapType): void {
-    super.replace(other)
+  assign(other: MapType): void {
+    super.assign(other)
     this.refInnerType = other.refInnerType
   }
 
@@ -320,8 +320,8 @@ export class Field extends Element {
     )
   }
 
-  replace(other: Field): void {
-    super.replace(other)
+  assign(other: Field): void {
+    super.assign(other)
     this.parent = other.parent
     this.name = other.name
     this.refType = other.refType
@@ -371,8 +371,8 @@ export class PrimitiveType<Primitive extends PrimitiveTypes = PrimitiveTypes> ex
     return res
   }
 
-  replace(other: PrimitiveType<Primitive>): void {
-    super.replace(other)
+  assign(other: PrimitiveType<Primitive>): void {
+    super.assign(other)
     this.primitive = other.primitive
   }
 }
@@ -498,8 +498,8 @@ export class ObjectType extends Element {
     return res
   }
 
-  replace(other: ObjectType): void {
-    super.replace(other)
+  assign(other: ObjectType): void {
+    super.assign(other)
     this.fields = other.fields
     this.metaType = other.metaType
     this.isSettings = other.isSettings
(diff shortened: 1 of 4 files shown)


packages/adapter-utils/src/compare.ts
```diff
@@ -377,7 +410,21 @@ const filterChangesForApply = (changes: DetailedChange[]): DetailedChange[] => {
...
+  const [baseElementModifications, nestedChanges] = _.partition(
+    detailedChanges,
+    change => isModificationChange(change) && change.id.isBaseID(),
+  )
+  baseElementModifications.forEach(change => {
+    const data = getChangeData(change)
+    if (isField(data)) {
+      replaceChangeData(resolvePath(element, data.elemID), data)
+      return
+    }
```
ori-moisis, line 422: I might be missing something, but if we got the object type and the modification is on a field, the code was supposed to work before this change as well (as it would just assign the `after` field to the object, which, I think is actually the better behavior in this case)

I think the only case we need to solve here is the case where the change is on the same ID as the provided element.
so, also no need for the `partition` in the beginning, just need a single check

packages/adapter-utils/src/compare.ts
```diff
@@ -368,6 +370,37 @@ const filterChangesForApply = (changes: DetailedChange[]): DetailedChange[] => {
...
+    dst.refType = src.refType
+    dst.value = src.value
+  }
+
+  if (isField(dst) && isField(src)) {
+    dst.parent = src.parent
+    dst.name = src.name
+    dst.refType = src.refType
+  }
+}
```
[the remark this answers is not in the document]
yelly: The problem with `_.assignIn` and all similar functions is that it won't override values that are set in destination but unassigned in the source. I've added a test case that fails when replacing `replaceChangeData` with `_.assignIn`.
I haven't managed to find a different approach to doing what we need in a way that is type safe. Open to further suggestions.

packages/adapter-utils/src/compare.ts
yelly: You're right, we only need to handle top level elements and not all base elements.
As for replacing the partition with a check - are you assuming that in the relevant cases we will get exactly one change which is a modification on the whole element? Are you sure we can assume that in all flows?

ori-moisis: I think it should be safe to assume.
that is supposed to be a basic assumption of detailed compare - it should not return changes that are nested in each other (there is one exception to that when comparing list items, but that is not relevant here I think)

yelly: Done.

packages/adapter-utils/src/compare.ts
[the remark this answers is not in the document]
ori-moisis: as much as I hate to say it, I think in that case we will have to implement this in the `Element` itself (similar to how `clone` is implemented per-type).
otherwise it would be hard to maintain the correctness of this function (being completely detached from the class definition...)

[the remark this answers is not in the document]
yelly: I feared it would come to this. Done.

packages/adapter-api/src/elements.ts
```diff
@@ -122,6 +122,17 @@ export abstract class Element {
    * @return {Type} the cloned instance
    */
   abstract clone(annotations?: Values): Element
+
+  /**
+   * Replace the content of the element with the content of other.
+   * Needs to be overridden by each subclass as this is structure dependent.
+   * Note that the element ID is not replaced.
+   */
+  replace(other: Element): void {
```
ori-moisis, line 131: nit: maybe `assign`?

packages/adapter-utils/src/compare.ts
```diff
@@ -377,6 +378,14 @@ const filterChangesForApply = (changes: DetailedChange[]): DetailedChange[] => {
  * function in a single call
  */
 export const applyDetailedChanges = (element: ChangeDataType, detailedChanges: DetailedChange[]): void => {
+  if (detailedChanges.length === 1) {
+    const change = detailedChanges[0]
+    if (isModificationChange(change) && change.id.isTopLevel()) {
```
ori-moisis, line 383: I think this is the wrong check, since `ChangeDataType` could be a field (and thus, not a top level ID)
it should be
```suggestion
    if (change.id.isEqual(element.elemID)) {
```

and then I wonder about the check for `isModificationChange`
we don't really support the case where we get a removal or addition on the ID of the element, should we fail more explicitly in that case?

packages/adapter-api/src/elements.ts
yelly: Sure.

packages/adapter-utils/src/compare.ts
yelly: The field case is covered by UT and works, is there a case you think is missing?
Anyway, went with your suggestion abut kept the `isModificationChange` test just to stay on the safe side and make it clear that we don't cover additions and removals in this flow.

ori-moisis: couldn't find the UT where a field is passed as the `element`, looks like in all the new UTs the `element` is an `ObjectType` (and I assume old UTs didn't cover it since it would not have worked before this PR).

ori-moisis (approved):

packages/adapter-utils/src/compare.ts
yelly: Is that even allowed? I thought there's no flow that does this.

yelly: Done.

PR #6226, opened 2024-07-17
SALTO-5917 - Salesforce: remove coordinate fields from auto-layout flows

tomsellek:
When flows are in auto-layout mode, the coordinates of their components are meaningless, yet when they change they cause many diffs in deployments.
This PR removes these coordinate values and re-adds them as 0 for deployment (since they are required)

---
Testing: 
 - Fetch a flow in auto-layout mode. Observe the fields not being present in the NaCl files.
 - Deploy a flow with no coordinate fields. Observe a successful deployment. Observe no coordinate fields in the NaCl file post-deploy.

---
_Release Notes_: 
Salesforce: coordinate fields of flow components that belong to a flow with AUTO_LAYOUT_CANVAS mode are now removed.

---
_User Notifications_: 
Salesforce: coordinate fields of flow components that belong to a flow with AUTO_LAYOUT_CANVAS mode are now removed.

### DIFF
--- a/packages/salesforce-adapter/src/filters/flow_coordinates.ts
+++ b/packages/salesforce-adapter/src/filters/flow_coordinates.ts
@@ -17,7 +17,7 @@ import { getChangeData, InstanceElement, isAdditionOrModificationChange } from '
 import { TransformFuncSync, transformValuesSync } from '@salto-io/adapter-utils'
 import { logger } from '@salto-io/logging'
 import { LocalFilterCreator } from '../filter'
-import { apiNameSync, isInstanceOfTypeChangeSync, isInstanceOfTypeSync } from './utils'
+import { apiNameSync, ensureSafeFilterFetch, isInstanceOfTypeChangeSync, isInstanceOfTypeSync } from './utils'
 import { FLOW_METADATA_TYPE } from '../constants'
 
 const log = logger(module)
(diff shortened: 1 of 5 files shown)


packages/salesforce-adapter/src/filters/flow_coordinates.ts
```diff
@@ -0,0 +1,148 @@
...
+]
+
+const isInCanvasAutoLayoutMode = (instance: InstanceElement): boolean => {
+  if (!instance.value.processMetadataValues) {
+    return false
+  }
+  const canvasMode = instance.value.processMetadataValues.find(
+    ({ name }: { name: string }) => name === 'CanvasMode',
+  )
+  return canvasMode?.value?.stringValue === 'AUTO_LAYOUT_CANVAS'
```
yelly, line 62: Isn't all of this just:
```
instance.value?.processMetadataValues?.find(
  ({ name }: { name: string }) => name === 'CanvasMode',
)?.value?.stringValue === 'AUTO_LAYOUT_CANVAS' ?? false
```
I guess it's a matter of preference but having this single boolean expression looks better to me.

packages/salesforce-adapter/src/filters/flow_coordinates.ts
yelly, line 1: Can you add some tests for this? I think just one matching case and one NOP case should be enough.

yelly (commented):
Looking good, just need to add tests.

packages/salesforce-adapter/src/filters/flow_coordinates.ts
```diff
@@ -0,0 +1,148 @@
...
+        instance.value = transformValuesSync({
+          values: instance.value,
+          type: instance.getTypeSync(),
+          transformFunc: removeCoordinatesFromAllSections,
+        })
+      })
+  },
+  preDeploy: async (changes) => {
+    changes
+      .filter(isInstanceOfTypeChangeSync(FLOW_METADATA_TYPE))
```
amitgoldarb, line 121: This is missing an `.filter(isInCanvasAutoLayoutMode)` section.

amitgoldarb (approved):
This LGTM!
Can we make sure we test this on flows manually before merging? Also, this needs to be behind an adapter.config flag with the default off. Sadly this needs to be behind a prod ticket as it will edit a lot of flows.
This probably also needs noise reduction before we turn it on.

tomsellek:
> Can we make sure we test this on flows manually before merging?

Is the testing I described in the PR notes not enough?

packages/salesforce-adapter/src/filters/flow_coordinates.ts
tomsellek: If it's not in auto-layout mode then the coordinates will be present and we won't add them. I figured I'd just save an iteration.

tomsellek:
> ... this needs to be behind an adapter.config flag with the default off. Sadly this needs to be behind a prod ticket as it will edit a lot of flows.
> This probably also needs noise reduction before we turn it on.


I will merge this code with the feature default-disabled. I created https://salto-io.atlassian.net/browse/PROD-341 to track the production rollout.

PR #6227, opened 2024-07-17
SALTO-6220: fixes to a problem with an undefined custom properties in AppUserSchema

SeggevHaimovich:
Deploying `AppUserSchema` with an undefined custom properties field created an error, this field is probably required in Okta's API.
Fixed this by deploying an empty object when the properties field is undefined.

---

---
_Release Notes_: 
Okta Adapter:
* fixed an issue when deploying an `AppUserSchema` with an undefined custom properties

---
_User Notifications_: 
_None_


packages/okta-adapter/src/filters/schema_field_removal.ts
```diff
@@ -55,7 +56,12 @@ const deletePropertiesWithNull = (instance: InstanceElement): void => {
   const path = SCHEMAS_TO_PATH[instance.elemID.typeName]
   const customPropertiesPath = instance.elemID.createNestedID(...path)
   const customProperties = resolvePath(instance, customPropertiesPath)
-
+  if (customProperties === undefined) {
+    return
+  }
```
shir-reifenberg, line 61: I think this is redundant if you're checking for is object

packages/okta-adapter/src/filters/schema_field_removal.ts
```diff
@@ -55,7 +56,12 @@ const deletePropertiesWithNull = (instance: InstanceElement): void => {
...
+  if (!values.isPlainRecord(customProperties)) {
+    throw new Error('Custom properties should be a record')
```
shir-reifenberg, line 63: let's just `log.error` and return, I think we should avoid throwing errors during `preDeploy` 
Also, is this really an issue in all schemas types? cause I think this filter also applies for `GroupSchema` and `UserSchema`

packages/okta-adapter/src/filters/app_user_schema_deployment.ts
```diff
@@ -97,7 +99,7 @@ const getAppUserSchemaInstance = (
   fieldsToOmit: string[],
 ): InstanceElement => {
   const createdAppUserSchemaInstance = appUserSchemaInstance.clone()
-  createdAppUserSchemaInstance.value = _.omit(autoCreatedappUserSchema, [...Object.keys(fieldsToOmit), 'name'])
```
shir-reifenberg, line 100: did you remove `name` on purpose?

packages/okta-adapter/src/filters/app_user_schema_deployment.ts
```diff
@@ -125,7 +127,10 @@ const makeModificationFromAddition = async (
     appUserSchemaInstance,
     fieldsToOmit,
   )
-  return { action: 'modify', data: { before: autoCreatedAppUserSchemaInstance, after: appUserSchemaInstance.clone() } }
+  return {
```
shir-reifenberg, line 130: nice catch 🙃

packages/okta-adapter/src/filters/app_user_schema_deployment.ts
```diff
@@ -140,6 +145,10 @@ const deployModificationChange = async (
   const after = change.data.after.clone()
   const beforeBaseElemId = getBaseElemID(before)
   setPath(after, getBaseElemID(after), resolvePath(before, beforeBaseElemId))
+  const afterCustomPropertiesElemID = after.elemID.createNestedID(...CUSTOM_PROPERTIES_PATH)
+  if (resolvePath(after, after.elemID.createNestedID(...CUSTOM_PROPERTIES_PATH)) === undefined) {
```
shir-reifenberg, line 149: nit: maybe add a comment about this ?

shir-reifenberg (commented):
Looks great! and thanks for the quick fix 👑 
I see you also changes the generic filter, can you please verify if this bug also applies to `UserSchema` and `GroupSchema` ? if so, maybe we should make the fix generic as well

packages/okta-adapter/src/filters/app_user_schema_deployment.ts
SeggevHaimovich: Yes, I saw that omitting the name field is unnecessarily

packages/okta-adapter/src/filters/schema_deployment.ts
```diff
@@ -0,0 +1,155 @@
...
+}
+// TODO change the documentation
+/**
+ * When a user wants to delete a custom property from a schema, the custom property is set to null.
+ * This is in order for Okta to delete the property from the schema.
+ * This filter removes the custom properties that are set to null in the onDeploy.
+ */
+const filterCreator: FilterCreator = () => {
+  const elemIdToAfterBaseFields: Record<string, Values> = {}
+  const elemIdToAfterCustomProperties: Record<string, Values> = {}
```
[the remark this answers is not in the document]
shir-reifenberg: even if we decide to keep this map, is there a reason not having a mapping to the original instance and then restoring directly from this mapping ? (instead having 2 different mappings)

shir-reifenberg (commented):
Looks good, thank you! 👑 
Added comments mainly around `deployModificationChange` that I think can be simplified?

packages/okta-adapter/src/change_validators/app_user_schema_base_properties.ts
```diff
@@ -70,12 +83,13 @@ const getErrorFromChange = (
 }
 
 /**
- * Verifies that AppUserSchema is not modified within the base field.
+ * Verifies that schema instances are not modified within the base field.
  */
-export const appUserSchemaBaseChangesValidator: ChangeValidator = async changes =>
+// TODO: change the file name to schema_base_changes.ts
```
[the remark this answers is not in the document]
SeggevHaimovich: I wanted to keep the same file name so you could see the modifications and not as a new file.
I will change it after you approve the PR

packages/okta-adapter/src/change_validators/app_user_schema_base_properties.ts
```diff
@@ -28,7 +28,21 @@ import {
...
+  APP_USER_SCHEMA_TYPE_NAME,
+  BASE_FIELD,
+  DEFINITIONS_FIELD,
+  GROUP_SCHEMA_TYPE_NAME,
+  USER_SCHEMA_TYPE_NAME,
+} from '../constants'
+
+const CUSTOM_PROPERTIES_PATH = ['definitions', 'custom', 'properties']
+
+const SCHEMAS_TO_PATH: Record<string, string[]> = {
```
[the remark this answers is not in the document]
SeggevHaimovich: Yes I agree, it was there befre so I thought you decided to do it like this for generalisation in the future.
I will change it

packages/okta-adapter/src/filters/schema_deployment.ts
```diff
@@ -0,0 +1,155 @@
...
+import _ from 'lodash'
+import { FilterCreator } from '../filter'
+import { APP_USER_SCHEMA_TYPE_NAME, GROUP_SCHEMA_TYPE_NAME, USER_SCHEMA_TYPE_NAME } from '../constants'
+import { BASE_PATH } from '../change_validators/app_user_schema_base_properties'
+
+const log = logger(module)
+
+const CUSTOM_PROPERTIES_PATH = ['definitions', 'custom', 'properties']
+
+export const SCHEMAS_TO_PATH: Record<string, string[]> = {
```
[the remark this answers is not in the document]
SeggevHaimovich: No worries, I will change it as you suggested

packages/okta-adapter/src/filters/schema_deployment.ts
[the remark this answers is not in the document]
SeggevHaimovich: I thought about it,
If we see in the after's custom properties an empty object, we can't know if the user deleted the properties field or kept it empty:
```ts
schema: {
  definition: {
    custom: {
      properties: {}
    }
  }
}

schema: {
  definition: {
    custom: {}
  }
}
```

You are right, we can keep only one map with the original instance and get this information from it.

packages/okta-adapter/src/filters/schema_deployment.ts
```diff
@@ -0,0 +1,152 @@
...
+ * This filter is used to make schema instances deployable, it effects the custom properties and the base fields.
+ *
+ * Custom Properties
+ * When deploying a schema, the schema must contain custom properties field.
+ * In addition, to delete a custom property from a schema, the custom property is set to null.
+ *
+ * Base
+ * Okta's API does not allow to deploy changes to base attributes.
+ * we deploy the base from the before and change it back in the onDeploy.
+ */
```
shir-reifenberg, line 132: thanks, this is helpful :)

packages/okta-adapter/src/filters/schema_deployment.ts
```diff
@@ -0,0 +1,152 @@
...
+  setPath(instance, customPropertiesElemID, originalCustomProperties)
+}
+
+const returnBaseToOriginal = (instance: InstanceElement, originalInstance: InstanceElement): void => {
+  const baseElemId = getBaseElemID(instance)
+  const originalBase = resolvePath(originalInstance, baseElemId)
+  setPath(instance, baseElemId, originalBase)
+}
+
+export const makeSchemaDeployable = (
```
shir-reifenberg, line 98: nit: as this function is not really short, and doesn't give much info re what it is doing, IMO it's more readable to have it inline in the `forEach` in line 142

packages/okta-adapter/src/filters/schema_deployment.ts
```diff
@@ -0,0 +1,152 @@
...
+    setBaseField(change)
+  }
+}
+
+const returnToOriginalForm = (
+  change: AdditionChange<InstanceElement> | ModificationChange<InstanceElement>,
+  elemIdToOriginalAfter: Record<string, InstanceElement>,
+): void => {
+  const { after } = change.data
+  const originalAfter = elemIdToOriginalAfter[getChangeData(change).elemID.getFullName()]
```
shir-reifenberg, line 115: I know we just set it before but let's please validate we have an instance in hand after this line ?

packages/okta-adapter/src/filters/schema_deployment.ts
```diff
@@ -0,0 +1,152 @@
...
+  const customPropertiesElemID = instance.elemID.createNestedID(...CUSTOM_PROPERTIES_PATH)
+  const originalCustomProperties = resolvePath(originalInstance, customPropertiesElemID)
+  setPath(instance, customPropertiesElemID, originalCustomProperties)
+}
+
+const returnBaseToOriginal = (instance: InstanceElement, originalInstance: InstanceElement): void => {
+  const baseElemId = getBaseElemID(instance)
+  const originalBase = resolvePath(originalInstance, baseElemId)
+  setPath(instance, baseElemId, originalBase)
+}
```
shir-reifenberg, line 96: super nit: I'm pretty sure those can share code by getting an elemID
but it's a nit so feel free to ignore me 🙃

shir-reifenberg (approved):
Looks great! The code looks much more cleaner and readable now IMO
thank you for fixing this 😃 👑

PR #6229, opened 2024-07-17
SALTO-6284: Salesforce LWC deploy XML parsing bug-fix

tamtamirr:
Salesforce LWC deploy XML parsing bug-fix

---

There's a bug when a property has the attribute `default` with the value true. This is demonstrated in the UTs and pretty straightforward.

---
_Release Notes_: 
Salesfore Adapter:
- Fixed a bug where deploy of LightningWebComponent failed on `Attribute name "default" associated with an element type "property" must be followed by the ' = ' character.`

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/src/transformers/xml_transformer.ts
+++ b/packages/salesforce-adapter/src/transformers/xml_transformer.ts
@@ -559,7 +559,8 @@ const cloneValuesWithAttributePrefixes = async (
     if (pathID && allAttributesPaths.has(pathID.getFullName())) {
       return XML_ATTRIBUTE_PREFIX + key
     }
-    // Special handling since the XML parser implementation will omit "true" or true values
+    // Special handling since the XML parser implementation will omit the value "true" for attributes
+    // As a workaround, we pass it as part of the key
     if (pathID && trueValueAttributePaths.has(pathID.getFullName())) {
       return `${XML_ATTRIBUTE_PREFIX}${key}="true"`
     }
--- a/packages/salesforce-adapter/src/transformers/xml_transformer.ts
+++ b/packages/salesforce-adapter/src/transformers/xml_transformer.ts
@@ -533,7 +533,6 @@ const cloneValuesWithAttributePrefixes = async (
   instance: InstanceElement,
 ): Promise<Values> => {
   const allAttributesPaths = new Set<string>()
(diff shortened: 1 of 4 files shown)


packages/salesforce-adapter/test/transformers/xml_transformer.test.ts
```diff
@@ -281,9 +281,26 @@ describe('XML Transformer', () => {
...
+              {
+                name: 'testTrueProp',
+                default: 'true',
+              },
+              {
+                name: 'testFalseProp',
+                default: 'false',
+              },
+            ],
+          )
```
yelly, line 300: Maybe instead of this just add these properties to the mock? This will also help ensure we don't have this problem anywhere else.

packages/salesforce-adapter/src/transformers/xml_transformer.ts
```diff
@@ -554,9 +560,11 @@ const cloneValuesWithAttributePrefixes = async (
     if (pathID && allAttributesPaths.has(pathID.getFullName())) {
       return XML_ATTRIBUTE_PREFIX + key
     }
+    if (pathID && trueValueAttributePaths.has(pathID.getFullName())) {
+      return `${XML_ATTRIBUTE_PREFIX}${key}="true"`
```
yelly, line 564: Please add a comment here that explains why we do this, it looks super weird.

packages/salesforce-adapter/test/transformers/xml_transformer.test.ts
tamtamirr: That was my previous commit. This broke the E2E deploy that used the same values which are mock and incomplete. I wouldn't complex it. We are testing edge-case for the XML parsing here

packages/salesforce-adapter/src/transformers/xml_transformer.ts
tamtamirr: I'll move the comment from L536 here

yelly: Please also specifically reference the issue with `fast-xml-parser` that causes us to do this. IIRC we have a comment about this in the other place we had this problem.

yelly (approved):

PR #6231, opened 2024-07-18
SALTO-6288: Fix bug in deployment of MultifactorEnrollment Policy order instance

shir-reifenberg:
bug fix

---

After creating policy with its rules, we create api calls to update the rules order according to the user's request.
To update the priority field we must update the policy rule, and the bug was due to the fact we didn't send the `actions` field for this policy (which is mandatory in the request) 

---
_Release Notes_: 

_Okta_adapter_:
- Bug fix for deployment of priority of `MultifactorEnrollmentPolicyRule` 

---
_User Notifications_: 
None


idoamit8 (approved):
Thanks For the bug fix! Looks fabulous 💯

PR #6232, opened 2024-07-18
SALTO-6264: Renaming CustomObject and CustomMetadata types

yelly:
Added an underscore before so we can free these element IDs up for meta types. Also syncified some logic during debugging which is nice to have generally.

---

I tested fetching and deploying and also deploying without fetching.
The latter works fine is because all of the places we rely on the new names are either during fetch (in which case we always already have them) or during deploy where we create them artificially and don't rely on what we got in fetch.

---
_Release Notes_: 
_Salesforce_:
- Renamed the internal and hidden `CustomObject` and `CustomMetadata` types to `_CustomObject` and `_CustomMetadata`.

---
_User Notifications_: 
None.

### DIFF
--- a/packages/salesforce-adapter/e2e_test/jest_environment.ts
+++ b/packages/salesforce-adapter/e2e_test/jest_environment.ts
@@ -61,7 +61,7 @@ export const credsSpec = (envName?: string): CredsSpec<UsernamePasswordCredentia
       }
     },
     typeName: 'salesforce',
-    globalProp: envName ? `salesforce_${envName}` : 'salseforce',
+    globalProp: envName ? `salesforce_${envName}` : 'salesforce',
   }
 }
 
--- a/packages/salesforce-adapter/e2e_test/jest_environment.ts
+++ b/packages/salesforce-adapter/e2e_test/jest_environment.ts
@@ -61,7 +61,7 @@ export const credsSpec = (envName?: string): CredsSpec<UsernamePasswordCredentia
       }
     },
     typeName: 'salesforce',
-    globalProp: envName ? `salesforce_${envName}` : 'salesforce',
+    globalProp: envName ? `salesforce_${envName}` : 'salseforce',
   }
 }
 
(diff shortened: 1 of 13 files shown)


packages/salesforce-adapter/src/transformers/transformer.ts
```diff
@@ -1658,9 +1663,10 @@ type CreateMetadataObjectTypeParams = Omit<ObjectTypeCtorParam, 'elemID'> & {
 }
 export const createMetadataObjectType = (
   params: CreateMetadataObjectTypeParams,
+  name?: string,
 ): MetadataObjectType =>
   new ObjectType({
-    elemID: new ElemID(SALESFORCE, params.annotations.metadataType),
+    elemID: new ElemID(SALESFORCE, name ?? params.annotations.metadataType),
```
tamtamirr, line 1669: I see this change is only relevant for UTs and has no use-case in our code. Not a fan of changing code for the purpose of UTs. Also, doesn't this implementation makes more sense?

```suggestion
    elemID: new ElemID(SALESFORCE, METADATA_TYPES_TO_RENAME.get(params.annotations.metadataType) ?? params.annotations.metadataType),
```

tamtamirr (commented):
Added some comments. Dangerous flows we must validate:
1. Include / Exclude of `CustomObject` and/or `CustomMetadata`
2. Targeted Fetch of `CustomObject` and/or `CustomMetadata`.

packages/salesforce-adapter/e2e_test/jest_environment.ts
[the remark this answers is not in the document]
yelly: Yeah, originally I touched this file but then didn't need to but this stuck around. Reverting.

packages/salesforce-adapter/src/transformers/transformer.ts
yelly: I just explicitly wrote it out in tests and reverted this change.

tamtamirr (approved):
LGTM

PR #6233, opened 2024-07-18
SLATO-6295: Simplify Salesforce Client listMetadataObjects caching

tamtamirr:
Simplify Salesforce Client listMetadataObjects caching

---

Required step to tackle throttling issues in Change Based Fetch where we execute the same listMetadataObjects request three times on Profiles related metadata types. 

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


yelly (approved):
Nice

PR #6234, opened 2024-07-18
SALTO-6293: [MS Entra] Convert http error messages to include the message from the response

shoshana-houri:
The error message returned in the response from Microsoft Graph when a request fails are informative (AFAICT ATM) and helpful in understanding the cause of the problem. We want the users to be able to read them as well and workaround small issues if such exist.

---
_Release Notes_: 
Microsoft Entra:
* Expand user-facing errors to include more informative messages when requests fail.

---
_User Notifications_:


Uri-Bar (approved):
esh

PR #6235, opened 2024-07-19
README.md: Update docker build command instruction

thilgen:
Add missing input path specifier to docker build command instruction


ori-moisis (approved):
Thanks for this fix

PR #6239, opened 2024-07-21
SALTO-6300: Avoid sending multiple listMetadataObjects requests on the same type in Salesforce fetch

tamtamirr:
Avoid sending multiple listMetadataObjects requests on the same type in Salesforce fetch

---

In Salesforce Changed Based Fetch, we have three promises that run concurrently and perform list on the same types:
1. Fetch metadata instances of all the types
2. Fetch of changed profiles with all the related types
3. Fetch of partial changed parts in all Profiles 

Since this code runs concurrently, the Metadata types that are related to Profiles (e.g. ApexClass, Layout) will be listed multiple times in Change Based Fetch for Environments with profiles.

---
_Release Notes_: 
Salesforce Adapter:
- Improved caching of listMetadataObjects: Same type won't be listed twice even when the requests are created concurrently.

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/test/client.test.ts
+++ b/packages/salesforce-adapter/test/client.test.ts
@@ -92,7 +92,7 @@ describe('salesforce client', () => {
           total: RATE_LIMIT_UNLIMITED_MAX_CONCURRENT_REQUESTS,
           retrieve: 3,
           read: RATE_LIMIT_UNLIMITED_MAX_CONCURRENT_REQUESTS,
-          list: 1,
+          list: undefined,
         },
       },
     })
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/test/client.test.ts
```diff
@@ -92,7 +92,7 @@ describe('salesforce client', () => {
           total: RATE_LIMIT_UNLIMITED_MAX_CONCURRENT_REQUESTS,
           retrieve: 3,
           read: RATE_LIMIT_UNLIMITED_MAX_CONCURRENT_REQUESTS,
-          list: 1,
+          list: undefined,
```
tamtamirr, line 95: Changed to undefined so the UT will use the default list bucket max concurrent request limit. Note: When the limit is 1, the issue does not reproduce

tamtamirr (commented):

yelly (approved):

PR #6240, opened 2024-07-21
SALTO-6255: Adding versioning for changed at singleton

yelly:
Now when we make major changes to workspaces (usually around PROD tickets) we can bump the version to force the first fetch to be a full fetch and to force users to fetch before deploying.

---

---
_Release Notes_: 
_Salesforce_:
- The first fetch after a major change to the adapter will be a full change even when change detections is requested.
- Deployments after a major change to the adapter will require a fetch.

---
_User Notifications_: 
None.

### DIFF
--- a/packages/salesforce-adapter/src/adapter.ts
+++ b/packages/salesforce-adapter/src/adapter.ts
@@ -138,7 +138,7 @@ import {
   addDefaults,
   apiNameSync,
   buildDataRecordsSoqlQueries,
-  changedAtOutdated,
+  isChangedAtSingletonOutdated,
   getFLSProfiles,
   getFullName,
   instanceInternalId,
@@ -573,20 +573,18 @@ export default class SalesforceAdapter implements AdapterOperations {
    * Account credentials were given in the constructor.
    */
   @logDuration('fetching account configuration')
-  async fetch({
-    progressReporter,
-    withChangesDetection: changesDetectionRequested = false,
-  }: FetchOptions): Promise<FetchResult> {
+  async fetch(fetchInput: FetchOptions): Promise<FetchResult> {
     const fetchParams = this.userConfig.fetch ?? {}
+    const { progressReporter } = fetchInput
     const baseQuery = buildMetadataQuery({ fetchParams })
     const lastChangeDateOfTypesWithNestedInstances =
       await getLastChangeDateOfTypesWithNestedInstances({
         client: this.client,
         metadataQuery: buildFilePropsMetadataQuery(baseQuery),
       })
-    const withChangesDetection = (await changedAtOutdated(this.elementsSource))
-      ? false
-      : changesDetectionRequested
+    const withChangesDetection =
+      fetchInput.withChangesDetection &&
+      !(await isChangedAtSingletonOutdated(this.elementsSource))
     const metadataQuery = withChangesDetection
       ? await buildMetadataQueryForFetchWithChangesDetection({
           fetchParams,
--- a/packages/salesforce-adapter/src/adapter.ts
+++ b/packages/salesforce-adapter/src/adapter.ts
@@ -138,6 +138,7 @@ import {
   addDefaults,
   apiNameSync,
   buildDataRecordsSoqlQueries,
+  changedAtOutdated,
   getFLSProfiles,
   getFullName,
   instanceInternalId,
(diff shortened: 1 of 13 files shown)


packages/salesforce-adapter/src/adapter.ts
```diff
@@ -583,6 +584,9 @@ export default class SalesforceAdapter implements AdapterOperations {
         client: this.client,
         metadataQuery: buildFilePropsMetadataQuery(baseQuery),
       })
+    const withChangesDetection = (await changedAtOutdated(this.elementsSource))
+      ? false
+      : changesDetectionRequested
```
tamtamirr, line 589:
```suggestion
    const withChangesDetection = changesDetectionRequested && await changedAtOutdated(this.elementsSource)
```
That way we avoid invoking `changedAtOutdated` on full fetch.

packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -876,6 +878,16 @@ export const getChangedAtSingletonInstance = async (
   return isInstanceElement(element) ? element : undefined
 }
 
+export const changedAtOutdated = async (
```
tamtamirr, line 881:
```suggestion
export const isChangedAtSingletonOutdated = async (
```

packages/salesforce-adapter/test/filters/utils.test.ts
```diff
@@ -513,6 +516,46 @@ describe('filter utils', () => {
       })
     })
   })
+  describe('changedAtOutdated', () => {
```
tamtamirr, line 519: We should also add a UT for the case where the `version` field does not exist in the singleton. Similar to your UT when the ChangedAtSingleton is undefined, but a different use-case. (Which will also break and expect to return true once we introduce version 1)

packages/salesforce-adapter/test/change_validators/changed_at_version.test.ts
```diff
@@ -0,0 +1,80 @@
...
+  const mockElementsSource = (
+    changedAtVersion: number,
+  ): ReadOnlyElementsSource => {
+    const changedAtSingleton = createInstanceElement(
+      {
+        fullName: 'whatever',
+        [CHANGED_AT_SINGLETON_VERSION_FIELD]: changedAtVersion,
+      },
+      ArtificialTypes[CHANGED_AT_SINGLETON],
+    )
```
tamtamirr, line 39: Why not use the mockInstance of ChangedAtSingleton here? Also not a fan of passing the `fullName` on a non Metadata instance, especially if we call it `whatever `.

tamtamirr (commented):
LGTM! Added a few comments.
Also we probably could implement the CV in a different PR, but IDM :)

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -574,7 +575,7 @@ export default class SalesforceAdapter implements AdapterOperations {
   @logDuration('fetching account configuration')
   async fetch({
     progressReporter,
-    withChangesDetection = false,
+    withChangesDetection: changesDetectionRequested = false,
```
tamtamirr, line 578: Not a fan of this name.
Possible alternatives I would consider:
Receive the args as `fetchInput`.
Within your code:
```
const { progressReporter } = fetchInput
```
Reference the input as `fetchInput.withChangesDetection`

packages/salesforce-adapter/src/adapter.ts
yelly: Done (just had to add a !).

packages/salesforce-adapter/src/adapter.ts
yelly: Better, done.

packages/salesforce-adapter/src/filters/utils.ts
yelly: Done.

packages/salesforce-adapter/test/change_validators/changed_at_version.test.ts
yelly: Good point, just copied from `account_settings.test.ts` but your suggestion is better.

packages/salesforce-adapter/test/filters/utils.test.ts
yelly: Good catch, done.

tamtamirr (approved):

PR #6241, opened 2024-07-21
SALTO-5692: Okta: Upgrade `BrandTheme` deploy definitions to new infra

Nurdok:
See title.

---

The migrated definitions still use the existing filters (for addition and removal). I've updated the filters to use the new infra.
We can potentially add an infra capability to allow specific HTTP status codes (404) to replace the removal filter, but that's out of scope for this PR.

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/okta-adapter/src/filters/brand_theme_addition.ts
```diff
@@ -41,50 +51,56 @@ const getThemeIdByBrand = async (
...
 
 /**
- * Deploy addition changes of BrandTheme instances, by finding the ID of the existing theme and updating it.
+ * Find and set the ID for an existing theme, to allow the deploy to succeed.
+ *
+ * The actual deploy of the BrandTheme is done by the standard deploy mechanism,
+ * this filter only adds the ID.
  */
-const filterCreator: FilterCreator = ({ definitions, oldApiDefinitions }) => ({
+const filterCreator: FilterCreator = ({ definitions }) => ({
```
shir-reifenberg, line 89: Can't we move this logic to the infra as well? 
I think we can add the additional request to `/api/v1/brands/${brandId}/themes` and then store the results in the `sharedContext` and use it, no ?

packages/okta-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -64,6 +65,62 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
...
+          ],
+          remove: [
+            {
+              request: {
+                endpoint: {
+                  // BrandThemes are removed automatically by Okta when the Brand is removed.
+                  // We use an empty URL here to mark this action as supported in case a user removed the theme
+                  // alongside its Brand.
+                  // A separate Change Validator ensures that mappings aren't removed by themselves.
+                  path: '/',
```
shir-reifenberg, line 115: why not actually making the `get` request, and then use `additionalValidStatuses` with 404 to mark this as success ?
(then we can delete `brand_theme_removal` filter as well ?)

shir-reifenberg (commented):
Looks good, added some questions inside 🙃 
Also please verify this works as well with brand theme change together with modification on the static files (which handled by `brandThemeFilesFilter`) cause this flow is not covered in E2E

packages/okta-adapter/src/definitions/deploy/deploy.ts
Nurdok: `additionalValidStatuses` only _adds_ valid statuses, but in this case a 200 status would be an error. This requires a new infra capability (see SALTO-6310).

shir-reifenberg: got it :)
so in this case, do we need the `path: `/`, or can we set `typesWithNoDeploy` in `typesWithNoDeploy` ?

packages/okta-adapter/src/filters/brand_theme_addition.ts
Nurdok: Done. It's even nicer than that, because the infra automatically copies ID to the instance.

packages/okta-adapter/src/definitions/deploy/deploy.ts
Nurdok: Bah, empty remove didn't really work. I changed this to `earlySuccess` for the same reason - I'm a little wary to hardcode this type into `typesWithNoDeploy` because it's an easy thing to miss when updating. Also, it will prevent validation for other actions as well.

packages/okta-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -64,6 +65,70 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
...
+            },
+          },
+        },
+        customizations: {
+          add: [
+            {
+              // BrandThemes are created automatically by Okta when a Brand is
+              // created, but we're allowing to "add" a theme alongside its
+              // Brand. We first send a GET query to get its ID, and then we
+              // perform a "modify" action.
```
shir-reifenberg, line 83: super nit - was this formatted this way? otherwise my OCD instincts will appreciate this 😅 

```suggestion
              // BrandThemes are created automatically by Okta when a Brand is created,
              // but we're allowing to "add" a theme alongside its Brand.
              // We first send a GET query to get its ID, and then we perform a "modify" action.
```

packages/okta-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -64,6 +65,70 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
...
+            },
+          ],
+          remove: [
+            // BrandThemes are removed automatically by Okta when the Brand is removed.
+            // We use an empty request list here to mark this action as supported in case a user removed the theme
+            // alongside its Brand.
+            // A separate Change Validator ensures that themes aren't removed by themselves.
+            {
+              request: {
+                earlySuccess: true,
```
shir-reifenberg, line 118: much cleaner :)

packages/okta-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -64,6 +65,70 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
...
+              },
+            },
+          ],
+        },
+      },
+      toActionNames: ({ change }) => (isAdditionChange(change) ? ['add', 'modify'] : [change.action]),
```
shir-reifenberg, line 124: big like!

packages/okta-adapter/src/definitions/requests/clients.ts
```diff
@@ -87,6 +87,11 @@ export const createClientDefinitions = (
               omitBody: true,
             },
           },
+          '/api/v1/brands/{brandId}/themes': {
+            get: {
+              omitBody: true,
+            },
```
shir-reifenberg, line 93: isn't it the default for `get` requests ? or is it different in deploy?

shir-reifenberg (approved):
🚀 🚀 🚀

packages/okta-adapter/src/definitions/deploy/deploy.ts
Nurdok: I'm used to wrap comments at 80 characters :/

packages/okta-adapter/src/definitions/requests/clients.ts
Nurdok: Removed. I recall it was needed for some reason, but maybe it was a temporary error in the config, dunno.

PR #6246, opened 2024-07-21
SALTO-5323 export getIndependentChanges

dantal4:

---

Rename `filterChangesForApply` to `getIndependentChanges` and export it.

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/adapter-utils/src/compare.ts
```diff
@@ -358,7 +358,7 @@ const isOrderChange = (change: DetailedChange): boolean =>
  * in that item that was changed. In that case we would want to ignore the inner change
  * since the item change already contains it.
  */
-const filterChangesForApply = (changes: DetailedChange[]): DetailedChange[] => {
+export const getIndependentChanges = (changes: DetailedChange[]): DetailedChange[] => {
   // For performance, avoiding the sort if no need to filter
```
yelly, line 362: What sort are we avoiding?

yelly (approved):
Just a tiny comment

packages/adapter-utils/src/compare.ts
dantal4: I *think* that it's about the sort of elemIds in getIndependentElemIDs.

PR #6248, opened 2024-07-21
SALTO-6304 Cache queryWithDefault queries

netama:
The def query is used in many places, and runs some computations in order to return `.query(typeName)` results. specifically, it makes many such requests as part of the `extractStandaloneInstances` logic which runs recursively on instances - and has caused some slowdown in zendesk fetches post upgrade.

---

adding a cache as a first step for optimizing that flow - which seems to reduce it significantly on large environments.
also adding some additional logs - if needed, we can optimize this flow further, e.g. by "pruning" paths that cannot reach standalone instances.

---
_Release Notes_: 
None

---
_User Notifications_: 
None


edenhassid (approved):
🔥

PR #6249, opened 2024-07-21
SALTO-6305 allow adjust to return multiple items

netama:
Allow the `adjust` function to return multiple items or values - needed for some more complex manipulations of returned responses

---

the "extended" `AdjustFunction` function can still be used, but since sometimes it helps to be able to rely on the returned type for further manipulation, converted most of the existing functions to `Single` (which was the only supported variant until this PR)

---
_Release Notes_: 
None

---
_User Notifications_: 
None


shir-reifenberg (approved):
👯

PR #6250, opened 2024-07-22
SALTO-6037 Extract string lexer, consolidate usage with template static file

adfineberg:
This PR wants to have a single string parser both for the regular flow and for template static files. 

Regarding the `values.ts` changes - I had to change the way the string value was created, as the parser now works for both single line and multi line strings. This means that multi line strings have escapes as tokens. So unescaping multiline strings doesn't work for escaped tokens, as they split the `\'''` into `\'` and `''`.

---

---
_Release Notes_: 
__Core:__
* Fixed a bug parsing template static files with string interpolations.

---
_User Notifications_: 

### DIFF
--- a/packages/parser/src/utils/template_static_file.ts
+++ b/packages/parser/src/utils/template_static_file.ts
@@ -13,60 +13,36 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { isReferenceExpression, StaticFile, TemplateExpression, TemplatePart } from '@salto-io/adapter-api'
+import { isReferenceExpression, StaticFile, TemplateExpression } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
+import type { Token } from 'moo'
 import { createTemplateExpression } from '@salto-io/adapter-utils'
 import { escapeTemplateMarker } from '../parser/internal/utils'
-import { createReferenceExpression, unescapeTemplateMarker } from '../parser/internal/native/helpers'
-import { IllegalReference } from '../parser'
-import { REFERENCE_PART } from '../parser/internal/native/lexer'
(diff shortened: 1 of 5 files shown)


packages/parser/src/utils/template_static_file.ts
```diff
@@ -13,60 +13,36 @@
...
+  buffer: Buffer,
+): { templateExpression: TemplateExpression; errors: string[] } => {
+  const tokens = stringLexerFromString(buffer.toString())
+  const context = { errors: [] as ParseError[], filename: '' }
+  const value = createStringValue(context, tokens, createSimpleStringValue)
+  const errors = context.errors.map(err => err.message)
+  if (typeof value === 'string') {
+    log.trace('Template static file is a string. Creating a TemplateExpression from it')
+    return {
+      templateExpression: createTemplateExpression({ parts: [value] }),
```
edenhassid, line 41: why do you create a template expression if the value is a string?

adfineberg: It shouldn't happen in a legitimate flow - as the function that calls this checks for the `isTemplate` flag, which is supposed to mean that it has references.
The parser function in general can return a string, so I handled it mostly for type safety and bugs

packages/parser/test/parser/dump.test.ts
edenhassid, line 1: why do you test the dump and not the parse?

edenhassid (approved):
looks good! added some small comments

packages/parser/test/parser/dump.test.ts
adfineberg: I didn't add any functionality to the parsing (just refactored there), but there was a missing test in the dump

PR #6254, opened 2024-07-22
SALTO-3676: remove useChangesDetection from NS config

Shulik95:
_We fully support `withChangesDetection` in the SAAS, there is no need for a config flag_

---

None

---
_Release Notes_: 
Netsuite Adapter:
* _useChangesDetection in the NS is not longer supported. Use changed-based fetch._

---
_User Notifications_: 


dantal4 (approved):
💯

PR #6266, opened 2024-07-23
SALTO-6276: add dumpElementsToFolder to adapter api and implement it for salesforce SFDX

ori-moisis:
---

There is still no flow which uses this new function (hence no release notes)

Some of the planned follow ups for this PR:
1. Change the way the client is passed into the remote filters so it will be type safe
2. Add a flow through core and a cli command that will actually use this capability

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


.vscode/settings.json
```diff
@@ -10,4 +10,5 @@
   "typescript.tsdk": "node_modules/typescript/lib",
   "typescript.preferences.importModuleSpecifier": "relative",
   "eslint.useFlatConfig": true,
+  "editor.tabSize": 2,
```
ori-moisis, line 13: unrelated to this PR, but didn't feel like it deserved its own PR.
let me know if you disagree and I'll split it

.vscode/tasks.json
```diff
@@ -4,11 +4,15 @@
...
-      ]
+      "type": "npm",
+      "script": "build-ts",
+      "group": {
+        "kind": "build",
+        "isDefault": true
+      },
+      "problemMatcher": [],
+      "label": "npm: build-ts",
+      "detail": "./build_utils/turbo_run.sh build-ts"
```
ori-moisis, line 15: unrelated to this PR, but didn't feel like it deserved its own PR.
let me know if you disagree and I'll split it

the task that was here previously didn't really work or do anything, replaced it with a default build task

packages/cli/webpack.config.js
```diff
@@ -60,6 +60,8 @@ module.exports = {
       SALTO_TELEMETRY_TOKEN: 'dev',
       SALTO_TELEMETRY_DISABLE: '0',
       SALTO_TELEMETRY_URL: 'https://telemetry.salto.io',
+      // DISABLE SFDX logging because their implementation uses relative paths and thus cannot work when packed
+      SF_DISABLE_LOG_FILE: 'true',
```
ori-moisis, line 64: The issue stems from this part of their code - https://github.com/forcedotcom/sfdx-core/blob/ed0b8c2ec521ab8a9dbc2210b8bdd0f42d88792e/src/logger/logger.ts#L183

packages/salesforce-adapter/src/sfdx_parser/sfdx_dump.ts
```diff
@@ -0,0 +1,166 @@
...
+      if (lastPath !== undefined && isSubDirectory(pathToDelete, lastPath)) {
+        return undefined
+      }
+      lastPath = pathToDelete
+      return pathToDelete
+    })
+    .filter(values.isDefined)
+}
+
+type DumpElementsToFolderFunc = Required<Adapter>['dumpElementsToFolder']
```
yelly, line 89: Nit: I think that this is more clear:
```suggestion
type DumpElementsToFolderFunc = NonNullable<Adapter['dumpElementsToFolder']>
```
(mainly because all of `Adapter` doesn't need to be required, only this type)

packages/salesforce-adapter/src/sfdx_parser/sfdx_dump.ts
```diff
@@ -0,0 +1,166 @@
...
+export const dumpElementsToFolder: DumpElementsToFolderFunc = async ({ baseDir, changes, elementsSource }) => {
+  const [customObjectInstanceChanges, metadataChanges] = _.partition(changes, isInstanceOfCustomObjectChangeSync)
```
yelly, line 91: We can also have other instances here, no?
For example: settings, custom metadata instances.

packages/salesforce-adapter/src/sfdx_parser/sfdx_dump.ts
```diff
@@ -0,0 +1,166 @@
...
+
+  const resolvedChanges = await awu(metadataChanges)
+    .map(change => resolveChangeElement(change, getLookUpName))
+    .toArray()
+
+  const fetchProfile = buildFetchProfile({ fetchParams: {} })
```
yelly, line 97: Don't we want to support the same optional features etc. as the normal flow?

packages/salesforce-adapter/src/sfdx_parser/sfdx_dump.ts
```diff
@@ -0,0 +1,166 @@
...
+    validChanges.map(change =>
+      addChangeToPackage(pkg, change, NESTED_METADATA_TYPES, {
+        forceAddToManifest: true,
+      }),
+    ),
+  )
+
+  // Load the components we wish to merge with the current project
+  const buffer = await pkg.getZip()
+  const zipTree = await ZipTreeContainer.create(buffer)
```
yelly, line 127: Since `buffer` is an unhelpful name I would:
```suggestion
  const zipTree = await ZipTreeContainer.create(await pkg.getZip())
```

packages/salesforce-adapter/test/sfdx_parser/sfdx_dump.test.ts
```diff
@@ -0,0 +1,274 @@
...
+          elementsSource: buildElementsSourceFromElements([object]),
+        })
+      })
+      it('should apply all changes and have no errors', () => {
+        expect(dumpResult.unappliedChanges).toHaveLength(0)
+        expect(dumpResult.errors).toHaveLength(0)
+      })
+      it('should create an XML for the new field', async () => {
+        await expect(
+          exists(path.join(project.name(), 'force-app/main/default/objects/Test__c/fields/New__c.field-meta.xml')),
```
yelly, line 216: Do we want to check the content?

packages/salesforce-adapter/test/sfdx_parser/sfdx_dump.test.ts
```diff
@@ -0,0 +1,274 @@
...
+    <fullName>All</fullName>
+    <filterScope>Everything</filterScope>
+    <label>Updated label</label>
+</ListView>
+`)
+      })
+    })
+  })
+
+  describe('with fields', () => {
```
yelly, line 195: Do we want to test field modify?

packages/salesforce-adapter/src/transformers/xml_transformer.ts
```diff
@@ -594,22 +595,29 @@ export const createDeployPackage = (deleteBeforeUpdate?: boolean): DeployPackage
...
 
+      const zip = new JSZip()
       // Set a constant date for all files in the zip in order to keep the zip hash constant when
       // the contents are the same.
       // this is important for the "quickDeploy" feature
       const date = new Date('2023-06-15T00:00:00.000+01:00')
-      Object.values(zip.files).forEach(info => {
-        info.date = date
-      })
+      Array.from(zipContent.entries()).forEach(([fileName, content]) => zip.file(fileName, content, { date }))
```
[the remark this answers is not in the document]
ori-moisis: map entries is an iterable, so this needs to be either `Array.from(...).forEach` or `wu(...).forEach`.

packages/adapter-api/src/adapter.ts
```diff
@@ -190,6 +190,17 @@ export type LoadElementsFromFolderArgs = {
...
 
+export type DumpElementsToFolderArgs = {
+  baseDir: string
+  changes: Change[]
+  elementsSource: ReadOnlyElementsSource
+}
+
+export type DumpElementsResult = {
+  unappliedChanges: Change[]
+  errors: ReadonlyArray<SaltoError | SaltoElementError>
```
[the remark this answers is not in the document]
ori-moisis: So, I copied this from the type used in `DeployResult`
There is a difference between the two things:
when the return type is defined as `SaltoError`, you cannot have a return literal which returns an elemID (which only exists on `SaltoElementError`), if you try to do that, you'd get a `Object literal may only specify known properties` error from typescript.
e.g, if I have this code in a function that is supposed to return a `DumpElementResult`:
```typescript
return {
      unappliedChanges: [],
      errors: [
        {
          elemID: myElemID,
          severity: 'Error',
          message: myErrorMessage,
        },
      ],
    }
```
I'd get:
```
Type '{ elemID: ElemID; severity: "Error"; message: any; }' is not assignable to type 'SaltoError'.
  Object literal may only specify known properties, and 'elemID' does not exist in type 'SaltoError'
```

packages/salesforce-adapter/src/sfdx_parser/sfdx_dump.ts
ori-moisis: all the examples you gave are metadata changes
this is essentially the same logic that determines the deploy code path when we deploy things (see logic [here](https://github.com/salto-io/salto/blob/92074712b14835ae1181943b59700055235d4f85/packages/salesforce-adapter/src/adapter.ts#L652))

packages/salesforce-adapter/test/sfdx_parser/sfdx_dump.test.ts
ori-moisis: we could
I didn't want to add another exact string comparison here because then if we make some formatting change it would be annoying to fix all the broken tests (I did want to validate the format at least in one test though because it is part of the expectation)

I can add a test which actually parses the XML and compares the values...

packages/salesforce-adapter/test/sfdx_parser/sfdx_dump.test.ts
ori-moisis: at this point it felt like it would be testing the SFDX code more than testing our code.
I added the non-trivial scenarios here mostly to make it go though code paths that make a difference in our code (e.g - deletion of fields which is implemented in our code, non-decomposed changes which makes it go through `readFileSync` that is implemented in our code).

having said that, I am not opposed to adding more cases, let me know which cases you think should be added here and I'll add them

packages/adapter-api/src/adapter.ts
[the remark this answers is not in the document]
yelly: Makes sense

packages/salesforce-adapter/test/sfdx_parser/sfdx_dump.test.ts
yelly: Nah I trust your judgement on this, just wanted to make sure you didn't overlook it by accident.

packages/salesforce-adapter/src/transformers/xml_transformer.ts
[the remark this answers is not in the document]
yelly: Personally I think `wu(...).forEach` is more clear since you don't care about the resulting array, but up to you.

yelly (approved):

PR #6268, opened 2024-07-23
SALTO-6318 support deploy scripts list when deploying policy

omrifarkash:
After research, modification and adding of policy API is supporting scripts but it expect to get it in a specific format inside the xml.

`<scripts><size>2</size><script> ...scriptData </script> ...scriptData  <script></script></scripts>`
Also, In realized that the data return on fetch of policy about the script is not duplication of the data of the script, but information of how to use the script in this policy.

To support this:
- Change how we reference a script in a policy instance. In the new way we are not losing information about how to use the script in this policy (we need this information for deploying late)
- Add adjust function to restructure scripts array before deploy so converting to xml will yield the wanted format (example above)

---

---

_Release Notes_: 
_Jamf Adapter_:
- Support deployment of scripts list on policy instance

---

_User Notifications_: 
_Jamf Adapter_:
- Support deployment of scripts list on policy instance


packages/jamf-adapter/src/definitions/deploy/policy.ts
```diff
@@ -0,0 +1,40 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { definitions } from '@salto-io/adapter-components'
+import { values } from '@salto-io/lowerdash'
+
+/*
+ * change scripts array to object with size and script fields, so when converting to xml it will be in the correct format
+ */
+export const adjustScriptStructureBeforeDeploy = (value: Record<string, unknown>): void => {
```
kami-delaroz, line 22: non export

kami-delaroz (approved):

PR #6269, opened 2024-07-23
SALT0-5323 add filterFunc to applyDetailedChanges

dantal4:
Add a `filterFunc` to `applyDetailedChanges`, that tells the function which changes should be applied. The reason for this `filterFunc`, instead of passing in advance only the list of changes that you want to apply, is that in list item changes, the indexes should be fixed when you apply only part of the changes - if you want to ignore an addition/removal, the indexes of the following changes should be fixed, so the right order is kept.
Because of the complexity in applying only part of the reorder changes in a list, reorder changes will be applied only when all of the changes in that list should be applied.

---

---
_Release Notes_: 
Core:
- add `filterFunc` to `applyDetailedChanges`

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-api/src/utils.ts
+++ b/packages/adapter-api/src/utils.ts
@@ -32,7 +32,7 @@ type SubElementSearchResult = {
   path: ReadonlyArray<string>
 }
 
-export const isIndexPathPart = (key: string): boolean => !Number.isNaN(Number(key))
+export const isIndexPathPart = (key: string): boolean => key !== '' && !Number.isNaN(Number(key))
 
 export const getDeepInnerType = async (
   type: TypeElement,
--- a/packages/adapter-api/src/utils.ts
+++ b/packages/adapter-api/src/utils.ts
@@ -32,7 +32,7 @@ type SubElementSearchResult = {
   path: ReadonlyArray<string>
 }
 
-export const isIndexPathPart = (key: string): boolean => key !== '' && !Number.isNaN(Number(key))
+export const isIndexPathPart = (key: string): boolean => /^\d+$/.test(key)
 
 export const getDeepInnerType = async (
   type: TypeElement,
--- a/packages/adapter-utils/src/compare.ts
+++ b/packages/adapter-utils/src/compare.ts
@@ -40,7 +40,7 @@ import {
 } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
 import { getIndependentElemIDs, resolvePath, setPath } from './utils'
-import { applyListChanges, getArrayIndexMapping, isOrderChange } from './list_comparison'
+import { applyListChanges, getArrayIndexMapping, getChangeRealId, isOrderChange } from './list_comparison'
 
 const log = logger(module)
 
@@ -390,21 +390,21 @@ export const applyDetailedChanges = (
   )
 
   const [realListItemGroup, otherGroups] = _.partition(Object.values(potentialListItemGroups), group =>
-    Array.isArray(resolvePath(element, group[0].id.createParentID())),
+    Array.isArray(resolvePath(element, getChangeRealId(group[0]).createParentID())),
   )
 
   _(otherChanges)
     .concat(otherGroups.flat())
     .filter(filterFunc)
     .forEach(detailedChange => {
-      const id = isRemovalChange(detailedChange)
-        ? detailedChange.elemIDs?.before ?? detailedChange.id
-        : detailedChange.elemIDs?.after ?? detailedChange.id
+      const id = getChangeRealId(detailedChange)
       const data = isRemovalChange(detailedChange) ? undefined : detailedChange.data.after
       setPath(element, id.replaceParentId(element.elemID), data)
     })
 
-  realListItemGroup.forEach(changes => {
+  // we want inner lists to be applied before outer lists, because the indexes may change
+  const orderedListItemGroups = _.orderBy(realListItemGroup, group => group[0].id.nestingLevel, 'desc')
+  orderedListItemGroups.forEach(changes => {
     applyListChanges(element, changes, filterFunc)
   })
 }
--- a/packages/adapter-utils/src/compare.ts
+++ b/packages/adapter-utils/src/compare.ts
@@ -16,7 +16,6 @@
 import _ from 'lodash'
 import {
   Change,
-  ChangeDataType,
   CompareOptions,
   DetailedChange,
   DetailedChangeWithBaseChange,
@@ -41,7 +40,7 @@ import {
 } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
 import { getIndependentElemIDs, resolvePath, setPath } from './utils'
-import { applyListChanges, getArrayIndexMapping } from './list_comparison'
+import { applyListChanges, getArrayIndexMapping, isOrderChange } from './list_comparison'
 
 const log = logger(module)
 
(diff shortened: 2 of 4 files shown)


packages/adapter-api/src/utils.ts
```diff
@@ -32,7 +32,7 @@ type SubElementSearchResult = {
   path: ReadonlyArray<string>
 }
 
-export const isIndexPathPart = (key: string): boolean => !Number.isNaN(Number(key))
+export const isIndexPathPart = (key: string): boolean => key !== '' && !Number.isNaN(Number(key))
```
yelly, line 35: While we're here, this isn't a particularly good test.
Examples for values that pass this which shouldn't: `0x10`, `12.55`, `00000012`, `-12`.
Instead I think we should go with a regex. Can be as simple as `\d+` or we can be fancier with `0|[1-9]\d*`.

packages/adapter-utils/src/compare.ts
```diff
@@ -213,7 +212,7 @@ const getAnnotationTypeChanges = ({
   beforeId: ElemID
   afterId: ElemID
 }): DetailedChange[] => {
-  const hasAnnotationTypes = (elem: ChangeDataType): elem is ObjectType | PrimitiveType =>
+  const hasAnnotationTypes = (elem: Element): elem is ObjectType | PrimitiveType =>
```
yelly, line 215: Why did we move from the specific `ChangeDataType` to the general `Element`? Did this function changes its purpose?

packages/adapter-utils/src/compare.ts
```diff
@@ -396,21 +390,22 @@ export const applyDetailedChanges = (element: ChangeDataType, detailedChanges: D
...
       const data = isRemovalChange(detailedChange) ? undefined : detailedChange.data.after
       setPath(element, id.replaceParentId(element.elemID), data)
     })
 
-  realListItemGroup.forEach(changes => {
-    applyListChanges(element, changes)
+  // we want inner lists to be applied before outer lists, because the indexes may change
+  const orderedListItemGroups = _.orderBy(realListItemGroup, group => group[0].id.nestingLevel, 'desc')
+  orderedListItemGroups.forEach(changes => {
+    applyListChanges(element, changes, filterFunc)
```
yelly, line 408:
```suggestion
  _.orderBy(realListItemGroup, group => group[0].id.nestingLevel, 'desc').forEach(changes => {
    applyListChanges(element, changes, filterFunc)
```
Just a matter of preference, take it or leave it.

packages/adapter-utils/src/list_comparison.ts
```diff
@@ -40,6 +41,42 @@ const log = logger(module)
 
 type KeyFunction = (value: Value) => string
 
+type TopLevelType = PrimitiveValue | ReferenceExpression | TemplateExpression | StaticFile
```
yelly, line 44: Maybe find a different name for this? `TopLevel` is overloaded.

packages/adapter-utils/src/list_comparison.ts
```diff
@@ -40,6 +41,42 @@ const log = logger(module)
...
+      afterValue: Value
+    }
+  | {
+      action: 'modify'
+      beforeIndex: number
+      afterIndex: number
+      afterValue: Value
+    }
+
+export const getChangeRealId = (change: DetailedChange): ElemID =>
```
yelly, line 68: Don't love this name but I'm not sure I have a better suggestion.

packages/adapter-utils/src/list_comparison.ts
```diff
@@ -40,6 +41,42 @@ const log = logger(module)
...
+  (isRemovalChange(change) ? change.elemIDs?.before : change.elemIDs?.after) ?? change.id
+
+/**
+ * This function returns if a change contains a moving of a item in a list for one index to another
```
yelly, line 72:
```suggestion
 * Returns whether a change contains a moving of an item in a list for one index to another
```

packages/adapter-utils/src/list_comparison.ts
```diff
@@ -40,6 +41,42 @@ const log = logger(module)
...
+ */
+export const isOrderChange = (change: DetailedChange): boolean =>
+  isIndexPathPart(change.id.name) &&
+  isIndexPathPart(change.elemIDs?.before?.name ?? '') &&
+  isIndexPathPart(change.elemIDs?.after?.name ?? '') &&
+  Number(change.elemIDs?.before?.name) !== Number(change.elemIDs?.after?.name)
```
yelly, line 78: Why do we need to convert to a number before comparing?

packages/adapter-utils/src/list_comparison.ts
```diff
@@ -274,6 +304,90 @@ export const getArrayIndexMapping = (before: Value[], after: Value[]): IndexMapp
...
 
+const hasBeforeIndex = (change: ListChange): change is ListChange & { action: 'remove' | 'modify' } =>
+  change.action !== 'add'
+
+const hasAfterIndex = (change: ListChange): change is ListChange & { action: 'add' | 'modify' } =>
+  change.action !== 'remove'
+
+// eslint-disable-next-line consistent-return
+const toListChange = (change: DetailedChange): ListChange => {
+  // eslint-disable-next-line default-case
```
yelly, line 315: I think it's better to refactor and get rid of these exceptions. Just make the `modify` case the default and you'll be fine. The type system means these are equivalent anyway.

packages/adapter-utils/src/list_comparison.ts
```diff
@@ -40,6 +41,42 @@ const log = logger(module)
...
 type KeyFunction = (value: Value) => string
 
+type TopLevelType = PrimitiveValue | ReferenceExpression | TemplateExpression | StaticFile
+
+type IndexMappingItem = {
+  beforeIndex?: number
+  afterIndex?: number
+}
+
+type ListChange =
```
yelly, line 51: I think it's better to define this with generics and get (some) type safety instead of using `Value` here and `updateList` and `updateListWithFilteredChanges` below.

packages/adapter-utils/src/list_comparison.ts
```diff
@@ -274,6 +304,90 @@ export const getArrayIndexMapping = (before: Value[], after: Value[]): IndexMapp
...
+}
+
+const updateList = (list: unknown[], changes: ListChange[]): void => {
+  const removalsAndModifications = changes.filter(hasBeforeIndex)
+  removalsAndModifications.forEach(({ beforeIndex }) => {
+    list[beforeIndex] = undefined
+  })
+
+  const sortedAdditionsAndModifications = _.sortBy(changes.filter(hasAfterIndex), change => change.afterIndex)
+  sortedAdditionsAndModifications.forEach(({ afterIndex, afterValue }) => {
```
yelly, line 341:
```suggestion
  _.sortBy(changes.filter(hasAfterIndex), change => change.afterIndex).forEach(({ afterIndex, afterValue }) => {
```
Also here, just personal preference.

packages/adapter-utils/src/list_comparison.ts
dantal4: `FlatValue` ?

packages/adapter-utils/src/list_comparison.ts
dantal4: generally I didn't change this function... just moved it from `compare.ts`. I think that we can avoid the conversion.

packages/adapter-utils/src/compare.ts
yelly: Yeah, but can't it instead more accurately receive `(before: ChangeDataType, after: ChangeDataType)`?
Keep in mind that `ChangeDataType = TopLevelElement | Field` which is more specific than `Element`.

packages/adapter-utils/src/list_comparison.ts
yelly: Maybe `TerminalValue` or `LeafValue`?
`Flat` makes me think of lists.

packages/adapter-utils/src/compare.ts
dantal4: fetch returns `Element[]`, so `detailedCompare` should receive `Element` too.

packages/adapter-utils/src/list_comparison.ts
dantal4: I'll change it to `LeafValue`

yelly (approved):

PR #6146, opened 2024-07-01
Salto 5959 bottleneck package performance is limiting adapter performance

IlayDavid95:
Switched bottleneck rate-limiting package for p-queue in the main adapter-component infrastructure. Added an abstraction over it, _RateLimiter_, for future ease of maintenance.

---

Bottleneck package had an inherent overhead of ~2.5-3.5 ms for each function call (hence its replacement). In order to make the transition as smooth as possible delayPerRequestMS was added to the generic adapter configuration (nacl) with the default value being 3ms.
 
It is assumed that not all adapters are aligned with the latest infrastructure in adapter-component, however I have made only a low-effort update to those that did. Any adapter not conforming to the infrastructure has been, sadly, left behind and will not be able to configure the delay on a per-user basis.

Some of the adapters actually were _reliant_ on the _faulty_ delay inherent in bottleneck, please note that. (for example, salesforce).

---
_Release Notes_: 
_Replace me with a short sentence that describes the effect of this change on Salto users_

---
_User Notifications_: 
_Replace me with a short sentence that describes changes that will appear in NaCls and are not caused by user actions (e.g. a new annotation, field values that are converted to references, etc). Hidden changes should not be listed._

PR #6270, opened 2024-07-23
SALTO-5959 Jira should use bottleneck by default

kami-delaroz:
SALTO-5959 Jira should use bottleneck by default

---

https://github.com/salto-io/salto/pull/6146/

---
_Release Notes_: 
none

---
_User Notifications_: 
none


IdoZakk (approved):
:great:

PR #6276, opened 2024-07-24
SALTO-6319: Add the ability to ignore errors during fetch

shir-reifenberg:
Added the option to ignore known errors during fetch and suppress them 

---

changes - 
- By default, all errors will not ignored and will trigger `log.error` (even when not providing error handler) with "failed to fetch type".
- When customizing `onError`, returning `ignoreError` will make the error to be suppressed (to debug level).

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/fetch/element/element.ts
+++ b/packages/adapter-components/src/fetch/element/element.ts
@@ -95,21 +95,22 @@ export const getElementGenerator = <Options extends FetchApiDefinitionsOptions>(
     const onErrorResult = onError?.custom?.(onError)({ error, typeName }) ?? onError
     switch (onErrorResult?.action) {
       case 'customSaltoError':
-        if (onErrorResult.ignore) {
-          log.debug('suppressing type failure for %s:%s, generating custom Salto error', adapterName, typeName)
-        } else {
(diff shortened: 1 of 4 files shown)


netama:
will think about it a bit more, but I'm not convinced adding this as another "dimension" is justified yet as I think it might require a bit too many decisions during development. I thought this would be another `OnErrorHandler`, and we'd only log an error as the default fallback if no handler caught it - but maybe I'm missing other use cases

packages/adapter-components/src/fetch/element/element.ts
```diff
@@ -110,7 +119,7 @@ export const getElementGenerator = <Options extends FetchApiDefinitionsOptions>(
       // eslint-disable-next-line no-fallthrough
       case undefined:
       default:
-        log.warn('failed to fetch type %s:%s: %s', adapterName, typeName, error.message)
+        log.error('failed to fetch type %s:%s: %s', adapterName, typeName, error.message)
```
netama, line 122: nit: maybe make the beginning of the sentence slightly more indicative so that it's easier to distinguish from the other cases? (like adding "unexpectedly")

netama (approved):
🎈 🎈

PR #6277, opened 2024-07-24
SALTO-6323 always support none pagination in endpoints

netama:
The default defined on the `Options` resolve functions did not support none when other pagination functions are defined, so there was no good way to "override" the pagination for specific endpoints (or to set it as the default while also supporting other functions).

---

Also cleaning up the use of the option in the adapter definitions - this can also be done for the other exported types but will do it separately.

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/fetch/request/requester.ts
+++ b/packages/adapter-components/src/fetch/request/requester.ts
@@ -28,6 +28,7 @@ import { computeArgCombinations } from '../resource/request_parameters'
 import {
   APIDefinitionsOptions,
   HTTPEndpointDetails,
+  PaginationDefinitions,
   ResolveClientOptionsType,
   ResolvePaginationOptionsType,
 } from '../../definitions/system'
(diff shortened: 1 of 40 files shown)


packages/adapter-components/src/fetch/request/requester.ts
```diff
@@ -126,11 +127,16 @@ export const getRequester = <Options extends APIDefinitionsOptions>({
...
-        : { funcCreator: noPagination, clientArgs: undefined }
+    const paginationOption = mergedRequestDef.endpoint.pagination ?? 'none'
+    const nonePaginationDef: PaginationDefinitions<ResolveClientOptionsType<Options>> = {
+      funcCreator: noPagination,
+      clientArgs: undefined,
+    }
+    const paginationWithNone = {
+      ...pagination,
+      none: nonePaginationDef,
+    } as Record<ResolvePaginationOptionsType<Options>, PaginationDefinitions<ResolveClientOptionsType<Options>>>
```
netama, line 138: not too happy with this cast but I think it's safe

shir-reifenberg (approved):
👯

PR #6281, opened 2024-07-24
SALTO-6324: Fix - default `custom` error handler overrides any type specific error handler definition

shoshana-houri:
**Problem:**
If a _custom_ error handler is defined as the default behavior in the fetch definition, and another non customized `onError` behavior is defined on a specific type - the latter should be used, not the default one. But it is not the case since nothing overrides the 'custom' field.

**Solution:**
Use the same structure of 'action' and 'value' for the customized handler as for the predefined ones. This way, the 'custom' field will be overridden the same way as the predefined ones.

---

---
_Release Notes_: 

---
_User Notifications_:

### DIFF
--- a/packages/adapter-components/src/definitions/system/fetch/resource.ts
+++ b/packages/adapter-components/src/definitions/system/fetch/resource.ts
@@ -15,7 +15,7 @@
  */
 import { SaltoError } from '@salto-io/adapter-api'
 import { ContextCombinationDefinition, RecurseIntoDefinition } from './dependencies'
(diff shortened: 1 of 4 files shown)


netama:
thanks! commented on the ticket - I'm not sure we should do this as I think we can always explicitly unset and I tend to think the more common behavior will actually be to want to keep the custom function - but happy to discuss :)

shoshana-houri:
Decided to explicitly unset the `custom` error handler as suggested by @netama. Closing.

PR #6282, opened 2024-07-24
SALTO-6299 fix fixElements of form fields that reference objects with appid

dantal4:
in case that a form field reference an object with `appid`, we should check if that app id is even being fetched, otherwise the object won't be in the `objectid_list` instance, and we'll remove it.

---

---
_Release Notes_: 
Netsuite Adapter:
- fix fixElements of form fields that reference objects with appid

---
_User Notifications_: 
None

### DIFF
--- a/packages/netsuite-adapter/src/custom_references/weak_references/fields_references.ts
+++ b/packages/netsuite-adapter/src/custom_references/weak_references/fields_references.ts
@@ -51,6 +51,8 @@ const { awu } = collections.asynciterable
 
 type MappedList = Record<string, { index: number; [key: string]: Value }>
 
+const NO_APPID = ''
+
 const formTypeNames = new Set([ADDRESS_FORM, ENTRY_FORM, TRANSACTION_FORM])
 
 const isFormInstanceElement = (element: Value): element is InstanceElement =>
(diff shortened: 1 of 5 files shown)


packages/netsuite-adapter/src/custom_references/weak_references/fields_references.ts
```diff
@@ -114,19 +114,25 @@ const isUnresolvedReferenceExpression = async (
...
-  envScriptIds: Set<string>,
+  envScriptIdsBySuiteAppId: Record<string, Set<string>>,
 ): boolean => {
   const capture = captureServiceIdInfo(value)
-  return capture
-    .map(serviceIdInfo => serviceIdInfo.serviceId.split('.')[0])
-    .some(serviceId => !generatedDependencies.has(serviceId) && !envScriptIds.has(serviceId))
+  return capture.some(serviceId => {
+    const objectServiceId = serviceId.serviceId.split('.')[0]
+    const suiteAppId = serviceId.appid ?? ''
```
omrilit, line 122: Please extract `''` to a const, so the usage of it won't be implicit.

omrilit (approved):
Looks great!
Just making sure that I am not missing anything - you solved the issue properly and not with a bandaid eventually, right?

PR #6283, opened 2024-07-24
SALTO-6264: Renaming CustomObject and CustomMetadata types

yelly:
Added an underscore before so we can free these element IDs up for meta types.
Also syncified some logic during debugging which is nice to have generally.
The rename for the type `CustomMetadata` is behind the meta types feature flag because it isn't hidden (yet).
This is round 2 of #6232.

---

I tested fetching and deploying and also deploying without fetching.
The latter works fine is because all of the places we rely on the new names are either during fetch (in which case we always already have them) or during deploy where we create them artificially and don't rely on what we got in fetch.

---
_Release Notes_: 
_Salesforce_:
* Renamed the internal and hidden `CustomObject` and `CustomMetadata` types to `_CustomObject` and `_CustomMetadata`.

---
_User Notifications_: 
None.

### DIFF
--- a/packages/salesforce-adapter/src/adapter.ts
+++ b/packages/salesforce-adapter/src/adapter.ts
@@ -33,8 +33,9 @@ import {
   Element,
   ProgressReporter,
   isInstanceElement,
+  ServiceIds,
 } from '@salto-io/adapter-api'
-import { filter, logDuration, safeJsonStringify } from '@salto-io/adapter-utils'
+import { filter, logDuration, naclCase, safeJsonStringify } from '@salto-io/adapter-utils'
 import { resolveChangeElement, restoreChangeElement } from '@salto-io/adapter-components'
 import { MetadataObject } from '@salto-io/jsforce'
 import _ from 'lodash'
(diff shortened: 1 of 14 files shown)


packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -224,7 +224,7 @@ const processDeployResponse = (
       log.debug('Unable to match deploy message for %s[%s] with an ElemID.', fullName, componentType)
       return undefined
     }
-    if (rawElemId.typeName === CUSTOM_OBJECT) {
+    if (rawElemId.typeName === CUSTOM_OBJECT_TYPE_NAME) {
```
tamtamirr, line 227: Won't this break when the metaTypes feature is disabled?

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -504,8 +505,18 @@ export default class SalesforceAdapter implements AdapterOperations {
...
+      ? (adapterName: string, serviceIds: ServiceIds, name: string): ElemID => {
+          const updatedName = name === constants.CUSTOM_METADATA ? constants.CUSTOM_METADATA_TYPE_NAME : name
+          return (
+            getElemIdFunc?.(adapterName, serviceIds, updatedName) ??
+            new ElemID(constants.SALESFORCE, naclCase(updatedName))
+          )
+        }
+      : getElemIdFunc
+    if (getElemIdFuncWithRename) {
+      Types.setElemIdGetter(getElemIdFuncWithRename)
```
tamtamirr, line 519: Can we add UT for the CustomMetadata rename?

tamtamirr (commented):
Added some comments

packages/salesforce-adapter/src/metadata_deploy.ts
yelly: No, the feature flag is only on custom metadata since custom object is hidden anyway.

packages/salesforce-adapter/src/adapter.ts
yelly: Done.

tamtamirr (approved):

PR #6284, opened 2024-07-24
SALTO-6317: Move Jsm forms to official api

idoamit8:
In this PR I moved the Form type in Jira to the new API 

---

The values ​​remained the same except for 3 new fields which I will add Noise Reduction.

image

  
From what I see through the CLI - fetch, create, modify and remove all worked great! 
It is important to note some point:
1. The id changed from a number to a string which is actually a uuid and I'm not sure if it has certain meanings or not
since the uuid was deleted, so the ServiceID changed to the id field. Whats going to happen in the next Fetch? 

---
_Release Notes_: 
_Jira Adapter_:
* Moved JSM forms to the new API. 

---
_User Notifications_: 
_Jira Adapter:_
* Changes will be applied in Form instances:
The fields `issueRequestTypeIds`, `newIssueIssueTypeIds`, `newIssueRequestTypeIds` will be replaced by  `recommendedIssueRequestTypeIds`, `issueCreateIssueTypeIds`, `issueCreateRequestTypeIds`.

### DIFF
--- a/packages/jira-adapter/src/client/client.ts
+++ b/packages/jira-adapter/src/client/client.ts
@@ -210,13 +210,42 @@ export default class JiraClient extends clientUtils.AdapterHTTPClient<Credential
     }
     throw new Error('Failed to get GQL response')
   }
-
-  public async atlassianApiSendRequest(
-    method: keyof clientUtils.HttpMethodToClientParams,
+  
+  public async atlassianApiGet(
     args: clientUtils.ClientDataParams,
   ): Promise<clientUtils.Response<clientUtils.ResponseValue | clientUtils.ResponseValue[]>> {
     const cloudId = await this.getCloudId()
-    return this.sendRequest(method, {
+    return this.get({
+      ...args,
+      url: `https://api.atlassian.com/jira/forms/cloud/${cloudId}/${args.url}`,
+    })
+  }
+
+  public async atlassianApiPost(
+    args: clientUtils.ClientDataParams,
+  ): Promise<clientUtils.Response<clientUtils.ResponseValue | clientUtils.ResponseValue[]>> {
+    const cloudId = await this.getCloudId()
+    return this.post({
+      ...args,
+      url: `https://api.atlassian.com/jira/forms/cloud/${cloudId}/${args.url}`,
+    })
+  }
+
+  public async atlassianApiPut(
+    args: clientUtils.ClientDataParams,
+  ): Promise<clientUtils.Response<clientUtils.ResponseValue | clientUtils.ResponseValue[]>> {
+    const cloudId = await this.getCloudId()
+    return this.put({
+      ...args,
+      url: `https://api.atlassian.com/jira/forms/cloud/${cloudId}/${args.url}`,
+    })
+  }
+
+  public async atlassianApiDelete(
+    args: clientUtils.ClientDataParams,
+  ): Promise<clientUtils.Response<clientUtils.ResponseValue | clientUtils.ResponseValue[]>> {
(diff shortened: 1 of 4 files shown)


packages/jira-adapter/src/client/client.ts
```diff
@@ -211,6 +211,17 @@ export default class JiraClient extends clientUtils.AdapterHTTPClient<Credential
...
 
+  public async atlassianApiSendRequest(
+    method: keyof clientUtils.HttpMethodToClientParams,
+    args: clientUtils.ClientDataParams,
+  ): Promise<clientUtils.Response<clientUtils.ResponseValue | clientUtils.ResponseValue[]>> {
+    const cloudId = await this.getCloudId()
+    return this.sendRequest(method, {
+      ...args,
+      url: `https://api.atlassian.com/jira/forms/cloud/${cloudId}/${args.url}`,
+    })
```
IdoZakk, line 222: When you use sendRequest you miss the decorators of the rate limit. 
We should have a call for each of the REST calls (you can add only the ones you currently use)

packages/jira-adapter/test/filters/forms/forms.test.ts
```diff
@@ -693,123 +658,33 @@ describe('forms filter', () => {
...
       elements = [projectInstance, projectType, formPortalType, formPublishType, formType, requestTypeType]
     })
     it('should add form', async () => {
       const res = await filter.deploy([{ action: 'add', data: { after: formInstance } }])
       expect(res.leftoverChanges).toHaveLength(0)
       expect(res.deployResult.errors).toHaveLength(0)
       expect(res.deployResult.appliedChanges).toHaveLength(1)
-      expect(connection.post).toHaveBeenCalledTimes(2)
-      expect(connection.put).toHaveBeenCalledTimes(1)
-      expect(connection.put).toHaveBeenCalledWith(
```
IdoZakk, line 729: Maybe put the equivalent expects here?

packages/jira-adapter/test/filters/forms/forms.test.ts
```diff
@@ -818,16 +693,12 @@ describe('forms filter', () => {
       expect(res.leftoverChanges).toHaveLength(0)
       expect(res.deployResult.errors).toHaveLength(0)
       expect(res.deployResult.appliedChanges).toHaveLength(1)
-      expect(connection.put).toHaveBeenCalledTimes(1)
-      expect(connection.post).toHaveBeenCalledTimes(1)
```
IdoZakk, line 822: I see you removed those in all the tests, I think we should verify the outside calls are called

IdoZakk (commented):
🆕 🥇 
Very exciting!
You mentioned the change of number to uuid. I do not think there is anything that might happen here, but we should notify the CEs that all deploys must be done after a new fetch
Also you mentioned a few fields changing, I did not see the removal of the templateId (which was removed in the code

packages/jira-adapter/src/client/client.ts
idoamit8: I use all of them. I will change it to use each call

idoamit8:
@IdoZakk I resolved your comments :) 
What do you mean by 

> you mentioned a few fields changing, I did not see the removal of the templateId (which was removed in the code

Where did you see it?

IdoZakk:
> @IdoZakk I resolved your comments :) What do you mean by
> 
> > you mentioned a few fields changing, I did not see the removal of the templateId (which was removed in the code
> 
> Where did you see it?

In the PR comment, under `Additional context for reviewer`

IdoZakk (approved):
💣

PR #6286, opened 2024-07-24
SALTO-5432 adjust logs

netama:
format logs and use safe stringify

---
_Release Notes_: 
None

---
_User Notifications_: 
None


shir-reifenberg (approved):
thanks!

PR #6288, opened 2024-07-25
SALTO-6329 upgrade axios

netama:
Upgrade axios so we can benefit from some new features.

There were some minor interface changes, I think they're safe but will appreciate another look.

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/netsuite-adapter/src/client/suiteapp_client/suiteapp_client.ts
```diff
@@ -127,7 +127,7 @@ type VersionFeatures = {
 }
 
 const getAxiosErrorDetailedMessage = (error: AxiosError): string | undefined => {
-  const errorDetails = error.response?.data?.['o:errorDetails']
+  const errorDetails = _.get(error.response?.data ?? {}, 'o:errorDetails')
```
netama, line 130: @dantal4 @omrilit any concerns with this change?

ori-moisis (approved):

PR #6291, opened 2024-07-25
SALTO-6330 allow adjusting serializations for query arg combinations

netama:
Allow serializing combinations of the same query arg in different ways using axios - 
`arg=val1&arg=val2`
`arg[]=val1&arg[]=val2`
`arg[0]=val1&arg[1]=val2`

for use in https://github.com/salto-io/salto/pull/6217

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/client/http_client.ts
+++ b/packages/adapter-components/src/client/http_client.ts
@@ -54,6 +54,13 @@ export type ClientBaseParams = {
   headers?: Record<string, string>
   responseType?: ResponseType
   params?: Record<string, Values>
+  // when multiple query args are provided:
+  queryParamsSerializer?: {
+    // false (default): &arg[]=val1&arg[]=val2
+    // true: &arg[0]=val1&arg[1]=val2
+    // null: &arg=val1&arg=val2
+    indexes?: boolean | null
+  }
 }
 
 export type ClientDataParams = ClientBaseParams & {
@@ -238,7 +245,7 @@ export abstract class AdapterHTTPClient<TCredentials, TRateLimitConfig extends C
       throw new Error(`uninitialized ${this.clientName} client`)
     }
 
-    const { url, queryParams, headers, responseType } = params
+    const { url, queryParams, headers, responseType, queryParamsSerializer: paramsSerializer } = params
     const data = isMethodWithData(params) ? params.data : undefined
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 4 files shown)


packages/adapter-components/src/fetch/request/requester.ts
```diff
@@ -151,7 +151,7 @@ export const getRequester = <Options extends APIDefinitionsOptions>({
         typeName,
       )
 
-    const allCallArgs = _.pick(mergedEndpointDef, ['queryArgs', 'headers', 'body', 'params'])
+    const allCallArgs = _.pick(mergedEndpointDef, ['queryArgs', 'headers', 'body', 'params', 'queryParamsSerializer'])
     const callArgs = mergedEndpointDef.omitBody ? _.omit(mergedEndpointDef, 'body') : allCallArgs
```
omrifarkash, line 155: Irrelevant to this PR.
Isn't this supposed to be ?
```suggestion
    const callArgs = mergedEndpointDef.omitBody ? _.omit(allCallArgs, 'body') : allCallArgs
```

packages/adapter-components/test/auth/oauth.test.ts
```diff
@@ -49,7 +49,7 @@ describe('oauth', () => {
       expect(req.url).toEqual('/oauth/token')
       expect(req.auth).toBeUndefined()
       expect(req.data).toEqual('client_id=client%20id&client_secret=secret&grant_type=client_credentials')
-      expect(req.headers).toEqual({
+      expect({ ...req.headers }).toEqual({
```
omrifarkash, line 52: nit.
It took me a minute to understand why.
Maybe use `expect.objectContaining` Or create `new AxiosHeaders` as you did on other tests?

packages/adapter-components/src/client/http_client.ts
```diff
@@ -54,6 +54,13 @@ export type ClientBaseParams = {
   headers?: Record<string, string>
   responseType?: ResponseType
   params?: Record<string, Values>
+  // when multiple query args are provided:
+  queryParamsSerializer?: {
+    // false (default): &arg[]=val1&arg[]=val2
```
omrifarkash, line 59: If `queryParamsSerializer` is undefined. It behave like `indexes` is false or null?
Do we want our default to behave like it is null? (from Confluence point of view this is what I would expect to be the default, but no idea what is the common case)

omrifarkash (approved):
LGTM

packages/adapter-components/src/client/http_client.ts
netama: false, like the default. undefined !== null. this is the axios default

packages/adapter-components/test/auth/oauth.test.ts
netama: this was from the bump PR (which is now merged), there was an interface change that required this (it wasn't in the last commit)

packages/adapter-components/src/fetch/request/requester.ts
netama: yes, thanks

PR #6302, opened 2024-07-28
SALTO-6350: Modify ./create_version_pr.sh script

idoamit8:
In this PR I modified ./create_version_pr.sh  script

---

@Eshed Gal-Or [PR](https://github.com/salto-io/salto/pull/6108) added dependency to our Salto release version in the yarn.lock

For example: 

"@salto-io/adapter-api": 0.3.60

Meaning that we should add yarn to modify this yarn.lock as part of the script of OSS release.

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/build_utils/create_version_pr.sh
+++ b/build_utils/create_version_pr.sh
@@ -41,7 +41,7 @@ fi
 BUMP=${1:-patch}
 
 yarn lerna-version -y $BUMP
-yarn; yarn; yarn 
+yarn --immutable
 
 DIRTY_FILES=$(git diff-index --name-only HEAD)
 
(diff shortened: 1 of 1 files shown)


build_utils/create_version_pr.sh
```diff
@@ -41,6 +41,7 @@ fi
 BUMP=${1:-patch}
 
 yarn lerna-version -y $BUMP
+yarn; yarn; yarn
```
dantal4, line 44: can be replaced with `yarn --immutable` (this was also updated in [this PR](https://github.com/salto-io/salto/pull/6108/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L7-R14))

dantal4 (approved):
💯

PR #6304, opened 2024-07-29
SALTO-6351: Okta: Add logs for `isCustomApp` matching methods.

Nurdok:
There is a bug in the current `isCustomApp` detection method when subdomains are changed. We suspect that checking for a suffix of underscore + counter (`_1`, `_2` etc.) would be sufficient and work around this bug. This PR adds some logs so we can gather some data on this.

---

I've also modified some tests to match what we know about custom app names.

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/definitions/fetch/types/application.ts
+++ b/packages/okta-adapter/src/definitions/fetch/types/application.ts
@@ -65,8 +65,34 @@ export const assignPolicyIdsToApplication = (value: unknown): Value => {
   }
 }
 
-export const isCustomApp = (value: Values, subdomain: string): boolean =>
-  [AUTO_LOGIN_APP, SAML_2_0_APP].includes(value.signOnMode) &&
-  value.name !== undefined &&
-  // custom app names starts with subdomain and '_'
-  _.startsWith(value.name, `${subdomain}_`)
+const endsWithNumberRegex = new RegExp(/_\d+$/)
+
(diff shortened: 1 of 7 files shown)


packages/okta-adapter/src/definitions/fetch/types/application.ts
```diff
@@ -65,8 +65,34 @@ export const assignPolicyIdsToApplication = (value: unknown): Value => {
...
+    value.name !== undefined &&
+    // custom app names starts with subdomain and '_'
+    _.startsWith(value.name, `${subdomain}_`)
+
+  const endsWithNumberMatch =
+    value.name !== undefined &&
+    // custom app names ends with a number
+    endsWithNumberRegex.test(value.name)
+
+  if (subdomainMatch !== endsWithNumberMatch) {
```
omrilit, line 81: Should we check it maybe only when `[AUTO_LOGIN_APP, SAML_2_0_APP].includes(value.signOnMode)` only evaluated to `true`? or is it interesting anyway?

omrilit (approved):
LGTM!

packages/okta-adapter/src/definitions/fetch/types/application.ts
Nurdok: It's interesting anyway, because we want data for non-custom apps too.

PR #6306, opened 2024-07-29
SALTO-2392 Identify ASVs in multiselect field

dantal4:
Identify ASVs in multiselect field - **only when there's a single value selected** (because we don't deal with `[ACCOUNT_SPECIFIC_VALUE]|[ACCOUNT_SPECIFIC_VALUE]...` values)

---

---
_Release Notes_: 
Netsuite Adapter:
- Identify ASVs in multiselect field

---
_User Notifications_: 
None


omrilit (approved):
LGTM!
Please merge only once the other ASV related identifications are ready so we will have a concentrated noisy fetch

PR #6307, opened 2024-07-29
SALTO-6353: Salesforce adapter referenceTo references from Lookup fields bug-fix

tamtamirr:
Salesforce adapter referenceTo references from Lookup fields bug-fix

---

Fixed a recent regression where referenceTo references were not created successfully in Lookup fields.

---

_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/src/filters/reference_annotations.ts
+++ b/packages/salesforce-adapter/src/filters/reference_annotations.ts
@@ -43,7 +43,10 @@ const convertAnnotationsToReferences = async (
   const resolveTypeReference = (ref: string | ReferenceExpression): string | ReferenceExpression => {
     if (_.isString(ref)) {
       // Try finding a metadata type and fallback to finding a custom object
-      const referenceElement = nameToElement.get(ref, ref) ?? nameToElement.get(CUSTOM_OBJECT_TYPE_NAME, ref)
+      const referenceElement =
+        nameToElement.get(ref, ref) ??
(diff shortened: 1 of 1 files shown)


packages/salesforce-adapter/src/filters/reference_annotations.ts
```diff
@@ -43,7 +43,10 @@ const convertAnnotationsToReferences = async (
   const resolveTypeReference = (ref: string | ReferenceExpression): string | ReferenceExpression => {
     if (_.isString(ref)) {
       // Try finding a metadata type and fallback to finding a custom object
-      const referenceElement = nameToElement.get(ref, ref) ?? nameToElement.get(CUSTOM_OBJECT_TYPE_NAME, ref)
+      const referenceElement =
+        nameToElement.get(ref, ref) ??
+        nameToElement.get(CUSTOM_OBJECT, ref) ??
+        nameToElement.get(CUSTOM_OBJECT_TYPE_NAME, ref)
```
tamtamirr, line 49: This should cover both cases, before and after the rename of the Metadata type `CustomObject`. Again, I was never a fun of this rename. 🤦

amitgoldarb: Can we add a small UT for this case.

amitgoldarb (approved):
Thanks for this

PR #6309, opened 2024-07-29
SALTO-6354 definition overrides env var work by account name

Uri-Bar:
SALTO-6354 defention overrides env var work by account name
---

---
_Release Notes_: 
none

---
_User Notifications_: 
none

### DIFF
--- a/packages/adapter-components/src/definitions/system/overrides.ts
+++ b/packages/adapter-components/src/definitions/system/overrides.ts
@@ -63,7 +63,7 @@ export const mergeDefinitionsWithOverrides = <Options extends APIDefinitionsOpti
   }
   log.debug('starting to merge definitions with overrides')
   const overrides = getParsedDefinitionsOverrides(account)
-  if (_.isEmpty(overrides)) {account
+  if (_.isEmpty(overrides)) {
     return definitions
   }
   log.debug('Definitions overrides: %s', safeJsonStringify(overrides))
--- a/packages/adapter-components/src/definitions/system/overrides.ts
+++ b/packages/adapter-components/src/definitions/system/overrides.ts
@@ -23,12 +23,12 @@ import { APIDefinitionsOptions } from './api'
 const log = logger(module)
 export const DEFINITIONS_OVERRIDES = 'SALTO_DEFINITIONS_OVERRIDES'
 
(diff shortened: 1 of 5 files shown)


packages/adapter-components/src/definitions/system/overrides.ts
```diff
@@ -23,12 +23,12 @@ import { APIDefinitionsOptions } from './api'
...
 export const DEFINITIONS_OVERRIDES = 'SALTO_DEFINITIONS_OVERRIDES'
 
-const getParsedDefinitionsOverrides = (): Values => {
+const getParsedDefinitionsOverrides = (account: string): Values => {
   const overrides = process.env[DEFINITIONS_OVERRIDES]
   try {
     const parsedOverrides = overrides === undefined ? undefined : JSON.parse(overrides)
     if (parsedOverrides !== undefined && typeof parsedOverrides === 'object') {
-      return parsedOverrides as Values
+      return parsedOverrides[account] as Values
```
netama, line 31: this is problematic, you are losing the safety that ensures it is undefined. please either move the checks to above, or add a "real" typeguard and avoid the cast

packages/adapter-components/src/definitions/system/overrides.ts
```diff
@@ -50,15 +50,19 @@ const getParsedDefinitionsOverrides = (): Values => {
  */
 export const mergeDefinitionsWithOverrides = <Options extends APIDefinitionsOptions>(
   definitions: RequiredDefinitions<Options>,
+  account?: string
```
netama, line 53: nit: can we avoid the naming inconsistency between `accountName` and `account`?

packages/adapter-components/src/definitions/system/overrides.ts
```diff
@@ -50,15 +50,19 @@ const getParsedDefinitionsOverrides = (): Values => {
...
 ): RequiredDefinitions<Options> => {
   const customMerge = (objValue: Value, srcValue: Value): Value => {
     if (_.isArray(objValue)) {
       return srcValue
     }
     return undefined
   }
+  if (account === undefined) {
+    return definitions
```
netama, line 62: is this an error flow? if so let's add a log error?

netama (approved):
thanks!! added some comments (one blocker), pre-approving

PR #6313, opened 2024-07-29
SALTO-5692, SALTO-6358: Okta: Upgrade `ApplicationGroupAssignment` deploy definitions to new infra

Nurdok:
This PR adds HTTP-recording tests to `ApplicationGroupAssignment` _and_ upgrades this type to the new infra. This also fixes a bug where `modify` always fails with HTTP 400 status code due to `appId` and `groupId` not being omitted.

---

None

---
_Release Notes_: 
_Okta adapter_:
- Fixed a bug where modifying `ApplicationGroupAssignment` failed with a 400 HTTP status code.

---
_User Notifications_: 
None


packages/okta-adapter/src/definitions/requests/clients.ts
```diff
@@ -87,6 +87,11 @@ export const createClientDefinitions = (
               omitBody: true,
             },
           },
+          '/api/v1/apps/{appId}/groups/{id}': {
```
shir-reifenberg, line 90: Should we add it under `default` ?

packages/okta-adapter/test/adapter.test.ts
```diff
@@ -1112,23 +1122,123 @@ describe('adapter', () => {
...
+      })
+
+      it('should successfully add an app group assignment', async () => {
+        loadMockReplies('app_group_assignment_add.json')
+        const appGroupAssignment = new InstanceElement(
+          'appGroupAssignment',
+          appGroupAssignmentType,
+          {
+            id: group.value.id,
+            priority: 2,
```
shir-reifenberg, line 1158: nit: let's add the `groupId` and `appId` fields to value to verify this test will break if somehow the code that currently omits those will change?

shir-reifenberg (approved):
Looks great with 1 question 😃

packages/okta-adapter/src/definitions/requests/clients.ts
Nurdok: I'll add as default in a separate PR.

PR #6321, opened 2024-07-30
Return nicer SOAP deploy error messages

dantal4:
Since each deploy error is attached to the failed change id, the error message we're returning for data instances deploy can be more friendly to the user:
instead of `SOAP api call <...> for instance <...> failed. error code: <...>, error message: <...>` we can just return the error message from SOAP. it will be much more readable for users.

In addition, make SOAP deploy return a nicer return type - `SoapDeployResult` instead of `number | Error`.

---

---
_Release Notes_: 
Netsuite Adapter:
- Return nicer SOAP deploy error messages

---
_User Notifications_: 
None

### DIFF
--- a/packages/netsuite-adapter/src/client/suiteapp_client/soap_client/types.ts
+++ b/packages/netsuite-adapter/src/client/suiteapp_client/soap_client/types.ts
@@ -72,6 +72,16 @@ export const isWriteResponseSuccess = (result: WriteResponse): result is WriteRe
 export const isWriteResponseError = (result: WriteResponse): result is WriteResponseError =>
   result.status.attributes.isSuccess === 'false'
 
+export type SoapDeployResult =
+  | {
+      isSuccess: true
+      internalId: string
+    }
+  | {
+      isSuccess: false
+      errorMessage: string
+    }
+
 export type DeployListSuccess = {
   writeResponseList: {
     writeResponse: WriteResponse[]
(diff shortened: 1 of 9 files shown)


packages/netsuite-adapter/src/client/suiteapp_client/soap_client/types.ts
```diff
@@ -72,6 +72,16 @@ export const isWriteResponseSuccess = (result: WriteResponse): result is WriteRe
 export const isWriteResponseError = (result: WriteResponse): result is WriteResponseError =>
   result.status.attributes.isSuccess === 'false'
 
+export type SoapDeployResult =
+  | {
+      isSuccess: true
+      internalId: string
```
omrilit, line 78: What is the effect of modifying the type from number to string?

omrilit (approved):
Looks great1

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/types.ts
dantal4: the only place we used the internalId as a number was to transform it back to string in [here](https://github.com/salto-io/salto/pull/6321/files#diff-16e4347c6fa5012d75fea713766c6853235dc29dc40651faa0a3ea58cb9bb44bR147)

PR #6325, opened 2024-07-31
SALTO-5692: Okta: Upgrade `AuthorizationServerPolicy` deploy definitions to new infra 

Nurdok:
See title.
This PR also fixes a bug where we send an unnecessary PUT request when only activating/deactivating without modifications.

---

None

---
_Release Notes_: 
_Okta adapter_:
- Removed an unnecessary API request when only activating / deactivating an `AuthorizationServerPolicy` instance without other modifications.

---
_User Notifications_: 
None


packages/okta-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -129,6 +130,115 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
...
+            },
+          ],
+          modify: [
+            {
+              condition: {
+                skipIfIdentical: true,
+                transformForCheck: {
+                  omit: ['status'],
+                },
+              },
```
shir-reifenberg, line 156: 🆒

packages/okta-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -129,6 +130,115 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
...
+        if (isAdditionChange(change) && getChangeData(change).value.status === INACTIVE_STATUS) {
+          return ['add', 'deactivate']
+        }
+        if (isModificationChange(change)) {
+          if (isActivationChange(change)) {
+            return ['modify', 'activate']
+          }
+          if (isDeactivationChange(change)) {
+            return ['deactivate', 'modify']
+          }
```
shir-reifenberg, line 219: just curiosity, is there a specific reason to make the changes on inactive status? (or this is just copied from the old implementation)

packages/okta-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -129,6 +130,115 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
...
+                },
+                context: {
+                  authorizationServerId: '{_parent.0.id}',
+                },
+              },
+            },
+          ],
+        },
+      },
+      toActionNames: ({ change }) => {
```
shir-reifenberg, line 209: I think this pattern might be useful in other cases as well? maybe consider extracting this to a function (can also do in the next time we encounter this)

shir-reifenberg: also I think we should add some logs

shir-reifenberg (commented):
LGTM 🚀

shir-reifenberg (approved):
Accidentally didn't approve in the previous comment

packages/okta-adapter/src/definitions/deploy/deploy.ts
Nurdok: It will definitely be useful in other cases, I just wanted to do the refactoring once there is more than one type that has the same logic, so that it's easier to see which parts are generic and which are specific to this type.

TL;DR: Soon :)

packages/okta-adapter/src/definitions/deploy/deploy.ts
Nurdok: I don't quite understand the question. If there's both a modification and a deactivation, I think we should do both, even if it "doesn't matter" since the instance is deactivated.

PR #6328, opened 2024-07-31
SALTO-6376: Add config creator with enable deploy option to Workato

shir-reifenberg:
Add config creator to workato to allow controlling deploy settings

---

Also added some code to adapter utils 

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-utils/src/config_options.ts
+++ b/packages/adapter-utils/src/config_options.ts
@@ -13,20 +13,21 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-export * from './src/change'
-export * from './src/compare'
-export * from './src/decorators'
-export * from './src/dependencies'
-export * from './src/element_source'
-export * from './src/element'
-export * as filter from './src/filter'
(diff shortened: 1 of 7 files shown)


packages/adapter-utils/src/config_options.ts
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ElemID, InstanceElement } from '@salto-io/adapter-api'
+import { logger } from '@salto-io/logging'
+
+const log = logger(module)
+
+export const isOptionsTypeInstance = <T>(
```
netama, line 22: a bit of a nit, but it seems odd that a type guard gets the elem id as input. can we either split this into two checks, or have a function that receives the elem id and returns the type guard? can be something like `createOptionsTypeGuard`, that receives an elem id and returns a guard

netama (approved):
LGTM code-wise with one nit.
I'm not in the details of the final plan though, assuming this is ironed out

PR #6330, opened 2024-08-01
SALTO-6357 fix serviceId is visible on target environment after deployment

omrifarkash:
- when creating type from entries and definitions. If there are no entries contains fields that defined to be hidden in the definitions: 
- Create field with type `unknown` and `_hidden_value` annotation
- If field is defined as `serviceIdField` in `resource` definitions, mark it as `serviceId` field instead of `unknown`

---

---

_Release Notes_: 
- Fix hidden fields are written to the nacl after deploy of first instance from a type

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/fetch/element/type_utils.ts
+++ b/packages/adapter-components/src/fetch/element/type_utils.ts
@@ -225,6 +225,8 @@ export const markServiceIdField = (
   }
 }
 
+// Verify that field is not undefined if it is a serviceId or field we want to hide.
+// This is to prevent writing to nacl (copyFromResponse on deployment) fields that supposed to be hidden.
 const overrideServiceIdOrHiddenFieldIfNotDefined = ({
   definedTypes,
   type,
@@ -239,7 +241,6 @@ const overrideServiceIdOrHiddenFieldIfNotDefined = ({
   isHiddenField?: boolean
 }): void => {
   if (type.fields[fieldName] === undefined) {
-    // we don't want field to be undefined if it is a serviceId or field we want to hide
     if (isServiceIdField) {
       overrideFieldType({
         type,
--- a/packages/adapter-components/src/fetch/element/type_utils.ts
+++ b/packages/adapter-components/src/fetch/element/type_utils.ts
@@ -225,6 +225,36 @@ export const markServiceIdField = (
   }
 }
 
+const overrideServiceIdOrHiddenFieldIfNotDefined = ({
+  definedTypes,
+  type,
+  fieldName,
+  isServiceIdField,
+  isHiddenField,
+}: {
+  definedTypes: Record<string, ObjectType>
+  type: ObjectType
+  fieldName: string
+  isServiceIdField?: boolean
+  isHiddenField?: boolean
+}): void => {
+  if (type.fields[fieldName] === undefined) {
+    // we don't want field to be undefined if it is a serviceId or field we want to hide
+    if (isServiceIdField) {
+      overrideFieldType({
+        type,
+        definedTypes,
+        fieldName,
+        fieldTypeName: BUILTIN_TYPE_NAMES.STRING, // fallback to string serviceId
+      })
+      return
+    }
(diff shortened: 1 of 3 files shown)


packages/adapter-components/src/fetch/element/type_utils.ts
```diff
@@ -246,17 +247,29 @@ export const overrideFieldTypes = <Options extends FetchApiDefinitionsOptions>({
...
+            overrideFieldType({
+              type,
+              definedTypes,
+              fieldName,
+              fieldTypeName: BUILTIN_TYPE_NAMES.STRING, // fallback to string serviceId
+            })
+          } else if (hide) {
+            overrideFieldType({ type, definedTypes, fieldName, fieldTypeName: BUILTIN_TYPE_NAMES.UNKNOWN })
+          }
+        }
```
shir-reifenberg, line 266: Can we please extract this logic to a separate function?

packages/adapter-components/src/fetch/element/type_utils.ts
```diff
@@ -246,17 +247,29 @@ export const overrideFieldTypes = <Options extends FetchApiDefinitionsOptions>({
...
-        const field = type.fields[fieldName]
-        if (field === undefined) {
+        if (type.fields[fieldName] === undefined) {
+          // we don't want field to be undefined if it is a serviceId or field we want to hide
+          if (resourceDef?.serviceIDFields?.includes(fieldName)) {
+            overrideFieldType({
+              type,
+              definedTypes,
+              fieldName,
+              fieldTypeName: BUILTIN_TYPE_NAMES.STRING, // fallback to string serviceId
```
shir-reifenberg, line 261: (not sure it actually matters) is there a specific reason to use `STRING` ? I think also service id can also be a number so `UNKOWN`?

shir-reifenberg (commented):
Looks great 🎉 
I wonder if we should apply the same logic even when `typeAreFinal` provided to the element definition (currently we skip the "fix" for this case)
LMK WDYT :)

packages/adapter-components/src/fetch/element/type_utils.ts
omrifarkash: There is a different. 
You are right, serviceId can be string or number,|
Here we will fallback to serviceId as string, and if this fallback is wrong we will have a workspace warning which will be fixed in the next fetch.

If we fallback to `UNKNOWN` as you suggested, we won't have a type mismatch warning, but, [here](https://github.com/salto-io/salto/pull/6330/files#diff-10fd46aea2a8f372c835a7ca5a3d62b78fa3e3cf5fed2b1d96ae263feed786e0R277) we are marking 'string' and 'number' field as serviceId field when needed, 'unknown' won't be marked as serviceId and we will lose the abilities of serviceId field (For example, we will recalculate elementID on the next fetch if this field value is different)

packages/adapter-components/src/fetch/element/type_utils.ts
```diff
@@ -224,6 +225,36 @@ export const markServiceIdField = (
   }
 }
 
+const overrideServiceIdOrHiddenFieldIfNotDefined = ({
```
shir-reifenberg, line 228: nit: maybe add a short comment explaining why we're doing it and what cases does it fix? (can replace the comment in line 242)

shir-reifenberg (approved):
Looks great 💯 
Let's catch-up with @netama when she's back re `typesAreFinal` definition
and please don't forget to add release notes 🙏

omrifarkash:
After f2f conversation
Due to the urgency of this PR @shir-reifenberg and I decided to merge it without moving the logic of `overrideServiceIdOrHiddenFieldIfNotDefined`
before `typesAreFinal` early exit. This means that if `typesAreFinal` is true in the definition we won't fix the potential bug we are attempting to address in this task. 

cc @netama WDYT? should we override serviceId and hidden field types even if `typesAreFinal` flag is on?

PR #6331, opened 2024-08-01
SALTO-5692: Okta: Upgrade `AppLogo` deploy definitions to new infra

Nurdok:
See title.

---

This PR includes both the tests and the actual upgrade, because filters still do all the work (config change is just a hack to keep the "operation not supported" CV from reporting this type).

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/definitions/deploy/deploy.ts
+++ b/packages/okta-adapter/src/definitions/deploy/deploy.ts
@@ -41,6 +41,7 @@ import {
   ID_FIELD,
   BRAND_THEME_TYPE_NAME,
   APP_GROUP_ASSIGNMENT_TYPE_NAME,
+  APP_LOGO_TYPE_NAME,
 } from '../../constants'
 import {
   APP_POLICIES,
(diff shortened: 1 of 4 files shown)


packages/okta-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -66,6 +67,7 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
   })
 
   const customDefinitions: Record<string, Partial<InstanceDeployApiDefinitions>> = {
+    [APP_LOGO_TYPE_NAME]: {},
```
shir-reifenberg, line 70: Is it needed ?

shir-reifenberg (approved):
🚢 
(please just update "release notes" and "users notifications" to "None", otherwise I think it screws things with release notes script)

packages/okta-adapter/src/definitions/deploy/deploy.ts
Nurdok: Removed.

PR #6332, opened 2024-08-01
SALTO-6378: Fix - combined addition of Application and its Service Principal fails

shoshana-houri:
1. On `application` addition - copy the `appId` from the response.
2. On `appRole` addition - generate internal id directly on the changes.

---
1. The `Service Principal` addition requires its corresponding `appId`, which is hidden on the `Application` instance.
We don't copy it automatically from the response, since it's not a serviceId, so when attempting to add the `Service Principal` right after adding the `Application` we can't resolve the `appId` ref.
When fixing it I found out that the `copyFromResponse` flow works on the original change, which means I had to adjust the `app_role` filter to not create a new `Application` change with the `appRole` field, but rather modify it in-place. 
2. In a similar way - when adding an instance that depends on an `appRole` - we have to get its `id`, which is hidden. Since the `appRole` is deployed as part of its parent we don't get the `copyFromResponse` behavior "for free" on the original changes, so we need to manually add the generated ids to the changes.

---
_Release Notes_: 
_Microsoft Entra_:
* Fix bug in combined addition of Application and its Service Principal.

---
_User Notifications_: 
_Microsoft Entra_:
* It's now possible to deploy addition of Application and its Service Principal in the same deployment.

### DIFF
--- a/packages/microsoft-entra-adapter/src/definitions/deploy/deploy.ts
+++ b/packages/microsoft-entra-adapter/src/definitions/deploy/deploy.ts
@@ -209,6 +209,9 @@ const graphV1CustomDefinitions: DeployCustomDefinitions = {
               toSharedContext: {
                 pick: ['id'],
               },
+              additional: {
+                pick: ['appId'],
+              },
             },
           },
           {
--- a/packages/microsoft-entra-adapter/src/definitions/deploy/deploy.ts
+++ b/packages/microsoft-entra-adapter/src/definitions/deploy/deploy.ts
@@ -210,6 +210,8 @@ const graphV1CustomDefinitions: DeployCustomDefinitions = {
                 pick: ['id'],
               },
               additional: {
+                // The appId is hidden, so it won't exist on addition.
+                // However, it is used to reference the application from other instances, so we should copy it to the applied change.
                 pick: ['appId'],
               },
             },
--- a/packages/microsoft-entra-adapter/src/filters/app_role.ts
+++ b/packages/microsoft-entra-adapter/src/filters/app_role.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 import _ from 'lodash'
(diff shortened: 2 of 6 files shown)


packages/microsoft-entra-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -209,6 +209,9 @@ const graphV1CustomDefinitions: DeployCustomDefinitions = {
               toSharedContext: {
                 pick: ['id'],
               },
+              additional: {
+                pick: ['appId'],
+              },
```
shir-reifenberg, line 214: nit: maybe add a comment explaining why this is needed?

packages/microsoft-entra-adapter/src/filters/app_role.ts
```diff
@@ -215,22 +229,41 @@ const deployAppRoleChangesViaParent = async ({
...
-      appliedParentChangesFullNames.includes(getParents(getChangeData(change))[0]?.elemID.getFullName()),
-    )
+    const appliedExistingParentChanges = await awu(parentChanges)
+      .filter(
+        ({ parentChange, existingParentChangeIdx }) =>
+          existingParentChangeIdx !== undefined &&
+          appliedParentChangesFullNames.includes(getChangeData(parentChange).elemID.getFullName()),
+      )
+      .map(({ parentChange }) =>
+        // We add the app roles field to the parent change in place, but we don't want to copy it to the workspace.
```
shir-reifenberg, line 248: mostly a nit - 
I think it's a bit odd to mention "workspace" in here (cause it's something that the adapter is not aware of)
maybe something like  
```suggestion
        // We add the app roles field to the parent change in place, but we don't include it in the applied change.
```

shir-reifenberg (approved):
💯

PR #6340, opened 2024-08-04
SALTO-6026: Identify more semantic references and commented serviceIds will not be created as generated dependencies

Shulik95:
This PR contains two tickets which are handled together:
* [SALTO-6056](https://salto-io.atlassian.net/browse/SALTO-6056): Deployment Should Not Be Blocked for Missing References in Commented-Out JS Code
* [SALTO-6026](https://salto-io.atlassian.net/jira/software/c/projects/SALTO/issues/SALTO-6026): Identify more semantic references in SuiteScripts

Currently the new references identified for SALTO-6026 are only logged and not added as dependencies.

---

None

---
_Release Notes_: 
Netsuite Adapter:
* Commented out JS code will not be added as a generated dependency in SuiteScripts

---
_User Notifications_: 

[SALTO-6056]: https://salto-io.atlassian.net/browse/SALTO-6056?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SALTO-6026]: https://salto-io.atlassian.net/browse/SALTO-6026?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -160,13 +197,54 @@ const getSuiteScriptReferences = async (
...
 
   const content = fileContent.toString()
-
   const nsConfigReferences = getGroupItemFromRegex(content, nsConfigRegex, OPTIONAL_REFS)
-  const semanticReferences = getGroupItemFromRegex(content, semanticReferenceRegex, OPTIONAL_REFS)
-    .filter(path => !path.startsWith(NETSUITE_MODULE_PREFIX))
-    .concat(nsConfigReferences)
 
-  return getServiceElemIDsFromPaths(semanticReferences, serviceIdToElemID, customRecordFieldsToServiceIds, element)
+  try {
```
dantal4, line 202: the following code ignores references in comments, so this will change existing nacls which we want to prevent.
I suggest to keep the old code, and then run the new code, and log the references that will be added, the references that will be deleted, and wrap it in `log.timeDebug` too.

dantal4 (commented):
Great work!

packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -65,6 +67,10 @@ const pathPrefixRegex = new RegExp(
   `^${FILE_CABINET_PATH_SEPARATOR}|^\\.${FILE_CABINET_PATH_SEPARATOR}|^\\.\\.${FILE_CABINET_PATH_SEPARATOR}`,
   'm',
 )
+// matches key strings in the format of 'key: value'
+const mappedReferenceRegex = new RegExp(`(?<${OPTIONAL_REFS}>\\w+):\\s*.+`, 'gm')
+// matches comments in js files
+const jsCommentsRegex = new RegExp('\\/\\*[\\s\\S]*?\\*\\/|^\\/\\/.*|(?<=[^:])\\/\\/.*', 'gm')
```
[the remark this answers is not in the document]
Shulik95: 1. like you said, `\\s\\S` matches multiline which `.*` doesn't
3. Its a positive lookbehind assertion, used to not match urls that contain `//`. It only matches the main pattern which is `//` in our case if a secondary pattern (here its `^:`) is matched before. this matches `//` only if its not preceded by a colon.

packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -144,13 +153,83 @@ const getServiceElemIDsFromPaths = (
     })
     .filter(isDefined)
 
+const parseAST = (ast: Program): { semanticReferencesArray: string[]; objectKeyReferencesArray: string[] } => {
+  const semanticReferences = new Set<string>()
+  const objectKeyReferences = new Set<string>()
```
dantal4, line 158: nit - it looks like both `parseAST` and `getReferencesWithRegex` can return `string[]`, and here we can have one `Set` to collect the results.

packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -160,12 +239,36 @@ const getSuiteScriptReferences = async (
...
     .concat(nsConfigReferences)
 
+  log.timeDebug(() => {
+    try {
+      const ast = parseScript(content)
+      const { semanticReferencesArray, objectKeyReferencesArray } = parseAST(ast)
+      // TODO: remove once SALTO-6026 is communicated
+      getAndLogReferencesDiff({
+        newReferences: semanticReferencesArray.concat(objectKeyReferencesArray),
+        existingReferences: semanticReferences,
```
dantal4, line 254: since `newReferences` won't include any of the `nsConfigReferences` strings, maybe pass in `existingReferences` the `semanticReferences` before concatenating them with `nsConfigReferences`

packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -160,12 +239,36 @@ const getSuiteScriptReferences = async (
...
+        serviceIdToElemID,
+        customRecordFieldsToServiceIds,
+      })
+    } catch (e) {
+      log.warn('Failed to parse file %s content with error %s', element.value[PATH], e)
+      const { semanticReferences: foundReferences, objectKeyReferences } = getReferencesWithRegex(content)
+      // TODO: remove once SALTO-6026 is communicated
+      getAndLogReferencesDiff({
+        newReferences: foundReferences.concat(objectKeyReferences),
+        existingReferences: semanticReferences,
```
dantal4, line 265: here too

packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -160,12 +239,36 @@ const getSuiteScriptReferences = async (
...
+      // TODO: remove once SALTO-6026 is communicated
+      getAndLogReferencesDiff({
+        newReferences: semanticReferencesArray.concat(objectKeyReferencesArray),
+        existingReferences: semanticReferences,
+        element,
+        serviceIdToElemID,
+        customRecordFieldsToServiceIds,
+      })
+    } catch (e) {
+      log.warn('Failed to parse file %s content with error %s', element.value[PATH], e)
```
dantal4, line 260:
```suggestion
      log.warn('Failed to parse file %s content with error: %o', element.value[PATH], e)
```

dantal4 (approved):
Great work!

PR #6343, opened 2024-08-04
SALTO-6279: (bug-fix) defaultDynamicContentItemVariantValidator - Cannot read properties of undefined (reading 'default')

tamtamirr:
defaultDynamicContentItemVariantValidator - Cannot read properties of undefined (reading 'default')

---

There was edge-case where the adapter threw this run-time error due to invalid resolved reference of a `Variant`.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


edenhassid (approved):
great! 💯

PR #6346, opened 2024-08-05
SALTO-6403 Pagerduty remove direct fetch from recurseInto serviceOrchestration type

Uri-Bar:
SALTO-6403 Pagerduty remove direct fetch from recurseInto serviceOrchestration type
---

---
_Release Notes_: 
none

---
_User Notifications_: 
none

### DIFF
--- a/packages/pagerduty-adapter/src/definitions/fetch/fetch.ts
+++ b/packages/pagerduty-adapter/src/definitions/fetch/fetch.ts
@@ -108,7 +108,7 @@ const createCustomizations = (): Record<string, definitions.fetch.InstanceFetchA
       },
     ],
     resource: {
-      directFetch: true,
+      directFetch: false,
     },
     element: {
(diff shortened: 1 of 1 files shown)


packages/pagerduty-adapter/src/definitions/fetch/fetch.ts
```diff
@@ -108,7 +108,7 @@ const createCustomizations = (): Record<string, definitions.fetch.InstanceFetchA
       },
     ],
     resource: {
-      directFetch: true,
+      directFetch: false,
     },
```
netama, line 112: nit: can just remove this section instead (not have a `resource` entry for this type)

Uri-Bar: I know, but I think explicit is fine for readability. IDK

PR #6349, opened 2024-08-05
SAAS-11750: CPQ Rules and Conditions Filter creates invalid TemplateExpressions when no conditions are fetched for Rule

tamtamirr:
CPQ Rules and Conditions Filter creates invalid TemplateExpressions when no conditions are fetched for Rule

---

When a `TemplateExpression` contains a single part that is string, the env will have non-semantic pending changes. (As we compare `TemplateExpression` with a string)

This is an edge-case that the filter can reach when no conditions were fetched for the `Custom Rule`.

---
_Release Notes_: 
Salesforce Adapter:
- Bugfix: CPQ Rules and Conditions Filter creates invalid TemplateExpressions when no conditions are fetched for Rule.

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/src/filters/cpq/rules_and_conditions_refs.ts
+++ b/packages/salesforce-adapter/src/filters/cpq/rules_and_conditions_refs.ts
@@ -26,6 +26,7 @@ import {
   TemplatePart,
 } from '@salto-io/adapter-api'
 import {
+  compactTemplate,
   createTemplateExpression,
   inspectValue,
   replaceTemplatesWithValues,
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/src/filters/cpq/rules_and_conditions_refs.ts
```diff
@@ -177,21 +178,23 @@ const setCustomConditionReferences = ({
...
-        return part
-      }
-      const condition = conditionsByIndex[index]
-      if (condition === undefined) {
-        log.warn(`Could not find condition with index ${index} for rule ${rule.elemID.getFullName()}`)
-        return part
-      }
-      createdReferences += 1
-      return new ReferenceExpression(condition.elemID, condition)
+  rule.value[def.rule.customConditionField] = compactTemplate(
```
tamtamirr, line 181: Simply wrapped the result of `createTemplateExpression` with `compactTemplate` which returns string when all parts are string.

edenhassid (approved):

PR #6350, opened 2024-08-05
SALTO-6407 Update config type to match the current Guide and Themes implementation.

adfineberg:

---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/zendesk-adapter/src/config_creator.ts
+++ b/packages/zendesk-adapter/src/config_creator.ts
@@ -14,7 +14,14 @@
  * limitations under the License.
  */
 
-import { BuiltinTypes, ConfigCreator, ElemID, InstanceElement } from '@salto-io/adapter-api'
+import {
+  BuiltinTypes,
+  ConfigCreator,
+  CORE_ANNOTATIONS,
+  createRestriction,
+  ElemID,
+  InstanceElement,
+} from '@salto-io/adapter-api'
 import {
   createDefaultInstanceFromType,
   createMatchingObjectType,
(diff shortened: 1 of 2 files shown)


packages/zendesk-adapter/src/config_creator.ts
```diff
@@ -56,13 +81,14 @@ export const getConfig = async (options?: InstanceElement): Promise<InstanceElem
   if (options === undefined || !createOptionsTypeGuard<ConfigOptionsType>(optionsElemId)(options)) {
     return defaultConf
   }
-  if (options.value.enableGuide === true || options.value.enableGuideThemes === true) {
+  const { enableGuide, enableGuideThemes, guideOptions } = options.value
+  if (enableGuide === true || enableGuideThemes === true || (guideOptions && guideOptions !== NO_GUIDE)) {
```
edenhassid, line 85: why do you have `guideOptions` twice?

packages/zendesk-adapter/src/config_creator.ts
```diff
@@ -56,13 +81,14 @@ export const getConfig = async (options?: InstanceElement): Promise<InstanceElem
...
     const configWithGuide = defaultConf.clone()
-    const guideOverride = options.value.enableGuide === true ? { brands: ['.*'] } : {}
```
edenhassid, line 61: why did you remove this whole logic?

edenhassid (commented):
looks good! added 2 small comments

packages/zendesk-adapter/src/config_creator.ts
adfineberg: because it's optional, if there isn't any `guideOptions` I don't want it to get in this clause

packages/zendesk-adapter/src/config_creator.ts
adfineberg: I'm going to remove the booleans in a bit, I'm leaving them now for backwards compatibility. 
Currently, the logic is - if any of the booleans is on, then we want guide.

edenhassid (approved):

PR #6352, opened 2024-08-05
SALTO-6400: Add reference from UserSchema properties sourced by apps to Application type (Okta)

shir-reifenberg:
Add reference from `UserSchema` to `Application`

---

I used context strategy cause the reference should only be created with `type = 'APP'`
This is a noisy change, will followup with noise reduction / prod ticket

---
_Release Notes_: 

_Okta_adapter_:
- Add reference from UserSchema type to `Application` type

---
_User Notifications_: 

_Okta_adapter_:
- `UserSchema` instances with properties sourced by external app will have `Application` reference added

### DIFF
--- a/packages/okta-adapter/src/reference_mapping.ts
+++ b/packages/okta-adapter/src/reference_mapping.ts
@@ -73,7 +73,7 @@ const getProfileMappingRefByType: referenceUtils.ContextValueMapperFunc = val =>
 const getProfileMappingRefByName: referenceUtils.ContextValueMapperFunc = val =>
   val.endsWith('_idp') ? IDENTITY_PROVIDER_TYPE_NAME : undefined
 
-const getUserSchemaPropertyOverrideType: referenceUtils.ContextValueMapperFunc = val => 
+const getUserSchemaPropertyOverrideType: referenceUtils.ContextValueMapperFunc = val =>
   val === 'APP' ? APPLICATION_TYPE_NAME : undefined
 
 export type ReferenceContextStrategyName = 'profileMappingType' | 'profileMappingName' | 'UserSchemaPropertyAppOverride'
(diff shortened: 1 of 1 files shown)


packages/okta-adapter/src/reference_mapping.ts
```diff
@@ -86,6 +89,11 @@ export const contextStrategyLookup: Record<ReferenceContextStrategyName, referen
     getLookUpName: async ({ ref }) => ref.elemID.name,
     contextValueMapper: getProfileMappingRefByName,
   }),
+  UserSchemaPropertyAppOverride: referenceUtils.neighborContextGetter({
+    contextFieldName: 'type',
+    getLookUpName: async ({ ref }) => ref.elemID.name,
+    contextValueMapper: getUserSchemaPropertyOverrideType,
+  }),
```
shoshana-houri, line 96: nit - the other context strategies names are camelCase while this one is PascalCase, is that on purpose?

Also, a suggestion: I think that this can be shared with `profileMappingType` as a more general context strategy, something like - `getMappingRefByType`, and them simply managing a `Record` from possible `type` values to their matching type names, i.e.:
```
VALUE_TO_MATCHING_TYPE: Record<string, string> = {
    user: USERTYPE_TYPE_NAME,
    appuser: APPLICATION_TYPE_NAME,
    APP: APPLICATION_TYPE_NAME,
}

and then -
contextValueMapper: value => VALUE_TO_MATCHING_TYPE[value]

```

kind of similar to what we do [here](https://github.com/salto-io/salto/blob/863f02348ee47a3e24cbd2fa9e719ab9696a456b/packages/microsoft-entra-adapter/src/definitions/references.ts#L230).

What do you think? Feel free to leave it as is if you think that changing it will be confusing :)

shoshana-houri (approved):
I had a small nit/suggestion, anyways approving on my side :)

packages/okta-adapter/src/reference_mapping.ts
shir-reifenberg: the PascalCase is a mistake, thanks !
re suggestion - I wouldn't want to consolidate for the unlikely yet possible scenario when I'll have for example `val = 'user'` but in `UserSchema` type... 
I think it's more accurate to keep it as is to avoid rules mix up

PR #6353, opened 2024-08-06
SALTO-6367: Adjust migrated MFA policies to Identity Engine format in Okta

shir-reifenberg:
Adjust MFA policies on fetch 

---

The change is needed to - 
1. Parse references correctly to authenticators
2. The nacl at its current state is not deployable for Identity Engine accounts

---
_Release Notes_: 

---
_User Notifications_: 


omrifarkash (commented):
I don't fully understand why is this change needed. I find the old code of excluding types from the definitions more readable and yield the same results?

Lets talk f2f :)

omrifarkash (approved):

Nurdok (approved):

PR #6354, opened 2024-08-06
SALTO-6415: Fix empty key for context options

idoamit8:
In this PR, I modified the empty key in the context option to be saved as @ instead of an empty string.

---

---
_Release Notes_: 
_Jira Adapter_:
* Fixed a bug that caused duplicate attributes due to an empty string in the context option's key.

---
_User Notifications_: 
_Jira Adapter_:
* The empty string inside the options field in `CustomFieldContext` will now be replaced by '@'.


netama (approved):
LGTM, but might be best to pick another (guaranteed-unique) string to be on the safe side

PR #6355, opened 2024-08-06
SALTO-6342: fix update nacl files to include static file changes continue

edenhassid:
fix update nacl files to include static file changes

---

---
_Release Notes_: 
core:
* fix bug that when the only changes in fetch are modifications to static resources, the fetch returns empty

---
_User Notifications_: 

### DIFF
--- a/packages/core/src/local-workspace/dir_store.ts
+++ b/packages/core/src/local-workspace/dir_store.ts
@@ -50,6 +50,7 @@ const buildLocalDirectoryStore = <T extends dirStore.ContentType>(
   let currentBaseDir = path.join(..._.compact([baseDir, storeName, nameSuffix]))
   let updated: FileMap<T> = initUpdated || {}
   let deleted: Set<string> = new Set(initDeleted || [])
+  let isEmptyCache: boolean | undefined
 
   const getAbsFileName = (filename: string, dir?: string): string => path.resolve(dir ?? currentBaseDir, filename)
 
@@ -160,7 +161,13 @@ const buildLocalDirectoryStore = <T extends dirStore.ContentType>(
       .uniq()
       .value()
 
-  const isEmpty = async (): Promise<boolean> => (await list()).length === 0
+  const isEmpty = async (): Promise<boolean> => {
+    if (isEmptyCache !== undefined) {
+      return isEmptyCache
+    }
+    isEmptyCache = (await list()).length === 0
+    return isEmptyCache
+  }
 
   const flush = async (): Promise<void> => {
     const deletesToHandle = deleted
@@ -219,6 +226,7 @@ const buildLocalDirectoryStore = <T extends dirStore.ContentType>(
       file.timestamp = Date.now()
       updated[relFilename] = file
       deleted.delete(relFilename)
+      isEmptyCache = undefined
       return Promise.resolve()
     },
 
@@ -230,6 +238,7 @@ const buildLocalDirectoryStore = <T extends dirStore.ContentType>(
         return Promise.reject(err)
       }
       deleted.add(relFilename)
+      isEmptyCache = undefined
       return Promise.resolve()
     },
 
--- a/packages/core/src/local-workspace/dir_store.ts
+++ b/packages/core/src/local-workspace/dir_store.ts
@@ -165,7 +165,7 @@ const buildLocalDirectoryStore = <T extends dirStore.ContentType>(
     if (cachedIsEmpty !== undefined) {
       return cachedIsEmpty
     }
-    cachedIsEmpty = Object.keys(updated).length > 0 ? false : (await list()).length === 0
+    cachedIsEmpty = (await list()).length === 0
     return cachedIsEmpty
   }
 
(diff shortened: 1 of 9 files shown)


packages/core/src/local-workspace/dir_store.ts
```diff
@@ -230,6 +238,7 @@ const buildLocalDirectoryStore = <T extends dirStore.ContentType>(
         return Promise.reject(err)
       }
       deleted.add(relFilename)
+      isEmptyCache = undefined
```
ori-moisis, line 241: the `clear` function should also unset this (or maybe set it to `true`?)

packages/core/src/local-workspace/dir_store.ts
```diff
@@ -50,6 +50,7 @@ const buildLocalDirectoryStore = <T extends dirStore.ContentType>(
   let currentBaseDir = path.join(..._.compact([baseDir, storeName, nameSuffix]))
   let updated: FileMap<T> = initUpdated || {}
   let deleted: Set<string> = new Set(initDeleted || [])
+  let isEmptyCache: boolean | undefined
```
ori-moisis, line 53: this name reads as if there is a cache and this says if that cache is empty

packages/core/src/local-workspace/dir_store.ts
```diff
@@ -219,6 +226,7 @@ const buildLocalDirectoryStore = <T extends dirStore.ContentType>(
       file.timestamp = Date.now()
       updated[relFilename] = file
       deleted.delete(relFilename)
+      isEmptyCache = undefined
```
adfineberg, line 229: why not set it here to be false? if a file is set, it means that the store isn't empty, right?

packages/core/src/local-workspace/dir_store.ts
```diff
@@ -160,7 +161,13 @@ const buildLocalDirectoryStore = <T extends dirStore.ContentType>(
       .uniq()
       .value()
 
-  const isEmpty = async (): Promise<boolean> => (await list()).length === 0
+  const isEmpty = async (): Promise<boolean> => {
+    if (isEmptyCache !== undefined) {
+      return isEmptyCache
+    }
+    isEmptyCache = (await list()).length === 0
```
adfineberg, line 168: why not add the optimization to look also at the updated if it exists?

packages/core/src/local-workspace/dir_store.ts
edenhassid: what do you think about `isemptyInfo` of just `isEmpty`?

packages/core/src/local-workspace/dir_store.ts
edenhassid: because Idon't know for sure if a file was set, it can also be only deletions

packages/core/src/local-workspace/dir_store.ts
ori-moisis: maybe just reversing the order - `cachedIsEmpty`?

packages/core/src/local-workspace/dir_store.ts
```diff
@@ -160,7 +161,13 @@ const buildLocalDirectoryStore = <T extends dirStore.ContentType>(
       .uniq()
       .value()
 
-  const isEmpty = async (): Promise<boolean> => (await list()).length === 0
+  const isEmpty = async (): Promise<boolean> => {
+    if (isEmptyInfo !== undefined) {
+      return isEmptyInfo
+    }
+    isEmptyInfo = Object.keys(updated).length > 0 ? false : (await list()).length === 0
```
ori-moisis: not sure I like relying on `updated` without considering `deleted` here
doing so means we need to make sure there are no conflicts between `updated` and `deleted`.
for example, with the current code AFAICT if something was updated and then deleted it will still be in the updated map (see the implementation of `delete`)

I know this was added due to the comment from @adfineberg but I am not sure the added complexity is justified by the (mostly theoretical) performance improvement here

edenhassid: ok so I think I'll just remove it

packages/core/src/local-workspace/dir_store.ts
```diff
@@ -160,7 +161,13 @@ const buildLocalDirectoryStore = <T extends dirStore.ContentType>(
       .uniq()
       .value()
 
-  const isEmpty = async (): Promise<boolean> => (await list()).length === 0
+  const isEmpty = async (): Promise<boolean> => {
+    if (cachedIsEmpty !== undefined) {
+      return cachedIsEmpty
+    }
+    cachedIsEmpty = Object.keys(updated).length > 0 ? false : (await list()).length === 0
+    return cachedIsEmpty
```
ori-moisis: very much a nit
```suggestion
    if (cachedIsEmpty === undefined) {
      cachedIsEmpty = // fixed code here, however you choose to fix it
    }
    return cachedIsEmpty
```

ori-moisis (approved):

packages/core/src/local-workspace/dir_store.ts
ori-moisis: if we do want a performance improvement we could stop relying on list, or making list an iterator and just getting the first entry from it when we check `isEmpty`

PR #6356, opened 2024-08-06
SALTO-6417: Change question type in forms

idoamit8:
We recently changed the API to fetch forms and the question field changes from ListType to mapType which resulted in not adding references. In this PR I fixed it. 

---

---
_Release Notes_: 
_Jira Adapter:_
* Inside `Form` type, the `jiraField` will change from string to reference. 

---
_User Notifications_: 
_Jira Adapter:_
* Inside `Form` type, the `jiraField` will change from string to reference.

### DIFF
--- a/packages/jira-adapter/src/filters/forms/forms_types.ts
+++ b/packages/jira-adapter/src/filters/forms/forms_types.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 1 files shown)


shir-reifenberg (approved):
💯 💯 💯 
Please update release notes + user notifications before merging.
This change is noisy, so leaving @omrilit @IdoZakk to decide wether worth adding a noise reduction / prod ticket

IdoZakk:
Yes, we should add noise reduction

idoamit8:
@IdoZakk Already added

PR #6360, opened 2024-08-06
SALTO-6418: support more than one resolution in transition properties

shayc331:
_support more than one resolution in transition properties_

---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/jira-adapter/src/filters/workflowV2/workflow_filter.ts
+++ b/packages/jira-adapter/src/filters/workflowV2/workflow_filter.ts
@@ -78,6 +78,7 @@ import {
   PayloadWorkflowStatus,
   EMPTY_STRINGS_PATH_NAME_TO_RECURSE,
   TRANSITION_LIST_FIELDS,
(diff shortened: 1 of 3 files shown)


packages/jira-adapter/src/filters/workflowV2/workflow_filter.ts
```diff
@@ -141,6 +143,26 @@ const fetchWorkflowData = async (paginator: clientUtils.Paginator): Promise<Work
 
 const convertIdsStringToList = (ids: string): string[] => ids.split(',')
 
+const splitResolutionProperties = (transitions: WorkflowV2Transition[]): void => {
+  transitions.forEach(transition => {
+    if (transition.properties !== undefined) {
```
kami-delaroz, line 148: nit: I think _filter_ followed by _forEach_ is more elegant then an _if_ condition within a _forEach_, but your choice

kami-delaroz (approved):
🥇

PR #6362, opened 2024-08-07
SALTO-6422: add AQL placeholder change validator

shayc331:
_add AQL placeholder change validator_

---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/jira-adapter/src/change_validators/field_contexts/assets_object_field_configuration_aql.ts
+++ b/packages/jira-adapter/src/change_validators/field_contexts/assets_object_field_configuration_aql.ts
@@ -17,7 +17,7 @@
 import { ChangeValidator, getChangeData, isAdditionOrModificationChange, isInstanceChange } from '@salto-io/adapter-api'
 import { FIELD_CONTEXT_TYPE_NAME } from '../../filters/fields/constants'
 
-const PLACEHOLDER_PATTERN = /\$\{(.*?)\}/g
+const PLACEHOLDER_PATTERN = /\$\{(.*)\}/
 
 const isAqlHasPlaceholder = (aql: string): boolean => {
   const matches = aql.match(PLACEHOLDER_PATTERN)
@@ -30,11 +30,7 @@ export const assetsObjectFieldConfigurationAqlValidator: ChangeValidator = async
     .filter(isAdditionOrModificationChange)
     .map(getChangeData)
     .filter(instance => instance.elemID.typeName === FIELD_CONTEXT_TYPE_NAME)
-    .filter(
-      instance =>
-        instance.value.assetsObjectFieldConfiguration !== undefined &&
-        instance.value.assetsObjectFieldConfiguration.issueScopeFilterQuery !== undefined,
-    )
+    .filter(instance => instance.value.assetsObjectFieldConfiguration?.issueScopeFilterQuery !== undefined)
     .filter(instance => isAqlHasPlaceholder(instance.value.assetsObjectFieldConfiguration.issueScopeFilterQuery))
     .map(instance => ({
       elemID: instance.elemID.createNestedID('assetsObjectFieldConfiguration', 'issueScopeFilterQuery'),
--- a/packages/jira-adapter/src/change_validators/field_contexts/assets_object_field_configuration_aql.ts
+++ b/packages/jira-adapter/src/change_validators/field_contexts/assets_object_field_configuration_aql.ts
@@ -1,0 +1,55 @@
+/*
+ *                      Copyright 2024 Salto Labs Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


packages/jira-adapter/src/change_validators/field_contexts/assets_object_field_configuration_aql.ts
```diff
@@ -0,0 +1,55 @@
...
+export const assetsObjectFieldConfigurationAqlValidator: ChangeValidator = async changes =>
+  changes
+    .filter(isInstanceChange)
+    .filter(isAdditionOrModificationChange)
+    .map(getChangeData)
+    .filter(instance => instance.elemID.typeName === FIELD_CONTEXT_TYPE_NAME)
+    .filter(
+      instance =>
+        instance.value.assetsObjectFieldConfiguration !== undefined &&
+        instance.value.assetsObjectFieldConfiguration.issueScopeFilterQuery !== undefined,
```
IdoZakk, line 36: You can do
instance.value.assetsObjectFieldConfiguration?.issueScopeFilterQuery !== undefined
Instead of those two lines

packages/jira-adapter/src/change_validators/field_contexts/assets_object_field_configuration_aql.ts
```diff
@@ -0,0 +1,55 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ChangeValidator, getChangeData, isAdditionOrModificationChange, isInstanceChange } from '@salto-io/adapter-api'
+import { FIELD_CONTEXT_TYPE_NAME } from '../../filters/fields/constants'
+
+const PLACEHOLDER_PATTERN = /\$\{(.*?)\}/g
```
IdoZakk, line 20: I think you can simplify:
* instead of .*? you can do .*. The difference is what inside the brackets if there are several matches, it does not really matter here
* you can remove the /g- it does not matter to us how many matches there are

packages/jira-adapter/src/change_validators/field_contexts/assets_object_field_configuration_aql.ts
```diff
@@ -0,0 +1,55 @@
...
+        'This AQL expression will be deployed as is. You may need to manually edit the ids later to match the target environment.',
+      deployActions: {
+        postAction: {
+          title: 'Edit AQL placeholders manually',
+          subActions: [
+            `In Jira, navigate to the field context "${instance.value.name}" > Edit Assets object/s field configuration`,
+            'Inside Filter issue scope section, fix the placeholder with the correct value',
+            'Click "Save"',
+          ],
+        },
```
IdoZakk, line 53: We can give the link in most cases, look at `workflow_scheme_migration` for example

packages/jira-adapter/src/change_validators/field_contexts/assets_object_field_configuration_aql.ts
```diff
@@ -0,0 +1,55 @@
...
+    .filter(isInstanceChange)
+    .filter(isAdditionOrModificationChange)
+    .map(getChangeData)
+    .filter(instance => instance.elemID.typeName === FIELD_CONTEXT_TYPE_NAME)
+    .filter(
+      instance =>
+        instance.value.assetsObjectFieldConfiguration !== undefined &&
+        instance.value.assetsObjectFieldConfiguration.issueScopeFilterQuery !== undefined,
+    )
+    .filter(instance => isAqlHasPlaceholder(instance.value.assetsObjectFieldConfiguration.issueScopeFilterQuery))
```
IdoZakk, line 38: I think that if the user will edit the NaCl and enter a non string value (number, array) in the issueScopeFilterQuery the code will crash. It is very unlikely, but to avoid you just need to change the filter before to make sure it is a string

IdoZakk (approved):
👑 
Please populate the release notes, and see some minor comments

packages/jira-adapter/src/change_validators/field_contexts/assets_object_field_configuration_aql.ts
shayc331: the problem is that this `assetObjectFieldConfiguration` service url is different than the context service url. It includes `atlToken` which we don't save

packages/jira-adapter/src/change_validators/field_contexts/assets_object_field_configuration_aql.ts
shayc331: 1- I dont understand your suggestion here (I did `.*?` and you suggested `..`)
2- agree

packages/jira-adapter/src/change_validators/field_contexts/assets_object_field_configuration_aql.ts
IdoZakk: I think giving them a link to the context and telling them to press the edit configuration... will be good

packages/jira-adapter/src/change_validators/field_contexts/assets_object_field_configuration_aql.ts
IdoZakk: Regarding 1 it was a typo, I meant `.*` (just remove the ?)

PR #6363, opened 2024-08-07
SALTO-6377 support blogpost version

omrifarkash:
On this PR:
- Change `get_users_and_groups` to `get_users` since [here](https://github.com/salto-io/salto/pull/6312) we added group as a valid type to fetch, so no need to re-fetch it for `group_and_user` filter
- Support `authorId` user display name on `blog_post` like we are doing on `space` and `page` types
- Support same `version` increasing logic on deployment of `blog_post` as we are doing with `page` type

---

---

_Release Notes_: 
_Confluence Adapter_:
- Support blog_post modification deployment

---

_User Notifications_: 
_Confluence Adapter_:
- Support blog_post modification deployment


packages/confluence-adapter/src/definitions/utils/blog_post.ts
```diff
@@ -0,0 +1,41 @@
...
+import { increaseVersion } from './version'
+
+const { reduceAsync } = collections.asynciterable
+
+export const adjustUserReferencesOnBlogPostReverse = createAdjustUserReferencesReverse(BLOG_POST_TYPE_NAME)
+
+/**
+ * AdjustFunction that runs all blog_post modification adjust functions.
+ */
+export const adjustBlogPostOnModification: definitions.AdjustFunctionSingle<
```
shir-reifenberg, line 31: We already have almost identical logic to `adjustPageOnModification`
can we consolidate ? (for example by creating a function that will get the list of adjust functions then return adjust function that runs all of them ?)

Also I think this lofic will be definitely useful in other adapters as well so consider moving it to adapter components (but really not mandatory 🙃 )

packages/confluence-adapter/src/definitions/utils/blog_post.ts
```diff
@@ -0,0 +1,41 @@
...
+  definitions.deploy.ChangeAndContext
+> = async args => {
+  const value = validateValue(args.value)
+  const argsWithValidatedValue = { ...args, value }
+  return reduceAsync(
+    [increaseVersion, adjustUserReferencesOnBlogPostReverse],
+    async (input, func) => ({ ...argsWithValidatedValue, ...(await func(input)) }),
+    argsWithValidatedValue,
+  )
```
shir-reifenberg, line 40: I find `reduce` a little difficult to read
WDYT about something like - 
```suggestion
adjustFuncs = [increaseVersion, adjustUserReferencesOnBlogPostReverse]
return args => 
   awu(adjustFuncs)
      .map(adjust => adjust(args)

File: packages/confluence-adapter/src/definitions/utils/version.ts [hunk 3]
```diff
@@ -0,0 +1,51 @@
...
+
+import { definitions } from '@salto-io/adapter-components'
+import { values } from '@salto-io/lowerdash'
+import _ from 'lodash'
+import { validateValue } from './generic'
+
+const isNumber = (value: unknown): value is number => typeof value === 'number'
+
+/**
+ * AdjustFunction that increases the version number of a page for deploy modification change.
```
shir-reifenberg, line 25: maybe adjust comment if it's not only in use for `page` ?

packages/confluence-adapter/src/definitions/utils/version.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+export const increaseVersion: definitions.AdjustFunctionSingle<definitions.deploy.ChangeAndContext> = async args => {
+  const value = validateValue(args.value)
+  const version = _.get(value, 'version')
+  if (!values.isPlainRecord(version) || !isNumber(version.number)) {
```
shir-reifenberg, line 30: nit

```suggestion
  if (!values.isPlainRecord(version) || !_.isNumber(version.number)) {
```

packages/confluence-adapter/src/definitions/utils/version.ts
```diff
@@ -0,0 +1,51 @@
...
+    return {
+      value: {
+        ...value,
+        version: {
+          // In case of homepage addition, we don't have a version number yet but it is "1" in the service
+          // It has been set to one when we created the space and the default homepage was created
+          number: 2,
```
shir-reifenberg, line 37: ignore me if not relevant but do we want to verify the object type before setting `number: 2` ? or does this relevant to all types ?

packages/confluence-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -193,6 +196,60 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
...
+                  path: '/wiki/api/v2/blogposts/{id}',
+                  method: 'put',
+                },
+                transformation: {
+                  adjust: adjustBlogPostOnModification,
+                },
+              },
+              copyFromResponse: {
+                additional: {
+                  pick: ['version.number'],
```
shir-reifenberg, line 235: nit: consider moving to const and share with page definition

packages/confluence-adapter/src/get_users.ts
shir-reifenberg, line 1: file name suggestion - maybe just `user.ts` to align with other adapters ?

packages/confluence-adapter/src/filters/groups_and_users_filter.ts
```diff
@@ -58,7 +69,7 @@ const filter: filterUtils.AdapterFilterCreator<UserConfig, filterUtils.FilterRes
             permission.displayName =
               (type === 'user'
                 ? usersIndex[permission.principalId]?.displayName
-                : groupsIndex[permission.principalId]?.name) ?? FALLBACK_DISPLAY_NAME
+                : groupsIndex[permission.principalId]) ?? FALLBACK_DISPLAY_NAME
```
shir-reifenberg, line 72: not sure I understand this part - If we get groups as instances, why replacing values instead of creating references ?

shir-reifenberg (approved):
Looks great! Added mainly small suggestions / nit
and one question ❓ 
but approving to unblock

packages/confluence-adapter/src/definitions/utils/blog_post.ts
omrifarkash: I don't mind if you prefer

packages/confluence-adapter/src/definitions/utils/version.ts
omrifarkash: Good point, I'll think of the edge cases

packages/confluence-adapter/src/filters/groups_and_users_filter.ts
omrifarkash: We do create references via reference_mapping. This is just for adding a display name to the object

PR #6364, opened 2024-08-07
SALTO-6423: statusMapping change validator should relies on elemID instead of name

shayc331:
_statusMapping change validator should relies on elemID instead of name_

---

---
_Release Notes_: 
Jira Adapter: 
* fix a bug in statusMapping change validator

---
_User Notifications_: 


IdoZakk (approved):
🦁 🦂 
Please update the release notes

PR #6367, opened 2024-08-08
SALTO-6228: (Bug-fix) Salesforce Profiles elements fixer results in wrong removals

tamtamirr:
(Bug-fix) Salesforce Profiles elements fixer results in wrong removals.

---

The `fieldPermissions` section of Profiles may contain values that do not refer to an actual Field. To avoid this, I've changed the implementation of the `Profiles fixer` to ignore non-custom fieldPermissions values in the Profile.

For reference, running the current fixer on my Profile instance resulted in the following changes (Attached is a `restore dry-run` post the fix to demonstrate the difference):
Screenshot 2024-08-08 at 17 20 49

The fixer shouldn't have modified anything in my Profile considering everything is expected to exist in the env 🙈 

---
_Release Notes_: 
_Salesforce Adapter_:
- Fixed an issue where the `profiles` Elements fixer resulted in wrong removals from the Profile.

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/custom_references/profiles.ts
```diff
@@ -252,6 +252,9 @@ const profileEntriesTargets = (profile: InstanceElement): Dictionary<ElemID> =>
     .fromPairs()
     .value()
 
+const isStandardFieldPermissionsPath = (path: string): boolean =>
+  path.startsWith('fieldPermissions') && !ENDS_WITH_CUSTOM_SUFFIX_REGEX.test(path)
```
adamro, line 256: nit: use the FIELD_PERMISSIONS constant?

adamro (approved):
Considering this is the decided logic the implementation LGTM :)

PR #6370, opened 2024-08-08
SALTO-6442: add dependency changer to allow deployment of MFA policy with authenticator

shir-reifenberg:
Add dependency changes to make sure MFA policies are deployed after their authenticators are activated

---

Each MFA policy has a list used authenticators. Authenticator must be activated before it can be assigned to MFA policy. 
The dependency changer is fixing a bug caused when trying to activate authenticator together with adding it to the MFA policy - in this case we must activate the authenticator before deploying the MFA policy changes.

In the next PR I'll add a validator to verify no inactive authenticator is added to mfa policy.

---
_Release Notes_: 

_Okta_adapter_:
- Fix a bug in MultifactorEnrollmentPolicy when deployed with Authenticator

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/dependency_changers/authenticator_to_mfa_policy.ts
+++ b/packages/okta-adapter/src/dependency_changers/authenticator_to_mfa_policy.ts
@@ -1,0 +1,82 @@
+/*
+ *                      Copyright 2024 Salto Labs Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


IdoZakk:
@shir-reifenberg I think you got the ticket number wrong, can you please update it?

packages/okta-adapter/src/dependency_changers/authenticator_to_mfa_policy.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  const authenticatorChangesById = _.keyBy(authenticatorChanges, change =>
+    getChangeData(change.change).elemID.getFullName(),
+  )
+
+  return mfaChanges
+    .flatMap(mfaChange => {
+      const usedAuthenticatorChanges = getAuthenticatorsFromMfaPolicy(getChangeData(mfaChange.change))
+        .map(authenticator => authenticator.key)
+        .map(reference => reference.elemID.getFullName())
```
idoamit8, line 69: Small nit: You can save `authenticatorChangesByIdSet` instead of the dictionary and filter to check if the authenticator is present before resolving the reference value. This might improve readability.

idoamit8 (approved):
Looks great! 💯 
Added one non-mandatory small nit 🙃

PR #6371, opened 2024-08-11
SALTO-6442: Validate all authenticators in MFA are enabled

shir-reifenberg:
Add change validator to block deployments of `MultifactorEnrollmentPolicy` with disabled Authenticators 

---

This rebased on top of https://github.com/salto-io/salto/pull/6370
**Please only review the second commits**  

---
_Release Notes_: 

_Okta_Adapter_:
- Add validation to prevent deployment of MultifactorEnrollmentPolicy with disabled Authenticator

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/change_validators/disabled_authenticators_in_mfa.ts
+++ b/packages/okta-adapter/src/change_validators/disabled_authenticators_in_mfa.ts
@@ -41,8 +41,8 @@ export const disabledAuthenticatorsInMfaPolicyValidator: ChangeValidator = async
   const mfaInstances = changes
     .filter(isInstanceChange)
     .filter(isAdditionOrModificationChange)
-    .filter(change => getChangeData(change).elemID.typeName === MFA_POLICY_TYPE_NAME)
     .map(getChangeData)
+    .filter(({ elemID }) => elemID.typeName === MFA_POLICY_TYPE_NAME)
 
   const disabledAuthenticatorsByMfaElemID: Record<string, InstanceElement[]> = Object.fromEntries(
     await awu(mfaInstances)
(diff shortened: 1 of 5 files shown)


packages/okta-adapter/src/change_validators/disabled_authenticators_in_mfa.ts
```diff
@@ -0,0 +1,73 @@
...
+    return []
+  }
+  const mfaInstances = changes
+    .filter(isInstanceChange)
+    .filter(isAdditionOrModificationChange)
+    .filter(change => getChangeData(change).elemID.typeName === MFA_POLICY_TYPE_NAME)
+    .map(getChangeData)
+
+  const disabledAuthenticatorsByMfaElemID: Record<string, InstanceElement[]> = Object.fromEntries(
+    await awu(mfaInstances)
```
idoamit8, line 48: `awu` processes each item in the iterable one by one, which can be useful when dealing with streams of data or when you need to process items sequentially rather than in parallel
`Promise.all` is used to run multiple promises in parallel.
Why do we use awu here and not promise.all?

packages/okta-adapter/src/change_validators/disabled_authenticators_in_mfa.ts
```diff
@@ -0,0 +1,73 @@
...
+      .map(async policy => {
+        const disabledAuthenticators = await awu(getAuthenticatorsFromMfaPolicy(policy))
```
idoamit8, line 50: Readability nit: at this point this list holds both active and inactive Authenticators (You are filter it in line 56) so I guess a better name would be `authenticators`

idoamit8 (commented):
Awesome work! 💯 🦾
Added two nits and one question 🙃

idoamit8 (approved):
Great!! 🥇

PR #6372, opened 2024-08-11
SALTO-6447: (Jira) Fix custom fields deletion

SeggevHaimovich:
The upgrade of Axios created a problem with deploying a deletion of a custom field.
The problem is happening because of a change in the way it handle redirection (it is no longer automatically pass the auth headers).

---

_None_

---
_Release Notes_: 
Jira adapter:
* Fix custom fields deletion

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/jira-adapter/src/filters/fields/field_deployment_filter.ts
+++ b/packages/jira-adapter/src/filters/fields/field_deployment_filter.ts
@@ -25,6 +25,7 @@ import {
 import { client as clientUtils } from '@salto-io/adapter-components'
 import _ from 'lodash'
 import { collections } from '@salto-io/lowerdash'
+import { logger } from '@salto-io/logging'
 import JiraClient from '../../client/client'
 import { FilterCreator } from '../../filter'
 import { deployContextChange, getContexts, getContextType } from './contexts'
@@ -32,6 +33,7 @@ import { defaultDeployChange, deployChanges } from '../../deployment/standard_de
 import { FIELD_TYPE_NAME } from './constants'
 import { JiraConfig } from '../../config/config'
 
+const log = logger(module)
 const { awu } = collections.asynciterable
 
 const REDIRECT_PATH_PREFIX = '/rest/api/3/task/'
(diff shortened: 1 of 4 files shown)


packages/jira-adapter/src/filters/fields/field_deployment_filter.ts
```diff
@@ -60,10 +75,7 @@ const filter: FilterCreator = ({ client, config }) => ({
   deploy: async changes => {
     const [relevantChanges, leftoverChanges] = _.partition(
       changes,
-      change =>
-        isInstanceChange(change) &&
-        isAdditionOrModificationChange(change) &&
-        getChangeData(change).elemID.typeName === FIELD_TYPE_NAME,
+      change => isInstanceChange(change) && getChangeData(change).elemID.typeName === FIELD_TYPE_NAME,
```
IdoZakk, line 78: Make sure there is no filter to handle fields deletion

packages/jira-adapter/src/filters/fields/field_deployment_filter.ts
```diff
@@ -33,13 +34,27 @@ import { JiraConfig } from '../../config/config'
...
+      change,
+      client,
+      apiDefinitions: config.apiDefinitions,
+      fieldsToIgnore: ['contexts'],
+    })
+  } catch (err) {
+    if (
+      isRemovalChange(change) &&
+      err instanceof clientUtils.HTTPError &&
+      err.response.status === 401 &&
```
IdoZakk, line 51: We want to ignore for any error

packages/jira-adapter/src/filters/fields/field_deployment_filter.ts
```diff
@@ -33,13 +34,27 @@ import { JiraConfig } from '../../config/config'
...
+      err.response.requestPath?.startsWith(REDIRECT_PATH_PREFIX)
+    ) {
+      return
```
IdoZakk, line 54: Please do not return here.
Also please log this in warning level, with relevant info like the redirect path

IdoZakk (approved):
👍 
Please look at my comments

PR #6374, opened 2024-08-11
SALTO-6453: Add missing standalone definition for appRoleAssignments

shoshana-houri:
Fix definitions for servicePrincipal - appRoleAssignments:
1. Define as standalone
2. Add reference definition

---
The `appRoleAssignments` defined on the Service Principal is supposed to be a standalone element (`isTopLevel: true` in the type definition) but this definition is missing on the paren't `fieldCustomization` field.
When reproducing this issue I also noticed that we're missing its reference definition as well.

---
_Release Notes_: 
_Microsoft Entra:_
* Fix issue with appRoleAssignments definition - set them as `standalone`

---
_User Notifications_: 
_Microsoft Entra:_
* Service Principal's `appRoleAssignments` instances will be added with references to the `appRole` + `resource`.

### DIFF
--- a/packages/microsoft-entra-adapter/src/definitions/fetch/fetch.ts
+++ b/packages/microsoft-entra-adapter/src/definitions/fetch/fetch.ts
@@ -347,6 +347,13 @@ const graphV1Customizations: FetchCustomizations = {
           },
         },
         [APP_ROLES_FIELD_NAME]: APP_ROLES_FIELD_CUSTOMIZATIONS,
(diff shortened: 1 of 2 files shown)


shir-reifenberg (approved):
💯 💯 💯 
user notifications nit - the purpose of this section is to state any expected changes to the environment. so I think you can you can write something like - `appRoleAssignments` instances will be added with a reference to their parent servicePrincipal

PR #6377, opened 2024-08-11
SALTO-5838: Deploy SLA addition

idoamit8:
In this PR added support of deployment for default SLA addition

---

when a user create a service desk project, automatically two SLAs are being created. When we try to deploy those SLA with salto with as addition, we are getting name collision error. We solved it by changing the addition change to modification change. 

---
_Release Notes_: 
_Jira Adapter:_
* fixed a bug where users can not deploy addition of default SLA. 

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/change_validators/index.ts
+++ b/packages/jira-adapter/src/change_validators/index.ts
@@ -73,7 +73,7 @@ import { jsmPermissionsValidator } from './jsm/jsm_permissions'
 import { referencedWorkflowDeletionChangeValidator } from './workflowsV2/referenced_workflow_deletion'
 import { missingExtensionsTransitionRulesChangeValidator } from './workflowsV2/missing_extensions_transition_rules'
 import { fieldContextOptionsValidator } from './field_contexts/field_context_options'
-import { ISSUE_TYPE_NAME, PORTAL_GROUP_TYPE, PROJECT_TYPE, SLA_TYPE_NAME } from '../constants'
+import { ISSUE_TYPE_NAME, PORTAL_GROUP_TYPE, PROJECT_TYPE } from '../constants'
 import { assetsObjectFieldConfigurationAqlValidator } from './field_contexts/assets_object_field_configuration_aql'
 
 const { deployTypesNotSupportedValidator, createChangeValidator, uniqueFieldsChangeValidatorCreator } =
@@ -82,7 +82,7 @@ const { deployTypesNotSupportedValidator, createChangeValidator, uniqueFieldsCha
 const TYPE_TO_UNIQUE_FIELD = {
   [ISSUE_TYPE_NAME]: ['name'],
   [PORTAL_GROUP_TYPE]: ['name'],
-  [SLA_TYPE_NAME]: ['name'],
+  // [SLA_TYPE_NAME]: ['name'], # TODO: uncomment when we add Scope to unique fields filter. SALTO-6424
   [PROJECT_TYPE]: ['name', 'key'],
 }
 
--- a/packages/jira-adapter/src/change_validators/index.ts
+++ b/packages/jira-adapter/src/change_validators/index.ts
@@ -82,7 +82,8 @@ const { deployTypesNotSupportedValidator, createChangeValidator, uniqueFieldsCha
 const TYPE_TO_UNIQUE_FIELD = {
   [ISSUE_TYPE_NAME]: ['name'],
   [PORTAL_GROUP_TYPE]: ['name'],
-  // [SLA_TYPE_NAME]: ['name'], # TODO: uncomment when we add Scope to unique fields filter. SALTO-6424
+  // TODO SALTO-6424 uncomment when we add Scope to unique fields filter.
+  // [SLA_TYPE_NAME]: ['name'],
   [PROJECT_TYPE]: ['name', 'key'],
 }
 
--- a/packages/jira-adapter/src/filters/sla_addition_deployment.ts
+++ b/packages/jira-adapter/src/filters/sla_addition_deployment.ts
@@ -18,6 +18,7 @@ import {
   AdditionChange,
   Change,
   InstanceElement,
+  ModificationChange,
   getChangeData,
   isAdditionChange,
   isInstanceChange,
@@ -26,7 +27,7 @@ import {
 import { createSchemeGuard, getParent, hasValidParent } from '@salto-io/adapter-utils'
 import _ from 'lodash'
 import { logger } from '@salto-io/logging'
-import { elements as elementUtils, config as configDeprecated } from '@salto-io/adapter-components'
+import { elements as elementUtils } from '@salto-io/adapter-components'
 import Joi from 'joi'
 import { defaultDeployChange, deployChanges } from '../deployment/standard_deployment'
 import { FilterCreator } from '../filter'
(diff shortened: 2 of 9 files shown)


packages/jira-adapter/src/filters/sla_addition_deployment.ts
```diff
@@ -0,0 +1,177 @@
...
+  parent: InstanceElement,
+  client: JiraClient,
+): Promise<(string | number)[][]> => {
+  try {
+    const response = await client.get({
+      url: `/rest/servicedesk/1/servicedesk/agent/${parent.value.key}/sla/metrics`,
+    })
+    if (!isSlaResponse(response.data)) {
+      return []
+    }
```
shir-reifenberg, line 72: it means that in case of un error we'll override all SLAs, right ? don't we want to throw in this scenario ?

packages/jira-adapter/src/filters/sla_addition_deployment.ts
```diff
@@ -0,0 +1,177 @@
...
+})
+  .unknown(true)
+  .required()
+
+const isSlaResponse = createSchemeGuard<SlaGetResponse>(SLA_RESPONSE_SCHEME)
+
+const getExistingSlaNamesAndIds = async (
+  parent: InstanceElement,
+  client: JiraClient,
+): Promise<(string | number)[][]> => {
```
shir-reifenberg, line 65: please return types objects - 
`Promise<{ id: string; name: string }[]>

packages/jira-adapter/src/filters/sla_addition_deployment.ts
```diff
@@ -0,0 +1,177 @@
...
+  try {
+    const response = await client.get({
+      url: `/rest/servicedesk/1/servicedesk/agent/${parent.value.key}/sla/metrics`,
+    })
+    if (!isSlaResponse(response.data)) {
+      return []
+    }
+    const existingSlas = response.data.timeMetrics.map(sla => [sla.name, sla.id])
```
shir-reifenberg, line 73: I think it's more safe to work with types objects
```suggestion
    const existingSlas = response.data.timeMetrics.map(({ name, id }) => ({ name, id }))
```

packages/jira-adapter/src/filters/sla_addition_deployment.ts
shir-reifenberg: looking again, can you reuse - `SlaParams`?

packages/jira-adapter/src/filters/sla_addition_deployment.ts
```diff
@@ -0,0 +1,177 @@
...
+    )
+
+    if (slaAdditionChanges.length === 0) {
+      return {
+        deployResult: { appliedChanges: [], errors: [] },
+        leftoverChanges: changes,
+      }
+    }
+
+    const projectToDefaultSlaAdditions = _.groupBy(slaAdditionChanges, change => {
```
shir-reifenberg, line 125: I don't see a reason to keep this map if you only use it to get the parent project in line 137
maybe just create an array of projects and then iterate it it instead?

packages/jira-adapter/src/filters/sla_addition_deployment.ts
```diff
@@ -0,0 +1,177 @@
...
+          instance,
+          config: jsmApiDefinitions,
+        }),
+      ),
+    })) as Change<InstanceElement>[]
+
+    const deployResult = await deployChanges(
+      typeFixedChanges.filter(isInstanceChange).filter(isAdditionChange),
+      async change => {
+        const serviceId = Object.fromEntries(
```
shir-reifenberg, line 155: I think `serviceId` is a little confusing in this context cause it has a different meaning in our code base

packages/jira-adapter/src/filters/sla_addition_deployment.ts
```diff
@@ -0,0 +1,177 @@
...
+          projectToServiceSlas[getParent(getChangeData(change)).elemID.getFullName()] ?? [],
+        )[change.data.after.value.name]
+        if (serviceId === undefined) {
+          await defaultDeployChange({
```
shir-reifenberg, line 159: readability nit - 
I think it would be easier to read if you create the change depends on `serviceId` and then just call `defaultDeployChange` once, something like - 
```suggestion
      const updatedChange = serviceId === undefined ? change : createSlaModificationChange(change)
       await defaultDeployChange({
          change: updatedChange,
          client,
          apiDefinitions: jsmApiDefinitions,
       })
```

packages/jira-adapter/src/filters/sla_addition_deployment.ts
```diff
@@ -0,0 +1,177 @@
...
+        id: Joi.number().required(),
+        name: Joi.string().required(),
+      }).unknown(true),
+    )
+    .required(),
+})
+  .unknown(true)
+  .required()
+
+const isSlaResponse = createSchemeGuard<SlaGetResponse>(SLA_RESPONSE_SCHEME)
```
shir-reifenberg, line 60: for logs :)
```suggestion
const isSlaResponse = createSchemeGuard<SlaGetResponse>(SLA_RESPONSE_SCHEME, 'Received invalid SLA response`)
```

packages/jira-adapter/src/filters/sla_addition_deployment.ts
```diff
@@ -0,0 +1,177 @@
...
+
+type SlaGetResponse = {
+  timeMetrics: SlaParams[]
+}
+const SLA_RESPONSE_SCHEME = Joi.object({
+  timeMetrics: Joi.array()
+    .items(
+      Joi.object({
+        id: Joi.number().required(),
+        name: Joi.string().required(),
```
shir-reifenberg, line 52: can names be empty? if so, please add `.allow('')` so the joi validation won't fail

packages/jira-adapter/src/change_validators/index.ts
```diff
@@ -82,7 +82,7 @@ const { deployTypesNotSupportedValidator, createChangeValidator, uniqueFieldsCha
 const TYPE_TO_UNIQUE_FIELD = {
   [ISSUE_TYPE_NAME]: ['name'],
   [PORTAL_GROUP_TYPE]: ['name'],
-  [SLA_TYPE_NAME]: ['name'],
+  // [SLA_TYPE_NAME]: ['name'], # TODO: uncomment when we add Scope to unique fields filter. SALTO-6424
```
shir-reifenberg, line 85: tiny nit
```suggestion
  // TODO SALTO-6424 uncomment when we add Scope to unique fields filter.
  // [SLA_TYPE_NAME]: ['name'],
```

packages/jira-adapter/test/filters/sla_addition_deployment.test.ts
```diff
@@ -0,0 +1,132 @@
...
+
+      expect(res.leftoverChanges).toHaveLength(0)
+      expect(res.deployResult.errors).toHaveLength(0)
+      expect(res.deployResult.appliedChanges).toHaveLength(2)
+      expect(mockPut).toHaveBeenCalledTimes(1)
+      expect(mockPost).toHaveBeenCalledTimes(1)
+    })
+  })
+
+  // Additional test cases could be added here, such as error handling scenarios
```
shir-reifenberg, line 131: let's remove if you don't intend to add those ?

shir-reifenberg (approved):
very cool and super fast fix 💪 💪 
Added some comments but approving to unblock

(also you can remove the user notifications, if it doesn't change anything in workspace it shouldn't be there)

packages/jira-adapter/src/filters/sla_addition_deployment.ts
idoamit8: Talked F2F

packages/jira-adapter/src/filters/sla_addition_deployment.ts
idoamit8: can not be

PR #6383, opened 2024-08-12
SALTO-5724: Avoid failing the entire plan for circular dependency error

shir-reifenberg:
When encounter circular dependencies in the grouped graph plan, remove the cycle from graph and create a valid plan without it.

---

Today, when the plan contains a circular dependency, we throw error and fail the entire plan, even for elements that could have been deployed. This PR changes this behavior, by removing the cycles from plan and creating change errors for the removed elements.

When first building the grouped graph, every cycle detected is removed from graph until no more cycles can be found.
After the first iteration of the grouped graph, we need to rerun the filter in order to adjust types and create additional (non circular, regular) dependency errors based on the dropped elements.
Then we try to build the group graph again, this time (hopefully), no cycles should exist.

The whole thing is controlled via `SALTO_FAIL_PLAN_ON_CIRCULAR_DEPENDENCY` env var 

---
_Release Notes_: 

_Core_:

- When deploy plan include elements which has circular dependency, the elements involved in the cycle will be removed from plan.

---
_User Notifications_: 
None

### DIFF
--- a/packages/core/src/core/plan/plan.ts
+++ b/packages/core/src/core/plan/plan.ts
@@ -450,7 +450,7 @@ const createGroupedGraphAndChangeErrors = async (
   after: ReadOnlyElementsSource,
   filterResult: FilterResult,
   customGroupIdFunctions: Record<string, ChangeGroupIdFunction>,
-): Promise<{ groupGraph: GroupDAG<Change>; changeErrors: ChangeError[] }> => {
+): Promise<{ groupedGraph: GroupDAG<Change>; changeErrors: ChangeError[] }> => {
   const { groupedGraph: firstIterationGraph, removedCycles } = await buildGroupedGraphFromFilterResult(
     before,
     after,
@@ -458,10 +458,10 @@ const createGroupedGraphAndChangeErrors = async (
     customGroupIdFunctions,
   )
   if (removedCycles.length === 0) {
-    return { groupGraph: firstIterationGraph, changeErrors: filterResult.changeErrors }
+    return { groupedGraph: firstIterationGraph, changeErrors: filterResult.changeErrors }
   }
 
-  // circular dependencies detected in the first iteration, build the graph again after all cycles were removed
+  log.error('detected circular dependencies in plan, rebuilding graph after cycles were removed')
 
   const circularDependencyErrors = removedCycles.flatMap(cycle => {
     const cycleIds = cycle.map(id => getChangeData(filterResult.validDiffGraph.getData(id)).elemID)
@@ -488,7 +488,7 @@ const createGroupedGraphAndChangeErrors = async (
   }
 
   return {
-    groupGraph: secondIterationGraph,
+    groupedGraph: secondIterationGraph,
     changeErrors: filterResult.changeErrors.concat(filteredCircularNodesResult.changeErrors),
   }
 }
@@ -538,14 +538,14 @@ export const getPlan = async ({
       const filterResult = await filterInvalidChanges(before, after, diffGraph, validatorsErrors)
 
       // build graph and add additional errors
-      const { groupGraph, changeErrors } = await createGroupedGraphAndChangeErrors(
+      const { groupedGraph, changeErrors } = await createGroupedGraphAndChangeErrors(
         before,
         after,
         filterResult,
         customGroupIdFunctions,
       )
       // build plan
-      return addPlanFunctions(groupGraph, changeErrors, compareOptions)
+      return addPlanFunctions(groupedGraph, changeErrors, compareOptions)
     },
     'get plan with %o -> %o elements',
     numBeforeElements,
--- a/packages/core/src/core/plan/plan.ts
+++ b/packages/core/src/core/plan/plan.ts
@@ -34,14 +34,16 @@ import {
   areReferencesEqual,
   CompareOptions,
   compareElementIDs,
+  getChangeData,
 } from '@salto-io/adapter-api'
 import { DataNodeMap, DiffNode, DiffGraph, GroupDAG } from '@salto-io/dag'
 import { logger } from '@salto-io/logging'
 import { expressions } from '@salto-io/workspace'
 import { collections, values } from '@salto-io/lowerdash'
+import { safeJsonStringify } from '@salto-io/adapter-utils'
 import { PlanItem, addPlanItemAccessors, PlanItemId } from './plan_item'
 import { buildGroupedGraphFromDiffGraph, getCustomGroupIds } from './group'
-import { filterInvalidChanges } from './filter'
+import { createCircularDependencyError, filterInvalidChanges, getChangeErrors, FilterResult } from './filter'
 import {
   addNodeDependencies,
   addFieldToObjectDependency,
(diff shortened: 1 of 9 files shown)


packages/core/test/core/plan/plan.test.ts
```diff
@@ -139,19 +140,31 @@ describe('getPlan', () => {
...
+  it('should omit fields from plan they create a dependency cycle on', async () => {
+    const plan = await planWithFieldDependencyCycle(false)
+    const planItems = [...plan.itemsByEvalOrder()]
+    expect(planItems).toHaveLength(6)
+    const [dependencyErrors, otherErrors] = _.partition(plan.changeErrors, isDependencyError)
+    expect(otherErrors).toHaveLength(2)
+    expect(dependencyErrors).toHaveLength(0)
   })
 
   it('should fail when plan has dependency cycle', async () => {
```
shir-reifenberg, line 161: This test still fails and I can't test if it's a real usecase and a bug in my code, or the test is set up incorrectly

packages/core/src/core/plan/plan.ts
```diff
@@ -423,6 +425,81 @@ const buildDiffGraph = (...transforms: ReadonlyArray<PlanTransformer>): Promise<
...
+    after,
+    filterResult,
+    customGroupIdFunctions,
+  )
+  if (removedCycles.length === 0) {
+    return { groupGraph: firstIterationGraph, changeErrors: filterResult.changeErrors }
+  }
+
+  // circular dependencies detected in the first iteration, build the graph again after all cycles were removed
+
```
dantal4, line 472: a log here would be usefull

packages/core/src/core/plan/plan.ts
```diff
@@ -423,6 +425,81 @@ const buildDiffGraph = (...transforms: ReadonlyArray<PlanTransformer>): Promise<
...
+  )
+  return { groupedGraph, removedCycles }
+}
+
+const createGroupedGraphAndChangeErrors = async (
+  before: ReadOnlyElementsSource,
+  after: ReadOnlyElementsSource,
+  filterResult: FilterResult,
+  customGroupIdFunctions: Record<string, ChangeGroupIdFunction>,
+): Promise<{ groupGraph: GroupDAG<Change>; changeErrors: ChangeError[] }> => {
```
dantal4, line 460: should probably be called `groupedGraph`

packages/dag/src/group.ts
```diff
@@ -191,22 +191,36 @@ const buildAcyclicGroupedGraphImpl = <T>(
...
+    return buildAcyclicGroupedGraphImpl(source, updatedGroupKey, origGroupKey, removedCycles)
+  } catch (error) {
+    if (error instanceof CircularDependencyError) {
+      const { causingNodeIds } = error
+      log.debug(
+        'detected circular dependency in group graph, removing the following nodes: %s',
+        causingNodeIds.map(toString).join(', '),
+      )
+      causingNodeIds.forEach(nodeId => source.deleteNode(nodeId))
+      return buildAcyclicGroupedGraphImpl(source, groupKey, origGroupKey, removedCycles.concat([causingNodeIds]))
```
dantal4, line 213: IMO instead of passing `removedCycles` list into `buildAcyclicGroupedGraphImpl`, the outer call should return its `causingNodeIds`, with the `removedCycles` of the inner call:
```suggestion
      const { graph, removedCycles } = buildAcyclicGroupedGraphImpl(source, groupKey, origGroupKey)
      return { graph, removedCycles: [[causingNodeIds]].concat(removedCycles) }
```

packages/dag/src/group.ts
```diff
@@ -191,22 +191,36 @@ const buildAcyclicGroupedGraphImpl = <T>(
...
   const possibleCycle = groupGraph.getCycle()
   if (possibleCycle === undefined) {
-    return new DAG(groupGraph.entries(), groupGraph.nodeData)
+    return { graph: new DAG(groupGraph.entries(), groupGraph.nodeData), removedCycles }
+  }
+  try {
+    const updatedGroupKey = modifyGroupKeyToRemoveCycle(groupGraph, groupKey, source, possibleCycle)
+    return buildAcyclicGroupedGraphImpl(source, updatedGroupKey, origGroupKey, removedCycles)
+  } catch (error) {
+    if (error instanceof CircularDependencyError) {
```
dantal4, line 206: I think that we'll want the ability to disable this feature using a core flag, so the current behavior will be a fallback in case of bugs.

packages/dag/src/group.ts
```diff
@@ -191,22 +191,36 @@ const buildAcyclicGroupedGraphImpl = <T>(
...
+      const { causingNodeIds } = error
+      log.debug(
+        'detected circular dependency in group graph, removing the following nodes: %s',
+        causingNodeIds.map(toString).join(', '),
```
dantal4, line 210: the `toString` is redundant

packages/dag/src/group.ts
```diff
@@ -191,22 +191,36 @@ const buildAcyclicGroupedGraphImpl = <T>(
...
+      )
+      causingNodeIds.forEach(nodeId => source.deleteNode(nodeId))
```
dantal4, line 212: are you sure that it's fine to mutate `source` ? I see that we use `filterResult.validDiffGraph` for it, and if it mutates then when we call `filterInvalidChanges` the second time then the cycled nodes won't be in it IIUC.

dantal4 (commented):
Great work!

packages/dag/src/group.ts
shir-reifenberg: It doesn't mutates the the original diff graph (`source`) because `removeRedundantFieldNodes` actually creates a new graph ([here](https://github.com/salto-io/salto/blob/main/packages/core/src/core/plan/group.ts#L135))

packages/dag/src/group.ts
shir-reifenberg: I found it harder to read this way 😅

packages/dag/src/group.ts
dantal4: since it's seating in a different package, I wonder whether we should clone the graph inside, or just add a comment that explains that we mutate the graph, and nodes that depend on cycles (but not part of it) are removed from the graph too.
@ori-moisis WDYT?

ori-moisis: I tried looking a bit the the PR, but it seems like this is fairly complicated and would take me a lot of time, so I'll just say that generally a function that changes its input and also returns a value is something we should avoid if possible.

packages/dag/src/group.ts
```diff
@@ -15,6 +15,7 @@ const { intersection } = collections.set
 const { DefaultMap } = collections.map
 
 const log = logger(module)
+const FAIL_ON_CIRCULAR_DEPENDENCY = 'SALTO_FAIL_PLAN_ON_CIRCULAR_DEPENDENCY'
```
dantal4, line 18: please use core flags for that, and pass the `shouldFailPlanOnCircularDependency` param to `buildAcyclicGroupedGraph` from `core/src/core/plan/group.ts` (IMO it should be required in `buildAcyclicGroupedGraph`)

packages/dag/src/group.ts
```diff
@@ -191,26 +192,59 @@ const buildAcyclicGroupedGraphImpl = <T>(
   source: DataNodeMap<T>,
   groupKey: GroupKeyFunc,
   origGroupKey: GroupKeyFunc,
-): GroupDAG<T> => {
+  removedCycles: collections.set.SetId[][] = [],
+  shouldFailOnCircularDependency = true,
```
dantal4, line 196: can this param be required ? maybe change the input here to be an object param?

dantal4 (commented):
💯 
last comments :)

dantal4 (approved):
Looks great 💯

PR #6385, opened 2024-08-12
SALTO-6421: Support Development Field In Issue Layout

idoamit8:
In this PR I support fetching and deploying development field in issue Layout

---

1. As we can see in the gira response we get from jira - there are two types of fields: `FIELD` and `PANEL`. 
Currently we supported only fields of type `FIELD` and I expended it to support also `PANEL` fields.  

---
_Release Notes_: 
_Jira Adapter:_
* [SALTO-6421](https://salto-io.atlassian.net/browse/SALTO-6421): Now support fetching and deploying development field in issueLayout. 

---
_User Notifications_: 
_Jira Adapter:_
* One or more fields of type `PANEL` will be added to the issueLayout type. 

[SALTO-6421]: https://salto-io.atlassian.net/browse/SALTO-6421?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/jira-adapter/src/filters/layouts/issue_layout.ts
+++ b/packages/jira-adapter/src/filters/layouts/issue_layout.ts
@@ -196,6 +196,15 @@ const deployLayoutChange = async (change: Change<InstanceElement>, client: JiraC
           },
         }
       }
+      if (item.type === 'PANEL') {
+        // Panel that is not a reference Expression
+        return {
+          type: item.type,
+          sectionType: item.sectionType.toLocaleLowerCase(),
+          key: item.key,
+          data: item.data,
+        }
+      }
       return undefined
     })
     .filter(isDefined)
--- a/packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
+++ b/packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
@@ -136,13 +136,13 @@ const fromLayoutConfigRespToLayoutConfig = (layoutConfig: IssueLayoutConfigurati
     .flatMap(container =>
       container.items.nodes.map(node => {
         const dataKey = node.panelItemId ?? node.fieldItemId
-        if (!dataKey) {
+        if (dataKey === undefined) {
           return undefined
         }
         return {
           type: node.fieldItemId ? 'FIELD' : 'PANEL',
           sectionType: container.containerType,
-          key: node.panelItemId ?? node.fieldItemId,
+          key: dataKey,
           data: fieldItemIdToMetaData[dataKey],
         }
       }),
--- a/packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
+++ b/packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
@@ -125,16 +125,27 @@ const fromLayoutConfigRespToLayoutConfig = (layoutConfig: IssueLayoutConfigurati
   const fieldItemIdToMetaData: Record<string, Value> = Object.fromEntries(
     (layoutConfig.metadata?.configuration.items.nodes ?? [])
       .filter(node => !_.isEmpty(node))
-      .map(node => [node.fieldItemId, _.omit(node, 'fieldItemId')]),
+      .map(node => {
+        if (node.fieldItemId) {
+          return [node.fieldItemId, _.omit(node, 'fieldItemId')]
+        }
+        return [node.panelItemId, _.omit(node, 'panelItemId')]
+      }),
   )
   const items = containers
     .flatMap(container =>
-      container.items.nodes.map(node => ({
-        type: 'FIELD',
-        sectionType: container.containerType,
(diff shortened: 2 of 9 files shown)


packages/jira-adapter/src/reference_mapping.ts
```diff
@@ -1194,7 +1194,6 @@ export const referencesRules: JiraFieldReferenceDefinition[] = [
   {
     src: { field: 'key', parentTypes: ['issueLayoutItem'] },
     serializationStrategy: 'id',
-    missingRefStrategy: 'typeAndValue',
```
idoamit8, line 1197: I removed the missingRefStrategy because the devSummary field is a locked field that reference the "Development" field. However, the key used is devSummary, while the ID of the "Development" field is `custom....` and the name is `Development` This mismatch results in the following configuration:
```
{
  type = "PANEL"
  sectionType = "PRIMARY"
  key = jira.Field.instance.missing_devSummary
  data = {
    name = "Development"
  }
}
```

Living this without reference is ok since its locked field and the service expects to get `devSummary`

packages/jira-adapter/src/filters/layouts/issue_layout.ts
```diff
@@ -196,6 +196,15 @@ const deployLayoutChange = async (change: Change<InstanceElement>, client: JiraC
           },
         }
       }
+      if (item.type === 'PANEL') {
```
IdoZakk, line 199: I think it will be safer to add a type check also in the other handling option above   (line 183), i.e. to check there that the type is indeed `Field`

IdoZakk, line 199: SuperNit: make 'PANEL` an constant

packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
```diff
@@ -125,16 +125,27 @@ const fromLayoutConfigRespToLayoutConfig = (layoutConfig: IssueLayoutConfigurati
...
     .flatMap(container =>
-      container.items.nodes.map(node => ({
-        type: 'FIELD',
-        sectionType: container.containerType,
-        key: node.fieldItemId,
-        data: fieldItemIdToMetaData[node.fieldItemId],
-      })),
+      container.items.nodes.map(node => {
+        const dataKey = node.panelItemId ?? node.fieldItemId
+        if (!dataKey) {
```
IdoZakk, line 139: nit: check undefined explicitly (=== undefined)

packages/jira-adapter/src/filters/layouts/layout_service_operations.ts
```diff
@@ -125,16 +125,27 @@ const fromLayoutConfigRespToLayoutConfig = (layoutConfig: IssueLayoutConfigurati
...
+          return undefined
+        }
+        return {
+          type: node.fieldItemId ? 'FIELD' : 'PANEL',
+          sectionType: container.containerType,
+          key: node.panelItemId ?? node.fieldItemId,
```
IdoZakk, line 145: You can use `dataKey` here

packages/jira-adapter/src/filters/layouts/layout_types.ts
```diff
@@ -141,7 +142,7 @@ export type LayoutConfigItem = {
 }
 
 export const ISSUE_LAYOUT_CONFIG_ITEM_SCHEME = Joi.object({
-  type: Joi.string().required(),
+  type: Joi.string().valid('FIELD', 'PANEL').required(),
```
IdoZakk, line 145: Are we sure there are no other types? otherwise we will screen out here issue layouts that were once passing the scheme guard

packages/jira-adapter/src/filters/request_type.ts
```diff
@@ -123,26 +123,32 @@ const deployRequestTypeLayout = async (
...
+      if (item.type === 'PANEL') {
+        // Panel that is not a reference Expression
+        return {
+          type: item.type,
+          sectionType: item.sectionType.toLocaleLowerCase(),
+          key: item.key,
+          data: item.data,
+        }
       }
+      return undefined
```
IdoZakk, line 151: You can move the original log error to here so we know it happened

packages/jira-adapter/src/reference_mapping.ts
IdoZakk: There were other ways to fix but I think it is the best one :-)

IdoZakk (approved):
💣 
See minor comments

packages/jira-adapter/src/filters/layouts/layout_types.ts
idoamit8: yes, there are any :)

PR #6386, opened 2024-08-12
SALTO-5912: (Jira) Support `AssigneeType` field in projects

SeggevHaimovich:
Support the `AsigneeType` field by fetching it from Jira using an API we haven't use so far in the fetch to get information about specific project.

---

---
_Release Notes_: 
_Jira adapter:_
* Support `AssigneeType` field in projects

---
_User Notifications_: 
_Jira adapter:_
* Project will contain `AsigneeType` which represents the Default Assignee

### DIFF
--- a/packages/jira-adapter/src/change_validators/projects/project_assignee_type.ts
+++ b/packages/jira-adapter/src/change_validators/projects/project_assignee_type.ts
@@ -1,0 +1,35 @@
+/*
+ *                      Copyright 2024 Salto Labs Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ChangeValidator, getChangeData, isAdditionOrModificationChange, isInstanceChange } from '@salto-io/adapter-api'
+import { ASSIGNEE_TYPE_FIELD, PROJECT_TYPE } from '../../constants'
+
+const VALID_ASSIGNEE_TYPES = ['PROJECT_LEAD', 'UNASSIGNED']
+
+export const projectAssigneeTypeValidator: ChangeValidator = async changes =>
+  changes
+    .filter(isInstanceChange)
+    .filter(isAdditionOrModificationChange)
+    .filter(change => change.data.after.elemID.typeName === PROJECT_TYPE)
+    .map(getChangeData)
+    .filter(project => project.value[ASSIGNEE_TYPE_FIELD] !== undefined)
+    .filter(project => !VALID_ASSIGNEE_TYPES.includes(project.value[ASSIGNEE_TYPE_FIELD]))
+    .map(instance => ({
+      elemID: instance.elemID,
+      severity: 'Error',
+      message: 'Invalid assignee type',
+      detailedMessage: `Project assignee type must be one of [${VALID_ASSIGNEE_TYPES.map(str => `'${str}'`).join(', ')}].`,
+    }))
--- a/packages/jira-adapter/src/filters/project.ts
+++ b/packages/jira-adapter/src/filters/project.ts
@@ -25,9 +25,10 @@ import {
   isInstanceElement,
   isModificationChange,
 } from '@salto-io/adapter-api'
-import { createSchemeGuard } from '@salto-io/adapter-utils'
+import { createSchemeGuard, setPath } from '@salto-io/adapter-utils'
+// import { resolveValues, config as configUtils } from '@salto-io/adapter-components'
 import { resolveValues } from '@salto-io/adapter-components'
-
+import { collections, values } from '@salto-io/lowerdash'
 import _ from 'lodash'
 import Joi from 'joi'
 import { logger } from '@salto-io/logging'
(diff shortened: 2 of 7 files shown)


packages/jira-adapter/src/filters/project.ts
```diff
@@ -288,6 +299,29 @@ const filter: FilterCreator = ({ config, client, elementsSource }) => ({
         instance.value.permissionScheme = instance.value.permissionScheme?.id?.toString()
         instance.value.issueSecurityScheme = instance.value.issueSecurityScheme?.id?.toString()
         changeProjectPath(instance, projectTypesKeysToFormattedKeys)
+        await setAssigneeTypeField(instance, client)
+        // TODO: should we add other fields from this response?
```
SeggevHaimovich, line 303: We get in the response many fields we currently do not support.
Maybe we should add them as well

SeggevHaimovich:
@IdoZakk Should we add noise reduction?

packages/jira-adapter/src/change_validators/projects/project_assignee_type.ts
```diff
@@ -0,0 +1,35 @@
...
+import { ASSIGNEE_TYPE_FIELD, PROJECT_TYPE } from '../../constants'
+
+const VALID_ASSIGNEE_TYPES = ['PROJECT_LEAD', 'UNASSIGNED']
+
+export const projectAssigneeTypeValidator: ChangeValidator = async changes =>
+  changes
+    .filter(isInstanceChange)
+    .filter(isAdditionOrModificationChange)
+    .filter(change => change.data.after.elemID.typeName === PROJECT_TYPE)
+    .map(getChangeData)
```
idoamit8, line 27:
```suggestion
.map(getChangeData)
.filter(instance => instance.elemID.typeName ===  PROJECT_TYPE)
```

packages/jira-adapter/src/change_validators/projects/project_assignee_type.ts
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ChangeValidator, getChangeData, isAdditionOrModificationChange, isInstanceChange } from '@salto-io/adapter-api'
+import { ASSIGNEE_TYPE_FIELD, PROJECT_TYPE } from '../../constants'
+
+const VALID_ASSIGNEE_TYPES = ['PROJECT_LEAD', 'UNASSIGNED']
+
+export const projectAssigneeTypeValidator: ChangeValidator = async changes =>
```
IdoZakk, line 22: Nit: add a description in a comment

packages/jira-adapter/src/filters/project.ts
```diff
@@ -269,10 +280,10 @@ const filter: FilterCreator = ({ config, client, elementsSource }) => ({
         .map(inst => [inst.value.key, inst.value.formattedKey]),
     )
 
-    elements
+    await awu(elements)
```
idoamit8, line 283: `awu` processes each item in the iterable one by one, which can be useful when dealing with streams of data or when you need to process items sequentially rather than in parallel
`Promise.all` is used to run multiple promises in parallel.
Why do we use `awu` here and not `promise.all`?

packages/jira-adapter/src/change_validators/projects/project_assignee_type.ts
```diff
@@ -0,0 +1,35 @@
...
+  changes
+    .filter(isInstanceChange)
+    .filter(isAdditionOrModificationChange)
+    .filter(change => change.data.after.elemID.typeName === PROJECT_TYPE)
```
IdoZakk, line 26: Nit: use getChangeData, it is cleaner

idoamit8 (commented):
Great Work! 💯 
Added some comment in order to save @IdoZakk some time.. hope It's ok  🙂

packages/jira-adapter/src/change_validators/projects/project_assignee_type.ts
```diff
@@ -0,0 +1,35 @@
...
+    .map(getChangeData)
+    .filter(project => project.value[ASSIGNEE_TYPE_FIELD] !== undefined)
+    .filter(project => !VALID_ASSIGNEE_TYPES.includes(project.value[ASSIGNEE_TYPE_FIELD]))
+    .map(instance => ({
+      elemID: instance.elemID,
+      severity: 'Error',
+      message: 'Invalid assignee type',
+      detailedMessage: `Project assignee type must be one of [${VALID_ASSIGNEE_TYPES.map(str => `'${str}'`).join(', ')}].`,
```
IdoZakk, line 34: You can add the received value, it will make it easier for the deployer

packages/jira-adapter/src/filters/project.ts
```diff
@@ -241,6 +243,15 @@ const deleteFieldConfigurationScheme = async (change: Change<InstanceElement>, c
...
 
+const setAssigneeTypeField = async (instance: InstanceElement, client: JiraClient): Promise<void> => {
+  const response = await client.get({ url: `/rest/api/3/project/${instance.value.key}` })
+  setPath(
+    instance,
+    instance.elemID.createNestedID(ASSIGNEE_TYPE_FIELD),
+    values.isPlainRecord(response.data) ? response.data[ASSIGNEE_TYPE_FIELD] : undefined,
+  )
+}
+
```
IdoZakk, line 254: You can achieve this with the api_config using recurse into, it will be safer and is a better practice

packages/jira-adapter/src/filters/project.ts
IdoZakk: We need to be more specific. Issue Types we have already. AvatarUrls we do not bring currently. and I think  we also have the lead. what type of roles and properties do we get?

IdoZakk (commented):
Please see my comment. I think we should use the api_config for the additional fetch

packages/jira-adapter/src/filters/project.ts
IdoZakk: After some research it will create more complications, leaving it as an additional call

PR #6388, opened 2024-08-12
SALTO-6419: Fix Axios error message

idoamit8:
In this PR I removed the HTTP errors that are thrown with the string `AxiosError` and are user visible.

---

---
_Release Notes_: 

---
_User Notifications_: 


netama (approved):
thanks!

PR #6389, opened 2024-08-12
SALTO-5610: (Salesforce) Support Type Exclusion/Inclusion in Change-Based fetch

tamtamirr:
(Salesforce) Support Type Exclusion/Inclusion in Change-Based fetch

---

Previous to this PR, when changing the adapter config to include/exclude a type it won't affect the scope of the fetched Elements.

- Will add UTs once the design of this PR is approved.

I've tested the following scenarios:
1. On Env with Profiles in exclude - Include Profile.
2. On the same Env, move Profiles back to exclude and make sure they are all removed.

---
_Release Notes_: 
Salesforce Adapter:
- Change-based fetch will now support including / excluding types in the adapter config.

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/adapter.ts
```diff
@@ -418,9 +417,7 @@ export const SYSTEM_FIELDS = [
 
 export const UNSUPPORTED_SYSTEM_FIELDS = ['LastReferencedDate', 'LastViewedDate']
 
-const getMetadataTypesFromElementsSource = async (
-  elementsSource: ReadOnlyElementsSource,
-): Promise<MetadataObjectType[]> =>
+const getMetadataTypesFromElementsSource = async (elementsSource: ReadOnlyElementsSource): Promise<TypeElement[]> =>
```
dantal4, line 420: why is this change?

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -559,9 +556,13 @@ export default class SalesforceAdapter implements AdapterOperations {
...
     const metadataTypeInfosPromise = this.listMetadataTypes(fetchProfile.metadataQuery)
-    const metadataTypesPromise = withChangesDetection
-      ? getMetadataTypesFromElementsSource(this.elementsSource)
-      : this.fetchMetadataTypes(metadataTypeInfosPromise, hardCodedTypes, metadataMetaType)
+    const metadataTypesPromise = this.fetchTypes({
+      metadataQuery,
+      withChangesDetection,
+      typeInfoPromise: metadataTypeInfosPromise,
+      knownMetadataTypes: hardCodedTypes,
+      metaType: metadataMetaType,
```
dantal4, line 564: these renames are a little confusing, and it would help to understand the flow if the var names will be consistent IMO

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -731,6 +732,40 @@ export default class SalesforceAdapter implements AdapterOperations {
...
+    typeInfoPromise: Promise<MetadataObject[]>
+    knownMetadataTypes: TypeElement[]
+    metaType?: ObjectType
+  }): Promise<TypeElement[]> {
+    if (!withChangesDetection) {
+      return this.fetchMetadataTypes(typeInfoPromise, knownMetadataTypes, metaType)
+    }
+    const metadataTypes = await typeInfoPromise
+    const typesFromSource = (await getMetadataTypesFromElementsSource(this.elementsSource)).filter(type =>
+      metadataQuery.isTypeMatch(apiNameSync(type) ?? ''),
```
dantal4, line 753: nit - since `getMetadataTypesFromElementsSource` is used only here - pass `metadataQuery` to `getMetadataTypesFromElementsSource` and filter in there.

maybe rename the function to `getMatchingMetadataTypesFromElementsSource`

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -731,6 +732,40 @@ export default class SalesforceAdapter implements AdapterOperations {
...
+    )
+    const typesFromSourceNames = new Set(typesFromSource.map(metadataType => apiNameSync(metadataType)))
```
dantal4, line 755: naming nit - `matchingTypeNamesFromSource`

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -731,6 +732,40 @@ export default class SalesforceAdapter implements AdapterOperations {
...
+    const missingTypes = metadataTypes.filter(type => !typesFromSourceNames.has(type.xmlName))
```
dantal4, line 756: don''t you need to check for `metadataQuery.isTypeMatch(type.xmlName)` too?

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -731,6 +732,40 @@ export default class SalesforceAdapter implements AdapterOperations {
...
+    if (missingTypes.length === 0) {
```
dantal4, line 757: aren't metadata types can change? is it fine to not fetch them?

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -824,20 +859,18 @@ export default class SalesforceAdapter implements AdapterOperations {
...
-  private async getElemIdsByTypeFromSource(): Promise<Record<string, ElemID[]>> {
+  private async getMetadataElementsByTypeFromSource(): Promise<Record<string, Element[]>> {
     const metadataElements = (await awu(await this.elementsSource.getAll()).toArray())
       .filter(element => !constants.NON_LISTED_ELEMENT_IDS.includes(element.elemID.getFullName()))
       .filter(element => isMetadataInstanceElementSync(element) || isCustomObjectSync(element))
-    return _(metadataElements)
-      .groupBy(metadataTypeSync)
-      .mapValues(elements => elements.map(e => e.elemID))
-      .value()
+    return _(metadataElements).groupBy(metadataTypeSync).value()
```
dantal4, line 866: nit
```suggestion
    return _.groupBy(metadataElements, metadataTypeSync)
```

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -849,34 +882,42 @@ export default class SalesforceAdapter implements AdapterOperations {
       fetchElements.filter(isMetadataObjectType),
       type => apiNameSync(type) ?? 'unknown',
     )
-    const elemIdsByTypeFromSource = await this.getElemIdsByTypeFromSource()
+    const metadataElementsByTypeFromSource = await this.getMetadataElementsByTypeFromSource()
     const deletedElemIds = new Set<ElemID>()
```
dantal4, line 886: I don't think that `Set` of elem ids works well. it should probably be `Set<string>`

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -849,34 +882,42 @@ export default class SalesforceAdapter implements AdapterOperations {
...
+        return
+      }
+      const listedInstancesFullNames = this.client.listedInstancesByType.getOrUndefined(typeName)
+      if (listedInstancesFullNames === undefined) {
+        log.warn('Skipping deletion detections for type %s since the type was not listed', typeName)
+        return
+      }
+      const listedElemIdsFullNames = new Set(
+        Array.from(listedInstancesFullNames)
+          // We invoke createInstanceElement to have the correct elemID that we calculate in fetch
```
dantal4, line 911: I think that this comment belong to line 879 (above `createInstanceElement`)

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -849,34 +882,42 @@ export default class SalesforceAdapter implements AdapterOperations {
...
+      if (isTargetedFetch && !fetchProfile.metadataQuery.isTypeMatch(typeName)) {
+        return
+      }
+      // Type was excluded, we should remove all of its Instances
+      if (!fetchProfile.metadataQuery.isTypeMatch(typeName)) {
+        elementsFromSource.forEach(elementFromSource => {
+          deletedElemIds.add(elementFromSource.elemID)
         })
+        return
+      }
```
dantal4, line 898: nit
```suggestion
      if (!fetchProfile.metadataQuery.isTypeMatch(typeName)) {
        // Instances of Type that is not part of the fetch targets should not be handled
        if (!isTargetedFetch) {
          // Type was excluded, we should remove all of its Instances
          elementsFromSource.forEach(elementFromSource => {
            deletedElemIds.add(elementFromSource.elemID)
          })
        }
        return
      }
```

dantal4 (commented):
Great work! LMK WDYT about my thoughts here 👍

packages/salesforce-adapter/src/adapter.ts
tamtamirr: This is existing code that works :)  Core also expects the adapter to return `ElemID[]` as part of the `partialFetchResult`, so the conversation seems unnecessary here

packages/salesforce-adapter/src/adapter.ts
tamtamirr: L766 :) To avoid type casting

packages/salesforce-adapter/src/adapter.ts
tamtamirr: MetadataTypes may change on release of new Salesforce API Version every couple months.
Full fetch will still retrieve all the new types.

dantal4 (approved):
💯

PR #6390, opened 2024-08-12
SALTO-6265 Add ticket form restricted_brand_ids sorting

adfineberg:
Add ticket form restricted_brand_ids sorting

---

---
_Release Notes_: 
__Zendesk Adapter:__
* Ticket form `restricted_brand_ids` are now sorted, as these cannot be updated by users.

---
_User Notifications_: 
__Zendesk Adapter:__
* Ticket form `restricted_brand_ids` are now sorted, as these cannot be updated by users.

### DIFF
--- a/packages/zendesk-adapter/src/filters/unordered_lists.ts
+++ b/packages/zendesk-adapter/src/filters/unordered_lists.ts
@@ -22,7 +22,7 @@ import {
   ReferenceExpression,
 } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
-import { isResolvedReferenceExpression } from '@salto-io/adapter-utils'
+import { inspectValue, isResolvedReferenceExpression } from '@salto-io/adapter-utils'
 import { FilterCreator } from '../filter'
 import { DYNAMIC_CONTENT_ITEM_VARIANT_TYPE_NAME } from './dynamic_content'
 import {
@@ -234,6 +234,7 @@ const sortRestrictedBrands = (formInstances: InstanceElement[]): void => {
       form.value.restricted_brand_ids = _.sortBy(restrictedBrands, brand => brand.elemID.getFullName())
     } else {
       log.warn(`could not sort restricted brands for ${form.elemID.getFullName()}`)
+      log.trace('restricted brands are: %s', inspectValue(restrictedBrands, { maxArrayLength: null }))
     }
   })
 }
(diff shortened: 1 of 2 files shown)


packages/zendesk-adapter/src/filters/unordered_lists.ts
```diff
@@ -227,6 +227,17 @@ const sortChildFields = (formInstances: InstanceElement[], ticketFieldById: Reco
   })
 }
 
+const sortRestrictedBrands = (formInstances: InstanceElement[]): void => {
+  formInstances.forEach(form => {
+    const restrictedBrands = form.value.restricted_brand_ids
+    if (_.isArray(restrictedBrands) && restrictedBrands.every(isReferenceExpression)) {
+      form.value.restricted_brand_ids = _.sortBy(restrictedBrands, brand => brand.value.value.name)
```
edenhassid, line 234: I would sort this by the elemId fullName and not the `.value.name` cause there reference may be unresolved

packages/zendesk-adapter/src/filters/unordered_lists.ts
```diff
@@ -227,6 +227,17 @@ const sortChildFields = (formInstances: InstanceElement[], ticketFieldById: Reco
   })
 }
 
+const sortRestrictedBrands = (formInstances: InstanceElement[]): void => {
+  formInstances.forEach(form => {
+    const restrictedBrands = form.value.restricted_brand_ids
```
tamtamirr, line 232: Can we const `restricted_brand_ids`?

packages/zendesk-adapter/src/filters/unordered_lists.ts
```diff
@@ -227,6 +227,17 @@ const sortChildFields = (formInstances: InstanceElement[], ticketFieldById: Reco
...
+    if (_.isArray(restrictedBrands) && restrictedBrands.every(isReferenceExpression)) {
+      form.value.restricted_brand_ids = _.sortBy(restrictedBrands, brand => brand.elemID.getFullName())
+    } else {
+      log.warn(`could not sort restricted brands for ${form.elemID.getFullName()}`)
```
tamtamirr, line 236: This is very unlikely to reach here, but if we do it'd be hard to debug. Let's add a trace on the actual values
```suggestion
      log.warn(`could not sort restricted brands for ${form.elemID.getFullName()}`)
      log.trace(`restricted brands are: %s`, inspectValue(restrictedBrands, { maxArrayLength: null }))
```

tamtamirr (commented):
LGTM! Added some small comments :)

packages/zendesk-adapter/src/filters/unordered_lists.ts
edenhassid: not sure I understand what you mean

tamtamirr: Add a constant
e.g. `RESTRICTED_BRAND_IDS = 'restricted_brand_ids`

tamtamirr (approved):

PR #6393, opened 2024-08-12
SALTO-6269 Add selected macros restriction ids sorting

adfineberg:
Add selected macros restriction ids sorting

---

---
_Release Notes_: 
__Zendesk Adapter:__
* Workspace selected macros restriction ids are now sorted, as these cannot be updated by users.

---
_User Notifications_: 
__Zendesk Adapter:__
* Workspace selected macros restriction ids are now sorted, as these cannot be updated by users.

### DIFF
--- a/packages/zendesk-adapter/src/filters/unordered_lists.ts
+++ b/packages/zendesk-adapter/src/filters/unordered_lists.ts
@@ -301,7 +301,7 @@ const sortSelectedMacros = (selectedMacros: { restriction?: MacroRestriction }[]
 
 const orderSelectedMacros = (workspace: InstanceElement): void => {
   const selectedMacros = workspace.value.selected_macros
-  if (_.isArray(selectedMacros) && selectedMacros.every(macro => macro.id !== undefined)) {
+  if (_.isArray(selectedMacros) && selectedMacros.every(macro => _.isPlainObject(macro) && macro.id !== undefined)) {
     sortSelectedMacros(selectedMacros)
     workspace.value.selected_macros = _.sortBy(selectedMacros, macro => {
       if (isReferenceExpression(macro.id)) {
--- a/packages/zendesk-adapter/src/filters/unordered_lists.ts
+++ b/packages/zendesk-adapter/src/filters/unordered_lists.ts
@@ -22,7 +22,7 @@ import {
   ReferenceExpression,
 } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
-import { isResolvedReferenceExpression } from '@salto-io/adapter-utils'
+import { inspectValue, isResolvedReferenceExpression } from '@salto-io/adapter-utils'
 import { FilterCreator } from '../filter'
 import { DYNAMIC_CONTENT_ITEM_VARIANT_TYPE_NAME } from './dynamic_content'
 import {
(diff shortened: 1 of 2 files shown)


packages/zendesk-adapter/src/filters/unordered_lists.ts
```diff
@@ -281,9 +281,27 @@ const orderMacroIds = (workspace: InstanceElement): void => {
...
+type MacroRestriction = {
+  ids?: ReferenceExpression[]
+}
+
+const sortSelectedMacros = (selectedMacros: { restriction?: MacroRestriction }[]): void =>
+  selectedMacros.forEach(macro => {
+    if (macro.restriction !== undefined && _.isObject(macro.restriction) && 'ids' in macro.restriction) {
+      const restrictionIds = _.get(macro, 'restriction.ids')
+      if (_.isArray(restrictionIds) && restrictionIds.every(isReferenceExpression)) {
+        macro.restriction.ids = _.sortBy(restrictionIds, 'value.value.name')
```
edenhassid, line 294: I would sort this by the elemId fullName and not the value inside the reference

packages/zendesk-adapter/src/filters/unordered_lists.ts
```diff
@@ -281,9 +281,27 @@ const orderMacroIds = (workspace: InstanceElement): void => {
...
+}
+
+const sortSelectedMacros = (selectedMacros: { restriction?: MacroRestriction }[]): void =>
+  selectedMacros.forEach(macro => {
+    if (macro.restriction !== undefined && _.isObject(macro.restriction) && 'ids' in macro.restriction) {
+      const restrictionIds = _.get(macro, 'restriction.ids')
+      if (_.isArray(restrictionIds) && restrictionIds.every(isReferenceExpression)) {
+        macro.restriction.ids = _.sortBy(restrictionIds, ref => ref.elemID.getFullName())
+      } else {
+        log.warn(`could not sort restriction ids for restriction ${macro.restriction}`)
```
tamtamirr, line 296: For easier debugging
```suggestion
        log.warn(`could not sort restriction ids for restriction ${macro.restriction}`)
        log.trace(`restriction ids are: ${inspectValue(restrictionIds, { maxArrayLength: null })}`)
```

packages/zendesk-adapter/src/filters/unordered_lists.ts
```diff
@@ -281,9 +281,27 @@ const orderMacroIds = (workspace: InstanceElement): void => {
...
+      }
+    }
+  })
+
 const orderSelectedMacros = (workspace: InstanceElement): void => {
   const selectedMacros = workspace.value.selected_macros
   if (_.isArray(selectedMacros) && selectedMacros.every(macro => macro.id !== undefined)) {
+    sortSelectedMacros(selectedMacros)
```
tamtamirr, line 304: Don't we deal with `any` here?
The type of the parameter `selectedMacros` is `{ restriction?: MacroRestriction }[]` 
I understand this will be handled in L393, but IMO it makes the code more confusing. We could implement the typeguard and run it in L303 as part of the `every` iteration.

tamtamirr (commented):
Added a few small comments :)

packages/zendesk-adapter/src/filters/unordered_lists.ts
edenhassid: no because not all macros will have restrictions, and we do want the other code to run

PR #6395, opened 2024-08-12
SALTO-6462: Fetch only OAuth2 Permission Grants for all users

shoshana-houri:
We discovered potential collisions with the elemIDs of `OAuth2PermissionGrant` instances when the `consentType` is `Principal` and the same permission is granted to different users. To address this, we've chosen to filter out these permissions in the request, as we currently don't manage users at all.

---
_Release Notes_: 
_Microsoft Entra:_
* Filter out `OAuth2PermissionGrant` instances that grant permissions to specific users.

---
_User Notifications_: 
_Microsoft Entra:_
* Filter out `OAuth2PermissionGrant` instances that grant permissions to specific users.
* Modify the elemID of `OAuth2PermissionGrant` by removing `consentType` from it.


shir-reifenberg (approved):
💪

PR #6397, opened 2024-08-13
SALTO-6468: Fix assetsObjectFieldConfigurationAqlValidator

idoamit8:
In this PR I fixed `assetsObjectFieldConfigurationAqlValidator`

---

There's an issue with the assetsObjectFieldConfigurationAqlValidator.

When we add both the context and its field parent, the field parent doesn't have an id attribute at this stage. However, we're trying to access this attribute, which leads to the undefined error.

For field and context additions we will add a general action item instead of specific URL.
I also validated that there is valid parent. 

---
_Release Notes_: 
_Jira Adapter:_
* Fixed Bug in `assetsObjectFieldConfigurationAqlValidator` when we are reading startsWith of an undefined object. 
---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/change_validators/field_contexts/assets_object_field_configuration_aql.ts
+++ b/packages/jira-adapter/src/change_validators/field_contexts/assets_object_field_configuration_aql.ts
@@ -14,15 +14,9 @@
  * limitations under the License.
  */
 
-import {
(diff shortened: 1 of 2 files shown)


kami-delaroz (approved):
🔥

PR #6399, opened 2024-08-13
SALTO-6470: Fix dependsOn definition for groupLifeCyclePolicy

shoshana-houri:
We defined the group lifecycle policy as dependent on the lifecycle policy to perform a conditional request. However, setting a sub-resource as dependent on another resource, which is not its parent, is not supported. Instead, we should define the parent as dependent on whatever we need and use the conditions under the recurseInto definition.

---
_Release Notes_: 
_Microsoft Entra:_
* Fix dependsOn definition for groupLifeCyclePolicy

---
_User Notifications_: 
_Microsoft Entra:_
* groupLifeCyclePolicy will now be referenced from the group to which it applies.


shir-reifenberg (approved):
Very cool 😎

PR #6401, opened 2024-08-13
SALTO-5891 Resolve the FlowDefinition type in preDeploy

dantal4:
In order to deactivate a `Flow` instance, we create a `FlowDefinition` instance with `activeVersionNumber` of 0. We do it by getting the `FlowDefinition` type from the `elementsSource` and create the instance with that type. The problem was that the type's fields weren't resolved, so while running `transformElement` on the `FlowDefinition` instance we got the following error- `Can not resolve value of reference with ElemID serviceid without elementsSource because value does not exist`.
The solution is to resolve the `FieldDefinition` type's fields & annotations when we get it from the `elementsSource`.

---

---
_Release Notes_: 
Salesforce:
- Fix the "Can not resolve ..." error on `Flow` deactivation changes.

---
_User Notifications_: 
None

### DIFF
--- a/packages/salesforce-adapter/src/filters/flows_filter.ts
+++ b/packages/salesforce-adapter/src/filters/flows_filter.ts
@@ -27,7 +27,7 @@ import {
   ObjectType,
   toChange,
 } from '@salto-io/adapter-api'
-import { findObjectType } from '@salto-io/adapter-utils'
+import { findObjectType, resolveTypeShallow } from '@salto-io/adapter-utils'
 import { values as lowerdashValues } from '@salto-io/lowerdash'
 import { FilterResult, RemoteFilterCreator } from '../filter'
 import {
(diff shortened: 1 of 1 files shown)


tamtamirr (approved):
LGTM! This PR raises the thought that perhaps we need core to pass a different implementation of **elementsSoruce** for deploy that have the types resolved (at-least shallowly) Can you create a ticket and add a comment with a link to the ticket above L170?
Approving to save cycles :)

dantal4:
> LGTM! This PR raises the thought that perhaps we need core to pass a different implementation of **elementsSoruce** that have the types resolved (at-least shallowly) Can you create a ticket and add a comment with a link to the ticket above L170? Approving to save cycles :)

@tamtamirr IIUC we intentionally don't want to resolve all types in deploy, maybe @ori-moisis can elaborate more.

PR #6402, opened 2024-08-13
SALTO-6472: Omit value set name annotation on object / field addition

ori-moisis:
Currently we only omit the value set name annotation on field modification This extends this behavior to field addition and to all fields inside an object addition

---

These new scenarios can't really happen in a real deployment (we cannot deploy a standard object / new standard field) but they can happen in a "dumpElementsToFolder" flow 

Manual tests:
- Tested setting / removing description on a standard picklist field (Account.Type), checked the change is deployed, the nacl does not change and there are no pending changes after the deploy 

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/src/filters/standard_value_sets.ts
+++ b/packages/salesforce-adapter/src/filters/standard_value_sets.ts
@@ -22,27 +22,25 @@ import {
   Field,
   ReferenceExpression,
   isObjectType,
-  isModificationChange,
-  isFieldChange,
-  getChangeData,
-  Change,
+  isAdditionOrModificationChange,
+  isField,
   getAllChangeData,
-  ModificationChange,
+  isObjectTypeChange,
+  isFieldChange,
+  Change,
 } from '@salto-io/adapter-api'
 import { buildElementsSourceFromElements } from '@salto-io/adapter-utils'
-import { collections, promises, types, values as lowerdashValues } from '@salto-io/lowerdash'
+import { collections, types } from '@salto-io/lowerdash'
 
 import { FilterResult, RemoteFilterCreator } from '../filter'
 import { FIELD_ANNOTATIONS, VALUE_SET_FIELDS } from '../constants'
-import { metadataType, apiName, isCustomObject, Types, isCustom } from '../transformers/transformer'
-import { extractFullNamesFromValueList, isInstanceOfTypeSync } from './utils'
+import { metadataType, isCustomObject, Types, isCustom } from '../transformers/transformer'
+import { apiNameSync, extractFullNamesFromValueList, isInstanceOfTypeSync } from './utils'
 import { ConfigChangeSuggestion } from '../types'
 import { fetchMetadataInstances } from '../fetch'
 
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/src/filters/standard_value_sets.ts
```diff
@@ -256,28 +250,25 @@ export const makeFilter =
...
-          .filter(isFieldChange)
-          .filter(change => isStandardPickList(getChangeData(change)))
-          .toArray()
-        originalChanges = await keyByAsync(standardPicklistFieldChanges, change => apiName(getChangeData(change)))
-        const deployableChanges = standardPicklistFieldChanges.map(toDeployableStandardPicklistFieldChange)
-        _.pullAll(changes, standardPicklistFieldChanges)
-        changes.push(...deployableChanges)
+        getAllStandardPicklistFields(changes).forEach(field => {
+          const svsName = field.annotations[VALUE_SET_FIELDS.VALUE_SET_NAME]
+          if (svsName !== undefined) {
```
tamtamirr, line 255: Are there any cases where the `svsName` won't be a string?
The code should work just as expected even if it's not a string as we simply omit the value and then set it back, just curious if we can improve on the fact we deal with `any` value here somehow.

packages/salesforce-adapter/test/filters/standard_value_sets.test.ts
```diff
@@ -260,42 +264,88 @@ describe('Standard Value Sets filter', () => {
...
+          description: 'after',
+        })
+        originalChange = toChange({
+          before: beforePicklistStandardField,
+          after: afterPicklistStandardField,
+        })
+        afterPreDeployChanges = [cloneDeepWithoutRefs(originalChange)]
+        await filter.preDeploy(afterPreDeployChanges)
+        afterOnDeployChanges = cloneDeepWithoutRefs(afterPreDeployChanges)
+        await filter.onDeploy(afterOnDeployChanges)
```
tamtamirr, line 294: Is there a reason to run the execution of `preDeploy` and `onDeploy` within the `beforeEach`?
I suggest only preparing the change in the `beforeEach` and execute the `preDeploy` and `onDeploy` within the test to avoid the clones. We could also test the annotations on the field change directly instead of needing to use `toSatisfyAll`. I think we can also simplify the other test case similarly.

tamtamirr (commented):
LGTM! Added some small comments

packages/salesforce-adapter/src/filters/standard_value_sets.ts
ori-moisis: there is no way of knowing what we are going to get here, users can edit the nacl manually for all we know

packages/salesforce-adapter/test/filters/standard_value_sets.test.ts
ori-moisis: no reason AFAICT, I was going with the existing tests' style

tamtamirr (approved):
LGTM

PR #6404, opened 2024-08-13
SALTO-4967 - Salesforce: Better handling for canceled metadata deploy

tomsellek:
If the deploy is canceled, return an error and no applied changes

---

---
_Release Notes_: 
Salesforce: will now show a meaningful error if the deploy is canceled

---
_User Notifications_: 
N/A


packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -247,6 +247,18 @@ const processDeployResponse = (
     return 'Warning'
   }
 
+  const canceledDeployError = (): SaltoError => ({
```
adamro, line 250: As is this doesn't require any function context - so define it outside of the fn please

Can we be more exact with the message please? Check if it's `checkOnly` and if yes write `Validation` and if not `Deployment`

tomsellek: Ack

adamro (approved):

PR #6406, opened 2024-08-14
SALTO-6479 Fix PrimitiveValue type to include null & undefined

dantal4:
`isPrimitiveValue` returns true for `null`, `undefined`, but those weren't part of the the `PrimitiveValue` type.

---

---
_Release Notes_: 
Core:
- List items comparison will not throw when there's `null` / `undefined` in an item.

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-utils/src/list_comparison.ts
+++ b/packages/adapter-utils/src/list_comparison.ts
@@ -109,7 +109,7 @@ const getSingleValueKey = (value: LeafValue): string => {
   if (isTemplateExpression(value)) {
     return value.parts.map(getSingleValueKey).join('')
   }
-  return value.toString()
+  return value?.toString() ?? ''
 }
 
 /**
--- a/packages/salesforce-adapter/src/transformers/transformer.ts
+++ b/packages/salesforce-adapter/src/transformers/transformer.ts
@@ -1232,9 +1232,9 @@ export const transformPrimitive: TransformFuncSync = ({ value, path, field }) =>
     case PrimitiveTypes.NUMBER:
       return Number(value)
     case PrimitiveTypes.BOOLEAN:
-      return value.toString().toLowerCase() === 'true'
+      return String(value).toLowerCase() === 'true'
     case PrimitiveTypes.STRING:
-      return value.toString()
+      return String(value)
     default:
       return value
   }
--- a/packages/salesforce-adapter/src/transformers/transformer.ts
+++ b/packages/salesforce-adapter/src/transformers/transformer.ts
@@ -1228,13 +1228,16 @@ export const transformPrimitive: TransformFuncSync = ({ value, path, field }) =>
   if (!isPrimitiveType(fieldType) || !isPrimitiveValue(value)) {
     return value
   }
+  if (value === null || value === undefined) {
+    return undefined
+  }
   switch (fieldType.primitive) {
     case PrimitiveTypes.NUMBER:
       return Number(value)
     case PrimitiveTypes.BOOLEAN:
-      return String(value).toLowerCase() === 'true'
+      return value.toString().toLowerCase() === 'true'
     case PrimitiveTypes.STRING:
-      return String(value)
+      return value.toString()
     default:
       return value
   }
--- a/packages/zendesk-adapter/src/filters/guide_service_url.ts
+++ b/packages/zendesk-adapter/src/filters/guide_service_url.ts
@@ -59,7 +59,7 @@ const replaceUrlParamsBrand = (url: string, instance: InstanceElement): string =
     if (!isPrimitiveValue(replacement)) {
       throw new Error(`Cannot replace param ${val} in ${url} with non-primitive value ${replacement}`)
     }
-    return replacement.toString()
+    return String(replacement)
   })
 
 const createServiceUrl = (instance: InstanceElement, baseUrl: string): void => {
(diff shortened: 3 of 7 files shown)


ori-moisis:
@dantal4  please create a ticket with some context for this and link this PR to it

packages/salesforce-adapter/src/transformers/transformer.ts
```diff
@@ -1232,9 +1232,9 @@ export const transformPrimitive: TransformFuncSync = ({ value, path, field }) =>
     case PrimitiveTypes.NUMBER:
       return Number(value)
     case PrimitiveTypes.BOOLEAN:
-      return value.toString().toLowerCase() === 'true'
+      return String(value).toLowerCase() === 'true'
     case PrimitiveTypes.STRING:
-      return value.toString()
+      return String(value)
```
ori-moisis, line 1237: this is potentially a real behavior change.
it will cause null / undefined values to become "null" or "undefined" which completely changes the meaning of the value.

technically I think we can't get here with `null` because of a check at the start of the function, but AFAICT there is nothing explicitly preventing us from getting here with `undefined`

I think the case for undefined may need to be covered in the condition above to ensure we don't get here with null / undefined

packages/zendesk-adapter/src/filters/guide_service_url.ts
```diff
@@ -59,7 +59,7 @@ const replaceUrlParamsBrand = (url: string, instance: InstanceElement): string =
     if (!isPrimitiveValue(replacement)) {
       throw new Error(`Cannot replace param ${val} in ${url} with non-primitive value ${replacement}`)
     }
-    return replacement.toString()
+    return String(replacement)
```
ori-moisis, line 62: similar issue as the comment on the salesforce transformer
maybe the check above should not allow null / undefined?

ori-moisis (commented):
there are more places where `isPrimitiveValue` is used and followed by a ".toString" (specifically a few places in adapter-components).

maybe we need two "isPrimitiveValue" functions? (one that includes null / undefined and one that doesn't)

packages/salesforce-adapter/src/transformers/transformer.ts
dantal4: This didn't happen until now, otherwise this function would have throw, right? So it's not really a behavior change....

dantal4:
> there are more places where `isPrimitiveValue` is used and followed by a ".toString" (specifically a few places in adapter-components).
> 
> maybe we need two "isPrimitiveValue" functions? (one that includes null / undefined and one that doesn't)

@ori-moisis in other places using `isPrimitiveValue(value)` we make sure that `value` isn't null/undefined, and TS is aware of that, and that's why the build isn't breaks.

ori-moisis:
what about cases where TS thinks the value is `any`?

dantal4:
> what about cases where TS thinks the value is `any`?

@ori-moisis TS would break too, because after `isPrimitiveValue` the `value` is not `any` anymore. That was the case [here](https://github.com/salto-io/salto/pull/6406/files#diff-31803eba82b498dd8bbf0835c3d9020a1d6c96d1c80bbfcefa5100ccc4896537R1237).
In addition, I checked all the places that use `isPrimitiveValue` and verified that we're not accessing the value properties without additional check.

packages/salesforce-adapter/src/transformers/transformer.ts
ori-moisis: the behavior change is that if we get here in the future we won't know it happened and just silently create another "ticking time bomb" in a user's configuration
not failing here is the behavior change

packages/adapter-utils/src/list_comparison.ts
```diff
@@ -109,7 +109,7 @@ const getSingleValueKey = (value: LeafValue): string => {
   if (isTemplateExpression(value)) {
     return value.parts.map(getSingleValueKey).join('')
   }
-  return value.toString()
+  return value?.toString() ?? ''
```
ori-moisis, line 112: there should be a unit test for this

packages/zendesk-adapter/src/filters/guide_service_url.ts
```diff
@@ -52,14 +52,14 @@ const replaceUrlParamsBrand = (url: string, instance: InstanceElement): string =
...
+      replacement = _.get(instance.annotations, val.slice(1, -1))
     } else {
-      replacement = instance.value[val.slice(1, -1)] ?? val
+      replacement = instance.value[val.slice(1, -1)]
     }
     if (!isPrimitiveValue(replacement)) {
       throw new Error(`Cannot replace param ${val} in ${url} with non-primitive value ${replacement}`)
     }
-    return replacement.toString()
+    return replacement?.toString() ?? val
```
ori-moisis, line 62: IIUC this fallback is not really valid
it means that we would potentially create a url of `/knowledge/articles/_parent.0.value.value.id/locale?brand_id=brand`.
I understand this is what the old code would do, but I am pretty sure that is not the intended behavior here.

unless I am missing something, it seems like if we are unable to resolve a real value (not null or undefined), we should fail here.
not sure throwing an exception is the correct behavior though as it might impact all instances / the entire fetch.
it would probably be better to just allow the resulting url to be undefined and in that case just not to set it on the instance that is missing a value...

ori-moisis (approved):
The remaining issue is technically not a regression from the current behavior
approving just in case you want to fix it in a separate PR

packages/zendesk-adapter/src/filters/guide_service_url.ts
dantal4: fixed 👍

PR #6409, opened 2024-08-14
SALTO-6478: component crash when no lead id

IdoZakk:
None
---

None

---
_Release Notes_: 
Jira Adapter:
* Fixed a bug in jira DC causing components with no lead to fail deployment

---
_User Notifications_: 
None


shayc331 (approved):
💯

PR #6411, opened 2024-08-14
SALTO-6478: deploy dc component with lead id

IdoZakk:
The deployment of project components with lead ID did not work- we send the id but we need to send the user name

---

None

---
_Release Notes_: 
Jira Adapter:
* Fixed a bug that prevented deployment of components with lead ID in dc

---
_User Notifications_: 
None


kami-delaroz (approved):
🔥

PR #6414, opened 2024-08-14
SALTO-6473: Support changing fields with unknown type

ori-moisis:
The metadata API technically supports deploying fields with unknown type if the type was already unknown This change:
- Removes the change validator that would prevent modifying existing fields with unknown type
- Moves the logic of putting an error on unknown field types to the custom_field_type validator to avoid having double warnings
- Fixes the deployment code / filters to not explicitly specify the field type when it is unknown

---

Manually tested with:
- Deploying a new object with a field that has an unknown type (the field with the unknown type is not deployed, deployment succeeds)
- Adding a field with an unknown type (gets change validation error)
- Modifying a field with an unknown type (adding a "description") - works

---
_Release Notes_: 
_Salesforce Adapter_
- Added support for deploying changes to fields with unknown type
- Removed double change validator when a field has an invalid type, leaving only one validation error

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/src/constants.ts
+++ b/packages/salesforce-adapter/src/constants.ts
@@ -91,11 +91,19 @@ export const COMPOUND_FIELDS_SOAP_TYPE_NAMES: Record<string, COMPOUND_FIELD_TYPE
   // name is handled differently with nameField
 }
 
-// target types for creating / updating custom fields:
-export const CUSTOM_FIELD_UPDATE_CREATE_ALLOWED_TYPES: string[] = [
+// target types for creating / updating custom fields
+// we can create fields of these types or update the types of existing fields to these types
+export const CUSTOM_FIELD_UPDATE_CREATE_ALLOWED_TYPES: (string | undefined)[] = [
   ...Object.values(FIELD_TYPE_NAMES),
   COMPOUND_FIELD_TYPE_NAMES.LOCATION,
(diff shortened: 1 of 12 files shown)


packages/salesforce-adapter/src/constants.ts
```diff
@@ -99,13 +99,21 @@ export const COMPOUND_FIELDS_SOAP_TYPE_NAMES: Record<string, COMPOUND_FIELD_TYPE
   // name is handled differently with nameField
 }
 
-// target types for creating / updating custom fields:
-export const CUSTOM_FIELD_UPDATE_CREATE_ALLOWED_TYPES: string[] = [
+// target types for creating / updating custom fields
+// we can create fields of these types or update the types of existing fields to these types
+export const CUSTOM_FIELD_UPDATE_CREATE_ALLOWED_TYPES: (string | undefined)[] = [
```
tamtamirr, line 104: Is any of the values undefined here?

tamtamirr (approved):
LGTM! Added one small question

packages/salesforce-adapter/src/constants.ts
ori-moisis: no, but the type needs to allow undefined in order for things like
```typescript
CUSTOM_FIELD_UPDATE_CREATE_ALLOWED_TYPES.includes(fieldType)
```
to work when `fieldType` can be undefined

PR #6416, opened 2024-08-15
SALTO-5984: Okta: Omit `EmailDomain` instances marked as `DELETED`

Nurdok:
This PR fixes a bug when fetching EmailDomain instances:

When deleting an email domain through the UI, the API still returns it, but with `validationStatus = "DELETED"`.

When recreating the same email domain through the UI (same email address and sender name), the API will now return two entries with the same details (one `DELETED` and one not)

When Salto performs a fetch, we get

```
Error merging okta.EmailDomain.instance.Team_Rachum@s: duplicate key <somekey>
[..]
dropped elements: okta.EmailDomain.instance.Team_Rachum@s
```

We solve this by omitting entries with DELETED status on fetch.

---

---
_Release Notes_: 
_Okta adapter_:
Fixed a bug where email domains are dropped after being deleted and recreated, by omitting all currently-deleted email domains in fetch

---
_User Notifications_: 
_Okta adapter_:
- Email domains that have `validationStatus = "DELETED"` will be omitted in future fetches.

### DIFF
--- a/packages/okta-adapter/src/definitions/fetch/types/email_domain.ts
+++ b/packages/okta-adapter/src/definitions/fetch/types/email_domain.ts
@@ -1,0 +1,42 @@
+/*
+ *                      Copyright 2024 Salto Labs Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { Values } from '@salto-io/adapter-api'
+import Joi from 'joi'
+import { createSchemeGuard } from '@salto-io/adapter-utils'
+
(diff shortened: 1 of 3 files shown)


packages/okta-adapter/src/definitions/fetch/types/email_domain.ts
```diff
@@ -0,0 +1,42 @@
...
+type EmailDomain = {
+  validationStatus: string
+}
+
+const hasValidationStatus = createSchemeGuard<EmailDomain>(EMAIL_DOMAIN_SCHEMA)
+
+/**
+ * Omit email domains with a DELETED status.
+ *
+ * When deleting an email domain, Okta marks it as deleted instead of removing it. This can cause a merge bug when
```
shir-reifenberg, line 34:
```suggestion
 * When deleting an email domain, Okta marks it as deleted instead of removing it. This can cause a merge error when
```

packages/okta-adapter/src/definitions/fetch/types/email_domain.ts
```diff
@@ -0,0 +1,42 @@
...
+ * recreating the email domain with the same name.
+ */
+export const isNotDeletedEmailDomain = (value: unknown): value is Values => {
+  if (hasValidationStatus(value)) {
+    return value.validationStatus !== 'DELETED'
+  }
+  return false
```
shir-reifenberg, line 41: maybe add a log to indicate value was removed because it didn't have `validationStatus`

shir-reifenberg (approved):
👯‍♀️

PR #6421, opened 2024-08-16
SALTO-6490: support group references change in automations

IdoZakk:
Following a change in the service, the group references are now IDs instead of names. Keeping both in case they revert

---

[Noise reduction](https://github.com/salto-io/salto_private/pull/9118)
None

---
_Release Notes_: 
Jira Adapter:
* Support Atlassian change in automation's email recipients from group name to group id
---
_User Notifications_: 
Jira Adapter:
* Automations with email group recipients will return to be references


omrilit (approved):
Thanks!

PR #6423, opened 2024-08-18
SALTO-6492: fix fieldConfigurationDescriptionLengthValidator bug

shayc331:
_fix fieldConfigurationDescriptionLengthValidator bug_

---

* the bug is that the validator fails when there is no description

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


SeggevHaimovich (approved):
:100:

PR #6425, opened 2024-08-18
SALTO-6424: (Jira) Support unique field checks within parent scope

SeggevHaimovich:
Change the unique fields CV to support uniqueness in different scopes.
We will support global and parent scopes.

Added types that need the parent scope check in Jira.

---

_None_

---
_Release Notes_: 
_Adapter Components:_
* Added a scope option in the unique fields CV

_Jira adapter:_
* Check unique names in SLAs and field contexts

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/adapter-components/src/deployment/change_validators/unique_fields.ts
+++ b/packages/adapter-components/src/deployment/change_validators/unique_fields.ts
@@ -107,7 +107,7 @@ const getErrorForChange = (
     elemID: instance.elemID,
     severity: 'Error',
     message: `The ${fieldNames.length > 1 ? 'fields' : 'field'} ${fieldNames.map(str => `'${str}'`).join(', ')} in type ${instance.elemID.typeName} must have ${fieldNames.length > 1 ? 'unique values' : 'a unique value'}`,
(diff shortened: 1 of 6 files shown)


packages/adapter-components/src/deployment/change_validators/unique_fields.ts
```diff
@@ -93,24 +107,28 @@ const getErrorForChange = (
     elemID: instance.elemID,
     severity: 'Error',
     message: `The ${fieldNames.length > 1 ? 'fields' : 'field'} ${fieldNames.map(str => `'${str}'`).join(', ')} in type ${instance.elemID.typeName} must have ${fieldNames.length > 1 ? 'unique values' : 'a unique value'}`,
-    detailedMessage: `This instance cannot be deployed due to non unique values in the following fields: ${nonUniqueFieldValues.map(str => `'${str}'`).join(', ')}.`,
+    detailedMessage: `This instance cannot be deployed due to non unique values, within the ${scope} scope, in the following fields: ${nonUniqueFieldValues.map(str => `'${str}'`).join(', ')}.`,
```
IdoZakk, line 110: The scope will not be understood by the users. We can consider a different message to the parent scope, or even the same message for both

IdoZakk (approved):
👍 Nice. Please fix the error message

PR #6427, opened 2024-08-18
Add docs to `RecurseIntoCondition`

Nurdok:
Just adding some comments to document the usage.

---

Inspired by #6426 

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/fetch/resource/subresources.ts
+++ b/packages/adapter-components/src/fetch/resource/subresources.ts
@@ -39,8 +39,8 @@ const extractRecurseIntoContext = async (
 }
 
 export type RecurseIntoConditionBase = {
-  // `match` strings are regex patterns that are matched against the tested value with search semantics with OR between them.
-// To express AND semantics, use multiple conditions instead.
+  // `match` strings are regex patterns that are matched against the tested value with search semantics.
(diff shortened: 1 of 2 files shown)


packages/adapter-components/src/fetch/resource/subresources.ts
```diff
@@ -38,7 +38,13 @@ const extractRecurseIntoContext = async (
   return context
 }
 
-export type RecurseIntoConditionBase = { match: string[] }
+export type RecurseIntoConditionBase = {
+  // `match` strings are regex patterns that are matched against the tested value with search semantics (i.e., partial
+  // matches are allowed). To match the entire value, use `^` and `$` in the pattern.
+  // These patterns have OR semantics (if any of the patterns match, the condition is considered to be met). To
+  // express AND semantics, use multiple conditions instead.
```
shir-reifenberg, line 45: I think we usually try to make comments relatively consie
so I think in this case there's not need to mention specifc regex patterns (like `^`) or explain what `OR` means

WDYT about something like this ?
```suggestion
  // `match` strings are regex patterns that are matched against the tested value with search semantics with OR between them.
// To express AND semantics, use multiple conditions instead.
```

shir-reifenberg (approved):
Thanks 💪

PR #6428, opened 2024-08-18
SALTO-2392 Improve condition formula parsing for ASVs

dantal4:
Improve the condition formula parsing for resolving ASVs in workflows:

1. extract only non-zero integers as internal ids.
2. deal with formulas that have the same param/s more than once.

---

---
_Release Notes_: 
Netsuite Adapter:
- Improve condition formula parsing for ASVs

---
_User Notifications_: 
Netsuite Adapter:
- unresolved `[ACCOUNT_SPECIFIC_VALUE]` values may be resolved


omrilit (approved):
Great work!

PR #6430, opened 2024-08-18
SALTO-6430: [Intune] Support fetch & deploy for applications

shoshana-houri:
Fetch Intune applications & deploy with the necessary basic adjustments / validations.

---
_Release Notes_: 
_Microsoft Security Adapter:_
* Support fetch & deploy for Intune applications

---
_User Notifications_: 
_Microsoft Security Adapter:_
* Support fetch & deploy for Intune applications


packages/microsoft-security-adapter/test/definitions/utils/intune/application.test.ts
```diff
@@ -0,0 +1,60 @@
...
+describe('Intune application fetch & deploy utils', () => {
+  describe(`${isApplicationOfType.name}`, () => {
+    const possibleTypes: PossibleAppType[] = [
+      'androidManagedStoreApp',
+      'androidStoreApp',
+      'iosStoreApp',
+      'managedAndroidStoreApp',
+      'managedIosStoreApp',
+    ]
+    it.each(possibleTypes)(
```
shir-reifenberg, line 22: Like 🔥

packages/microsoft-security-adapter/src/definitions/fetch/shared/utils/field_customizations.ts
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * CERTAIN THIRD PARTY SOFTWARE MAY BE CONTAINED IN PORTIONS OF THE SOFTWARE. See NOTICE FILE AT https://github.com/salto-io/salto/blob/main/NOTICES
+ */
+
+import { ElementFetchDefinition } from '../types'
+
+/**
+ * Convert a list of fields to a fieldCustomizations object with the given action
+ */
+const toOmittedOrHiddenFields = (
```
shir-reifenberg, line 14: TBH I think those utils are more complex than just writing the definitions and I prefer having the customizations explicitly (also I think it's not that of an effort to write...)

packages/microsoft-security-adapter/src/definitions/fetch/intune/fetch.ts
```diff
@@ -0,0 +1,58 @@
...
+        elemID: {
+          parts: [APPLICATION_TYPE_PART, ...APPLICATION_NAME_PARTS],
+        },
+        path: {
+          pathParts: [{ parts: [APPLICATION_TYPE_PART] }, { parts: APPLICATION_NAME_PARTS }],
+        },
+      },
+      fieldCustomizations: {
+        ...ID_FIELD_TO_HIDE,
+        ...toOmittedFields(APPLICATION_OMITTED_FIELDS),
```
shir-reifenberg, line 51: If you want to avoid writing it explicitly I prefer something like this (instead of tracking what the functions are doing..)

```suggestion
        ...APPLICATION_OMITTED_FIELDS.map(fieldName => { fieldName: { omit: true } }
```

packages/microsoft-security-adapter/src/definitions/fetch/shared/utils/adjust_utils.ts
```diff
@@ -0,0 +1,32 @@
...
+import { validatePlainObject } from '@salto-io/adapter-utils'
+import { AdjustFunctionSingle as AdjustFunctionSingleFetch } from '../types'
+
+const { awu } = collections.asynciterable
+
+/**
+ * Concatenates multiple adjust functions into a single adjust function.
+ * The adjust functions will be applied in the order they are provided.
+ */
+export const concatAdjustFunctions =
```
shir-reifenberg, line 18: Can we reuse the confluence implementation `createAdjustFunctionFromMultipleFunctions` ? (and move to adapter components 🙏)

packages/microsoft-security-adapter/src/definitions/fetch/intune/fetch.ts
```diff
@@ -0,0 +1,58 @@
...
+import { transformOdataTypeField } from '../../utils/shared'
+import { APPLICATION_OMITTED_FIELDS } from './constants'
+import { concatAdjustFunctions, createCustomizationsWithBasePathForFetch, toOmittedFields } from '../shared/utils'
+import { APPLICATION_NAME_PARTS, APPLICATION_TYPE_PART, omitApplicationRedundantFields } from './utils'
+
+const { APPLICATION_TYPE_NAME } = intuneConstants
+
+const graphBetaCustomizations: FetchCustomizations = {
+  [APPLICATION_TYPE_NAME]: {
+    resource: {
```
shir-reifenberg, line 23: aren't we missing a service id ? 
also  if it's a minimum effort, can you please also add alias, and service url ?

packages/microsoft-security-adapter/src/definitions/fetch/intune/fetch.ts
```diff
@@ -0,0 +1,58 @@
...
+    ],
+    element: {
+      topLevel: {
+        isTopLevel: true,
+        elemID: {
+          parts: [APPLICATION_TYPE_PART, ...APPLICATION_NAME_PARTS],
+        },
+        path: {
+          pathParts: [{ parts: [APPLICATION_TYPE_PART] }, { parts: APPLICATION_NAME_PARTS }],
+        },
```
shir-reifenberg, line 47: you added the `APPLICATION_TYPE_PART` so each type would have its own folder, right ?
but is this `{ parts: APPLICATION_NAME_PARTS }` actually needed? isn't the last "part" is taken from the `elemID.name` ?

packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/application.ts
```diff
@@ -0,0 +1,43 @@
...
+import { APP_STORE_URL_FIELD_NAME, APPLICATION_TYPE_NAME, PACKAGE_ID_FIELD_NAME } from '../../../../constants/intune'
+import { NAME_ID_FIELD } from '../../shared/defaults'
+import { getAdjustedOdataTypeFieldName } from '../../../utils/shared'
+
+/* The following parts are shared for the application elemID definition and its path definition */
+export const APPLICATION_TYPE_PART: definitions.fetch.FieldIDPart = {
+  fieldName: getAdjustedOdataTypeFieldName(APPLICATION_TYPE_NAME),
+}
+export const APPLICATION_NAME_PARTS: definitions.fetch.FieldIDPart[] = [
+  // Some ManagedAndroidStoreApp instances have the same name, so instead we use the packageId field,
```
shir-reifenberg, line 23: If `ManagedAndroidStoreApp` might have the same name, shouldn't we always have `name` and add `PACKAGE_ID_FIELD_NAME` just when needed? what a I missing ?

packages/microsoft-security-adapter/test/adapter.test.ts
```diff
@@ -84,8 +84,40 @@ describe('Microsoft Security adapter', () => {
...
+          'intune_application',
         ])
-        // TODO: Validate sub-types and structure of the elements
+        // TODO: Validate Entra sub-types and structure of the elements
+
+        /* Specific instances: */
+        const instances = elements.filter(isInstanceElement)
+
+        // Intune applications
+        const intuneApplications = instances.filter(e => e.elemID.typeName === 'intune_application')
```
shir-reifenberg, line 95: maybe do it in a separate `it` block ?

packages/microsoft-security-adapter/src/definitions/utils/shared/odata_type.ts
```diff
@@ -0,0 +1,61 @@
...
+import { AdjustFunctionSingle } from '../../fetch/shared/types'
+import { ODATA_PREFIX, ODATA_TYPE_FIELD } from '../../../constants'
+
+const log = logger(module)
+
+const extractTypeFromODataTypeField = (odataTypeField: string): string | undefined => {
+  // The odata type field is in the format of #microsoft.graph.{type}
+  const typeParts = odataTypeField.split('.')
+  if (typeParts.length < 3) {
+    log.warn(`Failed to extract type from odataTypeField: ${odataTypeField}`)
```
shir-reifenberg, line 21: nit: I think it would be useful for log to have the typeName here as well

packages/microsoft-security-adapter/src/definitions/utils/shared/odata_type.ts
```diff
@@ -0,0 +1,61 @@
...
+ */
+export const getAdjustedOdataTypeFieldName = (typeName: string): string => `${typeName}_type`
+
+/**
+ *  The OData type field is in the format of #microsoft.graph.{type} as received from the service,
+ *  and the field name is _odata_type@mv when converted to NACL case.
+ *  We want to transform these to more user-friendly values, and vice versa.
+ *  We don't simply omit this field since it's used for directory structure, deployment, etc.
+ */
+export const transformOdataTypeField: (operation: 'fetch' | 'deploy') => AdjustFunctionSingle =
```
shir-reifenberg, line 39: is this only for representation?

shir-reifenberg (commented):
Looks good 💪 
I added some comments, can also discuss IRL some of those :)

packages/microsoft-security-adapter/src/definitions/utils/intune/applications.ts
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * CERTAIN THIRD PARTY SOFTWARE MAY BE CONTAINED IN PORTIONS OF THE SOFTWARE. See NOTICE FILE AT https://github.com/salto-io/salto/blob/main/NOTICES
+ */
+
+import _ from 'lodash'
+import { Values } from '@salto-io/adapter-api'
+import { getAdjustedOdataTypeFieldName } from '../shared'
+import { APPLICATION_TYPE_NAME } from '../../../constants/intune'
+
+export type PossibleAppType =
```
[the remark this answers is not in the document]
shoshana-houri: tbh I couldn't find a proper list in their docs, only in the `openapi` file, which is to large to preview on github. I can still direct to it though with explicitly mentioning the line number 🤔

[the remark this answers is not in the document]
shoshana-houri: Also - according to the openapi I need to expand this list with some other types 😅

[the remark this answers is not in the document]
shoshana-houri: Giving it another thought - the full list of apps currently contains 40 different apps:
```
androidForWorkApp, androidManagedStoreApp, androidManagedStoreWebApp, androidStoreApp, iosiPadOSWebClip, iosStoreApp, iosVppApp, macOSMicrosoftDefenderApp, macOSMicrosoftEdgeApp, macOSOfficeSuiteApp, macOsVppApp, macOSWebClip, managedApp, managedAndroidStoreApp, managedIOSStoreApp, managedMobileLobApp, managedAndroidLobApp, managedIOSLobApp, microsoftStoreForBusinessApp, mobileLobApp, androidLobApp, iosLobApp, macOSDmgApp, macOSLobApp, macOSPkgApp, win32LobApp, win32CatalogApp, windowsAppX, windowsMobileMSI, windowsPhone81AppX, windowsPhone81AppXBundle, windowsPhoneXAP, windowsUniversalAppX, officeSuiteApp, webApp, windowsMicrosoftEdgeApp, windowsPhone81StoreApp, windowsStoreApp, windowsWebApp, winGetApp
```
ATM we only use `isApplicationOfType` for a single type. I originally thought that it'll be nice to generalize the validation, but I think it's ok to change it to `isXApplication` atm and only expand it in the future if/when we need it. WDYT?

packages/microsoft-security-adapter/src/definitions/utils/intune/applications.ts
```diff
@@ -0,0 +1,44 @@
...
+  | 'androidManagedStoreApp'
+  | 'androidStoreApp'
+  | 'iosStoreApp'
+  | 'managedAndroidStoreApp'
+  | 'managedIosStoreApp'
+
+const APPLICATION_TYPES: Record<PossibleAppType, string> = {
```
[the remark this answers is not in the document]
shoshana-houri: I just wanted it to be typed properly. We no longer use this mapping, but if you think I could've achieved it differently I would be happy to learn how :)

packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/application.ts
```diff
@@ -0,0 +1,43 @@
...
+  // which should be consistent across environments.
+  { fieldName: NAME_ID_FIELD.fieldName, condition: value => !value[PACKAGE_ID_FIELD_NAME] },
+  { fieldName: PACKAGE_ID_FIELD_NAME },
```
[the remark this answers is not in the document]
shoshana-houri: Good catch! we do omit it for a specific type of application 🤔 I will replace it with another identifier in this case, thanks!

packages/microsoft-security-adapter/src/definitions/deploy/intune/utils/application.ts
```diff
@@ -0,0 +1,29 @@
...
+import { inspect } from 'util'
+import { validatePlainObject } from '@salto-io/adapter-utils'
+import { AdjustFunctionSingle } from '../../shared/types'
+import { APP_IDENTIFIER_FIELD_NAME, APPLICATION_TYPE_NAME } from '../../../../constants/intune'
+
+/**
+ * Addition of an application with the field isSystemApp set to false is done
+ * by specifying the productIds field with the value app:<appId> and ignoring the rest of the fields.
+ * The non system apps installed apps that cannot be modified, so it's enough to specify their productIds field.
+ */
```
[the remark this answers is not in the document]
shoshana-houri: 1. I couldn't find any tbh :( I only found this endpoint by investigating their api calls.
2. They're not managed by the user - all these fields are determined by the app id that was added (I couldn't find a way to edit them via the ui)

packages/microsoft-security-adapter/src/definitions/fetch/intune/fetch.ts
shoshana-houri: The problem is that the mapping is less intuitive IMO, i,e, we need something like:
```suggestion
        ...APPLICATION_FIELDS_TO_OMIT.reduce((acc, fieldName) => ({ ...acc, [fieldName]: { omit: true } }), {}),
```
So, I thought that extracting it to a separate function with a straightforward name would actually make it a bit clearer. If it's confusing I'll just define them directly :)

packages/microsoft-security-adapter/src/definitions/fetch/shared/utils/adjust_utils.ts
shoshana-houri: Done, please LMK if you think I should locate the file some place else :)

packages/microsoft-security-adapter/src/definitions/fetch/intune/fetch.ts
shoshana-houri: It's 'id' by default (under `shared/deploy`). And sure :)

packages/microsoft-security-adapter/src/definitions/fetch/intune/fetch.ts
shoshana-houri: Sadly not 😬 When I only specified the `APPLICATION_TYPE_PART` the type was simply the file name and all the instances resided in the same file 🙈

packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/application.ts
shoshana-houri: I just thought it's too long 😬 wdyt?

packages/microsoft-security-adapter/src/definitions/utils/shared/odata_type.ts
shoshana-houri: Yes - both in the element and by using it in the path & elemID

packages/microsoft-security-adapter/src/definitions/utils/intune/applications.ts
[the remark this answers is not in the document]
shir-reifenberg: sounds good :)

packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/application.ts
shir-reifenberg: up too you, in most places the alias is what users see. if only the packageID makes it unique I'm ok with it :)

packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/application.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+export const APPLICATION_NAME_PARTS: definitions.fetch.FieldIDPart[] = [
+  // Some ManagedAndroidStoreApp instances have the same name, so instead we use the packageId field,
+  // which should be consistent across environments.
+  { fieldName: APP_IDENTIFIER_FIELD_NAME },
+  { fieldName: PACKAGE_ID_FIELD_NAME, condition: value => !value[APP_IDENTIFIER_FIELD_NAME] },
+  {
+    fieldName: NAME_ID_FIELD.fieldName,
+    condition: value => !value[APP_IDENTIFIER_FIELD_NAME] && !value[PACKAGE_ID_FIELD_NAME],
+  },
```
shir-reifenberg, line 49: nit: I think it would be useful to mention in a comment in what type of app we use each identifier
i.e something like (ignore the specific app name here)

```
  // identifier for managedAndroidApps
  {
    fieldName: NAME_ID_FIELD.fieldName,
    condition: value => !value[APP_IDENTIFIER_FIELD_NAME] && !value[PACKAGE_ID_FIELD_NAME],
  },
```

packages/microsoft-security-adapter/src/definitions/fetch/intune/fetch.ts
shir-reifenberg: got it, thanks!

packages/microsoft-security-adapter/src/definitions/deploy/intune/utils/application.ts
```diff
@@ -0,0 +1,29 @@
...
+export const transformNonSystemApp: AdjustFunctionSingle = async ({ value }) => {
```
shir-reifenberg, line 20: another question I have  😀 
This is currently only used for requests of `isManagedGooglePlayApp` - should we adjust the name & add a validation here? or you think this will apply in other cases as well ?

packages/microsoft-security-adapter/src/definitions/utils/intune/applications.ts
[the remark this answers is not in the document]
shir-reifenberg: I _think_ this can be done be defining something like 

```
const appTypes = ['ios', 'android'] as const
type AppType = (typeof appTypes)[number]

const isAppOfType = (value: Values): boolean => appTypes.includes(value[myField])

**Review by shir-reifenberg — APPROVED** [review]
NICE 🚢

File: packages/microsoft-security-adapter/src/definitions/utils/intune/applications.ts (hunk 12 above)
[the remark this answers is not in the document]
Comment (shir-reifenberg) [reply packages/microsoft-security-adapter/src/definitions/utils/intune/applications.ts]: I _think_ this can be done be defining something like 

```
const appTypes = ['ios', 'android'] as const
type AppType = (typeof appTypes)[number]

const isAppOfType = (value: Values): boolean => appTypes.includes(value[myField])

packages/microsoft-security-adapter/src/definitions/fetch/intune/fetch.ts
shir-reifenberg: got it, thanks!

packages/microsoft-security-adapter/src/definitions/deploy/intune/utils/application.ts
shoshana-houri: I tried creating a different types of apps in the service and that's the only one with a different endpoint, so I like your suggestion :)

PR #6432, opened 2024-08-19
SALTO-6491 Ignore FileCabinet errors on SDF validation

dantal4:
as described in SALTO-3016 we should include file cabinet changes as part of the SDF validation, but can we just ignore the validation errors that we have on them after the validation has ran.

---

---
_Release Notes_: 
Netsuite Adapter:
- Ignore FileCabinet errors on SDF validation

---
_User Notifications_: 
None


omrilit (approved):
Thanks for the quick implementation!

PR #6434, opened 2024-08-19
SALTO-5431: adjust definitions override logs

shir-reifenberg:
replaced `safeJsonStringify` with `inspect` cause we're logging all the `rateLimiters` config which contains a lot of informative data

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/definitions/system/overrides.ts
+++ b/packages/adapter-components/src/definitions/system/overrides.ts
@@ -7,7 +7,7 @@
  */
 import _ from 'lodash'
 import { Value, Values } from '@salto-io/adapter-api'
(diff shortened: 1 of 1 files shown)


netama (approved):
thanks!
LGTM - though I think in this case it might be better to add a replacer that doesn't log classes and other things that are not plain objects. but no harm in doing this :) (assuming 7 levels is the "deepest" real config goes right now?)

shir-reifenberg:
> thanks! LGTM - though I think in this case it might be better to add a replacer that doesn't log classes and other things that are not plain objects. but no harm in doing this :) (assuming 7 levels is the "deepest" real config goes right now?)

that's a better suggestion 😄 changed

PR #6435, opened 2024-08-19
SALTO-6501: Remove errors from logs

shir-reifenberg:

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/fetch/request/pagination/queue.ts
+++ b/packages/adapter-components/src/fetch/request/pagination/queue.ts
@@ -88,7 +88,7 @@ export class RequestQueue<ClientOptions extends string> {
       })
       nextArgs.forEach(arg => this.enqueue(arg))
     } catch (e) {
-      log.error('Error processing args (%s): %s, stack: %s', safeJsonStringify(args), safeJsonStringify(e), e.stack)
+      log.error('Error processing args (%s): stack: %s', safeJsonStringify(args), e.stack)
       throw e
     } finally {
       this.activePromises = this.activePromises.filter(p => p.id !== promiseID)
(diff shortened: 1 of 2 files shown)


packages/adapter-components/src/fetch/request/pagination/queue.ts
```diff
@@ -88,7 +88,7 @@ export class RequestQueue<ClientOptions extends string> {
       })
       nextArgs.forEach(arg => this.enqueue(arg))
     } catch (e) {
-      log.error('Error processing args (%s): %s, stack: %s', safeJsonStringify(args), safeJsonStringify(e), e.stack)
+      log.error('Error processing args (%s): stack: %s', safeJsonStringify(args), e.stack)
```
netama, line 91: we shouldn't remove it, we should revert to what it was in https://github.com/salto-io/salto/pull/6206
```suggestion
      log.error('Error processing args (%s): %s, stack: %s', safeJsonStringify(args), e, e.stack)
```

packages/adapter-components/src/fetch/resource/type_fetcher.ts
```diff
@@ -192,13 +191,7 @@ export const createTypeResourceFetcher = <ClientOptions extends string>({
...
       }
     } catch (e) {
-      log.error(
-        '[%s] Error caught while fetching %s: %s. stack: %s',
-        adapterName,
-        typeName,
-        safeJsonStringify(e),
-        (e as Error).stack,
-      )
+      log.error('[%s] Error caught while fetching %s: stack: %s', adapterName, typeName, (e as Error).stack)
```
netama, line 194: same
```suggestion
      log.error('[%s] Error caught while fetching %s: %s. stack: %s', adapterName, typeName, e, (e as Error).stack)
```

netama (approved):
thanks! added a comment but pre-approving

PR #6436, opened 2024-08-19
SALTO-6027 - Salesforce: create custom references from flow conditions

tomsellek:
Treat specific fields as containing formula-like identifiers and create custom references from them.
We can't use the reference mapper because we don't know a-priori the type of the referred types. We can't explicitly replace the field value with references because then we'd have to restore the original value before we deploy, and the field's content is a formula which we can't always construct from a reference expression. So we just create references directly in the reference cache using the custom references mechanism.

---

---
_Release Notes_: 
Salesforce: Will now identify more references from flows.

---
_User Notifications_: 
N/A


packages/salesforce-adapter/src/custom_references/handlers.ts
```diff
@@ -21,23 +20,27 @@ import {
...
   managedElements: managedElementsHandler,
   permisisonSets: permissionSetsHandler,
+  formulaRefs: formulaRefsHandler,
 }
 
 const defaultCustomReferencesConfiguration: Required<CustomReferencesSettings> = {
   profiles: true,
   managedElements: true,
   permisisonSets: true,
+  formulaRefs: false,
```
amitgoldarb, line 36: Why not on by default? Now that this is a custom reference it shouldn't be very dangerous right? What release mechanism did you imagine?

packages/salesforce-adapter/src/filters/formula_utils.ts
```diff
@@ -0,0 +1,101 @@
...
+      customObject: 'type',
+      customSetting: 'type',
+      standardField: 'field',
+      customField: 'field',
+      customMetadataTypeRecord: 'instance',
+      customLabel: 'instance',
+    }) as Record<IdentifierType, ElemIDType>
+  )[identifierInfo.type]
+
+export const referencesFromIdentifiers = async (typeInfos: FormulaIdentifierInfo[]): Promise<ElemID[]> =>
```
amitgoldarb, line 49: This doesn't have to be async right?

packages/salesforce-adapter/src/filters/formula_utils.ts
```diff
@@ -0,0 +1,101 @@
...
+    identifierInfo =>
+      new ElemID(
+        SALESFORCE,
+        naclCase(identifierTypeToElementType(identifierInfo)),
+        identifierTypeToElemIdType(identifierInfo),
+        ...identifierTypeToElementName(identifierInfo).map(naclCase),
+      ),
+  )
+
+const isValidReference = async (elemId: ElemID, allElements: ReadOnlyElementsSource): Promise<boolean> => {
```
amitgoldarb, line 60: If we load all elementIds once in the beginning, than this doesn't have to be async right?

packages/salesforce-adapter/src/custom_references/formula_refs.ts
```diff
@@ -0,0 +1,99 @@
...
+        return value
+      }
+      references.push(await referenceInfoFromFieldValue(instance, path, field, value, elements))
+      return value
+    }
+
+  const fetchedInstances = elements.filter(isInstanceOfTypeSync(...typesWithFieldsWithFormulaReferences))
+  await awu(fetchedInstances).forEach(async instance => {
+    instance.value =
+      (await transformValues({
```
amitgoldarb, line 85: I think we can remove the async down the line and then make this walkOn. It would probably be more efficeint.

packages/salesforce-adapter/src/custom_references/formula_refs.ts
```diff
@@ -0,0 +1,99 @@
...
+        transformFunc: transformInstanceFieldsToReference(instance),
+      })) ?? {}
+  })
+
+  return references.filter(isDefined)
+}
+
+export const formulaRefsHandler: WeakReferencesHandler = {
+  findWeakReferences,
+  removeWeakReferences: () => async () => ({ fixedElements: [], errors: [] }),
```
amitgoldarb, line 98: Can we open a ticket to track implementation of this one?

packages/salesforce-adapter/src/custom_references/formula_refs.ts
```diff
@@ -0,0 +1,99 @@
...
+    ref => ref.typeName !== apiNameSync(instance.getTypeSync()),
+  )
+
+  if (referencesToOtherTypes.length === 0) {
+    return undefined
+  }
+  return {
+    source: path,
+    target: referencesToOtherTypes[0],
+    type: 'weak',
```
amitgoldarb, line 59: I'm not sure this should be weak. @tamtamirr WDYT?

packages/salesforce-adapter/src/custom_references/formula_refs.ts
```diff
@@ -0,0 +1,99 @@
...
+const { isDefined } = values
+
+const typesWithFieldsWithFormulaReferences = ['Flow']
+
+const referenceFieldsWithFormulaIdentifiers: Record<string, string> = {
+  FlowCondition: 'leftValueReference',
+  FlowTestCondition: 'leftValueReference',
+  FlowTestParameter: 'leftValueReference',
+  FlowAssignmentItem: 'assignToReference',
+}
```
amitgoldarb, line 26: Can we organize this so that the types are closer to the fields? Isn't the format of the reference_mapper comfortable? We could reuse the "src" part.

amitgoldarb (commented):
Looks good! Added a few comments

packages/salesforce-adapter/src/filters/formula_utils.ts
tomsellek: Yes, but this function is used in the formula_deps filter as well. Loading everything in advance means increasing the memory consumption significantly in the quickfetch case, no?

packages/salesforce-adapter/src/custom_references/formula_refs.ts
tomsellek: https://salto-io.atlassian.net/browse/SALTO-6565

packages/salesforce-adapter/src/custom_references/formula_refs.ts
tomsellek: Having the top-level type allows us to not do any work for the 99% of elements that are irrelevant. I could make this record more complicated to turn it into something like:
```
{
  Flow: {
    FlowCondition: 'leftValueReference',
    FlowTestCondition: 'leftValueReference',
    ...
  }
}
```

Do you think this would be more readable?

amitgoldarb: Discussed f2f.

packages/salesforce-adapter/src/custom_references/handlers.ts
amitgoldarb: Let's wrap the critical part in try catch and see release with it on.

packages/salesforce-adapter/src/filters/formula_utils.ts
amitgoldarb: Discussed f2f, decided to load all elements once.

amitgoldarb (commented):
Discussed f2f, other than fixes we discussed, just need to decide on 'Strong'/'Weak'

packages/salesforce-adapter/src/custom_references/handlers.ts
tomsellek: ack

packages/salesforce-adapter/src/custom_references/formula_refs.ts
```diff
@@ -0,0 +1,171 @@
...
+  const referencesToOtherTypes = referencesWithValidity.valid.filter(
+    ref => ref.typeName !== apiNameSync(instance.getTypeSync()),
+  )
+
+  if (referencesToOtherTypes.length === 0) {
+    return undefined
+  }
+  return {
+    source: path,
+    target: referencesToOtherTypes[0],
```
amitgoldarb, line 51: Why are we only returning the first? Can't a formula reference a few types?

packages/salesforce-adapter/src/custom_references/formula_refs.ts
```diff
@@ -0,0 +1,171 @@
...
+  potentialReferenceTargets: Map<string, Element>,
+) => ReferenceInfo[]
+
+const referenceInfoFromFieldValue = (
+  instance: InstanceElement,
+  path: ElemID,
+  value: Value,
+  potentialReferenceTargets: Map<string, Element>,
+): ReferenceInfo | undefined => {
+  const topLevelParentInstanceElemId = instance.elemID
```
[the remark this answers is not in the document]
tomsellek: This is leftover from when the code handled nested types in `transformValues`. I will rename or remove.

packages/salesforce-adapter/src/custom_references/formula_refs.ts
tomsellek: Yeah, I suppose we could.

packages/salesforce-adapter/test/custom_references/formula_refs.test.ts
```diff
@@ -0,0 +1,208 @@
...
+                  'conditions',
+                  '0',
+                  'leftValueReference',
+                ),
+              ),
+            )
+            expect(refs[0].type).toEqual('strong')
+          })
+        })
+        describe('when the reference is not valid', () => {
```
amitgoldarb, line 61: What is not valid in this context?

packages/workspace/src/workspace/reference_indexes.ts
```diff
@@ -37,7 +37,7 @@ import { RemoteMap, RemoteMapEntry } from './remote_map'
 const log = logger(module)
 const { awu } = collections.asynciterable
 
-export const REFERENCE_INDEXES_VERSION = 11
+export const REFERENCE_INDEXES_VERSION = 12
```
amitgoldarb, line 40: If we're not turning this on by default. Then we actually don't need to bump this 😅

packages/salesforce-adapter/src/custom_references/formula_refs.ts
```diff
@@ -0,0 +1,131 @@
...
+
+type ReferenceExtractor = (instance: InstanceElement) => ReferenceInfo[]
+
+const referenceInfoFromFieldValue = (
+  instance: InstanceElement,
+  path: ElemID,
+  value: Value,
+): ReferenceInfo[] => {
+  const identifierInfo = parseFormulaIdentifier(value, instance.elemID.typeName)
+  const referenceElemIds = referencesFromIdentifiers(identifierInfo)
```
amitgoldarb, line 25: Can we still wrap this with a try catch?

packages/workspace/src/workspace/reference_indexes.ts
tomsellek: Ack.

PR #6440, opened 2024-08-20
Revert "SALTO-6026: Identify more semantic references and commented serviceIds will not be created as generated dependencies"

edenhassid:
Reverts salto-io/salto#6340


dantal4 (approved):
Thanks!

PR #6441, opened 2024-08-20
SALTO-6509 Add rename capability to transformation definitions

netama:
Add `rename` capability to the transformation definitions.

Adding it before setting the `root`, so that it will be possible to nest neighboring fields under the field that is going to be used as the root.

---

_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/definitions/system/shared/transformation.ts
+++ b/packages/adapter-components/src/definitions/system/shared/transformation.ts
@@ -51,10 +51,17 @@ export type AdjustFunction<TContext = ContextParams, TSourceVal = unknown, TTarg
   | AdjustFunctionSingle<TContext, TSourceVal, TTargetVal>
   | AdjustFunctionMulti<TContext, TSourceVal, TTargetVal>
 
+// note: "from" and "to" can contain nested paths (e.g. a.b.c)
 export type TransformationRenameDefinition = {
   from: string
   to: string
+  // behavior when the target already exists:
+  // - override: move "from" to "to" and lose the old "to" value
+  // - skip: keep the value in "from" and do not override "to"
+  // - omit: lose the value in "from" and keep the value in "to"
+  onConflict: 'override' | 'skip' | 'omit'
 }
+
 /**
  * transformation steps:
  * - if rename is specified, each "from" path is moved to the corresponding "to" path. Notes:
--- a/packages/adapter-components/src/definitions/system/shared/transformation.ts
+++ b/packages/adapter-components/src/definitions/system/shared/transformation.ts
@@ -51,8 +51,15 @@ export type AdjustFunction<TContext = ContextParams, TSourceVal = unknown, TTarg
   | AdjustFunctionSingle<TContext, TSourceVal, TTargetVal>
   | AdjustFunctionMulti<TContext, TSourceVal, TTargetVal>
 
+export type TransformationRenameDefinition = {
+  from: string
+  to: string
+}
 /**
  * transformation steps:
+ * - if rename is specified, each "from" path is moved to the corresponding "to" path. Notes:
+ *     - renames are done in order, so the "from" values should be based on the most recent state
+ *     - existing values will be overwritten
  * - if root is specified, look at the value under that path instead of the entire object
  * - if pick is specified, pick the specified paths
  * - if omit is specified, omit the specified paths
@@ -63,7 +70,7 @@ export type AdjustFunction<TContext = ContextParams, TSourceVal = unknown, TTarg
  *   (and undefined will be returned if the array is empty). if single is false, the result will be returned as-is.
  */
 export type TransformDefinition<TContext = ContextParams, TTargetVal = Values | Value> = {
-  // return field name (can customize e.g. to "type => types")
+  rename?: TransformationRenameDefinition[]
   root?: string
   pick?: string[]
   omit?: string[]
--- a/packages/adapter-components/src/fetch/utils.ts
+++ b/packages/adapter-components/src/fetch/utils.ts
@@ -7,6 +7,7 @@
  */
 import _ from 'lodash'
 import { Values } from '@salto-io/adapter-api'
+import { safeJsonStringify } from '@salto-io/adapter-utils'
 import { logger } from '@salto-io/logging'
 import { collections, values as lowerdashValues } from '@salto-io/lowerdash'
 import {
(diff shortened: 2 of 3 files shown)


packages/adapter-components/src/definitions/system/shared/transformation.ts
```diff
@@ -51,8 +51,15 @@ export type AdjustFunction<TContext = ContextParams, TSourceVal = unknown, TTarg
   | AdjustFunctionSingle<TContext, TSourceVal, TTargetVal>
   | AdjustFunctionMulti<TContext, TSourceVal, TTargetVal>
 
+export type TransformationRenameDefinition = {
+  from: string
+  to: string
```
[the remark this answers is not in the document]
shir-reifenberg: ok I later saw in tests you have example with `field.nestedField` 😅  maybe mention in comment a path can be provided using `.` ?

shir-reifenberg (approved):
🚀

packages/adapter-components/src/definitions/system/shared/transformation.ts
```diff
@@ -51,8 +51,15 @@ export type AdjustFunction<TContext = ContextParams, TSourceVal = unknown, TTarg
...
+}
 /**
  * transformation steps:
```
Nurdok, line 59: A general thought, not for this PR: instead of specifying a strict order, it could be more flexible by allowing a list of `TransformationDefinition` which would `OneOf` all of the current fields. That would allow to perform `rename` after, say, `root`, if needed. Specifically I've had some cases where I needed to read a field in `adjust` and then `omit` it, but I had to do the `omit` manually because it runs before `adjust`.

packages/adapter-components/src/fetch/utils.ts
```diff
@@ -9,22 +9,44 @@ import _ from 'lodash'
...
 
 const { awu } = collections.asynciterable
 const log = logger(module)
 
+const getObjectWithRenamedFields = (value: Values, renameDefs: TransformationRenameDefinition[]): Values => {
+  const newVal = _.cloneDeep(value)
+  renameDefs.forEach(({ from, to }) => {
+    if (_.get(newVal, to) !== undefined) {
+      log.warn('rename transformation results in overriding value in path %s', to)
+    }
```
Nurdok, line 28: This seems a bit dangerous if the override was a mistake (or the service happens to add a new field name which we use in `rename`). Maybe `rename` should have a flag `allowOverride` which defaults to `false`?

Nurdok (approved):

packages/adapter-components/src/fetch/utils.ts
```diff
@@ -9,22 +9,44 @@ import _ from 'lodash'
...
 import { DATA_FIELD_ENTIRE_OBJECT } from '../definitions'
 
 const { awu } = collections.asynciterable
 const log = logger(module)
 
+const getObjectWithRenamedFields = (value: Values, renameDefs: TransformationRenameDefinition[]): Values => {
+  const newVal = _.cloneDeep(value)
+  renameDefs.forEach(({ from, to }) => {
+    if (_.get(newVal, to) !== undefined) {
+      log.warn('rename transformation results in overriding value in path %s', to)
```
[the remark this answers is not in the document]
netama: I didn't want to pass it in all the way, will add it later if we see a need / if we see we use these for "generic" field names that repeat a lot

packages/adapter-components/src/definitions/system/shared/transformation.ts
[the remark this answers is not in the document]
netama: will add a comment (it's the same as for all the other similar paths, elem id field names etc)

packages/adapter-components/src/definitions/system/shared/transformation.ts
netama: that can be a nice extension :) can either be `OneOf` or converting the transformations into an array of these, need to think a bit through the use cases. if you have a specific example you needed can you please send it to me (slack)?

packages/adapter-components/src/fetch/utils.ts
netama: I started with the opposite of not doing the rename but realized it might be worse for downstream code relying on it. I'll add the flag, but I think it should default to true - so I'll add it without a default for now so devs would need to make the decision, wdyt?

packages/adapter-components/src/fetch/utils.ts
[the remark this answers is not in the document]
netama: added the full definition for now, hopefully that will be indicative enough but if not we can add more details

packages/adapter-components/src/fetch/utils.ts
Nurdok: LGTM! Forcing developers to make a conscious decision is much better than a default here, thanks!

packages/adapter-components/src/definitions/system/shared/transformation.ts
Nurdok: Discussed in Slack, that case is actually solved by `rename` so at least for now I don't see a strong reason to add this flexibility.

Nurdok (approved):

shoshana-houri:
I understand the reasoning in the description for allowing renaming before extracting the root, but I think it's not very user-friendly. My intuition was that it would be applied to each of the elements (same as omit, hide, etc.), and it took me a debugging session to figure out what was wrong.

Moreover, even if the usage was clear, having to specify the root both in the 'from' and 'to' paths when renaming a subfield seems redundant. Instead, we could introduce a flag like `useRootPrefix`. If false, it would directly use the `from` and `to` paths as provided, and if true, it would prepend the root path already specified in the configuration.

wdyt? @netama @shir-reifenberg @Nurdok

Nurdok:
> I understand the reasoning in the description for allowing renaming before extracting the root, but I think it's not very user-friendly. My intuition was that it would be applied to each of the elements (same as omit, hide, etc.), and it took me a debugging session to figure out what was wrong.
> 
> Moreover, even if the usage was clear, having to specify the root both in the 'from' and 'to' paths when renaming a subfield seems redundant. Instead, we could introduce a flag like `useRootPrefix`. If false, it would directly use the `from` and `to` paths as provided, and if true, it would prepend the root path already specified in the configuration.
> 
> wdyt? @netama @shir-reifenberg @Nurdok
> 
> Edited: TBH the current definition simply doesn't work when we receive an array of values under some root, which is a common use-case. This means that we can't use this renaming capability for a single element in this case. It’s not a blocker since we can always use ‘adjust’, I just think it’s a nice to have :)

Sounds like this is pushing us in the direction of flexible ordering :) I would be happy to work on this if priorities allow.

Assuming we're still using strict ordering for now - 

"Simple things should be simple, complex things should be possible" - if the common use case is renaming things under `root`, I'm +1 for putting `rename` after `root`. We can always use `adjust` for complex transformations like the one in the PR description.

@netama is there an existing use case for `rename` before `root` or is this justification hypothetical?

PR #6443, opened 2024-08-20
Fix saved search query filters

dantal4:
If there are no items to query using "anyof" filter in saved search- return empty result.

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


omrilit (approved):
Looks good!

PR #6444, opened 2024-08-21
SALTO-2392 Fix bug in workflow ASV resolving

dantal4:
An error was thrown because we ran `string.find()` instead of `string[].find()`.

---

---
_Release Notes_: 
Netsuite Adapter:
- Fix bug in workflow ASV resolving

---
_User Notifications_: 
None


shir-reifenberg (approved):
🚢 
can you please add the ticket number to the PR title?

PR #6452, opened 2024-08-21
SALTO-6512: fix Jira DC ruleScope bug

shayc331:
_fix Jira DC ruleScope bug_

---

* `convertRuleScopeValueToProjects` function was meant to be used in Cloud only but we were using it DC as well

---
_Release Notes_: 

---
_User Notifications_: 


IdoZakk (approved):
💯

PR #6456, opened 2024-08-22
SALTO-6432: [Intune] Support fetch & deploy for Application Configuration [Managed Apps]

shoshana-houri:
Support fetching & deploying `Application Configuration - Managed Apps`.
Including - settings and target apps.
image

---
_Release Notes_: 
_Microsoft Security Adapter:_
* Support fetch & deploy for Application Configuration [Managed Apps]

---
_User Notifications_: 
_Microsoft Security Adapter:_
* Support fetch & deploy for Application Configuration [Managed Apps]


packages/microsoft-security-adapter/src/definitions/deploy/intune/utils/apps_configuration.ts
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Definition utils for application configuration deployment
+ *
+ * The application configuration definition includes targeted apps, which are deployed separately from the main configuration.
+ * This module provides utils for handling the targeted apps field by utilizing a custom 'targetApps' action.
+ *
+ */
+
+export const modificationCondition: definitionUtils.deploy.DeployRequestCondition = {
```
shir-reifenberg, line 28: nit: if it's short and not shared by other definitions let's have it inline for readability

packages/microsoft-security-adapter/src/definitions/deploy/intune/deploy.ts
```diff
@@ -84,6 +87,59 @@ const graphBetaCustomDefinitions: DeployCustomDefinitions = {
...
+          {
+            request: {
+              endpoint: {
+                path: '/deviceAppManagement/targetedManagedAppConfigurations/{id}',
+                method: 'delete',
+              },
+            },
+          },
+        ],
+        targetApps: [
```
shir-reifenberg, line 124: Why did you choose to use action dependencies instead of another request to the `add` and `modify` definition?
I think `actionDependencies` adds a level of complexity, and if it's just an additional request that is specific to a single type, i'm not sure it's the right choice
LMK WDYT :)

packages/microsoft-security-adapter/src/definitions/deploy/intune/utils/apps_configuration.ts
```diff
@@ -0,0 +1,74 @@
...
+  custom:
+    () =>
+    ({ change }) => {
+      const changeData = getChangeData(change)
+      const targetedAppType = changeData.value.targetedManagedAppGroupType
+      // Turns out (not documented AFAICT) that changing the apps list when targetedManagedAppGroupType
+      // is different than 'selectedPublicApps' will silently set it to 'selectedPublicApps'.
+      // We prioritize this field definition assuming that the apps list may contain inconsistencies
+      // (hence changes) between envs when they have different applications instances.
+      // TODO SALTO-6432: Warn the user when this happens.
```
shir-reifenberg, line 44: I think it's the wrong ticket number

packages/microsoft-security-adapter/src/definitions/deploy/intune/utils/apps_configuration.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+export const targetAppsChangeCondition: definitionUtils.deploy.DeployRequestCondition = {
+  custom:
+    () =>
+    ({ change }) => {
+      const changeData = getChangeData(change)
+      const targetedAppType = changeData.value.targetedManagedAppGroupType
+      // Turns out (not documented AFAICT) that changing the apps list when targetedManagedAppGroupType
+      // is different than 'selectedPublicApps' will silently set it to 'selectedPublicApps'.
```
shir-reifenberg, line 41: not sure I follow 😅  both fields are references to apps? what's the difference? why do we need both ?

packages/microsoft-security-adapter/src/definitions/references/intune_reference_rules.ts
```diff
@@ -0,0 +1,46 @@
...
+  {
+    src: {
+      field: 'bundleId',
+      parentTypes: [
+        recursiveNestedTypeName(APPLICATION_CONFIGURATION_MANAGED_APP_APPS_TYPE_NAME, 'mobileAppIdentifier'),
+      ],
+    },
+    target: { type: APPLICATION_TYPE_NAME },
+    serializationStrategy: 'bundleId',
+    // TODO SALTO-6527: This definition doesn't actually work here because it only transforms the source value to lowercase,
```
shir-reifenberg, line 31: Can you please add a more concise comment in the spirit of (we can add any extra details in the ticket)

```suggestion
    // TODO SALTO-6527 partial coverage,
```

packages/microsoft-security-adapter/src/definitions/references/intune_reference_rules.ts
```diff
@@ -0,0 +1,46 @@
...
+const { APPLICATION_TYPE_NAME, APPLICATION_CONFIGURATION_MANAGED_APP_APPS_TYPE_NAME } = intuneConstants
+
+const { recursiveNestedTypeName } = fetchUtils.element
+
+export const REFERENCE_RULES: referenceUtils.FieldReferenceDefinition<
+  ReferenceContextStrategies,
+  CustomReferenceSerializationStrategyName
+>[] = [
+  // According to the openapi (beta) doc, the possible values for the mobileAppIdentifier field are: bundleId, packageId and windowsAppId.
+  // However, in practice, only bundleId and packageId are used, and the windowsAppId is not referenced in the doc, and does not appear in V1 /the UI.
```
shir-reifenberg, line 21: I think this comment is redundant ..

shir-reifenberg (commented):
Looks good! Added a suggestion that I think that might simplify the deploy definitions, rest are mostly nits :)

packages/microsoft-security-adapter/src/definitions/references/references.ts
```diff
@@ -8,18 +8,42 @@
...
+const fieldsToGroupBy: Record<'id' | CustomReferenceSerializationStrategyName, null> = {
+  id: null,
+  appId: null,
+  bundleId: null,
+  packageId: null,
+}
 
 export const REFERENCES: definitions.ApiDefinitions<Options>['references'] = {
   rules: REFERENCE_RULES,
-  fieldsToGroupBy: ['id', 'name', 'appId'],
```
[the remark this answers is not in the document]
shoshana-houri: Ohh tbh it wasn't intentional, but it's indeed not being used 😅

packages/microsoft-security-adapter/test/adapter.test.ts
```diff
@@ -87,47 +87,90 @@ describe('Microsoft Security adapter', () => {
           'entra_roleDefinition',
           'entra_servicePrincipal',
           'intune_application',
+          'intune_applicationConfigurationManagedApp',
```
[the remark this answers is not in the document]
shoshana-houri: TBH it was done on purpose on my side, I thought it differentiated them better 😬 
I don't have a strong objection though if you think we should change that. As for the snake_case suggestion - I would prefer not to, since the camelCase/PascalCase aligns with the Graph API/MS DSC.
wdyt?

packages/microsoft-security-adapter/src/definitions/deploy/intune/deploy.ts
```diff
@@ -23,7 +24,9 @@ const graphBetaCustomDefinitions: DeployCustomDefinitions = {
       default: {
         request: {
           transformation: {
-            adjust: omitReadOnlyFieldsWrapper(transformOdataTypeField('deploy')),
+            adjust: omitReadOnlyFieldsWrapper(
+              concatAdjustFunctions(transformOdataTypeField('deploy'), applications.omitApplicationRedundantFields),
```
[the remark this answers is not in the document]
shoshana-houri: Exactly—sorry, I forgot to add a comment here in advance. Since we use the `packageId` for reference purposes, I thought it might cause issues if we tried to omit it post-fetch. Instead, I omitted them pre-deploy.

packages/microsoft-security-adapter/src/definitions/deploy/intune/deploy.ts
shoshana-houri: TBH, I just liked the way you guys (cc @Nurdok) implemented the `simple_status` field in Okta. I thought it was clearer than adding the requests twice for addition and modification, as is currently done for applications in Entra. Let me know if I’m missing something, but couldn’t you have applied the same approach you’re suggesting here—just adding it as another requests (with matching conditions) for both addition and modification?

packages/microsoft-security-adapter/src/definitions/deploy/intune/utils/apps_configuration.ts
shoshana-houri: Not exactly. The apps field stores references to the targeted apps, while `targetedManagedAppGroupType` is an enum indicating the type of targeting being used. The only scenario where the user manually selects the apps is `selectedPublicApps`.

If `targetedManagedAppGroupType` is set to anything other than `selectedPublicApps` and we still manually update the apps list, it will silently change the `targetedManagedAppGroupType` value to `selectedPublicApps`, which is bad :)

packages/microsoft-security-adapter/src/definitions/deploy/intune/deploy.ts
shir-reifenberg: mm IMO it's not exactly the use case, the `simple_status` is a pattern being repeated in almost every type in Okta (unlike this case). In addition, we needed the flexibility given by `actionDependencies` as in some cases the order of those operations varies between types.

I think when there's a simpler alternative, we should prioritize it other `actionDependencies` (and IMO it's also more readable to have it twice, can extract to a const if needed).

packages/microsoft-security-adapter/test/adapter.test.ts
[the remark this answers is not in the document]
shir-reifenberg: don't have a really strong opinion re snake case vs pascal case, but I think the mix is strange.. can you please open a ticket for this? 🙏

packages/microsoft-security-adapter/src/definitions/references/references.ts
```diff
@@ -8,18 +8,42 @@
...
-import { REFERENCE_RULES } from './entra_reference_rules'
+import { CustomReferenceSerializationStrategyName, Options } from '../types'
+import { REFERENCE_RULES as EntraReferenceRules } from './entra_reference_rules'
+import { REFERENCE_RULES as IntuneReferenceRules } from './intune_reference_rules'
+
+const REFERENCE_RULES = [...EntraReferenceRules, ...IntuneReferenceRules]
+
+// We only use this record as a TS 'hack' to fail the build if we add a custom serialization strategy
+// and forget to add it to the fieldsToGroupBy.
+const fieldsToGroupBy: Record<'id' | CustomReferenceSerializationStrategyName, null> = {
```
shir-reifenberg, line 19: I think this should be 
```suggestion
const fieldsToGroupBy: Record<ReferenceIndexField | CustomReferenceSerializationStrategyName, null> = {
```

this way if we add something to adapter-components this map will have to be updated as well

packages/microsoft-security-adapter/src/definitions/types.ts
```diff
@@ -8,11 +8,11 @@
 import { definitions, filterUtils } from '@salto-io/adapter-components'
 import { UserConfig } from '../config'
 
-export type AdditionalAction = never
+export type AdditionalAction = 'targetApps'
```
shir-reifenberg, line 11: please remove 🙏

packages/microsoft-security-adapter/src/definitions/deploy/intune/utils/apps_configuration.ts
```diff
@@ -0,0 +1,54 @@
...
+  custom:
+    () =>
+    ({ change }) => {
+      const changeData = getChangeData(change)
+      const targetedAppType = changeData.value.targetedManagedAppGroupType
+      // The field targetedManagedAppGroupType is an enum indicating the type of targeting being used for the apps.
+      // The only scenario where the user manually selects the apps is when targetedManagedAppGroupType is set to selectedPublicApps.
+      // Turns out (not documented AFAICT) that changing the apps list when targetedManagedAppGroupType
+      // value is different than 'selectedPublicApps' will silently set it to 'selectedPublicApps'.
+      // TODO SALTO-6528: Warn the user when this happens.
```
shir-reifenberg, line 32: can we please update to a more concise comment if most of this info is documented in the ticket ? 🙏

shir-reifenberg (approved):
Great job! 🚀

packages/microsoft-security-adapter/src/definitions/deploy/intune/utils/apps_configuration.ts
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Definition utils for application configuration deployment
+ *
+ * The application configuration definition includes targeted apps, which are deployed separately from the main configuration.
+ * This module provides utils for handling the targeted apps field by utilizing a custom 'targetApps' action.
+ *
+ */
+
+export const targetAppsChangeCondition: definitionUtils.deploy.DeployRequestCondition = {
```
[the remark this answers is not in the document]
shoshana-houri: I wanted to be able to test it separately :)

PR #6457, opened 2024-08-22
SALTO-6503: Add logs for visibility to `validateElementsAndDependents`

shir-reifenberg:
Added some logs for visibility to understand why we get out of memory error

---

Also replaced `...` with concat as a first attempt to resolve the issue

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/workspace/src/workspace/workspace.ts
+++ b/packages/workspace/src/workspace/workspace.ts
@@ -684,25 +684,33 @@ export const loadWorkspace = async (
 
     const updateWorkspace = async (envName: string): Promise<void> => {
       const source = naclFilesSource
-      const getElementsDependents = async (elemIDs: ElemID[], addedIDs: Set<string>): Promise<ElemID[]> => {
-        elemIDs.forEach(id => addedIDs.add(id.getFullName()))
-        const filesWithDependencies = _.uniq(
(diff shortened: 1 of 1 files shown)


packages/workspace/src/workspace/workspace.ts
```diff
@@ -684,25 +684,33 @@ export const loadWorkspace = async (
...
-          id => id.getFullName(),
+      const getElementsDependents = async (elemIDs: ElemID[], addedIDs: Set<string>): Promise<ElemID[]> =>
+        log.timeDebug(
+          async () => {
+            elemIDs.forEach(id => addedIDs.add(id.getFullName()))
+            const filesWithDependencies = _.uniq(
+              await awu(elemIDs)
+                .flatMap(id => source.getElementReferencedFiles(envName, id))
+                .toArray(),
+            )
```
dantal4, line 695: worth adding time log here too
```suggestion
            const filesWithDependencies =  await log.timeDebug(
              async () => _.uniq(
                await awu(elemIDs)
                  .flatMap(id => source.getElementReferencedFiles(envName, id))
                  .toArray(),
              ),
              'running getElementReferencedFiles for %s elemIds',
              elemIDs.length,
            )
```

PR #6458, opened 2024-08-22
SALTO-6481 Allow controlling how full responses are logged

netama:
Add some controls for truncating/omitting the full response logs (can extend later as needed, e.g. more granular control on truncating / better buckets by pattern).

This was added as part of the user-facing config for simplicity, but does not have a user-facing default.

Default behavior: responses over ~10kb will be truncated.

---

Also consolidated some code duplication (that missed past cleanups) from the zendesk and okta adapters.

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/client/http_client.ts
+++ b/packages/adapter-components/src/client/http_client.ts
@@ -7,7 +7,7 @@
  */
 import _ from 'lodash'
 import { ResponseType } from 'axios'
-import { safeJsonStringify } from '@salto-io/adapter-utils'
+import { inspectValue, safeJsonStringify } from '@salto-io/adapter-utils'
 import { values } from '@salto-io/lowerdash'
 import { Values } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
@@ -29,6 +29,7 @@ import {
 } from '../definitions/user/client_config'
 import { requiresLogin, logDecorator } from './decorators'
 import { throttle } from './rate_limit'
+import { createResponseLogFilter, FullResponseLogFilter } from './logging_utils'
 
 const log = logger(module)
 
@@ -121,6 +122,7 @@ export abstract class AdapterHTTPClient<TCredentials, TRateLimitConfig extends C
 {
   protected readonly conn: Connection<TCredentials>
   protected readonly credentials: TCredentials
+  protected responseLogFilter: FullResponseLogFilter
 
   constructor(
     clientName: string,
@@ -139,6 +141,7 @@ export abstract class AdapterHTTPClient<TCredentials, TRateLimitConfig extends C
       createConnection,
     })
     this.credentials = credentials
+    this.responseLogFilter = createResponseLogFilter(config?.logging)
   }
 
   protected async ensureLoggedIn(): Promise<void> {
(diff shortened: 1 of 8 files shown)


packages/adapter-components/src/client/http_client.ts
```diff
@@ -228,52 +231,78 @@ export abstract class AdapterHTTPClient<TCredentials, TRateLimitConfig extends C
...
-        response: Buffer.isBuffer(res.data)
-          ? `<omitted buffer of length ${res.data.length}>`
-          : this.clearValuesFromResponseData(res.data, url),
-        headers: this.extractHeaders(res.headers),
-        data: Buffer.isBuffer(data) ? `<omitted buffer of length ${data.length}>` : data,
-      })
-
-      if (error === undefined) {
+    if (error === undefined) {
+      const strategy = this.responseLogFilter({ responseText, url })
```
shir-reifenberg, line 270: the name of the filter is `responseLogFilter` but we apply it on the entire data being logged (including the data sent in case of `PUT` / `POST` , and the rest of the params).

IMO if we apply it on everything then I'm not sure why we need the extra nesting level of `responseStrategies` and not just call it `loggingStrategies` in the config.

If we think we might want a different strategies on response vs request then maybe just apply it on `response`? WDYT ?

packages/adapter-components/src/client/http_client.ts
```diff
@@ -228,52 +231,78 @@ export abstract class AdapterHTTPClient<TCredentials, TRateLimitConfig extends C
...
+          'Truncated HTTP response for %s on %s (original size %d, truncated dize %d): %s',
+          method.toUpperCase(),
+          url,
+          responseText.length,
+          truncatedResponseText.length,
+          truncatedResponseText,
+        )
       }
+    } else {
+      log.warn(`failed to ${method} ${url} with error: ${error}, stack: ${error.stack}, ${responseText}`)
```
shir-reifenberg, line 291: I know this part wasn't changed, but why we're not using the same logging params for errors ? (headers, data..)
and sorry if I missed it and it's logged somewhere else

packages/adapter-components/src/definitions/user/client_config.ts
```diff
@@ -40,6 +41,26 @@ export type ClientTimeoutConfig = Partial<{
   lastRetryNoTimeout: boolean
 }>
 
+export type ResponseLogStrategy = 'truncate' | 'omit' | 'full'
+// configuration for controlling logging of full http responses.
+// if an endpoint matching the pattern (or any endpoint if pattern is not provided) matches (all of) the following:
+// - logged more than "numItems" times (default: 0)
+// - the current response size is over "size" (default: 0)
+// then the strategy is used - if `omit`, the "Full http response" log is omitted completely, and if `truncate` the response is truncateed.
```
shir-reifenberg, line 49:
```suggestion
// then the strategy is used - if `omit`, the "Full http response" log is omitted completely, and if `truncate` the response is truncated.
```

shir-reifenberg (commented):
Looks great 🚀 
I added some comments
and another thought I have - maybe we should apply those strategies only on fetch and not on deploy? (I used fetch & deploy instead of methods cause we do `GET` requests on deploy as well).

packages/adapter-components/src/client/logging_utils.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+  let allRequests = 0
+  const counters = Object.fromEntries(
+    (clientLoggingConfig.responseStrategies ?? []).map(def => def.pattern ?? '').map(p => [p, 0]),
+  )
+  return ({ responseText, url }) => {
+    allRequests += 1
+    clientLoggingConfig.responseStrategies?.forEach(({ pattern }) => {
+      if (pattern !== undefined && new RegExp(pattern).test(url)) {
+        counters[pattern] += 1
```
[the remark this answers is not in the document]
netama: yes it's intentional, do you think it's a problem?

packages/adapter-components/src/client/http_client.ts
netama: maybe the name is not ideal, I was trying to clarify it's only for the "Full HTTP response" logs without it being too specific, not that it only applies to specific values inside that log. do you have a better naming suggestion? :)

packages/adapter-components/src/client/http_client.ts
netama: probably some old leftover, we can add a trace log for these as well / extend the warn log with just the headers (I think the rest is there?), but I think we should see how we use this one first and not add it as part of this PR, wdyt?

packages/adapter-components/src/definitions/user/client_config.ts
```diff
@@ -40,6 +41,26 @@ export type ClientTimeoutConfig = Partial<{
...
+// note: strategies are calculated in order and the first match wins.
+type ClientLoggingStrategyByEndpointConfig = {
+  // when pattern is omitted, all endpoints match
+  pattern?: string
+  // note: there is an OR between maxItems and maxSize
```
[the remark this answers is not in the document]
netama: oops, old names (also `maxItems`), thanks :)

packages/adapter-components/src/client/http_client.ts
shir-reifenberg: got it. In this case why is the additional nesting level is needed? (`logging.responseStrategies`)

netama: mostly because I'm assuming there might be others in the future and to keep the upper level clean

packages/adapter-components/src/client/http_client.ts
shir-reifenberg: sounds good

shir-reifenberg (approved):
🚀 🚀 🚀

PR #6461, opened 2024-08-24
SALTO-6522 Extend stream serializer

dantal4:
Allow creating serialized stream with the following options:
- `maxLineLength`: writes the result in multiple lines, each line with length <= `maxLineLength`. for example:
```
createStreamSerializer({ maxLineLength: 5 })([1,2,3,4,5]) will result:
[1,2]
[3,4]
[5]
```
- `wrapWithKey`: wraps the output list with `{wrapWithKey:[]}`. for example:
```
createStreamSerializer({ wrapWithKey: 'elements' })([1,2,3,4,5]) will result:
{"elements":[1,2,3,4,5]}
```

---

---
_Release Notes_: 
Lowerdash:
- Extend stream serializer

---
_User Notifications_: 
None

### DIFF
--- a/packages/lowerdash/src/serialize.ts
+++ b/packages/lowerdash/src/serialize.ts
@@ -5,24 +5,58 @@
  *
  * CERTAIN THIRD PARTY SOFTWARE MAY BE CONTAINED IN PORTIONS OF THE SOFTWARE. See NOTICE FILE AT https://github.com/salto-io/salto/blob/main/NOTICES
  */
+import { EOL } from 'os'
+
+export type StreamSerializer = (items: (unknown[] | Record<string, unknown>)[]) => AsyncIterable<string>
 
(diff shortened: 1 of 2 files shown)


netama:
@dantal4 not sure I understand, don't we need the loading logic to be change as well (including backward compatibility)?

dantal4:
> @dantal4 not sure I understand, don't we need the loading logic to be change as well (including backward compatibility)?

@netama that is correct. I'll do the dumping & loading of the state file specifically in another PR.

netama:
@dantal4 what's the reason? don't these all tie together?

dantal4:
@netama  I thought that it would be easier to review, as this code is pretty independent.

edenhassid (approved):
cool! 😎

PR #6472, opened 2024-08-26
Bump to version 0.4.0

netama:
# Salto Version 0.4.0

Changed license to be under the [Salto Terms of Use](https://github.com/salto-io/salto/blob/main/LICENSE)

### Core
* Fix bug where fetch returns an empty response when the only changes in fetch are modifications to static resources.
* List items comparison will not throw when there's `null` / `undefined` in an item.
* Support auto-merge for lists.

### Salesforce Adapter
* Improved time of `Change-Based` fetches (`withChangesDetection`) for organizations with a large amount of Apex Classes.
* In ID collision errors - a link to the service will now appear if it exists.
* Bugfix: CPQ Rules and Conditions Filter creates invalid TemplateExpressions when no conditions are fetched for Rule.
* Fixed an issue where the `profiles` Elements fixer resulted in wrong removals from the Profile.
* Fix the "Can not resolve ..." error on `Flow` deactivation changes.
* Show an error message if the deploy is canceled.
* Removed coordinate fields of flow components that belong to a flow with AUTO_LAYOUT_CANVAS mode.
* `ManagedEventSubscription` is now excluded by default for new accounts.

### Zendesk Adapter
* Add support for custom layouts.
* In ID collision errors - a link to the service will now appear if it exists.
* Support deploying Custom Fields with Options.
* Ticket form `restricted_brand_ids` are now sorted, as these cannot be updated by users.
* Workspace selected macros restriction ids are now sorted, as these cannot be updated by users.

### Netsuite Adapter
* `useChangesDetection` is no longer supported. Use changed-based fetch.
* Support deploy of changes in nested custom fields in data instances (e.g. custom fields under the `subsidiary.mainAddress` field).
* Commented-out JS code will not be added as a generated dependency in SuiteScripts.
* Ignore FileCabinet errors on SDF validation.
* Identify ASVs in multiselect field.
* Recognize more fields and types for ASVs type.
* Improve condition formula parsing for ASVs.
* Prevent getting "Invalid login attempt" after a delayed request (e.g. due to concurrency limit).

### Jira Adapter
* Moved JSM forms to the new API. 
* Fixed a bug with FieldConfiguration deployment in Jira DC.
* Opened `enableNewWorkflowAPI` config flag by default.
* Fixed a bug that caused duplicate attributes due to an empty string in the context option's key.
* Inside `Form` type, the `jiraField` will change from string to reference.
* Validate uniqueness in Project's name and key.
* `ObjectTypeOrder` instances will no longer have `assetSchema` field.
* Enable the option to filter objects by `state` field in the fetch.
* Fix custom fields deletion.
* Fixed a bug where users can not deploy addition of default SLA.
* Support fetching and deploying development field in issue layout.
* Added support for JSM workflow approvals.
* Added support for assignee type in projects.
* Fixed a bug in Jira DC causing components with no lead to fail deployment.
* Fixed a bug that prevented deployment of components with lead ID in DC.
* Fixed a bug that caused some automation deployments to fail.
* Fixed a bug in fieldConfiguration deployment in Jira DC - to enable it you should upgrade Salto's plugin via the marketplace.
* Support Atlassian change in automation's email recipients from group name to group id.
* Check unique names in SLAs and field contexts.
* Fixed a bug caused by circular dependency between root `ObjectType`, `ObjectTypeOrder` and `ObjectSchema`.

### Okta Adapter
* Fixed a bug where modifying `ApplicationGroupAssignment` failed with a 400 HTTP status code.
* Fixed an issue when deploying an `AppUserSchema` with an undefined custom properties.
* Removed an unnecessary API request when only activating / deactivating an `AuthorizationServerPolicy` instance without other modifications.
* References to groups will be added to Identity Provider instances with assigned groups.
* Add reference from UserSchema type to `Application` type.
* Fix a bug in MultifactorEnrollmentPolicy when deployed with Authenticator.
* Add validation to prevent deployment of `MultifactorEnrollmentPolicy` with disabled Authenticator.
* Add deploy support for IdentityProvider type.
* Add option support in Identity Provider JsonWebKey type (excluded by default for all users).
* Fixed a bug where email domains are dropped after being deleted and recreated, by omitting all currently-deleted email domains in fetch.
* Exclude `ProfileMapping` type from Salto when the type is not enabled in the Okta account.

### Confluence Adapter
* Support excluding space specific types or statuses through user config.
* Support fetch and deploy of groups.
* Fix space deployment.
* Support `blog_post` modification deployment.
* Fix issue where hidden id fields were written to the nacl when deploying the first instance of a type.
* Stop fetching templates for archived spaces.

### Microsoft Security Adapter
* Added a shared adapter for Microsoft Entra and Microsoft Intune.
* Extended fetch coverage and deployment support, and added more references.

### JAMF Pro Adapter
* Convert error to Unauthorized error for statuses 401 and 403.

### Salesforce Adapter
* Removed coordinate fields of flow components that belong to a flow with AUTO_LAYOUT_CANVAS mode.

### Zendesk Adapter
* In contextual workspaces, the `layout_uuid` field will be updated to be a reference to a layout.
* Existing `custom_object_field__custom_field_options` instance will have their `name` field removed as it should be managed through its raw counterpart (`raw_name`).
* Ticket form `restricted_brand_ids` are now sorted, as these cannot be updated by users.
* Workspace selected macros restriction ids are now sorted, as these cannot be updated by users.

### Jira Adapter
* Changes will be applied in Form instances:
The fields `issueRequestTypeIds`, `newIssueIssueTypeIds`, `newIssueRequestTypeIds` will be replaced by  `recommendedIssueRequestTypeIds`, `issueCreateIssueTypeIds`, `issueCreateRequestTypeIds`.
* `Workflow` instances will change their structure and will be called `WorkflowConfiguration`.
* The empty string inside the options field in `CustomFieldContext` will now be replaced by '@'.
* Inside `Form` type, the `jiraField` will change from string to reference. 
* `ObjectTypeOrder` instances will no longer have `assetSchema` field.
* One or more fields of type `PANEL` will be added to the issueLayout type. 
* Projects will be added with assignee type.
* Automations will no longer contain their uuid.
* Automations with email group recipients will be references again.

### Okta Adapter
* References to groups will be added to Identity Provider instances with assigned groups.
* `UserSchema` instances with properties sourced by external app will have `Application` reference added.
* Additional fields will be added to IdentityProvider type.
* Email domains that have `validationStatus = "DELETED"` will be omitted.

### Netsuite Adapter
* Unresolved `[ACCOUNT_SPECIFIC_VALUE]` values may be resolved.


yarn.lock
```diff
@@ -4779,14 +4779,30 @@ __metadata:
   languageName: unknown
   linkType: soft
 
-"@salto-io/lowerdash@0.3.61, @salto-io/lowerdash@workspace:packages/lowerdash":
+"@salto-io/logging@npm:<0.4.0":
```
netama, line 4782: this is a dependency from the [suitecloud cli](https://github.com/salto-io/salto/blob/d2a3bb5ff851f1bc416c81914893d15d62b1e4d0/yarn.lock#L5405), fyi @omrilit

yarn.lock
```diff
@@ -4816,19 +4832,30 @@ __metadata:
   languageName: unknown
   linkType: soft
 
-"@salto-io/microsoft-entra-adapter@0.3.61, @salto-io/microsoft-entra-adapter@workspace:packages/microsoft-entra-adapter":
+"@salto-io/lowerdash@npm:0.3.61":
```
netama, line 4835: this is a side-effect of the lower-version logging library

yarn.lock
netama: discussed with @dantal4 , will be bumped afterwards

amitizle: What are we doing with that? Is it not explicitly required in the suitecloud project?

amitizle (approved):

yarn.lock
dantal4: I opened https://salto-io.atlassian.net/browse/SALTO-6546 to update the dependency in the `suitecloud` package.

PR #6473, opened 2024-08-26
SALTO-6481 truncate more

netama:
Small fixes - use smaller params when reaching truncate + typo fix + adjust extraction scripts

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/client/logging_utils.ts
+++ b/packages/adapter-components/src/client/logging_utils.ts
@@ -49,10 +49,10 @@ export const createResponseLogFilter = (
 
 export const truncateReplacer = (_key: string, value: Value): Value => {
   if (Array.isArray(value)) {
-    return value.slice(0, 50)
+    return value.slice(0, 10)
   }
-  if (_.isString(value) && value.length > 500) {
-    return value.slice(0, 500)
+  if (_.isString(value) && value.length > 50) {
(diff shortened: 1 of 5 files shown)


packages/adapter-components/src/client/logging_utils.ts
```diff
@@ -49,10 +49,10 @@ export const createResponseLogFilter = (
 
 export const truncateReplacer = (_key: string, value: Value): Value => {
   if (Array.isArray(value)) {
-    return value.slice(0, 50)
+    return value.slice(0, 10)
   }
-  if (_.isString(value) && value.length > 500) {
-    return value.slice(0, 500)
+  if (_.isString(value) && value.length > 50) {
```
shir-reifenberg, line 54: isn't 50 really short? I'm thinking of expressions and etc this can easily be more than 50
maybe 200?

shir-reifenberg (approved):

packages/adapter-components/src/client/logging_utils.ts
netama: I'll increase but this is truncating so we're always going to lose something (and if we don't truncate enough this is worthless), we can adjust this in the relevant adapters to not use the default rule and never reach this

PR #6474, opened 2024-08-26
SALTO-6419 change error text to avoid AxiosError prefix for additional errors

netama:
continuing from https://github.com/salto-io/salto/pull/6388

---
_Release Notes_: 
None

---
_User Notifications_: 
None


idoamit8 (approved):
💯

PR #6475, opened 2024-08-26
SALTO-6350 remove --immutable flag from yarn in version script

netama:
fix for https://github.com/salto-io/salto/pull/6302 since the yarn lockfile is mutated as part of version bumps

---
_Release Notes_: 
None

---
_User Notifications_: 
None


idoamit8 (approved):
Thanks! 🙏🏼

dantal4 (approved):
thanks!

PR #6476, opened 2024-08-26
SALTO-6544: Improve Data Instances References behavior

tamtamirr:
Improve Data Instances References behavior.

---

Implemented as optional feature for a gradual rollout. (`improvedDataBrokenReferences`)

Workspace Diff when feature is enabled: https://github.com/salto-io/tamir-sf/commit/806650b66b5c587cc64cf84cb09aa0f418b55bf7

When the `ReferenceBehavior` is not `ExcludeInstance`, instances that were referencing an invalid instance, would be omitted altogether, regardless to the defined reference behavior.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


amitgoldarb (approved):
Thanks for this!

PR #6477, opened 2024-08-26
SALTO-6515: Okta: Revert removal of underscores in `AppLogo` file names

Nurdok:
This behavior caused a bug where two `Application` instances with names that differ only with underscores results in their logos having the same file name and overwriting each other, causing a diff.

---

This will be accompanied by a noise reduction.

---
_Release Notes_: 
_Okta Adapter_:
- Fixed a bug where applications with similar names (differing only in underscores) caused a naming collision for their logo file names, resulting in pending changes after fetch / restore.
---
_User Notifications_: 
- `AppLogo` file names may change to include underscores upon fetch.

### DIFF
--- a/packages/okta-adapter/test/filters/app_logo.test.ts
+++ b/packages/okta-adapter/test/filters/app_logo.test.ts
@@ -32,19 +32,21 @@ describe('app logo filter', () => {
   const appLogoType = new ObjectType({ elemID: new ElemID(OKTA, APP_LOGO_TYPE_NAME) })
   const fileName = 'app1_logo.png'
   const content = Buffer.from('test')
-  const appInstance = new InstanceElement('app1', appType, {
-    id: '11',
-    label: 'app1',
-    [LINKS_FIELD]: {
-      logo: [
-        {
-          name: 'fileName',
-          href: 'https://ok12static.oktacdn.com/fs/bco/4/111',
-          type: 'image/png',
-        },
-      ],
-    },
-  })
+  const createAppInstance = (name: string): InstanceElement =>
+    new InstanceElement(name, appType, {
+      id: '11',
+      label: name,
+      [LINKS_FIELD]: {
+        logo: [
+          {
+            name: 'file_Name',
+            href: 'https://ok12static.oktacdn.com/fs/bco/4/111',
+            type: 'image/png',
+          },
+        ],
+      },
+    })
+  const appInstance = createAppInstance('app1')
   beforeEach(async () => {
     const mockCli = mockClient()
     client = mockCli.client
@@ -100,6 +102,17 @@ describe('app logo filter', () => {
       expect(logo).toBeUndefined()
       expect(res.errors).toHaveLength(1)
       expect(res.errors?.[0].message).toEqual('Failed to fetch App logo. Failed to find content type for app1')
+    })
+    // Reproduce an issue where the filter does not give unique file names to AppLogo instances if their names only
+    // differed by underscores.
+    it('should give AppLogo instances unique file names', async () => {
+      // Same name, but with underscore.
+      const anotherAppInstance = createAppInstance('app_1')
+      const elements = [appType, appInstance, anotherAppInstance].map(e => e.clone())
+      await filter.onFetch(elements)
+      const instances = elements.filter(isInstanceElement)
+      const fileNames = instances.filter(e => e.elemID.typeName === APP_LOGO_TYPE_NAME).map(e => e.value.fileName)
+      expect(fileNames).toEqual(['app1.png', 'app_1.png'])
     })
   })
   describe('deploy', () => {


packages/okta-adapter/test/filters/app_logo.test.ts
```diff
@@ -101,6 +103,17 @@ describe('app logo filter', () => {
       expect(res.errors).toHaveLength(1)
       expect(res.errors?.[0].message).toEqual('Failed to fetch App logo. Failed to find content type for app1')
     })
+    // Reproduce an issue where the filter does not give unique file names to AppLogo instances if their names only
+    // differed by underscores.
+    it('should give AppLogo instances unique file names', async () => {
+      // Same name, but with underscore.
+      const anotherAppInstance = createAppInstance('app_1')
+      const elements = [appType, appInstance, anotherAppInstance].map(e => e.clone())
```
idoamit8, line 111: nit: Can remove the clone() for each element 
```suggestion:
const elements = [appType, appInstance, anotherAppInstance]
```

idoamit8 (commented):
@Nurdok Great! 💯 Just a tiny nit.

@shir-reifenberg @Nurdok  There's another edge case we could address in this PR or in a separate one.

If the service has two apps: `dev_1` and `dev 1`.
The NaCl name for the first app is `dev_1`, and for the second app, it’s `dev_1@s`. However, their resourcePathName will be `dev_1` for both.

We might want to consider saving the app logo with the NaCl name. It’s not the prettiest solution, but it would ensure uniqueness. I'm not entirely sure about the file system constraints, though.

Nurdok:
> @Nurdok Great! 💯 Just a tiny nit.
> 
> @shir-reifenberg @Nurdok There's another edge case we could address in this PR or in a separate one.
> 
> If the service has two apps: `dev_1` and `dev 1`. The NaCl name for the first app is `dev_1`, and for the second app, it’s `dev_1@s`. However, their resourcePathName will be `dev_1` for both.
> 
> We might want to consider saving the app logo with the NaCl name. It’s not the prettiest solution, but it would ensure uniqueness. I'm not entirely sure about the file system constraints, though.


You're right. A few things:
1. The behavior for NaCL files that have the same name after normalization is a bit weird - both NaCLs are written to the same file, e.g.:
    ```
    okta.Group my_group {
      objectClass = [
        "okta:user_group",
      ]
      type = "OKTA_GROUP"
      profile = {
        name = "my_group"
      }
      _alias = "my_group"
    }
    okta.Group my_group@s {
      objectClass = [
        "okta:user_group",
      ]
      type = "OKTA_GROUP"
      profile = {
        name = "my group"
      }
      _alias = "my group"
    }
    ```
    Is this the intended behavior?
2. Full NaCL names including the `@` and suffix are valid file names. So I agree with @idoamit8 we can just use them for static file names.
3. I just realized my change also affected (fixed?) `BrandLogo` and `FavIcon` static files in Okta, so noise reduction should consider them too (and it also now shouldn't verify that the new name is identical except for underscored as the full NaCL name may not be) - I already merged that PR but I can draft another one with these changes.

@shir-reifenberg is using the full NaCL name including `@` and suffix SGTY?

shir-reifenberg:
> > @Nurdok Great! 💯 Just a tiny nit.
> > @shir-reifenberg @Nurdok There's another edge case we could address in this PR or in a separate one.
> > If the service has two apps: `dev_1` and `dev 1`. The NaCl name for the first app is `dev_1`, and for the second app, it’s `dev_1@s`. However, their resourcePathName will be `dev_1` for both.
> > We might want to consider saving the app logo with the NaCl name. It’s not the prettiest solution, but it would ensure uniqueness. I'm not entirely sure about the file system constraints, though.
> 
> You're right. A few things:
> 
> 1. The behavior for NaCL files that have the same name after normalization is a bit weird - both NaCLs are written to the same file, e.g.:
>    ```
>    okta.Group my_group {
>      objectClass = [
>        "okta:user_group",
>      ]
>      type = "OKTA_GROUP"
>      profile = {
>        name = "my_group"
>      }
>      _alias = "my_group"
>    }
>    okta.Group my_group@s {
>      objectClass = [
>        "okta:user_group",
>      ]
>      type = "OKTA_GROUP"
>      profile = {
>        name = "my group"
>      }
>      _alias = "my group"
>    }
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    Is this the intended behavior?
> 2. Full NaCL names including the `@` and suffix are valid file names. So I agree with @idoamit8 we can just use them for static file names.
> 3. I just realized my change also affected (fixed?) `BrandLogo` and `FavIcon` static files in Okta, so noise reduction should consider them too (and it also now shouldn't verify that the new name is identical except for underscored as the full NaCL name may not be) - I already merged that PR but I can draft another one with these changes.
> 
> @shir-reifenberg is using the full NaCL name including `@` and suffix SGTY?

thanks @Nurdok and @idoamit8 
I'm not sure I follow why we need the `normalizeFilePathPart` ?
shouldn't using `pathNaclCase(naclCase(logoName))` should give us uniqeness ?

Also I agree that we need a noise reduction for `FavIcon` and `BrandLogo` as well 🙏

Nurdok:
> thanks @Nurdok and @idoamit8 I'm not sure I follow why we need the `normalizeFilePathPart` ? wouldn't using `pathNaclCase(naclCase(logoName))` should give us uniqueness ?
> 
> Also I agree that we need a noise reduction for `FavIcon` and `BrandLogo` as well 🙏

`pathNaclCase` removes the suffix, so "dev_1" and "dev_1@s" would both result in "dev_1" so there's a file collision. The `normalizeFilePathPart` function handles truncation of long file names - we need to call it but it's not related to this issue.

I suggest using the original NaCL name, e.g., `dev_1@s.png` since it's both unique and valid as a file name.

Nurdok:
Discussed f2f with @netama, we'll use the full NaCL names but replace the `@` with `.` for OS / file system compatibility.

packages/okta-adapter/src/logo.ts
```diff
@@ -173,7 +174,7 @@ export const getLogo = async ({
         content: logoContent,
       }),
     },
-    [OKTA, RECORDS_PATH, ...nestedPath, logoType.elemID.typeName, pathName],
+    [OKTA, RECORDS_PATH, ...nestedPath, logoType.elemID.typeName, logoName],
```
idoamit8, line 177: I would prefer to keep the pathName for consistency and aesthetic reasons, especially since it's eventually nested under the parent app, so there won’t be any collisions. I'm aware there's currently a name collision in the Salto files tree, but it's across all of Salto, and we should address that later on.

This is the current view
image
@shir-reifenberg WDYT?

idoamit8 (commented):
Great work! 💯

packages/okta-adapter/src/logo.ts
Nurdok: This was unintended. Good catch! Done.

idoamit8 (approved):
Great 💯 Approving ✅

PR #6478, opened 2024-08-26
SALTO-6550: Allow transforming removed object fields to null

shoshana-houri:
The current implementation of transformRemovedValuesToNull recurses into subfields when a deleted field is an object, and it only transforms the lowest-level fields to null. In some cases, we want to allow the top-level deleted field to be set to null without including its subfields.

---
_Release Notes_: 

---
_User Notifications_:


shir-reifenberg (approved):
💯

PR #6479, opened 2024-08-26
SALTO-6551: Fix - transforming removed values to null creates side effects in the workspace

shoshana-houri:
The existing utility `transformRemovedValuesToNull` sets the removed fields to null in-place. This means that once the deployment succeeds, these null changes will appear in the workspace, which is not supposed to happen.

---

To fix this, we simply modify a clone of the change, rather than the original one.

---
_Release Notes_: 

---
_User Notifications_:


shir-reifenberg (approved):
Thank you for fixing this 👑 🕵️‍♀️ 
beyond the side affects, having null values in workspace leads to workspace error.. great catch!

PR #6480, opened 2024-08-26
SALTO-6538: add allow user call failure config flag

shayc331:
_add allow user call failure config flag_

---

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/adapter-components/src/client/pagination/pagination_async.ts
+++ b/packages/adapter-components/src/client/pagination/pagination_async.ts
@@ -175,6 +175,7 @@ export const traverseRequestsAsync: (
   customEntryExtractor?: PageEntriesExtractor,
 ) => GetAllItemsFunc = (paginationFunc, extractPageEntries, customEntryExtractor) =>
   async function* getPages({ client, pageSize, getParams }) {
+    const isAllowedFailure = getParams.allowFailure ?? false
     const usedParams = new Set<string>()
     const promisesQueue: PromisesQueue = new PromisesQueue()
     let numResults = 0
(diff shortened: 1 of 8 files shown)


packages/adapter-components/src/client/pagination/pagination_async.ts
```diff
@@ -201,6 +202,12 @@ export const traverseRequestsAsync: (
         // avoid leaking promises
         // eslint-disable-next-line no-await-in-loop
         await promisesQueue.clear()
+        if (isAllowedFailure) {
+          log.error(
+            `Failed to get the page starting at ${numResults} for endpoint ${getParams.url} with error: ${e.message}`,
+          )
+          return
```
IdoZakk, line 209: If you won't return here then the pagination will continue

IdoZakk (approved):
🛌 
I think you could have solved it also in the jira adapter by catching inside the paginator function (users.ts new line 61

```typescript
 const usersCallPromise = toArrayAsync(
    paginator(paginationArgs, 
       try {
          page => makeArray(page) as clientUtils.ResponseValue[]),
       } catch {
          ...
          }
  )
```

And sy my suggestion in case you decide to keep it in the adapter component code.
Currently only Jira is using the async pagination so no big impact

packages/adapter-components/src/client/pagination/pagination_async.ts
shayc331: it will not work with this call, when the users call throws it passed the code that enqueues another page

shayc331:
@IdoZakk it won't work because this function only extracts the entries after a successful call.
after talking f2f we agreed to stick with this implementation.

PR #6481, opened 2024-08-26
SALTO-6552: add elementsSource to assetObjectFieldConfiguration deployment

shayc331:
_add elementsSource to assetObjectFieldConfiguration deployment_

---

* while working I found a bug in the jql filter - unresolved references were not resolved.
* fix a bug in the elementsSource that would not resolve references of not top-level instances.
---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/jira-adapter/src/filters/assets/assets_object_field_configuration.ts
+++ b/packages/jira-adapter/src/filters/assets/assets_object_field_configuration.ts
@@ -7,7 +7,7 @@
  */
 
 import _ from 'lodash'
-import { getParent, isResolvedReferenceExpression } from '@salto-io/adapter-utils'
+import { createSchemeGuard, getParent, isResolvedReferenceExpression } from '@salto-io/adapter-utils'
 import { elements as elementUtils, config as configUtils, resolveValues } from '@salto-io/adapter-components'
 import { parse } from 'node-html-parser'
 import {
@@ -28,6 +28,7 @@ import {
   ReadOnlyElementsSource,
 } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
+import Joi from 'joi'
 import { DEFAULT_API_DEFINITIONS } from '../../config/api_config'
 import { FilterCreator } from '../../filter'
 import { FIELD_CONTEXT_TYPE_NAME, FIELD_TYPE_NAME } from '../fields/constants'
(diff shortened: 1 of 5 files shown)


IdoZakk (approved):
💯

packages/adapter-utils/test/element_source.test.ts
```diff
@@ -54,6 +59,11 @@ describe('elementSource', () => {
           expect(await elementsSource.get(new ElemID('adapter', 'type1'))).toBe(elements[0])
         })
 
+        it('should return a not top level element if exists', async () => {
```
IdoZakk, line 62: nit: non instead of not

packages/jira-adapter/src/filters/assets/assets_object_field_configuration.ts
```diff
@@ -46,6 +48,43 @@ const CMDB_OBJECT_FIELD_CONFIGURATION = 'com.atlassian.jira.plugins.cmdb:cmdb-ob
...
+      .required()
+      .unknown(true),
+  })
+    .required()
+    .unknown(true),
+}).unknown(true)
+
+const isCMBDErrorResponse = createSchemeGuard<CMBD_ERROR_RESPONSE>(
+  CMBD_ERROR_RESPONSE_SCHEME,
+  'Received an invalid CMBD error response',
```
IdoZakk, line 85: Seems you do not want a log error if the scheme guard fails

packages/jira-adapter/src/filters/assets/assets_object_field_configuration.ts
```diff
@@ -120,17 +160,19 @@ export const deployAssetObjectContext = async (
...
+  const resolvedInstance = await resolveValues(instance, getLookUpName, elementsSource)
   try {
     const configId = await getFieldConfigId(client, change)
     await client.putPrivate({
       url: `rest/servicedesk/cmdb/latest/fieldconfig/${configId}`,
       data: _.omit(resolvedInstance.value.assetsObjectFieldConfiguration, 'id'),
     })
   } catch (e) {
-    log.error(`Failed to deploy asset object field configuration for instance ${instance.elemID.getFullName()}: ${e}`)
+    const errorMessages = isCMBDErrorResponse(e) && e.response.data.errors.map(error => error.errorMessage)
```
IdoZakk, line 173: I do not like the `false or value` direction. It might be a personal flavor, but in general I think you should manage all the message logic in a single place, something like 

``` typescript
const errorMessages = isCMBDErrorResponse(e) ? e.response.data.errors.map(error => error.errorMessage).join(', ') : e.message
```

IdoZakk (commented):
See comments.
I would prefer another eye on the element source part, so not approving for now

packages/jira-adapter/src/filters/assets/assets_object_field_configuration.ts
shayc331: actually I do, IIUC all the errors should fit this schemeGuard and if they are not I want to know about it

packages/adapter-utils/src/element_source.ts
```diff
@@ -68,7 +68,9 @@ export const buildElementsSourceFromElements = (
   }
 
   const get = async (id: ElemID): Promise<Value> => {
-    const element = elementsMap[id.getFullName()]
+    const { parent } = id.createTopLevelParentID()
+    const topLevel = elementsMap[parent.getFullName()]
```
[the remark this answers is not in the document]
shayc331: implemented in another [pr](https://github.com/salto-io/salto/pull/6505)

PR #6482, opened 2024-08-26
SALTO-6481 adjust default rules for jira

netama:
See context in the ticket - truncate/omit full logs for heavy hitter endpoints

---
_Release Notes_: 
None

---
_User Notifications_: 
None


IdoZakk:
Just to make sure I follow this change- in case we have an incident in which we need this information we change the configuration and ask the customer to run the fetch again?
Also can you elaborate on the specific rules chosen for Jira? I did not find them in the ticket

IdoZakk (approved):
I added some comments in the jira ticket. In general it is OK, I guess we can refine also later on

PR #6483, opened 2024-08-27
SALTO-6435: [Intune] Support fetch & deploy for Device Configuration

shoshana-houri:
* Fetch & deploy Device Configuration policies.
* Add a default transformation on deploy that sets removed fields to null.

---

* The Device Configuration policies in the service:
image

* When testing deployment for this entity, I noticed that our attempts to remove any fields using the Graph API do not actually do anything because we use PATCH—which only updates the specified fields and leaves others unchanged. See [this comment](https://learn.microsoft.com/en-us/graph/api/application-update?view=graph-rest-1.0&tabs=http#request-body) on the application endpoint, for example (I couldn't find it as a global comment, but it appears in various resources).

---
_Release Notes_: 
_Microsoft Security Adapter:_
* Support fetch & deploy for Device Configurations.

---
_User Notifications_: 
_Microsoft Security Adapter:_
* Support fetch & deploy for Device Configurations.


packages/microsoft-security-adapter/src/definitions/deploy/shared/utils/adjust_wrapper.ts
```diff
@@ -0,0 +1,44 @@
...
+      value,
+    }
+  }
+  const adjustedChange = deployment.transformRemovedValuesToNull({ change, skipSubFields: true })
+  return {
+    context: {
+      ...context,
+      change: adjustedChange,
+    },
+    value: fetch.element.recursiveNaclCase(getChangeData(adjustedChange).value, true),
```
shoshana-houri, line 29: The fields in the change are nacl cased while the fields in the `value` are in their original form (see [ref](https://github.com/salto-io/salto/blob/7a234c813d373e3d6da5b101bb7ce8a0f013efc4/packages/adapter-components/src/deployment/deploy/requester.ts#L271)). So if we extract the value from the change we should make sure to be aligned.

packages/microsoft-security-adapter/src/definitions/fetch/intune/fetch.ts
```diff
@@ -137,6 +138,32 @@ const graphBetaCustomizations: FetchCustomizations = {
...
+    ],
+    resource: {
+      directFetch: true,
+    },
+    element: {
+      topLevel: {
+        isTopLevel: true,
+        serviceUrl: {
+          baseUrl: SERVICE_BASE_URL,
+          path: `https://intune.microsoft.com/#view/Microsoft_Intune_DeviceSettings/PolicySummaryReportBlade/policyId/{id}/policyName/${NAME_ID_FIELD.fieldName}/policyJourneyState~/0/policyType~/90/isAssigned~/{isAssigned}`,
```
shir-reifenberg, line 161: please notice you included the `baseUrl` in the path

packages/microsoft-security-adapter/src/definitions/deploy/shared/utils/adjust_wrapper.ts
```diff
@@ -0,0 +1,44 @@
...
+  }
+}
+
+/**
+ * Default adjust functions to apply to all values before deployment
+ * 1. Transform removed values to null
+ * 2. Omit read-only fields
+ */
+export const defaultAdjust = concatAdjustFunctions(adjustRemovedValuesToNull, omitReadOnlyFields)
```
shir-reifenberg, line 38: shouldn't `omitReadOnlyFields` run only for specific types ?

packages/microsoft-security-adapter/src/definitions/deploy/shared/utils/adjust_wrapper.ts
```diff
@@ -0,0 +1,44 @@
...
+import { AdjustFunctionSingle } from '../types'
+import { omitReadOnlyFields } from './read_only_fields'
+
+const adjustRemovedValuesToNull: AdjustFunctionSingle = async ({ typeName, value, context }) => {
+  const { change } = context
+  validatePlainObject(value, typeName)
+  if (!isModificationChange(change) || !isInstanceChange(change)) {
+    return {
+      value,
+    }
```
shir-reifenberg, line 21: shouldn't re also return `context` ? or is it needed only when the context is updated ?

packages/microsoft-security-adapter/src/definitions/deploy/shared/utils/adjust_wrapper.ts
shir-reifenberg: I think i'm missing something 🤔 
isn't `recursiveNaclCase(.., true)` will turn nacl case fields to non nacl case?  then it won't be  aligned? 
why not apply `recursiveNaclCase(.., true)` on the change before calling to `transformRemovedValuesToNull` ?

packages/microsoft-security-adapter/src/definitions/deploy/shared/utils/adjust_wrapper.ts
```diff
@@ -0,0 +1,44 @@
...
+ * Default adjust functions to apply to all values before deployment
+ * 1. Transform removed values to null
+ * 2. Omit read-only fields
+ */
+export const defaultAdjust = concatAdjustFunctions(adjustRemovedValuesToNull, omitReadOnlyFields)
+
+/*
+ * Adjust the value of the object using default adjust function, and then apply the provided adjust function
+ */
+export const adjustWrapper: (adjust: AdjustFunctionSingle) => AdjustFunctionSingle = adjust =>
```
shir-reifenberg, line 43: as this is exported and also used almost everywhere, I wonder if we would want to give it a more descriptive name, WDYT ?

shir-reifenberg (commented):
Looks great, added mainly questions :)

packages/microsoft-security-adapter/src/definitions/deploy/shared/utils/adjust_wrapper.ts
shoshana-houri: Similarly to what I commented [here](https://github.com/salto-io/salto/pull/6469#discussion_r1732584543), I thought it makes sense to run this by default and only require adding the type/fields to the read-only-fields map. I believe this approach is less mistake-prone, but if you have a strong objection I can restrict it to specific types :)

packages/microsoft-security-adapter/src/definitions/deploy/shared/utils/adjust_wrapper.ts
shoshana-houri: Only when it's updated :) see [ref](https://github.com/salto-io/salto/blob/592a7c76fa3d0b5f3df026e66272d4e194df75c7/packages/adapter-components/src/fetch/utils.ts#L97)

packages/microsoft-security-adapter/src/definitions/deploy/shared/utils/adjust_wrapper.ts
shoshana-houri: It used to be called `omitReadOnlyFields` wrapper, as you suggested in the past :) However, since it now applies two different adjust logics, I thought that a longer name would be too long and confusing. As I see it, these are just some default adjust functions that we always want to apply, and if we could have merged two function definitions using the 'default' in the deploy definitions, we would have just used that.

We can rename it to something like `wrapAdjustWithDefaultAdjustFunction` or something similar. I don't think that the name should include a reference to the functions that are being applied, once there's more than 1. wdyt?

[If, however, we decide not to apply `omitReadOnlyFields` by default to each type, then this will go back to applying a single function, and we can revert to a name that reflects that]

packages/microsoft-security-adapter/src/definitions/deploy/shared/utils/adjust_wrapper.ts
shoshana-houri: Sorry if it wasn't clear—the alignment here is regarding the normal form of the `value`. The `change` reflects exactly how the element is represented in the workspace, and we shouldn't modify that.

We could have sent a copy of the `change` and applied `transformRemovedValuesToNull` to both its before and after, but that would be 1) redundant, and 2) not accurate since the problem isn't with the `change` itself (which, as mentioned above, should remain in its original form), but rather with the `value`, which is the actual data we send in the request.

packages/microsoft-security-adapter/src/definitions/deploy/shared/utils/adjust_wrapper.ts
shir-reifenberg, line 60: nit: make it get `...adjust: AdjustFunctionSingle`
then when you have multiple adjust on the "outside" you won't need to call `concatAdjustFunctions` before calling `adjustWrapper`

packages/microsoft-security-adapter/src/definitions/deploy/shared/utils/adjust_wrapper.ts
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Default adjust functions to apply to all values before deployment
+ * 1. Transform removed values to null
+ * 2. Omit read-only fields
+ */
+export const defaultAdjust = concatAdjustFunctions(adjustRemovedValuesToNull, omitReadOnlyFields)
+
+/*
+ * Adjust the value of the object using default adjust function, and then apply the provided adjust function
+ */
```
shir-reifenberg, line 59: if the naming is not descriptive, can we please mention here specifically what functions are being applies?
mostly a nit - but you can just pass `adjustRemovedValuesToNull, omitReadOnlyFields` in line 44 to consolidate the functions

shir-reifenberg (approved):
looks great!

packages/microsoft-security-adapter/src/definitions/deploy/shared/utils/adjust_wrapper.ts
```diff
@@ -0,0 +1,61 @@
...
+import { getChangeData, isInstanceChange, isModificationChange, Values } from '@salto-io/adapter-api'
+import { invertNaclCase, validatePlainObject } from '@salto-io/adapter-utils'
+import { AdjustFunctionSingle } from '../types'
+import { omitReadOnlyFields } from './read_only_fields'
+
+/**
+ * Recursively merge null fields from source to target. Returns a new object.
+ */
+const mergeNullFields = (target: Values, source: Values): Values => {
+  const result = { ...target }
```
shir-reifenberg, line 20: isn't it a shallow copy? If we return a new don't we need to clone it ?

packages/microsoft-security-adapter/src/definitions/deploy/shared/utils/adjust_wrapper.ts
```diff
@@ -0,0 +1,61 @@
...
+  _.forEach(source, (srcField, naclCaseFieldName) => {
```
shir-reifenberg, line 21: nit: I think you can use Object.entries
something like - replace with Object.entries(source).map(([key, value]) => )

packages/microsoft-security-adapter/test/definitions/deploy/shared/utils/adjust_wrapper.test.ts
```diff
@@ -0,0 +1,99 @@
...
+  omitReadOnlyFields: jest.fn((...args) => omitReadOnlyFieldsMock(...args)),
+}))
+
+describe('adjust wrapper utils', () => {
+  describe(`${defaultAdjust.name}`, () => {
+    beforeEach(() => {
+      omitReadOnlyFieldsMock.mockImplementation(async ({ value }) => ({ value }))
+    })
+
+    it('should set removed fields as null and remove read-only fields', async () => {
```
shir-reifenberg, line 27: can we please add example with list as well ?

shir-reifenberg (commented):
Just realized I reviewed before u pushed some changes, so added few more comments but non blocking

packages/microsoft-security-adapter/src/definitions/deploy/shared/utils/adjust_wrapper.ts
shoshana-houri: The functions that are being applied are specifically mentioned in the `defaultAdjust` docstrings. Do you think we should repeat that? It seems like unnecessary duplication to me. The same goes for the comment on line 44—if we decide to add another function to `defaultAdjust`, we'll need to update both the docstrings and the corresponding code sections. 🤔

packages/microsoft-security-adapter/src/definitions/deploy/shared/utils/adjust_wrapper.ts
shoshana-houri: We never change the original value here since we recursively copy each level that may be modified.
As a more general comment, it's okay for the adjust functions to change the value in-place. I just didn't do that (AFAICR) in the adjust functions in this adapter, so I preferred keeping it this way.

packages/microsoft-security-adapter/src/definitions/deploy/shared/utils/adjust_wrapper.ts
shoshana-houri: AFAICT `Object.entries` won't give us the desired result, at least not without adding additional logic such as filter and concat. It's possible that there may be values in source that we don't want to return or set as null. For example, fields that were omitted from the value but not from the change.

shir-reifenberg (approved):

PR #6484, opened 2024-08-27
SALTO-6556: Salesforce customObjectInstanceReferencesFilter fails on Cannot read properties of null (reading 'slice')

tamtamirr:
Salesforce customObjectInstanceReferencesFilter fails on Cannot read properties of null (reading 'slice')

---

This is an edge-case where the Lookup field has either no `referenceTo` annotation or multiple `referenceTo` annotations, but the value of the Lookup is not a string. (null is possible for optional Lookup Field).
I simply validate the value of the reference field is a string as part of the transform function.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/test/filters/custom_object_instances_references.test.ts
+++ b/packages/salesforce-adapter/test/filters/custom_object_instances_references.test.ts
@@ -164,6 +164,16 @@ describe('Custom Object Instances References filter', () => {
           [FIELD_ANNOTATIONS.UPDATEABLE]: true,
         },
       },
+      LookupNoRefTo: {
+        refType: Types.primitiveDataTypes.Lookup,
+        annotations: {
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/test/filters/custom_object_instances_references.test.ts
```diff
@@ -236,6 +246,7 @@ describe('Custom Object Instances References filter', () => {
     Id: 'toDuplicate',
     LookupExample: 'duplicateId-1',
     MasterDetailExample: 'duplicateId-2',
+    LookupNoRefTo: null,
```
tamtamirr, line 249: Used to fail before my changes

Uri-Bar (approved):
🔥

PR #6486, opened 2024-08-27
SALTO-6549: Fix fetch isolated mode omitting field annotations

ori-moisis:
The change routing code would omit annotation values from fields when the field type changed. This is a hacky fix that does not really cover all cases, but should solve the most obvious ones at least

---

This issue was caught by the cli e2e, see ticket for more details on the scenario

---
_Release Notes_: 
Core:
- Fixed issue where fetch in "isolated" mode could potentially drop field attributes in some cases

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/workspace/src/workspace/nacl_files/multi_env/routers.ts
+++ b/packages/workspace/src/workspace/nacl_files/multi_env/routers.ts
@@ -110,6 +110,10 @@ const separateChangeByFiles = async (change: DetailedChange, source: NaclFilesSo
   if (_.isEmpty(elementNaclFiles)) {
     return [change]
   }
+  if (elementNaclFiles.length === 1) {
+    // Optimization - no need to parse files and go over all the values if there is only one possible location
+    return [{ ...change, path: toPathHint(elementNaclFiles[0]) }]
+  }
   const sortedChanges = (
     await Promise.all(
       elementNaclFiles.map(async filename => {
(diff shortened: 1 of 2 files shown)


packages/workspace/src/workspace/nacl_files/multi_env/routers.ts
```diff
@@ -130,6 +134,19 @@ const separateChangeByFiles = async (change: DetailedChange, source: NaclFilesSo
...
+    // The code in "filterByFile" can omit information if it did not exist in the source.
+    // There can be information that does not exist in the source if, for example, the change is a field type modification.
+    // in that case, the change would contain the whole field, but when we filter the annotation values of the field
+    // we would only know where to put the ones that existed before the type change, so the code above would omit
+    // all the new annotation values.
+    // Also, if all the annotation values are new, the entire change would be omitted because of "isEmptyChangeElement".
+    // The hacky bug fix here - assuming there is no case where we actually split the field to multiple files (a wrong assumption)
+    // we will get here with either 1 change that potentially is missing some values, or 0 changes (which is missing everything)
+    // either way, since there is no need to split anything, we can return the original change that contains all the values for sure
+    return [{ ...change, path: sortedChanges[0]?.path ?? change.path }]
```
netama, line 147: I'm wondering if we can instead add the following change in `filterByFile` instead of the current [EXCLUDE](https://github.com/salto-io/salto/blob/main/packages/workspace/src/workspace/nacl_files/multi_env/routers.ts#L83) - wdyt? 
```
    return (id.idType === 'field' && !id.isBaseID())
      ? FILTER_FUNC_NEXT_STEP.INCLUDE
      : FILTER_FUNC_NEXT_STEP.EXCLUDE
```
not sure how we want to handle cases where annotations are split across multiple files - but I guess we can add some check and keep them under the first if we need to?
anyway I think it might be a slightly safer change than this one, but I could be missing something (also any idea why the `.isAttrID()` elem id function doesn't handle the field annotation case correctly? or am I just not reading it right?)

netama (approved):

PR #6494, opened 2024-08-28
Allow ignoring empty query args, fix precedence during recursion

netama:
* Add an option to omit empty query args - which will allow unsetting values more easily when placeholders are defined (this still won't cover everything - we can later add conditions on fetch requests as well)
* Reverse order of precedence between initial context and overrides received from recurseInto args, to allow self-recursion

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/definitions/system/requests/types.ts
+++ b/packages/adapter-components/src/definitions/system/requests/types.ts
@@ -30,6 +30,8 @@ export type RequestArgs = {
     // true: &arg[0]=val1&arg[1]=val2
     // null: &arg=val1&arg=val2
     indexes?: boolean | null
+    // when true, empty strings will be omitted from the query params
+    omitEmpty?: boolean
   }
   // TODO support x-www-form-urlencoded + URLSearchParams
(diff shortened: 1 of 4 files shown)


packages/adapter-components/src/definitions/system/requests/types.ts
```diff
@@ -30,6 +30,8 @@ export type RequestArgs = {
     // true: &arg[0]=val1&arg[1]=val2
     // null: &arg=val1&arg=val2
     indexes?: boolean | null
+    // when true, empty strings will be omitted from the query params
```
shir-reifenberg, line 33: this is also true for empty arrays, right ?

shir-reifenberg (approved):
thank you 👑

packages/adapter-components/src/definitions/system/requests/types.ts
netama: yes, forgot to update the comment after renaming the field, thanks :)

PR #6495, opened 2024-08-28
SALTO-6271: Upgrade Workato fetch to new adapters infra

shir-reifenberg:
Upgrade workato definitions to use the new infra

---

Did the following tests - 

1. fetch before -> git add -> fetch after, with the following options:
-- no additional services
-- with netsuite connected and enableDeploy flag disabled
-- with enableDeploy flag enabled
2. deploy
3. `SALTO_LOG_LEVEL=error salto fetch -f` - didn't see additional errors 

Still TODO -
adjust adapter.test

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/workato-adapter/src/adapter.ts
```diff
@@ -65,57 +65,77 @@ export const DEFAULT_FILTERS = [
...
-      types: this.userConfig.apiDefinitions.types,
-      supportedTypes: this.userConfig.apiDefinitions.supportedTypes,
       fetchQuery: this.fetchQuery,
-      paginator: this.paginator,
-      nestedFieldFinder: returnFullEntry,
-      computeGetArgs: simpleGetArgs,
-      typeDefaults: this.userConfig.apiDefinitions.typeDefaults,
+      definitions: this.definitions,
       getElemIdFunc: this.getElemIdFunc,
+      additionalRequestContext: { parent_id: [''] },
```
netama, line 138: maybe add a comment explaining that this is for the folders recursion?

packages/workato-adapter/src/client/client.ts
```diff
@@ -36,7 +36,7 @@ export default class WorkatoClient extends clientUtils.AdapterHTTPClient<
 > {
   constructor(clientOpts: clientUtils.ClientOpts<Credentials, definitions.ClientRateLimitConfig>) {
     super(WORKATO, clientOpts, createConnection, {
-      pageSize: DEFAULT_PAGE_SIZE,
+      pageSize: DEFAULT_PAGE_SIZE, // TODOS deprecated ? ask neta
```
netama, line 39: pretty sure it is, but I think you have to pass it in. do you need it for anything? I think we'll remove it once all adapters are migrated

netama (approved):
LGTM 🎈 
might be worth testing with two connections, though it's probably ok since the filter was not modified
I only sampled the config-vs-fetch-definitions for a few types, if you want a deeper review there lmk

shir-reifenberg:
> LGTM 🎈 might be worth testing with two connections, though it's probably ok since the filter was not modified I only sampled the config-vs-fetch-definitions for a few types, if you want a deeper review there lmk

tested with one SF one NS connection, looks good.

PR #6496, opened 2024-08-28
SALTO-6577: support OAuth server scopes

shir-reifenberg:
Adde deploy support for `OAuth2Scope` type, and a missing ref from ` AuthorizationServerPolicyRule` to `OAuth2Scope`

---

will add tests & noise reduction later

---
_Release Notes_: 
_Okta_adapter_:
- Support deployment of Authorization Server scopes 
- Fix a bug in deployment of Authorization Server Policy Rules that references scopes

---
_User Notifications_: 

_Okta_adapter_:
- Reference will be added in Authorization Server Policy Rule that references Auth Scopes

### DIFF
--- a/packages/okta-adapter/src/definitions/deploy/deploy.ts
+++ b/packages/okta-adapter/src/definitions/deploy/deploy.ts
@@ -804,6 +804,51 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
         },
       },
     },
+    OAuth2Scope: {
+      requestsByAction: {
+        customizations: {
+          add: [
+            {
+              request: {
+                endpoint: {
+                  path: '/api/v1/authorizationServers/{authServerId}/scopes',
+                  method: 'post',
+                },
+                context: {
+                  authServerId: '{_parent.0.id}',
+                },
+              },
+            },
+          ],
+          modify: [
+            {
+              request: {
+                endpoint: {
+                  path: '/api/v1/authorizationServers/{authServerId}/scopes/{id}',
+                  method: 'put',
+                },
+                context: {
+                  authServerId: '{_parent.0.id}',
+                },
+              },
+            },
+          ],
+          remove: [
+            {
+              request: {
+                endpoint: {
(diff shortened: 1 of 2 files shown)


packages/okta-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -804,6 +804,51 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
...
       },
     },
+    OAuth2Scope: {
+      requestsByAction: {
+        customizations: {
+          add: [
+            {
+              request: {
+                endpoint: {
+                  path: '/api/v1/authorizationServers/{authServerId}/scopes',
```
shoshana-houri, line 814: nit (same for all 3 requests):
```suggestion
                  path: '/api/v1/authorizationServers/{parent_id}/scopes',
```
And this way there's no need to specify the context.
See [ref](https://github.com/salto-io/salto/blob/b987c68615c0c50fa8e667b81da0c40c2f6939a2/packages/adapter-components/src/deployment/definition_helpers.ts#L12)

shoshana-houri (approved):
🚀 
The requests can be simplified a little bit. Approving to save cycle :)

packages/okta-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -804,6 +804,42 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
         },
       },
     },
+    OAuth2Scope: {
```
Nurdok, line 807: Would this work if you added it to the `createStandardDeployDefinitions` list above? Not use if `{parent_id}` would work, but worth a try.

Nurdok (approved):

packages/okta-adapter/src/definitions/deploy/deploy.ts
shir-reifenberg: it doesn't work cause it expect for a single id field and we have 2 here 🙃

PR #6499, opened 2024-08-29
SAAS-10995: add logs to fallback_user fixer

Shulik95:
_Adds a log to fallback_user fixer to increase visibility_

---

_None_

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/zendesk-adapter/src/fix_elements/fallback_user.ts
+++ b/packages/zendesk-adapter/src/fix_elements/fallback_user.ts
@@ -171,6 +171,9 @@ export const fallbackUsersHandler: FixElementsHandler =
       .filter(isRelevantElement)
       .map(replaceMissingUsers(userEmails, fallbackValue))
       .filter(values.isDefined)
+    if (fixedElementsWithUserCount.length !== 0) {
+      log.info('Setting the user for %d elements: %o', fixedElementsWithUserCount.length, fixedElementsWithUserCount)
+    }
     const errors = fixedElementsWithUserCount.map(({ fixedInstance, missingUsers }) =>
       missingUsersChangeWarning(fixedInstance, missingUsers, fallbackValue),
     )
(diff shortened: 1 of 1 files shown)


packages/zendesk-adapter/src/fix_elements/fallback_user.ts
```diff
@@ -171,6 +171,9 @@ export const fallbackUsersHandler: FixElementsHandler =
       .filter(isRelevantElement)
       .map(replaceMissingUsers(userEmails, fallbackValue))
       .filter(values.isDefined)
+    if (fixedElementsWithUserCount.length !== 0) {
+      log.info('Setting the user for %d elements: %o', fixedElementsWithUserCount.length, fixedElementsWithUserCount)
```
shir-reifenberg, line 175: few comments / suggestions :)
1. `fixedElementsWithUserCount` might contain references so printing it with `%o` can make the code crash due to circular refs. For those cases u can use `inspectValue` which converts object to string in a safe way, and also can be provided with `maxArrayLength` to avoid logging too much data (large logs can get truncated).
but - I'm not sure we care about the elements themselves, rather than the number of changed elements, WDYT?
2. I think it will be useful to include the "fallback" strategy from the config, cause in our case we only care about `##DEPLOYER##` fallback (I think including the `defaultMissingUserFallback` is what we want)
3. mostly nit - maybe just `log.debug` ?

shir-reifenberg (approved):
Added some comments but approving to save a cycle 🚲

packages/zendesk-adapter/src/fix_elements/fallback_user.ts
Shulik95: 1. Agreed, the amount of changed elements is enough here.
2. 👍 
3. 👍

PR #6500, opened 2024-08-29
SALTO-6531: Change custom name logic in Okta to be less strict

shir-reifenberg:
Our current custom app logic didn't take into account domain changes, this PR fixes this

---

When the subdomain of the tenant changes, apps created before the changes will keep the name of the original domain and will not get updated. This leads to cases when we false detect certain apps as non custom, though they actually are.

This PR introduces a new logic to determine if an app is custom - 
1. If the OrgSetting type exist, we will use the domain
2. If there's no match, we will use the tenant_appName_number pattern to detect it

---
_Release Notes_: 

_Okta_adapter_:
- Fix a bug in the deployment of custom apps when subdomain has changed.

---
_User Notifications_: 

_Okta_adapter_:
- Name field will be removed for custom SAML and AUTO_LOGIN applications.

### DIFF
--- a/packages/okta-adapter/src/definitions/fetch/types/application.ts
+++ b/packages/okta-adapter/src/definitions/fetch/types/application.ts
@@ -88,7 +88,7 @@ export const isCustomApp = (value: Values, subdomain?: string): boolean => {
     )
   }
 
-  // when subdomain was replaced in the tenant, existing apps will include the prevoius subdomain in their name, so we rely on the regex as a fallback
+  // when subdomain was replaced in the tenant, existing apps will include the previous subdomain in their name, so we rely on the regex as a fallback
   const isCustomAppName = subdomainMatch || customAppNamePatternMatch
   return [AUTO_LOGIN_APP, SAML_2_0_APP].includes(value.signOnMode) && isCustomAppName
 }
--- a/packages/okta-adapter/src/definitions/fetch/types/application.ts
+++ b/packages/okta-adapter/src/definitions/fetch/types/application.ts
@@ -57,34 +57,38 @@ export const assignPolicyIdsToApplication = (value: unknown): Value => {
   }
 }
 
-const endsWithNumberRegex = new RegExp(/_\d+$/)
+// custom app name pattern is: subdomain_appName_number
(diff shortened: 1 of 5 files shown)


packages/okta-adapter/src/definitions/fetch/types/application.ts
```diff
@@ -57,34 +57,38 @@ export const assignPolicyIdsToApplication = (value: unknown): Value => {
   }
 }
 
-const endsWithNumberRegex = new RegExp(/_\d+$/)
+// custom app name pattern is: subdomain_appName_number
```
Nurdok, line 60: Are you sure there are _exactly_ two underscores? What if the subdomain itself contains underscores? Might be safer to allow underscores in the middle part too.

packages/okta-adapter/src/definitions/fetch/types/application.ts
```diff
@@ -57,34 +57,38 @@ export const assignPolicyIdsToApplication = (value: unknown): Value => {
...
+      'isCustomApp matching methods agree for %s: subdomainMatch=%s, customAppNamePatternMatch=%s',
       value.name,
       subdomainMatch,
-      endsWithNumberMatch,
+      customAppNamePatternMatch,
     )
   }
 
-  return [AUTO_LOGIN_APP, SAML_2_0_APP].includes(value.signOnMode) && subdomainMatch
+  // when subdomain was replaced in the tenant, existing apps will include the prevoius subdomain in their name, so we rely on the regex as a fallback
```
Nurdok, line 91: previous (typo)

Nurdok (approved):
LGTM, but let's just make sure what happens with subdomain that have underscores in them.

packages/okta-adapter/src/definitions/fetch/types/application.ts
shir-reifenberg: I tried to create an app with underscore in its name and Okta seems to be removing those for the `name` field. Didn't validated it but I guess it would be the same for a subdomain containing underscore? (also I think it's a little unlikely)

PR #6501, opened 2024-08-29
SALTO-6573: Avoid redundant read of elements when rebuilding cache

ori-moisis:
This avoids loading another copy of the elements from the state
It also avoids re-calculating the hidden parts for each such element

---

The main point of the change here is this:
There is code that needs to decide which elements to merge (dropStateOnlyElementsRecovery)
before this change, that code gets two iterators of elements, it needs to filter the second iterator such that it only contains elements that exist in the first iterator (by element ID) or elements that are fully hidden.
to implement the first condition, it has to fully consume the first iterator just to get the element IDs, so essentially it loads all the elements into memory.

In this PR we change the interface such that it works with iterators of element IDs and not full elements, this mainly improves memory consumption since there is no need to hold all elements in memory in `dropStateOnlyElementsRecovery` anymore, it only holds the element IDs.

In the second commit we add a performance improvement such that if an element is overridden, the code will not try to read it from the original source.
This is significant because the original source in this case is a "map element source" ([here](https://github.com/salto-io/salto/blob/main/packages/workspace/src/workspace/workspace.ts/#L642)) that is heavy to calculate

---
_Release Notes_: 
Core:
- Slight performance improvement when loading workspaces without a valid workspace cache

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/workspace/src/workspace/elements_source.ts
+++ b/packages/workspace/src/workspace/elements_source.ts
@@ -158,3 +158,41 @@ export const mapReadOnlyElementsSource = (
   has: id => source.has(id),
   list: () => source.list(),
 })
+
+export const createOverrideReadOnlyElementsSource = (
+  source: ReadOnlyElementsSource,
+  overrideElements: Record<string, Element | undefined>,
+): ReadOnlyElementsSource => {
+  const compareStrings = (id1: string, id2: string): number => {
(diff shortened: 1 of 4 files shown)


ori-moisis:
renamed the branch, the PR is now in #6512

PR #6502, opened 2024-08-29
SALTO-6480 make sure there are no dups before cloning fields parent

omrifarkash:
On `getClonedElements` in `workspace` package, we are cloning the parent of each Field. This might cause an OOM error, we should first filter out duplications before cloning

---

---

_Release Notes_: 
_Core_:
- fix OOM error when cloning large amount of Fields

---

_User Notifications_: 
None


dantal4 (approved):
thanks!

PR #6503, opened 2024-08-29
Improve memory usage of custom references calculation in Jira

dantal4:
In `getFieldConfigurationItemsReferences`:
1. I removed the usage of `joi` schema for `fields` in `FieldConfiguration` instances, because I suspected that it has memory/runtime implications, and we actually only care about the key of each field in `fields`
2. I reused the `elemIDs` of the `Field` instances, because there's no reason to create a new `ElemID` object in memory for each field in each fieldConfiguration instance, as all fieldConfiguration instances reference the same field instances (`ElemID` objects are immutable, so there's no problem sharing them).

Those improvements made it possible to run `salto workspace cache update` on a really large Jira environment - with more that 1M references of `Field` instances in `FieldConfiguration` instances.

---

---
_Release Notes_: 
Jira Adapter:
- Improve memory usage of custom references calculation

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/field_configuration/replace_field_configuration_references.ts
+++ b/packages/jira-adapter/src/filters/field_configuration/replace_field_configuration_references.ts
@@ -6,7 +6,6 @@
  * CERTAIN THIRD PARTY SOFTWARE MAY BE CONTAINED IN PORTIONS OF THE SOFTWARE. See NOTICE FILE AT https://github.com/salto-io/salto/blob/main/NOTICES
  */
 import _ from 'lodash'
-import Joi from 'joi'
 import {
   CORE_ANNOTATIONS,
   ElemID,
@@ -18,8 +17,9 @@ import {
   Values,
   ReadOnlyElementsSource,
   ReferenceExpression,
+  Value,
 } from '@salto-io/adapter-api'
-import { createSchemeGuard, isResolvedReferenceExpression } from '@salto-io/adapter-utils'
+import { isResolvedReferenceExpression } from '@salto-io/adapter-utils'
 import { collections, values } from '@salto-io/lowerdash'
 import { logger } from '@salto-io/logging'
 import { findObject } from '../../utils'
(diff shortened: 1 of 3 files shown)


packages/jira-adapter/src/weak_references/field_configuration_items.ts
```diff
@@ -26,60 +25,48 @@ const { pickAsync } = promises.object
...
-const FIELD_CONFIGURATION_ITEMS_SCHEME = Joi.object().pattern(/.*/, FIELD_CONFIGURATION_ITEM_SCHEME)
-
-export const isFieldConfigurationItems = createSchemeGuard<FieldConfigurationItems>(FIELD_CONFIGURATION_ITEMS_SCHEME)
-
-const getFieldReferences = (instance: InstanceElement): ReferenceInfo[] => {
+const getFieldReferences = (instance: InstanceElement, fieldElemIdsMap: Record<string, ElemID>): ReferenceInfo[] => {
   const fieldConfigurationItems = instance.value.fields
-  if (fieldConfigurationItems === undefined || !isFieldConfigurationItems(fieldConfigurationItems)) {
+  if (fieldConfigurationItems === undefined) {
+    return []
```
shayc331, line 31: I think we should log this warning in both cases

packages/jira-adapter/src/filters/field_configuration/replace_field_configuration_references.ts
```diff
@@ -25,20 +26,21 @@ import { findObject } from '../../utils'
...
-
 const enrichFieldItem = async (
   fieldName: string,
-  fieldItem: FieldItem,
+  fieldItem: Value,
   elementSource: ReadOnlyElementsSource,
   instanceName: string,
-): Promise<EnrichedFieldItem | undefined> => {
+): Promise<Value> => {
+  if (!_.isPlainObject(fieldItem)) {
```
shayc331, line 40: please add a ut for this scenario

packages/jira-adapter/src/filters/field_configuration/replace_field_configuration_references.ts
```diff
@@ -64,8 +66,15 @@ const replaceToMap = (instance: InstanceElement): void => {
 
 const replaceFromMap = async (instance: InstanceElement, elementSource: ReadOnlyElementsSource): Promise<void> => {
   const fieldConfigurationItems = instance.value.fields
-  if (fieldConfigurationItems === undefined || !isFieldConfigurationItems(fieldConfigurationItems)) {
-    log.warn(`fields value is corrupted in instance ${instance.elemID.getFullName()}, hence not changing fields format`)
+  if (fieldConfigurationItems === undefined) {
+    return
+  }
+  if (!_.isPlainObject(fieldConfigurationItems)) {
```
shayc331, line 72: please add a ut for this scenario as well

shayc331 (approved):
🥇 looking great! 🥇 
add some comments, approving to save a cycle

PR #6505, opened 2024-09-01
SALTO-6592:  fix elementsSource bug that cause failure when getting a non top level element

shayc331:
_fix elementsSource bug that cause failure when getting a non top level element_

---

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/adapter-utils/src/element_source.ts
+++ b/packages/adapter-utils/src/element_source.ts
@@ -71,7 +71,7 @@ export const buildElementsSourceFromElements = (
     const { parent } = id.createTopLevelParentID()
     const topLevelElement = elementsMap[parent.getFullName()]
     const element = elementsMap[id.getFullName()] ?? (topLevelElement && resolvePath(topLevelElement, id))
-    if (element) {
+    if (element !== undefined) {
       return element
     }
 
--- a/packages/adapter-utils/src/element_source.ts
+++ b/packages/adapter-utils/src/element_source.ts
@@ -9,7 +9,7 @@ import { ReadOnlyElementsSource, Element, ElemID, Value, isElement } from '@salt
 import { logger } from '@salto-io/logging'
 import _ from 'lodash'
 import { collections, values } from '@salto-io/lowerdash'
-import { resolveTypeShallow } from './utils'
+import { resolvePath, resolveTypeShallow } from './utils'
 
 const { awu } = collections.asynciterable
 
@@ -68,8 +68,10 @@ export const buildElementsSourceFromElements = (
   }
 
   const get = async (id: ElemID): Promise<Value> => {
-    const element = elementsMap[id.getFullName()]
-    if (element !== undefined) {
+    const { parent } = id.createTopLevelParentID()
+    const topLevelElement = elementsMap[parent.getFullName()]
+    const element = elementsMap[id.getFullName()] ?? (topLevelElement && resolvePath(topLevelElement, id))
+    if (element) {
       return element
     }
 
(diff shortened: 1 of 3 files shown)


packages/adapter-utils/src/element_source.ts
```diff
@@ -68,8 +68,10 @@ export const buildElementsSourceFromElements = (
   }
 
   const get = async (id: ElemID): Promise<Value> => {
-    const element = elementsMap[id.getFullName()]
-    if (element !== undefined) {
+    const { parent } = id.createTopLevelParentID()
+    const topLevelElement = elementsMap[parent.getFullName()]
+    const element = elementsMap[id.getFullName()] ?? (topLevelElement && resolvePath(topLevelElement, id))
+    if (element) {
```
dantal4, line 74: element can be falsy - 0, '', false

```suggestion
    if (element !== undefined) {
```

packages/adapter-utils/test/element_source.test.ts
```diff
@@ -72,7 +100,18 @@ describe('elementSource', () => {
...
+
+        it('should return true for a non top level element if exists', async () => {
+          const nestedElemId = instanceElement.elemID.createNestedID('key')
+          expect(await elementsSource.has(nestedElemId)).toBeTruthy()
+        })
+
+        it('should return true for nested field when the field is in the source but the parent is not', async () => {
+          const field = objectType.fields.key
+          const source = buildElementsSourceFromElements([field])
+          expect(await source.get(field.elemID)).toBeTruthy()
```
dantal4, line 114:
```suggestion
          expect(await source.has(field.elemID)).toBeTruthy()
```

dantal4 (approved):
Great work!
Approving to save a cycle 🙏

packages/adapter-utils/src/element_source.ts
```diff
@@ -78,8 +80,7 @@ export const buildElementsSourceFromElements = (
       .find(values.isDefined)
   }
 
-  const has = async (id: ElemID): Promise<boolean> =>
-    elementsMap[id.getFullName()] !== undefined || awu(fallbackSources).some(async source => source.has(id))
+  const has = async (id: ElemID): Promise<boolean> => (await get(id)) !== undefined
```
ori-moisis, line 83: this can create a very significant degradation in performance (reading the value takes significantly more time than checking only keys).
not sure how much we actually use `has`, but if we do, this change could be bad, especially for large workspaces

dantal4: @ori-moisis is it expensive doing it on the input `elements` or only on the `fallbackSources`?

dantal4: This is not expensive, right?
```typescript
    const topLevelElement = elementsMap[parent.getFullName()]
    const element = elementsMap[id.getFullName()] ?? (topLevelElement && resolvePath(topLevelElement, id))
```

ori-moisis: yes

dantal4: So we can extract the part in get that goes to the elements (instead of the fallback sources), and use it:
```typeacript
 const has = id =>
   getFromElements(id) !== undefined || awu(fallbackSources).some(...)  getFromElements(id) !== undefined || awu(fallbackSources).some(...)
```

@ori-moisis  right?

ori-moisis: hard to review like this, open a PR?

dantal4: @shayc331 opened https://github.com/salto-io/salto/pull/6510

PR #6506, opened 2024-09-01
SALTO-6434: [Intune] Support fetch & deploy for Device Compliance policies

shoshana-houri:
Fetch & deploy Device Compliance policies.

---

image

See nacls structure example [here](https://github.com/salto-io/microsoft-security/tree/main/microsoft_security/Records/intune_deviceCompliance).

---
_Release Notes_: 
_Microsoft Security Adapter:_
* Support fetch & deploy for Device Configurations.

---
_User Notifications_: 
_Microsoft Security Adapter:_
* Support fetch & deploy for Device Configurations.


packages/microsoft-security-adapter/src/definitions/deploy/intune/deploy.ts
```diff
@@ -250,6 +254,73 @@ const graphBetaCustomDefinitions: DeployCustomDefinitions = {
...
+                path: '/deviceManagement/deviceCompliancePolicies/{id}/scheduleActionsForRules',
+                method: 'post',
+              },
+              transformation: {
+                adjust: adjustWrapper(async ({ value }) => {
+                  validatePlainObject(value, DEVICE_COMPLIANCE_TYPE_NAME)
+                  return {
+                    value: {
+                      deviceComplianceScheduledActionForRules: value[SCHEDULED_ACTIONS_FIELD_NAME] ?? [],
+                    },
```
shir-reifenberg, line 299: Can `rename` work?

packages/microsoft-security-adapter/src/definitions/fetch/intune/fetch.ts
```diff
@@ -226,6 +230,62 @@ const graphBetaCustomizations: FetchCustomizations = {
...
+          path: `/#view/Microsoft_Intune_DeviceSettings/CompliancePolicyOverview.ReactView/policyId/{id}/policyName/{${NAME_ID_FIELD.fieldName}}/platform~/0/policyType~/0`,
+        },
+      },
+      fieldCustomizations: ID_FIELD_TO_HIDE,
+    },
+  },
+  [DEVICE_COMPLIANCE_SCHEDULED_ACTIONS_TYPE_NAME]: {
+    resource: {
+      directFetch: false,
+    },
```
shir-reifenberg, line 265: I think this is redundant if the type is not top level

packages/microsoft-security-adapter/src/definitions/references/cross_reference_rules.ts
shir-reifenberg, line 1: That's the reason I initially thought we shouldn't have have 2 separate files cause I think at some point it will be a bit hard to decide what should be where, but up to you

packages/microsoft-security-adapter/src/definitions/deploy/intune/deploy.ts
```diff
@@ -20,6 +21,9 @@ const {
   APPLICATION_CONFIGURATION_MANAGED_DEVICE_TYPE_NAME,
   DEVICE_CONFIGURATION_TYPE_NAME,
   DEVICE_CONFIGURATION_SETTING_CATALOG_TYPE_NAME,
+  DEVICE_COMPLIANCE_TYPE_NAME,
+
```
shir-reifenberg, line 25:
```suggestion
```

shir-reifenberg (approved):
Great work! 🚀

packages/microsoft-security-adapter/src/definitions/fetch/intune/fetch.ts
```diff
@@ -226,6 +230,62 @@ const graphBetaCustomizations: FetchCustomizations = {
...
+      {
+        endpoint: {
+          path: '/deviceManagement/deviceCompliancePolicies',
+          queryArgs: {
+            $expand: 'scheduledActionsForRule($expand=scheduledActionConfigurations)',
+          },
+        },
+        transformation: {
+          ...DEFAULT_TRANSFORMATION,
+          omit: ['version', 'scheduledActionsForRule@odata.context'],
```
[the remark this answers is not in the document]
shoshana-houri: There isn't really a specific reason :)
My intuition was the opposite—"if we can omit them early in the first transformation, let's do that." For example, if we omit `scheduledActionsForRule@odata.context` under `fieldDefinitions`, it means it will already be transformed to NACL case, which seems a bit redundant. What do you think?

packages/microsoft-security-adapter/src/definitions/deploy/intune/deploy.ts
shoshana-houri: Sadly no, because it's being executed before `pick` is executed, and we have to only specify this field in the request.
I also tried combining `root` & `nestUnderField`, but `root` transformation uses `flatMap`, which is problematic here since this value is an array and we want it to remain such 🥲

packages/microsoft-security-adapter/src/definitions/fetch/intune/fetch.ts
shoshana-houri: It is, I just thought it would be clearer this way 😅

packages/microsoft-security-adapter/src/definitions/deploy/intune/deploy.ts
shoshana-houri: Taking this back 🤦‍♀️ I can simply pick the new field name 😅

PR #6507, opened 2024-09-01
SALTO-6428: Change Microsoft Security type names to PascalCase

shoshana-houri:
Change all Microsoft Security type names to PascalCase, following the discussion here

---
This was done following the discussion [here](https://github.com/salto-io/salto/pull/6456#discussion_r1727124764).

---
_Release Notes_: 

---
_User Notifications_:

### DIFF
--- a/packages/lowerdash/src/types.ts
+++ b/packages/lowerdash/src/types.ts
@@ -105,31 +105,3 @@ export type RecursivePartial<T> = {
       ? RecursivePartial<T[P]>
       : T[P]
 }
-
-export type UPPER_CASE_LETTERS =
-  | 'A'
-  | 'B'
-  | 'C'
-  | 'D'
-  | 'E'
-  | 'F'
-  | 'G'
-  | 'H'
-  | 'I'
-  | 'J'
-  | 'K'
-  | 'L'
-  | 'M'
-  | 'N'
-  | 'O'
-  | 'P'
-  | 'Q'
-  | 'R'
-  | 'S'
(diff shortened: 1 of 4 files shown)


packages/lowerdash/src/types.ts
```diff
@@ -105,3 +105,31 @@ export type RecursivePartial<T> = {
       ? RecursivePartial<T[P]>
       : T[P]
 }
+
+export type UPPER_CASE_LETTERS =
```
shoshana-houri, line 109: i know it's kind of a lot 😅 but I think it's useful.
Let me know if you disagree or if you think we should locate this someplace else :)

shir-reifenberg (approved):
Added one comment, pre approving 👯

shoshana-houri:
Thanks @shir-reifenberg! For some reason I can't see the comment  😅

packages/lowerdash/src/types.ts
shir-reifenberg: IMO it's a bit of a overkill 😅 We can just make `toEntraTypeName`  to get a string

shir-reifenberg:
> Thanks @shir-reifenberg! For some reason I can't see the comment 😅

Oops not sure why is wasn't submitted, added now 👀

packages/lowerdash/src/types.ts
shoshana-houri: Oh well, it was worth the shot 😄

PR #6510, opened 2024-09-01
SALTO-6592: fix elementsSource has implementation

shayc331:
_fix elementsSource has implementation_

---

* it is a fix for the change from this [pr](https://github.com/salto-io/salto/pull/6505). I changed `elementsSource.has` implementation to use `elementsSource.get` inside, which may cause performance issues.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/adapter-utils/src/element_source.ts
+++ b/packages/adapter-utils/src/element_source.ts
@@ -67,14 +67,14 @@ export const buildElementsSourceFromElements = (
     }
   }
 
-  const getElement = (id: ElemID): Value => {
+  const retrieveValueByID = (id: ElemID): Value => {
     const { parent } = id.createTopLevelParentID()
     const topLevelElement = elementsMap[parent.getFullName()]
     return elementsMap[id.getFullName()] ?? (topLevelElement && resolvePath(topLevelElement, id))
   }
 
   const get = async (id: ElemID): Promise<Value> => {
-    const element = getElement(id)
+    const element = retrieveValueByID(id)
     if (element !== undefined) {
       return element
     }
@@ -85,7 +85,7 @@ export const buildElementsSourceFromElements = (
   }
 
   const has = async (id: ElemID): Promise<boolean> =>
-    getElement(id) !== undefined || awu(fallbackSources).some(async source => source.has(id))
+    retrieveValueByID(id) !== undefined || awu(fallbackSources).some(source => source.has(id))
 
   self = {
     getAll: async () => getElements(),
--- a/packages/adapter-utils/src/element_source.ts
+++ b/packages/adapter-utils/src/element_source.ts
@@ -67,10 +67,14 @@ export const buildElementsSourceFromElements = (
     }
   }
 
-  const get = async (id: ElemID): Promise<Value> => {
+  const getElement = (id: ElemID): Value => {
     const { parent } = id.createTopLevelParentID()
     const topLevelElement = elementsMap[parent.getFullName()]
-    const element = elementsMap[id.getFullName()] ?? (topLevelElement && resolvePath(topLevelElement, id))
+    return elementsMap[id.getFullName()] ?? (topLevelElement && resolvePath(topLevelElement, id))
+  }
+
+  const get = async (id: ElemID): Promise<Value> => {
+    const element = getElement(id)
     if (element !== undefined) {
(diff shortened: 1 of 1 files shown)


packages/adapter-utils/src/element_source.ts
```diff
@@ -67,10 +67,14 @@ export const buildElementsSourceFromElements = (
     }
   }
 
-  const get = async (id: ElemID): Promise<Value> => {
+  const getElement = (id: ElemID): Value => {
     const { parent } = id.createTopLevelParentID()
     const topLevelElement = elementsMap[parent.getFullName()]
-    const element = elementsMap[id.getFullName()] ?? (topLevelElement && resolvePath(topLevelElement, id))
+    return elementsMap[id.getFullName()] ?? (topLevelElement && resolvePath(topLevelElement, id))
```
ori-moisis, line 73: `resolvePath` can handle the ID being the top level ID, no need to perform another lookup in the elements map
```suggestion
    return topLevelElement && resolvePath(topLevelElement, id)
```

packages/adapter-utils/src/element_source.ts
```diff
@@ -67,10 +67,14 @@ export const buildElementsSourceFromElements = (
     }
   }
 
-  const get = async (id: ElemID): Promise<Value> => {
+  const getElement = (id: ElemID): Value => {
```
ori-moisis, line 70: this doesn't return an element, the name is misleading

packages/adapter-utils/src/element_source.ts
```diff
@@ -80,7 +84,8 @@ export const buildElementsSourceFromElements = (
       .find(values.isDefined)
   }
 
-  const has = async (id: ElemID): Promise<boolean> => (await get(id)) !== undefined
+  const has = async (id: ElemID): Promise<boolean> =>
+    getElement(id) !== undefined || awu(fallbackSources).some(async source => source.has(id))
```
ori-moisis, line 88: nit: redundant `async` definition
```suggestion
    getElement(id) !== undefined || awu(fallbackSources).some(source => source.has(id))
```
this actually has an effect on performance in large enough numbers, not the case here, but still...

packages/adapter-utils/src/element_source.ts
dantal4: Theoretically, `elements` may contain fields without their parent types.

ori-moisis: really?
if that is the case, then the code should have taken `baseID`

ori-moisis: AFAICT all places in the code only pass in top level elements.

maybe worth adding a warn log if someone tries to build an element source here with something that isn't a top level element?

dantal4: I prefer checking them both, as I see a situation when somebody uses this function in deploy, where changes have fields in them.
```typescript
    const baseId = id.createBaseID().parent
    const topLevelParentId = id.createTopLevelParentID().parent
    const element = elementsMap[baseId.getFullName()] ?? elementsMap[topLevelParentId.getFullName()]
    return element && resolvePath(element, id)
```

shayc331: I agree with @dantal4

ori-moisis (commented):
please fill in the PR description

packages/adapter-utils/src/element_source.ts
```diff
@@ -67,10 +67,15 @@ export const buildElementsSourceFromElements = (
...
+    const topLevelParentId = id.createTopLevelParentID().parent
+    const element = elementsMap[baseId.getFullName()] ?? elementsMap[topLevelParentId.getFullName()]
+    return element && resolvePath(element, id)
+  }
+
   const get = async (id: ElemID): Promise<Value> => {
-    const { parent } = id.createTopLevelParentID()
-    const topLevelElement = elementsMap[parent.getFullName()]
-    const element = elementsMap[id.getFullName()] ?? (topLevelElement && resolvePath(topLevelElement, id))
+    const element = getValueByID(id)
```
ori-moisis, line 78: nit: not an element

ori-moisis (approved):

PR #6515, opened 2024-09-02
SAAS-11985: fix getImportantValues

edenhassid:
fix getImportantValues

---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/adapter-utils/src/important_values.ts
+++ b/packages/adapter-utils/src/important_values.ts
@@ -14,6 +14,7 @@ import {
   isField,
   isInstanceElement,
   isObjectType,
+  isPlaceholderObjectType,
   isPrimitiveValue,
   isReferenceExpression,
   ObjectType,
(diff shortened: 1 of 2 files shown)


packages/adapter-utils/src/important_values.ts
```diff
@@ -127,9 +128,19 @@ export const getImportantValues = async ({
     return extractImportantValuesFromElement({ importantValues, element, indexedOnly, highlightedOnly })
   }
   if (isField(element) || isInstanceElement(element)) {
-    try {
+    const getTypeObj = async (): Promise<ObjectType> => {
       const typeObj = await element.getType(elementSource)
-      const importantValues = typeObj?.annotations[CORE_ANNOTATIONS.IMPORTANT_VALUES]
+      return isPlaceholderObjectType(typeObj) ? elementSource?.get(typeObj.elemID) : typeObj
```
netama, line 133: nit: maybe add a comment explaining that we reach this flow for types that are hidden?

packages/adapter-utils/src/important_values.ts
```diff
@@ -127,9 +128,19 @@ export const getImportantValues = async ({
...
+    }
+    try {
+      const typeObj = await getTypeObj()
+      if (typeObj === undefined) {
+        log.trace(
+          `could not get important values as type is undefined returning [] for element ${element.elemID.getFullName()}`,
```
netama, line 139: tiny nit:
```suggestion
          `could not get important values as type is undefined, returning [] for element ${element.elemID.getFullName()}`,
```

packages/adapter-utils/test/important_values.test.ts
```diff
@@ -160,6 +161,30 @@ describe('getImportantValues', () => {
       { key: 'doesNotExist', value: undefined },
     ])
   })
+  it('should get the right important values for an instance when obj is', async () => {
```
netama, line 164: nit: unfinished sentence?

netama (approved):

packages/adapter-utils/test/important_values.test.ts
edenhassid: missed it 😅

PR #6518, opened 2024-09-02
SALTO-6517: fix static file path collision

IdoZakk:
Fixed the static file path collision in Jira

---

* It makes more sense that all adapters will do it similarly, so I added a function to adapter-utils and moved the okta solution to use it.
* The solution is based on the[ solution agreed upon in Okta](https://github.com/salto-io/salto/pull/6477) following a conversation with @netama 
* A change was done in jira to use the elemID name instead of the name in the value of the instance. It seems safer, as in some cases in Jira to API allows for two elements with the same name, and we can have different elemIDs for them (for instance if one of them was changed after starting to use Salto)

Added [noise reduction](https://github.com/salto-io/salto_private/pull/9244)

---
_Release Notes_: 
Jira Adapter:
* Fixed a potential bug with static file path for icons

---
_User Notifications_: 
Jira Adapter:
* The static file path of icons will change to avoid collisions

### DIFF
--- a/packages/adapter-utils/src/nacl_case_utils.ts
+++ b/packages/adapter-utils/src/nacl_case_utils.ts
@@ -25,6 +25,9 @@ const allCapsCamelCaseRegex = /[A-Z]([A-Z][a-z])/g
 export const pathNaclCase = (name?: string): string =>
   (name ? name.split(NACL_ESCAPING_SUFFIX_SEPARATOR)[0] : '').slice(0, MAX_PATH_LENGTH)
 
+// Converts a nacl case to a file system safe name. Use it (instead of pathNaclCase) if the file name must be unique
+export const fileNameNaclCase = (name: string): string => name.replace('@', '.')
+
 // Trim part of a file name to comply with filesystem restrictions
 // This assumes the filesystem does not allow path parts to be over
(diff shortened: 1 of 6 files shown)


idoamit8 (approved):
Nice work! Thanks for consolidating the code 💯

packages/adapter-utils/src/nacl_case_utils.ts
```diff
@@ -25,6 +25,9 @@ const allCapsCamelCaseRegex = /[A-Z]([A-Z][a-z])/g
 export const pathNaclCase = (name?: string): string =>
   (name ? name.split(NACL_ESCAPING_SUFFIX_SEPARATOR)[0] : '').slice(0, MAX_PATH_LENGTH)
 
+// Converts a nacl case to a file system safe name. Use it (instead of pathNaclCase) if the file name must be unique
+export const fileNameNaclCase = (name: string): string => name.replace('@', '.')
+
```
shoshana-houri, line 30: Added it in this PR https://github.com/salto-io/salto/pull/6625
As [suggested](https://github.com/salto-io/salto/pull/6625#discussion_r1775669948) by @shir-reifenberg

PR #6520, opened 2024-09-03
SALTO-6605 SOAP WSDL version config

dantal4:
Allow users to config the SOAP WSDL version used by `SoapClient` - using the `suiteAppClient.wsdlVersion` adapter config.

---

---
_Release Notes_: 
Netsuite Adapter:
- Allow users to config the SOAP WSDL version used by `SoapClient` - using the `suiteAppClient.wsdlVersion` adapter config

---
_User Notifications_: 
None

### DIFF
--- a/packages/netsuite-adapter/src/client/suiteapp_client/soap_client/types.ts
+++ b/packages/netsuite-adapter/src/client/suiteapp_client/soap_client/types.ts
@@ -5,6 +5,11 @@
  *
  * CERTAIN THIRD PARTY SOFTWARE MAY BE CONTAINED IN PORTIONS OF THE SOFTWARE. See NOTICE FILE AT https://github.com/salto-io/salto/blob/main/NOTICES
  */
+
+export const SUPPORTED_WSDL_VERSIONS = ['2020_2', '2023_1', '2023_2', '2024_1'] as const
+
+export type WSDLVersion = (typeof SUPPORTED_WSDL_VERSIONS)[number]
+
 type StatusSuccess = {
   attributes: {
     isSuccess: 'true'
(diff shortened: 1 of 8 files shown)


packages/netsuite-adapter/src/client/suiteapp_client/soap_client/types.ts
```diff
@@ -5,6 +5,11 @@
  *
  * CERTAIN THIRD PARTY SOFTWARE MAY BE CONTAINED IN PORTIONS OF THE SOFTWARE. See NOTICE FILE AT https://github.com/salto-io/salto/blob/main/NOTICES
  */
+
+export const SUPPORTED_WSDL_VERSIONS = ['2020_2', '2023_1', '2023_2', '2024_1'] as const
```
omrilit, line 9: Do you think that we should allow also future versions by having a regex validation in the adapter_creator?
Just so we will have the option to upgrade to newer versions without code changes.
This is probably not needed for now and we can consider doing it in the future if this is something you'd prefer

packages/netsuite-adapter/src/config/validations.ts
```diff
@@ -72,6 +73,26 @@ function validateNumber(value: unknown, configPath: string | string[]): asserts
   }
 }
 
+function validateEnumValue(
+  value: unknown,
+  enumValues: ReadonlyArray<string>,
+  configPath: string | string[],
+): asserts value is string | undefined {
+  if (value === undefined) {
```
omrilit, line 81: You already check outside whether the value is undefined, maybe one of them is redundant?

omrilit (approved):
NICE!

packages/netsuite-adapter/src/client/suiteapp_client/soap_client/types.ts
dantal4: if a user will use a version that match the regex and doesn't exist, they'll get an exception from SOAP, so I think that I prefer that we'll manage the allowed versions.

PR #6521, opened 2024-09-03
SALTO-6344 delay after creation or deletion on Jamf

omrifarkash:
We need to add some busy waiting after POST or DELETE api calls on Jamf classicAPI
Please see https://community.jamf.com/t5/jamf-pro/delay-time-after-post-or-delete-api-calls/m-p/323383/emcs_t/S2h8ZW1haWx8dG9waWNfc3Vic2NyaXB0aW9ufE0wM1dLNk4xNU4yMlBFfDMyMzM4M3xTVUJTQ1JJUFRJT05TfGhL#M278490

---

---

_Release Notes_: 
_Jamf Adapter_:
- Fix deployment fails when there are dependencies on creation or deletion

---

_User Notifications_: 
_Jamf Adapter_:
- Fix deployment fails when there are dependencies on creation or deletion

### DIFF
--- a/packages/jamf-adapter/src/config.ts
+++ b/packages/jamf-adapter/src/config.ts
@@ -12,11 +12,15 @@ export type UserFetchConfig = definitions.UserFetchConfig<{
   fetchCriteria: definitions.DefaultFetchCriteria
 }>
 
+export type UserDeployConfig = definitions.UserDeployConfig & {
+  delayAfterDeploy: number
+}
+
 export type UserConfig = definitions.UserConfig<
   never,
   definitions.ClientBaseConfig<definitions.ClientRateLimitConfig>,
   UserFetchConfig,
-  definitions.UserDeployConfig
+  UserDeployConfig
 >
 
 export const DEFAULT_CONFIG: UserConfig = {
--- a/packages/jamf-adapter/src/config.ts
+++ b/packages/jamf-adapter/src/config.ts
@@ -13,7 +13,7 @@ export type UserFetchConfig = definitions.UserFetchConfig<{
 }>
 
 export type UserDeployConfig = definitions.UserDeployConfig & {
-  delayAfterDeploy: number
+  delayAfterDeploy?: number
 }
 
 export type UserConfig = definitions.UserConfig<
--- a/packages/jamf-adapter/src/constants.ts
+++ b/packages/jamf-adapter/src/constants.ts
@@ -19,3 +19,4 @@ export const RESULTS = 'results'
 export const OS_X_CONFIGURATION_PROFILE_TYPE_NAME = 'os_x_configuration_profile'
 export const MOBILE_DEVICE_CONFIGURATION_PROFILE_TYPE_NAME = 'mobile_device_configuration_profile'
 export const MAC_APPLICATION_TYPE_NAME = 'mac_application'
+export const DUMMY_DELAY_TIMEOUT = 50 * 1000 // 50 seconds in ms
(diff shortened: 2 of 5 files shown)


packages/jamf-adapter/src/definitions/deploy/classic_api_utils.ts
```diff
@@ -58,6 +65,7 @@ export const createClassicApiDefinitionsForType = (
                   throw new Error('Expected value to be a record')
                 }
                 const xmlVal = xmljs.js2xml({ [typeName]: { ...value } }, { compact: true })
+                await dummyDelay(userConfig.deploy?.delayAfterDeploy)
```
shir-reifenberg, line 68: Isn't this an issue that the `adjust` function is actually called before the API call and we want to add the delay **after** the API call ?
For example if we have a single change in the deployment, we'll delay the request and then sent it and return, and still the subsequent fetch might delete the recently added element, no?

shir-reifenberg (commented):
Added one question, I think we might end up wanting to add something to the client definition instead

packages/jamf-adapter/src/definitions/deploy/classic_api_utils.ts
omrifarkash: You're this is not what I meant to do. Changing it

packages/jamf-adapter/src/config.ts
```diff
@@ -12,11 +12,15 @@ export type UserFetchConfig = definitions.UserFetchConfig<{
   fetchCriteria: definitions.DefaultFetchCriteria
 }>
 
+export type UserDeployConfig = definitions.UserDeployConfig & {
+  delayAfterDeploy?: number
```
shir-reifenberg, line 16: Can you please verify this value don't get written to the adapter.nacl for new envs?
(can check this by running `salto workspace clean -g`, to reset to config to its default state)

packages/jamf-adapter/src/constants.ts
```diff
@@ -19,3 +19,4 @@ export const RESULTS = 'results'
 export const OS_X_CONFIGURATION_PROFILE_TYPE_NAME = 'os_x_configuration_profile'
 export const MOBILE_DEVICE_CONFIGURATION_PROFILE_TYPE_NAME = 'mobile_device_configuration_profile'
 export const MAC_APPLICATION_TYPE_NAME = 'mac_application'
+export const DEFAULT_POST_DEPLOY_DELAY = 50 * 1000 // 50 seconds in ms
```
shir-reifenberg, line 22: nit: move to classic_api_utils since it's not used anywhere else

packages/jamf-adapter/src/definitions/deploy/classic_api_utils.ts
```diff
@@ -102,6 +111,16 @@ export const createClassicApiDefinitionsForType = (
               },
             },
           },
+          copyFromResponse: {
+            additional: {
+              adjust: async () => {
+                // We do not really copy from response
+                // This is a hack to run postDeployDelay function after the API call
```
shir-reifenberg, line 118: can you please have this comment also in line 48?

shir-reifenberg (approved):
LGTM 👯 
(you don't need to have user notification in this PR cause there are no expected changes to env)

packages/jamf-adapter/src/definitions/deploy/classic_api_utils.ts
omrifarkash: On this line we do copyFromResponse

PR #6524, opened 2024-09-03
SALTO-6614: Create fetch warning on Brand Logo 403

shir-reifenberg:
Until now, when we got 403 while trying to get brand logos, the entire fetch failed, we now convert it into a fetch warning, like we do to other types fetched with the infra.

---

---
_Release Notes_: 

_Zendesk_adapter_:
- Create fetch warning when failing to get brand logos due to insufficient permissions, instead of failing the entire fetch

---
_User Notifications_: 
None


edenhassid (approved):
LGTM! 💯 🔥

PR #6525, opened 2024-09-04
SAAS-6615 - Added option to configure extraDependencies initial chunk size

amitgoldarb:
Added option to configure extraDependencies initial chunk size. It seems like the API sometimes returns partial information and we don't catch it and chunk to smaller requests. I've added a log to see the response size which currently isn't visibile and added an option to allow to configure the default chunk size.

---

None.

---
_Release Notes_: 
Salesforce: 
* Add option to configure extraDependencies requests chunk size possibly allowing to stabalize generated dependencies results.

---
_User Notifications_: 
None.


packages/salesforce-adapter/src/types.ts
```diff
@@ -824,6 +829,16 @@ const optionalFeaturesType = createMatchingObjectType<OptionalFeatures>({
   },
 })
 
+const optionalDefaultsType = createMatchingObjectType<OptionalDefaults>({
+  elemID: new ElemID(constants.SALESFORCE, 'optionalDefaults'),
+  fields: {
+    extraDependenciesChunkSize: { refType: BuiltinTypes.NUMBER },
```
tamtamirr, line 835: Please move this into the `client` section of the config as this is advanced setting that we don't expect customers to configure by themselves

packages/salesforce-adapter/src/fetch_profile/fetch_profile.ts
```diff
@@ -80,6 +81,7 @@ export const buildFetchProfile = ({
     metadataQuery,
     maxItemsInRetrieveRequest,
     importantValues: mergeWithDefaultImportantValues(additionalImportantValues),
+    optionalDefaults: optionalDefaults ?? {},
```
tamtamirr, line 84: Receive **extraDependenciesChunkSize**  and expose it through the fetchProfile directly.
```suggestion
    extraDependenciesChunkSize,
```

packages/salesforce-adapter/src/filters/extra_dependencies.ts
```diff
@@ -159,13 +161,19 @@ const queryDeps = async ({ client, elements }: QueryDepsParams): Promise<Depende
     RefMetadataComponentId, RefMetadataComponentType, RefMetadataComponentName 
   FROM MetadataComponentDependency WHERE MetadataComponentId IN (${idsChunk.map(id => `'${id}'`).join(', ')})`
           const allRecords = (await toArrayAsync(await client.queryAll(query, true))).flat()
+          log.debug(
+            'Queried %d dependencies for %d elements, the last 3 elements are',
+            allRecords.length,
+            idsChunk.length,
+            idsChunk.slice(-3),
+          )
```
tamtamirr, line 169:
```suggestion
          log.debug(
            'Queried %d dependencies for %d elements, the last 3 elements are %s',
            allRecords.length,
            idsChunk.length,
            inspectValue(idsChunk.slice(-3)),
          )
```

packages/salesforce-adapter/src/filters/extra_dependencies.ts
```diff
@@ -62,6 +62,7 @@ type Dependency = {
 type QueryDepsParams = {
   client: SalesforceClient
   elements: Element[]
+  initialChunkSize?: number
```
tamtamirr, line 65:
```suggestion
  initialChunkSize: number
```
Shouldn't be optional at this point IMO. Pass the default when undefined.

packages/salesforce-adapter/src/filters/extra_dependencies.ts
```diff
@@ -323,7 +334,11 @@ const creator: RemoteFilterCreator = ({ client, config }) => ({
     filterName: 'extraDependencies',
     fetchFilterFunc: async (elements: Element[]) => {
       const groupedDeps = config.fetchProfile.isFeatureEnabled('extraDependenciesV2')
-        ? await getDependenciesV2({ client, elements })
+        ? await getDependenciesV2({
+            client,
+            elements,
+            initialChunkSize: config.fetchProfile.optionalDefaults.extraDependenciesChunkSize,
```
tamtamirr, line 340: Please add UT that makes sure the filter uses the configured chunk size when it's defined in the fetch profile.

tamtamirr (commented):
Great Work! 👑 Added some comments and suggestions

packages/salesforce-adapter/src/types.ts
amitgoldarb: I aligned a bit differently. I added a limits section in the fetch config and I think we can move all of the max<%s> things under there. I'll open a ticket if you like it also.

tamtamirr (approved):
LGTM!

PR #6526, opened 2024-09-04
SALTO-6618: Avoid using persistent workspace for read only commands

ori-moisis:
---

This was mildly annoying that I couldn't run `salto element print` when I tried to debug something...

---
_Release Notes_: 
Cli:
- Enable running commands that do not change the workspace while other commands are running

---
_User Notifications_: 
_None_


tamtamirr (approved):
LGTM!

PR #6527, opened 2024-09-04
SALTO-6599: [Intune] Add fetch & deploy for application group assignments

shoshana-houri:
* Support fetching and deploying group assignments for Intune application types.
* Update the OdataType renaming to camelCase to align with our new naming in Microsoft Security adapter (avoiding the mix of snake_case and xCase :) ).

---
This PR introduces group assignments and is followed by another one that generalizes this concept for other types.
Handling application group assignments required special treatment—we encountered an issue when adding a `managedGooglePlayApp` app type and its assignments—the endpoint does not return the newly created object id, which required us to manually fetch it (using polling, since it doesn't update instantly 🥲) and then use it to assign the groups.

---
_Release Notes_: 
_Microsoft Security adapter:_
* Support fetch & deploy for application group assignments.

---
_User Notifications_: 
_Microsoft Security adapter:_
* Support fetch & deploy for application group assignments.


packages/microsoft-security-adapter/src/definitions/deploy/intune/utils/application.ts
```diff
@@ -50,3 +59,29 @@ export const transformManagedGooglePlayApp: AdjustFunctionSingle = async ({ valu
...
+          to: ASSIGNMENT_ROOT_FIELD_NAME,
+          onConflict: 'skip',
+        },
+      ],
+      pick: [ASSIGNMENT_ROOT_FIELD_NAME],
+    },
+  },
+  // We can't simply use transformToCheck.pick since this request is also used for addition changes,
+  // while transformToCheck is only valid for modification changes.
+  condition: createCustomConditionCheckChangesInFields([ASSIGNMENTS_FIELD_NAME]),
```
shir-reifenberg, line 86: I think you don't need to share this part, no?
you can use `transformForCheck` with `pick` for the modification request and use a different condition for the addition one

packages/microsoft-security-adapter/src/definitions/deploy/intune/utils/application.ts
```diff
@@ -9,15 +9,24 @@
...
+  APPLICATION_TYPE_NAME,
+  APP_IDENTIFIER_FIELD_NAME,
+  APP_STORE_URL_FIELD_NAME,
+  PACKAGE_ID_FIELD_NAME,
+  ASSIGNMENTS_FIELD_NAME,
+} = intuneConstants
 
 const { isManagedGooglePlayApp, isAndroidEnterpriseSystemApp } = intuneUtils.application
 
+export const GET_MANAGED_STORE_APP_POST_DEPLOY_PATH: EndpointPath = `/deviceAppManagement/mobileApps?$filter=(isof('microsoft.graph.androidManagedStoreApp') and (microsoft.graph.androidManagedStoreApp/${APP_IDENTIFIER_FIELD_NAME} eq '{${APP_IDENTIFIER_FIELD_NAME}}'))`
```
shir-reifenberg, line 28: Is there any chance this request might return more than one element ?

packages/microsoft-security-adapter/src/definitions/requests/clients.ts
```diff
@@ -44,6 +46,15 @@ export const createClientDefinitions = (
               },
             },
           },
+          [`/beta${GET_MANAGED_STORE_APP_POST_DEPLOY_PATH}` as EndpointPath]: {
+            get: {
+              polling: {
+                interval: 5000,
+                retries: 6,
```
shir-reifenberg, line 53: are those numbers based on some documentation? if so, can you please share it?

shir-reifenberg (commented):
Looks great 🕺 added some comments / questions
Few additional questions -
1. Can you please share what were the consideration of keeping the assignments inside the application instance vs extracting to standalone ?
2. I saw we omit the assignment id, are those not needed for updates ?

packages/microsoft-security-adapter/src/definitions/deploy/intune/deploy.ts
```diff
@@ -16,13 +16,17 @@ import { application as applicationDeployUtils, appsConfiguration } from './util
...
 
 const {
+  // Type names
   APPLICATION_TYPE_NAME,
   APPLICATION_CONFIGURATION_MANAGED_APP_TYPE_NAME,
   APPLICATION_CONFIGURATION_MANAGED_DEVICE_TYPE_NAME,
   DEVICE_CONFIGURATION_TYPE_NAME,
   DEVICE_CONFIGURATION_SETTING_CATALOG_TYPE_NAME,
   DEVICE_COMPLIANCE_TYPE_NAME,
+
```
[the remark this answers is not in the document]
shoshana-houri: As always lol

image

packages/microsoft-security-adapter/src/definitions/deploy/intune/deploy.ts
```diff
@@ -71,6 +79,29 @@ const graphBetaCustomDefinitions: DeployCustomDefinitions = {
...
+            condition: {
+              custom:
+                () =>
+                ({ change }) =>
+                  application.isManagedGooglePlayApp(getChangeData(change).value),
+            },
+            copyFromResponse: {
+              additional: {
+                root: 'value',
+                pick: ['id'],
```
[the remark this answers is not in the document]
shoshana-houri: AFAIK it won't:
`peak` accepts an array, but it simply picks all the fields specified in the array. It's different than `get` which extract the specified fields level after level.
i.e.:
```
const obj = {
  value: {
    id: 'nested-field',
  },
  id: 'top-level'
  otherField: 'top-level'
}

_.pick(obj, ['value', 'id]) ->> {value: {id: 'nested-field'}, id: 'top-level'}

File: packages/microsoft-security-adapter/src/definitions/deploy/intune/utils/application.ts (hunk 1 above)
Comment (shoshana-houri) [reply packages/microsoft-security-adapter/src/definitions/deploy/intune/utils/application.ts:86]: We can, but then we'll need to differentiate between the actions, which is something I preferred to avoid 🤔

File: packages/microsoft-security-adapter/src/definitions/deploy/intune/utils/application.ts (hunk 2 above)
Comment (shoshana-houri) [reply packages/microsoft-security-adapter/src/definitions/deploy/intune/utils/application.ts:28]: Not AFAIK, the app Id is unique
`.../${APP_IDENTIFIER_FIELD_NAME} eq '{${APP_IDENTIFIER_FIELD_NAME}`

**Comment by shoshana-houri** [comment]
> Looks great 🕺 added some comments / questions Few additional questions -
> 
> 1. Can you please share what were the consideration of keeping the assignments inside the application instance vs extracting to standalone ?
> 2. I saw we omit the assignment id, are those not needed for updates ?

Thanks @shir-reifenberg!
Re standalone:
1. There are two options for a standalone structure in this case (I'm not sure which one you're referring to, so I'll cover both):
Each assignment as a top-level element: This could be problematic since they are deployed together.
2. All assignments as a single top-level element: This is possible, and I don't have a strong objection to it. Honestly, I just didn't consider this option. Some considerations:
    1. A clear advantage of this approach is that if the assignment request fails, we can still correctly indicate that the parent's request succeeded and update the state accordingly.
    2. It would still require specifying each type that has assignments since they wouldn't share the same type name (e.g., ${parentTypeName}__Assignments). 
    3. UX wise: It might be "easier" for the user to deploy the parent without its assignments—whether intentionally or not. I'm not sure if that’s a good or bad thing :) In the UI, the assignments are treated as part of the other properties of the parent.

Re the assignment id:
They're not needed :) The received id is simply `${parentId}_${groupId}`

File: packages/microsoft-security-adapter/src/definitions/deploy/intune/deploy.ts (hunk 4 above)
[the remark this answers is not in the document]
Comment (shir-reifenberg) [reply packages/microsoft-security-adapter/src/definitions/deploy/intune/deploy.ts:26]: 🤣 🤣 🤣 LOL

File: packages/microsoft-security-adapter/src/definitions/deploy/intune/deploy.ts (hunk 5 above)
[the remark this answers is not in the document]
Comment (shir-reifenberg) [reply packages/microsoft-security-adapter/src/definitions/deploy/intune/deploy.ts:100]: you're correct, sorry :)

File: packages/microsoft-security-adapter/src/definitions/deploy/intune/utils/application.ts (hunk 1 above)
Comment (shir-reifenberg) [reply packages/microsoft-security-adapter/src/definitions/deploy/intune/utils/application.ts:86]: IMO we should prefer using the shared infra code over custom when possible, and I not sure there's a justification for splitting this one ?
(also IMO this is a little more readable)
```
{
  ...ASSIGNMENT_REQUEST,
  conditions: { transformForCheck: { pick: [ASSIGNMENTS_FIELD_NAME] } }  // and have a custom func for addition field
}
```

shir-reifenberg:
> > Looks great 🕺 added some comments / questions Few additional questions -
> > 
> > 1. Can you please share what were the consideration of keeping the assignments inside the application instance vs extracting to standalone ?
> > 2. I saw we omit the assignment id, are those not needed for updates ?
> 
> Thanks @shir-reifenberg!
> 
> **Re standalone:**
> 
> 1. There are two options for a standalone structure in this case (I'm not sure which one you're referring to, so I'll cover both):
>    Each assignment as a top-level element: This could be problematic since they are deployed together.
> 2. All assignments as a single top-level element: This is possible, and I don't have a strong objection to it. Honestly, I just didn't consider this option. Some considerations:
>    
>    1. A clear advantage of this approach is that if the assignment request fails, we can still correctly indicate that the parent's request succeeded and update the state accordingly.
>    2. It would still require specifying each type that has assignments since they wouldn't share the same type name (e.g., ${parentTypeName}__Assignments).
>    3. UX wise: It might be "easier" for the user to deploy the parent without its assignments—whether intentionally or not. I'm not sure if that’s a good or bad thing :) In the UI, the assignments are treated as part of the other properties of the parent.
> 
> What do you think?
> 
> **Re the assignment id:** They're not needed :) The received id is simply `${parentId}_${groupId}`

Actually I first thought about the first option, but I agree that considering how things being deployed it makes more sense to think about the second option you mentioned :)
I don't have a have a strong opinion, was mainly curious to understand what were the considerations. 
another advantage for 2 I have in mind is that it's easier to exclude if needed.

but I think we can start this way and revisit if and when it becomes relevant 🚀

shir-reifenberg (approved):

packages/microsoft-security-adapter/src/definitions/deploy/intune/utils/application.ts
shoshana-houri: I agree about using the shared infra code. I just think that in this case it's a bit more confusing having "different" conditions in requests that should be identical. It also adds logic which we can avoid (the `condition: createCustomConditionCheckChangesInFields([ASSIGNMENTS_FIELD_NAME]),` will appear anyway in one place or the other).

shir-reifenberg: you can keep it like this, just my personal preference :)

PR #6530, opened 2024-09-05
SALTO-6600: [Intune] Support fetch & deploy for Filters

shoshana-houri:
Support fetching & deploying Intune filters - including filters assignment (ref).

---

See Filter nacls structure example [here](https://github.com/salto-io/microsoft-security/tree/main/microsoft_security/Records/IntuneFilter).
See assignment ref structure example [here](https://github.com/salto-io/microsoft-security/blob/9b91f94ad320211601d7e06331b39c7888901164/microsoft_security/Records/IntuneApplicationConfigurationManagedApp/AppConfigurationPolicy.nacl#L91).

---
_Release Notes_: 
_Microsoft Security adapter:_
* Support fetch & deploy for Intune Filters.

---
_User Notifications_: 
_Microsoft Security adapter:_
* Support fetch & deploy for Intune Filters.


packages/microsoft-security-adapter/src/definitions/fetch/intune/fetch.ts
```diff
@@ -305,6 +306,32 @@ const graphBetaCustomizations: FetchCustomizations = {
...
+      },
+    ],
+    resource: {
+      directFetch: true,
+    },
+    element: {
+      topLevel: {
+        isTopLevel: true,
+        serviceUrl: {
+          baseUrl: SERVICE_BASE_URL,
```
shir-reifenberg, line 328: nit: if this applies to all / most types, I think you can move this part to the `default` config

packages/adapter-components/src/deployment/definition_helpers.ts
```diff
@@ -16,6 +16,7 @@ export const DEFAULT_CONTEXT = {
 type StandardDeployArgs<AdditionalAction extends string, ClientOptions extends string> = {
   bulkPath: string
   idField?: string
+  modificationMethod?: 'put' | 'patch'
```
shir-reifenberg, line 19: please just update this comment as well - 
`* - update: PUT on <base_path>/<typename>/<:id_field>`

packages/microsoft-security-adapter/src/definitions/deploy/shared/deploy.ts
```diff
@@ -13,17 +13,8 @@ import { createEntraCustomizations } from '../entra/deploy'
...
-    ClientOptions
-  >({})
-
-  return _.merge(standardRequestDefinitions, {
-    ...createEntraCustomizations(),
-    ...createIntuneCustomizations(),
-  })
-}
+const createCustomizations = (): DeployCustomDefinitions =>
+  _.merge(createEntraCustomizations(), createIntuneCustomizations())
```
shir-reifenberg, line 17: Not for this PR but just something to think about and avoid when possible - I think the issue with `_.merge` is that it will override duplicate definitions.
In comparison using plain assignments would have create error and make the developer aware

shir-reifenberg (approved):
Looks great 🎈  approving with one question - is filter's `displayName` unique across platforms? i.e is it valid / common to have 2 filters with the same name but for different platforms ? if so, we might want to consider adding this to the elemID

packages/microsoft-security-adapter/src/definitions/deploy/shared/deploy.ts
shoshana-houri: By plain assignments you mean creating Intune and Entra definitions in the same object?

shoshana-houri:
> Looks great 🎈 approving with one question - is filter's `displayName` unique across platforms? i.e is it valid / common to have 2 filters with the same name but for different platforms ? if so, we might want to consider adding this to the elemID

Thanks @shir-reifenberg, good point! Will update the elemID as you suggested :)

packages/microsoft-security-adapter/src/definitions/deploy/shared/deploy.ts
shir-reifenberg: yap, but we can also validate it just before merge (with code)
anyway not important enough for now :)

PR #6534, opened 2024-09-08
SAAS-12017: Add updateElementFolder

dorsev:
Add a new function `dumpElementsToFolder` which updates a “hybrid” folder - a folder containing a Salto workspace and an SFDX project.

It will get the changes to apply, it will call `dumpElementsToFolder` to update the SFDX part. And in the future, it should use the leftover changes to the Salto workspace

Additional context for the reviewer
No release notes

Release Notes:
None

User Notifications:
None

### DIFF
--- a/packages/core/src/core/adapter_format.ts
+++ b/packages/core/src/core/adapter_format.ts
@@ -253,7 +253,7 @@ export type UpdateElementFolderResult = {
   errors: ReadonlyArray<SaltoError>
 }
 
-export const updateElementFolder = async ({
+export const updateElementFolder = ({
   workspace,
   baseDir,
   changes,
--- a/packages/core/src/core/adapter_format.ts
+++ b/packages/core/src/core/adapter_format.ts
@@ -6,7 +6,7 @@
  * CERTAIN THIRD PARTY SOFTWARE MAY BE CONTAINED IN PORTIONS OF THE SOFTWARE. See NOTICE FILE AT https://github.com/salto-io/salto/blob/main/NOTICES
  */
 import _ from 'lodash'
-import { Adapter, AdapterOperationsContext, Element, SaltoError } from '@salto-io/adapter-api'
+import { Adapter, AdapterOperationsContext, Change, Element, SaltoError } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
 import { collections } from '@salto-io/lowerdash'
 import { merger, Workspace, ElementSelector, expressions, elementSource } from '@salto-io/workspace'
@@ -183,25 +183,23 @@ export const syncWorkspaceToFolder = ({
       })
       const { loadElementsFromFolder, dumpElementsToFolder } = adapter
       if (loadElementsFromFolder === undefined) {
-        const message = `Account ${accountName}'s adapter does not support loading a non-nacl format`
         return {
           errors: [
             {
               severity: 'Error' as const,
-              message,
-              detailedMessage: message,
+              message: 'Format not supported',
+              detailedMessage: `Account ${accountName}'s adapter does not support loading a non-nacl format`,
             },
           ],
         }
       }
       if (dumpElementsToFolder === undefined) {
-        const message = `Account ${accountName}'s adapter does not support writing a non-nacl format`
         return {
           errors: [
             {
               severity: 'Error' as const,
-              message,
-              detailedMessage: message,
+              message: 'Format not supported',
+              detailedMessage: `Account ${accountName}'s adapter does not support writing a non-nacl format`,
             },
           ],
         }
@@ -241,5 +239,46 @@ export const syncWorkspaceToFolder = ({
       return dumpElementsToFolder({ baseDir, changes, elementsSource: adapterContext.elementsSource })
     },
     'syncWorkspaceToFolder %s',
+    baseDir,
+  )
+
+type UpdateElementFolderArgs = {
+  workspace: Workspace
+  baseDir: string
+  accountName: string
+  changes: ReadonlyArray<Change>
+}
+
+export type UpdateElementFolderResult = {
+  errors: ReadonlyArray<SaltoError>
+}
+
+export const updateElementFolder = async ({
+  workspace,
+  baseDir,
+  changes,
+  accountName,
+}: UpdateElementFolderArgs): Promise<UpdateElementFolderResult> =>
+  log.time(
+    async () => {
+      const { adapter, adapterContext } = await getAdapterAndContext({
+        workspace,
+        accountName,
+      })
+      const { dumpElementsToFolder } = adapter
+      if (dumpElementsToFolder === undefined) {
+        return {
+          errors: [
+            {
+              severity: 'Error' as const,
+              message: 'Format not supported',
+              detailedMessage: `Account ${accountName}'s adapter does not support writing a non-nacl format`,
+            },
+          ],
+        }
+      }
+      return dumpElementsToFolder({ baseDir, changes, elementsSource: adapterContext.elementsSource })
+    },
+    'updateElementFolder %s',
     baseDir,
   )
--- a/packages/core/test/core/adapter_format.test.ts
+++ b/packages/core/test/core/adapter_format.test.ts
@@ -14,13 +14,20 @@ import {
   Adapter,
   toChange,
   SaltoError,
+  Change,
 } from '@salto-io/adapter-api'
 import { Workspace } from '@salto-io/workspace'
 import { collections } from '@salto-io/lowerdash'
 import { adapterCreators } from '../../src/core/adapters'
 import { mockWorkspace } from '../common/workspace'
 import { createMockAdapter } from '../common/helpers'
-import { calculatePatch, syncWorkspaceToFolder, SyncWorkspaceToFolderResult } from '../../src/core/adapter_format'
+import {
+  calculatePatch,
+  syncWorkspaceToFolder,
+  SyncWorkspaceToFolderResult,
+  updateElementFolder,
+  UpdateElementFolderResult,
+} from '../../src/core/adapter_format'
 
 const { awu } = collections.asynciterable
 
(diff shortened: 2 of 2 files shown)


packages/core/src/core/adapter_format.ts
```diff
@@ -243,3 +243,57 @@ export const syncWorkspaceToFolder = ({
...
+          errors: [
+            {
+              severity: 'Error' as const,
+              message,
+              detailedMessage: message,
+            },
+          ],
+        }
+      }
+      return dumpElementsToFolder({ baseDir, changes, elementsSource: adapterContext.elementsSource })
```
dorsev, line 295: should I implement now the ` and then apply the leftover changes to the salto workspace` ? or should this be done in a separate PR (together with sync method?)

packages/core/src/core/adapter_format.ts
```diff
@@ -243,3 +243,57 @@ export const syncWorkspaceToFolder = ({
...
+        return {
+          errors: [
+            {
+              severity: 'Error' as const,
+              message,
+              detailedMessage: message,
+            },
+          ],
+        }
+      }
```
dorsev, line 282: DRY

packages/core/src/core/adapter_format.ts
```diff
@@ -243,3 +243,57 @@ export const syncWorkspaceToFolder = ({
...
+export type UpdateElementFolderResult = {
+  errors: ReadonlyArray<SaltoError>
+}
+
+export const updateElementFolder = async ({
+  workspace,
+  baseDir,
+  changes,
+  accountName,
+}: UpdateElementFolderArgs): Promise<UpdateElementFolderResult> =>
```
dorsev, line 263: this currently does not return the `unAppliedChanges`. Is that okay? should I change this API currently?

packages/core/src/core/adapter_format.ts
dorsev, line 282: I tried to "DRY" this but I could not think of a good enough API here.

packages/core/src/core/adapter_format.ts
```diff
@@ -243,3 +243,57 @@ export const syncWorkspaceToFolder = ({
...
+        return {
+          errors: [
+            {
+              severity: 'Error' as const,
+              message,
+              detailedMessage: message,
+            },
+          ],
+        }
+      }
```
ori-moisis, line 294: 1. why is this an error?
2. if there is really a need to check for this use case, maybe do it before the whole initialization of the adapter?

packages/core/src/core/adapter_format.ts
```diff
@@ -243,3 +243,57 @@ export const syncWorkspaceToFolder = ({
...
+      })
+      const { dumpElementsToFolder } = adapter
+      if (dumpElementsToFolder === undefined) {
+        const message = `Account ${accountName}'s adapter does not support writing a non-nacl format`
+        return {
+          errors: [
+            {
+              severity: 'Error' as const,
+              message,
+              detailedMessage: message,
```
ori-moisis, line 278: it doesn't make sense that the message and the detailed message are the same here (cc: @shayc331 
@dantal4  @kami-delaroz  - not sure if you did this in more places, this is not great, and if people continue to extend code as if these are always supposed to be the same, it will be harder to fix in the future...)

in my branch I changed this (in the other function) to
```typescript
const { loadElementsFromFolder, dumpElementsToFolder } = adapter
      if (loadElementsFromFolder === undefined) {
        return {
          errors: [
            {
              severity: 'Error' as const,
              message: 'Format not supported',
              detailedMessage: `Account ${accountName}'s adapter does not support loading a non-nacl format`,
            },
          ],
        }
      }
      if (dumpElementsToFolder === undefined) {
        return {
          errors: [
            {
              severity: 'Error' as const,
              message: 'Format not supported',
              detailedMessage: `Account ${accountName}'s adapter does not support writing a non-nacl format`,
            },
          ],
        }
      }
```

my branch is not ready yet, so could you please make this change here?

packages/core/test/core/adapter_format.test.ts
```diff
@@ -353,3 +371,137 @@ describe('syncWorkspaceToFolder', () => {
...
+  const mockAdapterName = 'mock'
+  const type = new ObjectType({
+    elemID: new ElemID(mockAdapterName, 'type'),
+    fields: {
+      f: { refType: BuiltinTypes.STRING },
+    },
+  })
+
+  const instance = new InstanceElement('instance', type, { f: 'v' })
+  const changes = [toChange({ after: instance })]
```
ori-moisis, line 385: we should not create consts outside of tests / `beforeEach`, these create implicitly shared context between tests

packages/core/test/core/adapter_format.test.ts
```diff
@@ -353,3 +371,137 @@ describe('syncWorkspaceToFolder', () => {
...
+  beforeEach(() => {
+    mockAdapter = createMockAdapter(mockAdapterName)
+    adapterCreators[mockAdapterName] = mockAdapter
+  })
+  afterEach(() => {
+    delete adapterCreators[mockAdapterName]
+  })
+  describe('when called with valid parameters', () => {
+    let workspace: Workspace
+    let accountName: string
```
ori-moisis, line 397: why is this needed in this scope?

packages/core/test/core/adapter_format.test.ts
```diff
@@ -353,3 +371,137 @@ describe('syncWorkspaceToFolder', () => {
...
+        elementsSource: expect.anything(),
+      })
+    })
+  })
+
+  describe('when updating elements to folder returns an error', () => {
+    let result: SyncWorkspaceToFolderResult
+    let errors: SaltoError[]
+    let workspace: Workspace
+    const accountName = mockAdapterName
```
ori-moisis, line 420: inconsistent style (in the previous block this was an unneeded `let`, now it seems like an unneeded `const`)
(unneeded because it is only used in the `beforeEach`, no point in having it defined in a larger context)

packages/core/test/core/adapter_format.test.ts
```diff
@@ -353,3 +371,137 @@ describe('syncWorkspaceToFolder', () => {
...
+          severity: 'Error',
+          message: 'something failed',
+          detailedMessage: 'something failed',
+        },
+      ]
+      workspace = mockWorkspace({
+        name: 'workspace',
+        elements: [],
+        accountToServiceName: { [accountName]: mockAdapterName },
+      })
```
ori-moisis, line 433: seems like this initialization of an empty workspace is repeated in multiple blocks
maybe do it once in an outer `beforeEach` block instead?

packages/core/src/core/adapter_format.ts
shayc331: @ori-moisis it is an intermediate stage, currently the saas reads the `message` and I am changing it to read the `detailedMessage`. In the end, the UI will group errors by the `message` and it will be different than the `detailedMessage`

packages/core/src/core/adapter_format.ts
dorsev: I wasn't sure of this. good point. removed

packages/core/test/core/adapter_format.test.ts
dorsev: good point. changed

packages/core/test/core/adapter_format.test.ts
dorsev: good point. changed

packages/core/test/core/adapter_format.test.ts
dorsev: it's not. good point. changed

packages/core/test/core/adapter_format.test.ts
dorsev: 👍 ack. changed. 
(moved to beforeEach)

dorsev:
Thank you for the thorough review @ori-moisis 
I think I'm ready for round 2.

packages/core/src/core/adapter_format.ts
dorsev: opened - SAAS-12078

packages/core/src/core/adapter_format.ts
dorsev: yes 👍

packages/core/src/core/adapter_format.ts
dorsev: good point @ori-moisis , I agree. added you suggested code there.

packages/core/src/core/adapter_format.ts
```diff
@@ -243,3 +241,44 @@ export const syncWorkspaceToFolder = ({
...
+  baseDir: string
+  accountName: string
+  changes: ReadonlyArray<Change>
+}
+
+export type UpdateElementFolderResult = {
+  errors: ReadonlyArray<SaltoError>
+}
+
+export const updateElementFolder = async ({
```
ori-moisis, line 256: I missed this on the first round, but this doesn't need to be `async`
```suggestion
export const updateElementFolder = ({
```

packages/core/test/core/adapter_format.test.ts
```diff
@@ -353,3 +372,94 @@ describe('syncWorkspaceToFolder', () => {
...
+      },
+    })
+
+    const instance = new InstanceElement('instance', type, { f: 'v' })
+    changes = [toChange({ after: instance })]
+    workspace = mockWorkspace({
+      name: 'workspace',
+      elements: [],
+      accountToServiceName: { [mockAdapterName]: mockAdapterName },
+    })
```
ori-moisis, line 397: note that you create an addition change of an instance, but the type of that instance is not defined in the workspace (i.e - it is not added, and does not exist in the workspace)
this is technically a possible scenario, but not a very valid one.
maybe it doesn't currently affect the results of this specific test, but I don't think you meant to test this...

moreover, in this test you are supposed to update a folder with changes from a workspace, but in this case the workspace does not contain the elements that were supposedly added to it.
this is actually not valid input here, right?

ori-moisis (approved):
a few nitpicks, other than that, code looks good

packages/core/src/core/adapter_format.ts
dorsev: that's a good catch

packages/core/test/core/adapter_format.test.ts
dorsev: good input. I've changed this. LMK if this makes more sense now? 🤔

ori-moisis: 👍

PR #6537, opened 2024-09-09
SALTO-6623: add perform side effect removals config option

ori-moisis:
This is needed for SFDX dump code to actually delete side effects from the project

---

---
_Release Notes_: 
_Salesforce Adapter_:
- Added optional feature `performSideEffectDeletes` that allows explicitly deleting child instances of custom object when the custom object is removed (default: false)

---
_User Notifications_: 
_None_


tamtamirr (approved):
LGTM. Are you aware to why we implemented this behavior in first place? I couldn't find the relevant ticket. I wonder if we really want two different implementations here and not make your code the default behavior. (By deleting the old code)

PR #6538, opened 2024-09-09
SALTO-6601: [Intune] Support fetch & deploy for Linux Platform Scripts

shoshana-houri:
Support fetching & deploying Intune Linux Platform Scripts - including assignments & settings. Extract the script value to a static file.

---
This is the first of three types of Intune Platform Scripts. Each type has different endpoints and structures, so I'll split the PRs accordingly.

The type of Linux Platform Scripts is actually a specific variation of the device configuration policy (with a settings catalog). In the UI, the Linux scripts do not appear in the device configuration section, so in this PR, we mirrored this behavior by splitting them into two different types.

See .nacls structure example [here](https://github.com/salto-io/microsoft-security/blob/main/microsoft_security/Records/IntunePlatformScriptLinux/svh_linux_script.nacl).
See the matching static resource example [here](https://github.com/salto-io/microsoft-security/blob/main/static-resources/microsoft_security/IntunePlatformScriptLinux/svh%20linux%20script/linux_customconfig_script.sh). 

---
_Release Notes_: 
_Microsoft Security adapter:_
* Support fetch & deploy for Intune Linux Platform Scripts.

---
_User Notifications_: 
_Microsoft Security adapter:_
* Support fetch & deploy for Intune Linux Platform Scripts.


packages/serviceplaceholder-adapter/test/adapter.test.ts
```diff
@@ -28,6 +28,7 @@ import { DEFAULT_CONFIG } from '../src/config'
 import { ADAPTER_NAME } from '../src/constants'
 // TODO update mock file -
 // for fetch: run fetch with trace-level logs:
+// [Please make sure to first set the strategy to 'full' in the DEFAULT_LOGGING_CONFIG under logging.utils.ts]
```
shir-reifenberg, line 31: Thanks!
Can you please just put the needed configuration here ?

packages/microsoft-security-adapter/src/definitions/deploy/intune/utils/device_configuration_settings.ts
```diff
@@ -0,0 +1,69 @@
...
+ * CERTAIN THIRD PARTY SOFTWARE MAY BE CONTAINED IN PORTIONS OF THE SOFTWARE. See NOTICE FILE AT https://github.com/salto-io/salto/blob/main/NOTICES
+ */
+
+import { groupAssignments } from '.'
+import { intuneConstants } from '../../../../constants'
+import { InstanceDeployApiDefinitions } from '../../shared/types'
+
+const { ASSIGNMENTS_FIELD_NAME } = intuneConstants
+
+// In the UI, some of the device configurations (setting catalog) are located in the Device Configuration settings tab
```
shir-reifenberg, line 15: "the UI" sounds a bit odd to me. maybe something like "in intune admin console" (that's the name for the Okta admin console, please adjust with the MS phrasing)

packages/microsoft-security-adapter/src/definitions/fetch/intune/fetch.ts
```diff
@@ -332,6 +323,55 @@ const graphBetaCustomizations: FetchCustomizations = {
       fieldCustomizations: ID_FIELD_TO_HIDE,
     },
   },
+  [PLATFORM_SCRIPT_LINUX_TYPE_NAME]: {
```
Nurdok, line 326: This seems _almost_ identical to `DEVICE_CONFIGURATION_SETTING_CATALOG_TYPE_NAME`.

I assume that if we find we need to change one of these, we'll probably change both? If so, maybe extract these to a function that accepts the differences between them as params and returns the configuration?

Something like `getDeviceConfigurationSettingsFetchDefinitions(typename, query)`?

packages/microsoft-security-adapter/test/adapter.test.ts
```diff
@@ -491,6 +493,51 @@ describe('Microsoft Security adapter', () => {
...
+
+              const platformScriptLinuxNames = platformScriptsLinux.map(e => e.elemID.name)
+              expect(platformScriptLinuxNames).toEqual(['test_linux_script@s'])
+            })
+
+            it('should include settings field with the correct values', async () => {
+              const platformScriptLinux = platformScriptsLinux[0]
+              expect(platformScriptLinux.value.settings).toHaveLength(4)
+              expect(Object.keys(platformScriptLinux.value.settings[0])).toEqual(['settingInstance'])
+              const linuxScript = _.get(platformScriptLinux.value.settings[3], [
```
Nurdok, line 516: Any way to avoid accessing the array with hardcoded indexes?

packages/microsoft-security-adapter/test/definitions/fetch/intune/utils/device_configuration_settings.test.ts
```diff
@@ -0,0 +1,129 @@
...
+          }),
+        ).toEqual({ value })
+      })
+    })
+
+    describe('when the instance has settings', () => {
+      it('should set the script value as a static file', async () => {
+        const value = PLATFORM_SCRIPT_VALUE_WITH_SETTINGS
+        const expectedSettings = [
+          ...PLATFORM_SCRIPT_VALUE_WITH_SETTINGS.settings.slice(0, 3),
```
Nurdok, line 104: This is difficult to read - I don't really understand what makes the fourth setting unique (it has a script?)
I suggest you give each setting a descriptive name, that will make it both easier to read and have less index manipulation.

```
const PLATFORM_SCRIPT_VALUE_WITH_SETTINGS = {
      platforms: 'linux',
      settings: [ SETTING_WITHOUT_A_SCRIPT_0, DESCRIPTIVE_SETTING_1, DESCRIPTIVE_SETTING_2, SETTING_WITH_A_SCRIPT ],
}
```
or perhaps `[ ...SETTINGS_WITHOUT_A_SCRIPT, SETTING_WITH_A_SCRIPT]`

Then the test would also make more sense:

```
const expectedSettings = {
    ...SETTINGS_WITHOUT_A_SCRIPT,
   { 
      settingsInstance: {...SETTING_WITH_A_SCRIPT.settingInstace,
      // etc.

```

shoshana-houri: That's a good idea, thanks :)

packages/microsoft-security-adapter/src/definitions/fetch/intune/fetch.ts
shoshana-houri: That's a great suggestion, thanks :)

packages/serviceplaceholder-adapter/test/adapter.test.ts
shir-reifenberg: no, I meant copy-paste it here in the comment so it would be easy to copy-paste from the comment to my local config file

shir-reifenberg: something like (not sure I wrote here the actual accurate config) - 
```suggestion
// add this to you adapter.nacl file to capture full logs of the responses:
  client = {
    logging = {
      responseStrategies = [
        {
          strategy = "full"
        },
      ]
    }
  }
 //
```

packages/microsoft-security-adapter/src/definitions/deploy/intune/utils/device_configuration_settings.ts
Nurdok: +1

Nurdok (approved):

PR #6540, opened 2024-09-09
SAAS-12079: Extend plan API to include "All Changes"

dorsev:

---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/core/src/core/plan/plan.ts
+++ b/packages/core/src/core/plan/plan.ts
@@ -39,7 +39,7 @@ import { DataNodeMap, DiffNode, DiffGraph, GroupDAG } from '@salto-io/dag'
 import { logger } from '@salto-io/logging'
 import { expressions } from '@salto-io/workspace'
 import { collections, values } from '@salto-io/lowerdash'
-import { PlanItem, addPlanItemAccessors, PlanItemId } from './plan_item'
+import { PlanItem, addPlanItemAccessors, PlanItemId, ChangeWithDetails, toChangeWithDetails } from './plan_item'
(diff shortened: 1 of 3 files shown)


packages/core/src/core/plan/plan.ts
```diff
@@ -391,13 +392,20 @@ export type Plan = GroupDAG<Change> & {
   getItem: (id: PlanItemId) => PlanItem
   changeErrors: ReadonlyArray<ChangeError>
 }
+export type PlanWithAllChanges = Plan & { all: Iterable<ChangeWithDetails> }
```
dorsev, line 395: I don't think this is a good enough name. I would appreciate your input or suggestions here.
Im thinking maybe `allChanges` or `diffChanges`?

packages/core/src/core/plan/plan.ts
```diff
@@ -474,7 +482,7 @@ export const getPlan = async ({
       // build graph
       const groupedGraph = buildGroupedGraphFromDiffGraph(filterResult.validDiffGraph, changeGroupIdMap, disjointGroups)
       // build plan
-      return addPlanFunctions(groupedGraph, filterResult.changeErrors, compareOptions)
+      return addPlanFunctions(groupedGraph, filterResult.changeErrors, diffGraph, compareOptions)
```
dorsev, line 485: Question: where should I add the unit tests? do they belong to `plan.test.ts` or `filter.test.ts`?

dorsev:
Will reopen this under a different ticket.

PR #6541, opened 2024-09-09
SALTO-6638: Extend Plan interface to allow "consuming" all changes, instead of "plan changes"

dorsev:

---

---
_Release Notes_: 

---
_User Notifications_: 


packages/core/src/core/plan/plan.ts
```diff
@@ -448,36 +123,14 @@ export const getPlan = async ({
   topLevelFilters = [],
   compareOptions,
 }: GetPlanParameters): Promise<Plan> => {
-  const numBeforeElements = await awu(await before.list()).length()
-  const numAfterElements = await awu(await after.list()).length()
-  return log.timeDebug(
-    async () => {
-      const diffGraph = await buildDiffGraph(
-        addDifferentElements(before, after, topLevelFilters, numBeforeElements + numAfterElements, compareOptions),
-        resolveNodeElements(before, after),
```
dorsev, line 457: Q: why does `resolveNodeElements` run in getDiffChanges vs `buildPlan`.

This is probably a nit.

dorsev:
Commenting for future readers:
We separated `getDiffChanges` from `plan` API, and tried to re-use it in `createDiffChanges`.
This fails since createDiffChanges assumes some grouping logic that `buildPlan` does.
This led us to believe this is the wrong tree to chase, and we will go about this in a different way.

PR #6542, opened 2024-09-10
SALTO-6026: Update mappedReferenceRegex to not be greedy

Shulik95:
_The existing version of the regex was greedy and caused parsing to misbehave, The regex now matches the key in key: value and verifies there something after the colon._

---

_None_

---
_Release Notes_: 
_Netsuite Adapter_:
* Failing to parse suitescripts for references logs error instead of failing fetch..

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/netsuite-adapter/src/filters/element_references.ts
+++ b/packages/netsuite-adapter/src/filters/element_references.ts
@@ -60,7 +60,7 @@ const pathPrefixRegex = new RegExp(
   'm',
 )
 // matches key strings in format of 'key': value or key: value
-const mappedReferenceRegex = new RegExp(`['"]?(?<${OPTIONAL_REFS}>\\w+)['"]?\\s*:\\s*.+`, 'gm')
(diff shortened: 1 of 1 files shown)


packages/netsuite-adapter/src/filters/element_references.ts
```diff
@@ -224,31 +224,35 @@ const getSuiteScriptReferences = async (
...
-      skippedFileExtensions.add(osPath.extname(element.value[PATH]))
+      getAndLogReferencesDiff({
+        newReferences: foundReferences,
+        existingReferences: semanticReferences,
+        element,
+        serviceIdToElemID,
+        customRecordFieldsToServiceIds,
+      })
+    } catch (e) {
+      log.error('Failed extracting references from file %s with error %s', element.value[PATH], e)
```
dantal4, line 254:
```suggestion
      log.error('Failed extracting references from file %s with error: %o', element.value[PATH], e)
```

dantal4 (approved):
Great work!

PR #6543, opened 2024-09-11
SALTO-6642 serialize ConflictingMetaTypeError

dantal4:
We should serialize the `ConflictingMetaTypeError` merge error, to be able loading it back as a merge error (currently we fail with `Deserialization failed. At least one element did not deserialize to a MergeError`)

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


ori-moisis (approved):
do we know if there are more bugs like this?
is there a way to enforce that this doesn't happen again?

PR #6544, opened 2024-09-11
SALTO-6643 add calculatePendingChanges param to calcFetchChanges

dantal4:
We can have conflicts even when there are no before elements. We currently don’t calculate the pending changes in this case (because we treat it as "empty state"), which makes us accept all the service changes even if there are already nacls with elements that conflict that.

In fetch we shouldn't do that, because we can fetch an environment for the first time (so there's no state) and already have elements in common - we don't want pending changes in that case.

In `calculatePatch` it can happen, if there are no elements in the `fromDir`, which we use as the state when calculating the pending changes.

In `calculatePatchFromChanges` it can happen, when we have only addition changes, because then there won’t be any `before` elements to use as the state when calculating the pending changes - this is a bug.

---

---
_Release Notes_: 
Core:
- Calculate pending changes when patch has only addition changes

---
_User Notifications_: 
None

### DIFF
--- a/packages/core/src/core/fetch.ts
+++ b/packages/core/src/core/fetch.ts
@@ -793,7 +793,7 @@ export const calcFetchChanges = async ({
   workspaceElements,
   partiallyFetchedAccounts,
   allFetchedAccounts,
-  calculatePendingChanges,
+  calculatePendingChanges = true,
 }: {
   accountElements: ReadonlyArray<Element>
   mergedAccountElements: ReadonlyArray<Element>
@@ -801,7 +801,7 @@ export const calcFetchChanges = async ({
   workspaceElements: ReadOnlyElementsSource
   partiallyFetchedAccounts: Map<string, PartiallyFetchedAccountData>
   allFetchedAccounts: Set<string>
-  calculatePendingChanges: boolean
+  calculatePendingChanges?: boolean
 }): Promise<CalcFetchChangesResult> => {
   const mergedAccountElementsSource = elementSource.createInMemoryElementSource(mergedAccountElements)
 
@@ -963,7 +963,10 @@ const createFetchChanges = async ({
     .filter(e => !e.isConfigType())
     .isEmpty()
 
-  const isStateEmpty = await stateElements.isEmpty()
+  // When we init a new env, the state will be empty, but the workspace can already have elements in common.
+  // In that case we shouldn't calculate pending changes, otherwise there would be conflicts
+  // between the fetched elements and the elements in common.
+  const calculatePendingChanges = !(await stateElements.isEmpty())
 
   const { changes, serviceToStateChanges } = isFirstFetch
     ? await createFirstFetchChanges(unmergedElements, processErrorsResult.keptElements)
@@ -974,7 +977,7 @@ const createFetchChanges = async ({
         workspaceElements,
         partiallyFetchedAccounts: partiallyFetchedAccountData,
         allFetchedAccounts: new Set(adapterNames),
-        calculatePendingChanges: !isStateEmpty,
+        calculatePendingChanges,
       })
   log.debug('finished to calculate fetch changes')
   if (progressEmitter) {
--- a/packages/core/src/core/fetch.ts
+++ b/packages/core/src/core/fetch.ts
@@ -786,14 +786,23 @@ type DetailedChangeTreesResults = {
 
 // Calculate the fetch changes - calculation should be done only if workspace has data,
 // o/w all account elements should be consider as "add" changes.
-export const calcFetchChanges = async (
-  accountElements: ReadonlyArray<Element>,
-  mergedAccountElements: ReadonlyArray<Element>,
-  stateElements: elementSource.ElementsSource,
-  workspaceElements: ReadOnlyElementsSource,
-  partiallyFetchedAccounts: Map<string, PartiallyFetchedAccountData>,
-  allFetchedAccounts: Set<string>,
-): Promise<CalcFetchChangesResult> => {
+export const calcFetchChanges = async ({
+  accountElements,
+  mergedAccountElements,
+  stateElements,
+  workspaceElements,
+  partiallyFetchedAccounts,
+  allFetchedAccounts,
+  calculatePendingChanges,
+}: {
+  accountElements: ReadonlyArray<Element>
+  mergedAccountElements: ReadonlyArray<Element>
+  stateElements: elementSource.ElementsSource
+  workspaceElements: ReadOnlyElementsSource
+  partiallyFetchedAccounts: Map<string, PartiallyFetchedAccountData>
+  allFetchedAccounts: Set<string>
+  calculatePendingChanges: boolean
+}): Promise<CalcFetchChangesResult> => {
   const mergedAccountElementsSource = elementSource.createInMemoryElementSource(mergedAccountElements)
 
   const partialFetchFilter: IDFilter = id =>
@@ -892,13 +901,9 @@ export const calcFetchChanges = async (
     }
   }
 
-  // When we init a new env, state will be empty. We fallback to the workspace
-  // elements since they should be considered a part of the env and the diff
-  // should be calculated with them in mind.
-  const isStateEmpty = await stateElements.isEmpty()
-  const { serviceChanges, pendingChanges, workspaceToServiceChanges, serviceToStateChanges } = isStateEmpty
-    ? await calculateChangesWithEmptyState()
-    : await calculateChangesWithState()
+  const { serviceChanges, pendingChanges, workspaceToServiceChanges, serviceToStateChanges } = calculatePendingChanges
+    ? await calculateChangesWithState()
+    : await calculateChangesWithEmptyState()
 
   // Merge pending changes and service changes into one tree so we can find conflicts between them
   serviceChanges.merge(pendingChanges)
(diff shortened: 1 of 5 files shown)


packages/core/src/core/fetch.ts
```diff
@@ -892,13 +892,21 @@ export const calcFetchChanges = async (
...
-  // When we init a new env, state will be empty. We fallback to the workspace
-  // elements since they should be considered a part of the env and the diff
-  // should be calculated with them in mind.
-  const isStateEmpty = await stateElements.isEmpty()
-  const { serviceChanges, pendingChanges, workspaceToServiceChanges, serviceToStateChanges } = isStateEmpty
-    ? await calculateChangesWithEmptyState()
-    : await calculateChangesWithState()
+  const calculateChanges = async (): Promise<DetailedChangeTreesResults> => {
+    if (await stateElements.isEmpty()) {
+      if (getCoreFlagBool(CORE_FLAGS.skipCalculatingAllChangesWithEmptyState)) {
```
ori-moisis, line 897: having this flag off by default means that by default this doesn't really check if this is breaking anything (i.e - all the unit tests run as if this isn't here).
did you run the unit tests with this enabled to see if anything breaks?

dantal4: The flag is off by default so default behavior has changed - we'll calculate all changes even if the state is empty.

ori-moisis: ok, good to see we have tests that failed then

packages/core/test/core/fetch.test.ts
```diff
@@ -284,10 +284,11 @@ describe('fetch', () => {
           mockAdapters[newTypeDifferentAdapterID.adapter].fetch.mockResolvedValueOnce({
             elements: [newTypeBaseModified],
           })
+          const elementsSource = createInMemoryElementSource([newTypeBaseModifiedDifferentId, typeWithFieldDifferentID])
           const fetchChangesResult = await fetchChanges(
             mockAdapters,
-            createInMemoryElementSource([newTypeBaseModifiedDifferentId, typeWithFieldDifferentID]),
-            createInMemoryElementSource([]),
```
dantal4, line 290: this didn't make sense - when there are no state elements there are no deletions.
@ori-moisis FYI

packages/core/test/core/fetch.test.ts
```diff
@@ -389,13 +390,14 @@ describe('fetch', () => {
...
+            new ObjectType({ elemID: new ElemID('dummy1', 'type') }),
+            new ObjectType({ elemID: new ElemID('dummy2', 'type') }),
+          ])
           const fetchChangesResult = await fetchChanges(
             adapters,
-            createInMemoryElementSource([
-              new ObjectType({ elemID: new ElemID('dummy1', 'type') }),
-              new ObjectType({ elemID: new ElemID('dummy2', 'type') }),
-            ]),
-            createInMemoryElementSource([]),
```
dantal4, line 398: this didn't make sense - when there are no state elements there are no deletions.
@ori-moisis FYI

packages/core/src/core/fetch.ts
```diff
@@ -892,13 +901,21 @@ export const calcFetchChanges = async (
...
-  const { serviceChanges, pendingChanges, workspaceToServiceChanges, serviceToStateChanges } = isStateEmpty
-    ? await calculateChangesWithEmptyState()
-    : await calculateChangesWithState()
+  const calculateChanges = async (): Promise<DetailedChangeTreesResults> => {
+    if (await stateElements.isEmpty()) {
+      if (!calculateAllChangesWithEmptyState) {
+        // When we init a new env, state will be empty. We fallback to the workspace
+        // elements since they should be considered a part of the env and the diff
+        // should be calculated with them in mind.
+        log.info('skipping calculating all changes because state is empty')
```
ori-moisis, line 910: this doesn't feel like an "info" level message as it talks about a very specific decision point in the code, I think "info" level messages should be about entering / leaving significant blocks of code...

this is at most a debug log

packages/core/src/core/fetch.ts
```diff
@@ -892,13 +901,21 @@ export const calcFetchChanges = async (
...
-  // When we init a new env, state will be empty. We fallback to the workspace
-  // elements since they should be considered a part of the env and the diff
-  // should be calculated with them in mind.
-  const isStateEmpty = await stateElements.isEmpty()
-  const { serviceChanges, pendingChanges, workspaceToServiceChanges, serviceToStateChanges } = isStateEmpty
-    ? await calculateChangesWithEmptyState()
-    : await calculateChangesWithState()
+  const calculateChanges = async (): Promise<DetailedChangeTreesResults> => {
+    if (await stateElements.isEmpty()) {
+      if (!calculateAllChangesWithEmptyState) {
```
ori-moisis, line 906: I don't understand this name...
I am not sure what a good name would be, this name seems like it relies too much on the implementation details that outside callers don't need to know.

what is the difference in the behavior of the function when this is `true` vs. `false`?

packages/core/src/core/fetch.ts
```diff
@@ -786,14 +786,23 @@ type DetailedChangeTreesResults = {
 
 // Calculate the fetch changes - calculation should be done only if workspace has data,
```
ori-moisis, line 787: is the comment still correct with the new behavior?

packages/core/src/core/fetch.ts
```diff
@@ -892,13 +901,21 @@ export const calcFetchChanges = async (
...
   }
 
-  // When we init a new env, state will be empty. We fallback to the workspace
-  // elements since they should be considered a part of the env and the diff
-  // should be calculated with them in mind.
-  const isStateEmpty = await stateElements.isEmpty()
-  const { serviceChanges, pendingChanges, workspaceToServiceChanges, serviceToStateChanges } = isStateEmpty
-    ? await calculateChangesWithEmptyState()
-    : await calculateChangesWithState()
+  const calculateChanges = async (): Promise<DetailedChangeTreesResults> => {
```
ori-moisis, line 904: I think this extra function makes the code less readable
is there a reason this can't condition can't go into the original boolean that decides which function to use?

packages/core/test/core/fetch.test.ts
ori-moisis: The test was actually valid since when the state is empty, the workspace elements are used.
when you take away this test, you are removing a use case that used to be covered (the same one that caused the e2e to fail).

I agree the name of the test was not good and it did not explicitly say what the scenario under test is (good example of why the name of the test is so important), but I don't want to have the e2e be the only place that tests the behavior of fetch when there is a common folder.

if you remove the use case from here, I think we need to add another unit test for that scenario (empty state, non empty workspace)

packages/core/src/core/fetch.ts
dantal4: indeed, otherwise we're not going in this function:
https://github.com/salto-io/salto/blob/180c916705dd9593b7803989f86456338524fe5d/packages/core/src/core/fetch.ts#L961-L963

packages/core/src/core/fetch.ts
dantal4: I did the opposite - added a required `calculatePendingChanges` param, and for fetch passed there `!isStateEmpty` .
WDYT?

dantal4: (maybe make it optional with a default `true` value?)

packages/core/src/core/fetch.ts
```diff
@@ -892,13 +901,9 @@ export const calcFetchChanges = async (
     }
   }
 
-  // When we init a new env, state will be empty. We fallback to the workspace
-  // elements since they should be considered a part of the env and the diff
-  // should be calculated with them in mind.
```
ori-moisis, line 897: I think we should keep the comment that says why we need this behavior (probably move it to the call where the boolean is passed in)
maybe try to make the comment more clear so the next dev won't try to remove the case like we did (be more explicit about the scenario)

packages/core/test/core/fetch.test.ts
```diff
@@ -2322,14 +2322,15 @@ describe('calc fetch changes', () => {
...
-      new Set(['salto']),
-    )
+    const { changes, serviceToStateChanges } = await calcFetchChanges({
+      accountElements: [existingElement],
+      mergedAccountElements: [existingElement],
+      stateElements: createInMemoryElementSource([existingElement, instanceA, instanceB]),
+      workspaceElements: createInMemoryElementSource([existingElement, instanceA, instanceB]),
+      partiallyFetchedAccounts: new Map([['salto', { deletedElements: new Set([instanceA.elemID.getFullName()]) }]]),
+      allFetchedAccounts: new Set(['salto']),
+      calculatePendingChanges: true,
```
ori-moisis, line 2332: there should also be a test somewhere with `false`.
we need to test if for coverage, but more importantly, we need a test in order to explicitly describe the expected behavior and what that flag is supposed to do

packages/core/src/core/fetch.ts
ori-moisis: having a default true is a hint for future users of the function.
if you think we should basically almost always pass in true, then having it be the default makes sense

packages/core/test/core/fetch.test.ts
```diff
@@ -2315,34 +2315,100 @@ describe('fetch from workspace', () => {
...
+          serviceChanges: [
+            expect.objectContaining({
+              id: elemID.createNestedID('attr', 'test'),
+              action: 'modify',
+              data: { before: false, after: true },
+            }),
+          ],
+          pendingChanges: [],
+          metadata: {},
+        },
```
ori-moisis, line 2410: these two test say conflicting things.
reading the description it says that when `calculating pending changes` we should both `calculate pending changes` and `not calculate pending changes`
how can both of these pass at the same time?

if these is not in the same scenario, then there should be something that says what is the scenario when we expect to have pending changes and what is the scenario where we shouldn't get them

ori-moisis (approved):
I think the code is ok, the tests are confusing IMO and should be improved

PR #6545, opened 2024-09-11
SALTO-6350 Add elemId hash to create uniqueness of static file paths

adfineberg:
Currently in the relevant places in the Zendesk adapter

---

---
_Release Notes_: 

_Zendesk Adapter:_
* Some static file paths will change in order to ensure uniqueness. Also the pointers to these static files will be updated to the correct path.

---
_User Notifications_: 

_Zendesk Adapter:_
* Some static file paths will change in order to ensure uniqueness. Also the pointers to these static files will be updated to the correct path.

### DIFF
--- a/packages/zendesk-adapter/src/filters/macro_attachments.ts
+++ b/packages/zendesk-adapter/src/filters/macro_attachments.ts
@@ -42,7 +42,7 @@ import { addId, deployChange, deployChanges } from '../deployment'
 import { getZendeskError } from '../errors'
 import { lookupFunc } from './field_references'
 import ZendeskClient from '../client/client'
-import { createAdditionalParentChanges } from './utils'
+import { createAdditionalParentChanges, shortElemIdHash } from './utils'
 
 const log = logger(module)
 const { awu } = collections.asynciterable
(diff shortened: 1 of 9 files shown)


packages/zendesk-adapter/src/filters/macro_attachments.ts
```diff
@@ -140,21 +140,22 @@ const createAttachmentInstance = ({
...
       filename: attachment.filename,
       contentType: attachment.content_type,
-      content: content
-        ? new StaticFile({ filepath: `${ZENDESK}/${attachmentType.elemID.name}/${resourcePathName}`, content })
-        : undefined,
     },
     [ZENDESK, RECORDS_PATH, MACRO_ATTACHMENT_TYPE_NAME, pathName],
     { [CORE_ANNOTATIONS.PARENT]: [new ReferenceExpression(macro.elemID, macro)] },
   )
+  const resourcePathName = normalizeFilePathPart(`${shortElemIdHash(attachmentInstance.elemID)}_${name}`)
```
edenhassid, line 154: we should check how this affects the deploy of a macro attachment, if we upload it with the new name or not

edenhassid (approved):
looks good! 🥳 added one small thing we should check, pre-approving :)

packages/zendesk-adapter/src/filters/macro_attachments.ts
adfineberg: The deploy is fine :)

PR #6549, opened 2024-09-11
SALTO-6644: Exclude non retrievable Profile related instances

tamtamirr:
Exclude non retrievable Profile related instances

---

Additional context an reasoning can be found on the linked ticket.

---
_Release Notes_: 
_Salesforce Adapter_:
- Suggest to exclude non-retrievable Profile related instances to avoid broken references and improve Change-Based fetch times.

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/fetch.ts
```diff
@@ -466,6 +466,25 @@ export const retrieveMetadataInstances = async ({
       typesWithContent,
       fetchProfile.isFeatureEnabled('fixRetrieveFilePaths'),
     )
+    // Exclude Profile related instances we fail to retrieve for envs that manage Profiles to improve performance
+    // in subsequent fetches and avoid broken references in Profiles.
+    if (
+      metadataQuery.isTypeMatch(PROFILE_METADATA_TYPE) &&
+      fetchProfile.isFeatureEnabled('excludeNonRetrievedProfilesRelatedInstances')
+    ) {
+      const retrievedFileNames = allValues.map(({ file }) => file.fileName)
```
amitgoldarb, line 475: This will be more efficeint if this is a Set

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -466,6 +466,25 @@ export const retrieveMetadataInstances = async ({
...
+      allFileProps
```
amitgoldarb, line 476: Can we also log all of the instances we are excluding for this reason?

amitgoldarb (approved):
Approving to save cycles. Thanks for this

PR #6550, opened 2024-09-11
SALTO-6646: use detailedMessage instead of message in saltoError

shayc331:
_use detailedMessage instead of message in saltoError_

---

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/workspace/src/workspace/workspace.ts
+++ b/packages/workspace/src/workspace/workspace.ts
@@ -1096,6 +1096,8 @@ export const loadWorkspace = async (
 
     return {
       ...saltoElemErr,
+      message: saltoElemErr.message,
+      detailedMessage: saltoElemErr.detailedMessage,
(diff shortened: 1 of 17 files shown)


packages/workspace/src/workspace/workspace.ts
```diff
@@ -1097,6 +1097,7 @@ export const loadWorkspace = async (
     return {
       ...saltoElemErr,
       message: saltoElemErr.message,
+      detailedMessage: saltoElemErr.detailedMessage,
```
dantal4, line 1100: how this didn't break the code in the previous PR?

shayc331:
@dantal4 these changes came after I renamed `message` to `message2`, so I changed every place that used `message` to `detailedMessage`. Am I missing something?

dantal4 (approved):
💯

PR #6551, opened 2024-09-11
SALTO-6438: Support configuring Microsoft Security adapter to selectively manage services

shoshana-houri:
Allow users to select whether they wish to manage Entra, Intune or both.
This selection is done **during authentication** and is stored as part of the user's credentials (not the user config), since it has an impact on the scopes we use to authenticate.

---
_Release Notes_: 
_Microsoft Security adapter:_
* Allow users to selectively choose the services they wish to manage (Entra, Intune)

---
_User Notifications_: 
_Microsoft Security adapter:_
* Allow users to selectively choose the services they wish to manage (Entra, Intune)


packages/microsoft-security-adapter/src/definitions/fetch/entra/fetch.ts
```diff
@@ -835,10 +835,20 @@ const graphBetaCustomizations: FetchCustomizations = {
...
-  definitions.fetch.InstanceFetchApiDefinitions<Options>
-> => ({
-  ...createCustomizationsWithBasePathForFetch(graphV1Customizations, GRAPH_V1_PATH),
-  ...createCustomizationsWithBasePathForFetch(graphBetaCustomizations, GRAPH_BETA_PATH),
-})
+// These types are always included in the fetch, regardless of the chosen services to manage
+const BASIC_ENTRA_TYPES = [GROUP_TYPE_NAME, GROUP_ADDITIONAL_DATA_TYPE_NAME]
+export const basicEntraCustomizations = Object.fromEntries(
+  Object.entries(graphV1Customizations).filter(([typeName, _customization]) => BASIC_ENTRA_TYPES.includes(typeName)),
+)
```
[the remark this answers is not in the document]
shir-reifenberg: also, I think the `export` can be removed

packages/microsoft-security-adapter/src/client/constants.ts
```diff
@@ -0,0 +1,108 @@
...
+    clientSecret: {
+      refType: BuiltinTypes.STRING,
+      annotations: { _required: true, message: 'Client Secret' },
+    },
+    refreshToken: {
+      refType: BuiltinTypes.STRING,
+      annotations: { _required: true, message: 'Refresh Token' },
+    },
+    servicesToManage: {
+      refType: new ListType(BuiltinTypes.STRING),
```
shir-reifenberg, line 104: this is not typed, let's create an `ObjectType` from it / 2 booleans

packages/microsoft-security-adapter/src/client/oauth.ts
```diff
@@ -7,121 +7,40 @@
...
 
+const getServicesToManageFromResponse = (response: Values): AvailableMicrosoftSecurityServices[] =>
+  Object.entries(response)
+    .filter(([, shouldManage]) => shouldManage)
+    .map(([serviceName]) => serviceName as AvailableMicrosoftSecurityServices)
+
 export const createOAuthRequest = (userInput: InstanceElement): OAuthRequestParameters => {
-  const { tenantId, clientId, port } = userInput.value
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const { tenantId, clientId, clientSecret: _clientSecret, port, ...servicesToManageResponse } = userInput.value
```
shir-reifenberg, line 36: leftover?

packages/microsoft-security-adapter/src/client/oauth.ts
```diff
@@ -7,121 +7,40 @@
...
   const baseUrl = getAuthenticationBaseUrl(tenantId)
   const redirectUri = getRedirectUri(port)
-  const scope = ['offline_access', ...OAUTH_REQUIRED_SCOPES].join(' ')
+  const servicesToManage = getServicesToManageFromResponse(servicesToManageResponse)
+  if (servicesToManage.length === 0) {
+    throw new Error('At least one service should be selected to be managed by Salto')
+  }
```
shir-reifenberg, line 42: 🔥

shir-reifenberg (commented):
Looks great 🔥 🚒 
added some comments
also, if the effort is not too big I think we should add to `adapter.test.ts` test cases for the intune, entra, and intune+entra case

packages/microsoft-security-adapter/src/client/oauth.ts
shoshana-houri: Actually, no—I wanted to use the spread notation (...) to capture all the services to manage without specifying them explicitly. I had to unpack `clientSecret` as well in order to do that.

packages/microsoft-security-adapter/src/client/constants.ts
shoshana-houri: Not sure I understand/agree--let's talk f2f

shoshana-houri: Talked f2f - decided to manage as an object

packages/microsoft-security-adapter/src/client/oauth.ts
```diff
@@ -7,121 +7,53 @@
...
-      refType: BuiltinTypes.NUMBER,
-      annotations: {
-        message: 'Port',
-        _required: true,
-      },
-    },
-  },
-})
+const extractServicesToManageFromInputAsArray = (userInput: Values): AvailableMicrosoftSecurityServices[] =>
+  AVAILABLE_MICROSOFT_SECURITY_SERVICES.filter(service => userInput[service])
```
shir-reifenberg, line 26: nit: can we be explicit here?
```suggestion
  AVAILABLE_MICROSOFT_SECURITY_SERVICES.filter(service => userInput[service] === true)
```

packages/microsoft-security-adapter/src/client/oauth.ts
```diff
@@ -7,121 +7,53 @@
...
 
-export type Credentials = Omit<OauthRequestParameters, 'port'> & {
-  refreshToken: string
+export const getOAuthRequiredScopes = (userInput: Values): string => {
+  const servicesToManage = extractServicesToManageFromInputAsArray(userInput)
+  const scopes = servicesToManage.flatMap(service => SCOPE_MAPPING[service])
+  return [...new Set([...BASIC_OAUTH_REQUIRED_SCOPES, ...scopes])].join(' ')
```
shir-reifenberg, line 31: nit: add a `log.trace` with requested scopes

packages/microsoft-security-adapter/src/client/oauth.ts
```diff
@@ -7,121 +7,53 @@
...
-    },
-  },
-})
-
 export const getAuthenticationBaseUrl = (tenantId: string): string =>
   `https://login.microsoftonline.com/${tenantId}/oauth2/v2.0`
 
 const getRedirectUri = (port: number): string => `http://localhost:${port}/extract`
 
+const extractServicesToManageFromInputAsObject = (userInput: Values): MicrosoftServicesToManage =>
```
shir-reifenberg, line 39: readability nit: I'm not sure why we convert it to array if we convert it back to object 🤔 
also, we prefer avoid `reduce` when possible as it's more difficult to read IMO
should _.pick(userInput, AVAILABLE_MICROSOFT_SECURITY_SERVICES) work?

shir-reifenberg (approved):
🎉
added a few more comments, nothing is a blocker

packages/microsoft-security-adapter/src/client/oauth.ts
shoshana-houri: The `pick` is a great suggestion, thanks!
We use the array format because, otherwise, we would need to manually manipulate the object multiple times to: 1) determine if at least one service is selected, and 2) choose the relevant scopes. Transforming it to an array makes these checks and mappings a bit cleaner IMO

packages/microsoft-security-adapter/src/auth.ts
```diff
@@ -0,0 +1,135 @@
...
+      annotations: { _required: true, message: 'Client Secret' },
+    },
+    refreshToken: {
+      refType: BuiltinTypes.STRING,
+      annotations: { _required: true, message: 'Refresh Token' },
+    },
+    servicesToManage: {
+      refType: new ObjectType({
+        elemID: new ElemID(ADAPTER_NAME, 'servicesToManage'),
+        fields: AVAILABLE_MICROSOFT_SECURITY_SERVICES.reduce(
```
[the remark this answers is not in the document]
shoshana-houri: TS infers the type of `fields` as `Record<string, FieldDefinition>`, meaning it's not specifically a mapping of `AvailableMicrosoftSecurityServices`. This means the build won't break if we add a new service and forget to include it here. Therefore, I prefer using this general mapping to ensure all services are covered.

As for `reduce` - I replaced it with `map` and `Object.fromEntries`, hope it's clearer now :)

PR #6552, opened 2024-09-12
SALTO-6652: Improve Salesforce Profiles & PS Elements fixer

tamtamirr:
Improve Salesforce Profiles & PS Elements fixer

---

Improved the implementation to not omit values of Instances that are not passing a dedicated MetadataQuery implementation that operates on `ElemID`.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


packages/salesforce-adapter/test/fetch_profile/metadata_query.test.ts
```diff
@@ -837,4 +838,87 @@ describe('buildMetadataQuery', () => {
...
+    ]
+
+    describe.each(testCases)(
+      'isInstanceMatch & isInstanceIncluded on "$elemFullName"',
+      ({ elemFullName, expectedMetadataInstance }) => {
+        let elemID: ElemID
+        beforeEach(() => {
+          elemID = ElemID.fromFullName(elemFullName)
+        })
+        it('should invoke the underlying MetadataQuery with correct MetadataInstance', () => {
```
amitgoldarb, line 915: This tests an implementation detail and not a side effect. I think it would be better to actually test the result.

packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
```diff
@@ -318,3 +318,27 @@ export const buildFilePropsMetadataQuery = (metadataQuery: MetadataQuery): Metad
...
   }
 }
+
+type TopLevelElemID = Omit<ElemID, 'idType'> & {
+  idType: 'type' | 'instance'
+}
+export const buildElemIDMetadataQuery = (metadataQuery: MetadataQuery): MetadataQuery<ElemID> => {
+  const getTopLevelElemID = (elemID: ElemID): TopLevelElemID => elemID.createTopLevelParentID().parent as TopLevelElemID
+  const elemIDToMetadataInstance = (id: TopLevelElemID): MetadataInstance => {
+    const fullName = invertNaclCase(id.name)
```
amitgoldarb, line 328: This logic seems slightly fragile and should probably be placed next to the getElemId funciton in transformers. We might want to place a comment above this that this shouldn't be widely used.
For example what about types that we map them? (Flexipage for example)

amitgoldarb (commented):
Looks good. Added some small implementation comments. Let me know what you think

packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
tamtamirr: I've added disclaimer on the implementation 🙏

amitgoldarb: Interesting. Then maybe even move this to the customReferences file? Making it appear less generic?

amitgoldarb (approved):
Approving to save cycles. Added one more small comment about the placement of the functions

packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
tamtamirr: Will move 🙏

PR #6555, opened 2024-09-12
SALTO-6607: Fix wrong transformation of number strings in Salesforce Adapter

tamtamirr:
Fix wrong transformation of number strings in Salesforce Adapter
 
---

Workspace diff: https://github.com/salto-io/tamir-sf/commit/d07e163cbe92370366c7faad1a68635623323926

---
_Release Notes_: 
Salesforce Adapter:
- Fixed an issue where the adapter transformed number strings incorrectly. This behavior requires the `skipParsingXmlNumbers` 

---
_User Notifications_: 
Salesforce:
- Turning on the`skipParsingXmlNumbers` feature may change some values in existing instances.


packages/salesforce-adapter/src/sfdx_parser/filters/custom_fields.ts
```diff
@@ -52,7 +52,7 @@ const getFieldsOfCustomObject = async (
         log.warn('skipping %s because we could not get its content', fileName)
         return undefined
       }
-      return xmlToValues(fileContent).values
+      return xmlToValues(fileContent, true).values
```
amitgoldarb, line 55: Shouldn't this be managed by the config?

packages/salesforce-adapter/src/sfdx_parser/sfdx_parser.ts
```diff
@@ -87,7 +87,7 @@ const getElementsFromFile = async (
     log.warn('skipping %s because we could not get its content', fileName)
     return []
   }
-  const { typeName, values } = xmlToValues(fileContent)
+  const { typeName, values } = xmlToValues(fileContent, true)
```
amitgoldarb, line 90: Shouldn't this be managed by the config?

packages/salesforce-adapter/src/transformers/xml_transformer.ts
```diff
@@ -280,14 +282,21 @@ export const complexTypesMap: ComplexTypesMap = {
...
   ignoreAttributes: false,
   attributeNamePrefix: XML_ATTRIBUTE_PREFIX,
   ignoreDeclaration: true,
   tagValueProcessor: (_name, val) => val.replace(/&#xD;/g, '\r'),
+}
+
+const parserV1 = new XMLParser(parserOptions)
+const parserV2 = new XMLParser({
+  ...parserOptions,
+  numberParseOptions: { hex: false, leadingZeros: false, eNotation: false, skipLike: /.*/ },
```
amitgoldarb, line 295: We don't want hex numbers? I beleive the only thing we have received complaints about is the leadingZeros no?
skipLike - https://github.com/NaturalIntelligence/strnum/blob/3fcb88602a08c103bccffc8c36041c3f3d0f8317/strnum.js#L40, Doesn't this mean we won't do any parsing? Are you sure about this?

packages/salesforce-adapter/src/transformers/xml_transformer.ts
```diff
@@ -352,12 +361,24 @@ const extractFileNameToData = async ({
       )
 }
 
+const getDiffValues = (oldValues: Values, newValues: Values): Values => {
+  const diff: Values = {}
+  Object.entries(newValues).forEach(([key, value]) => {
+    if (!_.isEqual(oldValues[key], value)) {
```
amitgoldarb, line 367: I wonder if this will create performance impact. We've had bad performance with _isEqual before.

packages/salesforce-adapter/src/transformers/xml_transformer.ts
```diff
@@ -352,12 +361,24 @@ const extractFileNameToData = async ({
       )
 }
 
+const getDiffValues = (oldValues: Values, newValues: Values): Values => {
```
amitgoldarb, line 364: Let's discuss this. I wonder if we could be slightly more aggressive with the rollout of this feature.

amitgoldarb (commented):
Looks good! Added a few small comments

packages/salesforce-adapter/src/fetch_profile/fetch_profile.ts
```diff
@@ -43,6 +43,8 @@ const optionalFeaturesDefaultValues: OptionalFeaturesDefaultValues = {
   taskAndEventCustomFields: false,
   sharingRulesMaps: false,
   excludeNonRetrievedProfilesRelatedInstances: true,
+  useXmlParserV2: false,
+  logXmlParserV2Diff: true,
```
ori-moisis, line 47: this is not a parser version.
I think the flag should be something like `skipParsingXmlNumbers`, generally, something that actually has some kind of meaning / hint towards what the flag actually does
the log one can be something like `logDiffsFromParsingXmlNumber`

packages/salesforce-adapter/src/sfdx_parser/filters/custom_fields.ts
ori-moisis: I wouldn't spend too much effort on this, I hope to remove this file soon

packages/salesforce-adapter/src/sfdx_parser/sfdx_parser.ts
ori-moisis: similar, this file would be completely changed.

packages/salesforce-adapter/src/transformers/xml_transformer.ts
ori-moisis: why would we keep the bug with hex?
if it is a string, it should be a string, randomly changing string values is not good...

packages/salesforce-adapter/src/transformers/xml_transformer.ts
ori-moisis, line 364: why create a new implementation of this and not use `getValuesChanges`?
seems like that is already used elsewhere in the adapter...

packages/salesforce-adapter/src/transformers/xml_transformer.ts
```diff
@@ -376,9 +397,30 @@ export const fromRetrieveResult = async (
...
+          inspectValue(diffValues),
+        )
+      }
+    }
+    const metadataValues = Object.assign(
+      xmlToValues(xmlString, fetchProfile.isFeatureEnabled('useXmlParserV2')).values,
+      {
+        [INSTANCE_FULL_NAME_FIELD]: file.fullName,
+      },
+    )
```
ori-moisis, line 423: I think the "main" parsing should happen first, and then only if the "main" parsing was with the old config and the flag to log diffs is on, there would be a need to parse again.

with the way this is now, the default config would have us parse everything 3 times (and two of those times would be parsing exactly the same thing with the same config)

packages/salesforce-adapter/test/transformers/xml_transformer.test.ts
```diff
@@ -815,13 +824,37 @@ describe('XML Transformer', () => {
...
             values: expect.objectContaining({ fullName: 'MyReportType' }),
           }),
         )
       })
     })
+    describe('with number values', () => {
+      const XML_STRING =
+        '<?xml version="1.0" encoding="UTF-8"?>\n' +
+        '<AuraDefinitionBundle xmlns="http://soap.sforce.com/2006/04/metadata">\n' +
+        '    <apiVersion>047.0</apiVersion>\n' +
```
ori-moisis, line 839: this is not really a possible example, right?
maybe we should do this with an example that is actually possible?
I assume we will want to keep this test for the long run even after the roll-out is done and the flag is removed, the test should make sense

packages/salesforce-adapter/test/transformers/xml_transformer.test.ts
```diff
@@ -815,13 +824,37 @@ describe('XML Transformer', () => {
...
     })
+    describe('with number values', () => {
+      const XML_STRING =
+        '<?xml version="1.0" encoding="UTF-8"?>\n' +
+        '<InstalledPackage xmlns="http://soap.sforce.com/2006/04/metadata">\n' +
+        '    <version>47.0</version>\n' +
+        '    <fullName>SBQQ</fullName>\n' +
+        '</AuraDefinitionBundle>\n'
+      describe('when skipParsingXmlNumbers is false', () => {
+        it('should attempt to convert the string to number', () => {
```
ori-moisis, line 843: not loving that fact that this adds a unit test that asserts our code behaves in a way that we consider a bug...
maybe change the wording on the test or add a comment saying this is a bug and not desired behavior?

ori-moisis (commented):
code LGTM
will let @amitgoldarb do the final approval as I assume you will be covering the next steps here (i.e - seeing the compare results, deciding how to migrate the code to the new configuration, etc...)

packages/salesforce-adapter/src/transformers/xml_transformer.ts
```diff
@@ -376,7 +391,31 @@ export const fromRetrieveResult = async (
...
+        after: valuesFromXml,
+        beforeId: undefined,
+        afterId: undefined,
+      })
+      if (detailedChanges.length > 0) {
+        typesWithDiff.add(file.type)
+        log.trace(
+          'Found differences in the xml parsing of instance of type %s: %s',
+          file.fullName,
+          inspectValue(detailedChanges),
```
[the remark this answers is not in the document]
tamtamirr: Of-course 🙈  Here are some examples for the trace logs:
```
2024-09-16T10:46:52.469Z trace salesforce-adapter/transformers/xml_transformer Found differences in the xml parsing of instance of type sbaa: [
  {
    id: ElemID(salesforce.InstalledPackage.instance.compare.versionNumber),
    action: 'modify',
    data: { before: '244.0', after: 244 },
    elemIDs: { before: undefined, after: undefined }
  }
]
2024-09-16T10:46:52.471Z trace salesforce-adapter/transformers/xml_transformer Found differences in the xml parsing of instance of type TestBundle: [
  {
    id: ElemID(salesforce.AuraDefinitionBundle.instance.compare.apiVersion),
    action: 'modify',
    data: { before: '61.0', after: 61 },
    elemIDs: { before: undefined, after: undefined }
  }
]
2024-09-16T10:46:52.479Z trace salesforce-adapter/transformers/xml_transformer Found differences in the xml parsing of instance of type Campaign: [
  {
    id: ElemID(salesforce.CustomObject.instance.compare.listViews.0.filters.value),
    action: 'modify',
    data: { before: '1', after: 1 },
    elemIDs: { before: undefined, after: undefined }
  },
  {
    id: ElemID(salesforce.CustomObject.instance.compare.listViews.1.filters.value),
    action: 'modify',
    data: { before: '1', after: 1 },
    elemIDs: { before: undefined, after: undefined }
  },
  {
    id: ElemID(salesforce.CustomObject.instance.compare.webLinks.height),
    action: 'modify',
    data: { before: '500', after: 500 },
    elemIDs: { before: undefined, after: undefined }
  },
  {
    id: ElemID(salesforce.CustomObject.instance.compare.webLinks.width),
    action: 'modify',
    data: { before: '500', after: 500 },
    elemIDs: { before: undefined, after: undefined }
  }
]
```

amitgoldarb (approved):
LGTM!
Thanks for this

PR #6558, opened 2024-09-12
SALTO-5922 use provided type name for recurseInto types instead of generating default ducktype nested type name

netama:
When adding a field with `resource.recurseInto`, the user provides the name of the type they are creating, and various downstream places rely on it. this was ignored when generating the ducktype type name, causing non-standalone types to not be defined correctly (and e.g. miss field customizations and not support defining nested references)

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/test/fetch/fetch.test.ts
+++ b/packages/adapter-components/test/fetch/fetch.test.ts
@@ -6,7 +6,7 @@
  * CERTAIN THIRD PARTY SOFTWARE MAY BE CONTAINED IN PORTIONS OF THE SOFTWARE. See NOTICE FILE AT https://github.com/salto-io/salto/blob/main/NOTICES
  */
 import { MockInterface, mockFunction } from '@salto-io/test-utils'
-import { SaltoError, isInstanceElement } from '@salto-io/adapter-api'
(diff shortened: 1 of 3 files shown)


packages/adapter-components/test/fetch/fetch.test.ts
```diff
@@ -89,6 +89,7 @@ describe('fetch', () => {
     it('should generate elements correctly', async () => {
       const customSaltoError: SaltoError = {
         message: 'error fetching default option',
+        detailedMessage: 'error fetching default option',
```
netama, line 92: unrelated

shir-reifenberg (approved):
LGTM 😄 
are there any real scenarios where we use `recurseInto` without `standalone`? if so we should verify this change doesn't break any references

netama:
it can only add new references because currently these fields have type `list<unknown>` and are not traversed when searching for references (this is why I needed this fix now)
there does seem to be a broken rule in [confluence](https://github.com/salto-io/salto/blob/main/packages/confluence-adapter/src/definitions/references.ts#L40) that can now be fixed to use the correct part type and actually include the references, but will do that separately

PR #6559, opened 2024-09-14
SALTO-3986: Support Brand Email template

shir-reifenberg:
Added support for email customizations

---

- Each brand has a fixed type of `EmailTemplate` instances, for each type of email ("forget password", "user activation", etc). The email templates are auto created when the brand is added (similar to brand theme)
- For each type of email, Okta allows customizing it in different languages.
- Planned to add E2E, but our E2E account doesn't allow customizations. Will update tests in `adapter.test.ts` soon.

---
_Release Notes_: 

_Okta adapter_:
- Add support in Email Templates and customizations for Brands

---
_User Notifications_: 

_Okta adapter_:
- `EmailCustomization` and `EmailTemplate` elements will be added for workspaces with brands customizations.


packages/okta-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -840,6 +844,106 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
...
+                transformation: {
+                  root: 'settings',
+                },
+              },
+              condition: { skipIfIdentical: true, transformForCheck: { root: 'settings' } },
+            },
+          ],
+          // EmailTemplates are removed automatically by Okta when the Brand is removed.
+          // A separate Change Validator ensures that templates aren't removed by themselves.
+          remove: [{ request: { earlySuccess: true } }],
```
Nurdok, line 893: In other types that are automatically removed (`ProfileMapping`, `BrandTheme`) we perform a GET to make sure that the instance was deleted. Should we do this here, too?

packages/okta-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -840,6 +844,106 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
...
+        },
+      },
+      toActionNames: ({ change }) => (isAdditionChange(change) ? ['add', 'modify'] : [change.action]),
+      actionDependencies: [{ first: 'add', second: 'modify' }],
+    },
+    EmailCustomization: {
```
Nurdok, line 899: Please put this in a `const` too.

packages/okta-adapter/src/definitions/fetch/fetch.ts
```diff
@@ -705,6 +718,83 @@ const createCustomizations = ({
...
   },
+  [EMAIL_TEMPLATE]: {
+    requests: [
+      {
+        endpoint: {
+          path: '/api/v1/brands/{brandId}/templates/email',
+          queryArgs: { expand: 'settings' },
+        },
+        transformation: {
+          rename: [{ from: '_embedded.settings', to: 'settings', onConflict: 'override' }],
```
Nurdok, line 729: nice

packages/okta-adapter/e2e_test/adapter.test.ts
Nurdok, line 1: Please also add some HTTP-recording tests (the other `adapter.test.ts` 😄)

Nurdok (changes requested):
Looks great! That was really fast!

packages/okta-adapter/src/definitions/deploy/deploy.ts
shir-reifenberg: true, wanted to implement it with `additionalStatusCode` but I think there's might be an issue with it (opened SALTO-6664), I think we can start like this and adjust once fixed

packages/okta-adapter/e2e_test/adapter.test.ts
shir-reifenberg: Added 😄 seems like i'm doing something wrong cause the entire thing fails in my added tests, let's discuss tomorrow IRL

Nurdok (approved):

PR #6560, opened 2024-09-15
SALTO-6651: (Salesforce) Fetch missing WaveDataflows Instances

tamtamirr:
Fetch missing Salesforce WaveDataflows Instances.

---

Workspace Diff: https://github.com/salto-io/tamir-sf/commit/a36f0fc2d64d55c61b91e39d09e927d5a7533798

Some WaveDataflows instances are not fetched, since they are not listed in Salesforce.
This PR extends the capability of the client `Custom List Functions` with a new mode `extendsOriginal`, where the list result for the type is extended by the custom list function.

- Implemented under the optional feature `waveMetadataSupport` that is turned off by default.

---
_Release Notes_: 
_Salesforce Adapter_:
- Fetch missing Salesforce WaveDataflows Instances. (Requires the `waveMetadataSupport` optional feature).

---
_User Notifications_: 
_Salesforce_:
- Enabling the `waveMetadataSupport` optional feature may result in new `WaveDataflow` Instances and references in existing Workspaces.


packages/salesforce-adapter/src/client/client.ts
```diff
@@ -721,19 +724,32 @@ export default class SalesforceClient implements ISalesforceClient {
...
           type,
           inspectValue(e),
         )
         return this.fullListPromisesByType[type] ?? this.sendChunkedList([{ type }], isUnhandledError)
-      }) ?? this.sendChunkedList([{ type }], isUnhandledError)
+      })
+      // In this mode, we run the custom list function and then extend the result with the original list call
+      if (customListFuncDef.mode === 'extendsOriginal') {
+        const customListResult = await request
+        const sendChunkedResult = await this.sendChunkedList([{ type }], isUnhandledError)
```
amitgoldarb, line 744: Can we do this in parallel? Or should this be appended to the fullListPromisesByType?

packages/salesforce-adapter/src/client/custom_list_funcs.ts
```diff
@@ -63,5 +71,35 @@ export const createListApexClassesDef = (elementsSource: ReadOnlyElementsSource)
...
+    const recipePropsToDataflowProps = (recipeProps: FileProperties): FileProperties => {
+      const clonedProps = _.clone(recipeProps)
+      return {
+        ...clonedProps,
+        id: '',
+        fileName: clonedProps.fileName.replace(WAVE_RECIPE_FILE_EXTENSION, WAVE_DATAFLOW_FILE_EXTENSION),
+        type: WAVE_DATAFLOW_METADATA_TYPE,
+      }
+    }
+    const { result: waveRecipesProps } = await client.listMetadataObjects([{ type: WAVE_RECIPE_METADATA_TYPE }])
```
amitgoldarb, line 89: Can we somehow reuse the result from the previous list?

packages/salesforce-adapter/src/client/custom_list_funcs.ts
```diff
@@ -63,5 +71,35 @@ export const createListApexClassesDef = (elementsSource: ReadOnlyElementsSource)
...
+    const missingWaveDataflowsFileProps = waveRecipesProps.map(recipePropsToDataflowProps)
```
amitgoldarb, line 90: It's not the end of the world to get this wrong, but I'm not entirely sure about this logic. There are some WaveFlows that we get, not sure that all of the recipies actually create a missing one, it might be a bit more complicated than that. I think we can start like this and maybe over request a bit, just need to monitor and than we can figure out a better heuristic later.

amitgoldarb (approved):
Approving to save cycles. Added a few smaller comments.
Thanks for this!

packages/salesforce-adapter/src/client/custom_list_funcs.ts
tamtamirr: There's caching in the client :) On going request to list on this type will actually return a promise.

packages/salesforce-adapter/src/client/custom_list_funcs.ts
tamtamirr: My initial implementation listed the `waveRecipe` props to avoid duplicates. This caused infinite loop since we've overridden the implementation of the list to use this custom list function. If there are duplications we will see warnings, in the adapter and it will choose the first property, however we can handle that in the client to prefer file properties from the original list (unique by the full name). Will handle and add UT 🙏

packages/salesforce-adapter/src/client/client.ts
tamtamirr: Can be done in parallel 🙏 good point

packages/salesforce-adapter/src/client/client.ts
```diff
@@ -721,19 +724,37 @@ export default class SalesforceClient implements ISalesforceClient {
...
         )
         return this.fullListPromisesByType[type] ?? this.sendChunkedList([{ type }], isUnhandledError)
-      }) ?? this.sendChunkedList([{ type }], isUnhandledError)
+      })
+      // In this mode, we run the original list function and then extend the result with the custom list function result
+      if (customListFuncDef.mode === 'extendsOriginal') {
+        const [originalListResult, customListResult] = await Promise.all([
+          this.sendChunkedList([{ type }], isUnhandledError),
+          request,
+        ])
```
amitgoldarb, line 746: Can we somehow append this to the fullListPromisesByType? This means less parallel lists right?

amitgoldarb (commented):
Added a small question

PR #6561, opened 2024-09-15
SALTO-6601: [Intune] Support fetch & deploy for Windows Platform Scripts

shoshana-houri:
Support fetching & deploying Intune Linux Platform Scripts - including assignments. Extract the script value to a static file.

---
This is the second out of three types of Intune Platform Scripts. Each type has different endpoints and structures, so the PRs are divided accordingly.

See .nacls structure example [here](https://github.com/salto-io/microsoft-security/tree/main/microsoft_security/Records/IntunePlatformScriptWindows).
See the matching static resource example [here](https://github.com/salto-io/microsoft-security/tree/main/static-resources/microsoft_security/IntunePlatformScriptWindows). 

---
_Release Notes_: 
_Microsoft Security adapter:_
* Support fetch & deploy for Intune Windows Platform Scripts.

---
_User Notifications_: 
_Microsoft Security adapter:_
* Support fetch & deploy for Intune Windows Platform Scripts.


packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/platform_script.ts
```diff
@@ -0,0 +1,114 @@
...
+}: {
+  typeName: string
+  uniquePath: FilePath
+  fileExtension: 'sh' | 'ps1'
+  content: string
+}): StaticFile => {
+  const formattedUniquePath = uniquePath
+    .split('/')
+    .map(naclCase)
+    .map(name => name.replace('@', '_'))
```
Nurdok, line 44: Since the original value can contain underscores, and those will remain after applying `naclCase`, it's possible (though very unlikely) that this can result in file path collisions - e.g., for `my path/script` (which will turn into `my_path_s/script`) and `my_path_s/script` which will remain as-is.

I suggest replacing `@` with `.` instead, since they cannot exist after applying `naclCase` and they are file-system friendly. We used the same replacement in Okta, see https://github.com/salto-io/salto/blob/5f47ca1deb994e3170f03410db62d874dd135517/packages/okta-adapter/src/logo.ts#L162

packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/platform_script.ts
```diff
@@ -0,0 +1,114 @@
...
+
+const createStaticFileFromScript = ({
+  typeName,
+  uniquePath,
+  fileExtension,
+  content,
+}: {
+  typeName: string
+  uniquePath: FilePath
+  fileExtension: 'sh' | 'ps1'
```
Nurdok, line 38: What's the reason to limit these and not use `string`?

Nurdok (approved):
LG! Left a note about static file naming, approving to save cycles.

packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/platform_script.ts
shoshana-houri: Cool, thanks!

packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/platform_script.ts
shoshana-houri: I just thought it's worth limiting the options if we know what they are, and if the limitation is straightforward enough 😅

packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/platform_script.ts
shoshana-houri: Actually, I think I'll use `-` instead of `.` to avoid confusion with the file extension. 🤔 wdyt?

Nurdok: @IdoZakk is working on a PR (https://github.com/salto-io/salto/pull/6518) to extract our Okta logic here to `adapter-utils` so IMO if this becomes a cross-adapter convention, I'd say we use it here too. There shouldn't be an issue with the file extensions, dots in filenames are valid.

packages/microsoft-security-adapter/src/definitions/deploy/shared/utils/condition_changes_in_fields.ts
```diff
@@ -23,7 +23,9 @@ export const createCustomConditionCheckChangesInFields = (
     () =>
     ({ change }) => {
       if (isAdditionChange(change)) {
-        return !_.isEmpty(_.pick(getChangeData(change).value, fieldNames))
+        return Object.entries(_.pick(getChangeData(change).value, fieldNames)).some(
+          ([_fieldName, value]) => !(Array.isArray(value) && _.isEmpty(value)),
```
Nurdok, line 27: If you don't need the key, why not call `object.values` instead?

Nurdok (approved):

PR #6562, opened 2024-09-15
SALTO-2511: (Salesforce) Convert Wave Metadata Instances content to Static Files

tamtamirr:
(Salesforce) Convert Wave Metadata Instances content to Static Files

---

Workspace diff: https://github.com/salto-io/tamir-sf/commit/cb64a7e0bd1cd509f399d798abc77e5c15915642

This PR tackles two issues:
1. The API behaves weirdly on deploy for these instances. It actually accepts the decoded base64 content in order to deploy successfully. For that reason, I didn't implement any transformation for deploy.
2. Moved the content into static files.

---
_Release Notes_: 
_Salesforce Adapter_:
- Moved Wave Metadata Instances content into static files. This should also fix deploy issues when attempting to modify / add these instances. Requires the optional feature `waveMetadataSupport`

---
_User Notifications_: 
_Salesforce_:
- Enabling the `waveMetadataSupport` optional feature, will change existing Wave Metadata instances to have a static file instead of raw content in the nacl.


packages/salesforce-adapter/src/filters/wave_static_files.ts
```diff
@@ -0,0 +1,57 @@
...
+// is need before deploying back to the service. The transformation in fetch
+// should allow deploying these instances successfully.
+const filter: LocalFilterCreator = ({ config }) => ({
+  name: 'waveStaticFiles',
+  onFetch: ensureSafeFilterFetch({
+    warningMessage: 'Error occurred while attempting to convert Wave Metadata content to static files',
+    config,
+    filterName: 'waveMetadataSupport',
+    fetchFilterFunc: async elements => {
+      elements.filter(isInstanceOfTypeSync(...WAVE_TYPES_WITH_STATIC_FILES)).forEach(convertContentToStaticFile)
```
amitgoldarb, line 52: This function needs to be behind the config option right? Theoretically that means we are missing a UT if this passed.

amitgoldarb (approved):
Approving to save cycles, but I think there is a bug here. Please fix it before merging 🙏

packages/salesforce-adapter/src/filters/wave_static_files.ts
tamtamirr: It is behind the FF, this is not a bug (See ensureSafeFilterFetch). Added UTs to the scenario where the feature is disabled as-well

PR #6567, opened 2024-09-16
SALTO-6665: fix updateElemIDDefinitions

edenhassid:
fix updateElemIDDefinitions

---

---
_Release Notes_: 

---
_User Notifications_: 


packages/zendesk-adapter/src/adapter_creator.ts
```diff
@@ -114,6 +113,7 @@ const adapterConfigFromConfig = (config: Readonly<InstanceElement> | undefined):
   ) as configUtils.AdapterDuckTypeApiConfig
 
   const fetch = mergeWithDefaultConfig(DEFAULT_CONFIG.fetch, config?.value.fetch) as ZendeskFetchConfig
+  const useNewInfra = configValue.fetch?.useNewInfra
   if (useNewInfra === true) {
```
netama, line 117: nit: if `fetch` could still be undefined, might be safer to change these to `useNewInfra !== false`

netama (approved):
LGTM with a suggestion (though I assume there might be other places that should be checked?)

PR #6568, opened 2024-09-16
SALTO-6663: [Intune] Support fetch & deploy for Scope Tags

shoshana-houri:
Support fetching & deploying Intune Scope Tags - including assignments.

---

Although it's not a particularly complex type, all the current Intune types reference it, making it a significant addition.

You can see an example of the .nacl structure [here](https://github.com/salto-io/microsoft-security/tree/main/microsoft_security/Records/IntuneScopeTag). 
As mentioned, references can be found across all types. For example, see [this one](https://github.com/salto-io/microsoft-security/blob/main/microsoft_security/Records/IntuneFilter/svh_filter_apps_windows_windowsMobileApplicationManagement.nacl).

---
_Release Notes_: 
_Microsoft Security adapter:_
* Support fetch & deploy for Intune Scope Tags.

---
_User Notifications_: 
_Microsoft Security adapter:_
* Support fetch & deploy for Intune Scope Tags.


packages/microsoft-security-adapter/src/definitions/fetch/intune/fetch.ts
```diff
@@ -302,8 +304,68 @@ const graphBetaCustomizations: FetchCustomizations = {
...
+    },
+  },
   ...TYPES_WITH_GROUP_ASSIGNMENTS_ASSIGNMENTS.map(typeName => ({
     [typeName]: {
+      ...(typeName === SCOPE_TAG_ASSIGNMENTS_TYPE_NAME
+        ? {
+            requests: [
+              {
+                endpoint: {
+                  path: '/deviceManagement/roleScopeTags/{id}/assignments' as EndpointPath,
```
shoshana-houri, line 362: For some reason, specifying `$expand=assignments` in the query params of the parent request, as we do for other types, always returns an empty assignments list for scope tags. Instead, we fetch the assignments directly for them.

packages/microsoft-security-adapter/src/definitions/fetch/intune/fetch.ts
```diff
@@ -302,8 +304,68 @@ const graphBetaCustomizations: FetchCustomizations = {
...
+                },
+                transformation: DEFAULT_TRANSFORMATION,
+              },
+            ],
+          }
+        : {}),
```
shoshana-houri, line 368: For some reason, specifying $expand=assignments in the query params of the parent request, as we do for other types, always returns an empty assignments list for scope tags. Instead, we fetch the assignments directly for them.

packages/microsoft-security-adapter/src/definitions/fetch/intune/fetch.ts
```diff
@@ -302,8 +304,68 @@ const graphBetaCustomizations: FetchCustomizations = {
...
+        },
+      },
+      mergeAndTransform: {
+        adjust: async ({ value }) => {
+          validatePlainObject(value, SCOPE_TAG_TYPE_NAME)
+          return {
+            value: {
+              ...value,
+              // Workaround to allow empty arrays for a recurseInto field
+              [ASSIGNMENTS_FIELD_NAME]: value[ASSIGNMENTS_FIELD_NAME] ?? [],
```
shir-reifenberg, line 337: isn't `allowEmptyArrays` enough here?

packages/microsoft-security-adapter/src/definitions/fetch/intune/fetch.ts
shir-reifenberg, line 362: is the casting needed ?

packages/microsoft-security-adapter/src/definitions/fetch/intune/fetch.ts
```diff
@@ -302,8 +304,68 @@ const graphBetaCustomizations: FetchCustomizations = {
...
+          path: '/#view/Microsoft_Intune_DeviceSettings/ScopeTagSummaryBlade/roleScopeTagId/{id}/roleScopeTagDisplayName/{displayName}',
+        },
+        allowEmptyArrays: true,
+      },
+      fieldCustomizations: ID_FIELD_TO_HIDE,
+    },
+  },
   ...TYPES_WITH_GROUP_ASSIGNMENTS_ASSIGNMENTS.map(typeName => ({
     [typeName]: {
+      ...(typeName === SCOPE_TAG_ASSIGNMENTS_TYPE_NAME
```
shir-reifenberg, line 357: nit: maybe just move to it out of `TYPES_WITH_GROUP_ASSIGNMENTS_ASSIGNMENTS` for readability?

shir-reifenberg (approved):
Looks great 💃

packages/microsoft-security-adapter/src/definitions/fetch/intune/fetch.ts
shoshana-houri: No, see: https://github.com/salto-io/salto/blob/8123eea26067123d50a5ade0d09bd3f1e45efbbc/packages/adapter-components/src/fetch/resource/subresources.ts#L116
This field is not added at all if the `recurseInto` result returns an empty array. That's why, as a workaround, we explicitly add it as an empty array **and** set `allowEmptyArray = true` to retain this value when generating the instances.

packages/microsoft-security-adapter/src/definitions/fetch/intune/fetch.ts
shoshana-houri: `TYPES_WITH_GROUP_ASSIGNMENTS_ASSIGNMENTS` is used not only for the fetch definition but also for reference purposes, so I'd prefer to keep it there. (If more types will require specific handling, we can update `TYPES_WITH_GROUP_ASSIGNMENTS_ASSIGNMENTS` to be a map with the appropriate values or something similar, but I think it's too much for a single use-case)

PR #6569, opened 2024-09-16
SALTO-6464 do not run actions if earlier actions failed

netama:
add option (at the request level) to skip a request if we already know the change is going to be marked as failed, and default to skipping.

---

this is currently needed for okta.

allowing the customization to allow for cases where a cleanup is needed, or for more tailored conditions (e.g. only fail on specific error types / treat some warnings as errors / take into account errors from other changes - though this is limited and can be extended in SALTO-6669).

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/adapter-components/src/deployment/deploy/deploy.ts
```diff
@@ -122,18 +122,28 @@ export const deployChanges = async <TOptions extends APIDefinitionsOptions>({
...
-            log.error('Deployment of %s failed: %o', getChangeData(change).elemID.getFullName(), err)
+            const { elemID } = getChangeData(change)
+            if (errors[elemID.getFullName()] === undefined) {
+              errors[elemID.getFullName()] = []
+            }
+            log.error('Deployment of %s failed: %o', elemID.getFullName(), err)
             if (isSaltoError(err)) {
-              errors.push({
+              errors[elemID.getFullName()].push({
+                // TODON index errors by elem id so that it's easier to look up?
```
shir-reifenberg, line 140: leftover? :)

packages/adapter-components/src/deployment/deploy/deploy.ts
```diff
@@ -122,18 +122,28 @@ export const deployChanges = async <TOptions extends APIDefinitionsOptions>({
       await Promise.all(
         typeActionChanges.map(async change => {
           try {
-            await limitedDeployChange({ ...changeContext, change, action })
+            await limitedDeployChange({
```
shir-reifenberg, line 125: sorry if i'm missing something here (didn't test my usecase locally), but isn't the fact we're returning value when `limitedDeployChange` skipped the request means it will get into `appliedChanges` list?

packages/adapter-components/src/deployment/deploy/requester.ts
```diff
@@ -67,19 +66,25 @@ const createExtractor = (transformationDef?: TransformDefinition<ChangeAndContex
...
   // since the comparison will return the same value when working with two arrays vs two individual items
   const transform = createValueTransformer(transformForCheck)
   return async args => {
-    const { change } = args
+    const { change, errors } = args
+    if (
+      failIfChangeHasErrors !== false &&
+      !_.isEmpty(errors[getChangeData(change).elemID.getFullName()]?.filter(e => e.severity === 'Error'))
+    ) {
+      return false
```
shir-reifenberg, line 83: might be useful to have a `log.debug` here ?

shir-reifenberg (commented):
Nice 💯 
I have one question about the applied changes

packages/adapter-components/src/deployment/deploy/requester.ts
```diff
@@ -67,19 +66,25 @@ const createExtractor = (transformationDef?: TransformDefinition<ChangeAndContex
...
-const createCheck = (conditionDef?: DeployRequestCondition): ((args: ChangeAndContext) => Promise<boolean>) => {
-  const { custom, transformForCheck, skipIfIdentical } = conditionDef ?? {}
+const createCheck = (conditionDef?: DeployRequestCondition): ((args: ChangeAndExtendedContext) => Promise<boolean>) => {
+  const { custom, transformForCheck, skipIfIdentical, failIfChangeHasErrors } = conditionDef ?? {}
   if (custom !== undefined) {
-    return async input => custom({ skipIfIdentical, transformForCheck })(input)
-  }
-  if (skipIfIdentical === false) {
-    return async () => true
+    return async input => custom({ skipIfIdentical, transformForCheck, failIfChangeHasErrors })(input)
```
shir-reifenberg, line 72: another question - should `custom` overrides the `failIfChangeHasErrors` ? cause in this we are asking the developer to be aware to this behavior

packages/adapter-components/src/deployment/deploy/deploy.ts
netama: you're right, my bad :) moved this to be higher and keeping an option to use this in the custom functions for special cases (so this is now closer to what you were aiming for originally I think, but still with an option to opt-out)

packages/adapter-components/src/deployment/deploy/requester.ts
netama: wdym by override? all the `custom` functions until now were assumed to be able to call/mimic the original logic if needed, wanted to keep it as flexible as possible (but we also haven't seen that many cases yet)

packages/adapter-components/src/deployment/deploy/deploy.ts
```diff
@@ -115,25 +119,42 @@ export const deployChanges = async <TOptions extends APIDefinitionsOptions>({
     )
     const { concurrency } = defQuery.query(String(typeName)) ?? {}
 
-    const bucket = new RateLimiter({ maxConcurrentCalls: concurrency })
-    const limitedDeployChange = bucket.wrap(deploySingleChange)
+    const limiter = new RateLimiter({ maxConcurrentCalls: concurrency })
```
netama, line 122: unrelated - changing the name back, looks like it was changed but I don't understand the reason for the new name 😅

shir-reifenberg (approved):
LGTM! thanks 🔥 🔥

PR #6570, opened 2024-09-16
SALTO-6611: Targeted CBF fetches modified Flow instances even though the Flow MetadataType is not in the fetch targets

tamtamirr:
Targeted CBF fetches modified Flow instances even though the Flow MetadataType is not in the fetch targets

---

The filter implementation relied on the existence of the Flow MetadataType to understand if the Flow type should be retrieved or not. This caused a weird behavior in Change-Based Fetches where we get the types from the Elements Source and not filter them. This will be fixed in the major changes from the way we handle types in CBF here: https://github.com/salto-io/salto/pull/6389.

---

_Release Notes_: 
_Salesforce Adapter_:
- Bug-fix: Targeted Change-based fetch fetches modified Flow instances even though the Flow MetadataType is not in the fetch targets.

---
_User Notifications_: 
_None_


amitgoldarb (approved):
Thanks for this

PR #6574, opened 2024-09-17
SALTO-6675: add detailedMessage to salesforce custom object deploy error

shayc331:
_add detailedMessage to salesforce custom object deploy error_

---

* remove other casting I found as well

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


dantal4 (approved):
Thanks!

PR #6575, opened 2024-09-17
SALTO-6677: (Salesforce) Simplify metadataQuery for targeted fetch

tamtamirr:
(Salesforce) Simplify metadataQuery for targeted fetch.

---

Moved metadata targets dependencies from the metadataQuery implementation to the dependencies definitions.
Tested targeted fetch on both Flow and CustomObject.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
+++ b/packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
@@ -14,11 +14,8 @@ import { safeJsonStringify } from '@salto-io/adapter-utils'
 import {
   CUSTOM_OBJECT,
   DEFAULT_NAMESPACE,
-  FLOW_DEFINITION_METADATA_TYPE,
-  FLOW_METADATA_TYPE,
   MAX_TYPES_TO_SEPARATE_TO_FILE_PER_FIELD,
   SETTINGS_METADATA_TYPE,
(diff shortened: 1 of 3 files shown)


packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
```diff
@@ -98,27 +95,16 @@ export const buildMetadataQuery = ({ fetchParams }: BuildMetadataQueryParams): M
...
-      return true
-    }
-    if (type === TOPICS_FOR_OBJECTS_METADATA_TYPE && target.includes(CUSTOM_OBJECT)) {
-      return true
-    }
-    // We should really do this only when config.preferActiveFlowVersions is true
-    // if you have another use-case to pass the config here also handle this please
-    if (type === FLOW_DEFINITION_METADATA_TYPE && target.includes(FLOW_METADATA_TYPE)) {
-      return true
-    }
```
tamtamirr, line 115: This comment is outdated. No reason not to fetch FlowDefinitions, especially since they affect values in Profiles. These instances are not managed in the Workspace either way so there shouldn't be any implications and this won't be a regression as current targeted fetch of Flows included them.

amitgoldarb (approved):

PR #6576, opened 2024-09-17
SALTO-6674: fix scriptRunner workflow for old formats

IdoZakk:
Fixed a bug in ScriptRunner wokrflow for DC
---

The original implementation considered the following structure for some of the fields for scripts:
{
    script: "...."
}
The entire object is encoded.
We saw cases where there was not such structure, and others where the script was not encoded.
From now on all scripts will have the same structure (like the new format).
noise reduction: https://github.com/salto-io/salto_private/pull/9348

---
_Release Notes_: 
Jira Adapter: 
* Fixed a bug that prevented deployment of DC workflows with old format scriptRunner scripts

---
_User Notifications_: 
Jira Adapter:
* ScriptRunner scripts in old formats for DC will look like new format scripts inte NaCls

### DIFF
--- a/packages/jira-adapter/src/filters/script_runner/workflow/workflow_dc.ts
+++ b/packages/jira-adapter/src/filters/script_runner/workflow/workflow_dc.ts
@@ -24,13 +24,13 @@ const decodeBase64 = (base64: string): string => {
   try {
     const decoded = Buffer.from(base64, 'base64').toString('utf8')
     if (!decoded.startsWith(DC_ENCODE_PREFIX)) {
-      log.warn(`Could not decode DC ScriptRunner script, expected to start with ${DC_ENCODE_PREFIX}, got: ${decoded}`)
+      log.info(`Could not decode DC ScriptRunner script, expected to start with ${DC_ENCODE_PREFIX}, got: ${decoded}`)
       return base64
     }
     // all base64 strings of DC ScriptRunner scripts start with `!` (or YCFg in base 64)
     return decoded.substring(DC_ENCODE_PREFIX.length)
   } catch (e) {
-    log.warn(`Could not decode DC ScriptRunner script, expected base64, got: ${base64}`)
+    log.info(`Could not decode DC ScriptRunner script, expected base64, got: ${base64}`)
     return base64
   }
 }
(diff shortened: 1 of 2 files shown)


packages/jira-adapter/test/filters/script_runner/workflow/workflow_dc.test.ts
```diff
@@ -160,13 +160,6 @@ describe('Scriptrunner DC Workflow', () => {
         await filterOff.onFetch([instance])
         expect(instance.value.transitions.tran1.rules.postFunctions[0].configuration.FIELD_NOTES).toEqual(goodBase64)
       })
-      it('should fail if script not in json format', async () => {
```
kami-delaroz, line 163: do we want to ensure the input is in either one of: plain-text, base64 and json with the "scriptId" field?

kami-delaroz (approved):
🥇

packages/jira-adapter/src/filters/script_runner/workflow/workflow_dc.ts
```diff
@@ -24,13 +24,13 @@ const decodeBase64 = (base64: string): string => {
   try {
     const decoded = Buffer.from(base64, 'base64').toString('utf8')
     if (!decoded.startsWith(DC_ENCODE_PREFIX)) {
-      log.warn(`Could not decode DC ScriptRunner script, expected to start with ${DC_ENCODE_PREFIX}, got: ${decoded}`)
+      log.info(`Could not decode DC ScriptRunner script, expected to start with ${DC_ENCODE_PREFIX}, got: ${decoded}`)
```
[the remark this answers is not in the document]
IdoZakk: It can happen, we know the scenario, it is not a warn anymore.
I do not want to move it to trace as there still might be issues around it, so someone inspecting the logs around this area should see them even if not looking for them

packages/jira-adapter/test/filters/script_runner/workflow/workflow_dc.test.ts
IdoZakk: Not sure. I think the code is safe now for things we do not expect, we do not ensure it in other places. WDYT?

PR #6579, opened 2024-09-17
SALTO-6679: add timeout only for safe requests

edenhassid:
add timeout only for safe requests

---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/adapter-components/test/client/http_connection.test.ts
+++ b/packages/adapter-components/test/client/http_connection.test.ts
@@ -134,7 +134,7 @@ describe('client_http_connection', () => {
 
       const httpClient = await connection.login(mockCredentials)
 
-      // Mock a POST request
+      // Mock a GET request
       mockAxiosAdapter.onGet('/test').reply(200, { data: 'success' })
 
       // Make a POST request
--- a/packages/adapter-components/test/client/http_connection.test.ts
+++ b/packages/adapter-components/test/client/http_connection.test.ts
@@ -85,6 +85,64 @@ describe('client_http_connection', () => {
       ).rejects.toThrow(new Error('Login failed with error: aaa'))
     })
   })
+  describe('axiosConnection with timeout', () => {
+    let retryOptions: RetryOptions
+    beforeEach(() => {
+      retryOptions = createRetryOptions(
+        {
+          maxAttempts: 3,
+          retryDelay: 100,
+          additionalStatusCodesToRetry: [],
+        },
+        {
+          retryOnTimeout: true,
+          lastRetryNoTimeout: true,
+        },
+      )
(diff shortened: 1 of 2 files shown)


packages/adapter-components/test/client/http_connection.test.ts
```diff
@@ -85,6 +85,64 @@ describe('client_http_connection', () => {
...
+
+      const mockCredentials = { username: 'test', password: 'test' }
+
+      const httpClient = await connection.login(mockCredentials)
+
+      // Mock a POST request
+      mockAxiosAdapter.onPost('/test').reply(200, { data: 'success' })
+
+      // Make a POST request
+      const response = await httpClient.post('/test', {})
```
adfineberg, line 120: A bit surprising that you're looking at the POST response, not the request. Shouldn't the config be on the request?

packages/adapter-components/test/client/http_connection.test.ts
```diff
@@ -85,6 +85,64 @@ describe('client_http_connection', () => {
...
+        baseURLFunc: async () => BASE_URL,
+        credValidateFunc: async () => ({ accountId: '1' }),
+        timeout: 5000,
+      })
+
+      const mockCredentials = { username: 'test', password: 'test' }
+
+      const httpClient = await connection.login(mockCredentials)
+
+      // Mock a POST request
```
adfineberg, line 137:
```suggestion
      // Mock a GET request
```

packages/adapter-components/test/client/http_connection.test.ts
edenhassid: from GPT:
Yes, you're correct that looking at the response's configuration (response.config.timeout) to verify a request's timeout can be unintuitive, and it might seem surprising. However, this is a side effect of how Axios works—when you make a request, Axios attaches the config object (which includes details like the timeout, headers, and other request-specific settings) to the response object that it returns.

this is the only thing that worked for me in this test 🤷‍♀️

packages/adapter-components/test/client/http_connection.test.ts
```diff
@@ -85,6 +85,64 @@ describe('client_http_connection', () => {
...
+      })
+
+      const mockCredentials = { username: 'test', password: 'test' }
+
+      const httpClient = await connection.login(mockCredentials)
+
+      // Mock a GET request
+      mockAxiosAdapter.onGet('/test').reply(200, { data: 'success' })
+
+      // Make a POST request
```
adfineberg, line 140:
```suggestion
      // Make a GET request
```

adfineberg (approved):
Small nit, approved!

PR #6580, opened 2024-09-17
SALTO-4678: (Salesforce) Support targeted fetch for specific Custom Objects 

tamtamirr:
(Salesforce) Support targeted fetch for specific Custom Objects 

---

Allow fetching specific Custom Objects and their data instances by providing the CustomObject apiName as part of the fetch targets.

---
_Release Notes_: 
_Salesforce Adapter_:
- Support targeted fetch on specific Custom Objects (OSS only)

---
_User Notifications_: 
_None_


amitgoldarb:
This includes several commits that were merged already right? Can we rebase this PR?

packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -954,3 +963,94 @@ export const isCustomField = (field: Field): boolean => field.name.endsWith(SALE
...
+    )
+    log.trace('Singleton values are: %s', inspectValue(orgSettingsValues))
+    return {
+      metadataTypes: SUPPORTED_METADATA_TYPES,
+      customObjects: [],
+      customObjectsLookups: {},
+    }
+  }
+  return {
+    metadataTypes: SUPPORTED_METADATA_TYPES,
```
amitgoldarb, line 1005: Should this be hard coded? Shouldn't it be what we get from listMetadataTypes, or also some permutation of that with the adapter.config?
Doesn't have to be in the first iteration of this PR though, but if we are calculating things dynamically we can improve the metadata list as well to not be hard coded.

packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -954,3 +963,94 @@ export const isCustomField = (field: Field): boolean => field.name.endsWith(SALE
...
+      getCustomObjectDependenciesRecursively(target, orgFetchTargets.customObjectsLookups, handledTypes),
+    ),
+  )
+  metadataTypeTargets.forEach(typeName => {
+    includeParams.push({
+      metadataType: typeName,
+    })
+  })
+  customObjectTargetsWithDependencies.forEach(customObjectName => {
+    includeParams.push({
```
amitgoldarb, line 1050: Nit: This can be a list of one with a regex that matches for all of the types right?

packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -954,3 +963,94 @@ export const isCustomField = (field: Field): boolean => field.name.endsWith(SALE
...
+  target: string,
+  lookups: Record<string, readonly string[]>,
+  handledTypes: Set<string>,
+): string[] => {
+  if (handledTypes.has(target)) {
+    return []
+  }
+  handledTypes.add(target)
+  const dependencies = lookups[target] ?? []
+  return dependencies.reduce(
```
amitgoldarb, line 1021: Theoretically this list can have duplicates right? Possibly let's set and then array it

packages/salesforce-adapter/src/filters/organization_settings.ts
```diff
@@ -171,6 +235,10 @@ const filterCreator: RemoteFilterCreator = ({ client, config }) => ({
         client,
         organizationInstance,
       })
+      populateCustomObjects({
+        elements: await toArrayAsync(await buildElementsSourceForFetch(elements, config).getAll()),
```
amitgoldarb, line 239: You can do this with list right? No need to get the entire object?

amitgoldarb (commented):
This looks good!
Added some small comments

packages/salesforce-adapter/src/filters/utils.ts
amitgoldarb: Do you want to open a ticket for this?

amitgoldarb: @tamtamirr Did you follow up here with a ticket? Which one?

packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -968,3 +975,96 @@ export const getProfilesAndPermissionSetsBrokenPaths = async (
...
+  const dependencies = lookups[target] ?? []
+  return _.uniq(
+    dependencies.reduce(
+      (acc, dep) => acc.concat(dep, getCustomObjectDependenciesRecursively(dep, lookups, handledTypes)),
+      [target],
+    ),
+  )
+}
+
+export const getMetadataIncludeFromFetchTargets = async (
```
amitgoldarb, line 1041: Nit: This function is also for data.
```suggestion
export const getMetadataQueryIncludeFromFetchTargets = async (
```

packages/salesforce-adapter/src/filters/utils.ts
```diff
@@ -968,3 +975,96 @@ export const getProfilesAndPermissionSetsBrokenPaths = async (
...
+  targets: string[],
+  elementsSource: ReadOnlyElementsSource,
+): Promise<MetadataQueryParams[]> => {
+  const targetsWithDependencies = getFetchTargetsWithDependencies(targets)
+  const includeParams: MetadataQueryParams[] = []
+  const orgFetchTargets = await getOrgFetchTargets(elementsSource)
```
amitgoldarb, line 1047: The fact that we use the element source here has a surprising side effect right? If I change the adapter.config and run a targeted fetch on that specific type also in the same fetch It won't work as promised. Considering we want to use this list as the base for the custom fetches later down the line I think this is relatively low risk. Just thinking out loud here right?
Is there anyway where this could be supported? Would we have to calculate the entire filter again here?

packages/salesforce-adapter/src/filters/fetch_targets.ts
```diff
@@ -0,0 +1,77 @@
...
+  const customObjects = elements.filter(isCustomObjectSync)
+  const customObjectNames: string[] = []
+  const customObjectsLookups: Record<string, string[]> = {}
+  customObjects.forEach(customObject => {
+    const objectApiName = apiNameSync(customObject)
+    if (objectApiName === undefined) {
+      return
+    }
+    customObjectNames.push(objectApiName)
+    const customObjectLookupTypes = getCustomObjectLookupTypes(customObject)
```
amitgoldarb, line 50: Another thought I had here. This means that the targets are only based on what we have successfully fetched right? What about if someone didn't have any instances of a custom object and then added some?
I think this use case isn't interesting, but just wondering your thoughts as well

amitgoldarb (commented):
Looks great! Added some small comments/thoughts before approving wondering your thoughts.

packages/salesforce-adapter/src/filters/utils.ts
tamtamirr: yes https://salto-io.atlassian.net/browse/SALTO-6804

packages/salesforce-adapter/src/filters/utils.ts
tamtamirr: Fetch targets are referring to the include section within `Metadata` though 🤔

packages/salesforce-adapter/src/filters/utils.ts
tamtamirr: As long as the user didn't run fetch which added the type (partial/full) he won't be suggested to include the type in the fetch targets. I think this is expected behavior and we can ignore this use-case.

packages/salesforce-adapter/src/filters/fetch_targets.ts
tamtamirr: Not sure I understand the concern. As long as the CustomObject itself was fetched it will be part of the `fetch targets` we calculate here (unrelated to wether we've fetched data instances of this type).

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -534,14 +538,18 @@ export default class SalesforceAdapter implements SalesforceAdapterOperations {
       client: this.client,
       metadataQuery: buildFilePropsMetadataQuery(baseQuery),
     })
+    const targetedFetchInclude = fetchParams.target
```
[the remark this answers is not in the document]
tamtamirr: Wondering why the suggested rename. The value here is `metadata include` params. Why `included`?

amitgoldarb (approved):
Thanks for this! Looks great!

PR #6581, opened 2024-09-17
SALTO-6683: (Salesforce) Extended Apex Triggers Metadata 

tamtamirr:
(Salesforce) Extended Apex Triggers Metadata 

---

Workspace Diff: https://github.com/salto-io/tamir-sf/commit/c31d1fe901d966ad962115437c347f145bc69ca1

Implemented as optional feature. This should improve our IA and dependencies between ApexTriggers and the CustomObjects they operate on.

---
_Release Notes_: 
_Salesforce Adapter_:
- Added more metadata on ApexTrigger instances. Now every trigger Instance will contain the **triggerType(s)** and a parent relationship to the CustomObject the trigger is on. Requires the optional feature `extendedTriggersMetadata`

---
_User Notifications_: 
_Salesforce_:
- Enabling the optional feature `extendedTriggersMetadata` will modify existing trigger Instances to include more metadata.


packages/salesforce-adapter/src/filters/extended_triggers_metadata.ts
```diff
@@ -0,0 +1,194 @@
...
+  customObjectsByInternalId,
+  customObjectsByApiName,
+}: {
+  trigger: InstanceElement
+  record: SalesforceRecord
+  customObjectsByInternalId: Record<string, ObjectType>
+  customObjectsByApiName: Record<string, ObjectType>
+}): void => {
+  const tableEnumOrId = record.TableEnumOrId
+  const parentObject = customObjectsByInternalId[tableEnumOrId] ?? customObjectsByApiName[tableEnumOrId]
```
amitgoldarb, line 92: Can you possibly explain this?

packages/salesforce-adapter/src/filters/extended_triggers_metadata.ts
```diff
@@ -0,0 +1,194 @@
...
+    await Promise.all(
+      _.chunk(internalIds, chunkSize).map(async chunk => {
+        const query = `SELECT Id, TableEnumOrId, ${TRIGGER_TYPE_FIELDS.join(',')} FROM ApexTrigger WHERE Id IN ('${chunk.join("','")}')`
+        return (await toArrayAsync(await client.queryAll(query))).flat()
+      }),
+    ),
+  )
+
+const updateApexTriggerMetadataType = (metadataType: ObjectType): void => {
+  metadataType.fields[TRIGGER_TYPE_FIELD] = new Field(metadataType, TRIGGER_TYPE_FIELD, BuiltinTypes.STRING)
```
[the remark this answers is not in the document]
amitgoldarb: I'm also concerned this is incorrect, shouldn't this be an array? Triggers can trigger on delete and update right?

packages/salesforce-adapter/src/filters/extended_triggers_metadata.ts
```diff
@@ -0,0 +1,194 @@
...
+    },
+    onDeploy: async changes => {
+      const appliedApexTriggerChanges = changes
+        .filter(isInstanceOfTypeChangeSync(APEX_TRIGGER_METADATA_TYPE))
+        .filter(isAdditionOrModificationChange)
+      const appliedElemIds = appliedApexTriggerChanges.map(change => getChangeData(change).elemID.getFullName())
+      _.pullAll(changes, appliedApexTriggerChanges)
+      Object.values(_.pick(originalApexTriggerChangesByElemId, appliedElemIds)).forEach(originalChange => {
+        changes.push(originalChange)
+      })
```
amitgoldarb, line 189: Isn't this overriding too much? If a different filter later did a different manipulation, this will override it right?

amitgoldarb (commented):
Looks good!
Added some comments
Also this is missing some UT right?

packages/salesforce-adapter/src/filters/extended_triggers_metadata.ts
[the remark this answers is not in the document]
tamtamirr: yes sorry, this is outdated code. Please re-review 🙏

packages/salesforce-adapter/src/filters/extended_triggers_metadata.ts
```diff
@@ -0,0 +1,194 @@
...
+  chunkSize,
+  internalIds,
+}: {
+  client: SalesforceClient
+  chunkSize: number
+  internalIds: string[]
+}): Promise<SalesforceRecord[]> =>
+  _.flatten(
+    await Promise.all(
+      _.chunk(internalIds, chunkSize).map(async chunk => {
```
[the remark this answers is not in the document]
tamtamirr: We are using the Data API for this query. Unlike the Tooling API queryAll works here. The only limitation is the MAX_QUERY_LENGTH. I don't think we should configure anything here and make this more complex. This shouldn't  be configurable.

packages/salesforce-adapter/src/filters/extended_triggers_metadata.ts
```diff
@@ -0,0 +1,194 @@
...
+        const query = `SELECT Id, TableEnumOrId, ${TRIGGER_TYPE_FIELDS.join(',')} FROM ApexTrigger WHERE Id IN ('${chunk.join("','")}')`
+        return (await toArrayAsync(await client.queryAll(query))).flat()
```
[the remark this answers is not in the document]
tamtamirr: Response size has to be the same size as the constant chunk size :) We query by Id

packages/salesforce-adapter/src/filters/extended_triggers_metadata.ts
tamtamirr: The value of TableEnumOrId is either the CustomObject API name or its internal Id.
From what I've noticed, here are some differences:
1. CustomObjects from installed packages (managed) will have the API name here.
2. Standard Objects will have the API name here.
3. Unmanaged Custom Objects will have the Internal Id here.

It'd be easier to think of this implementation as a really simple `multiIndex` :)

packages/salesforce-adapter/src/filters/extended_triggers_metadata.ts
tamtamirr: This is true, but it's a very bad practice to have two different filters operating on the same changes. We should avoid that. I wouldn't make the implementation of this filter more complex just for that sake. If you disagree, lets discuss this 🙏

amitgoldarb: I don't disagree, only that it's difficult to monitor that. Going the extra mile of applying only the part of the change we remove shouldn't be too difficult no?

packages/salesforce-adapter/src/filters/extend_triggers_metadata.ts
```diff
@@ -0,0 +1,202 @@
...
+        return (await toArrayAsync(await client.queryAll(query))).flat()
+      }),
+    ),
+  )
+
+const updateApexTriggerMetadataType = (metadataType: ObjectType): void => {
+  metadataType.fields[TRIGGER_TYPES_FIELD_NAME] = new Field(
+    metadataType,
+    TRIGGER_TYPES_FIELD_NAME,
+    new ListType(BuiltinTypes.STRING),
```
amitgoldarb, line 81: Nit: What about core restrictions for this value?

packages/salesforce-adapter/src/filters/extend_triggers_metadata.ts
```diff
@@ -0,0 +1,202 @@
...
+  customObjectsByApiName,
+}: {
+  trigger: InstanceElement
+  record: SalesforceRecord
+  customObjectsByInternalId: Record<string, ObjectType>
+  customObjectsByApiName: Record<string, ObjectType>
+}): void => {
+  const tableEnumOrId = record.TableEnumOrId
+  // The value of TableEnumOrId is either the CustomObject API name or its internal Id.
+  const parentObject = customObjectsByInternalId[tableEnumOrId] ?? customObjectsByApiName[tableEnumOrId]
```
amitgoldarb, line 98: Can we log a warning if the parentObject is null?

packages/salesforce-adapter/src/filters/extend_triggers_metadata.ts
```diff
@@ -0,0 +1,202 @@
...
+
+const ON_UPDATE_TRIGGER_TYPES = ['UsageBeforeUpdate', 'UsageAfterUpdate'] as const
+
+export const TRIGGER_TYPE_FIELDS = [
+  ...ON_DELETE_TRIGGER_TYPES,
+  ...ON_INSERT_TRIGGER_TYPES,
+  ...ON_UPDATE_TRIGGER_TYPES,
+  'UsageAfterUndelete',
+] as const
+
```
amitgoldarb, line 53: I think this is slightly more readable
```suggestion
enum TriggerType {
  UsageBeforeDelete = 'UsageBeforeDelete',
  UsageAfterDelete = 'UsageAfterDelete',
  UsageBeforeInsert = 'UsageBeforeInsert',
  UsageAfterInsert = 'UsageAfterInsert',
  UsageBeforeUpdate = 'UsageBeforeUpdate',
  UsageAfterUpdate = 'UsageAfterUpdate',
  UsageAfterUndelete = 'UsageAfterUndelete',
}

const TRIGGER_TYPE_FIELDS = Object.values(TriggerType);
```

amitgoldarb (commented):
Looks good. Added just some more small comments

packages/salesforce-adapter/src/filters/extend_triggers_metadata.ts
tamtamirr: These values should never be modified manually by users so no point in setting the restriction.

packages/salesforce-adapter/src/filters/extended_triggers_metadata.ts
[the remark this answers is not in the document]
amitgoldarb: We have lower layers for throttling, so it shouldn't concern this. We can have this as 1K ids that's fine. I think we should allow this to be configured though via the limit section if we ever want to play around with this

amitgoldarb (approved):
LGTM!
Thanks for this

PR #6583, opened 2024-09-18
Improve dangling static files removal

dantal4:
this is an improvement for fetches with mainly additions (like first fetch), or when no static file was removed - we can then avoid going over all of the changes' after and return an empty result.

---

---
_Release Notes_: 
Workspace:
- Improve dangling static files removal

---
_User Notifications_: 
None


adfineberg (approved):
Looks good!

PR #6584, opened 2024-09-18
SALTO-6687: Omit ORN field from Okta applications

shir-reifenberg:
Okta added ORN (=Okta resource name) to all apps

---

deploy works without this field so it should be safe to remove

---
_Release Notes_: 
None

---
_User Notifications_: 

_Okta_adapter_:
- `orn` field will be deleted from all applications


idoamit8 (approved):
💯

shoshana-houri (approved):

PR #6585, opened 2024-09-18
SALTO-6688: (Bugfix) Change-Based fetch deletes sub-instances that their type is not included explicitly in the adapter config

tamtamirr:
(Bugfix) Change-Based fetch deletes sub-instances that their type is not included explicitly in the adapter config
 
---

Made the implementation of `metadataQuery.isTypeMatch` more correct and take into consideration the relationship between the nested instances and their parent.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/test/fetch_profile/metadata_query.test.ts
+++ b/packages/salesforce-adapter/test/fetch_profile/metadata_query.test.ts
@@ -398,6 +398,42 @@ describe('buildMetadataQuery', () => {
     })
   })
 
+  describe('isTypeMatch on nested types', () => {
+    let metadataQuery: MetadataQuery
+    describe('when parent type is included', () => {
+      beforeEach(() => {
+        metadataQuery = buildMetadataQuery({
+          fetchParams: {
+            metadata: {
+              include: [{ metadataType: CUSTOM_OBJECT }],
+            },
+          },
+        })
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/test/fetch_profile/metadata_query.test.ts
```diff
@@ -398,6 +398,72 @@ describe('buildMetadataQuery', () => {
...
+        metadataQuery = buildMetadataQuery({
+          fetchParams: {
+            metadata: {
+              exclude: [{ metadataType: CUSTOM_OBJECT }],
+            },
+          },
+        })
+      })
+      it('should return false for nested types', () => {
+        expect(metadataQuery.isTypeMatch('ValidationRule')).toBeFalse()
```
amitgoldarb, line 430: I find this behaviour confusing. Was it always this way? The only way to include ValidationRule is if I include CustomObject as well? I don't find this to be an intuitive interface.
Was it always like that?

tamtamirr: Yes. This is the current behavior actually, that's why I've added these UT cases 🙏

amitgoldarb (approved):

PR #6587, opened 2024-09-18
SALTO-6522 Fix serializer for undefined

dantal4:
`JSON.stringify(undefined)` returns `undefined` and not `string`, which cause us to throw when accessing `serializedItem.length`

---

---
_Release Notes_: 
Lowerdash:
- Fix serializer for undefined

---
_User Notifications_: 
None


shoshana-houri (approved):
LGTM, thanks!

PR #6592, opened 2024-09-19
SALTO-6671: Add dependency between each Policy to prior

idoamit8:
In this PR I added dependency changer between each policy\policyRule to its prior policy. 

---

---
_Release Notes_: 
_Okta Adapter:_
* Added a dependency changer between each policy\policyRule to its prior policy\policyRule in order to  prevent race conditions that could result in policies being assigned the same priority.
---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/dependency_changers/index.ts
+++ b/packages/okta-adapter/src/dependency_changers/index.ts
@@ -14,6 +14,7 @@ import { removeProfileMappingAfterDeps } from './remove_profile_mapping_after_de
 import { changeDependenciesFromPoliciesAndRulesToPriority } from './policy_and_rules_to_priority'
 import { defaultMultifactorEnrollmentPolicyDependency } from './default_multi_factor_enrollment_policy'
 import { addAuthenticatorToMfaPolicyDependency } from './authenticator_to_mfa_policy'
+import { addDependenciesFromPolicyToPriorPolicy } from './policy_to_prior_policy'
 
 const { awu } = collections.asynciterable
 
@@ -25,6 +26,7 @@ const DEPENDENCY_CHANGERS: DependencyChanger[] = [
   changeDependenciesFromPoliciesAndRulesToPriority,
   defaultMultifactorEnrollmentPolicyDependency,
   addAuthenticatorToMfaPolicyDependency,
+  addDependenciesFromPolicyToPriorPolicy,
 ]
 
 export const dependencyChanger: DependencyChanger = async (changes, deps) =>
--- a/packages/okta-adapter/src/dependency_changers/index.ts
+++ b/packages/okta-adapter/src/dependency_changers/index.ts
@@ -14,7 +14,7 @@ import { removeProfileMappingAfterDeps } from './remove_profile_mapping_after_de
 import { changeDependenciesFromPoliciesAndRulesToPriority } from './policy_and_rules_to_priority'
 import { defaultMultifactorEnrollmentPolicyDependency } from './default_multi_factor_enrollment_policy'
 import { addAuthenticatorToMfaPolicyDependency } from './authenticator_to_mfa_policy'
-import { addDependenciesFromPolicyToPriorPolicy } from './policy_to_prior_policy'
+import { addDependenciesFromPolicyToPriorPolicy } from './order_policies_by_priority'
 
 const { awu } = collections.asynciterable
 
--- a/packages/okta-adapter/src/dependency_changers/order_policies_by_priority.ts
+++ b/packages/okta-adapter/src/dependency_changers/order_policies_by_priority.ts
@@ -20,7 +20,7 @@ import {
   isReferenceExpression,
 } from '@salto-io/adapter-api'
 import { deployment } from '@salto-io/adapter-components'
-import { getParent } from '@salto-io/adapter-utils'
+import { getParent, hasValidParent } from '@salto-io/adapter-utils'
 import { logger } from '@salto-io/logging'
 import { POLICY_PRIORITY_TYPE_NAMES, POLICY_RULE_PRIORITY_TYPE_NAMES } from '../constants'
 import { ALL_SUPPORTED_POLICY_NAMES, POLICY_RULE_TYPES_WITH_PRIORITY_INSTANCE } from '../filters/policy_priority'
@@ -31,38 +31,17 @@ type ChangeWithKey = deployment.dependency.ChangeWithKey<Change<InstanceElement>
 const createDependencyChange = (prev: ChangeWithKey, current: ChangeWithKey): DependencyChange[] => [
   dependencyChange('add', current.key, prev.key),
 ]
-
 const groupChangesByType = (changes: ChangeWithKey[]): Record<string, ChangeWithKey[]> =>
-  changes.reduce(
-    (acc, change) => {
-      const instance = getChangeData(change.change)
-      const { typeName } = instance.elemID
-      if (!acc[typeName]) {
-        acc[typeName] = []
-      }
-      acc[typeName].push(change)
-      return acc
-    },
-    {} as Record<string, ChangeWithKey[]>,
-  )
+  _.groupBy(changes, change => getChangeData(change.change).elemID.typeName)
 
 const groupRuleChangesByPolicy = (changes: ChangeWithKey[]): Record<string, ChangeWithKey[]> =>
-  changes.reduce(
-    (acc, change) => {
+  _.groupBy(
+    changes.filter(change => hasValidParent(getChangeData(change.change))),
+    change => {
       const instance = getChangeData(change.change)
-      try {
-        const parent = getParent(instance)
-        const parentFullName = parent.elemID.getFullName()
-        if (!acc[parentFullName]) {
-          acc[parentFullName] = []
-        }
-        acc[parentFullName].push(change)
-      } catch (e) {
-        log.error('Failed to get parent for %s', instance.elemID.getFullName())
-      }
-      return acc
+      const parent = getParent(instance)
+      return parent.elemID.getFullName()
     },
-    {} as Record<string, ChangeWithKey[]>,
   )
 
 const getGroupedPoliciesAndRules = (
@@ -83,9 +62,19 @@ const sortByPriority = (group: ChangeWithKey[], priorityMap: Record<string, numb
     return priority1 - priority2
   })
 
+const getPriorityMap = (priorityNameToPoliciesOrRuleNames: Record<string, string[]>): Record<string, number> =>
+  Object.fromEntries(
+    Object.entries(priorityNameToPoliciesOrRuleNames).flatMap(([priorityFullName, policyNames]) =>
+      policyNames.map(policyName => [
+        policyName,
+        priorityNameToPoliciesOrRuleNames[priorityFullName]?.indexOf(policyName) ?? Infinity,
+      ]),
+    ),
+  )
+
 /*
  * This dependency handler ensures that dependencies are added between each policy or policyRule during addition changes.
- * The dependency is added based on the priority change and if its none then by order by name
+ * The dependency is added based on the priority change and if it's none then by the order of the changes.
  * This prevents race conditions that could result in policies being assigned the same priority.
  */
 export const addDependenciesFromPolicyToPriorPolicy: DependencyChanger = async changes => {
@@ -129,15 +118,7 @@ export const addDependenciesFromPolicyToPriorPolicy: DependencyChanger = async c
     }),
   )
 
-  // Map each policy or rule name to its priority position
-  const policyNameToPriority = Object.fromEntries(
-    Object.entries(priorityNameToPoliciesOrRuleNames).flatMap(([priorityFullName, policyNames]) =>
-      policyNames.map(policyName => [
-        policyName,
-        priorityNameToPoliciesOrRuleNames[priorityFullName]?.indexOf(policyName) ?? Infinity,
-      ]),
-    ),
-  )
+  const policyNameToPriority = getPriorityMap(priorityNameToPoliciesOrRuleNames)
 
   const { policies, rules } = getGroupedPoliciesAndRules(additionPolicyChanges, additionRuleChanges)
 
--- a/packages/okta-adapter/src/dependency_changers/order_policies_by_priority.ts
+++ b/packages/okta-adapter/src/dependency_changers/order_policies_by_priority.ts
@@ -6,6 +6,7 @@
  * CERTAIN THIRD PARTY SOFTWARE MAY BE CONTAINED IN PORTIONS OF THE SOFTWARE. See NOTICE FILE AT https://github.com/salto-io/salto/blob/main/NOTICES
  */
 
+import _ from 'lodash'
 import {
   Change,
   DependencyChange,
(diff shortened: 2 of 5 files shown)


packages/okta-adapter/src/dependency_changers/policy_to_prior_policy.ts
```diff
@@ -0,0 +1,114 @@
...
+          acc[typeName] = []
+        }
+
+        acc[typeName].push(policyChange)
+        return acc
+      },
+      {} as Record<string, typeof additionPoliciesAndRulesChanges>,
+    )
+
+  const policiesGroups = Object.values(typeNameToPolicyChanges).concat(Object.values(priorityFullNameToPoliciesChanges))
```
idoamit8, line 104: @shir-reifenberg Currently, I haven't sorted the policies in typeNameToPolicyChanges by name or any specific criteria, as I don't think it's necessary. What do you think?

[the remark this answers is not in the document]
shir-reifenberg: also, the deploy order should be 
1. deploy all policies that are included in priority instance by their order
2. deploy all other policy changes

packages/okta-adapter/test/dependency_changers/policy_to_prior_policy.test.ts
```diff
@@ -0,0 +1,167 @@
...
+      const inputDeps = new Map<collections.set.SetId, Set<collections.set.SetId>>([])
+      dependencyChanges = [...(await addDependenciesFromPolicyToPriorPolicy(inputChanges, inputDeps))]
+      expect(dependencyChanges).toHaveLength(1)
+      expect(dependencyChanges[0].action).toEqual('add')
+      expect(dependencyChanges[0].dependency.source).toEqual(1)
+      expect(dependencyChanges[0].dependency.target).toEqual(0)
+    },
+  )
+  it.each([...POLICY_RULE_TYPES_WITH_PRIORITY_INSTANCE, ...ALL_SUPPORTED_POLICY_NAMES])(
+    "should add dependencies between each %s addition change when they don't have priority change",
```
shir-reifenberg, line 110: can we verify we don't add dependency between rules of different policies?

shir-reifenberg (commented):
Great work! some comments / suggestions -
1. I think we should address the use case you mentioned (good catch!), and I think that taking a slightly different approach would handle this with not too big of an effort - 
For each policy type ->
- find it's matching priority instance
- order all changes by: 1. their order in the priority instance 2. their name (you can do it with `_.sortBy` and provide a custom sort function
Then do the same for policy rule, (but by their parent and not by type)
WDYT?

2. IIUC there might be a bug here with `defaultMultifactorEnrollmentPolicyDependency` that can cause circular dependency. `defaultMultifactorEnrollmentPolicyDependency` add dependency from custom policies to the default one, then this dependency changer takes all policies (including the default) and add a dependency to policy in the order instance.

4. in general I think we might not want to include the default policy, or have it always be added last? (if it's priority isn't handled in all policy / rules types, I think we can just exclude it from this dependency changer)

packages/okta-adapter/src/dependency_changers/policy_to_prior_policy.ts
```diff
@@ -0,0 +1,114 @@
...
+  if (_.isEmpty(additionPoliciesAndRulesChanges)) {
+    return []
+  }
+
+  const policyNameToChange = Object.fromEntries(
+    additionPoliciesAndRulesChanges.map(policyChange => {
+      const instance = getChangeData(policyChange.change)
+      return [instance.elemID.getFullName(), policyChange]
+    }),
+  )
```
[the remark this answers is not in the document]
idoamit8: Not sure I understand this

packages/okta-adapter/src/dependency_changers/policy_to_prior_policy.ts
```diff
@@ -0,0 +1,114 @@
...
+      return [priorityInstance.elemID.getFullName(), policies]
+    }),
+  )
+
+  const policiesWithPriorityChange = new Set(Object.values(priorityFullNameToPoliciesChanges).flat())
+
+  // Organize changes by typeName excluding those with priority changes
+  const typeNameToPolicyChanges = additionPoliciesAndRulesChanges
+    .filter(policyChange => !policiesWithPriorityChange.has(policyChange))
+    .reduce(
```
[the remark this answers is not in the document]
idoamit8: why?
I can go with this code. Do you think its better?
```
  const typeNamesWithPriorityChanges: Record<string, typeof additionPoliciesAndRulesChanges> = {}
  additionPoliciesAndRulesChanges.forEach(policyChange => {
    const { typeName } = getChangeData(policyChange.change).elemID
    if (!typeNamesWithPriorityChanges[typeName]) {
      typeNamesWithPriorityChanges[typeName] = []
    }
    typeNamesWithPriorityChanges[typeName].push(policyChange)
  })
```

packages/okta-adapter/src/dependency_changers/index.ts
```diff
@@ -25,6 +26,7 @@ const DEPENDENCY_CHANGERS: DependencyChanger[] = [
   changeDependenciesFromPoliciesAndRulesToPriority,
   defaultMultifactorEnrollmentPolicyDependency,
   addAuthenticatorToMfaPolicyDependency,
+  addDependenciesFromPolicyToPriorPolicy,
```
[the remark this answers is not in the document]
shir-reifenberg: maybe align with the file name?

packages/okta-adapter/src/dependency_changers/order_policies_by_priority.ts
```diff
@@ -0,0 +1,136 @@
...
+export const addDependenciesFromPolicyToPriorPolicy: DependencyChanger = async changes => {
+  const instanceChanges = Array.from(changes.entries())
+    .map(([key, change]) => ({ key, change }))
+    .filter((change): change is deployment.dependency.ChangeWithKey<Change<InstanceElement>> =>
+      isInstanceChange(change.change),
+    )
+
+  const additionPoliciesAndRulesChanges = instanceChanges.filter(
+    change =>
+      new Set([...ALL_SUPPORTED_POLICY_NAMES, ...POLICY_RULE_TYPES_WITH_PRIORITY_INSTANCE]).has(
```
shir-reifenberg, line 91: please create the `Set` once and not for each change
(same for line 97)

packages/okta-adapter/src/dependency_changers/order_policies_by_priority.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  const policyNameToChange = Object.fromEntries(
+    additionPoliciesAndRulesChanges.map(policyChange => {
+      const instance = getChangeData(policyChange.change)
+      return [instance.elemID.getFullName(), policyChange]
+    }),
+  )
+
+  // Map priority full names to related policy changes
+  const priorityFullNameToPoliciesChanges = Object.fromEntries(
```
shir-reifenberg, line 114:
```suggestion
  const priorityFullNameToPolicyChanges = Object.fromEntries(
```

packages/okta-adapter/src/dependency_changers/order_policies_by_priority.ts
```diff
@@ -0,0 +1,136 @@
...
+const createDependencyChange = (
+  prevPolicyOrPolicyRule: deployment.dependency.ChangeWithKey<Change<InstanceElement>>,
+  policyOrPolicyRule: deployment.dependency.ChangeWithKey<Change<InstanceElement>>,
+): DependencyChange[] => [dependencyChange('add', policyOrPolicyRule.key, prevPolicyOrPolicyRule.key)]
+
+/*
+ * This dependency handler ensures that dependencies are added between each policy or policyRule during addition changes.
+ * The dependency is added based on the priority change and if its none than by order by name
+ * This prevents race conditions that could result in policies being assigned the same priority.
+ */
```
shir-reifenberg, line 41: please put the comment above `addDependenciesFromPolicyToPriorPolicy`
also `than` -> `then`

packages/okta-adapter/src/dependency_changers/order_policies_by_priority.ts
```diff
@@ -0,0 +1,136 @@
...
+  additionPoliciesAndRulesChanges: deployment.dependency.ChangeWithKey<Change<InstanceElement>>[],
+  priorityFullNameToPoliciesChanges: Record<string, deployment.dependency.ChangeWithKey<Change<InstanceElement>>[]>,
+): deployment.dependency.ChangeWithKey<Change<InstanceElement>>[][] => {
+  const policiesWithPriorityChange = new Set(
+    Object.values(priorityFullNameToPoliciesChanges)
+      .flat()
+      .map(policyChange => policyChange.key),
+  )
+
+  const typeNameToPolicyChanges: Record<string, typeof additionPoliciesAndRulesChanges> = {}
```
shir-reifenberg, line 53: I think we shouldn't use `typeof` here

packages/okta-adapter/src/dependency_changers/order_policies_by_priority.ts
```diff
@@ -0,0 +1,136 @@
...
+  policyOrPolicyRule: deployment.dependency.ChangeWithKey<Change<InstanceElement>>,
+): DependencyChange[] => [dependencyChange('add', policyOrPolicyRule.key, prevPolicyOrPolicyRule.key)]
+
+/*
+ * This dependency handler ensures that dependencies are added between each policy or policyRule during addition changes.
+ * The dependency is added based on the priority change and if its none than by order by name
+ * This prevents race conditions that could result in policies being assigned the same priority.
+ */
+
+const getGroupedPoliciesAndRules = (
```
shir-reifenberg, line 43: I think this name hides a very important thing this function is doing - it's grouping policies with the same "parent" (either a real parent + a type), but it also orders the policies by their priority?

I think this also indicates a wider issue, that we mix the grouping and the ordering within the same logic and it makes it more difficult to read IMO
It would be clearer to have 2 functions here - one for the grouping, and another one for the ordering

packages/okta-adapter/src/dependency_changers/order_policies_by_priority.ts
```diff
@@ -0,0 +1,136 @@
...
+    Object.values(priorityFullNameToPoliciesChanges)
+      .flat()
+      .map(policyChange => policyChange.key),
+  )
+
+  const typeNameToPolicyChanges: Record<string, typeof additionPoliciesAndRulesChanges> = {}
+  const policyToPolicyRulesChanges: Record<string, typeof additionPoliciesAndRulesChanges> = {}
+  additionPoliciesAndRulesChanges
+    .filter(policyChange => !policiesWithPriorityChange.has(policyChange.key))
+    .forEach(policyOrRuleChange => {
```
shir-reifenberg, line 57: in general we prefer function style, and I think you can create it by using `_.groupBy` with something like - 

```
const changesGroupedByTypeOrParent = _.groupBy(
  additionPoliciesAndRulesChanges,
  change => {
     if (isPolicyRule) {
        return getParent.elemID.getFullName
     } else { // policy
       return change.elemID.typeName
     }
  }
)
```

packages/okta-adapter/src/dependency_changers/order_policies_by_priority.ts
```diff
@@ -0,0 +1,136 @@
...
+  const policiesWithPriorityChange = new Set(
+    Object.values(priorityFullNameToPoliciesChanges)
+      .flat()
+      .map(policyChange => policyChange.key),
+  )
+
+  const typeNameToPolicyChanges: Record<string, typeof additionPoliciesAndRulesChanges> = {}
+  const policyToPolicyRulesChanges: Record<string, typeof additionPoliciesAndRulesChanges> = {}
+  additionPoliciesAndRulesChanges
+    .filter(policyChange => !policiesWithPriorityChange.has(policyChange.key))
```
shir-reifenberg, line 56: maybe I'm missing something, but why do we have this filter?
isn't this means we might put policy rule from the same "parent" in different groups? and then create no dependency between them ?

packages/okta-adapter/src/dependency_changers/order_policies_by_priority.ts
shir-reifenberg: another suggestion - do we really need the changes themselves here? isn't it enough to have the the mapping from the priority instance to the full name of its policies / rules?

packages/okta-adapter/src/dependency_changers/order_policies_by_priority.ts
```diff
@@ -0,0 +1,136 @@
...
+        .filter(isDefined)
+      return [priorityInstance.elemID.getFullName(), policies]
+    }),
+  )
+
+  const policiesGroups = getGroupedPoliciesAndRules(additionPoliciesAndRulesChanges, priorityFullNameToPoliciesChanges)
+
+  return policiesGroups
+    .flatMap(group =>
+      group.slice(1).map((currentPolicyChange, index) => {
```
shir-reifenberg, line 130: nit: I think it would be useful to have a debug log indicating that we're going to add dependencies between the following policies with their full names (already ordered)

packages/okta-adapter/test/dependency_changers/order_policies_by_priority.test.ts
```diff
@@ -0,0 +1,259 @@
...
+  ReferenceExpression,
+  CORE_ANNOTATIONS,
+} from '@salto-io/adapter-api'
+import { collections } from '@salto-io/lowerdash'
+import { OKTA } from '../../src/constants'
+import { changeDependenciesFromPoliciesAndRulesToPriority } from '../../src/dependency_changers/policy_and_rules_to_priority'
+import { ALL_SUPPORTED_POLICY_NAMES, POLICY_RULE_TYPES_WITH_PRIORITY_INSTANCE } from '../../src/filters/policy_priority'
+import { addDependenciesFromPolicyToPriorPolicy } from '../../src/dependency_changers/order_policies_by_priority'
+
+describe('addDependenciesFromPolicyToPriorPolicy', () => {
```
shir-reifenberg, line 24: can we please add tests for the scenarios - 
1. policies / rules from different types are added and not have dependencies between them (e.g addition ACCESS_POLICY_RULE and PASSWORD RULE should have dependency between them)
2. that we create dependency even if the order instance is missing / not pointing the policies
(it's ok not to add it generic and just sample a certain type and test for it)

shir-reifenberg (commented):
Great job!! I added a few more comments but I might be missing something here
LMK if you'd like to discuss IRL

packages/okta-adapter/src/dependency_changers/policy_to_prior_policy.ts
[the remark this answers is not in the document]
shir-reifenberg: yeah is doesn't work as I wrote it
it should work like

```
  const policyNameToChange =   _.keyBy(
    additionPoliciesAndRulesChanges
    change => getChangeData(change).elemID.getFullName()
  )
```

packages/okta-adapter/src/dependency_changers/order_policies_by_priority.ts
idoamit8: Its not ordering, its just grouping. Why you think its ordering? 
It gets the ordered instances as a parameter and groups it with the one that doesn't have priority.

packages/okta-adapter/test/dependency_changers/order_policies_by_priority.test.ts
idoamit8: 1. Why do want to have dependencies between them?
2. I think I already have this

shir-reifenberg: 1. realized it wasn't clear enough :) I meant a test checking we **don't** add those dependencies

packages/okta-adapter/src/dependency_changers/order_policies_by_priority.ts
```diff
@@ -0,0 +1,160 @@
...
+    )
+    .filter(change => getChangeData(change.change).value.system !== true)
+
+  const additionRuleChanges = instanceChanges
+    .filter(
+      change =>
+        POLICY_RULE_TYPES_WITH_PRIORITY_INSTANCE.includes(getChangeData(change.change).elemID.typeName) &&
+        isAdditionChange(change.change),
+    )
+    .filter(change => getChangeData(change.change).value.system !== true)
```
shir-reifenberg, line 110: I realized some policy rules doesn't have the `default` field which means it can be ordered and therefore included in here (for example `MultifactorEnrollmentPolicyRulePriority` and `AuthorizationServerPolicyRulePriority`)
There should be some piece of code to determine in which policy types we should include the default and in which we shouldn't 
(the same code that decides in there is a `default` field in the priority instance

packages/okta-adapter/src/dependency_changers/order_policies_by_priority.ts
```diff
@@ -0,0 +1,160 @@
...
+
+  // Map each policy or rule name to its priority position
+  const policyNameToPriority = Object.fromEntries(
+    Object.entries(priorityNameToPoliciesOrRuleNames).flatMap(([priorityFullName, policyNames]) =>
+      policyNames.map(policyName => [
+        policyName,
+        priorityNameToPoliciesOrRuleNames[priorityFullName]?.indexOf(policyName) ?? Infinity,
+      ]),
+    ),
+  )
```
shir-reifenberg, line 140: please extract to function (then the comment can be removed)

packages/okta-adapter/src/dependency_changers/order_policies_by_priority.ts
```diff
@@ -0,0 +1,160 @@
...
+    (acc, change) => {
+      const instance = getChangeData(change.change)
+      const { typeName } = instance.elemID
+      if (!acc[typeName]) {
+        acc[typeName] = []
+      }
+      acc[typeName].push(change)
+      return acc
+    },
+    {} as Record<string, ChangeWithKey[]>,
```
shir-reifenberg, line 46: another reason I don't like reduce is the cast

packages/okta-adapter/src/dependency_changers/order_policies_by_priority.ts
```diff
@@ -0,0 +1,160 @@
...
+  )
```
shir-reifenberg, line 47: just because we had this discussion - this IMO is a case where there is not a real reason to use `reduce` and `_.groupBy` would have been clearer
(same applies for the other `reduce`, though you can first filter for changes with parent using `hasValidParent`)

WDYT ?
```suggestion
const groupChangesByType = (changes: ChangeWithKey[]): Record<string, ChangeWithKey[]> =>
  _.groupBy(
    changes,
    change => getChangeData(change.change).elemID.typeName
  )
```

packages/okta-adapter/src/dependency_changers/order_policies_by_priority.ts
```diff
@@ -0,0 +1,160 @@
...
+  policies: Object.values(groupChangesByType(additionPolicyChanges)),
+  rules: Object.values(groupRuleChangesByPolicy(additionRuleChanges)),
+})
+
+const sortByPriority = (group: ChangeWithKey[], priorityMap: Record<string, number>): ChangeWithKey[] =>
+  group.sort((change1, change2) => {
+    const priority1 = priorityMap[getChangeData(change1.change).elemID.getFullName()] ?? Infinity
+    const priority2 = priorityMap[getChangeData(change2.change).elemID.getFullName()] ?? Infinity
+    return priority1 - priority2
+  })
```
shir-reifenberg, line 84: I think this can be replaced with - 

```suggestion
const sortByPriority = (group: ChangeWithKey[], priorityMap: Record<string, number>): ChangeWithKey[] =>
  _.sortBy(
    group,
    change => priorityMap[getChangeData(change1.change).elemID.getFullName()] ?? Infinity
  )
```

packages/okta-adapter/src/dependency_changers/order_policies_by_priority.ts
```diff
@@ -0,0 +1,160 @@
...
+
+/*
+ * This dependency handler ensures that dependencies are added between each policy or policyRule during addition changes.
+ * The dependency is added based on the priority change and if its none then by order by name
```
shir-reifenberg, line 88: the by name is not accurate now, right? cause we default to just being last and don't care about the order
also I think it should be `it's`

shir-reifenberg (approved):
Looks great! This is much more correct & readable IMO
I added one question I thought about, and some suggestions for better readability
LMK what you think about those but I'm pre approving 💪 👑

packages/okta-adapter/src/dependency_changers/order_policies_by_priority.ts
idoamit8: @shir-reifenberg  But for those policies there isn't any change that his `system` field is true? This is the reason I used `!== true` and not `=== false`

PR #6595, opened 2024-09-19
SALTO-6056 Add logs for new SuiteScript references logic

dantal4:
Add logs for new SuiteScript references logic, and avoid running the calculation (regex) for files with invalid extension.

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


omrilit (approved):
Thanks!

PR #6596, opened 2024-09-19
SALTO-6671: Avoid overriding policy settings when updating policy priority

shir-reifenberg:
When updating policy / rules priorities, by sending partial data using `PUT`, the policy settings were overridden with Okta's default.

---

There is no `PATCH` request to update policy priority, therefore, in order to update it's priority we send an additional request in `policyPrioritiesFilter` to update this field. The issue was, we sent partial data, and in some cases it cause Okta to fail the request (in MFA enrollment policies), and in others to "reset" the policy settings into the default settings.

To avoid this, I adjusted how we make the policy update - instead of sending specific fields, we'll make a `GET` request to get the current policy data, and the merge of the current data + the required priority (and omit any non deployable fields).

due to the references and `changeDependenciesFromPoliciesAndRulesToPriority`, guarantee no race condition will occur (i.e update to policy data in between the `GET` and `POST` to update the priority).

On policy addition cases, there were some scenarios when I got 404 in the `GET` after addition, so added polling to tackle this.

---
_Release Notes_: 

_Okta adapter_:
- Avoid reseting policy settings when updating priority updates.

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/filters/policy_priority.ts
+++ b/packages/okta-adapter/src/filters/policy_priority.ts
@@ -22,8 +22,13 @@ import {
 } from '@salto-io/adapter-api'
 import _ from 'lodash'
 import { collections } from '@salto-io/lowerdash'
-import { getParents, invertNaclCase, naclCase, pathNaclCase } from '@salto-io/adapter-utils'
-import { fetch, elements as adapterElements, client as clientUtils } from '@salto-io/adapter-components'
+import { getParents, invertNaclCase, naclCase, pathNaclCase, safeJsonStringify } from '@salto-io/adapter-utils'
+import {
+  fetch,
+  elements as adapterElements,
+  client as clientUtils,
+  definitions as definitionsUtils,
+} from '@salto-io/adapter-components'
 import { logger } from '@salto-io/logging'
 import { FilterCreator } from '../filter'
 import {
@@ -50,11 +55,10 @@ import {
   SIGN_ON_RULE_TYPE_NAME,
 } from '../constants'
 import { deployChanges } from '../deprecated_deployment'
-import { API_DEFINITIONS_CONFIG, OktaSwaggerApiConfig } from '../config'
 
 const log = logger(module)
 const { awu } = collections.asynciterable
-const { createUrl } = fetch.resource
+const { replaceAllArgs } = fetch.request
 export const POLICY_RULE_TYPES_WITH_PRIORITY_INSTANCE = [
   ACCESS_POLICY_RULE_TYPE_NAME,
   IDP_RULE_TYPE_NAME,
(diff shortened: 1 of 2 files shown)


packages/okta-adapter/src/filters/policy_priority.ts
```diff
@@ -168,49 +168,86 @@ const getParentPolicy = (rule: InstanceElement): InstanceElement | undefined =>
...
-  // SignOn and MultiFactorEnrollment rules must include the `actions` field in the payload
-  const data = [SIGN_ON_RULE_TYPE_NAME, MFA_RULE_TYPE_NAME].includes(instance.elemID.typeName)
-    ? { ...baseData, actions: instance.value.actions }
-    : baseData
-  const typeDefinition = apiDefinitions.types[instance.elemID.typeName]
-  const deployRequest = typeDefinition.deployRequests ? typeDefinition.deployRequests.modify : undefined
-  const deployUrl = deployRequest?.url
-  if (deployUrl === undefined) {
-    throw new Error('Failed to deploy priority change due to missing url')
+  const pathContext = _.merge({}, instance.value, { parent_id: getParentPolicy(instance)?.value?.id })
```
idoamit8, line 207: nit
```suggestion
const pathContext = { ...instance.value, parent_id: getParentPolicy(instance)?.value?.id }
```
I think that Using the spread operator { ...instance.value, parent_id: parentPolicyId } is better than _.merge({}, instance.value, { parent_id: ... }) because it simplifies the operation, avoids unnecessary deep merging, and guarantees immutability without needing an extra empty object.

packages/okta-adapter/src/filters/policy_priority.ts
```diff
@@ -168,49 +168,86 @@ const getParentPolicy = (rule: InstanceElement): InstanceElement | undefined =>
...
+    retryOnStatus: [404],
+  }
+  const clientGet = async ({
+    url,
+  }: clientUtils.ClientBaseParams): Promise<
+    clientUtils.Response<clientUtils.ResponseValue | clientUtils.ResponseValue[]>
+  > => client.get({ url })
+  const res = await clientUtils.executeWithPolling(args, pollingArgs, clientGet)
+  return res
+}
```
idoamit8, line 192: This function appears to be quite generic and could be useful in more places in the future. Consider adding `pollingArgs` as a parameter so it can be reused where applicable. (Can also remove it to a better location for that matter)

idoamit8 (approved):
Great work!! 💯 
Added two small nits  🕵🏻‍♂️
Approving ✅

packages/okta-adapter/src/filters/policy_priority.ts
shir-reifenberg: There is a generic implementation for it and I'm using it here ('executeWithPolling')😃 
The outer function just sets the params for this specific use case

PR #6597, opened 2024-09-19
SALTO-6433: (Intune) Support fetch & deploy for IOS Application Protection policies

shoshana-houri:
Support fetching & deploying Intune IOS Application Protection policies - including target apps & assignments.

---
This is the 2nd out of 4 types of Intune Application Protection policies. Each type has different endpoints and structures, so the PRs are divided accordingly.

See .nacls structure example [here](https://github.com/salto-io/microsoft-security/tree/main/microsoft_security/Records/IntuneApplicationProtectionIOS).

---
_Release Notes_: 
_Microsoft Security adapter:_
* Support fetch & deploy for Intune IOS Application Protection policies.

---
_User Notifications_: 
_Microsoft Security adapter:_
* Support fetch & deploy for Intune IOS Application Protection policies.


packages/microsoft-security-adapter/src/definitions/deploy/intune/utils/target_apps.ts
```diff
@@ -6,43 +6,17 @@
...
-      // TODO SALTO-6528: Warn the user when this occurs.
-      if (isRemovalChange(change) || (targetedAppType && targetedAppType !== 'selectedPublicApps')) {
-        return false
-      }
-
-      return isAdditionChange(change)
-        ? !_.isEmpty(changeData.value.apps)
-        : change.data.before.value.apps !== change.data.after.value.apps
-    },
-})
```
shoshana-houri: When further investigating how this update is handled in the Intune admin center, I found that we do need to call the targetApps endpoint, but we should also include the `targetedAppType` in the request.

PR #6603, opened 2024-09-22
SALTO-6672: Add policy priority E2E

shir-reifenberg:
Add policy priority E2E for policy and policy rules

---

added coverage for password policy and access policy for now, will extend later

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/e2e_test/adapter.test.ts
+++ b/packages/okta-adapter/e2e_test/adapter.test.ts
@@ -33,6 +33,7 @@ import {
   detailedCompare,
   getParents,
   inspectValue,
+  invertNaclCase,
   naclCase,
   safeJsonStringify,
 } from '@salto-io/adapter-utils'
@@ -58,6 +59,9 @@ import {
   INACTIVE_STATUS,
   NETWORK_ZONE_TYPE_NAME,
   ORG_SETTING_TYPE_NAME,
+  PASSWORD_POLICY_PRIORITY_TYPE_NAME,
+  PASSWORD_POLICY_TYPE_NAME,
+  PASSWORD_RULE_TYPE_NAME,
   PROFILE_ENROLLMENT_POLICY_TYPE_NAME,
   PROFILE_ENROLLMENT_RULE_TYPE_NAME,
   ROLE_TYPE_NAME,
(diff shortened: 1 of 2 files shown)


packages/okta-adapter/e2e_test/adapter.test.ts
```diff
@@ -93,10 +97,7 @@ const createInstance = ({
   parent?: InstanceElement
   name?: string
 }): InstanceElement => {
-  const instValues = {
-    ...mockDefaultValues[typeName],
-    ...valuesOverride,
-  }
+  const instValues = _.merge({}, mockDefaultValues[typeName], valuesOverride)
```
idoamit8, line 100: nit: 
Why are you using `_.mrege` and not spread operator? 
you can also use `  Object.assign({}, mockDefaultValues[typeName], valuesOverride)`
I think shallow copy is enough here and we don't need deep copy. 
but its up to you

packages/okta-adapter/e2e_test/adapter.test.ts
```diff
@@ -173,23 +174,175 @@ const createChangesForDeploy = async (types: ObjectType[], testSuffix: string):
...
+              },
+            ],
+          },
+        },
+      },
+    },
+    parent: accessPolicy,
+  })
+  const accessPolicyRulePriority = createInstance({
+    typeName: 'AccessPolicyRulePriority',
```
idoamit8, line 249: Can we use `ACCESS_POLICY_RULE_PRIORITY_TYPE_NAME` const from `constants.ts`?

packages/okta-adapter/e2e_test/adapter.test.ts
```diff
@@ -173,23 +174,175 @@ const createChangesForDeploy = async (types: ObjectType[], testSuffix: string):
...
+  const passwordPolicyRuleB = createInstance({
+    typeName: PASSWORD_RULE_TYPE_NAME,
+    types,
+    valuesOverride: {
+      name: createName('passwordRuleB'),
+    },
+    parent: passwordPolicyA,
+  })
+  const passwordPolicyRulePriority = createInstance({
+    typeName: 'PasswordPolicyRulePriority',
```
idoamit8, line 335: Please use `PASSWORD_RULE_PRIORITY_TYPE_NAME` const

idoamit8 (approved):
Great Work! 💯 
Approving to save cycle ✅
Added some small comments

packages/okta-adapter/e2e_test/adapter.test.ts
shir-reifenberg: I added this cause in some cases I only wanted to pass partial path and I wanted it to actually do the merge
(see line 228 in `defaultRule`)

PR #6605, opened 2024-09-22
SALTO-6696: Increase rocksdb write buffer size when opening a new DB

ori-moisis:
Currently there is a mismatch between opening a new DB with "open" and creating a new DB with "replicate" the "open" code uses a much smaller write buffer size than "replicate". We want the bigger buffer size since we tend to write a lot of information

---

Documentation about the write buffer size option:
https://github.com/facebook/rocksdb/wiki/Setup-Options-and-Basic-Tuning#write-buffer-size

The place in the binding code that sets the different default value:
https://github.com/salto-io/rocksdb/blob/9863eafd245cefa2fa2a70f6b82734b194259457/binding.cc/#L831

---
_Release Notes_: 
Core:
- Increase write buffer size in rocksdb to improve performance

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/core/src/local-workspace/remote_map/remote_map.ts
+++ b/packages/core/src/local-workspace/remote_map/remote_map.ts
@@ -32,6 +32,11 @@ const log = logger(module)
 export const TMP_DB_DIR = 'tmp-dbs'
 export type RocksDBValue = string | Buffer | undefined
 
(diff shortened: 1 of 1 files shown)


dorsev (approved):
LGTM, thank you for this PR! 🥇

PR #6607, opened 2024-09-22
SALTO-6701: Do not omit objectPermissions of standard objects in the Profiles & PS Elements fixer

tamtamirr:
Do not omit objectPermissions of standard objects in the Profiles & PS Elements fixer

---

Some Standard Objects are not in the workspace and are being omitted from the PermissionSet. This is most likely due to the fact they are not managed in the Metadata API. See this for example as a result of running the fixer on Weave’s environment:

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


amitgoldarb (approved):
Thanks for this

PR #6609, opened 2024-09-22
SALTO-6056 add word boundaries to regex

dantal4:
add word boundaries to regex to improve its performance by to preventing partial matches that could lead to more backtracking.

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


Shulik95 (approved):
💯

PR #6612, opened 2024-09-23
SALTO-6707: Allow customizing prefix for element path

shir-reifenberg:
Allow customizing prefix for element path

---

hardcoded path added to the `topLevel`.`path` definition, allowing to consolidate few types under the same path.

example (not a real usecase):
image

will add tests later

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/definitions/system/fetch/element.ts
+++ b/packages/adapter-components/src/definitions/system/fetch/element.ts
@@ -55,6 +55,8 @@ export type ElemIDDefinition<TCustomNameMappingOptions extends string = never> =
 
 export type PathDefinition<TCustomNameMappingOptions extends string = never> = {
   pathParts?: IDPartsDefinition<TCustomNameMappingOptions>[]
+  // allow adding hard coded baseDir for path that will be added before the type name
+  baseDir?: string[]
 }
 
 type StandaloneFieldDefinition = {
--- a/packages/adapter-components/src/fetch/element/id_utils.ts
+++ b/packages/adapter-components/src/fetch/element/id_utils.ts
@@ -207,6 +207,25 @@ export const createElemIDFunc =
     return computedName
   }
 
+const verifyNestedPathIncludesBaseDir = ({
+  nestUnderPath,
+  pathBaseDir,
+  typeName,
+}: {
+  nestUnderPath: string[]
+  pathBaseDir: string[]
+  typeName: string
+}): void => {
+  if (!_.isEqual(pathBaseDir, nestUnderPath.slice(0, pathBaseDir.length))) {
+    log.warn(
+      'detected inconsistency between pathBaseDir and nestUnderPath for type %s, expected base dir to be %o, but received nested path: %o',
+      typeName,
+      pathBaseDir,
+      nestUnderPath,
+    )
+  }
+}
+
 export const getElemPath =
   <TCustomNameMappingOptions extends string = never>({
     def,
@@ -242,12 +261,9 @@ export const getElemPath =
 
     const { adapter: adapterName, typeName } = typeID
     const lastPart = pathParts[pathParts.length - 1]
-    return [
-      adapterName,
-      RECORDS_PATH,
-      ...pathBaseDir,
-      ...(nestUnderPath ?? [pathNaclCase(typeName)]),
-      ...pathParts,
-      ...(createSelfFolder && lastPart ? [lastPart] : []),
-    ]
+    const typePath = nestUnderPath ?? [...pathBaseDir, pathNaclCase(typeName)]
+    if (nestUnderPath !== undefined && !_.isEmpty(pathBaseDir)) {
+      verifyNestedPathIncludesBaseDir({ nestUnderPath, pathBaseDir, typeName })
+    }
+    return [adapterName, RECORDS_PATH, ...typePath, ...pathParts, ...(createSelfFolder && lastPart ? [lastPart] : [])]
   }
--- a/packages/adapter-components/src/fetch/element/id_utils.ts
+++ b/packages/adapter-components/src/fetch/element/id_utils.ts
@@ -226,9 +226,9 @@ export const getElemPath =
     customNameMappingFunctions?: NameMappingFunctionMap<TCustomNameMappingOptions>
   }): PartsCreator =>
   ({ entry, parent, defaultName }) => {
-    const pathPrefix = def?.prefix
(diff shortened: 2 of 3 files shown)


packages/adapter-components/src/definitions/system/fetch/element.ts
```diff
@@ -56,6 +56,8 @@ export type ElemIDDefinition<TCustomNameMappingOptions extends string = never> =
 export type PathDefinition<TCustomNameMappingOptions extends string = never> = {
   // when id parts info is missing, inherited from elemID (but the values are path-nacl-cased)
   pathParts?: IDPartsDefinition<TCustomNameMappingOptions>[] & { extendsParent?: boolean }
+  // allow adding hard coded prefix for path that will be added before the type name
+  prefix?: string
```
shoshana-houri, line 60: What do you think about allowing multiple levels of nesting?
```suggestion
  prefix?: string[]
```

shoshana-houri (approved):
LGTM! Thank you! 🚀
I’m wondering if we should allow providing a prefix with more than one directory. Either way, approving to save cycle.
Also, can we please add a UT that provides a prefix?

netama:
please wait with merging this, commenting on the ticket

packages/adapter-components/src/definitions/system/fetch/element.ts
netama: I think it's better not to use `prefix` here as it's usually used inside a specific string. maybe `baseDir` or `root`?

netama: also why not add this next to `extendsParent`?

packages/adapter-components/src/fetch/element/id_utils.ts
```diff
@@ -226,8 +226,9 @@ export const getElemPath =
     customNameMappingFunctions?: NameMappingFunctionMap<TCustomNameMappingOptions>
   }): PartsCreator =>
   ({ entry, parent, defaultName }) => {
+    const pathPrefix = def?.prefix
     if (singleton) {
-      return [typeID.adapter, RECORDS_PATH, SETTINGS_NESTED_PATH, pathNaclCase(typeID.typeName)]
+      return [typeID.adapter, RECORDS_PATH, ...(pathPrefix ? [pathPrefix] : []), SETTINGS_NESTED_PATH, pathNaclCase(typeID.typeName)]
```
netama, line 231: we should check how this looks in the SAAS, not sure this will work

packages/adapter-components/src/definitions/system/fetch/element.ts
shir-reifenberg: `extendsParent` is on the elemID and not on path ?

shir-reifenberg: my bad, I see it's also defined on the path, but in each element of the array
can you please remind me why `pathParts` is an array of `parts` ?

packages/adapter-components/src/fetch/element/id_utils.ts
shir-reifenberg: I think it's OK (discussed IRL)

packages/adapter-components/src/definitions/system/fetch/element.ts
netama: oops I see I misread the definition of `extendsParent` - so it's the other way around, we should move it to be next to this and *not* on `pathParts`, I see this was adde by @edenhassid  in https://github.com/salto-io/salto/pull/6115 with your review so I'm not sure I fully follow why it was defined next to the array this way, maybe one of you will remember? we don't have to fix it as part of this PR but I don't think it's usable the way it's defined, there's no way to do this in json even if typescript allows it.

the path contains multiple folders (`a/b/c`) and each path part defines a folder, which itself can be built based on parts similarly to the elem id (`c = x_y_z`)

shir-reifenberg: thanks, IIRC I think the use case for  `pathBaseDir` to be under parts is to allow flexibility in each part to extend the parent path and it was done intentionally
I kind of agree this is a bit strange and we might want to change it

anyway I think `baseDir` should be directly under `PathDefinition`

packages/adapter-components/src/fetch/element/id_utils.ts
```diff
@@ -244,6 +245,7 @@ export const getElemPath =
     return [
       adapterName,
       RECORDS_PATH,
+      ...pathBaseDir,
       ...(nestUnderPath ?? [pathNaclCase(typeName)]),
```
netama, line 249: just making sure this also works for nested paths? there's this part
```
    const nestUnderPath = standaloneDef.nestPathUnderParent
      ? [...(parent.path?.slice(2, parent.path?.length - 1) ?? []), field.name]
      : undefined
```
which makes me think we would get the `pathBaseDir` again under the parent type in these cases, can you please add a test for it? (I might be wrong as this would have probably surfaced in the ms case)

packages/adapter-components/src/definitions/system/fetch/element.ts
netama: @edenhassid is there a reason to keep it "next to the array" the way it's defined now? or was that a mistake and we can move it to be next to the new param? :)

packages/adapter-components/src/fetch/element/id_utils.ts
netama: I think the question is whether you want to add `pathBaseDir` as the default for a set of definitions, or manage it type by type :) 
I'd think it's easier to allow the developer to "not worry about it" (some types might sometimes be standalone and sometimes not, e.g. folders defined recursively) - and it also makes sense not to use it when not defining the base dir - so IMO the code behavior should ignore the base dir in these cases (and maybe log a warning in cases where the base dir is different from the final path's base dir)? wdyt?

shir-reifenberg: sounds legit, will change + test
thanks !

packages/adapter-components/src/definitions/system/fetch/element.ts
edenhassid: it wasn't a mistake, I thought at the time that there is no reason to limit this only to the file name, and to add the possibility to support this for folders as well.

PR #6613, opened 2024-09-24
SALTO-6699: Cache whether a remote map is empty

ori-moisis:
Checking if a remote map is empty requires creating a DB iterator with a prefix. If we know that a namespace is empty, we can avoid doing that and hopefully reduce the amount of DB iterators we create. This in turn will hopefully reduce the chances that we trigger the crazy memory consumption issue in rocksdb

---

---
_Release Notes_: 
Core:
- Improve performance when reading many elements with static files for a workspace

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/core/src/local-workspace/remote_map/remote_map.ts
+++ b/packages/core/src/local-workspace/remote_map/remote_map.ts
@@ -637,12 +637,10 @@ export const createRemoteMapCreator = (
                 resolve(undefined)
                 return
               }
-              isNamespaceEmpty = false
               await resolveRet(innerValue)
               statCounters.RemoteMapHit.inc()
             })
           } else {
-            isNamespaceEmpty = false
             await resolveRet(value)
             statCounters.RemoteMapHit.inc()
           }
--- a/packages/core/src/local-workspace/remote_map/remote_map.ts
+++ b/packages/core/src/local-workspace/remote_map/remote_map.ts
@@ -398,7 +398,22 @@ const createFilteredReadIterator = (
   }
 }
 
-const createIterator = (prefix: string, opts: CreateIteratorOpts, connection: rocksdb): ReadIterator => {
+type CreateIteratorArgs = {
+  prefix: string
+  opts: CreateIteratorOpts
+  connection: rocksdb
+  isEmpty?: boolean | undefined
+}
+const createIterator = ({ prefix, opts, connection, isEmpty }: CreateIteratorArgs): ReadIterator => {
+  if (isEmpty) {
+    // If we know the result of the iterator is going to be empty, we can skip creating the connection iterator
+    // This is more efficient because the low level iterator works with a prefix, it doesn't really know about namespaces
+    // and therefore potentially has to do much more work to figure out that the iterator is empty
+    return {
+      next: async () => undefined,
+      nextPage: async () => undefined,
+    }
+  }
   // Cannot use inner limit when filtering because if we filter anything out we will need
   // to get additional results from the inner iterator
   const limit = opts.filter === undefined ? opts.first : undefined
@@ -430,6 +445,7 @@ export const createRemoteMapCreator = (
     deserialize,
   }: remoteMap.CreateRemoteMapParams<T>): Promise<remoteMap.RemoteMap<T, K>> => {
     let wasClearCalled = false
+    let isNamespaceEmpty: boolean | undefined
     const delKeys = new Set<string>()
     const locationTmpDir = getDBTmpDir(location)
     if (!(await fileUtils.exists(location))) {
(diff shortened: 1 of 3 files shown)


packages/core/src/local-workspace/remote_map/remote_map.ts
```diff
@@ -601,10 +637,12 @@ export const createRemoteMapCreator = (
                 resolve(undefined)
                 return
               }
+              isNamespaceEmpty = false
```
dantal4, line 640: do we need it here and in `resolveRet` too?

packages/core/src/local-workspace/remote_map/remote_map.ts
```diff
@@ -601,10 +637,12 @@ export const createRemoteMapCreator = (
...
               await resolveRet(innerValue)
               statCounters.RemoteMapHit.inc()
             })
           } else {
+            isNamespaceEmpty = false
```
dantal4, line 645: same question

packages/core/src/local-workspace/remote_map/remote_map.ts
```diff
@@ -491,6 +522,9 @@ export const createRemoteMapCreator = (
       if (i % batchInterval !== 0) {
         await promisify(batch.write.bind(batch))()
       }
+      if (i > 0) {
```
dantal4, line 525: can you explain this condition ?

packages/core/src/local-workspace/remote_map/remote_map.ts
```diff
@@ -753,17 +795,25 @@ export const createRemoteMapCreator = (
...
+        return found
       },
       close: async (): Promise<void> => {
         // Do nothing - we can not close the connection here
         //  because we share the connection across multiple namespaces
         log.warn('cannot close connection of remote map with close method - use closeRemoteMapsOfLocation')
       },
-      isEmpty: async (): Promise<boolean> => awu(keysImpl({ first: 1 })).isEmpty(),
+      isEmpty: async (): Promise<boolean> => {
+        if (isNamespaceEmpty === undefined) {
```
dantal4, line 812: maybe we want to update `isNamespaceEmpty` in `aggregatedIterable` too?  so in multiple calls to entries/keys/values we'll avoid creating an iterator if the namespace is empty.

```typescript
let done = false
let isEmpty = true
while (!done) {
  ...
  if (minEntry === undefined) {
    done = true
  } else {
    isEmpty = false
    ...
  }
}

isNamespaceEmpty = isEmpty
```

packages/core/src/local-workspace/remote_map/remote_map.ts
ori-moisis: if `i` stays `0` then the `batchInsertIterator` was empty and nothing actually happened, in which case, we didn't change the information we had about whether the namespace is empty

packages/core/src/local-workspace/remote_map/remote_map.ts
ori-moisis: no, not needed

packages/core/src/local-workspace/remote_map/remote_map.ts
ori-moisis: I think `aggregatedIterable` is too detached from the specific namespace that we are working with.
I thought about doing this in the iterator, but that required exposing more internal implementation in a way I think did not justify the potential gain since the main scenario we are trying to improve here is with a lot of calls to `isEmpty`...

packages/core/src/local-workspace/remote_map/remote_map.ts
```diff
@@ -398,7 +399,23 @@ const createFilteredReadIterator = (
   }
 }
 
-const createIterator = (prefix: string, opts: CreateIteratorOpts, connection: rocksdb): ReadIterator => {
+type CreateIteratorArgs = {
+  prefix: string
+  opts: CreateIteratorOpts
+  connection: rocksdb
+  statCounters?: LocationCounters
+  isEmpty?: boolean | undefined
```
dantal4, line 407: why are those optional?

dantal4 (approved):
💯

packages/core/src/local-workspace/remote_map/remote_map.ts
ori-moisis: Because there is another remote map implementation [here](https://github.com/salto-io/salto/pull/6613/files#diff-ffae2ae6f7ae4caad47d8c509cdeacdbdfd278a6a55cc30e91f8bf7346484b38R836) (createReadOnlyRemoteMapCreator) that does not use any caches...

PR #6614, opened 2024-09-24
SALTO-6056 Improve SuiteScripts object keys regex

dantal4:
We don't need to extract object keys in the form of `{ "key": value }` / `{ 'key': value }`, because they are extracted from the general strings regex (`semanticReferenceRegex`)

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


Shulik95 (approved):
🥇

PR #6616, opened 2024-09-24
SALTO-5955: (Bug-fix) Salesforce Profiles fetch results in incorrect values when retrieve requests are being chunked

tamtamirr:
(Bug-fix) Salesforce Profiles fetch results in incorrect values when retrieve requests are being chunked

---

The implementation of `lodash merge` isn't correct when merging array values, as it replaces them by index. There's example on the ticket. 
- I've removed the old UTs which were incorrect and didn't replicate the fetch flow correctly, hence never caught this bug.
- For clarity and future use-cases, created a dedicated test file for Profiles.

---
_Release Notes_: 
_Salesforce Adapter_:
- (Bug-fix) Salesforce Profiles fetch results in incorrect values when retrieve requests are being chunked

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/fetch.ts
```diff
@@ -359,8 +360,33 @@ export const retrieveMetadataInstances = async ({
...
+      [ProfileSection.ApplicationVisibilities]: ['application'],
+      [ProfileSection.ClassAccesses]: ['apexClass'],
+      [ProfileSection.FlowAccesses]: ['flow'],
+      [ProfileSection.LayoutAssignments]: ['layout', 'recordType'],
+      [ProfileSection.PageAccesses]: ['apexPage'],
+    }
+    const mergeSection = (sectionName: string, fieldsToUniqueBy: string[]): Values =>
+      _.uniqBy(
+        instances.flatMap(instance => instance.value[sectionName] ?? []),
+        value => fieldsToUniqueBy.map(fieldName => value[fieldName] ?? '').join('@'),
```
tamtamirr, line 378: Using `@` as the separator here as its an invalid character for metadata instances api names.

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -359,8 +360,33 @@ export const retrieveMetadataInstances = async ({
...
+      )
     const result = instances[0].clone()
-    result.value = _.merge({}, ...instances.map(instance => instance.value))
+    result.value = {
+      ..._.merge({}, ...instances.map(instance => instance.value)),
```
amitgoldarb, line 382: Do we still need this merge?

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -359,8 +360,33 @@ export const retrieveMetadataInstances = async ({
   const typesWithContent = await getTypesWithContent(types)
 
   const mergeProfileInstances = (instances: ReadonlyArray<InstanceElement>): InstanceElement => {
+    const fieldsToUniqueByPerSection: Record<ProfileSection, string[]> = {
```
amitgoldarb, line 363: Does this have to be so hard coded?

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -359,8 +360,33 @@ export const retrieveMetadataInstances = async ({
...
+      [ProfileSection.TabVisibilities]: ['tab'],
+      [ProfileSection.UserPermissions]: ['name'],
+      [ProfileSection.ApplicationVisibilities]: ['application'],
+      [ProfileSection.ClassAccesses]: ['apexClass'],
+      [ProfileSection.FlowAccesses]: ['flow'],
+      [ProfileSection.LayoutAssignments]: ['layout', 'recordType'],
+      [ProfileSection.PageAccesses]: ['apexPage'],
+    }
+    const mergeSection = (sectionName: string, fieldsToUniqueBy: string[]): Values =>
+      _.uniqBy(
```
amitgoldarb, line 376: Is lowerdash concatObjects good enough? I think from looking at the code it should be sufficient.

tamtamirr: Nope, as concatObjects won't remove duplicates which will result in incorrect values. No reason to use it considering we unique by different fields per section in the Profile.

packages/salesforce-adapter/src/fetch.ts
tamtamirr: Yes. What's the alternative? 🤔

amitgoldarb: Can't we deduce it from the profile itself?

packages/salesforce-adapter/src/fetch.ts
amitgoldarb: What do you think about this:
```
export const concatObjects = <
  T extends Record<string, ReadonlyArray<unknown> | unknown[] | undefined>,
>(
  objects: T[],
  deduplicationFns?: Record<string, (values: any[]) => any[]>,
): T =>
  _.mapValues(
    _.groupBy(objects.flatMap(Object.entries), ([key]) => key),
    (lists, key) => {
      const concatenated = lists
        .map(([_key, list]) => list)
        .filter(isDefined)
        .flat();

      if (deduplicationFns && deduplicationFns[key]) {
        return deduplicationFns[key](concatenated);
      }
      return concatenated;
    },
  ) as T;
```

  And then providing the dedup function that you provided?

amitgoldarb (commented):
Added small suggestion and question

amitgoldarb (approved):
Nice catch here!
Thanks for this

PR #6617, opened 2024-09-24
SALTO-6708: (Bug-fix) Profile layoutAssignments may result in missing assignments of RecordTypes when requests are being chunked

tamtamirr:
(Bug-fix) Profile layoutAssignments may result in missing assignments of RecordTypes when requests are being chunked

---

Issue is explained in details in the ticket.
- Will add UTs as soon as [this](https://github.com/salto-io/salto/pull/6616) PR is merged to utilize the `profiles.test.ts` module. Didn't want to couple the PRs though so it'd be easier to review without the previous PR changes.
---
_Release Notes_: 
_Salesforce Adapter_:
- (Bug-fix) Profile layoutAssignments may result in missing assignments of RecordTypes when requests are being chunked
---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/fetch.ts
```diff
@@ -388,8 +395,33 @@ export const retrieveMetadataInstances = async ({
   const retrieveInstances = async (
     fileProps: ReadonlyArray<FileProperties>,
     filePropsToSendWithEveryChunk: ReadonlyArray<FileProperties> = [],
+    customObjectFilePropsByName: Record<string, FileProperties>,
   ): Promise<InstanceElement[]> => {
-    const allFileProps = fileProps.concat(filePropsToSendWithEveryChunk)
+    // When fetching Profiles the layoutAssignments of RecordTypes require the parent CustomObject to be retrieved as part of the retrieve request.
+    const getLayoutParentsToRetrieve = (): FileProperties[] => {
```
amitgoldarb, line 401: Can we move this logic to the caller in the profile specific section?

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -388,8 +395,33 @@ export const retrieveMetadataInstances = async ({
...
+      if (parentFileProps.length > 0) {
+        log.debug(
+          'Adding parent CustomObjects to retrieve request: %s',
+          inspectValue(parentFileProps.map(prop => prop.fullName)),
+        )
+      }
+      return parentFileProps
+    }
+    const layoutParentsToRetrieve = getLayoutParentsToRetrieve()
+    const customObjectsToOmitFromResult = new Set(layoutParentsToRetrieve.map(prop => prop.fullName))
```
amitgoldarb, line 423: Why do we need to omit this in the result?

amitgoldarb (commented):
Looks good! Nice catch again.
Added some comments 🙏

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -479,19 +511,29 @@ export const retrieveMetadataInstances = async ({
...
+      .filter(
+        instance => !isInstanceOfTypeSync(CUSTOM_OBJECT) || !customObjectsToOmitFromResult.has(instance.value.fullName),
+      )
   }
 
   const retrieveProfilesWithContextTypes = async (
     profileFileProps: ReadonlyArray<FileProperties>,
     contextFileProps: ReadonlyArray<FileProperties>,
   ): Promise<Array<InstanceElement>> => {
+    const customObjectFilePropsByName = _.keyBy(
```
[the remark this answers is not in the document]
tamtamirr: We must use the list result of `CustomObjects` for change-based fetch 💘 (Since the context type will only contain file properties of modified CustomObjects. A layout can be modified / added regardless to the parent object)

packages/salesforce-adapter/src/fetch.ts
tamtamirr: The CustomObjects are not supposed to be retrieved for this chunk. They will be retrieved as part of the chunk they are part of, and we should remove them here to avoid duplicates.

packages/salesforce-adapter/src/fetch.ts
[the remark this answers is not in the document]
amitgoldarb: Don't we want only the modified custom objects in change based fetch? Or are you discussing the scenario where the profile was modified? It should be good to differentiate between the two.

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -479,19 +511,29 @@ export const retrieveMetadataInstances = async ({
...
+      (await client.listMetadataObjects([{ type: CUSTOM_OBJECT }])).result,
+      props => props.fullName,
+    )
     const allInstances = await Promise.all(
       _.chunk(contextFileProps, maxItemsInRetrieveRequest - profileFileProps.length)
         .filter(filesChunk => filesChunk.length > 0)
-        .map(filesChunk => retrieveInstances(filesChunk, profileFileProps)),
+        .map(filesChunk => retrieveInstances(filesChunk, profileFileProps, customObjectFilePropsByName)),
```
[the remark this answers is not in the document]
amitgoldarb: ?

packages/salesforce-adapter/src/fetch.ts
amitgoldarb: This means that we are fetching them n+1 times right? Im not sure how big of a difference it is, but maybe we can save the result of the last time?

packages/salesforce-adapter/src/fetch.ts
amitgoldarb: Then isn't that exactly `FilePropsToSendWithEveryChunk`?

amitgoldarb (commented):
Added some questions/comments

packages/salesforce-adapter/src/fetch.ts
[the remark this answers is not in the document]
tamtamirr: We want the parents of all of the modified layouts. The parent wont necessarily exist in the `contextFileProps` in change-based fetch as only modified CustomObjects will appear there.

PR #6621, opened 2024-09-25
SALTO-6026 Add adapter config for using new references in SuiteScripts

dantal4:
Add the `useNewReferencesInSuiteScripts` adapter config in order to calculate & set the `_generated_dependencies` of SuiteScripts.

- Ignores invalid file extensions
- Ignores commented code
- Extracts references from object keys

---

---
_Release Notes_: 
Netsuite Adapter:
- Add the `useNewReferencesInSuiteScripts` adapter config in order to calculate & set the `_generated_dependencies` of SuiteScripts

---
_User Notifications_: 
Netsuite Adapter:
- When the `useNewReferencesInSuiteScripts` adapter config is `true`, the `_generated_dependencies` annotation of file instances will be modified

### DIFF
--- a/packages/netsuite-adapter/src/config/types.ts
+++ b/packages/netsuite-adapter/src/config/types.ts
@@ -113,6 +113,7 @@ export type FetchParams = {
   // SALTO-6145: should be removed
   forceFileCabinetExclude?: boolean
   calculateNewReferencesInSuiteScripts?: boolean
+  useNewReferencesInSuiteScripts?: boolean
   findReferencesInFilesWithExtension?: string[]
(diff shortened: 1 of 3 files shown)


omrilit:
> Looks great 💯 
> 
> Do we want to add any extensions to the default list? the logs from last week:
> 
> output (1)
> 
> I think the only relevant extension is maybe `.ts`. WDYT?
> 
> @omrilit 
> 
> 

Sounds like a good idea

dantal4:
@Shulik95 @omrilit I already opened https://github.com/salto-io/salto/pull/6619 :)

PR #6625, opened 2024-09-25
SALTO-6721: [Intune] Create a static file from the Device Configuration template payload

shoshana-houri:
Some device configuration policies may contain configuration profile files returned as base64 blobs. This PR extracts them to static files.

---

See example .nacl structure [here](https://github.com/salto-io/microsoft-security/blob/main/microsoft_security/Records/IntuneDeviceConfiguration/svh_custom_template.nacl).
And the matching [static file](https://github.com/salto-io/microsoft-security/blob/main/static-resources/microsoft_security/IntuneDeviceConfiguration/svh_custom_template.s/example.xml).

---
_Release Notes_: 

---
_User Notifications_:


packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/device_configuration.ts
```diff
@@ -0,0 +1,35 @@
...
+import { createStaticFileFromBase64Blob } from '../../shared/utils'
+import { NAME_ID_FIELD } from '../../shared/defaults'
+
+/**
+ * Extracts the payload from the device configuration and creates a static file from it, if exists
+ */
+export const extractPayloadToStaticFile: AdjustFunctionSingle = async ({ value }) => {
+  validatePlainObject(value, DEVICE_CONFIGURATION_TYPE_NAME)
+  const { payload } = value
+  if (!payload) {
```
shir-reifenberg, line 21: we prefer explicit checks 

```suggestion
  if (payload !== undined) {
```
or even better ? 
```suggestion
  if (_.isString(payload)) {
```

packages/microsoft-security-adapter/src/definitions/fetch/shared/utils/static_file.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+import { StaticFile } from '@salto-io/adapter-api'
+import { naclCase } from '@salto-io/adapter-utils'
+import { ADAPTER_NAME } from '../../../../constants'
+
+/**
+ * Creates a static file from a base64 blob, using the element's full name as the folder name for uniqueness
+ */
+export const createStaticFileFromBase64Blob = ({
```
shir-reifenberg, line 16: nit: consider moving to `device_configuration.ts` if this is used in a single place ?

packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/device_configuration.ts
```diff
@@ -0,0 +1,35 @@
...
+    return { value }
+  }
+  return {
+    value: {
+      ...value,
+      payload: createStaticFileFromBase64Blob({
+        typeName: DEVICE_CONFIGURATION_TYPE_NAME,
+        fullName: value[NAME_ID_FIELD.fieldName],
```
shir-reifenberg, line 29: (I think I missed it in the previous reviews, sorry about this)
you're making an assumption here that `elemID` fields are always `NAME_ID_FIELD`, while it can be configured by the user and then the instance and the static file will be misaligned / we'll get few static files in the same file

I think the alternatives are 
1. calling the function to calculate the elemID here (see similar example [here](https://github.com/salto-io/salto/blob/main/packages/okta-adapter/src/filters/app_logo.ts#L97))
2. use `custom` to create the instance (I don't really like it 😅 

Any other suggestions ? WDYT?

packages/microsoft-security-adapter/src/definitions/fetch/shared/utils/static_file.ts
```diff
@@ -0,0 +1,34 @@
...
+  fullName,
+  fileName,
+  content,
+}: {
+  typeName: string
+  fullName: string
+  fileName: string
+  content: string
+}): StaticFile => {
+  const formattedFullName = naclCase(fullName).replace('@', '.')
```
shir-reifenberg, line 27: can we extract to a shared place something so that the other MS type and the Okta implementation will share code? 
@IdoZakk started [here](https://github.com/salto-io/salto/pull/6518/files) but didn't merge ultimately due to the Jira changes

shir-reifenberg (commented):
Great job (and super fast)
added some comments, LMK WDYT 😄

Nurdok (commented):
LGTM. I'll leave the approval for Shir :)

packages/microsoft-security-adapter/src/definitions/fetch/shared/utils/static_file.ts
shoshana-houri: It's also used in `platform_script.ts` :)

packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/device_configuration.ts
shoshana-houri: Ohh, that's a good point, thanks!
Can the elemID still be configured by the user with the new infra? I don't think we’re accounting for the user config when creating the fetch definitions (at least not explicitly in the Microsoft Security adapter)

packages/microsoft-security-adapter/test/definitions/fetch/intune/utils/device_configuration.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    it('should return the value as is if payload is missing', async () => {
+      const value = { displayName: 'name' }
+      expect(await extractPayloadToStaticFile({ value, typeName: 'test', context: contextMock })).toEqual({ value })
+    })
+
+    it('should return the value with the payload as a static file', async () => {
+      const value = {
+        displayName: 'name',
+        payload:
+          'PD94bWwgdmVyc2lvbj0iMS4wIiBlbmNvZGluZz0iVVRGLTgiPz4NCjxub3RlPg0KICA8dG8+VGVzdDwvdG8+DQogIDxmcm9tPlRlc3RlcjwvZnJvbT4NCiAgPGhlYWRpbmc+UGxlYXNlIHdvcmsgbmljZWx5PC9oZWFkaW5nPg0KICA8Ym9keT5UaGFuayB5b3U8L2JvZHk+DQo8L25vdGU+',
```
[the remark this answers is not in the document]
shoshana-houri: image
My kind of easter egg 🥚 😅 
You're right though, I'll use the shorter one

packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/device_configuration.ts
shir-reifenberg: Yap :) via `elemID` config (under `fetch`), see [this](https://github.com/salto-io/salto/blob/main/packages/adapter-components/src/adapter-wrapper/adapter/adapter.ts#L122) part

packages/adapter-utils/test/nacl_case_utils.test.ts
```diff
@@ -6,12 +6,20 @@
...
+  pathNaclCase,
+  prettifyName,
+} from '../src/nacl_case_utils'
 
 describe('naclCase utils', () => {
   const generateRandomChar = (): string => String.fromCharCode(Math.random() * 65535)
 
-  describe('naclCase func', () => {
+  describe(`${naclCase} func`, () => {
     it('should return empty string for undefined', () => {
```
Nurdok, line 23: Shouldn't this be `naclCase.name`? Ditto for the others.

Nurdok (approved):

packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/device_configuration.ts
shoshana-houri: Cool, thanks @shir-reifenberg! The issue with the first method is that AFAIU it can only be used after the definitions are resolved, whereas we need it when setting the fetch definitions, since this is called inside an adjust function in the fetch definitions.
WDYS? Perhaps we should consider allowing the user config to be propagated to the custom adjust functions?

shir-reifenberg: You're right @shoshana-houri 😄 , even if we pass the userConfig, it won't take into account `SALTO_DEFINITIONS_OVERRIDES` as well... and it's actually quite strange to calculate elem ID related things in the `adjust` function of the request

I guess that's the issue with creating static files while we don't have the parent instance yet in hand..
I think you can merge this for now, but we should address it differently as it wouldn't work with elemIDs customizations on the device configuration.
Can you please open a ticket for this ? (and also tag @netama, as she might have a suggestion / thought about it before)

PR #6627, opened 2024-09-26
SALTO-6724: (Bug-fix) Salesforce Fetch fails on TypeError: Cannot read properties of undefined (reading 'annotations')

tamtamirr:
(Bug-fix) Salesforce Fetch fails on TypeError: Cannot read properties of undefined (reading 'annotations')

---

Type safety issues when dealing with Records, again 🤦 
Not SOS as this is the first time we encounter issue like this and it's not a regression. Added logs to understand these cases in future.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/src/fetch.ts
+++ b/packages/salesforce-adapter/src/fetch.ts
@@ -391,6 +391,7 @@ export const retrieveMetadataInstances = async ({
     return false
   }
 
+  const missingTypes = new Set<string>()
   const retrieveInstances = async (
     fileProps: ReadonlyArray<FileProperties>,
     filePropsToSendWithEveryChunk: ReadonlyArray<FileProperties> = [],
@@ -400,7 +401,7 @@ export const retrieveMetadataInstances = async ({
     const filesToRetrieve = allFileProps.map(inst => {
       const metadataType = typesByName[inst.type]
       if (metadataType === undefined) {
-        log.warn('Could not find metadata type for %s', inst.type)
+        missingTypes.add(inst.type)
         return inst
       }
       return {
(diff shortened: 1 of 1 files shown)


packages/salesforce-adapter/src/fetch.ts
```diff
@@ -389,16 +391,24 @@ export const retrieveMetadataInstances = async ({
...
     const allFileProps = fileProps.concat(filePropsToSendWithEveryChunk)
     // Salesforce quirk - folder instances are listed under their content's type in the manifest
-    const filesToRetrieve = allFileProps.map(inst => ({
-      ...inst,
-      type: getManifestTypeName(typesByName[inst.type]),
-    }))
+    const filesToRetrieve = allFileProps.map(inst => {
+      const metadataType = typesByName[inst.type]
+      if (metadataType === undefined) {
+        missingTypes.add(inst.type)
```
amitgoldarb, line 404: Maybe log the entire fileProp? It could be useful

packages/salesforce-adapter/src/fetch.ts
```diff
@@ -532,6 +542,9 @@ export const retrieveMetadataInstances = async ({
   log.info('going to retrieve %d files', filesToRetrieve.length)
 
   const instances = await retrieveProfilesWithContextTypes(profileFiles, nonProfileFiles)
+  if (missingTypes.size > 0) {
+    log.warn('Missing metadata types in fetch: %s', inspectValue(Array.from(missingTypes)))
```
amitgoldarb, line 546: This might be a bit "loud", can we maybe make this debug? This might be happening right now no?

amitgoldarb (approved):
Approving to save cycles. Added 2 small nits

packages/salesforce-adapter/src/fetch.ts
tamtamirr: Mustn't happen often as the previous code would crash fetch. We've seen one instance of this happening

packages/salesforce-adapter/src/fetch.ts
tamtamirr: This log should help us understand edge-case where the adapter listed a type that does not exist. The instances don't matter and logging them may be too noisy.

amitgoldarb (approved):
Thanks for this

PR #6632, opened 2024-09-29
SALTO-6730: (Microsoft Security) Fix inconsistent order of array fields

shoshana-houri:
MS Graph doesn’t return the following fields in a specific order (between different tenants), so we sort them to maintain consistency:
* Assignments: Found in various Intune types.
* Application Tags: Found in Entra Service Principals.

---
_Release Notes_: 

---
_User Notifications_:


Nurdok (commented):
Woohoo for `sort` usage :)

Nurdok (approved):

PR #6633, opened 2024-09-29
SALTO-6735: (MS Security) Fix reference rule for Conditional Access Policy

shoshana-houri:
1. Add polling for creation/update of a Conditional Access Policy.
2. Fix reference rule for Conditional Access policies

---

1. When we deploy a conditional access policy and the service principal it references within the same request, the service might throw an error stating that the service principal does not exist. This issue resolved (i.e. the service principal is updated) after 5-15 seconds.

2. A Conditional access policy may contain the following structure for application references:
```
  conditions = {
    ...otherFields,
    applications = {
      includeApplications = [
        microsoft_security.EntraServicePrincipal.instance.someSPInstanceA,
        ...
      ]
      excludeApplications = [
        microsoft_security.EntraServicePrincipal.instance.someSPInstanceB,
        ...,
      ]
    }
  }
```

The current reference rule simply serialize the application reference by returning the `appId` field on the Service Principal.
However, this `appId` may (and in most cases _will_) be a reference itself - to the application instance.
In this case we should resolve this reference as well, and return the `appId` uuid.

Note that the reference here is to the Service Principal and not directly to the application. This is because that even if the application exists in the tenant - we still need its matching service principal to exist as well in order to reference it in the policy. 

---
_Release Notes_: 

---
_User Notifications_:

### DIFF
--- a/packages/microsoft-security-adapter/src/definitions/references/references.ts
+++ b/packages/microsoft-security-adapter/src/definitions/references/references.ts
@@ -17,18 +17,21 @@ const REFERENCE_RULES = [...EntraReferenceRules, ...IntuneReferenceRules, ...Cro
 
 // We only use this record as a TS 'hack' to fail the build if we add a custom serialization strategy
 // and forget to add it to the fieldsToGroupBy.
-const fieldsToGroupBy: Record<referenceUtils.ReferenceIndexField | CustomReferenceSerializationStrategyName, null> = {
-  id: null,
-  name: null,
-  appId: null,
-  bundleId: null,
-  packageId: null,
-  servicePrincipalAppId: null,
-}
+const fieldsToGroupBy: Record<referenceUtils.ReferenceIndexField | CustomReferenceSerializationStrategyName, boolean> =
+  {
+    id: true,
+    name: true,
+    appId: true,
+    bundleId: true,
+    packageId: true,
+    servicePrincipalAppId: false,
+  }
 
 export const REFERENCES: definitions.ApiDefinitions<Options>['references'] = {
   rules: REFERENCE_RULES,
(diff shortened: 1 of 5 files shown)


packages/microsoft-security-adapter/src/definitions/references/references.ts
```diff
@@ -35,6 +39,18 @@ export const REFERENCES: definitions.ApiDefinitions<Options>['references'] = {
       lookup: referenceUtils.basicLookUp,
       lookupIndexName: 'appId',
     },
+    servicePrincipalAppId: {
+      serialize: ({ ref }) => {
+        const { appId } = ref.value.value
+        if (isReferenceExpression(appId)) {
```
shir-reifenberg, line 45: nit: I would verify the type 
eg 
```suggestion
        if (isReferenceExpression(appId) && ref.value.elemID.typeName === 'ServicePrincipal') {
```

same in line 46, otherwise log error and throw

shir-reifenberg (approved):
🎉 
Can we please cover with a unit test?

packages/microsoft-security-adapter/src/definitions/references/references.ts
shoshana-houri: @shir-reifenberg I think we should only log and throw an error if it is a reference, but not a reference to an application. This is because we may have Service Principals where this field was not resolved as a reference if the application does not exist—for example, for built-in Microsoft applications.

shoshana-houri: Also, it looks like the serializer is running both on deploy and fetch (to validate the reference). We discussed this f2f and decided not to throw an error, but to only log it.

shoshana-houri:
@shir-reifenberg 
We will add deploy UTs for the Application type, using the library @Nurdok introduced in Okta.
However, we'll do this in a separate PR to ensure this change can be deployed as soon as possible :)

PR #6637, opened 2024-09-30
SAAS-5602: Dummy: Support loadElementsFromFolder

YuvalGivon:
Support loadElementsFromFolder in dummy adapter as a first step towards running "Deploy from PR" tests using the dummy adapter instead of using Netsuite.

---

Per the requirements specified in the ticket, the adapter will look for a dummy.nacl file in the provided directory and use it as a configuration file to generate elements.

---
_Release Notes_: 

- Dummy adapter supports loadElementsFromFolder function

---
_User Notifications_:

### DIFF
--- a/packages/dummy-adapter/src/adapter_creator.ts
+++ b/packages/dummy-adapter/src/adapter_creator.ts
@@ -15,12 +15,21 @@ import {
   GetCustomReferencesFunc,
   ConfigCreator,
   createRestriction,
+  FetchResult,
+  LoadElementsFromFolderArgs,
 } from '@salto-io/adapter-api'
 import { createDefaultInstanceFromType, inspectValue } from '@salto-io/adapter-utils'
 import { logger } from '@salto-io/logging'
 import _ from 'lodash'
 import DummyAdapter from './adapter'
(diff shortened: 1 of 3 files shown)


packages/dummy-adapter/src/adapter_creator.ts
```diff
@@ -57,6 +66,10 @@ const getCustomReferences: GetCustomReferencesFunc = async elements =>
       ]
     : []
 
+const loadElementsFromFolder = async ({ baseDir }: LoadElementsFromFolderArgs): Promise<FetchResult> => ({
```
omrifarkash, line 69: No need to re-define `loadElementsFromFolder` function type

```suggestion
const loadElementsFromFolder: Adapter['loadElementsFromFolder'] = async ({ baseDir }) => ({
```

omrifarkash (commented):
Looking good. Did you try to play with it locally and see how it works?

Also please add release notes for dummy adapter that it implements now `loadElementsFromFolder` dunction

omrifarkash (approved):
👑  Perfect

PR #6638, opened 2024-09-30
SALTO-6504: Fix memory leak in remote map cache

ori-moisis:
Remove reference counting logic from remote map cache. Unfortunately, we can't really count references because the interface for closing remote maps is not symmetric with the interface for creating them. Currently when we close remote maps, we just close them, no matter how many references they had, so the cache has to have the same behavior.

Also, removed the unused "revert" function from the remote map interface (will need another PR to fully remove it)

---

---
_Release Notes_: 
Core:
- Fix potential memory leak in long running processes using workspaces

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/core/src/local-workspace/remote_map/counters.ts
+++ b/packages/core/src/local-workspace/remote_map/counters.ts
@@ -32,9 +32,7 @@ type Counter = {
   value: () => number
 }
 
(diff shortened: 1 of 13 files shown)


tamtamirr (approved):
LGTM!

PR #6640, opened 2024-09-30
SALTO-6739: Fixing hiding elements in top level modifications

yelly:
Fix a bug which broke fetches which caused type elements to be hidden in a top-level modification, such as when hiding a type and adding a meta type in the same fetch.

---

Example env: https://github.com/salto-io/ariel-sf/commit/3a761290a2d4427610a319551c923d411bbc56f6

Not testing for unhiding types because that doesn't work due to an unrelated bug (SALTO-6741).

---
_Release Notes_: 
None.

---
_User Notifications_: 
None.


tamtamirr (approved):
LGTM!

PR #6646, opened 2024-10-06
SALTO-6750: Sort verification methods fields in Okta verify authenticator

shir-reifenberg:
Sort a field causing noise between fetches

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


Nurdok (approved):
image

PR #6649, opened 2024-10-06
Change NS features config defaults

dantal4:
- `fetch.forceFileCabinetExclude` - change to `true` by default
- `fetch.addLockedCustomRecordTypes` - change to `true` by default
- `fetch.resolveAccountSpecificValues` - change to `true` by default
- `fetch.useNewReferencesInSuiteScripts` - change to `true` by default
- `suiteAppClient.wsdlVersion` - change to `"2024_1"` by default

---

---
_Release Notes_: 
Netsuite Adapter:
- Change features config defaults

---
_User Notifications_:

### DIFF
--- a/packages/netsuite-adapter/e2e_test/adapter.test.ts
+++ b/packages/netsuite-adapter/e2e_test/adapter.test.ts
@@ -323,7 +323,7 @@ describe('Netsuite adapter E2E with real account', () => {
       name: randomString,
       state: 'AZ',
       mainAddress: {
-        country: '_uSMinorOutlyingIslands',
+        country: '_unitedStatesMinorOutlyingIslands',
         addressee: 'testSub',
         state: 'AZ',
         addrText: 'testSub<br>AZ <br>US Minor Outlying Islands',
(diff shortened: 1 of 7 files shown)


packages/netsuite-adapter/src/change_validators/data_account_specific_values.ts
```diff
@@ -94,25 +72,10 @@ const changeValidator: NetsuiteChangeValidator = async (
     return []
   }
 
-  if (config.fetch.resolveAccountSpecificValues !== false) {
```
omrilit, line 97: do we really want to remove that protection? maybe we should allow disabling it for few weeks just to make sure that we have this safety net?

omrilit (commented):
Just making sure that this PR is planned to be merged after October 8th, right?

packages/netsuite-adapter/src/change_validators/data_account_specific_values.ts
dantal4: We can do that, but in that case we should make sure that no customer has that disabled in their config.

dantal4:
> Just making sure that this PR is planned to be merged after October 8th, right?

@omrilit indeed

packages/netsuite-adapter/src/change_validators/data_account_specific_values.ts
omrilit: I think that it would be better to set an explicit override in LDK for them for few more weeks and keep that safety net. WDYT?

dantal4: ok. sounds good.

omrilit (approved):
Thanks!

PR #6652, opened 2024-10-06
SALTO-5846: Okta: Add support for brand customized pages

Nurdok:

---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/okta-adapter/src/definitions/fetch/fetch.ts
+++ b/packages/okta-adapter/src/definitions/fetch/fetch.ts
@@ -33,6 +33,8 @@ import {
   EMAIL_CUSTOMIZATION_TYPE_NAME,
   EMAIL_TEMPLATE_TYPE_NAME,
   AUTOMATION_RULE_TYPE_NAME,
+  SIGN_IN_PAGE_TYPE_NAME,
+  ERROR_PAGE_TYPE_NAME,
 } from '../../constants'
 import { isGroupPushEntry } from '../../filters/group_push'
 import { extractSchemaIdFromUserType } from './types/user_type'
@@ -680,6 +682,14 @@ const createCustomizations = ({
           typeName: 'EmailTemplate',
           context: { args: { brandId: { root: 'id' } } },
         },
+        SignInPage: {
+          typeName: 'SignInPage',
+          context: { args: { brandId: { root: 'id' } } },
+        },
+        ErrorPage: {
+          typeName: 'ErrorPage',
+          context: { args: { brandId: { root: 'id' } } },
+        },
       },
     },
     element: {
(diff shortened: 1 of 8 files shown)


packages/okta-adapter/src/definitions/requests/clients.ts
```diff
@@ -91,6 +91,16 @@ export const createClientDefinitions = (
               omitBody: true,
             },
           },
+          '/api/v1/brands/{brandId}/pages/sign-in/customized': {
+            get: {
+              additionalValidStatuses: [404],
```
shir-reifenberg, line 96: This is the response when the default page is defined and there's no customization, right ?

packages/okta-adapter/test/mock_replies/custom_pages_modify.json
```diff
@@ -0,0 +1,67 @@
...
+        "signInLabel": "Sign in test",
+        "usernameLabel": "what's your name",
+        "usernameInfoTip": "ok",
+        "passwordLabel": "what's you password",
+        "passwordInfoTip": "okok",
+        "showPasswordVisibilityToggle": true,
+        "showUserIdentifier": true,
+        "forgotPasswordLabel": "Did you forget your password????",
+        "forgotPasswordUrl": "https://help.salto.io/en/articles/8721163-connecting-okta-adapter-using-custom-oauth-integration",
+        "unlockAccountLabel": null,
```
shir-reifenberg, line 41: Does the fact it is returned as `null` means we need to set it to `null` on modifications for removed fields? (similar to what we did with apps with `transformRemovedValuesToNull`)

shir-reifenberg (approved):
Looks great 🔥 🔥  
with one question (about the `null` fields) and a comment 

Also, can you please add noise reduction (for new added instances) + update user notifications that instances of those types will be added ?

packages/okta-adapter/src/definitions/requests/clients.ts
Nurdok: Yes, added a comment.

packages/okta-adapter/src/definitions/fetch/fetch.ts
```diff
@@ -803,6 +829,34 @@ const createCustomizations = ({
...
+  [ERROR_PAGE_TYPE_NAME]: {
+    requests: [
+      {
+        endpoint: { path: '/api/v1/brands/{brandId}/pages/error/customized' },
+      },
+    ],
+    element: {
+      topLevel: {
+        isTopLevel: true,
+        elemID: { extendsParent: true },
```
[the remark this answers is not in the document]
Nurdok: AFAICT it can be omitted (https://github.com/salto-io/salto/blob/1e3e395586497d24d2bbf8fdef04d9bb274ab4b2/packages/adapter-components/src/fetch/element/id_utils.ts#L141). I searched and didn't find a log entry complaining.

packages/okta-adapter/test/mock_replies/custom_pages_modify.json
Nurdok: Nope, I've tried to delete a few of these and they revert to the default just fine without any changes.

PR #6657, opened 2024-10-08
Throw the right error on InvalidSuiteAppCredentialsError

dantal4:
When we receive an invalid creds error inside the `listBundles` query we throw this error:
`listBundles operation failed. Received the following error: Invalid SuiteApp credentials`
We should instead throw the `InvalidSuiteAppCredentialsError` itself.

---

---
_Release Notes_: 
Netsuite Adapter:
- Throw a right error on InvalidSuiteAppCredentialsError

---
_User Notifications_: 
None


Shulik95 (approved):
💯

PR #6658, opened 2024-10-08
SALTO-5887 JSM - support Field RequestType name change

kami-delaroz:
Temp solution till we handle id-alignment generally

---

---
_Release Notes_: 
_Jira Adapter_:
- SALTO-5887: Fix a bug where RequestType field name in JSM is different for some old envs

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/config/config.ts
+++ b/packages/jira-adapter/src/config/config.ts
@@ -75,6 +75,7 @@ type JiraFetchConfig = definitions.UserFetchConfig<{ fetchCriteria: JiraFetchFil
   enableAssetsObjectFieldConfiguration?: boolean
   automationPageSize?: number
   splitFieldContextOptions?: boolean
+  enableRequestTypeFieldNameAlignment?: boolean
 }
 
 export type MaskingConfig = {
@@ -328,6 +329,7 @@ const fetchConfigType = definitions.createUserFetchConfigType({
     enableAssetsObjectFieldConfiguration: { refType: BuiltinTypes.BOOLEAN },
     automationPageSize: { refType: BuiltinTypes.NUMBER },
     splitFieldContextOptions: { refType: BuiltinTypes.BOOLEAN },
+    enableRequestTypeFieldNameAlignment: { refType: BuiltinTypes.BOOLEAN },
   },
   fetchCriteriaType: fetchFiltersType,
   omitElemID: true,
--- a/packages/jira-adapter/src/config/config.ts
+++ b/packages/jira-adapter/src/config/config.ts
@@ -385,6 +385,7 @@ export const configType = createMatchingObjectType<Partial<JiraConfig>>({
       'fetch.enableAssetsObjectFieldConfiguration',
       'fetch.automationPageSize',
       'fetch.parseAdditionalAutomationExpressions',
+      'fetch.enableRequestTypeFieldNameAlignment',
       'deploy.taskMaxRetries',
       'deploy.taskRetryDelay',
       'deploy.ignoreMissingExtensions',
--- a/packages/jira-adapter/src/filters/fields/field_name_filter.ts
+++ b/packages/jira-adapter/src/filters/fields/field_name_filter.ts
@@ -29,15 +29,16 @@ const getFieldType = (instance: InstanceElement): string | undefined =>
 
 const isCustomField = (instance: InstanceElement): boolean => instance.value.schema?.custom !== undefined
 
-const nameNaclCase = (baseName: string, instance: InstanceElement, config: JiraConfig): string => naclCase(
-  [
-    baseName,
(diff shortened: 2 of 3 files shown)


kami-delaroz:
Tailor made for them, I'll add a FF around it soon. please comment on the logic, it behaved nicely in the SAAS

packages/jira-adapter/src/filters/fields/field_name_filter.ts
```diff
@@ -26,22 +29,32 @@ const getFieldType = (instance: InstanceElement): string | undefined =>
 
 const isCustomField = (instance: InstanceElement): boolean => instance.value.schema?.custom !== undefined
 
+const nameNaclCase = (baseName: string, instance: InstanceElement, config: JiraConfig): string => naclCase(
```
dantal4, line 32: naming nit - maybe `getFieldElementName` ?

packages/jira-adapter/src/filters/fields/field_name_filter.ts
```diff
@@ -26,22 +29,32 @@ const getFieldType = (instance: InstanceElement): string | undefined =>
...
+  ]
+    .filter(values.isDefined)
+    .join('__'),
+)
+
 const getInstanceName = (instance: InstanceElement, config: JiraConfig, getElemIdFunc?: ElemIdGetter): string => {
   const baseName = generateInstanceNameFromConfig(instance.value, instance.elemID.typeName, config.apiDefinitions)
 
   if (baseName === undefined) {
     return instance.elemID.name
```
dantal4, line 46: can we get here with `instance.elemID.name === 'Customer_Request_Type@ss'` ?

packages/jira-adapter/src/filters/fields/field_name_filter.ts
```diff
@@ -26,22 +29,32 @@ const getFieldType = (instance: InstanceElement): string | undefined =>
...
-      baseName,
-      config.fetch.addTypeToFieldName ?? true ? getFieldType(instance) : undefined,
-      isCustomField(instance) ? CUSTOM_FIELDS_SUFFIX : undefined,
-    ]
-      .filter(values.isDefined)
-      .join('__'),
-  )
+  // SALTO-5887: JSM CustomerRequestType was changed to RequestType - support same id for both
+  // we do not support id stickiness (through get element from state) in this case
+  if (baseName === 'Customer Request Type' && instance.value.isLocked) {
```
dantal4, line 51: what about `instance.value.name` ? it will remain 'Customer Request Type', and therefore will appear as a diff between the environments.

dantal4 (commented):
Looks great!
please add a test for that too (that checks that we don't take the name from `getElemIdFunc`

packages/jira-adapter/src/filters/fields/field_name_filter.ts
kami-delaroz: Indeed. I wondered regarding it, and thought it might be better to allow the user to have the "real" field name in his Salto env. could be less confusing. The real issue was the deployment blocker - and once we recognize this element as the same with a name difference, *and* we block deploying this name diff as it's a locked field, I think it's the best we could have here

packages/jira-adapter/src/config/config.ts
```diff
@@ -328,6 +329,7 @@ const fetchConfigType = definitions.createUserFetchConfigType({
     enableAssetsObjectFieldConfiguration: { refType: BuiltinTypes.BOOLEAN },
     automationPageSize: { refType: BuiltinTypes.NUMBER },
     splitFieldContextOptions: { refType: BuiltinTypes.BOOLEAN },
+    enableRequestTypeFieldNameAlignment: { refType: BuiltinTypes.BOOLEAN },
```
shir-reifenberg, line 332: Didn't review the rest of the code, but I think you might want to remove it from the defaultConfig ? otherwise every workspace will be created with this field visible in the adapter config nacl (see [this](https://github.com/salto-io/salto/blob/main/packages/jira-adapter/src/config/config.ts#L385) for example)

packages/jira-adapter/src/filters/fields/field_name_filter.ts
kami-delaroz: nope

packages/jira-adapter/test/filters/fields/field_name_filter.test.ts
```diff
@@ -156,4 +156,69 @@ describe('field_name_filter', () => {
     await filter.onFetch(elements)
     expect(elements.map(e => e.elemID.name)).toEqual(['custom2'])
   })
+
+  describe('SALTO-5887: JSM CustomerRequestType was changed to RequestType - support same id for both', () => {
```
dantal4, line 160: all of the `it.only` should be `it`.

dantal4 (approved):
💯

PR #6659, opened 2024-10-09
SALTO-6726 Flag to deploy articles as drafts

adfineberg:
Add a flag that will deploy all Zendesk articles as drafts (implemented through article_translations, as per [API ref](https://developer.zendesk.com/api-reference/help_center/help-center-api/articles/#update-article)).

---

---
_Release Notes_: 
_Zendesk Adapter:_
* Add a flag that will deploy all Zendesk articles as drafts: `{ deploy: { deployArticlesAsDraft: true } }`

---
_User Notifications_: 
_Zendesk Adapter:_
* Article instances will have the fields `draft` and `locale` omitted, as they are controlled via their translations.


packages/zendesk-adapter/src/filters/deploy_articles_as_drafts.ts
```diff
@@ -0,0 +1,32 @@
...
+import { ARTICLE_TRANSLATION_TYPE_NAME } from '../constants'
+import { FilterCreator } from '../filter'
+
+/**
+ * Update articles to be deployed to draft if the flag deployArticlesAsDraft is true.
+ * These updates are done through article translations.
+ */
+const filterCreator: FilterCreator = ({ config }) => ({
+  name: 'deployBrandedGuideTypesFilter',
+  preDeploy: async (changes: Change<InstanceElement>[]) => {
```
edenhassid, line 19: what about onDeploy? to return it to the way it was? I think if we don't have it we will get partially successful. maybe the right solution is a fixer and not a filter if we don't want to return it to the way that it was.

edenhassid (commented):
nice! 🔥 🚀 added a comment :)

packages/zendesk-adapter/src/fix_elements/deploy_articles_as_drafts.ts
```diff
@@ -0,0 +1,56 @@
...
+ * These updates are done through article translations.
+ */
+export const deployArticlesAsDraftHandler: FixElementsHandler =
+  ({ config }) =>
+  async elements => {
+    if (config[FIX_ELEMENTS_CONFIG]?.deployArticlesAsDraft !== true) {
+      return {
+        fixedElements: [],
+        errors: [],
+      }
```
edenhassid, line 43: I don't think you need this part, the fixer will not run if the config flag is false

packages/zendesk-adapter/src/fix_elements/deploy_articles_as_drafts.ts
```diff
@@ -0,0 +1,56 @@
...
+import { FIX_ELEMENTS_CONFIG } from '../config'
+import { ARTICLE_TRANSLATION_TYPE_NAME } from '../constants'
+import { FixElementsHandler } from './types'
+
+const isRelevantElement = (element: unknown): element is InstanceElement =>
+  isInstanceElement(element) &&
+  element.elemID.typeName === ARTICLE_TRANSLATION_TYPE_NAME &&
+  element.value.draft === false
+
+const updateDraftToFalse = (element: InstanceElement): InstanceElement | undefined => {
```
edenhassid, line 19: nit: why can this return undefined?

packages/zendesk-adapter/src/fix_elements/deploy_articles_as_drafts.ts
```diff
@@ -0,0 +1,56 @@
...
+    if (config[FIX_ELEMENTS_CONFIG]?.deployArticlesAsDraft !== true) {
+      return {
+        fixedElements: [],
+        errors: [],
+      }
+    }
+    const fixedArticleTranslationsWithDraft = elements
+      .filter(isRelevantElement)
+      .map(updateDraftToFalse)
+      .filter(values.isDefined)
```
edenhassid, line 48: if it can't return undefined you don't need this

edenhassid (approved):
looks good!! added some small comments, pre-approving to save a cycle 😄

PR #6664, opened 2024-10-10
Salto-6621: Skip minifying profiles when dumping SFDX

yelly:
When dumping SFDX we need to pass entire profiles and not minified.

---

This required changing the interface for SF filters so that we can selectively skip them in the SFDX flow (Local/Remote doesn't really cover all the bases here).

---
_Release Notes_: 
* None.

---
_User Notifications_: 
* None.

### DIFF
--- a/packages/salesforce-adapter/src/filters/minify_deploy.ts
+++ b/packages/salesforce-adapter/src/filters/minify_deploy.ts
@@ -20,7 +20,7 @@ import {
 import { collections, values } from '@salto-io/lowerdash'
 import _ from 'lodash'
 import { detailedCompare, getPath } from '@salto-io/adapter-utils'
-import { LocalFilterCreator } from '../filter'
+import { FilterCreator } from '../filter'
 import { isInstanceOfTypeChange } from './utils'
 import { PROFILE_METADATA_TYPE, INSTANCE_FULL_NAME_FIELD } from '../constants'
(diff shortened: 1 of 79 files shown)


packages/salesforce-adapter/src/filters/minify_deploy.ts
```diff
@@ -117,6 +122,11 @@ const filterCreator: LocalFilterCreator = () => {
       changes.push(...(await Promise.all(relatedChanges.map(toMinifiedChange))))
     },
     onDeploy: async changes => {
+      if (client === undefined) {
+        // We don't want to run this filter when the results aren't being sent to the service.
```
tamtamirr, line 126:
```suggestion
        // There's no need to minify the Profile Elements in the SFDX Flow
```

tamtamirr (approved):
LGTM! Added one small suggestion

packages/salesforce-adapter/src/filters/minify_deploy.ts
yelly: Done.

ori-moisis:
can we please have the giant refactor of all filters as a separate PR and not a "side effect" of this seemingly tiny feature?

yelly:
Sorry @ori-moisis - saw your comment too late...

PR #6666, opened 2024-10-10
SALTO-6621: Sort profiles

yelly:
Reordering profiles pre deploy to maintain order in SFDX.
This is the last issue for profiles so enabling them.

---

For elements split between files, the NaCl source loads only the files that have been changed and then merges them with the element in the state. This means that the order of the value depends on which files have been changed.
In order to avoid fundementally changing the way the NaCl source works I instead opted for just fixing the issue for Profiles which as far as I can tell is the only type affected in this way which is merged back into a single file in SFDX.

---
_Release Notes_: 
* None.

---
_User Notifications_: 
* None.

### DIFF
--- a/packages/salesforce-adapter/src/filters/profile_instance_split.ts
+++ b/packages/salesforce-adapter/src/filters/profile_instance_split.ts
@@ -7,11 +7,19 @@
  */
 import _ from 'lodash'
 import { strings, collections, promises } from '@salto-io/lowerdash'
-import { Element, ObjectType, isMapType, InstanceElement, isInstanceElement } from '@salto-io/adapter-api'
+import {
+  Element,
+  ObjectType,
+  isMapType,
+  InstanceElement,
+  Change,
+  isInstanceChange,
+  isAdditionOrModificationChange,
+} from '@salto-io/adapter-api'
 
 import { FilterCreator } from '../filter'
-import { metadataType } from '../transformers/transformer'
 import { PROFILE_METADATA_TYPE } from '../constants'
+import { isInstanceOfTypeChangeSync, isInstanceOfTypeSync } from './utils'
 
 const { groupByAsync, awu } = collections.asynciterable
 const { removeAsync } = promises.array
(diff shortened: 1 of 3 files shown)


packages/salesforce-adapter/src/filters/profile_instance_split.ts
```diff
@@ -46,24 +54,49 @@ const splitProfile = async (profile: InstanceElement): Promise<InstanceElement[]
...
+      isInstanceOfTypeSync(PROFILE_METADATA_TYPE),
+    )) as InstanceElement[]
     if (profileInstances.length === 0) {
       return
     }
     const newProfileInstances = await awu(profileInstances).flatMap(splitProfile).toArray()
     elements.push(...newProfileInstances)
   },
+  preDeploy: async (changes: Change[]) => {
+    // Splitting into files can cause the internal order of the profile to be affected by which files are changed.
```
amitgoldarb, line 88: Super nit - I think this comment should be on line 93 (above the actual sorting), or above the pre-deploy alltogether.

packages/salesforce-adapter/src/filters/profile_instance_split.ts
```diff
@@ -46,24 +54,49 @@ const splitProfile = async (profile: InstanceElement): Promise<InstanceElement[]
...
+    // In order to maintain the order in SFDX we sort the value.
+    if (client !== undefined) {
```
amitgoldarb, line 90: I find this rather cryptic, is this the convention for how we know if the predeploy filter is running for SFDX purposes or not?

amitgoldarb (approved):
Approving to save cycles. Added a small comment

packages/salesforce-adapter/src/filters/profile_instance_split.ts
yelly: Addressing in next PR.

packages/salesforce-adapter/src/filters/profile_instance_split.ts
yelly: Addressing in next PR.

yelly: Oh actually I can address here - yes, this is our way for now to differentiate the SFDX flow. Not ideal but better than what we had before and any superior solution will likely require a huge redesign in adapter API etc.

PR #6667, opened 2024-10-10
SALTO-6711 save fields and annotations type references

dantal4:
In SALTO-6697 we want to delete the nacl source’s referenced index. one place that it was used was to create validation errors on field values in instances/annotations in case that a field type has changed. In order to do that using the workspace’s reference sources index we need to save there the field to field type references, and type to its annotation ref types references.

In this PR there are several changes:
- Save field/annotation refType dependencies to the reference sources&targets indexes
- Add an env var that controls how do we calculate the dependent elements that should be validated (using nalc source's referenced index OR referenceSources index)

---

I calculated the dependets for each element in a large netsuite+salesforce environment and checked that the results makes sense- that we're removing only false dependats, and not adding to much dependents that may cause a regression in the calculation time.

---
_Release Notes_: 
Workspace:
- Save field/annotation refType dependencies to the reference sources&targets indexes.
- Add an env var that controls how do we calculate the dependent elements that should be validated.

---
_User Notifications_: 
None


packages/workspace/src/workspace/workspace.ts
```diff
@@ -506,6 +506,7 @@ export const loadWorkspace = async (
   persistent = true,
   mergedRecoveryMode: MergedRecoveryMode = 'rebuild',
   getCustomReferences: WorkspaceGetCustomReferencesFunc = async () => [],
+  validateDependentsMode: 'old' | 'new' | 'log-validations-diff' = 'old',
```
ori-moisis, line 509: I don't think we should expose this in the workspace interface, this should not be in the code long term.

personally I think this is a bit overly defensive and potentially computationally expensive to the point that we won't actually use it, but even if I let that part go, it should not be in the external interface. if you really think this is needed, it can be a core flag maybe?

packages/workspace/src/validator.ts
```diff
@@ -856,3 +863,149 @@ export const validateElements = async (
...
+    // top level type changes are dependencies in all groups
+    const groupDedendencies = ids.concat(typeTopLevelChangeIDs)
+    const topLevelIDs = groupDedendencies.map(id => id.createTopLevelParentID().parent)
+    return _.uniqBy(topLevelIDs, id => id.getFullName())
+  })
+
+  return topLevelDependencies
+}
+
+export const validateElementsAndDependents = async (
```
ori-moisis, line 969: it is very strange that the new implementation and the old implementation are in different files.

I am ok with putting the new implementation here, but if you choose to do this and keep the old implementation, the old implementation should move here as well

ori-moisis (commented):
didn't fully review the logic, mostly looked at the things around it, but I see I am delaying this too much, so posting the comments I have so far

packages/adapter-api/src/element_id.ts
```diff
@@ -325,6 +325,15 @@ export class ElemID {
...
   }
 
+  getInnerTypeElemID(): ElemID {
+    const container = this.getContainerPrefixAndInnerType()
+    if (container === undefined) {
+      return this
+    }
+    const innerTypeElemID = ElemID.fromFullName(container.innerTypeName)
+    return innerTypeElemID.getInnerTypeElemID()
+  }
```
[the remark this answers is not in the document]
dantal4: it indeed returns the same result. I'll use the existing function.

packages/workspace/src/workspace/reference_indexes.ts
```diff
@@ -68,48 +72,90 @@ export type ReferenceTargetIndexValue = collections.treeMap.TreeMap<ReferenceInd
...
+  customReferences.forEach(addReferenceInfo)
   const templateExpressions = await Promise.all(
     templateStaticFiles.map(async fileObj => {
       const expression = await parserUtils.staticFileToTemplateExpression(fileObj.value)
       return { expression, source: fileObj.source }
     }),
   )
   templateExpressions.forEach(fileObj => {
-    getReferencesFromTemplateExpression(fileObj.source, fileObj.expression)
+    getReferencesFromTemplateExpression(fileObj.source, fileObj.expression, addReferenceInfo)
```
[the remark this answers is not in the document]
dantal4: you're right. changing it to consider custom references first.

packages/workspace/src/workspace/workspace.ts
dantal4: I think that if the new logic will miss some dependencies in the calculation, it will take a long time until someone will notice it, so IMO having the `log-validations-diff` mode running for two weeks will let us find relevant&common edge cases that the new logic misses.
as for core flag - it's not in the `core` package so we'll need to pass a parameter to `loadWorkspace` anyway, or use an env var directly in here - something like `SALTO_VALIDATE_DEPENDENTS_MODE` - WDYT?

ori-moisis: if these are the only options, I would opt for env var (maybe we need to move the "core flag" mechanism to somewhere more generic?)

as for running `log-validations-diff` - I think creating the test scenarios "locally" is just better, I don't think the unit tests are enough, and that is OK, I think this requires "manual" testing - taking real / imagined scenarios and running them (e.g - random sample a list of IDs from a workspace and see that you get the same result from both algorithms, repeat many times with different sample sizes etc...)

I am ok with keeping the old code available through an env var as a last resort if needed, but given that this can take up to 3 minutes, I don't think it would make sense to run it twice in a real scenario...

packages/workspace/src/validator.ts
```diff
@@ -856,3 +863,149 @@ export const validateElements = async (
...
+    instancesDependencies,
+    elementsSource,
+    referenceSourcesIndex,
+  )
+  const dependentTypeIDs = await getDependentIDs(typesDependencies, referenceSourcesIndex, isDependentAnnotationType)
+  const dependentByReferencesIDs = await getDependentIDs(
+    referencesDependencies,
+    referenceSourcesIndex,
+    isDependentReference,
+  )
```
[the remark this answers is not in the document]
dantal4: removed from the PR.

packages/workspace/src/workspace/workspace.ts
```diff
@@ -908,14 +859,22 @@ export const loadWorkspace = async (
...
-      const changeIDs = changes.map(getChangeData).map(elem => elem.elemID)
       if (validate) {
-        const { errors: validationErrors, validatedElementsIDs } = await validateElementsAndDependents(
-          changedElements,
-          stateToBuild.states[envName].merged,
+        const changeIDs = changes.map(change => getChangeData(change).elemID)
+        const changedElements: Element[] = changes.filter(isAdditionOrModificationChange).map(getChangeData)
+        const dependents = await getDependents(
           changeIDs,
+          stateToBuild.states[envName].merged,
```
dantal4, line 867: Should we pass here the visible elements source, so we'll validate only visible elements?

packages/workspace/src/workspace/dependents.ts
```diff
@@ -0,0 +1,144 @@
...
+        ids.forEach(id => {
+          addedIDs.add(id.getFullName())
+        })
+
+        const dependentIDs = await log.timeTrace(
+          async () =>
+            awu(ids)
+              // TODO: should we filter out weak referenecs or references that aren't in the element?
+              .map(id => referenceSourcesIndex.get(id.getFullName()))
+              .flatMap(references => references ?? [])
```
ori-moisis, line 37: nit - can just replace this with
```suggestion
              .flatMap(id => referenceSourcesIndex.get(id.getFullName()) ?? [])
```
?

packages/workspace/src/workspace/dependents.ts
```diff
@@ -0,0 +1,144 @@
...
+              .toArray(),
+          'getDependentIDs for %d ids',
+          ids.length,
+        )
+
+        return dependentIDs.length === 0 ? dependentIDs : dependentIDs.concat(await getDependentIDs(dependentIDs))
+      }
+
+      const dependentIDs = await getDependentIDs(elemIDs)
+      const additionalDependentInstanceIDs = await awu(await elementsSource.list())
```
ori-moisis, line 50: comment about what is going on here and why it is needed?

packages/workspace/src/workspace/dependents.ts
```diff
@@ -0,0 +1,144 @@
...
+        )
+        .toArray()
+
+      return dependentIDs.concat(additionalDependentInstanceIDs)
+    },
+    'getDependentIDsFromReferenceSourceIndex for %d elemIDs',
+    elemIDs.length,
+  )
+
+const getDependentIDsFromReferencedFiles = async (
```
ori-moisis, line 65: can I assume this is an exact copy-paste from the original code or was this changed and needs to be reviewed?

packages/workspace/src/workspace/dependents.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  const dependents = (
+    await log.timeDebug(
+      () => Promise.all(dependentIDs.map(id => elementsSource.get(id))),
+      'getting %d dependents from elements source',
+      dependentIDs.length,
+    )
+  ).filter(isElement)
+
+  log.debug('found %d dependents of %d elements', dependents.length, elemIDs.length)
```
ori-moisis, line 141: `dependents.length` is after filtering out `dependents` that are not `isElement`.
may want to log the original number of `dependentIDs` as well (or if you want to be really defensive, if it is not the same number, log a warning and monitor it because it should never happen IIUC)

packages/workspace/src/workspace/reference_indexes.ts
```diff
@@ -99,20 +145,21 @@ const getReferences = async (element: Element, customReferences: ReferenceInfo[]
...
-      const expression = await parserUtils.staticFileToTemplateExpression(fileObj.value)
-      return { expression, source: fileObj.source }
-    }),
-  )
-  templateExpressions.forEach(fileObj => {
-    getReferencesFromTemplateExpression(fileObj.source, fileObj.expression)
-  })
+  const templateExpressionReferences = await Promise.all(
+    templateStaticFiles.map(getReferencesFromTemplateStaticFile),
+  ).then(res => res.flat())
```
ori-moisis, line 150: style nit - don't mix `await` and `then`

packages/workspace/src/workspace/reference_indexes.ts
```diff
@@ -99,20 +145,21 @@ const getReferences = async (element: Element, customReferences: ReferenceInfo[]
...
 
-  return Object.values(references)
+  const fieldRefTypesReferences = getReferencesFromFieldRefTypes(element)
+  const annotationRefTypesReferences = getReferencesFromAnnotationRefTypes(element)
```
ori-moisis, line 153: also need to consider meta type if one exists?

packages/workspace/src/workspace/reference_indexes.ts
```diff
@@ -99,20 +145,21 @@ const getReferences = async (element: Element, customReferences: ReferenceInfo[]
...
+
+  return _.uniqBy(
+    customReferences
```
ori-moisis, line 156: IIUC, the order is important here (i.e - `customReferences` must be the first)
if this is true, worth adding a comment saying that it is important and why

packages/workspace/src/workspace/workspace.ts
ori-moisis: that could lead to incorrect errors if, for example, there is a hidden value that is required.
also, if you pass in only the visible elements, you won't have access to any of the types in most adapters

packages/workspace/src/workspace/workspace.ts
```diff
@@ -908,14 +859,22 @@ export const loadWorkspace = async (
...
+          stateToBuild.states[envName].referenceSources,
```
ori-moisis, line 868: are the reference source already updated at this point?
is it good if they are? (will it cause issues for removal changes?)

packages/workspace/src/workspace/workspace.ts
dantal4: But a hidden value will be missing in an addition, so maybe making it required is the problem...

ori-moisis: > if you pass in only the visible elements, you won't have access to any of the types in most adapters

dantal4: why do we need the access to the types? to `validateElements` we pass the `merged` source, so in the validation itself the types are accessible. the question is why to validate a hidden type?

ori-moisis: mixing different element sources here adds complexity which I don't fully understand.
If I don't understand something and I don't see the value in doing it, I opt not to do it.

if you do understand it and think there is a logical reason why this needs to use a different element source and the is value in using a different element source, I won't prevent you from doing it, I just can't anticipate what will happen and what edge cases will arise from this...

packages/workspace/src/workspace/dependents.ts
```diff
@@ -0,0 +1,144 @@
...
+
+      const getDependentIDs = async (ids: ElemID[]): Promise<ElemID[]> => {
+        ids.forEach(id => {
+          addedIDs.add(id.getFullName())
+        })
+
+        const dependentIDs = await log.timeTrace(
+          async () =>
+            awu(ids)
+              // TODO: should we filter out weak referenecs or references that aren't in the element?
```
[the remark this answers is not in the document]
dantal4: that's what I thought, about adding a field of `source: 'nacl' | 'custom'` to the reference info, but maybe it's not that crucial.

packages/workspace/src/workspace/dependents.ts
dantal4: `referenceSourcesIndex.get(..)` returns a promise..

packages/workspace/src/workspace/workspace.ts
dantal4: I will leave it as it is for now 👍

ori-moisis (approved):

packages/workspace/src/workspace/dependents.ts
```diff
@@ -0,0 +1,150 @@
...
+      () => Promise.all(dependentIDs.map(id => elementsSource.get(id))),
+      'getting %d dependents from elements source',
+      dependentIDs.length,
+    )
+  ).filter(isElement)
+
+  log.debug('found %d dependents of %d elements', dependents.length, elemIDs.length)
+  if (dependentIDs.length !== dependents.length) {
+    log.warn(
+      'there is a mismatch between the num of requested dependent IDs and the num of dependents in the elements source. missing elements: %s',
```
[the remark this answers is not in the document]
dantal4: I actually prefer using `${list}` in this case so the result will be: 'elemId1,elemId2,elemId3,elemId4...' .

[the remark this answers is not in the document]
ori-moisis: my performance brain does not allow me to support this style of logging (since it does all the string manipulation calculations and memory allocations if the log level is not active) but I know it is a lost battle...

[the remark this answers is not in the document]
dantal4: 😮

PR #6670, opened 2024-10-12
SALTO-6737: Limit the number of users fetched in Okta

shir-reifenberg:
We want to limit the number of users fetched to 200,000

---

Set the default max users number to 200,000, when exceeding this, the pagination func will throw and this error will be caught by the custom error handler that will convert the error into fetch warning.
The max number is configurable through the adapter config, and can also be set to `-1` to cancel this limitation.

---
_Release Notes_: 

_Okta adapter_:
- Add safety to avoid fetching users in account with more than 200,000 users

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/fetch/request/pagination/pagination_functions.ts
+++ b/packages/adapter-components/src/fetch/request/pagination/pagination_functions.ts
@@ -13,13 +13,16 @@ import { collections } from '@salto-io/lowerdash'
 import { PaginationFunction } from '../../../definitions/system/requests/pagination'
 import { DATA_FIELD_ENTIRE_OBJECT } from '../../../definitions'
 import { ResponseValue } from '../../../client'
+import { MaxResultsExceeded } from '../../errors'
 
 const log = logger(module)
 
-const getItems = (value: ResponseValue | ResponseValue[], dataField: string): unknown[] =>
+export const getItems = (value: ResponseValue | ResponseValue[], dataField: string): unknown[] =>
   collections.array
     .makeArray(value)
-    .map(item => (dataField === DATA_FIELD_ENTIRE_OBJECT ? item : _.get(item, dataField)))
+    .flatMap(item => (dataField === DATA_FIELD_ENTIRE_OBJECT ? item : _.get(item, dataField)))
+
+const defaultGetItems = (value: ResponseValue | ResponseValue[]): unknown[] => getItems(value, DATA_FIELD_ENTIRE_OBJECT)
 
 /**
  * Make paginated requests using the specified pagination field
@@ -297,4 +300,30 @@ export const tokenPagination = ({
     ]
   }
   return nextPageTokenPages
+}
+
+export const getPaginationWithLimitedResults = ({
+  maxResultsNumber,
+  paginationFunc,
+  getItemsFunc = defaultGetItems,
+}: {
+  maxResultsNumber: number
+  paginationFunc: PaginationFunction
+  getItemsFunc?: (value: ResponseValue | ResponseValue[]) => unknown[]
+}): PaginationFunction => {
+  const UNLIMITED_RESULTS = -1
+  let totalResults = 0
+  log.debug('creating pagination function with max results: %d', maxResultsNumber)
+  const paginationWithLimitedResults: PaginationFunction = args => {
+    if (maxResultsNumber !== UNLIMITED_RESULTS) {
+      const currentPageResults = getItemsFunc(args.responseData).length
+      totalResults += currentPageResults
+      if (totalResults > maxResultsNumber) {
+        log.error('reached max results for endpoint %s, stopping pagination', args.endpointIdentifier.path)
+        throw new MaxResultsExceeded({ endpoint: args.endpointIdentifier.path, maxResults: maxResultsNumber })
+      }
+    }
+    return paginationFunc(args)
+  }
+  return paginationWithLimitedResults
 }
--- a/packages/okta-adapter/src/definitions/fetch/fetch.ts
+++ b/packages/okta-adapter/src/definitions/fetch/fetch.ts
@@ -1256,7 +1256,7 @@ const createCustomizations = ({
           () =>
           ({ error }) => {
             if (error instanceof fetchUtils.errors.MaxResultsExceeded) {
-              const message = `The number of users fetched exceeded the maximum allowed: ${error.maxResults}. Consider excluding this type or filtering users by specific status.`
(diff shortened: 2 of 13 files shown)


packages/adapter-components/src/fetch/request/pagination/pagination_functions.ts
```diff
@@ -13,13 +13,16 @@ import { collections } from '@salto-io/lowerdash'
...
+import { MaxResultsExceeded } from '../../errors'
 
 const log = logger(module)
 
-const getItems = (value: ResponseValue | ResponseValue[], dataField: string): unknown[] =>
+export const getItems = (value: ResponseValue | ResponseValue[], dataField: string): unknown[] =>
   collections.array
     .makeArray(value)
-    .map(item => (dataField === DATA_FIELD_ENTIRE_OBJECT ? item : _.get(item, dataField)))
+    .flatMap(item => (dataField === DATA_FIELD_ENTIRE_OBJECT ? item : _.get(item, dataField)))
```
shir-reifenberg, line 23: I think this fixes a bug for scenarios when `value` is not an array
for example in this case -
```
{
   nextPage: '/foo/bar'
   value: [{ id: 'a'}, {id: 'b'}]
}
```

luckily this was only used in confluence and workato, and from what I checked the responses in both cases were array so it had no affect

Nurdok: Doesn't the `makeArray` call take care of that?

packages/adapter-components/src/fetch/request/pagination/pagination_functions.ts
```diff
@@ -298,3 +301,29 @@ export const tokenPagination = ({
...
+}): PaginationFunction => {
+  const UNLIMITED_RESULTS = -1
+  let totalResults = 0
+  log.debug('creating pagination function with max results: %d', maxResultsNumber)
+  const paginationWithLimitedResults: PaginationFunction = args => {
+    if (maxResultsNumber !== UNLIMITED_RESULTS) {
+      const currentPageResults = getItemsFunc(args.responseData).length
+      totalResults += currentPageResults
+      if (totalResults > maxResultsNumber) {
+        log.error('reached max results for endpoint %s, stopping pagination', args.endpointIdentifier.path)
```
Nurdok, line 322: If the overall result here is a fetch warning, I'd say the log should be at warning level, too. This helps keeps the logs cleaner.

packages/okta-adapter/src/definitions/fetch/fetch.ts
```diff
@@ -1249,7 +1249,29 @@ const createCustomizations = ({
...
     ],
-    resource: { directFetch: true },
+    resource: {
+      directFetch: true,
+      onError: {
+        custom:
+          () =>
+          ({ error }) => {
+            if (error instanceof fetchUtils.errors.MaxResultsExceeded) {
+              const message = `The number of users fetched exceeded the maximum allowed: ${error.maxResults}. Consider excluding this type or filtering users by specific status.`
```
Nurdok, line 1259:
```suggestion
              const message = `The number of users fetched exceeded the maximum allowed: ${error.maxResults}. Consider excluding this type or filtering users by a specific status.`
```

packages/okta-adapter/src/definitions/fetch/fetch.ts
```diff
@@ -1249,7 +1249,29 @@ const createCustomizations = ({
...
+              return {
+                action: 'customSaltoError',
+                value: {
+                  message,
+                  detailedMessage: message,
+                  severity: 'Warning',
+                },
+              }
+            }
+            return { action: 'failEntireFetch', value: false }
```
Nurdok, line 1269: Other errors shouldn't fail the fetch? This is a separate behavior change, right?

packages/okta-adapter/src/definitions/requests/pagination.ts
```diff
@@ -7,14 +7,27 @@
...
+  config: OktaUserConfig,
+): definitions.ApiDefinitions<OktaOptions>['pagination'] => ({
   cursorHeader: {
     funcCreator: () => cursorHeaderPagination({ pathChecker: defaultPathChecker }),
   },
   cursor: {
     funcCreator: () => cursorPagination({ paginationField: 'nextMappingsPageUrl', pathChecker: defaultPathChecker }),
   },
-}
+  limitedCursorHeader: {
```
Nurdok, line 26: This is specific to `User` (it uses `maxUsersResults`) so IMO it should have 'user' in the name.

Nurdok (approved):

packages/adapter-components/src/fetch/request/pagination/pagination_functions.ts
shir-reifenberg: mmm IMO it should be `error`, cause the user's intention was to fetch users but we weren't able to do so..
We usually return fetch warning and not errors (I think there's might be a code that fails the fetch if we create an fetch warning with error severity, but not sure about it) but I wouldn't expect this log to be noisy and I definitely would want this to pop up when looking on logs for env
WDYT?

packages/adapter-components/src/fetch/request/pagination/pagination_functions.ts
shir-reifenberg: `makeArray` would put the entire response in an array
so 
 `{ nextPage: '/foo/bar', value: [{ id: 'a'}, { id: 'b'}] }` will be -> `[{ nextPage: '/foo/bar', value: [{ id: 'a'}, { id: 'b'}] }]`
```
and `getItems` will return `[[{ id: 'a' }, { id: 'b' }]]` instead of `[{ id: 'a' }, { id: 'b' }]`

packages/okta-adapter/src/definitions/fetch/fetch.ts
shir-reifenberg: Good catch, the default should be fallback to [this](https://github.com/salto-io/salto/blob/main/packages/adapter-components/src/fetch/element/element.ts#L114), just log error (we have a monitor for it, but the default shouldn't fail fetch IMO as it's quite radical).
but the current implementation will just ignore any other error

I changed the `custom` handler definition so it can return undefined and fallback to the default behavior, @Nurdok @shoshana-houri  - WDYT ?

shoshana-houri: @shir-reifenberg Cool! Makes sense to me :)

PR #6671, opened 2024-10-13
SALTO-6697 remove nacl files referenced index

dantal4:
This index isn't needed anymore.

---

---
_Release Notes_: 
Workspace:
- Remove nacl files referenced index

---
_User Notifications_: 
None

### DIFF
--- a/packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
+++ b/packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
@@ -15,18 +15,11 @@ import {
   ElemID,
   isAdditionChange,
   isAdditionOrModificationChange,
-  isContainerType,
   isElement,
-  isIndexPathPart,
-  isInstanceElement,
   isModificationChange,
-  isObjectType,
   isObjectTypeChange,
-  isReferenceExpression,
   isRemovalOrModificationChange,
   isStaticFile,
-  isVariable,
-  placeholderReadonlyElementsSource,
   StaticFile,
   toChange,
   Value,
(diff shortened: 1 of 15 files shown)


packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
```diff
@@ -154,110 +146,53 @@ const createFilenameToElementIDFullNamesMapping = (
...
+    }),
   )
 
-export const toParsedNaclFile = (naclFile: NaclFile, parseResult: parser.ParseResult): SyncParsedNaclFile => {
-  const { elements, errors, sourceMap } = parseResult
+  return Array.from(staticFiles)
+}
 
-  let referenced: string[]
+const loadStaticFiles = (elements: Element[]): (() => string[]) => {
```
ori-moisis, line 169: it looks like this is removing an optimization that the previous code had where if you called the method more than once, it would not calculate everything again.
is this not needed anymore? was it only here because it would find the references and static files together?

also, load is kind of a strange name here
maybe `getStaticFiles` or `createGetStaticFilesFunc` or something like that?

ori-moisis (approved):

packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
dantal4: it's still caching the `staticFIles`:
```typescript
const getStaticFiles = (): string[] => {
  if (staticFiles === undefined) {
    staticFiles = getElementsStaticFiles(elements)
  }
  return staticFiles
}
return getStaticFiles 
```

ori-moisis: right, sorry, got lost in all the diffs

dantal4:
Merging this PR, after verifying that the [dependents calculation](https://github.com/salto-io/salto/blob/c4963ea1a916d61b21302094a2e3437b30da8f89/packages/workspace/src/workspace/dependents.ts#L137-L139) using the `referencesSource` index takes a reasonable amount of time, that's also quicker than the calculation using the `getElementReferencedFiles` API.

PR #6673, opened 2024-10-13
SALTO-6667 Zendesk trigger skill priority

adfineberg:
Support a new API format of skills within Zendesk triggers.

---

---
_Release Notes_: 
_Zendesk Adapter:_
* set_skills or add_skills under Trigger actions will now have a priority of required or optional as per https://developer.zendesk.com/documentation/ticketing/using-the-zendesk-api/setting-skill-priority-with-skills-in-trigger-action/

---
_User Notifications_: 
_Zendesk Adapter:_
* Trigger nacls with set_skills or add_skills will now include a priority key if this is received from Zendesk.

### DIFF
--- a/packages/zendesk-adapter/src/filters/deploy_trigger_skills.ts
+++ b/packages/zendesk-adapter/src/filters/deploy_trigger_skills.ts
@@ -20,7 +20,7 @@ const restoreTriggerSkillToApi = async (
     .forEach((action: { value?: string; priority?: string }) => {
       if ('priority' in action && 'value' in action) {
         const { value, priority } = action as { value: string; priority: string }
-        action.value = `${value}#${priority === 'optional' ? '0' : '1'}`
+        action.value = `${value}#${priority === 'optional' ? '1' : '0'}`
         skillMapping[action.value] = { value, priority }
         delete action.priority
       }
--- a/packages/zendesk-adapter/src/filters/deploy_trigger_skills.ts
+++ b/packages/zendesk-adapter/src/filters/deploy_trigger_skills.ts
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2024 Salto Labs Ltd.
+ * Licensed under the Salto Terms of Use (the "License");
+ * You may not use this file except in compliance with the License.  You may obtain a copy of the License at https://www.salto.io/terms-of-use
+ *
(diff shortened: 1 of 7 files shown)


packages/zendesk-adapter/src/filters/deploy_trigger_skills.ts
```diff
@@ -0,0 +1,65 @@
...
+  instance: InstanceElement,
+  skillMapping: Record<string, { value: string; priority: string }>,
+): Promise<void> => {
+  instance.value?.actions
+    .filter((action: unknown) => TRIGGER_SKILL_FIELDS.includes(_.get(action, 'field')))
+    .forEach((action: { value?: string; priority?: string }) => {
+      if ('priority' in action && 'value' in action) {
+        const { value, priority } = action as { value: string; priority: string }
+        action.value = `${value}#${priority === 'optional' ? '1' : '0'}`
+        skillMapping[action.value] = { value, priority }
```
edenhassid, line 24: can you have duplicates here? meaning that `action.value` will appear twice?

packages/zendesk-adapter/src/filters/deploy_trigger_skills.ts
```diff
@@ -0,0 +1,65 @@
...
+          .filter(instance => instance.elemID.typeName === TRIGGER_TYPE_NAME)
+          .map(instance => restoreTriggerSkillToApi(instance, skillMapping)),
+      )
+    },
+    onDeploy: async (changes: Change<InstanceElement>[]): Promise<void> =>
+      changes
+        .map(getChangeData)
+        .filter(instance => instance.elemID.typeName === TRIGGER_TYPE_NAME && Array.isArray(instance.value?.actions))
+        .forEach(async instance => {
+          instance.value.actions = instance.value.actions.map((action: { value: string }) => {
```
edenhassid, line 51: can you check that action is of the correct type? that it has a value?

edenhassid (commented):
looks good! are we sure that this is only possible for triggers and no other types?

packages/zendesk-adapter/src/filters/deploy_trigger_skills.ts
adfineberg: It's ok, they're always the same mapping

edenhassid (approved):

PR #6674, opened 2024-10-14
SALTO-6753: Supporting workflows in SFDX

yelly:
Since the SFDX code overrides entire workflows on any workflow change, we need to collect all workflow sub-instances on any workflow change.

---

None.

---
_Release Notes_: 
None.

---
_User Notifications_: 
None.

### DIFF
--- a/packages/salesforce-adapter/src/filters/workflow.ts
+++ b/packages/salesforce-adapter/src/filters/workflow.ts
@@ -236,7 +236,7 @@ const filterCreator: FilterCreator = ({ config, client }) => {
       originalWorkflowChanges = await groupByAsync(allWorkflowRelatedChanges, getWorkflowApiName)
 
       if (client === undefined) {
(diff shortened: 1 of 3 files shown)


packages/salesforce-adapter/src/filters/workflow.ts
```diff
@@ -218,6 +235,24 @@ const filterCreator: FilterCreator = () => {
 
       originalWorkflowChanges = await groupByAsync(allWorkflowRelatedChanges, getWorkflowApiName)
 
+      if (client === undefined) {
+        // In the SFDX flow we need to get the entire flow to avoid part of it being dropped.
```
amitgoldarb, line 239: Nit: this took me an extra second to orient myself
```suggestion
        // In the SFDX flow we need to get the entire workflow to avoid part of it being dropped.
```

amitgoldarb (approved):
Thanks for this one as well

PR #6676, opened 2024-10-14
SALTO-6621: Skipping additional profile permissions in SFDX flow

yelly:
We add some extra permissions to profiles when deploying to make sure users have access, but we don't want to do this in the SFDX dump flow.

---

None.

---
_Release Notes_: 
None.

---
_User Notifications_: 
None.


amitgoldarb (approved):
Thanks for this one!

PR #6679, opened 2024-10-15
SALTO-6759 omit password from configurationProfile instances

omrifarkash:
---

---
_Release Notes_: 
_Jamf Adapter_:
- omit password from configurationProfile instances

---
_User Notifications_: 
_Jamf Adapter_:
- omit password from configurationProfile instances


kami-delaroz (approved):
💯

PR #6681, opened 2024-10-15
SALTO-6783: Do not return remote map cache if it was not closed

ori-moisis:
If closeRemoteMapsOfLocation is directly called but no remote map was opened we should not try to close the cache of that location.
This is mostly to avoid the redundant warning log

---

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


tamtamirr (approved):
Code LGTM! Can you please add UT for this scenario?

PR #6683, opened 2024-10-20
SALTO-6787 Return a detailed change on the whole primitive type when …

dantal4:
…its primitive was changed

In detailedCompare of a primitive type, return a change on the whole type if the primitive was changed. This is similar to a change in field refType, or an objectType's metaType. I ran into this case in the tests, were we change a primitive type's primitive and expect validations to be calculated on an instance that have a field of that primitive.

---

---
_Release Notes_: 
Adapter Utils:
- Return a detailed change on the whole primitive type when its primitive was changed

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-utils/src/compare.ts
+++ b/packages/adapter-utils/src/compare.ts
@@ -269,9 +269,11 @@ export const detailedCompare = (
 
   // A special case to handle type changes.
   // In fields, we have to modify the whole field.
+  // For primitive types, we have to modify the entire type.
   // For object type meta types, we have to modify the entire object.
   if (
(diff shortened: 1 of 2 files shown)


ori-moisis (approved):
did you check that this works in `updateNaclFiles` as well?
maybe it would be good to add a test for that as well?

also might be worth at least manually testing that the state update flow works well with this kind of change now

dantal4:
> did you check that this works in `updateNaclFiles` as well? maybe it would be good to add a test for that as well?
> 
> also might be worth at least manually testing that the state update flow works well with this kind of change now

I tested it manually in fetch, deploy & restore. it works well in all of them (which doesn't work in the previous behavior)

PR #6684, opened 2024-10-20
SALTO-6747: Change InitFolder params

dorsev:
---

_Release Notes_: 
Core: Change the parameters of `isInitializedFolder` and `initFolder` not to include `workspace`

---

_User Notifications_:

### DIFF
--- a/packages/cli/src/commands/adapter_format.ts
+++ b/packages/cli/src/commands/adapter_format.ts
@@ -171,7 +171,7 @@ export const syncWorkspaceToFolderAction: WorkspaceCommandAction<SyncWorkspaceTo
 }) => {
   const { accountName, toDir, force } = input
 
-  const initializedResult = await isInitializedFolder({ workspace, accountName, baseDir: toDir })
+  const initializedResult = await isInitializedFolder({ adapterName: accountName, baseDir: toDir })
   if (initializedResult.errors.length > 0) {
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/adapter_format.ts
```diff
@@ -171,7 +171,7 @@ export const syncWorkspaceToFolderAction: WorkspaceCommandAction<SyncWorkspaceTo
 }) => {
   const { accountName, toDir, force } = input
```
ori-moisis, line 172: if there is no translation done from account name to adapter, then the input should be adapter name

dorsev: This fails, as syncWorkspaceToFolder needs `accountName`

dorsev: line 195.

should I send `result = await syncWorkspaceToFolder({ workspace, accountName: adapterName`

there?

dorsev: I've Implemented your last suggestion.

dorsev: LMK if this makes sense

dorsev:
Thank you for the super quick review cycles @ori-moisis 
I'm ready for another round

ori-moisis (approved):

PR #6685, opened 2024-10-21
SALTO-6793: Fix SFDX broken UT

tamtamirr:
Wrong invocation of `log.error`. Was called with `Error` object and not string as expected.

---

_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/src/sfdx_parser/project.ts
+++ b/packages/salesforce-adapter/src/sfdx_parser/project.ts
@@ -7,6 +7,7 @@
  */
 
 import path from 'path'
+import { inspectValue } from '@salto-io/adapter-utils'
 import { logger } from '@salto-io/logging'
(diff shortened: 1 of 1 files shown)


amitgoldarb (approved):

packages/salesforce-adapter/src/sfdx_parser/project.ts
```diff
@@ -62,7 +63,7 @@ export const createProject: InitFolderFunc = async ({ baseDir }) => {
     await templateService.create(TemplateType.Project, opts)
     return { errors: [] }
   } catch (error) {
-    log.error(error)
+    log.error('Error occurred when creating SFDX project: %s', inspectValue(error))
```
ori-moisis, line 66: what about the exact same code in line 34?
is it not a problem there?

PR #6686, opened 2024-10-21
SALTO-6791: (Bug-fix) FlowDefinition instances have pending changes when Flow type is excluded

tamtamirr:
(Bug-fix) FlowDefinition instances have pending changes when Flow type is excluded

---

The previous implementation made the FlowDefinition hidden only if the `Flow` type was included. We should hide them regardless.

---
_Release Notes_: 
Salesforce Adapter:
- (Bug-fix) FlowDefinition instances have pending changes when Flow type is excluded

---
_User Notifications_: 
_None_


ori-moisis:
why do we retrieve FlowDefinition if Flow is excluded?

tamtamirr:
@ori-moisis As FlowDefinition instances affect the `flowAccesses` section in Profiles. Theoretically, environment can choose to manage `Flow Accesses` in Profiles without managing the Flow instances.

packages/salesforce-adapter/src/fetch_profile/metadata_query.ts
```diff
@@ -104,7 +106,9 @@ export const buildMetadataQuery = ({ fetchParams }: BuildMetadataQueryParams): M
         })
       return acc
     },
-    {},
+    {
+      [FLOW_DEFINITION_METADATA_TYPE]: FLOW_METADATA_TYPE,
```
yelly, line 110: Why not include this in `TYPE_TO_NESTED_TYPES`? As far as I can tell, this is the only place it's used.

yelly: Also, while we're at it, why are we inverting this every time `buildMetadataQuery` is called and not just once in a const next to the definition of `TYPE_TO_NESTED_TYPES`?

tamtamirr: Re 1st comment: This is used in the `last_change_date_of_types_with_nested_instances`, which `Flow & FlowDefinition` are not handled in.

Re 2nd comment: Exported to a const

yelly (approved):

PR #6690, opened 2024-10-28
SALTO-6779: remove boolean query param from adapter component client interface

shayc331:
_remove boolean query param from adapter component client interface_

---

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


netama (approved):
thanks!

PR #6693, opened 2024-10-29
SALTO-6812: Using empty SFDX project template

yelly:
The standard template is a bit more spammy than we would like.

---

_Salesforce adapter_: 
* Moved to using the empty template for creating SFDX projects instead of the standard template which has some extra files.

---

_User Notifications_: 
None.


dorsev (approved):
Big Like. thanks for this pr 🙏

PR #6697, opened 2024-10-30
SALTO-6817: Fix AddJsmProjectValidator to address only jira cloud 

idoamit8:
In this PR I fixed AddJsmProjectValidator to address only jira cloud 

---

During deployment of a JSM project to our Jira DC instance, the operation failed due to the addJsmProjectValidator being incorrectly applied to both DC and Cloud environments. After disabling the validator, the deployment succeeded.
The validator logic needs to be updated to address only Cloud environments.

---
_Release Notes_: 
* [SALTO-6817](https://salto-io.atlassian.net/browse/SALTO-6817) - Users will be able to deploy JSM projects (Only the projects, without subTypes) to their DC instances. 
---
_User Notifications_: 
None

[SALTO-6817]: https://salto-io.atlassian.net/browse/SALTO-6817?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/jira-adapter/src/change_validators/adding_jsm_project.ts
+++ b/packages/jira-adapter/src/change_validators/adding_jsm_project.ts
@@ -14,31 +14,33 @@ import {
 } from '@salto-io/adapter-api'
 import { PROJECT_TYPE, SERVICE_DESK } from '../constants'
 import { hasJiraServiceDeskLicense } from '../utils'
+import JiraClient from '../client/client'
 
 /*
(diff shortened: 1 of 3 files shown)


shayc331 (approved):
💯 
can you please check if there are more JSM related change validators or filters that we should deactivate when using Jira DC?

PR #6698, opened 2024-11-03
SALTO-6654: Core: Allow RemoteMap cache to serve concurrent reads for the same key

Nurdok:
### Changes in this PR

Before this PR, concurrent reads for the same key (that isn't in the cache) would result in each `get` call to retrieve the value independently, resulting in 0% cache hits for all reads. This PR modifies caching behavior to so that the first `get` will immediately store a `Promise` in the case and subsequent (concurrent) reads will immediately find the promise and wait for the result, resulting in more cache hits.

In addition, this change includes caching missing values (e.g., returning `undefined` from the DB) as these can help avoid DB operations as well.

See attached ticket for more details.

### Analysis
I ran some side by side salto runs, comparing logged countersfor the old vs the new implementation of the cache. I did this with the following scenarios:

1. A small Salesforce run that’s similar to the multi_env e2e test: running `fetch` on a second environment when there are already elements from an existing environment
2. Another run of the above scenario, reusing the cache built from the first fetch
3. A larger fetch on an Okta workspace, ran after `workspace clean`

Findings:

- Adding caching for missing values (`undefined`) resulted in 8-12% increase in cache hits for a fetch with an clean cache
- Reusing an existing cache with cached missing values resulted in ~17% increase in cache hits
- When I removed counting of missing value cache hits (keep counting only existing key’s cache - e.g., disable missing value cache hits from interfering with concurrent read improvements):
    - For the first two scenarios, counters were exactly the same between the implementations (and between runs). This means that there were probably no concurrent reads and the run was pretty deterministic. This shows (IMO) that the new implementation doesn’t contain a regression in correctness.
    - For the third scenario, where the fetch was larger, cache hits were *decreased* when not counting missing value hits. This seems to happen because with the added cached `undefined` values, we reach the cache limit and reduce cache hits for keys that were removed from the LRU cache - I showed this by increasing the cache size such that we don’t reach the limit. When that happens, counters are identical again.

In summary, this change seems to be valid and correct and improves cache hits by caching missing values. We might consider increasing the cache size as cached missing values don’t take up a lot of memory (the cache is more “sparse” for lack of a better word).

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/core/test/workspace/local/remote_map/remote_map.test.ts
```diff
@@ -124,6 +124,42 @@ describe('test operations on remote db', () => {
...
+      expect(await remoteMap.get(elements[0].elemID.getFullName())).toEqual(expectedElementFromMap)
+      expect(await remoteMap.get(elements[0].elemID.getFullName())).toEqual(expectedElementFromMap)
+
+      const locationInfo = remoteMapLocations.get(DB_LOCATION)
+      expect(locationInfo.counters.LocationCacheMiss.value()).toEqual(1)
+      expect(locationInfo.counters.LocationCacheHit.value()).toEqual(3)
+      expect(locationInfo.cache.length).toEqual(1)
+    })
+    it('should read from cache in concurrent reads', async () => {
+      await Promise.all([
```
ori-moisis, line 139: nit - since it shouldn't be too difficult, maybe check that the result is correct here as well?

packages/core/src/local-workspace/remote_map/remote_map.ts
```diff
@@ -566,7 +566,7 @@ export const createRemoteMapCreator = (
       value,
     }: remoteMap.RemoteMapEntry<string>): Promise<remoteMap.RemoteMapEntry<T, K>> => {
       const cacheKey = keyToTempDBKey(key)
-      const cacheValue = locationCache.get(cacheKey) as T | undefined
+      const cacheValue = await (locationCache.get(cacheKey) as Promise<T | undefined>)
```
ori-moisis, line 569: this cast is incorrect, the type here is actually `Promise<T | undefined> | undefined` (the second `undefined` is because it is possible that the `cacheKey` is not in the cache...

it might be better to leave the cast as it was and put it outside the `await`
```suggestion
      const cacheValue = (await locationCache.get(cacheKey)) as T | undefined
```

packages/core/src/local-workspace/remote_map/remote_map.ts
```diff
@@ -618,42 +618,56 @@ export const createRemoteMapCreator = (
           return
         }
         if (locationCache.has(keyToTempDBKey(key))) {
-          statCounters.LocationCacheHit.inc()
-          statCounters.RemoteMapHit.inc()
-          isNamespaceEmpty = false
-          resolve(locationCache.get(keyToTempDBKey(key)) as T)
-          return
+          const cached = locationCache.get(keyToTempDBKey(key)) as Promise<T | undefined>
```
ori-moisis, line 621: I think maybe the `LocationCache` type needs to be updated to know that it will always store promises, these incorrect casts are misleading...
e.g, change
```typescript
export class LocationCache extends LRU<string, unknown> {
```
to
```typescript
export class LocationCache extends LRU<string, Promise<unknown>> {
```

packages/core/src/local-workspace/remote_map/remote_map.ts
```diff
@@ -618,42 +618,56 @@ export const createRemoteMapCreator = (
...
+            statCounters.LocationCacheHit.inc()
+            statCounters.RemoteMapHit.inc()
+            resolve(
+              cached.then(value => {
+                if (value !== undefined) {
+                  isNamespaceEmpty = false
+                }
+                return value
+              }),
+            )
```
ori-moisis, line 632: I know this is not your doing initially, but I think this mix of using `async` functions in most places and using `new Promise(resolve => ...)` and `cached.then(...` here is making the code less readable (and this PR is adding more complexity to this)

I will make the effort to try and follow this code as it is here, but maybe it would be worth seeing if this code can be refactored to using async functions instead of this `new Promise` style?
(probably worth starting this out as a separate PR on top of this one, but if it looks good, maybe we'll merge that one directly)

packages/core/src/local-workspace/remote_map/remote_map.ts
```diff
@@ -618,42 +618,56 @@ export const createRemoteMapCreator = (
           return
         }
         if (locationCache.has(keyToTempDBKey(key))) {
-          statCounters.LocationCacheHit.inc()
-          statCounters.RemoteMapHit.inc()
-          isNamespaceEmpty = false
-          resolve(locationCache.get(keyToTempDBKey(key)) as T)
-          return
+          const cached = locationCache.get(keyToTempDBKey(key)) as Promise<T | undefined>
+          if (cached !== undefined) {
```
ori-moisis, line 622: not loving the double nesting here of using the `locationCache.has(keyToTempDBKey(key)` condition and then also having another condition on `cached !== undefined`.
if we need the second one anyway, lets remove the first check...

packages/core/src/local-workspace/remote_map/remote_map.ts
```diff
@@ -618,42 +618,56 @@ export const createRemoteMapCreator = (
...
+            statCounters.LocationCacheHit.inc()
+            statCounters.RemoteMapHit.inc()
```
ori-moisis, line 624: IIUC, `RemoteMapHit` is supposed to mean that we found the value in the map, so incrementing it here is incorrect (the resolved value could still end up being `undefined`)

packages/core/src/local-workspace/remote_map/remote_map.ts
```diff
@@ -618,42 +618,56 @@ export const createRemoteMapCreator = (
...
-          if (error) {
-            if (wasClearCalled) {
-              statCounters.RemoteMapMiss.inc()
-              resolve(undefined)
-              return
-            }
-            // eslint-disable-next-line @typescript-eslint/no-misused-promises
-            persistentDB.get(keyToDBKey(key), async (innerError, innerValue) => {
-              if (innerError) {
+        const cachePromise = new Promise<T | undefined>(resolveInner => {
```
ori-moisis, line 637: this time it is your doing, we should not mix this style, use async functions instead.

in this example (which in unfortunate since we are already in a function called `getImpl`), this would mean having a separate async function (with the very unfortunate name of `getImplInner` or `getInnerImpl` or `innerGetImpl` or something along those lines...) that would take care of actually getting the value.

then the code here would be
```typescript
const cachePromise = getImplInner(...)
locationCache.set(keyToTempDBKey(key), cachePromise)
resolve(cachePromise)
```

packages/core/src/local-workspace/remote_map/remote_map.ts
```diff
@@ -618,42 +618,56 @@ export const createRemoteMapCreator = (
...
+            statCounters.RemoteMapHit.inc()
+            resolve(
+              cached.then(value => {
+                if (value !== undefined) {
+                  isNamespaceEmpty = false
+                }
+                return value
+              }),
+            )
+            return
```
ori-moisis, line 633: another reason to avoid the double condition (checking both `.has` and that `.get` didn't return undefined).
because of that double condition, it appears as though there can be a scenario where we entered the `if` and we won't `return` here (which I really hope is not actually possible)

packages/core/src/local-workspace/remote_map/remote_map.ts
```diff
@@ -786,15 +800,16 @@ export const createRemoteMapCreator = (
       },
       delete: deleteImpl,
       has: async (key: string): Promise<boolean> => {
-        if (locationCache.has(keyToTempDBKey(key))) {
+        if ((await getImpl(key)) !== undefined) {
```
ori-moisis, line 803: calling `getImpl` here would cause a significant degradation in performance since `getImpl` fully reads and deserializes the value, which takes significantly longer than just checking if the value exists

packages/core/src/local-workspace/remote_map/remote_map.ts
```diff
@@ -786,15 +800,16 @@ export const createRemoteMapCreator = (
...
           isNamespaceEmpty = false
           return true
         }
         const hasKeyImpl = async (k: string, db: rocksdb): Promise<boolean> =>
           new Promise(resolve => {
             // eslint-disable-next-line @typescript-eslint/no-misused-promises
             db.get(k, async (error, value) => {
-              resolve(!error && value !== undefined)
+              const ret = !error && value !== undefined
+              resolve(ret)
```
ori-moisis, line 812: I assume this was for debugging?
if you want to keep this change, please give it a better name

also, if you are already changing things here, this function shouldn't by `async` (it has no `await` in it...)

packages/core/src/local-workspace/remote_map/remote_map.ts
```diff
@@ -611,50 +613,59 @@ export const createRemoteMapCreator = (
...
+      valueFromDb = await getAwaitable(tmpDB, keyToTempDBKey(key))
+      if (valueFromDb === undefined) {
+        if (wasClearCalled) {
+          statCounters.RemoteMapMiss.inc()
+          locationCache.del(keyToTempDBKey(key))
+          return undefined
+        }
+        valueFromDb = await getAwaitable(persistentDB, keyToDBKey(key))
+      }
+      if (valueFromDb !== undefined) {
```
ori-moisis, line 641: in the previous code the counter for `RemoteMapMiss` would be incremented if the value was not found, seems to be missing in the new code.

maybe make some kind of run with the old code and the new code and compare the counters at the end? (I guess they don't have the be identical, but they should be pretty close...)

packages/core/src/local-workspace/remote_map/remote_map.ts
```diff
@@ -611,50 +613,59 @@ export const createRemoteMapCreator = (
...
       })
+    // Retrieve a value from the DBs directly, ignoring the cache.
+    // If the value is not found in the temporary db, it will be retrieved from the persistent db.
+    const getFromDb = async (key: string): Promise<T | undefined> => {
+      let valueFromDb: T | undefined
+      valueFromDb = await getAwaitable(tmpDB, keyToTempDBKey(key))
+      if (valueFromDb === undefined) {
+        if (wasClearCalled) {
+          statCounters.RemoteMapMiss.inc()
+          locationCache.del(keyToTempDBKey(key))
```
ori-moisis, line 636: this feels wrong
if `clear` was called, it would have cleared the location cache as well.

if the intention of this to avoid having promises that resolve to `undefined` in the cache, then this is just one (fairly rare IIUC) use case where this can happen, so having it only here is probably not right...

packages/core/src/local-workspace/remote_map/remote_map.ts
```diff
@@ -787,10 +798,13 @@ export const createRemoteMapCreator = (
       delete: deleteImpl,
       has: async (key: string): Promise<boolean> => {
         if (locationCache.has(keyToTempDBKey(key))) {
-          isNamespaceEmpty = false
-          return true
+          // If the key is already in the cache, then we won't incur additional cost by calling getImpl here.
+          if ((await getImpl(key)) !== undefined) {
+            isNamespaceEmpty = false
+            return true
+          }
```
ori-moisis, line 805: if we really waiting for a pending `get`, shouldn't we be able to fully answer if the key exists or not?

naively this would be returning `false` if we got `undefined`, I am not sure if `undefined` is a valid value in a remote map though (and if it is, does it mean `has` should answer `true` on a key that has a value `undefined`?).
might be worth checking what the current behavior is...

ori-moisis: if you think it is fine (I tend to agree), then we should be able to answer `false` if we got `undefined` here, right?

packages/core/src/local-workspace/remote_map/remote_map.ts
```diff
@@ -611,50 +613,57 @@ export const createRemoteMapCreator = (
       const opts = { ...(iterationOpts ?? {}), keys: true, values: false }
       return awu(getDataIterableWithPages(opts)).map(entries => entries.map(entry => entry.key as K))
     }
-    const getImpl = (key: string): Promise<T | undefined> =>
-      new Promise(resolve => {
-        if (delKeys.has(key)) {
-          resolve(undefined)
-          return
+    // Wrapper for the db.get function that is awaitable.
+    const getAwaitable = async (db: rocksdb, key: string): Promise<rocksdb.Bytes | undefined> => {
```
ori-moisis, line 617: AFAICT this function does not have to be defined in this scope, it can be moved to be a utility function (similar to `getOpenDBConnection`)

packages/core/src/local-workspace/remote_map/remote_map.ts
```diff
@@ -611,50 +613,57 @@ export const createRemoteMapCreator = (
...
+        if (error.message.includes('NotFound') === true) {
+          return undefined
         }
-        if (locationCache.has(keyToTempDBKey(key))) {
-          statCounters.LocationCacheHit.inc()
-          statCounters.RemoteMapHit.inc()
-          isNamespaceEmpty = false
-          resolve(locationCache.get(keyToTempDBKey(key)) as T)
-          return
+        throw error
```
ori-moisis, line 624: this is a little scary as a behavior change, I suggest we do this in two steps.
for now we maintain the old behavior and return `undefined` on any error like the old code did, but log if we ever get an error that isn't `NotFound`.
only after we see the new log doesn't actually happen I think we can be comfortable with changing the behavior to this one

wdyt?

packages/core/src/local-workspace/remote_map/remote_map.ts
```diff
@@ -611,50 +613,57 @@ export const createRemoteMapCreator = (
...
+          statCounters.RemoteMapMiss.inc()
+          return undefined
         }
-        statCounters.LocationCacheMiss.inc()
-        const resolveRet = async (value: Buffer | string): Promise<void> => {
-          const ret = await deserialize(value.toString())
-          locationCache.set(keyToTempDBKey(key), ret)
+        valueFromDb = await getAwaitable(persistentDB, keyToDBKey(key))
+      }
+      if (valueFromDb !== undefined) {
```
ori-moisis, line 640: nit - I would flip this condition (put the shorter branch of `return undefined` in the `if`)

packages/core/src/local-workspace/remote_map/remote_map.ts
```diff
@@ -787,10 +796,14 @@ export const createRemoteMapCreator = (
       delete: deleteImpl,
       has: async (key: string): Promise<boolean> => {
         if (locationCache.has(keyToTempDBKey(key))) {
-          isNamespaceEmpty = false
-          return true
+          // If the key is already in the cache, then we won't incur additional cost by calling getImpl here.
```
ori-moisis, line 799: I think that instead of this comment we could do this:
```typescript
const cachedPromise = locationCache.get(keyToTempDBKey(key))
if (cachedPromise !== undefined) {
  if ((await cachedPromise) === undefined) {
    return false
  }
  isNamespaceEmpty = false
  return true
}
```

that way we are not relying on the implementation of `getImpl`, and I don't feel like we are adding much code duplication here

packages/core/src/local-workspace/remote_map/remote_map.ts
```diff
@@ -787,10 +796,14 @@ export const createRemoteMapCreator = (
...
+            isNamespaceEmpty = false
+            return true
+          }
+          return false
         }
-        const hasKeyImpl = async (k: string, db: rocksdb): Promise<boolean> =>
+        const hasKeyImpl = (k: string, db: rocksdb): Promise<boolean> =>
           new Promise(resolve => {
             // eslint-disable-next-line @typescript-eslint/no-misused-promises
             db.get(k, async (error, value) => {
```
ori-moisis, line 809: we can just remove the `async` here, right?

ori-moisis (commented):
code looks good, not approving just because of the one comment about the behavior change that I am a little scared of

packages/core/src/local-workspace/remote_map/remote_map.ts
Nurdok: Done

ori-moisis (approved):

ori-moisis (approved):

PR #6701, opened 2024-11-03
SALTO-6171: Create "Info" message when adding a user in status PROVISIONED (Okta)

shir-reifenberg:
Indicate admins that added users will receive an email when adding those using Salto.

---

---
_Release Notes_: 

_Okta adapter_:
- Add a change validator to indicate that users added in status provisioned will receive an email to complete activation.  

---
_User Notifications_: 
None

### DIFF
--- a/packages/okta-adapter/src/change_validators/index.ts
+++ b/packages/okta-adapter/src/change_validators/index.ts
@@ -36,6 +36,7 @@ import { disabledAuthenticatorsInMfaPolicyValidator } from './disabled_authentic
 import { oidcIdentityProviderValidator } from './oidc_idp'
 import { everyoneGroupAssignments } from './everyone_group_assignments'
 import { emailDomainAdditionValidator } from './email_domain_addition'
+import { provisionedUserAdditions } from './provisioned_user_addition'
 import OktaClient from '../client/client'
 import {
   API_DEFINITIONS_CONFIG,
@@ -114,6 +115,7 @@ export default ({
     oidcIdentityProvider: oidcIdentityProviderValidator,
     everyoneGroupAssignments,
     emailDomainAddition: emailDomainAdditionValidator,
+    provisionedUserAdditions,
   }
 
   return createChangeValidator({
--- a/packages/okta-adapter/src/change_validators/provisioned_user_addition.ts
+++ b/packages/okta-adapter/src/change_validators/provisioned_user_addition.ts
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2024 Salto Labs Ltd.
+ * Licensed under the Salto Terms of Use (the "License");
+ * You may not use this file except in compliance with the License.  You may obtain a copy of the License at https://www.salto.io/terms-of-use
+ *
+ * CERTAIN THIRD PARTY SOFTWARE MAY BE CONTAINED IN PORTIONS OF THE SOFTWARE. See NOTICE FILE AT https://github.com/salto-io/salto/blob/main/NOTICES
+ */
+import {
+  ChangeValidator,
+  getChangeData,
+  isInstanceChange,
+  isAdditionChange,
+  AdditionChange,
+  InstanceElement,
+} from '@salto-io/adapter-api'
+import { USER_TYPE_NAME } from '../constants'
+
+const isProvisionedUser = (change: AdditionChange<InstanceElement>): boolean =>
+  getChangeData(change).elemID.typeName === USER_TYPE_NAME && getChangeData(change).value.status === 'PROVISIONED'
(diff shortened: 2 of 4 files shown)


packages/okta-adapter/src/change_validators/provisioned_user_addition.ts
```diff
@@ -0,0 +1,35 @@
...
+  ChangeValidator,
+  getChangeData,
+  isInstanceChange,
+  isAdditionChange,
+  AdditionChange,
+  InstanceElement,
+} from '@salto-io/adapter-api'
+import { USER_TYPE_NAME } from '../constants'
+
+const isProvisionedUser = (change: AdditionChange<InstanceElement>): boolean =>
```
Nurdok, line 18: Nit: I would change the `AdditionChange` to `Change` here because:
- It's not needed (you're not using any addition-specific API here)
- You're already checking for `isAdditionChange` below (this isn't providing any additional type safety)
- It gives the impression that only an addition change can be a provisioned user, which is weird - the name of this function is generic, but the signature is unnecessarily specific

packages/okta-adapter/test/change_validators/provisioned_user_addition.test.ts
```diff
@@ -0,0 +1,49 @@
...
+describe('provisionedUserAdditions', () => {
+  const userType = new ObjectType({ elemID: new ElemID(OKTA, USER_TYPE_NAME) })
+
+  describe('addition changes', () => {
+    it('should create change error with level Info when user is created with PROVISIONED status', async () => {
+      const instance = new InstanceElement('user1', userType, { status: 'PROVISIONED' })
+      const change = toChange({ after: instance })
+      const result = await provisionedUserAdditions([change])
+      expect(result).toHaveLength(1)
+      expect(result[0]).toMatchObject({
```
Nurdok, line 21: Also a nit, but more general: I would do this in one `expect` - `expect(result).toEqual([ /* your error here */])`
In the current implementation, if the test fails, you'll get "expected 1, got /whatever/" and you'll have to debug it manually to get the actual contents. With a single `expect`, you'll get the contents immediately.

packages/okta-adapter/test/change_validators/provisioned_user_addition.test.ts
```diff
@@ -0,0 +1,49 @@
...
+        detailedMessage:
+          'Salto does not configure authentication for newly added users. The user will receive an email with instructions to complete the activation process.',
+      })
+    })
+
+    it('should not create change errors for users with STAGED status', async () => {
+      const instance = new InstanceElement('user2', userType, { status: 'STAGED' })
+      const change = toChange({ after: instance })
+      const result = await provisionedUserAdditions([change])
+      expect(result).toHaveLength(0)
```
Nurdok, line 34: Similarly, I would `expect(result).toBeEmpty()`, so that you'll get the contents if the test fails.
(in the next test, too)

Nurdok (approved):
LGTM!
Added some nits, but feel free to ignore.

packages/okta-adapter/test/change_validators/provisioned_user_addition.test.ts
shir-reifenberg: good point, thanks

packages/okta-adapter/test/change_validators/provisioned_user_addition.test.ts
shir-reifenberg: for some reason, the function doesn't exist 🤔 
replaced with `.toMatchObject([])`

packages/okta-adapter/src/change_validators/provisioned_user_addition.ts
shir-reifenberg: I prefer the function to match the type.. but you're right about naming, renamed 😄

packages/okta-adapter/src/change_validators/index.ts
```diff
@@ -36,6 +36,7 @@ import { disabledAuthenticatorsInMfaPolicyValidator } from './disabled_authentic
 import { oidcIdentityProviderValidator } from './oidc_idp'
 import { everyoneGroupAssignments } from './everyone_group_assignments'
 import { emailDomainAdditionValidator } from './email_domain_addition'
+import { provisionedUserAdditions } from './provisioned_user_addition'
 import OktaClient from '../client/client'
```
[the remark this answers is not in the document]
shir-reifenberg: In general I agree, but I preferred all change validators to be next to each other (and I prefer not to mess with the git blame...)

PR #6703, opened 2024-11-04
SALTO-6843: Filtering out invalid keys in convert maps

yelly:
Being more defensive about values we use in this filter since we've seen it fail in the wild.

---

Not sure this is the best approach, open to other ideas.

---
_Release Notes_: 
None.

---
_User Notifications_: 
None.

### DIFF
--- a/packages/salesforce-adapter/src/filters/convert_maps.ts
+++ b/packages/salesforce-adapter/src/filters/convert_maps.ts
@@ -264,11 +264,11 @@ const convertArraysToMaps = (
       .filter(([fieldName]) => _.get(getElementValueOrAnnotations(element), fieldName) !== undefined)
       .forEach(([fieldName, mapDef]) => {
         const mapper = mapDef.mapper ?? defaultMapper
-        const mapDefFilter = (item: Values): boolean => typeof item[mapDef.key] === 'string'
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/src/filters/convert_maps.ts
```diff
@@ -264,30 +264,37 @@ const convertArraysToMaps = (
       .filter(([fieldName]) => _.get(getElementValueOrAnnotations(element), fieldName) !== undefined)
       .forEach(([fieldName, mapDef]) => {
         const mapper = mapDef.mapper ?? defaultMapper
+        const mapDefFilter = (item: Values): boolean => typeof item[mapDef.key] === 'string'
```
tamtamirr, line 267:
```suggestion
        const isConvertableValue = (item: Values): boolean => _.isString(item[mapDef.key])
```
Also, to be extra safe I would implement the following:
const isConvertableValue = (item: unknown): boolean =>_.isPlainObject(item) &&  _.isString(item[mapDef.key])

tamtamirr (commented):
Added one small comment 🙏

packages/salesforce-adapter/src/filters/convert_maps.ts
yelly: Done.

tamtamirr (approved):

PR #6708, opened 2024-11-05
SALTO-6819: Excluding changes on hidden elements from dump to folder

yelly:
The way things currently stand (and likely forever), the adapter logic for dumping to folder never depends on these chages, so we can omit them.
This is becuase dumping to folder is (in many ways) like deploying, and we will never try to deploy changes on hidden elements because they cannot be selected in the plan.
We took a naive approach for filtering hidden changes which is good enough for now but we might choose to expand in the future.

---

_Release Notes_: 
Salesforce:
* Exclude hidden elements in SFDX.

---
_User Notifications_: 
None.

### DIFF
--- a/packages/core/src/core/adapter_format.ts
+++ b/packages/core/src/core/adapter_format.ts
@@ -11,14 +11,14 @@ import {
   AdapterFormat,
   AdapterOperationsContext,
   Change,
-  CORE_ANNOTATIONS,
   Element,
   getChangeData,
+  ReadOnlyElementsSource,
   SaltoError,
 } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
 import { collections } from '@salto-io/lowerdash'
-import { merger, Workspace, ElementSelector, expressions, elementSource } from '@salto-io/workspace'
+import { merger, Workspace, ElementSelector, expressions, elementSource, hiddenValues } from '@salto-io/workspace'
 import { FetchResult } from '../types'
 import { adapterCreators } from './adapters'
 import { MergeErrorWithElements, getFetchAdapterAndServicesSetup, calcFetchChanges } from './fetch'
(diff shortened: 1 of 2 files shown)


packages/core/src/core/adapter_format.ts
```diff
@@ -168,6 +177,12 @@ const loadElementsAndMerge = (
     dir,
   )
 
+// This is a naive approach, for a more complete implementations see workspace.filterOutHiddenChanges.
+// This is good enough for now since hidden value (etc.) changes will not affect adapter format (as far as we can tell).
+// For mixed mode, we need to partition on this test and add all the hidden changes to the unapplied changes.
```
ori-moisis, line 182: I don't think the last line is correct here.
the separation XML and NaCl in mixed mode is supposed to be at the element level (i.e - we do not expect to have to have the same element both in NaCl and XML as that would be more challenging to maintain and understand)

packages/core/src/core/adapter_format.ts
```diff
@@ -168,6 +177,12 @@ const loadElementsAndMerge = (
...
+const filterHiddenChanges = (changes: ReadonlyArray<Change>): ReadonlyArray<Change> =>
+  changes.filter(change => !getChangeData(change).annotations[CORE_ANNOTATIONS.HIDDEN])
```
ori-moisis, line 184: is this enough or do we also need to consider if the type has a HIDDEN_VALUE annotation?
please check the hidden values code to see what the behavior is. this code needs to be consistent with that behavior

yelly: I've moved to using the implemenation from `hiddenValues`. We might take a performance hit here (if we need to resolve types) but that's life.

packages/core/src/core/adapter_format.ts
```diff
@@ -307,7 +327,7 @@ export const syncWorkspaceToFolder = ({
       // re-dump elements that are equal, so even though we do redundant work, the end result should be correct
       const plan = await getPlan({
         before: elementSource.createInMemoryElementSource(folderElements),
-        after: elementSource.createInMemoryElementSource(workspaceElements),
```
yelly, line 310: @ori-moisis note that I've moved to using the elements source we already have here instead of creating a new one. Is this OK?

ori-moisis: no
the workspace elements are already resolved and already in memory, not using that would hurt performance

why did you change it?

yelly: To align with the element source I used below.
Do I need to use an in memory element source also for checking hidden?

ori-moisis: if you have all the elements resolved in memory and you already "paid the tax" of loading, resolving and keeping them in memory, it is always better to use that and not create new copies of the elements by using the original element source

ori-moisis: looking at the code though, I think I might be wrong about this

ori-moisis: ok, I had to read it again, but it looks like this is creating essentially the same element source, so it doesn't matter

ori-moisis (approved):

packages/core/src/core/adapter_format.ts
yelly: I guess I was confused because you used `adapterContext.elementsSource` in the calls to `dumpElementsToFolder`.
I've aligned to always use the in memory source - is this OK?

ori-moisis: as we discussed f2f - I was wrong initially and they are pretty much equivalent.
the reason to keep the code as it was is just readability

PR #6712, opened 2024-11-06
SALTO-5282: Add deploy support for authorization servers

shir-reifenberg:
Added deploy support for authorization servers.

---

I added deploy support for auth servers. Okta accounts come with a default servers, and additional servers can be configured. Each server comes with scopes, claims the policies.
On auth server addition, some scopes and claims might be created (there's no consistent list of the scopes and claims that will be created, but they will have `system = true` property. 

Included in this PR -
- Support for auth servers action (add, activate..)
- Handle auto creation for scopes additions - In case of addition of system scopes, we'll fetch the added scope and then modify it to ensure the scope is updated with the requested values
- remove the reference from auth servers to their scopes, claims and policies. It was needed to avoid the circular dependency + it adds no value.

This PR doesn't handle system claims (as it doesn't work exactly like scopes), I plan to add it in a separate one.

---
_Release Notes_: 

_Okta_adapter_:
- Add deploy support for authorization servers

---
_User Notifications_: 

_Okta adapter_:
- Authorization servers will no longer contain reference to their scopes, claims and policies (the reference exists in the other way)

### DIFF
--- a/packages/okta-adapter/src/definitions/deploy/deploy.ts
+++ b/packages/okta-adapter/src/definitions/deploy/deploy.ts
@@ -45,6 +45,7 @@ import {
   PROFILE_MAPPING_TYPE_NAME,
   SIGN_IN_PAGE_TYPE_NAME,
   ERROR_PAGE_TYPE_NAME,
+  AUTHORIZATION_SERVER,
 } from '../../constants'
 import {
   APP_POLICIES,
@@ -56,6 +57,7 @@ import { isActivationChange, isDeactivationChange } from './utils/status'
 import * as simpleStatus from './utils/simple_status'
 import { isCustomApp } from '../fetch/types/application'
 import { addBrandIdToRequest } from './types/email_domain'
+import { isSystemScope } from './types/authorization_servers'
 
 const log = logger(module)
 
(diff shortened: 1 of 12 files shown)


packages/okta-adapter/test/adapter.test.ts
```diff
@@ -3216,5 +3224,139 @@ describe('adapter', () => {
         expect(nock.pendingMocks()).toHaveLength(0)
       })
     })
+    describe('deploy authorization servers', () => {
+      it('should successfully add an authorization server', async () => {
+        loadMockReplies('authorization_server_add.json')
+        const authServerAddition = authServerInstance.clone()
```
Nurdok, line 3230: Why use `clone` here?
This is different enough from the clones instance that IMO it would be more understandable to create a new instance here. 
Also maybe rename `authServerInstance` to `defaultAuthServer` so that it'll be clear that it's a specific flavor of auth servers? (I'm guessing this is why the values are changed for this test?)

packages/okta-adapter/test/adapter.test.ts
```diff
@@ -3216,5 +3224,139 @@ describe('adapter', () => {
...
+        expect(getChangeData(result.appliedChanges[0] as Change<InstanceElement>).value.id).toEqual(
+          'authorizationserver-fakeid1',
+        )
+        expect(nock.pendingMocks()).toHaveLength(0)
+      })
+      it('should succeessfully modify and activate an authorization server', async () => {
+        loadMockReplies('authorization_server_modify_and_activate.json')
+        const afterAuthServerInstance = authServerInstance.clone()
+        afterAuthServerInstance.value.status = 'ACTIVE'
+        afterAuthServerInstance.value.audiences = ['api://default/activate']
```
Nurdok, line 3256: This is just for "modify any value" but using something with "activate" / "deactivate" in the name is really confusing. Please pick any other value to change :)
(for the tests below as well)

Nurdok (approved):
LGTM!

Left some comments for documentation and clarity.

packages/okta-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -927,6 +984,27 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
...
+                  ({ change }) =>
+                    !isSystemScope(change),
+              },
+            },
+            {
+              request: {
+                endpoint: {
+                  path: '/api/v1/authorizationServers/{parent_id}/scopes',
+                  method: 'get',
+                  queryArgs: { q: '{name}' },
```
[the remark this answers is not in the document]
shir-reifenberg: will add
and just validated 
image

PR #6714, opened 2024-11-06
SALTO-6853: Fixing LWC regression in convert maps

yelly:
---
_Release Notes_: 
_Salesforce_:
* Fixed a regression in deploying Lightning Web Components.

---
_User Notifications_: 
None.


amitgoldarb (approved):
Nice catch!
Let's be sure to add UT in another PR to cover this slippery one

PR #6718, opened 2024-11-07
SALTO-6841: Log an error when getAccessToken fails for Microsoft Security

shoshana-houri:
Header is self explanatory

---

---
_Release Notes_: 

---
_User Notifications_:

### DIFF
--- a/packages/microsoft-security-adapter/src/client/connection.ts
+++ b/packages/microsoft-security-adapter/src/client/connection.ts
@@ -48,7 +48,7 @@ const getAccessToken = async ({ tenantId, clientId, clientSecret, refreshToken }
     const res = await httpClient.post(`${getAuthenticationBaseUrl(tenantId)}/token`, data)
     return res.data.access_token
   } catch (e) {
-    log.error('Failed to get access token: %s', e)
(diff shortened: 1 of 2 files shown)


packages/microsoft-security-adapter/src/client/connection.ts
```diff
@@ -44,8 +44,13 @@ const getAccessToken = async ({ tenantId, clientId, clientSecret, refreshToken }
...
     refresh_token: refreshToken,
   })
-  const res = await httpClient.post(`${getAuthenticationBaseUrl(tenantId)}/token`, data)
-  return res.data.access_token
+  try {
+    const res = await httpClient.post(`${getAuthenticationBaseUrl(tenantId)}/token`, data)
+    return res.data.access_token
+  } catch (e) {
+    log.error(`Failed to get access token: ${e}`)
+    throw new clientUtils.UnauthorizedError(e)
```
shir-reifenberg, line 52: any specific reason to throw `UnauthorizedError` ?
If not, maybe let's keep the behavior as today and just `throw e` ?

shir-reifenberg (approved):
😄

PR #6724, opened 2024-11-07
SALTO-6765: Fail fetch if credentials are invalid

shir-reifenberg:
Also added throw `Unauthorized Error` when generating the token in OAuth flow (cause it's called before validate credentials)

---

additionally, I wrapped `login` func in try-catch as well, as `authParamsFunc` might make network calls

---
_Release Notes_: 

_Okta adapter_:
- Fix a bug causing fetch not to fail on credentials error

_Zendesk adapter_:
- Fix a bug causing fetch not to fail on credentials error

_Microsoft adapter_:
- Fix a bug causing fetch not to fail on credentials error

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/auth/oauth.ts
+++ b/packages/adapter-components/src/auth/oauth.ts
@@ -9,8 +9,9 @@ import _ from 'lodash'
 import qs from 'qs'
 import axios, { AxiosRequestHeaders } from 'axios'
 import axiosRetry from 'axios-retry'
+import { safeJsonStringify } from '@salto-io/adapter-utils'
 import { logger } from '@salto-io/logging'
-import { RetryOptions } from '../client/http_connection'
+import { RetryOptions, UnauthorizedError } from '../client/http_connection'
 
 const log = logger(module)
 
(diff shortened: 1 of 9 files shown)


packages/adapter-components/src/auth/oauth.ts
```diff
@@ -99,21 +100,30 @@ export const oauthAccessTokenRefresh = async ({
...
+    }
+    return {
+      headers: {
+        Authorization: `Bearer ${accessToken}`,
+      },
+    }
+  } catch (error) {
+    log.error(
+      'Failed to get access token, error: %s, stack: %s',
+      safeJsonStringify({ data: error?.response?.data, status: error?.response?.status }),
```
netama, line 124: not sure we should log the data here without an option to omit sensitive credential details from the response body, let's start with just the error itself (which it seems we're not currently logging?)

packages/microsoft-security-adapter/src/client/connection.ts
```diff
@@ -54,7 +54,7 @@ const getAccessToken = async ({ tenantId, clientId, clientSecret, refreshToken }
       safeJsonStringify({ data: e?.response?.data, status: e?.response?.status }),
       e.stack,
     )
-    throw e
+    throw new clientUtils.UnauthorizedError('Invalid Credentials')
```
netama, line 57: leaving this part to @shoshana-houri

netama (approved):
thanks! added one comment re the info in the log message

packages/adapter-components/src/auth/oauth.ts
shir-reifenberg: AFAIK the response data should be safe, but logging the entire error can contain headers sent
by error itself you mean the message ?

packages/microsoft-security-adapter/src/client/connection.ts
shoshana-houri: LGTM :)

packages/adapter-components/src/auth/oauth.ts
shir-reifenberg: changed to `message`

PR #6726, opened 2024-11-10
Bump to version 0.4.5

shir-reifenberg:
# Salto Version 0.4.5
	
### NetSuite adapter

- Update the message and detailedMessage for 'SDF validation error'

### Salesforce adapter
- (Bug-fix) FlowDefinition instances have pending changes when Flow type is excluded
- Moved to using the empty template for creating SFDX projects instead of the standard template which has some extra files.
- Improved the Metadata Deploy progress to include the elapsed time since the deployment was created.
- Added references from LightningPage instances to CustomFields (requires the lightningPageFieldItemReference optional feature)
- RecordType picklist values are now represented as references to the original picklist values (protected by picklistsAsMaps feature flag).
- Return errors from parsing and dumping SFDX instead of throwing.
- Removed coordinates from flows in auto layout mode, reducing noise in comparisons and fetches.
- Improved CPQ Broken References Policy to include instances referencing invalid ones.
- Enabled more seamless edits to task and event objects by connecting them with their respective parents.
- Converted sharing rules values from arrays to maps.
- Enhanced support for all wave types.
- Addressed collision issue with same-name EmailTemplate attachments in file paths.
- Fixed bug in value conversion from strings to numbers.
- Added more detailed information to Apex triggers.
- Enhanced experience with missing Data instances.
- Added new references to Lightning page definition
- Exclude hidden elements in SFDX.
- Fixed a regression in deploying Lightning Web Components.

### Jira adapter
- Users will be able to deploy JSM projects (Only the projects, without subTypes) to their DC instances.

### Okta adapter
- Add a change validator to indicate that users added in status provisioned will receive an email to complete activation.
- Add safety to avoid fetching users in accounts with more than 200,000 users.
- Add deploy support for authorization servers

### Jamf adapter
- Omit password from configurationProfile instances

### Salesforce adapter
- When enabling the feature lightningPageFieldItemReference, expect changes to existing LightningPage nacls.
- Expect changes to many nacls in existing Workspaces

### Okta adapter
- Authorization servers will no longer contain reference to their scopes, claims and policies (the reference exists in the other way).


dantal4 (approved):
Thanks!

PR #6728, opened 2024-11-10
[WIP] SALTO-6874 Allow customizing authentication methods and credential validation by account name

netama:
Make sure the adapter receives the account name when determining the auth flow.
Impacted interfaces:
* authenticationMethods
* validateCredentials

Marking as WIP until I finish checking the implications of this change outside this repo, but happy for feedback on the code in the PR.

---

More details are in the ticket - this is in preparation for using the account name to load runtime overrides that might impact the authentication and credential validation flows.

---
_Release Notes_: 
None

---
_User Notifications_: 
None


packages/adapter-api/src/adapter.ts
```diff
@@ -261,8 +261,8 @@ export type AdapterFormat = {
 
 export type Adapter = {
   operations: (context: AdapterOperationsContext) => AdapterOperations
-  validateCredentials: (config: Readonly<InstanceElement>) => Promise<AccountInfo>
-  authenticationMethods: AdapterAuthentication
+  validateCredentials: (config: Readonly<InstanceElement>, accountName?: string) => Promise<AccountInfo>
+  authenticationMethods: (accountName?: string) => AdapterAuthentication
```
ori-moisis, line 265: I wonder if maybe we want to do something like `operations` where we get the account name as context for a bunch of methods (all of them?) instead of passing it as a parameter to each one individually? like, have `AdapterCreator` as a function that returns an `Adapter` given an account name.

also, (alternatively I guess), I think we should aim to use named arguments in all APIs.
for `validateCredentials` this can be a little annoying as it will be a breaking change but I think it shouldn't be too bad to do it as the effect of this API shouldn't go beyond this repo currently

packages/core/src/api.ts
```diff
@@ -511,7 +541,7 @@ class AdapterInstallError extends Error {
   }
 }
 
-const getAdapterCreator = (adapterName: string): Adapter => {
+export const getAdapterCreator = (adapterName: string): Adapter => {
```
ori-moisis, line 544: why export? is this used anywhere outside this file?

PR #6729, opened 2024-11-10
SALTO-6686: Okta: Prevent multiple "deploy not supported" validation errors

Nurdok:
This PR adds an option to ignore unknown types in the config-based validator, so that it will only report on unsupported deploy actions in types that have some config. Types with no deploy support at all will be caught by the definitions-based validator.

---

None

---
_Release Notes_: 
_Okta Adapter_:
- Fixed a bug where multiple "deploy not supported" validation errors are produced for the same change.

---
_User Notifications_: 
None


shir-reifenberg (commented):
Though this works, if possible, I think we should avoid adding complexity to `createCheckDeploymentBasedOnConfigValidator`.
Can we maybe create a validator that will consolidate both `createCheckDeploymentBasedOnDefinitionsValidator` and `createCheckDeploymentBasedOnConfigValidator`, and will only run the latter validator for changes with no change error from the first validator? WDYT ?
also in this way we can property test we're not generating an error twice

shir-reifenberg (approved):
Please don't forget updating `User Notifications` section in PR description to `None` 🙏

PR #6731, opened 2024-11-10
SALTO-6776: (Salesforce) Add the ability to cancel previous validations

tamtamirr:
(Salesforce) Add the ability to cancel previous validations

---

See the E2E to understand the desired Flow for this feature.

---
_Release Notes_: 
_Salesforce Adapter_:
- (Salesforce) Add the ability to cancel previous validations

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/client/client.ts
```diff
@@ -1110,4 +1111,32 @@ export default class SalesforceClient implements ISalesforceClient {
...
+          body: inspectValue({
+            deployResult: {
+              status: 'Canceling',
+            },
+          }),
+        })
+        if (_.get(cancelDeployResult, ['deployResult', 'status']) === 'Canceling') {
+          await new Promise(resolve => setTimeout(resolve, this.conn.metadata.pollInterval))
+          await checkStatus()
+        }
```
tamtamirr, line 1135: Possible optimization: Do not wait until the deployment was canceled and simply invoke the path request once. This may affect the progress UX badly. I think we can see how it behaves first before deciding that

PR #6737, opened 2024-11-12
SALTO-2432: Log the time it takes to process references for type

shir-reifenberg:

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/references/field_references.ts
+++ b/packages/adapter-components/src/references/field_references.ts
@@ -275,27 +275,37 @@ export const addReferences = async <
   const { elemByElemID, ...fieldLookups } = await indexer.process(awu(contextElements))
 
   const fieldsWithResolvedReferences = new Set<string>()
-  const proccessTimeByType = new Map<string, { time: number; elements: number }>()
-  await awu(instances).forEach(async instance => {
-    const startTime = Date.now()
-    instance.value = await replaceReferenceValues({
-      instance,
-      resolverFinder,
-      elemLookupMaps: fieldLookups as Record<string, multiIndex.Index<[string, string], Element>>,
-      fieldsWithResolvedReferences,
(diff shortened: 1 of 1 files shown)


packages/adapter-components/src/references/field_references.ts
```diff
@@ -283,8 +286,16 @@ export const addReferences = async <
...
+
+    const processTime = Date.now() - startTime
+    const current = proccessTimeByType.get(instance.elemID.typeName) ?? { time: 0, elements: 0 }
+    proccessTimeByType.set(instance.elemID.typeName, {
+      time: current.time + processTime,
+      elements: current.elements + 1,
+    })
   })
   log.debug('added references in the following fields: %s', [...fieldsWithResolvedReferences])
+  log.debug('references processing time by type: %s', inspectValue(proccessTimeByType))
```
netama, line 298: might want to set `maxArrayLength: null` in case there are many types (the default is 100 and I _think_ this also applies to maps), but up to you

also - maybe add a comment linking to the ticket so that we know it can later be removed?

packages/adapter-components/src/references/field_references.ts
```diff
@@ -283,8 +286,16 @@ export const addReferences = async <
       elemByElemID,
       contextStrategyLookup,
     })
+
+    const processTime = Date.now() - startTime
+    const current = proccessTimeByType.get(instance.elemID.typeName) ?? { time: 0, elements: 0 }
+    proccessTimeByType.set(instance.elemID.typeName, {
+      time: current.time + processTime,
+      elements: current.elements + 1,
+    })
```
netama, line 295: wondering if it's worth wrapping in try+catch just to be on the safe side?

netama (approved):
thanks! added very minor nits

PR #6738, opened 2024-11-13
SALTO-6896: (Bugfix) Salesforce convertMaps filter does not convert the types of Picklist and MultiselectPicklist

tamtamirr:
(Bugfix) Salesforce convertMaps filter does not convert the types of Picklist and MultiselectPicklist

---

The Primitive Field types were not updated directly which resulted in incorrect annotationRefType for `valueSet`. (List<valueSet> instead of OrderedMap<valueSet>)

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/src/filters/convert_maps.ts
+++ b/packages/salesforce-adapter/src/filters/convert_maps.ts
@@ -33,6 +33,7 @@ import {
   ElemID,
   isObjectTypeChange,
   BuiltinTypes,
+  isPrimitiveType,
 } from '@salto-io/adapter-api'
 import { collections, values as lowerdashValues } from '@salto-io/lowerdash'
 import { naclCase, applyFunctionToChangeData } from '@salto-io/adapter-utils'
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/src/filters/convert_maps.ts
```diff
@@ -627,13 +628,18 @@ const filter: FilterCreator = ({ config }) => ({
     })
 
     const fields = elements.filter(isObjectType).flatMap(obj => Object.values(obj.fields))
-    await awu(Object.entries(annotationDefsByType)).forEach(async ([fieldType, annotationToMapDef]) => {
-      const fieldsToConvert = fields.filter(field => field.refType.elemID.typeName === fieldType)
+    await awu(Object.entries(annotationDefsByType)).forEach(async ([fieldTypeName, annotationToMapDef]) => {
+      const fieldType = elements.filter(isPrimitiveType).find(e => e.elemID.name === fieldTypeName)
```
yelly: Maybe filter out the primitive types out once at the top of the `onFetch` instead of needing to iterate over all the elements so many times? You can also go the extra mile and map them by name for fast lookup here.

yelly: Looks like you already got to this...

packages/salesforce-adapter/src/filters/convert_maps.ts
```diff
@@ -627,13 +628,19 @@ const filter: FilterCreator = ({ config }) => ({
     })
 
     const fields = elements.filter(isObjectType).flatMap(obj => Object.values(obj.fields))
-    await awu(Object.entries(annotationDefsByType)).forEach(async ([fieldType, annotationToMapDef]) => {
-      const fieldsToConvert = fields.filter(field => field.refType.elemID.typeName === fieldType)
+    const primitiveTypesByName = _.keyBy(elements.filter(isPrimitiveType), e => e.elemID.name)
```
yelly, line 631: These can happen outside of this `forEach`.

tamtamirr: I'm a-bit confused. This keyBy happens outside of a loop..? 🤔

PR #6750, opened 2024-11-13
SALTO-6845: (Bugfix) Salesforce `picklistAsMaps` feature does not work for environments with CPQ

tamtamirr:
(Bugfix) Salesforce picklistAsMaps feature does not work for environments with CPQ 

---

For environments that manage CPQ metadata, turning on the `picklistAsMaps` feature makes fetch fail violently. This PR exposes two issues in the adapter:

1. `convert_maps` filter should be able to key-by `ReferenceExpression` value.
2. If `OrderedMap` order value references value with unresolved references, the Element won't be fully resolved.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/src/adapter.ts
+++ b/packages/salesforce-adapter/src/adapter.ts
@@ -23,9 +23,13 @@ import {
   PartialFetchData,
   Element,
   isInstanceElement,
+  Field,
+  isField,
+  isObjectType,
+  TypeReference,
 } from '@salto-io/adapter-api'
-import { filter, inspectValue, logDuration, safeJsonStringify } from '@salto-io/adapter-utils'
-import { resolveChangeElement, restoreChangeElement } from '@salto-io/adapter-components'
+import { filter, inspectValue, logDuration, ResolveValuesFunc, safeJsonStringify } from '@salto-io/adapter-utils'
+import { resolveChangeElement, resolveValues, restoreChangeElement } from '@salto-io/adapter-components'
 import { MetadataObject } from '@salto-io/jsforce'
 import _ from 'lodash'
 import { logger } from '@salto-io/logging'
@@ -201,6 +205,7 @@ export const allFilters: Array<FilterCreator> = [
   emailTemplateFilter,
   // standardValueSetFilter should run before convertMapsFilter
   standardValueSetFilter,
+  cpqLookupFieldsFilter,
   // convertMapsFilter should run before profile fieldReferencesFilter
   convertMapsFilter,
   // picklistReferences should run after convertMapsFilter and before fieldReferencesFilter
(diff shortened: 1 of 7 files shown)


packages/salesforce-adapter/src/adapter.ts
```diff
@@ -432,6 +436,40 @@ type CreateFiltersRunnerParams = {
   contextOverrides?: Partial<FilterContext>
 }
 
+const isOrderedMapTypeOrRefType = (typeRef: TypeElement | TypeReference): boolean =>
+  typeRef.elemID.name.startsWith('OrderedMap<')
```
yelly, line 440: Didn't you just rename the type? Or is that a different PR?
Either way, look out.

packages/salesforce-adapter/src/filters/convert_maps.ts
```diff
@@ -99,21 +111,22 @@ const createOrderedMapType = <T extends TypeElement>(innerType: T): ObjectType =
...
- */
-export const defaultMapper = (val: string): string[] => val.split(API_NAME_SEPARATOR).map(v => naclCase(v))
+const stringifyMapperInput = (mapperInput: MapperInput): string =>
+  _.isString(mapperInput) ? mapperInput : apiNameSync(mapperInput.value) ?? ''
+
+const isMapperInput = (val: unknown): val is MapperInput =>
+  _.isString(val) || isResolvedReferenceExpressionToElement(val)
+
+export const defaultMapper: MapperFunc = val =>
+  (_.isString(val) ? val : apiNameSync(val.value) ?? '').split(API_NAME_SEPARATOR).map(v => naclCase(v))
```
yelly, line 121:
```suggestion
  stringifyMapperInput(val).split(API_NAME_SEPARATOR).map(v => naclCase(v))
```
(I'm assuming this is what you meant to do)

packages/salesforce-adapter/src/filters/cpq/lookup_fields.ts
```diff
@@ -142,16 +143,12 @@ const transformObjectLookupValueSetFullNames = async (
   ) => ReferenceExpression | string | undefined,
 ): Promise<ObjectType> => {
   const lookupFields = await getLookupFields(object)
-  lookupFields.forEach(field => transformLookupValueSetFullNames(field, transformFullNameFn))
+  await awu(lookupFields).forEach(field => transformLookupValueSetFullNames(field, transformFullNameFn))
```
yelly, line 146: What? How did this work before?

packages/salesforce-adapter/src/adapter.ts
tamtamirr: different PR

packages/salesforce-adapter/src/filters/cpq/lookup_fields.ts
tamtamirr: race condition. Found this with debugger which is significantly slower.

packages/salesforce-adapter/src/filters/convert_maps.ts
yelly: Did you miss this?

yelly (approved):

PR #6754, opened 2024-11-14
SALTO-6905 Update trigger skill priorities

adfineberg:
Update support for trigger priorities according to https://developer.zendesk.com/documentation/ticketing/using-the-zendesk-api/setting-skill-priority-with-skills-in-trigger-action/

---

---
_Release Notes_: 

---
_User Notifications_: 
_Zendesk Adapter:_
* Trigger skill priorities will change from `optional` to `optional high` as per  https://developer.zendesk.com/documentation/ticketing/using-the-zendesk-api/setting-skill-priority-with-skills-in-trigger-action/


edenhassid (approved):
nice! 🕺

PR #6761, opened 2024-11-17
SALTO-6901: Deprecating extra dependencies optional features

yelly:
This also allows completely removing the v1 implementation.

---

None.

---

_Release Notes_: 
_Salesforce_:
* Removing the v1 extra dependencies filter implementation.

---

_User Notifications_: 
None.


tamtamirr (approved):
LGTM

PR #6765, opened 2024-11-17
SALTO-6917 Jira: run sort-lists filter on relevant types only

kami-delaroz:
SALTO-6917 Jira: run sort-lists filter on relevant types only

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/sort_lists.ts
+++ b/packages/jira-adapter/src/filters/sort_lists.ts
@@ -11,9 +11,13 @@ import _ from 'lodash'
 import {
   AUTOMATION_TYPE,
   DASHBOARD_TYPE,
+  FIELD_CONFIGURATION_SCHEME_TYPE,
+  ISSUE_TYPE_SCREEN_SCHEME_TYPE,
   NOTIFICATION_EVENT_TYPE_NAME,
   NOTIFICATION_SCHEME_TYPE_NAME,
+  PERMISSION_SCHEME_TYPE_NAME,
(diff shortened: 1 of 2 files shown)


packages/jira-adapter/src/filters/sort_lists.ts
```diff
@@ -132,6 +136,19 @@ const VALUES_TO_SORT: Record<string, Record<string, string[]>> = {
...
+  ISSUE_TYPE_SCREEN_SCHEME_TYPE,
+  AUTOMATION_TYPE,
+  FIELD_CONFIGURATION_SCHEME_TYPE,
+  NOTIFICATION_SCHEME_TYPE_NAME,
+  DASHBOARD_TYPE,
+  PROJECT_ROLE_TYPE,
+  REQUEST_TYPE_NAME,
+  WORKFLOW_TYPE_NAME,
+  WORKFLOW_CONFIGURATION_TYPE,
+])
```
IdoZakk, line 151: This is a bit error-prone, as we define the VALUES_TO_SORT in two different places.
I think we should change the VALUES_TO_SORT to have 3 layers. The first is the upper type, the other two are the existing ones. 
Then here we can just take the keys of the records (turned into a set)

IdoZakk (approved):
🥇 
Approved for time reasons, please see my comment. I think we should approach this differently 
Also please note the tests fail, they do not have a specific type for the instance and should be fixed

shayc331 (commented):
please add constants to the upper list

PR #6767, opened 2024-11-17
SALTO-5289: reduce fieldConfiguration nacl size

shayc331:
_reduce fieldConfiguration nacl size_

---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/jira-adapter/src/adapter.ts
+++ b/packages/jira-adapter/src/adapter.ts
@@ -68,6 +68,7 @@ import fieldConfigurationSplitFilter from './filters/field_configuration/field_c
 import fieldConfigurationItemsFilter from './filters/field_configuration/field_configuration_items'
 import missingFieldDescriptionsFilter from './filters/field_configuration/missing_field_descriptions'
 import fieldConfigurationDependenciesFilter from './filters/field_configuration/field_configuration_dependencies'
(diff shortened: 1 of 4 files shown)


packages/jira-adapter/src/adapter.ts
```diff
@@ -352,6 +353,8 @@ export const DEFAULT_FILTERS = [
   fieldConfigurationIrrelevantFields,
   // Must run after fieldConfigurationIrrelevantFields
   fieldConfigurationSplitFilter,
+  // Must run before replaceFieldConfigurationReferences
+  fieldConfigurationDefaultValuesFilter,
```
shir-reifenberg, line 357: How does this behaves with `fieldConfigurationSplitFilter` ?
I think if you want `removeFieldConfigurationDefaultValues` to work with `splitFieldContextOptions` it needs to be running before `fieldConfigurationSplitFilter` as well 
(and anyway we should verify they both work together / block this case if it doesn't)

kami-delaroz: I believe fieldConfigurationSplitFilter should be removed from the code, it's not in used, and we know it's dangerous to take this approach.

kami-delaroz (approved):

PR #6771, opened 2024-11-17
SALTO-6915: create placeholder types for types with no entries

shir-reifenberg:
types with `dependsOn` were not created when they have no entries, this PR fix this to ensure all we create types for all type level types

---

---
_Release Notes_: 

_Jamf adapter_:
- Fix a bug that might cause merge errors (workspace errors) after deployment of instance of type that doesn't exist 

_Google workspace adapter_:
- Fix a bug that might cause merge errors (workspace errors) after deployment of instance of type that doesn't exist 

_Pager Duty adapter_:
- Fix a bug that might cause merge errors (workspace errors) after deployment of instance of type that doesn't exist 

_Confluence adapter_:
- Fix a bug that might cause merge errors (workspace errors) after deployment of instance of type that doesn't exist 

_Workato adapter_:
- Fix a bug that might cause merge errors (workspace errors) after deployment of instance of type that doesn't exist 

_Microsoft Security adapter_:
- Fix a bug that might cause merge errors (workspace errors) after deployment of instance of type that doesn't exist 

---

_User Notifications_: 
None


netama (approved):
thanks!
re the release notes - we shouldn't write project58 there, can you please add a section for each impacted adapter instead?

packages/adapter-components/src/fetch/element/type_utils.ts
```diff
@@ -379,3 +380,38 @@ export const removeNullValues = ({
...
+  adapterName,
+  definedTypes,
+  defQuery,
+}: {
+  adapterName: string
+  definedTypes: Record<string, ObjectType>
+  defQuery: ElementAndResourceDefFinder<Options>
+}): Record<string, ObjectType> => {
+  const topLevelTypeNames = Object.keys(_.pickBy(defQuery.getAll(), def => def.element?.topLevel?.isTopLevel))
+  const missingTypes = topLevelTypeNames.filter(typeName => definedTypes[typeName] === undefined)
```
netama, line 400: nit: may be best to use 
```suggestion
  const missingTypes = topLevelTypeNames.filter(typeName => !Object.prototype.hasOwnProperty.call(definedTypes, typeName))
```
or alternatively 
```suggestion
  const missingTypes = topLevelTypeNames.filter(typeName => !isElement(definedTypes[typeName]))
```
to avoid failing on types such as `toString` (not very likely to be typenames and not fieldnames, but...)

packages/adapter-components/src/fetch/element/type_utils.ts
```diff
@@ -379,3 +380,38 @@ export const removeNullValues = ({
...
+  if (missingTypes.length > 0) {
+    log.debug('creating empty types for the following types: %s', inspectValue(missingTypes))
```
netama, line 402: nit: since `inspectValue` can truncate, maybe include in the log the number of types / extend `maxArrayLength` ?

netama (approved):
thanks! 🎈

packages/adapter-components/src/fetch/element/type_utils.ts
shir-reifenberg: good point, thanks

PR #6773, opened 2024-11-17
Reduce memory consumption when updating nacl files

dantal4:
In `nacl_files_source.ts` we need the current element IDs in each modified file. Instead of getting them by calling `elements()` that reads and deserialize all of the elements in the nacl file, we can use the `elementsIndex` that holds a mapping between `ElemID` and the files it's in.
This is currently disabled by default and can be enabled with an env var - `SALTO_CREATE_FILENAMES_TO_ELEMENT_IDS_MAPPING`

---

The change in the tests is the addition of `describe.each([false, true])` with the env var's value on all of the tests in each file.

---
_Release Notes_: 
Workspace:
- Reduce memory consumption when updating nacl files (enabled with an env var - `SALTO_CREATE_FILENAMES_TO_ELEMENT_IDS_MAPPING`)

---
_User Notifications_: 
None

### DIFF
--- a/packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
+++ b/packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
@@ -137,6 +137,22 @@ export const toPathHint = (filename: string): string[] => {
   return [...dirPathSplitted, osPath.basename(filename, osPath.extname(filename))]
 }
 
+const createFilenamesToElementIDsMapping = async (
+  currentState: NaclFilesState,
+  naclFilesByName: Record<string, ParsedNaclFile>,
+): Promise<Record<string, ElemID[]>> =>
+  log.timeDebug(
+    async () =>
+      _.mapValues(
+        await awu(currentState.elementsIndex.entries())
+          .flatMap(row => row.value.map(filename => ({ path: row.key, filename })))
+          .filter(row => naclFilesByName[row.filename] !== undefined)
+          .groupBy(row => row.filename),
+        group => group.map(row => ElemID.fromFullName(row.path)),
+      ),
+    'create filenames to element ids mapping from elements index',
+  )
+
 export const getElementReferenced = async (
   element: Element,
 ): Promise<{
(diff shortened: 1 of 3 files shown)


packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
```diff
@@ -137,6 +140,22 @@ export const toPathHint = (filename: string): string[] => {
...
 
+const createFilenamesToElementIDsMapping = async (
+  currentState: NaclFilesState,
+  naclFilesByName: Record<string, ParsedNaclFile>,
+): Promise<Record<string, ElemID[]>> =>
+  log.timeDebug(
+    async () =>
+      _.mapValues(
+        await awu(currentState.elementsIndex.entries())
+          .flatMap(row => row.value.map(filename => ({ path: row.key, filename })))
```
shir-reifenberg, line 151: readability nit:
```suggestion
          .flatMap(({ key, value }) => row.value.map(filename => ({ path: row.key, filename })))
```

packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
```diff
@@ -137,6 +140,22 @@ export const toPathHint = (filename: string): string[] => {
...
+          .filter(row => naclFilesByName[row.filename] !== undefined)
+          .groupBy(row => row.filename),
+        group => group.map(row => ElemID.fromFullName(row.path)),
```
shir-reifenberg, line 154: probably won't make much of a difference, but should we maybe keep those as strings and convert to `ElemID` later ? (i.e `getElementIDsFromNaclFile`)

packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
```diff
@@ -364,6 +383,20 @@ const buildNaclFilesState = async ({
   const relevantElementIDs: ElemID[] = []
   const newElementsToMerge: AsyncIterable<Element>[] = []
 
+  const shouldCreateFilenamesToElementIDsMapping = getSaltoFlagBool(CREATE_FILENAMES_TO_ELEMENT_IDS_MAPPING_FLAG)
```
shir-reifenberg, line 386: can we please log `shouldCreateFilenamesToElementIDsMapping` ?

shir-reifenberg (approved):
Looks great! added mainly nits

PR #6775, opened 2024-11-18
SALTO-6048: Remove adding JWK to exclude list migration

shir-reifenberg:
As part of SALTO-6048, We added migration to migrate all existing envs to have `JWK` type in their exclude list
It has been long enough since, and all active envs are migrated, so it's safe to remove this now

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None


shoshana-houri (approved):
LGTM!

PR #6777, opened 2024-11-18
DOC-191 - Update the CPQ doc to include an example with Internal Id as Salto Id

adamro:
Update the CPQ doc to include an example with Internal Id as Salto Id
---

None

---
_Release Notes_: 
None
---
_User Notifications_: 
None


yelly (approved):
Thanks

PR #6781, opened 2024-11-18
SALTO 6824: Add workspace validate command to the CLI

AlmogMesilaty:
Added workspace validate command to the CLI

---

The CLI output should appear as follows:

Valid workspace:
image

Workspace with errors:
image

Workspace with warnings:
image

---

_Release Notes_:
CLI: 
Users can use the validate command as part of the workspace sub-commands. This command will indicate whether the workspace is valid or, if not, specify the errors or warnings and provide links to their locations.

---
_None_

### DIFF
--- a/packages/cli/src/command_builder.ts
+++ b/packages/cli/src/command_builder.ts
@@ -65,7 +65,7 @@ type CommandInnerDef<T> = {
 
 export type WorkspaceCommandArgs<T> = Omit<DefActionInput<T>, 'workspacePath'> & { workspace: Workspace }
 
-export type WorkspaceCommandAction<T = {}> = (args: WorkspaceCommandArgs<T>) => Promise<CliExitCode>
+export type WorkspaceCommandAction<T> = (args: WorkspaceCommandArgs<T>) => Promise<CliExitCode>
 
 export type WorkspaceCommandDef<T> = {
   properties: CommandOptions<T>
--- a/packages/cli/src/command_builder.ts
+++ b/packages/cli/src/command_builder.ts
@@ -65,7 +65,7 @@ type CommandInnerDef<T> = {
 
 export type WorkspaceCommandArgs<T> = Omit<DefActionInput<T>, 'workspacePath'> & { workspace: Workspace }
 
-export type WorkspaceCommandAction<T> = (args: WorkspaceCommandArgs<T>) => Promise<CliExitCode>
+export type WorkspaceCommandAction<T = {}> = (args: WorkspaceCommandArgs<T>) => Promise<CliExitCode>
 
 export type WorkspaceCommandDef<T> = {
   properties: CommandOptions<T>
--- a/packages/cli/src/commands/workspace.ts
+++ b/packages/cli/src/commands/workspace.ts
@@ -16,8 +16,6 @@ import {
   formatStepStart,
   formatStepFailed,
   formatStepCompleted,
-  error,
-  warn,
 } from '../formatter'
 import { outputLine, errorOutputLine } from '../outputer'
 import Prompts from '../prompts'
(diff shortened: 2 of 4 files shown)


packages/cli/src/commands/workspace.ts
```diff
@@ -201,13 +202,41 @@ const setStateProviderDef = createWorkspaceCommand({
...
+  }
+
+  spinner.succeed('Workspace is valid')
+  return CliExitCode.Success
+}
+
+const wsValidateDef = createWorkspaceCommand({
+  properties: {
+    name: 'validate',
+    description: 'Log the errors on the workspace',
```
Uri-Bar, line 228:
```suggestion
    description: 'Log the workspace errors',
```;'

File: packages/cli/src/command_builder.ts [hunk 2]
```diff
@@ -65,7 +65,7 @@ type CommandInnerDef<T> = {
 
 export type WorkspaceCommandArgs<T> = Omit<DefActionInput<T>, 'workspacePath'> & { workspace: Workspace }
 
-export type WorkspaceCommandAction<T> = (args: WorkspaceCommandArgs<T>) => Promise<CliExitCode>
+export type WorkspaceCommandAction<T = {}> = (args: WorkspaceCommandArgs<T>) => Promise<CliExitCode>
```
Uri-Bar, line 68: dont change it here, check how cacheUpdateAction is dealing with that

packages/cli/test/commands/workspace.test.ts
```diff
@@ -308,4 +317,62 @@ describe('workspace command group', () => {
...
     })
   })
+
+  describe('validate command', () => {
+    const commandName = 'validate'
+    const cliCommandArgs: mocks.MockCommandArgs = mocks.mockCliCommandArgs(commandName, cliArgs)
+    const workspace: mocks.MockWorkspace = mocks.mockWorkspace({})
+
+    describe('when workspace is valid', () => {
+      it('should indicate the workspace is valid', async () => {
```
Uri-Bar, line 327: please add something that validate the Workspace is valid text

packages/cli/test/commands/workspace.test.ts
```diff
@@ -308,4 +317,62 @@ describe('workspace command group', () => {
...
+            errors: [],
+          }),
+        )
+        const result = await wsValidateAction({
+          ...cliCommandArgs,
+          input: {},
+          workspace,
+        })
+        expect(workspaceFunctions.printWorkspaceErrors).not.toHaveBeenCalled()
+        expect(result).toEqual(CliExitCode.Success)
```
Uri-Bar, line 340: I see in all cases we return CliExitCode.Success so there is no need to check them (you can remove it from all the test)

packages/cli/test/commands/workspace.test.ts
Uri-Bar: this relavant to all the tests, we want to see some difference between warning and error, as it has different outcome in the code

packages/cli/test/commands/workspace.test.ts
```diff
@@ -308,4 +317,62 @@ describe('workspace command group', () => {
...
+      })
+    })
```
Uri-Bar, line 342:
```suggestion
    describe('when workspace is valid', () => {
    beforeEach(async () => {
            mockedWorkspaceFunctions.validateWorkspace.mockResolvedValueOnce(
          Promise.resolve({
            status: 'Valid',
            errors: [],
          }),
        )
        await wsValidateAction({
          ...cliCommandArgs,
          input: {},
          workspace,
        })
      }
      it('should indicate the workspace is valid', () => {
        expect(workspaceFunctions.printWorkspaceErrors).not.toHaveBeenCalled()
      })
    })
```

Uri-Bar: This is a best practice to run the logic in a beforeEach/BeforeAll and use the it to only check to only check to outcome.
this relavant to all the describes here, and because we dont want to check the CliExitCode we dont need to save the result

packages/cli/src/commands/workspace.ts
```diff
@@ -201,13 +202,41 @@ const setStateProviderDef = createWorkspaceCommand({
...
+
+  if (status === 'Warning') {
+    spinner.fail(`Workspace has ${errors.length === 1 ? 'a warning' : 'warnings'}:`)
+    await printWorkspaceErrors(status, await formatWorkspaceErrors(workspace, errors), output)
+    return CliExitCode.Success
+  }
+
+  spinner.succeed('Workspace is valid')
+  return CliExitCode.Success
+}
```
Uri-Bar, line 223: I see you return return CliExitCode.Success in all the cases, maybe use switch case when the success is in the default section and after all return the exitCode

Uri-Bar (commented):
looks very good, added few questions

PR #6782, opened 2024-11-18
SALTO-2905: Supporting settings in SFDX

yelly:
The new settings filter no longer relies on the client so it works for SFDX.

---

Interestingly, we don't need to do anything pre-deploy because they way settings are deployed is super weird in SF.

---

_Release Notes_: 
_Salesforce_:
* Add support for settings types in SFDX.

---

_User Notifications_: 
None.

### DIFF
--- a/packages/salesforce-adapter/src/sfdx_parser/sfdx_dump.ts
+++ b/packages/salesforce-adapter/src/sfdx_parser/sfdx_dump.ts
@@ -44,9 +44,6 @@ const { withLimitedConcurrency } = promises.array
 const FILE_DELETE_CONCURRENCY = 100
 
 export const UNSUPPORTED_TYPES = new Set([
-  // Salto uses non-standard type names here (SFDX names them all "Settings", we have a separate type for each one)
-  // This causes us to always think the settings in the project need to be deleted
-  'Settings',
   // For documents with a file extension (e.g. bla.txt) the SF API returns their fullName with the extension (so "bla.txt")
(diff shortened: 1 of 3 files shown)


packages/salesforce-adapter/src/sfdx_parser/sfdx_dump.ts
```diff
@@ -44,9 +44,6 @@ const { withLimitedConcurrency } = promises.array
 const FILE_DELETE_CONCURRENCY = 100
 
 export const UNSUPPORTED_TYPES = new Set([
-  // Salto uses non-standard type names here (SFDX names them all "Settings", we have a separate type for each one)
-  // This causes us to always think the settings in the project need to be deleted
-  'Settings',
```
amitgoldarb, line 49: This is only actually supported after we GA the `retrieveSettings` feature right?

yelly: Nope, in the PR I selectively enable the feature for the SFDX flow which is good enough.

amitgoldarb: Meaning you've tested this on both permutations? On a workspace with this feature on & off?

yelly: Yes.
Obviously when using `read` in the workspace there's a bigger difference between what we get from salto and what we get from the `sf` cli, but it works.

amitgoldarb (approved):
LGTM! Thanks for this!

PR #6785, opened 2024-11-18
SALTO-6903: Avoid creating a unified source map for many files

ori-moisis:
When finding locations for changes, try to avoid creating a single source map. We suspect this single source map consumes a lot of memory.

Instead, try to group changes by which files that could be located in, and for each group create a source map just from the relevant file(s).

This is all behind a flag: SALTO_USE_SPLIT_SOURCE_MAP_IN_UPDATE

---

Some results from a reproduction with the dummy adapter - having 70 large files (~50k lines each).
Fetch with a single change per file:

  | Peak memory in relevant part | Time in code | Total fetch time
-- | -- | -- | --
Old code | before: 1014, peak: 2441.8, after: 965.7 | 16.25 | 122505 ms |  
New code | before: 948.5, peak: 1905.1 (mid run GC to 926.1), after: 1027.7 | 15.04 | 121357 ms |  

Fetch with many changes per file (~25k changes)

  | Peak memory in relevant part | Time in code | Total fetch time
-- | -- | -- | --
Old code | before: 8123.4, peak: 8649.7→(GC to 7523.4)→8774.6, after: 8677.2 | 63.25 | 555964 ms |  
New code | before: 8129.2, peak: 8681.9 (mid run GC to 7577.6), after: 8066.5 | 76.24 | 527406 ms |  

---
_Release Notes_: 
Core:
- Add flag SALTO_USE_SPLIT_SOURCE_MAP_IN_UPDATE which should improve memory consumption when updating many large NaCl files, currently off by default

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
+++ b/packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
@@ -45,7 +45,7 @@ export type DetailedChangeWithSource = DetailedChange & {
   requiresIndent?: boolean
 }
 
-export const createFileNameFromPath = (pathParts?: ReadonlyArray<string>): string =>
+const createFileNameFromPath = (pathParts?: ReadonlyArray<string>): string =>
   `${path.join(...(pathParts ?? ['unsorted']))}${FILE_EXTENSION}`
 
 type PositionInParent = {
--- a/packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
+++ b/packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
@@ -45,7 +45,7 @@ export type DetailedChangeWithSource = DetailedChange & {
   requiresIndent?: boolean
 }
 
-const createFileNameFromPath = (pathParts?: ReadonlyArray<string>): string =>
+export const createFileNameFromPath = (pathParts?: ReadonlyArray<string>): string =>
   `${path.join(...(pathParts ?? ['unsorted']))}${FILE_EXTENSION}`
 
 type PositionInParent = {
--- a/packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
+++ b/packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
@@ -868,7 +868,6 @@ const buildNaclFilesSource = (
     naclFiles: string[],
   ): Promise<DetailedChangeWithSource[]> => {
     const { parsedNaclFiles } = await getState()
-    log.debug('Nacl source %s getting source maps for %d nacl files', sourceName, naclFiles.length)
     const changedFileSourceMaps = (
       await withLimitedConcurrency(
         naclFiles.map(naclFile => async () => {
@@ -922,6 +921,11 @@ const buildNaclFilesSource = (
       })),
       ({ potentialFiles }) => potentialFiles?.join(','),
     )
+    log.debug(
+      'Nacl source %s getting source maps for %d nacl file groups',
+      sourceName,
+      Object.keys(changesByFiles).length,
+    )
     const changesWithLocationsByFiles = await withLimitedConcurrency(
       Object.values(changesByFiles).map(
         changeGroups => () =>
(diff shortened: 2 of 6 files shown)


packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
```diff
@@ -380,9 +379,9 @@ const buildNaclFilesState = async ({
   const staticFilesIndexDeletions: Record<string, Set<string>> = {}
   // We need to iterate over this twice - so no point in making this iterable :/
   const relevantElementIDs: ElemID[] = []
-  const newElementsToMerge: AsyncIterable<Element>[] = []
+  const newElementsToMerge: Element[] = []
```
ori-moisis, line 382: this change is kind of unrelated, did it here because it made debugging things harder.
keeping it in the PR because it should also slightly improve performance (removing unnecessary async)

if this makes it harder to review let me know and I'll split it to a separate PR

packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
```diff
@@ -884,18 +875,87 @@ const buildNaclFilesSource = (
           const parsedNaclFile = await parsedNaclFiles.get(naclFile)
           return values.isDefined(parsedNaclFile) ? getSourceMap(parsedNaclFile.filename, false) : undefined
         }),
-        CACHE_READ_CONCURRENCY,
+        SOURCE_MAP_READ_CONCURRENCY,
```
ori-moisis, line 878: reduced concurrency here for two reasons
1. less concurrency could help lower the peak in memory consumption
2. when using split source maps, this code might be called concurrently

packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
```diff
@@ -1191,23 +1251,5 @@ export const naclFilesSource = async (
   staticFilesSource: StaticFilesSource,
   remoteMapCreator: RemoteMapCreator,
   persistent: boolean,
-  parsedFiles?: ParsedNaclFile[],
```
ori-moisis, line 1194: removed this option
it was only used in unit tests, but those unit tests created unrealistic scenarios and would not allow properly testing the new behavior, so I had to change the tests and this option was no longer needed

packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
shir-reifenberg: This changes the current behavior of the unified source map, no? What is the impact of this ? maybe worth passing `SOURCE_MAP_READ_CONCURRENCY` as a param to `getChangeLocationsForFiles` so we can pass the current value as today for the unified map ? WDYT?

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
```diff
@@ -45,7 +45,7 @@ export type DetailedChangeWithSource = DetailedChange & {
   requiresIndent?: boolean
 }
 
-const createFileNameFromPath = (pathParts?: ReadonlyArray<string>): string =>
+export const createFileNameFromPath = (pathParts?: ReadonlyArray<string>): string =>
```
shir-reifenberg, line 48: Is this used outside of this file?

packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
```diff
@@ -866,43 +863,114 @@ const buildNaclFilesSource = (
...
+        .toArray(),
     )
+    log.debug('Nacl source %s getting source maps for %d nacl files', sourceName, naclFiles.length)
+    return getChangeLocationsForFiles(changes, naclFiles)
   }
 
+  const groupChangesByFilename = (changes: DetailedChange[]): Promise<Record<string, DetailedChangeWithSource[]>> =>
+    log.timeDebug(
+      async () => {
+        const changesWithLocation = getSaltoFlagBool(WORKSPACE_FLAGS.useSplitSourceMapInUpdate)
```
shir-reifenberg, line 960: Can we please log in which option we're using ?

shir-reifenberg (approved):
🔥 🔥 
added one question, but pre approving

packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
ori-moisis: I thought about it, but felt like 100 is too much anyway for such a large computation, so I don't feel too bad about reducing it here.
and AFAICT this can't really change the result of the computation, only the concurrency, so I felt that it is safe.

packages/workspace/src/workspace/nacl_files/nacl_file_update.ts
ori-moisis: good catch, it was in one iteration of the code but now I realize I removed the usage.
will remove it, thanks!

packages/workspace/src/workspace/nacl_files/nacl_files_source.ts
ori-moisis: When I moved the logs around I made them different
```
Nacl source %s getting source maps for %d nacl file groups
```
vs
```
Nacl source %s getting source maps for %d nacl files
```

figured it would be good enough, do you think we need an extra log?

PR #6786, opened 2024-11-19
SALTO-6711 skip getting additional dependent instances when there are…

dantal4:
… no dependent types

When getting elements dependents using the `referenceSources` index we run `elementsSource.list()` in order to get the instances of the dependent types. When there are no dependent types we can skip it.

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/workspace/src/workspace/dependents.ts
+++ b/packages/workspace/src/workspace/dependents.ts
@@ -51,6 +51,12 @@ const getDependentIDsFromReferenceSourceIndex = async (
 
       const dependentIDs = await getDependentIDs(elemIDs)
 
+      // if there are no dependent types we can return `dependentIDs` and avoid
+      // iterating `elementsSource.list()` to get the additional dependent instances.
+      if (!dependentIDs.some(id => id.idType === 'type')) {
+        return dependentIDs
+      }
+
       // in `referenceSourcesIndex` there are no references between types and their instances
       // so we should add the instances of the types that are in `addedIDs` as well.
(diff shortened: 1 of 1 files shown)


packages/workspace/test/workspace/dependents.test.ts
```diff
@@ -116,12 +125,15 @@ describe('dependents', () => {
...
       })
 
-      describe('reference dependents', () => {
+      describe('instance dependents', () => {
         let dependentIDs: ElemID[]
 
         beforeAll(async () => {
           dependentIDs = await getDependentIDs(new ElemID('salto', 'base', 'instance', 'aBaseInst'))
         })
+        it('should not call elementsSource.list', () => {
```
ori-moisis, line 134: this is an internal implementation detail, I don't think it should be enforced by a test.

I am not sure the tests need to change given the code change is not supposed to change anything in the behavior, unless I a missing something and it is crucial that this function isn't called?

ori-moisis (approved):

packages/workspace/test/workspace/dependents.test.ts
dantal4: it's not crucial, just a running time improvement. I'd like the next developer to be aware of it when changing code in that area, so we won't lose that improvement.

ori-moisis: I don't think a test makes sense for this purpose, maybe a comment in the code would do it

PR #6792, opened 2024-11-19
SALTO-6932: Fix implementation of Salesforce `picklistReferences` filter.

tamtamirr:
SALTO-6932: Fix implementation of Salesforce `picklistReferences` filter.

---

Workspace Diff:https://github.com/salto-io/tamir-sf/commit/bfe349bc385b6f74e9b66887406bdbfffb294f67
The previous implementation created broken references which made the Workspace invalid.

---

_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/filters/picklist_references.ts
```diff
@@ -36,124 +40,128 @@ const getValueSetFieldName = (typeName: string): string => {
...
-        ORDERED_MAP_VALUES_FIELD,
-        naclCase(decodeURIComponent(value.fullName!)),
-      ),
-    ),
-  }))
+  recordTypePicklistValues.forEach(({ picklist, values }) => {
+    const field = picklist.value
+    const valueSetElement = getValueSetElementFromPicklistField(field)
+    values.forEach(value => {
+      const ref: ReferenceExpression | undefined =
```
yelly, line 100: Why do you need the explicit typing? Doesn't the `?.` tell TS that this can be `undefined`?

packages/salesforce-adapter/src/filters/picklist_references.ts
```diff
@@ -36,124 +40,128 @@ const getValueSetFieldName = (typeName: string): string => {
...
-      ...rest,
+// Create reference index from baseElements full names (either Field, GlobalValueSet or StandardValueSet instances)
+// with references to each of their valueSet values.
+const createPicklistValuesReferenceIndex = (elements: Element[]): PicklistValuesReferenceIndex => {
+  const result: PicklistValuesReferenceIndex = {}
+  elements.filter(isInstanceOfTypeSync(GLOBAL_VALUE_SET, STANDARD_VALUE_SET)).forEach(instance => {
+    const fieldName = getValueSetFieldName(metadataTypeSync(instance))
+    const values = instance.value[fieldName]
+    if (!isOrderedValueSet(values)) {
+      return
```
yelly, line 119: Is this expected? Do we want to log?

packages/salesforce-adapter/src/filters/picklist_references.ts
```diff
@@ -36,124 +40,128 @@ const getValueSetFieldName = (typeName: string): string => {
...
     }
+    const refIndex: Record<string, ReferenceExpression> = {}
+    Object.entries(values.values).forEach(([key, value]) => {
+      const fullName = _.isString(value.fullName) ? value.fullName : apiNameSync(value.fullName.value) ?? ''
+      refIndex[fullName] = new ReferenceExpression(
```
yelly, line 124: What if `fullName` is `''`?

packages/salesforce-adapter/src/filters/picklist_references.ts
```diff
@@ -36,124 +40,128 @@ const getValueSetFieldName = (typeName: string): string => {
...
+    })
+    result[instance.elemID.getFullName()] = refIndex
   })
+  elements
+    .filter(isObjectType)
+    .flatMap(object => Object.values(object.fields))
+    .forEach(field => {
+      const valueSet = field.annotations[FIELD_ANNOTATIONS.VALUE_SET]
+      if (!isOrderedValueSet(valueSet)) {
+        return
```
yelly, line 137: Is this expected? Do we want to log?

packages/salesforce-adapter/src/filters/picklist_references.ts
```diff
@@ -36,124 +40,128 @@ const getValueSetFieldName = (typeName: string): string => {
...
+      }
+      const refIndex: Record<string, ReferenceExpression> = {}
+      Object.entries(valueSet.values).forEach(([key, value]) => {
+        const fullName = _.isString(value.fullName) ? value.fullName : apiNameSync(value.fullName.value) ?? ''
+        refIndex[fullName] = new ReferenceExpression(
+          field.elemID.createNestedID(FIELD_ANNOTATIONS.VALUE_SET, ORDERED_MAP_VALUES_FIELD, key, 'fullName'),
+          fullName,
+        )
+      })
+      result[field.elemID.getFullName()] = refIndex
```
yelly, line 147: This looks a lot like the previous block. Can we maybe share the code?

packages/salesforce-adapter/src/filters/picklist_references.ts
tamtamirr: Nope
Screenshot 2024-11-20 at 12 07 03

packages/salesforce-adapter/src/filters/picklist_references.ts
tamtamirr: This is expected when the feature `picklistAsMaps` is disabled. A log here will be noisy. We can discuss adding the log when the feature is enabled and non optional anymore

packages/salesforce-adapter/src/filters/picklist_references.ts
tamtamirr: not an actual scenario. Don't worth special handling IMO

packages/salesforce-adapter/src/filters/picklist_references.ts
tamtamirr: Same comment as above

packages/salesforce-adapter/src/filters/picklist_references.ts
yelly: Walla

packages/salesforce-adapter/src/filters/picklist_references.ts
```diff
@@ -36,124 +40,125 @@ const getValueSetFieldName = (typeName: string): string => {
...
-    ),
-  }))
+  recordTypePicklistValues.forEach(({ picklist, values }) => {
+    const field = picklist.value
+    const valueSetElement = getValueSetElementFromPicklistField(field)
+    values.forEach(value => {
+      const ref: ReferenceExpression | undefined =
+        // Using URI decode since RecordType Picklist values return as encoded URI string.
+        // e.g. 'You %26 Me' instead of 'You & Me'.
+        picklistValuesReferenceIndex[valueSetElement.elemID.getFullName()]?.[decodeURIComponent(value.fullName)]
```
yelly, line 103: `decodeURIComponent` can throw an error if the input is invalid - maybe we should catch it?
IMO if we get an error here we should just use the value without decoding.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent#catching_errors

yelly (approved):
Just a small comment, take it or leave it.

PR #6793, opened 2024-11-20
SALTO-6939: Add the ability to disable custom references from config (Jira)

shir-reifenberg:
Add the ability to disable / enable specific custom references funcs from Jira config

---
Added this ability as in certain large Jira envs we might create too many references
By default all custom reference func are enabled

Usage example - 
```
jira {
  fetch = { ... }
  customReferences = {
    fieldConfigurationsHandler = false
  }
}
```

---
_Release Notes_: 
None

---
_User Notifications_: 
None


kami-delaroz (approved):
🔥

PR #6801, opened 2024-11-20
SALTO-6939: Add reference to SLA's conditionId

IdoZakk:
Added references the the SLA's conditionId
---

See ticket.
JSM SLA has some conditions that can be defined. Some of them are for specific statuses, and they require reference in the conditioned field.
Also, I changed the deployment annotation for the name field, as changes to it will not change anything in the service

Noise reduction will follow

---
_Release Notes_: 
Jira Adapter:
* Added reference support to SLA's status conditions

---
_User Notifications_: 
Jira Adapter:
* SLA's conditioned can turn from a number to a reference

### DIFF
--- a/packages/jira-adapter/src/filters/jsm_types_fetch_filter.ts
+++ b/packages/jira-adapter/src/filters/jsm_types_fetch_filter.ts
@@ -6,7 +6,7 @@
  * CERTAIN THIRD PARTY SOFTWARE MAY BE CONTAINED IN PORTIONS OF THE SOFTWARE. See NOTICE FILE AT https://github.com/salto-io/salto/blob/main/NOTICES
  */
 
-import { isObjectType, CORE_ANNOTATIONS, isInstanceElement } from '@salto-io/adapter-api'
+import { isObjectType, CORE_ANNOTATIONS, isInstanceElement, TypeReference, BuiltinTypes } from '@salto-io/adapter-api'
 import { collections } from '@salto-io/lowerdash'
 import { FilterCreator } from '../filter'
 import {
@@ -25,8 +25,11 @@ import {
   OBJECT_SCHMEA_DEFAULT_REFERENCE_TYPE_TYPE,
   OBJECT_TYPE_ICON_TYPE,
   OBJECT_SCHEMA_GLOBAL_STATUS_TYPE,
+  SLA_CONDITIONS_STOP_TYPE,
+  SLA_CONDITIONS_START_TYPE,
+  SLA_CONDITIONS_PAUSE_TYPE,
 } from '../constants'
-import { setTypeDeploymentAnnotations, addAnnotationRecursively, findObject } from '../utils'
+import { setTypeDeploymentAnnotations, addAnnotationRecursively, findObject, findObjects } from '../utils'
 
 const { awu } = collections.asynciterable
 const jsmSupportedTypes = [
@@ -76,6 +79,24 @@ const filterCreator: FilterCreator = ({ config }) => ({
     if (objectTypeIconType !== undefined) {
       objectTypeIconType.fields.icon.annotations[CORE_ANNOTATIONS.UPDATABLE] = false
     }
+
+    // SLA
+    const slaConditionTypes = findObjects(elements, [
+      SLA_CONDITIONS_STOP_TYPE,
+      SLA_CONDITIONS_START_TYPE,
+      SLA_CONDITIONS_PAUSE_TYPE,
+    ])
+    if (slaConditionTypes === undefined) {
+      return
+    }
+    slaConditionTypes.forEach(slaType => {
+      slaType.fields.conditionId.refType = new TypeReference(BuiltinTypes.UNKNOWN.elemID, BuiltinTypes.UNKNOWN)
+      // prevent deployment of name as it does nothing
+      if (slaType.fields.name !== undefined) {
+        slaType.fields.name.annotations[CORE_ANNOTATIONS.CREATABLE] = false
+        slaType.fields.name.annotations[CORE_ANNOTATIONS.UPDATABLE] = false
+      }
+    })
   },
 })
 export default filterCreator
--- a/packages/jira-adapter/src/utils.ts
+++ b/packages/jira-adapter/src/utils.ts
@@ -63,7 +63,8 @@ export const findObjects = (elements: Element[], names: string[]): ObjectType[] 
   const types = elements.filter(isObjectType).filter(element => names.includes(element.elemID.name))
 
   if (types.length !== names.length) {
-    log.warn(`some of ${names.join(',')} types not found`)
+    const missingTypes = names.filter(name => !types.map(t => t.elemID.name).includes(name)).join(', ')
+    log.warn(`could not find the following objects: ${missingTypes}`)
     return undefined
   }
   return types
--- a/packages/jira-adapter/src/utils.ts
+++ b/packages/jira-adapter/src/utils.ts
@@ -59,6 +59,16 @@ export const findObject = (elements: Element[], name: string): ObjectType | unde
   return type
 }
 
+export const findObjects = (elements: Element[], names: string[]): ObjectType[] | undefined => {
+  const types = elements.filter(isObjectType).filter(element => names.includes(element.elemID.name))
+
+  if (types.length !== names.length) {
+    log.warn(`some of ${names.join(',')} types not found`)
+    return undefined
+  }
+  return types
+}
+
 export const addAnnotationRecursively = async (type: ObjectType, annotation: string): Promise<void> =>
   awu(Object.values(type.fields)).forEach(async field => {
     if (!field.annotations[annotation]) {
--- a/packages/jira-adapter/test/filters/jsm_types_fetch_filter.test.ts
+++ b/packages/jira-adapter/test/filters/jsm_types_fetch_filter.test.ts
@@ -8,11 +8,26 @@
 
 import { filterUtils } from '@salto-io/adapter-components'
 import _ from 'lodash'
-import { InstanceElement, ReferenceExpression, Element, CORE_ANNOTATIONS } from '@salto-io/adapter-api'
+import {
+  InstanceElement,
+  ReferenceExpression,
+  Element,
+  CORE_ANNOTATIONS,
+  ObjectType,
+  ElemID,
+  BuiltinTypes,
+} from '@salto-io/adapter-api'
 import { getDefaultConfig } from '../../src/config/config'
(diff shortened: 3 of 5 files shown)


packages/jira-adapter/src/filters/jsm_types_fetch_filter.ts
```diff
@@ -76,6 +79,24 @@ const filterCreator: FilterCreator = ({ config }) => ({
...
+    ])
+    if (slaConditionTypes === undefined) {
+      return
+    }
+    slaConditionTypes.forEach(slaType => {
+      slaType.fields.conditionId.refType = new TypeReference(BuiltinTypes.UNKNOWN.elemID, BuiltinTypes.UNKNOWN)
+      // prevent deployment of name as it does nothing
+      if (slaType.fields.name !== undefined) {
+        slaType.fields.name.annotations[CORE_ANNOTATIONS.CREATABLE] = false
+        slaType.fields.name.annotations[CORE_ANNOTATIONS.UPDATABLE] = false
```
idoamit8, line 97: Did you check it? I am not sure it works

packages/jira-adapter/test/filters/jsm_types_fetch_filter.test.ts
```diff
@@ -79,5 +94,88 @@ describe('jsmTypesFetchFilter', () => {
...
+        fields: {
+          icon: {
+            refType: BuiltinTypes.STRING,
+            annotations: { [CORE_ANNOTATIONS.UPDATABLE]: true },
+          },
+        },
+      })
+      elements.push(objectTypeIconType)
+      await filter.onFetch(elements)
+      expect(objectTypeIconType.fields.icon.annotations[CORE_ANNOTATIONS.UPDATABLE]).toBe(false)
```
idoamit8, line 109: Is this intentional?

packages/jira-adapter/src/utils.ts
```diff
@@ -59,6 +59,16 @@ export const findObject = (elements: Element[], name: string): ObjectType | unde
   return type
 }
 
+export const findObjects = (elements: Element[], names: string[]): ObjectType[] | undefined => {
+  const types = elements.filter(isObjectType).filter(element => names.includes(element.elemID.name))
+
+  if (types.length !== names.length) {
```
idoamit8, line 65: We can be more explicit here and add 
```suggestion
if (types.length < names.length) {
```

packages/jira-adapter/src/utils.ts
```diff
@@ -59,6 +59,16 @@ export const findObject = (elements: Element[], name: string): ObjectType | unde
...
+    log.warn(`some of ${names.join(',')} types not found`)
```
idoamit8, line 66: nit: 
I think its better to return the the actual types that were not found

packages/jira-adapter/src/utils.ts
```diff
@@ -59,6 +59,16 @@ export const findObject = (elements: Element[], name: string): ObjectType | unde
   return type
 }
 
+export const findObjects = (elements: Element[], names: string[]): ObjectType[] | undefined => {
+  const types = elements.filter(isObjectType).filter(element => names.includes(element.elemID.name))
```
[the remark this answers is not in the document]
IdoZakk: What is the benefit of a Set here?

packages/jira-adapter/src/filters/jsm_types_fetch_filter.ts
IdoZakk: Yes. It issues a warning when you try to modify or create the name

packages/jira-adapter/test/filters/jsm_types_fetch_filter.test.ts
IdoZakk: Yes, I added it to increase coverage

packages/jira-adapter/src/utils.ts
IdoZakk: Do you mean the log? if so I agree and fixing
If you mean the function I think the error handling should be done in the actual function. The purpose of this function is all-or-nothing

idoamit8: I meant the log 🙂

packages/jira-adapter/src/utils.ts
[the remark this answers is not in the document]
idoamit8: Using a Set instead of an array has several benefits.
In this case, I think the biggest advantage is clarity: a Set makes it clear that names represents a collection of **unique** values.
I understand that the list is small and there is no significant difference in performance, but I believe it’s better practice because utility functions often evolve, and starting with the right data structure avoids the need for refactoring later.

packages/jira-adapter/test/filters/jsm_types_fetch_filter.test.ts
```diff
@@ -79,5 +94,88 @@ describe('jsmTypesFetchFilter', () => {
...
+              },
+            },
+          }),
+      )
+      elements.push(...slaTypes)
+      await filter.onFetch(elements)
+      expect(slaTypes[0].fields.conditionId.refType.elemID.name).toEqual(BuiltinTypes.STRING.elemID.name)
+      expect(slaTypes[1].fields.conditionId.refType.elemID.name).toEqual(BuiltinTypes.STRING.elemID.name)
+    })
+    it('should disable deployment of name field in SLA types', async () => {
```
idoamit8, line 156: nit: If it’s not actually disabling it but just adding a warning, this title is misleading.

packages/jira-adapter/test/filters/jsm_types_fetch_filter.test.ts
```diff
@@ -79,5 +94,88 @@ describe('jsmTypesFetchFilter', () => {
...
+      const slaTypeNames = [SLA_CONDITIONS_STOP_TYPE, SLA_CONDITIONS_START_TYPE, SLA_CONDITIONS_PAUSE_TYPE]
+      const slaTypes = slaTypeNames.map(
```
idoamit8, line 158: nit: Lets export this to a function since you are using it 3 times

packages/jira-adapter/src/utils.ts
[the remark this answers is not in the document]
IdoZakk: I agree that the right DS should be used in utility functions. However, I do not feel a set is the right DS here. 
* The input is an array, so moving to a Set has a cost (that will increase with size)
* There should be no duplicates in this case, as there should be no two object types which are similar
* Searching a set and an array has the same complexity (assuming the same size)
* A hash will have a search advantage. The goal of this function is for bulk actions on types. If specific types are needed, several calls to the  single function will be more appropriate IMO.

packages/jira-adapter/src/utils.ts
IdoZakk: As the types cannot be bigger than names (object type names are unique) I do not think it makes a difference, and IMO it is more readable with `!==`

packages/jira-adapter/src/utils.ts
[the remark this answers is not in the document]
idoamit8: I see things differently. I believe the input can also be a Set for better clarity and to enforce uniqueness. Set.has() offers O(1) performance compared to Array.includes() at O(n). That said, it’s your code, and what ever you choose here is acceptable 🙂

packages/jira-adapter/test/filters/jsm_types_fetch_filter.test.ts
```diff
@@ -79,5 +95,86 @@ describe('jsmTypesFetchFilter', () => {
...
+              conditionId: {
+                refType: BuiltinTypes.STRING,
+              },
+              name: {
+                refType: BuiltinTypes.STRING,
+                annotations: { [CORE_ANNOTATIONS.CREATABLE]: true, [CORE_ANNOTATIONS.UPDATABLE]: true },
+              },
+            },
+          }),
+      )
```
idoamit8, line 127: nit: lets move it to be function
```suggestion
const createSlaType = (typeName: string): ObjectType => 
  new ObjectType({
    elemID: new ElemID(JIRA, typeName),
    fields: {
      conditionId: {
        refType: BuiltinTypes.STRING,
      },
      name: {
        refType: BuiltinTypes.STRING,
        annotations: {
          [CORE_ANNOTATIONS.CREATABLE]: true,
          [CORE_ANNOTATIONS.UPDATABLE]: true,
        },
      },
    },
  });

const slaTypes = slaTypeNames.map(createSlaType);
```

packages/jira-adapter/src/filters/jsm_types_fetch_filter.ts
```diff
@@ -76,6 +79,24 @@ const filterCreator: FilterCreator = ({ config }) => ({
...
+    const slaConditionTypes = findObjects(elements, [
+      SLA_CONDITIONS_STOP_TYPE,
+      SLA_CONDITIONS_START_TYPE,
+      SLA_CONDITIONS_PAUSE_TYPE,
+    ])
+    if (slaConditionTypes === undefined) {
+      return
+    }
+    slaConditionTypes.forEach(slaType => {
+      slaType.fields.conditionId.refType = new TypeReference(BuiltinTypes.UNKNOWN.elemID, BuiltinTypes.UNKNOWN)
```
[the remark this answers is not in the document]
idoamit8: Does it work with a `String`?

idoamit8 (approved):
Great work! 💯 
Added few last comments! 
Approving to save cycle 🍏

packages/jira-adapter/src/filters/jsm_types_fetch_filter.ts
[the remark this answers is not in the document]
IdoZakk: It is originally with a string. The reason I added it was to change it.
I did some testing, I did not find a problem if I keep it a string. I do remember having problems in the past with such things, I do not remember the details.
The best practice is to keep the type aligned with the possible data options. As we do not have an `Or` on types we use UNKNOWN for those cases.
What is the advantage of keeping it string?

packages/jira-adapter/src/utils.ts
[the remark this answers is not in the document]
IdoZakk: `Set.has` loops on elements (in [JS](https://tc39.es/ecma262/multipage/keyed-collections.html#sec-set.prototype.has) in particular, but also in general)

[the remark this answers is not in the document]
ori-moisis: @IdoZakk this is from the link you sent:

> Set objects must be implemented using either hash tables or other mechanisms that, on average, provide access times that are sublinear on the number of elements in the collection. The data structure used in this specification is only intended to describe the required observable semantics of Set objects. It is not intended to be a viable implementation model.

PR #6802, opened 2024-11-21
Salto 6917 sort lists by relevant types

kami-delaroz:
Salto 6917 sort lists by relevant types

---

None

---
_Release Notes_: 
None
---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/sort_lists.ts
+++ b/packages/jira-adapter/src/filters/sort_lists.ts
@@ -136,7 +136,7 @@ const TYPES_AND_VALUES_TO_SORT: Record<string, Record<string, Record<string, str
     WorkflowTransitions: {
       properties: ['key'],
     },
-    //TODO: add also WorkflowCondition
+    // TODO: add also WorkflowCondition
   }
 }
 
--- a/packages/jira-adapter/src/filters/sort_lists.ts
+++ b/packages/jira-adapter/src/filters/sort_lists.ts
@@ -137,7 +137,7 @@ const TYPES_AND_VALUES_TO_SORT: Record<string, Record<string, Record<string, str
       properties: ['key'],
     },
     // TODO: add also WorkflowCondition
-  }
+  },
 }
 
 const getValue = (value: Value): Value => (isResolvedReferenceExpression(value) ? value.elemID.getFullName() : value)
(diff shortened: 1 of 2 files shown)


packages/jira-adapter/test/filters/sort_lists.test.ts
```diff
@@ -228,7 +234,7 @@ describe('sortListsFilter', () => {
 
     it('should sort inner lists', async () => {
       const type = new ObjectType({
-        elemID: new ElemID(JIRA, DASHBOARD_TYPE),
+        elemID: new ElemID(JIRA, PERMISSION_SCHEME_TYPE_NAME),
```
shayc331, line 237: why do we need this change?

packages/jira-adapter/src/filters/sort_lists.ts
```diff
@@ -48,114 +47,102 @@ const WORKFLOW_CONDITION_SORT_BY = [
   'configuration.comparisonType',
 ]
 
-const VALUES_TO_SORT: Record<string, Record<string, string[]>> = {
-  PermissionScheme: {
-    permissions: ['permission', 'holder.type', 'holder.parameter'],
+const TYPES_AND_VALUES_TO_SORT: Record<string, Record<string, Record<string, string[]>>> = {
+  [PERMISSION_SCHEME_TYPE_NAME]: {
+    PermissionScheme: {
```
shayc331, line 52: we can use the constants here

shayc331 (commented):
looking good, had a small comment

packages/jira-adapter/src/filters/sort_lists.ts
kami-delaroz: Yep, I guess we could use more constants, just kept the orig info and added the mapping above

packages/jira-adapter/test/filters/sort_lists.test.ts
kami-delaroz: because the types they used within the test are relevant to PERMISSION_SCHEME_TYPE_NAME, and now with the mapping it won't work without the right type above

packages/jira-adapter/src/filters/sort_lists.ts
shayc331: I see, it just looks weird that some of the entries use constant and the others dont

shayc331 (approved):

packages/jira-adapter/src/filters/sort_lists.ts
kami-delaroz: changed :)

PR #6803, opened 2024-11-21
SALTO-1143 add a fallback logic for element open command

rotem531:
In the CLI "salto element open <element>", if the element doesn't have service URL, it will open the parent's service URL

---

when running in the CLI "salto element open <element>", if the element doesn't have service URL, than if the parent of the element has service URL ,it will open the parent's service URL

---
_Release Notes_: 
CLI:
upgraded behavior to salto element open <elementId> to open parent's service URL if <elementId> has no service URL
---
_User Notifications_: 
none


packages/cli/src/commands/element.ts
```diff
@@ -528,8 +529,26 @@ export const openAction: WorkspaceCommandAction<OpenActionArgs> = async ({ input
...
+    return CliExitCode.AppError
+  }
 
-  const serviceUrl = isElement(element) ? element.annotations[CORE_ANNOTATIONS.SERVICE_URL] : undefined
+  const getParentUrl = async (childElement: Value): Promise<string | undefined> => {
+    const parentsArray = getParents(childElement)
+    if (parentsArray.length === 0) {
+      return undefined
+    }
+    const parent = parentsArray[0]
```
shir-reifenberg, line 542: Didn't review the entire PR, but in some cases we have more than a single parent. If we don't find a service url for the first parent, can we please iterate on the other parents ? (I think getting the first url found is a good enough behavior)

Uri-Bar: can you share an example for this use case?
when you have multiple parents and you dont want to take the url from the first one (or the first one does not have url)
I know parents annotation is an array, but we have many places when we assume it has only one parent

rotem531: certainly possible,
i will note that in the utils, the functions that use getParents, throw error if the length of the list of parents isn't 1, this is why I used this logic

Uri-Bar (commented):
looks good, added few comments

packages/cli/src/commands/element.ts
shir-reifenberg: In Okta Authorization Server policies, see [this code](https://github.com/salto-io/salto/blob/main/packages/okta-adapter/src/filters/authorization_server_rule.ts#L31) (but I honestly not familiar with all cases)
We should only use `getParents` when we're certain that there is one parent (as this can throw, as you said)

Also core supports this, so I don't see a reason to assume there's only 1 parent
you can use .find to iterate parents array and finds the first one

packages/cli/src/commands/element.ts
```diff
@@ -514,6 +515,23 @@ const safeGetElementId = (maybeElementIdPath: string, output: CliOutput): ElemID
...
+  const parentRef = await awu(parentsArray).find(async parent => {
+    const parentElement = await workspace.getValue(parent.elemID)
+    if (!isElement(parentElement)) {
+      return false
+    }
+    return parentElement.annotations[CORE_ANNOTATIONS.SERVICE_URL] !== undefined
+  })
+
+  const parentElem = parentRef !== undefined ? await workspace.getValue(parentRef.elemID) : undefined
+  return isElement(parentElem) ? parentElem.annotations[CORE_ANNOTATIONS.SERVICE_URL] : undefined
```
Uri-Bar, line 532:
```suggestion
  const url = await awu(parentsArray)
    .map(async parentEntry => {
      const parent = await workspace.getValue(parentEntry.elemID)
      return parent.annotations[CORE_ANNOTATIONS.SERVICE_URL]
    })
    .find(serviceUrl => {
      return serviceUrl !== undefined
    })
  return url
```
the point of doing it in that way was to not use the "expenisive" func of getValue more than we need, so let try to think of a way that we dont need to call it again after the iteration.
this is only a suggestion of direction

rotem531: ill try to figure it out
i tried some ways to avoid the calliing of the getValue function again after the iteration
but had some issues trying to do so

Uri-Bar (approved):
thanks for that

PR #6804, opened 2024-11-21
removedLog

rotem531:
removed log from formula_deps

---

removed noisy log from formula_deps 

---
_Release Notes_: 
salseforce-adapter:
removed noisy log

---
_User Notifications_: 
none


Uri-Bar (approved):
great

PR #6806, opened 2024-11-21
SALTO-6946: Protecting unsafe call to toMD5

yelly:
Looks like we can get `undefined` here in some flows.

---

None.

---

_Release Notes_: 
None.

---

_User Notifications_: 
None.

### DIFF
--- a/packages/salesforce-adapter/src/client/client.ts
+++ b/packages/salesforce-adapter/src/client/client.ts
@@ -252,6 +252,10 @@ const oauthConnection = (params: OauthConnectionParams): Connection => {
   })
 
   conn.on('refresh', accessToken => {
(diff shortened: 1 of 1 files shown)


packages/salesforce-adapter/src/client/client.ts
```diff
@@ -252,6 +252,10 @@ const oauthConnection = (params: OauthConnectionParams): Connection => {
   })
 
   conn.on('refresh', accessToken => {
+    if (typeof accessToken !== 'string') {
```
tamtamirr, line 255:
```suggestion
    if (!_.isString(accessToken)) {
```

tamtamirr (approved):
small comment.

PR #6808, opened 2024-11-21
SALTO-5888 Jira DC: Fail on API-config with JSM on

kami-delaroz:
SALTO-5888 Jira DC: Fail on API-config with JSM on

---

None

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/adapter_creator.ts
+++ b/packages/jira-adapter/src/adapter_creator.ts
@@ -39,10 +39,12 @@ function validateConfig(config: Values, isDataCenter: boolean): asserts config i
   Object.values(getApiDefinitions(apiDefinitions)).forEach(swaggerDef => {
     validateSwaggerApiDefinitionConfig('apiDefinitions', swaggerDef)
   })
-  if(isDataCenter) {
-    if(fetch.enableJSM || fetch.enableJSMPremium || fetch.enableJsmExperimental) {
+  if (isDataCenter) {
+    if (fetch.enableJSM || fetch.enableJSMPremium || fetch.enableJsmExperimental) {
       log.error('JSM is not supported in Jira DC config')
-      throw new Error('Failed to load Jira config. JSM is not supported for Jira DC, please remove enableJSM flag from the config file and try again')
+      throw new Error(
+        'Failed to load Jira config. JSM is not supported for Jira DC, please remove enableJSM flag from the config file and try again',
+      )
     }
   }
   validateJiraFetchConfig({
--- a/packages/jira-adapter/src/adapter_creator.ts
+++ b/packages/jira-adapter/src/adapter_creator.ts
@@ -44,7 +44,7 @@ function validateConfig(config: Values, isDataCenter: boolean): asserts config i
       log.error('JSM is not supported in Jira DC config')
       throw new Error(
         'Failed to load Jira config. JSM is not supported for Jira DC, please remove enableJSM flag from the config file and try again\n' +
-        'More information about Jira configuration options in Salto can be found here: https://github.com/salto-io/salto/blob/main/packages/jira-adapter/config_doc.md'
+          'More information about Jira configuration options in Salto can be found here: https://github.com/salto-io/salto/blob/main/packages/jira-adapter/config_doc.md',
       )
     }
   }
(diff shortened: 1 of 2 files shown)


shayc331 (commented):
I am not sure if this is the change we need to solve it. IIUC The user experience does not change much after this pr, did you check how it looks from the UI side?
IMO the change should be in the SAAS, we need to remove the jsm checkbox when connecting new jira DC service. lets talk f2f

packages/jira-adapter/src/adapter_creator.ts
```diff
@@ -39,6 +39,11 @@ function validateConfig(config: Values): asserts config is JiraConfig {
   Object.values(getApiDefinitions(apiDefinitions)).forEach(swaggerDef => {
     validateSwaggerApiDefinitionConfig('apiDefinitions', swaggerDef)
   })
+  if(isDataCenter) {
+    if(fetch.enableJSM || fetch.enableJSMPremium || fetch.enableJsmExperimental) {
```
shir-reifenberg, line 43: nit: extract to a function in the spirit of `isDCWithJsmEnabled()`

packages/jira-adapter/src/adapter_creator.ts
```diff
@@ -39,6 +39,11 @@ function validateConfig(config: Values): asserts config is JiraConfig {
...
+      throw new Error('JSM is not supported for Jira DC')
```
shir-reifenberg, line 44: 1. Can we please add a log?
2. This is a user visible error. I think the error should be a little more informative and guide the user how to resolve it

packages/jira-adapter/src/adapter_creator.ts
```diff
@@ -53,13 +58,14 @@ function validateConfig(config: Values): asserts config is JiraConfig {
...
   defaultConfig: JiraConfig,
+  isDataCenter: boolean,
 ): JiraConfig => {
   const configWithoutFetch = mergeWithDefaultConfig(
     _.omit(defaultConfig, 'fetch'),
     _.omit(config?.value ?? {}, 'fetch'),
   )
   const fetch = _.defaults({}, config?.value.fetch, defaultConfig.fetch)
-  const fullConfig = { ...configWithoutFetch, fetch }
+  const fullConfig = { ...configWithoutFetch, fetch, isDataCenter }
```
shir-reifenberg, line 68: why are we assigning this to the config ? I find it a little strange to assign it to the config when there is no `isDataCenter` field in it
I think we should just add `isDataCenter` as an additional param

shir-reifenberg (commented):
Added some comments 🙃

shir-reifenberg:
> I am not sure if this is the change we need to solve it. IIUC The user experience does not change much after this pr, did you check how it looks from the UI side? IMO the change should be in the SAAS, we need to remove the jsm checkbox when connecting new jira DC service. lets talk f2f

@shayc331 IIRC the error is visible (though not in an ideal way)

packages/jira-adapter/src/adapter_creator.ts
kami-delaroz: these JSM flags are redundant and I'm planning to get rid of soon

kami-delaroz:
@shayc331 @shir-reifenberg indeed the error is visible

packages/jira-adapter/src/adapter_creator.ts
```diff
@@ -39,6 +39,14 @@ function validateConfig(config: Values): asserts config is JiraConfig {
   Object.values(getApiDefinitions(apiDefinitions)).forEach(swaggerDef => {
     validateSwaggerApiDefinitionConfig('apiDefinitions', swaggerDef)
   })
+  if (isDataCenter) {
+    if (fetch.enableJSM || fetch.enableJSMPremium || fetch.enableJsmExperimental) {
+      log.error('JSM is not supported in Jira DC config')
+      throw new Error(
+        'Failed to load Jira config. JSM is not supported for Jira DC, please remove enableJSM flag from the config file and try again',
```
shir-reifenberg, line 46: I think the part of `Failed to load Jira config` is not really user friendly and can probably be removed
Also we should probably point to https://help.salto.io/en/articles/7439324-salto-adapter-configuration-file

shir-reifenberg (approved):
LGTM! 🔥 
Can you please add a test covering this ?

packages/jira-adapter/src/adapter_creator.ts
kami-delaroz: I copied it from NS validation error :)
but the config doc is a good idea

PR #6810, opened 2024-11-21
SALTO-6828: Convert MS security permission errors into warning

shir-reifenberg:
Convert 403 errors to fetch warnings

---

I saw we get 403 when we lack of permissions, we can iterate once we see more relevant status codes.

---

_Release Notes_: 

_Microsoft Security Adapter_:
- Create fetch warning for while detecting there are missing permissions during fetch

---
_User Notifications_: 
None


shoshana-houri (approved):
💯 💯 NICE! Thank you!

PR #6813, opened 2024-11-21
SALTO-1380 post deployment summary

rotem531:
added post deployment summary to the cli
image
image
image
image

---

when deployment isn't fully successful, post deployment summary will be printed listing which elements where deployed successfully, which were partially successful and which have failed

---
_Release Notes_: 
CLI:
added post deployment summary

---
_User Notifications_: 
none

### DIFF
--- a/packages/cli/src/commands/deploy.ts
+++ b/packages/cli/src/commands/deploy.ts
@@ -92,27 +92,10 @@ const printDeploymentSummary = async (
     },
     {} as Record<DeploySummaryResult, DetailedChangeId[]>,
   )
-  outputLine(formatDeploymentSummary(dictionaryB), output)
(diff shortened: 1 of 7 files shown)


Uri-Bar:
please fix the title and fill the PR description (you can use the same ones as in the original PR)

packages/cli/src/commands/deploy.ts
```diff
@@ -78,6 +81,23 @@ const printStartDeploy = async (output: CliOutput, executingDeploy: boolean, che
   }
 }
 
+export const getReversedSummarizeDeployChanges = (
```
Uri-Bar, line 84: no export needed right?

Uri-Bar (approved):
approved with small nit.
please fix the title and the description before merging

PR #6815, opened 2024-11-21
SALTO-6213: Adding knip to CI

yelly:
Disabled the two most important issue types (exports and types) since we will onboard them gradually.

---

---

_Release Notes_: 
None.

---

_User Notifications_: 
None.

### DIFF
--- a/.circleci/config_template.yml
+++ b/.circleci/config_template.yml
@@ -1,4 +1,3 @@
-
 version: 2.1
 
 orbs:
@@ -12,7 +11,6 @@ environment:
   TURBO_CACHE_DIR: /mnt/ramdisk/project/.turbo
 
 commands:
-  
   restore_intermediate:
     description: Restore intermediate artifacts between steps
     steps:
@@ -46,7 +44,7 @@ commands:
             - &node-modules-cache node-modules-v2-{{ .Branch }}-{{ checksum "yarn.lock" }}
             - node-modules-v2-{{ .Branch }}
             - node-modules-v2-main
-  
+
   store_node_modules:
     description: Cache store node_modules
     steps:
@@ -199,12 +197,12 @@ commands:
 
   test_cli:
     parameters:
-        arch:
-            description: Linux, mac, win
-            type: string
-        exe_name:
-            description: Name of executable (e.g. salto, salto.exe)
-            type: string
+      arch:
+        description: Linux, mac, win
+        type: string
+      exe_name:
+        description: Name of executable (e.g. salto, salto.exe)
+        type: string
     steps:
       - aws-cli/install
       - aws-cli/setup
@@ -215,8 +213,8 @@ commands:
       - run: echo << parameters.arch >>
       - when:
           condition:
-            not: 
-              equal: [ win, << parameters.arch >> ]
+            not:
+              equal: [win, << parameters.arch >>]
           steps:
             - run: chmod +x ./<< parameters.exe_name >>
       - run:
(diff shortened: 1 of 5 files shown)


.circleci/config_template.yml
```diff
@@ -264,9 +262,24 @@ jobs:
           cache_name: build
       - store_intermediate
 
+  knip:
+    docker:
+      - image: cimg/node:18.9
+
+    resource_class: xlarge
```
ori-moisis, line 269: do we really need this to be `xlarge`?

.knip.json
ori-moisis, line 1: can we use a non-hidden file name? seems like it is possible for the documentation

.knip.json
```diff
@@ -0,0 +1,49 @@
+{
+  "$schema": "https://unpkg.com/knip@5/schema.json",
+  "rules": {
+    "exports": "off",
+    "types": "off"
+  },
+  "workspaces": {
+    ".": {
+      "entry": ["*.{js,mjs}", ".circleci/scripts/*.js", "build_utils/*.js"],
+      "project": ["*.{js,mjs}", ".circleci/**/*.js", "build_utils/**/*.js"],
```
ori-moisis, line 10: I get why we would want to define every file as an entry here, but why is it done in different ways?

.knip.json
```diff
@@ -0,0 +1,49 @@
...
+      "ignoreDependencies": ["jest-junit", "turbo"]
+    },
+    ".circleci": {
+      "entry": "scripts/*.js"
+    },
+    "build_utils": {
+      "entry": "*.js"
+    },
```
ori-moisis, line 18: isn't this a duplicate of the definition on `"."` above?
do we really need to have both?

.knip.json
```diff
@@ -0,0 +1,49 @@
...
+    "packages/cli": {
+      "entry": [
+        "index.ts",
+        "package_native.js",
+        "src/main.ts",
+        "e2e_test/**/*.test.ts",
+        "e2e_test/jest_environment.ts"
+      ],
+      "ignoreDependencies": ["webpack-cli"]
```
ori-moisis, line 34: isn't this supposed to be covered by the [webpack plugin](https://knip.dev/reference/plugins/webpack)?
is there a reason we need to do something non-default here? (beyond the package_native.js thing which I understand)

.circleci/config_template.yml
yelly: Yeah, I tried large and it didn't work.
It loads the whole repo so that makes some sense.

.knip.json
yelly: I used the hidden name to align with our eslint config, but I'll remove the dot.

.knip.json
yelly: Theoretically we could add some helper scripts deeper in these directories (I'm imagining something like `build_utils/helpers/something.js` or something like that).
In that case I want them to be covered by knip but not to be considered entry points.

.knip.json
yelly: Yes, nice catch.

.circleci/config_template.yml
ori-moisis: k

.knip.json
ori-moisis: k

package.json
```diff
@@ -24,9 +24,12 @@
     "globals": "^15.8.0",
     "jest-junit": "^16.0.0",
     "jsonc-parser": "^2.2.1",
+    "knip": "^5.37.1",
     "lerna": "^5.6.2",
     "lodash": "^4.17.21",
-    "turbo": "^2.0.6"
+    "tmp": "^0.0.33",
+    "turbo": "^2.0.6",
+    "yargs": "^13.3.0"
```
ori-moisis, line 32: why were `tmp` and `yargs` added here?

ori-moisis: nvm, I see they are used by the build utils
how come this was not caught by anything?

ori-moisis (approved):

package.json
yelly: Because I previously broke out the build utils into a separate workspace but there's no `package.json` so I guess knip just ignored dependencies (which is a reasonable approach).

PR #6818, opened 2024-11-24
SALTO-6947: Okta: Change ElemID computation for default domain and default brand

Nurdok:
See attached ticket for ~~more~~ too much details.

[Workspace diff](https://github.com/salto-io/amir-okta/commit/abb8015445ae4d80781ddd2f0c68da12fdf63b88#diff-058b213e882f65dfd2905b14b606eb3973dafbfddbc362857998118c702197b0) after running fetch with `--regenerate-salto-ids`

---

None

---
_Release Notes_: 
_Okta Adapter_:
- The default brand and domain for each account will now be named "Default Brand" and "Default Domain" respectively, to allow deployment of these instances between environments. This will happen in new environments, or in existing ones after regenerating Salto element IDs

---
_User Notifications_: 
- After regenerating Salto element IDs, the default brand and domain for each account will be renamed "Default Brand" and "Default Domain" respectively


packages/okta-adapter/src/definitions/fetch/fetch.ts
```diff
@@ -116,6 +116,30 @@ const getPrivateAPISettingsDefinitions = ({
   }
 }
 
+const brandCustomizer: definitions.fetch.FetchTopLevelElementDefinition['elemID'] = {
+  custom:
+    args =>
+    ({ entry, defaultName }) => {
+      if (entry?.isDefault) {
```
shir-reifenberg, line 123: We prefer explicit checks
```suggestion
      if (entry?.isDefault === true) {
```

packages/okta-adapter/src/definitions/fetch/fetch.ts
```diff
@@ -1077,7 +1102,11 @@ const createCustomizations = ({
     requests: [{ endpoint: { path: '/api/v1/domains' }, transformation: { root: 'domains' } }],
     resource: { directFetch: true },
     element: {
-      topLevel: { isTopLevel: true, elemID: { parts: [{ fieldName: 'domain' }], extendsParent: true } },
+      topLevel: {
+        isTopLevel: true,
+        elemID: { ...domainCustomizer, parts: [{ fieldName: 'domain' }], extendsParent: true },
```
shir-reifenberg, line 1107: just verifying we need both the customizer and the `parts`

shir-reifenberg (approved):
Great job! thanks (and thanks for adding the workspace diff)🙏 
Can you please update release notes to indicate it will only happen with `regenerate` ?

packages/okta-adapter/src/definitions/fetch/fetch.ts
Nurdok: Done

packages/okta-adapter/src/definitions/fetch/fetch.ts
Nurdok: We still need them, yes - the customizer only changes elemID computation for default domains, the rest are computed with the non-`custom` args

PR #6819, opened 2024-11-24
SALTO-5023: Do not allow Activating Flow of type InvalidDraft

AlmogMesilaty:
SALTO-5023: Do not allow Activating Flow of type InvalidDraft

---
image

---
_None_ 

---
_None_

### DIFF
--- a/packages/salesforce-adapter/src/change_validators/flows.ts
+++ b/packages/salesforce-adapter/src/change_validators/flows.ts
@@ -24,7 +24,7 @@ import {
 import { collections } from '@salto-io/lowerdash'
 import _, { isEmpty, isUndefined } from 'lodash'
 import { detailedCompare } from '@salto-io/adapter-utils'
-import { ACTIVE, FLOW_METADATA_TYPE, SALESFORCE, STATUS } from '../constants'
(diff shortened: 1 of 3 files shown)


packages/salesforce-adapter/src/change_validators/flows.ts
```diff
@@ -263,6 +270,9 @@ const activeFlowValidator =
       .filter(isActivatingChange)
       .map(change => {
         if (isActivatingChangeOnly(change)) {
+          if (getFlowStatus(change.data.before) === INVALID_DRAFT) {
```
Uri-Bar, line 273: are you sure you want to throw this error only if its activating change only? 
I think that we want to error this even if the user did other changes, so this needs to be outside this if

Uri-Bar (commented):
looks good, just one important comment

packages/salesforce-adapter/src/change_validators/flows.ts
AlmogMesilaty: I agree, done.

Uri-Bar (approved):
🔥

PR #6827, opened 2024-11-25
SALTO-6954: Removing some async calls from validators

yelly:
Won't solve all the problems but hopefully will lead to some performance gains.

---

---

_Release Notes_: 
None.

---

_User Notifications_: 
None.

### DIFF
--- a/packages/salesforce-adapter/src/change_validators/map_keys.ts
+++ b/packages/salesforce-adapter/src/change_validators/map_keys.ts
@@ -16,10 +16,8 @@ import {
   isObjectType,
   isAdditionOrModificationChange,
 } from '@salto-io/adapter-api'
-import { GetLookupNameFunc, TransformFunc, transformValues } from '@salto-io/adapter-utils'
+import { GetLookupNameFunc, TransformFuncSync, transformValuesSync } from '@salto-io/adapter-utils'
 import { resolveValues } from '@salto-io/adapter-components'
-
-import { collections } from '@salto-io/lowerdash'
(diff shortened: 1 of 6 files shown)


packages/salesforce-adapter/src/change_validators/map_keys.ts
```diff
@@ -44,72 +42,72 @@ const isNum = (str: string | undefined): boolean => !_.isEmpty(str) && !Number.i
...
-          const previewPrefix = actualPath.slice(0, actualPath.findIndex((val, idx) => val !== expectedPath[idx]) + 1)
-          if (!_.isEqual(actualPath, expectedPath)) {
-            errors.push({
-              elemID: after.elemID.createNestedID(fieldName, ...previewPrefix),
-              severity: 'Error',
-              message: 'Incorrect map key in field',
-              detailedMessage: `${typeName} ${after.value.fullName} field ${fieldName}: Incorrect map key ${actualPath?.join(API_NAME_SEPARATOR)}, should be ${expectedPath.join(API_NAME_SEPARATOR)}`,
-            })
-          }
+  await awu(
```
tamtamirr, line 48: We can avoid the async code here by using `transformValuesSync`

tamtamirr (commented):
LGTM! We can still avoid more promises. See my comment

packages/salesforce-adapter/src/change_validators/map_keys.ts
yelly: Done.

tamtamirr (approved):
LGTM

PR #6832, opened 2024-11-25
SALTO-6963: fix resolve types when passing multiple fragments

ori-moisis:
When multiple fragments of the same element were passed, only one of them was resolved This fixes the code so we will resolve all the fragments

---

This still leaves a bit of "unstable" behavior in that if there are references point to the element which is split, we will resolve them to point to one of the fragments, we do not guarantee which one

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/workspace/src/expressions.ts
+++ b/packages/workspace/src/expressions.ts
@@ -306,19 +306,16 @@ export const resolve = (
 
       // Since fields technically reference their parent type with the .parent property
       // we need to make sure to resolve all field parents as well
-      const fieldParents = elementsToResolve.filter(isField).map(field => field.parent)
+      const fieldParents = _.uniqBy(
+        elementsToResolve.filter(isField).map(field => field.parent),
+        elem => elem.elemID.getFullName(),
+      )
 
       // We assume that if we got a field and its parent type in the same resolve call, the field's
       // parent will point to the same object type that we got to resolve, so we don't need to resolve
       // both.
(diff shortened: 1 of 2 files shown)


packages/workspace/src/expressions.ts
```diff
@@ -311,7 +311,15 @@ export const resolve = (
...
       // both.
-      const allElementsToResolve = _.uniqBy(elementsToResolve.concat(fieldParents), elem => elem.elemID.getFullName())
+      const elementIDsToResolve = new Set(elements.map(elem => elem.elemID.getFullName()))
+      const fieldParentsToResolve = fieldParents.filter(elem => {
+        if (!elementIDsToResolve.has(elem.elemID.getFullName())) {
+          elementIDsToResolve.add(elem.elemID.getFullName())
+          return true
+        }
+        return false
+      })
```
Nurdok, line 321: IIUC, the issue with the previous implementation was that it dropped elements from `elementsToResolve`, and the new implementation always keep all of `elementsToResolve` while dropping elements from `fieldParentsToResolve` based on elemID uniqueness across _both_ groups?

I believe this works too, with the benefit that it doesn't have side effects in the `filter` callback:

```ts
const fieldParentsToResolve =
  _.uniqBy(fieldParents, elem => elem.elemID.getFullName())
  .filter(elem => !elementIDsToResolve.has(elem.elemID.getFullName()))
```

Nurdok (approved):

PR #6833, opened 2024-11-25
SALTO-6964: avoid generating source maps where they are not needed 

ori-moisis:
Also, refactor dummy adapter a bit to allow it to handle scale better

---

Based on #6832 

---
_Release Notes_: 
Core:
- Small performance improvements by reducing source map calculations

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/cli/src/main.ts
+++ b/packages/cli/src/main.ts
@@ -61,4 +61,6 @@ const main = async (): Promise<CliExitCode> => {
   }
 }
 // eslint-disable-next-line @typescript-eslint/no-floating-promises
-main().then(exitCode => process.exit(exitCode))
+main().then(exitCode => {
+  process.exitCode = exitCode
+})
--- a/packages/core/src/app_config.ts
+++ b/packages/core/src/app_config.ts
@@ -125,7 +125,7 @@ const mergeConfigWithEnv = async (config: AppConfig): Promise<AppConfig> => {
 
 const configFromNaclFile = async (filepath: string): Promise<AppConfig> => {
   const buf = await readFile(filepath)
-  const configInstance = (await awu((await parse(buf, filepath)).elements).peek()) as InstanceElement
+  const configInstance = (await awu((await parse(buf, filepath, {}, false)).elements).peek()) as InstanceElement
   if (!configInstance) throw new AppConfigParseError()
 
   configInstance.refType = createRefToElmWithValue(saltoAppConfigType)
--- a/packages/dummy-adapter/src/generator.ts
+++ b/packages/dummy-adapter/src/generator.ts
@@ -40,6 +40,7 @@ import {
   isReferenceExpression,
   SaltoError,
   SaltoElementError,
+  ReadOnlyElementsSource,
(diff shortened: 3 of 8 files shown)


packages/cli/src/main.ts
```diff
@@ -61,4 +61,6 @@ const main = async (): Promise<CliExitCode> => {
   }
 }
 // eslint-disable-next-line @typescript-eslint/no-floating-promises
-main().then(exitCode => process.exit(exitCode))
+main().then(exitCode => {
```
ori-moisis, line 64: unrelated fix - when calling `process.exit` we prevent `beforeExit` hooks to do async work.
this causes some injected tools like [clinic's heap profiler](https://clinicjs.org/heapprofiler/) to not work properly

packages/dummy-adapter/src/generator.ts
```diff
@@ -840,7 +847,11 @@ export const generateElements = async (
     }).flat()
   }
 
-  const generateUsersLike = ({ numOfUsers, numOfGroups, distributionFactor }: UsersGenerationParams): Element[] => {
+  const generateUsersLike = (): Element[] => {
```
ori-moisis, line 850: unrelated change - align the interface of this generate function to be the same as the other ones

packages/dummy-adapter/src/generator.ts
```diff
@@ -1102,28 +1114,30 @@ export const generateElements = async (
...
+  const allElements = defaultTypes
+    .concat(primitiveTypes)
+    .concat(types)
+    .concat(records)
+    .concat(objects)
+    .concat(profiles)
+    .concat(users)
+    .concat([new ObjectType({ elemID: new ElemID(DUMMY_ADAPTER, 'noPath'), fields: {} })])
+    .concat(envObjects)
+    .filter(e => !elementsToExclude.has(e.elemID.getFullName()))
```
ori-moisis, line 1131: the reason for the change in the code here is to avoid the `...`

packages/core/src/app_config.ts
```diff
@@ -125,7 +125,7 @@ const mergeConfigWithEnv = async (config: AppConfig): Promise<AppConfig> => {
 
 const configFromNaclFile = async (filepath: string): Promise<AppConfig> => {
   const buf = await readFile(filepath)
-  const configInstance = (await awu((await parse(buf, filepath)).elements).peek()) as InstanceElement
+  const configInstance = (await awu((await parse(buf, filepath, {}, false)).elements).peek()) as InstanceElement
```
Nurdok, line 128: IMO passing constants like this is forcing the reader to read the original signature, I suggest moving from a list of parameters to a single object parameter so this'll read as `functions: {}, calcSourceMap: false`.

Feel free to ignore.

Nurdok (approved):

packages/core/src/app_config.ts
ori-moisis: I fully agree
this is an external interface, so changing it / changing its behavior I think requires a separate PR

PR #6835, opened 2024-11-25
SALTO-6866: Switch MS Entra Administrative Unit deployment definition to use the beta API

shoshana-houri:
The current deploy definition, which uses the MS Graph v1.0 API, no longer works.

---

---
_Release Notes_: 
_Microsoft Security adapter:_
* Fix Administrative Unit deployment.

---
_User Notifications_:


shir-reifenberg (approved):
🔥 
(No need for user notifications, this is only relevant when there's a visible change in the workspace)

PR #6836, opened 2024-11-25
SALTO-6933: Microsoft Security - Add missing references under Entra app

shoshana-houri:
We have missing references in the `EntraApplication` under the `requiredResourceAccess` field. This PR addresses that by adding references to other apps/service principals/app roles.

Under this field we may also have references to OAuthScopes. These references are not included yet. Before adding them, we need to extract the OAuthScopes, which reside directly in a field under the application instance, into a standalone instance. This PR focuses only on the existing references for now.

You can see the before/after changes in this [commit](https://github.com/salto-io/microsoft-security/commit/e227545ed53f4c93cc595fa7af93aa96d02c880b).

---

---
_Release Notes_: 
_Microsoft Security adapter:_
* Add missing references under Entra Application.

---
_User Notifications_: 
_Microsoft Security adapter:_
* Add missing references under Entra Application.


packages/microsoft-security-adapter/src/definitions/references/entra_reference_rules.ts
```diff
@@ -209,4 +210,30 @@ export const REFERENCE_RULES: referenceUtils.FieldReferenceDefinition<
...
+    target: { type: APPLICATION_TYPE_NAME },
+    serializationStrategy: 'appId',
+  },
+  {
+    src: {
+      field: 'resourceAppId',
+      parentTypes: [recursiveNestedTypeName(APPLICATION_TYPE_NAME, REQUIRED_RESOURCE_ACCESS_FIELD_NAME)],
+    },
+    target: { type: SERVICE_PRINCIPAL_TYPE_NAME },
+    serializationStrategy: 'servicePrincipalAppId',
```
shoshana-houri, line 227: Please note that both of these reference rules apply to the same field but target different types.

This distinction is necessary because, in some cases, we don’t receive the application instance. However, if we reference its app roles or scopes, it means we do receive its Service Principal instance.

As a result, we prioritize referencing the application where possible, but we allow referencing the Service Principal if the application instance doesn’t exist.

This approach is similar to how we handle app roles that we’re extracting (and referencing here). If the Application instance is available, the app roles will reside there. Otherwise, if we only have the Service Principal, the app roles are extracted from it.

Here’s an example from our workspace referencing [an application](https://github.com/salto-io/microsoft-security/commit/e227545ed53f4c93cc595fa7af93aa96d02c880b#diff-8a4c0a0e8c40a3e8764392623d9ead29e845c7fe428f5f8ffdc7e8b977ffb1eaR10). And here’s one referencing [a Service Principal](https://github.com/salto-io/microsoft-security/commit/e227545ed53f4c93cc595fa7af93aa96d02c880b#diff-2f1da6d80e6bae3cd6d76ed2b5903505c31519674c01eb97e681dd7768cc2102R10).

packages/microsoft-security-adapter/src/definitions/references/references.ts
```diff
@@ -75,6 +75,12 @@ export const REFERENCES: definitions.ApiDefinitions<Options>['references'] = {
       getLookUpName: async ({ ref }) => ref.elemID.name,
       contextValueMapper: refType => entraConstants.SUPPORTED_DIRECTORY_OBJECTS_ODATA_TYPE_NAME_TO_TYPE_NAME[refType],
     }),
+    resourceAccessType: referenceUtils.neighborContextGetter({
+      contextFieldName: 'type',
+      getLookUpName: async ({ ref }) => ref.elemID.name,
+      // TODO SALTO-6933: Cover 'Scope' type
+      contextValueMapper: refType => (refType === 'Role' ? entraConstants.APP_ROLE_TYPE_NAME : undefined),
```
shir-reifenberg, line 82: nit: `refType` means something else in our code base, maybe just `value`?

packages/microsoft-security-adapter/src/definitions/references/entra_reference_rules.ts
shir-reifenberg: This is based on the assumption the references rules run by this order ?
Can we maybe achieve the dynamic target type using `ContextFunc`? 
I think we can dynamically calculate the serialization field using `GetLookupNameFunc`

shir-reifenberg (commented):
Added one question :)

packages/microsoft-security-adapter/src/definitions/references/entra_reference_rules.ts
shoshana-houri: @shir-reifenberg Yes, unfortunately it is.

Unless I’m missing something, the issue with dynamically determining the referenced type is that it doesn’t depend on an available field in the element itself. Instead, it relies on other fetched elements and whether the app was returned or not.

One way to handle this would be to make an additional call during fetch, using the `appId` returned on the `Service Principal`. This would allow us to create a hidden field on the `Service Principal` to indicate whether its app was successfully retrieved. We could then use this information when dynamically resolving the reference.

That said, I think the current approach, while implicit, should be fine for now. If any changes to this assumption are made in the reference calculation method, we should be able to detect them immediately in our `track-adapter-changes`.

shir-reifenberg (approved):
🚒

packages/microsoft-security-adapter/src/definitions/references/entra_reference_rules.ts
shoshana-houri: Good ideas, thanks! :)

PR #6838, opened 2024-11-26
SALTO-6976: Fix wrong "unsupported deploy" errors caused by default definitions

Nurdok:
This PR fixes a bug where types deployed by the old config were reported as not supporting deploy. The issue was that we checked if they were supported by the new definition config first, and instead of returning `undefined`, the query returned the default definitions which aren't enough to support deploy.
 
---

---
_Release Notes_: 
_Okta adapter_:
- Fixed a bug where some types were reported as not supporting deploy actions.

---
_User Notifications_: 
None


shir-reifenberg (approved):
Thanks!

PR #6841, opened 2024-11-26
SALTO-6978: Fix an SLA bug

IdoZakk:
Crashed as ducktypes might not contain the conditioned field
---

None

---
_Release Notes_: 
Jira Adapter:
* fixed a bug in SLA references

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/jsm_types_fetch_filter.ts
+++ b/packages/jira-adapter/src/filters/jsm_types_fetch_filter.ts
@@ -89,14 +89,16 @@ const filterCreator: FilterCreator = ({ config }) => ({
     if (slaConditionTypes === undefined) {
       return
     }
-    slaConditionTypes.forEach(slaType => {
-      slaType.fields.conditionId.refType = new TypeReference(BuiltinTypes.UNKNOWN.elemID, BuiltinTypes.UNKNOWN)
-      // prevent deployment of name as it does nothing
-      if (slaType.fields.name !== undefined) {
-        slaType.fields.name.annotations[CORE_ANNOTATIONS.CREATABLE] = false
-        slaType.fields.name.annotations[CORE_ANNOTATIONS.UPDATABLE] = false
-      }
-    })
+    slaConditionTypes
+      .filter(slaType => slaType.fields.conditionId !== undefined)
+      .forEach(slaType => {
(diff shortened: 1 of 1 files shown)


idoamit8 (commented):
Great work ! 💯 
Thanks for the quick fix! 
Please add a test when `conditionId` is undefined.
What about adding another check if `!fetchQuery.isTypeMatch(SLA_TYPE)` so we can exclude SLA and not get to this code.

IdoZakk:
@idoamit8 I did not want to delay the development on UT, as we will issue a hotfix as you suggested.
Regarding the isTypeMatch I would add it if was general for the whole filter

idoamit8:
@IdoZakk and @kami-delaroz, I know this has already been merged, and it works well as a hotfix. However, there are some points we should address tomorrow:

1. What happens when deploying an SLA from envA to envB, where the SLA type in envB doesn't include a conditionId? Could this cause any issues?
2. I’m not a fan of filtering by conditionId and then checking inside if name is undefined. It feels like the name is **depends** on conditionId, which isn't true in the case ( I do agree that they are come together)

packages/jira-adapter/src/filters/jsm_types_fetch_filter.ts
```diff
@@ -89,14 +89,16 @@ const filterCreator: FilterCreator = ({ config }) => ({
...
-    slaConditionTypes.forEach(slaType => {
-      slaType.fields.conditionId.refType = new TypeReference(BuiltinTypes.UNKNOWN.elemID, BuiltinTypes.UNKNOWN)
-      // prevent deployment of name as it does nothing
-      if (slaType.fields.name !== undefined) {
-        slaType.fields.name.annotations[CORE_ANNOTATIONS.CREATABLE] = false
-        slaType.fields.name.annotations[CORE_ANNOTATIONS.UPDATABLE] = false
-      }
-    })
+    slaConditionTypes
+      .filter(slaType => slaType.fields.conditionId !== undefined)
```
ori-moisis, line 93: I know this is already merged, but IIUC this means that if there is no `conditionId` field, then the code that sets `creatable / updatable` on the `name` field won't run as well.

is this on purpose?
just looks a bit strange that these two fields are tied together this way...

IdoZakk: @idoamit8 commented on it. I fixed it in [here](https://github.com/salto-io/salto/pull/6844)

PR #6848, opened 2024-11-27
SALTO-6776: Salesforce Cancel Validation/Deployment

tamtamirr:
SALTO-6776: Salesforce Cancel Validation/Deployment

---

- Added the ability to cancel a serviceAsyncTask in the adapter's core API.
- Implemented support for canceling task of type `validation` or `deployment` in Salesforce Adapter.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


packages/salesforce-adapter/src/client/client.ts
```diff
@@ -1002,6 +1003,34 @@ export default class SalesforceClient implements ISalesforceClient {
...
+              status: 'Canceling',
+            },
+          }),
+        })
+        if (_.get(cancelDeployResult, ['deployResult', 'status']) === 'Canceling') {
+          await new Promise(resolve => setTimeout(resolve, this.conn.metadata.pollInterval))
+          await checkStatus()
+        }
+      } catch (e) {
+        log.warn('Failed to cancel %s with id %s: %s', taskType, taskId, inspectValue(e))
```
yelly, line 1028: I think we should return this error instead of just logging it.

packages/salesforce-adapter/src/client/client.ts
```diff
@@ -1002,6 +1003,34 @@ export default class SalesforceClient implements ISalesforceClient {
...
+        const cancelDeployResult = await this.conn.request({
+          method: 'PATCH',
+          url: `/services/data/v${API_VERSION}/metadata/deployRequest/${taskId}`,
+          body: inspectValue({
+            deployResult: {
+              status: 'Canceling',
+            },
+          }),
+        })
+        if (_.get(cancelDeployResult, ['deployResult', 'status']) === 'Canceling') {
```
yelly, line 1023: I don't really like using `_.get` to get around doing a type guard.

packages/salesforce-adapter/src/client/client.ts
```diff
@@ -1002,6 +1003,34 @@ export default class SalesforceClient implements ISalesforceClient {
...
+          await new Promise(resolve => setTimeout(resolve, this.conn.metadata.pollInterval))
```
yelly, line 1024: Do we really need this weird syntax?
Very possible the answer is "yes", it's just the first time I'm seeing it.

packages/salesforce-adapter/src/metadata_deploy.ts
```diff
@@ -442,15 +442,20 @@ const quickDeployOrDeploy = async (
...
+  let wasTaskIdReported = false
   const createProgressReporterCallback =
     (suffix?: string) =>
     (deployResult: SFDeployResult): void => {
       if (!progressReporter) {
         return
       }
       progressReporter.reportMetadataProgress({ result: deployResult, suffix })
+      if (!wasTaskIdReported && progressReporter.reportServiceAsyncTaskId) {
+        progressReporter.reportServiceAsyncTaskId(deployResult.id)
```
yelly, line 456: Will we always have an ID on the first deploy result we get?

packages/salesforce-adapter/src/client/client.ts
```diff
@@ -1002,6 +1003,34 @@ export default class SalesforceClient implements ISalesforceClient {
...
+          await checkStatus()
+        }
```
yelly, line 1026: I think we should check if the terminal status is `Canceled` (or whatever it should be) and return an error if it's anything else.

packages/salesforce-adapter/src/metadata_deploy.ts
yelly, line 456: Let's cover this with a test.

packages/salesforce-adapter/src/adapter.ts
```diff
@@ -794,6 +795,12 @@ export default class SalesforceAdapter implements SalesforceAdapterOperations {
     return this.deployOrValidate(deployOptions, true)
   }
 
+  async cancelServiceAsyncTask(input: CancelServiceAsyncTaskInput): Promise<void> {
+    if (input.taskType === 'deployment' || input.taskType === 'validation') {
+      await this.client.cancelMetadataValidateOrDeployTask(input)
+    } else throw new Error(`Salesforce cancelServiceAsyncTask is not supported for task type: ${input.taskType}`)
```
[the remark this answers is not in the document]
tamtamirr: This is the prettier format

packages/salesforce-adapter/src/client/client.ts
tamtamirr: This is the suggestions I've seen online as-well sadly.

packages/salesforce-adapter/src/metadata_deploy.ts
tamtamirr: Yes :) It's a required field on the type `DeployResult`

packages/adapter-api/src/adapter.ts
```diff
@@ -143,6 +154,7 @@ export type AdapterOperations = {
   fetch: (opts: FetchOptions) => Promise<FetchResult>
   deploy: (opts: DeployOptions) => Promise<DeployResult>
   validate?: (opts: DeployOptions) => Promise<DeployResult>
+  cancelServiceAsyncTask?: (opts: CancelServiceAsyncTaskInput) => Promise<void>
```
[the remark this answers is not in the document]
tamtamirr: Discussed F2F

packages/salesforce-adapter/src/client/client.ts
tamtamirr: In a race condition where we attempt to cancel a validation that is already `Done` Error is not expected IMO.

packages/salesforce-adapter/src/client/client.ts
tamtamirr: discussed F2F. Will throw the Error here instead of returning it

yelly (approved):
@ori-moisis please sign off on the adapter API changes.

packages/adapter-api/src/adapter.ts
```diff
@@ -63,16 +63,27 @@ export type ProgressReporter = {
   reportProgress: (progress: Progress) => void
 }
 
+export type AsyncTaskProgressReporter = ProgressReporter & {
+  reportServiceAsyncTaskId?: (id: string) => void
+}
```
ori-moisis, line 68: not sure it makes sense for this to be a separate type, especially if you have the function optional anyway.
we wouldn't want to create a new progress reporter type every time we add a new event type

packages/adapter-api/src/adapter.ts
```diff
@@ -63,16 +63,27 @@ export type ProgressReporter = {
...
-  progressReporter: ProgressReporter
+  progressReporter: AsyncTaskProgressReporter
   changeGroup: ChangeGroup
 }
 
+export type ServiceAsyncTaskType = 'deployment' | 'validation'
+
+export type CancelServiceAsyncTaskInput = {
+  readonly taskId: string
+  readonly taskType: ServiceAsyncTaskType
```
ori-moisis, line 84: this doesn't feel like something that should be here, I think we don't need this (I see it is only used for logs, but we can know what is happening from context, we don't need this)

packages/cli/src/commands/deploy.ts
```diff
@@ -187,6 +187,9 @@ const deployPlan = async (
         (item: PlanItem, step: ItemStatus, details?: string) => updateAction(item, step, details),
         accounts,
         checkOnly,
+        asyncTaskId => {
+          outputLine(`Reported async task id: ${asyncTaskId}`, output)
```
ori-moisis, line 191: I am not sure it makes sense to write this to the output like this...
I don't know if a user really has anything to do with this kind of log.

packages/core/src/api.ts
```diff
@@ -162,6 +164,7 @@ export const deploy = async (
   reportProgress: (item: PlanItem, status: ItemStatus, details?: string) => void,
   accounts = workspace.accounts(),
   checkOnly = false,
+  reportServiceAsyncTaskId: AsyncTaskProgressReporter['reportServiceAsyncTaskId'] = () => undefined,
```
ori-moisis, line 167: it doesn't make sense to me to have a separate function for every type of event

packages/adapter-api/src/adapter.ts
```diff
@@ -74,6 +74,14 @@ export type DeployOptions = {
   changeGroup: ChangeGroup
 }
 
+export type CancelServiceAsyncTaskInput = {
+  readonly taskId: string
```
yelly, line 78: I don't think `readonly` does anything here.

packages/adapter-components/test/adapter-wrapper/adapter.test.ts
```diff
@@ -17,8 +17,7 @@ import {
   getChangeData,
   InstanceElement,
   isInstanceElement,
-  ObjectType,
-  ProgressReporter,
+  ObjectType, ProgressReporter,
```
yelly, line 20: No way the linter allows this...

packages/salesforce-adapter/src/client/client.ts
```diff
@@ -586,6 +593,15 @@ export type CustomListFuncDef = {
...
 }
 
+type CanceledDeployResult = {
+  deployResult: {
+    status: string
+  }
+}
+
+const isCancelDeployResult = (result: unknown): result is CanceledDeployResult =>
+  _.isString(_.get(result, ['deployResult', 'status']))
```
yelly, line 603: I prefer to avoid `_.get` in cases like this - you can just access the fields using normal dot access.

packages/salesforce-adapter/src/client/client.ts
```diff
@@ -1002,6 +1018,55 @@ export default class SalesforceClient implements ISalesforceClient {
     )
   }
 
+  @mapToUserFriendlyErrorMessages
+  @logDecorator()
+  @requiresLogin()
+  public async cancelMetadataValidateOrDeployTask({
+    taskId,
+  }: CancelServiceAsyncTaskInput): Promise<CancelServiceAsyncTaskResult> {
+    const checkStatus = async (): Promise<CancelServiceAsyncTaskResult> => {
```
yelly, line 1027: I'm confused - this doesn't check the status, it cancels the operation, no?

packages/salesforce-adapter/src/client/client.ts
tamtamirr: I can't, I'm dealing with unknown value here though 🤔

packages/salesforce-adapter/src/client/client.ts
tamtamirr: It does so recursively, I've renamed 🙏

packages/salesforce-adapter/src/client/client.ts
```diff
@@ -1002,6 +1018,55 @@ export default class SalesforceClient implements ISalesforceClient {
...
+                message: `Failed to cancel async deployment with id ${taskId}`,
+                detailedMessage: 'Salesforce cancelDeployResult value does not contain status',
+                severity: 'Error',
+              },
+            ],
+          }
+        }
+        if (cancelDeployResult.deployResult.status === 'Canceling') {
+          await new Promise(resolve => setTimeout(resolve, this.conn.metadata.pollInterval))
+          await cancelSalesforceDeployment()
```
yelly, line 1051: Let's `GET` for polling and not `PATCH`.

packages/salesforce-adapter/e2e_test/cancel_running_validation.test.ts
```diff
@@ -0,0 +1,147 @@
...
+    const validationResultPromise = firstAdapter.validate({
+      changeGroup: {
+        groupID: 'Metadata Deploy (Validation)',
+        changes: [toChange({ after: apexClassInstance })],
+      },
+      progressReporter: {
+        ...nullProgressReporter,
+        reportMetadataProgress: args => {
+          if (cancelValidationResultPromise === undefined) {
+            cancelValidationResultPromise = secondAdapter.cancelServiceAsyncTask({ taskId: args.result.id })
```
yelly, line 129: Maybe make sure that `args.result.id` is defined in case we add other events that are reported before we have the ID in the future.

tamtamirr: `reportMetadataProgress` is always invoked on a valid `deployResult` though. `id` is always defined.

packages/core/src/api.ts
```diff
@@ -729,3 +731,30 @@ export const fixElements = async (
...
+    workspace.accountConfig.bind(workspace),
+    await workspace.elements(),
+    getAccountToServiceNameMap(workspace, accounts),
+  )
+  const adapter = adaptersMap[account]
+  if (adapter === undefined) {
+    throw new Error(`No adapter found for account ${account}`)
+  }
+  if (adapter.cancelServiceAsyncTask === undefined) {
+    throw new Error(`cancelServiceAsyncTask is not supported for account ${account}`)
```
tamtamirr, line 757: Not sure what's the preferred approach here. Should we return a `SaltoError` instead of throwing Error here?

packages/cli/src/commands/cancel_task.ts
```diff
@@ -0,0 +1,60 @@
...
+  taskId: string
+  account: string
+}
+export const action: WorkspaceCommandAction<CancelTaskInput> = async ({ input, workspace, output }) => {
+  const { taskId, account } = input
+  try {
+    const result = await cancelServiceAsyncTask({ workspace, account, input: { taskId } })
+    const errors = result.errors.filter(e => e.severity === 'Error')
+    if (errors.length > 0) {
+      errorOutputLine(`Failed to cancel async task ${taskId} with Errors: ${inspectValue(errors)}`, output)
```
ori-moisis, line 24: I don't think this ends up being a reasonable format for errors
we have functions that format errors in the CLI, maybe another one needs to be added...

packages/cli/src/commands/cancel_task.ts
```diff
@@ -0,0 +1,60 @@
...
+export type CancelTaskInput = {
+  taskId: string
+  account: string
+}
+export const action: WorkspaceCommandAction<CancelTaskInput> = async ({ input, workspace, output }) => {
+  const { taskId, account } = input
+  try {
+    const result = await cancelServiceAsyncTask({ workspace, account, input: { taskId } })
+    const errors = result.errors.filter(e => e.severity === 'Error')
+    if (errors.length > 0) {
```
ori-moisis, line 23: this means that warnings will not be in the output, not even in the log
feels like bad logic

packages/salesforce-adapter/e2e_test/cancel_running_validation.test.ts
ori-moisis, line 1: did we test canceling a task that is already done / a task that does not exist?

PR #6851, opened 2024-11-27
SALTO-6956: Remove deprecated loadWorkspace api

dorsev:
None

---

_Release Notes_: 

Remove deprecated API of loadWorkspace - that was added here - https://github.com/salto-io/salto/pull/2895 ( over 2 years ago)

---

_User Notifications_: 
None

### DIFF
--- a/packages/core/src/local-workspace/workspace.ts
+++ b/packages/core/src/local-workspace/workspace.ts
@@ -175,7 +175,6 @@ export const loadLocalElementsSources = async ({
   persistent = true,
 }: {
   baseDir: string
-  localStorage?: string // TODO: remove unused argument (kept backwards compatibility)
   envs: ReadonlyArray<string>
   remoteMapCreator: remoteMap.RemoteMapCreator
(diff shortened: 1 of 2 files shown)


packages/core/src/local-workspace/workspace.ts
```diff
@@ -284,14 +284,14 @@ type LoadLocalWorkspaceArgs = {
   ignoreFileChanges?: boolean
 }
 
-const loadLocalWorkspaceImpl = async ({
+export async function loadLocalWorkspace({
```
dorsev, line 287: Also, while I'm at it, should I also remove `localStorage?: string // TODO: remove unused argument (kept backward compatibility)`? 

cc: @ori-moisis

ori-moisis (approved):

packages/core/src/local-workspace/workspace.ts
dorsev: removed 👍

PR #6861, opened 2024-11-28
SALTO-6990: Add parseValue function in parser

idoamit8:
In this PR I added parseValue function in parser.

---

I added this function in order to get the env-var values from the DB and parse them by their type. 

---
_Release Notes_: 

---
_User Notifications_: 


dantal4 (commented):
Great work!

packages/parser/src/parser/internal/native/consumers/values.ts
```diff
@@ -459,3 +460,18 @@ export const consumeValue = (
       throw new Error('Unknown value')
   }
 }
+
+export const parseValue = (content: string): Value => {
```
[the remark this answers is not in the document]
ori-moisis: on @dantal4 's comments:
1. +1
2. +1, I think it would be better to return the errors for consistency, and because there can be more than one error (also, this needs to be in the unit tests)
3. I think this would be a confusing option because it won't actually prevent promises. in fact, IIUC, the code you suggested would do the opposite of what I would expect. if I call the function with "allowPromiseValues: false" I would get promises in the returned value.
is it really that important for this function to be sync? if so, maybe we have to return an error if `valuePromiseWatchers` is not empty?

also:
4. this being an external API - used named arguments so it can be extended more easily (so, `parseValue({ content, /* other args... */ }: ParseValueArgs)`)
5. I think we will want to be able to pass in the functions, similar to the API of the "regular" parse (though, if point 1 is done, this doesn't have to happen in this PR if you think it is not needed)

packages/parser/test/parser/parse.test.ts
```diff
@@ -1744,4 +1746,66 @@ multiline
       ])
     })
   })
+  /*
+   * Copyright 2024 Salto Labs Ltd.
+   * Licensed under the Salto Terms of Use (the "License");
+   * You may not use this file except in compliance with the License.  You may obtain a copy of the License at https://www.salto.io/terms-of-use
+   *
+   * CERTAIN THIRD PARTY SOFTWARE MAY BE CONTAINED IN PORTIONS OF THE SOFTWARE. See NOTICE FILE AT https://github.com/salto-io/salto/blob/main/NOTICES
+   */
```
dantal4, line 1755: ?

packages/parser/test/parser/parse.test.ts
```diff
@@ -1744,4 +1746,66 @@ multiline
...
+
+  describe('parseValue', () => {
+    const mockConsumeValue = jest.spyOn(values, 'consumeValue')
```
dantal4, line 1758: I think that we should test `consumeValue`'s real usage, and not mocking it.
for parse error - you can use `"value` (without closing quote sign)
for a promise - pass a `file` function that returns promise in `parse`, and give the value `file('filename')`

packages/parser/src/parser/parse.ts
```diff
@@ -73,3 +77,31 @@ export function* tokenizeContent(content: string): IterableIterator<Token> {
     log.error('Error occured while getting token: %o', e)
   }
 }
+
+export const parseValue = (
+  content: string,
+  functions?: Functions,
```
dantal4, line 83: as @ori-moisis suggested, let's change this to get named parameters (in an object).
nit for `functions` - use `functions = {}` instead of `functions ?? {}` in line 91.

packages/parser/src/parser/parse.ts
```diff
@@ -73,3 +77,31 @@ export function* tokenizeContent(content: string): IterableIterator<Token> {
     log.error('Error occured while getting token: %o', e)
   }
 }
+
+export const parseValue = (
```
dantal4, line 81: can you add a documentation to this function (similar to the documentation of `parse`) - and write that it doesn't support values that are asynchronously parsed?

packages/parser/test/parser/parse.test.ts
```diff
@@ -1744,4 +1746,66 @@ multiline
...
+
+    describe('when value is string', () => {
+      it('should return string value', () => {
+        const value = '"value"'
+        expect(parseValue(value).value).toEqual('value')
```
dantal4, line 1763: in each call to `parseValue` we should check that `errors` is empty.

dantal4 (commented):
Great work!

packages/parser/test/parser/parse.test.ts
```diff
@@ -1744,4 +1746,66 @@ multiline
...
+      })
+    })
+    describe('when consumeValue adds promises to context', () => {
+      it('should return the errors in the result', () => {
+        const mockResult = { value: 'mockValue', range: {} as Omit<SourceRange, 'filename'> }
+        mockConsumeValue.mockImplementation(context => {
+          context.valuePromiseWatchers.push({ parent: 'some value', key: 'some key' })
+          return mockResult
+        })
+        expect(() => parseValue('mock input')).toThrow('Unexpected promise')
```
ori-moisis, line 1807: the test says it should return the errors in the result (which would be the correct behavior in our APIs), but this actually asserts that the code throws an error instead of returning it

packages/parser/test/parser/parse.test.ts
```diff
@@ -1744,4 +1746,66 @@ multiline
...
   })
+  /*
+   * Copyright 2024 Salto Labs Ltd.
+   * Licensed under the Salto Terms of Use (the "License");
+   * You may not use this file except in compliance with the License.  You may obtain a copy of the License at https://www.salto.io/terms-of-use
+   *
+   * CERTAIN THIRD PARTY SOFTWARE MAY BE CONTAINED IN PORTIONS OF THE SOFTWARE. See NOTICE FILE AT https://github.com/salto-io/salto/blob/main/NOTICES
+   */
+
+  describe('parseValue', () => {
```
ori-moisis, line 1757: missing tests on complex values (arrays, objects)

packages/parser/test/parser/parse.test.ts
```diff
@@ -1744,4 +1746,72 @@ multiline
...
+      expect(result.value).toEqual([1, 2, 3])
+      expect(result.errors).toEqual([])
+    })
+
+    describe('when consumeValue adds errors to context', () => {
+      it('should return the errors in the result', () => {
+        const result = parseValue({ content: '"value' })
+        expect(result.value).toBeUndefined()
+        expect(result.errors).toEqual(
+          expect.arrayContaining([
```
dantal4, line 1798: nit - you know that there's only one error in the array, so the `arrayContaining` is redundant.

packages/parser/src/parser/parse.ts
```diff
@@ -73,3 +78,58 @@ export function* tokenizeContent(content: string): IterableIterator<Token> {
...
+    sourceMap: new SourceMap(),
+    valuePromiseWatchers: [],
+  }
+  let result: ConsumerReturnType<Value> | undefined
+  try {
+    result = consumeValue(context)
+  } catch (e) {
+    processParseError(context, e)
+  }
+  if (context.valuePromiseWatchers.length > 0) {
```
dantal4, line 123: since we saw that for `myFunc("some.png")`, `valuePromiseWatchers`is empty, and `value` is a `Promise`:
```suggestion
  if (context.valuePromiseWatchers.length > 0 || isPromise(result?.value)) {
```
and I would also add a test for that.

dantal4 (approved):
💯 
thanks for that!

PR #6863, opened 2024-11-28
SALTO-6994: Avoid redundant copy of fetch changes

ori-moisis:
---

in the past the changes here were an iterator
they are an array now, so there is no point in creating this copy

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/core/src/core/fetch.ts
+++ b/packages/core/src/core/fetch.ts
@@ -204,7 +204,7 @@ const toChangesWithPath =
 
     const originalElements = accountElementByFullName(changeID)
     if (originalElements.length === 0) {
-      log.trace(`no original elements found for change element id ${changeID.getFullName()}`)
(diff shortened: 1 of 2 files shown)


packages/core/src/core/fetch.ts
```diff
@@ -204,7 +204,7 @@ const toChangesWithPath =
 
     const originalElements = accountElementByFullName(changeID)
     if (originalElements.length === 0) {
-      log.trace(`no original elements found for change element id ${changeID.getFullName()}`)
+      log.debug('no original elements found for change element id %s', changeID.getFullName())
```
ori-moisis, line 207: this is unrelated, but since #6784 this is no longer a super spammy log, so we can bring it back to the a `debug` log and not just a `trace` log

yelly (approved):

PR #6866, opened 2024-11-28
SALTO-6995: Add element-test-utils package

ori-moisis:
The first thing this package implements is an override for jest expect that compares element IDs using a custom matcher

---

Required for #6862

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/adapter-api/test/setup.ts
+++ b/packages/adapter-api/test/setup.ts
@@ -1,0 +1,21 @@
+/*
+ * Copyright 2024 Salto Labs Ltd.
+ * Licensed under the Salto Terms of Use (the "License");
+ * You may not use this file except in compliance with the License.  You may obtain a copy of the License at https://www.salto.io/terms-of-use
+ *
+ * CERTAIN THIRD PARTY SOFTWARE MAY BE CONTAINED IN PORTIONS OF THE SOFTWARE. See NOTICE FILE AT https://github.com/salto-io/salto/blob/main/NOTICES
+ */
+import type { Tester } from '@jest/expect-utils'
+import { expect } from '@jest/globals'
+import { ElemID } from '../src/element_id'
+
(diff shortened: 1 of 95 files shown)


packages/adapter-api/test/setup.ts
```diff
@@ -0,0 +1,21 @@
...
+import { ElemID } from '../src/element_id'
+
+// We cannot import this from @salto-io/element-test-utils because it would cause a cyclic dependency
+// so we have to define this here separately
+export const isEqualElemID: Tester = (a, b) => {
+  if (a instanceof ElemID && b instanceof ElemID) {
+    return a.isEqual(b)
+  }
+  return undefined
+}
```
tamtamirr, line 19: - Do we a have future ticket to separate the types from `adapter-api` to another package in order to break this cycle? 
- Is there any way to enforce having all the matchers from `elements-test-utils` here?

packages/element-test-utils/README.md
```diff
@@ -0,0 +1,5 @@
+# Salto - test utilities
+
+---
+
+This package contains a few useful shortcuts and utilities for the [jest framework](https://jestjs.io/)
```
tamtamirr, line 5: I think we should write a simple conventions guide about how to add matchers to this package.

tamtamirr (commented):
Nice 👑 Added some comments

packages/adapter-api/test/setup.ts
ori-moisis: I can't think of any way that would allow this given that this code requires the class (for `instanceof`, for calling `isEqual`, etc...).
AFAICT there is no way to break the cycle, because the whole point of this new package is that it can use things from adapter-api (unlike the general `test-utils`).

PR #6867, opened 2024-11-28
SALTO-6973: Parse scripts in Intune application

shoshana-houri:
According to [MS Graph openAPI](https://github.com/microsoftgraph/msgraph-metadata/blob/master/openapi/beta/openapi.yaml), the following Intune applications may include a binary64 encoded `scriptContent` field:
```
macOSPkgApp: {
  preInstallScript: {
    ...
    scriptContent,
  },
  postInstallScript: {
    ...
    scriptContent,
  },
}

win32LobApp: {
  detectionRules: Array of {
    ...
    scriptContent,
  },
  requirementRules: Array of {
    ...
    scriptContent,
  },
  rules: Array of {
    ...
    scriptContent,
  },
},
```
This PR extracts those scriptContent fields to binary64 static files.

---
_Release Notes_: 
_Microsoft Security adapter:_
* Parse scripts in Intune application

---
_User Notifications_: 
_Microsoft Security adapter:_
* Parse scripts in Intune application


packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/application.ts
```diff
@@ -42,3 +49,54 @@ export const APPLICATION_NAME_PARTS: definitions.fetch.FieldIDPart[] = [
...
+const odataTypeToScriptsExtractionParams: Record<string, ScriptsExtractionParams> = {
+  macOSPkgApp: {
+    scriptsRootFieldNames: ['preInstallScript', 'postInstallScript'],
+    validateFunc: validatePlainObject,
+    toFileName: ({ scriptsRootFieldName }) => `${scriptsRootFieldName}.sh`,
+  },
+  win32LobApp: {
+    scriptsRootFieldNames: ['detectionRules', 'requirementRules', 'rules'],
+    validateFunc: validateArray,
+    toFileName: ({ scriptsRootFieldName, scriptField, index }) =>
```
shir-reifenberg, line 68: are we sure scripts have `displayName` ? using the index here means that if the order changes, the static file moves path

packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/script_content.ts
```diff
@@ -0,0 +1,88 @@
...
+      log.debug(`No script content found for ${typeName}.${scriptsRootFieldName} (ID: ${value.id})`)
+      return
+    }
+
+    if (!_.isString(scriptContent)) {
+      const message = `Invalid script content for ${typeName}.${scriptsRootFieldName}. Expected string, got ${safeJsonStringify(scriptContent)}`
+      log.error(message)
+      throw new Error(message)
+    }
+
```
shir-reifenberg, line 74: should we aggregate into _.isString() with `log.trace` ? (as this is a valid scenario)

packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/script_content.ts
```diff
@@ -0,0 +1,88 @@
...
+  scriptsRootFieldName: string
+  scriptValuePath?: string[]
+  elemIDFieldName?: string
+  staticFileSubDirectory?: string
+  throwOnEmptyField?: boolean
+  validateFunc: ValidateScriptsRootFieldFunc
+  toFileName: ToFileNameFunc
+}
+
+export const extractStaticFileFromBinaryScript = ({
```
shir-reifenberg, line 35: are we sure there is enough in common between platform scripts to app scripts to consolidate the logic creating them ?
asking cause -
1. one is created in the `adjust` and the other in `mergeAndTransform` and I find it a little surprising that the `setScriptValueAsStaticFile` in platform scripts create the scropts then copying it to the new value
2. the `staticFileSubDirectory` and `throwOnEmptyField` are not needed by both?

Do you think we share enough code here? WDYT?

packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/application.ts
shir-reifenberg: also just verifying we think the file names (in both cases) are unique cause ending up with two static files in the same file is never good 😅

packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/application.ts
```diff
@@ -42,3 +49,54 @@ export const APPLICATION_NAME_PARTS: definitions.fetch.FieldIDPart[] = [
...
+  scriptsRootFieldNames: string[]
+  validateFunc: ValidateScriptsRootFieldFunc
+  toFileName: ToFileNameFunc
+}
+
+const odataTypeToScriptsExtractionParams: Record<string, ScriptsExtractionParams> = {
+  macOSPkgApp: {
+    scriptsRootFieldNames: ['preInstallScript', 'postInstallScript'],
+    validateFunc: validatePlainObject,
+    toFileName: ({ scriptsRootFieldName }) => `${scriptsRootFieldName}.sh`,
```
shir-reifenberg, line 63: are we sure it's always a bash script? (same question for the .ps1)

shir-reifenberg (commented):
Great job ! (and great idea looking for those in the swagger 😄)
Added some questions and suggestions, LMK if you'd like to discuss IRL

packages/microsoft-security-adapter/src/definitions/fetch/intune/fetch.ts
```diff
@@ -62,7 +62,10 @@ const graphBetaCustomizations: FetchCustomizations = {
           ...DEFAULT_TRANSFORMATION,
           // TODO SALTO-6483: We need to store the largeIcon as a static file, for now we omit it
           omit: ['largeIcon', ASSIGNMENTS_ODATA_CONTEXT],
-          adjust: odataType.transformOdataTypeField('fetch'),
+          adjust: concatAdjustFunctions(
```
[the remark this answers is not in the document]
shoshana-houri: 🥲 
Until we'll implement https://salto-io.atlassian.net/browse/SALTO-6935

packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/application.ts
```diff
@@ -42,3 +49,54 @@ export const APPLICATION_NAME_PARTS: definitions.fetch.FieldIDPart[] = [
...
+ * Adjust function to set script values as static files for different intune application types.
+ */
+export const setApplicationScriptValueAsStaticFile: AdjustFunctionSingle = async ({ value }) => {
+  validatePlainObject(value, APPLICATION_TYPE_NAME)
+
+  const appOdataType = value[odataType.getAdjustedOdataTypeFieldName(APPLICATION_TYPE_NAME)]
+  if (!_.isString(appOdataType)) {
+    const message = `Missing odataType for ${APPLICATION_TYPE_NAME} (ID: ${value.id}).`
+    log.error(message)
+    throw new Error(message)
```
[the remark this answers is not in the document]
shoshana-houri: I agree, we can probably be more forgiving here and just log a warning as we do in `transformOdataTypeField`

packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/script_content.ts
shoshana-houri: I'm not sure I understand :)

packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/application.ts
shoshana-houri: Yes :)

From [MS Graph doc](https://learn.microsoft.com/en-us/graph/api/resources/intune-apps-macosappscript?view=graph-rest-beta) for both pre and post install scrips:
> The base64 encoded shell script (.sh) that assists managing macOS apps.

packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/script_content.ts
shoshana-houri: I agree that there are some minor differences in behavior between application scripts and platform scripts, but it seems there is more in common than not, which is why I think we should share the code.

1. The reason we use `mergeAndTransform` in the platform script instead of just `adjust` is that, for some reason, the Graph API returns `scriptContent` as null by default. We need to explicitly request it inside `recurseInto`. It’s possible to modify `extractStaticFileFromBinaryScript` to accept an optional param that allows creating the script content in another field (instead of manually replacing it in the caller function), but I thought that approach might be too specific and implicit. That’s why I decided to leave it in the caller function. What do you think?

2. Existing optional params: 
   1. The `subDirectory` is only required for applications because the application instance resides in two "unique" parent directories: `IntuneApplication(typeName)/appOdataType`. Most of our types, including `platformScript`, are only located under their `typeName` folder, which seems enough.
   2. Regarding `throwOnEmptyField`: The `scriptContent` field doesn’t always exist for the different fields in `win32LobApps`, but it should exist for `macOSPkgApps`. I’ll add it there.

shir-reifenberg (approved):
Didn't review the entire thing but approving to unblock 😄

packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/application.ts
shoshana-houri: I managed to create the app in Intune. However, when adding script files, it doesn't allow you to add files that don't end with .ps1 (see screenshot), so seems like adding .ps1 is safe here :)
image

Anyway, I've adjusted the logic for determining the file name so that it doesn't add an additional .ps1 if it's already present in the displayName. I noticed that when working directly with the Intune Admin Center and uploading a file, the displayName is automatically set to the file name, which may already include .ps1.

packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/script_content.ts
shoshana-houri: Talked f2f - decided not to throw an error on unexpected format that is returned from the service, and change the log level to trace.

shoshana-houri:
I managed to create `win32apps` in our MS Intune tenant, and here’s what I learned:

1. The rules field is redundant—it’s essentially a concatenation of the detection rules and the requirement rules.
2. You can only create a single detection rule that uses a script.
3. The `displayName` for requirement rules is both required and must be unique.
4. When adding script files in the Intune Admin Center, it only allows files with a `.ps1` extension (see screenshot). Based on this, it seems safe to append the `.ps1` extension where necessary. That said, I’ve adjusted the logic for determining the file name to avoid adding an extra .ps1 if it’s already present in the displayName (when uploading a file directly in the Intune Admin Center, the displayName is automatically set to the file name)

image


I've updated the PR to reflect all of that. 
In addition, I've tested deployment locally and it works :)

packages/microsoft-security-adapter/src/definitions/fetch/intune/utils/application.ts
shoshana-houri: Added more accurate details here https://github.com/salto-io/salto/pull/6867#issuecomment-2509784244

PR #6853, opened 2024-11-27
SALTO-6955: Move local-workspace files to a new OSS package

dorsev:
### TODO
- [x] Self-review this code.
- [x] Validate next prs work properly rebased on this PR

### Summary
This PR introduces the `local-workspace` package, which handles operations related to local workspace management, such as state management, file storage, configuration, and telemetry.

It will be utilized in a follow-up PR [#6868](https://github.com/salto-io/salto/pull/6868) to implement adapter-format methods for the dummy-adapter. This was previously infeasible due to a cyclic dependency.

---

### Additional Context for Reviewers
I’ve ensured that no changes break the `core` API package. Please double-check that nothing was missed.

---

### Release Notes
None.

---

### User Notifications
None.

PR #6868, opened 2024-12-01
SALTO-6955: Support all adapter format methods in dummy adapter

dorsev:
This is rebased on [https://github.com/salto-io/salto/pull/6853 &](https://github.com/salto-io/salto/pull/6944)

Keeping this in draft until the above is merged.

So please only review the last commit

---

None

---

_Release Notes_: 
None

---

_User Notifications_: 
None

### DIFF
--- a/packages/cli/src/commands/adapter_format.ts
+++ b/packages/cli/src/commands/adapter_format.ts
@@ -21,10 +21,12 @@ import { getUserBooleanInput } from '../callbacks'
 const log = logger(module)
 const { awu } = collections.asynciterable
 
+const APPLY_PATCH_ADAPTERS = ['salesforce', 'netsuite', 'dummy'] as const
+type ApplyPatchAdapters = (typeof APPLY_PATCH_ADAPTERS)[number]
 type ApplyPatchArgs = {
   fromDir: string
   toDir: string
-  accountName: 'salesforce' | 'netsuite'
+  accountName: ApplyPatchAdapters
   targetEnvs?: string[]
   updateStateInEnvs?: string[]
 } & UpdateModeArg
@@ -124,7 +126,7 @@ const applyPatchCmd = createWorkspaceCommand({
           'The account name for elements, this determines the expected format of the elements in the directories',
         type: 'string',
         required: true,
-        choices: ['salesforce', 'netsuite'],
+        choices: [...APPLY_PATCH_ADAPTERS],
       },
       {
         name: 'targetEnvs',
@@ -159,9 +161,12 @@ const applyPatchCmd = createWorkspaceCommand({
   action: applyPatchAction,
 })
 
+const SYNC_WORKSPACE_ADAPTERS = ['salesforce', 'dummy'] as const
+type SyncWorkspaceAdapters = (typeof SYNC_WORKSPACE_ADAPTERS)[number]
+
 type SyncWorkspaceToFolderArgs = {
   toDir: string
-  accountName: 'salesforce'
+  accountName: SyncWorkspaceAdapters
   force: boolean
 }
 export const syncWorkspaceToFolderAction: WorkspaceCommandAction<SyncWorkspaceToFolderArgs> = async ({
@@ -218,7 +223,7 @@ const syncToWorkspaceCmd = createWorkspaceCommand({
         type: 'string',
         alias: 'a',
         description: 'The name of the account to synchronize to the project',
-        choices: ['salesforce'],
+        choices: [...SYNC_WORKSPACE_ADAPTERS],
         default: 'salesforce',
       },
       {
--- a/packages/dummy-adapter/jest.config.js
+++ b/packages/dummy-adapter/jest.config.js
@@ -13,10 +13,10 @@ module.exports = deepMerge(require('../../jest.base.config.js'), {
   collectCoverageFrom: ['!<rootDir>/index.ts'],
   coverageThreshold: {
     global: {
-      branches: 83.78,
-      functions: 94.45,
-      lines: 97.28,
-      statements: 97.26,
+      branches: 82.7,
+      functions: 92.3,
+      lines: 97.1,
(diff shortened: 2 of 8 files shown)


packages/cli/src/commands/adapter_format.ts
```diff
@@ -124,7 +124,7 @@ const applyPatchCmd = createWorkspaceCommand({
           'The account name for elements, this determines the expected format of the elements in the directories',
         type: 'string',
         required: true,
-        choices: ['salesforce', 'netsuite'],
+        choices: ['salesforce', 'netsuite', 'dummy'],
```
tamtamirr, line 127: nit: Can we export this type to a const and reuse it both here and in L27?
```
const APPLY_PATCH_ADAPTERS = ['salesforce', 'netsuite', 'dummy']

accountName: typeof APPLY_PATCH_ADAPTERS[number] // type def

choices: APPLY_PATCH_ADAPTERS
```

packages/cli/src/commands/adapter_format.ts
```diff
@@ -161,7 +161,7 @@ const applyPatchCmd = createWorkspaceCommand({
 
 type SyncWorkspaceToFolderArgs = {
   toDir: string
-  accountName: 'salesforce'
+  accountName: 'salesforce' | 'dummy'
```
tamtamirr, line 164: here too, same suggestion for `applyPatch`

packages/dummy-adapter/src/adapter_creator.ts
```diff
@@ -65,9 +62,57 @@ const getCustomReferences: GetCustomReferencesFunc = async elements =>
...
+    persistent: false,
+    getConfigTypes: async () => [],
+    getCustomReferences: async elements => getCustomReferences(elements),
+  })
+  const elements = await workspace.elements()
+  return { elements: await awu(await elements.getAll()).toArray() }
+}
+
+const dumpElementsToFolder: AdapterFormat['dumpElementsToFolder'] = async ({ baseDir, changes }) => {
+  const workspace = await loadLocalWorkspace({
```
tamtamirr, line 77: 1. Why we chose o use `persistent: false` for the two other methods and not on this one?
2. Can we re-use the calls to `loadLocalWorkspace` by exporting it to a method considering they receive the same params?

tamtamirr (commented):
Added a few comments

packages/dummy-adapter/src/adapter_creator.ts
dorsev: We use persistent=false for read-only connections to RocksDB, right?
So, during "load," there's no need for "write."
But during "dump," I need to write, so persistent=true is required.
Does this make sense? Did I miss anything?

dorsev: 2. ok 👍

packages/cli/src/commands/adapter_format.ts
```diff
@@ -124,7 +126,7 @@ const applyPatchCmd = createWorkspaceCommand({
           'The account name for elements, this determines the expected format of the elements in the directories',
         type: 'string',
         required: true,
-        choices: ['salesforce', 'netsuite'],
+        choices: [...APPLY_PATCH_ADAPTERS],
```
dorsev, line 129: Note: I tried to avoid the `...` here, but got 
image

packages/dummy-adapter/test/generator.test.ts
```diff
@@ -73,8 +73,7 @@ describe('generator', () => {
         expect(_.uniq(profiles.map(p => p.elemID.getFullName()))).toHaveLength(testParams.numOfProfiles)
         expect(records).toHaveLength(testParams.numOfRecords + 5) // 5 default instance fragments
       })
-      // eslint-disable-next-line
-      it.skip('should create list and map types', async () => {
+      it('should create list and map types', async () => {
```
dorsev, line 76: Note: I am not sure why was this "skipped".
enabling this works, so I did it.

dorsev:
Thank you for the review @tamtamirr , Im ready for round 2

packages/dummy-adapter/jest.config.js
```diff
@@ -14,7 +14,7 @@ module.exports = deepMerge(require('../../jest.base.config.js'), {
   coverageThreshold: {
     global: {
       branches: 74,
-      functions: 93,
+      functions: 89.1,
```
tamtamirr, line 17: Why did the coverage went significantly down?

tamtamirr (commented):
LGTM! Why did we decrease the coverage though?

packages/dummy-adapter/jest.config.js
```diff
@@ -13,10 +13,10 @@ module.exports = deepMerge(require('../../jest.base.config.js'), {
   collectCoverageFrom: ['!<rootDir>/index.ts'],
   coverageThreshold: {
     global: {
-      branches: 74,
-      functions: 93,
-      lines: 93,
-      statements: 93,
+      branches: 83.7,
+      functions: 92.3,
```
dorsev, line 17: this went down due to these methods not being tested


```

const getCustomReferences: GetCustomReferencesFunc = async elements =>
  elements.find(e => e.elemID.getFullName() === 'dummy.Full.instance.FullInst1')
    ? [
        {
          source: ElemID.fromFullName('dummy.Full.instance.FullInst1.strField'),
          target: ElemID.fromFullName('dummy.Full.instance.FullInst2.strField'),
          type: 'weak',
        },
      ]
    : []
```

And the `getConfigTypes` and `getCustomReferences` in loadWorkspace in adapter_Creator.ts


I don't think they are worth the trouble.

packages/dummy-adapter/jest.config.js
dorsev: increased it back up

tamtamirr (approved):

PR #6869, opened 2024-12-01
SALTO-6947: Okta: Change ElemID computation for default domain and default brand

Nurdok:
See attached ticket for ~~more~~ too much details.

There was an existing solution for `AccessPolicy` and `AccessPolicyRule`, but we realized it regenerates IDs even without `--regenerate-salto-ids`, so this PR fixes those too.

Workspace diffs:
- [Fetch with the new code, **not** regenerating Salto IDs](https://github.com/salto-io/amir-okta/commit/8eeb97c11e1ac392a6e246cc7b6cd9d2610cd569)
  - The diff has only `_alias` additions to domains.
- [Fetch with the new code with `--regenerate-salto-ids` (after the above fetch)](https://github.com/salto-io/amir-okta/commit/4c433a857caebef292adfd62924087a337a62d91)
- Note that the reimplementation of the access policy default name results in "old format" diff when regenerating - default policy rule names will get an added `_s`.

---

None

---
_Release Notes_: 
_Okta Adapter_:
- The default brand and domain for each account will now be named "Default Brand" and "Default Domain" respectively, to allow deployment of these instances between environments. This will happen in new environments, or in existing ones after regenerating Salto element IDs

---
_User Notifications_: 
_Okta Adapter_:
- After regenerating Salto element IDs, the default brand and domain for each account will be renamed "Default Brand" and "Default Domain" respectively


shir-reifenberg (commented):
I'm not sure adding `defaultName` is the correct generalization for this scenario. I think this introduces a few questions / issues like - what happen in cases where we have empty string on purpose (for example when relying on the parent id), or how do we keep uniqueness across entries (the `unnamed_${index}` is there for a reason). 

In general, before adding infra capabilities, and specifically changes to the definitions, I think we should first discuss it in a ticket 😄 

Do you think making the change in Okta can work here ?

Nurdok:
> I'm not sure adding `defaultName` is the correct generalization for this scenario. I think this introduces a few questions / issues like - what happen in cases where we have empty string on purpose (for example when relying on the parent id), or how do we keep uniqueness across entries (the `unnamed_${index}` is there for a reason).
> 
> In general, before adding infra capabilities, and specifically changes to the definitions, I think we should first discuss it in a ticket 😄
> 
> Do you think making the change in Okta can work here ?

Discussed f2f, PTAL.

packages/okta-adapter/src/definitions/fetch/fetch.ts
```diff
@@ -696,6 +679,12 @@ const createCustomizations = ({
       topLevel: {
         isTopLevel: true,
         serviceUrl: { path: '/admin/customizations/footer' },
+        elemID: {
+          parts: [
+            { fieldName: 'name', condition: value => !isDefaultBrand(value) },
+            { fieldName: '', condition: isDefaultBrand, custom: () => () => 'Default Brand' },
```
shir-reifenberg, line 685: nit: maybe add a comment in the spirit of `fieldName is ignored when defining custom function` ?

packages/okta-adapter/test/adapter.test.ts
```diff
@@ -240,8 +240,8 @@ describe('adapter', () => {
           'okta.AccessPolicy.instance.Seamless_access_based_on_network_context@s',
           'okta.AccessPolicy.instance.custom_policy@s',
           'okta.AccessPolicyRule.instance.Classic_Migrated_s__Catch_all_Rule@umuubs',
-          'okta.AccessPolicyRule.instance.Default_Policy__Catch_all_Rule@suubs',
-          'okta.AccessPolicyRule.instance.Default_Policy__shir_test@suus',
+          'okta.AccessPolicyRule.instance.Default_Policy_s__Catch_all_Rule@umuubs',
```
shir-reifenberg, line 243: just making sure the diff is the extra nacl case 🙏

shir-reifenberg (approved):
LGTM 🔥 
Please update the user notification section to have _Okta adapter_ header

PR #6871, opened 2024-12-01
SALT-6999: (Salesforce) Profiles & PS Elements fixer should omit default values from Instance

tamtamirr:
(Salesforce) Profiles & PS Elements fixer should omit default values from Instance

---

Prior to this PR the fixer didn't omit values where we don't create references for.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/salesforce-adapter/src/custom_references/profiles_and_permission_sets.ts
+++ b/packages/salesforce-adapter/src/custom_references/profiles_and_permission_sets.ts
@@ -310,10 +310,10 @@ export const getProfilesAndPsBrokenReferenceFields = async ({
           [sectionName, sectionEntryKey, ...makeArray(sourceField)].join('.'),
           target,
         ],
-        // Unlike in the findWeakReferences, we don't want to filter out the entries that don't have a target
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/src/custom_references/profiles_and_permission_sets.ts
```diff
@@ -246,10 +260,14 @@ const findWeakReferences: WeakReferencesHandler['findWeakReferences'] = async (
 
 const instanceEntriesTargets = (instance: InstanceElement, metadataQuery: MetadataQuery<ElemID>): Dictionary<ElemID> =>
```
yelly, line 261: It's not so clear why in the function with this name we don't want to apply the filter.
I think we should either rename the function or make the decision to skip higher up the call stack.

tamtamirr: Good point 👑 I've moved this private function to where it's being used for better context. Also added a comment.

yelly (approved):

PR #6872, opened 2024-12-01
SALTO-6954: Removing some async calls from CVs

yelly:
Hopefully this should lead to some performance gains.

---

I tested all of the CVs using the CLI in a real world example and they all worked.

---

_Release Notes_: 
None.

---

_User Notifications_: 
None.


tamtamirr (approved):
LGTM. There's a core UT failing, is it a regression from this PR?

yelly:
> LGTM. There's a core UT failing, is it a regression from this PR?

Looks unrelated. I'll rebase and retry.

PR #6873, opened 2024-12-01
SALTO-6954:  Removing some async calls from maps filter and CV

yelly:
It should slightly improve performance.

---

This was originally implemented in https://github.com/salto-io/salto/pull/6827 but was reverted due to a bug in fetching profiles.
The bug was the extra `async` keyword in https://github.com/salto-io/salto/pull/6827/files#diff-616666620c2c1d68e89a7ac6aa5f3734f3a73fd036150009bca8124f97faa0bfR610. This is now removed.

---

_Release Notes_: 
None.

---

_User Notifications_: 
None.


tamtamirr (approved):
This bug is the definition of **bastard**. Let's open a ticket to catch such issues in lint 🙏

PR #6874, opened 2024-12-01
SALTO-1830-cancel-deployment-fix

rotem531:
fixed case of cancelled deployment for deployment summary

---

fixed case of cancelled deployment

---
_Release Notes_: 
none

---
_User Notifications_: 
none

### DIFF
--- a/packages/cli/src/commands/deploy.ts
+++ b/packages/cli/src/commands/deploy.ts
@@ -119,7 +119,6 @@ const deployPlan = async (
   workspace: Workspace,
   cliTelemetry: CliTelemetry,
   output: CliOutput,
-  executingDeploy: boolean,
   checkOnly: boolean,
   accounts?: string[],
 ): Promise<DeployResult> => {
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/deploy.ts
```diff
@@ -259,10 +259,10 @@ export const action: WorkspaceCommandAction<DeployArgs> = async ({
 
   const actionPlan = await preview(workspace, actualAccounts, checkOnly)
   await printPlan(actionPlan, output, workspace, detailedPlan)
-
+  const executingDeploy = !dryRun ? force || (await shouldDeploy(actionPlan, checkOnly)) : false
```
Uri-Bar, line 262: this is quite hard to read, let think of a different way of stating this 
see how it was before I think we can use the same stature without ? :

packages/cli/src/commands/deploy.ts
```diff
@@ -259,10 +259,10 @@ export const action: WorkspaceCommandAction<DeployArgs> = async ({
...
   const result = dryRun
     ? { success: true, errors: [] }
-    : await deployPlan(actionPlan, workspace, cliTelemetry, output, force, checkOnly, actualAccounts)
+    : await deployPlan(actionPlan, workspace, cliTelemetry, output, executingDeploy, checkOnly, actualAccounts)
```
Uri-Bar, line 265: do we really need to call deployPlan when executingDeploy? what will happen?

Uri-Bar (commented):
thanks for that, I have few comments, and please change the release notes to non as this is a minor bugfix and not a new feature

Uri-Bar (approved):

PR #6878, opened 2024-12-02
SALTO-6495 Add useFieldValueAsFallback option to alias components

adfineberg:
Allow fallback to the string of the alias field if it is not a reference.

---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/adapter-components/src/add_alias.ts
+++ b/packages/adapter-components/src/add_alias.ts
@@ -23,6 +23,7 @@ const log = logger(module)
 export type AliasComponent = {
   fieldName: string
   referenceFieldName?: string
+  useFieldValueAsFallback?: boolean
 }
 
 export type ConstantComponent = {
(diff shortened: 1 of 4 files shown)


packages/adapter-components/src/add_alias.ts
```diff
@@ -70,6 +71,9 @@ const getAliasFromField = ({
     return _.isString(fieldValue) ? fieldValue : undefined
   }
   if (!isReferenceExpression(fieldValue)) {
+    if (component.useFieldValueAsFallback === true) {
+      return _.isString(fieldValue) ? fieldValue : undefined
```
edenhassid, line 75: i think if its not a string we should log an error, not sure that we should return undefined, WDYT?

edenhassid (commented):
looks good! 🔥  added one small comment

edenhassid (approved):

PR #6879, opened 2024-12-02
SALTO-7007 Omit inactive (not activated) workspaces

adfineberg:
Omit explicitly inactive workspaces as they have a different indicator field

---

---
_Release Notes_: 

---
_User Notifications_: 
__Zendesk Adapter:__
* Bug fix: inactive workspaces will be removed if omitInactive is set.


edenhassid (approved):
nice!!

PR #6884, opened 2024-12-03
SALTO-7009: Upgrading eslint

yelly:
Upgrading eslit and typescript-eslint and fixing all the errors that are raised.

---

---

_Release Notes_: 
None.

---

_User Notifications_: 
None.


packages/jira-adapter/test/filters/fields/contexts.test.ts
```diff
@@ -38,27 +37,17 @@ jest.mock('@salto-io/adapter-components', () => {
...
-  let paginator: clientUtils.Paginator
 
   beforeEach(() => {
     deployChangeMock.mockClear()
 
     const mockCli = mockClient()
     client = mockCli.client
-    paginator = mockCli.paginator
-
-    filter = fieldsDeploymentFilter(
```
dorsev, line 56: Why wasn't this caught earlier by the linter? did the eslint upgrade find this?

dorsev (approved):
GJ! LGTM

packages/jira-adapter/test/filters/fields/contexts.test.ts
yelly: Yes. The new rule catches values that are only used as types (as was the case here).

PR #6887, opened 2024-12-03
SALTO-6969: Add service URL support for Microsoft Entra elements

shoshana-houri:
Add service url to specific Entra instances.
Some types requires service url customization (for example we need the negation of some value in the url), which is not supported yet.

---
_Release Notes_: 
_Microsoft Security adapter:_
* Add service url to specific Entra instances.

---
_User Notifications_:


shir-reifenberg (approved):
🔥

PR #6893, opened 2024-12-04
SALTO-5728: Support array fields deployments in adapters infra

shir-reifenberg:
Add definitions to allow deployment of nested fields, usually needed for deployment of array fields.

---

How deploy of items is supported - 
- We create "fake" changes for each subresource item, each is either addition / modification / removal change
- We add default dependencies in graph between each subresource to its original change, with the same rules we use in core. additional dependencies can be configures using `dependencies` or `actionDependencies` cusomizations
- Each sub resource change is deployed the same way any other change is deployed
- If a sub resource change fails, we don't continue with the deployment of additional items (unless `failIfChangeHasErrors` is false), and we mark the entire "parent" change as failed.

Not included in this PR (but should be included in a following PR / in the future)
- Support for array items which are primitives
- Allow more complex logic in `getChangeId`, ideally using `TransformDefinition` that returns a string
- Add "copy to parent" functionality, needed to "sync" changes done to the sub resources on the "parent" element
- deployment of all array items from the same change/action in parallel (not sure if this is actually a requirement)

usage example can be seen here https://github.com/salto-io/salto/pull/6933

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/definitions/system/deploy/deploy.ts
+++ b/packages/adapter-components/src/definitions/system/deploy/deploy.ts
@@ -64,6 +64,14 @@ export type ActionDependency<AdditionalAction extends string> = {
   second: ActionName | AdditionalAction
 }
 
+export type RecurseIntoTypeDef = {
+  fieldPath: string[]
+  changeIdFields: string[]
+  condition?: DeployRequestCondition
+  // when missing, all actions are used. currently only standard actions are supported.
+  onActions?: ActionName[]
+}
+
 export type InstanceDeployApiDefinitions<AdditionalAction extends string, ClientOptions extends string> = {
   // a sorted list of requests to make in order to deploy a change of this type
   requestsByAction: DefaultWithCustomizations<
@@ -103,6 +111,8 @@ export type InstanceDeployApiDefinitions<AdditionalAction extends string, Client
   // fail if the change already has (error-level) errors from previous actions
   // default: true
   failIfChangeHasErrors?: boolean
+
+  recurseIntoTypes?: Record<string, RecurseIntoTypeDef>
 }
 
 export type DeployApiDefinitions<AdditionalAction extends string, ClientOptions extends string> = {
--- a/packages/adapter-components/src/definitions/system/deploy/deploy.ts
+++ b/packages/adapter-components/src/definitions/system/deploy/deploy.ts
@@ -64,11 +64,14 @@ export type ActionDependency<AdditionalAction extends string> = {
   second: ActionName | AdditionalAction
 }
 
-export type RecurseIntoTypeDef = {
+export type RecurseIntoPath = {
   fieldPath: string[]
+  // the type name in the specified path, the type should have its own deploy definition
+  typeName: string
   changeIdFields: string[]
+  // when provided, only changes matching the condition will recurse into specified path
   condition?: DeployRequestCondition
-  // when missing, all actions are used. currently only standard actions are supported.
+  // when missing, all actions are used. currently only standard actions are supported
   onActions?: ActionName[]
 }
 
@@ -78,6 +81,12 @@ export type InstanceDeployApiDefinitions<AdditionalAction extends string, Client
     DeployableRequestDefinition<ClientOptions>[],
     ActionName | AdditionalAction
   >
+
+  // when provided, changes will be calculated from the data at the specified path,
+  // and deployed according to the definition of the provided type.
+  // this is mainly used for deploying array fields, where each array item must be deployed separately.
+  // note: the data is not removed from the main change, but can be excluded using the main change's deploy definition.
+  recurseIntoPath?: RecurseIntoPath[]
 
   // how many changes of this type can be deployed in parallel
   // default = unlimited (max concurrency)
@@ -111,8 +120,6 @@ export type InstanceDeployApiDefinitions<AdditionalAction extends string, Client
   // fail if the change already has (error-level) errors from previous actions
   // default: true
   failIfChangeHasErrors?: boolean
-
-  recurseIntoTypes?: Record<string, RecurseIntoTypeDef>
 }
 
 export type DeployApiDefinitions<AdditionalAction extends string, ClientOptions extends string> = {
--- a/packages/adapter-components/src/deployment/deploy/deploy.ts
+++ b/packages/adapter-components/src/deployment/deploy/deploy.ts
@@ -28,7 +28,7 @@ import { createDependencyGraph } from './graph'
 import { ChangeAndExtendedContext, DeployChangeInput } from '../../definitions/system/deploy/types'
 import { ChangeElementResolver } from '../../resolve_utils'
 import { ResolveAdditionalActionType } from '../../definitions/system/api'
-import { createChangesForSubResources } from './deploy_subresources'
+import { createChangesForSubResources } from './subresources'
 
 const log = logger(module)
 const { mapValuesAsync } = promises.object
--- a/packages/adapter-components/src/deployment/deploy/deploy.ts
+++ b/packages/adapter-components/src/deployment/deploy/deploy.ts
@@ -133,6 +133,27 @@ export const deployChanges = async <TOptions extends APIDefinitionsOptions>({
     defQuery.query(getFailedChangeID(elemID).typeName)?.failIfChangeHasErrors !== false &&
     !_.isEmpty(errors[getFailedChangeID(elemID).getFullName()]?.filter(e => e.severity === 'Error'))
 
+  const handleChangeFailure = (change: Change<InstanceElement>, err: Error): void => {
+    const { elemID } = getChangeData(change)
+    log.error('Deployment of %s (action %s) failed: %o', elemID, change.action, err)
+    const failedElemID = getFailedChangeID(elemID)
+    errors[failedElemID.getFullName()] ??= []
+    if (isSaltoError(err)) {
+      errors[failedElemID.getFullName()].push({
+        ...err,
+        elemID: failedElemID,
+      })
+    } else {
+      const message = `${err}`
+      errors[failedElemID.getFullName()].push({
+        message,
+        detailedMessage: message,
+        severity: 'Error',
+        elemID: failedElemID,
+      })
+    }
+  }
+
   await graph.walkAsync(async nodeID => {
     const { typeName, action, typeActionChanges } = graph.getData(nodeID)
 
@@ -169,23 +190,7 @@ export const deployChanges = async <TOptions extends APIDefinitionsOptions>({
             })
             return change
           } catch (err) {
-            log.error('Deployment of %s (action %s) failed: %o', elemID, change.action, err)
-            const failedElemID = getFailedChangeID(getChangeData(change).elemID)
-            errors[failedElemID.getFullName()] ??= []
-            if (isSaltoError(err)) {
-              errors[failedElemID.getFullName()].push({
-                ...err,
-                elemID: failedElemID,
-              })
-            } else {
-              const message = `${err}`
-              errors[failedElemID.getFullName()].push({
-                message,
-                detailedMessage: message,
-                severity: 'Error',
-                elemID: failedElemID,
-              })
-            }
+            handleChangeFailure(change, err)
             return undefined
           }
         }),
--- a/packages/adapter-components/src/deployment/deploy/deploy.ts
+++ b/packages/adapter-components/src/deployment/deploy/deploy.ts
@@ -19,7 +19,7 @@ import {
   createSaltoElementErrorFromError,
 } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
-import { types, values } from '@salto-io/lowerdash'
+import { types, values, promises } from '@salto-io/lowerdash'
 import { APIDefinitionsOptions, ApiDefinitions, queryWithDefault } from '../../definitions'
 import { ChangeAndContext } from '../../definitions/system/deploy'
 import { getRequester } from './requester'
@@ -28,8 +28,10 @@ import { createDependencyGraph } from './graph'
 import { ChangeAndExtendedContext, DeployChangeInput } from '../../definitions/system/deploy/types'
 import { ChangeElementResolver } from '../../resolve_utils'
 import { ResolveAdditionalActionType } from '../../definitions/system/api'
+import { createChangesForSubResources } from './deploy_subresources'
 
 const log = logger(module)
+const { mapValuesAsync } = promises.object
 
 export type ConvertError = (elemID: ElemID, error: Error) => Error | SaltoElementError | undefined
 
@@ -95,7 +97,28 @@ export const deployChanges = async <TOptions extends APIDefinitionsOptions>({
   const defQuery = queryWithDefault(definitions.deploy.instances)
   const { dependencies } = definitions.deploy
 
-  const graph = createDependencyGraph({ defQuery, dependencies, changes, ...changeContext })
+  const changesByID = _.keyBy(changes, change => getChangeData(change).elemID.getFullName())
+  const subResourceChangesByChangeID = await mapValuesAsync(changesByID, async change =>
+    createChangesForSubResources({ change, definitions, context: changeContext }),
+  )
+  const subResourceChangeIDToOriginChangeID = _.mapValues(
+    Object.fromEntries(
+      Object.entries(subResourceChangesByChangeID).flatMap(([originChangeID, subResourceChanges]) =>
+        subResourceChanges.map(subResourceChange => [
+          getChangeData(subResourceChange).elemID.getFullName(),
+          changesByID[originChangeID],
+        ]),
+      ),
+    ),
+    change => getChangeData(change).elemID,
+  )
+
+  const graph = createDependencyGraph({
+    defQuery,
+    dependencies,
+    changes: changes.concat(Object.values(subResourceChangesByChangeID).flat()),
+    ...changeContext,
+  })
 
   const errors: Record<string, SaltoElementError[]> = {}
   const appliedChanges: Change<InstanceElement>[] = []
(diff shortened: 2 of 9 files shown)


packages/adapter-components/src/deployment/deploy/deploy.ts
```diff
@@ -75,6 +85,102 @@ const createSingleChangeDeployer = <TOptions extends APIDefinitionsOptions>({
   }
 }
 
+const createChangesForRecurseInto = async <TOptions extends APIDefinitionsOptions>(
```
shir-reifenberg, line 88: I will move all of this to a different file

packages/adapter-components/src/deployment/deploy/graph.ts
```diff
@@ -99,8 +99,17 @@ export const createDependencyGraph = <ClientOptions extends string, AdditionalAc
         }
       })
       graph.addEdge(toNodeID(typeName, second), toNodeID(typeName, first))
+      Object.keys(defQuery.query(typeName)?.recurseIntoTypes ?? []).forEach(recurseIntoTypeName => {
```
shir-reifenberg, line 102: currently creating dependencies between every "parent" change/operation to the "child" change/operation
might be better to create only in some cases like
parent/add -> child/add
child/remove -> parent/remove

packages/adapter-components/src/deployment/deploy/deploy_subresources.ts
```diff
@@ -0,0 +1,232 @@
...
+    return toChange({ before: beforeInstance, after: afterInstance })
+  }
+
+  const changes: Change<InstanceElement>[] = (
+    await Promise.all(
+      Object.entries(recurseIntoTypes).map(
+        async ([recurseIntoTypeName, { condition, fieldPath, changeIdFields, onActions }]) => {
+          if (onActions !== undefined && !onActions.includes(change.action)) {
+            log.trace(
+              'skipping recurce into %s in type %s and action %s',
```
shir-reifenberg, line 146: typo
```suggestion
              'skipping recurse into %s in type %s and action %s',
```

packages/adapter-components/src/definitions/system/deploy/deploy.ts
```diff
@@ -103,6 +111,8 @@ export type InstanceDeployApiDefinitions<AdditionalAction extends string, Client
   // fail if the change already has (error-level) errors from previous actions
   // default: true
   failIfChangeHasErrors?: boolean
+
+  recurseIntoTypes?: Record<string, RecurseIntoTypeDef>
```
netama, line 115: * not sure I understand what it means that the key is the type - what if there are multiple fields with the same type (think of a linked list with `before` and `after`)? I also don't know that we want to assume all the fields of the same nested type would have identical `recurseInto` definitions
* I think we should place this higher up (right after/before `requestsByAction`) with details explaining how this impacts what goes into which deploy definition and what's expected from the developer?

packages/adapter-components/src/definitions/system/deploy/deploy.ts
```diff
@@ -64,6 +64,14 @@ export type ActionDependency<AdditionalAction extends string> = {
   second: ActionName | AdditionalAction
 }
 
+export type RecurseIntoTypeDef = {
```
netama, line 67: I think for clarity for the reader (and consistency with the `recurseInto` fetch support) maybe we should mention the type name explicitly + throw / log an error if it doesn't match the field type?

packages/adapter-components/src/deployment/deploy/deploy_subresources.ts
```diff
@@ -0,0 +1,232 @@
+/*
```
netama, line 1: nit: I think there's no need for `deploy_` in the filename as it's under a `deploy` folder

packages/adapter-components/src/deployment/deploy/deploy_subresources.ts
```diff
@@ -0,0 +1,232 @@
...
+ * Returns the type of the sub resource at a given path, as the original type is required for resolving references.
+ * If the type is not found, returns an empty type with the expected type name.
+ */
+const getSubResourceType = (instance: InstanceElement, path: string[], recurseIntoTypeName: string): ObjectType => {
+  const topLevelType = instance.getTypeSync()
+  const getTypeInPath = (type: ObjectType | undefined, fieldPath: string[]): ObjectType | undefined => {
+    if (type === undefined || fieldPath.length === 0) {
+      return type
+    }
+    const field = type.fields[fieldPath[0]]
```
netama, line 46: we need all the usual handling here around things like `toString` that would require checking `hasOwnProperty` or using lodash - can you please copy the logic from one of the other places accessing by path?

packages/adapter-components/src/deployment/deploy/deploy_subresources.ts
```diff
@@ -0,0 +1,232 @@
...
+    if (!field) {
```
netama, line 47: please use explicit checks (e.g. `!== undefined`)

packages/adapter-components/src/deployment/deploy/deploy_subresources.ts
```diff
@@ -0,0 +1,232 @@
...
+      log.warn('failed to find type for field %s in type %s', fieldPath[0], type.elemID.getFullName())
+      return undefined
+    }
+
+    // we currently ignore paths that contains any container types
```
netama, line 52: not sure I understand the comment, isn't this for arrays?

packages/adapter-components/src/deployment/deploy/deploy_subresources.ts
```diff
@@ -0,0 +1,232 @@
...
+    // TODO SALTO-7041 support primitive types
+    if (!isObjectType(fieldType)) {
+      log.warn('field %s in type %s is not an object type', fieldPath[0], type.elemID.getFullName())
+      return undefined
+    }
+    return getTypeInPath(fieldType, path.slice(1))
+  }
+  const typeInPath = getTypeInPath(topLevelType, path)
+  if (typeInPath?.elemID?.name !== recurseIntoTypeName) {
+    log.error(
```
netama, line 63: are you sure we shouldn't throw an error in this case? fixing could involve just overriding the definitions so it'd be quick to get to something like this if we decided this was the fix

packages/adapter-components/src/deployment/deploy/deploy_subresources.ts
```diff
@@ -0,0 +1,232 @@
...
+
+  const getSubResourceId = (item: Values, changeIdFields: string[]): string => {
+    const nameParts = changeIdFields.map(fieldName => ({
+      fieldName,
+      fieldValue: _.get(item, fieldName),
+    }))
+    const [fields, missingFields] = _.partition(nameParts, ({ fieldValue }) => fieldValue !== undefined)
+    if (missingFields.length > 0) {
+      log.warn(
+        'failed to fields %s in item %s of change %s',
```
netama, line 103: missing some word?

packages/adapter-components/src/deployment/deploy/deploy_subresources.ts
```diff
@@ -0,0 +1,232 @@
...
+  definitions: types.PickyRequired<ApiDefinitions<TOptions>, 'deploy'>
+  context: Omit<ChangeAndContext, 'change'>
+}): Promise<Change<InstanceElement>[]> => {
+  const defQuery = queryWithDefault(definitions.deploy.instances)
+  const { recurseIntoTypes } = defQuery.query(getChangeData(change).elemID.typeName) ?? {}
+  if (recurseIntoTypes === undefined) {
+    return []
+  }
+
+  const getSubResourceId = (item: Values, changeIdFields: string[]): string => {
```
netama, line 95: is there a way to reuse the elem id logic?

packages/adapter-components/src/deployment/deploy/deploy_subresources.ts
```diff
@@ -0,0 +1,232 @@
...
+            [CORE_ANNOTATIONS.PARENT]: [originalChangeRef],
+          })
+        : undefined
+    return toChange({ before: beforeInstance, after: afterInstance })
+  }
+
+  const changes: Change<InstanceElement>[] = (
+    await Promise.all(
+      Object.entries(recurseIntoTypes).map(
+        async ([recurseIntoTypeName, { condition, fieldPath, changeIdFields, onActions }]) => {
```
netama, line 143: continuing from the comment on the definition - I'm not sure I understand the decision to do this by nested type name and assume the field path is "secondary", I _think_ I'd expect this to be the other way around

netama (commented):
thanks! sorry about the delay - did a partial pass and added some questions

packages/adapter-components/src/definitions/system/deploy/deploy.ts
shir-reifenberg: Good point, I didn't think of the case when a few fields might have the same type. 1
So the alternative IMO is making `recurseIntoTypes` an array
I think the key shouldn't be the field (as in `standalone` definition), cause there might be scenarios when the field is nested

so something like 
```suggestion
  recurseIntoTypes?: RecurseIntoTypeDef[]
  
  
export type RecurseIntoTypeDef = {
  typeName: string
  fieldPath: string[]
  changeIdFields: string[]
  condition?: DeployRequestCondition
  // when missing, all actions are used. currently only standard actions are supported.
  onActions?: ActionName[]
}
  
```

WDYT @netama ?

packages/adapter-components/src/deployment/deploy/deploy_subresources.ts
shir-reifenberg: I think it was relevant at some point while I was working this (for cases that contain more than one level of container types) but I think it's not relevant anymore, will verify

packages/adapter-components/src/deployment/deploy/deploy_subresources.ts
shir-reifenberg: mmm not sure, I think it's important to have the type only for reference resolution. Do we always have type at this point?
don't have a strong opinion here, so what ever you think is better 😄

packages/adapter-components/src/deployment/deploy/deploy_subresources.ts
shir-reifenberg: you mean for the definitions of the implementation ?
if the first - I think it's a little too complex to what we need here and includes many properties that are irrelevant so it can be confusing

packages/adapter-components/src/definitions/system/deploy/deploy.ts
netama: we can start with that (should maybe rename `recurseIntoTypes`to just `recurseInto` with some explanation?), might want to revisit later but it shouldn't block us for now. 
does it have an effect on the "main" change or are users expected to omit the fields in the main deploy?

I think we should still make the `fieldPath` the more "central" thing here - this is about breaking down the parent element, and we're pointing to the child type that contains the actual definition.

packages/adapter-components/src/deployment/deploy/deploy_subresources.ts
shir-reifenberg: If we want the field to be the key it means it can't be an array
I guess it can include `.` but I think it's kind of implicit

packages/adapter-components/src/deployment/deploy/subresources.ts
```diff
@@ -0,0 +1,233 @@
...
+        const [beforeItemsById, afterItemsById] = [
+          isRemovalOrModificationChange(change) ? change.data.before : undefined,
+          isAdditionOrModificationChange(change) ? change.data.after : undefined,
+        ]
+          .map(instance => makeArray(_.get(instance?.value, fieldPath)))
+          .map(items => _.keyBy(items, item => getSubResourceId(item, changeIdFields)))
+
+        const recurseIntoType = getSubResourceType(getChangeData(change), fieldPath, typeName)
+
+        const additions = Object.entries(afterItemsById)
```
Nurdok, line 176: I think you can make this code a lot shorter by using `lowerdash`'s `iterateTogether` function, something like:

```ts
const [beforeItems, afterItems] = [
          isRemovalOrModificationChange(change) ? change.data.before : undefined,
          isAdditionOrModificationChange(change) ? change.data.after : undefined,
        ]
  .map(instance => makeArray(_.get(instance?.value, fieldPath)))

const subResourceChanges = iterateTogether(beforeItems, afterItems, item => getSubResourceId(item, changeIdFields))
  .map(({before, after}) => createChangeFromSubResource({id, data: {before, after} }, recurseIntoType, change))
```

`iterateTogether` will returns pairs of `before`/`after` based on the ID - if there's are items with the same ID on both iteratees, then they will be returned together, otherwise you'll get an item and `undefined`.

packages/adapter-components/src/deployment/deploy/subresources.ts
```diff
@@ -0,0 +1,233 @@
...
+          log.trace(
+            'skipping recurse into %s in type %s and action %s',
+            typeName,
+            getChangeData(change).elemID.typeName,
+            change.action,
+          )
+          return []
+        }
+
+        const checkFunc = createCheck(condition, fieldPath)
```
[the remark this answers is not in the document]
shir-reifenberg: not sure I understand the suggestion 🤔 you mean editing the `condition` inplace and add a root to it?

packages/adapter-components/src/deployment/deploy/subresources.ts
shir-reifenberg: That's really cool! wasn't familiar with `iterateTogether`
Though I do see it's async though this code isn't, do you think it's worth it ?

packages/adapter-components/src/deployment/deploy/subresources.ts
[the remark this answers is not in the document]
Nurdok: I actually forgot `condition` has `tranformForCheck`, but using it is a better idea! Note that it's delicate if the original condition already uses `root` (this might be a good use case to add to SALTO-6583). I think you can do something like:


```ts
const subResourceCondition = condition.clone()
subResourceCondition.transformForCheck = { 
  ...subResourceCondition.transformForCheck, 
  root: subResourceCondition.transformForCheck.root === undefined ? fieldPath : [fieldPath, subResourceCondition.transformForCheck.root].join('.')
}
```

---

I originally meant to do something like this, which is more complicated:

```ts
const checkFunc = createCheck(condition)
const transformForCheck = createValueTransformer({ root: fieldPath })
// build a new change by running `transformForCheck` on the values of `before` and `after`
if (!await checkFunc({transformedChange, ...context, errors: {} }))) {
  [..]
}
```

---

But you can do it any other way, I just think that you shouldn't change `createCheck` but rather give it a modified input.

packages/adapter-components/src/deployment/deploy/subresources.ts
Nurdok: Hadn't noticed the async there. I'd copy `iterateTogether` to `lowerdash`'s `iterable.ts` file and make a sync version, but if you want to leave it as is, that's okay too.

packages/adapter-components/src/deployment/deploy/deploy.ts
```diff
@@ -95,17 +97,41 @@ export const deployChanges = async <TOptions extends APIDefinitionsOptions>({
   const defQuery = queryWithDefault(definitions.deploy.instances)
   const { dependencies } = definitions.deploy
 
-  const graph = createDependencyGraph({ defQuery, dependencies, changes, ...changeContext })
+  const changesByID = _.keyBy(changes, change => getChangeData(change).elemID.getFullName())
```
Nurdok, line 100: General note: I'm a little worried about how long this function is getting... over 100 lines with this change. I don't have a specific suggestion, but if you can see a way to extract some of the new logic into a function I think it'll be more readable.

packages/adapter-components/src/deployment/deploy/graph.ts
```diff
@@ -100,7 +104,28 @@ export const createDependencyGraph = <ClientOptions extends string, AdditionalAc
       })
       graph.addEdge(toNodeID(typeName, second), toNodeID(typeName, first))
     })
+
+    // Add dependencies from original changes to subresource changes created for deployment
```
Nurdok, line 108: This function, too, is getting pretty long... Again, not sure what to do about this

packages/adapter-components/src/definitions/system/deploy/deploy.ts
```diff
@@ -64,13 +64,30 @@ export type ActionDependency<AdditionalAction extends string> = {
...
 }
 
+export type RecurseIntoPath = {
+  fieldPath: string[]
+  // the type name in the specified path, the type should have its own deploy definition
+  typeName: string
+  changeIdFields: string[]
+  // when provided, only changes matching the condition will recurse into specified path
+  condition?: DeployRequestCondition
+  // when missing, all actions are used. currently only standard actions are supported
```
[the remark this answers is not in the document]
Nurdok: Oh, because the graph only works on standard actions when computing dependencies? It's alright to leave as is.

packages/adapter-components/src/deployment/deploy/subresources.ts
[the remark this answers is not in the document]
shir-reifenberg: ahh just realized it doesn't work as `root` is `string` and `fieldPath` an array

packages/adapter-components/src/deployment/deploy/subresources.ts
shir-reifenberg: gonna skip this time 😅

packages/adapter-components/src/fetch/resource/type_fetcher.ts
```diff
@@ -124,6 +124,15 @@ export const createTypeResourceFetcher = <ClientOptions extends string>({
       return { success: true }
     }
 
+    const dependsOnDef = def.context?.dependsOn
```
Nurdok, line 127: Seems unrelated to this PR. Please remove :)

Nurdok (approved):

PR #6894, opened 2024-12-04
SALTO-7024: Using custom resolver in SFDX dump

yelly:
This is required for OrderedMaps.

---

---

_Release Notes_: 
None.

---

_User Notifications_:
None.

### DIFF
--- a/packages/salesforce-adapter/src/adapter.ts
+++ b/packages/salesforce-adapter/src/adapter.ts
@@ -486,7 +486,10 @@ export const salesforceAdapterResolveValues: ResolveValuesFunc = async (
     : resolvedElement
 }
 
-export const resolveChanges = (changes: readonly Change[], getLookupNameFunc: GetLookupNameFunc): Promise<Change[]> =>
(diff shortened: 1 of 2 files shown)


packages/salesforce-adapter/src/adapter.ts
```diff
@@ -479,6 +486,9 @@ export const salesforceAdapterResolveValues: ResolveValuesFunc = async (
     : resolvedElement
 }
 
+export const resolveChanges = (changes: readonly Change[], getLookupNameFunc: GetLookupNameFunc): Promise<Change[]> =>
```
tamtamirr, line 489:
```suggestion
export const resolveSalesforceChanges = (changes: readonly Change[], getLookupNameFunc: GetLookupNameFunc): Promise<Change[]> =>
```
nit: As `resolveChanges` might be a too generic export that may lead to wrong auto imports

tamtamirr (approved):
LGTM! One small nit :) Approving to save cycles

PR #6897, opened 2024-12-04
SALTO-6166: Avoid serializing redundant values

ori-moisis:
When serializing element IDs and Fields, only write the attributes that are needed to deserialize

---

On an example workspace (relatively small salesforce environment) - this reduced the overall size of the cache (rocksdb)  from 21MB to 14MB
most likely, this would mostly scale with the number of references in the workspace...

---
_Release Notes_: 
Core:
- Reduced size of state file and cache

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/workspace/src/serializer/elements.ts
+++ b/packages/workspace/src/serializer/elements.ts
@@ -34,6 +34,7 @@ import {
   isTypeReference,
   isVariableExpression,
   PlaceholderObjectType,
+  isField,
 } from '@salto-io/adapter-api'
 import { DuplicateAnnotationError, MergeError, isMergeError } from '../merger/internal/common'
 import { DuplicateInstanceKeyError } from '../merger/internal/instances'
@@ -191,7 +192,7 @@ export const serializeStream = async <T = Element>({
 
   const elemIdReplacer = (
     id: ElemID,
-  ): Omit<types.PickDataFields<ElemID>, 'nestingLevel' | 'name'> & { readonly nameParts: ReadonlyArray<string> } => ({
+  ): Pick<ElemID, 'adapter' | 'typeName' | 'idType'> & { readonly nameParts: ReadonlyArray<string> } => ({
     adapter: id.adapter,
     typeName: id.typeName,
     idType: id.idType,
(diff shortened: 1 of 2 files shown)


packages/workspace/src/serializer/elements.ts
```diff
@@ -200,12 +201,14 @@ export const serializeStream = async <T = Element>({
 
   const referenceExpressionReplacer = (e: ReferenceExpression): ReferenceExpression & SerializedClass => {
     if (e.value === undefined || referenceSerializerMode === 'keepRef' || !isVariableExpression(e)) {
-      return saltoClassReplacer(e.createWithValue(undefined))
+      // eslint-disable-next-line no-use-before-define
+      return saltoClassReplacer(_.cloneDeepWith(e.createWithValue(undefined), replacer))
```
ori-moisis, line 205: by not going over this with a replacer previously, the code for `elemIdReplacer` would not be called, and the element ID in the reference would be serialized with all of its members (including the `fullName`)

tamtamirr (approved):

packages/workspace/test/serializer/serializer.test.ts
```diff
@@ -355,6 +357,16 @@ describe('State/cache serialization', () => {
         new StaticFile({ filepath: staticFile.filepath, hash: staticFile.hash }),
       )
     })
+    it('should serialize fields without values that are not used for deserialize', async () => {
```
Nurdok, line 360: The nittiest of nits: there's grammatical ambiguity here, I read this as "should serialize fields that don't have values". Can you rephrase to, e.g., "should drop values that are not use for deserialization while serializing fields" or something like that?

packages/workspace/src/serializer/elements.ts
```diff
@@ -200,12 +201,14 @@ export const serializeStream = async <T = Element>({
 
   const referenceExpressionReplacer = (e: ReferenceExpression): ReferenceExpression & SerializedClass => {
     if (e.value === undefined || referenceSerializerMode === 'keepRef' || !isVariableExpression(e)) {
-      return saltoClassReplacer(e.createWithValue(undefined))
+      // eslint-disable-next-line no-use-before-define
```
Nurdok, line 204: Can you avoid all these use-before-defines by wrapping these functions in a class?
Feel free to ignore

Nurdok (approved):

packages/workspace/src/serializer/elements.ts
ori-moisis: merging this for now, feel free to comment anyway, I'll open a separate refactor PR if we need to change this

PR #6900, opened 2024-12-05
Fix NS config suggestions when there are config overrides

dantal4:
The config suggestions should be on the config instance that the adapter receives from core, without the modifications that are done in `config_creator`, otherwise we might return a modification to a field that was override using a config override, which isn't allowed.

---

---
_Release Notes_: 
Netsuite Adapter:
- Fix NS config suggestions when there are config overrides

---
_User Notifications_: 
None

### DIFF
--- a/packages/netsuite-adapter/e2e_test/adapter.ts
+++ b/packages/netsuite-adapter/e2e_test/adapter.ts
@@ -63,7 +63,7 @@ export const realAdapter = (
     client,
     elementsSource: buildElementsSourceFromElements(elements),
     config: config ?? { fetch: fullFetchConfig() },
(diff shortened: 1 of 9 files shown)


omrilit (approved):
The code looks good so I am approving.
The only problem that I have is that when I see `config` & `userConfig` I don't really understand what do they represent and what they differ from each other. Could you please think of a better naming?

dantal4:
> The code looks good so I am approving. The only problem that I have is that when I see `config` & `userConfig` I don't really understand what do they represent and what they differ from each other. Could you please think of a better naming?

@omrilit can you suggest better names?

PR #6902, opened 2024-12-05
SALTO-6995: fix element test utils types path

ori-moisis:
Also export "allEqualityTesters" as an array so if more are added in the future, depending packages won't have to change

---

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/element-test-utils/package.json
+++ b/packages/element-test-utils/package.json
@@ -14,7 +14,7 @@
     "dist/src",
     "dist/index.*"
   ],
-  "types": "dist/src/index.d.ts",
+  "types": "dist/index.d.ts",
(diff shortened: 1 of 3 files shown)


packages/element-test-utils/package.json
```diff
@@ -14,7 +14,7 @@
     "dist/src",
     "dist/index.*"
   ],
-  "types": "dist/src/index.d.ts",
```
ori-moisis, line 17: the old path doesn't exist, which means if a package tries to import this, it will fail and say it cannot find the types...

Nurdok (approved):

PR #6904, opened 2024-12-05
SALTO-6197 Salesforce adapter - default_rules CV will catch profiles with no default record type

rotem531:
SALTO-6197 Salesforce adapter - profiles with no default record type will be caught at default_rules CV, formerly multiple_defaults
---
additional context for reviewer:
When a Profile has recordTypes, exactly one entry must be both default and visible
error when default is not visible:
image
error when there is at least one visible entry but no default entry at all:
image

---
_Release Notes_: 
SalesforceAdapter:
multiple_defaults Change Validator changed to default_rules, and catches when a recordType have no default=true

---
_User Notifications_: 
none


packages/salesforce-adapter/src/change_validators/recordTypeVisibility_no_default.ts
```diff
@@ -0,0 +1,119 @@
...
+        return true
+      }
+      const noDefaultNoVisible = ans.noDefaultNoVisibleFlag ?? true
+      const oneDefaultIsVisible = ans.oneDefaultIsVisibleFlag ?? true
+      return noDefaultNoVisible || oneDefaultIsVisible
+    }
+    return true
+  })
+
+const getRecordTypeVisibilityNoDefaultError = async (change: InstanceElement | undefined): Promise<ChangeError[]> => {
```
Uri-Bar, line 85: this shoudnlt get InstanceElement | undefined you are able to know before you can call this func

packages/salesforce-adapter/src/change_validators/recordTypeVisibility_no_default.ts
```diff
@@ -0,0 +1,119 @@
...
+    return []
+  }
+  const recordEntries = _.entries(record)
+  if (recordEntries.length === 0) {
+    return []
+  }
+  const recordsOnly = recordEntries.map(getRecEntryFields).filter(a => a !== undefined)
+  if (!recordsOnly) {
+    return []
+  }
```
Uri-Bar, line 97: cant you replace all of this with one liner map + filter?

packages/salesforce-adapter/src/change_validators/recordTypeVisibility_no_default.ts
```diff
@@ -0,0 +1,119 @@
...
+  const elementsToCheck = recordsOnly.map(getKeys).filter(a => a !== undefined)
```
Uri-Bar, line 98: some of the keys are undefined? we sure about it? why when does that happen?

packages/salesforce-adapter/src/change_validators/recordTypeVisibility_no_default.ts
```diff
@@ -0,0 +1,119 @@
...
+  const changeErrors = isValidRecord(recordsOnly, elementsToCheck)
```
Uri-Bar, line 99: instead of holding an array of boolean its should be in a filter step here

packages/salesforce-adapter/src/change_validators/recordTypeVisibility_no_default.ts
```diff
@@ -0,0 +1,119 @@
...
+  return undefined
+}
+
+/*
+ * this function checks the relevant fields if they're valid
+ * valid option 1: no default field is true and no visible field is true (marked initially as true [valid] in res[0], and changed to false [invalid] in case of default or visible field being true)
+ * valid option 2: there is exactly one element with default field as true and it's visible field is also true (marked initially as false [invalid] in res[1], and changed to true [valid] in case one element's default and visible fields are true)
+ * NOTICE: although it is not allowed to have more than one element with default field set to true, this CV doesn't check for this case because it is caught in the 'multiple_defaults' CV
+ */
+const isValidRecord = (records: (Value | undefined)[], fieldsToCheck: (string[] | undefined)[]): boolean[] =>
```
Uri-Bar, line 54: as I wrote in the previous comment, you should change this func to run on a single record and return bool.
another comment I have it that this func is really hard to read (I didnt read the comment) we usually only use comments on stuff more complicated than that.
you can see in the CV I sent you that I devided this func into several fucs that calls each other and it makes it much readable

packages/salesforce-adapter/src/change_validators/recordTypeVisibility_no_default.ts
```diff
@@ -0,0 +1,119 @@
...
+
+type fieldsFlags = { noDefaultNoVisibleFlag: boolean; oneDefaultIsVisibleFlag: boolean }
+
+const TYPES_WITH_RECORD_TYPE_VISIBILITY = new Set([PROFILE_METADATA_TYPE])
+
+const createNoDefaultError = (id: ElemID, recordTypeVisibility: string): ChangeError => ({
+  elemID: id,
+  severity: 'Error',
+  message: 'Must have one field that is set as default and visible',
+  detailedMessage: `Must have one field that is set as default and visible, or no default fields and no visible fields, under ${id.getFullName()}.${recordTypeVisibility}`,
```
Uri-Bar, line 31: the elem id is in his own filed, you dont need print it again here

packages/salesforce-adapter/src/change_validators/recordTypeVisibility_no_default.ts
```diff
@@ -0,0 +1,119 @@
...
+const { isDefined } = values
+
+type fieldsFlags = { noDefaultNoVisibleFlag: boolean; oneDefaultIsVisibleFlag: boolean }
+
+const TYPES_WITH_RECORD_TYPE_VISIBILITY = new Set([PROFILE_METADATA_TYPE])
+
+const createNoDefaultError = (id: ElemID, recordTypeVisibility: string): ChangeError => ({
+  elemID: id,
+  severity: 'Error',
+  message: 'Must have one field that is set as default and visible',
```
Uri-Bar, line 30: I think you can find a better way of stating this, think that a user reads that and need to understand what is the problem and how to fix it

packages/salesforce-adapter/src/change_validators/recordTypeVisibility_no_default.ts
Uri-Bar: why is this func async?

packages/salesforce-adapter/src/change_validators/recordTypeVisibility_no_default.ts
```diff
@@ -0,0 +1,119 @@
...
+    .filter(isDefined)
+}
+
+const changeValidator: ChangeValidator = async changes => {
+  const errors = changes
+    .map(getChangeIfInstanceChange)
+    .filter(change => (change !== undefined ? TYPES_WITH_RECORD_TYPE_VISIBILITY.has(change?.elemID.typeName) : false))
+    .map(getRecordTypeVisibilityNoDefaultError)
+  const resolved = await Promise.all(errors)
+  return resolved.flat()
```
Uri-Bar, line 116: we are not using this method, we are using awu when we need to iterate over an array with async func

Uri-Bar: in second look, all of the func you call are sync, so we dont need that at all

packages/salesforce-adapter/src/change_validators/recordTypeVisibility_no_default.ts
```diff
@@ -0,0 +1,119 @@
...
+      }
+      return undefined
+    })
+    .filter(isDefined)
+}
+
+const changeValidator: ChangeValidator = async changes => {
+  const errors = changes
+    .map(getChangeIfInstanceChange)
+    .filter(change => (change !== undefined ? TYPES_WITH_RECORD_TYPE_VISIBILITY.has(change?.elemID.typeName) : false))
```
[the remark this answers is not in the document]
Uri-Bar: add isAdditionOrModification

packages/salesforce-adapter/src/change_validators/recordTypeVisibility_no_default.ts
```diff
@@ -0,0 +1,119 @@
...
+        return createNoDefaultError(change.elemID, `recordTypeVisibilities.${val[0]}`)
+      }
+      return undefined
+    })
+    .filter(isDefined)
+}
+
+const changeValidator: ChangeValidator = async changes => {
+  const errors = changes
+    .map(getChangeIfInstanceChange)
```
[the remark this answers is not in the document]
Uri-Bar: add isAdditionOr<odification

packages/salesforce-adapter/src/change_validators/recordTypeVisibility_no_default.ts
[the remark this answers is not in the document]
Uri-Bar: look for getType/ getTypeSync, and dont reach .typeName directly

packages/salesforce-adapter/src/change_validators/recordTypeVisibility_no_default.ts
```diff
@@ -0,0 +1,119 @@
...
+      }
+      const noDefaultNoVisible = ans.noDefaultNoVisibleFlag ?? true
+      const oneDefaultIsVisible = ans.oneDefaultIsVisibleFlag ?? true
+      return noDefaultNoVisible || oneDefaultIsVisible
+    }
+    return true
+  })
+
+const getRecordTypeVisibilityNoDefaultError = async (change: InstanceElement | undefined): Promise<ChangeError[]> => {
+  const record = change?.value?.recordTypeVisibilities
```
Uri-Bar, line 86: change the name to recordTypeVisibilities maybe use { recordTypeVisibilities } = change?.value?

packages/salesforce-adapter/src/change_validators/recordTypeVisibility_no_default.ts
Uri-Bar: use object.keys directly

packages/salesforce-adapter/src/change_validators/recordTypeVisibility_no_default.ts
Uri-Bar: that runs separately on each recordType

packages/salesforce-adapter/src/change_validators/recordTypeVisibility_no_default.ts
```diff
@@ -0,0 +1,119 @@
...
+    return []
+  }
+  const recordsOnly = recordEntries.map(getRecEntryFields).filter(a => a !== undefined)
+  if (!recordsOnly) {
+    return []
+  }
+  const elementsToCheck = recordsOnly.map(getKeys).filter(a => a !== undefined)
+  const changeErrors = isValidRecord(recordsOnly, elementsToCheck)
+  return recordEntries
+    .map((val, index) => {
```
Uri-Bar, line 101: we shoudnlt use ((val, index) here, after you will implement all the above changes all this line will turn into a map to the change error

Uri-Bar (commented):
more comments f2f

packages/salesforce-adapter/config_doc.md
```diff
@@ -329,28 +329,28 @@ For more details see the DeployOptions section in the [salesforce documentation
...
+| managedPackage               | true                   | Disallow changes to objects and fields that are part of a managed package                  |
+| picklistStandardField        | true                   | It is forbidden to modify a picklist on a standard field. Only StandardValueSet is allowed |
+| customObjectInstances        | true                   | Validate permissions of creating / update data records                                     |
+| customFieldType              | true                   | Ensure the type given to a custom field is a valid type for custom fields                  |
+| standardFieldLabel           | true                   | Disallow changing a label of a standard field                                              |
+| mapKeys                      | true                   | Ensure proper structure of profiles before deploying                                       |
+| defaults_rules               | true                   | Check for default rules violations such as multiple defaults, no default etc.              |
```
Uri-Bar, line 340: please make it defaultsRules in camelCase

packages/salesforce-adapter/src/change_validator.ts
```diff
@@ -15,7 +15,7 @@ import customObjectInstancesValidator from './change_validators/custom_object_in
 import customFieldTypeValidator from './change_validators/custom_field_type'
 import standardFieldLabelValidator from './change_validators/standard_field_label'
 import mapKeysValidator from './change_validators/map_keys'
-import multipleDefaultsValidator from './change_validators/multiple_defaults'
+import multipleDefaultsValidator from './change_validators/default_rules'
```
Uri-Bar, line 18: you need to change the name of the import as well

Uri-Bar: and also in the map object in this file.
and in the types files in the adapter package

packages/salesforce-adapter/src/change_validators/default_rules.ts
```diff
@@ -169,7 +205,13 @@ const getInstancesMultipleDefaultsErrors = async (after: InstanceElement): Promi
         return awu(Object.entries(value)).flatMap(async ([_key, innerValue]) => {
           const startLevelType = isMapType(fieldType) ? fieldType.getInnerTypeSync() : fieldType
           const defaultsContexts = await findMultipleDefaults(innerValue, startLevelType, valueName)
-          return createChangeErrorFromContext(field, defaultsContexts, after)
+          if (defaultsContexts !== undefined) {
+            return createChangeErrorFromContext(field, defaultsContexts, after)
```
Uri-Bar, line 209: you can change this func to get context that is not undefined as you check here

Uri-Bar: actually maybe you can delete this func and call createInstanceChangeError directly now

packages/salesforce-adapter/src/change_validators/default_rules.ts
```diff
@@ -99,6 +99,20 @@ const createInstanceChangeError = (field: Field, contexts: string[], instance: I
   }
 }
 
+const createInstanceChangeErrorSingleDefault = (
+  fieldPath: string,
+  field: string,
+  instance: InstanceElement,
+): ChangeError => {
+  const elementId = instance.elemID.createNestedID(fieldPath, field, 'visible')
```
Uri-Bar, line 107: why did you change the error message from the one in the description of the PR, I think its better

packages/salesforce-adapter/src/change_validators/default_rules.ts
```diff
@@ -143,6 +157,28 @@ const getInstancesMultipleDefaultsErrors = async (after: InstanceElement): Promi
...
+      const visibleField = _.get(curr, 'visible')
+      if (defaultField) {
+        if (visibleField) {
+          return ''
+        }
+        return curr.recordType
+      }
+      return undefined
+    }, undefined)
+    return isValid !== undefined && isValid !== '' ? isValid : undefined
```
Uri-Bar, line 179:
```suggestion
    return isValid !== '' ? isValid : undefined
```

packages/salesforce-adapter/src/change_validators/default_rules.ts
```diff
@@ -169,7 +205,13 @@ const getInstancesMultipleDefaultsErrors = async (after: InstanceElement): Promi
         return awu(Object.entries(value)).flatMap(async ([_key, innerValue]) => {
           const startLevelType = isMapType(fieldType) ? fieldType.getInnerTypeSync() : fieldType
           const defaultsContexts = await findMultipleDefaults(innerValue, startLevelType, valueName)
-          return createChangeErrorFromContext(field, defaultsContexts, after)
+          if (defaultsContexts !== undefined) {
+            return createChangeErrorFromContext(field, defaultsContexts, after)
+          }
+          const singleDefaultisValid = checkSingleDefault(innerValue, startLevelType)
+          return singleDefaultisValid && singleDefaultisValid !== ''
```
Uri-Bar, line 212: how can singleDefaultisValid be '' if you change it to undefined?

packages/salesforce-adapter/src/change_validators/default_rules.ts
```diff
@@ -169,7 +205,13 @@ const getInstancesMultipleDefaultsErrors = async (after: InstanceElement): Promi
         return awu(Object.entries(value)).flatMap(async ([_key, innerValue]) => {
           const startLevelType = isMapType(fieldType) ? fieldType.getInnerTypeSync() : fieldType
           const defaultsContexts = await findMultipleDefaults(innerValue, startLevelType, valueName)
-          return createChangeErrorFromContext(field, defaultsContexts, after)
+          if (defaultsContexts !== undefined) {
+            return createChangeErrorFromContext(field, defaultsContexts, after)
+          }
+          const singleDefaultisValid = checkSingleDefault(innerValue, startLevelType)
```
Uri-Bar, line 211: nit: change the check into find

packages/salesforce-adapter/src/change_validators/default_rules.ts
Uri-Bar: it does not have to be the same as in the pic (also change the pic afterwards) but it should say something like bla bla should have at least one set default recordType which is visible

Uri-Bar (commented):
looks much better, added few comments

packages/salesforce-adapter/src/types.ts
```diff
@@ -148,7 +148,7 @@ export type ChangeValidatorName =
   | 'customFieldType'
   | 'standardFieldLabel'
   | 'mapKeys'
-  | 'multipleDefaults'
+  | 'DefaultRules'
```
Uri-Bar, line 151: no need for capital D

packages/salesforce-adapter/src/types.ts
```diff
@@ -836,7 +836,7 @@ const changeValidatorConfigType = createMatchingObjectType<ChangeValidatorConfig
     customFieldType: { refType: BuiltinTypes.BOOLEAN },
     standardFieldLabel: { refType: BuiltinTypes.BOOLEAN },
     mapKeys: { refType: BuiltinTypes.BOOLEAN },
-    multipleDefaults: { refType: BuiltinTypes.BOOLEAN },
+    DefaultRules: { refType: BuiltinTypes.BOOLEAN },
```
Uri-Bar, line 839: same

packages/salesforce-adapter/src/change_validators/default_rules.ts
```diff
@@ -99,6 +99,20 @@ const createInstanceChangeError = (field: Field, contexts: string[], instance: I
...
+const createInstanceChangeErrorSingleDefault = (
+  fieldPath: string,
+  field: string,
+  instance: InstanceElement,
+): ChangeError => {
+  const elementId = instance.elemID.createNestedID(fieldPath, field, 'visible')
+  return {
+    elemID: elementId,
+    severity: 'Error',
+    message: 'Default entry must be visible',
```
Uri-Bar, line 111: maybe Im missing something but what about the case of no defaults at all are we catching it?

Uri-Bar (commented):
I have a filling that we are missing one case here, lets talk f2f

packages/salesforce-adapter/src/change_validators/default_rules.ts
Uri-Bar: talked f2f need to add more change error for visible without default and catch it on your code

Uri-Bar (commented):
fixed my comment

packages/salesforce-adapter/src/change_validators/default_rules.ts
```diff
@@ -99,6 +99,34 @@ const createInstanceChangeError = (field: Field, contexts: string[], instance: I
   }
 }
 
+const createInstanceChangeErrorSingleDefault = (
```
Uri-Bar, line 102: the word visible needs to be in the change error name

packages/salesforce-adapter/src/change_validators/default_rules.ts
```diff
@@ -169,7 +238,17 @@ const getInstancesMultipleDefaultsErrors = async (after: InstanceElement): Promi
         return awu(Object.entries(value)).flatMap(async ([_key, innerValue]) => {
           const startLevelType = isMapType(fieldType) ? fieldType.getInnerTypeSync() : fieldType
           const defaultsContexts = await findMultipleDefaults(innerValue, startLevelType, valueName)
-          return createChangeErrorFromContext(field, defaultsContexts, after)
+          if (defaultsContexts !== undefined) {
+            return [createInstanceChangeError(field, defaultsContexts, after)]
+          }
+          const singleDefaultisValid = findSingleDefault(innerValue, startLevelType)
+          if (singleDefaultisValid !== undefined) {
+            return singleDefaultisValid !== ''
```
Uri-Bar, line 246: I have already asked you about it in other versions of this code. singleDefaultisValid shouldnt be able to be undefined and also empty string, it should fallback only to one of them and remove the double validation

packages/salesforce-adapter/src/change_validators/default_rules.ts
```diff
@@ -31,6 +31,8 @@ import { isFieldOfCustomObject } from '../transformers/transformer'
 
 const { awu } = collections.asynciterable
 
+const NO_ERRORS_FOUND: ChangeError[] = []
```
Uri-Bar, line 34: no need to set this var, just use []

packages/salesforce-adapter/src/change_validators/default_rules.ts
```diff
@@ -161,19 +221,41 @@ const getInstancesMultipleDefaultsErrors = async (after: InstanceElement): Promi
...
         return awu(Object.entries(value)).flatMap(async ([_key, innerValue]) => {
           const startLevelType = isMapType(fieldType) ? fieldType.getInnerTypeSync() : fieldType
-          const defaultsContexts = await findMultipleDefaults(innerValue, startLevelType, valueName)
-          return createChangeErrorFromContext(field, defaultsContexts, after)
+          const { defaults: defaultsContexts, count: defaultsCount } = await findMultipleDefaults(
+            innerValue,
+            startLevelType,
+            valueName,
+          )
+          if (defaultsContexts === undefined) {
```
Uri-Bar, line 236: i think you are missing this line here:
          return createChangeErrorFromContext(field, defaultsContexts, after)
          
only if defaultsContexts !== undefined

packages/salesforce-adapter/src/change_validators/default_rules.ts
```diff
@@ -161,19 +221,41 @@ const getInstancesMultipleDefaultsErrors = async (after: InstanceElement): Promi
...
+          }
+          if (defaultsCount === 1) {
+            const singleDefaultIsValid = findNotVisibleDefault(innerValue, startLevelType)
+            if (singleDefaultIsValid !== undefined) {
+              return [createInstanceChangeErrorSingleDefaultNoVisible(fieldPath, singleDefaultIsValid, after)]
+            }
+            return NO_ERRORS_FOUND
+          }
+          const noDefaultValidation = findVisibleNoDefault(innerValue, startLevelType)
+          return !noDefaultValidation ? [createInstanceChangeErrorNoDefault(fieldPath, _key, after)] : NO_ERRORS_FOUND
```
Uri-Bar, line 250: maybe we ned to change the func name to `is..` or `has...` as it return boolean and not finding anything and then the next line will be without !

Uri-Bar (commented):
very small stuff. looking much better and cleaner

packages/salesforce-adapter/src/change_validators/default_rules.ts
Uri-Bar: not sure, in what cases would defaultsContexts be undefined?

Uri-Bar (approved):
Great job

PR #6905, opened 2024-12-05
SALTO-6759 Jamf: mask passwords

kami-delaroz:
Mask passwords within policy scripts and configuration_profiles payloads

---

---
_Release Notes_: 

_Jamf_adapter_:
- Mask passwords within policy scripts and configuration_profiles payloads

---
_User Notifications_: 
Passwords within policy scripts and configuration_profiles payloads will be masked

### DIFF
--- a/packages/jamf-adapter/src/definitions/fetch/transforms/utils.ts
+++ b/packages/jamf-adapter/src/definitions/fetch/transforms/utils.ts
@@ -10,7 +10,7 @@ import { Values } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
 import _ from 'lodash'
 
-export const MASK_VALUE = '**MASKED_PASSWORD**'
+export const MASK_VALUE = '**SALTO_MASKED_VALUE**'
 
 const PASSWORD_REGEX = /<key>Password<\/key><string>.*?<\/string>/
 const MASKED_PASSWORD = `<key>Password</key><string>${MASK_VALUE}</string>`
@@ -95,10 +95,12 @@ export const removeSelfServiceSecurityPassword = (value: Values): void => {
 export const maskPayloadsPassword = (value: Values): void => {
   const payloads = _.get(value, 'general.payloads')
   if (typeof payloads === 'string') {
-    if (PASSWORD_REGEX.test(payloads)) {
+    const maskedPayloads = payloads.replace(PASSWORD_REGEX, () => {
       log.trace(`Masked value of payloads in '${_.get(value, 'general.name')}'`)
-      _.set(value, 'general.payloads', payloads.replace(PASSWORD_REGEX, MASKED_PASSWORD))
-    }
+      return MASKED_PASSWORD
+    })
+    _.set(value, 'general.payloads', maskedPayloads)
+    // TODO: change to replaceAll to support all occasions in the payloads
   }
 }
 
--- a/packages/jamf-adapter/src/definitions/fetch/transforms/utils.ts
+++ b/packages/jamf-adapter/src/definitions/fetch/transforms/utils.ts
@@ -10,13 +10,13 @@ import { Values } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
 import _ from 'lodash'
 
-export const MASK_VALUE = '**SALTO_MASKED_VALUE**'
+export const SALTO_MASKED_VALUE = '**SALTO_MASKED_VALUE**'
 
 const PASSWORD_REGEX = /<key>Password<\/key><string>.*?<\/string>/
-const MASKED_PASSWORD = `<key>Password</key><string>${MASK_VALUE}</string>`
+const MASKED_PASSWORD = `<key>Password</key><string>${SALTO_MASKED_VALUE}</string>`
 
 const CLIENT_SECRET_REGEX = '^client_secret=.*'
-const MASKED_CLIENT_SECRET = `client_secret=${MASK_VALUE}`
+const MASKED_CLIENT_SECRET = `client_secret=${SALTO_MASKED_VALUE}`
 
 const log = logger(module)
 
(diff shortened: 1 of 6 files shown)


kami-delaroz:
CV warning will follow in a separated PR

packages/jamf-adapter/src/definitions/fetch/transforms/utils.ts
```diff
@@ -5,9 +5,17 @@
  *
  * CERTAIN THIRD PARTY SOFTWARE MAY BE CONTAINED IN PORTIONS OF THE SOFTWARE. See NOTICE FILE AT https://github.com/salto-io/salto/blob/main/NOTICES
  */
-import { values } from '@salto-io/lowerdash'
+import { regex as lowerdashRegex, values } from '@salto-io/lowerdash'
 import _ from 'lodash'
 
+export const MASK_VALUE = '**MASKED_PASSWORD**'
```
shir-reifenberg, line 11: IMO worth aligning with the same value we have in jira which is `<SECRET_TOKEN>`

packages/jamf-adapter/src/definitions/fetch/transforms/utils.ts
```diff
@@ -79,3 +87,23 @@ export const removeSelfServiceSecurityPassword = (value: Record<string, unknown>
...
+  }
+}
+
+export const maskPasswordsForScriptsObjectArray = (value: Record<string, unknown>): void => {
+  const { scripts } = value
+  if (Array.isArray(scripts)) {
+    scripts.forEach(script =>
+      Object.keys(script).forEach(key => {
+        if (lowerdashRegex.isFullRegexMatch(script[key], CLIENT_SECRET_REGEX)) {
+          script[key] = MASKED_CLIENT_SECRET
```
shir-reifenberg, line 104: Can we add a `log.trace` indicating we masked a value with elem id / other identifier if we have ? (same for other replacements)

packages/jamf-adapter/src/definitions/fetch/transforms/utils.ts
```diff
@@ -79,3 +87,23 @@ export const removeSelfServiceSecurityPassword = (value: Record<string, unknown>
     }
   }
 }
+
+export const maskPayloadsPassword = (value: Record<string, unknown>): void => {
```
shir-reifenberg, line 91: shouldn't this be (same for `maskPasswordsForScriptsObjectArray`) ?
```suggestion
export const maskPayloadsPassword = (value: Values): void => {
```

shir-reifenberg (commented):
Looks good! 🔥 
How is it going to affect deployments ?
and fyi you wrote "Jira" in the release notes instead of jamf

packages/jamf-adapter/src/definitions/fetch/transforms/utils.ts
kami-delaroz: I prefer not to, as I have it within xmls.
and I prefer not to change staff users might be regular with in Jira

packages/jamf-adapter/src/definitions/fetch/transforms/utils.ts
kami-delaroz: (I've changed in all the original functions in this page as well)

packages/jamf-adapter/test/definitions/fetch/transform/configuration_profile.test.ts
```diff
@@ -76,4 +76,22 @@ describe('adjust configuration profile', () => {
       })
     })
   })
+  describe('maskPayloadsPassword', () => {
+    it('should mask passwords within payloads', async () => {
+      const value = {
+        general: {
+          payloads:
+            '<?xml version="1.0" encoding="UTF-8"?><plist version="1"><dict><key>Password</key><string>1234</string></dict></plist>',
```
shir-reifenberg, line 84: can you please add a test that has 2 `Password` values in it to make sure we cover replacement of all values and not only the first ?

packages/jamf-adapter/src/definitions/fetch/transforms/utils.ts
```diff
@@ -79,3 +91,27 @@ export const removeSelfServiceSecurityPassword = (value: Record<string, unknown>
     }
   }
 }
+
+export const maskPayloadsPassword = (value: Values): void => {
+  const payloads = _.get(value, 'general.payloads')
+  if (typeof payloads === 'string') {
+    if (PASSWORD_REGEX.test(payloads)) {
+      log.trace(`Masked value of payloads in '${_.get(value, 'general.name')}'`)
+      _.set(value, 'general.payloads', payloads.replace(PASSWORD_REGEX, MASKED_PASSWORD))
```
shir-reifenberg, line 100: We're now testing the regex twice which isn't ideal for long strings.
I think you can provide `replace` with a callback, so you can log once you detect the regex. should be something like -
```suggestion
const maskedPayloads = payloads.replace(PASSWORD_REGEX, () => {
    log.trace(`Masked value of ...`)
    return MASKED_PASSWORD
  })
  _.set(value, 'general.payloads', maskedPayloads)
```

if it doesn't work probably best to just skip the log 😅

shir-reifenberg (approved):
Looks good! 🔥 added one comment about the masking value that I think we should think about, but approving to unblock

packages/jamf-adapter/test/definitions/fetch/transform/configuration_profile.test.ts
kami-delaroz: good idea

packages/jamf-adapter/src/definitions/fetch/transforms/utils.ts
kami-delaroz: yep, actually I had mixed feelings regarding this log initially, even after seeing we have such a log in Jira. But I added it as your request became a tie breaker 😅
Looks like it could work with _replace_ but not with _replaceAll_, and I don't want to postpone the rollout of this, will revisit it while working on the CV 🙏

PR #6907, opened 2024-12-08
SALTO-7030 SALTO-7035: (Bug-fix) Salesforce OrderedMaps type are incorrect after partial fetch

tamtamirr:
(Bug-fix) Salesforce OrderedMaps type are incorrect after partial fetch.

---

The issue occurred since we get the MetadataTypes from the Workspace on partial fetch. These types were already converted.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


amitgoldarb (approved):
Thanks for this

packages/salesforce-adapter/test/filters/convert_maps.test.ts
```diff
@@ -731,6 +732,13 @@ describe('Convert maps filter', () => {
       expect(fieldType.elemID.typeName).toEqual(`${ORDERED_MAP_PREFIX}CustomValue`)
     })
 
+    describe('when fetch is partial', () => {
```
amitgoldarb, line 735: I don't think fetch is actually partial in this example, just gave a very specific type no?

tamtamirr: This is partial in a sense where we pass a type that was already converted

PR #6909, opened 2024-12-08
SALTO-7037: add end_user_conditions and agent_conditions fields if they are missing from the form during deploy

edenhassid:
add end_user_conditions and agent_conditions fields if they are missing from the form during deploy

---

---
_Release Notes_: 

---
_User Notifications_: 


packages/zendesk-adapter/src/filters/ticket_form.ts
```diff
@@ -148,6 +148,8 @@ const getChangeWithoutRemovedFields = (change: ModificationChange<InstanceElemen
   }
   const clonedInst = after.clone()
   clonedInst.value.ticket_field_ids = (clonedInst.value.ticket_field_ids ?? []).concat(finalRemovedFields)
+  clonedInst.value.end_user_conditions = clonedInst.value.end_user_conditions ?? []
```
adfineberg, line 151: you can go with 
```suggestion
  clonedInst.value.end_user_conditions ??= []
```

adfineberg (approved):
👩‍🎤

PR #6913, opened 2024-12-08
SALTO-7039: Enforce MS Security OAuth flow to select an account

shoshana-houri:
Currently, the OAuth flow for connecting to MS Security fails if the user is logged into Microsoft with a different account than the one they wish to use for authentication.

This PR enforces the OAuth flow to prompt the user to select the account they want to authenticate with.

---

---
_Release Notes_: 
_Microsoft Security adapter:_
* Enforce MS Security OAuth flow to select an account

---
_User Notifications_: 
None


shir-reifenberg (approved):
🔥 🚒

PR #6915, opened 2024-12-08
SALTO-7040: Salesforce fetchTargets filter should not run in partial fetch

tamtamirr:
SALTO-7040: Salesforce fetchTargets filter should not run in partial fetch

---

_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


amitgoldarb (approved):
Thanks for this

PR #6916, opened 2024-12-08
SALTO-6934: Support configuring dummy adapter to return deploy failure

shoshana-houri:
Allow specifying a new dummy config param with the desired deployment result - failure or success.

---
_Release Notes_: 
_Dummy adapter:_
* Allow specifying `failDeploy: boolean` as a config param.

---
_User Notifications_: 
None

### DIFF
--- a/packages/core/src/core/deploy/deploy_summary.ts
+++ b/packages/core/src/core/deploy/deploy_summary.ts
@@ -26,7 +26,6 @@ import _ from 'lodash'
 
 const log = logger(module)
 
-// TODO SALTO-6934: Remove DeploySummaryResult after updating downstreams to import it from adapter-api.
 export type DeploySummaryResult = 'success' | 'failure' | 'partial-success'
 export type DetailedChangeId = string
 export type DetailedChangeDeploySummaryResult = [DetailedChangeId, DeploySummaryResult]
--- a/packages/core/src/core/deploy/deploy_summary.ts
+++ b/packages/core/src/core/deploy/deploy_summary.ts
@@ -26,6 +26,7 @@ import _ from 'lodash'
 
 const log = logger(module)
 
+// TODO SALTO-6934: Remove DeploySummaryResult after updating downstreams to import it from adapter-api.
 export type DeploySummaryResult = 'success' | 'failure' | 'partial-success'
 export type DetailedChangeId = string
 export type DetailedChangeDeploySummaryResult = [DetailedChangeId, DeploySummaryResult]
--- a/packages/dummy-adapter/src/adapter.ts
+++ b/packages/dummy-adapter/src/adapter.ts
@@ -5,6 +5,7 @@
  *
  * CERTAIN THIRD PARTY SOFTWARE MAY BE CONTAINED IN PORTIONS OF THE SOFTWARE. See NOTICE FILE AT https://github.com/salto-io/salto/blob/main/NOTICES
  */
+import _ from 'lodash'
 import {
   FetchResult,
   AdapterOperations,
@@ -16,8 +17,7 @@ import {
   isInstanceElement,
   FixElementsFunc,
 } from '@salto-io/adapter-api'
-import _ from 'lodash'
-import { generateElements, generateFetchErrorsFromConfig, GeneratorParams } from './generator'
+import { generateDeployResult, generateElements, generateFetchErrorsFromConfig, GeneratorParams } from './generator'
 import { changeValidator } from './change_validator'
 
 export default class DummyAdapter implements AdapterOperations {
@@ -42,10 +42,7 @@ export default class DummyAdapter implements AdapterOperations {
         instance.value = _.omit(instance.value, this.genParams.fieldsToOmitOnDeploy ?? [])
       })
 
-    return {
-      appliedChanges: changeGroup.changes,
-      errors: [],
-    }
+    return generateDeployResult(changeGroup.changes, this.genParams.deployResult ?? 'success')
   }
 
   // eslint-disable-next-line class-methods-use-this
@@ -58,6 +55,13 @@ export default class DummyAdapter implements AdapterOperations {
 
   public deployModifiers: DeployModifiers = {
     changeValidator: changeValidator(this.genParams),
+    getChangeGroupIds:
+      this.genParams.deployResult === 'partial-success'
+        ? // In order to return partial success, we need all the changes to reside in the same group
+          async changesMap => ({
+            changeGroupIdMap: new Map(Array.from(changesMap.entries()).map(([id]) => [id, 'dummy.ChangeGroup'])),
+          })
+        : undefined,
   }
 
   // eslint-disable-next-line class-methods-use-this
--- a/packages/dummy-adapter/src/adapter.ts
+++ b/packages/dummy-adapter/src/adapter.ts
@@ -42,7 +42,7 @@ export default class DummyAdapter implements AdapterOperations {
         instance.value = _.omit(instance.value, this.genParams.fieldsToOmitOnDeploy ?? [])
       })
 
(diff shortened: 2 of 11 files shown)


packages/core/src/core/deploy/deploy_summary.ts
```diff
@@ -26,6 +26,7 @@ import _ from 'lodash'
 
 const log = logger(module)
 
+// TODO SALTO-6934: Remove DeploySummaryResult after updating downstreams to import it from adapter-api.
```
shoshana-houri, line 29: I'm not sure if there are side effects to removing this import from `core` and replacing it with one from `adapter-api` in two iterations.
If there are any, or if you think this shouldn't be located under `adapter-api`, please let me know.

I felt it was better for this result to be typed appropriately, and we cannot import from `core` to `dummy`. However, if this introduces issues, we can manage without it being typed.

packages/dummy-adapter/src/generator.ts
```diff
@@ -1148,3 +1156,77 @@ export const generateFetchErrorsFromConfig = (
         ...error,
         elemID: ElemID.fromFullName(error.elemID),
       }))
+
+const generateChangeError = (change: Change<ChangeDataType>): SaltoElementError => ({
+  elemID: getChangeData(change).elemID,
+  severity: 'Error',
+  message: 'Failed to deploy',
```
Nurdok, line 1163: Suggestion: maybe add a hint into why this was generated as part of the message. 
"Failed to deploy - this was intentional, see deplyResult adapter config flag"
(but probably with better phrasing...)

packages/dummy-adapter/src/generator.ts
```diff
@@ -1148,3 +1156,77 @@ export const generateFetchErrorsFromConfig = (
...
+
+  if (hasRemovalChange) {
+    return {
+      appliedChanges: changesWithPartialSuccess,
+      errors: [],
+    }
+  }
+
+  if (changes.length < 2) {
+    log.error('Expected at removal change or at least 2 changes for partial success. Falling back to error result')
```
Nurdok, line 1203: at -> a

packages/dummy-adapter/src/generator.ts
```diff
@@ -1148,3 +1156,77 @@ export const generateFetchErrorsFromConfig = (
...
+        errors: [],
+      }
+    case 'failure':
+      return {
+        appliedChanges: [],
+        errors: changes.map(generateChangeError),
+      }
+    case 'partial-success':
+      return generatePartialSuccessDeployResult(changes)
+    default:
```
Nurdok, line 1229: Is this needed if `dummyResult` is typed?

packages/dummy-adapter/test/generator.test.ts
```diff
@@ -356,4 +373,191 @@ describe('generator', () => {
...
+      instanceModificationChange,
+      objectTypeRemovalChange,
+      objectTypeAdditionChange,
+      objectTypeModificationChange,
+    ]
+
+    describe('when deploy result is success', () => {
+      it('should return a deploy result with the applied changes and no errors', () => {
+        const deployResult = generateDeployResult(changes, 'success')
+        expect(deployResult.appliedChanges).toHaveLength(changes.length)
```
Nurdok, line 422: This `expect` is unnecessary, and it will also make this test harder to debug if it fails - you'll get something like "length was 5, expected 7" instead of the actual diff. Please remove.

Ditto for tests below.

packages/dummy-adapter/test/generator.test.ts
```diff
@@ -356,4 +373,191 @@ describe('generator', () => {
...
+        expect(deployResult.appliedChanges).toEqual(changes)
+        expect(deployResult.errors).toHaveLength(0)
```
Nurdok, line 424: For the same reason, I'd use `toEqual([])`, so that you'll actually see the error in the test results.
I'd also put this above the previous line - if there's a failure, most likely both `expect`s will fail, and I think seeing the error will be more helpful for debugging. Another option is to put these as separate tests inside a wrapping `describe`.

Nurdok (commented):
LGTM, not approving because I'm not sure about the `core` to `adapter-api` change, and I think someone more knowledgable should see this too :)

packages/dummy-adapter/src/generator.ts
shoshana-houri: We have some lint rules configured in this repo which prevents us from removing it -
When I remove the "default" I get 
> Expected a default case.eslint[default-case](https://eslint.org/docs/latest/rules/default-case)

And when I ignore this lint rule I get
> Expected to return a value at the end of arrow function.eslint[consistent-return](https://eslint.org/docs/latest/rules/consistent-return)

When I ignore both of them we do receive the desired result of a build error when some type is missing, so I guess that I can do that...?

packages/dummy-adapter/test/generator.test.ts
shoshana-houri: Got it. I actually thought that it's somewhat easier to first see changes in the number of elements and only then dive into details if necessary, but I don't mind changing that :)

packages/dummy-adapter/src/adapter.ts
```diff
@@ -42,10 +42,7 @@ export default class DummyAdapter implements AdapterOperations {
         instance.value = _.omit(instance.value, this.genParams.fieldsToOmitOnDeploy ?? [])
       })
 
-    return {
-      appliedChanges: changeGroup.changes,
-      errors: [],
-    }
+    return generateDeployResult(changeGroup.changes, Boolean(this.genParams.failDeploy))
```
ori-moisis, line 45: 1. this seems overly complicated, if we want to return no applied changes, we don't need to run the code above, we just need to return an empty array
2. Returning errors should be separate, I don't think we should reduce it to a boolean, the config should be able to control which error returns like it is in other places that inject errors (e.g - in the dummy change validators)

shoshana-houri: 1. 100%, missed that in the 2nd iteration, thanks!!
2. Got it. So do you think that we should return an empty array both for `appliedChanges` and for `errors`? The behavior in this case is a bit weird IMO since all changes appear to be successful, while the deploy result is failure. Perhaps returning a general `SaltoError` instead of `SaltoElementError` would be better?

Returning an empty array  for `appliedChanges` and for `errors`:
image

Returning a general `SaltoError`:
image

ori-moisis: they appear to be successful because of the bug, no?

shoshana-houri: AFAIU no - we only update the `planItem` [with the applied changes](https://github.com/salto-io/salto/blob/fad42b019fdd1370af9c3c689d37862da1cf6d7b/packages/core/src/core/deploy/deploy_actions.ts#L107-L113), and since there aren't any - the plan item remains the same - which means that the behavior in the CLI has not effect on it.
These are [the lines](https://github.com/salto-io/salto/blob/fad42b019fdd1370af9c3c689d37862da1cf6d7b/packages/core/src/core/deploy/deploy_actions.ts#L168-L180) where we determine what to print - if no errors are returned then the deployment of this change is printed as success.

ori-moisis: that is just the progress report
I am not sure it is the most important thing (or that it is even a problem), but even if you think there is a problem there, it is a CLI problem, the thing you say is wrong there is the UX...
it should not determine the implementation of the dummy adapter.
on the contrary, if the dummy adapter is not flexible, it would be impossible to expose such UX issues

shoshana-houri: Got it. Returning no applied changes and no errors makes sense to me then, I'll update the PR. Thanks!

ori-moisis (approved):

PR #6918, opened 2024-12-09
SALTO-7043: Fix VS Code Jest tests debug setup

shir-reifenberg:
It was looking for Jest's binary in the wrong place 

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-api/.vscode/launch.json
+++ b/packages/adapter-api/.vscode/launch.json
@@ -9,7 +9,12 @@
       "request": "launch",
       "name": "adapter-api Debug Jest Tests",
       "cwd": "${workspaceFolder}",
-      "args": ["${workspaceRoot}/node_modules/.bin/jest", "--runInBand", "--config", "${workspaceRoot}/jest.config.js"],
+      "args": [
+        "${workspaceFolder}/../../node_modules/.bin/jest",
+        "--runInBand",
+        "--config",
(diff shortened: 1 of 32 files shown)


packages/adapter-api/.vscode/launch.json
```diff
@@ -9,7 +9,7 @@
       "request": "launch",
       "name": "adapter-api Debug Jest Tests",
       "cwd": "${workspaceFolder}",
-      "args": ["${workspaceRoot}/node_modules/.bin/jest", "--runInBand", "--config", "${workspaceRoot}/jest.config.js"],
+      "args": ["${workspaceRoot}/../../node_modules/.bin/jest", "--runInBand", "--config", "${workspaceRoot}/jest.config.js"],
```
Nurdok, line 12: Looks like the `workspaceRoot` variable name is deprecated in favor of `workspaceFolder`.
Not really important, but might save us some trouble in the future by switching now.

https://code.visualstudio.com/docs/editor/variables-reference#_why-isnt-workspaceroot-documented

Nurdok (approved):

packages/adapter-api/.vscode/launch.json
shir-reifenberg: Thanks, will replace that

PR #6919, opened 2024-12-09
Add s3 logs for error visibility

adfineberg:
Related to an issue where S3 returned an empty content.

---

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/local-workspace/src/s3_dir_store.ts
+++ b/packages/local-workspace/src/s3_dir_store.ts
@@ -9,7 +9,7 @@ import path from 'path'
 import { logger } from '@salto-io/logging'
 import * as AWS from '@aws-sdk/client-s3'
 import { createS3Client } from '@salto-io/aws-utils'
-import { safeJsonStringify } from '@salto-io/adapter-utils'
+import { inspectValue, safeJsonStringify } from '@salto-io/adapter-utils'
 import { dirStore, staticFiles } from '@salto-io/workspace'
 import Bottleneck from 'bottleneck'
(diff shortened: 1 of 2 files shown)


dantal4 (approved):
thanks!

packages/local-workspace/src/s3_dir_store.ts
```diff
@@ -55,10 +55,18 @@ export const buildS3DirectoryStore = ({
       }
 
       const buffer = await getStream.buffer(s3Obj.Body)
-
+      if (buffer === undefined) {
+        log.warn(
+          'Failed to read file %s from S3 bucket %s with body %s',
+          fullFilePath,
+          bucketName,
+          safeJsonStringify(s3Obj.Body),
```
ori-moisis, line 63: this could lead to a very very very very large log
use `inspectValue` instead, unless it is important that you get all of the content

ori-moisis: actually, what are you expecting to see here, isn't this a stream object?

adfineberg: It should be empty given that the buffer is undefined, but no need to risk it

ori-moisis: no need to risk it, and yet you merged it with `safeJsonStringify`

adfineberg: No, I changed it!

PR #6923, opened 2024-12-10
SALTO-7047: Improve performance of workspace update

ori-moisis:
The override mechanism is replaced by fixing the original element source to return the correct result The issue that the override mechanism was there to solve was that if a field was removed from NaCl, we should remove the hidden part of the field as well.

This didn't work in the original code because the state element source would try to merge it with the element from the workspace cache but the element in the workspace cache was not updated yet, so it still had that field.

Changing the element source to use the element from the nacl cache instead (which is updated before) should solve the issue

---

It seems like the original issue that the overrides code solves was never actually covered in unit tests, I did verify it manually and added the missing unit test in this PR.

The scenario is to have a field with a hidden annotation in an environment, and then delete that field from the nacl (by editing the nacl manually).
before this fix, without the override code, the field would not be fully deleted from the workspace (the hidden part would remain)

---
_Release Notes_: 
Core:
- Improve performance of updating workspace cache

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/workspace/src/workspace/workspace.ts
+++ b/packages/workspace/src/workspace/workspace.ts
@@ -30,7 +30,6 @@ import {
   isModificationChange,
   TypeReference,
   DEFAULT_SOURCE_SCOPE,
-  isElement,
   isAdditionOrModificationChange,
 } from '@salto-io/adapter-api'
 import { logger } from '@salto-io/logging'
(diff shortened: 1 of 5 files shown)


packages/workspace/src/workspace/workspace.ts
```diff
@@ -632,9 +631,10 @@ export const loadWorkspace = async (
         )
         const sources: Record<string, ReadOnlyElementsSource> = {}
         await awu(envs()).forEach(async envName => {
-          sources[MULTI_ENV_SOURCE_PREFIX + envName] = await naclFilesSource.getElementsSource(envName)
+          const naclSource = await naclFilesSource.getElementsSource(envName)
+          sources[MULTI_ENV_SOURCE_PREFIX + envName] = naclSource
           sources[STATE_SOURCE_PREFIX + envName] = mapReadOnlyElementsSource(state(envName), async element =>
-            getElementHiddenParts(element, state(envName), await states[envName].merged.get(element.elemID)),
```
ori-moisis, line 637: the fix is just here, using the multi-env source means we are using the source that is already updated before us, which means if anything is removed, by the time we get to updating our cache, it will already by updated in the multi-env.
before this code would use the workspace cache (which is the thing that is being updated), so it would have elements that are not updated yet...

packages/workspace/src/workspace/workspace.ts
```diff
@@ -752,23 +752,6 @@ export const loadWorkspace = async (
           }
         }
 
-        const workspaceChangedElements = Object.fromEntries(
```
ori-moisis, line 755: removing this section is the main motivation from this change - on top of it being a significant time consuming calculation, the fact that it would hold all the results in memory created a local memory peak that was just unnecessary

Nurdok (commented):
LGTM, though I admit I'm struggling to understand the high-level flow where `loadWorkspace` is used. Waiting to see the additional unit test before approving.

Nurdok (approved):

PR #6929, opened 2024-12-10
SALTO-7049: add pariallyFetchedAccounts to FetchResult

shayc331:
_add pariallyFetchedAccounts to FetchResult_

---

The motivation is to differ between fetch operation types in the backend (full, custom or changed based).
partiallyFetchAccounts is filled with accounts that do any kind of a partial fetch, and we want to expose it to the SAAS.

---
_Release Notes_: 
_Core_:
* `FetchResult` will contain `partiallyFetchedAccounts` to differentiate between partial and full fetches.

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/cli/test/commands/adapter_format.test.ts
+++ b/packages/cli/test/commands/adapter_format.test.ts
@@ -73,6 +73,7 @@ describe('apply-patch command', () => {
         fetchErrors: [],
         success: true,
         updatedConfig: {},
+        partiallyFetchedAccounts: new Set(['salesforce']),
       })
 
       exitCode = await applyPatchAction({
@@ -114,6 +115,7 @@ describe('apply-patch command', () => {
         fetchErrors: [],
         success: true,
         updatedConfig: {},
+        partiallyFetchedAccounts: new Set(['salesforce']),
       })
       exitCode = await applyPatchAction({
         ...cliCommandArgs,
@@ -167,6 +169,7 @@ describe('apply-patch command', () => {
         fetchErrors: [],
         success: true,
         updatedConfig: {},
+        partiallyFetchedAccounts: new Set(['salesforce']),
       })
       exitCode = await applyPatchAction({
         ...cliCommandArgs,
@@ -207,6 +210,7 @@ describe('apply-patch command', () => {
         fetchErrors: [],
         success: true,
         updatedConfig: {},
+        partiallyFetchedAccounts: new Set(['salesforce']),
       })
       exitCode = await applyPatchAction({
         ...cliCommandArgs,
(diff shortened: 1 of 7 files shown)


packages/core/src/core/adapter_format.ts
```diff
@@ -224,6 +224,7 @@ export const calculatePatch = async ({
       fetchErrors: [],
       success: false,
       updatedConfig: {},
+      partiallyFetchedAccounts: new Set([accountName]),
```
dantal4, line 227: nit - set `const partiallyFetchedAccounts = new Set([accountName])` and use it in the three returns in this function.

packages/cli/test/commands/adapter_format.test.ts
```diff
@@ -167,6 +169,7 @@ describe('apply-patch command', () => {
         fetchErrors: [],
         success: true,
         updatedConfig: {},
+        partiallyFetchedAccounts: new Set(['salesforce']),
```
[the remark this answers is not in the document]
dantal4: your impl is good IMO

dantal4 (commented):
Looks good! please add tests for it in `core/fetch.test.ts` & `core/adapter_format.test.ts`

dantal4 (approved):
💯

packages/core/src/core/adapter_format.ts
```diff
@@ -217,13 +217,15 @@ export const calculatePatch = async ({
     mergeErrors: beforeMergeErrors,
     mergedElements: mergedBeforeElements,
   } = await loadElementsAndMerge(fromDir, loadElementsFromFolder, adapterContext)
+  const partiallyFetchedAccounts = new Set([accountName])
```
ori-moisis, line 220: why is a patch always marked as a partial fetch?
not sure I understand the logic behind this decision...

packages/core/src/types.ts
```diff
@@ -37,6 +37,7 @@ export type FetchResult = {
   configChanges?: Plan
   updatedConfig: Record<string, InstanceElement[]>
   accountNameToConfigMessage?: Record<string, string>
+  partiallyFetchedAccounts: Set<string>
```
ori-moisis, line 40: is it really ok for this to be required?
can't it break anything? (reminder - this is an external core API)

dantal4: only possible on tests, or if somewhere a `FetchResult` object is created but not by calling one of the core fetch APIs.

packages/core/src/core/adapter_format.ts
dantal4: this was based on the fact that we pass `calcFetchChanges` a similar information ([here](https://github.com/salto-io/salto/pull/6929/files/5204036ffe2db3c1d45dc6fbec0e2173563e10d9#diff-f98013d1f83b5156009ad134bacfb88963cc400e45b00da03566f12ea90ae489R252-R253)).
It kind of makes sense, because we're only "fetching" the elements in the from&to directories, that in most case don't hold all of the elements in the workspace.

ori-moisis: ok

PR #6931, opened 2024-12-10
SALTO-7052 Add a CV for invalid target types

adfineberg:
As per [here](https://support.zendesk.com/hc/en-us/articles/6468124845210-Announcing-the-deprecation-of-URL-targets-and-branded-targets), we are blocking invalid target types from being deployed.

---

---
_Release Notes_: 

---
_User Notifications_: 


edenhassid (approved):
LGTM!

PR #6932, opened 2024-12-11
SALTO-7054: NS: Avoid having RegexMismatchValidationError warning for apparently valid scriptIds

omrilit:
The code that was added to `types_generator.py` wasn't really tested, but only added so it won't get overridden next time we will generate types.

I used https://app.datadoghq.com/logs?query=status%3Awarn%20RegexMismatchValidationError%20-%2Areportdefinition%2A%20-%2Acustomsearch%2A%20-%2Acustomrole%2A&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1733300418316&to_ts=1733905218316&live=true in order to find which elements suffer from the issue.

---

---
_Release Notes_: 

---
_User Notifications_: 


dantal4 (approved):
Thanks!

PR #6933, opened 2024-12-11
SALTO-4062: Support Custom Roles in Okta

shir-reifenberg:
Add a proper support in Okta's custom roles

---

Until now, we only fetched the name and description of the role, with no deploy support.
This PR adds the permission field for roles + support deploying those
Standard roles are not deployable and blocked using a validator.

This PR rebased on top of https://github.com/salto-io/salto/pull/6933, only the last commit is relevant

---
_Release Notes_: 

_Okta Adapter_:
- Add deploy support for Custom Roles

---
_User Notifications_: 
_Okta Adapter_:
- Custom roles elements will now include their permissions

### DIFF
--- a/packages/okta-adapter/src/definitions/deploy/deploy.ts
+++ b/packages/okta-adapter/src/definitions/deploy/deploy.ts
@@ -17,7 +17,7 @@ import {
   isRemovalChange,
   Values,
 } from '@salto-io/adapter-api'
-import { getParents, validatePlainObject } from '@salto-io/adapter-utils'
+import { getParents, naclCase, validatePlainObject } from '@salto-io/adapter-utils'
 import { AdditionalAction, ClientOptions } from '../types'
 import {
   APPLICATION_TYPE_NAME,
@@ -47,6 +47,7 @@ import {
   ERROR_PAGE_TYPE_NAME,
   AUTHORIZATION_SERVER,
   GROUP_RULE_TYPE_NAME,
+  ROLE_TYPE_NAME,
 } from '../../constants'
 import {
   APP_POLICIES,
@@ -60,6 +61,7 @@ import { isCustomApp } from '../fetch/types/application'
 import { addBrandIdToRequest } from './types/email_domain'
 import { isSystemScope } from './types/authorization_servers'
 import { isActiveGroupRuleChange } from './types/group_rules'
+import { adjustRoleAdditionChange, isPermissionChangeOfAddedRole } from './types/roles'
 
 const log = logger(module)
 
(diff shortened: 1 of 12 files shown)


packages/okta-adapter/src/definitions/deploy/types/roles.ts
```diff
@@ -0,0 +1,63 @@
...
+  const { sharedContext } = context
+  validatePlainObject(value, ROLE_TYPE_NAME)
+  const permissions = _.get(value, 'permissions')
+  if (!_.isArray(permissions)) {
+    log.error('expected permissions to be an array, instead got %s', safeJsonStringify(permissions))
+    throw new Error('missing permissions for role')
+  }
+  const mappedPermissions = permissions.map(permission => {
+    if (_.isPlainObject(permission?.conditions)) {
+      // naclCase permission labal because the label includes dots
```
shir-reifenberg, line 42: fix type `labal`

packages/okta-adapter/src/definitions/deploy/types/roles.ts
```diff
@@ -0,0 +1,63 @@
...
+  validatePlainObject(value, ROLE_TYPE_NAME)
+  const permissions = _.get(value, 'permissions')
+  if (!_.isArray(permissions)) {
+    log.error('expected permissions to be an array, instead got %s', safeJsonStringify(permissions))
+    throw new Error('missing permissions for role')
+  }
+  const mappedPermissions = permissions.map(permission => {
+    if (_.isPlainObject(permission?.conditions)) {
+      // naclCase permission.label because the label includes dots
+      _.set(sharedContext, naclCase(permission.label), true)
```
Nurdok, line 43: I _think_ this is wrong because the shared context is the same for all request, and there's no keying by the role ID here, so when you look for a permission later, you'll see `true` on permission from _all_ roles. Did you try deploying two distinct roles with different permissions in the same deployment?

I think you should either:
1. Key permission labels under the role ID here, and when reading in the permission request.
2. In this function, only flatten the permission array and don't save to the shared context. Rename to, e.g., `flattenPermissionArray`. Then use `copyFromResponse.toSharedContext`, set it to `pick: 'permissions'` and set `nestUnderElemID` to `true` (this will only work if the response contains the permissions, which IIUC it does). It'll be a bit clearer to read directly from the definitions.
3. Only flatten here and rename like above, but don't use the shared context at all, rather use the element source in the permission request to figure out which permissions have conditions.

I think the last option is best because using the shared context between requests is harder to read and understand - it requires cooperation between two seemingly unrelated functions and is more prone to errors if you're not careful about using the same keying logic, etc.

packages/okta-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -1316,6 +1318,95 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
         },
       },
     },
+    [ROLE_TYPE_NAME]: {
```
Nurdok, line 1321: Can you please add an HTTP-recording test for the new supported actions (the other `adapter.test.ts`)?

Nurdok (changes requested):

packages/okta-adapter/src/definitions/deploy/types/roles.ts
shir-reifenberg: Good catch! thank you
I think I didn't encountered this as each element is on a different change group, and we didn't fixed sharedContext to share info between groups..
I'll see what's goes best with the deployment order and will make the adjustments 🤞 (though I kind of prefer avoiding the element source as I have the parent change in hand..)

packages/okta-adapter/src/definitions/fetch/fetch.ts
```diff
@@ -371,6 +371,16 @@ const createCustomizations = ({
             },
           },
         },
+        AppProvisioning: {
```
Nurdok: Remove :)

Nurdok (approved):

PR #6939, opened 2024-12-12
SALTO-7058: Fix CLI deploy summary

ori-moisis:
The deploy summary in the CLI was incorrect when changes were not fully applied

---

---
_Release Notes_: 
CLI:
- Fixed issue where deploy summary would not show failed or partially successful changes correctly

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/cli/src/commands/deploy.ts
+++ b/packages/cli/src/commands/deploy.ts
@@ -7,6 +7,7 @@
  */
 import _ from 'lodash'
 import { collections, promises } from '@salto-io/lowerdash'
+import { applyFunctionToChangeDataSync } from '@salto-io/adapter-utils'
 import {
   PlanItem,
   Plan,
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/deploy.ts
Nurdok, line 1: It looks like there's no unit test that captures the bug that this PR is fixing - there are unit tests for `summartizeDeployChanges`, and unit tests for the CLI command (this file) that mock `summartizeDeployChanges` and test the output prints for the CLI. We need a test that makes sure `summartizeDeployChanges` actually gets called with the correct input. Can you modify an existing test for this, or add a new one?

ori-moisis: I am not sure this is a unit test though...
the bug was in the integration - because the real `deploy` implementation has a side effect.
so, if we wanted to have a unit test for this, it would have to create a mock of the side effect that running a real `deploy` has.

Nurdok (approved):

ori-moisis:
tried figuring out if I could add a unit test but it seems like the example plan we are working with in the current unit tests would not be easy to change or utilize to have a side effect.

given that this is a somewhat "edge" issue (only affects the CLI, only prints information and cannot cause actual corruption of data), I prefer merging the fix...
not setting a great example here :(

PR #6940, opened 2024-12-12
SALTO-7062: Deployment of filters with user share permissions

IdoZakk:
The deployments failed, which is very weird

---

None

---
_Release Notes_: 
Jira Adapter:
* Fixed a bug in deployment of Filters with user share permissions

---
_User Notifications_: 
None

### DIFF
--- a/packages/jira-adapter/src/filters/filter.ts
+++ b/packages/jira-adapter/src/filters/filter.ts
@@ -14,49 +14,43 @@ import {
   Value,
 } from '@salto-io/adapter-api'
 import { FilterCreator } from '../filter'
+import { SHARE_PERMISSION_FIELDS } from '../constants'
+
+type ChangeFormatFunction = (permission: Value) => void
+
+const toDeploymentFormat = (permission: Value): void => {
+  permission.user = {
+    accountId: permission.user,
+  }
+}
+
+const toNaclFormat = (permission: Value): void => {
+  permission.user = permission.user.accountId
+}
+
+const changePermissionFieldFormat = (changes: Change<ChangeDataType>[], changeFunction: ChangeFormatFunction): void => {
+  changes
+    .filter(isInstanceChange)
+    .filter(isAdditionOrModificationChange)
+    .map(getChangeData)
+    .filter(instance => instance.elemID.typeName === 'Filter')
+    .forEach(instance => {
+      SHARE_PERMISSION_FIELDS.forEach(fieldName => {
+        if (Array.isArray(instance.value[fieldName])) {
+          instance.value[fieldName].filter((permission: Value) => permission.type === 'user').forEach(changeFunction)
+        }
+      })
+    })
+}
 
 const filter: FilterCreator = () => ({
   name: 'filtersFilter',
   preDeploy: async (changes: Change<ChangeDataType>[]) => {
-    changes
-      .filter(isInstanceChange)
-      .filter(isAdditionOrModificationChange)
-      .map(change => getChangeData(change))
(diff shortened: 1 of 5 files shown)


packages/jira-adapter/src/filters/filter.ts
```diff
@@ -14,49 +14,43 @@ import {
...
+const changePermissionFieldFormat = (changes: Change<ChangeDataType>[], changeFunction: ChangeFormatFunction): void => {
+  changes
+    .filter(isInstanceChange)
+    .filter(isAdditionOrModificationChange)
+    .map(getChangeData)
+    .filter(instance => instance.elemID.typeName === 'Filter')
+    .forEach(instance => {
+      SHARE_PERMISSION_FIELDS.forEach(fieldName => {
+        if (Array.isArray(instance.value[fieldName])) {
+          instance.value[fieldName].filter((permission: Value) => permission.type === 'user').forEach(changeFunction)
```
shayc331, line 40: I think permission might be `undefined`, we can add a simple type guard and use a type instead of `Value` in the rest of the functions

shayc331 (commented):
looking good, had some comments
did you check it using DC?

packages/jira-adapter/src/filters/filter.ts
```diff
@@ -14,49 +14,43 @@ import {
   Value,
 } from '@salto-io/adapter-api'
 import { FilterCreator } from '../filter'
+import { SHARE_PERMISSION_FIELDS } from '../constants'
+
+type ChangeFormatFunction = (permission: Value) => void
+
+const toDeploymentFormat = (permission: Value): void => {
+  permission.user = {
+    accountId: permission.user,
```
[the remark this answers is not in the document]
IdoZakk: We change it to reflect also the actual name of the user not just the accountId

packages/jira-adapter/src/filters/filter.ts
IdoZakk: type guard is a good idea

packages/jira-adapter/src/filters/filter.ts
```diff
@@ -13,50 +13,61 @@ import {
...
+import Joi from 'joi'
 import { FilterCreator } from '../filter'
+import { FILTER_TYPE_NAME, SHARE_PERMISSION_FIELDS } from '../constants'
 
+const { makeArray } = collections.array
+const USER = 'user'
+
+type Permission = {
+  type: string
+  user: Value
```
shayc331, line 27: I think we should add a type for the user as well:
``` 
type User {
  accountId: string
}
type Permission = {
  type: string
  user: User
}

File: packages/jira-adapter/src/filters/filter.ts [hunk 4]
```diff
@@ -13,50 +13,61 @@ import {
...
+}
+
+const USER_PERMISSION_SCHEME = Joi.object({
+  type: Joi.string().valid(USER).required(),
+  user: Joi.any().required(),
```
shayc331, line 32: after we add a type for the user we can improve this type guard

shayc331 (commented):
looking great!
had one comment

packages/jira-adapter/src/filters/filter.ts
IdoZakk: The user can also be a string. It makes things complicated. For instance
image

The red line is because permission.user can also be a User type, so it does not fit a string

shayc331: I see, lets move on it is not worth it

shayc331 (approved):

PR #6942, opened 2024-12-15
SALTO-6992: Skip calculating workspace-service changes if there are no pending changes

Nurdok:
When computing "workspace <--> service" diff in fetch, limit the comparison only to elements that have pending changes - all other incoming changes from the service will have similar diffs to already-computed "service <--> state" diff, so we can return these diffs instead.

There's a caveat though - "service <--> state" diffs have resolved references that are based on the _state_ source instead of the workspace. To solve this, this PR also refactors the way we compute the diffs - this was done with `getPlan` which resolves values, but this isn't needed for fetch. The refactor extracts the basic code to compute a list of changes for two element sources, and now fetch uses this function directly instead of `getPlan`.

---

Memory / runtime comparison:
newplot

---
_Release Notes_: 
- Improved runtime and memory consumption performance for fetch operations

---
_User Notifications_: 
None


packages/core/src/core/fetch.ts
```diff
@@ -358,21 +364,61 @@ const getChangesNestedUnderID = (
...
+    }
+    if (isStaticFile(value)) {
+      return value
+    }
+    if (_.isArray(value)) {
+      return value.map(item => unresolveReferencesSingleValue({ value: item }))
+    }
+    if (_.isObject(value)) {
+      return _.mapValues(value, item => unresolveReferencesSingleValue({ value: item }))
+    }
```
ori-moisis, line 391: are you sure these are needed?
if they are, it probably means we have a significant bug in transform values

packages/core/src/core/fetch.ts
```diff
@@ -358,21 +364,61 @@ const getChangesNestedUnderID = (
...
+      const relatedChanges = getChangesNestedUnderID(elemId, serviceAndPendingChanges)
+      const [serviceChanges, pendingChanges] = _.partition(relatedChanges, change => change.origin === 'service').map(
+        changeList => changeList.map(change => change.change),
+      )
       const wsChanges = getChangesNestedUnderID(elemId, workspaceToServiceChanges).map(({ change }) => change)
+      if (pendingChanges.length === 0 && wsChanges.length === 0) {
+        const unresolvedServiceChanges = await awu(serviceChanges).map(unresolveReferences).toArray()
+        return unresolvedServiceChanges.map(
+          (change): FetchChange => ({ change, serviceChanges: unresolvedServiceChanges, pendingChanges: [] }),
+        )
```
ori-moisis, line 419: are you sure we need to return the cloned service changes? we can't use the original ones here? (is the difference in memory consumption really significant?)

packages/core/src/core/fetch.ts
```diff
@@ -358,21 +364,61 @@ const getChangesNestedUnderID = (
...
     .toArray()
 
-const toFetchChanges = (
+const toFetchChanges = async (
   serviceAndPendingChanges: collections.treeMap.TreeMap<WorkspaceDetailedChange>,
   workspaceToServiceChanges: collections.treeMap.TreeMap<WorkspaceDetailedChange>,
-): Iterable<FetchChange> => {
+): Promise<Iterable<FetchChange>> => {
+  const unresolveReferencesSingleValue: TransformFunc = ({ value }): Value => {
+    if (isElement(value)) {
```
ori-moisis, line 372: is there a case where this is actually possible?
it really shouldn't be

packages/core/src/core/plan/diff.ts
```diff
@@ -52,3 +75,286 @@ export const getDiffInstance = (change: Change<InstanceElement>): InstanceElemen
...
+        }
+      }
+
+      const addElementsNodes = async (comparison: BeforeAfter<ChangeDataType>): Promise<void> => {
+        const beforeElement = comparison.before
+        const afterElement = comparison.after
+        if (!isVariable(beforeElement) && !isVariable(afterElement)) {
+          await addChangeIfDifferent(beforeElement, afterElement)
+        }
+        if (
```
Nurdok, line 304: This `if` is the only real difference between this and the original implementation from `plan.ts` - this code would emit separate changes for each field of an added / deleted type. In `getPlan`, this is then fixed by [`removeRedundantFieldNodes`](https://github.com/salto-io/salto/blob/1710e7d748037b2edfa6111aa1c7225dfcd9b42a/packages/core/src/core/plan/group.ts#L73). For deploy, these "redundant" nodes are required, because a specific field may have its own dependencies on other nodes, which would require the parent type and that field to be deployed separately. 

For fetch, this isn't an issue, and not adding the redundant changes here is much easier than fixing the list of the changes later, so I added a flag to remove them, which will only be used in the fetch code path.

packages/core/src/core/fetch.ts
```diff
@@ -372,7 +373,31 @@ const toFetchChanges = (
...
+      const wsChangeIds = new Set(wsChanges.map(change => change.id.getFullName()))
+      // Service-to-workspace change were only computed for elements that have pending changes, because no pending
+      // changes means that service-to-workspace changes are identical to service-to-state changes. Since this
+      // function's output should include all changes, we need to copy the service-to-workspace changes that were not
+      // computed before. We are copying references to the changes, so this is still memory efficient.
+      if (pendingChanges.length === 0) {
+        serviceChanges
+          .filter(change => !wsChangeIds.has(change.id.getFullName()))
+          .forEach(change => {
+            if (!isAdditionChange(change) && isReferenceExpression(change.data.before)) {
```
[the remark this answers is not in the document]
Nurdok: Done by not resolving references in the first place.

packages/core/src/core/fetch.ts
Nurdok: Removed

packages/core/src/core/fetch.ts
Nurdok: Removed

packages/core/src/core/fetch.ts
Nurdok: Removed

packages/core/src/core/plan/diff.ts
ori-moisis: can you have a version of the code where this isn't refactored in a way that makes it hard to see the differences?
not saying the refactor is bad, just that it makes it much harder to review...

packages/core/src/core/fetch.ts
```diff
@@ -132,19 +134,11 @@ export const getDetailedChanges = async (
...
-        topLevelFilters,
-      })
-    ).itemsByEvalOrder(),
-  )
-    .map(item => item.detailedChanges())
-    .flatten()
+): Promise<DetailedChangeWithBaseChange[]> => {
+  const numElements = (await awu(await before.list()).length()) + (await awu(await after.list()).length())
+  const changes = await calculateDiff(before, after, topLevelFilters, numElements, undefined, true)
+  return changes.flatMap(change => getDetailedChangesFromChange(change))
```
ori-moisis, line 140: 1. with functions that get a lot of arguments, especially booleans and optionals, we should prefer named arguments over positional
2. I think we should have this return an async iterable instead of having all of these in memory at the same time (this might be minor ATM because the next step will pull all of it into memory anyway, but I think it is still worth pushing that part to be as late as possible)

packages/core/src/core/fetch.ts
```diff
@@ -132,19 +134,11 @@ export const getDetailedChanges = async (
...
-        after,
-        dependencyChangers: [],
-        topLevelFilters,
-      })
-    ).itemsByEvalOrder(),
-  )
-    .map(item => item.detailedChanges())
-    .flatten()
+): Promise<DetailedChangeWithBaseChange[]> => {
+  const numElements = (await awu(await before.list()).length()) + (await awu(await after.list()).length())
```
ori-moisis, line 138: I think we should not do this
IIUC, it is only used for a log, and that log is incorrect anyway if this is the calculation (given that it does not take the filters into account)

packages/core/src/core/fetch.ts
Nurdok: 1. Done
2. Done. This affected the order of changes slightly, I modified some tests to not check for the exact order. AFAICT this is still correct, but please look it over to see that it makes sense.

packages/core/src/core/plan/diff.ts
Nurdok: I've squashed and split this PR into commits that should make it easier for reviewing. The main issue is that GitHub doesn't track code moves into different files (what year is it??), so I did the move in a commit of its own without any logic changes. The commit descriptions are hopefully, well, descriptive.

packages/core/src/core/fetch.ts
```diff
@@ -358,21 +354,30 @@ const getChangesNestedUnderID = (
...
       }
 
       const elemId = ElemID.fromFullName(id)
+
+      const relatedChanges = getChangesNestedUnderID(elemId, serviceAndPendingChanges)
+      const [serviceChanges, pendingChanges] = _.partition(relatedChanges, change => change.origin === 'service').map(
+        changeList => changeList.map(change => change.change),
+      )
       const wsChanges = getChangesNestedUnderID(elemId, workspaceToServiceChanges).map(({ change }) => change)
+      if (pendingChanges.length === 0 && wsChanges.length === 0) {
```
ori-moisis, line 377: not sure I understand why we need to search in the `workspaceToServiceChanges` here
isn't the condition of `pendingChanges.length === 0` enough?

also, I think this deserves a comment explaining why what we are doing is correct

packages/core/src/core/fetch.ts
```diff
@@ -358,21 +354,30 @@ const getChangesNestedUnderID = (
...
+        return serviceChanges.map((change): FetchChange => ({ change, serviceChanges, pendingChanges: [] }))
```
ori-moisis, line 378: is the explicit return type necessary here? I think it is implied by the `return`...

packages/core/src/core/fetch.ts
```diff
@@ -383,14 +388,9 @@ const toFetchChanges = (
         return undefined
       }
 
-      // Find all changes that relate to the current ID and mark them as handled
-      const relatedChanges = getChangesNestedUnderID(elemId, serviceAndPendingChanges)
+      // Mark all changes that relate to the current ID as handled
```
ori-moisis, line 391: this feels too late...
we should mark the related changes as handled as soon as we get them

packages/core/src/core/plan/diff.ts
```diff
@@ -52,3 +77,295 @@ export const getDiffInstance = (change: Change<InstanceElement>): InstanceElemen
...
+          after: elementPair.after ?? (await after.get(id)),
+        }
+      }
+      return elementPair
+    }
+
+    const sieve = new Set<string>()
+
+    const isDifferent = async (change: Change): Promise<boolean> => {
+      // We can cast to string, at least one of the changes should be defined.
```
ori-moisis, line 338: comment doesn't really make sense given there is no cast here...

packages/core/src/core/plan/diff.ts
```diff
@@ -52,3 +77,295 @@ export const getDiffInstance = (change: Change<InstanceElement>): InstanceElemen
...
+  removeRedundantChanges = false,
+}: {
+  before: ReadOnlyElementsSource
+  after: ReadOnlyElementsSource
+  topLevelFilters: IDFilter[]
+  compareOptions?: CompareOptions
+  removeRedundantChanges?: boolean
+}): Promise<AsyncIterable<Change>> =>
+  log.timeDebug(async () => {
+    const addElementsNodes = async (comparison: BeforeAfter<ChangeDataType>): Promise<Change[]> => {
```
ori-moisis, line 290: this doesn't actually add anything, and there are no nodes.

it also feels like this function actually does 3 separate things:
1. filters out variables
2. creates a change from the "BeforeAfter" type thing
3. creates field changes from object changes

it might be more readable to split it into 3 "functions" (the first two can probably be anonymous single-line functions, only the one that adds fields would probably require more than a single line)

so the code at the end would be something like
```typescript
    return awu(
      iterateTogether(
        await getFilteredElements(before, topLevelFilters),
        await getFilteredElements(after, topLevelFilters),
        compareChangeDataType,
      ),
    )
      .map(handleSpecialIds)
      .filter(({ before, after }) => !isVariable(before) && !isVariable(after))
      .map(toChange)
      .flatMap(change => (createFieldChanges ? splitFieldChanges(change) : change))
      .filter(isDifferent)
```

packages/core/src/core/plan/diff.ts
```diff
@@ -52,3 +77,295 @@ export const getDiffInstance = (change: Change<InstanceElement>): InstanceElemen
...
+      return elementPair
+    }
+
+    const sieve = new Set<string>()
+
+    const isDifferent = async (change: Change): Promise<boolean> => {
+      // We can cast to string, at least one of the changes should be defined.
+      const fullName = getChangeData(change).elemID.getFullName()
+      if (!sieve.has(fullName)) {
+        sieve.add(fullName)
```
ori-moisis, line 341: does it make sense to have this here?
I am not sure I understand the use case.

we don't have to remove it in this PR, but might be worth checking and simplifying the code in a follow up PR

packages/core/src/core/plan/diff.ts
```diff
@@ -52,3 +77,295 @@ export const getDiffInstance = (change: Change<InstanceElement>): InstanceElemen
...
+      const beforeElement = comparison.before
+      const afterElement = comparison.after
+      if (!isVariable(beforeElement) && !isVariable(afterElement)) {
+        changes.push(toChange({ before: beforeElement, after: afterElement }))
+      }
+      if (
+        removeRedundantChanges &&
+        (beforeElement === undefined || afterElement === undefined) &&
+        (isObjectType(beforeElement) || isObjectType(afterElement))
+      ) {
```
ori-moisis, line 301: in terms of naming and readability, when it is possible to have a "positive" option it is usually better than having a "negative" option.
in this case, this isn't `removeRedundantChanges` because we are not actually removing anything, there were no changes to begin with so it is kind of confusing to say this is removing anything.
I think this should be called something like `addFieldChanges` or `createFieldChanges` (this aligns with the terminology in [CompareOptions](https://github.com/salto-io/salto/blob/ae8b0a9373f9d1ab4e32a669e1ec0d0b19781433/packages/adapter-api/src/values.ts/#L28)).
this would also make it clearer in the code IMO, instead of having this fairly large condition that checks multiple things, we could have a simple condition that surrounds the `await Promise.all` below

```typescript
    if (createFieldChanges) {
      await Promise.all(
        allFieldNames.map(fieldName =>
          addChangeIfDifferent(
            // We check `hasOwnProperty` and don't just do `beforeFields[fieldName]`
            // because fieldName might be a builtin function name such as
            // `toString` and in that case `beforeFields[fieldName]` will
            // unexpectedly return a function
            Object.prototype.hasOwnProperty.call(beforeFields, fieldName) ? beforeFields[fieldName] : undefined,
            Object.prototype.hasOwnProperty.call(afterFields, fieldName) ? afterFields[fieldName] : undefined,
          ),
        ),
      )
}
```

not sure if it would be more or less confusing but we could also just use the `createFieldChanges` in the compare options that we get here anyway (override it to be `true` in the call from `getPlan`)

packages/core/src/core/plan/diff.ts
```diff
@@ -52,3 +77,295 @@ export const getDiffInstance = (change: Change<InstanceElement>): InstanceElemen
...
+          return true
+        }
+      }
+      return false
+    }
+
+    /**
+     * Ids that represent types or containers need to be handled separately,
+     * because they would not necessarily be included in getAll.
+     */
```
ori-moisis, line 360: this comment doesn't make sense here, I guess it should be on `handleSpecialIds`

Nurdok:
Do you think maybe I should've guarded the new behavior with a feature flag?

packages/core/src/core/plan/diff.ts
```diff
@@ -52,3 +79,282 @@ export const getDiffInstance = (change: Change<InstanceElement>): InstanceElemen
...
+      ),
+    )
+      .map(handleSpecialIds)
+      .filter(
+        ({ before: beforeElement, after: afterElement }) => !isVariable(beforeElement) && !isVariable(afterElement),
+      )
+      .map(toChange)
+      .flatMap(splitFieldChanges)
+      .filter(isDifferent)
+  }, 'calculate diff between element sources')
```
ori-moisis, line 360: I might be wrong, but IIUC, this `logTime` won't return a meaningful result, because it just build an iterator (and does not consume it), it would essentially always return almost immediately. no?

the logTime should be on the functions that use this...

Nurdok: Done

packages/core/src/core/fetch.ts
```diff
@@ -372,7 +386,32 @@ const toFetchChanges = (
...
       const wsChanges = getChangesNestedUnderID(elemId, workspaceToServiceChanges).map(({ change }) => change)
+
+      // Service-to-workspace diffs for a given element are only computed when pending changes exist for that element.
+      // When there are no pending changes, the state and workspace are already aligned, so we can reuse service changes
+      // as workspace changes. We are guaranteed no conflicts, so we can return early here. This reuse relies on the
+      // assumption that reference expressions in the state element source used to compute the diff are *not* resolved.
+      //
+      // Note: There is another optimization for diff computation when the state itself is empty, in which case only the
+      // service-to-workspace diffs are computed (and pending changes are empty). We need to check that no workspace
+      // changes exist to avoid catching that edge case in the condition below.
```
ori-moisis, line 406: asking again, I guess you missed the last times
why is this part of the comment true?
in the case described here, the "service-to-workspace" changes are passed as "serviceChanges" and the correct thing would be to return them AFAICT.
why do we need a condition to explicitly avoid doing this?
what am I missing?

packages/core/src/core/plan/plan.ts
```diff
@@ -53,292 +41,58 @@ import {
...
-      second,
-      firstSrc,
-      secondSrc,
-      firstVisitedReferences: firstVisited,
-      secondVisitedReferences: secondVisited,
-      compareOptions,
-    })
-    if (referencesCompareResult.returnCode === 'return') {
-      return referencesCompareResult.returnValue
+const changeToDiffNode = (change: Change<ChangeDataType>): DiffNode<ChangeDataType> => {
```
ori-moisis, line 53: https://github.com/salto-io/salto/pull/6942#discussion_r1900352520

packages/core/src/core/plan/plan.ts
```diff
@@ -53,292 +41,58 @@ import {
...
+        before,
+        after,
+        topLevelFilters,
+        compareOptions: { ...compareOptions, createFieldChanges: true },
+      })
+      await awu(changes).forEach(change => {
+        outputGraph.addNode(changeId(change), [], changeToDiffNode(change))
+      })
+      return outputGraph
+    }, 'add differing elements to graph')
```
ori-moisis, line 95: is there a reason not to keep the old wording?
just to have an easier time comparing things...

packages/core/src/core/fetch.ts
```diff
@@ -132,19 +136,29 @@ export const getDetailedChanges = async (
...
-        dependencyChangers: [],
-        topLevelFilters,
-      })
-    ).itemsByEvalOrder(),
-  )
-    .map(item => item.detailedChanges())
-    .flatten()
+): Promise<DetailedChangeWithBaseChange[]> => {
+  if (getSaltoFlagBool(WORKSPACE_FLAGS.computePlanOnFetch)) {
+    log.info('fetching detailed changes with getPlan')
```
ori-moisis, line 141: technically this is not needed because `getPlan` does plenty of logging on its own (I don't love info logs that are actually for debugging purposes and appear in a fairly minor part of the code that runs multiple times)

packages/core/src/core/fetch.ts
Nurdok: You are correct, removed.

packages/core/src/core/fetch.ts
```diff
@@ -429,7 +460,7 @@ const toFetchChanges = (
       return wsChanges.map(createFetchChange)
     })
     .filter(values.isDefined)
-    .flatten(true)
+    .flatten()
```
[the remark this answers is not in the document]
Nurdok: Too many edits to this file back and forth... done.

packages/core/src/core/fetch.ts
```diff
@@ -872,13 +895,25 @@ export const calcFetchChanges = async ({
...
     const { changesTree: workspaceToServiceChanges } = await log.timeDebug(
       () =>
         getDetailedChangeTree(
           workspaceElements,
           partialFetchElementSource,
-          [accountFetchFilter, partialFetchFilter, serviceChangeIdsFilter],
+          [
+            accountFetchFilter,
+            partialFetchFilter,
+            getSaltoFlagBool(WORKSPACE_FLAGS.computePlanOnFetch) ? serviceChangeIdsFilter : pendingChangeIdsFilter,
```
ori-moisis, line 915: comment to explain?
not sure how long the flag is going to be here...

packages/core/src/core/plan/plan.ts
Nurdok: Done (without the numElements count)

ori-moisis: technically it is available here (the calling function has it anyway)

packages/core/src/core/plan/plan.ts
Nurdok: Done

packages/core/src/core/fetch.ts
Nurdok: Since I use this in several places, I put the explanation on the flag, WDYT?

packages/core/src/core/plan/plan.ts
Nurdok: In https://github.com/salto-io/salto/pull/6942#discussion_r1897354615 you said "that log is incorrect anyway if this is the calculation (given that it does not take the filters into account)" so I assumed you didn't want me to bring it back. Should I?

ori-moisis: not a strong opinion, but as a general rule of thumb, when I don't have to change something, I try not to

ori-moisis (approved):

PR #6945, opened 2024-12-15
SALTO-6639 - Salseforce - added error mapping for when user is missing permissions

rotem531:
added error mapping for when user is missing permissions
image

---

added error mapping for when user is missing permissions

---
_Release Notes_: 
none

---
_User Notifications_: 
none


Uri-Bar (approved):
lgtm

PR #6946, opened 2024-12-15
SALTO-7065: (Bug-fix) Fetch-From with `fromState` with addition of static file causes pending change

tamtamirr:
(Bug-fix) Fetch-From with `fromState` with addition of static file causes pending change

---

See comments.

---
_Release Notes_: 
_Core_:
- (Bug-fix) Fetch-From with `fromState` with addition of static file causes pending change.

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/core/src/core/fetch.ts
+++ b/packages/core/src/core/fetch.ts
@@ -1141,7 +1141,7 @@ const fixStaticFilesForFromStateChanges = async (
         return
       }
       const staticFilePath = staticFileValElemID.createTopLevelParentID().path
-      const relativePath = staticFilePath.filter(pathPart => !changePath.includes(pathPart))
+      const relativePath = staticFilePath.slice(changePath.length)
       _.set(change.data.after, relativePath, actualStaticFile)
     })
   })
--- a/packages/core/src/core/fetch.ts
+++ b/packages/core/src/core/fetch.ts
@@ -1141,7 +1141,7 @@ const fixStaticFilesForFromStateChanges = async (
         return
       }
       const staticFilePath = staticFileValElemID.createTopLevelParentID().path
-      const relativePath = staticFilePath.slice(changePath.length - 1)
+      const relativePath = staticFilePath.filter(pathPart => !changePath.includes(pathPart))
       _.set(change.data.after, relativePath, actualStaticFile)
     })
   })
--- a/packages/core/test/core/fetch.test.ts
+++ b/packages/core/test/core/fetch.test.ts
@@ -2188,7 +2188,7 @@ describe('fetch from workspace', () => {
           ) as FetchChange
           expect(nestedStaticFileChange).toBeDefined()
           const complexFieldValue = getChangeData(nestedStaticFileChange.change)
-          expect(complexFieldValue).toEqual({
+          expect(complexFieldValue).toStrictEqual({
             staticFileField: fileTwo,
             staticFilesArr: [fileThree],
           })
--- a/packages/core/test/core/fetch.test.ts
+++ b/packages/core/test/core/fetch.test.ts
@@ -35,6 +35,7 @@ import {
   toServiceIdsString,
   ElemIdGetter,
   ServiceIds,
+  isElement,
 } from '@salto-io/adapter-api'
 import * as utils from '@salto-io/adapter-utils'
 import { collections } from '@salto-io/lowerdash'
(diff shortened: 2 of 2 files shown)


packages/core/src/core/fetch.ts
```diff
@@ -1141,7 +1141,7 @@ const fixStaticFilesForFromStateChanges = async (
         return
       }
       const staticFilePath = staticFileValElemID.createTopLevelParentID().path
-      const relativePath = staticFilePath.slice(changePath.length - 1)
+      const relativePath = staticFilePath.filter(pathPart => !changePath.includes(pathPart))
```
tamtamirr, line 1144: The relativePath was incorrect, which later we use `lodash set` to the wrong path. That's what caused the bug. It wasn't covered in UTs either.

packages/core/test/core/fetch.test.ts
```diff
@@ -2174,10 +2177,23 @@ describe('fetch from workspace', () => {
...
+          const changes = [...fetchRes.changes]
+          const nestedStaticFileChange = changes.find(c =>
+            c.change.id.isEqual(editStateExistingInstance.elemID.createNestedID('complexField')),
+          ) as FetchChange
+          expect(nestedStaticFileChange).toBeDefined()
+          const complexFieldValue = getChangeData(nestedStaticFileChange.change)
+          expect(complexFieldValue).toEqual({
+            staticFileField: fileTwo,
+            staticFilesArr: [fileThree],
+          })
```
tamtamirr, line 2194: This will fail with the code in main. The `complexFieldValue` would be:
```
{
  staticFileField: ...,
  staticFilesArr: ....,
  complexFieldValue: ... // bug
}
```

packages/core/test/core/fetch.test.ts
```diff
@@ -2196,9 +2212,9 @@ describe('fetch from workspace', () => {
               change.change.id.createTopLevelParentID().parent.isEqual(editStateExistingInstance.elemID),
             )
             .map(change => getChangeData(change.change))
-          expect(modifyStaticVals).toHaveLength(3)
+          expect(modifyStaticVals).toHaveLength(2)
```
tamtamirr, line 2215: Instead of modification on each `Static File` within `complexField`, there's a single addition change with the `complexField`.

packages/core/test/core/fetch.test.ts
```diff
@@ -2196,9 +2212,9 @@ describe('fetch from workspace', () => {
...
           const staticFileModifies = modifyStaticVals.filter(val => isStaticFile(val))
-          expect(staticFileModifies).toHaveLength(3)
+          expect(staticFileModifies).toHaveLength(1)
```
tamtamirr, line 2217: There's only a single direct change on `StaticFile` now: `staticFileField`

packages/core/test/core/fetch.test.ts
tamtamirr, line 2215: instead of two modifications within `complexField`, there's a single addition change now.

packages/core/src/core/fetch.ts
ori-moisis: this can't be right
consider if the static file path is `children.0.value.children.1.value` and the change path is `children.0` - you will remove the `children` twice

packages/core/test/core/fetch.test.ts
```diff
@@ -2174,10 +2177,23 @@ describe('fetch from workspace', () => {
...
+          const nestedStaticFileChange = changes.find(c =>
+            c.change.id.isEqual(editStateExistingInstance.elemID.createNestedID('complexField')),
+          ) as FetchChange
+          expect(nestedStaticFileChange).toBeDefined()
+          const complexFieldValue = getChangeData(nestedStaticFileChange.change)
+          expect(complexFieldValue).toStrictEqual({
+            staticFileField: fileTwo,
+            staticFilesArr: [fileThree],
+          })
+        })
```
tamtamirr, line 2195: This UT covers two nesting levels from the base change of `complexField` addition:
1. Relative path: `['staticFileField']`
2. Relative path: `['staticFilesArr', '0']`

packages/core/src/core/fetch.ts
tamtamirr: fixed. Thanks 🙏

ori-moisis (approved):

PR #6949, opened 2024-12-16
SALTO-6542 - salesforce - update error message when type is unknown

rotem531:
update error message when type is unknown

---

update error message when type is unknown to suggest the customer look at account/org settings

---
_Release Notes_: 
none

---
_User Notifications_: 
none


Uri-Bar (approved):
nice

PR #6951, opened 2024-12-16
SALTO-7071: Export NS fetch targets

tamtamirr:
Export NS fetch targets

---

_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


dantal4:
@ori-moisis @netama are we fine with exporting adapter-specific information from the adapter package ?

ori-moisis:
we already crossed that bridge I think

dantal4 (approved):
💯

PR #6952, opened 2024-12-16
SALTO-7068: Set a hardcoded element ID for the Okta Dashboard app

Nurdok:
The Okta Dashboard app is a singleton - only one of its type is allowed. However, it has an autogenerated `id` from the service, and we use its label as the element ID, which users can change. We want the element ID to be fixed, so that deploying Okta Dashboard app changes between environments where that app's label was changed, to modify the correct app and not apply an add+remove change (which is disallowed by the service).

This PR also fixes a bug in computing `AppLogo` element IDs (which are based on their parent app ID) - before this fix, element IDs for `AppLogo`s were always recomputed regardless of whether `--regenerate-salto-ids` was supplied or not.

---

I manually tested that running `fetch` without `--regenerate-salto-ids` results in zero changes.
And [here's the workspace diff](https://github.com/salto-io/amir-okta/commit/33799f1d8f1b0ed4668fad8a48b02829baa36d42) for running _with_ `--regenerate-salto-ids`.

---
_Release Notes_: 
_Okta adapter_:
- Fixed a bug where modification deployments of the "Okta Dashboard" application would fail if the app had different labels between the environments in the deployment.

---
_User Notifications_: 
- When regenerating Salto IDs, the "Okta Dashboard" application instance's element ID will change to "Okta Dashboard" even if its label was changed.

### DIFF
--- a/packages/okta-adapter/src/definitions/fetch/fetch.ts
+++ b/packages/okta-adapter/src/definitions/fetch/fetch.ts
@@ -40,7 +40,7 @@ import {
 import { isGroupPushEntry } from '../../filters/group_push'
 import { extractSchemaIdFromUserType } from './types/user_type'
 import { isNotMappingToAuthenticatorApp } from './types/profile_mapping'
-import { assignPolicyIdsToApplication } from './types/application'
+import { assignPolicyIdsToApplication, isOktaDashboard } from './types/application'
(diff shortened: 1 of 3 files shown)


shir-reifenberg (approved):
Thanks! 👑

PR #6954, opened 2024-12-17
SALTO-7076: Add lint rules to prevent invalid imports

ori-moisis:
Add rules to prevent importing from files that are not distributed to NPM

---

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


dorsev (approved):
LGTM! thank you for this!

PR #6955, opened 2024-12-17
SALTO-6594: Add a CV to prevent deployment of JIM field

YuvalGivon:
Added a field validator to prevent deployment of a searcherKey value for the field type 'textfield':
searcherKey = 'com.atlassian.jira.plugin.system.customfieldtypes:exacttextsearcher'
type = 'com.atlassian.jira.plugin.system.customfieldtypes:textfield'

---

---
_Release Notes_: 
_Jira_adapter_:
- Added new CV to prevent deployment of JIM field

---
_User Notifications_:

### DIFF
--- a/packages/jira-adapter/src/change_validators/field.ts
+++ b/packages/jira-adapter/src/change_validators/field.ts
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2024 Salto Labs Ltd.
+ * Licensed under the Salto Terms of Use (the "License");
+ * You may not use this file except in compliance with the License.  You may obtain a copy of the License at https://www.salto.io/terms-of-use
+ *
+ * CERTAIN THIRD PARTY SOFTWARE MAY BE CONTAINED IN PORTIONS OF THE SOFTWARE. See NOTICE FILE AT https://github.com/salto-io/salto/blob/main/NOTICES
+ */
+import {
+  ChangeValidator,
+  getChangeData,
+  isAdditionOrModificationChange,
+  isInstanceChange,
+  SeverityLevel,
+} from '@salto-io/adapter-api'
+
+const SEARCHER_KEY = 'com.atlassian.jira.plugin.system.customfieldtypes:exacttextsearcher'
+const TYPE = 'com.atlassian.jira.plugin.system.customfieldtypes:textfield'
+
+export const fieldValidator: ChangeValidator = async changes =>
+  changes
+    .filter(isInstanceChange)
+    .filter(isAdditionOrModificationChange)
+    .filter(change => getChangeData(change).elemID.typeName === 'Field')
+    .filter(change => change.data.after.value.searcherKey === SEARCHER_KEY && change.data.after.value.type === TYPE)
+    .map(change => ({
+      elemID: getChangeData(change).elemID,
+      severity: 'Error' as SeverityLevel,
+      message: 'SearcherKey is invalid for the field type',
+      detailedMessage: 'SearcherKey is invalid for the field type.',
+    }))
--- a/packages/jira-adapter/src/change_validators/field.ts
+++ b/packages/jira-adapter/src/change_validators/field.ts
@@ -12,19 +12,24 @@ import {
   isInstanceChange,
   SeverityLevel,
 } from '@salto-io/adapter-api'
+import { FIELD_TYPE_NAME } from '../filters/fields/constants'
 
-const SEARCHER_KEY = 'com.atlassian.jira.plugin.system.customfieldtypes:exacttextsearcher'
-const TYPE = 'com.atlassian.jira.plugin.system.customfieldtypes:textfield'
+export const EXACT_TEXT_SEARCHER_KEY = 'com.atlassian.jira.plugin.system.customfieldtypes:exacttextsearcher'
+export const TEXT_FIELD_TYPE = 'com.atlassian.jira.plugin.system.customfieldtypes:textfield'
 
 export const fieldValidator: ChangeValidator = async changes =>
   changes
     .filter(isInstanceChange)
     .filter(isAdditionOrModificationChange)
-    .filter(change => getChangeData(change).elemID.typeName === 'Field')
-    .filter(change => change.data.after.value.searcherKey === SEARCHER_KEY && change.data.after.value.type === TYPE)
-    .map(change => ({
-      elemID: getChangeData(change).elemID,
+    .map(getChangeData)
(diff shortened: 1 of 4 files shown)


packages/jira-adapter/src/change_validators/field.ts
```diff
@@ -0,0 +1,30 @@
...
+} from '@salto-io/adapter-api'
+
+const SEARCHER_KEY = 'com.atlassian.jira.plugin.system.customfieldtypes:exacttextsearcher'
+const TYPE = 'com.atlassian.jira.plugin.system.customfieldtypes:textfield'
+
+export const fieldValidator: ChangeValidator = async changes =>
+  changes
+    .filter(isInstanceChange)
+    .filter(isAdditionOrModificationChange)
+    .filter(change => getChangeData(change).elemID.typeName === 'Field')
```
IdoZakk, line 23: We try to avoid writing strings in our code. For type names (which are used in several places) we keep those consts in a[ file named constants](https://github.com/salto-io/salto/blob/main/packages/jira-adapter/src/constants.ts)

packages/jira-adapter/src/change_validators/field.ts
```diff
@@ -0,0 +1,30 @@
...
+    .filter(change => change.data.after.value.searcherKey === SEARCHER_KEY && change.data.after.value.type === TYPE)
```
IdoZakk, line 24: You are using `change.data.after` twice (and also getChangeData above)
You can map using getChangeData and avoid those (you might also need to push down the check for instanceElement

packages/jira-adapter/src/change_validators/field.ts
```diff
@@ -0,0 +1,30 @@
...
+    .map(change => ({
+      elemID: getChangeData(change).elemID,
+      severity: 'Error' as SeverityLevel,
+      message: 'SearcherKey is invalid for the field type',
```
IdoZakk, line 28: We try to keep the errors as informative for the users as possible, so they will know what happened and what to do. 
The detailed message should suggest that this field was created using JIT, that it will fail deployment, and suggest to edit the nacl and change `exacttextsearcher` to `textsearcher`

packages/jira-adapter/test/change_validators/field.test.ts
```diff
@@ -0,0 +1,64 @@
...
+const EXACT_TEXT_SEARCHER_KEY = 'com.atlassian.jira.plugin.system.customfieldtypes:exacttextsearcher'
+const TEXT_FIELD_TYPE = 'com.atlassian.jira.plugin.system.customfieldtypes:textfield'
+
+describe('fieldValidator', () => {
+  let type: ObjectType
+  let instance: InstanceElement
+  let afterInstance: InstanceElement
+
+  beforeEach(() => {
+    type = new ObjectType({ elemID: new ElemID(JIRA, 'Field') })
```
IdoZakk, line 21: Nit: we have a test util function called createEmptyType for those creations

packages/jira-adapter/test/change_validators/field.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    ).toEqual([
+      {
+        elemID: instance.elemID,
+        severity: 'Error',
+        message: 'SearcherKey is invalid for the field type',
+        detailedMessage: 'SearcherKey is invalid for the field type.',
+      },
+    ])
+  })
+})
```
IdoZakk, line 64: Please add a test to the normal flow, to make sure we do not block legit filters (you can run on the instance you created)
Also a nit, I think when we handle an array we should make a test of at least two items to make sure we properly go over the array.

IdoZakk (commented):
Very nice 🎸 
Please see me notes below :-)

shayc331 (commented):
looking good! 
had some comments

packages/jira-adapter/src/change_validators/field.ts
```diff
@@ -12,19 +12,24 @@ import {
...
+    .filter(
+      instance => instance.value.searcherKey === EXACT_TEXT_SEARCHER_KEY && instance.value.type === TEXT_FIELD_TYPE,
+    )
+    .map(instance => ({
+      elemID: instance.elemID,
       severity: 'Error' as SeverityLevel,
       message: 'SearcherKey is invalid for the field type',
-      detailedMessage: 'SearcherKey is invalid for the field type.',
+      detailedMessage:
+        'The field created using JIT cannot be deployed. To resolve this issue, edit the NaCl file and replace in searchKey field "exacttextsearcher" with "textsearcher".',
```
IdoZakk, line 34: Good! Some minor phrasing suggestions:
`This field was created using JIT and cannot be deployed. To resolve this issue, edit the NaCl file, and in the searchKey field, replace "exacttextsearcher" with "textsearcher"`

packages/jira-adapter/test/change_validators/field.test.ts
```diff
@@ -5,29 +5,35 @@
...
+    expect(
+      await fieldValidator([
+        toChange({
+          after: instance,
+        }),
+      ]),
+    ).toEqual([])
+  })
+
+  it('Should not return an error when adding an instance with an invalid searcherKey for this type', async () => {
```
IdoZakk, line 36: remove the 'not' from the title

packages/jira-adapter/test/change_validators/field.test.ts
```diff
@@ -39,12 +45,13 @@ describe('fieldValidator', () => {
...
         message: 'SearcherKey is invalid for the field type',
-        detailedMessage: 'SearcherKey is invalid for the field type.',
+        detailedMessage:
+          'The field created using JIT cannot be deployed. To resolve this issue, edit the NaCl file and replace in searchKey field "exacttextsearcher" with "textsearcher".',
       },
     ])
   })
 
-  it('should return an error if changed to the exact text searcher key', async () => {
+  it('Should not return an error when modifying an instance with an invalid searcherKey for this type', async () => {
```
IdoZakk, line 54: remove the 'not' here as well

IdoZakk (commented):
Almost done 😃 
See comments below.
Also see my comment from the previous CR about testing more than one message. I usually try to add two instances that will trigger the CV and one that doesn't in the same array

IdoZakk (approved):
💯 
Two small notes:
- After a CR please do not squash new changes with previous commits, sometimes, it is easier to look at the last commit 
- Regarding the tests: it is OK like this, I meant adding another test so if something fails it is easier to understand what the problem is (based on the descriptions of each test). But no need to change for this PR

PR #6963, opened 2024-12-17
SALTO-7067: add `accountElements` utility to Workspace

tamtamirr:
Add **accountElements** utility to Workspace.

---

- Moved useful implementations of **ElementSources** from `core` to `adapter-utils`.
- Implemented a Workspace utility that returns elements for the provided account. 

---
_Release Notes_: 
_Workspace_:
- Added method `accountElements` that returns **ElementsSource** of the provided account.

---
_User Notifications_: 
_None_


packages/workspace/src/workspace/workspace.ts
```diff
@@ -185,6 +187,12 @@ export type Workspace = {
   uid: string
 
   elements: (includeHidden?: boolean, env?: string) => Promise<ElementsSource>
+  accountElements: (args: {
```
adamro, line 190: What's the reasoning behind this being a new function and not another param to `elements`?

tamtamirr: I think it made sense to have a dedicated implementation for a specific account's ElementsSource, where the account name is a required parameter.
Also, while I could add more optional parameters to elements, I think this would make the elements method harder to use, especially since it doesn't use an object parameter, and changing it would cause a breaking change.

ori-moisis (changes requested):
I don't see a reason this should be added to the workspace interface
if this is part of a bigger plan, I would like to see what the bigger plan is and why this has to be a part of it

PR #6969, opened 2024-12-18
SALTO-7089 wait for workspaceState to finish building in workspace.cl…

dantal4:
…ose()

When we run a workspace method that rebuilds the workspaceState (e.g updateNaclFiles), we need to wait for the workspaceState to finish rebuilding before calling remoteMapCreator.close() , else the remote map calls in the building process will fail on closed connection.

---

---
_Release Notes_: 
Workspace:
- prevent a future bug of closing a workspace while a workspaceState is being built, causing to remoteMap calls to fail on closed connections.

---
_User Notifications_: 
None

### DIFF
--- a/packages/workspace/src/workspace/workspace.ts
+++ b/packages/workspace/src/workspace/workspace.ts
@@ -1212,9 +1212,11 @@ export const loadWorkspace = async (
     state,
     close: async () => {
       if (workspaceState !== undefined) {
-        // in case that `workspaceState` is currently beind built, we'd like to wait until it finishes.
+        // in case that `workspaceState` is currently being built, we'd like to wait until it finishes.
+        log.debug('waiting for workspaceState to finish building before closing the workspace')
         await workspaceState
       }
+      log.debug('closing the workspace')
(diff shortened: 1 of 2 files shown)


dantal4:
@ori-moisis  do you have a suggestion how to test it?

packages/workspace/src/workspace/workspace.ts
```diff
@@ -1211,6 +1211,10 @@ export const loadWorkspace = async (
     elements: elementsImpl,
     state,
     close: async () => {
+      if (workspaceState !== undefined) {
+        // in case that `workspaceState` is currently beind built, we'd like to wait until it finishes.
+        await workspaceState
```
omrilit, line 1216: Maybe worth adding a log so we could understand why it may take some time to close it?

ori-moisis:
you can spy on something that is used when building the workspace state and then use `makeResolvablePromise` or `stepManager` from `@salto-io/test-utils` to manage when each step happens and ensure the close is called before the workspace state is built

omrilit (approved):
🔥

PR #6972, opened 2024-12-18
SALTO-7095: add getAndLogCollisionWarningsV2

shayc331:

---

an example of the new detailedMessage: 
```
    2 salto elements and their child elements were not fetched, as they were mapped to a single ID salto.obj.instance.test:
    [aliasName](someUrl), [aliasName](someUrl)
    
    Usually, this happens because of duplicate configuration names in the service. Make sure these element names are unique, and try fetching again.
    [Learn about additional ways to resolve this issue](https://help.salto.io/en/articles/6927157-salto-id-collisions)
```

---
_Release Notes_: 

---
_User Notifications_: 

### DIFF
--- a/packages/adapter-utils/src/collisions.ts
+++ b/packages/adapter-utils/src/collisions.ts
@@ -173,7 +173,6 @@ ${getInstancesMarkdownLinks(collideInstances)}
 Usually, this happens because of duplicate configuration names in the service. Make sure these element names are unique, and try fetching again.
 ${createMarkdownLink({ text: 'Learn about additional ways to resolve this issue', url: 'https://help.salto.io/en/articles/6927157-salto-id-collisions' })}`,
   )
-  console.log(warningMessages[0])
   return warningMessages.map(warningMessage =>
     createWarningFromMsg({
       message: 'Some elements were not fetched due to Salto ID collisions',
--- a/packages/adapter-utils/src/collisions.ts
+++ b/packages/adapter-utils/src/collisions.ts
@@ -87,9 +87,7 @@ const logInstancesWithCollidingElemID = async (
   })
 }
 // after SALTO-7088 is done, this function will replace logInstancesWithCollidingElemID
-const logInstancesWithCollidingElemIDV2 = async (
-  elemIDtoInstances: Record<string, InstanceElement[]>,
-): Promise<void> => {
+const logInstancesWithCollidingElemIDV2 = (elemIDtoInstances: Record<string, InstanceElement[]>): void => {
   Object.entries(elemIDtoInstances).forEach(([elemID, collideInstances]) => {
     if (collideInstances.length <= 1) {
       log.error('Collide instances length is less than or equal to 1 for elemID %s', elemID)
@@ -137,33 +135,29 @@ ${getInstancesDetailsMsg(instanceDetails, baseUrl, maxBreakdownDetailsElements)}
   return collisionMessages
 }
 
-const createMarkdownLink = ({ text, url }: { text: string; url: string }): string => `[${text}](${url})`
+const createMarkdownLinkOrText = ({ text, url }: { text: string; url?: string }): string =>
+  url !== undefined ? `[${text}](${url})` : text
 
 const getInstancesMarkdownLinks = (instances: InstanceElement[]): string =>
   instances
     .map(instance =>
-      createMarkdownLink({
+      createMarkdownLinkOrText({
         text: instance.annotations[CORE_ANNOTATIONS.ALIAS] ?? instance.elemID.name,
         url: instance.annotations[CORE_ANNOTATIONS.SERVICE_URL],
       }),
     )
     .join(', ')
 
-// after SALTO-7088 is done, this function will replace getAndLogCollisionWarnings
-export const getAndLogCollisionWarningsV2 = async ({
-  instances,
+const createWarningMessages = ({
+  elemIDtoInstances,
+  adapterName,
   addChildrenMessage,
 }: {
-  instances: InstanceElement[]
+  elemIDtoInstances: Record<string, InstanceElement[]>
+  adapterName: string
   addChildrenMessage?: boolean
-}): Promise<SaltoError[]> => {
-  if (instances.length === 0) {
-    return []
-  }
-  const adapterName = instances[0].elemID.adapter
-  const elemIDtoInstances = _.groupBy(instances, instance => instance.elemID.getFullName())
-  await logInstancesWithCollidingElemIDV2(elemIDtoInstances)
-  const warningMessages = Object.entries(elemIDtoInstances).map(
+}): string[] =>
+  Object.entries(elemIDtoInstances).map(
     ([
       elemId,
       collideInstances,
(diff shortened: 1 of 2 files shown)


packages/adapter-utils/src/collisions.ts
```diff
@@ -119,6 +137,50 @@ ${getInstancesDetailsMsg(instanceDetails, baseUrl, maxBreakdownDetailsElements)}
...
+  const warningMessages = Object.entries(elemIDtoInstances).map(
+    ([
+      elemId,
+      collideInstances,
+    ]) => `${collideInstances.length} ${adapterName} elements ${addChildrenMessage ? 'and their child elements ' : ''}were not fetched, as they were mapped to a single ID ${elemId}:
+${getInstancesMarkdownLinks(collideInstances)}
+
+Usually, this happens because of duplicate configuration names in the service. Make sure these element names are unique, and try fetching again.
+${createMarkdownLink({ text: 'Learn about additional ways to resolve this issue', url: 'https://help.salto.io/en/articles/6927157-salto-id-collisions' })}`,
+  )
```
dantal4, line 175: please extract to a function

packages/adapter-utils/src/collisions.ts
```diff
@@ -119,6 +137,50 @@ ${getInstancesDetailsMsg(instanceDetails, baseUrl, maxBreakdownDetailsElements)}
   return collisionMessages
 }
 
+const createMarkdownLink = ({ text, url }: { text: string; url: string }): string => `[${text}](${url})`
+
+const getInstancesMarkdownLinks = (instances: InstanceElement[]): string =>
+  instances
+    .map(instance =>
+      createMarkdownLink({
+        text: instance.annotations[CORE_ANNOTATIONS.ALIAS] ?? instance.elemID.name,
```
dantal4, line 146: in SF, they were passing a custom `getInstanceName` function, so I think that it should be an optional param for `getAndLogCollisionWarningsV2` with a default value of `instance.annotations[CORE_ANNOTATIONS.ALIAS] ?? instance.elemID.name`.

packages/adapter-utils/src/collisions.ts
```diff
@@ -119,6 +137,50 @@ ${getInstancesDetailsMsg(instanceDetails, baseUrl, maxBreakdownDetailsElements)}
...
+        url: instance.annotations[CORE_ANNOTATIONS.SERVICE_URL],
```
dantal4, line 147: what should happen if `_service_url` is undefined ?

packages/adapter-utils/src/collisions.ts
```diff
@@ -119,6 +137,50 @@ ${getInstancesDetailsMsg(instanceDetails, baseUrl, maxBreakdownDetailsElements)}
...
+  await logInstancesWithCollidingElemIDV2(elemIDtoInstances)
+  const warningMessages = Object.entries(elemIDtoInstances).map(
+    ([
+      elemId,
+      collideInstances,
+    ]) => `${collideInstances.length} ${adapterName} elements ${addChildrenMessage ? 'and their child elements ' : ''}were not fetched, as they were mapped to a single ID ${elemId}:
+${getInstancesMarkdownLinks(collideInstances)}
+
+Usually, this happens because of duplicate configuration names in the service. Make sure these element names are unique, and try fetching again.
+${createMarkdownLink({ text: 'Learn about additional ways to resolve this issue', url: 'https://help.salto.io/en/articles/6927157-salto-id-collisions' })}`,
```
dantal4, line 174: we should have an optional `docsUrl` too. see SF usage.

packages/adapter-utils/src/collisions.ts
```diff
@@ -86,6 +86,24 @@ const logInstancesWithCollidingElemID = async (
     })
   })
 }
+// after SALTO-7088 is done, this function will replace logInstancesWithCollidingElemID
+const logInstancesWithCollidingElemIDV2 = async (
```
dantal4, line 90: why do we need another function? and why is it async?

dantal4 (commented):
Great work!

packages/adapter-utils/src/collisions.ts
shayc331: good catch. In the new detailed message we want to use the aliases, I think the instance's name should be:
```
instance.annotations[CORE_ANNOTATIONS.ALIAS] ?? getInstanceName(instance)
```
and the default for getInstanceName is `instance.elemID.name`
WDYT?

packages/adapter-utils/src/collisions.ts
shayc331: salesforce specific docs url is: https://help.salto.io/en/articles/6927217-salto-for-salesforce-cpq-support, and it is broken. I talked @tomermevorach about it, and we agreed to show the general article

packages/adapter-utils/src/collisions.ts
shayc331: talked with @tamtamirr about it, he said that all the instances have aliases, so it makes this function redundant

packages/adapter-utils/src/collisions.ts
shayc331: I will change it not to create a link when serviceUrl is not provided

packages/adapter-utils/src/collisions.ts
shayc331: The old function gets a different (and unnecessary) data structure.
IMO the simplest solution is to use another function and delete the old one when we stop using it.

the async is redundant, removed it

packages/adapter-utils/test/collisions.test.ts
```diff
@@ -138,4 +154,117 @@ Alternatively, you can exclude obj from the default configuration in salto.nacl`
...
     })
   })
+  describe('getAndLogCollisionWarningsV2', () => {
+    const prefix = (x: string): string => `${x} salto elements `
+    const andTheirChildElements = 'and their child elements '
+    const wereNotFetched = (elemID: string): string => `were not fetched, as they were mapped to a single ID ${elemID}:`
+    const usuallyThisHappens =
+      'Usually, this happens because of duplicate configuration names in the service. Make sure these element names are unique, and try fetching again.'
+    const learnMore =
+      '[Learn about additional ways to resolve this issue](https://help.salto.io/en/articles/6927157-salto-id-collisions)'
```
dantal4, line 164: all of those partial parts makes it harder to understand, and it will be harder to debug in the future. I think that it will be better to write the whole message directly in each of the tests.

packages/adapter-utils/test/collisions.test.ts
```diff
@@ -138,4 +154,117 @@ Alternatively, you can exclude obj from the default configuration in salto.nacl`
...
+
+    it('should return the correct warning messages', async () => {
+      const instancesLinks = '[aliasName](someUrl), [aliasName](someUrl)\n'
+      const detailedMessage = `${prefix('2')}${wereNotFetched(instance.elemID.getFullName())}
+${instancesLinks}
+${usuallyThisHappens}
+${learnMore}`
+      const errors = getAndLogCollisionWarningsV2({
+        instances: [instance, instance.clone()],
```
dantal4, line 173: can we run in on different instances, that have different alias/url , so it will be a more real use case?

packages/adapter-utils/test/collisions.test.ts
```diff
@@ -138,4 +154,117 @@ Alternatively, you can exclude obj from the default configuration in salto.nacl`
...
+      expect(errors).toHaveLength(1)
+      expect(errors[0]).toEqual({
+        severity: 'Warning',
+        message: COLLISION_MESSAGE,
+        detailedMessage,
+      })
+    })
+    it('should return no errors when there are no collided instances', async () => {
+      const errors = getAndLogCollisionWarningsV2({
+        instances: [],
```
dantal4, line 238: why are we passing an empty list and not a list of unique instances?

dantal4 (commented):
looks great !

packages/adapter-utils/src/collisions.ts
```diff
@@ -86,6 +86,22 @@ const logInstancesWithCollidingElemID = async (
     })
   })
 }
+// after SALTO-7088 is done, this function will replace logInstancesWithCollidingElemID
+const logInstancesWithCollidingElemIDV2 = (elemIDtoInstances: Record<string, InstanceElement[]>): void => {
+  Object.entries(elemIDtoInstances).forEach(([elemID, collideInstances]) => {
+    if (collideInstances.length <= 1) {
+      log.error('Collide instances length is less than or equal to 1 for elemID %s', elemID)
```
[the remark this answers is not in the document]
shayc331: yes, it may point to a bug in the adapter

packages/adapter-utils/src/collisions.ts
```diff
@@ -86,6 +86,22 @@ const logInstancesWithCollidingElemID = async (
     })
   })
 }
+// after SALTO-7088 is done, this function will replace logInstancesWithCollidingElemID
+const logInstancesWithCollidingElemIDV2 = (elemIDtoInstances: Record<string, InstanceElement[]>): void => {
+  Object.entries(elemIDtoInstances).forEach(([elemID, collideInstances]) => {
+    if (collideInstances.length <= 1) {
```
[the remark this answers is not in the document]
dantal4: you're right, but this is an inner function, that you call after `groupBy`, that won't return a key with empty list of instances.
we can leave it this way if you prefer.

packages/adapter-utils/src/collisions.ts
```diff
@@ -86,6 +86,22 @@ const logInstancesWithCollidingElemID = async (
...
+      log.error('Collide instances length is less than or equal to 1 for elemID %s', elemID)
+      return
+    }
+    const type = collideInstances[0].elemID.typeName
+    const instancesCount = collideInstances.length
+    log.debug(`Omitted ${instancesCount} instances of type ${type} due to Salto ID collisions`)
```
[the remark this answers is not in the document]
shayc331: On second thought, I will remove this function, I kept it because the old code used it, but it does not make sense to me because the collision errors are user-visible, and the purpose of this function is to log very similar messages.

dantal4 (approved):
💯 💯 💯

PR #6973, opened 2024-12-18
SALTO-7087: keep statusMigrations when deploying workflowScheme

IdoZakk:
Current situation creates partial success for deployments and triggers an incident

---

I decided to do the long way, and properly clone the change before deployment. That is the correct behavior and the one used in the deployment function, and I think it reduces future potential issues
[Noise reduction](https://github.com/salto-io/salto_private/pull/10078)

---
_Release Notes_: 
Jira Adapter:
* Deployments of Workflow Schemes with migration will not end in partial success

---
_User Notifications_: 
None


shayc331 (approved):
clean and elegant  💯

PR #6975, opened 2024-12-19
SALTO-7097: add more references to json body

edenhassid:
add more references to json body

---

---
_Release Notes_: 
zendesk: 
* add more references to json body when `convertJsonIdsToReferences` = true

---
_User Notifications_: 


adfineberg (approved):
🔥

PR #6977, opened 2024-12-19
SALTO-6069: CV for invalid WorkflowV2 transition names

YuvalGivon:
Added CV for a limitation for global transitions names- global transitions must have unique names.

---

---
_Release Notes_: 
_Jira_adapter_:
- Added new CV to limit duplication of global transitions
---
_User Notifications_:


packages/jira-adapter/src/change_validators/workflowsV2/outbound_transition.ts
```diff
@@ -0,0 +1,89 @@
...
+): boolean => {
+  if (map.has(key) && map.get(key)?.has(name)) {
+    return false
+  }
+  if (!map.has(key)) {
+    map.set(key, new Set<string>())
+  }
+  map.get(key)?.add(name)
+  return true
+}
```
idoamit8, line 40: I think this will be cleaner. There is no reason to add an empty set to the map and then get it to add an element to it.
```suggestion
const addTransitionToMap = (
  map: Map<string | ReferenceExpression, Set<string>>,
  key: string | ReferenceExpression,
  name: string,
): boolean => {
  const keySet = map.get(key)
  if (keySet !== undefined) {
    return keySet.has(name)
  }
  const addedSet = new Set<string>([name])
  map.set(key, addedSet)
  return true
}
```

idoamit8 (commented):
Great job! 💯
I hope you don't mind me jumping in to review the code. 🙂
I’ve added a few comments, but I’ll leave the rest of the review to @shayc331

packages/jira-adapter/src/change_validators/workflowsV2/outbound_transition.ts
```diff
@@ -0,0 +1,89 @@
...
+import { isWorkflowV2Instance, WorkflowTransitionLinks, WorkflowV2Instance } from '../../filters/workflowV2/types'
+
+let fromStatusReference: string | ReferenceExpression | undefined
+let transitionName: string
+let transitionType: string
+let fromStatusReferenceName: string | undefined
+const { isDefined } = values
+const { awu } = collections.asynciterable
+
+const addTransitionToMap = (
```
idoamit8, line 27: nit: The function name `addTransitionToMap` feels a bit unintuitive since it returns a boolean. Typically, I find it helpful when functions that return booleans start with `is`, `has`, `does`, etc., as it makes the return value feel like an answer to a question.

You might consider splitting this into two functions – one to handle the boolean check and another for the insertion.

packages/jira-adapter/src/change_validators/workflowsV2/outbound_transition.ts
```diff
@@ -0,0 +1,89 @@
...
+    fromStatusReferenceName = isReferenceExpression(fromStatusReference)
+      ? fromStatusReference.elemID.getFullName()
+      : undefined
+
+    return fromStatusReferenceName ? addTransitionToMap(map, fromStatusReferenceName, transitionName) : true
+  })
+
+const isValidTransitions = (instance: WorkflowV2Instance): boolean => {
+  const transitionMap = new Map<string | ReferenceExpression, Set<string>>()
+  const transitions = Object.values(instance.value.transitions)
```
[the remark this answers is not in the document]
shayc331: it can not, `isValidWorkflowV2` validates it

packages/jira-adapter/src/change_validators/workflowsV2/outbound_transition.ts
YuvalGivon: Thanks i will try to make it clearer, 
your implementation does not cover all cases (key can have more than one name)

shayc331 (commented):
while reviewing I faced another validator that covers some of this pr content.
Even though the code is going to change Im leaving the comments I wrote, I think they may be useful

packages/jira-adapter/src/change_validators/workflowsV2/global_transition.ts
```diff
@@ -0,0 +1,57 @@
...
+  SeverityLevel,
+  getChangeData,
+  isInstanceChange,
+  isAdditionOrModificationChange,
+} from '@salto-io/adapter-api'
+import { isWorkflowV2Instance, WorkflowV2Instance } from '../../filters/workflowV2/types'
+
+const { isDefined } = values
+
+const getInvalidGlobalTransitions = (instance: WorkflowV2Instance): string[] => {
```
shayc331, line 20: you can filter the global transitions and then use _.groupBy name to find global transitions with the same name. I think It will be cleaner

packages/jira-adapter/src/change_validators/workflowsV2/global_transition.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export const globalTransitionValidator: ChangeValidator = async changes =>
+  changes
+    .filter(isInstanceChange)
+    .filter(isAdditionOrModificationChange)
+    .map(getChangeData)
+    .filter(isWorkflowV2Instance)
+    .map(instance => {
+      const invalidGlobalTransitions = getInvalidGlobalTransitions(instance)
+      const isInvalidGlobalTransitions = invalidGlobalTransitions.length > 0
```
shayc331, line 47: you can use _.isEmpty, it wont fail in a case of null, undefined or non array input

packages/jira-adapter/test/change_validators/workflowsV2/global_transition.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    changes = [toChange({ after: validWorkflowGlobalInstance })]
+    expect(await globalTransitionValidator(changes)).toEqual([])
+  })
+
+  it('should return an error when theres an addition with a duplicated global transition name', async () => {
+    changes = [toChange({ after: workflowGlobalInstanceAfter })]
+    expect(await globalTransitionValidator(changes)).toEqual([
+      {
+        elemID: workflowGlobalInstanceAfter.elemID,
+        severity: 'Error' as SeverityLevel,
```
shayc331, line 57: I think you can remove `as SeverityLevel`

packages/jira-adapter/src/config/config.ts
```diff
@@ -297,6 +297,7 @@ const CHANGE_VALIDATOR_NAMES = [
   'emptyProjectScopedContext',
   'filter',
   'kanbanBoardBacklog',
+  'outboundTransition',
```
shayc331, line 300: I think it should be `globalTransition`, I dont understand how the build works

shayc331 (commented):
looking great! 🥇 
had some comments

packages/jira-adapter/src/change_validators/workflowsV2/global_transition.ts
```diff
@@ -0,0 +1,51 @@
...
+  getChangeData,
+  isInstanceChange,
+  isAdditionOrModificationChange,
+} from '@salto-io/adapter-api'
+import { isWorkflowV2Instance, WorkflowV2Instance } from '../../filters/workflowV2/types'
+
+const { isDefined } = values
+
+const getInvalidGlobalTransitions = (instance: WorkflowV2Instance): string[] => {
+  const globalTransitionGroupByName = _.groupBy(
```
shayc331, line 22: nit: 
```suggestion
  const globalTransitionsGroupByName = _.groupBy(
```

shayc331 (approved):
💯

PR #6978, opened 2024-12-19
SALTO-6493: Change Validator that validates Flows have the referenced Elements

AlmogMesilaty:
SALTO-6493: Change Validator that validates Flows have the referenced Elements

---
Error message format:
image

---
_Release Notes_: 
_None_

---
_User Notifications_:
_None_


packages/salesforce-adapter/src/change_validators/flow_referenced_elements.ts
```diff
@@ -0,0 +1,122 @@
...
+  FieldMap,
+} from '@salto-io/adapter-api'
+import { TransformFuncSync, transformValuesSync } from '@salto-io/adapter-utils'
+import _ from 'lodash'
+import { isInstanceOfTypeSync } from '../filters/utils'
+import { FLOW_METADATA_TYPE } from '../constants'
+
+const isFlowNode = (fields: FieldMap): boolean => 'locationX' in fields && 'locationY' in fields
+
+const isFlowNodeName = (value: unknown, path: ElemID, parent: ObjectType): value is string =>
```
idoamit8, line 27: I am might missing something, but I don't like the return value here because it doesn't just check if the value is a string; it also includes additional logic. I think the function should return a boolean instead.

packages/salesforce-adapter/src/change_validators/flow_referenced_elements.ts
```diff
@@ -0,0 +1,122 @@
...
+  isFlowNode(parent.fields) && path.name === 'name' && _.isString(value)
+
+const isTargetReference = (value: unknown, path: ElemID): value is string =>
```
idoamit8, line 30: same here

packages/salesforce-adapter/src/change_validators/flow_referenced_elements.ts
```diff
@@ -0,0 +1,122 @@
...
+  path.name === 'targetReference' && _.isString(value)
+
+const getFlowNodesAndTargetReferences = (
+  element: InstanceElement,
+): { targetReferences: Map<string, ElemID>; flowNodes: Map<string, ElemID> } => {
+  const flowNodes = new Map<string, ElemID>()
+  const targetReferences = new Map<string, ElemID>()
+  const findFlowNodesAndTargetReferences: TransformFuncSync = ({ value, path, field }) => {
+    if (!field || !path) return value
```
idoamit8, line 39: When a variable is not a boolean, avoid using ! to check if it's undefined, as it will also return the same result for values like 0, an empty string, or other falsy values.

packages/salesforce-adapter/src/change_validators/flow_referenced_elements.ts
```diff
@@ -0,0 +1,122 @@
...
+    missingReferencedElementsErrors.forEach(([targetReference, elemId]) =>
+      errors.push(createMissingReferencedElementChangeError(targetReference, elemId)),
+    )
+  }
+  if (unreferencedElementsErrors.length > 0) {
+    unreferencedElementsErrors.forEach(([flowNode, elemId]) =>
+      errors.push(createUnreferencedElementChangeError(flowNode, elemId)),
+    )
+  }
+  return errors
```
idoamit8, line 109: I find this a bit clearer
```suggestion
  const targetEntries = Array.from(targetReferences)
  const flowEntries = Array.from(flowNodes)

  const missingReferenceErrors = targetEntries
    .filter(([reference]) => !flowNodes.has(reference))
    .map(([reference, elemId]) => createMissingReferencedElementChangeError(reference, elemId))

  const unreferencedErrors = flowEntries
    .filter(([reference]) => !targetReferences.has(reference))
    .map(([reference, elemId]) => createUnreferencedElementChangeError(reference, elemId))

  return [...missingReferenceErrors, ...unreferencedErrors]
```
(But if the lists can be very long sometimes `...` is not a good approach and its better to push)

idoamit8 (commented):
Great job!!  💯
I added some comments but will leave the rest of the review to @Uri-Bar  and @tamtamirr 

As we discussed, I noticed you're not using any get methods from the Map and are treating it purely as key-value pairs.
In this case, I would personally recommend using Record. While both have similar capabilities, Record is a simpler data structure and should be more than sufficient here.

packages/salesforce-adapter/test/change_validators/flow_referenced_elements.test.ts
```diff
@@ -0,0 +1,163 @@
...
+      recordCreates: { refType: new ListType(FlowNode), annotations: { required: false } },
+    },
+  })
+  describe('when all flow elements are existing and referenced', () => {
+    beforeEach(() => {
+      const flow = createInstanceElement(
+        {
+          fullName: 'TestFlow',
+          start: {
+            connector: { targetReference: 'ActionCall' },
```
tamtamirr, line 52: Can we const `targetReference` and use it both in the code and UTs?

packages/salesforce-adapter/test/change_validators/flow_referenced_elements.test.ts
```diff
@@ -0,0 +1,163 @@
...
+            },
+          ],
+        },
+        Flow,
+      )
+      flowChange = toChange({ after: flow })
+    })
+    it('should not return any errors', async () => {
+      const errors = await flowReferencedElements([flowChange])
+      expect(errors).toHaveLength(0)
```
tamtamirr, line 84:
```suggestion
      expect(errors).toBeEmpty()
```

packages/salesforce-adapter/test/change_validators/flow_referenced_elements.test.ts
```diff
@@ -0,0 +1,163 @@
...
+      )
+      flowChange = toChange({ after: flow })
+    })
+    it('should not return any errors', async () => {
+      const errors = await flowReferencedElements([flowChange])
+      expect(errors).toHaveLength(1)
+      const [error] = errors
+      expect(error.severity).toEqual('Error')
+      expect(error.message).toEqual('Reference to missing Flow Element')
+      expect(error.detailedMessage).toEqual(`The Flow Element "${'ActionCall'}" does not exist.`)
```
tamtamirr, line 106:
```suggestion
      expect(errors).toEqual([{
        message: '...',
        detailedMessge: '...',
        severity: 'Error',
        elemID: missingRefElemID,
      }])
```
Nit: Can be greatly simplified

packages/salesforce-adapter/test/change_validators/flow_referenced_elements.test.ts
```diff
@@ -0,0 +1,163 @@
...
+      )
+      flowChange = toChange({ after: flow })
+    })
+    it('should not return any errors', async () => {
+      const errors = await flowReferencedElements([flowChange])
+      expect(errors).toHaveLength(1)
+      const [error] = errors
+      expect(error.severity).toEqual('Info')
+      expect(error.message).toEqual('Unused Flow Element')
+      expect(error.detailedMessage).toEqual(`The Flow Element “${'ActionCall'}” isn’t being used in the Flow.`)
```
tamtamirr, line 130: Nit: Can be greatly simplified

packages/salesforce-adapter/test/change_validators/flow_referenced_elements.test.ts
```diff
@@ -0,0 +1,163 @@
...
+              name: 'Decision',
+              connector: { targetReference: 'RecordCreate' },
+            },
+          ],
+        },
+        Flow,
+      )
+      flowChange = toChange({ after: flow })
+    })
+    it('should create change error per error', async () => {
```
tamtamirr, line 158:
```suggestion
    it('should create change error per issue', async () => {
```
Nit:  I think it's confusing to use the term `error` on both sides

packages/salesforce-adapter/test/change_validators/flow_referenced_elements.test.ts
```diff
@@ -0,0 +1,163 @@
...
+      const errors = await flowReferencedElements([flowChange])
+      expect(errors).toHaveLength(3)
```
tamtamirr, line 160: Let's validate the errors explicitly :)

packages/salesforce-adapter/config_doc.md
```diff
@@ -355,3 +355,4 @@ For more details see the DeployOptions section in the [salesforce documentation
 | accountSettings              | true                   | Cannot set a value for enableAccountOwnerReport without proper org setting.                       |
 | unknownPicklistValues        | true                   | Disallow any usage of unknown pickList values.                                                    |
 | customApplications           | true                   | Check for multiple action overrides for the same action in customApplications instances.          |
+| flowReferencedElements       | true                   | Check that the referenced elements in Flow exist.                                                    |
```
tamtamirr, line 358: nit: Shouldn't the `|` be aligned with the rest? (I think prettier should solve that)

tamtamirr (commented):
Great work 👑  Added some comments

packages/salesforce-adapter/src/change_validators/flow_referenced_elements.ts
```diff
@@ -0,0 +1,98 @@
...
+} from '@salto-io/adapter-api'
+import { TransformFuncSync, transformValuesSync } from '@salto-io/adapter-utils'
+import _ from 'lodash'
+import { isInstanceOfTypeSync } from '../filters/utils'
+import { FLOW_METADATA_TYPE, FLOW_NODE, TARGET_REFERENCE } from '../constants'
+
+type ReferenceToElemId = {
+  referenceOrName: string
+  elemId: ElemID
+  kind?: string
```
idoamit8, line 27: Can kind ever be a string other than FLOW_NODE or TARGET_REFERENCE?
If not, it might be clearer to make this explicit.
```suggestion
  kind?: typeof FLOW_NODE | typeof TARGET_REFERENCE
```

idoamit8, line 27: Why do you allow `kind` to be undefined in the type? It seems it always defined in your code

packages/salesforce-adapter/src/change_validators/flow_referenced_elements.ts
```diff
@@ -0,0 +1,111 @@
...
+  return [...unusedElementErrors, ...missingReferenceErrors]
+}
+const changeValidator: ChangeValidator = async changes =>
+  changes
+    .filter(isAdditionOrModificationChange)
+    .filter(isInstanceChange)
+    .map(getChangeData)
+    .filter(isInstanceOfTypeSync(FLOW_METADATA_TYPE))
+    .map(getFlowNodesAndTargetReferences)
+    .map(getUnusedElementsAndMissingReferences)
```
tamtamirr, line 108: I think we can avoid this map and invoke `getUnusedElementsAndMissingReferences` from within `createChangeErrors`

packages/salesforce-adapter/src/change_validators/flow_referenced_elements.ts
```diff
@@ -0,0 +1,111 @@
...
+})
+
+const createUnusedElementChangeError = (elemId: ElemID, elemName: string): ChangeError => ({
+  elemID: elemId,
+  severity: 'Info',
+  message: 'Unused Flow Element',
+  detailedMessage: `The Flow Element "${elemName}" isn’t being used in the Flow.`,
+})
+
+const createChangeError = ({
```
tamtamirr, line 86:
```suggestion
const createChangeErrors = ({
```

packages/salesforce-adapter/src/constants.ts
```diff
@@ -76,6 +76,19 @@ export enum FIELD_TYPE_NAMES {
...
+// FlowStart field name
+export const TARGET_REFERENCE = 'targetReference'
+
+// FlowNode field names
+export enum FLOW_NODE_FIELD_NAMES {
+  NAME = 'name',
+  LOCATION_X = 'locationX',
+  LOCATION_Y = 'locationY',
+}
+
```
tamtamirr, line 91:
```suggestion
// Flow constants
export const FLOW_NODE = 'FlowNode'
export const TARGET_REFERENCE = 'targetReference'
export enum FlowNodeField {
  Name = 'name',
  LocationX = 'locationX',
  LocationY = 'locationY',
}
```
1. I think we should follow `PascalCase` for enums. It's more readable IMO.
2. These comments are confusing to me. I'd simply put them all under section of `Flow constants`

packages/salesforce-adapter/test/change_validators/flow_referenced_elements.test.ts
tamtamirr: We didn't use the constant 🙈

packages/salesforce-adapter/test/change_validators/flow_referenced_elements.test.ts
```diff
@@ -0,0 +1,246 @@
...
+            connector: { targetReference: 'ActionCall' },
+          },
+        },
+        Flow,
+      )
+      flowChange = toChange({ after: flow })
+    })
+    it('should not return any errors', async () => {
+      const errors = await flowReferencedElements([flowChange])
+      expect(errors).toHaveLength(1)
```
tamtamirr, line 114: nit: This check is redundant, the next expect covers that. Remove other occurrences as-well

packages/salesforce-adapter/test/change_validators/flow_referenced_elements.test.ts
```diff
@@ -0,0 +1,246 @@
...
+    beforeEach(() => {
+      flow = createInstanceElement(
+        {
+          fullName: 'TestFlow',
+          start: {
+            connector: { targetReference: 'ActionCall' },
+          },
+          assignments: [
+            {
+              name: 'Assignment',
```
tamtamirr, line 165: Use the `name` constant you've defined

packages/salesforce-adapter/test/change_validators/flow_referenced_elements.test.ts
```diff
@@ -0,0 +1,246 @@
...
+        message: 'Reference to missing Flow Element',
+        detailedMessage: `The Flow Element "${'ActionCall'}" does not exist.`,
+        elemID: flow.elemID.createNestedID('assignments', '0', 'connector', TARGET_REFERENCE),
+      })
+      expect(errors[2]).toEqual({
+        severity: 'Error',
+        message: 'Reference to missing Flow Element',
+        detailedMessage: `The Flow Element "${'ActionCall'}" does not exist.`,
+        elemID: flow.elemID.createNestedID('decisions', '0', 'connector', TARGET_REFERENCE),
+      })
```
tamtamirr, line 243: What happened to `error[0]` 😨 ?
Also, we can combine all these assertions to one `expect`

tamtamirr (commented):
Great work 🦚 ! Added some comments

packages/salesforce-adapter/src/constants.ts
AlmogMesilaty: I fixed the comments.
Regarding the enum, all the enums in the module use upper snake case, Isn't it best to maintain consistency?

tamtamirr: Some of them do some don't, you can keep it as UPPER_SNAKE_CASE.

tamtamirr (approved):
Great work! 🌵

PR #6981, opened 2024-12-19
SAAS-11932: Fix workspace functions return too early when updating inner state

ori-moisis:
When a workspace function needs to rebuild the inner state, it should not return before the state is updated

---

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/workspace/src/workspace/workspace.ts
+++ b/packages/workspace/src/workspace/workspace.ts
@@ -886,6 +886,11 @@ export const loadWorkspace = async (
     return workspaceState
   }
 
+  const updateWorkspaceState = async (...args: Parameters<typeof buildWorkspaceState>): Promise<void> => {
+    workspaceState = buildWorkspaceState(...args)
+    await workspaceState
+  }
+
   const getLoadedNaclFilesSource = async (): Promise<MultiEnvSource> => {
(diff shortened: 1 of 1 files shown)


packages/workspace/src/workspace/workspace.ts
```diff
@@ -968,7 +973,7 @@ export const loadWorkspace = async (
     const postChangeHash = await state(currentEnv()).getHash()
     const stateOnlyChanges = await getStateOnlyChanges(hiddenChanges)
     fixStateOnlyChangesFieldTypes(stateOnlyChanges, changes)
-    workspaceState = buildWorkspaceState({
+    await updateWorkspaceState({
```
dantal4, line 976: just to make sure that I understand, the `await` will happen only after the assignment of `workspaceState`,right? so this is equivalent to:
```typescript
const updatePromise = updateWorkspaceState(...)
await updatePromise
```

dantal4 (approved):
Thanks!

packages/workspace/src/workspace/workspace.ts
ori-moisis: yes

PR #6982, opened 2024-12-19
SALTO-7092: call workspace close method in case loadWorkspace fails

dorsev:
Suggestion: View this PR using
image

---

_Release Notes_: 
None

---

_User Notifications_: 
None

### DIFF
--- a/packages/workspace/src/workspace/workspace.ts
+++ b/packages/workspace/src/workspace/workspace.ts
@@ -500,1100 +500,1120 @@ export const loadWorkspace = async (
   mergedRecoveryMode: MergedRecoveryMode = 'rebuild',
   getCustomReferences: WorkspaceGetCustomReferencesFunc = async () => [],
 ): Promise<Workspace> => {
-  const workspaceConfig = await config.getWorkspaceConfig()
-  log.debug('Loading workspace with id: %s', workspaceConfig.uid)
-
-  if (_.isEmpty(workspaceConfig.envs)) {
-    throw new Error('Workspace with no environments is illegal')
-  }
-  const envs = (): ReadonlyArray<string> => workspaceConfig.envs.map(e => e.name)
-  let overrideEnv: string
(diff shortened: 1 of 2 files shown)


packages/workspace/src/workspace/workspace.ts
```diff
@@ -500,1095 +500,1116 @@ export const loadWorkspace = async (
...
-                        async staticFile =>
-                          (await naclFilesSource.getStaticFile({
-                            filePath: staticFile.filepath,
-                            encoding: staticFile.encoding,
-                            env: envName,
-                            isTemplate: staticFile.isTemplate,
-                            hash: staticFile.hash,
-                          })) ?? staticFile,
-                      ),
+  const close = async (): Promise<void> => {
```
yelly, line 504: This PR indents 1.1K lines, not great IMO.
A different approach can be renaming `loadWorkspace` to something else (`unsafeLoadWorksapce`, `loadWorkspaceInternal` etc.) and then implement `loadWorkspace` to just `try` calling the new function and closing in `except`.

dorsev: Your comment makes sense, but if I do as you suggest, I end up with the issue I sought to solve in the first place.

```
const loadWorkspace = {
try {
  const w = loadWorkspaceInternal()
} catch (e) {
// What do I call here? workspace is not 100% initialized.
}
```


So Im not sure how to address this comment at the moment. 

thoughts @ori-moisis @yelly ?

yelly: Yeah I guess the only real solution is breaking up this massive function but that's definitely out of scope.

yelly (approved):

packages/workspace/src/workspace/workspace.ts
ori-moisis: tried to think of an alternative as well and could not come up with a good one

ori-moisis (approved):

dorsev:
Thank you for the reviews @ori-moisis @yelly

PR #6983, opened 2024-12-19
SALTO-7104 Kanban board backlog

IdoZakk:
Fixed a bug in which we used to delete the backlog column of boards

---

This PR is split to 2 commits, the first one is relocation for some functions.

In kanban boards there are columns. They have a special built-in column called `Backlog` that is always the first column. Unfortunately, you can also add another column called the same way. 
The bug here resulted from a wrong bug fix in the past.
Fetch for Boards is the official API, but modify is private API.  As part of the private API there is a field called `isKanPlanColumn`, which is true for the built-in Backlog, and false for the rest (from what I saw).
If you deploy without `isKanPlanColumn` you will deploy that Backlog column as a user added `Backlog` column.
In the past we did that and created unwanted columns for users. The fix removed the backlog column altogether.
Other than to remove useful information this behavior caused deployment failures. The deployment fails only if during the put command the build-in `Backlog` column have some statuses, and that is why our tests did not fail.

* I have tested DC, the behavior and API there are the same (and that is a very surprising result. If I had to change the DC adapter for that I might have ran to Syria), and it works
* I have added a CV. It will also protect us from users who deploy from the old version (no backlog column in the start of the array)

---
_Release Notes_: 
Jira Adapter:
* Fixed a bug with some Board Deployments. Also, Board will now include also the Backlog built-in column

---
_User Notifications_: 
Jira Adapter:
* Board NaCls will now include also the built-in backlog column

### DIFF
--- a/packages/jira-adapter/src/change_validators/boards/kanban_board_backlog.ts
+++ b/packages/jira-adapter/src/change_validators/boards/kanban_board_backlog.ts
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2024 Salto Labs Ltd.
+ * Licensed under the Salto Terms of Use (the "License");
+ * You may not use this file except in compliance with the License.  You may obtain a copy of the License at https://www.salto.io/terms-of-use
+ *
+ * CERTAIN THIRD PARTY SOFTWARE MAY BE CONTAINED IN PORTIONS OF THE SOFTWARE. See NOTICE FILE AT https://github.com/salto-io/salto/blob/main/NOTICES
+ */
+import {
+  ChangeValidator,
+  getChangeData,
+  isAdditionOrModificationChange,
+  isInstanceChange,
+  SeverityLevel,
+} from '@salto-io/adapter-api'
+import Joi from 'joi'
+import { createSchemeGuard } from '@salto-io/adapter-utils'
+import { BOARD_TYPE_NAME, KANBAN_TYPE } from '../../constants'
+
+const BACKLOG = 'Backlog'
+type BoardColumn = {
+  name: string
+}
+
+type BoardColumnConfig = {
+  columns: BoardColumn[]
+}
+
+const COLUMN_CONFIG_SCHEME = Joi.object({
+  columns: Joi.array()
+    .items(
+      Joi.object({
+        name: Joi.string().required(),
+      }).unknown(true),
+    )
+    .min(1)
+    .required(),
+})
+  .required()
+  .unknown(true)
+
+const isBoardColumnConfig = createSchemeGuard<BoardColumnConfig>(COLUMN_CONFIG_SCHEME)
+
+export const kanbanBoardBacklogValidator: ChangeValidator = async changes =>
+  changes
+    .filter(isInstanceChange)
+    .filter(isAdditionOrModificationChange)
+    .map(getChangeData)
+    .filter(instance => instance.elemID.typeName === BOARD_TYPE_NAME)
+    .filter(instance => instance.value.type === KANBAN_TYPE)
+    .filter(
+      instance =>
+        !isBoardColumnConfig(instance.value.columnConfig) || instance.value.columnConfig.columns[0].name !== BACKLOG,
+    )
+    .map(instance => ({
+      elemID: instance.elemID,
+      severity: 'Error' as SeverityLevel,
+      message: 'Unable to deploy a Kanban Board When the first column is not named "Backlog"',
+      detailedMessage:
+        // Change this message in 1 month, the fetch part is relevant only for deployments before the fix
+        'A Kanban board must have a first column named Backlog. If you did not edit the board manually please fetch both envs and try again',
+    }))
--- a/packages/jira-adapter/src/change_validators/boards/kanban_board_backlog.ts
+++ b/packages/jira-adapter/src/change_validators/boards/kanban_board_backlog.ts
@@ -17,8 +17,11 @@ import { createSchemeGuard } from '@salto-io/adapter-utils'
 import { BOARD_TYPE_NAME, KANBAN_TYPE } from '../../constants'
 
 const BACKLOG = 'Backlog'
-type BoardColumn = {
+export type BoardColumn = {
   name: string
+  min?: number
+  max?: number
+  statuses?: string[]
 }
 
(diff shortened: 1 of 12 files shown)


packages/jira-adapter/src/change_validators/boards/kanban_board_backlog.ts
```diff
@@ -0,0 +1,61 @@
...
+    .filter(
+      instance =>
+        !isBoardColumnConfig(instance.value.columnConfig) || instance.value.columnConfig.columns[0].name !== BACKLOG,
+    )
+    .map(instance => ({
+      elemID: instance.elemID,
+      severity: 'Error' as SeverityLevel,
+      message: 'Unable to deploy a Kanban Board When the first column is not named "Backlog"',
+      detailedMessage:
+        // Change this message in 1 month, the fetch part is relevant only for deployments before the fix
```
shayc331, line 59: please add a ticket

packages/jira-adapter/src/change_validators/boards/kanban_board_backlog.ts
```diff
@@ -0,0 +1,61 @@
...
+
+const COLUMN_CONFIG_SCHEME = Joi.object({
+  columns: Joi.array()
+    .items(
+      Joi.object({
+        name: Joi.string().required(),
+      }).unknown(true),
+    )
+    .min(1)
+    .required(),
```
shayc331, line 36: I think it is better to add `.unkown(true)` here as well

packages/jira-adapter/src/change_validators/boards/kanban_board_backlog.ts
```diff
@@ -0,0 +1,61 @@
...
+})
+  .required()
+  .unknown(true)
+
+const isBoardColumnConfig = createSchemeGuard<BoardColumnConfig>(COLUMN_CONFIG_SCHEME)
```
shayc331, line 41: if you add a parameter of `errorMessage` it will log an error when this schemeGuard fails, I think it may be useful.

packages/jira-adapter/src/filters/board/board_columns.ts
```diff
@@ -105,7 +105,7 @@ export const deployColumns = async (
   }
```
shayc331, line 105: please look at line 103, is it possible to not have columnConfig at all?
I think it contradicts the CV.

packages/jira-adapter/src/filters/board/board_columns.ts
```diff
@@ -76,11 +75,12 @@ const getColumnsName = async (id: string, client: JiraClient): Promise<string[]
   return response.data[COLUMNS_CONFIG_FIELD].columns.map(({ name }) => name)
 }
 
-const convertColumn = (column: Values): Values => ({
+const convertColumn = (column: Values, isKanPlanColumn: boolean): Values => ({
```
shayc331, line 78: we can use `BoardColumn` instead of Values

packages/jira-adapter/src/change_validators/boards/kanban_board_backlog.ts
shayc331: why not add `mappedStatuses`, `min` and `max`?

shayc331 (commented):
looking good!
had some comments

packages/jira-adapter/test/change_validators/boards/kanban_board_backlog.test.ts
```diff
@@ -0,0 +1,154 @@
...
+          after: noErrorInstance,
+        }),
+      ]),
+    ).toEqual([
+      {
+        elemID: instance.elemID,
+        severity: 'Error',
+        message: ERROR_TITLE,
+        detailedMessage: ERROR_MESSAGE,
+      },
```
shayc331, line 151: you wrote in the `it` clause that `it should return several errors if they exist` but the result is only one error.

packages/jira-adapter/test/filters/board/board_columns.test.ts
```diff
@@ -112,57 +95,8 @@ describe('boardColumnsFilter', () => {
...
-              },
-              {
-                name: 'someColumn',
-              },
-            ],
-          },
-        },
-      })
-
+    it('should not remove first column if scrum', async () => {
```
shayc331, line 98: I think `should not remove any column` or something similar is better since Scrum and Kanban are all the options for board types (right?)

packages/jira-adapter/src/change_validators/boards/kanban_board_backlog.ts
IdoZakk: The `items` has unknown, not sure I follow the logic in having unknown to an array
As this is a private PR I am not sure what is mandatory and what is not, so I only added what is needed for this CV to work

packages/jira-adapter/src/change_validators/boards/kanban_board_backlog.ts
IdoZakk: I think the error will be redundant, we will issue an error for the CV

packages/jira-adapter/src/change_validators/boards/kanban_board_backlog.ts
shayc331: about my first comment: I was wrong, I meant to add `.unkown(true)` for the object as you did.
about the second comment: I agree

packages/jira-adapter/src/change_validators/boards/kanban_board_backlog.ts
shayc331: There isn't an indication in Datadog when the user gets a CV error. IMO it has a value

IdoZakk: I think that since the CV is an error CV it will only be noise- we tell the user what to do, and if the user will have a problem and will issue we can see it in retool

packages/jira-adapter/src/filters/board/board_columns.ts
IdoZakk: Good catch

shayc331 (approved):
💯

PR #6992, opened 2024-12-23
SALTO-7119 use async iterable in mergeManager cache update

dantal4:
Includes a small readability refactor in `elements_cache.ts`.
In the code we have `Promise.all` that gets elements from each of the sources and returns `Element[]`, but it should return `AsyncIterable<Element>`, so we don’t need to have all elements in the memory in the same time.

---

---
_Release Notes_: 
Workspace:
- Reduce memory consumption in mergeManager cache update

---
_User Notifications_: 
None

### DIFF
--- a/packages/workspace/src/workspace/nacl_files/elements_cache.ts
+++ b/packages/workspace/src/workspace/nacl_files/elements_cache.ts
@@ -78,29 +78,26 @@ export type CacheChangeSetUpdate = {
   currentErrors: RemoteMap<SaltoError[]>
 }
 
-const getElementsToMergeFromChanges = async (
+const getElementsToMergeFromChanges = (
   srcChanges: ChangeSet<Change<Element>>,
   idsExistingOnSecondSourceChanges: Set<string>,
   otherSrcChanges: ReadOnlyElementsSource,
-): Promise<Element[]> => {
-  const elementsToMerge: (Element | undefined)[] = []
-  await Promise.all(
-    srcChanges.changes.map(async change => {
-      const element = getChangeData(change)
-      const id = element.elemID
-      if (isAdditionOrModificationChange(change)) {
-        elementsToMerge.push(element)
+): AsyncIterable<Element> =>
+  awu(srcChanges.changes).flatMap(async change => {
+    const elementsToMerge: Element[] = []
+    const element = getChangeData(change)
+    const id = element.elemID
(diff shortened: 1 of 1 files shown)


packages/workspace/src/workspace/nacl_files/elements_cache.ts
```diff
@@ -78,28 +72,53 @@ export type CacheChangeSetUpdate = {
...
+      const unmodifiedFragment = await otherSrcChanges.get(id)
+      if (unmodifiedFragment !== undefined) {
+        elementsToMerge.push(unmodifiedFragment)
       }
-    }),
-  )
-  return elementsToMerge.filter(values.isDefined)
+    }
+    return elementsToMerge
+  })
```
dantal4, line 104: the main change is here - use `awu` instead of an `Promise.all`

ori-moisis: but this is iterating over something that is already in-memory (`srcChanges.changes`) and anything it gets from the `otherSrcChanges` is:
1. also already in memory IIUC
2. gets pushed into an array anyway, so even if it wasn't in memory before, it will be at the end of this iteration

ori-moisis:
any chance you could undo the refactor?
it is really hard to figure out what happened here...

packages/workspace/src/workspace/nacl_files/elements_cache.ts
ori-moisis: never mind, I think I understand why this changes something
but I am not sure with the whole refactor where this fits into the logic

dantal4: `otherSrcChanges` is `src1` / `src2` - that comes from `sources` - https://github.com/salto-io/salto/blob/a79a6eb1f36b531ccb4a0c409424a36fd2e1f2f4/packages/workspace/src/workspace/workspace.ts#L633-L640

dantal4: re 2: the list contains up to 2 elements - `elementsToMerge` is initialized in each async iteration.

dantal4:
> any chance you could undo the refactor? it is really hard to figure out what happened here...

I moved the refactor to https://github.com/salto-io/salto/pull/6993

ori-moisis (approved):

PR #6996, opened 2024-12-24
SALTO-7134: Calculate Salesforce Data Instances service URL before creating collision warnings

tamtamirr:
Calculate Salesforce Data Instances service URL before creating collision warnings

---

Simply reordered the filters.

---
_Release Notes_: 
_None_

---
_User Notifications_: 
_None_


shayc331 (approved):
thanks! 🥇

PR #6997, opened 2024-12-25
Remove warnings from SDF error message

dantal4:
Return a cleaner SDF error message - without warnings that aren't part of the error.

---

---
_Release Notes_: 
Netsuite Adapter:
- Remove warnings from SDF error message

---
_User Notifications_: 
None


omrilit (approved):
👑

PR #7001, opened 2024-12-25
SALTO-7139 SALTO-6602: Elements Fixer should receive Elements with resolved types

tamtamirr:
Elements Fixer should receive Elements with resolved types.

---

Simply invoking `resolveTypeShallow` on the Workspace Elements.

---
_Release Notes_: 
_Core_:
- Element fixers will now receive Elements with resolved types.

---
_User Notifications_: 
_None_

### DIFF
--- a/packages/core/test/api.test.ts
+++ b/packages/core/test/api.test.ts
@@ -36,6 +36,7 @@ import {
   ReferenceExpression,
   SeverityLevel,
   toChange,
+  TypeReference,
 } from '@salto-io/adapter-api'
 import * as workspace from '@salto-io/workspace'
 import { collections } from '@salto-io/lowerdash'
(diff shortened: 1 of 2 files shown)


packages/core/test/api.test.ts
```diff
@@ -1311,7 +1312,7 @@ describe('api.ts', () => {
       type = new ObjectType({
         elemID: new ElemID('test1', 'test'),
       })
-      instance = new InstanceElement('test', type, { c: 4 })
+      instance = new InstanceElement('test', new TypeReference(type.elemID), { c: 4 })
```
tamtamirr, line 1315: The type shouldn't have been resolved here in first place (As mock workspace will always return an instance with resolved type)

yelly (approved):

PR #7002, opened 2024-12-26
SALTO-5932: change split options in Fields to default true

IdoZakk:
None

---

[Noise reduction](https://github.com/salto-io/salto_private/pull/10145)

---
_Release Notes_: 
Jira Adapter:
* Options in Fields will now be a separate element. It allows modifying options without changing their IDs

---
_User Notifications_: 
Jira Adapter:
* Field contexts will now have a list of references to their options, options will now be in standalone elements

### DIFF
--- a/packages/jira-adapter/e2e_test/instances/index.ts
+++ b/packages/jira-adapter/e2e_test/instances/index.ts
@@ -18,7 +18,7 @@ import { naclCase } from '@salto-io/adapter-utils'
 import { CUSTOM_FIELDS_SUFFIX } from '../../src/filters/fields/field_name_filter'
 import { JIRA, WEBHOOK_TYPE, STATUS_TYPE_NAME } from '../../src/constants'
 import { createReference, findType } from '../utils'
-import { createContextValues, createFieldValues } from './field'
+import { createContextValues, createFieldValues, createOptionsAndOrders } from './field'
 import { createFieldConfigurationSchemeValues } from './fieldConfigurationScheme'
 import { createIssueTypeScreenSchemeValues } from './issueTypeScreenScheme'
 import { createScreenValues } from './screen'
@@ -45,6 +45,12 @@ export const createInstances = (fetchedElements: Element[], isDataCenter: boolea
     undefined,
     { [CORE_ANNOTATIONS.PARENT]: [new ReferenceExpression(field.elemID, field)] },
   )
+
+  const fieldContextOptionsAndOrders = createOptionsAndOrders({
+    optionsType: findType('CustomFieldContextOption', fetchedElements),
+    orderType: findType('FieldContextOptionsOrder', fetchedElements),
+    contextInstance: fieldContext,
+  })
 
   const screen = new InstanceElement(
     randomString,
@@ -108,7 +114,7 @@ export const createInstances = (fetchedElements: Element[], isDataCenter: boolea
       ? createDataCenterInstances(randomString, fetchedElements)
       : createCloudInstances(randomString, uuid, fetchedElements)),
     [field],
-    [fieldContext],
+    fieldContextOptionsAndOrders,
     [screen],
     [screenScheme],
     [issueTypeScreenScheme],
--- a/packages/jira-adapter/e2e_test/instances/index.ts
+++ b/packages/jira-adapter/e2e_test/instances/index.ts
@@ -14,7 +14,8 @@ import {
   ReferenceExpression,
   ModificationChange,
 } from '@salto-io/adapter-api'
-import { naclCase } from '@salto-io/adapter-utils'
+import { inspectValue, naclCase } from '@salto-io/adapter-utils'
+import { logger } from '@salto-io/logging'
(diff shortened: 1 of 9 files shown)


packages/jira-adapter/e2e_test/instances/index.ts
```diff
@@ -46,6 +46,12 @@ export const createInstances = (fetchedElements: Element[], isDataCenter: boolea
     { [CORE_ANNOTATIONS.PARENT]: [new ReferenceExpression(field.elemID, field)] },
   )
 
+  const fieldContextOptionsAndOrders = createOptionsAndOrders({
+    optionsType: findType('CustomFieldContextOption', fetchedElements),
+    orderType: findType('FieldContextOptionsOrder', fetchedElements),
```
shayc331, line 51: Can the `optionsType` and the `orderType` be different than those types? 
we can simplify it by creating them inside the function

packages/jira-adapter/test/change_validators/field_contexts/custom_field_with_10K_options.test.ts
```diff
@@ -56,6 +56,7 @@ describe('customFieldsWith10KOptionValidator', () => {
       },
     )
     config = _.cloneDeep(getDefaultConfig({ isDataCenter: false }))
+    config.fetch.splitFieldContextOptions = false
```
shayc331, line 59: I think this line should be inside the "without splitFieldContextOptions" suite

packages/jira-adapter/test/filters/fields/default_values.test.ts
```diff
@@ -39,6 +39,7 @@ describe('default values', () => {
         patch: mockFunction<clientUtils.HTTPWriteClientInterface['patch']>(),
       }
       config = _.cloneDeep(getDefaultConfig({ isDataCenter: false }))
+      config.fetch.splitFieldContextOptions = false
```
shayc331, line 42: `updateDefaultValues` should work with ([code](https://github.com/salto-io/salto/blob/main/packages/jira-adapter/src/filters/fields/context_default_value_deployment_filter.ts#L26)) and without ([code](https://github.com/salto-io/salto/blob/main/packages/jira-adapter/src/filters/fields/contexts.ts#L96)) the flag. I think we should test it in both ways

shayc331 (commented):
looking great! 
had some small comments 🎖️

packages/jira-adapter/e2e_test/instances/index.ts
IdoZakk: The types are created in the fetch from the swagger, this is the deploy phase.
I do not think it is wise to create empty types here

packages/jira-adapter/test/filters/fields/default_values.test.ts
IdoZakk: It is tested, in a a different file `field_context_option_split.test`
Once the flag will be on for 3 weeks we will delete it completely, and we will clean the code which is made very complex in order to support both versions

packages/jira-adapter/e2e_test/instances/index.ts
```diff
@@ -119,6 +129,12 @@ export const createInstances = (fetchedElements: Element[], isDataCenter: boolea
     // [group],
     [status],
   ]
+  instances.forEach(instGroup =>
+    instGroup.forEach(instance =>
+      log.debug(`E2E created instance ${instance.elemID.getFullName()} with values ${inspectValue(instance.value)}`),
+    ),
+  )
```
IdoZakk, line 136: I am keeping this as I think it can really help debugging the E2E

packages/jira-adapter/e2e_test/instances/index.ts
shayc331: yes I see, I wonder why not passing `fetchedElements` to `createOptionsAndOrders`.
but it is completely a nit, we can move on

packages/jira-adapter/src/filters/fields/context_options_deployment_filter.ts
```diff
@@ -83,17 +84,27 @@ const filter: FilterCreator = ({ config, client, paginator, elementsSource }) =>
...
-      )
-      appliedChanges = []
+      if (
+        addChanges.length === 0 &&
+        modifyChanges.length === 0 &&
+        err instanceof clientUtils.HTTPError &&
+        err.response.status === 404
+      ) {
+        // We delete the context before the options, so if the context is already deleted, we can ignore the error
+        log.debug('All field context options were already deleted')
```
shayc331, line 94: are you sure that `addChanges` and `modifyChanges` should be empty in this case?

shayc331 (commented):
looking good, had one question

packages/jira-adapter/src/filters/fields/context_options_deployment_filter.ts
IdoZakk: If the context was removed the options cannot stay, the context is the parent

shayc331 (approved):
🤞 🤞

PR #7003, opened 2024-12-26
SALTO-4116: Support fetch and deploy for provisioning in apps

idoamit8:
In this PR I support fetch and deploy for provisioning in apps

---

Some public APIs support User Provisioning (referred to as "ToApp" in the UI) and Inbound Provisioning (referred to as "To Okta" in the UI). These endpoints currently support a limited number of applications, as listed below:
match: ['^google$', '^office365$', '^okta_org2org$', '^slack$', '^zoomus$', '^zscalerbyz$']
Additionally, all applications have a private API that supports the To Okta page.
I plan to use the public API for the supported applications to enable both fetching and deploying their provisioning. For the remaining applications, I will support fetching and deploying only for what is represented by To Okta.

_After first review I will add:_
**1. CV to say that we are not support enablement of provisioning** (Although we will be able to support it if we turn off the CV (but its a matter of authentication))

Some screenshots:
||applicationUserProvisioning|applicationInboundProvisioning|applicationProvisioningGeneral|applicationProvisioningUsers|
|---|---|---|---|---|
||image|image|image|image|

 
---
_Release Notes_: 

---
_User Notifications_: 


packages/okta-adapter/src/definitions/deploy/deploy.ts
```diff
@@ -483,14 +490,156 @@ const createCustomizations = (): Record<string, InstanceDeployApiDefinitions> =>
...
+                          'applicationProvisioningGeneral',
+                        ]),
                       },
                     }
                   },
                 },
               },
             },
             ...createDeployAppPolicyRequests(),
+            {
```
Nurdok, line 510: Discussed f2f, let's try migrating one of these to use `RecurseIntoPath` and see if it works.

packages/okta-adapter/test/mock_replies/application_activate_provisioning_supported_app.json
```diff
@@ -0,0 +1,146 @@
...
+          "password": {
+            "status": "DISABLED",
+            "seed": "RANDOM",
+            "change": "KEEP_EXISTING"
+          }
+        }
+      },
+      "_links": {
+        "self": {
+          "href": "https://salto.okta.com/api/v1/apps/app-fakeid1/features/USER_PROVISIONING",
```
shir-reifenberg, line 46: please clean our base url from here 🙏  (I think you can remove the entire `_links` object

packages/okta-adapter/src/definitions/fetch/fetch.ts
```diff
@@ -371,6 +371,44 @@ const createCustomizations = ({
             },
           },
         },
+        applicationUserProvisioning: {
+          typeName: 'ApplicationUserProvisioning',
+          conditions: [
+            {
+              fromField: 'name',
+              match: ['^google$', '^office365$', '^okta_org2org$', '^slack$', '^zoomus$', '^zscalerbyz$'],
```
shir-reifenberg, line 379: 2 comments here -
1. can you please add (here or in a better place) a link to the documentation stating only those apps are supported ?
2. Any chance we can use this list as const in all places? and then use `.map` to turn this into regex according to the pattern ?

also it would be useful to have examples (like what we're trying to detect with this pattern

packages/okta-adapter/test/mock_replies/application_activate_provisioning_supported_app.json
idoamit8: Cleaned it. (But we have it in a lot of this mock_replies)

packages/okta-adapter/src/definitions/fetch/fetch.ts
```diff
@@ -371,6 +372,44 @@ const createCustomizations = ({
             },
           },
         },
+        applicationUserProvisioning: {
+          typeName: 'ApplicationUserProvisioning',
+          conditions: [
+            {
+              fromField: 'name',
+              match: INBOUND_PROVISIONING_SUPPORTED_APP_NAMES.map(name => `^${name}$`).concat(['^zscalerbyz$']),
```
Nurdok, line 380: Instead of using `INBOUND_PROVISIONING_SUPPORTED_APP_NAMES` and concatenating `zscalerbyz`, I'd create a new constant

```ts
USER_PROVISIONING_SUPPORTED_APP_NAMES = [...INBOUND_PROVISIONING_SUPPORTED_APP_NAMES, 'zscalerbyz']
```
and use it here.

Nurdok (approved):

PR #7006, opened 2024-12-28
SALTO-7148: Change toActionNames to be async, so it can work with element source

shir-reifenberg:
The function was sync which made element source useless

---

---
_Release Notes_: 
None

---
_User Notifications_: 
None

### DIFF
--- a/packages/adapter-components/src/deployment/deploy/graph.ts
+++ b/packages/adapter-components/src/deployment/deploy/graph.ts
@@ -47,30 +47,33 @@ const isStandardDependentAction = <AdditionalAction extends string = never>(
 ): boolean =>
   isStandardAction(sourceAction) && isStandardAction(targetAction) && isDependentAction(sourceAction, targetAction)
 
-const groupChangesByTypeAndAction = <ClientOptions extends string, AdditionalAction extends string = never>(
+const groupChangesByTypeAndAction = async <ClientOptions extends string, AdditionalAction extends string = never>(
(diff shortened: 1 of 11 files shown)


packages/adapter-components/src/deployment/deploy/graph.ts
```diff
@@ -47,38 +47,41 @@ const isStandardDependentAction = <AdditionalAction extends string = never>(
...
       }
-    })
-  })
+      const typeChanges = changesByTypeAndAction[typeName]
+      actions.forEach(a => {
+        if (typeChanges[a] !== undefined) {
+          typeChanges[a]?.push(c)
+        } else {
+          typeChanges[a] = [c]
+        }
```
Nurdok, line 73: Feel free to ignore, but this can be shortened to:

```ts
const typeChanges = changesByTypeAndAction[typeName] ??= {}
actions.forEach(a => (typeChanges[a] ??= []).push(c))
```

The _nullish coalescing assignment_ assigns the righthand side to the lefthand side only if the lefthand side is nullish (`undefined` or `null`) and it returns the value of the lefthand side (after the assignment if the assignment was needed).

Nurdok (approved):
Thanks for this!

packages/adapter-components/src/deployment/deploy/graph.ts
shir-reifenberg: Thanks!! I prefer leaving as is for better readability 🙏

